PR #2, opened 2025-05-15
Release Lemonade v7

jeremyfowers:
## Breaking Changes

The Lemonade SDK project has graduated from ONNX/TurnkeyML!
- Official link for the new repo: https://github.com/lemonade-sdk/lemonade
- New PyPI package: `lemonade-sdk` at https://pypi.org/project/lemonade-sdk/
- The `turnkey` CLI and related code has been removed from this repo.
- The `turnkeyml` module `src/turnkeyml` has been removed from the package. Any submodules needed for Lemonade have been moved into `src/lemonade`.
  - To update your code, change the imports like this `from turnkeyml.state import State` to `from lemonade.state import State`

## New Features
- Enable `--log-level` argument on `lemonade-server` CLI (@danielholanda, @kovtcharov)

## Documentation
- New Lemonade Server guides for getting started, the `lemonade-server` CLI, and local LLM server FAQs (@jeremyfowers)
- Add information about LLM management on disk (location, deletion procedure) to server_models.md (@jeremyfowers)

## Bugs Fixed
- `lemonade-install --ryzenai` no longer rejects Strix Halo devices (@danielholanda)

## Additional Changes
- Refactor profilers into a new base class (@amd-pworfolk)

## Thank You!!!

A huge thank you to everyone who has contributed to the Lemonade SDK project so far! We are looking forward to a great new chapter in this dedicated repository.

@danielholanda @ramkrishna2910 @vgodsoe @amd-pworfolk @apsonawane @jiafatom @pcolange @kovtcharov @iswaryaalex @gabeweisz @itomek

Issue #23, opened 2025-06-10
Knowledge Base: How to apply black formatting


jeremyfowers:
We require that all python files in this repo adhere to black formatting. This is enforced with a black check in some CI workflows.

Here's an example of code that is not properly formatted (left) vs. formatted (right):

![Image](https://github.com/user-attachments/assets/a43be4ff-0ec4-4d2e-8aca-b269723d483f)

The easiest way to get past this is to enable black formatter in your vscode to format-on-save.

To enable black in your own editor:

1. Install the extension

![Image](https://github.com/user-attachments/assets/6b3a2054-5169-45a8-b1fd-08d5e6a299ad)

2. Set black as the default formatter

![Image](https://github.com/user-attachments/assets/f10f3738-dc40-4ce6-961a-a962c148adef)

3. Make sure format-on-save is enabled

![Image](https://github.com/user-attachments/assets/a4fa1a5b-f5d8-4da7-9658-29608ff79a56)

PR #233, opened 2025-08-14
Fix model deletion bugs

tylerstraub:
## Summary
Fixes two critical bugs in model deletion functionality that caused data loss and CLI crashes.

### 🔧 Fix #1: Handle manual cache deletion gracefully
- **Issue**: CLI crashed with `LocalEntryNotFoundError` when cache files were manually removed
- **Root Cause**: `delete_model()` relied on `custom_snapshot_download()` to find cache paths, which failed when files were manually deleted
- **Solution**: Added fallback cache path construction using standard HuggingFace naming conventions and graceful error handling
- **Impact**: Eliminates CLI crashes, provides clear user feedback, maintains registry consistency

### 🔧 Fix #2: Prevent GGUF variant cross-deletion data loss
- **Issue**: Deleting one GGUF variant would accidentally delete ALL variants from the same repository
- **Example**: Deleting `gpt-oss-120b-GGUF:Q4_K_M` would also delete `gpt-oss-120b-GGUF:F16`, `gpt-oss-120b-GGUF:Q8_0`, etc.
- **Root Cause**: `delete_model()` used `parse_checkpoint()[0]` (base only) and deleted entire repository directory
- **Solution**: Selective variant deletion using existing `identify_gguf_models()` utility with smart repository cleanup
- **Impact**: Prevents data loss, enables proper multi-variant GGUF model management

## Files Changed
- `src/lemonade_server/model_manager.py` - Enhanced `delete_model()` method
- `test/model_manager_test.py` - Comprehensive test suite (new file)

## Verification
```bash
# Run the new test suite
python test/model_manager_test.py
# Result: 11/11 tests passed

# Verify no regressions  
python test/server_unit.py
# Result: All existing tests still pass
```

### DIFF
--- a/src/lemonade_server/model_manager.py
+++ b/src/lemonade_server/model_manager.py
@@ -232,32 +232,53 @@ class ModelManager:
         # Handle GGUF models that have the format "checkpoint:variant"
         base_checkpoint = parse_checkpoint(checkpoint)[0]
 
+        model_path = None
+        
         try:
-            # Get the local path using snapshot_download with local_files_only=True
+            # First, try to get the local path using snapshot_download with local_files_only=True
             snapshot_path = custom_snapshot_download(
                 base_checkpoint, local_files_only=True
             )
-
             # Navigate up to the model directory (parent of snapshots directory)
             model_path = os.path.dirname(os.path.dirname(snapshot_path))
 
-            # Delete the entire model directory (including all snapshots)
-            if os.path.exists(model_path):
-                shutil.rmtree(model_path)
-                print(f"Successfully deleted model {model_name} from {model_path}")
-            else:
-                raise ValueError(
-                    f"Model {model_name} not found locally at {model_path}"
-                )
-
         except Exception as e:
+            # If snapshot_download fails, try to construct the cache path manually
             if (
                 "not found in cache" in str(e).lower()
-                or "no such file" in str(e).lower()
+                or "localentrynotfounderror" in str(e).lower()
+                or "cannot find an appropriate cached snapshot" in str(e).lower()
             ):
-                raise ValueError(f"Model {model_name} is not installed locally")
+                # Construct the Hugging Face cache path manually
+                cache_home = huggingface_hub.constants.HF_HUB_CACHE
+                # Convert repo format (e.g., "unsloth/GLM-4.5-Air-GGUF") to cache format
+                repo_cache_name = base_checkpoint.replace("/", "--")
+                model_path = os.path.join(cache_home, f"models--{repo_cache_name}")
             else:
                 raise ValueError(f"Failed to delete model {model_name}: {str(e)}")
+
+        # Try to delete the model directory
+        if model_path and os.path.exists(model_path):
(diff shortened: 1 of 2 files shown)


jeremyfowers:
Thanks very much for working on this @tylerstraub! FYI there's a guide here for how to apply `black` and get past that CI check: https://github.com/lemonade-sdk/lemonade/issues/23

test/model_manager_test.py
```diff
@@ -0,0 +1,412 @@
+"""
```
danielholanda: Really like this test!
Can you please name `test/model_manager.py` instead of `test/model_manager_test.py` and add it to `test_server.yml`? This way it will always run when our CI tests run.

src/lemonade_server/model_manager.py
```diff
@@ -229,36 +230,115 @@ def delete_model(self, model_name: str):
...
+                        snapshot_dirs = [d for d in os.listdir(snapshots_dir)
+                                       if os.path.isdir(os.path.join(snapshots_dir, d))]
+                        if snapshot_dirs:
+                            # Use the first (likely only) snapshot directory
+                            snapshot_path = os.path.join(snapshots_dir, snapshot_dirs[0])
             else:
                 raise ValueError(f"Failed to delete model {model_name}: {str(e)}")
 
+        # Handle deletion based on whether this is a GGUF model with variants
+        if variant and snapshot_path and os.path.exists(snapshot_path):
```
danielholanda, line 287: Tested locally and it works great. That said, one issue still remains:
If we have two variants of the same model listed in `server_models.json` and one of those is downloaded, both will appear as "available" on the model manager. My recommendation is to merge this PR as is and target this second bug in a second PR.
https://github.com/lemonade-sdk/lemonade/issues/234

tylerstraub: Good catch! Perhaps we can add the assertion to the new tests as well.

test/model_manager_test.py
tylerstraub: Ah that makes sense. I can make these changes but it might be delayed a few days there are some other urgencies I need to tend to first.

danielholanda: Changes are too awesome to wait. Pushed this minor change to ensure we can merge it :)

src/lemonade_server/model_manager.py
danielholanda: Looks like we already have all assertions needed. Feel free to add more later if needed

danielholanda (approved):
Too awesome to wait.
Thanks for the contribution!

.github/workflows/test_server.yml
```diff
@@ -47,6 +47,10 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
         run: |
           python test/server_unit.py
+      - name: Run Model Manager Tests
+        shell: bash -el {0}
+        run: |
+          python test/model_manager_unit.py
```
danielholanda:
```suggestion
          python test/model_manager.py
```

Issue #229, opened 2025-08-14
Allow models to run during model download


jeremyfowers:
## Summarized Issue

Problems:
- Model loading is a blocking call.
- While streaming, health checks still show the blocking state for model load.

Suggested solution:
- Implement model load status API to prevent duplicate loading.

## Original Issue

"Ah, lemonade is synchronous so even when downloading a model I can't do anything else 😛" - Primal on discord

We should let completions and other endpoints go through during downloads.

ramkrishna2910:
+1 heard this request from another customer

jeremyfowers:
This is supported now!

PR #241, opened 2025-08-17
WIP: add OCI container

FuchtelJockel:
Can't test/add ROCm because I have no compatible GPU.


container/build-oci-image
```diff
@@ -0,0 +1,63 @@
...
+
+ctr=$(buildah from docker.io/library/ubuntu:rolling)
+buildah add "$ctr" entrypoint.sh /
+buildah config --cmd '["lemonade-server-dev"]' "$ctr"
+buildah config --entrypoint '["/entrypoint.sh"]' "$ctr"
+buildah config --port 8000 "$ctr"
+buildah run "$ctr" /bin/sh -c 'ln -sf /usr/share/zoneinfo/Europe/London /etc/localtime; \
+     apt update; \
+     apt install -y \
+     git \
```
mcowger: This is missing curl and libcurl4-openssl-dev to be able to start the vulkan backend.

FuchtelJockel: Works just fine without it on my end. It uses libcurl, it gets pulled in through python package dependencies. But I understand why a LLM would be likely to recommend it.

FuchtelJockel:
@mcowger please append files instead of pasting long texts (logs).
I converted the script into a Containerfile that produces a working container on my machine.

MatthK:
I just tried to build the docker image and fire up the container. While the build worked, and I can access the Webgui, I can't load a model. It keeps saying `Server Offline` and shows the following error message:

```
Error loading models: NetworkError when attempting to fetch resource.
Check the Lemonade Server logs via the system tray app for more information.
```

However, the docker log doesn't show anything, and I'm not sure what other log I could check.
And the next question then would be, how to take advantage of the Strix Halo GPU of my AMD RYZEN AI MAX+ 395 w/Radeon 8060S x 32

FuchtelJockel:
@MatthK You are probably running into this https://github.com/lemonade-sdk/lemonade/issues/229

The web UI cannot connect to the server while a model is being fetched.

Your GPU supports Vulkan and the instructions are in the README included in this commit. The NPU is only supported on Windows.

MatthK:
I'm not sure it's that problem. I see the Error loading models message from the very beginning, before I even do anything. I just go to the WebGUI and the error message already shows up. I would assume Lemonade is not trying to download some standard models by default?

jeremyfowers:
We now have an official docker image: https://github.com/lemonade-sdk/lemonade/pkgs/container/lemonade-server

So I am closing container-related issues and PRs.

Please shoot me a message if I am closing anything erroneously.

PR #242, opened 2025-08-17
Update styles.css so tooltips will show correctly

RobertAgee:
Tooltips previously obscured by other page elements. They now display correctly to inform users.

Changed CSS style for .model-browser-container and .model-browser-main from hidden to visible. I didn't notice any negative impacts on the display of any page elements.

before:

image
image

after:

image
image

Hope this helps!


RobertAgee:
Let's go!🚀

PR #244, opened 2025-08-18
System Tray Menu - Toggle debug logging, store user setting

RobertAgee:
Problem: Non-cli users can't intuitively toggle debug logging, monitor system performance. Server must be restarted to turn on/off

Solution: 
- Make debug logging directly accessible from the system tray menu

image

image

- Takes effect immediately, no need for server restart. Toggle on/off as desired.
- Toast notification confirmation
- User preference is persistently stored between sessions, reloaded automatically (set it and forget it)
- - Stored: C:\Users\\[username]\\.cache\lemonade in user_settings.json (alongside existing user_models.json)
- - **Future User Setting Idea: Always load XYZ model on Lemonade start-up?**


src/lemonade/tools/server/static/styles.css
RobertAgee, line 1: snuck in here from PR #242

src/lemonade_server/cli.py
RobertAgee, line 1: Majority of localhost mentions in the repo are 'localhost' as opposed to more explicit '127.0.0.1'. Happy with either, defaulted to most common occurrence in repo.

src/lemonade/tools/server/static/styles.css
RobertAgee: PR #242 is merged, so all good.

src/lemonade/tools/server/tray.py
```diff
@@ -417,6 +438,19 @@ def create_port_handler(port):
 
         port_submenu = Menu(*port_menu_items)
 
+        # Create the Logs submenu
+        debug_log_text = (
+            "Disable Debug Logs" if self.debug_logs_enabled else "Enable Debug Logs"
+        )
```
danielholanda:
```suggestion
        debug_log_text = "Enable Debug Logs"
```
I recommend not changing this name since we already have the `✓` "tick" on the side of the name to indicate whether the debug logs are enabled.

src/lemonade/tools/server/serve.py
```diff
@@ -67,7 +67,9 @@
     ResponsesRequest,
```
danielholanda, line 67: Note: You can run `pip install black` followed by `black src` to fix the lint issues.

RobertAgee: hahah, great minds and all that, thank you!

relinted everything to black, so 🤞 it'll go through this time 😅

RobertAgee (commented):
Reviewed

danielholanda (approved):
Awesome! Thanks for the great contribution!

PR #255, opened 2025-08-19
The load and completions/response endpoints do not upgrade models

### LINKED ISSUE #169: Lemonade re-downloading model after upgrade?
I have upgraded Lemonade to 8.1.1; after querying `gpt-oss-20b-GGUF`, the model got re-downloaded again. Is there a reason for this behaviour and if yes, can the old model be deleted (automatically)? For large models the space/bandwidth constraints can become significant.

Image

jeremyfowers:
Closes #169 

New rules:
1. The `pull` endpoint has the same behavior as before; it downloads the latest version of the model, even if you already had an older version.
2. The `load` endpoint has new behavior: if a local copy of the model is available, we load that (regardless of what is available on the server). Otherwise, we download the latest version.
    - Note: completions, chat/completions, and responses all call `load`, so this rule applies to those endpoints as well.

Also, this PR exposed and fixed a testing bug: we have "offline" server tests, but we were not actually ensuring all models were available offline before starting that test.

Design notes:
- There are quite a lot of code paths for loading/pulling a model, as you can see in this PR.
    - Most loads/pulls go through ModelManager's `download_models()` function
    - Some go through `lemonade.api` into the respective `Load Tool`.
- I deliberately didn't apply the rules above to Hugging Face PyTorch models, because it would have been too much work and I don't think anyone cares.
- It could behoove us to clean up load/pull code so that there's only 1 right way to download models in the future.

### DIFF
--- a/test/utils/server_base.py
+++ b/test/utils/server_base.py
@@ -37,6 +37,11 @@ except ImportError:
 
 
 MODEL_NAME = "Qwen2.5-0.5B-Instruct-CPU"
+# This list must include all models that could be accessed in offline testing
+MODELS_UNDER_TEST = [
+    MODEL_NAME,
+    "Llama-3.2-1B-Instruct-CPU",  # used in test_001_test_simultaneous_load_requests
+]
 MODEL_CHECKPOINT = "amd/Qwen2.5-0.5B-Instruct-quantized_int4-float16-cpu-onnx"
 PORT = 8000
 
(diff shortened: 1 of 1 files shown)


jeremyfowers:
How I tested:

1. Go into my HF cache, see 2 qwen3-coder snapshots
2. Delete the newer one (in all 3 subfolders)
3. Push the `load` button for qwen3-coder in the model manager. loads right away!
4. Unload the model
5. `lemonade-server-dev pull Qwen3-Coder-30B-A3B-Instruct-GGUF` -> this triggers a download

danielholanda (approved):
Works great! I just have one request that can become an issue if you don't want to add it to this PR: 

Users should be alerted when a new version of their model is available. Adding a warning to the log stating that and how to refresh the model would be great.

PR #258, opened 2025-08-19
Revert change that caused a bug in `lemonade-server-dev`

danielholanda:


jeremyfowers (approved):
I wound up disabling the stale test so that we can release the fix. See https://github.com/lemonade-sdk/lemonade/issues/262

PR #265, opened 2025-08-20
Refactor WrappedServer out of LlamaServer

jeremyfowers:
image

This PR paves the way for Lemonade to wrap additional types of servers.

The major change is that `server/llamacpp.py`
- used to be 1) largely functional, and 2) pass this `LlamaTelemetry` class around to all functions
- is now 1) object oriented, 2) much cleaner, 3) based on an `ABC` called `WrappedServer`

The `WrappedServer` actually took most of the code (almost 500 LoC) with only minor changes, since they turned out to be generic to any kind wrapped server. For example, the completions/chat/responses/embedding/rerank code was all generic.

Now `LlamaServer` is only responsible for a few llama-specific things like:
- Parsing llama-server.exe output logs
- Downloading a GGUF
- Downloading and installing llama-server.exe
- Launching llama-server.exe

### DIFF
--- a/src/lemonade/tools/server/llamacpp.py
+++ b/src/lemonade/tools/server/llamacpp.py
@@ -1,143 +1,27 @@
 import os
 import logging
-import time
 import subprocess
 import re
 import threading
 import platform
 
-import requests
-from tabulate import tabulate
 from dotenv import load_dotenv
-from fastapi import HTTPException, status
-from fastapi.responses import StreamingResponse
-
-from openai import OpenAI
 
 from lemonade_server.pydantic_models import (
-    ChatCompletionRequest,
-    CompletionRequest,
     PullConfig,
-    EmbeddingsRequest,
-    RerankingRequest,
 )
-from lemonade_server.model_manager import ModelManager
-from lemonade.tools.server.utils.port import find_free_port
 from lemonade.tools.llamacpp.utils import (
     get_llama_server_exe_path,
     install_llamacpp,
     download_gguf,
 )
+from lemonade.tools.server.wrapped_server import WrappedServerTelemetry, WrappedServer
 
 
-def llamacpp_address(port: int) -> str:
-    """
-    Generate the base URL for the llamacpp server.
-
-    Args:
-        port: The port number the llamacpp server is running on
-
-    Returns:
-        The base URL for the llamacpp server
-    """
-    return f"http://127.0.0.1:{port}/v1"
-
-
-def _separate_openai_params(request_dict: dict, endpoint_type: str = "chat") -> dict:
-    """
-    Separate standard OpenAI parameters from custom llama.cpp parameters.
-
-    Args:
-        request_dict: Dictionary of all request parameters
-        endpoint_type: Type of endpoint ("chat" or "completion")
-
-    Returns:
-        Dictionary with parameters properly separated for OpenAI client
-    """
-    openai_client_params = {}
-    extra_params = {}
-
-    # Common OpenAI parameters for both endpoint types
-    common_params = {
-        "model",
-        "frequency_penalty",
-        "logit_bias",
-        "logprobs",
-        "max_tokens",
-        "n",
-        "presence_penalty",
-        "seed",
-        "stop",
-        "stream",
-        "temperature",
-        "top_p",
-        "user",
-    }
-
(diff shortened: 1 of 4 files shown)


src/lemonade/tools/server/wrapped_server.py
```diff
@@ -0,0 +1,489 @@
...
+        """
+        Launch wrapped server subprocess with appropriate configuration.
+
+        Args:
+            snapshot_files: Dictionary of model files to load
+            supports_embeddings: Whether the model supports embeddings
+            supports_reranking: Whether the model supports reranking
+
+        Returns:
+            Subprocess handle for the wrapped server
```
danielholanda: The concrete implementation doesn't return anything - it sets self.process instead
```suggestion
    ):
        """
        Launch wrapped server subprocess with appropriate configuration.

        Args:
            snapshot_files: Dictionary of model files to load
            supports_embeddings: Whether the model supports embeddings
            supports_reranking: Whether the model supports reranking
```

src/lemonade/tools/server/wrapped_server.py
```diff
@@ -0,0 +1,489 @@
...
+    Manages telemetry data collection and display for wrapped server.
+    """
+
+    def __init__(self):
+        self.input_tokens = None
+        self.output_tokens = None
+        self.time_to_first_token = None
+        self.tokens_per_second = None
+        self.prompt_eval_time = None
+        self.eval_time = None
```
danielholanda, line 35: Works, but could be cleaner if we use `dataclasses` for the telemetry data instead of individual attributes.

src/lemonade/tools/server/wrapped_server.py
```diff
@@ -0,0 +1,489 @@
...
+
+    @property
+    def address(self) -> str:
+        """
+        Generate the base URL for the server.
+
+        Returns:
+            The base URL for the wrapped server
+        """
+        return f"http://127.0.0.1:{self.port}/v1"
```
danielholanda: [If you want the ‘textbook-style’ implementation] I think we should avoid using @property for the address method since it's computing a string each time rather than accessing stored state.

danielholanda (approved):
I can confirm this is a clean refactor with no functional changes. This makes the code a bit more complex, but will be important when adding new wrapped servers.

danielholanda:
Probably no overlap, but I if you can please merge this test first to make sure we have better coverage.
https://github.com/lemonade-sdk/lemonade/pull/267

src/lemonade/tools/server/llamacpp.py
```diff
@@ -1,144 +1,28 @@
 import os
```
danielholanda, line 1: Blocking: Please enable `test-server-llamacpp-ubuntu` (part of `test_llamacpp_oss.yml`) before continuing. You can skip rocm and do vulkan only. If this can't be done with a github runner, might be worth waiting for us to move the rocm ubuntu machines to this repo.

src/lemonade/tools/server/llamacpp.py
```diff
@@ -1,144 +1,28 @@
...
 import logging
```
danielholanda, line 2: Can you please remind me why we commented out tests 001-006 of `server_llamacpp.py`? Highly recommend uncommenting those before merging this refactor.

jeremyfowers: An accident! Good catch. Uncommenting.

src/lemonade/tools/server/wrapped_server.py
jeremyfowers: I thought dataclass was only cleaner if downstream code was initializing the members, which isn't the case here?

src/lemonade/tools/server/wrapped_server.py
jeremyfowers: Changed it back to a regular function

src/lemonade/tools/server/llamacpp.py
jeremyfowers: The way the test is written, it requires a self hosted runner and wont work on a VM. Will wait for you to bring the ubuntu machine online.

jeremyfowers: Done!

PR #287, opened 2025-08-22
Fix the run command and add test

danielholanda:
Closes https://github.com/lemonade-sdk/lemonade/issues/280


jeremyfowers (approved):
goooood

PR #289, opened 2025-08-22
Improve custom model registration process

danielholanda:
# Description

Previously, if someone tried to register a model with the same name but different parameters, it would silently use the existing configuration instead of validating or warning the user.

This PR fixes this by properly warning the user when that is the case.


src/lemonade_server/model_manager.py
```diff
@@ -201,6 +201,53 @@ def download_models(
                     reasoning=reasoning,
                 )
             else:
+                # Model is already registered - check if trying to register with different parameters
```
jeremyfowers, line 204: I'm pretty sure you could cut the LoC down by a lot (and improve maintainability in case we add more parameters) by using list comparison. But this works.

jeremyfowers (approved):
I literally just ran into this testing your other PR. good stuff.

PR #302, opened 2025-08-25
llamacpp errors should be visible to users

### LINKED ISSUE #75: llamacpp/GGUF model load fails (ERROR: vkDestroyFence: Invalid device)
Lemonade Server can fail to load GGUF models on certain AMD PCs. In debug mode (`lemonade-server serve --log-level debug`) the error looks like this:

Image

This error is caused by the AMD graphics driver being too old for the version of llama.cpp that Lemonade is using.

The solution is to download and install a driver update from the AMD website: https://www.amd.com/en/support/download/drivers.html

If a lot of people run into this, we can catch this error in the Lemonade code and raise a message with the solution above.

jeremyfowers:
Closes #75 

We have been hiding some errors under the llamacpp debug logs, when they would be really useful to users in debugging.

Additionally, the vkDestroyFence error is common, so this PR adds special handling for that.

Finally, we had some bugs with respect to showing errors in the webapp. This PR fixes those.

## Before:

No feedback whatsoever! Silent failure.

image

## After (this specific error):

"vkDestroyFence" raises a specific httpexception that gets elevated to the webapp.

image

image

## After (generic error):

Anything from the llamacpp log with the word "error" will get elevated to "logging.error"

image

### DIFF
--- a/src/lemonade/tools/server/llamacpp.py
+++ b/src/lemonade/tools/server/llamacpp.py
@@ -28,6 +28,9 @@ class LlamaTelemetry(WrappedServerTelemetry):
         Parse telemetry data from llama server output lines.
         """
 
+        if "error" in line.lower():
+            logging.error(line)
+
         # Parse Vulkan device detection
         vulkan_match = re.search(r"ggml_vulkan: Found (\d+) Vulkan devices?:", line)
         if vulkan_match:
--- a/src/lemonade/tools/server/llamacpp.py
+++ b/src/lemonade/tools/server/llamacpp.py
@@ -6,6 +6,7 @@ import threading
 import platform
 
 from dotenv import load_dotenv
+from fastapi import HTTPException, status
 
 from lemonade_server.pydantic_models import (
     PullConfig,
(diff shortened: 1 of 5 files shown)


src/lemonade/tools/server/wrapped_server.py
```diff
@@ -217,6 +220,12 @@ def _wait_for_load(self):
                 )
             time.sleep(1)
 
+        if self.log_thread_exception:
+            # Check if the
```
danielholanda: Check if the... ?
```suggestion
```

src/lemonade/tools/server/llamacpp.py
```diff
@@ -28,6 +29,20 @@ def parse_telemetry_line(self, line: str):
         Parse telemetry data from llama server output lines.
         """
 
+        if "vk::PhysicalDevice::createDevice: ErrorExtensionNotPresent" in line:
```
ramkrishna2910, line 32: Lets hope we dont have to track more errors like this :D

ramkrishna2910 (approved):

PR #303, opened 2025-08-25
Power profiling

amd-pworfolk:
These changes add power profiling to Lemonade.  There are two options `--power-hwinfo` and `--power-agt`.  The former uses the HWiNFO tool while the latter uses the AMD private AGT tool.

Please see the docs for configuration instructions and remember that these commands need to be executed as Administrator.  The following are example usages:

```
lemonade --power-hwinfo -i amd/Llama-3.2-1B-Instruct-awq-g128-int4-asym-fp16-onnx-hybrid oga-load --device hybrid --dtype int4 llm-prompt -p "Alice and Bob"
```

```
lemonade --power-agt -i amd/Llama-3.2-1B-Instruct-awq-g128-int4-asym-fp16-onnx-hybrid oga-load --device hybrid --dtype int4 llm-prompt -p "Alice and Bob"
```


src/lemonade/profilers/agt_power.py
```diff
@@ -0,0 +1,437 @@
+#
+# This power profiler uses the AMD internal AGT tool.
```
ramkrishna2910:
```suggestion
# This power profiler uses a tool that is not publicly available yet.
```

PR #310, opened 2025-08-26
Add getting started system messages to the web ui

### LINKED ISSUE #306: New pypi/source user experience: no model, no hints
If a new user installs from pypi or source they will have no model. If they open the web ui, the first thing they see is this:

Image

There's no model and no hint as to where a model might come from.

Suggested solution: if no models are installed chat box should have a system message that says: "head over to the Model Management tab and download a model with the 📥 button to get started."

cc @ramkrishna2910 relevant for macOS, noticed this during my PR review.

jeremyfowers:
Closes #306 

## How it used to look
Uhg, what am I supposed to do?
image

## New message for when you have no model
Test by setting `$env:HF_HOME="test_cache"`, which will make it seem like you have no models.
image

## New message for when you have a model
But you still need to load it...
image

### DIFF
--- a/src/lemonade/tools/server/static/js/chat.js
+++ b/src/lemonade/tools/server/static/js/chat.js
@@ -1,7 +1,7 @@
 // Chat logic and functionality
 let messages = [];
 let attachedFiles = [];
-let systemMessageElement = null; // Track the current system message element
+let systemMessageElement = null;
 
 // Default model configuration
 const DEFAULT_MODEL = 'Qwen2.5-0.5B-Instruct-CPU';
@@ -493,7 +493,7 @@ function displaySystemMessage() {
     if (messageText) {
         const div = document.createElement('div');
         div.className = 'chat-message system';
-        div.setAttribute('data-system-message', 'true'); // Mark as system message
+        div.setAttribute('data-system-message', 'true');
         
         const bubble = document.createElement('div');
         bubble.className = 'chat-bubble system';


src/lemonade/tools/server/static/js/chat.js
```diff
@@ -451,6 +460,53 @@ function appendMessage(role, text, isMarkdown = false) {
...
+        // No models installed - show first message
+        messageText = `Welcome to Lemonade! To get started:
+1. Head over to the Model Management tab.
+2. Use the 📥Download button to download a model.
+3. Use the 🚀Load button to load the model.
+4. Come back to this tab, and you are ready to chat with the model.`;
+    } else if (!currentLoadedModel) {
+        // Models available but none loaded - show second message
+        messageText = 'Welcome to Lemonade! Choose a model from the dropdown menu below to load it and start chatting.';
+    }
```
danielholanda, line 491: This second message is not needed since we already have "pick a model" written in the dropdown menu.
I would personally remove it, since I like UIs that are more minimalistic, but will let you make this call.

src/lemonade/tools/server/static/js/chat.js
```diff
@@ -451,6 +460,53 @@ function appendMessage(role, text, isMarkdown = false) {
...
+    
+    if (messageText) {
+        const div = document.createElement('div');
+        div.className = 'chat-message system';
+        div.setAttribute('data-system-message', 'true'); // Mark as system message
```
danielholanda:
```suggestion
        div.setAttribute('data-system-message', 'true');
```

src/lemonade/tools/server/static/js/chat.js
```diff
@@ -1,6 +1,7 @@
 // Chat logic and functionality
 let messages = [];
 let attachedFiles = [];
+let systemMessageElement = null; // Track the current system message element
```
danielholanda:
```suggestion
let systemMessageElement = null;
```

danielholanda (approved):
Works great! I find the initial message really helpful when no models are available.

src/lemonade/tools/server/static/js/chat.js
jeremyfowers: Ideally it shows on the first usage, and then never again

but we'd need a concept of memory in the web ui, which we do not have, so thats another PR

Issue #288, opened 2025-08-22
Updating custom models


danielholanda:
# Description

Currently, if a user tries to pull a custom model and the name is already taken the checkpoint is completely ignored

<img width="1156" height="96" alt="Image" src="https://github.com/user-attachments/assets/0ade7e53-baaf-4abc-8b2e-e7abc0114599" />

## Desired Behavior

User should be notified that the model name is already taken when the recipe/checkpoint name do not match and asked to either use a different name OR manually delete the model before updating it.

PR #316, opened 2025-08-27
Handle model pulling with conflicting parameters

danielholanda:
Closes https://github.com/lemonade-sdk/lemonade/issues/288

### DIFF
--- a/src/lemonade_server/cli.py
+++ b/src/lemonade_server/cli.py
@@ -197,15 +197,15 @@ def pull(
         for model_name in model_names:
             payload = {"model_name": model_name}
 
-            if checkpoint and recipe:
-                # Add the parameters for registering a new model
-                payload["checkpoint"] = checkpoint
-                payload["recipe"] = recipe
-
-                if reasoning:
-                    payload["reasoning"] = reasoning
-                if mmproj:
-                    payload["mmproj"] = mmproj
+            # Add the parameters to the payload
+            for key, value in [
+                ("checkpoint", checkpoint),
+                ("recipe", recipe),
+                ("reasoning", reasoning),
+                ("mmproj", mmproj),
+            ]:
+                if value:
+                    payload[key] = value
 
             # Install the model
             pull_response = requests.post(f"{base_url}/pull", json=payload)
@@ -225,6 +225,9 @@ def pull(
             recipe=recipe,
             reasoning=reasoning,
             mmproj=mmproj,
+            # The pull command will download an upgraded model if available, even
+            # if we already have a local copy of the model
+            do_not_upgrade=False,
         )
 
 
--- a/src/lemonade_server/model_manager.py
+++ b/src/lemonade_server/model_manager.py
@@ -146,7 +146,7 @@ class ModelManager:
         do_not_upgrade: prioritize any local copy of the model over any updated copy
             from the Hugging Face Hub.
         """
-        for model in models:
+        for model in models:        
             if model not in self.supported_models:
                 # Register the model as a user model if the model name
                 # is not already registered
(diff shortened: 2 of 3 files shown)


src/lemonade_server/model_manager.py
```diff
@@ -201,6 +201,54 @@ def download_models(
                     reasoning=reasoning,
                 )
             else:
+                # Model is already registered - check if trying to register with different parameters
```
danielholanda, line 204: Please do not review if this is still marked as a draft

test/model_manager.py
```diff
@@ -296,6 +296,44 @@ def test_user_model_registry_cleanup(self, mock_snapshot_download, mock_supporte
...
+                "model_name": "user.test-model"
+            }
+        }
+        mock_supported_models.return_value = existing_models
+
+        # Ensure we can pull a registered model if the same exact parameters are provided
+        self.model_manager.download_models(["user.test-model"], checkpoint="test/checkpoint1", recipe="oga-cpu", reasoning=False)
+
+        # Ensure we can pull a registered model, even if no parameters are provided
+        self.model_manager.download_models(["user.test-model"])
```
danielholanda, line 319: This is important to ensure users can pull models without providing all args all the time. We only raise an assertion if conflicting parameters have been found.

src/lemonade_server/model_manager.py
```diff
@@ -201,6 +201,54 @@ def download_models(
...
+                existing_checkpoint = existing_model.get("checkpoint")
+                existing_recipe = existing_model.get("recipe")
+                existing_reasoning = "reasoning" in existing_model.get("labels", [])
+                existing_mmproj = existing_model.get("mmproj", "")
+
+                # Compare parameters (handle None/empty string equivalence for mmproj)
+                checkpoint_differs = checkpoint and checkpoint != existing_checkpoint
+                recipe_differs = recipe and recipe != existing_recipe
+                reasoning_differs = reasoning and reasoning != existing_reasoning
+                mmproj_differs = mmproj and mmproj != existing_mmproj
```
danielholanda, line 215: Do we have to handle None/empty string equivalence for mmproj?

danielholanda: Tested in multiple scenarios. Please also bash it while reviewing.

src/lemonade_server/cli.py
```diff
@@ -225,6 +225,9 @@ def pull(
             recipe=recipe,
             reasoning=reasoning,
             mmproj=mmproj,
+            # The pull command will download an upgraded model if available, even
+            # if we already have a local copy of the model
+            do_not_upgrade=False,
```
danielholanda, line 230: Bug fix! We were only upgrading the models while pulling if the server was already running!

src/lemonade_server/model_manager.py
```diff
@@ -201,6 +201,54 @@ def download_models(
...
+                existing_checkpoint = existing_model.get("checkpoint")
+                existing_recipe = existing_model.get("recipe")
+                existing_reasoning = "reasoning" in existing_model.get("labels", [])
+                existing_mmproj = existing_model.get("mmproj", "")
+
+                # Compare parameters
+                checkpoint_differs = checkpoint and checkpoint != existing_checkpoint
+                recipe_differs = recipe and recipe != existing_recipe
+                reasoning_differs = reasoning and reasoning != existing_reasoning
+                mmproj_differs = mmproj and mmproj != existing_mmproj
```
danielholanda, line 215: We only raise an assertion if there is already a model registered with the provided name AND conflicting parameters have been found. We intentionally do not check for parameters that have been omitted by the user.

Example: Assume `user.Gemma-3-8b` has been previously registered with mmproj `file.mmproj`
* `lemonade-server pull user.Gemma-3-8b --mmproj file.mmproj` -> No assertion raised
* `lemonade-server pull user.Gemma-3-8b` -> No assertion raised
* `lemonade-server pull user.Gemma-3-8b --mmproj another_file.mmproj` -> Assertion raised

src/lemonade_server/cli.py
```diff
@@ -197,15 +197,15 @@ def pull(
...
 
-            if checkpoint and recipe:
-                # Add the parameters for registering a new model
-                payload["checkpoint"] = checkpoint
-                payload["recipe"] = recipe
-
-                if reasoning:
-                    payload["reasoning"] = reasoning
-                if mmproj:
-                    payload["mmproj"] = mmproj
```
jeremyfowers, line 208: What's the difference between the old and new? `checkpoint` and `recipe` are no longer required? 

Shouldn't `checkpoint` and `recipe` be required?

src/lemonade_server/cli.py
jeremyfowers: 😮

jeremyfowers: On this note... when you added the `run` command, you added the ability for the CLI to start the server, then do a command.

That capability didn't exist before `run`, which is why `pull`, `list`, `delete`, etc. all have 2 code paths.

Should those commands all be updated to behave just like `run`, and then we wouldn't ever have this kind of bug again?

danielholanda: Reducing code paths is always good. :)

src/lemonade_server/cli.py
danielholanda: yeah, but when pulling a model we want the ability to check for those separately. If you have any thoughts about this potentially generating a bug let me know and we should look into it.

Issue #320, opened 2025-08-28
Prevent the browser from caching the Lemonade web ui


jeremyfowers:
A frequent source of problems is that the web ui has been updated, but the browser caches the CSS or JavaScript, resulting in weird behavior.

Some code like this will prevent such caching:

```
from fastapi.staticfiles import StaticFiles

class NoCacheStaticFiles(StaticFiles):
    """Custom StaticFiles class with no-cache headers"""

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def file_response(self, *args, **kwargs) -> Response:
        response = super().file_response(*args, **kwargs)
        # Add no-cache headers for all static files
        response.headers["Cache-Control"] = "no-cache, no-store, must-revalidate"
        response.headers["Pragma"] = "no-cache"
        response.headers["Expires"] = "0"
        return response

app.mount("/static", NoCacheStaticFiles(directory=str(STATIC_DIR)), name="static")
```

The app.mount in Lemonade web ui is here: https://github.com/lemonade-sdk/lemonade/blob/c4ba594c343de69881aae64261a23cab04cdc103/src/lemonade/tools/server/serve.py#L200

p55d2k:
Hi, I'd like to work on this, please assign me this issue.

jeremyfowers:
Assigned, thanks!

jeremyfowers:
Closed by #325

PR #325, opened 2025-08-30
fix: prevent browser from caching lemonade ui

p55d2k:
Fix to Issue #320. Added a NoCacheStaticFiles class, to prevent browser from caching lemonade UI.

Changed files:
[src/lemonade/tools/server/serve.py](https://github.com/lemonade-sdk/lemonade/blob/main/src/lemonade/tools/server/serve.py#L201)


jeremyfowers (approved):
Thanks for making this improvement!

PR #347, opened 2025-09-04
llama.cpp prefill progress tracking

### LINKED ISSUE #259: Prefill progress tracking through tool-calling
## Problem

In local-ai setting, it can take lot of time to process the input tokens.
As application developer, it makes hard to deliver good UX. User should keep waiting.

## Solution(with llama-cpp-rs & hyprnote)

In [Hyprnote](https://github.com/fastrepl/hyprnote), we expose prefill progress through tool-calling. (Which is kind of only way to expose this while doing openai compat, I think)

https://github.com/user-attachments/assets/edc57851-48d6-4d42-90c6-3be477d2e513

Related code:

https://github.com/fastrepl/hyprnote/blob/b4b9544ed864093a9dab3f8ddfd802fe328889b5/plugins/local-llm/src/server.rs#L397-L420
https://github.com/fastrepl/hyprnote/blob/b4b9544ed864093a9dab3f8ddfd802fe328889b5/crates/llama/src/lib.rs#L111-L229
https://github.com/fastrepl/hyprnote/blob/b4b9544ed864093a9dab3f8ddfd802fe328889b5/plugins/local-llm/src/server.rs#L204-L232

## Solution(with lemonade)

Currently not implemented. But something like above can work.

## Other context

@ramkrishna2910 and I talked about this during the call.

itomek:
# Add Prefill Progress Tracking for LlamaCpp Streaming

Resolves #259

## Summary
Adds real-time progress tracking during the prefill (prompt processing) phase for llama.cpp streaming responses. Users now see progress updates (0% → 100%) before token generation begins.

## Changes
- Parse llama.cpp progress output and track prefill progress in `LlamaTelemetry`
- Stream progress updates as OpenAI-compatible tool calls before content chunks
- Use async streaming to monitor telemetry without blocking
- Ensure final 100% progress is always reported

## Testing
- Added test `test_007_test_prefill_progress_in_streaming` 
- Verifies progress updates are sent and reach 100% before content
- All existing tests pass
- Manually tested on Windows 11 and Ubuntu 24

## Example
```
Progress: 0.0%
Progress: 51.4%
Progress: 100.0%
[Content generation begins...]
```

image


yujonglee:
Thanks. Just found out that things like Vercel AI SDK warn about the tool that is never specified in the client SDK.
We might want to make this behavior optional/configurable.

ramkrishna2910:
@yujonglee Please take a look at this PR and let us know if this meets your requirements on reporting prefill progress.

src/lemonade/tools/server/llamacpp.py
```diff
@@ -24,11 +25,59 @@ class LlamaTelemetry(WrappedServerTelemetry):
...
+        # Prefill progress tracking
+        self.prefill_progress = 0.0  # Current progress (0.0 to 1.0)
+        self.last_reported_progress = 0.0  # Track last reported to avoid spam
+        self.prefill_total_tokens = 0  # Total tokens to process
+        self.prefill_processed_tokens = 0  # Tokens processed so far
+        self.last_progress_time = 0.0  # Track last update time for throttling
+        self.prefill_complete = False  # Flag to indicate prefill completion
+
+    def reset_prefill_progress(self):
+        """Reset prefill progress tracking for a new request."""
```
ramkrishna2910:
```suggestion
        """
        Reset prefill progress tracking for a new request.
        """
```
Following the same pattern as other functions

src/lemonade/tools/server/llamacpp.py
```diff
@@ -24,11 +25,59 @@ class LlamaTelemetry(WrappedServerTelemetry):
...
+        self.prefill_progress = 0.0
+        self.last_reported_progress = 0.0
+        self.prefill_total_tokens = 0
+        self.prefill_processed_tokens = 0
+        self.last_progress_time = 0.0
+        self.prefill_complete = False
+
+    def should_report_progress(self) -> bool:
+        """Check if progress should be reported based on throttling rules."""
```
ramkrishna2910:
```suggestion
        """
        Check if progress should be reported based on throttling rules.
        """
```

src/lemonade/tools/server/llamacpp.py
```diff
@@ -24,11 +25,59 @@ class LlamaTelemetry(WrappedServerTelemetry):
...
+
     def parse_telemetry_line(self, line: str):
         """
         Parse telemetry data from llama server output lines.
         """
 
+        # Parse prefill progress from llama.cpp server output
+        # Look for "prompt processing progress" with progress value
+        progress_match = re.search(
+            r"prompt processing progress.*progress\s*=\s*([\d.]+)", line
```
ramkrishna2910, line 72: We are relying on a specific format in the llamacpp output. We have seen changes to this in the past.
Can we add a test to make sure this pattern exists.
This would ensure progress doesnt break as we upgrade versions of llamacpp.

src/lemonade/tools/server/wrapped_server.py
```diff
@@ -300,6 +341,141 @@ def load(
                 detail=f"Failed to load {model_config.model_name} with {self.server_name}",
             )
 
+    def _supports_prefill_progress(self) -> bool:
+        """Check if telemetry supports prefill progress tracking."""
```
ramkrishna2910:
```suggestion
        """
        Check if telemetry supports prefill progress tracking.
        """
```
Please fix the other doc strings as well

src/lemonade/tools/server/wrapped_server.py
```diff
@@ -300,6 +341,141 @@ def load(
...
+
+    async def _monitor_prefill_progress_async(self):
+        """
+        Async generator that yields prefill progress updates from telemetry.
+        Used to report progress during the prefill phase before chunks arrive.
+        """
+        if not self._supports_prefill_progress():
+            return
+
+        timeout = 30  # Maximum seconds to wait
```
ramkrishna2910: I would suggest increasing this to 60s as with increasing context length support we could see greater than 30s prefill times.
Also suggest making this a class level constant.

danielholanda:
> Thanks. Just found out that things like Vercel AI SDK warn about the tool that is never specified in the client SDK. We might want to make this behavior optional/configurable.

Please make sure to address this request

src/lemonade/tools/server/wrapped_server.py
```diff
@@ -317,13 +493,11 @@ def chat_completion(self, chat_completion_request: ChatCompletionRequest):
...
+            async def event_stream():
                 try:
-                    # Enable streaming
-                    # pylint: disable=missing-kwoa
-                    for chunk in client.chat.completions.create(**openai_client_params):
-                        yield f"data: {chunk.model_dump_json()}\n\n"
-                    yield "data: [DONE]\n\n"
+                    # Stream with progress monitoring
+                    async for data in self._stream_with_progress(openai_client_params):
+                        yield data
```
danielholanda, line 506: Do we have any way of measuring whether this is adding any noticeable overhead to streaming?
Maybe run Qwen 0.6B  (or smaller) on main and using this branch to see if we get any performance degradation?

src/lemonade/tools/server/wrapped_server.py
danielholanda: Also, why do we need a timeout at all? Can't we completely skip this timeout logic and simply wait for the first token regardless of how long it takes?

jeremyfowers: Agree with Daniel, multi-minute prefills are totally possible on client if the context length is dialed way up.

itomek: Yes! It was meant to avoid mem leaks and timeout issues where I was experiening initially and it hung on parsing but i haven't seen it hang since so I agree, we should remove.

src/lemonade/tools/server/llamacpp.py
itomek: Added `test_020_llamacpp_prefill_progress_integration` in `server_llamacpp.py`

itomek:
> Thanks. Just found out that things like Vercel AI SDK warn about the tool that is never specified in the client SDK. We might want to make this behavior optional/configurable.

@yujonglee  is this necessary to do? We don't have a need to hide the prefill progress, so it just a matter of documenting it somewhere or are you saying that you can't proceed without a configurable flag to toggle prefill indication feature?

Issue #172, opened 2025-08-09
Investigate llamacpp think flag


jeremyfowers:
To see if it helps handle gpt-oss thinking behavior.

Labyricorn:
I'm getting the following from gpt-oss 20b through open-webui..
The left chat is using LM Studio to server, the center is using Ollama to serve and the right is using Lemonade to serve.
Is this an issue with Lemonade or with Open WebUI?

Thanks much,
Chris

<img width="1623" height="901" alt="Image" src="https://github.com/user-attachments/assets/c8f2af75-fb40-4c07-9906-254f3cf54162" />

jeremyfowers:
Thanks for reporting @Labyricorn! The last time I had checked, Open WebUI didn't support the OpenAI thinking format regardless of server. That probably means it's time to do something on the Lemonade side.

jeremyfowers:
Reference: OpenAI defined this new template format called Harmony. It has a Python lib, so perhaps we could cast it to regular thinking format within Lemonade's Python: https://github.com/openai/harmony

kpoineal:
I'm working on the --reasoning-mode auto fix. Once I test that I will see if it's easy to inttegrate harmony.

PR #356, opened 2025-09-08
Feat: Unified reasoning stream handling, animated thought UI, llama.cpp bump, and reasoning format auto-switch

kpoineal:
Closes https://github.com/lemonade-sdk/lemonade/issues/172

This PR upgrades both backend and frontend reasoning support. It standardizes llama.cpp invocation to use automatic reasoning formatting, bumps the Vulkan build version, removes a brittle Jinja disable path, and delivers a significantly more resilient chat UI that cleanly parses, streams, and animates chain-of-thought / reasoning tokens

# Backend changes:

- llamacpp.py: "--reasoning-format" none -> "--reasoning-format" auto to allow adaptive emission (legacy vs structured reasoning) without client breakage.
- Removed special-case Jinja disable logic for gpt-oss-120b on Vulkan
- utils.py: Vulkan llama.cpp version bump b6097 -> b6392 (aligns with upstream improvements / likely reasoning or stability fixes).

# Frontend (chat.js) reasoning & UX overhaul:

Introduced generalized reasoning parser supporting:
- think.../think (full or streaming, partial closures).
- Harmony channel variants: <|channel|>analysis<|channel|>, <|channel|>analysis<|message|>, <|channel|>analysis<|assistant|> with <|end|>.
- Converts streamed reasoning into a collapsible panel; auto-expanded while thinking, collapsible afterward.
- Reliable closure if the model stops mid-thought (auto-inserts closing tag for display).
- Splits stored assistant message into { content, reasoning_content } (no need for downstream template parsing).
- Incremental DOM updates (preserves user-expand/collapse state during streaming).
- Animated “Thinking…” header with optional 🍋 frame cycle; automatically stops when reasoning completes.

# User-visible behavior improvements:

- Real-time animated reasoning header gives clearer progress feedback.
- Thought process no longer wipes user toggle state while streaming tokens.
- Stable rendering even with mixed or malformed vendor reasoning markers.

### DIFF
--- a/src/lemonade/tools/server/static/js/chat.js
+++ b/src/lemonade/tools/server/static/js/chat.js
@@ -214,6 +214,8 @@ function handleChatInputKeydown(e) {
         e.preventDefault();
         clearAttachments();
     } else if (e.key === 'Enter' && !e.shiftKey) {
+        e.preventDefault();
+        // Check if we have a loaded model
         if (currentLoadedModel && modelSelect.value !== '' && !modelSelect.disabled) {
             sendMessage();
         } else if (!currentLoadedModel) {
--- a/docs/dev_cli/README.md
+++ b/docs/dev_cli/README.md
@@ -160,6 +160,11 @@ Hugging Face:
 This generates a PNG file that is stored in the current folder and the build folder.  This file
 contains a figure plotting the memory usage over the Lemonade tool sequence.  Learn more by running `lemonade -h`.
 
+### Power Profiling
+
+Power profiling is enabled through external tools.  Refer to [Power Profiling](power_profiling.md).
+
+
 ### System Information
 
 To view system information and available devices, use the `system-info` tool:
--- a/docs/dev_cli/power_profiling.md
+++ b/docs/dev_cli/power_profiling.md
@@ -1,0 +1,58 @@
+# Introduction
+
+Power profiling is a profiler functionality that allows power consumption
+measurements to be taken during Lemonade execution. It is only available
+for Windows machines. Power profiling can be done using the 3rd party tool HWiNFO or the
+private AMD AGT tool (AMD platforms only).  Instructions for each are below.
+
+## Power Profiling with HWiNFO
+
+To capture power data during the lemonade execution sequence with the HWiNFO tool:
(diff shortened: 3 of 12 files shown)


src/lemonade/tools/llamacpp/utils.py
```diff
@@ -14,7 +14,7 @@
```
jeremyfowers: image

I'm seeing this when I load models. Happened on Qwen3-0.6B-GGUF and gpt-oss-20b-GGUF so far. Happening to you @kpoineal?

I need to check if this happens for me on main.

jeremyfowers: PS. everything works after I see this error, so it's a false negative.
PPS. works fine on `main`.

src/lemonade/tools/server/static/js/chat.js
```diff
@@ -431,208 +324,330 @@ function updateInputPlaceholder() {
...
+
     div.appendChild(bubble);
     chatHistory.appendChild(div);
     chatHistory.scrollTop = chatHistory.scrollHeight;
-    return bubble; // Return the bubble element for streaming updates
+    return bubble;
 }
 
-// Display system message based on current state
-function displaySystemMessage() {
```
jeremyfowers, line 466: Was this function removed by mistake?

src/lemonade/tools/llamacpp/utils.py
jeremyfowers: Looks like this is the culprit https://github.com/lemonade-sdk/lemonade/pull/356/files#r2331111838

kpoineal: I've re-added this function and validated the correct functionality.

src/lemonade/tools/server/static/js/chat.js
kpoineal: Yes. It has been re-added.

src/lemonade/tools/llamacpp/utils.py
jeremyfowers: Resolved!

Issue #274, opened 2025-08-21
lllamacpp --seed argument doesn't seem to work with llamacpp-rocm


jeremyfowers:
This prevents the test `test_007_test_generation_parameters_with_llamacpp` from working, since it relies on requests with identical arguments to produce identical results.

Example failing workflow: https://github.com/lemonade-sdk/lemonade/actions/runs/17114851893/job/48543625025?pr=271

That test is skipped on rocm until this is resolved; to close this issue, remove the lines of code for skipping the test.

jeremyfowers:
Looks like this is specific to newer llamacpp versions and not specific to ROCm.

The problem has been reproduced on vulkan with `LLAMA_VERSION_VULKAN = "b6392"`

danielholanda:
Fully working now! 

To reproduce:
```
llama-cli.exe -mu "https://huggingface.co/unsloth/Qwen3-0.6B-GGUF/resolve/main/Qwen3-0.6B-Q4_0.gguf" -ngl 99 -p "Hello, how are you? Tell me more about..." -n 30 -st --seed 321
```
> Note: I used our new [fancy curl support](https://github.com/lemonade-sdk/llamacpp-rocm/pulls) for this test

<img width="1214" height="1047" alt="Image" src="https://github.com/user-attachments/assets/82ab84ae-6e9d-4266-9e6f-a776b8c5f6d2" />

jeremyfowers:
That’s great! But let’s only close the issue when we re enable the generation params CI tests.

PR #359, opened 2025-09-08
Update LLAMA_VERSION_VULKAN to b6392

jeremyfowers:
This PR revs our llamacpp vulkan version to `b6392`, which carries these improvements:
1. Support for JINA embeddings model requested by user
2. Adds a new "performance stats chunk" at the end of streaming messages (see below)
3. Fixes the gpt-oss-120b + vulkan jinja error, allowing tool calling with that model (a different PR will clean that up in the server)
4. Helpful Vulkan OOM messages:
image

With regards to the performance stats chunk, it looks like this and shows up at the end of a response stream after all message chunks have arrived:

```
ChatCompletionChunk(id='chatcmpl-FcbtiTqDpL8xekTjeSKBuxJi2HL8jBAo', choices=[], created=1757427579, model='Qwen3-0.6B-GGUF', object='chat.completion.chunk', service_tier=None, system_fingerprint='b6392-5143fa89', usage=CompletionUsage(completion_tokens=292, prompt_tokens=59, total_tokens=351, completion_tokens_details=None, prompt_tokens_details=None), timings={'prompt_n': 1, 'prompt_ms': 15.434, 'prompt_per_token_ms': 15.434, 'prompt_per_second': 64.79201762342879, 'predicted_n': 292, 'predicted_ms': 2923.172, 'predicted_per_token_ms': 10.01086301369863, 'predicted_per_second': 99.89148774003036})
```

This is pretty cool and we should consider using it for perf instead of snooping on log telemetry. However, it also breaks our llamacpp streaming chat completions test, since that test expects all chunks to include `choices[0]`, which is no longer a safe assumption.

This PR also skips testing for "with a locked seed, two responses to the same prompt should be identical". See #274. We should resolve that bug outside of this PR.

### DIFF
--- a/test/server_llamacpp.py
+++ b/test/server_llamacpp.py
@@ -104,7 +104,7 @@ class LlamaCppTesting(ServerTestingBase):
         complete_response = ""
         chunk_count = 0
         for chunk in stream:
-            if chunk.choices[0].delta.content is not None:
+            if chunk.choices and chunk.choices[0].delta.content is not None:
                 complete_response += chunk.choices[0].delta.content
                 print(chunk.choices[0].delta.content, end="")
                 chunk_count += 1


src/lemonade/tools/llamacpp/utils.py
```diff
@@ -14,7 +14,7 @@
 
 from dotenv import set_key, load_dotenv
 
-LLAMA_VERSION_VULKAN = "b6097"
+LLAMA_VERSION_VULKAN = "b6392"
 LLAMA_VERSION_ROCM = "b1021"
```
danielholanda, line 18: PR looks good, my only suggestion is to update this to LLAMA_VERSION_VULKAN  `b6431` and LLAMA_VERSION_ROCM  `b1057`. Both of those correspond to Llama.cpp Commit Hash: 4f63c (latest).

test/server_llamacpp.py
```diff
@@ -104,7 +104,7 @@ def test_001_test_llamacpp_chat_completion_streaming(self):
         complete_response = ""
         chunk_count = 0
         for chunk in stream:
-            if chunk.choices[0].delta.content is not None:
+            if chunk.choices and chunk.choices[0].delta.content is not None:
```
danielholanda, line 107: I'm trusting you tested streaming here and you are seeing a response.

jeremyfowers: yes i did

src/lemonade/tools/llamacpp/utils.py
jeremyfowers: follow up PR, since it may involve yet more work

PR #360, opened 2025-09-09
block default Enter behavior

RobertAgee:
Simple fix to block default Enter key behavior on chatInput.value when sending a message


RobertAgee:
@jeremyfowers Cool, just sent you an email

PR #372, opened 2025-09-10
Fix wildcard load and `gpt-oss-20b-mxfp4-GGUF` model load

danielholanda:
# Description

This PR fixes loading/downloading models with the new ":*" syntax.

Before, all files would be downloaded and our chat interface would often try to load files like `.gitignore` instead of the appropriate .gguf files.

This PR fixes this bug by only downloading the required gguf files.

As a result, models like `gpt-oss-20b-mxfp4-GGUF` can now be correctly loaded.


src/lemonade/tools/llamacpp/utils.py
```diff
@@ -531,13 +531,14 @@ def identify_gguf_models(
 
     # (case 0) Wildcard, download everything
     if variant and variant == "*":
-        sharded_files = repo_files
+        sharded_files = [f for f in repo_files if f.endswith(".gguf")]
```
jeremyfowers, line 534: If you have a little more time to spend on this, I suggest changing it to just make sure the first GGUF gets loaded instead of not downloading non-GGUF files.

PR #373, opened 2025-09-10
Implement individual model retrieval endpoint `(/models/{model_id})`

danielholanda:
# Description

This PR implements OpenAI's individual model retrieval endpoint `(/models/{model_id})`.


src/lemonade/tools/server/serve.py
```diff
@@ -1590,6 +1592,36 @@ async def models(self):
...
+            # Mimic the error format of the OpenAI API
+            raise HTTPException(
+                status_code=404,
+                detail={
+                    "message": f"model {model_id} not found",
+                    "type": "api_error",
+                    "param": None,
+                    "code": None,
+                },
+            )
```
danielholanda, line 1610: Here is proof that we are raising exactly the same error as Ollama.

```
from openai import OpenAI

use_lemonade=True

if use_lemonade:

    client = OpenAI(
        base_url="http://localhost:8000/api/v1",
        api_key="-",  # required, but unused
    )
else:
    client = OpenAI(
        base_url="http://localhost:11434/v1",
        api_key="-",  # required, but unused
    )
client.models.retrieve("non-existent-model")
```

test/server.py
```diff
@@ -173,6 +175,18 @@ def test_004_test_models(self):
...
+        test_model = l.data[0]
+        model = client.models.retrieve(test_model.id)
+        assert model.id == test_model.id
+        assert model.checkpoint == test_model.checkpoint
+        assert model.recipe == test_model.recipe
+
+        # Check that the individual model endpoint returns a NotFoundError for a non-existent model
+        # Do not modify this test as this exact error type might be checked by different ISVs
+        with pytest.raises(NotFoundError):
+            client.models.retrieve("non-existent-model")
```
danielholanda: We are raising a `HTTPException` exception and the openai library is responsible for turning it into a NotFoundError.

test/server.py
```diff
@@ -173,6 +175,18 @@ def test_004_test_models(self):
...
+        # Check the individual model endpoint
+        test_model = l.data[0]
+        model = client.models.retrieve(test_model.id)
+        assert model.id == test_model.id
+        assert model.checkpoint == test_model.checkpoint
+        assert model.recipe == test_model.recipe
+
+        # Check that the individual model endpoint returns a NotFoundError for a non-existent model
+        # Do not modify this test as this exact error type might be checked by different ISVs
+        with pytest.raises(NotFoundError):
```
jeremyfowers: Why install the whole pytest framework just for this? There's no way to do it in unittest framework?

docs/server/server_spec.md
```diff
@@ -356,6 +357,48 @@ curl http://localhost:8000/api/v1/models
...
+  "object": "model",
+  "owned_by": "lemonade",
+  "checkpoint": "amd/Llama-3.2-1B-Instruct-awq-g128-int4-asym-fp16-onnx-hybrid",
+  "recipe": "oga-hybrid"
+}
+```
+
+#### Error responses
+
+If the model is not found, the endpoint returns a 404 error:
```
jeremyfowers, line 391: I tested on some models that have real model names, but aren't downloaded, and they raise the 404 error:

```
C:\Users\jfowe>curl http://localhost:8000/api/v1/models/bge-reranker-v2-m3-GGUF
{"detail":{"message":"model bge-reranker-v2-m3-GGUF not found","type":"api_error","param":null,"code":null}}
```

Is that intended behavior? Or should that model be found, but with a `"downloaded": false` key?

.github/workflows/test_server.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           python -m pip install --upgrade pip
```
jeremyfowers: This will be nice to have in LemonadeClient/Arcade, opened https://github.com/lemonade-sdk/lemonade-arcade/issues/45

.github/workflows/test_server.yml
```diff
@@ -40,6 +40,7 @@ jobs:
...
           python -m pip check
           pip install -e .[dev,oga-cpu]
+          python -m pip install pytest
```
jeremyfowers:
```suggestion
```

Please consider removing, see other comment

docs/server/server_spec.md
jeremyfowers: BTW these 404s on real models will make this endpoint useless for retrieving details in the web ui model manager. To be useful, the endpoint needs to work on every model displayed by the manager.

.github/workflows/test_server.yml
danielholanda: Removed and replaced by unittest!

PR #378, opened 2025-09-11
[Discussion] Should we deprecate the developer entrypoint?

danielholanda:
# Description

Can we remove this now that the log level is a persistent setting?


danielholanda:
Actually, the dev entrypoint still has a benefit on Linux, since there is no Tray there.
Closing it for now.

Issue #388, opened 2025-09-14
Add server restart functionality to Lemonade tray


siavashhub:
### Description
The Lemonade tray server currently lacks a restart feature. This enhancement request proposes adding a "restart server" capability to improve usability and allow users to quickly restart the Lemonade server from the tray interface.

### Benefits
-- Improves user experience by providing a quick way to restart Lemonade when needed.

jeremyfowers:
Hi @siavashhub thanks for the issue and PR!

Would you kindly help us better understand how you plan to use this feature? ie, why does Lemonade Server need to be restarted?

siavashhub:
Hi @jeremyfowers 
Sure, I don’t have an AMD platform, so I’ve been testing Lemonade on my Intel-based machine. Not sure if it’s related, but sometimes the server halts or sometimes the UI refresh stops working and the only way to recover is by quitting and restarting it with the shortcut, which gets a bit annoying. I worked around it by automating the restart with a PowerShell script for myself. I know that’s not a real fix and it could be just an issue from my side, but I thought having this as a built-in feature could be useful for others as well. It would also be handy for dev contributions that require a server reload to test new code locally.

jeremyfowers:
Gotcha! It would be nice if we could fix the hang/halt issues for you.

> having this as a built-in feature could be useful for others as well. It would also be handy for dev contributions that require a server reload to test new code locally

Yeah I can see it being useful as a developer-focused feature. I'm just not sure about exposing it to all users. What would you think of hiding it under a developer env var or something?

siavashhub:
> What would you think of hiding it under a developer env var or something?

Sure, I refactored it by loading env on top:
https://github.com/lemonade-sdk/lemonade/pull/389/commits/205260aa02328b2f59e921f073d6759a1dc6cffa#diff-a2250f18b7d4024b45d6c4c8754ae50fba600412b760cb889f23185d7e10521aR19-R26

and hid the item behind the environment variable:
https://github.com/lemonade-sdk/lemonade/pull/389/commits/205260aa02328b2f59e921f073d6759a1dc6cffa#diff-a2250f18b7d4024b45d6c4c8754ae50fba600412b760cb889f23185d7e10521aR576-R577

This way it will only show up if someone place a .env file under Lemonade folder with below value 
`LEMONADE_ENV=Developer`

Would this be acceptable?

PR #389, opened 2025-09-14
Add Lemonade server restart functionality to tray

siavashhub:
### Description 
This PR adds a restart feature to the Lemonade tray server. Users can now restart the Lemonade server directly from the tray, improving workflow efficiency and convenience.

### Changes
- Added restart server logic to `src\lemonade\tools\server\tray.py`
- Ensured clean shutdown and restart of server processes

Closes https://github.com/lemonade-sdk/lemonade/issues/388

### DIFF
--- a/src/lemonade/tools/server/tray.py
+++ b/src/lemonade/tools/server/tray.py
@@ -7,6 +7,7 @@ import webbrowser
 from pathlib import Path
 import logging
 import tempfile
+import platform
 
 import requests
 from packaging.version import parse as parse_version
@@ -14,7 +15,20 @@ from packaging.version import parse as parse_version
 from lemonade_server.pydantic_models import DEFAULT_CTX_SIZE
 
 from lemonade.version import __version__
-from lemonade.tools.server.utils.system_tray import SystemTray, Menu, MenuItem
+
+# Import the appropriate tray implementation based on platform
+if platform.system() == "Darwin":  # macOS
+    from lemonade.tools.server.utils.macos_tray import (
+        MacOSSystemTray as SystemTray,
+        Menu,
+        MenuItem,
+    )
+else:  # Windows/Linux
+    from lemonade.tools.server.utils.windows_tray import (
+        SystemTray,
+        Menu,
+        MenuItem,
+    )
 
 # adding environemnt variable support
 from dotenv import load_dotenv
@@ -94,6 +108,9 @@ class LemonadeTray(SystemTray):
         # Background thread for version checking
         self.version_check_thread = None
         self.stop_version_check = threading.Event()
+
+        # Hook function for platform-specific initialization callback
+        self.on_ready = None
 
     def get_latest_version(self):
         """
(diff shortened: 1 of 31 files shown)


src/lemonade/tools/server/tray.py
```diff
@@ -531,6 +573,8 @@ def create_ctx_handler(ctx_size):
         items.append(MenuItem("Model Manager", self.open_model_manager))
         items.append(MenuItem("Logs", None, submenu=logs_submenu))
         items.append(Menu.SEPARATOR)
+        if LEMONADE_ENV == "Developer":
```
danielholanda, line 619: Can we use `LEMONADE_ADVANCED_TRAY_OPTIONS` and just check if it was set instead of `LEMONADE_ENV`?

src/lemonade/tools/server/tray.py
```diff
@@ -401,6 +409,40 @@ def toggle_debug_logs(self, _, __):
...
+                "subprocess.run([r'{}', '-m', 'lemonade_server.cli', 'stop'], shell=True); "
+                "time.sleep(2); "
+                "subprocess.Popen(['wscript.exe', r'{}'], shell=True)"
+            ).format(sys.executable.replace("\\", "/"), vbs_path)
+            subprocess.Popen(
+                [sys.executable, "-c", restart_script],
+                shell=True,
+                creationflags=subprocess.DETACHED_PROCESS
+                | subprocess.CREATE_NEW_PROCESS_GROUP,
+            )
```
danielholanda, line 477: Do we really need to create a whole new detached process like this? Can't we simply call `change_port`?
Then, if `self.port==new_port` you simply show  the message "Server Rebooted" instead of "Port Changed".

danielholanda (commented):
Thanks for the PR! I suggested a simpler approach that may have the same outcome. Please let me know if this works.

src/lemonade/tools/server/tray.py
siavashhub: So instead of using `LEMONADE_ENV=Developer` in the .env, we’d switch to `LEMONADE_ADVANCED_TRAY_OPTIONS=True`? If that works better for the team, I’m happy to update it, just let me know and I’ll make the change.

src/lemonade/tools/server/tray.py
siavashhub: The issue I ran into is that without starting a new detached process, the Lemonade server itself doesn’t fully restart, only Uvicorn and the web UI do. For example, if this happens during a model download, the download continues, so it’s not a true restart. On the other hand, if I force a full exit, there’s no process left to restart. That’s why I opted for creating a new detached process.

src/lemonade/tools/server/tray.py
danielholanda: Yes, I think this other env var would be a better fit.

src/lemonade/tools/server/tray.py
danielholanda: The current logic unfortunately doesn't work for me (developer installation flow)
image

The proposed solution also ignores certain parameters that the user may have set when starting the server e.g., ctx-size).

A proper solution would likely be a restart command inside cli.py. That said, making sure that all corners are round could be a significant amount of effort.

I would recommend trying a different approach that may work for all users regardless of installation flow (and ideally also regardless of OS - as we have MacOS support now).

Another option is to identify the issues that are causing the need to reboot. If we can fix those directly, we might be able to remove the need for a restart button altogether.

jeremyfowers:
Closing Python PRs as Python codebase will be mostly removed in the next week or two.

PR #396, opened 2025-09-18
Update Llama.cpp+ROCm to use ROCWMMA

danielholanda:
# Description

This PR updates ROCm to use ROCWMMA and syncs all Llama.cpp versions to commit 703f9e3.

### DIFF
--- a/.github/workflows/test_llamacpp_oss.yml
+++ b/.github/workflows/test_llamacpp_oss.yml
@@ -7,9 +7,8 @@ on:
   push:
     branches: ["main"]
   pull_request:
-    branches:
-      - '**'
   workflow_dispatch:
+  merge_group:
 
 
 permissions:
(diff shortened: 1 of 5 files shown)


src/lemonade/version.py
```diff
@@ -1 +1 @@
-__version__ = "8.1.10"
+__version__ = "8.1.11"
```
ramkrishna2910, line 1: Should this be 8.2.0 as the next release would include the macos changes?

ramkrishna2910 (approved):

src/lemonade/version.py
danielholanda: I don't think macOS is causing any breaking changes. If we are still following semantic versioning, there is no need to bump to 8.2.0. My recommendation is to keep it this way for now and bump to 8.2.0 before the Wednesday release if needed.

jeremyfowers: Hehe marketing version numbers :)

Let's have a quick sync today to discuss our rollouts for the next 2 weeks? Should be hype!

jeremyfowers: (quick, someone make an extremely minor breaking change!)

PR #419, opened 2025-09-26
Detect AMD GPUs without lspci

superm1:
shelling out to lspci and relying upon pciids.txt is fragile, especially when trying to run on an old distro.  To avoid risk with this instead parse amdgpu sysfs files to determine if an APU or dGPU and build the graphics architecture from the HW IP version.

NOTE: this is not a perfect solution still because HW IP and SW IP version might not match.  But it's certainly A LOT closer to right.

There might need to be a matrix to map out HW IP to SW IP versions as another layer.


superm1:
So I was thinking about this and I have a MUCH BETTER approach.  There is a sysfs file called `properties` that will have the matching ISA.  Parse this.

```
❯ pwd
/sys/class/kfd/kfd/topology/nodes/1
❯ cat properties
cpu_cores_count 0
simd_count 64
mem_banks_count 1
caches_count 70
io_links_count 1
p2p_links_count 0
cpu_core_id_base 0
simd_id_base 2147487744
max_waves_per_simd 16
lds_size_in_kb 64
gds_size_in_kb 0
num_gws 64
wave_front_size 32
array_count 4
simd_arrays_per_engine 2
cu_per_simd_array 10
simd_per_cu 2
max_slots_scratch_cu 32
gfx_target_version 110501
vendor_id 4098
device_id 5510
location_id 49664
domain 0
drm_render_minor 128
hive_id 0
num_sdma_engines 1
num_sdma_xgmi_engines 0
num_sdma_queues_per_engine 6
num_cp_queues 8
max_engine_clk_fcompute 2700
local_mem_size 0
fw_version 32
capability 671326848
capability2 0
debug_prop 1495
sdma_fw_version 17
unique_id 0
num_xcc 1
max_engine_clk_ccompute 3600
```

superm1:
@danielholanda 

Here's what I've got for using KFD to get the info and link it back to the PCI ID.  I'm confused; it seems like there are both Python and C++ implementations now though.  Why are both there?  Should this be redone in C++ instead?  Or it needs to be in both?

```python
    def _detect_amd_gpus(self, gpu_type: str, include_inference_engines: bool = False):
        """
        Shared AMD GPU detection logic for both integrated and discrete GPUs.
        Uses keyword-based classification for simplicity and reliability.

        Args:
            gpu_type: Either "integrated" or "discrete".

        Returns:
            list: List of detected GPU info dictionaries.
        """
        gpu_devices = []
        try:
            gpu_info = {}
            kfd_dir = "/sys/class/kfd/kfd/topology/nodes/"
            if not os.path.isdir(kfd_dir):
                return gpu_devices

            for node in os.listdir(kfd_dir):
                properties_path = os.path.join(kfd_dir, node, "properties")
                if os.path.isfile(properties_path):
                    try:
                        with open(properties_path, "r", encoding="utf-8") as f:
                            for line in f:
                                if "gfx_target_version" in line:
                                    gpu_info["isa"] = line.split()[1].strip()
                                if "drm_render_minor" in line:
                                    gpu_info["drm_render_minor"] = line.split()[1].strip()
                    except (OSError, ValueError):
                        continue

            device_path = f"/sys/class/drm/renderD{gpu_info['drm_render_minor']}/device/"
            board_info_path = os.path.join(device_path, "board_info")

            is_discrete = os.path.isfile(board_info_path)
            is_integrated = not is_discrete

            # Filter based on requested type
            if (gpu_type == "integrated" and is_integrated) or (
                gpu_type == "discrete" and is_discrete
            ):
                device_type = "amd_igpu" if is_integrated else "amd_dgpu"

            # Get VRAM information for discrete GPUs
            if is_discrete:
                gpu_info["vram_gb"] = self._get_amd_vram_sysfs(device_path)

            if include_inference_engines:
                gpu_info["inference_engines"] = (
                    self._detect_inference_engines(device_type, gpu_info["isa]"])
                )
            gpu_devices.append(gpu_info)

        except Exception as e:  # pylint: disable=broad-except
            error_msg = f"AMD {gpu_type} GPU detection failed: {e}"
            return [{"available": False, "error": error_msg}]
```

superm1:
@danielholanda OK - I've come up with a much more flexible approach.  I parse the ISA from KFD.  This should mean zero hardcoding for Linux in the future.

src/cpp/server/system_info.cpp
```diff
@@ -1290,92 +1300,85 @@ NPUInfo LinuxSystemInfo::get_npu_device() {
...
-    // Execute lspci to find GPUs
-    FILE* pipe = popen("lspci 2>/dev/null | grep -iE 'vga|3d|display'", "r");
-    if (!pipe) {
+    std::string kfd_path = "/sys/class/kfd/kfd/topology/nodes";
+
+    if (!fs::exists(kfd_path)) {
         GPUInfo gpu;
         gpu.available = false;
-        gpu.error = "Failed to execute lspci command";
+        gpu.error = "No KFD nodes found (AMD GPU driver not loaded or no GPU present)";
```
danielholanda, line 1280: What is the guidance for the user in this scenario? Download the latest driver from Adrenalin?

src/lemonade/common/system_info.py
```diff
@@ -1216,79 +1220,21 @@ def _get_nvidia_vram_smi_linux(self) -> float:
...
 
         Returns:
-            float: VRAM in GB, or 0.0 if detection fails
+            float: VRAM in GBfails
         """
-        try:
-            # Try different sysfs paths for VRAM information
-            sysfs_paths = [
-                f"/sys/bus/pci/devices/{pci_id}/mem_info_vram_total",
-                "/sys/class/drm/card*/device/mem_info_vram_total",
```
danielholanda, line 1268: Would this ever fail due to permission issues?

src/lemonade/common/system_info.py
```diff
@@ -1216,79 +1220,21 @@ def _get_nvidia_vram_smi_linux(self) -> float:
...
-                    # Extract numeric value (assuming it's in MB or GB)
-                    numbers = re.findall(r"\d+", line)
-                    if numbers:
-                        vram_value = int(numbers[0])
-                        # Assume MB if value is large, GB if small
-                        if vram_value > 100:  # Likely MB
-                            vram_gb = round(vram_value / 1024, 1)
-                        else:  # Likely GB
-                            vram_gb = float(vram_value)
-                        return vram_gb
```
danielholanda, line 1249: @ramkrishna2910 Would be great if you could take a look here

src/cpp/server/system_info.cpp
superm1: Most users should never see this.   If you don't have a GPU driver you're probably going to have a pretty bad experience 😛 

Here's a few reasons it might come up:
* Brand new hardware, missing driver.  This case means they need to update their kernel (or their distro)
* Manually blocked amdgpu from loading.  This case they need to unblock it.
* Older hardware, older distro.  Update hardware.

src/lemonade/common/system_info.py
superm1: Nope!  Permissions are openly readable for all users for this file.

```
❯ ls -alh /sys/class/drm/card0/device/mem_info_vram_total
-r--r--r-- 1 root root 4.0K Dec  8 16:02 /sys/class/drm/card0/device/mem_info_vram_total
```

superm1:
@danielholanda @ramkrishna2910 can you review this?  It will need to be rebased as there are new conflicts, but I'd like to make sure I don't have anything fundamental wrong before I rebase if possible.

jeremyfowers:
> @danielholanda @ramkrishna2910 can you review this? It will need to be rebased as there are new conflicts, but I'd like to make sure I don't have anything fundamental wrong before I rebase if possible.

Hi @superm1 sorry for the very delayed response. We are stretched pretty thin right now, and this kind of falls under “if it ain’t broke, don’t fix it” for me (I’m not aware of any Issue or Discord conversation that has a lot of comments about this).

If this fixes a specific issue you are having on your own system I am open to merging it, however I do not want to replace the existing methodology that works for most people. We could add this methodology as a fallback in case lspci fails, though.

Finally, I’ll add that system_info.py (and the vast majority of other Python code in this repo) is deprecated and will be removed in the next week or two.

superm1:
> Hi @superm1 sorry for the very delayed response. We are stretched pretty thin right now, and this kind of falls under “if it ain’t broke, don’t fix it” for me (I’m not aware of any Issue or Discord conversation that has a lot of comments about this).

It won't be broke for current hardware; but the current solution will not scale for future hardware.  I'll get more to it below.

> If this fixes a specific issue you are having on your own system I am open to merging it, however I do not want to replace the existing methodology that works for most people. We could add this methodology as a fallback in case lspci fails, though.

Future hardware will have new PCI IDs and the strings are updated out of band by manually modifying https://admin.pci-ids.ucw.cz/.  Then distros need to pull down the updates PCI IDs, which doesn't happen quickly automatically.

So for example if you have some new hardware that works with an IDENTICAL ISA (which is the case for multiple products this year) Lemonade wouldn't be able to recognize it in mainline LTS distros like Ubuntu without a lot of extra work.

But if you switch to my proposed approach Lemonade will work on the new hardware with no other changes to the distros, PCI IDs database, etc.  You can take a ROCm 7.x stack and run on anything launching with an identical ISA to existing hardware but a new PCI ID.

> Finally, I’ll add that system_info.py (and the vast majority of other Python code in this repo) is deprecated and will be removed in the next week or two.

So if I redo the PR to _not touch any Python code_ would it be more acceptable?

superm1:
I have rebased, and confirmed the C++ implementation still downloads the correct binary on my SUT.

src/lemonade/common/system_info.py
superm1: As Jeremy suggested to drop python implementation this comment is no longer relevant.

src/lemonade/common/system_info.py
superm1: As Jeremy suggested to drop python implementation this comment is no longer relevant.

jeremyfowers:
> I have rebased, and confirmed the C++ implementation still downloads the correct binary on my SUT.

Thanks @superm1! Looking forward to reviewing this.

src/cpp/server/system_info.cpp
```diff
@@ -363,7 +363,22 @@ bool SystemInfo::check_vulkan_support() {
 std::string identify_rocm_arch_from_name(const std::string& device_name) {
```
jeremyfowers, line 376: @superm1 there is a problem on here somewhere.

When I try to install rocm via lemonade on my Radeon 9070 xt system on `main`:

```
[LlamaCpp] Detected ROCm architecture: gfx120X
[LlamaCpp] Downloading from: https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1162/llama-b1162-ubuntu-rocm-gfx120X-x64.zip
[LlamaCpp] Downloading to: /home/jfowers/.cache/huggingface/hub/llama-b1162-ubuntu-rocm-gfx120X-x64.zip
  Progress: 6% (45.8/661.3 MB)
  Progress: 16% (108.3/661.3 MB)
```

When I try on your branch:

```
[LlamaCpp] Detected ROCm architecture: gfx1201
[LlamaCpp] Downloading from: https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1162/llama-b1162-ubuntu-rocm-gfx1201-x64.zip
[LlamaCpp] Downloading to: /home/jfowers/.cache/huggingface/hub/llama-b1162-ubuntu-rocm-gfx1201-x64.zip

[Download] Found partial file (0.0 MB), resuming...

[Download] Error (attempt 1): HTTP error 404 for URL: https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1162/llama-b1162-ubuntu-rocm-gfx1201-x64.zip
```

The download needs to use a name that matches the builds here: https://github.com/lemonade-sdk/llamacpp-rocm/releases/tag/b1162

But your way returns 1201 (surely more precise), when  120X is what is expected.

superm1: Ah it looks to me that TheRock has some optimization that they decided to group a few together for the dGPUs.

https://github.com/ROCm/TheRock/blob/main/RELEASES.md#index-page-listing

I guess I see two approaches.

1. If 404 then "fall back" to replacing last digit with an "X" and try again.
2. Have a mapping table that groups GFX targets.

I prefer the former (I feel like it scales better) but what do you think?

jeremyfowers: > If 404 then "fall back" to replacing last digit with an "X" and try again.

This sounds ok, except that we have some retry logic on downloads. Would not want to wait for the full timeout and all retries before putting the X.

> Have a mapping table that groups GFX targets.

This sounds like less effort IMO. The mapping is on llamacpp-rocm readme:

```
gfx1151 (STX Halo APU) - Ryzen AI MAX+ Pro 395
gfx1150 (STX Point APU) - Ryzen AI 300
gfx120X (RDNA4 GPUs) - includes AMD Radeon RX 9070 XT/GRE/9070, RX 9060 XT/9060
gfx110X (RDNA3 GPUs) - includes AMD Radeon PRO W7900/W7800/W7700/W7600, RX 7900 XTX/XT/GRE, RX 7800 XT, RX 7700 XT/7700, RX 7600 XT/7600
```

superm1: OK, added a mapping database for the ones grouped as families.  Hopefully this works for you.

PR #424, opened 2025-09-29
Fix model manager button status during long tasks

jeremyfowers:


kpoineal:
I am not seeing an hour glass when deleting, loading, or installing a model after the button has been clicked. I *do* see an hourglass when I use the model drop down. 

Can you confirm this is working for you?

PR #435, opened 2025-10-01
Add addition acknowledgement for FLM 

ZaneNi:
FLM used [mlir-aie](https://github.com/Xilinx/mlir-aie) repo from AMD.


jeremyfowers (approved):
Thanks @ZaneNi!

Issue #344, opened 2025-09-04
Set think/no_think in LLM Chat and HTTP requests


jeremyfowers:
Hybrid reasoning models like Qwen3-8B-GGUF are supposed to be able to take a `think` vs. `no_think` parameter at runtime that determines whether they will perform reasoning or just directly answer the question.

Currently, Lemonade does not pass this parameter to the model and we always get thinking.

This feature would be impactful because `Qwen3-8B-GGUF` is still the best 8B model, so unlocking the no_think mode would make it usable in apps like Lemonade Aracde.

This PR is a good blueprint for how to add parameters in general: https://github.com/lemonade-sdk/lemonade/pull/186

jeremyfowers:
https://github.com/Kritik-07 is planning to work on this

PR #438, opened 2025-10-02
Enable thinking flag support for completions and chat/completions (#344)

Kritik-07:
### Summary
- This PR addresses issue #344 by adding support for the enable_thinking flag in the completions and chat/completions APIs.

### Changes
- Updated src/lemonade/tools/server/serve.py to handle the enable_thinking parameter.
- Updated src/lemonade_server/pydantic_models.py to add the corresponding field.
- Added a test script test_thinking.py for basic verification.

### Testing
- Verified locally with Qwen3-4B-Instruct-2507-GGUF: request succeeds but reasoning is ignored (as expected, since this model does not support reasoning).
- Verified with Qwen3-4B-GGUF (hybrid reasoning model): reasoning works correctly when enable_thinking: true is passed.

### Notes
- The Responses API is not supported for GGUF models (thanks @jeremyfowers for clarifying). This PR only covers completions and chat/completions, as discussed.

### DIFF
--- a/src/lemonade/tools/oga/migration.py
+++ b/src/lemonade/tools/oga/migration.py
@@ -401,4 +401,3 @@ def delete_incompatible_models(
         "freed_size_formatted": format_size(freed_size),
         "cleaned_user_models": cleaned_user_models,
     }
-    


jeremyfowers:
Thanks for the PR, @Kritik-07!

Looking at the failed Test Lemonade Server CI job: 

```
[stderr] TypeError: OrtGenaiModel.generate() got an unexpected keyword argument 'enable_thinking'
[stderr] ERROR:    Exception raised in generate thread: OrtGenaiModel.generate() got an unexpected keyword argument 'enable_thinking'
```

I think you only implemented `enable_thinking` for GGUF/FLM models, and not OGA models. This is fine, because we dont have any hybrid reasoning OGA models.

However, the OGA models are getting confused by the enable_thinking parameter. You need to make sure that is handled correctly.

jeremyfowers:
FYI the Test Lemonade on NPU failure is unrelated to your PR, #443 will fix

Kritik-07:
Hi @jeremyfowers , I’m running into a couple of question while fixing the enable_thinking tests:

Can you help me where is this model OGA is defined or how I should import it? I couldn’t find its location in the repo.

Right now, GGUF models (like Qwen hybrids) work fine with enable_thinking, but OGA models fail if we include that parameter. Should I:

Only keep enable_thinking for GGUF/Qwen models, and strip it for OGA?

Or is there already a standard way in the codebase to handle unsupported args for OGA models?

Just want to be sure I’m handling this in the simplest correct way.

jeremyfowers:
> Hi @jeremyfowers , I’m running into a couple of question while fixing the enable_thinking tests:
> 
> Can you help me where is this model OGA is defined or how I should import it? I couldn’t find its location in the repo.
> 
> Right now, GGUF models (like Qwen hybrids) work fine with enable_thinking, but OGA models fail if we include that parameter. Should I:
> 
> Only keep enable_thinking for GGUF/Qwen models, and strip it for OGA?
> 
> Or is there already a standard way in the codebase to handle unsupported args for OGA models?
> 
> Just want to be sure I’m handling this in the simplest correct way.

Sorry about my very slow reply @Kritik-07!

> Can you help me where is this model OGA is defined or how I should import it? I couldn’t find its location in the repo.

OGA models get loaded here: https://github.com/lemonade-sdk/lemonade/blob/6de207926ec7fe55c9ff79e70226ad2b7e15fad8/src/lemonade/tools/server/serve.py#L1641

Basically any time you see `if self.llm_loaded.recipe == "llamacpp" or self.llm_loaded.recipe == "flm"`, the `else` is OGA

> Only keep enable_thinking for GGUF/Qwen models, and strip it for OGA?

Yes

> Or is there already a standard way in the codebase to handle unsupported args for OGA models?

I don't think so. This kind of situation hasn't come up too often, so we don't have a formal solution.

Kritik-07:
Hi @jeremyfowers
Update: The OGA `enable_thinking` parameter issue is fixed! 
Current CI status:
- **Main fix verified**: No more `TypeError` for OGA models
-  **Ubuntu**: 1 timeout in `test_013_test_load_checkpoint_completion` 
-  **Windows**: NPU test canceled due to incompatible hardware (AMD EPYC vs Ryzen AI)

Both remaining failures appear unrelated to the `enable_thinking` changes. The feature logic is working as intended - only passing the parameter to non-OGA Qwen models.
Let me know if there is something I should work on now.

jeremyfowers:
Ok hopefully this was a transient issue. I'm rerunning the jobs now to see if that helps.

Kritik-07:
Hi @jeremyfowers, 
Thanks for re-running! Unfortunately both tests failed again with the same issues:

**Ubuntu**: Timeout in `test_013_test_load_checkpoint_completion` (openai.APITimeoutError: Request timed out.)
**Windows**: NPU hardware incompatibility
Should we merge as-is and address these test flakes separately?
Happy to make any adjustments you'd like.

jeremyfowers:
> Hi @jeremyfowers, Thanks for re-running! Unfortunately both tests failed again with the same issues:
> 
> **Ubuntu**: Timeout in `test_013_test_load_checkpoint_completion` (openai.APITimeoutError: Request timed out.) **Windows**: NPU hardware incompatibility Should we merge as-is and address these test flakes separately? Happy to make any adjustments you'd like.

I'm taking a deeper look now!

jeremyfowers:
@Kritik-07 I'm open to dealing with the test_server issue in a separate PR.

However, I ran your test script (with one modification) and found that it is still outputting reasoning content.

If you change `print(completion.choices[0].message.content)` to `print(completion.choices[0].message)` in your 3 tests, that will print the whole message object, which includes both `content` and `reasoning_content`.

When I do this, your test 2 (no reasoning) looks like this. Note that there is `reasoning_content`. Can you please check your implementation or let me know what I'm doing wrong?

> Test 2: enable_thinking=False (fast mode)
> Response without thinking:
> ChatCompletionMessage(content='The product of 15 and 23 is calculated as follows:\n\n1. **Breakdown using distributive property**:  \n   $ 15 \\times 23 = 15 \\times (20 + 3) = (15 \\times 20) + (15 \\times 3) = 300 + 45 = 345 $.\n\n2. **Alternative method**:  \n   $ 15 \\times 23 = (10 + 5) \\times 23 = (10 \\times 23) + (5 \\times 23) = 230 + 115 = 345 $.\n\nBoth methods confirm the result. \n\n**Answer:**  \n$$\n\\boxed{345}\n$$', refusal=None, role='assistant', annotations=None, audio=None, function_call=None, tool_calls=None, reasoning_content="Okay, let's see. The user is asking for 15 multiplied by 23. Hmm, I need to calculate that. Let me think about the best way to do this. I remember that multiplication can be done using different methods, like the standard algorithm or maybe breaking it down into smaller parts. \n\nFirst, maybe I can use the standard multiplication method. Let me write it out. So, 15 times 23. Let me write it vertically:\n\n    15\n  x23\n  ----\n\nStarting with the 3 in the ones place of 23. Multiply 15 by 3. 15*3 is 45. Then, move to the 2 in the tens place of 23. That's actually 20, so 15*20. Let me calculate that. 15*20 is 300. Then add those two results together: 45 + 300. That gives 345. Wait, so 15*23 is 345?\n\nAlternatively, maybe I can break it down using distributive property. 23 can be thought of as 20 + 3. So 15*(20 + 3) = 15*20 + 15*3. Which is 300 + 45 = 345. Same result. That seems right.\n\nAlternatively, another way is to think of 15*23. Maybe 15 is a nice number. 15 is 10 + 5. So 23*10 + 23*5. 23*10 is 230, and 23*5 is 115. Adding those together: 230 + 115 = 345. Yep, same answer again. \n\nWait, let me check with another method just to be sure. Maybe using the standard algorithm step by step. Multiply 15 by 23. Let's do it as 15 x 23. \n\nStarting from the right, 15 x 3 is 45. Then 15 x 2 is 30, but since the 2 is in the tens place, it's actually 300. So adding 45 + 300 gives 345. \n\nAlternatively, maybe using a calculator? Wait, but since I need to do this manually, let me think. Another way is to use the fact that 15 is 3*5, and 23 is a prime number. But maybe that's complicating. \n\nAlternatively, think of 15*23 as (10 + 5)*23 = 10*23 + 5*23. 10*23 is 230, 5*23 is 115. 230 + 115 is indeed 345. \n\nHmm, seems consistent. Maybe I can check with a different approach. Let's do 23*15. 23*10 is 230, 23*5 is 115, total 345. \n\nAlternatively, using the formula for multiplying numbers. Let me recall. Or maybe using the method of multiplying 23 by 15. \n\nAlternatively, maybe using the standard multiplication steps. \n\nAnother way: 23*15. Let me write 23*15. \n\n23\nx15\n----\nFirst, multiply 23 by 5: 23*5 = 115. Then multiply 23 by 10: 23*10 = 230. Then add them: 115 + 230 = 345. \n\nYes, same answer. \n\nWait, just to make sure I didn't make a mistake in any of these steps. Let me check 23*5. 20*5 is 100, 3*5 is 15, so 100+15=115. Correct. Then 23*10 is 230. 230 + 115 is 345. Correct. \n\nAlternatively, think of 15*23 as 20*23 - 5*23. 20*23 is 460, 5*23 is 115. 460 - 115 = 345. That also works. \n\nAnother way: 15*23 is the same as 15*20 + 15*3 = 300 + 45 = 345. \n\nAll these methods lead to 345. So I think that's the right answer. I don't see any errors in my reasoning. Let me just verify with a calculator-like approach. 15 times 20 is 300, 15 times 3 is 45, so 300 + 45 is 345. Yep. \n\nAlternatively, if I think of 15*23 as (16 -1)*23. 16*23 is 368, minus 1*23 is 23, so 368 -23 = 345. That also works. \n\nSo, all methods converge to 345. I think that's the correct answer. So the answer should be 345.")

Kritik-07:
Thanks @jeremyfowers  for pointing out.

 There is no problem from your side, this is actually a bug.
 So I have done some try and testing and now I'm in little confusion and need your help:
 
 I was curious how think and no_think are defined so I came up with below code:
 ```
**Code**
  `from transformers import AutoTokenizer
  
  tokenizer = AutoTokenizer.from_pretrained( "Qwen/Qwen3-4B",
  use_fast=True,
  trust_remote_code=True)
  
  print("Vocab size:", len(tokenizer))
  print("/think in vocab?", "/think" in tokenizer.get_vocab())
  print("/no_think in vocab?", "/no_think" in tokenizer.get_vocab())`
```

  **output:**
  Vocab size: 151669
  /think in vocab? True
  /no_think in vocab? False

This means:

- Lemonade’s current logic of prepending  /no_think does not activate any special behavior in the model.
- The model treats these as ordinary text tokens,  so it have no effect.

 **Question 1:** So I'm confused since **think** is in vocab but why not **no_think** . And my earlier logic which I said it above is correct or not?

Second thing I tested out:
I've been trying to fix the `enable_thinking=False` issue but I'm running into some confusion. 

**Current Implementation:**
I'm modifying the message before tokenization and passing `enable_thinking` to `generation_kwargs`:
```
python
# Before tokenization
if enable_thinking is not None and "qwen" in model_name.lower():
    if enable_thinking:
        message = f"/think\n{message}"
    else:
        message = f"/no_think\n{message}"

# Later in generation_kwargs (for non-OGA models only)
generation_kwargs["enable_thinking"] = enable_thinking

```

**Question 2:** My above logic is correct?  

**Question 3:**  Do llamacpp/transformers actually accepts** `enable_thinking` parameter

**Question 4:**  Since no_think is not working, I thought can we use a NLP for prompting
Updated _generate_tokens() like this:

```
if "qwen" in self.llm_loaded.model_name.lower():
    if enable_thinking is True:
        message = "Let's think step by step.\n\n" + message
    elif enable_thinking is False:
        message = "Answer directly and concisely without any explanation.\n\n" + message
    # If enable_thinking is None, leave message unchanged (preserves default behavior)
```
    
  My logic behind this code:
  This will actually influence Qwen3’s behavior, because it’s trained on such instructions.
  So do you think this logic is correct?
  
  
  ---

**Summary of what I need:**

To properly implement this feature, I need to understand:
1. Is `/no_think` supposed to be in the vocab, or am I misunderstanding how it works?
2. Should I be passing `enable_thinking` as a parameter to the model engine, or handling it purely through prompting?
3. What's the correct way to disable thinking for Qwen3 models in Lemonade?

Would you be able to point me to the specific Qwen3 documentation you mentioned, or share how you expected this to work? That would help me implement it correctly.

Thanks for your patience as I work through this!

jeremyfowers:
Thanks for the deep thoughts @Kritik-07!

The documentation I referenced is here: https://huggingface.co/Qwen/Qwen3-4B#enable_thinkingtrue

My suggestion is to first try and get things working with Transformers and llama.cpp, since those are the reference designs. Once you have that, you will know that your requests are properly formatted, and that will remove a lot of ambiguity from the debugging process.

Also remember that for GGUFs, Lemonade is just forwarding an OpenAI http request to a llama-server subprocess. So if you get things working with regular llama.cpp/llama-server then it should be easy to get it working in Lemonade too.

Kritik-07:
Hi @jeremyfowers,
I was trying to figure out the problem and looks like the request never went to correct place.I tried to run some tests and I found out that The request is going to http://127.0.0.1:49678 - the llama-server subprocess not the Lemonade's main server.
`INFO:     HTTP Request: POST http://127.0.0.1:49678/v1/chat/completions "HTTP/1.1 200 OK"`

So, for GGUF models, requests go directly to llama-server subprocess. How should we handle enable_thinking in this case? Do we need to apply the chat template before forwarding, or does llama-server support this parameter?

jeremyfowers:
> Hi @jeremyfowers, I was trying to figure out the problem and looks like the request never went to correct place.I tried to run some tests and I found out that The request is going to http://127.0.0.1:49678 - the llama-server subprocess not the Lemonade's main server. `INFO: HTTP Request: POST http://127.0.0.1:49678/v1/chat/completions "HTTP/1.1 200 OK"`
> 
> So, for GGUF models, requests go directly to llama-server subprocess. How should we handle enable_thinking in this case? Do we need to apply the chat template before forwarding, or does llama-server support this parameter?

FYI the way that Lemonade supports llamacpp is:
1. Start a llama-server subprocess on a random free port (in this case, 49678)
2. Forward all chat/completions requests from the client to that port.
3. Route the responses back to the client.

> does llama-server support this parameter?

I don't actually know! Best way to find out is probably to try it and see if it has an effect.

BTW we are merging support for Ryzen AI SW 1.6 this week, and that will add new Qwen3 hybrid reasoning NPU models to Lemonade :) So your work is becoming more important.

Kritik-07:
Hi @jeremyfowers,

**Update:**
 I confirmed llama-server doesn't support `enable_thinking` - it ignores the parameter and still generates reasoning content.

**The Core Problem:**
For GGUF models, Lemonade bypasses `apply_chat_template()` and forwards requests directly to llama-server. This means the chat template (which is where `enable_thinking` should be applied according to Qwen3 docs) never gets called.

**My Question:**
What's the simplest way to make `enable_thinking` work for Qwen3 GGUF models? 

Should I:
1. Format the messages with `apply_chat_template(enable_thinking=...)` and send the formatted text to llama-server instead of the messages array?
2. Or is there a different approach you'd recommend?

I want to make sure I implement this correctly, especially with the new Ryzen AI NPU Qwen3 models coming soon. I'm committed to getting this feature working properly!

Thanks for your guidance.

jeremyfowers:
How does this sound for a llamacpp solution... if the user/client puts `enable_thinking=False` into their request, you append `/no_think` to the prompt?

Appending `/no_think` to the prompt works for me with the Qwen3 hybrid reasoning GGUF:

image

Kritik-07:
Hi @jeremyfowers,

I'm happy to report that the `enable_thinking` feature is now working! 🎉

**Final Implementation:**
- For llamacpp/GGUF models: appends `/no_think` to the prompt when `enable_thinking=False`
- Tested with Qwen3-4B-GGUF: both thinking and no-thinking modes work perfectly
- Fixed the OGA model compatibility issue we encountered earlier

**Test Results:**
- ✅ `enable_thinking=True`: Shows full reasoning_content
- ✅ `enable_thinking=False`: Direct answers, no reasoning_content
- ✅ Default behavior: Thinking mode (maintains backward compatibility)

Thank you so much for your patience and guidance throughout this process! Your suggestion to append `/no_think` to the prompt was exactly what was needed. It took some time to understand the llamacpp architecture, but we got there in the end.

I'm excited that this will be ready for the upcoming Ryzen AI NPU Qwen3 models! Looking forward to seeing this feature in action.

Thanks again for all your help!

jeremyfowers:
hi @Kritik-07 I just pulled your branch, and the feature is working for me! Awesome!

Before we can merge though, there's a lot of cleanup needed here so that your branch only contains the lines of code for your specific change. I might suggest just making a clean PR based on `main`.

PR #445, opened 2025-10-03
Add gpt-oss-20b to FLM

Tetramatrix:


jeremyfowers:
image

Shows up in model manager

PR #446, opened 2025-10-05
Fix typo in example script in concepts.md

jasonhernandez:


jeremyfowers (approved):
Thanks very much for catching this, @jasonhernandez!

Issue #275, opened 2025-08-21
Lemonade-server can access a local model folder and correctly load models downloaded from different model communities(ModelScope)


Jays-1111:
The background of this matter is that some clients encounter network restrictions when accessing huggingface to download models. Currently, the default download method of lemonade server is to download from huggingface.  there are some communities that can provide model downloads without network restrictions. so I wonder if it is possible to support the selection of the source for model download, so as to help ISVs use and promote them better.
such as ：download from https://modelscope.cn/

@jeremyfowers 
It can be summarized into the following two schemes
1. allow a path-on-disk to be an alternative to a checkpoint name, when pulling/installing a model
2. fully enable modelscope as an alternative source of models to hugging face

Note：Using a disk path is a straightforward solution to this problem. Maybe we need to consider the sources of the models being read—such as those downloaded from Hugging Face or Model Scope—to ensure they can be correctly identified.

function example：
Lmstudio：Open download directory

The desired functions might be in two stages:
1. The first is that users can download the model unrestrictedly from the model scope by themselves first and then open the folder on the lemonade-server to read and load the model correctly.
2. The final stage is that lemonade-server supports downloading models from multiple model communities such as (huggingface, modelscope).

jeremyfowers:
Also getting requested in the discord: https://discord.com/channels/1392562559122407535/1394403929973985413/1408388342227140699

jeremyfowers:
@meghsat is working on this

GCHarry12530:
Is this improvement still under developing? Currently lemonade would download models to the default directory, which is at the C drive. This consumes a great amount of space on the C drive. so I hope that lemonade could freely load models from any local directory. Or, it is also nice that lemonade would provide a option that users can change the folder that local models are downloaded.

jeremyfowers:
@GCHarry12530 the download location is determined by the `HF_HOME` environment variable, which you can set to whatever you like.

hj8e45:
@jeremyfowers I noticed this feature has implemented in #456, but there are no instructions or docs about how to load local model, if any?

jeremyfowers:
> [@jeremyfowers](https://github.com/jeremyfowers) I noticed this feature has implemented in [#456](https://github.com/lemonade-sdk/lemonade/pull/456), but there are no instructions or docs about how to load local model, if any?

@hj8e45 it's in the help menu for `pull`:

```
(base) PS C:\Users\Jeremy> lemonade-server.exe pull --help
Download a model
[...]

PR #456, opened 2025-10-09
Feature to upload local models to the UI

meghsat:
Closes https://github.com/lemonade-sdk/lemonade/issues/275

Changes:
1. Automatically refresh the models list after creating new custom models or deleting a model, without requiring a page refresh. This is referenced from SERVER_MODELS. Model deletion removes the model from user_models.json and the HF cache directory, so it’s permanently deleted unless you have it saved elsewhere.
2. Upload a local model at any folder level.
    Steps:
    a. After performing these initial checks:
       ONNX: Accepts the folder if it finds any valid ONNX files after searching recursively   (src/lemonade/tools/oga/load.py/find_onnx_files_recursively).
       Llamacpp: Accepts the folder if there is at least one .gguf file in it.
       MMproj: The frontend automatically populates the MMPROJ filename wherever it is found in the selected folder (the filename must contain the text “mmproj” and end with .gguf). The user can edit this field as well.
3. Copies the uploaded folder to the HF cache directory so that the local folder can be safely deleted. The checkpoint points to this HF cache/model_folder base location.
4. Registers the model locally (src/lemonade_server/model_manager.py).
5. When the user loads an LLM:
    OGA: The logic recursively searches for genai_config.json and loads that file’s root directory as the checkpoint.
    Llamacpp: The logic recursively searches for .gguf files and loads that file’s root directory as the checkpoint.
    MMproj: The submitted MMPROJ filename is located in the uploaded directory, and that path is used as the checkpoint.

### DIFF
--- a/src/lemonade_server/model_manager.py
+++ b/src/lemonade_server/model_manager.py
@@ -123,10 +123,17 @@ class ModelManager:
                 checkpoint = model_info["checkpoint"]
                 base_checkpoint, variant = parse_checkpoint(checkpoint)
                 if model.startswith("user."):
-                    # Parse checkpoint to get base path (models--xxx) without variant
-                    local_model_path = os.path.join(HF_HUB_CACHE, base_checkpoint)
-                    if os.path.exists(local_model_path):
-                        downloaded_models[model] = model_info
+                    # Check if this is a locally uploaded model or internet-downloaded model
+                    if model_info.get("source") == "local_upload":
+                        # Locally uploaded model: checkpoint is in cache directory format (models--xxx)
+                        local_model_path = os.path.join(HF_HUB_CACHE, base_checkpoint)
+                        if os.path.exists(local_model_path):
+                            downloaded_models[model] = model_info
+                    else:
+                        # Internet-downloaded model: checkpoint is in HuggingFace format (org/repo)
+                        # Check if it's in the downloaded checkpoints list
+                        if base_checkpoint in downloaded_checkpoints:
+                            downloaded_models[model] = model_info
                     continue
 
                 if base_checkpoint in downloaded_checkpoints:
@@ -426,41 +433,10 @@ class ModelManager:
             # We do this registration after the download so that we don't register
             # any incorrectly configured models where the download would fail
             if new_registration_model_config:
-                from huggingface_hub.constants import HF_HUB_CACHE
-
-                # Parse checkpoint to get base checkpoint without variant
-                base_checkpoint, variant = parse_checkpoint(checkpoint)
-
-                if current_recipe and current_recipe.startswith("hf-"):
-                    cache_checkpoint = checkpoint
-                else:
-                    # For GGUF and OGA models, convert to cache directory format
-                    repo_cache_name = base_checkpoint.replace("/", "--")
-                    cache_checkpoint = f"models--{repo_cache_name}"
-
-                    # For GGUF models, preserve variant so the loader knows which file to use
-                    if variant:
-                        cache_checkpoint = f"{cache_checkpoint}:{variant}"
-
-                    # For OGA models, resolve the path to genai_config.json
-                    elif current_recipe and current_recipe.startswith("oga-"):
-                        model_cache_dir = os.path.join(HF_HUB_CACHE, cache_checkpoint)
-
-                        # Search for genai_config.json in the downloaded model
-                        resolved_checkpoint = None
-                        if os.path.exists(model_cache_dir):
-                            for root, _, files in os.walk(model_cache_dir):
-                                if "genai_config.json" in files:
-                                    # Store relative path from HF_HUB_CACHE for portability
-                                    resolved_checkpoint = os.path.relpath(
-                                        root, HF_HUB_CACHE
-                                    )
-                                    break
-
-                        if resolved_checkpoint:
-                            cache_checkpoint = resolved_checkpoint
-
-                new_user_model["checkpoint"] = cache_checkpoint
+                # For models downloaded from the internet (HuggingFace),
+                # keep the original checkpoint format (e.g., "amd/Llama-3.2-1B-Instruct-...")
+                # Do NOT convert to cache directory format - that's only for locally uploaded models
+                new_user_model["checkpoint"] = checkpoint
 
                 if os.path.exists(USER_MODELS_FILE):
                     with open(USER_MODELS_FILE, "r", encoding="utf-8") as file:


src/lemonade/tools/llamacpp/utils.py
```diff
@@ -646,14 +646,24 @@ def identify_gguf_models(
...
-def download_gguf(config_checkpoint, config_mmproj=None, do_not_upgrade=False) -> dict:
+def download_gguf(config_checkpoint: str, config_mmproj=None, do_not_upgrade: bool = False) -> dict:
     """
     Downloads the GGUF file for the given model configuration.
 
     For sharded models, if the variant points to a folder (e.g. Q4_0), all files in that folder
     will be downloaded but only the first file will be returned for loading.
     """
+    # Check if this is a user-uploaded model (path contains models-- pattern or is already a full path)
+    if "models--" in config_checkpoint and os.path.exists(config_checkpoint):
```
jeremyfowers: is the `models--` naming check really required? What is the downside to accepting any `os.path.exists` here?

src/lemonade/tools/llamacpp/utils.py
```diff
@@ -646,14 +646,24 @@ def identify_gguf_models(
...
+        result = {"variant": config_checkpoint}
+        if config_mmproj:
+            result["mmproj"] = config_mmproj
+        return result
```
jeremyfowers: Can those whole code block be removed without any functional change? Unless I am missing something, it looks like:

```python
if CONDITION:
    do X
    return Y

do X
return Y
```

jeremyfowers: It also looks like on line 666 you have a `return`, which makes it impossible to reach line 669. Something is funky here, please revise :)

src/lemonade/tools/server/serve.py
```diff
@@ -7,12 +7,14 @@
...
-from typing import Optional, Union
+from typing import Optional, Union, List
 import json
 from pathlib import Path
 import os
-
-from fastapi import FastAPI, HTTPException, status, Request, WebSocket
+import shutil
+from pathlib import Path
+from huggingface_hub.constants import HF_HUB_CACHE
```
jeremyfowers: We rely on lazy imports to help this module load fast. Please move the `from huggingface_hub.constants import HF_HUB_CACHE` into the function that needs it (`upload_model()`)?

src/lemonade/tools/llamacpp/utils.py
```diff
@@ -9,7 +9,7 @@
 import requests
```
jeremyfowers, line 9: Please see the code formatting guidelines: https://github.com/lemonade-sdk/lemonade/blob/main/docs/contribute.md#code-formatting

They will help you get past the failed formatting checks!

I also sent you an invite to the lemonade-sdk org so that these checks will run for you automatically next time.

src/lemonade/tools/server/serve.py
```diff
@@ -336,6 +339,94 @@ def setup_routes(self, api_prefixes: list[str]):
             self.app.post(f"{prefix}/reranking")(self.reranking)
             self.app.post(f"{prefix}/rerank")(self.reranking)
 
+    async def upload_model(
+        self,
+        model_name: str = Form(...),
+        recipe: str = Form(...),
+        reasoning: bool = Form(False),
+        vision: bool = Form(False),
+        mmproj: str = Form(None),
```
jeremyfowers, line 376: Does the use of `Form` here mean that the feature only works from the web ui, and not from the CLI? Or can I do `lemonade-server pull user.local_model --checkpoint C:\path\to\model`?

src/lemonade/tools/server/serve.py
```diff
@@ -336,6 +339,94 @@ def setup_routes(self, api_prefixes: list[str]):
...
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="At least one .gguf file is required for llamacpp"
+                )
+
+            # Create a unique directory for the model in the HF cache            
+            model_name_clean = model_name.replace("user.", "")
+            repo_cache_name = f"{model_name_clean.replace('/', '--')}"
+            snapshot_path = os.path.join(HF_HUB_CACHE, f"models--{repo_cache_name}")
+            os.makedirs(snapshot_path, exist_ok=True)
```
jeremyfowers, line 427: `exist_ok=True` what is the expected behavior in the event of a collision? Like if I put a `model_name` that happened to already match a directory in the HF_HUB_CACHE?

jeremyfowers: Should this raise an http exception instead of doing exist_ok=True?

src/lemonade/tools/server/serve.py
```diff
@@ -336,6 +339,94 @@ def setup_routes(self, api_prefixes: list[str]):
             self.app.post(f"{prefix}/reranking")(self.reranking)
             self.app.post(f"{prefix}/rerank")(self.reranking)
 
+    async def upload_model(
```
jeremyfowers: UI/UX question: is "upload" the right name for a feature that takes a local file on disk and copies it to another location on disk?

Would you consider renaming it to "add local model" or similar to avoid confusion?

I don't want people to think we're uploading their model to the internet.

src/lemonade/tools/server/serve.py
```diff
@@ -1597,6 +1688,62 @@ async def load_llm(self, config: LoadConfig):
                 # Get additional properties from the model registry
                 config_to_use = LoadConfig(**supported_models[config.model_name])
 
+
+            if config_to_use.model_name.startswith("user."):
```
jeremyfowers: Is there any way you could prepare the model folder or user_models.json at add-time so that all of this work isn't needed at load-time?

We try to avoid putting code on the critical path of loading and running models whenever possible.

src/lemonade/tools/server/static/js/models.js
```diff
@@ -427,6 +427,19 @@ function createModelItem(modelId, modelData, container) {
 
 // Install model
 async function installModel(modelId) {
+    if (modelId.startsWith('user.')) {
```
jeremyfowers: `modelId.startsWith('user.')` is possible for non-local Add a Model as well. Is this going to create strange behavior for Add a Model that triggers a long download from Hugging Face?

src/lemonade/tools/server/static/js/models.js
```diff
@@ -505,7 +518,11 @@ async function deleteModel(modelId) {
             headers: { 'Content-Type': 'application/json' },
             body: JSON.stringify({ model_name: modelId })
         });
-        
+        installedModels.delete(modelId);
+        // Remove custom models from SERVER_MODELS to prevent them from reappearing without having to do a manual refresh
+        if (modelId.startsWith('user.')) {
```
jeremyfowers, line 533: Same question about `user.` models downloaded from huggingface.

src/lemonade_server/model_manager.py
```diff
@@ -104,6 +105,11 @@ def downloaded_models(self) -> dict:
             else:
                 # Handle other models
                 checkpoint = model_info["checkpoint"]
+                if model.startswith('user.'):
```
jeremyfowers: Same comment about `user.`

src/lemonade_server/model_manager.py
```diff
@@ -147,6 +153,76 @@ def downloaded_models(self) -> dict:
...
+    snapshot_path: str = ""
+    ):
+        if not model_name.startswith("user."):
+            raise ValueError("Model name must start with 'user.'")
+
+        model_name_clean = model_name[5:]
+
+        valid_recipes = ["llamacpp", "oga-npu", "oga-hybrid", "oga-cpu"]
+        if recipe not in valid_recipes:
+            raise ValueError(f"Invalid recipe. Must be one of: {', '.join(valid_recipes)}")
```
jeremyfowers: You have these checks in the endpoint as well, right? Is it possible to consolidate?

src/lemonade_server/model_manager.py
```diff
@@ -147,6 +153,76 @@ def downloaded_models(self) -> dict:
...
+            new_user_model["mmproj"] = mmproj
+
+        # Load existing user models
+        user_models = {}
+        if os.path.exists(USER_MODELS_FILE):
+            with open(USER_MODELS_FILE, "r", encoding="utf-8") as file:
+                user_models = json.load(file)
+
+        # Check for conflicts
+        if model_name_clean in user_models:
```
jeremyfowers, line 217: Would it be simpler to just do:

```python
if model_name_clean in user_models:
    raise ValueError(
                    f"Model {model_name} is already registered. "
                    f"Please use a different model name or delete the existing model."
                )
```

The existing logic seems tuned for allowing re-registering a model if it already exists with the exact same properties, but I don't know why we need to allow that?

src/lemonade/tools/llamacpp/utils.py
meghsat: Added it as an extra layer of filtering for local models but it's not needed as long as the local folder path is valid. Removed it in the latest commit.

src/lemonade/tools/llamacpp/utils.py
meghsat: Looks like a duplicate paste, removed it.

src/lemonade/tools/server/serve.py
meghsat: Moved all the HF_HUB_CACHE imports to functions

src/lemonade/tools/llamacpp/utils.py
meghsat: Enabled the Black formatter and also joined the lemonade-sdk org

src/lemonade/tools/server/serve.py
meghsat: Yes. The current cli.py/pull logic sends JSON data and it doesn't accept files. So, the current Form logic fails. I will work on the terminal implementation in the next version.

src/lemonade/tools/server/serve.py
meghsat: Added a check on form submit to ensure the model_name doesn't already exist in the window.SERVER_MODELS. Additionally, added an extra check at HTTPExcep inside the add_local_model function to check for existing model names. 
image

src/lemonade/tools/server/serve.py
meghsat: Makes sense, renamed it to 'add_local_model'

src/lemonade/tools/server/serve.py
meghsat: You're right, the current logic checks the paths every time the model is loaded. Moved all of this logic into the load time function: add_local_model. However, if it is a HF direct download, the path is updated during the download time[utils.py/download_gguf for gguf models] and for OGA models, the user_models.json is updated in the download_models/new_registration_model_config.

src/lemonade/tools/server/static/js/models.js
meghsat: This check is not needed and the logic inside the condition is available out side too, so removed it.

src/lemonade/tools/server/static/js/models.js
meghsat: This check is needed here and it is relevant for any user added model. This way we don't remove the built-in models from the list.

src/lemonade_server/model_manager.py
meghsat: Similar to previous comment, this check works for any user added model could be through folder upload or HF download. If it's not a custom model then the previous flow is triggered.

src/lemonade_server/model_manager.py
meghsat: These are duplicate checks, removed them.

src/lemonade_server/model_manager.py
meghsat: Now that we have checks to ensure a model_name is unique and can't be overwritten adding the ValueError here makes sense. Updated it.

src/lemonade/tools/server/serve.py
```diff
@@ -1618,6 +1792,35 @@ async def load_llm(self, config: LoadConfig):
                 # Get additional properties from the model registry
                 config_to_use = LoadConfig(**supported_models[config.model_name])
 
+            # For user models, convert the relative checkpoint path to absolute path
```
jeremyfowers: I'm confused by this code block because it appears to apply to all `user.` models, not only local models.

Would it simplify the code a lot if you added a `local` label to the models, and then tested for that instead of testing for all of these other conditions?

src/lemonade/tools/llamacpp/utils.py
```diff
@@ -656,16 +654,71 @@ def identify_gguf_models(
...
+def download_gguf(
+    config_checkpoint: str, config_mmproj=None, do_not_upgrade: bool = False
+) -> dict:
     """
     Downloads the GGUF file for the given model configuration.
 
     For sharded models, if the variant points to a folder (e.g. Q4_0), all files in that folder
     will be downloaded but only the first file will be returned for loading.
     """
+    if os.path.exists(config_checkpoint):
```
jeremyfowers, line 734: I'm concerned that all of this code is being added to the main codepath for downloading a GGUF, when it is actually all specific to registering a local model. Is it possible to keep this new logic more separated to reduce the complexity of the heavily-used code path for downloading GGUFs from the internet?

src/lemonade/tools/server/serve.py
meghsat: My understanding is all the user downloaded/ uploaded models have a `user.` tag, so this condition would pass for all of them.

jeremyfowers: Right, but downloaded `user.` models already worked before this PR. So I'm concerned that there's a new code path that activates for downloaded `user.` models. Can you make the new code completely specific to local models?

src/lemonade/tools/llamacpp/utils.py
meghsat: Moved all of this logic into a separate function: utils.py/resolve_local_gguf_model

src/lemonade/tools/server/serve.py
meghsat: Added an extra label for locally uploaded models.

PR #457, opened 2025-10-10
Enable reliable Explorer context-menu launch for lemonade-server.bat

Tetramatrix:
**Summary**

Default lemonade-server.bat to serve when no args are provided so launching from the Explorer context menu starts the server with the tray icon. Argument handling is preserved when args are present.

**Changes**

- Use an internal CLI_ARGS to default to serve on empty invocation.
- Set TRAY=1 when serve is detected to show the tray icon.

**Behavior**

- Right-click → Open from Explorer now starts the server with the tray icon.
- CLI usage with explicit arguments remains unchanged.

### DIFF
--- a/installer/lemonade-gui.bat
+++ b/installer/lemonade-gui.bat
@@ -1,0 +1,14 @@
+@echo off
+setlocal enabledelayedexpansion
+
+rem This is a simple wrapper to launch the Lemonade Server in GUI mode.
+rem It calls the main batch script with the "serve" command, ensuring the tray icon appears.
+
+REM Change to parent directory where conda env and bin folders are located
+pushd "%~dp0.."
+
+REM Run the Python CLI script, passing filtered arguments
+call "%CD%\python\Scripts\lemonade-server-dev" serve
+set SERVER_ERRORLEVEL=%ERRORLEVEL%
+popd
+
--- a/installer/lemonade-server.bat
+++ b/installer/lemonade-server.bat
@@ -9,21 +9,18 @@ for /f "tokens=1-4 delims=:.," %%a in ("!time!") do (
 REM Use temp directory for the lock file
 set "LOCK_FILE=%TEMP%\lemonade_server.lock"
 
-REM Default to "serve" if no arguments are provided
-set "CLI_ARGS=%*"
-if "%CLI_ARGS%"=="" set "CLI_ARGS=serve"
+set "ARGS=%*"
 
-REM Show a notification and run the server in tray mode.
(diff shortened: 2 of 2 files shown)


danielholanda:
Hi @Tetramatrix, thanks for your PR.

What your changes are effectively doing is causing the `lemonade-server` CLI command to call `lemonade-server serve` when no args are provided instead of showing the `help` menu. This is not ideal, as many of our users do use the CLI and not showing the help menu could be confusing for this audience.

I can also confirm that the server launches correctly (with the tray) when launched from `Windows Search` (earch bar at the bottom of Windows) and from Explorer context menu (the right-click menu that appears when you click on files).

Is there any way in which you can still achieve your goal without changing the default behavior of our CLI? Maybe ask your app to open the actual `lemonade-server.lnk` file or the `vbs` instead of the underlying bat file?

Tetramatrix:
Hi @danielholanda 

Thanks for the feedback!

Just to clarify — my PR wasn’t about the file context menu (right-clicking the .bat itself), but rather the desktop context menu (as shown in the attached screenshot). When launching from there, it seems the .vbs doesn’t get executed at all, while the .bat works reliably.

I didn’t realize that running lemonade-server without arguments shows the help menu — that makes sense now. My intention was only to make the server start correctly when triggered from the desktop context menu, not to alter the default CLI behavior for users launching it manually.

Would you be open to a solution that detects when the process is launched from that specific context (desktop context menu) and delegates to the normal serve behavior, while keeping the CLI unchanged in other cases?

i8wOWqnVSJ

danielholanda:
Hi Tetramatrix,

Thanks for clarifying this! As long as the solution is fairly lean and it doesn't cause issues elsewhere, then it makes sense to me!

jeremyfowers:
@danielholanda can you clarify if you are approving here, or if changes are required?

Tetramatrix:
Hi @danielholanda,

In my opinion, the simplest approach would be to create an additional GUI start batch file to prevent issues elsewhere.

Cheers

danielholanda:
@Tetramatrix In the interest of keeping the codebase lean and avoiding additional init scripts would you mind keeping that as a private solution for now, until we see other people needing it? sorry about any confusion from my prior message

Tetramatrix:
@danielholanda Thanks for the clarity—totally fine to keep this external for now; I’ll maintain a separate script and we can revisit if more users ask for it.

jeremyfowers:
Thanks all for the discussion. Closing this one for now, but feel free to re-open later.

PR #465, opened 2025-10-14
Add support for Granite 4.0 and LFM2 moes (and rev llamacpp versions)

jeremyfowers:
New llamacpp is required to support LFM2 moe

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -21,10 +21,6 @@ limit-inference-results = 100
 # Pickle collected data for later comparisons.
 persistent = no
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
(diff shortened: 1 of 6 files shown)


danielholanda:
Additional context: Llamacpp doesn't show any errors, it just exists immediately. This suggest an out-of-memory (OOM) issue. I’ve confirmed that this doesn’t occur on the main branch.

jeremyfowers:
Converting to draft until we have some new llamacpp builds to try

jeremyfowers:
Just tested https://github.com/lemonade-sdk/llamacpp-rocm/releases/tag/b1079 and that one doesn't work either.

PR #476, opened 2025-10-16
Fix major testing issues in server.py

jeremyfowers:
`python test/server.py` was not working on local windows PCs because it would hang forever soon after starting.

It was also rather hard to tell what was going on.

This PR makes the following improvements:

1. Fixes the hang! Turns out the problem was spamming v0/system-info followed by v1/system-info in test_000.
    - Fixed this by using HEAD requests instead of GET requests in test_000. HEAD doesn't actually run the endpoint, it just checks that it is registered. This is much more appropriate for a test called "test_000_endpoints_available"
    - Made sure this still detects endpoint registration by commenting on the pull endpoint in serve.py and confirming that the test failed.
2. Added a timeout of all the `requests.get()` calls so that we don't have hanging tests anymore.
3. Changed the test settings in test/server_base.py:
    - Now "fails fast", so as you as you fail a test it will exit and give you the stack trace. That makes it massively easier to figure out what failed.
    - Increased print verbosity and removed buffering so that we can get all the info needed to debug. For example, the name of the test now prints right before the log output for that test.

### DIFF
--- a/test/server.py
+++ b/test/server.py
@@ -13,7 +13,6 @@ If you get the `ImportError: cannot import name 'TypeIs' from 'typing_extensions
 """
 
 import asyncio
-import numpy as np
 import requests
 from openai import NotFoundError
 
@@ -48,7 +47,7 @@ class Testing(ServerTestingBase):
     """Main testing class that inherits shared functionality from ServerTestingBase."""
 
     def test_000_endpoints_available(self):
-        # List of endpoints to check
+        # List of endpoints to check for registration
         valid_endpoints = [
             "chat/completions",
             "completions",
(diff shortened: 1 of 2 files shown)


test/server.py
```diff
@@ -67,14 +66,15 @@ def test_000_endpoints_available(self):
 
         # Ensure that we get a 404 error when the endpoint is not registered
         url = f"http://localhost:{PORT}/api/v0/nonexistent"
-        response = requests.get(url)
+        response = requests.head(url, timeout=60)
```
danielholanda, line 69: [non-blocking]
A 5s timeout is likely more than enough here.

PR #479, opened 2025-10-21
Fix server spec bugs

jeremyfowers:
When planning lemon.cpp, we noticed that server_spec.md has some bugs.

For example, a lot of the endpoints are actually POST and not GET.

Example:

```
Jeremy@stx-halo MINGW64 ~
$ curl http://localhost:8000/api/v1/unload
{"detail":"Method Not Allowed"}
```

```
Jeremy@stx-halo MINGW64 ~
$ curl -X POST http://localhost:8000/api/v1/unload
{"status":"success","message":"Unloaded model"}


danielholanda (approved):
Thanks for addressing those!

PR #486, opened 2025-10-27
Fix thought process content display

ramkrishna2910:
image

This PR makes the following changes:
- The streaming chat completions endpoint now separates reasoning tokens from regular response tokens. State tracking detects the </think> boundary to transition from reasoning phase to normal content.
- Chat completion chunks now use separate reasoning_content and content fields instead of mixing both in a single content field.


danielholanda (commented):
Can you provide an example where the currently logic on main does not work? The version on main works well on me for Qwen3-0.6B

Issue #482, opened 2025-10-24
lemon.cpp todo list


jeremyfowers:
- [x] max_tokens is being ignored
- [ ] max_completion_tokens is being ignored on completions endpoint
- [x] add a tray application
- [x] system-info
- [x] pass server.py
- [ ] send the log to the ws endpoint for the web ui
- [x] enable and test rocm
- [x] make sure gguf model download works the same as the python (handles all the corner cases)
- [ ] `list` command is really slow
- [ ] non-serve/run CLI commands should not show the server log
- [x] ONNX model pull puts a ton of progress spam into the log
- [ ] "[stdout] [ProcessManager] Process started successfully, PID: 55616" is being double printed
- [ ] Weird text on successful model download "[ModelManager] âœ“ Downloaded: model.onnx"
- [x] Two lines of "FLM" in the model manager printout when models endpoint is queried.
- [ ] system-info caches the NPU driver version
- [ ] split serve.py into OGA vs. generic (e.g., system-info) tests. This is needed for Linux support.
- [ ] FLM model delete not implemented
- [ ] Closing the tray app should close the log window too... or at least the log window being open shouldnt prevent uninstall. Also the log window opens really slow.
- [ ] The webui stop button crashes lemonade-router.exe during a ryzenai-serve and llamacpp chat/completion requests
- [ ] Support for "local model uploads" introduced in #456 
- [ ] Support Linux
- [ ] Support macOS

danielholanda:
Will keep updating this here as I review
- [x] Tray has name "Lemonade Server Beta - Tray Application" instead of "Lemonade Server Beta"
- [x] Tray has big "info" sign. Should remove or replace with Lemon icon to be clean
- [x] Tray menu does not show up while loading a model (try to load a large model like gpt-oss-20)
- [x] Debug logs do not show llamacpp logs
- [x] Multiple instances of Lemonade can be opened at once, causing confusion on the tray menu. This should only be allowed under very specific conditions and definitely not when using the tray. Recommendation is to mimic the behavior of main for now.
  <img width="112" height="80" alt="Image" src="https://github.com/user-attachments/assets/e3d58305-1181-40c6-968a-550cf97dd54b" />
- [x] Both right AND left clicks should trigger the tray. Currently, only right-clicking works.
- [x] `lemonade-server-beta` is the same as `lemonade-server serve`. Stop, list, and other commands are missing.

ramkrishna2910:
Will add comments here:

[...]

PR #487, opened 2025-10-28
lemon.cpp

jeremyfowers:
## Overview

This PR introduces a C++ version of Lemonade Server. It is not fully complete, but the core feature set is expected to work nicely.

**Please see the Limitations section for info about what doesn't work yet.**

The implementation is extensively documented in:
- Tray app and server-router: https://github.com/lemonade-sdk/lemonade/blob/lemoncpp/src/cpp/README.md
- RAI OGA serving: https://github.com/lemonade-sdk/lemonade/blob/lemoncpp/src/ryzenai-serve/README.md

The purpose of this PR is to merge the existing progress to enable more agility in future development. The C++ server is not intended for production use or to replace the Python at this time.

## Limitations

There is a running list of known issues here: https://github.com/lemonade-sdk/lemonade/issues/482

The most important limitations to be aware of are:
1. Linux and macOS support are not yet implemented.
2. Ryzen AI 1.6 is supported, but you must either:
    a. Build from source (see doc links above), or
    b. add a lemonade repo PAT to your user environment variables with the key `GITHUB_TOKEN`

> Note: limitation 2b will be relaxed when we have a formal release artifact that contains ryzenai-serve.zip. Until then, a PAT is required to access CI artifacts.

New limitations can be noted as comments on this PR, comments on the issue linked above, or new issues with the cpp label.

## Guidance for trying it out

1. The new `C++ Server Build, Test, and Release` action extensively tests the C++ server. Please review the test logs to get acclimated to the capabilities.
2. That action also produces a Lemonade_Server_Installer_beta.exe. This is the easiest way to get started. Just see the note above in Limitations about the temporary PAT requirement.
3. There are instructions to build and run from source in the docs. Please also try this out so that you are comfortable to do your own development in the future! It's easy :)

In terms of usage... just replace the python server with the C++ one for whatever you are currently doing. The C++ is interface-compatible and should be a drop in replacement for Open WebUI and any other app.

> note: the server binary on your PATH is `lemonade-server-beta.exe`, so make sure to take that into account for any app that is looking for `lemonade-server.exe`

## Cool new stuff

Yes that's right, there are some improvements over the python here you can try out!

1. The installer registers Lemonade Server Beta with your Start Menu and Add/Remove Programs menu (for uninstallation).
2. Better multi-threading: clients can run `health` and `models` endpoints while completions and loading are running.
3. Cleaner architecture: the tray app, server-router, and RAI server are three separate, modular executables now.

## Changes to existing files

This PR tries to absolutely minimize changes to existing files. However, there are a few:
1. Fixes some bugs with the web ui where the web ui was expecting a single-threaded server (models endpoint blocks on load/pull endpoint).
2. Fixes bugs in server_base.py that could hang the test if stderr/stdout came from the server process in just the right way.
3. test_011_test_simultaneous_load_requests now allows the models to be loaded in either order, since we are truly multithreaded now.
4. test_012_test_load_by_name uses a GGUF model instead of an HF model. We haven't been supporting the HF recipe anyways.


danielholanda:
Will add all minor issues to https://github.com/lemonade-sdk/lemonade/issues/482

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -0,0 +1,985 @@
...
+          ls -lh ryzenai-serve.zip
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v2
+        with:
+          files: |
+            Lemonade_Server_Installer_beta.exe
+            ryzenai-serve.zip
+          fail_on_unmatched_files: true
+          generate_release_notes: true
```
danielholanda, line 984: We should be very careful here to avoid unintentional releases

src/lemonade/tools/server/static/js/shared.js
```diff
@@ -358,6 +368,11 @@ async function loadModelStandardized(modelId, options = {}) {
         // Always show error banner to ensure user sees the error
         showErrorBanner('Failed to load model: ' + error.message);
         
+        // Remove from active operations on error too
+        if (window.activeOperations) {
+            window.activeOperations.delete(modelId);
+        }
```
danielholanda, line 374: I would prefer if we avoided touching those JS files in this PR, unless you’re confident they don’t interfere with our legacy Python lemonade functionality.

jeremyfowers: I would have too, but this concurrency bug is a killer for testing out the multi-threaded C++ server.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: I removed the `generate_release_notes: true`, that shouldnt have been there.

Other than that, this job is identical to our code that releases the python installer. It just appends some files to an existing release. There's no new surface for accidents.

ramkrishna2910 (approved):
Added a couple of minor comments to #482. Looks great and works well!

PR #492, opened 2025-10-30
Fix all known C++ tray, installer, and cli bugs

jeremyfowers:
# C++ User Experience Improvements

This PR polishes the user experience by addressing **all** UI/UX related feedback across the installer, tray, start menu, CLI etc. in https://github.com/lemonade-sdk/lemonade/issues/482

## Code/Arch changes

See the changes to src/cpp/README.md for implementation details!

## Fixes

The fixes in this PR include, but are not limited to, the following feedbacks:
- list command is really slow
- non-serve/run CLI commands should not show the server log
- "[stdout] [ProcessManager] Process started successfully, PID: 55616" is being double printed
- Weird text on successful model download "[ModelManager] âœ“ Downloaded: model.onnx"
- Closing the tray app should close the log window too... or at least the log window being open shouldnt prevent uninstall. Also the log window opens really slow.
- Tray has name "Lemonade Server Beta - Tray Application" instead of "Lemonade Server Beta"
- Tray has big "info" sign. Should remove or replace with Lemon icon to be clean
- Tray menu does not show up while loading a model (try to load a large model like gpt-oss-20)
- Debug logs do not show llamacpp logs
- Multiple instances of Lemonade can be opened at once, causing confusion on the tray menu. This should only be allowed under very specific conditions and definitely not when using the tray. Recommendation is to mimic the behavior of main for now.
- Both right AND left clicks should trigger the tray. Currently, only right-clicking works.
- lemonade-server-beta is the same as lemonade-server serve. Stop, list, and other commands are missing.
- The c++ installer does not close the currently open lemonade server. The python installer used to do this.
- Able to open multiple instances of the server concurrently.
- cpp release job should not generate release notes
- lemonade-router.exe --help shows the help, but also says Unknown command:  and gives a non-zero exit code. lemonade-router.exe just says unknown command.
- run command doesnt work

## Bottom Line

With this changeset, every aspect of the Installer, Tray, and CLI UI/UX should be the same or better than the Python!

### DIFF
--- a/src/cpp/tray/tray_app.cpp
+++ b/src/cpp/tray/tray_app.cpp
@@ -33,6 +33,61 @@ namespace lemon_tray {
         std::cout << "DEBUG: " << msg << std::endl; \
     }
 
+#ifdef _WIN32
+// Helper function to show a simple Windows notification without tray
+static void show_simple_notification(const std::string& title, const std::string& message) {
+    // Convert UTF-8 to wide string
+    auto utf8_to_wstring = [](const std::string& str) -> std::wstring {
+        if (str.empty()) return std::wstring();
+        int size_needed = MultiByteToWideChar(CP_UTF8, 0, str.c_str(), -1, nullptr, 0);
+        std::wstring result(size_needed, 0);
+        MultiByteToWideChar(CP_UTF8, 0, str.c_str(), -1, &result[0], size_needed);
+        if (!result.empty() && result.back() == L'\0') {
+            result.pop_back();
+        }
+        return result;
+    };
+    
+    // Create a temporary window class and window for the notification
+    WNDCLASSW wc = {};
+    wc.lpfnWndProc = DefWindowProcW;
+    wc.hInstance = GetModuleHandle(nullptr);
+    wc.lpszClassName = L"LemonadeNotifyClass";
+    RegisterClassW(&wc);
+    
+    HWND hwnd = CreateWindowW(L"LemonadeNotifyClass", L"", 0, 0, 0, 0, 0, nullptr, nullptr, wc.hInstance, nullptr);
+    
+    if (hwnd) {
+        NOTIFYICONDATAW nid = {};
+        nid.cbSize = sizeof(nid);
+        nid.hWnd = hwnd;
+        nid.uID = 1;
+        nid.uFlags = NIF_INFO | NIF_ICON;
+        nid.dwInfoFlags = NIIF_INFO;
+        
+        // Use default icon
+        nid.hIcon = LoadIcon(nullptr, IDI_INFORMATION);
+        
+        std::wstring title_wide = utf8_to_wstring(title);
+        std::wstring message_wide = utf8_to_wstring(message);
+        
+        wcsncpy_s(nid.szInfoTitle, title_wide.c_str(), _TRUNCATE);
+        wcsncpy_s(nid.szInfo, message_wide.c_str(), _TRUNCATE);
+        wcsncpy_s(nid.szTip, L"Lemonade Server", _TRUNCATE);
+        
+        // Add the icon and show notification
+        Shell_NotifyIconW(NIM_ADD, &nid);
+        
+        // Keep it displayed briefly then clean up
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        Shell_NotifyIconW(NIM_DELETE, &nid);
+        
+        DestroyWindow(hwnd);
+    }
+    UnregisterClassW(L"LemonadeNotifyClass", GetModuleHandle(nullptr));
+}
+#endif
+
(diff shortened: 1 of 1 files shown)


.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
               exit 1
```
danielholanda, line 115: > Multiple instances of Lemonade can be opened at once, causing confusion on the tray menu. This should only be allowed under very specific conditions and definitely not when using the tray. Recommendation is to mimic the behavior of main for now.

Glad this was fixed? But maybe show a notification saying that the server is already running?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
...
           }
```
danielholanda, line 116: > lemonade-server-beta

I personally find the examples at the bottom too verbose and would simply remove those. CLI is already clear enough.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
...
```
danielholanda, line 117: > lemonade-server-beta --server-binary PATH

What is the use of this? Would recommend simplifying and removing this if 99% of users will never touch it. Simplicity first.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
...
           
+          if (-not (Test-Path "Release\lemonade-tray.exe")) {
```
danielholanda, line 118: > lemonade-server-beta --help

To make it more modular and cleaner, I recommend only showing the server options if people type `lemonade-server-beta serve --help`

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
...
+              Write-Host "ERROR: lemonade-tray.exe not found!" -ForegroundColor Red
```
danielholanda, line 119: Lemonade server run should work regardless of whether the server is already running if we want to maintain the same behavior as before.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
...
+              exit 1
```
danielholanda, line 120: HEALTH!!!!
I get a health check every 0.1s while loading a model. This is very disruptive when looking at the logs. Maybe tune that down a bit?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
...
+          }
```
danielholanda, line 121: Interesting behavior:
When I start lemonade using the cli, I see a ton of "debug" commands that do not show up if I go to `tray -> show logs`. Looks like we are not forwarding all logs there?

danielholanda (approved):
Amazing improvements! Please either address the added comments or add those to our list of know nits to fix.

Btw, I love that both `stop` and `status` show exactly the same text as the old version! great job there.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: There are both tray logs and server logs, but the Show Logs will only show the server logs. Tray logs should be pretty irrelevant to users.

jeremyfowers: I'll just suppress the tray logs

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: fixed (will push later)

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: fixed (will push later)

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: fixed (will push later)

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: fixed (will push later)

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: fixed (will push later)

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: fixed, but had to remake the PR

PR #525, opened 2025-11-05
Added a missing --device support for llamacpp

meghsat:
The command ```lemonade -i <model-path> llamacpp-load --device cpu llamacpp-bench --iterations 10 --prompts 64``` is targeting the iGPU even though the CPU is specified. 
**Fix**: Add the ngl_value linked to --device in the cmd variable.


amd-pworfolk (approved):
Thanks for finding and fixing this bug!  I've verified it.

PR #531, opened 2025-11-06
Add enable_thinking toggle to web UI model settings

Kritik-07:
Adds a checkbox in the Model Settings menu to control the `enable_thinking` parameter for API requests.

## Testing
- Checkbox state persists across page refreshes
- "Reset to Defaults" button resets checkbox to enabled (default state)
- `enable_thinking` parameter is correctly sent in API requests
- Tested with Qwen3 model - `/no_think` prefix is added when checkbox is disabled
- Help text displays correctly below checkbox


Kritik-07:
Thank you @jeremyfowers, Glad it works well. It was great adding both the backend and frontend for this feature. Thanks for the guidance and quick reviews!

PR #549, opened 2025-11-11
C++: General Availability

jeremyfowers:
This PR has a lot of file changes, but the overview is simple:
1. Remove the Python-based NSIS installer at //installer, and the associated Python installer tests
2. Remove the "beta" tag from everywhere in the C++
3. This has the effect that the C++ is now responsible for creating Lemoande_Server_Installer.exe and lemonade-server.exe, instead of the Python
4. Rev python and C++ version numbers to `9.0.2`
5. Update the installation instructions html
6. Remove `lemonade-server-dev` references from most documentation (it's always `lemonade-server` now!)
7. Add a deprecation notice that prints when using the python server

Additional C++ changes:
1. Changes default `--host` from `127.0.0.1` to `localhost` (closes https://github.com/lemonade-sdk/lemonade/issues/554)

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -6,6 +6,7 @@
 #include <lemon/system_info.h>
 #include <filesystem>
 #include <iostream>
+#include <fstream>
 #include <algorithm>
 #include <cstdlib>
 #include <sstream>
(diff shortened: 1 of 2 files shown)


danielholanda:
Not passing tests yet. Please take a look

PR #568, opened 2025-11-14
Improve the user experience for embeddings

jeremyfowers:
Requested by @kovtcharov 

## Embedding-Specific Improvements

1. Enable users to register custom embedding models via the web ui and `pull` CLI command (by adding an embedding checkbox/arg)
    - Also `reranking`, for good measure
2. Help users understand how to use embeddings by documenting the endpoint in server_spec.md
    - Also `reranking`, for good measure
3. Add Qwen3-*B-Embedding-GGUF models to server_models.json
4. Only apply the `--ubatch-size 8192` arg on nomic models (since it crashes Qwen models)

## Additional Improvements

1. Add a `lemonade-server pull --help` help page that defines the args
2. Document the arguments to `pull` on lemonade-server-cli.md
3. Fix a bug where model sizes weren't being shared with the web ui

### DIFF
--- a/src/cpp/tray/tray_app.cpp
+++ b/src/cpp/tray/tray_app.cpp
@@ -627,11 +627,7 @@ void TrayApp::print_pull_help() {
     std::cout << "  For custom models, use the registration options below.\n\n";
     std::cout << "Registration Options (for custom models):\n";
     std::cout << "  --checkpoint CHECKPOINT  Hugging Face checkpoint (format: org/model:variant)\n";
-    std::cout << "                           Required for registering new models.\n";
-    std::cout << "                           Examples:\n";
-    std::cout << "                             - unsloth/Qwen3-8B-GGUF:Q4_0\n";
-    std::cout << "                             - microsoft/Phi-3-mini-4k-instruct-onnx\n\n";
-    std::cout << "  --recipe RECIPE          Inference recipe to use. Required for new models.\n";
+    std::cout << "  --recipe RECIPE          Inference recipe to use\n";
     std::cout << "                           Options: llamacpp, flm, oga-cpu, oga-hybrid, oga-npu\n\n";
     std::cout << "  --reasoning              Mark model as a reasoning model (e.g., DeepSeek-R1)\n";
     std::cout << "                           Adds 'reasoning' label to model metadata.\n\n";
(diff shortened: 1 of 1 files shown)


src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -607,7 +607,13 @@ void LlamaCppServer::load(const std::string& model_name,
         // ubatch-size: physical batch size (tokens processed in a single forward pass)
         push_arg(args, reserved_flags, "--embeddings");
         push_arg(args, reserved_flags, "--batch-size", std::to_string(EMBEDDING_BATCH_SIZE));
-        push_arg(args, reserved_flags, "--ubatch-size", std::to_string(EMBEDDING_UBATCH_SIZE));
+        
+        // Only set ubatch-size for nomic models (case insensitive check)
```
danielholanda, line 611: Are nomic models special and require this? Or are Qwen models special and can't handle it?

danielholanda: Just curious, but this here is good for now

src/cpp/tray/tray_app.cpp
```diff
@@ -612,6 +617,59 @@ void TrayApp::print_version() {
...
+    std::cout << "    --mmproj mmproj-model-f16.gguf\n\n";
+    std::cout << "  # Register and install an embedding model\n";
+    std::cout << "  lemonade-server pull user.nomic-embed \\\n";
+    std::cout << "    --checkpoint nomic-ai/nomic-embed-text-v1-GGUF:Q4_K_S \\\n";
+    std::cout << "    --recipe llamacpp \\\n";
+    std::cout << "    --embedding\n\n";
+    std::cout << "Notes:\n";
+    std::cout << "  - Custom model names must use the 'user.' namespace (e.g., user.MyModel)\n";
+    std::cout << "  - GGUF models require a variant in the checkpoint (after the colon)\n";
+    std::cout << "  - See available models: lemonade-server list\n";
```
danielholanda: Adding excessive documentation on a CLI like this is not standard and will require the user to scroll up to see information that actually matters. Recommend removing this chunk to avoid being excessively verbose.

src/cpp/tray/tray_app.cpp
```diff
@@ -612,6 +617,59 @@ void TrayApp::print_version() {
...
+    std::cout << "Description:\n";
+    std::cout << "  Downloads a model from the Lemonade Server registry or Hugging Face.\n";
+    std::cout << "  For registered models, only the model name is required.\n";
+    std::cout << "  For custom models, use the registration options below.\n\n";
+    std::cout << "Registration Options (for custom models):\n";
+    std::cout << "  --checkpoint CHECKPOINT  Hugging Face checkpoint (format: org/model:variant)\n";
+    std::cout << "                           Required for registering new models.\n";
+    std::cout << "                           Examples:\n";
+    std::cout << "                             - unsloth/Qwen3-8B-GGUF:Q4_0\n";
+    std::cout << "                             - microsoft/Phi-3-mini-4k-instruct-onnx\n\n";
```
danielholanda: So verbose. Isn't something like this enough?
```suggestion
    std::cout << "  --checkpoint CHECKPOINT  Hugging Face checkpoint (format: org/model:variant)\n";
    std::cout << "                           Example: unsloth/Qwen3-8B-GGUF:Q4_0.\n";
```

danielholanda (approved):
Looks good. Added a few minor suggestions

src/cpp/server/backends/llamacpp_server.cpp
jeremyfowers: @ramkrishna2910 put that arg in there because Morphik needed it

But it crashes Qwen models

This is a compromise until we understand well enough to put a nice solution

kovtcharov (approved):

lostdisc:
Is there more info on why ONNX/OGA does not support embeddings output?  What would need to change to make it possible?

For context, Z-Image uses Qwen 3-4B as its text encoder, but Qwen is too big to keep in VRAM across image-gen runs, and rather slow to swap in/out or run on CPU, so I was looking for a way to run it on NPU.  ComfyUI doesn't have built-in support for AMD NPUs or ONNX, but I thought of making it pass the prompt to Lemonade to run text encode on NPU, then retrieve those results (embeddings?) for next steps.

jeremyfowers:
> Is there more info on why ONNX/OGA does not support embeddings output? What would need to change to make it possible?

@lostdisc it is a matter of prioritization for the OGA engineering team.

FastFlowLM has some support for embeddings on NPU: https://fastflowlm.com/docs/models/embeddinggemma/

That is not enabled in Lemonade yet but you could check it out.

lostdisc:
Neat, thanks; I didn't know FastFlowLM was targeted at AMD NPUs as well.

One thing I don't get yet is that there are specialized embedding versions of models, e.g. [Qwen3-Embedding-4B](https://huggingface.co/Qwen/Qwen3-Embedding-4B), but Z-Image is using regular Qwen3-4B as its text encoder.  When I looked at Lemonade's [server spec](https://lemonade-server.ai/docs/server/server_spec/), I initially thought calling `POST /api/v1/embeddings` on regular Qwen3-4B must do the same thing as text-encoding, but the server spec also mentions "the embeddings endpoint requires embedding-specific models" a couple of times, so it may not be as straightforward as I thought.

lostdisc:
Or it could be that the Qwen3-4B version that Z-Image uses is modified to spit out embeddings; [their](https://huggingface.co/Tongyi-MAI/Z-Image-Turbo/tree/main/text_encoder) third safetensors file has a different hash from the [stock](https://huggingface.co/Qwen/Qwen3-4B/tree/main) version's.

PR #590, opened 2025-11-20
Infinite timeout for inference requests

### LINKED ISSUE #572: Lemonade 9.0.#: Long prompt responses timing out
Tested via WebUI and with POML VS Code extension.

Scenario:
When provided with a long prompt, LLMs running in Lemonade Server will begin providing response output, then cease output at random. This is typically mid-sentence.

Rerunning the prompt does **not** generate a complete response, and will end in the same error.

Attached is pretty much the only error I could find in the log:
[lemonade-9.0.2-error.txt](https://github.com/user-attachments/files/23555196/lemonade-9.0.2-error.txt)

jeremyfowers:
Fixes #572 

Turns out we had a 300 second timeout for inference requests. This PR change it to infinite, and also introduces a new centralized constant `INFERENCE_TIMEOUT_SECONDS` in case anyone wants to change this value later.

fyi @kovtcharov that this bug exists in the 9.0.3 release

### DIFF
--- a/src/cpp/include/lemon/streaming_proxy.h
+++ b/src/cpp/include/lemon/streaming_proxy.h
@@ -37,7 +37,8 @@ public:
         const std::string& backend_url,
         const std::string& request_body,
         httplib::DataSink& sink,
-        std::function<void(const TelemetryData&)> on_complete = nullptr
+        std::function<void(const TelemetryData&)> on_complete = nullptr,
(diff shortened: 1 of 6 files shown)


kovtcharov:
A nice to have would be able to configure the timeout from the client via the server api.

kovtcharov (approved):

PR #616, opened 2025-11-25
Extend Lemonade CLI to support FastFlow models

### LINKED ISSUE #485: FLM: Add support for FastFlow Models in Lemonade CLI
Extend Lemonade CLI to support FastFlow models in Lemonade CLI. This would allow easy performance and quality benchmarking.

The implementation could be similar to what we do for llama.cpp models.

amd-pworfolk:
This PR extends the lemonade CLI to support FastFlowLM models:

Sample commands are:
```
lemonade -d .\cache -i qwen3:0.6b flm-load llm-prompt -p "Why is the sky blue?" --output-tokens 32
lemonade -d .\cache -i qwen3:0.6b flm-load flm-bench -p 32 -w 1 -i 2 --output-tokens 32
```

Since FastFlowLM runs models on the NPU, I added the backend statistic so that we can easily differentiate between using OGA on the NPU vs. using FastFlow on the NPU.  This stat is included for all backends (OGA, Hugging Face, Llama.cpp, and FLM).  Sample output for FLM is:

```
> lemonade -d .\cache -i qwen3:0.6b flm-load flm-bench -p 32 -w 0 -i 1 --output-tokens 32

Building "qwen3-0.6b_2025y_11m_25d_02h_35m_57s"
    ✓ Starting FLM server and loading FLM model
    ✓ Benchmarking LLM

qwen3:0.6b:
        Build dir:      C:\Users\pworfolk\projects\2025_10_flm\cache\builds\qwen3-0.6b_2025y_11m_25d_02h_35m_57s
        Peak memory:    0.163 GB
        Status:         Successful build!
        Device:                             npu
        Backend:                            FastFlowLM
        Seconds To First Token:             0.494
        Token Generation Tokens Per Second: 56.284
        Prefill Tokens Per Second:          91.118
        Prompt Tokens:                      45
        Response Tokens:                    32
        Max Memory Used Gbyte:              4.631



Woohoo! Saved to ~\projects\2025_10_flm\cache\builds\qwen3-0.6b_2025y_11m_25d_02h_35m_57s
```

This closes #485.


src/lemonade/tools/flm/load.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        from lemonade.tools.flm.utils import (
+            install_flm,
+            check_flm_version,
+            FLMTokenizerAdapter,
+            FLMAdapter,
+        )
+
+        # Install FLM if needed
+        install_flm()
```
ramkrishna2910, line 74: I would recommend against auto upgrading FLM as we have had some driver issues with this method and had to lock down the version of flm we use on the server.
FLM might also need a driver compatibility check.

src/lemonade/tools/huggingface/load.py
```diff
@@ -212,12 +216,13 @@ def run(
         state.tokenizer = HuggingfaceTokenizerAdapter(tokenizer, device)
         state.dtype = dtype_to_use
         state.checkpoint = checkpoint
-        state.device = device
+        state.device = device if device else "cpu"
```
ramkrishna2910, line 219: Why is this part of the FLM PR? Is this adding support for GPU on huggingface?

amd-pworfolk: I added the `Backend` statistic to each type of load function because this is critical information for understanding the result.   Device is no longer sufficient since there are multiple different ways to run on each device.  In testing this for each load function, I discovered that by default, the HuggingFace device value is `None`, so I fixed that.  So now whichever model load function is used, the backend and device statistics are set correctly.

src/lemonade/tools/flm/load.py
amd-pworfolk: I created a `backend_versions.json` file (like the server) used to specify the FLM version and minimum NPU driver version.

ramkrishna2910 (approved):

PR #622, opened 2025-11-26
fix: Serve favicon from root

VladimirVLF:
This PR ixes serving the favicon from the root url as `<LEMONADE-URL>/favicon.ico`.

This location is expected by most browsers and therefore fixes the icon for the Lemonade bookmark, at least in Firefox.


jeremyfowers:
Should be easy to merge, will try it out locally.

JFYI the web ui will be removed soon when the new app comes out.

VladimirVLF:
I hope it will still be possible to run Lemonade without graphics just with command line, otherwise it will likely break the possibility of running it in docker containers.

jeremyfowers:
> I hope it will still be possible to run Lemonade without graphics just with command line, otherwise it will likely break the possibility of running it in docker containers.

Yes lemonade-server-minimal installer is going to be a no-UI backend-only installer going forwards.

PS. as you can see, our CI system took a dive today. I'll have it fixed by tomorrow morning and then can resume reviewing PRs.

jeremyfowers (approved):
I already had the favicon displaying, but I can confirm that this didn't break anything on my system. I will take your word that it helps on other systems!

PR #624, opened 2025-11-27
Lemonade Manager: Phase 1

danielholanda:
# Description

The App PR is distributed in two phases to facilitate review:

**Phase 1:** All artifacts **inside** `src\app` are added to the repo. No integration is done at this point. GitHub will likely choke while trying to display all changes, which is not an issue given that a detailed review is only expected in phase 2.

**Phase 2:**  All files **outside** `src\app` are added. Those changes are minimal compared to the previous ones. GitHub is expected to properly allow reviewers to navigate changes. This will integrate the app into our software and replace the WebUI on both Windows and Linux.

### DIFF
--- a/src/app/.gitignore
+++ b/src/app/.gitignore
@@ -1,0 +1,39 @@
+# Node modules
+node_modules/
+
+# Build output
+dist/
+dist-app/
+build/
+
+# OS files
+.DS_Store
+Thumbs.db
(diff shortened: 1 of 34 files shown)


danielholanda:
Too complicated to explain over text, so I will explain during scrum and reopen :)

PR #627, opened 2025-11-29
Adding CPU Backend Support to LlamaCPP Server for Lemonade CPP

siavashhub:
### Description
This PR adds CPU parameters support to the C++ LlamaCPP server, aligning its behavior with the existing Python implementation.
It introduces a new configuration flag (cpuParameters) that allows users to pass CPU-only parameters without requiring GPU-specific backend flags. 
This is especially useful for systems lacking GPU support or when testing CPU-only behavior.

### Changes
1. Configuration Update (`src/cpp/resources/backend_versions.json`)
    - Added a new boolean field: "cpuParameters".
    - This flag allows enabling/disabling GPU parameter usage globally from the config file.

2.  CPU Fallback Logic in Server (`src/cpp/server/backends/llamacpp_server.cpp`)
    - Added get_cpu_param_enabled() helper to load the new config field with proper error handling.
    - Added a use_gpu decision variable that defaults to GPU usage but switches to CPU when cpuParameters is true.
    - Added detailed logging to indicate backend selection and fallback status.
    - Added conditional behavior for:
      - MMProj loading (skip or disable offload when GPU is off)
      - GPU layer configuration (-ngl 99 for GPU, -ngl 0 for CPU-only)

### DIFF
--- a/docs/server/lemonade-server-cli.md
+++ b/docs/server/lemonade-server-cli.md
@@ -47,6 +47,8 @@ lemonade-server run MODEL_NAME [options]
 
 These settings can also be provided via environment variables that Lemonade Server recognizes regardless of launch method: `LEMONADE_HOST`, `LEMONADE_PORT`, `LEMONADE_LOG_LEVEL`, `LEMONADE_LLAMACPP`, `LEMONADE_CTX_SIZE`, and `LEMONADE_LLAMACPP_ARGS`.
 
+Additionally, you can provide your own `llama-server` binary by giving the full path to it via the following environment variables: `LEMONADE_LLAMACPP_ROCM_BIN`, `LEMONADE_LLAMACPP_VULKAN_BIN`. Note that this does not override the `--llamacpp` option, rather it allows to provide an alternative binary for specific backends.
+
 ## `pull` Command Options
 
 The `pull` command downloads and installs models. For models already in the [Lemonade Server registry](./server_models.md), only the model name is required. To register and install custom models from Hugging Face, use the registration options below:
--- a/src/cpp/include/lemon/backends/llamacpp_server.h
+++ b/src/cpp/include/lemon/backends/llamacpp_server.h
@@ -44,6 +44,7 @@ public:
 private:
     std::string get_llama_server_path();
     std::string find_executable_in_install_dir(const std::string& install_dir);
+    std::string find_external_llama_server(const std::string& backend);
     
     std::string backend_;  // vulkan, rocm, metal, cpu
     std::string model_path_;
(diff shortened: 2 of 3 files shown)


siavashhub:
> Hey @siavashhub good to see you contributing more!
> 
> Since the intent of this PR is to allow the user to globally disable llamacpp's GPU access, I wonder if a different solution might be better?
> 
> ggml-org/llama.cpp releases provide CPU-only `llama-b7217-bin-win-cpu-x64.zip` and `llama-b7217-bin-ubuntu-x64.zip` artifacts.
> 
> If those were supported, as alternatives to llamacpp rocm/vulkan, then users could do `lemonade-server serve --llamacpp cpu` or use `LEMONADE_LLAMACPP_BACKEND="cpu"` env var. And they would not need to go into any config file.
> 
> What do you think?

Hey @jeremyfowers,
Sure I can change the router to support cpu as a backend value and then have the logic in llamacpp file to check backend value or the env var instead of cpuParameter that I added and download the cpu artifacts if it's set to "cpu" and pass correct parameters.
Give me a couple of days and I will work on it.

jeremyfowers:
> > Hey @siavashhub good to see you contributing more!
> > Since the intent of this PR is to allow the user to globally disable llamacpp's GPU access, I wonder if a different solution might be better?
> > ggml-org/llama.cpp releases provide CPU-only `llama-b7217-bin-win-cpu-x64.zip` and `llama-b7217-bin-ubuntu-x64.zip` artifacts.
> > If those were supported, as alternatives to llamacpp rocm/vulkan, then users could do `lemonade-server serve --llamacpp cpu` or use `LEMONADE_LLAMACPP_BACKEND="cpu"` env var. And they would not need to go into any config file.
> > What do you think?
> 
> Hey @jeremyfowers, Sure I can change the router to support cpu as a backend value and then have the logic in llamacpp file to check backend value or the env var instead of cpuParameter that I added and download the cpu artifacts if it's set to "cpu" and pass correct parameters. Give me a couple of days and I will work on it.

Awesome! I’m setting PR status to draft, please take it out of draft when I should review :)

siavashhub:
Seems there has been changes to the code causing conflicts, let me fix them first.

siavashhub:
@jeremyfowers  all done and ready for your review.
tested it on windows and ubuntu.

siavashhub:
> Tested locally and it worked perfectly!
> 
> Would you kindly mention this option in the highlighted line of these two docs:
> 
> https://github.com/lemonade-sdk/lemonade/blob/c3d9293a1163700a7f9d6f1ea4544d40dfbc14a5/docs/server/server_spec.md?plain=1#L804
> 
> https://github.com/lemonade-sdk/lemonade/blob/c3d9293a1163700a7f9d6f1ea4544d40dfbc14a5/docs/server/lemonade-server-cli.md?plain=1#L43
> 
> Also FYI #651 is going to merge soon, and may have conflicts with this PR. I can help resolve if thats the case.

done!
thank you!

jeremyfowers (approved):
Thanks for another great contribution @siavashhub!

PR #642, opened 2025-12-03
Bug Fix: Allow run command to work when server is already running

danielholanda:
## The bug

The C++ implementation was acting different from the Python implementation (expected behavior).

When you tried to run `lemonade-server run <model>` while Lemonade was already running, it would just say "Error: Another instance is already running" and exit. This was bad because you couldn't load a model and open the UI without stopping and restarting everything.

## The fix
Now when you run `lemonade-server run <model>` and the server is already up:
* It detects the running server and connects to it 
* Loads the model you specified
* Opens the browser to the chat interface


src/cpp/tray/tray_app.cpp
```diff
@@ -313,10 +313,36 @@ int TrayApp::run() {
         // Check for single instance - only for 'serve' and 'run' commands
         // Other commands (status, list, pull, delete, stop) can run alongside a server
         if (lemon::SingleInstance::IsAnotherInstanceRunning("Server")) {
+            // If 'run' command and server is already running, connect to it and execute the run command
```
danielholanda, line 316: Note: I found it odd that this change had to be done on a file called `tray_app.cpp`. Looks like a lot of the old `cli.py` functionality landed here?

jeremyfowers: Yeah…. needs a refactor

jeremyfowers: `lemonade-server` is in “cpp/tray”
`lemonade-router` is in “cpp/server”

🤦‍♂️

PR #651, opened 2025-12-04
Allow overriding llama-server bin path through the environment

bitgamma:
Give the user the ability to override through environment variables the location of the llama-server binary. This setting is per-backend. For example:

LEMONADE_LLAMACPP_ROCM_BIN
LEMONADE_LLAMACPP_VULKAN_BIN

this allow using the auto-installed binary for some backends and only override others. This also skips version checks usually done by the installer. 

In my case this is useful because it makes the ROCm backend work on Arch (tested on gfx1150). I guess there could be several use cases where a custom built version of llama-server is desirable


src/cpp/include/lemon/backends/llamacpp_server.h
```diff
@@ -44,6 +44,7 @@ class LlamaCppServer : public WrappedServer, public IEmbeddingsServer, public IR
 private:
```
jeremyfowers, line 44: I'm testing this locally now!

Would you kindly document the new environment variable? Here: https://github.com/lemonade-sdk/lemonade/blob/main/docs/server/lemonade-server-cli.md

bitgamma: added documentation

bitgamma:
> It works!
> 
> ```
> [ProcessManager] Starting process with filtered output: "C:\Users\Jeremy\Downloads\llama-b7268-bin-win-vulkan-x64\llama-server.exe" "-m" "C:\Users\Jeremy\.cache\huggingface\hub/models--unsloth--Qwen3-0.6B-GGUF\snapshots\50968a4468ef4233ed78cd7c3de230dd1d61a56b\Qwen3-0.6B-Q4_0.gguf" "--ctx-size" "4096" "--port" "8001" "--jinja" "--context-shift" "--keep" "16" "--reasoning-format" "auto" "-ngl" "99"
> ```
> 
> However, definitely needs documentation. My first instinct about how to set the env var was wrong (I passed the folder path instead of the executable path).
> 
> Will merge this as soon as the doc is updated.

Great! Documentation added!

src/cpp/include/lemon/backends/llamacpp_server.h
jeremyfowers: looks good!

jeremyfowers (approved):
Tested locally and worked like a charm. I think power users will really appreciate this.

PR #658, opened 2025-12-04
Dev Containers - Docker build system implemented and includes cmake p…

Geramy:
I have implemented a configuration for Visual Studio Code which uses Dev Containers and Docker to setup a build system without having to do anything but have the Dev Containers extension and docker in order to build. It also includes the CMake plugin. This allows you to build "remotely" via docker and not have to worry about your local system / computer having all the dependencies. It also reduces the issue of version mismatch on dependencies and allows for the programmer and build system to be platform agnostic, since docker is universal.


Geramy:
Done, created a readme.md with details of how to use the new implemented configurations.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM mcr.microsoft.com/devcontainers/cpp:1-debian-12
+
+ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="3.22.2"
+
+# [FIX IS HERE] We must specify the path relative to the Project Root
+# because devcontainer.json context is set to ".."
```
jeremyfowers:
```suggestion
```

Removing AI self-referential comments

src/cpp/README.md
```diff
@@ -36,6 +36,48 @@ sudo apt install build-essential cmake libcurl4-openssl-dev libssl-dev pkg-confi
...
+### Developer IDE & IDE Build Steps
+#### Visual Studio Code Setup Guide
+1. Clone the repository into a blank folder locally on your computer.
+2. Open the folder in visual studio code.
+3. Install Dev Containers extension in visual studio code by using
+  control + p to open the command bar at the top of the IDE or if on mac with Cmd + p.
+4. Type "> Extensions: Install Extensions" which will open the Extensions side panel.
+5. in the extensions search type ```Dev Containers``` and install it.
+6. Once completed with the prior steps you may run command
+```>Dev Containers: Open Workspace in Container``` which you can do in the command bar in the IDE and it should reopen the visual studio code project.
```
jeremyfowers: I don't see "open workspace in container" but I do see "open folder in container", is that expected?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+WORKDIR /usr/src/app
+
+# Copy the Project Root into the container
+COPY . .
+
+# Install the package
+RUN pip install --upgrade pip setuptools wheel && \
+    pip install ".[dev]"
```
jeremyfowers: The `.[dev]` extra is rather heavy, and includes a bunch of python tools for things like benchmarking safetensors models with CUDA, and converting safetensors models to ONNX, etc.

What's the intent behind this installation? Hopefully we can do something lighter.

If you just want to be able to test the C++ server with the python tests, all you need is to `pip install -r` this file https://github.com/lemonade-sdk/lemonade/blob/main/test/requirements.txt

src/cpp/README.md
```diff
@@ -36,6 +36,48 @@ sudo apt install build-essential cmake libcurl4-openssl-dev libssl-dev pkg-confi
 xcode-select --install
 ```
 
+### Developer IDE & IDE Build Steps
```
jeremyfowers, line 39: This failed for me in Windows, but is working for me in WSL. Should the instructions say that Linux is a requirement?

This is my first time trying dev containers so I'm a bit of a noob here.

.devcontainer/Dockerfile
jeremyfowers: >  => [dev_container_auto_added_stage_label 8/8] RUN pip install --upgrad  300.0s

Yeah this step took 5 minutes on my system. We can cut this way down unless [dev] is needed for something.

src/cpp/README.md
Geramy: That should be fine, it may be different because I have it already open in my system. I will notate and phrase for both options.

.devcontainer/Dockerfile
Geramy: No thats fine we can do that. Technically this should be a run once setup for new developers to be able to start doing C++ development in a uniform environment. Continual development shouldn't have to deal with any build time unless the docker file changes and needs to be rebuilt. But 5 minutes is a long time.

src/cpp/README.md
Geramy: It should be working with windows, but I have mac so its hard for me to test for the windows use case.
Did you end up getting this to work under windows with docker?

jeremyfowers: It worked for me on WSL but not Windows - didn't try too hard to debug. I figure most people on regular windows will just use the msi installer and not any dev container. Let's merge.

.devcontainer/Dockerfile
Geramy: done.

PR #673, opened 2025-12-06
Allow overriding whisper-server bin path via the environment

bitgamma:
Specular to #651 but for `whisper-server`. The variable name is `LEMONADE_WHISPERCPP_BIN`. This seems particularly useful on non-Windows systems at the moment because the automatically downloaded release seems to be Windows only and [on the release page of the project](https://github.com/ggml-org/whisper.cpp/releases/tag/v1.8.2) I couldn't find precompiled linux or Mac binaries

### DIFF
--- a/src/cpp/server/backends/whisper_server.cpp
+++ b/src/cpp/server/backends/whisper_server.cpp
@@ -188,15 +188,15 @@ void WhisperServer::install(const std::string& backend) {
     bool needs_install = exe_path.empty();
 
     if (needs_install) {
-        std::string install_dir = get_whisper_install_dir();
-        std::string version_file = (fs::path(install_dir) / "version.txt").string();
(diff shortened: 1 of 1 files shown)


jeremyfowers:
Thanks for doing this @bitgamma!

> at the moment because the automatically downloaded release seems to be Windows only and [on the release page of the project](https://github.com/ggml-org/whisper.cpp/releases/tag/v1.8.2) I couldn't find precompiled linux or Mac binaries

yeah this is a puzzle for us. there are also no GPU/vulkan builds in the release. We need to consider some PRs to enable more release builds there, or start publishing our own builds.

jeremyfowers (approved):
Tested locally and confirmed that it works as expected.

Thanks for another nice contribution @bitgamma!

PR #676, opened 2025-12-08
Add download manager to UI

danielholanda:
Closes https://github.com/lemonade-sdk/lemonade/pull/676
Closes https://github.com/lemonade-sdk/lemonade/issues/691

# Description

This PR adds a download manager to the UI.
download manager (1)

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -11,6 +11,7 @@
 #include <fstream>
 #include <memory>
 #include <thread>
+#include <chrono>
 #include <mutex>
 #include <condition_variable>
 #include <atomic>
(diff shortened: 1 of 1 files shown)


jeremyfowers:
where did this come from 😂

awesome!

src/app/src/renderer/App.tsx
```diff
@@ -5,6 +5,7 @@ import ModelManager from './ModelManager';
 import LogsWindow from './LogsWindow';
```
jeremyfowers, line 5: https://github.com/user-attachments/assets/7c39b4aa-b21f-4586-bec2-2511ee699dcc

it would be nice if clicking outside the download manager would close the download manager. right now I'm forced to click the X

src/app/src/renderer/App.tsx
```diff
@@ -5,6 +5,7 @@ import ModelManager from './ModelManager';
...
 import CenterPanel from './CenterPanel';
```
jeremyfowers, line 6: image

The "total size" reported in the download manager appears to be the size of the last file downloaded. In this case its the config.json, which is very small. Total size should be the GGUF size + JSON size in this case.

Otherwise users will think the download was bugged.

src/app/src/renderer/App.tsx
danielholanda: Done! This is now enabled!

src/app/src/renderer/App.tsx
danielholanda: Great catch! Tracking cumulative download size now!
image

src/app/src/renderer/App.tsx
```diff
@@ -5,6 +5,7 @@ import ModelManager from './ModelManager';
...
 import ResizableDivider from './ResizableDivider';
```
jeremyfowers, line 7: image

it would be nice to get a "resume" button on canceled/partial downloads.

src/app/src/renderer/App.tsx
```diff
@@ -5,6 +5,7 @@ import ModelManager from './ModelManager';
...
+import DownloadManager from './DownloadManager';
 import '../../styles.css';
```
jeremyfowers, line 9: image

Deleting a loaded model is majorly bugged in the UI

the server backend does delete the model, but it still shows as loaded in the UI

pressing delete multiple times can even produce multiple "model was deleted" popups

jeremyfowers: image

hah it even shows the eject, download, and delete buttons all at the same time in the model manager

danielholanda: Oh man! That is funny. I don't see that. My guess is that you hit a corner case due to the issue with the models endpoint incorrectly showing partially downloaded models as downloaded.

src/app/src/renderer/App.tsx
```diff
@@ -5,6 +5,7 @@ import ModelManager from './ModelManager';
...
 
 const LAYOUT_CONSTANTS = {
```
jeremyfowers, line 11: image

If I push the red X (cancel download) and then the retry button in quick succession, it starts pulling again before the cancel's delete has a chance to finish. There seems to be a race condition.

src/app/src/renderer/App.tsx
jeremyfowers: fixed!

src/app/src/renderer/App.tsx
danielholanda: Solved! We are now preventing retry until cleanup is complete

jeremyfowers: image

image

Looks like `delete` is being called before the client disconnects. Taking a quick look...

jeremyfowers: It's a server-side race condition. Will push a fix.

jeremyfowers: Solved!

jeremyfowers (approved):
"LGTM"

PR #683, opened 2025-12-09
Add pre-commit support

superm1:


superm1:
The second commit is the output of `pre-commit run --all-files` if you would rather do it yourself.

superm1:
@Geramy @jeremyfowers @danielholanda can you look this over?  If you don't want to to review all the files closely you can reproduce what I did in the second commit and compare by `pre-commit run --all-files` on your side too.

PR #693, opened 2025-12-09
Improve FLM logic and add LFM2-1.2B-FLM

### LINKED ISSUE #689: flm: find on PATH, instead of pre-determined paths

jeremyfowers:
Closes #689 

Fixes:
- Avoid bugs where `flm.exe` was not found, resulting in the models list not including downloaded FLM models

Improvements:
- Added `LFM2-1.2B-FLM` to the models list

### DIFF
--- a/src/lemonade_server/server_models.json
+++ b/src/lemonade_server/server_models.json
@@ -592,6 +592,12 @@
         "suggested": true,
         "size": 2.62
     },
+    "LFM2-1.2B-FLM": {
+        "checkpoint": "lfm2:1.2b",
+        "recipe": "flm",
+        "suggested": true,
+        "size": 0.96
+    },
     "Whisper-Tiny": {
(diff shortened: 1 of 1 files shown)


src/cpp/include/lemon/utils/path_utils.h
```diff
@@ -18,6 +18,14 @@ std::string get_executable_dir();
  */
 std::string get_resource_path(const std::string& relative_path);
 
+/**
+ * Find the FLM executable (flm.exe on Windows, flm on Unix).
```
danielholanda, line 22: Unix? Isn't FLM Windows-only?
```suggestion
 * Find the FLM executable.
```

jeremyfowers: Future proofing ;)

Issue #685, opened 2025-12-09
Sync app version number with the rest of Lemonade


jeremyfowers:
Right now the app has its own version number, 1.0.0.

IMO this should be the same as the Lemonade C++ project, which is currently set in src/cpp/cmakelists

It is fine if this version is only applied when building the app installer via cmake (ie, if `npm start` still shows version 1.0.0 thats ok).

danielholanda:
This is possible, but would require tons of new lines of code to sync both.

This version number only shows up when `npm start` is run. What about simply adding a `.npmrc` file to `src/app` containing `loglevel=silent` to suppress unimportant info messages? This would effectively get rid of the issue with minimal code changes.

> Note: The correct version is already shown in the app's "about" section

danielholanda:
Here is a PR with this change if you want to try it:
https://github.com/lemonade-sdk/lemonade/pull/695

jeremyfowers:
Eh nevermind

PR #695, opened 2025-12-09
Suppress App info to avoid need to dynamically sync versions [proposal]

danielholanda:
Closes https://github.com/lemonade-sdk/lemonade/issues/685

# Description

This PR supporesses app info to avoid the need to sync versions dynamically. Warnings and Errors are still shown.


jeremyfowers:
Nevermind :)

PR #696, opened 2025-12-10
Filter available models by size and system ram

### LINKED ISSUE #687: hide gpt-oss-20b-FLM on <32 GB systems

jeremyfowers:
closes #687 

This PR filters out models from the `models` endpoint according to these two rules:
- Model size is >= 80% of system RAM
- `gpt-oss-120b` is only allowed when System RAM is >= 64GB

This will save user confusion. We've had a lot of instances of people pulling large models they can't actually run.

To test on an arbitrary system:
- Open server_models.json and change `size` of a model to be very large
- cmake ..
- cmake --build . --config release
- .\Release\lemonade-server serve 

to test on a 32 GB STX system:
- the biggest model you see should be Qwen3-30B
- gpt-oss-20b-FLM should not appear


danielholanda (approved):

flst01:
I have a question on this: 
Why is only RAM and not VRAM or combined RAM+VRAM considered?
On a Strix Halo machine, I can allocate 96 GB of the shared 128 GB to VRAM. This means I am able to load a model like gpt-oss-120b fully into VRAM, if I allocate 96 GB.
However now I am prohibited to do so because of having unsufficient RAM - Which shouldn't really be used in interference?

flst01:
The RAM condition can be bypassed on a Strix Halo Linux machine by editing GRUB like presented [here](https://github.com/technigmaai/technigmaai-wiki/wiki/AMD-Ryzen-AI-Max--395:-GTT--Memory-Step%E2%80%90by%E2%80%90Step-Instructions-%28Ubuntu-24.04%29). 
This way, all "unified RAM" is seen as RAM by lemonade-server.

jeremyfowers:
> I have a question on this: Why is only RAM and not VRAM or combined RAM+VRAM considered? On a Strix Halo machine, I can allocate 96 GB of the shared 128 GB to VRAM. This means I am able to load a model like gpt-oss-120b fully into VRAM, if I allocate 96 GB. However now I am prohibited to do so because of having unsufficient RAM - Which shouldn't really be used in interference?

@flst01 thanks for the question and sorry about the slow reply, I was on vacation.

What operating system are you on? On Windows Strix Halo, I have 64 GB VRAM and 64 GB RAM, but Lemonade reports my "physical memory" as 128 GB.

If you would like to continue the discussion please open an issue so we don't lose track of it.

PR #699, opened 2025-12-10
Route index.html to `api/v1`

danielholanda:
Index.html is now also served on `api/v1`. Please test this with apps before merging.


jeremyfowers (approved):
Works great!

PR #741, opened 2025-12-16
Add --extra-models-dir option

### LINKED ISSUE #710: Support llamacpp and lmstudio model directories
Most optimal way of storing models for me is in a seperate drive with LM Studio structure (e.g. vendor/model/*.gguf).

Event llama-server now supports multiple model loading with --models-path

jeremyfowers:
Closes #710 

image

This PR adds a highly requested feature: let me use any GGUF I want, especially:
- GGUFs that aren't on Hugging Face
- GGUFs I downloaded through some non huggingface-hub means (like `llama-cli -hf CHECKPOINT`)

The main feature added here is a `--extra-models-dir DIR` argument. GGUF files in that directory are added to the `models` endpoint and can be loaded and inferenced. See the new spec file, FAQ, and CLI doc update for details.

This isn't perfect yet, but it's a start that will unblock the majority of use-cases.

This is only LLM GGUFs. Support for embeddings, reranking, audio, OGA, etc. could be added at a later time if there's demand.


src/cpp/server/model_manager.cpp
```diff
@@ -263,6 +263,169 @@ std::string ModelManager::get_hf_cache_dir() const {
...
+        info.source = "extra_models_dir";
+        info.labels.push_back("custom");
+        
+        // Calculate size in GB
+        try {
+            uintmax_t file_size = fs::file_size(gguf_path);
+            info.size = static_cast<double>(file_size) / (1024.0 * 1024.0 * 1024.0);
+        } catch (...) {
+            info.size = 0.0;
+        }
```
danielholanda, line 351: Why not also reuse this logic for `user.` models?

docs/server/lemonade-server-cli.md
```diff
@@ -43,9 +43,10 @@ lemonade-server run MODEL_NAME [options]
 | `--llamacpp [vulkan\|rocm\cpu]`    | Default LlamaCpp backend to use when loading models. Can be overridden per-model via the `/api/v1/load` endpoint. | vulkan |
 | `--ctx-size [size]`            | Default context size for models. For llamacpp recipes, this sets the `--ctx-size` parameter for the llama server. For other recipes, prompts exceeding this size will be truncated. Can be overridden per-model via the `/api/v1/load` endpoint. | 4096 |
 | `--llamacpp-args [args]`       | Default custom arguments to pass to llama-server. Must not conflict with arguments managed by Lemonade (e.g., `-m`, `--port`, `--ctx-size`, `-ngl`). Can be overridden per-model via the `/api/v1/load` endpoint. Example: `--llamacpp-args "--flash-attn on --no-mmap"` | "" |
+| `--extra-models-dir [path]`    | Secondary directory to scan for GGUF model files. Models found here are automatically added to the model list with the `llamacpp` recipe. Supports multimodal and multi-shard models. Suggested paths: `%LOCALAPPDATA%\llama.cpp` (Windows) or `~/.cache/llama.cpp` (Linux/macOS). | None |
```
danielholanda: This flag is specific to llama.cpp. If you don't want to enable this option for all backends, please call it `--llamacpp-extra-models-dir`.

src/cpp/Extra-Models-Dir-Spec.md
```diff
@@ -0,0 +1,66 @@
+# Extra Models Directory Specification
```
danielholanda, line 1: This seems like an extremely niche feature to deserve a full spec doc just for itself. This will make users question why do we have such a detailed spec for this, but not for other features that are a lot more visible.

Suggest either (1) removing it or (2) moving this and the multiModel spec to a `Spec` folder if we want to start having a bunch of these for every single feature we add.

docs/server/lemonade-server-cli.md
danielholanda: Make sure to clearly mention the limitation here (LLMs only) and create an issue for completing this task.

danielholanda (changes requested):
Please take a look at the comments. The main recommendation is to change the option name to `-llamacpp-models-dir`.

That said, please consider the possibility of enabling a similar feature that does not require as many code changes and doesn't create tech debt in terms of unsuported backends of model types (e.g., completely replacing the default models directory instead of adding an extra one [`--models-dir`]).

src/cpp/Extra-Models-Dir-Spec.md
danielholanda: Chatted offline. We will move this to a folder when we go beyond 5

docs/server/lemonade-server-cli.md
danielholanda: Chatted offline: Please add more details to the description before merging.

PR #763, opened 2025-12-27
Make some llamacpp-server args optional defaults

bitgamma:
This PR makes some of the arguments set by lemonade when launching llama-server optional defaults. It works like this: if no custom `llamacpp_args` are present, apply the default flags otherwise skip all default flags completely. The affected flags are

`--context-shift`
`--keep`
`--reasoning-format`

the rationale for this is that those arguments are not strictly needed and power users might want to customize them, so making them strictly reserved seems too restrictive. These flags are also not mentioned as being reserved anywhere in the documentation.

Also by introducing the concept of optional defaults, it is easier to balance user-friendliness without sacrificing flexibility for power users.

### DIFF
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -98,7 +98,7 @@ static void push_overridable_arg(std::vector<std::string>& args,
     }
 }
 
-// Helper to add a flag-value overridable pair (e.g., --port 8000, -m model.gguf)
+// Helper to add a flag-value overridable pair (e.g., --keep 16)
 static void push_overridable_arg(std::vector<std::string>& args,
                     const std::string& custom_args,
                     const std::string& key,


src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -639,6 +627,18 @@ void LlamaCppServer::load(const std::string& model_name,
         std::cout << "[LlamaCpp] Adding custom arguments: " << custom_args_ << std::endl;
         std::vector<std::string> custom_args_vec = parse_custom_args(custom_args_);
         args.insert(args.end(), custom_args_vec.begin(), custom_args_vec.end());
+    } else {
```
jeremyfowers: I like the intent of this change, but I don't think I can merge it as-is because it is a breaking change that will impact downstream apps.

Consider a pre-existing app is doing a load with `--no-mmap` with today's Lemonade v9.1.1. 

In Lemonade v9.1.2 with this change, by loading with `--no-mmap` they will inadvertently disable `--context-shift --keep 16` unless they manually append those options to their load command. In turn this can cause their completions to crash.

I would be in favor of merging a different implementation that doesn't incur a breaking change, though.

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -595,18 +595,6 @@ void LlamaCppServer::load(const std::string& model_name,
         }
     }
     
-    // Enable context shift for vulkan/rocm (not supported on Metal)
```
jeremyfowers, line 598: > These flags are also not mentioned as being reserved anywhere in the documentation.

image

True! This part of the server spec should mention those flags.

I opened a doc bug PR to fix the docs. I'll wait to see what happens with this PR; if we merge this PR then I'll just close that issue.

https://github.com/lemonade-sdk/lemonade/issues/789

src/cpp/server/backends/llamacpp_server.cpp
bitgamma: I was hoping a breaking change at this point could still be somewhat acceptable. I could add a `--no-llamacpp-defaults` flag to explicitly trigger this behavior but for the long term it seemed more clean to have this way. wdyt?

jeremyfowers: We already have a few production apps like GAIA and Morphik that use this feature, so I would prefer not to mess with them.

I know it would be more complex to implement, but what would you think of this: if the user provides `--keep X` or `--reasoning-format Y`, allow those to override the default values, rather than erroring on them being reserved.

`--context-shift` is a special case, but it is possible as well. If the user puts `--no-context-shift` (which is a real `llama-server` arg), replace `--context-shift` with `--no-context-shift`.

That would involve checking whether `llamacpp_args` contains any of those special arg names and replacing them as needed. Would you be up for that implementation?

bitgamma: ok, I'll reimplement it that way then

bitgamma: done. The way the flags are searched in `push_overridable_arg` is naive but with the long-form flags (i.e. starting with `--`) there is no possible ambiguity and if the command line is malformed it will be detected during the regular validation step.

Note that for flags that have both a long-form and short-form we would need special handling, but the same is true about the way reserved flags are currently checked. None of the current overridable flags has this problem as they don't have alternative forms.

The problem instead exists for most reserved flags (for example `--ctx-size` can also be written as `-c`) meaning the user could accidentally override them by using one of their alternative forms, but this is something outside the scope of this PR, it could be handled in a separate one.

PR #774, opened 2025-12-29
Lemonade API key

### LINKED ISSUE #769: Support API key
Since one might want to expose their lemonade server over a network, it wouldn't hurt having the ability to add a single API key functioning as a password of a sort to avoid intentional or accidental DoS attacks.

Since it is quite usual to pass this kind of secrets over the env, I'd add an optional `LEMONADE_API_KEY` env. If it is present, all requests will need to have this API key.

If you think this is a good idea let me know and I'll provide a PR

bitgamma:
closes #769 

defines the LEMONADE_API_KEY which will be required to serve any request. The authentication is done in a way compatible with OpenAI, meaning all tools that have a "API key" field in their connection settings will work out of the box

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -102,7 +102,7 @@ void Server::log_request(const httplib::Request& req) {
 }
 
 httplib::Server::HandlerResponse Server::authenticate_request(const httplib::Request& req, httplib::Response& res) {
-    if (api_key_ != "") {
+    if ((api_key_ != "") && (req.method != "OPTIONS")) {
         if (api_key_ != httplib::get_bearer_token_auth(req)) {
             res.status = 401;
             res.set_content("{\"error\": \"Invalid or missing API key\"}", "application/json");


src/cpp/server/server.cpp
```diff
@@ -91,15 +94,31 @@ Server::~Server() {
...
+    if (req.path != "/api/v0/health" && req.path != "/api/v1/health") {
+        std::cout << "[Server PRE-ROUTE] " << req.method << " " << req.path << std::endl;
+        std::cout.flush();
+    }
+}
+
+httplib::Server::HandlerResponse Server::authenticate_request(const httplib::Request& req, httplib::Response& res) {
+    if (api_key_ != "") {
+        if (api_key_ != httplib::get_bearer_token_auth(req)) {
+            res.status = 401;
```
jeremyfowers, line 107: @bitgamma what am I doing wrong here?

image

I set my API key to "bob", then launched the server, then tried to connect with Open WebUI and got a 401 error.

bitgamma: please try removing the quotes from bob when settings the environment variable. I tested today with Open WebUI and everything works as expected so this is likely the problem.

jeremyfowers (approved):
Tested locally in Continue and Open WebUI, works great!

PR #788, opened 2026-01-05
Steamline the new user experience

### LINKED ISSUE #765: App new user UX: no models in chat
User story:
1. Install lemonade for the first time. I have no models.
2. Open the app
3. The models dropdown the chat section is blank, and raises a 404 if I send a message
4. I download a model
5. The models dropdown the chat section is *still* blank and *still* raises a 404 if I send a message
6. I manually click the little model load button in the model manager
7. Now things work

This is too unintuitive and error-prone for my taste. I'm not a new user, just using a new computer, and ran into this.

To repro on a computer with models, set HF_HOME to some directory with no models and then start the server.

Suggested solution: new users clicking around the UI should get helpful hints and never a generic 404 error.

cc @danielholanda

jeremyfowers:
Closes #765 

# Overview

- Fixes aspects of the new user experience that could result in 404 errors for trying basic actions (see below)
- Consolidates the chat's fetchModels() and fetchSupportedModels(), as well as model manager's fetchModels(), and associated data structures.
    - We were having a lot of strange behaviors because functionality was triplicated and data structures could get out of sync.

# New user experience

## Prompt

I am working on the App at src/app @src/app/README.md 

Right now, the behavior for when no model is downloaded is a bit undefined. We are going to turn this into a nice experience.

If the models endpoint check returns an empty list, add Qwen3-0.6B-GGUF to the LLM Chat models dropdown. This way there is a default model, so that if people type a message and push send something will happen.

We need a good UX for this case as well: the model download should show up in the download manager, and then when the download completes the model the message is sent to it.

This means we cant just send a chat/completions request to the model when the user hits send. In this special case, we need to do a streaming pull of the model first, then chat/completions.

From the user's perspective:
1. They install lemonade on a fresh PC and open the app for the first time
2. The LLM Chat panel has Qwen3-0.6B-GGUF selected (even though no models are downloaded yet)
3. User types a message and pushes send.
4. Qwen3-0.6B-GGUF downloads in the Download Manager
5. when the download completes, the Download Manager minimizes, the user's message is sent to the LLM, and the response is printed in LLM Chat

Note that the "send" button in LLM Chat should be disabled during steps 4 and 5. 

## Agent's Code Changes

image

### DIFF
--- a/src/app/src/renderer/hooks/useModels.tsx
+++ b/src/app/src/renderer/hooks/useModels.tsx
@@ -1,0 +1,174 @@
+import React, { createContext, useContext, useState, useEffect, useCallback, useMemo, useRef } from 'react';
+import { ModelsData, ModelInfo, USER_MODEL_PREFIX, fetchSupportedModelsData } from '../utils/modelData';
+import { onServerPortChange } from '../utils/serverConfig';
+
+// Default model to use when no models are downloaded (first-time user experience)
+export const DEFAULT_MODEL_ID = 'Qwen3-0.6B-GGUF';
+
+export interface DownloadedModel {
+  id: string;
+  info: ModelInfo;
+}
+
+export interface SuggestedModel {
+  name: string;
+  info: ModelInfo;
+}
+
+interface ModelsContextValue {
+  // All models with full metadata (keyed by model ID)
+  modelsData: ModelsData;
+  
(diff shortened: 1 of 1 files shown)


twaelti:
StReamline, not steamline ;-)

PR #793, opened 2026-01-06
Extra models dir models have an extra. prefix

### LINKED ISSUE #759: Compatibility with LM Studio models directory layout
https://discord.com/channels/1392562559122407535/1394403929973985413/1452827404744527953

jeremyfowers:
Closes #759 

Also consolidates some redundant constants.

image

Installed all 4 models via LM Studio. Imported them to Lemonade and tested all 4, including image input in the Gemma.

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -714,8 +714,15 @@ void ModelManager::build_cache() {
     
     // Step 1.5: Discover models from extra_models_dir
     // All discovered models are prefixed with "extra." to avoid conflicts
+    // We still gracefully handle conflicts just in case, though
     auto discovered_models = discover_extra_models();
     for (const auto& [name, info] : discovered_models) {
+        // Check for conflicts with registered models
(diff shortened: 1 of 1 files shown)


src/cpp/server/model_manager.cpp
```diff
@@ -715,14 +713,9 @@ void ModelManager::build_cache() {
     }
     
     // Step 1.5: Discover models from extra_models_dir
+    // All discovered models are prefixed with "extra." to avoid conflicts
     auto discovered_models = discover_extra_models();
     for (const auto& [name, info] : discovered_models) {
-        // Check for conflicts with registered models
```
Geramy, line 724: Since extra is user controlled or thats my assumption, I probably would leave this in the case they accidentally have multiple models in directories, they could accidentally have two of the same models, thats my first thought when reading this. Its not the programs job to organize the users directories of course but it could be useful.

jeremyfowers: Good call, done!

PR #796, opened 2026-01-06
Improve the UX when FLM updates invalidate models

### LINKED ISSUE #794: Poor FLM UX when models are invalidated
FLM 0.9.24 invalidates the old LFM2-1.2b-FLM model and requires it to be re-download. `flm list` auto-deletes the old model, which is good, but the UX in Lemonade is still not appropriate.

## Current State

in the general case of an flm 0.9.23 install:

1. User clicks to load a model
2. FLM auto-updates
3. We don't refresh the models list based on FLM update alone
4. It looks like lfm2-1.2b is still there
5. User clicks to load lfm2-1.2b
6. lfm2-1.2b auto-re-downloads
7. lfm2-1.2b loads

in the specific case of a flm 0.9.23 install clicking on lfm2-1.2b:

1. User clicks to load lfm2-1.2b
2. FLM auto-updates
3. lfm2-1.2b auto-re-downloads
4. lfm2-1.2b loads

## Solution

1. FLM updates should immediately call `flm list` and update lemonade's models cache
2. If FLM is being updated because a specific model load is requested, and the FLM update invalidates that model (it now shows as not available in `flm list`), then send an HTTP error to the user telling them they need to download the model again.

jeremyfowers:
Closes #794 

image

New features:
1. Server: Check for model invalidations when upgrading FLM (model changes from available to unavailable when calling `flm list`)
2. App: When trying to load an invalidated model, first download it with the download manager (see screenshot above), then load it.

### DIFF
--- a/src/app/src/renderer/ModelManager.tsx
+++ b/src/app/src/renderer/ModelManager.tsx
@@ -541,7 +541,7 @@ const ModelManager: React.FC<ModelManagerProps> = ({ isVisible, width = 280 }) =
     };
   }, [handleDownloadModel]);
 
-  const handleLoadModel = async (modelName: string) => {
+  const handleLoadModel = async (modelName: string, autoLoadAfterDownload: boolean = false) => {
     try {
       const modelData = modelsData[modelName];
       if (!modelData) {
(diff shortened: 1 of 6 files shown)


jeremyfowers:
fyi @danielholanda a couple small app changes

PR #798, opened 2026-01-06
Add a --quiet flag to serve and run that disables the startup balloon

### LINKED ISSUE #783: Only show ‘Lemonade Started’ modification when manually launched
# Description

We should only show the ‘Lemonade Started’ notification when Lemonade is manually launched.

jeremyfowers:
Closes #783


danielholanda (approved):
Love how simple those changes were!

PR #804, opened 2026-01-07
implement save_options

### LINKED ISSUE #791: `save_options` flag for /load API and `--save-options` CLI option
#775 added support for a recipe_options file. We want to make it easier to modify that file by exposing it through the CLI and HTTP API.

The /load HTTP method should get a `save_options` flag that saves any user-set `llamacpp_args`, `llamacpp_backend`, and `ctx_size` options to `recipe_options.json`.

Additionally, the `lemonade-server` CLI should expose `--save-options` for the `run` command.

Notes:
- Settings saved this way should completely override any prior options saved for that model in `recipe_options.json`.
- `lemonade-server serve` does NOT get `--save-options`.
    - You'll need to introduce the concept of a run-only arg in the CLI parser and docs, since right now `serve` and `run` share all args.
- The CLI should not directly write the JSON file. It should only set `save_options: true` in the /load HTTP method, so that lemonade-router is the only entity that reads or writes the JSON file.

For example:

`lemonade-server run Qwen3-0.6B-GGUF --llamacpp-args --no-mmap --save-options` should result in this entry in `recipe_options.json`:

```
"Qwen3-0.6B-GGUF" : {
    "llamacpp_args" : "--no-mmap"
}
```

And then, if the user does this: `lemonade-server run Qwen3-0.6B-GGUF --llamacpp rocm --save-options`, then the entry in `recipe_options.json` should be overwritten to:

```
"Qwen3-0.6B-GGUF" : {
    "llamacpp_backend" : "rocm"
}
```

Please be sure to update the docs at both https://github.com/lemonade-sdk/lemonade/blob/main/docs/server/lemonade-server-cli.md and https://github.com/lemonade-sdk/lemonade/blob/main/docs/server/server_spec.md

### LINKED ISSUE #809: Model info endpoint should return saved recipe options for that model
Required for #792 

The [model info endpoint](https://lemonade-server.ai/docs/server/server_spec/#get-apiv1modelsmodel_id) `/models/MODEL_NAME` to return all saved recipe options for that model.

bitgamma:
Closes #791
Closes #809 

- [x] Implement save_options in the load API
- [x] Implement the CLI switch `--save-options`
- [x] Extend model info endpoint to return recipe options
- [x] Update `server_specs.md`
- [x] Update `lemonade-server-cli.md`

### DIFF
--- a/docs/server/server_spec.md
+++ b/docs/server/server_spec.md
@@ -685,9 +685,11 @@ Returns a single model object with the same fields as described in the [models l
   "downloaded": true,
   "suggested": true,
   "labels": ["reasoning"],
-  "ctx_size": 8192,
-  "llamacpp_args": "--no-mmap",
-  "llamacpp_backend": "rocm"
+  "recipe_options" {
+    "ctx_size": 8192,
+    "llamacpp_args": "--no-mmap",
+    "llamacpp_backend": "rocm"
+  }
 }
 ```
 
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -762,6 +762,20 @@ void Server::handle_models(const httplib::Request& req, httplib::Response& res) 
 }
 
 nlohmann::json Server::model_info_to_json(const std::string& model_id, const ModelInfo& info) {
+    nlohmann::json recipe_options = nlohmann::json::object();
+
+    if (info.ctx_size >= 0) {
+        recipe_options["ctx_size"] = info.ctx_size;
+    }
+
+    if (!info.llamacpp_backend.empty()) {
+        recipe_options["llamacpp_backend"] = info.llamacpp_backend;
+    }
+
+    if (!info.llamacpp_args.empty()) {
+        recipe_options["llamacpp_args"] = info.llamacpp_args;
+    }
+
     nlohmann::json model_json = {
         {"id", model_id},
         {"object", "model"},
@@ -772,9 +786,7 @@ nlohmann::json Server::model_info_to_json(const std::string& model_id, const Mod
         {"downloaded", info.downloaded},
         {"suggested", info.suggested},
         {"labels", info.labels},
-        {"ctx_size", info.ctx_size},
-        {"llamacpp_backend", info.llamacpp_backend},
-        {"llamacpp_args", info.llamacpp_args},
+        {"recipe_options", recipe_options},
     };
     
     // Add size if available


docs/server/server_spec.md
```diff
@@ -684,7 +684,10 @@ Returns a single model object with the same fields as described in the [models l
   "size": 0.38,
   "downloaded": true,
   "suggested": true,
-  "labels": ["reasoning"]
+  "labels": ["reasoning"],
+  "ctx_size": 8192,
+  "llamacpp_args": "--no-mmap",
+  "llamacpp_backend": "rocm"
```
jeremyfowers: What's the behavior if these aren't set in `recipe_options.json`? 

It might be better to have a key called `recipe_options`?

That way if nothing is set, its an empty dict

```
"recipe_option" : {}
```

And if it is set, its easy to tell whats an option (and therefore optional / may not always be in the response)

```
"recipe_options" : {
    "ctx_size": 8192,
    "llamacpp_args": "--no-mmap"
}
```

bitgamma: the current behavior is to return their "empty" value (ctx_size -1, empty string for the other). But I like your idea better.

src/cpp/tray/server_manager.cpp
```diff
@@ -333,9 +333,19 @@ nlohmann::json ServerManager::get_models() {
...
     try {
-        std::string body = "{\"model_name\": \"" + model_name + "\"}";
+        nlohmann::json load_req = nlohmann::json::object();
+        load_req["model_name"] = model_name;
+
+        if (save_options) {
+            load_req["save_options"] = true;
+            load_req["ctx_size"] = ctx_size_;  
+            load_req["llamacpp_backend"] = llamacpp_backend_;  
+            load_req["llamacpp_args"] = llamacpp_args_;
```
jeremyfowers, line 345: Maintainability note: I noticed that `ctx_size`, `llamacpp_backend`, and `llamacpp_args` are hardcoded multiple times in the code as the recipe options that can be set.

What happens if we add another save-able option? It would be a lot of code edits to add it.

Is it possible to abstract this into an `options` struct that can be passed between functions, so that the struct can be programmatically loaded and saved without explicitly referencing each key by name?

bitgamma: it is possible but to do it right it would require refactoring the existing code in many places. Would it be OK to do it in a separate PR? Not to mix new features with refactoring

src/cpp/server/model_manager.cpp
```diff
@@ -636,14 +636,40 @@ json ModelManager::load_optional_json(const std::string& path) {
...
+}
+
+void ModelManager::save_model_options(const ModelInfo& info) {
+    std::cout << "[ModelManager] Saving options for model: " << info.model_name << std::endl;
+
+    json model_options = json::object();
+
+    if (info.ctx_size >= 0) {
+        model_options["ctx_size"] = info.ctx_size;
+    }
```
jeremyfowers, line 659: I ran 

```
C:\work\forks\bitgamma\lemonade\src\cpp\build> .\Release\lemonade-server.exe run user.Qwen3-0.6B-GGUF --llamacpp vulkan --save-options
```

And my `recipe_options.json` updated to:

```
  "user.Qwen3-0.6B-GGUF": {
    "ctx_size": 4096,
    "llamacpp_backend": "vulkan"
  }
```

It's a bug that `"ctx_size": 4096,` was saved?

bitgamma: no, that is intentional. It is how we agreed to do it on Discord

src/cpp/tray/server_manager.cpp
jeremyfowers: I'll leave it up to you, if you want to do it in this PR, another PR, or not at all. Since you're doing so much work with the feature. I just like to point these things out when I notice them.

src/cpp/server/model_manager.cpp
jeremyfowers: Haha I'm reading the discord again, and I realize I said I'm fine with whatever you pick. And then I didn't remember what you picked :)

docs/server/server_spec.md
bitgamma: done

src/cpp/tray/server_manager.cpp
bitgamma: I definitely think it is a good idea as it makes maintenance easier. I'll do it in a separate PR

docs/server/server_spec.md
jeremyfowers: looks great!

PR #812, opened 2026-01-08
Added ability to build rpm packages for Fedora users

SidShaytay:
Updated the readme and the .cmake to easily allow the creation of rpm packages for Fedora users. 

This is done just by 
```
cd src/cpp/build
cpack -G RPM
```

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -510,5 +510,7 @@ if(UNIX AND NOT APPLE)
         endif()
     endif()
     
+    # RPM specific variables defined within
+    include(${CMAKE_CURRENT_SOURCE_DIR}/CPackRPM.cmake)
     include(CPack)
 endif()
--- a/src/cpp/CPackRPM.cmake
+++ b/src/cpp/CPackRPM.cmake
@@ -1,0 +1,21 @@
+# Optional CPack RPM configuration for lemonade-server
+# Include this file before include(CPack) in CMakeLists.txt
+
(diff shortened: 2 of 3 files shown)


jeremyfowers (commented):
Thanks for the PR, @SidShetye!

I don't have a Fedora system to test with, but as long as our CI checks for Ubuntu still pass then I am happy to merge this.

SidShaytay:
Thanks @jeremyfowers. I mostly did this so you can reach Fedora users too. @kyuz0's toolboxes has really helped here). 

It should fine if your CI is Ubuntu-based. Ubuntu runners should be able build Fedora RPMs via CPack by installing rpm-build (or running the build inside a Fedora container; docker run fedora). If you ever feel like shipping Fedora RPMs, the existing CI can get you there with minimal tweaks. 

Would be good to keep the momentum for your Fedora users too!

jeremyfowers:
> Thanks @jeremyfowers. I mostly did this so you can reach Fedora users too. @kyuz0's toolboxes has really helped here).
> 
> It should fine if your CI is Ubuntu-based. Ubuntu runners should be able build Fedora RPMs via CPack by installing rpm-build (or running the build inside a Fedora container; docker run fedora). If you ever feel like shipping Fedora RPMs, the existing CI can get you there with minimal tweaks.
> 
> Would be good to keep the momentum for your Fedora users too!

Awesome, appreciate it @SidShetye! I want Lemonade to be for everyone, but I only have so much bandwidth to support all platforms. It is super helpful when community members come in and contribute more support <3

I’m definitely open to shipping RPMs in the future if people want it.

Issue #706, opened 2025-12-10
`run` command does not open app on Linux if the server is already running


jeremyfowers:
In one terminal:

```
lemonade-server serve
```

In a second terminal:

```
lemonade-server run Qwen3-0.6B-GGUF --log-level debug
DEBUG: Signal handlers installed
DEBUG: TrayApp::run() starting...
DEBUG: Command: run
DEBUG: Searching for server binary...
DEBUG: Found server binary: /usr/local/bin/lemonade-router
DEBUG: Using server binary: /usr/local/bin/lemonade-router
Lemonade Server is already running. Connecting to it...
Running model: Qwen3-0.6B-GGUF
Loading model Qwen3-0.6B-GGUF...
Model loaded successfully!
Launching Lemonade app...
Found Electron app at: /usr/local/share/lemonade-server/app/lemonade
Launched Electron app (PID: 19786)
Shutting down server...
DEBUG: Shutting down gracefully...
Terminating Electron app (PID: 19786)...
```

Expected behavior: app opens

Current behavior: the electron app is killed as the second lemonade-server process self-exits.

cc @danielholanda

anditherobot:
Should we close this issue?
 Verified in **Lemonade 11.8.0**: `lemonade run` loads models via HTTP and opens the UI with `xdg-open` on
  Linux, so exiting the CLI no longer closes the window.
**Resolution PR:** [#1336](https://github.com/lemonade-sdk/lemonade/pull/1336) 
(Commit[`8c010710`](https://github.com/lemonade-sdk/lemonade/commit/8c0107107ab739bdb21a27e667fa8e13b4cb0300))

fl0rianr:
I agree and see this as resolved, thanks.

Issue #704, opened 2025-12-10
Two+ copies of the Lemonade App can open at the same time on Linux


jeremyfowers:
This behavior is supposed to be prevented.

Easiest repro:

```
/usr/local/share/lemonade-server/app/lemonade
/usr/local/share/lemonade-server/app/lemonade
/usr/local/share/lemonade-server/app/lemonade
```

Will open 3 copies of the app.

cc @danielholanda

jeremyfowers:
Closing as don’t care

PR #815, opened 2026-01-09
fix server detection on Linux

bitgamma:
This patch fixes behavior of the cli on Linux. In particular:

1. Correctly detect running server even if started by another user
2. Do not attempt to shutdown server if it wasn't started by the current command (on Linux the current implementation would cause all commands, including `list` to delete the PID file on exit)
3. Do not attempt to start the electron app if `run` connects to an existing server, because since the function returns immediately it would be shutdown the next second anyway (this also applies to Windows I think)

### DIFF
--- a/src/cpp/include/lemon_tray/tray_app.h
+++ b/src/cpp/include/lemon_tray/tray_app.h
@@ -135,7 +135,7 @@ private:
     std::string loaded_model_;
     std::vector<ModelInfo> downloaded_models_;
     bool should_exit_;
-    bool owns_server_ = false;
+    bool process_owns_server_ = false;
     
     // Model loading state
     std::atomic<bool> is_loading_model_{false};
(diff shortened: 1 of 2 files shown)


jeremyfowers:
There are some related bugs, will test whether these are now resolved

https://github.com/lemonade-sdk/lemonade/issues/706
https://github.com/lemonade-sdk/lemonade/issues/704

bitgamma:
> There are some related bugs, will test whether these are now resolved
> 
> #706 #704

well this patch makes the behavior in #706 become canon (it stops trying to start the electron app if the server is already running and you do a `run`). The rationale is that 

1. if you are just starting the server locally with the run, you might want the electron app
2. if the server is started and want to use the GUI, you just start the GUI
3. if you are using cli anyway, you probably don't want stuff to pop up at screen (it might also be impossible if it is a remote ssh session

PR #820, opened 2026-01-09
Add --convert flag when launching whisper-server

poplk:
## Summary
This PR adds the `--convert` flag when launching whisper-server to fix crashes in binaries compiled with ffmpeg support.

## Problem
When whisper-server binaries are compiled with ffmpeg support, they crash because whisper.cpp takes the content of the file instead of the filename (different behaviour than with the default miniaudio integrated in whisper).

## Solution
Add the `--convert` flag when launching whisper-server. This flag has been tested and is safe for:
- Binaries built with ffmpeg support (fixes the crash)
- Binaries using only the default built-in miniaudio library (no side effects)

## Test plan
- Test with whisper-server binary compiled with ffmpeg support (with standards wav/mp3 and with m4a)
- Test with whisper-server binary using miniaudio only
- Verify transcription works correctly in both cases

### DIFF
--- a/src/cpp/server/backends/whisper_server.cpp
+++ b/src/cpp/server/backends/whisper_server.cpp
@@ -403,7 +403,8 @@ void WhisperServer::load(const std::string& model_name,
     // Note: Don't include exe_path here - ProcessManager::start_process already handles it
     std::vector<std::string> args = {
         "-m", model_path_,
-        "--port", std::to_string(port_)
+        "--port", std::to_string(port_),
+        "--convert"
     };
 
     // Note: whisper-server doesn't support --debug flag


src/cpp/server/backends/whisper_server.cpp
```diff
@@ -403,7 +403,8 @@ void WhisperServer::load(const std::string& model_name,
     // Note: Don't include exe_path here - ProcessManager::start_process already handles it
     std::vector<std::string> args = {
         "-m", model_path_,
-        "--port", std::to_string(port_)
+        "--port", std::to_string(port_),
+        "--convert"
```
ramkrishna2910, line 407: Is there an issue or discussion that references this behavior? Does this cause any performance impact for the non-ffmpeg case?

jeremyfowers: there's some discussion here: https://discordapp.com/channels/1392562559122407535/1394403929973985413/1459094565939056804

ramkrishna2910:
Looks like we are also facing this CI failure, do we need to install ffmpeg as a dependency?
```
=== Stopping Lemonade ===
Error: ] ffmpeg is not found. Please ensure that ffmpeg is installed and that its executable is included in your system's PATH. [ERROR] whisper-server process has terminated with exit code: 0
[stdout] 
```

src/cpp/server/backends/whisper_server.cpp
ramkrishna2910: Ok, this does not have a negative impact on performance. But ffmpeg needs to be installed per [this](https://github.com/ggml-org/whisper.cpp/blob/master/examples/server/README.md):
` --convert,                     [false  ] Convert audio to WAV, requires ffmpeg on the server`

poplk: Hum that's weird I was sure I tested with a whisper compiled without ffmpeg and it worked also. So I guess we have to detect if the whisper binary has ffmpeg built in at start up and only pass the parameter if yes

jeremyfowers:
Closing as abandoned. @poplk we'd love for your come re-open and merge these if you have a chance!

PR #828, opened 2026-01-09
Stable diffusion cpp integration

ramkrishna2910:
This PR adds text-to-image generation capabilities to Lemonade Server using [sd-cpp](https://github.com/leejet/stable-diffusion.cpp) as the backend.

## Core Implementation
  - New sd-cpp recipe and SDServer backend wrapping sd-server
  - OpenAI-compatible /api/v1/images/generations endpoint
  - Auto-download of sd-cpp binaries on first use

 ## Models
  - SD-Turbo, SDXL-Turbo, SD-1.5, SDXL-Base-1.0 added to model registry

##  Tests & Docs
  - test/server_sd.py - CI tests for image generation
  - examples/api_image_generation.py - Usage examples
  - Updated README, server_spec.md, and CLI docs

## Follow-up PRs

  - Vulkan/ Rocm GPU acceleration (currently CPU-only)
  - Support for Z-image, Flux and Qwen-Image
  - Lemonade desktop app UI integration

## To test

  - Start the server and try the command below.
 ```powershell
    Invoke-WebRequest `
      -Uri "http://localhost:8000/api/v1/images/generations" `
      -Method POST `
      -Headers @{ "Content-Type" = "application/json" } `
      -Body '{
        "model": "SD-Turbo",
        "prompt": "A serene mountain landscape at sunset",
        "size": "512x512",
        "steps": 4,
        "response_format": "b64_json"
      }'

### DIFF
--- a/docs/server/server_spec.md
+++ b/docs/server/server_spec.md
@@ -609,20 +609,18 @@ Image Generation API. You provide a text prompt and receive a generated image. T
 
 #### Example request
 
-=== "PowerShell"
+=== "Bash"
 
-    ```powershell
-    Invoke-WebRequest `
-      -Uri "http://localhost:8000/api/v1/images/generations" `
-      -Method POST `
-      -Headers @{ "Content-Type" = "application/json" } `
-      -Body '{
-        "model": "SD-Turbo",
-        "prompt": "A serene mountain landscape at sunset",
-        "size": "512x512",
-        "steps": 4,
-        "response_format": "b64_json"
-      }'
+    ```bash
+    curl -X POST http://localhost:8000/api/v1/images/generations \
+      -H "Content-Type: application/json" \
+      -d '{
+            "model": "SD-Turbo",
+            "prompt": "A serene mountain landscape at sunset",
+            "size": "512x512",
+            "steps": 4,
+            "response_format": "b64_json"
+          }'
     ```
 
 #### Response format


bitgamma:
This is amazing! From my limited testing, here is what I found out. Consider I am on Linux (Arch) on gfx1150

1. Generation causes this error:

> /home/mikkoph/git/tmp/lemonade/src/cpp/build/sd-cpp/sd-cli: error while loading shared libraries: libstable-diffusion.so: cannot open shared object file: No such file or directory

I worked around this setting `LD_LIBRARY_PATH`. I think this should be done automatically, like it is done for `llama-server`

2. `--save-images` is ignored => the tmp image is deleted and no file is stored

3. The `sd-cpp` binaries are stored in the old location. To fix this `get_sd_base_dir()` should to be removed and `get_downloaded_bin_dir()` from `lemon/utils/path_utils.h` should be called instead (might need a rebase though since this function was added yesterday)

examples/api_image_generation.py
```diff
@@ -0,0 +1,74 @@
+"""
```
jeremyfowers, line 1: All of the other examples in this folder are using `lemonade.api`, which is deprecated. I need to clear them all out...

Are you think that we should repopulate the folder with server examples that call http endpoints?

examples/api_image_generation.py
```diff
@@ -0,0 +1,74 @@
...
+This example demonstrates how to use the lemonade server API to generate
+images using Stable Diffusion models via the OpenAI Python client.
```
jeremyfowers, line 3: Execute this example in CI testing

examples/api_image_generation.py
ramkrishna2910: Yes, this set of specific examples would be valuable. I think we should repopulate it.

docs/server/lemonade-server-cli.md
```diff
@@ -46,6 +46,8 @@ lemonade-server run MODEL_NAME [options]
 | `--llamacpp-args [args]`       | Default custom arguments to pass to llama-server. Must not conflict with arguments managed by Lemonade (e.g., `-m`, `--port`, `--ctx-size`, `-ngl`). Can be overridden per-model via the `/api/v1/load` endpoint. Example: `--llamacpp-args "--flash-attn on --no-mmap"` | "" |
 | `--extra-models-dir [path]`    | Experimental feature. Secondary directory to scan for LLM GGUF model files. Audio, embedding, reranking, and non-GGUF files are not supported, yet. | None |
 | `--max-loaded-models [LLMS] [EMBEDDINGS] [RERANKINGS] [AUDIO]` | Maximum number of models to keep loaded simultaneously. Accepts 1, 3, or 4 values for LLM, embedding, reranking, and audio models respectively. Unspecified values default to 1. Example: `--max-loaded-models 3 2 1 1` loads up to 3 LLMs, 2 embedding models, 1 reranking model, and 1 audio model. | `1 1 1 1` |
+| `--save-images`                | Save generated images to disk instead of cleaning up after returning base64 response. Useful for debugging or keeping generated images. | False |
+| `--images-dir [path]`          | Directory to save generated images when `--save-images` is enabled. | `./generated_images` |
```
jeremyfowers: Thoughts about simplifying this by always saving images to `~/.cache/lemonade/generated_images`?

I'm a little worried about CLI bloat.

Or, never saving images directly from the server, and leave that functionality up to the application layer. Like including a "save image" button the Lemonade App.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1112,6 +1112,240 @@ jobs:
...
+          install-path: ${{ env.LEMONADE_INSTALL_PATH }}
+
+      - name: Setup Python and virtual environment
+        uses: ./.github/actions/setup-venv
+        with:
+          venv-name: '.venv'
+          python-version: '3.10'
+          requirements-file: 'test/requirements.txt'
+
+      - name: Install Visual C++ Redistributable (for sd-cpp AVX2 build)
```
jeremyfowers: Does this mean that SD support does not work on Lemonade unless the user manually installs Visual C++?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1112,6 +1112,240 @@ jobs:
...
+        run: |
+          $ErrorActionPreference = "Stop"
+          $venvPython = ".\.venv\Scripts\python.exe"
+
+          $serverExe = Join-Path $env:LEMONADE_INSTALL_PATH "bin\lemonade-server.exe"
+
+          Write-Host "Running Stable Diffusion --save-images tests with C++ server CLI: $serverExe" -ForegroundColor Cyan
+          Write-Host "HF_HOME is set to: $env:HF_HOME" -ForegroundColor Cyan
+
+          & $venvPython test/server_sd.py --server-binary $serverExe --test-save-images
```
jeremyfowers: If you moved this one line of code up to the prior step would it save the 20 other lines of code in this step?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1112,6 +1112,240 @@ jobs:
...
+          # Extract the .deb package
+          mkdir -p lemonade-extracted
+          dpkg-deb -x "$DEB_FILE" ./lemonade-extracted
+
+          # Copy resources to user's local share (path_utils.cpp checks here!)
+          mkdir -p ~/.local/share/lemonade-server
+          cp -r ./lemonade-extracted/usr/local/share/lemonade-server/* ~/.local/share/lemonade-server/
+
+          echo "Extraction complete. Contents:"
+          find ./lemonade-extracted -type f -name "lemonade-*" | head -20
```
jeremyfowers: Is there not an action defined for this? This block be repeated a lot in this file lol

jeremyfowers: And the "Verify binaries work" step below - they could be combined into a single Action

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1112,6 +1112,240 @@ jobs:
...
+
+          echo "Running Stable Diffusion tests with C++ server CLI: $SERVER_BIN"
+          echo "HF_HOME is set to: $HF_HOME"
+          echo "Note: Image generation may take 2-3 minutes with CPU backend"
+
+          .venv/bin/python test/server_sd.py --server-binary $SERVER_BIN
+
+          echo "Stable Diffusion tests PASSED!"
+
+      - name: Run Stable Diffusion --save-images tests
```
jeremyfowers: Same comment about converging test steps

README.md
```diff
@@ -89,6 +89,38 @@ Use `lemonade-server pull` or the built-in **Model Manager** to download models.
 
 <br clear="right"/>
 
+## Image Generation
```
jeremyfowers, line 92: You added an imagegen section but not transcription?

jeremyfowers: Lets discuss how we want to reconfigure the top README around multi-modal.

docs/server/server_spec.md
```diff
@@ -9,6 +9,7 @@ Lemonade Server currently supports these backends:
 | [Llama.cpp](https://github.com/ggml-org/llama.cpp)    | `.GGUF`      | Uses llama.cpp's `llama-server` backend. More details [here](#gguf-support).                    |
 | [ONNX Runtime GenAI (OGA)](https://github.com/microsoft/onnxruntime-genai) | `.ONNX`      | Uses Lemonade's own `ryzenai-server` backend.                                                |
 | [FastFlowLM](https://github.com/FastFlowLM/FastFlowLM)    | `.q4nx`      | Uses FLM's `flm serve` backend. More details [here](#fastflowlm-support).                    |
+| [stable-diffusion.cpp](https://github.com/leejet/stable-diffusion.cpp) | `.safetensors` | Uses sd.cpp's `sd-cli` backend for image generation. Models: SD-Turbo, SDXL-Turbo, etc. |
```
jeremyfowers, line 13: Where is whisper.cpp?

docs/server/server_spec.md
```diff
@@ -582,6 +584,73 @@ Audio Transcription API. You provide an audio file and receive a text transcript
...
+> **Performance:** CPU inference takes ~4-5 minutes per image. GPU (Vulkan) is faster but may have compatibility issues with some hardware.
+
+#### Parameters
+
+| Parameter | Required | Description | Status |
+|-----------|----------|-------------|--------|
+| `prompt` | Yes | The text description of the image to generate. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `model` | Yes | The Stable Diffusion model to use (e.g., `SD-Turbo`, `SDXL-Turbo`). | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `size` | No | The size of the generated image. Format: `WIDTHxHEIGHT` (e.g., `512x512`, `256x256`). Default: `512x512`. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `n` | No | Number of images to generate. Currently only `1` is supported. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
```
jeremyfowers: "Only 1 is supported" - I would call this partially available lol

docs/server/server_spec.md
```diff
@@ -582,6 +584,73 @@ Audio Transcription API. You provide an audio file and receive a text transcript
...
+| `response_format` | No | Format of the response. Only `b64_json` (base64-encoded image) is supported. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
```
jeremyfowers: partially supported

README.md
```diff
@@ -89,6 +89,38 @@ Use `lemonade-server pull` or the built-in **Model Manager** to download models.
...
+lemonade-server pull SD-Turbo
+
+# Start the server (optionally save generated images)
+lemonade-server --save-images
+```
+
+Generate images using the OpenAI-compatible API:
+
+```python
+from openai import OpenAI
```
jeremyfowers: I'm pretty torn on these python examples. I know we also provide them for LLM, but they take up a lot of space and I don't know how many people use them vs. calling `curl`

docs/server/server_spec.md
```diff
@@ -582,6 +584,73 @@ Audio Transcription API. You provide an audio file and receive a text transcript
...
+      -Body '{
+        "model": "SD-Turbo",
+        "prompt": "A serene mountain landscape at sunset",
+        "size": "512x512",
+        "steps": 4,
+        "response_format": "b64_json"
+      }'
+    ```
+
+#### Response format
```
jeremyfowers, line 626: Should the response include the location of the image file on disk, if applicable?

README.md
ramkrishna2910: Fair. Given we have one in the examples folder, we can get rid of this to save space.

src/cpp/include/lemon/backends/sd_server.h
```diff
@@ -0,0 +1,68 @@
...
+    // WrappedServer interface
+    void install(const std::string& backend = "") override;
+
+    std::string download_model(const std::string& checkpoint,
+                              const std::string& mmproj = "",
+                              bool do_not_upgrade = false) override;
+
+    void load(const std::string& model_name,
+             const ModelInfo& model_info,
+             int ctx_size,
```
jeremyfowers: this will need to be replaced by the new RecipeOptions class, see #837 for how this modification looks in other WrappedServers

src/cpp/include/lemon/backends/sd_server.h
```diff
@@ -0,0 +1,68 @@
...
+
+    void unload() override;
+
+    // ICompletionServer implementation (not supported - return errors)
+    json chat_completion(const json& request) override;
+    json completion(const json& request) override;
+    json responses(const json& request) override;
+
+    // IImageServer implementation
+    json image_generations(const json& request) override;
```
jeremyfowers, line 40: Any reason this doesnt mention other capabilities like embedding, reranking, and transcription?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -0,0 +1,816 @@
...
+    }
+}
+
+// Helper to get the install directory for sd executable
+static std::string get_sd_install_dir() {
+    return (fs::path(get_downloaded_bin_dir()) / "sd-cpp").string();
+}
+
+// Helper to run a process safely without shell injection vulnerabilities
+static int run_process_safe(const std::vector<std::string>& args) {
```
jeremyfowers: there isnt a helper for this kind of thing in utils? I'm surprised if its unique to SD - if so, why?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -0,0 +1,816 @@
...
+        if (WIFEXITED(status)) {
+            return WEXITSTATUS(status);
+        }
+        return -1;
+    }
+#endif
+}
+
+// Helper to extract ZIP files
+static bool extract_zip(const std::string& zip_path, const std::string& dest_dir) {
```
jeremyfowers: Same question about utils - is there not a shared function for this? Most backends handle zips.

src/cpp/server/backends/sd_server.cpp
```diff
@@ -0,0 +1,816 @@
...
+
+    if (model_path.empty() || !fs::exists(model_path)) {
+        throw std::runtime_error("Failed to download SD model: " + checkpoint);
+    }
+
+    std::cout << "[SDServer] Model downloaded to: " << model_path << std::endl;
+    return model_path;
+}
+
+void SDServer::load(const std::string& model_name,
```
jeremyfowers, line 266: I don't really understand what's going on here - there's a lot of lines of code, but its not loading a model into memory?

Does each call to the imagegen endpoint has to load the model from scratch? Will that not add a lot of latency to generating the images, especially when we move to GPU and the generation itself gets faster?

ramkrishna2910: Yes, sd-cpp server did not accept steps, sample_method, or scheduler as JSON parameters.
But as of 2 days ago sd-cpp has added [support for these features](https://github.com/leejet/stable-diffusion.cpp/releases/tag/master-470-48d3161) in their server.
So I am taking this PR back to draft and moving to the server.

docs/server/server_spec.md
ramkrishna2910: Yup fixed.

docs/server/server_spec.md
ramkrishna2910: Fixed.

README.md
ramkrishna2910: Done.

src/cpp/include/lemon/backends/sd_server.h
ramkrishna2910: Rebased to main.

src/cpp/server/backends/sd_server.cpp
ramkrishna2910: Moved to sd-server

src/cpp/server/backends/sd_server.cpp
ramkrishna2910: Moving to sd-server removed this function

docs/server/server_spec.md
ramkrishna2910: Added whisper cpp.

src/cpp/server/backends/sd_server.cpp
ramkrishna2910: Moved to utils and updated all backends.

src/cpp/include/lemon/backends/sd_server.h
ramkrishna2910: WrappedServer declares these as pure virtual methods, so all derived classes must implement them even if they just return errors. Ideally sd_server needs to only define `image_generations`.

.github/workflows/cpp_server_build_test_release.yml
ramkrishna2910: Got rid of it.

.github/workflows/cpp_server_build_test_release.yml
ramkrishna2910: This test has been removed

docs/server/server_spec.md
ramkrishna2910: Its not a part of the OpenAI spec. We can add it, it wouldn't break OpenAI compatibility since clients would just ignore unknown fields. Currently the user is responsible for saving the image returned. Maybe wait for user feedback?

.github/workflows/cpp_server_build_test_release.yml
ramkrishna2910: Done

.github/workflows/cpp_server_build_test_release.yml
ramkrishna2910: Remove the save image test

examples/api_image_generation.py
ramkrishna2910: CI uses a similar example but uses smaller resolutions and step counts to optimize for CI time. I would prefer the example to have usable resolutions and the test run fast. Duplication worth the cost?

jeremyfowers: Gotcha... ok the policy will be: if the example breaks once, it goes into CI.

docs/server/server_spec.md
```diff
@@ -583,6 +586,65 @@ Audio Transcription API. You provide an audio file and receive a text transcript
...
+| `size` | No | The size of the generated image. Format: `WIDTHxHEIGHT` (e.g., `512x512`, `256x256`). Default: `512x512`. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `n` | No | Number of images to generate. Currently only `1` is supported. | <sub>![Status](https://img.shields.io/badge/partial-yellow)</sub> |
+| `response_format` | No | Format of the response. Only `b64_json` (base64-encoded image) is supported. | <sub>![Status](https://img.shields.io/badge/partial-yellow)</sub> |
+| `steps` | No | Number of inference steps. SD-Turbo works well with 4 steps. Default varies by model. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `cfg_scale` | No | Classifier-free guidance scale. SD-Turbo uses low values (~1.0). Default varies by model. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `seed` | No | Random seed for reproducibility. If not specified, a random seed is used. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+
+#### Example request
+
+=== "PowerShell"
```
jeremyfowers: bash version of the command plz ;)

src/lemonade_server/server_models.json
```diff
@@ -699,5 +699,33 @@
...
+        "suggested": false,
+        "labels": ["image"],
+        "size": 4.3
+    },
+    "SDXL-Base-1.0": {
+        "checkpoint": "stabilityai/stable-diffusion-xl-base-1.0:sd_xl_base_1.0.safetensors",
+        "recipe": "sd-cpp",
+        "suggested": false,
+        "labels": ["image"],
+        "size": 6.9
```
jeremyfowers, line 729: Why bother adding these if they are suggested=false?

src/lemonade_server/server_models.json
```diff
@@ -699,5 +699,33 @@
...
+        "labels": ["image"],
+        "size": 5.2
+    },
+    "SDXL-Turbo": {
+        "checkpoint": "stabilityai/sdxl-turbo:sd_xl_turbo_1.0_fp16.safetensors",
+        "recipe": "sd-cpp",
+        "suggested": true,
+        "labels": ["image"],
+        "size": 6.9
+    },
```
jeremyfowers, line 716: These show up in the app and, when loaded, use the llm chat interface, which obviously doesn't work:

image

Maybe set suggested=false so they don't show up in the app for now?

src/lemonade_server/server_models.json
jeremyfowers: Unless you're taking the suggestion above ;) https://github.com/lemonade-sdk/lemonade/pull/828/files#r2724876944

examples/api_image_generation.py
```diff
@@ -0,0 +1,74 @@
...
+    print()
+
+    # Generate using OpenAI client
+    result = generate_with_openai_client()
+
+    print()
+    print("=" * 60)
+    print("Done!")
+    if result:
+        print(f"Generated image saved to: {result}")
```
jeremyfowers, line 74: image

"A guy who looks like a lemon, painting a lemonade on canvas mounted to an easel"

jeremyfowers: So trippy I love it

src/cpp/include/lemon/backends/sd_server.h
jeremyfowers: Got it

src/lemonade_server/server_models.json
ramkrishna2910: Done

Issue #144, opened 2025-08-07
Lemonade on Arch Linux


jeremyfowers:
I have an Ubuntu install on my 9070xt box and I just get a generic "no rocm device found" error. Pretty opaque, unfortunately.

jeremyfowers:
cc @danielholanda

sofiageo:
Hey, I'm the guy that asked on reddit. I tested it on my PC (Arch Linux) with a 9070xt I just bought and it works fine (and it's super fast as well).

```
INFO:     Loading llm: Gemma-3-4b-it-GGUF
WARNING:  llamacpp binaries are built for Ubuntu. Compatibility with other Linux distributions is not guaranteed.
INFO:     Using backend: rocm
INFO:     Downloading llama.cpp server from https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1021/llama-b1021-ubuntu-rocm-gfx120X-x64.zip
INFO:     Extracting llama-b1021-ubuntu-rocm-gfx120X-x64.zip to /home/xxx/.local/share/pipx/venvs/lemonade-sdk/bin/rocm/llama_server
INFO:     Set executable permissions for /home/xxx/.local/share/pipx/venvs/lemonade-sdk/bin/rocm/llama_server/llama-server
INFO:     Set executable permissions for /home/xxx/.local/share/pipx/venvs/lemonade-sdk/bin/rocm/llama_server/llama-cli
```

I had some issues but not related to the GPU:

-  I had to manually change the host to 0.0.0.0 instead of localhost
-  and comment out the if tray: code block completely from serve.py file

jeremyfowers:
Awesome! Welcome to the github :)

>I had to manually change the host to 0.0.0.0 instead of localhost

interesting, hadnt seen that before, but there is a PR open right now to at least make it convenient to change the host #142 

>and comment out the if tray: code block completely from serve.py file

PR to fix that: https://github.com/lemonade-sdk/lemonade/pull/148

>I tested it on my PC (Arch Linux) with a 9070xt I just bought and it works fine (and it's super fast as well).

Interesting! I was using a 100% clean install of Ubuntu. Perhaps you have some package installed that I did not. Did you happen to have already installed rocm or similar, before trying Lemonade? cc @danielholanda

sofiageo:
Cool, I forgot to check the PRs before I started digging but it all worked out in the end ! :) 

[...]

PR #829, opened 2026-01-09
Added Linux install instructions

sofiageo:
fixes: #144

### DIFF
--- a/docs/assets/install-selector.js
+++ b/docs/assets/install-selector.js
@@ -17,7 +17,7 @@ const NPU_DRIVER_URL = 'https://account.amd.com/en/forms/downloads/ryzenai-eula-
 const GITHUB_RELEASES_API = 'https://api.github.com/repos/lemonade-sdk/lemonade/releases/latest';
 
 // State: os, type, fw, dev, latestVersion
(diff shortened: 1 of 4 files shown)


docs/assets/install-selector.js
```diff
@@ -17,7 +17,7 @@ const NPU_DRIVER_URL = 'https://account.amd.com/en/forms/downloads/ryzenai-eula-
 const GITHUB_RELEASES_API = 'https://api.github.com/repos/lemonade-sdk/lemonade/releases/latest';
 
 // State: os, type, fw, dev, latestVersion
-window.lmnState = { os: 'win', type: 'app', fw: 'oga', dev: 'npu' };
+window.lmnState = { os: 'win', distro: 'win', type: 'app', fw: 'oga', dev: 'npu' };
```
jeremyfowers, line 20: If I come into install_options.html and click "Linux" it doesn't auto-select a distro, and the download button is still for the MSI.

image

How do you feel about auto-selecting Ubuntu? Alternatively, showing "select a distro to get started" instead of the MSI button.

sofiageo: Yes, auto-selecting Ubuntu is fine

sofiageo:
should be fine now

PR #835, opened 2026-01-10
Add: unversioned /live liveness endpoint & align Docker health check

### LINKED ISSUE #834: Docker: health check fails when LEMONADE_API_KEY is being used
Reported by SzczechzCzech on discord:

hi all, it is great to have the docker image build instructions and with 9.1.3 now the prebuilt image - thanks. One thing I noticed, when LEMONADE_API_KEY is used in the env, then the health check that runs inside the container fails with 401 error. A thought would have to be given also to pass the KEY to the client, or the health check client to use the env variable too.

% Total % Received % Xferd Average Speed Time Time Time Current Dload Upload Total Spent Left Speed 0 0 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0 0 39 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0 curl: (22) The requested URL returned error: 401

This does not happen obviously when API key is not used. I.e. Healthcheck passes.

siavashhub:
fixes #834 

### Problem

When the LEMONADE_API_KEY environment variable is set, all API routes required authentication.
This caused the container to fail its Docker HEALTHCHECK, since the health probe did not (and should not) provide an API key.

### Changes

- `Dockerfile`: Updated Docker HEALTHCHECK to target `/live` instead of `/`
- `src/cpp/include/lemon/server.h`: added `live` handler
- `src/cpp/server/server.cpp`: 
  - Added `/live` endpoint
  - Excluded `/live` from API key authentication and logging

### Result

- Health checks no longer fail when LEMONADE_API_KEY is configured
- Authentication remains enforced for all other API routes
- Live endpoint will act as an unauthenticated liveness probe

### Notes

This aligns with common container orchestration expectations (Docker, Kubernetes, etc.) where live endpoints are intentionally unauthenticated.


Dockerfile
```diff
@@ -70,7 +70,7 @@ EXPOSE 8000
```
jeremyfowers, line 70: Would you kindly add `/live` to docs/server/server_spec.md?

jeremyfowers:
@bitgamma would you mind taking a look here to ensure it matches the patterns you expect for LEMONADE_API_KEY?

src/cpp/server/server.cpp
```diff
@@ -102,6 +102,11 @@ void Server::log_request(const httplib::Request& req) {
 }
 
 httplib::Server::HandlerResponse Server::authenticate_request(const httplib::Request& req, httplib::Response& res) {
+    // Skip auth for live checks
```
bitgamma: the check in `set_pre_routing_handler` is sufficient, I'd remove this duplicated check (or at least it could be integrated in the `if` below)

bitgamma:
everything works as expected on my system

jeremyfowers (changes requested):
Would you kindly add /live to docs/server/server_spec.md?

src/cpp/server/server.cpp
siavashhub: done

Dockerfile
siavashhub: done

siavashhub:
> Would you kindly add /live to docs/server/server_spec.md?

done

docs/server/server_spec.md
```diff
@@ -50,6 +50,7 @@ The additional endpoints are:
 - GET `/api/v1/health` - Check server health
```
jeremyfowers:
```suggestion
- GET `/api/v1/health` - Check server status, such as models loaded
```
To clarify the distinction with the new live endpoint

Dockerfile
```diff
@@ -70,7 +70,7 @@ EXPOSE 8000
 
 # Health check
 HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-    CMD curl -f http://localhost:8000/ || exit 1
+    CMD curl -f http://localhost:8000/live || exit 1
```
jeremyfowers, line 73: Should the docker smoketest call `live` instead to match the documetentation?

This LoC here: https://github.com/lemonade-sdk/lemonade/blob/f80404e764b6eba70f7af4be578128784a8f59e1/.github/workflows/docker-build-smoke-test.yml#L32

jeremyfowers (changes requested):
Two very small requests and then good to merge!

Dockerfile
siavashhub: Yeah sorry I missed that, I will fix it.

siavashhub: fixed

siavashhub:
> Two very small requests and then good to merge!

done

PR #837, opened 2026-01-12
Pack all recipe options into a class

### LINKED ISSUE #832: Refactoring: put recipe options into a structure
Across the codebase the `ctx_size`, `llamacpp_backend` and `llamacpp_args` parameters (collectively known as the recipe options) are handled as separate variables, duplicating the serialization/deserialization logic.

An improvement would be to create a `LlamaCppOptions` structure (or class, tbd) that encapsulates these parameters and handles their serialization/deserialization (mostly to/from json). Later, similar structures could be created for other backends (WhisperCpp, RyzenAI) in case we need to handle custom options for those as well.

The changes proposed must not cause any change to current behavior or API.

After the refactoring is complete, however we could consider wrapping those parameters into a `recipe_options` object also when passing them to the load api (with a transition period where both forms are accepted). This could be desirable since the content of the `recipe_options` object could be made backend-specific (LLamaCpp has its options, RyzanAI might have others, etc).

bitgamma:
Closes #832 

This PR tries to remove all references to `ctx_size`, `llamacpp_backend` and `llamacpp_args` across the codebase with a collective `RecipeOptions` class. The `RecipeOptions` class should be the *only* place where options and especially their default values are defined.

The pros of this approach are:

- Default values are only defined in one place, so they can be changed easily if needed
- If other recipe/load options are added, there there is no need anymore to edit WrappedServer and all backend. Only the backends that need to make use of said options
- The router/server do not need to be aware of options
- The CLI args for recipe options are also handled in `RecipeOptions`

## Future improvements

- Also refactor the tray app (`lemonade-server`). I see that currently the source of the router and the server seem to be completely separated so I didn't want to change this further increasing the scope of this PR. We can discuss on how to change this (another improvement for the tray app would be to also use CLI11 for arg parsing)

### DIFF
--- a/src/cpp/server/main.cpp
+++ b/src/cpp/server/main.cpp
@@ -60,8 +60,7 @@ int main(int argc, char** argv) {
         std::cout << "  Port: " << config.port << std::endl;
         std::cout << "  Host: " << config.host << std::endl;
         std::cout << "  Log level: " << config.log_level << std::endl;
-        //TODO: reinsert
-        //std::cout << "  Context size: " << config.ctx_size << std::endl;
+        std::cout << "  Context size: " << RecipeOptions::get_ctx_size_from_cli_options(config.recipe_options) << std::endl;
         if (!config.extra_models_dir.empty()) {
             std::cout << "  Extra models dir: " << config.extra_models_dir << std::endl;
         }
--- a/src/cpp/include/lemon/recipe_options.h
+++ b/src/cpp/include/lemon/recipe_options.h
@@ -17,6 +17,7 @@ public:
     json get_option(const std::string& opt) const;
 
     static void add_cli_options(CLI::App& app, json& storage);
+    static int get_ctx_size_from_cli_options(json options_) { return RecipeOptions("llamacpp", options_).get_option("ctx_size"); }
 private:
     json options_ = json::object();
     std::string recipe_ = "";
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -718,8 +718,7 @@ void Server::handle_health(const httplib::Request& req, httplib::Response& res) 
     response["max_models"] = router_->get_max_model_limits();
     
     // Add context size
-    //TODO: re-add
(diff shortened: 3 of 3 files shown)


src/cpp/server/main.cpp
```diff
@@ -60,14 +60,13 @@ int main(int argc, char** argv) {
         std::cout << "  Port: " << config.port << std::endl;
         std::cout << "  Host: " << config.host << std::endl;
         std::cout << "  Log level: " << config.log_level << std::endl;
-        std::cout << "  Context size: " << config.ctx_size << std::endl;
+        std::cout << "  Context size: " << RecipeOptions::get_ctx_size_from_cli_options(config.recipe_options) << std::endl;
```
jeremyfowers, line 63: Why is a dedicated function needed to retrieve context size, when other options can be retrieved with the generic get_option() method?

bitgamma: this is just a small convenience function, it invokes get_option internally. Since a goal of this PR was to keep functionality unchanged, the ctx_size returned is the one passed at command line or the default one not the one used to load any model.

My thought is as follows: the CLI options are stored all together, independent of the recipe and thus they are not a RecipeOptions, but a repository of raw options from the command line stored in a JSON object. From that, when loading the model the base RecipeOption object is created, dependent on recipe (so that we know what to log, among other things) and then eventually overridden by model and load options.

As a shortcut for the CTX size which is logged in places that are not related to actual model loading, I created this simple function.

bitgamma: to be clear, this is only used in /health and this log entry, all other places that need ctx_size use get_option normally

jeremyfowers: Got it! I like the goal of minimally disrupting the rest of the codebase.

jeremyfowers (approved):
Tested locally with a few recipe configurations and everything worked as expected :)

This feels like an important refactor to me - it will save a dozen or more files from being touched the next time we add a recipe option. Thanks @bitgamma!

PR #842, opened 2026-01-13
Add strix halo dedicated gpu memory to the FAQ

### LINKED ISSUE #841: Add STX Halo windows memory configuration to FAQ
96 GB dedicated memory means:
- CPU can only access 32 GB
- GPU can access ~112 GB

This starves the CPU of RAM, which can cause model loading to fail.

We recommend to set the dedicated memory to "64/64 (auto)" to prevent these issues. That way
- CPU can access 64 GB
- GPU can access 112 GB

(the dedicated memory only prevents CPU access, it doesn't enhance GPU access)

jeremyfowers:
Closes #841

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -146,7 +146,7 @@
 
    We suggest setting dedicated GPU RAM to `64/64 (auto)`.
 
-   > Note: On Windows, the GPU can access both unified RAM and dedicated GPU RAM, but the CPU is blocked from accessing dedicated GPU RAM. For this reason, allocating too much dedicated GPU RAM can interere with model loading, which requires the CPU to access a substaintial amount unified RAM.
+   > Note: On Windows, the GPU can access both unified RAM and dedicated GPU RAM, but the CPU is blocked from accessing dedicated GPU RAM. For this reason, allocating too much dedicated GPU RAM can interfere with model loading, which requires the CPU to access a substantial amount unified RAM.
 
 
 ## Hybrid and NPU Questions


docs/faq.md
```diff
@@ -140,6 +140,15 @@
 
    File a detailed issue on TheRock repo for support: https://github.com/ROCm/TheRock
 
+### 4. **How should dedicated GPU RAM be allocated on Strix Halo**
+
+   Strix Halo PCs can have up to 128 GB of unified RAM and Windows allows the user to allocated a portion of this to dedicated GPU RAM.
```
ramkrishna2910:
```suggestion
   Strix Halo PCs can have up to 128 GB of unified RAM and Windows allows the user to allocate a portion of this to dedicated GPU RAM.
```

docs/faq.md
```diff
@@ -140,6 +140,15 @@
...
+
+   We suggest setting dedicated GPU RAM to `64/64 (auto)`.
+
+   > Note: On Windows, the GPU can access both unified RAM and dedicated GPU RAM, but the CPU is blocked from accessing dedicated GPU RAM. For this reason, allocating too much dedicated GPU RAM can interere with model loading, which requires the CPU to access a substaintial amount unified RAM.
```
ramkrishna2910:
```suggestion
   > Note: On Windows, the GPU can access both unified RAM and dedicated GPU RAM, but the CPU is blocked from accessing dedicated GPU RAM. For this reason, allocating too much dedicated GPU RAM can interfere with model loading, which requires the CPU to access a substantial amount unified RAM.
```

PR #843, opened 2026-01-13
Add workflows to test and monitor our runners

vgodsoe:
Our self-hosted runners are removed after 14 days of inactivity. This PR adds 2 workflows: 
1) Pulses each self-hosted runner to make sure it's alive. Generates an artifact if it is. This is scheduled to run on Sunday's at 1AM.
2) Pulls the artifacts and checks if any is missing. If one is missing, a message is created to notify the Lemonade team that a runner is down and needs maintenance. This runs Monday at 1AM so reports are ready for review.

### DIFF
--- a/.github/workflows/monitor_selfhosted_runners.yml
+++ b/.github/workflows/monitor_selfhosted_runners.yml
@@ -4,11 +4,6 @@ on:
   schedule:
     - cron: "0 1 * * 1"   # every Monday at 1AM UTC
   workflow_dispatch:
-  push:
-    branches: 
-      - vgodsoe/ci-runner-heartbeat  # Temporary: remove before merging
-    paths:
-      - '.github/workflows/monitor_selfhosted_runners.yml'
 
 permissions:
   actions: read  # Required to read artifacts
--- a/.github/workflows/runner_heartbeat.yml
+++ b/.github/workflows/runner_heartbeat.yml
@@ -1,16 +1,9 @@
 name: Runner Heartbeat
 
-# Change to trigger push.
-
 on:
   schedule:
     - cron: "0 1 * * 0"    # every Sunday at 1AM UTC
   workflow_dispatch:
-  push:
-    branches: 
-      - vgodsoe/ci-runner-heartbeat  # Temporary: remove before merging
-    paths:
-      - '.github/workflows/runner_heartbeat.yml'
 
 permissions:
   actions: write  # Required to upload artifacts


.github/workflows/runner_heartbeat.yml
```diff
@@ -0,0 +1,27 @@
...
+  schedule:
+    - cron: "0 1 * * 0"    # every Sunday at 1AM UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  heartbeat:
+    runs-on: [self-hosted]
```
jeremyfowers: Won't this run on exactly 1 runner each week?

You might need to spawn a number of `jobs` equal to the number of runners to make sure they all get hit (and keep that counter up to date).

vgodsoe: Totally right. fixed in a recent push.

.github/workflows/monitor_selfhosted_runners.yml
```diff
@@ -0,0 +1,133 @@
...
+
+permissions:
+  actions: read  # Required to read artifacts
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get list of runners
```
jeremyfowers, line 16: If a runner gets de-registered does that mean it wont appear in this list, which in turn means we wont check its heartbeat?

.github/workflows/monitor_selfhosted_runners.yml
```diff
@@ -0,0 +1,133 @@
...
+            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            https://api.github.com/orgs/${{ github.repository_owner }}/actions/runners \
+            > org_runners.json
+          
+          # Combine and deduplicate runner names
+          repo_names=$(jq -r '.runners[]?.name // empty' repo_runners.json 2>/dev/null || echo "")
+          org_names=$(jq -r '.runners[]?.name // empty' org_runners.json 2>/dev/null || echo "")
+          
+          # Fallback to known runner list if API fails
+          known_runners="jfowers-stx-01,jfowers-stx-02,sjlab-stx-halo-06,sjlab-stx-halo-07,sjlab-stx-halo-08,sjlab-stx-halo-09,sjlab-stx-halo-10,sjlab-stx-halo-11,sjlab-stx-halo-12,sjlab-stx-halo-13,sjlab-stx-0"
```
jeremyfowers, line 36: two lists to maintain

vgodsoe: only if the API fails, is this list used. Will look at better optimizations in future iterations.

.github/workflows/monitor_selfhosted_runners.yml
vgodsoe: Correct, which means we have <14 days to fix a broken runner. The notification we get will let us know what runners died if we wait longer than 14 days.

PR #854, opened 2026-01-14
Doc Docker documentation

### LINKED ISSUE #823: Docker documentation follow up
From #803, that we should resolve before too long:
1. "Quick Start with Docker" contents on the top-level README could be moved to their own document to keep the top README more concise
2. Discussion: is there any redundancy between the checked-in Dockerfile and build flow vs. the DOCKER_GUIDE.md contents @siavashhub ?

siavashhub:
Resolves #823


docs/assets/install-selector.js
```diff
@@ -235,6 +241,51 @@ function renderDownload() {
...
+
+    if (installType) installType.style.display = 'none';
+
+    if (downloadArea) {
+      downloadArea.style.display = 'none';
+    }
+    if (installCmdDiv) {
+      installCmdDiv.style.display = 'block';
+      const commands = [
+        `docker run -d \\`,
```
jeremyfowers, line 259: The on-hover "copy code" button copies only one line of this command. Ideally it should copy the whole command.

Would you like to add some special handling for copying a whole multi-line command? Or I can in a follow-up.

docs/assets/install-selector.js
```diff
@@ -85,7 +85,7 @@ window.lmnRender = function() {
```
jeremyfowers, line 88: image

The Quick Start section is totally disabled when Docker is selected, which means that the NPU quick start instructions are displayed.

What backends actually do work with docker? I'm sure CPU does. Does llamacpp Vulkan work as well?

Ideally:
1. Supported configurations are click-able
2. A good default configuration is auto-selected when Docker is clicked, so that relevant quick start instructions are displayed

jeremyfowers (changes requested):
Please see comments on install-selector.js

The changes to README.md and DOCKER_GUIDE.md look good to me :)

docs/assets/install-selector.js
siavashhub: Sure, I will update it tomorrow.

siavashhub: Sure, I will fix it tomorrow.

siavashhub: Sure, I will fix it tomorrow.

siavashhub: > The on-hover "copy code" button copies only one line of this command. Ideally it should copy the whole command.
> 
> Would you like to add some special handling for copying a whole multi-line command? Or I can in a follow-up.

done

docs/assets/install-selector.js
siavashhub: > The Quick Start section is totally disabled when Docker is selected, which means that the NPU quick start instructions are displayed.
> 

fixed

> What backends actually do work with docker? I'm sure CPU does. Does llamacpp Vulkan work as well?
> 

It should work with Vulkan, later I can work on adding rocm as well.

> Ideally:
> 
> 1. Supported configurations are click-able
> 2. A good default configuration is auto-selected when Docker is clicked, so that relevant quick start instructions are displayed

done

siavashhub:
> Please see comments on install-selector.js
> 
> The changes to README.md and DOCKER_GUIDE.md look good to me :)

all done

PR #857, opened 2026-01-15
remove checkpoint_loaded and context_size from /health and add recipe and options in `all_models_loaded`

### LINKED ISSUE #824: Return the backend in the `all_models_loaded` field of `/health`
Needed to show a model's backend in the App

bitgamma:
Closes #824 

Example output:

```
{
    "all_models_loaded": [
        {
            "backend_url": "http://127.0.0.1:8001/v1",
            "checkpoint": "unsloth/Qwen3-0.6B-GGUF:Q4_0",
            "device": "gpu",
            "last_use": 4343774,
            "model_name": "Qwen3-0.6B-GGUF",
            "recipe": "llamacpp",
            "recipe_options": {
                "ctx_size": 8192,
                "llamacpp_args": "--no-mmap",
                "llamacpp_backend": "rocm"
            },
            "type": "llm"
        }
    ],
    "log_streaming": {
        "sse": true,
        "websocket": false
    },
    "max_models": {
        "embedding": 1,
        "llm": 1,
        "reranking": 1
    },
    "model_loaded": "Qwen3-0.6B-GGUF",
    "status": "ok",
    "version": "9.1.4"
}
```

### DIFF
--- a/docs/server/server_spec.md
+++ b/docs/server/server_spec.md
@@ -47,9 +47,10 @@ The additional endpoints are:
 - POST `/api/v1/delete` - Delete a model
 - POST `/api/v1/load` - Load a model
 - POST `/api/v1/unload` - Unload a model
-- GET `/api/v1/health` - Check server health
+- GET `/api/v1/health` - Check server status, such as models loaded
 - GET `/api/v1/stats` - Performance statistics from the last request
 - GET `/api/v1/system-info` - System information and device enumeration
+- GET `/live` - Check server liveness for load balancers and orchestrators
 
 ## Multi-Model Support
 
--- a/.github/workflows/docker-build-smoke-test.yml
+++ b/.github/workflows/docker-build-smoke-test.yml
@@ -29,7 +29,7 @@ jobs:
         run: |
           echo "Waiting for Lemonade to start..."
           for i in {1..30}; do
-            if curl -sf http://localhost:8000/ > /dev/null; then
+            if curl -sf http://localhost:8000/live > /dev/null; then
               echo "Server is up"
               exit 0
             fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,7 +70,7 @@ EXPOSE 8000
 
 # Health check
 HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-    CMD curl -f http://localhost:8000/ || exit 1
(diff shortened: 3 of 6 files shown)


jeremyfowers:
> BTW: is the context_size value outside of all_models_loaded still needed? I see it is not a documented field in server_spec.md and its value is just the default that will be applied when loading a new model if no load or model-specific option exists

it was there for backwards compatibility to the days when we could only load one model at a time. Similar to a couple other fields there. I’d be comfortable to remove those soon since they’ve been deprecated for at least a month at this point, and they aren’t good to use.

bitgamma:
so should we remove those (I guess you are also talking about `checkpoint_loaded` and `model_loaded`) as part of this PR or do we do that separately at a later time?

jeremyfowers:
> so should we remove those (I guess you are also talking about `checkpoint_loaded` and `model_loaded`) as part of this PR or do we do that separately at a later time?

I’m fine with removing those 3 in this PR if you want. You might fail some tests, which would then need to be updated too.

bitgamma:
> > so should we remove those (I guess you are also talking about `checkpoint_loaded` and `model_loaded`) as part of this PR or do we do that separately at a later time?
> 
> I’m fine with removing those 3 in this PR if you want. You might fail some tests, which would then need to be updated too.

I removed `checkpoint_loaded` and `context_size`. Regarding `model_loaded`, I see that beside tests it is also used by the Electron app to select the default model in the chat window, which seems a reasonable use case.

docs/server/server_spec.md
```diff
@@ -1014,6 +1014,10 @@ curl http://localhost:8000/api/v1/health
       "last_use": 1732123456.789,
```
jeremyfowers, line 1014: "checkpoint_loaded" is still in the example response a few lines above this one, please remove.

bitgamma: oops

jeremyfowers: If you dont mind editing the PR description as well that would be helpful. I'm going to send to folks so they can review the breaking change.

jeremyfowers: Actually I have the power to do it myself! Thanks for the doc fix.

jeremyfowers:
> Regarding `model_loaded`, I see that beside tests it is also used by the Electron app to select the default model in the chat window, which seems a reasonable use case.

Good catch!

PR #862, opened 2026-01-16
fix recipe options creation for newly added models

bitgamma:
Fixes the following issue:

if adding a new model through GUI or pull command and running it immediately the default options will be applied, regardless of command line switches or existing recipe options. This fixes it


jeremyfowers (approved):
Did not test locally, but change looks reasonable. Thanks @bitgamma !

PR #874, opened 2026-01-20
unify CLI parsing between router and server

### LINKED ISSUE #771: Ensure the CLI raises an error on invalid args
For example, if someone puts “—rocm”

Is there a library we could use for this instead of coding for every corner case?

bitgamma:
Closes #771

Use the CLI11 library for argument/environment variable parsing in both router and server

Pros:
- Robust CLI handling
- All options and default defined in a single place
- Run commands options are not ignored anymore if the server was already running

Possible follow up PR:
- Ideally ServerManager should forward the entire `ServerConfig` struct without having to know the flag names or meaning.

### DIFF
--- a/src/cpp/server/recipe_options.cpp
+++ b/src/cpp/server/recipe_options.cpp
@@ -65,7 +65,7 @@ void RecipeOptions::add_cli_options(CLI::App& app, json& storage) {
         o->envname(opt["envname"]);
         o->type_name(opt["type_name"]);
         if (opt.contains("allowed_values")) {
-            o->check(CLI::IsMember((const std::vector<std::string>)opt["allowed_values"]));
+            o->check(CLI::IsMember(opt["allowed_values"].get<std::vector<std::string>>()));
         }
     }
 }


jeremyfowers:
@bitgamma I marked the PR as a draft since you have TODOs in the description. Please mark Ready for Review when its time for me to take a look. Or LMK if you want early feedback.

bitgamma:
@jeremyfowers ready for review

src/cpp/include/lemon/cli_parser.h
```diff
@@ -24,6 +23,30 @@ struct ServerConfig {
     int max_audio_models = 1;
 };
 
+struct TrayConfig {
+    std::string command;  // No default - must be explicitly specified
+    // Default to headless mode on Linux (no tray support), tray mode on other platforms
+#if defined(__linux__) && !defined(__ANDROID__)
```
jeremyfowers, line 29: Android?

src/cpp/include/lemon/cli_parser.h
```diff
@@ -24,6 +23,30 @@ struct ServerConfig {
...
+    bool save_options = false;
+
+    // Pull options
+    std::string checkpoint = "";
+    std::string recipe = "";
+    std::string mmproj = "";
+    bool is_reasoning = false;
+    bool is_vision = false;
+    bool is_embedding = false;
+    bool is_reranking = false;
```
jeremyfowers, line 47: I know we have a bit of a naming soup at the moment in the codebase at large.

But why would pull options be part of a "TrayConfig"?

src/cpp/server/cli_parser.cpp
```diff
@@ -52,8 +78,63 @@ CLIParser::CLIParser()
...
+    // Serve
+    CLI::App* serve = app_.add_subcommand("serve", "Start the server");
+    add_serve_options(serve, config_, max_models_vec_);
+    serve->add_flag("--no-tray", tray_config_.no_tray, "Start server without tray (headless mode, default on Linux)");
+
+    // Run
+    CLI::App* run = app_.add_subcommand("run", "Run a model");
+    run->add_option("model", tray_config_.model, "The model to run")->required();
+    add_serve_options(run, config_, max_models_vec_);
+    run->add_flag("--no-tray", tray_config_.no_tray, "Start server without tray (headless mode, default on Linux)");
```
jeremyfowers, line 102: This is manually added to both serve and run options, could it just be in `add_serve_options()`?

src/cpp/server/cli_parser.cpp
```diff
@@ -52,8 +78,63 @@ CLIParser::CLIParser()
...
+    del->add_option("model", tray_config_.model, "The model to delete")->required();
+
+    // Status
+    CLI::App* status = app_.add_subcommand("status", "Check server status");
+
+    // Stop
+    CLI::App* stop = app_.add_subcommand("stop", "Stop the server");
+#else
+    add_serve_options(&app_, config_, max_models_vec_);
+#endif
```
jeremyfowers, line 137: I'm not sure I understand the if/else here. It's saying if we're on a no-tray platform (Linux) then the pull/delete/run/list/status/stop commands are not available?

src/cpp/include/lemon/cli_parser.h
bitgamma: I just moved these lines as they were, I can remove Android

src/cpp/include/lemon/cli_parser.h
bitgamma: cause these are handled by the tray app (lemonade-server). I could put them separately but not sure if it is really worth it. Should I rename TrayConfig to something else?

src/cpp/server/cli_parser.cpp
bitgamma: add_serve_options is also used for router, which doesn't need the tray flag

src/cpp/server/cli_parser.cpp
bitgamma: no this is the "tray app" (lemonade-server) vs router

src/cpp/include/lemon/cli_parser.h
jeremyfowers: I had a branch where I renamed this kind of thing to be easier to understand, but ran out of bandwidth to fully test it :/

I should definitely revisit that later. In the meantime, this is fine, no changes needed.

src/cpp/server/cli_parser.cpp
jeremyfowers: gotcha, thanks!

src/cpp/include/lemon/cli_parser.h
jeremyfowers: no worries

PR #884, opened 2026-01-21
Improve env var and customization documentation

### LINKED ISSUE #875: Add to FAQ: how do I use my own llamacpp or whispercpp binaries?

jeremyfowers:
Closes #875

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -85,7 +85,7 @@
    Lemonade supports a wide range of LLMs including LLaMA, DeepSeek, Qwen, Gemma, Phi, gpt-oss, LFM, and many more. Most GGUF models can also be added to Lemonade Server by users using the Model Manager interface in the app or the `pull` command on the CLI.
    
    👉 [Supported Models List](https://lemonade-server.ai/docs/server/server_models/)
-   👉 [pull command](https://lemonade-server.ai/docs/server/lemonade-server-cli/#pull-command-options)
+   👉 [pull command](https://lemonade-server.ai/docs/server/lemonade-server-cli/#options-for-pull)
 
 ### 3. **How do I know what size model will work with my setup?**
 
@@ -103,7 +103,7 @@
    
    You can:
 
-   - Add a custom model manually via the app's "Add a Model" interface or the [CLI pull command](https://lemonade-server.ai/docs/server/lemonade-server-cli/#pull-command-options).
+   - Add a custom model manually via the app's "Add a Model" interface or the [CLI pull command](https://lemonade-server.ai/docs/server/lemonade-server-cli/#options-for-pull).
    - Use a pull request to add the model to the built-in `server_models.json` file.
    - Request support by opening a [GitHub issue](https://github.com/lemonade-sdk/lemonade/issues).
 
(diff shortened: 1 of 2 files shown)


prodkt:
@jeremyfowers Implemented bad url in this pr.

Page: https://lemonade-server.ai/docs/server/server_integration/#installing-additional-models
Bad url can be found at "[A JSON file with the list of supported models](https://github.com/lemonade-sdk/lemonade/tree/main/src/lemonade_server/server_models.json) is included in every Lemonade Server installation."

What should also be very clearly called out is any customization via this recommended path will be removed upon any update or new install of lemonade server. Give the chance for those readers to backup beforehand.

More ideal would be having the ability to set an extra path that could be searched against and indexed if set, and then if the schema matches that of the server_models.json  and passes standard syntax validation. Minor things like prepend any additional model found within this custom json with "user" to match the result of what doing this via the UI would output to the user. This way there be no reason for a backup or headache of reimplementation upon loss. If there be breaking changes as far as changes to the schema down the line I don't see this an issue calling it out in the release notes - those of whom add many additional models via custom JSON will have no problem updating in adherence --- far more ideal than any potential loss of data or disruption an overwrite would cost.

jeremyfowers:
> @jeremyfowers Implemented bad url in this pr.
> 
> Page: https://lemonade-server.ai/docs/server/server_integration/#installing-additional-models Bad url can be found at "[A JSON file with the list of supported models](https://github.com/lemonade-sdk/lemonade/tree/main/src/lemonade_server/server_models.json) is included in every Lemonade Server installation."
> 
> What should also be very clearly called out is any customization via this recommended path will be removed upon any update or new install of lemonade server. Give the chance for those readers to backup beforehand.
> 
> More ideal would be having the ability to set an extra path that could be searched against and indexed if set, and then if the schema matches that of the server_models.json and passes standard syntax validation. Minor things like prepend any additional model found within this custom json with "user" to match the result of what doing this via the UI would output to the user. This way there be no reason for a backup or headache of reimplementation upon loss. If there be breaking changes as far as changes to the schema down the line I don't see this an issue calling it out in the release notes - those of whom add many additional models via custom JSON will have no problem updating in adherence --- far more ideal than any potential loss of data or disruption an overwrite would cost.

Hi @prodkt thanks for pointing out the broken link! I have made a PR to fix it, and overhaul that whole section, here: https://github.com/lemonade-sdk/lemonade/pull/993

> What should also be very clearly called out is any customization via this recommended path will be removed upon any update or new install of lemonade server. Give the chance for those readers to backup beforehand.

I have made it clear on that PR that the `server_models.json` file should **never** be modified.

FYI we provide a `user_models.json` file that works exactly like `server_models.json`, except that it is there for you to modify and it persists across updates. You can also automatically add models to `user_models.json` with the `/pull` endpoint, `lemonade-server pull` command, and the Add a Model form in the app.

prodkt:
@jeremyfowers For sure! I've tried to use the user_models.json in the past. Unfortunately in my specific & personal use case I heavily rely on workflows/agents/context I've built out in Dify(self-hosted). At the moment models added via user_models aren't seen as available or acceptable via model name when attempting to add them in Dify Model Provider/Lemonade Plugin. I am though able to consume all my user models if I pack them within server_models. For little old me this workaround is hardly a hassle when updates comes available I've got my process. Still, I can imagine with Lemonades increasing popularity means active users☝️, soon enough they'll be others. :)

Thank you, sir!

Issue #722, opened 2025-12-12
Physical RAM is reported incorrectly on Linux


jeremyfowers:
Reported by NineBall on discord

"model usable ram is 32+60GB"

however, system info is reporting 9.7 GB available RAM.

System info:

```
{"OS Version":"Linux-6.12.57+deb13-amd64 (Debian GNU/Linux 13)","Physical Memory":"9.70 GB","Processor":"AMD RYZEN AI MAX+ 395 w/ Radeon 8060S","devices":{"amd_dgpu":[{"available":false,"error":"No AMD discrete GPU found","name":""}],"amd_igpu":{"available":true,"inference_engines":{"llamacpp-rocm":{"available":false,"error":"rocm binaries not installed"},"llamacpp-vulkan":{"available":false,"error":"vulkan binaries not installed"}},"name":"Advanced Micro Devices, Inc. [AMD/ATI] Strix Halo [Radeon Graphics / Radeon 8050S Graphics / Radeon 8060S Graphics] (rev c1)"},"cpu":{"available":true,"cores":16,"inference_engines":{"llamacpp-vulkan":{"available":false,"error":"vulkan binaries not installed"}},"name":"AMD RYZEN AI MAX+ 395 w/ Radeon 8060S","threads":16},"npu":{"available":false,"error":"NPU detection not yet implemented for Linux","inference_engines":{"oga":{"available":false}},"name":"AMD NPU"},"nvidia_dgpu":[{"available":false,"error":"No NVIDIA discrete GPU found","name":""}]}}
```

<img width="732" height="235" alt="Image" src="https://github.com/user-attachments/assets/91f3231b-e05f-4c05-b9da-fd15e2967fee" />

<img width="988" height="95" alt="Image" src="https://github.com/user-attachments/assets/a1db2afc-dbd9-4fc4-88cd-9f5dab224563" />

NineBallo:
Just adding the update here. It is detecting system ram correctly but does not account for the VRAM in this calculation.

The primary issue in this case was that the VM was setup with 10GB of ram, then updated to 60GB. This didn't trigger a regeneration of the hardware_info.json and thus it was very very incorrect.

An easy fix could just be regenerating this file on startup. As for accounting for the VRAM, that is a separate issue. 

```{
  "hardware": {
    "OS Version": "Linux-6.17.11-061711-generic (Debian GNU/Linux 13)",
    "Physical Memory": "57.45 GB",
[...]

PR #901, opened 2026-01-21
Add AMD GTT detection support

NineBallo:
Copied from #722 for visibility, since this is working. 

What do you think about this sort of thing?

I've tried to keep it fairly consistent with how we currently handle things. However my idea is that we will eventually move "Physical Memory" into the "cpu" section of the devices and treat all devices as having some form of attached memory. This idea could also make sense in the future when dealing with multi-gpu vram-pools.

I currently only added AMD support on linux. Nvidia does it's own memory management to my understanding. I did add a general TTM-pages check. This should be used for general linux devices, but this is only really relevant for intel chips for now. Because we don't really support that right now it's unused as the AMD specific one is more reliable IMO.

I also cleaned up and sped up some of the checks. It does the full device scan in under a second so it may be viable to just regenerate it each startup. The trick I'm guessing is also getting the windows speed to be similar, as @jeremyfowers had mentioned it takes >15 seconds in some cases.

The ram check specifically is now one file call so it's effectively instant for this use case.


src/cpp/server/system_info.cpp
```diff
@@ -79,6 +79,8 @@ json SystemInfo::get_device_dict() {
         auto amd_igpu = get_amd_igpu_device();
         devices["amd_igpu"] = {
             {"name", amd_igpu.name},
+            {"vram_gb", amd_igpu.vram_gb},
+            {"dynamic_mem_gb", amd_igpu.dynamic_gb},
```
superm1: not my place to make this call, but as a bystander to the project with a few PR's open I don't like `dynamic_mem_gb`.  It matches nothing in the kernel or sysfs files.  I would rather see it as `gtt_gb`.

src/cpp/server/system_info.cpp
```diff
@@ -1292,18 +1297,24 @@ std::vector<GPUInfo> LinuxSystemInfo::detect_amd_gpus(const std::string& gpu_typ
...
+
+                // Get Dynamic VRAM for device
+                double dynamic_vram = get_amd_dynamic_vram(pci_id);
+
+                if (vram > 0.0) {
+                    gpu.vram_gb = vram;
+                }
+
+                if (dynamic_vram > 0.0) {
+                    gpu.dynamic_gb = dynamic_vram;
```
superm1: FYI this is going to clash with my PR https://github.com/lemonade-sdk/lemonade/pull/419.  So whichever of us gets merged first (if any) the other will need to update.

src/cpp/server/system_info.cpp
```diff
@@ -1428,14 +1439,35 @@ double LinuxSystemInfo::get_amd_vram_rocm_smi() {
...
+    }
+
+    std::string vram_str;
+    std::getline(file, vram_str);
+    file.close();
+    
+    try {
+        uint64_t vram_bytes = std::stoull(vram_str);
+        return std::round(vram_bytes / (1024.0 * 1024.0 * 1024.0) * 10.0) / 10.0;
+    } catch (...) {
```
superm1, line 1429: AFAICT this math isn't right.  The unit for pages_limit is pages, but you're not doing anything with `PAGE_SIZE` (4096).

Here is what I use in `amd-ttm`:
https://github.com/superm1/amd-debug-tools/blob/master/src/amd_debug/common.py#L233

src/cpp/server/system_info.cpp
```diff
@@ -1862,5 +1883,4 @@ json SystemInfoCache::get_system_info_with_cache(bool verbose) {
 }
 
 
-} // namespace lemon
-
+} // namespace lemon
```
superm1: missing newline

src/cpp/server/system_info.cpp
NineBallo: I originally had the same idea, however the project aims to be device agnostic and if we open it to non-graphics devices gtt doesn't make sense. I'm open for other suggestions here cause I also don't like the current naming but I don't think gtt is the ideal move here. 

Even with CPU models for example we would already have this issue. If we have CXL memory what would we consider that? Its not necessarily system ram but its not GTT either. I named it so it would still make sense in cases like that.

src/cpp/server/system_info.cpp
NineBallo: Yeah I saw that, changes should be simple in either case but if you could give me a ping if yours is merged first I'd appreciate it and I'll do the same

src/cpp/server/system_info.cpp
superm1: Got it; I see what you mean.  Considering that background I think your name selection makes a lot of sense.

src/cpp/server/system_info.cpp
```diff
@@ -1428,14 +1439,35 @@ double LinuxSystemInfo::get_amd_vram_rocm_smi() {
...
+    std::ifstream file(ttm_path);
+
+    if (!file.is_open()) {
+        return 0.0;
+    }
+
+    std::string vram_str;
+    std::getline(file, vram_str);
+    file.close();
+
```
superm1: this shouldn't be called `vram_str`, that's pretty confusing (I think it's from copy paste).

NineBallo: Yes that is correct. It's just taken from the old mem function. I'll take a bit of a better look at the TTM code and see if I can resolve the other concern with it you had.

src/cpp/server/system_info.cpp
NineBallo: Fixed

src/cpp/server/system_info.cpp
NineBallo: Good catch, fixed it and changed up the naming. That function isn't really used right now, and it honestly didn't get nearly enough attention. Unsure if it will be kept in the final pull; going to see what the maintainers think of it since it is only really relevant for intel/misc vendors to my understanding?

src/cpp/server/system_info.cpp
NineBallo: Using page_limit and page_limit_str now. Should be a bit easier to understand

src/cpp/server/system_info.cpp
```diff
@@ -1428,14 +1439,35 @@ double LinuxSystemInfo::get_amd_vram_rocm_smi() {
...
     // Try device-specific path first
-    std::string vram_path = "/sys/bus/pci/devices/" + pci_id + "/mem_info_vram_total";
+    std::string vram_path = "/sys/bus/pci/devices/" + pci_id + "/" + deviceFile;
     std::ifstream file(vram_path);
     
     if (!file.is_open()) {
         // Try wildcard path
-        FILE* pipe = popen("cat /sys/class/drm/card*/device/mem_info_vram_total 2>/dev/null | head -1", "r");
+        std::string wildcard_path = "cat /sys/class/drm/card*/device/" + deviceFile + " 2>/dev/null | head -1";
+        FILE* pipe = popen(wildcard_path.c_str(), "r");
```
superm1: this feels a bit hacky to shell out to `cat`.  Could you do it in a native way where the file is opened/read/closed instead?

NineBallo: This entire fallback section is a bit hacky generally, it was part of the original function but should we look at just removing it? I pushed a change either way but ideally this code path isn't touched in prod as it just grabs the first set of memory it can find, not necessarily matching the device in question.

superm1:
Okay my PR merged, can you please rebase on it?

NineBallo:
@superm1 I noticed a bug in your implementation, not a massive one but you are doing the isIntegrated check, then setting the device_type within that block. I fixed it during the merge though. 

I moved your code for vram into the general parse_vram_sysfs helper as it makes more sense there, and cleaned up that weird fallback path entirely. I also updated the names to reflect everything a bit better.
I also renamed the get_amd_vram_sysfs to just get_amd_vram as your code is now our default pathway.

I also moved the isIntegrated check into it's own file, as again I think it makes a bit more sense to be separated given everything else. 

Let me know if you see any issues, I don't love the new function name "get_amd_is_igpu" . I'm stuck on how else to word it while still sticking to convention. 

This does leave get_amd_vram_rocm_smi() hanging, but I imagine we can trim that off later if it ends up having no real use.

NineBallo:
Just noticed a bug, it's now throwing the device in both dgpu and igpu json locations. Fixing now sorry

NineBallo:
Noticed something else in the upstream code: you set the isIntegrated check to pass with the following conditions `
!fs::exists(board_info_path) && fs::is_regular_file(board_info_path)` I'm presuming you meant something more like `!(fs::exists(board_info_path) && fs::is_regular_file(board_info_path))`?

Also the behaviour of the gpu.isEmpty() check on upstream lines 1346-1360 don't seem right, don't we want to do that check right before returning?

I moved the gpu.isEmpty() to where I think it should go instead, and fixed the isIntegrated check as described. I also am properly filtering igpu from dgpu devices. Should be good for review now.

src/cpp/server/model_manager.cpp
```diff
@@ -1055,6 +1056,22 @@ static double parse_physical_memory_gb(const std::string& memory_str) {
     return 0.0;
 }
 
+
+double get_max_memory_of_device(json device, std::optional<std::reference_wrapper<double>> current_max = std::nullopt) {
```
superm1: I'm not sure I agree with the premise of this function.  I don't believe KFD will let you schedule work on both GTT or VRAM.  It's going to pick whichever is bigger for this purpose.  But you don't get a giant pool of both IIUC.

I feel this function should just return the bigger of the two.

src/cpp/server/system_info.cpp
```diff
@@ -1457,34 +1452,44 @@ double LinuxSystemInfo::get_amd_vram_rocm_smi() {
...
+    }   
+}
+
+bool LinuxSystemInfo::get_amd_is_igpu(const std::string& drm_render_minor) {
+    std::string device_path = "/sys/class/drm/renderD" + drm_render_minor + "/device/";
+    std::string board_info_path = device_path + "board_info";
+    return !(fs::exists(board_info_path) && fs::is_regular_file(board_info_path));
+}
+
+double LinuxSystemInfo::parse_vram_sysfs(const std::string& drm_render_minor, const std::string& deviceFile){
```
superm1: I think this function should be renamed, these are not vram files it processes since it's getting memory info for both types.

Likewise the variable used in the function should not be vram.

Maybe `parse_amdgpu_mem_sysfs()` for the function name?

I also disagree with the argument being `deviceFile`.  That makes me think it's looking for a `/dev/dri` node.  Could you maybe make it `sysfs_file` or `fname` or something more representative.

src/cpp/server/system_info.cpp
```diff
@@ -1493,6 +1498,14 @@ double LinuxSystemInfo::get_amd_vram_sysfs(const std::string& pci_id) {
     }
 }
 
+double LinuxSystemInfo::get_amd_dynamic_vram(const std::string& drm_render_minor){
```
superm1: It's not `dynamic_vram` though.  since this is an AMD method could it be `get_amd_gtt`?

superm1:
> @superm1 I noticed a bug in your implementation, not a massive one but you are doing the isIntegrated check, then setting the device_type within that block. I fixed it during the merge though.
> 

Thanks.

> I moved your code for vram into the general parse_vram_sysfs helper as it makes more sense there, and cleaned up that weird fallback path entirely. I also updated the names to reflect everything a bit better. I also renamed the get_amd_vram_sysfs to just get_amd_vram as your code is now our default pathway.
> 
> I also moved the isIntegrated check into it's own file, as again I think it makes a bit more sense to be separated given everything else.
> 
> Let me know if you see any issues, I don't love the new function name "get_amd_is_igpu" . I'm stuck on how else to word it while still sticking to convention.
> 
> This does leave get_amd_vram_rocm_smi() hanging, but I imagine we can trim that off later if it ends up having no real use.

Axe it - I don't see a use for it when we're getting it from real sysfs files.  Those are always there.

src/cpp/include/lemon/system_info.h
```diff
@@ -116,13 +117,19 @@ class LinuxSystemInfo : public SystemInfo {
...
     std::vector<GPUInfo> detect_amd_gpus(const std::string& gpu_type);
     std::string get_nvidia_driver_version();
     double get_nvidia_vram();
     double get_amd_vram_rocm_smi();
-    double get_amd_vram_sysfs(const std::string& pci_id);
+    double get_amd_vram(const std::string& drm_render_minor);
+    double get_amd_dynamic_vram(const std::string& drm_render_minor);
+    bool get_amd_is_igpu(const std::string& drm_render_minor);
+
+private:
```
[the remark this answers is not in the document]
NineBallo: This was done to separate purely internal helper function, from actual externally usable functions. I could just move it out of the namespace but then it loses the linux-specific designation so I think it's best to stay like this

src/cpp/server/model_manager.cpp
```diff
@@ -1087,14 +1104,31 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
...
-    
+
+    // Get largest VRAM object for memory-based filtering
+    double largest_vram_target_gb = 0.0;
+    for (const auto& [dev_type, dev_type_info] : hardware.items()) {
+
+        // Possible multiple entries per type
+        if (dev_type_info.is_array()) {
+            for (const auto& dev : dev_type_info) {
+                get_max_memory_of_device(dev, largest_vram_target_gb);
```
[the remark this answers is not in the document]
NineBallo: I think copilot is just wrong here?

src/cpp/server/model_manager.cpp
```diff
@@ -1087,14 +1104,31 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
...
+            }
+        }
+        else {
+            get_max_memory_of_device(dev_type_info, largest_vram_target_gb);
```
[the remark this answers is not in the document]
NineBallo: Again its just wrong idk

src/cpp/server/model_manager.cpp
```diff
@@ -1116,6 +1150,9 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
             std::cout << "  - System RAM: " << std::fixed << std::setprecision(1) << system_ram_gb
```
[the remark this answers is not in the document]
NineBallo: This was how Jeremy F did it, I just kept the same naming scheme, don't see an issue here.

src/cpp/server/model_manager.cpp
NineBallo: I thought it could go across for Vulkan at least, I think rocm is a bit more strict. I'll do a test tomorrow and update it accordingly if that's the case.

src/cpp/include/lemon/system_info.h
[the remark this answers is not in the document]
superm1: Got it.  This makes sense.

src/cpp/server/model_manager.cpp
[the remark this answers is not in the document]
superm1: > I think copilot is just wrong here?

Yeah, I think it's too far here.  We can double check compiler doesn't complain about it, but I agree you're probably right.

src/cpp/server/model_manager.cpp
superm1: Yeah; my comment was specifically with KFD (what ROCm uses) in mind.

NineBallo: I tested and vulkan does span both. So I'll have to figure something out to make this make sense on both backends. I'm thinking of two pathways depending on the backend in question, just because the check will restrict visible models, so defaulting to the more restrictive option is not ideal.

src/cpp/server/system_info.cpp
NineBallo: Yeah that's a good idea, I was trying to stay away from gtt just because it is sorta superseded by ttm (to end users at least), but I think everyone working on the project should be able to get the gist of what we are going for there. I'll add the rename to the next commit here.

src/cpp/server/system_info.cpp
NineBallo: Renamed

src/cpp/server/model_manager.cpp
NineBallo: This is a bit weird, I added a toggle to either take the largest or to take the pool. I don't think we have a great way of determining which backend is currently in use, just that it is llama.cpp (of some kind). I think we should keep it like this for now, and when we expand the definitions there is already code to account for it. Especially as this is just a restrictor on what models are available and I'd prefer OOM to not giving users the option, especially because vulkan is the default right now. (and most devices with vram have no gtt, and vice-versa)

I want to make another PR and move all the string definitions over to enums, because it's getting to be a lot lol. I'm thinking that separation might be better done there, along with adding a llamacpp-vk, and llamacpp-rocm, instead of just the general base we have now. 

I was supposed to sleep like 4 hours ago though... so I may just be missing something obvious when figuring out if we are using vk or rocm.

superm1: > I want to make another PR and move all the string definitions over to enums, because it's getting to be a lot lol. I'm thinking that separation might be better done there, along with adding a llamacpp-vk, and llamacpp-rocm, instead of just the general base we have now.

OK.

> I tested and vulkan does span both. So I'll have to figure something out to make this make sense on both backends. I'm thinking of two pathways depending on the backend in question, just because the check will restrict visible models, so defaulting to the more restrictive option is not ideal.

That's interesting.  I wonder what it means for a dGPU then?  Because on a dGPU the performance is **significantly** different for GTT and VRAM. 

> Especially as this is just a restrictor on what models are available and I'd prefer OOM to not giving users the option, especially because vulkan is the default right now. (and most devices with vram have no gtt, and vice-versa)

Well you can get GTT on a dGPU and I'm really worried about implications for dGPU by making changes that expose people to poor performance.

src/cpp/server/model_manager.cpp
```diff
@@ -1055,6 +1056,37 @@ static double parse_physical_memory_gb(const std::string& memory_str) {
     return 0.0;
 }
 
+
+double get_max_memory_of_device(json device, bool unifiedMemory, std::optional<std::reference_wrapper<double>> current_max = std::nullopt) {
```
superm1: Why is this returning an _unused_ double but also setting a reference?  It feels more logical to me to drop the second argument and just return the double.

The two callers can just take the return value then.

src/cpp/server/model_manager.cpp
NineBallo: I agree about the performance, but setting up GTT is a manual process I thought? At least on the STX Halo and Debian 13 I needed to set it through ttm pages. If they have that set I think it's valid for it to be accounted for as they might be intentionally pooling with system ram to run larger models.

I'll look at adding a warning if using a dGPU and pooling with system ram?

superm1: > I agree about the performance, but setting up GTT is a manual process I thought? At least on the STX Halo and Debian 13 I needed to set it through ttm pages. If they have that set I think it's valid for it to be accounted for as they might be intentionally pooling with system ram to run larger models.

Well so the default kernel policy is that the number of pages available for GTT will use 50% of your "total system memory".  This is the same for all GPUs that use TTM. It's a statement across vendors, both integrated and discrete.

But so on AMD we have a [policy in the kernel](https://github.com/torvalds/linux/commit/759e764f7d587283b4e0b01ff930faca64370e59) that the KFD pool for dGPU comes from VRAM and for APU comes from whichever is bigger (VRAM or GTT).

But workloads that don't go through KFD will go to VRAM first and GTT next.  This will mean that if you have a 20G model and a GPU with 8G VRAM you end up spanning both if you use Vulkan.

> I'll look at adding a warning if using a dGPU and pooling with system ram?

At a high level that sounds reasonable, but it's going to be a bit complicated to plumb through I'd think.  How would you feel about splitting this part out (the pooling both together) into another PR and land everything else you're doing in this one.

src/cpp/server/model_manager.cpp
NineBallo: Yeah, the thought process was based around not being sure about codebase standards and trying to reduce main loop complexity + a bit of nervousness. I'll just swap that out.

src/cpp/server/model_manager.cpp
NineBallo: Ah I see, that makes sense then, largest pools until we can figure out the plumbing. Definitely outside of scope for this one agreed.

I think I found a good balance here, it will default to the behaviour you described unless the env variable LEMONADE_DGPU_GTT is enabled. If it is then a warning will be displayed on startup explaining the performance impact.

src/cpp/server/model_manager.cpp
```diff
@@ -1055,17 +1055,61 @@ static double parse_physical_memory_gb(const std::string& memory_str) {
     return 0.0;
 }
 
+
+double get_max_memory_of_device(json device, bool unifiedMemory) {
+    // Get the maximum POSSIBLE accessible memory of the device in question, then optionally replace an existing max.
```
superm1: this comment should be updated for the new behavior (no longer does the optional replace)

src/cpp/server/model_manager.cpp
```diff
@@ -1055,17 +1055,61 @@ static double parse_physical_memory_gb(const std::string& memory_str) {
...
     
     // Check if model filtering is disabled via environment variable
-    const char* disable_filtering = std::getenv("LEMONADE_DISABLE_MODEL_FILTERING");
-    if (disable_filtering && (std::string(disable_filtering) == "1" || 
-                              std::string(disable_filtering) == "true" || 
-                              std::string(disable_filtering) == "yes")) {
+    bool disable_filtering = parse_TF_env_var("LEMONADE_DISABLE_MODEL_FILTERING");
+
+    // Check if dGPUs should use GTT
+    bool enable_dgpu_gtt = parse_TF_env_var("LEMONADE_ENABLE_DGPU_GTT");
```
superm1, line 1109: this new environment variable needs to be documented.

possible places (read through and decide if they make sense contextually):

* `docs/server/lemonade-server-cli.md`
* `docs/faq.md`

src/cpp/server/model_manager.cpp
```diff
@@ -1055,17 +1055,61 @@ static double parse_physical_memory_gb(const std::string& memory_str) {
...
         filtered_out_models_.clear();
         return models;
     }
+
+    if (enable_dgpu_gtt)
+    {
+      std::cout << "[ModelManager]: LEMONADE_ENABLE_DGPU_GTT has been set to true." << std::endl
+                << "     Models are being filtered assuming GTT memory." << std::endl
+                << "     Using GTT on a dGPU will have a significant performance impact." << std::endl;
+    }
```
superm1, line 1121: But so this message will show up even if you don't have a dGPU but set the env variable; right?  Could you look at whether it's integrated or not to decide to show it?

superm1:
In your next push can you squash some commits and force push the branch instead?  I think it's getting close to merging, so may as well get it in shape to actually do so.

src/cpp/server/model_manager.cpp
NineBallo: I was conflicted with this one as the end goal is for this to only be applicable to dgpu's but also the rocm allocation thing was worrying until the supporting plumbing is added. I think you are right though, I'll bring back the igpu filter for it and push a change in a second here

jeremyfowers:
> In your next push can you squash some commits and force push the branch instead? I think it's getting close to merging, so may as well get it in shape to actually do so.

FYI @superm1 we squash merge all PRs into main so this isn’t strictly necessary

NineBallo:
Ok so I gave up trying to do the memory stuff the hacky way and just added an enum with all the possible cases. I was going to save that for later but it just makes sense to get started here.

I know we had some troubles with the naming of GTT memory and I think this naming scheme could make sense. Using Virtual instead of Dynamic that is.

I also did a rebase/squash set and cleaned things up a bit.

src/cpp/server/model_manager.cpp
NineBallo: Actually I'm not sure if that's the best option.
The logic here is that yes currently this would be displayed for both IGPU/DGPU/etc... 
but fundamentally it's just is a warning message for the environment variable, not for the device.

I was experimenting with a message more like this:
`[ModelManager]: Device "110501" of type "amd_igpu" is marked as a dGPU but models are being filtered assuming GTT memory.
Using GTT on a dGPU will have a significant performance impact.`

And while I think this is good, I think this should be part of the later expansion. This won't break rocm by default but still gives both parties the option to use unified memory if desired.

IGPU behavior is by default to now pick the largest pool, with dGPUs defaulting to vram. This GTT variable is an override to this default behavior in both cases (for now) where it takes both pools. 

I know the naming is somewhat misleading for now, but I think it's best to avoid introducing a bunch of new env variables, since once we do that llamacpp plumbing it will ultimately only be needed for the dGPUs.

NineBallo:
Oh and the new memory allocation code should match what ROCM expects, unless that new env variable is set.

NineBallo:
Rebased for #921

jeremyfowers:
> The code looks good to me now. I'll find some time to test it tomorrow.

Shall we merge @superm1 ? 

edit: he says yes :)

PR #910, opened 2026-01-23
refactor(backends): extract archive extraction utilities to shared ba…

Geramy:
…ckend_utils module

Moved duplicate archive extraction functions (extract_zip, extract_tarball, extract_archive) from llamacpp_server.cpp and ryzenaiserver.cpp into a new backend_utils.cpp file to eliminate code duplication and improve maintainability across backend implementations. Updated CMakeLists.txt to include the new source file and modified backends to use the shared utility functions.

### DIFF
--- a/.signpath/policies/lemonade/test-signing.yml
+++ b/.signpath/policies/lemonade/test-signing.yml
@@ -1,0 +1,8 @@
+# SignPath policy for test-signing
+# See: https://docs.signpath.io/trusted-build-systems/github
+
+github-policies:
+  runners:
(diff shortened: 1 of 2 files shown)


Geramy:
@jeremyfowers thanks I do see an issue in the path by the way, not sure why its being auto corrected by windows.

```
C:\Users\Jeremy\.cache\lemonade/bin\ryzenai-server
```

Windows is primarily backtick but somehow there is a forward tick prior to bin.
Maybe we should open a issue request for that to be fixed?

PR #913, opened 2026-01-23
Use system libraries if available

superm1:
Rather than relying upon pulling libraries from the web, many distros have them already available in the repos.  Use those if possible.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -579,9 +579,11 @@ jobs:
           if [ "${{ matrix.test_type }}" = "cli" ]; then
               echo "Running CLI tests..."
               $VENV_PYTHON test/server_cli.py --server-binary "$SERVER_BINARY"
+              $VENV_PYTHON test/server_cli.py --server-binary "$SERVER_BINARY" --ephemeral
           else
               echo "Running endpoint tests..."
               $VENV_PYTHON test/server_endpoints.py --server-binary "$SERVER_BINARY"
+              $VENV_PYTHON test/server_endpoints.py --server-binary "$SERVER_BINARY" --server-per-test
           fi
           
           echo "${{ matrix.test_type }} tests PASSED!"
--- a/.github/workflows/test_lemonade_eval.yml
+++ b/.github/workflows/test_lemonade_eval.yml
@@ -6,7 +6,15 @@ name: Test Lemonade Eval
 on:
   push:
     branches: ["main"]
+    paths:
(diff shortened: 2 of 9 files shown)


superm1:
NOTE: I expect that this won't work until https://github.com/lemonade-sdk/lemonade/pull/911 is merged.

jeremyfowers:
> NOTE: I expect that this won't work until #911 is merged.

merged! and updated here.

src/cpp/CMakeLists.txt
```diff
@@ -28,45 +28,55 @@ endif()
...
+set(USE_SYSTEM_CURL ${CURL_FOUND})
+set(USE_SYSTEM_ZSTD ${ZSTD_FOUND})
+set(USE_SYSTEM_CLI11 ${CLI11_INCLUDE_DIRS})
+set(USE_SYSTEM_HTTPLIB ${HTTPLIB_FOUND})
+
+if(USE_SYSTEM_JSON)
+    message(STATUS "Using system nlohmann_json")
+else()
+    message(STATUS "System nlohmann_json not found, will use FetchContent")
+endif()
```
jeremyfowers, line 93: There's no notion of required version numbers for the system packages (unless I missed something).

What happens here there's a very old version of the dependence installed on the system?

superm1: You're right.  This is missing.  Let me see if I can come up with a way to detect the versions from pkg config and only allow system ones if new enough.

FWIW I was able to build on both Debian unstable and Ubuntu resolute as long as I pulled an updated cpp-httplib, but on Ubuntu noble or Debian stable I expect failures, so this will be a good unit test to confirm it works.

superm1:
> Needs to handle the corner case: What happens here there's a very old version of the dependence installed on the system?

Should be sorted now. Introspecting all dep versions.

src/cpp/CMakeLists.txt
```diff
@@ -77,63 +133,93 @@ set(BUILD_SHARED_LIBS OFF CACHE INTERNAL "")  # Build static libraries
...
+
 # === nlohmann/json ===
-set(JSON_BuildTests OFF CACHE INTERNAL "")
-set(JSON_Install OFF CACHE INTERNAL "")
-FetchContent_MakeAvailable(json)
+if(NOT USE_SYSTEM_JSON)
+    FetchContent_Declare(json
+        GIT_REPOSITORY https://github.com/nlohmann/json.git
+        GIT_TAG v3.11.3
```
jeremyfowers: Can you use the constant from line 35 instead of having two sources for this number?

Comment applies to all 5 packages.

PR #880, opened 2026-01-20
Add load options menu

sofiageo:
an attempt to fix #792

PR #914, opened 2026-01-23
Add electron app linking instructions

jeremyfowers:
I realized I wasn't sure how to do this while reviewing #880 so I added it to the docs.

### DIFF
--- a/src/cpp/README.md
+++ b/src/cpp/README.md
@@ -79,6 +79,23 @@ cmake --build . --config Release -j
   - `build/lemonade-server` - Console CLI client
 - **Resources:** Automatically copied to `build/Release/resources/` (web UI files, model registry, backend version configuration)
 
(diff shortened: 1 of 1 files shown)


src/cpp/README.md
```diff
@@ -79,6 +79,23 @@ cmake --build . --config Release -j
...
 
+### Building the Electron Desktop App (Optional)
+
+The tray menu's "Open app" option and the `lemonade-server run` command can launch the Electron desktop app. To include it in your build:
+
+```powershell
+# Build the Electron app (requires Node.js 20+)
+cd src\app
+npm install
+npm run build:win
```
ramkrishna2910, line 90: Do we need instructions for Linux as well?

sofiageo: My opinion is because there is no tray at the moment for Linux there is no point to add these instructions to the cpp readme.

Issue #915, opened 2026-01-23
Repo formatting rules


jeremyfowers:
After macOS and lemonade-eval PRs merge:
- Run formatting on the repo and commit
- Add contrib helpers so people can easily run formatting
- Add CI job to enforce formatting

cc @superm1

superm1:
So in terms of the contrib helper - I think the project could do with a build system overhaul first.  Right now; you need to use `python -m build` (for python content), `npm` (for the electron app), and `cmake` (for lemonade server).  If this can be wrapped up into a single build system that calls out into the different components maybe we can also use a contrib helper to set that up.

One idea I have (which I haven't evaluated viability) is using meson + node-api for the build system.

superm1:
> Run formatting on the repo and commit
> Add CI job to enforce formatting

https://github.com/lemonade-sdk/lemonade/pull/917

> Add contrib helpers so people can easily run formatting

https://github.com/lemonade-sdk/lemonade/pull/957

Geramy:
Should be closed both PRs merged.

PR #917, opened 2026-01-23
Add pre-commit CI job

superm1:
This is a draft until ready (a few other big jobs need to merge and we need to reformat the "whole" codebase).

#915

### DIFF
--- a/.github/workflows/docs_and_style.yml
+++ b/.github/workflows/docs_and_style.yml
@@ -31,3 +31,20 @@ jobs:
           set -euo pipefail
           find . -type f -name "*.md" -not -path "./node_modules/*" \
             -exec markdown-link-check -c .github/workflows/mlc_config.json {} \;
+
+  pre-commit:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    concurrency:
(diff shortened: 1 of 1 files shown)


superm1:
Since you merged your big python drop, I rebased, re-ran and updated the CI.

superm1:
But we might want to wait for https://github.com/lemonade-sdk/lemonade/pull/868 to merge first.  Up to you.

superm1:
This job should now only run on merge.

.github/workflows/docs_and_style.yml
```diff
@@ -31,3 +31,20 @@ jobs:
           set -euo pipefail
           find . -type f -name "*.md" -not -path "./node_modules/*" \
             -exec markdown-link-check -c .github/workflows/mlc_config.json {} \;
+
+  pre-commit:
```
jeremyfowers, line 35: I don't think this will have the intended effect, since it doesn't commit the changes.

You could have it commit the changes, but that's a little scary because they would be committed post-CI.

https://github.com/marketplace/pre-commit-ci?tab=readme could be an option (free for OSS repos). Mainly I would want to make sure it wouldn't double-run our CI checks, once for the user's push and once for its fixes.

PR #924, opened 2026-01-25
Restore default port and host values in ServerManager

### LINKED ISSUE #923: list and delete commands don't work while server is running
Reproduction:

1. In terminal A: `.\Release\lemonade-server.exe serve`
2. In terminal B: `.\Release\lemonade-server.exe list` or `.\Release\lemonade-server.exe delete Qwen3-0.6B-GGUF`

In step 2, the command exits with non-zero exit code without printing any error message. The terminal A server log also doesn't show any endpoint being accessed.

The `status` and `pull` commands seem to work fine though.

I'm pretty sure commit `4a4bfb1a6451f1de886b671e8ea3d33075804f96` (https://github.com/lemonade-sdk/lemonade/pull/874) broke this. It works on `2051a5c6ae0b9d9e05004f173862d6abee9e5d40`, which is the commit immediately prior on main.

#922 will add coverage for these commands to avoid this break in the future.

We should not release the next lemonade version until this is fixed.

jeremyfowers:
Fixes #923 

@bitgamma this fixes an important Windows-only bug (see issue linked above). Any reason this change is not a good fix, and the bug should be solved a different way?

### DIFF
--- a/src/cpp/tray/server_manager.cpp
+++ b/src/cpp/tray/server_manager.cpp
@@ -50,6 +50,8 @@ namespace lemon_tray {
 
 ServerManager::ServerManager()
     : server_pid_(0)
+    , port_(8000)
+    , host_("localhost")
     , show_console_(false)
     , is_ephemeral_(false)
     , server_started_(false)


bitgamma:
Having multiple places where default values are defined is not a good practice imo. Those looking at the code for the first time might be confused by which one takes precedence and when you need to change the default (unlikely in this case, but very possible in general) you might forget to change a spot causing hard to spot bugs.

If `ServerManager` requires host and port to be passed, these parameters could just become explicit constructor arguments. We always have them available where we construct the object anyway. This also has the nice side effect that you can remove the line `server_manager_->set_port(port);` which is almost always the next line after instantiating the `ServerManager` object

src/cpp/tray/server_manager.cpp
```diff
@@ -50,6 +50,8 @@ namespace lemon_tray {
 
 ServerManager::ServerManager()
     : server_pid_(0)
+    , port_(8000)
```
bitgamma, line 53: let's pass these to the constructor instead

jeremyfowers:
> Having multiple places where default values are defined is not a good practice imo. Those looking at the code for the first time might be confused by which one takes precedence and when you need to change the default (unlikely in this case, but very possible in general) you might forget to change a spot causing hard to spot bugs.
> 
> If `ServerManager` requires host and port to be passed, these parameters could just become explicit constructor arguments. We always have them available where we construct the object anyway. This also has the nice side effect that you can remove the line `server_manager_->set_port(port);` which is almost always the next line after instantiating the `ServerManager` object

This makes sense. Ok, wont merge the changes as-is.

The `lemonade-server` commands are *supposed* to locate any running `lemonade-router` and talk to it. A default value like this shouldn’t be necessary. If that isn’t happening, there is some other bug in the code that I should track down.

PR #927, opened 2026-01-25
feat: Add status bar with server and system stats

kenvandine:
Add a status bar at the bottom of the app, displaying:
    - Input tokens and output tokens from server /stats endpoint
    - Time to first token (TTFT)
    - Tokens per second (TPS)
    - RAM usage
    - CPU usage
    - GPU utilization (platform-specific)
    - NPU utilization (platform-specific)
    
    Cross-platform support:
    - Linux: sysfs for GPU/NPU stats
    - Windows: PowerShell Get-Counter for GPU, xrt-smi for NPU
    - macOS: ioreg for GPU stats

AI disclaimer: I did use Claude to help with Windows and macOS support, as I have no access to those platforms or the expertise, so any assistance there would be greatly appreciated.

Screenshot From 2026-01-25 09-14-33

### DIFF
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -595,6 +595,7 @@ const getGpuUsage = async () => {
 // Platform-specific VRAM/GTT usage detection (returns used memory in GB)
 // For dGPU: returns dedicated VRAM only
 // For APU: returns VRAM + GTT combined (since GTT is primary GPU memory)
+// On multi-GPU systems, returns memory from the GPU with highest utilization
 const getVramUsage = async () => {
   if (process.platform === 'linux') {
     // Linux: Read from AMD sysfs
@@ -605,10 +606,23 @@ const getVramUsage = async () => {
         return [];
       });
 
+      let highestUsage = -1;
+      let highestCard = null;
+      let highestCardMemory = null;
+
       for (const card of cards) {
         if (!card.match(/^card\d+$/)) continue;
 
         const devicePath = path.join(drmPath, card, 'device');
+
+        // Read GPU utilization to find the most active GPU
+        let gpuUsage = 0;
+        try {
+          const content = await fs.promises.readFile(path.join(devicePath, 'gpu_busy_percent'), 'utf-8');
+          gpuUsage = parseFloat(content.trim()) || 0;
+        } catch (err) {
+          console.debug(`VRAM detection: ${card} gpu_busy_percent not available:`, err.code || err.message);
+        }
 
         let vramUsed = 0;
         let gttUsed = 0;
@@ -648,21 +662,33 @@ const getVramUsage = async () => {
           continue;
         }
 
+        // Calculate memory for this card
+        let cardMemory = 0;
         if (isDGPU) {
-          // Discrete GPU: report only dedicated VRAM
-          if (vramUsed > 0) {
-            console.debug(`VRAM detection: dGPU ${card}, VRAM used: ${(vramUsed / (1024 * 1024 * 1024)).toFixed(2)} GB`);
-            return vramUsed / (1024 * 1024 * 1024);
-          }
+          cardMemory = vramUsed;
         } else {
-          // APU: report VRAM + GTT combined (GTT is primary GPU memory carved from system RAM)
-          const totalUsed = vramUsed + gttUsed;
-          if (totalUsed > 0) {
-            console.debug(`VRAM detection: APU ${card}, VRAM+GTT used: ${(totalUsed / (1024 * 1024 * 1024)).toFixed(2)} GB (VRAM: ${(vramUsed / (1024 * 1024 * 1024)).toFixed(2)}, GTT: ${(gttUsed / (1024 * 1024 * 1024)).toFixed(2)})`);
-            return totalUsed / (1024 * 1024 * 1024);
-          }
+          cardMemory = vramUsed + gttUsed;
+        }
+
+        // Track the GPU with highest utilization
+        if (gpuUsage > highestUsage || highestCard === null) {
+          highestUsage = gpuUsage;
+          highestCard = card;
+          highestCardMemory = { vramUsed, gttUsed, isDGPU, cardMemory };
         }
       }
+
+      // Return memory from the GPU with highest utilization
+      if (highestCard !== null && highestCardMemory !== null) {
+        const memGb = highestCardMemory.cardMemory / (1024 * 1024 * 1024);
+        if (highestCardMemory.isDGPU) {
+          console.debug(`VRAM detection: dGPU ${highestCard} (${highestUsage}% usage), VRAM used: ${memGb.toFixed(2)} GB`);
+        } else {
+          console.debug(`VRAM detection: APU ${highestCard} (${highestUsage}% usage), VRAM+GTT used: ${memGb.toFixed(2)} GB`);
+        }
+        return memGb;
+      }
+
       console.debug('VRAM detection: No GPU with memory info found');
     } catch (error) {
       console.error('Failed to read VRAM/GTT stats from sysfs:', error);


src/app/src/renderer/StatusBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [systemStats, setSystemStats] = useState<SystemStats>({
+    cpu_percent: 0,
+    memory_gb: 0,
+    gpu_percent: null,
+    npu_percent: null,
+  });
+
+  const fetchStats = useCallback(async () => {
+    try {
+      const response = await serverConfig.fetch('/stats');
```
jeremyfowers, line 34: This seems to get called about every second, which seems unnecessary considering the data is only updated on the server side after an inference request.

Thoughts about tying the callback to inference requests finishing?

src/app/main.js
```diff
@@ -424,6 +424,170 @@ ipcMain.handle('get-server-port', async () => {
...
+
+  return null;
+};
+
+ipcMain.handle('get-system-stats', async () => {
+  try {
+    // Get memory info
+    const totalMemory = os.totalmem();
+    const freeMemory = os.freemem();
+    const usedMemory = totalMemory - freeMemory;
```
jeremyfowers, line 718: This seems to be returning the CPU memory used on my Windows system, but not the GPU memory, so it doesn't change when I load a model.

src/app/main.js
```diff
@@ -424,6 +424,170 @@ ipcMain.handle('get-server-port', async () => {
...
+      }
+    } catch (error) {
+      console.error('Failed to read NPU stats from sysfs:', error);
+    }
+    return null;
+
+  } else if (process.platform === 'win32') {
+    // Windows: Use xrt-smi for AMD Ryzen AI NPU
+    return new Promise((resolve) => {
+      const xrtSmiPath = path.join('C:', 'Windows', 'System32', 'AMD', 'xrt-smi.exe');
```
jeremyfowers: Some logs for this command:

```
(base) PS C:\Users\Jeremy> C:\Windows\System32\AMD\xrt-smi.exe examine -r aie
ERROR: No report generator found for report: 'aie'
: invalid argument

COMMAND: examine

DESCRIPTION: This command will 'examine' the state of the system/device and will generate a report of interest
               in a text or JSON format.

USAGE: xrt-smi examine [-h] [-d arg] [-f arg] [-o arg] [-r arg]

OPTIONS:
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest
  -f, --format       - Report output format. Valid values are:
                        JSON        - Latest JSON schema
                        JSON-2020.2 - JSON 2020.2 schema
  -h, --help         - Help to use this sub-command
  -o, --output       - Direct the output to the given file
  -r, --report       - The type of report to be produced. Reports currently available are:
                         aie-partitions - AIE partition information
                         all            - All known reports are produced
                         host           - Host information
                         platform       - Platforms flashed on the device


GLOBAL OPTIONS:
  --verbose          - Turn on verbosity
  --batch            - Enable batch mode (disables escape characters)
  --force            - When possible, force an operation
(base) PS C:\Users\Jeremy> C:\Windows\System32\AMD\xrt-smi.exe examine -r
ERROR: No report given to be produced: invalid argument

COMMAND: examine

DESCRIPTION: This command will 'examine' the state of the system/device and will generate a report of interest
               in a text or JSON format.

USAGE: xrt-smi examine [-h] [-d arg] [-f arg] [-o arg] [-r arg]

OPTIONS:
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest
  -f, --format       - Report output format. Valid values are:
                        JSON        - Latest JSON schema
                        JSON-2020.2 - JSON 2020.2 schema
  -h, --help         - Help to use this sub-command
  -o, --output       - Direct the output to the given file
  -r, --report       - The type of report to be produced. Reports currently available are:
                         aie-partitions - AIE partition information
                         all            - All known reports are produced
                         host           - Host information
                         platform       - Platforms flashed on the device


GLOBAL OPTIONS:
  --verbose          - Turn on verbosity
  --batch            - Enable batch mode (disables escape characters)
  --force            - When possible, force an operation
(base) PS C:\Users\Jeremy> C:\Windows\System32\AMD\xrt-smi.exe examine -r all
System Configuration
  OS Name              : Windows NT
  Release              : 26200
  Machine              : x86_64
  CPU Cores            : 32
  Memory               : 65174 MB
  Distribution         : Microsoft Windows 11 Pro
  Model                : NucBox_EVO-X2
  BIOS Vendor          : American Megatrends International, LLC.
  BIOS Version         : EVO-X2 1.05

XRT
  Version              : 2.19.0
  Branch               : HEAD
  Hash                 : 77c7088d804602a53c3eb489b9cb37b709bcd751
  Hash Date            : 2025-12-04 16:52:22
  NPU Driver Version   : 32.0.203.329
  NPU Firmware Version : 1.0.21.44

Device(s) Present
|BDF             |Name            |
|----------------|----------------|
|[00c6:00:01.1]  |NPU Strix Halo  |



--------------------------------
[00c6:00:01.1] : NPU Strix Halo
--------------------------------
AIE Partitions
  No hardware contexts running on device

Platform
  Name                   : NPU Strix Halo
  Power Mode             : Performance
  Total Columns          : 8

Power                    : N/A
```

I dont see any utilization metric in there.

jeremyfowers: NPU utilization shows as N/A on my system.

src/app/src/renderer/StatusBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  return (
+    <div className="status-bar">
+      <div className="status-bar-item">
+        <span className="status-bar-label">INPUT TOKENS:</span>
+        <span className="status-bar-value">{formatTokens(serverStats.input_tokens)}</span>
+      </div>
+      <div className="status-bar-item">
+        <span className="status-bar-label">TOKENS:</span>
```
jeremyfowers:
```suggestion
        <span className="status-bar-label">OUTPUT TOKENS:</span>
```

To make the distinction from INPUT TOKENS clear.

src/app/src/renderer/StatusBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+        <span className="status-bar-value">{formatTokens(serverStats.output_tokens)}</span>
+      </div>
+      <div className="status-bar-item">
+        <span className="status-bar-label">TTFT:</span>
```
jeremyfowers, line 138: Want to add tokens per second (TPS) metric? It's in stats and is very useful.

src/app/main.js
kenvandine: I think I've addressed this.

src/app/src/renderer/StatusBar.tsx
kenvandine: I hooked it up to inference-complete but also poll less frequenetly to ensure there's data for the UI, let me know what you think.

src/app/main.js
kenvandine: It might not, I added some more debug output for a variety of hardware related calls to gather more info

src/app/src/renderer/StatusBar.tsx
kenvandine: > Want to add tokens per second (TPS) metric? It's in stats and is very useful.

Great idea, done!

jeremyfowers: Works great, thanks!

src/app/main.js
jeremyfowers: image

Not working for me on Windows. Want to just hide the RAM and VRAM fields on Windows?

GPU utilization also not working for me on Windows.

CPU utilization does work on Windows!

src/app/src/renderer/StatusBar.tsx
jeremyfowers: The /stats spam is substantially reduced, thanks!

src/app/main.js
jeremyfowers: The NPU is showing as N/A on GUI, even when running NPU models right now.

Want to just remove this UI element for now?

> It might not, I added some more debug output for a variety of hardware related calls to gather more info

Where are these outputs going? My console is pretty empty

image

src/app/main.js
```diff
@@ -424,6 +424,325 @@ ipcMain.handle('get-server-port', async () => {
...
+
+  console.debug('NPU detection: Unsupported platform:', process.platform);
+  return null;
+};
+
+ipcMain.handle('get-system-stats', async () => {
+  try {
+    // Get memory info
+    const totalMemory = os.totalmem();
+    const freeMemory = os.freemem();
```
jeremyfowers, line 717: On Linux, after the update, GPU models load into "RAM" not "VRAM" on my Strix Point. So VRAM looks unused. 

GPU % utilization works nicely on Linux.

Suggest hiding the NPU field on unsupported systems, such as all Linux systems and Windows systems that don't have XDNA2 NPUs.

kenvandine: > On Linux, after the update, GPU models load into "RAM" not "VRAM" on my Strix Point. So VRAM looks unused.

For systems with an iGPU using GTT, it shouldn't show VRAM utilization, right? 
> 
> GPU % utilization works nicely on Linux.
> 
> Suggest hiding the NPU field on unsupported systems, such as all Linux systems and Windows systems that don't have XDNA2 NPUs.

jeremyfowers: I’m just picturing the people coming into the discord and asking why their model doesn’t use VRAM lol

kenvandine: > I’m just picturing the people coming into the discord and asking why their model doesn’t use VRAM lol

Understood.... trying to decide what we should do with it. I think I've addressed everything else and I resolved the conflicts.

superm1: > Understood.... trying to decide what we should do with it. I think I've addressed everything else and I resolved the conflicts.

One more conflict; jeremey tore out all the python stuff (it's going to be it's own project).  So rebase once more for that torn out stuff.

> I’m just picturing the people coming into the discord and asking why their model doesn’t use VRAM lol

Need an FAQ page and then to link to the relevant kernel commits?  
https://github.com/torvalds/linux/commit/759e764f7d587283b4e0b01ff930faca64370e59

src/app/main.js
```diff
@@ -424,6 +424,326 @@ ipcMain.handle('get-server-port', async () => {
...
+        return [];
+      });
+
+      for (const device of devices) {
+        const devicePath = path.join(accelPath, device, 'device');
+        const possiblePaths = [
+          path.join(devicePath, 'npu_busy_percent'),
+          path.join(devicePath, 'device_busy_percent'),
+          path.join(accelPath, device, 'busy_percent'),
+        ];
```
superm1: I'm guessing claude made these up, they're not what is used for XDNA.  The way that NPU metrics work (today) is that they are exported from the PMF driver:

https://git.kernel.org/pub/scm/linux/kernel/git/pdx86/platform-drivers-x86.git/commit/?h=for-next&id=118222e20d16caf38264b850d7a386e5f063008c

XDNA driver will eventually pick them up and advertise them through an IOCTL so that xrt-smi can pick them up.

So long story short - I think you should keep NPU out of this for now until the rest of that stack has all pieces landed.

kenvandine: @superm1 good idea, I dropped NPU stats for now and can resurrect it when we get NPU support on Linux and I can test it myself.

src/app/main.js
```diff
@@ -424,6 +424,256 @@ ipcMain.handle('get-server-port', async () => {
...
+
+        const amdBusyPath = path.join(drmPath, card, 'device', 'gpu_busy_percent');
+        try {
+          const content = await fs.promises.readFile(amdBusyPath, 'utf-8');
+          const percent = parseFloat(content.trim());
+          if (!isNaN(percent)) {
+            return percent;
+          }
+        } catch (err) {
+          console.debug(`GPU detection: ${amdBusyPath} not available:`, err.code || err.message);
```
superm1, line 540: I guess this will catch `-EBUSY`, right?  That's what we would hit in the case of a dGPU that is turned off at the time of the read with a modern kernel.

On an old kernel it will wake the dGPU.

src/app/main.js
```diff
@@ -424,6 +424,256 @@ ipcMain.handle('get-server-port', async () => {
...
+
+      for (const card of cards) {
+        if (!card.match(/^card\d+$/)) continue;
+
+        const amdBusyPath = path.join(drmPath, card, 'device', 'gpu_busy_percent');
+        try {
+          const content = await fs.promises.readFile(amdBusyPath, 'utf-8');
+          const percent = parseFloat(content.trim());
+          if (!isNaN(percent)) {
+            return percent;
```
superm1: On a multi AMD GPU system this is just going to return the first value it finds.  For example if you're on a Desktop APU with an dGPU plugged in but the APU comes earlier in the PCIe topology then it will return the APU.

I see a few ways to fix this.

1. Average the percent across all awake GPUs (see below comment about dGPU that is off).
2. Add an argument to `getGpuUsage` to actually use the "correct" GPU.

src/app/main.js
```diff
@@ -424,6 +424,256 @@ ipcMain.handle('get-server-port', async () => {
...
+    // Linux: Read from AMD sysfs
+    try {
+      const drmPath = '/sys/class/drm';
+      const cards = await fs.promises.readdir(drmPath).catch((err) => {
+        console.debug('VRAM detection: Failed to read /sys/class/drm:', err.code || err.message);
+        return [];
+      });
+
+      for (const card of cards) {
+        if (!card.match(/^card\d+$/)) continue;
```
superm1, line 614: This is going to return the wrong values on multi-GPU systems potentially (similar to the above comment about GPU utilization).

src/app/main.js
```diff
@@ -424,6 +424,256 @@ ipcMain.handle('get-server-port', async () => {
...
+          if (!isNaN(bytes)) {
+            return bytes / (1024 * 1024 * 1024); // Convert to GB
+          }
+        } catch (err) {
+          console.debug(`VRAM detection: ${gttUsedPath} not available:`, err.code || err.message);
+        }
+      }
+      console.debug('VRAM detection: No GPU with VRAM/GTT memory info found');
+    } catch (error) {
+      console.error('Failed to read VRAM/GTT stats from sysfs:', error);
```
superm1: There is an implicit problem here that VRAM is preference over GTT but that's not how the kernel actually works.  In the kernel for KFD it's:

dGPU: VRAM is used.
APU: VRAM is used if smaller than GTT, GTT is used otherwise.

Considering this I think the most sane policy is:
* dGPU: return VRAM
* APU: return VRAM + GTT

src/app/styles.css
```diff
@@ -4267,3 +4267,169 @@ footer {
...
+
+.status-bar-label {
+    color: #555;
+    font-size: 11px;
+}
+
+.status-bar-value {
+    color: #888;
+    font-variant-numeric: tabular-nums;
+}
```
superm1: missing newline at end of file (I suggest installing `pre-commit`, it catches this)

src/app/main.js
superm1: One more comment to justify my above proposal.  There is a major performance hit using GTT on a dGPU.  APUs have don't have a very big performance hit for using GTT which is why that kernel policy exists (like single digit %).

src/app/main.js
jeremyfowers: > Need an FAQ page and then to link to the relevant kernel commits?

Never choose to document something vs. automating it away ;)

superm1: > Never choose to document something vs. automating it away ;)

What do you think of my proposal below then?  On APU show VRAM + GTT, on dGPU show VRAM.

superm1: FWIW - that's what I convinced nvtop to do: https://github.com/Syllo/nvtop/pull/428

jeremyfowers: repeating what you said in my own words for validation purposes: yes, showing a single pool of RAM on unified memory systems seems like the right approach

src/app/main.js
kenvandine: > There is an implicit problem here that VRAM is preference over GTT but that's not how the kernel actually works. In the kernel for KFD it's:
> 
> dGPU: VRAM is used. APU: VRAM is used if smaller than GTT, GTT is used otherwise.
> 
> Considering this I think the most sane policy is:
> 
> * dGPU: return VRAM
> * APU: return VRAM + GTT

What's the best way to determine if it's a dGPU here? Would it be some heuristic related to the size of mem_info_vram_total and/or mem_info_gtt_used?

superm1: Definitely not a heuristic like that. The presence of `board_info` is a good way though. This will work on consumer products.

The best way is what libdrm uses though - there is information in the ioctl.

superm1: Another possible proxy is the presence of the `rom` sysfs file.  That's what fwupd uses at least.
https://github.com/fwupd/fwupd/blob/main/plugins/amd-gpu/fu-amd-gpu-device.c#L128

superm1: This looks good now, thanks!

jeremyfowers:
> I'm mostly happy with it, but I'm worried about what happens in multi-GPU contexts.
> 
> If plumbing through which GPU you're interacting with is too much how would you feel about just hiding the tooltips when a system has more than one GPU with the files looked for?

I would love to merge this PR ASAP, so I like the strategy of "if in doubt, hide the stat". We can always improve/add stats in subsequent PRs!

Will test the PR now :)

jeremyfowers:
Tested with Whisper and Stable Diffusion: they display `Input Tokens: 0 Output Tokens: 0` after inference, which is fine.

src/app/main.js
jeremyfowers: image

@kenvandine would it be possible to hide "VRAM" instead of displaying "VRAM: N/A" on unified memory systems? This will help minimize "is my brand new computer broken?" questions.

jeremyfowers:
GPU utilization is still locked to 0% on Windows. I suggest just hiding that stat on Windows devices for now.

superm1 (approved):

jeremyfowers (approved):
Tested on Windows and Linux, it works great!

Thanks @kenvandine for your first contribution, I think it will be a hit. And thanks for persevering through multiple rounds of feedback too, appreciate you.

PR #928, opened 2026-01-25
Allow running on python 3.14

superm1:


superm1:
(Although if you're tearing all the python stuff out maybe this doesn't matter)

jeremyfowers:
> (Although if you're tearing all the python stuff out maybe this doesn't matter)

The PR that removed 22k lines of Python just merged :)

Is there a specific usecase you have in mind for the newly-renamed `lemonade-eval` tool on Python 3.14?

My main consideration is just to ensure we have test coverage for what we claim to support in the setup.py.

superm1:
Running it on Arch at all. Python 3.13 is gone

superm1:
I suppose I could make a new CI target with either fedora or arch (both of which picked up 3.14 already) and run exact same as some other CI targets. That good enough?

superm1:
Matrix doesn't seem to make sense for arch and windows, but it can make sense to do a combination of something like Ubuntu and Arch or Ubuntu, Fedora and Arch.

Nonetheless the Arch CI target makes sense and shows a **real problem.**  

```
ERROR: Ignored the following versions that require a different python version: 1.21.2 Requires-Python >=3.7,<3.11; 1.21.3 Requires-Python >=3.7,<3.11; 1.21.4 Requires-Python >=3.7,<3.11; 1.21.5 Requires-Python >=3.7,<3.11; 1.21.6 Requires-Python >=3.7,<3.11; 1.26.0 Requires-Python >=3.9,<3.13; 1.26.1 Requires-Python >=3.9,<3.13
ERROR: Could not find a version that satisfies the requirement onnxruntime-genai==0.9.2; extra == "oga-cpu" (from lemonade-sdk[dev,oga-cpu]) (from versions: none)
ERROR: No matching distribution found for onnxruntime-genai==0.9.2; extra == "oga-cpu"
```

This shouldn't be merged until this can be fixed.

jeremyfowers:
> Matrix doesn't seem to make sense for arch and windows, but it can make sense to do a combination of something like Ubuntu and Arch or Ubuntu, Fedora and Arch.
> 
> Nonetheless the Arch CI target makes sense and shows a **real problem.**
> 
> ```
> ERROR: Ignored the following versions that require a different python version: 1.21.2 Requires-Python >=3.7,<3.11; 1.21.3 Requires-Python >=3.7,<3.11; 1.21.4 Requires-Python >=3.7,<3.11; 1.21.5 Requires-Python >=3.7,<3.11; 1.21.6 Requires-Python >=3.7,<3.11; 1.26.0 Requires-Python >=3.9,<3.13; 1.26.1 Requires-Python >=3.9,<3.13
> ERROR: Could not find a version that satisfies the requirement onnxruntime-genai==0.9.2; extra == "oga-cpu" (from lemonade-sdk[dev,oga-cpu]) (from versions: none)
> ERROR: No matching distribution found for onnxruntime-genai==0.9.2; extra == "oga-cpu"
> ```
> 
> This shouldn't be merged until this can be fixed.

Are you suggesting to enable OGA on Linux, or to use only llamacpp test cases on Linux?

The easiest solution is to skip the `Test OGA-Load (CPU ONNX model)` on Linux since there's not much use case for OGA-on-Linux today.

superm1:
> The easiest solution is to skip the Test OGA-Load (CPU ONNX model) on Linux since there's not much use case for OGA-on-Linux today.

That sounds like a good balance.

superm1:
Closed in favor of https://github.com/lemonade-sdk/lemonade/pull/938

Issue #877, opened 2026-01-20
Minor app bug: log viewer stops auto-scrolling


jeremyfowers:
Sometimes when a large amount of log contents are streamed from the server the Server Logs view in the App will stop auto-scrolling.

Pushing Jump to Bottom will restore auto-scrolling, but it can still break again.

Ideally, auto-scroll would never stop unless the user actually scrolls.

jeremyfowers:
The LLM Chat interface can also break autoscroll. Maybe the same bug, maybe not.

Usually it happens when the LLM is outputting a lot of text in a longer chat.

Repro:
1. Load `Phi-4-mini-instruct-GGUF`
2. "hi there"
3. "tell a short joke"
4. "tell a longer joke"
5. "longer"
6. "LONGER" <- stopped auto-scrolling here, my message stayed at the bottom of the message viewer

PR #930, opened 2026-01-25
Stable diffusion cpp app integration

ramkrishna2910:
This PR enables SD support in the electron app:
  - Add UI controls for image generation parameters (steps, cfg_scale, width, height, seed) in the Electron app
  - Store model-specific defaults in server_models.json and expose via /models API
  - Settings auto-populate when switching models (e.g., SD-Turbo: 4 steps, cfg=1.0 vs SDXL-Base: 20 steps, cfg=7.5)

image

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -579,9 +579,11 @@ jobs:
           if [ "${{ matrix.test_type }}" = "cli" ]; then
               echo "Running CLI tests..."
               $VENV_PYTHON test/server_cli.py --server-binary "$SERVER_BINARY"
+              $VENV_PYTHON test/server_cli.py --server-binary "$SERVER_BINARY" --ephemeral
           else
               echo "Running endpoint tests..."
               $VENV_PYTHON test/server_endpoints.py --server-binary "$SERVER_BINARY"
+              $VENV_PYTHON test/server_endpoints.py --server-binary "$SERVER_BINARY" --server-per-test
           fi
           
           echo "${{ matrix.test_type }} tests PASSED!"
--- a/.github/workflows/test_lemonade_eval.yml
+++ b/.github/workflows/test_lemonade_eval.yml
@@ -6,7 +6,15 @@ name: Test Lemonade Eval
 on:
   push:
     branches: ["main"]
+    paths:
+      - '.github/workflows/test_lemonade_eval.yml'
+      - 'setup.py'
+      - 'src/lemonade/**'
   pull_request:
+    paths:
+      - '.github/workflows/test_lemonade_eval.yml'
+      - 'setup.py'
+      - 'src/lemonade/**'
   merge_group:
 
 permissions:
--- a/test/server_cli.py
+++ b/test/server_cli.py
@@ -16,8 +16,9 @@ Two test modes:
 2. Ephemeral mode (--ephemeral): Each command that needs a server starts its own
 
 Usage:
-    python server_cli.py --server-binary lemonade-server
-    python server_cli.py --server-binary lemonade-server --ephemeral
+    python server_cli.py
+    python server_cli.py --ephemeral
+    python server_cli.py --server-binary /path/to/lemonade-server
 """
 
 import unittest
@@ -27,13 +28,13 @@ import socket
 import sys
 import os
 import argparse
-import platform
 
 from utils.test_models import (
     PORT,
     ENDPOINT_TEST_MODEL,
     TIMEOUT_MODEL_OPERATION,
     TIMEOUT_DEFAULT,
+    get_default_server_binary,
 )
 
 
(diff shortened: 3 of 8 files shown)


docs/server/server_spec.md
```diff
@@ -684,16 +684,22 @@ curl http://localhost:8000/api/v1/models?show_all=true
...
-      "recipe": "llamacpp",
-      "size": 3.61,
+      "checkpoint": "stabilityai/sd-turbo:sd_turbo.safetensors",
+      "recipe": "sd-cpp",
+      "size": 5.2,
       "downloaded": true,
       "suggested": true,
-      "labels": ["hot", "vision"]
+      "labels": ["image"],
+      "image_defaults": {
```
ramkrishna2910, line 709: This PR introduces image_defaults because unlike LLMs, SD models have per model recommended settings that dont work across the board and these settings have major impact on the result.

jeremyfowers: Interesting, same is true for LLM except that LLMs put their temperature, etc. into a config.json file with the model.

jeremyfowers:
image

hehe

docs/server/server_models.md
```diff
@@ -1530,6 +1530,91 @@ lemonade-server pull LFM2.5-1.2B-Thinking-FLM
 </details>
```
jeremyfowers, line 1530: https://github.com/user-attachments/assets/0453ccd7-56ae-4c96-9b1f-c4a4c26d9043

Seems to be a significant bug. Repro:

1. Generate an image
2. Change the prompt and generate another image

jeremyfowers: Oh... the images are there, I just need to scroll, but I didn't know that (as you can see in the video there's no indicator)

Maybe the same bug as this comment here: https://github.com/lemonade-sdk/lemonade/issues/877#issuecomment-3773491892

docs/server/server_models.md
```diff
@@ -1530,6 +1530,91 @@ lemonade-server pull LFM2.5-1.2B-Thinking-FLM
...
 
 
+### Image Generation (Stable Diffusion CPP)
+
```
jeremyfowers, line 1534: image


image


Dunno if you want to address in this PR, but it feels odd to me that the LLM settings are in the gear-icon settings menu, but the SD settings are in the main image generation pane.

I'm torn... on the one hand, I like uniformity. On the other hand, its undoubtedly useful to have the SD settings easier to access.

I leave the decision to you.

Dunno if

docs/server/server_models.md
jeremyfowers: But yeah, the "Generating Image..." spinner and the next image are consistently completely hidden.

docs/server/server_models.md
```diff
@@ -1530,6 +1530,91 @@ lemonade-server pull LFM2.5-1.2B-Thinking-FLM
...
+Image generation models use the [stable-diffusion.cpp](https://github.com/leejet/stable-diffusion.cpp) backend. Each model includes `image_defaults` that specify recommended generation parameters (steps, cfg_scale, width, height).
+
+<details>
```
jeremyfowers, line 1537: https://github.com/user-attachments/assets/66631d94-0bf3-4eb5-858c-b24dedb1c31d

the reset session button does nothing. I would expect it to clear the prior image from the session.

docs/server/server_models.md
ramkrishna2910: Ah yes, I did try multiple images. I was on a bigger screen and saw the image. I will add auto scroll to bottom.

docs/server/server_models.md
ramkrishna2910: Yeah I was torn on this as well. The LLM settings dont change the output as significantly as it does with SD. The workflow is more iterative on image gen than in text gen. 
Generate an image and keep tweaking until u get a satisfying result. Thats why i put it right under the image.

jeremyfowers: SGTM

bitgamma:
I just had to test it too, this is fun :D 

BTW: can't the ImageDefaults be handled like the other RecipeOptions? This would make them fit in the current architecture, making it easy to make them user-overridable (per-model) + optionally add cli args, env vars etc

docs/server/server_models.md
ramkrishna2910: Added auto scroll to bottom.

docs/server/server_models.md
ramkrishna2910: Reset button now clears chat window.

ramkrishna2910:
> I just had to test it too, this is fun :D
> 
> BTW: can't the ImageDefaults be handled like the other RecipeOptions? This would make them fit in the current architecture, making it easy to make them user-overridable (per-model) + optionally add cli args, env vars etc

Yes, thats a fair point.
I moved to integrate them with RecipeOptions internally - the image_defaults from server_models.json now get converted to RecipeOptions during model cache build, so you get CLI args (--steps, --cfg-scale) and per-model user persistence for free.

I kept image_defaults as a field in server_models.json because a) makes it obvious these are image generation defaults vs generic options and image gen options are more model specific and used more to tune the results than in LLMs. b) For future extension like Flux, Qwen-Image, and other newer architectures need a components field to specify multiple checkpoint files (VAE, text encoders, etc.). Keeping model-specific config in server_models.json with explicit fields lets us extend the schema cleanly.

PR #937, opened 2026-01-26
Add support for meson build system

superm1:
This wraps up the 2 build systems into one at the top level.  Meson is a lot faster than cmake.  On my system meson can build about 90% faster than cmake does.


superm1:
I played with meson a bunch and although I got it working really well on Linux I got very frustrated with Windows.  I then realized the _biggest_ performance advantage for meson is actually the ninja backend and this is usable with `cmake`.

I've opened up a PR for that instead.
https://github.com/lemonade-sdk/lemonade/pull/957

I'll close this one in favor of that.

PR #938, opened 2026-01-26
Move lemonade-eval to its own repo

jeremyfowers:
This PR removes all `lemonade-eval` docs, code, and infra from this repo.

All files removed in this PR are located here: https://github.com/lemonade-sdk/lemonade-eval

The reason for this change is to allow both repos to be flattened and clarified.

For example:
- `src/lemonade` in this repo will no longer confuse people looking for C++ source code
- Server docs dont need to be nested into `docs/server`
- The appropriate tests can be marked as `required` in each repos
- C++ server releases don't need to trigger `pypi` releases, and vice versa

Please note these to-do's are needed to finalize the first version of the new repo:
1. https://github.com/lemonade-sdk/lemonade-eval/issues/2
2. https://github.com/lemonade-sdk/lemonade-eval/issues/1


superm1:
But I still see a job `Build and publish Python distributions to PyPI` in the checks, this might need a workflow to be torn out still.

jeremyfowers:
> But I still see a job `Build and publish Python distributions to PyPI` in the checks, this might need a workflow to be torn out still.

It is still marked as "required" in the repo settings, but the workflow file itself is gone. That is why it isn't running. If folks approve this PR then I will un-require the action, and it will disappear from here.

PR #940, opened 2026-01-26
Rev the version number for release

jeremyfowers:
Unless we want to do 9.2.0 because we're very happy about stablediffusion @ramkrishna2910 ;)


ramkrishna2910:
Some would say a new modality makes this release worthy of 9.2.0 ;)

jeremyfowers:
> Some would say a new modality makes this release worthy of 9.2.0 ;)

So be it!

PR #943, opened 2026-01-27
Use older nodejs semver

superm1:
If there isn't a strong reason to take the newer one, this is one less package in Debian to solve.

### DIFF
--- a/src/app/package-lock.json
+++ b/src/app/package-lock.json
@@ -26,7 +26,7 @@
         "copy-webpack-plugin": "^13.0.1",
         "css-loader": "^7.1.2",
         "electron": "^39.2.4",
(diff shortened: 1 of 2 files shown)


jeremyfowers:
Can you run `npm audit` and see if it complains? Many of our package version are chosen by `npm audit fix`.

superm1:
> Can you run `npm audit` and see if it complains? Many of our package version are chosen by `npm audit fix`.

There was no vulnerability for that, but there was one for tar.  Let me roll it back and then add a CI job to catch vulnerabilities for the future.

superm1:
It seems that there are problems in some deps, OK this isn't a rabbit hole worth fighting.

PR #950, opened 2026-01-27
Upgrade arch job to also build fedora and debian

### LINKED ISSUE #932: Debian
Debian Linux is not supported according to https://lemonade-server.ai/

Is that correct? What needs to be done to change that?

jeremyfowers:
Fixes #932 

- We now test builds for Arch, Fedora, and Debian
- Debian-specific instructions added to src/cpp/README.md and linked from install_options.html

image

### DIFF
--- a/docs/assets/install-selector.js
+++ b/docs/assets/install-selector.js
@@ -89,7 +89,7 @@ window.lmnRender = function() {
   // Update active states and disabled states
   const cells = {
     os: ['win', 'linux', 'macos', 'docker'],
-    distro: ['win', 'macos', 'ubuntu', 'arch', 'fedora'],
+    distro: ['win', 'macos', 'ubuntu', 'arch', 'fedora', 'debian'],
     type: ['app', 'server'],
     fw: ['oga', 'llama', 'flm'],
     dev: ['npu', 'gpu', 'cpu']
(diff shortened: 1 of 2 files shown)


.github/workflows/linux_distro_builds.yml
```diff
@@ -11,39 +11,95 @@ permissions:
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - distro: Arch
+            image: archlinux:latest
+            python: python
+          - distro: Debian
+            image: debian:bookworm
```
superm1: make this trixie

.github/workflows/linux_distro_builds.yml
```diff
@@ -11,39 +11,95 @@ permissions:
...
+          elif [ "${{ matrix.distro }}" = "Fedora" ]; then
+            echo "Updating Fedora package database..."
+            dnf update -y
+            
+            echo "Installing build dependencies..."
+            dnf install -y \
+              @development-tools \
+              cmake \
+              git \
+              curl \
```
superm1, line 88: need `gcc-c++` as well

.github/workflows/linux_distro_builds.yml
jeremyfowers: done

.github/workflows/linux_distro_builds.yml
jeremyfowers: done

superm1:
Looks like fedora needs `which` package and AFAICT debian needs to explicitly run `bash` instead of `/bin/sh` which is provided by `dash` (so it can use `source`)

superm1:
> Looks like fedora needs `which` package and AFAICT debian needs to explicitly run `bash` instead of `/bin/sh` which is provided by `dash` (so it can use `source`)

nvm, you found another approach

PR #970, opened 2026-01-29
Add a new `/api/v1/system-stats` endpoint and use it

superm1:
By using it on the backend then any client can get this information not just the electron app.

CC @kenvandine

### DIFF
--- a/src/cpp/include/lemon/server.h
+++ b/src/cpp/include/lemon/server.h
@@ -125,8 +125,8 @@ private:
 
     std::string api_key_;
 
-    // CPU usage tracking (for Linux)
-#ifdef __linux__
+    // CPU usage tracking
+#if defined(__linux__) || defined(_WIN32)
     struct CpuStats {
         uint64_t total_idle = 0;
         uint64_t total = 0;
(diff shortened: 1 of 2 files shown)


src/app/main.js
```diff
@@ -424,324 +424,38 @@ ipcMain.handle('get-server-port', async () => {
   return cachedServerPort;
```
jeremyfowers, line 424: image

image

CPU reported as N/A on Windows. It's ok with me if its not supported, but I'd prefer the CPU key be hidden than say N/A.

superm1: Is this a regression from what the electron app/JS app did?

jeremyfowers: Yes, CPU % was working on Windows before

But not a regression against the last release since the stats bar hasnt released yet ;)

kenvandine: I don't think it's a regression, we hide the GPU info if it's null but not the CPU. We should do that consistently, though.

superm1: Fixed so that if it's null it's hidden.  But I also had copilot wire up what it thinks for reading Windows CPU stats.  We'll see how that works!

jeremyfowers (approved):
Tested on windows and linux. Works great!

The CPU utilization displays on Windows now too :)

Note: we are intentionally not documenting this now because we intend to make breaking changes.

Thanks @superm1!

PR #971, opened 2026-01-29
Add support to prefer ROCm as preference for llamacpp

superm1:
Add a new "auto" mode.  In this mode:
1. Detect the GFX ISA.
2. Try to download a ROCm binary for that ISA
3. If the error code is 404, fall back to vulkan

### DIFF
--- a/docs/server/lemonade-server-cli.md
+++ b/docs/server/lemonade-server-cli.md
@@ -51,7 +51,7 @@ lemonade-server run MODEL_NAME [options]
 | `--host [host]`                | Specify the host address for where to listen connections | `localhost` |
 | `--log-level [level]`          | Set the logging level               | info |
 | `--no-tray`                    | Start server without the tray app (headless mode) | False |
-| `--llamacpp [vulkan\|rocm\cpu]`    | Default LlamaCpp backend to use when loading models. Can be overridden per-model via the `/api/v1/load` endpoint. | vulkan |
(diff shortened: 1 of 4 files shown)


superm1:
From offline discussion, will revisit this later.

sofiageo:
Will this PR make the server download ROCm binary when server starts? If yes, it can probably help with #880 missing ROCm as an option at the moment.

superm1:
> Will this PR make the server download ROCm binary when server starts? If yes, it can probably help with #880 missing ROCm as an option at the moment.

Yes it would.  Basically if ROCm is available for the APU/dGPU it's downloaded at startup.  If it's missing (404) then it falls back to Vulkan.

superm1:
If that load options one is going to merge, I can rework it so that it downloads all applicable backends for the hardware at startup.  It will mean downloading both ROCm and vulkan at startup, but I think that's a net benefit.

sofiageo:
> If that load options one is going to merge, I can rework it so that it downloads all applicable backends for the hardware at startup. It will mean downloading both ROCm and vulkan at startup, but I think that's a net benefit.

That would help a bit in the current situation, but, Jeremy proposed to add another value in #978. So I've hidden the backend dropdown for now until that issue is resolved. thanks for the suggestion anyway!

superm1:
OK!  I'll leave the branch around as a draft at least until we explicitly decide to do ROCm as first preference in the future.  If you see value on any of it before then, feel free to lift and shift.

Issue #866, opened 2026-01-16
Rev the llamacpp version and add glm-4.7-flash


jeremyfowers:
- Problems on Arch linux
- Problems with Qwen3-Next
    - To reproduce, use the pull command to update Qwen3-Next to its latest commit
- add glm-4.7-flash to models list and test

bitgamma:
I did some tests on Arch and Strix Halo. I downloaded the binaries that we currently autodownload (`b1136`) and the latest version in the `llamacpp-rocm` repo (`b1160`). Additionally I downloaded a recent build of ROCm  (filename `therock-dist-linux-gfx1151-7.11.0a20260115.tar.gz`) and put it in `/opt/rocm`

## Results

### Version b1136  (bundled ROCm)
CMD: `LD_LIBRARY_PATH=/home/mikkoph/llama1136 ./llama-server -m /path/to/model.gguf`
RES: starts, then crashes after logging `common_init_result: fitting params to device memory, to report bugs during this step use -fit off (or --verbose if you can't)`

This error is exactly what happens when running with Lemonade

### Version b1136  (system ROCm)
CMD: `LD_LIBRARY_PATH=/opt/rocm/lib ./llama-server -m /path/to/model.gguf`
RES: Works!

### Version b1160 (bundled ROCm)
CMD: `LD_LIBRARY_PATH=/home/mikkoph/llama1160 ./llama-server -m /path/to/model.gguf`
RES: `implib-gen: libamd_comgr.so.3: failed to load library 'libamd_comgr.so.3' via callback 'amd_comgr_stub_dlopen'`

Note that the file unable to be loaded exists

### Version b1160  (system ROCm)
CMD: `LD_LIBRARY_PATH=/opt/rocm/lib ./llama-server -m /path/to/model.gguf`
RES: Works!

## Conclusion

The problem seems to lie with the bundled version of ROCm.

additional note: using `tar.gz` instead of `zip` for linux binaries would preserve symlinks (zip support for symlinks is non-standard) thus making the distribution much smaller (because of the way libraries are named/linked)

DamirLisak:
I'm having problems with the model released today: https://huggingface.co/unsloth/GLM-4.7-Flash-GGUF
The new model "GLM-4.7-Flash-GGUF" fails to start for **tool-calling** use cases on Lemonade servers.

The following command (using ROCm or the Vulkan backend) fails to start the server:
[...]

PR #974, opened 2026-01-29
llamacpp: also reserve flag aliases/negatives

bitgamma:
We currently mark as reserved only flags we explicitly pass. However many of these flags have several aliases and equivalents. Letting them through causes unpredictable behavior. This PR adds all flags variants. It also makes sure flags are reserved in all instances (for example we do not reserve the `--embedding` flag for non-embedding models right now, which seems wrong)

I also added `--no-webui` as an overridable default, because I don't think users are supposed to interact with the server directly. But this is up to discussion

Other things to discuss:

- Should we list in documentation the entire explicit list of flags? This PR doesn't conceptually reserve any new flag, just aliases or directly connected flags (antiflags, variations)
- Should we try to filter out the flags (and their arguments, which makes the logic a bit more complex) + print a warning instead of failing? The problem is for CLI users since they do not see the server log by default and might not understand why it is failing. OTOH the warning is easily ignored and people might be confused why their arguments have no effect

@jeremyfowers wdyt?

Reference for flags: https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md


jeremyfowers:
> Should we list in documentation the entire explicit list of flags? This PR doesn't conceptually reserve any new flag, just aliases or directly connected flags (antiflags, variations)

I tend to write documentation when people are already confused, since pre-emptive exhaustive documentation is a lot of work. I don't get too many questions about this, so I think we can put it off for now.

> Should we try to filter out the flags (and their arguments, which makes the logic a bit more complex) + print a warning instead of failing? The problem is for CLI users since they do not see the server log by default and might not understand why it is failing. OTOH the warning is easily ignored and people might be confused why their arguments have no effect

Same deal - I would wait for the bug reports on discord or github before putting too much time into it.

bitgamma:
yup. What prompted me to do this (with a bit of delay, I admit) was this report: https://github.com/lemonade-sdk/lemonade/issues/866#issuecomment-3772477304

see how both -c and --jinja are passed in the llamacpp_args (this was the cause of the problems)

bitgamma:
spotted another example of reserved option slipping through today (example from discord)

`Environment="LEMONADE_LLAMACPP_ARGS=--flash-attn on --no-mmap --n-gpu-layers 99 --threads 32 --threads-batch 32"`

bitgamma:
> Quick question, do you think it's advisable we clean up the arguments on exit of the backend or unload? We probably currently just throw away the backend? But it could be a good idea to clear the arrays before the destructor? Other than that this looks good, great work!

Thanks for your review. All involved arrays are locals so they get destroyed as soon as they go out of scope

Issue #817, opened 2026-01-09
Feature Request: LAN Service Discovery (Auto-Connect)


Geramy:
Clients currently require manual IP configuration. 
This is fragile because local IP addresses often change (DHCP), 
causing connection failures that force users to manually look up the new server IP and reconfigure the client.

Server: Periodically broadcasts its presence (IP/Port) to the local network.
Client: Listens for broadcasts to automatically detect and connect to available instances.
Additional Context Enables "zero-config" usage and headless deployments.

How it works (Step-by-Step):
Server Beacons (The Shout) The Lemonade Server sends a UDP broadcast packet to 255.255.255.255 on a specific port (e.g., 8888) every 5 seconds.

Payload: A simple JSON signature to prove it's a Lemonade server.

{"service": "lemonade", "port": 8000, "endpoints": ["/api/v1"]}

Note: The server does NOT need to know its own IP.

Client Listens (The Ear) The client (VS Code / WebUI) listens on UDP port 8888.
Extraction (The "Clean" Part) When the client receives the packet, the network socket provides two things:

The Data (payload)
The Sender's Address (Source IP) <--- This is the key.
The client ignores DNS names entirely. It simply sees: "I received a valid Lemonade packet from 192.168.1.15."

Auto-Connection The client adds this endpoint into the list of available servers.

Geramy:
Check https://github.com/lemonade-sdk/lemonade/tree/geramy/announce-to-lan for updates

anditherobot:
Hi @Geramy! Looks like your PR #988 (commit 8c010710
  via UdpBeaconBroadcaster in src/cpp/cli/discovery.cpp implemented LAN auto-discovery. Since this is live across CLI and app,
Should this issue be closed?

Geramy:
thanks yeah ill close it.

PR #988, opened 2026-01-29
Geramy/announce to lan

Geramy:
PR request to close and finish feature request.
https://github.com/lemonade-sdk/lemonade/issues/817


CMakeLists.txt
```diff
@@ -395,12 +395,14 @@ set(SOURCES
     src/cpp/server/utils/process_manager.cpp
     src/cpp/server/utils/path_utils.cpp
     src/cpp/server/utils/wmi_helper.cpp
+    server/utils/network_beacon.cpp
```
jeremyfowers: This needs to be rebased like the other paths

CMakeLists.txt
```diff
@@ -395,12 +395,14 @@ set(SOURCES
     src/cpp/server/utils/process_manager.cpp
```
jeremyfowers, line 403: @Geramy can you add a reviewer note to the PR description for a noob like me: how do I know this is working?

Like what commands would I run on windows or linux to observe that the service has been announced.

Geramy: You would need wireshark to view it broadcasting. I will on my next PR create some way for us to test it :)

CMakeLists.txt
```diff
@@ -395,12 +395,14 @@ set(SOURCES
...
     src/cpp/server/utils/path_utils.cpp
     src/cpp/server/utils/wmi_helper.cpp
+    src/cpp/server/utils/network_beacon.cpp
     src/cpp/server/backends/llamacpp_server.cpp
     src/cpp/server/backends/fastflowlm_server.cpp
     src/cpp/server/backends/ryzenaiserver.cpp
     src/cpp/server/backends/whisper_server.cpp
     src/cpp/server/backends/sd_server.cpp
     src/cpp/server/backends/backend_utils.cpp
+
```
superm1: why the extra whitespace here?

src/cpp/include/lemon/server.h
```diff
@@ -99,7 +100,7 @@ class Server {
 
     // Helper function to generate detailed model error responses (not found, not supported, load failure)
     nlohmann::json create_model_error(const std::string& requested_model, const std::string& exception_msg);
-
+
```
superm1: please don't introduce new whitespace after we just fixed this, can you install pre-commit and run it?

src/cpp/include/lemon/utils/network_beacon.h
```diff
@@ -0,0 +1,48 @@
+#ifndef NETWORK_BEACON_H
+#define NETWORK_BEACON_H
+
+#include <cstdint>
+#include <string>
+#include <thread>
+#include <mutex>
+#include <atomic>
```
superm1: this should be sorted

src/cpp/include/lemon/utils/network_beacon.h
```diff
@@ -0,0 +1,48 @@
...
+    bool _isInitialized;
+    uint16_t _broadcastIntervalSeconds;
+    std::string _payload;
+
+    void cleanup();
+    void createSocket();
+    void broadcastThreadLoop();
+};
+
+#endif
```
superm1: missing newline at end of file

src/cpp/server/utils/network_beacon.cpp
```diff
@@ -0,0 +1,168 @@
+#include "lemon/utils/network_beacon.h"
+
+#include <chrono>
+#include <mutex>
+#include <thread>
+#include <stdexcept>
+#include <sstream>
+
```
superm1, line 8: sort

src/cpp/server/utils/network_beacon.cpp
```diff
@@ -0,0 +1,168 @@
...
+        {
+            std::lock_guard<std::mutex> lock(_netMtx);
+            if (!_netThreadRunning) break;
+            currentPayload = _payload; // Allow payload updates on the fly
+        }
+
+        sendto(_socket, currentPayload.c_str(), (int)currentPayload.size(), 0, (sockaddr*)&addr, sizeof(addr));
+        std::this_thread::sleep_for(std::chrono::seconds(interval));
+    }
+}
```
superm1: missing newline end of file

src/cpp/server/server.cpp
```diff
@@ -519,6 +519,25 @@ void Server::run() {
             http_server_v6_->listen_after_bind();
         });
     }
+    //For now we will use getLocalHostname to get the machines hostname.
+    //This allows external devices to not have to do a rDNS lookup. 
+    bool RFC1918_IP = udp_beacon_.isRFC1918(ipv4);
+    if(RFC1918_IP) {
+        udp_beacon_.startBroadcasting(
+            8000, //Broadcast port best to not make it adjustable, so clients dont have to scan.
```
superm1, line 528: if this port is not configurable, shouldn't it be documented somewhere user facing?

Also; how does the user turn this off?  Shouldn't there be a command line option?  Or it's a mandatory feature in your mind?

CMakeLists.txt
Geramy: Was a merge conflict white space, I'll remove it.

src/cpp/include/lemon/utils/network_beacon.h
Geramy: What's the reason for new lines at the end of files? I did notice files had that it seemed odd to me.

src/cpp/server/server.cpp
Geramy: Touché, what a great set of point I didn't really put that much thought into it except for the feature, Yeah we could document this somewhere. I didn't think about having the ability to turn it off but you are right maybe there should be a --no-broadcast flag!

src/cpp/include/lemon/utils/network_beacon.h
superm1: It's to avoid problems with tools that parse files.  It's this hook that pre-commit would do for you so you don't need to think about it when committing.

https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#end-of-file-fixer

src/cpp/server/server.cpp
```diff
@@ -519,6 +519,29 @@ void Server::run() {
             http_server_v6_->listen_after_bind();
         });
     }
+    
+    //For now we will use getLocalHostname to get the machines hostname.
+    //This allows external devices to not have to do a rDNS lookup.
+    bool RFC1918_IP = udp_beacon_.isRFC1918(ipv4);
+    if(RFC1918_IP && !no_broadcast_) {
+        udp_beacon_.startBroadcasting(
+            8000, //Broadcast port best to not make it adjustable, so clients dont have to scan.
```
superm1, line 528: this is the default port lemonade runs on isn't it?  Is that intended and workable?

src/cpp/include/lemon/utils/network_beacon.h
superm1: still missing newline

src/cpp/server/cli_parser.cpp
```diff
@@ -51,6 +51,12 @@ static void add_serve_options(CLI::App* serve, ServerConfig& config, std::vector
         ->type_name("PATH")
         ->default_val(config.extra_models_dir);
 
+    serve->add_flag("--no-broadcast", config.no_broadcast,
+                   "Disable UDP broadcasting on private networks")
+        ->envname("LEMONADE_NO_BROADCAST")
```
superm1, line 56: this needs to be added to https://github.com/lemonade-sdk/lemonade/blob/main/data/lemonade.conf

src/cpp/server/server.cpp
Geramy: Its intended it uses udp and doesnt conflict with tcp because its a different protocl, we can change it to 8001 if you like.

superm1: Ah ok.  no concerns then.

PR #989, opened 2026-01-29
fix: Don't attempt to preserve ownership when extracting backends

kenvandine:
This fixes errors such as:

Jan 29 18:28:52 monster lemonade-server.service[84101]: tar: libggml-vulkan.so: Cannot change ownership to uid 1001, gid 1001: Operation not permitted
Jan 29 18:28:52 monster lemonade-server.service[84101]: tar: Exiting with failure status due to previous errors


superm1:
This is quite obviously correct, thanks!

PR #991, opened 2026-01-30
Add a helper script that uses Ninja build for building 

superm1:
Add a helper script that users can run which will setup a build directory and configure cmake to use ninja as a generator which is way faster than the traditional make generators.

Now you get started from the repo root with 

```
bash setup.sh # linux
./setup.ps1 # windows
```

And then build from the repo root with

```
cmake --build --preset default # linux
cmake --build --preset windows # windows
```


docs/dev-getting-started.md
```diff
@@ -49,7 +83,7 @@ cmake --build --preset default
 
 **Windows**
 ```powershell
-cmake --build --preset windows
+cmake --build --preset windows --config Release
```
superm1: I'm a bit surprised this is needed.

Can you add:

```
      "cacheVariables": {
        "CMAKE_BUILD_TYPE": "Release"
      }
```

into the `CMakePresets.json` to avoid it?

superm1: (assuming that fixes it - I expect you can roll back all the places that use `--config Release`)

jeremyfowers: It was doing a Debug build by default before. I'll try this.

jeremyfowers: It worked!

PR #992, opened 2026-01-30
Re-do the system info code and endpoint

### LINKED ISSUE #942: Raise an error when the user tries to pull an NPU model on an unsupported system
If a user does something like `lemonade-server pull NPU_MODEL_NAME` on an unsupported system, the model downloads but then doesn’t show up in the models endpoint, can’t load, etc.

It would be much better to raise an informative error prior to downloading the model.

Note: I’m not sure if this happens with built-in model names, or only with `user.` models. Exact repro case here: https://discord.com/channels/1392562559122407535/1394403929973985413/1465512998901776459

### LINKED ISSUE #968: App: Filter Whisper models in Linux

### LINKED ISSUE #978: system-info endpoint: each inference engine needs a "supported" field
Right now, system-info has an "available" field indicating whether something is installed. But there is no way to tell if we _could_ install it, e.g., whether it should be shown as an option at all. This is needed for GUI work to show the right buttons.

Could be helpful in the `lemonade-server` CLI as well to hide unsupported options.

My current system-info field for igpu:

```
amd_igpu":{
    "available":true,
    "inference_engines":{
        "llamacpp-rocm":
            {"available":true,"backend":"rocm","version":"b1162"},
        "llamacpp-vulkan":
            {"available":true,"backend":"vulkan","version":"b7869"}
    },"name":"AMD Radeon(TM) 8060S Graphics","virtual_mem_gb":0.0,"vram_gb":0.0}
```

Each inference engine needs a "supported" field that indicates whether installing it is supported or not:

```
amd_igpu":{
    "available":true,
    "inference_engines":{
        "llamacpp-rocm":
            {"supported":true,"available":true,"backend":"rocm","version":"b1162"},
        "llamacpp-vulkan":
            {"supported":true,"available":true,"backend":"vulkan","version":"b7869"}
    },"name":"AMD Radeon(TM) 8060S Graphics","virtual_mem_gb":0.0,"vram_gb":0.0}
```

Examples where `"supported":false` would be the case:
    - NPU on Linux
    - ROCm when the detected GPU is not one of the 4 supported families
    - Whisper on Linux (may require #968)

This filtering should already be happening somewhere within system-info or model-manager in the c++ and just needs to be exposed.

jeremyfowers:
Closes #968 
Closes #942 
Closes #978 
Also fixes a bug where we would try downloading llamacpp-rocm binaries for unsupported GPUs

## Overview

* Breaking change: Complete overhaul of system-info to be much more useful
    * Now reports which recipes, and backends of recipes, are installed, supported, or unsupported.
    * Removed the `verbose` mode; there is now a single mode that has all information.
    * Includes `whisper` and `sd` recipes, which were previously missing.
* `system_info.cpp` now has a compatibility table for recipes, backends, operating systems and hardware.
    * All compatibility checks, error messages, endpoint responses, etc. are now programmatically automated.
    * Going forward it is forbidden to add device-specific checks anywhere outside of this table.
* Added a `lemonade-server recipes` command to easily review recipes

## Demo

image

## Spec Update

Please see the server_spec.md changes in this PR to understand the new system-info format.

Namely:
 - there is a new `recipes` list in the response that clearly states what recipes/backends are supported on the system
 - the `devices` list is still there, but it only contains device info and not software info

## Testing

This PR adds a new test file, `test/server_system_info.py`, that iterates over several mock hardware configurations to ensure that system-info is producing the expected recipe support response.

## Reviewer note

Some of the code is more verbose than it needs to be because we dont have a standard mapping of recipes to `bin/RECIPE_BINARY` folder names.

For example:
- `llamacpp` -> `llama`
- `whispercpp` -> `whisper`
- `sd-cpp` -> `sd-cpp` ✅
- `oga-*` -> `ryzenai-server`

Going forwards lets make sure they are identical, and perhaps we can clean this up later.

Also, the directory structure should be `bin/RECIPE/BACKEND/BINARIES`, whereas sometimes right now it is `bin/RECIPE/BINARES` and sometimes `bin/RECIPE/BACKEND/BINARIES`.


src/cpp/server/recipe_options.cpp
```diff
@@ -82,10 +89,26 @@ static const bool is_empty_option(json option) {
...
+    static std::vector<std::string> supported_backends = llamacpp_result.backends;
+    static std::string default_backend = get_default_backend("llamacpp");
+
     for (auto& [key, opt] : CLI_OPTIONS.items()) {
         const std::string opt_name = opt["option_name"];
         CLI::Option* o;
         json defval = DEFAULTS[opt_name];
+
+        // Special handling for llamacpp_backend
+        if (opt_name == "llamacpp_backend") {
```
bitgamma: since we are going to have more backends with similar pattern, what if instead of opt_name we check "type_name" to be "BACKEND"?

we could then either (ab)use the DEFAULTS and put there the recipe name (so that we know what to call get_default_backend with) or add a field "recipe" in the option itself with the recipe name.

this also mean that we don't need to short-circuit the insertion, just handle this case in the if-else that currently checks the type of defval

src/cpp/server/system_info.cpp
```diff
@@ -48,6 +50,272 @@ const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
...
+// ============================================================================
+// Device family to human-readable name mapping
+// ============================================================================
+
+// Maps device family codes to human-readable descriptions
+// Format: {family_code, human_readable_name}
+static const std::map<std::string, std::string> DEVICE_FAMILY_NAMES = {
+    // CPU architectures
+    {"x86_64", "x86-64 processors"},
+    {"arm64", "ARM64 processors"},
```
superm1, line 129: lemonade-server is getting built for more architectures in Debian and Ubuntu.  Would it make sense to have more listed here?

Ubuntu is adding these:
* [ppc64el](https://launchpad.net/ubuntu/+source/lemonade/9.2.0-0ubuntu1/+build/32210182) (
* [s390x](https://launchpad.net/ubuntu/+source/lemonade/9.2.0-0ubuntu1/+build/32210184)
* [riscv64](https://launchpad.net/ubuntu/+source/lemonade/9.2.0-0ubuntu1/+build/32210183)

Debian will have even more.

jeremyfowers: My thought is that if lemonade itself wouldn't compile or work on something like RISC-V then there's no need to check for it in system info?

superm1: Well it's compiling on all those architectures.  I don't know how _useful_ it will be, but this is something distros.

jeremyfowers: We don't support any inference engines for those processors. Can we keep it simple and ignore them for now?

superm1: yeah let's do that .

src/cpp/server/system_info.cpp
```diff
@@ -48,6 +50,272 @@ const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
...
+
+    // whisper.cpp - Windows x86_64 only
+    {"whispercpp", "default", {"windows"}, {
+        {"cpu", {"x86_64"}},
+    }},
+
+    // stable-diffusion.cpp - Windows/Linux x86_64
+    {"sd-cpp", "default", {"windows", "linux"}, {
+        {"cpu", {"x86_64"}},
+    }},
```
iswaryaalex: Ah, so here would be the one place to extend backend, platforms, OS etc ?

src/cpp/server/recipe_options.cpp
```diff
@@ -82,10 +89,26 @@ static const bool is_empty_option(json option) {
...
+            o = app.add_option_function<std::string>(key, [opt_name, &storage = storage](const std::string& val) { storage[opt_name] = val; }, opt["help"]);
+            o->default_val(default_backend);
+            o->envname(opt["envname"]);
```
iswaryaalex: Going forward, should env_variables be deprecated?  What's the intended behavior here? Should users be able to override hardware detection via environment variables, or should we enforce the smart defaults (based on priority order)?

src/cpp/server/recipe_options.cpp
jeremyfowers: Great suggestion, added in https://github.com/lemonade-sdk/lemonade/pull/992/commits/ee7cd63eced075fb99754edb946a533a7fbe6b0e

src/cpp/server/recipe_options.cpp
```diff
@@ -82,10 +84,38 @@ static const bool is_empty_option(json option) {
...
+            }
+            const auto& result = backend_cache[recipe];
+            std::string default_backend = result.backends.empty() ? "" : result.backends[0];
+
+            o = app.add_option_function<std::string>(key, [opt_name, &storage = storage](const std::string& val) { storage[opt_name] = val; }, opt["help"]);
+            o->default_val(default_backend);
+            o->envname(opt["envname"]);
+            o->type_name(opt["type_name"]);
+            o->check(CLI::IsMember(result.backends));
+            continue;
```
bitgamma: can be merged as is, but "continue" is easy to miss I think. You could just chain this to the next if  with an else and let envname and type_name be set by the already existing code path. Calling "check" here is totally fine, there is no specific order to follow.

bitgamma (approved):
looks good to me!

src/cpp/server/recipe_options.cpp
jeremyfowers: Solved in https://github.com/lemonade-sdk/lemonade/pull/992/commits/54811b7aaa21ab8a8be64dcbaa1703b94b22c062

src/cpp/server/recipe_options.cpp
bitgamma: looks great

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -523,7 +523,7 @@ jobs:
       fail-fast: false
```
jeremyfowers, line 523: Blocked models should give the real reason they are not downloading, not

```
(base) PS C:\Users\jerem> lemonade-server pull Qwen2.5-0.5B-Instruct-CPU
Pulling model: Qwen2.5-0.5B-Instruct-CPU
Error pulling model: When registering a new model, the model name must include the `user` namespace, for example `user.Phi-4-Mini-GGUF`. Received: Qwen2.5-0.5B-Instruct-CPU
```

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -523,7 +523,7 @@ jobs:
...
       matrix:
         os: [ubuntu-latest, windows-latest]
```
jeremyfowers, line 525: Say "Requires XDNA2", not

```
(base) PS C:\Users\jerem> lemonade-server recipes
Recipe              Backend     Status        Version/Error
---------------------------------------------------------------------------
flm                 default     unsupported   Requires NPU
llamacpp            vulkan      supported     -
                    rocm        unsupported   Requires AMD dGPU
                    metal       unsupported   Requires macOS
                    cpu         supported     -
oga-cpu             default     unsupported   Requires NPU
oga-hybrid          default     unsupported   Requires NPU
oga-npu             default     unsupported   Requires NPU
sd-cpp              default     supported     -
whispercpp          default     supported     -
---------------------------------------------------------------------------
```

makn87amd (approved):

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers, line 523: Add a CLI test for the `recipes` command

jeremyfowers: Also update the system-info endpoint test to check for the recipes list

PR #997, opened 2026-01-30
Whisper streaming audio

ramkrishna2910:
Adds live microphone transcription via a WebSocket streaming API, alongside the existing file upload transcription. Users click a mic button to stream audio in real-time to Whisper and see transcripts appear live.

## Architecture
image

### VAD Configuration

  | Variable | Value | Effect of Changing |
  |----------|-------|-------------------|
  | `energy_threshold` | 0.01 | Lower = more sensitive to quiet speech; Higher = ignores background noise |
  | `min_speech_ms` | 250 | Lower = faster speech detection; Higher = filters out short sounds |
  | `min_silence_ms` | 800 | Lower = faster speech-end; Higher = bridges longer pauses |
  | `onset_frames` | 2 | Lower = faster trigger; Higher = fewer false starts |
  | `hangover_frames` | 6 | Lower = tighter cuts; Higher = captures trailing words |

  ### Audio Processing

  | Variable | Location | Value | Effect of Changing |
  |----------|----------|-------|-------------------|
  | `wav_padding_ms` | realtime_session.cpp | 500 | Lower = less silence for Whisper; Higher = avoids "too short" errors |
  | Buffer size | useAudioCapture.ts | 4096 | Lower = more responsive; Higher = fewer chunks, less overhead |
  | Target rate | useAudioCapture.ts | 16000 | Must match Whisper's expected sample rate |
  | WS close delay | ChatWindow.tsx | 3000 | Lower = faster cleanup; Higher = more time for final transcript |

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -560,6 +560,7 @@ set(SOURCES
     src/cpp/server/backends/kokoro_server.cpp
     src/cpp/server/backends/sd_server.cpp
     src/cpp/server/backends/backend_utils.cpp
+    src/cpp/server/ollama_api.cpp
     $<$<PLATFORM_ID:Windows>:src/cpp/server/streaming_audio_buffer.cpp>
     $<$<PLATFORM_ID:Windows>:src/cpp/server/vad.cpp>
     $<$<PLATFORM_ID:Windows>:src/cpp/server/realtime_session.cpp>
--- a/docs/server/server_spec.md
+++ b/docs/server/server_spec.md
@@ -58,6 +58,32 @@ The additional endpoints are:
 - GET `/api/v1/system-info` - System information and device enumeration
 - GET `/live` - Check server liveness for load balancers and orchestrators
 
+### Ollama-Compatible API
+
+Lemonade supports the [Ollama API](https://github.com/ollama/ollama/blob/main/docs/api.md), allowing applications built for Ollama to work with Lemonade without modification.
+
+To enable auto-detection by Ollama-integrated apps, launch the server on the Ollama default port:
+
+```bash
+lemonade-server serve --port 11434
+```
+
+| Endpoint | Status | Notes |
+|----------|--------|-------|
+| `POST /api/chat` | Supported | Streaming and non-streaming |
+| `POST /api/generate` | Supported | Text completion + image generation |
+| `GET /api/tags` | Supported | Lists downloaded models |
+| `POST /api/show` | Supported | Model details |
+| `DELETE /api/delete` | Supported | |
+| `POST /api/pull` | Supported | Download with progress |
+| `POST /api/embed` | Supported | New embeddings format |
+| `POST /api/embeddings` | Supported | Legacy embeddings |
+| `GET /api/ps` | Supported | Running models |
+| `GET /api/version` | Supported | |
+| `POST /api/create` | Not supported | Returns 501 |
+| `POST /api/copy` | Not supported | Returns 501 |
+| `POST /api/push` | Not supported | Returns 501 |
+
 ## Multi-Model Support
 
 Lemonade Server supports loading multiple models simultaneously, allowing you to keep frequently-used models in memory for faster switching. The server uses a Least Recently Used (LRU) cache policy to automatically manage model eviction when limits are reached.
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -643,7 +643,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        test_type: [cli, endpoints, system-info]
+        test_type: [cli, endpoints, system-info, ollama]
     env:
       LEMONADE_CI_MODE: "True"
       PYTHONIOENCODING: utf-8
@@ -743,6 +743,9 @@ jobs:
               # Unset LEMONADE_CI_MODE so the server uses mock cache files
               unset LEMONADE_CI_MODE
               $VENV_PYTHON test/server_system_info.py --server-binary "$SERVER_BINARY"
+          elif [ "${{ matrix.test_type }}" = "ollama" ]; then
+              echo "Running Ollama API tests..."
+              $VENV_PYTHON test/test_ollama.py --server-binary "$SERVER_BINARY"
           fi
 
           echo "${{ matrix.test_type }} tests PASSED!"
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -28,6 +28,32 @@ Lemonade SDK used the [ONNX TurnkeyML](https://github.com/onnx/turnkeyml) projec
 > OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 > SOFTWARE.
 
+## Ollama
+
+Lemonade's Ollama-compatible API is based on the [Ollama API specification](https://github.com/ollama/ollama/blob/main/docs/api.md), licensed under the MIT license from https://github.com/ollama/ollama.
+
+> MIT License
+>
+> Copyright (c) Ollama
+>
+> Permission is hereby granted, free of charge, to any person obtaining a copy
+> of this software and associated documentation files (the "Software"), to deal
+> in the Software without restriction, including without limitation the rights
+> to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+> copies of the Software, and to permit persons to whom the Software is
+> furnished to do so, subject to the following conditions:
+>
+> The above copyright notice and this permission notice shall be included in all
+> copies or substantial portions of the Software.
+>
+> THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+> IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+> FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+> AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+> LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+> OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+> SOFTWARE.
+
 ## TurnkeyML Attribution
 
 TurnkeyML used code from other open source projects as a starting point (see [NOTICE.md](NOTICE.md)). Thank you Philip Colangelo, Derek Elkins, Jeremy Fowers, Dan Gard, Victoria Godsoe, Mark Heaps, Daniel Holanda, Brian Kurtz, Mariah Larwood, Philip Lassen, Andrew Ling, Adrian Macias, Gary Malik, Sarah Massengill, Ashwin Murthy, Hatice Ozen, Tim Sears, Sean Settle, Krishna Sivakumar, Aviv Weinstein, Xueli Xao, Bill Xing, and Lev Zlotnik for your contributions to that work.
--- a/docs/server/lemonade-server-cli.md
+++ b/docs/server/lemonade-server-cli.md
@@ -72,6 +72,7 @@ These settings can also be provided via environment variables that Lemonade Serv
 | `LEMONADE_CTX_SIZE`                | Default context size for models                                                                                                                         |
 | `LEMONADE_LLAMACPP_ARGS`           | Custom arguments to pass to llama-server                                                                                                                |
 | `LEMONADE_EXTRA_MODELS_DIR`        | Secondary directory to scan for GGUF model files                                                                                                        |
+| `LEMONADE_MAX_LOADED_MODELS`       | Maximum number of models to keep loaded per type slot (LLMs, audio, image, etc.). Use `-1` for unlimited, or a positive integer. Default: `1`           |
 | `LEMONADE_DISABLE_MODEL_FILTERING` | Set to `1` to disable hardware-based model filtering (e.g., RAM amount, NPU availability) and show all models regardless of system capabilities         |
 | `LEMONADE_ENABLE_DGPU_GTT`         | Set to `1` to include GTT for hardware-based model filtering |
 
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -552,7 +552,7 @@ ipcMain.handle('get-version', async () => {
 });
 
 ipcMain.handle('discover-server-port', async () => {
-  let baseURL = getBaseURLFromConfig();
+  let baseURL = await getBaseURLFromConfig();
   if (baseURL) {
     console.log('Port discovery skipped - using explicit server URL:', baseURL);
     ensureTrayRunning();
--- a/src/cpp/include/lemon/ollama_api.h
+++ b/src/cpp/include/lemon/ollama_api.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include <string>
+#include <memory>
+#include <functional>
+#include <nlohmann/json.hpp>
+#include <httplib.h>
+#include "router.h"
+#include "model_manager.h"
+#include "model_types.h"
+
+namespace lemon {
+
+using json = nlohmann::json;
+
+class OllamaApi : public std::enable_shared_from_this<OllamaApi> {
+public:
+    OllamaApi(Router* router, ModelManager* model_manager);
+
+    // Must be called on a shared_ptr instance (uses shared_from_this internally)
+    void register_routes(httplib::Server& server);
+
+private:
+    Router* router_;
+    ModelManager* model_manager_;
+
+    // Endpoint handlers
+    void handle_chat(const httplib::Request& req, httplib::Response& res);
+    void handle_generate(const httplib::Request& req, httplib::Response& res);
+    void handle_generate_image(const json& request_json, httplib::Response& res, const std::string& model);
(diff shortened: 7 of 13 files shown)


CMakeLists.txt
```diff
@@ -226,6 +226,21 @@ if(NOT USE_SYSTEM_HTTPLIB)
...
+    GIT_TAG v${MIN_IXWEBSOCKET_VERSION}
+)
+# Disable TLS and ZLIB for simpler setup (localhost only, no compression needed)
+set(USE_TLS OFF CACHE INTERNAL "")
+set(USE_OPEN_SSL OFF CACHE INTERNAL "")
+set(USE_MBED_TLS OFF CACHE INTERNAL "")
+set(USE_ZLIB OFF CACHE INTERNAL "")
+set(IXWEBSOCKET_INSTALL OFF CACHE INTERNAL "")
+FetchContent_MakeAvailable(ixwebsocket)
+message(STATUS "IXWebSocket will be fetched (v${MIN_IXWEBSOCKET_VERSION})")
```
jeremyfowers: @superm1 is this going to mess you up?

@ramkrishna2910 please see the pattern of using system libraries where applicable.

~~Also curious @ramkrishna2910: why introduce websocket instead of using SSE streams like most of the other code for this?~~

docs/server/server_spec.md
```diff
@@ -586,6 +587,127 @@ Audio Transcription API. You provide an audio file and receive a text transcript
 
 
 
+### `WS /api/v1/realtime` <sub>![Status](https://img.shields.io/badge/status-partial-yellow)</sub>
+
+Realtime Audio Transcription API via WebSocket. Stream audio from a microphone and receive transcriptions in real-time with Voice Activity Detection (VAD).
+
+> **Note:** The WebSocket server runs on HTTP port + 100 (e.g., if HTTP is on port 8000, WebSocket is on port 8100).
```
jeremyfowers: Curious about the potential for collisions here? Would it be better to bind a free port when the model is loaded, and then use /health to look up the model's websocket port?

Also, why can't WS share a port with the other endpoints?

CMakeLists.txt
jeremyfowers: Oh, duh, the OpenAI API is websocket. That answers that.

examples/README.md
```diff
@@ -4,14 +4,43 @@ Interactive demos that work with Lemonade Server.
...
+### Audio Transcription
+- **realtime_transcription.py**: Stream microphone audio for real-time transcription (+ test mode for WAV files)
+
+### LLM Demos
 - **llm-debate.html**: A debate arena where multiple LLMs can debate each other on any topic
 - **multi-model-tester.html**: Test prompts across multiple models side-by-side
 
+### Image Generation
+- **api_image_generation.py**: Generate images using the Stable Diffusion API
+
```
jeremyfowers, line 19:
```suggestion
### Image Generation
- **api_image_generation.py**: Generate images using the `images/generations` endpoint

### Speech Synthesis (TTS)
- **api_text_to_speech.py**: Generate speech from a prompt using `audio/speech` endpoint

```

examples/realtime_transcription.py
```diff
@@ -0,0 +1,410 @@
...
+    except urllib.error.URLError as e:
+        print(f"Error: Cannot connect to server at {server_url}")
+        print(f"Reason: {e.reason}")
+        print("\nMake sure Lemonade Server is running:")
+        print("  lemonade-server serve")
+        return False
+    except TimeoutError:
+        print("Error: Timeout while loading model (this can happen for large models on first load)")
+        print("The model may still be downloading. Try again in a few minutes.")
+        return False
```
jeremyfowers: So much try-except, want to simplify? For example there's already a check_server function.

If you're worried about timeout on download during load you could make an explicit pull function

examples/realtime_transcription.py
```diff
@@ -0,0 +1,410 @@
...
+        print("\nOn Windows, you may need to install from wheel:")
+        print("  pip install pipwin && pipwin install pyaudio")
+        return
+
+    # Check server is running
+    if not check_server(server_url):
+        print(f"Error: Cannot connect to Lemonade Server at {server_url}")
+        print("\nMake sure the server is running:")
+        print("  lemonade-server serve")
+        return
```
jeremyfowers: check_server raises an exception on failure, so i dont think it can return `false`?

examples/realtime_transcription.py
```diff
@@ -0,0 +1,410 @@
+"""
```
jeremyfowers, line 1: Complex example - would it be simpler if it used https://github.com/openai/openai-python ?

src/app/src/renderer/utils/websocketClient.ts
```diff
@@ -0,0 +1,238 @@
+/**
```
jeremyfowers, line 1: Why not use https://github.com/openai/openai-node ?

CMakeLists.txt
superm1: > @superm1 is this going to mess you up?
> @ramkrishna2910 please see the pattern of using system libraries where applicable.

If it's not a system library yes it can mess up Linux.  If you need to use the library, please make the feature conditional.

examples/realtime_transcription.py
jeremyfowers: Works for me - a bit slow though even with whisper tiny. Need #1030 to merge so that it runs on NPU by default!

Great simplification of the code.

CMakeLists.txt
```diff
@@ -272,6 +272,21 @@ if(NOT USE_SYSTEM_HTTPLIB)
     endif()
```
jeremyfowers, line 272: https://github.com/user-attachments/assets/840ca2fa-7758-4a92-aed6-732de50b3dab

Bug: if you change models in the dropdown, it doesnt load the new model and fails.

> [RealtimeSession] Whisper response: {"error":{"message":"No model loaded: Whisper-Base","type":"model_not_loaded"}}

CMakeLists.txt
ramkrishna2910: IXWebSocket has no apt package on Ubuntu/Debian. `libwebsockets-dev` exists but that's a completely different C library.
So I guesss the FetchContent approach in the CMakeLists.txt is the correct way to handle this.
The current setup is fine for Linux; it will clone and build IXWebSocket from source during the CMake build, same as it does on Windows.
Thoughts?

examples/realtime_transcription.py
ramkrishna2910: Simplified,

examples/realtime_transcription.py
ramkrishna2910: Fixed

CMakeLists.txt
```diff
@@ -272,6 +272,21 @@ if(NOT USE_SYSTEM_HTTPLIB)
...
 endif()
 
+# === IXWebSocket (for WebSocket support) ===
+set(MIN_IXWEBSOCKET_VERSION "11.4.4")
```
jeremyfowers: https://github.com/user-attachments/assets/c1ef738f-193b-40e9-9daf-b6ac3b306f0c


Does this seem like reasonable latency on NPU?

You can see in the vid I start talking at 0:02 and finish at 0:09. Result appears at 0:12.

Is this 7 seconds of streaming buffering followed by 3 seconds of processing?

ramkrishna2910: @iswaryaalex is this expected?

CMakeLists.txt
```diff
@@ -272,6 +272,21 @@ if(NOT USE_SYSTEM_HTTPLIB)
...
+FetchContent_Declare(ixwebsocket
+    GIT_REPOSITORY https://github.com/machinezone/IXWebSocket.git
```
jeremyfowers: image

Can you refine how this looks to meet the clean aesthetic of the rest of the app?

Can you eliminate "connected"? Seems redundant if we know the model is loaded.

"Speaking" also seems redundant with the volume bar.

jeremyfowers: Also maybe reduce the amount of buttons? Not sure why need upload, mic, and send. Can upload just auto-send, and then we dont need a send button?

CMakeLists.txt
jeremyfowers: Also if I push the mic and say "whats up" as soon as it says "connected" it takes 6-7 seconds to produce the transcription. That seems much too long? 

Your test_speech.wav file only takes about 4 seconds to transcribe, and that is 11 seconds of audio compared to ~1 second for my "whats up".

CMakeLists.txt
jeremyfowers: Lets just if(WINDOWS) it for now since we're not supporting whisper on linux yet anyways

CMakeLists.txt
jeremyfowers: This can be deferred to an issue

docs/server/server_spec.md
```diff
@@ -588,6 +589,129 @@ Audio Transcription API. You provide an audio file and receive a text transcript
 
 
 
+### `WS /realtime` <sub>![Status](https://img.shields.io/badge/status-partial-yellow)</sub>
+
+Realtime Audio Transcription API via WebSocket (OpenAI SDK compatible). Stream audio from a microphone and receive transcriptions in real-time with Voice Activity Detection (VAD).
+
+> **Note:** The WebSocket server runs on HTTP port + 100 (e.g., if HTTP is on port 8000, WebSocket is on port 8100).
+>
+> **Limitations:** Only 16kHz mono PCM16 audio format is supported. Uses the same Whisper models as the HTTP transcription endpoint.
```
jeremyfowers, line 645: Add a note like "we are still tuning the chunking to balance latency vs. accuracy"?

test/server_whisper.py
```diff
@@ -239,6 +251,188 @@ def test_005_transcription_npu_backend(self):
...
+                for i in range(new_len)
+            ]
+
+        return struct.pack(f"<{len(samples)}h", *samples)
+
+    def test_006_realtime_websocket_connect(self):
+        """Test WebSocket connection and session creation."""
+        asyncio.run(self._test_006_realtime_websocket_connect())
+
+    async def _test_006_realtime_websocket_connect(self):
```
jeremyfowers, line 319: Can these tests use OpenAI API like the example? That would give much more assurance that we are OAI compliant.

CMakeLists.txt
iswaryaalex: There is a lot of initializations, preproc, loading onto NPU etc overhead that gets amortized when you transcribe a 5s or greater audio and get ~2-3x faster than realtime.

But for short chunks <2-3s, this tends to be too long. However, if the model is fully loaded after the first run, this should be much faster

CMakeLists.txt
ramkrishna2910: @iswaryaalex is fixing this in #1041

src/app/src/renderer/utils/websocketClient.ts
ramkrishna2910: Moved to openai-node instead of custom websocket implementation

docs/server/server_spec.md
ramkrishna2910: Moved away from using a hardcoded port. 
As to sharing ports, the server uses two separate libraries cpp-httplib for HTTP/REST endpoints and ixwebsocket for WebSocket, cpp-httplib doesn't support WebSocket. Using this 2 port approach is simpler as we dont need an bridging code between the libraries.

CMakeLists.txt
ramkrishna2910: Improved transcription outputs to be realtime. Give it another shot.

CMakeLists.txt
ramkrishna2910: How does this look: 
image

test/server_whisper.py
ramkrishna2910: Moved to OAI API.

CMakeLists.txt
ramkrishna2910: Aligned on experience offline

CMakeLists.txt
ramkrishna2910: Fixed.

src/app/src/renderer/utils/websocketClient.ts
ramkrishna2910: The Electron app intentionally uses a raw WebSocket instead of the OpenAI JS SDK's OpenAIRealtimeWebSocket for the following reasons:

1. Forced wss:// protocol - The JS SDK hardcodes `url.protocol = 'wss'` during URL construction with no option to override it. Since we support running the server on a remote machine over the network, we need the flexibility to connect via plain ws://. The SDK would break any non-TLS remote setup.
2. The native WebSocket API is already very small. 

The Python SDK avoids both issues: websocket_base_url gives full control over scheme and host, and the websockets library it uses under the hood doesn't impose protocol restrictions.

jeremyfowers: okie dokie

CMakeLists.txt
jeremyfowers: Definitely better! I still think it could be cleaner, but this is ok if you want to ship.

docs/server/server_spec.md
jeremyfowers: "Moved away from using a hardcoded port."

There are at least two places here that mention the hardcoded port

>The WebSocket server runs on HTTP port + 100 (e.g., if HTTP is on port 8000, WebSocket is on port 8100).

> you only need to change the `base_url` to point to your local Lemonade server (`ws://localhost:8100`).

docs/server/server_spec.md
```diff
@@ -589,6 +590,130 @@ Audio Transcription API. You provide an audio file and receive a text transcript
 
 
 
+### `WS /realtime` <sub>![Status](https://img.shields.io/badge/status-partial-yellow)</sub>
```
jeremyfowers, line 641: Out of scope for this PR but we really need to move each endpoint into its own sub-file!

examples/README.md
```diff
@@ -4,14 +4,45 @@ Interactive demos that work with Lemonade Server.
...
+### LLM Demos
 - **llm-debate.html**: A debate arena where multiple LLMs can debate each other on any topic
 - **multi-model-tester.html**: Test prompts across multiple models side-by-side
 
+### Image Generation
+- **api_image_generation.py**: Generate images using the `images/generations` endpoint
+
+### Speech Synthesis (TTS)
+- **api_text_to_speech.py**: Generate speech from a prompt using `audio/speech` endpoint
```
jeremyfowers, line 18: My experience running this was not nearly as good as in the app itself. Is there some settings skew affecting quality?

examples/realtime_transcription.py
```diff
@@ -0,0 +1,232 @@
...
+        with urllib.request.urlopen(req, timeout=120) as resp:
+            print(f"Model loaded: {model}")
+    except Exception as e:
+        print(f"Error loading model: {e}")
+        print("Make sure Lemonade Server is running: lemonade-server serve")
+        return
+
+    # Get WebSocket port from /health endpoint
+    try:
+        health_url = server_url.replace("/api/v1", "") + "/api/v1/health"
```
jeremyfowers:
```suggestion
        health_url = server_url + "/health"
```

?

examples/realtime_transcription.py
```diff
@@ -0,0 +1,232 @@
...
+        with urllib.request.urlopen(health_url, timeout=10) as resp:
+            health = json.loads(resp.read().decode())
+            ws_port = health.get("websocket_port")
```
jeremyfowers, line 110: Looks like you changed the /health endpoint? Please update the /health spec.

examples/realtime_transcription.py
```diff
@@ -0,0 +1,232 @@
...
+
+    args = parser.parse_args()
+
+    if args.mic:
+        transcribe_microphone(args.model, args.server)
+    else:
+        parser.print_help()
+        print("\nExample:")
+        print("  python realtime_transcription.py --mic")
+        print("  python realtime_transcription.py --mic --model Whisper-Small")
```
jeremyfowers: I don't understand this. If `--mic` is not optional, why is it an argument at all?

CMakeLists.txt
```diff
@@ -272,6 +272,23 @@ if(NOT USE_SYSTEM_HTTPLIB)
     endif()
 endif()
 
+# === IXWebSocket (for WebSocket support, Windows only) ===
```
jeremyfowers, line 275: Reminder that this hack is only valid until we have linux support for whipser... #1118 is open

jeremyfowers (changes requested):
Done reviewing! Our app works great, which is the most important.

Two major things:
1. the example should work as well as our app, so that downstream lemonade apps work well too
2. server spec /health needs an update

CMakeLists.txt
ramkrishna2910: @iswaryaalex can you handle this in your PR?

iswaryaalex: @ramkrishna2910 This can be handled in my PR that brings Linux support. IXWebSocket is supported on Linux and if it is using plain tcp, no concerns. I'll need to test this on Linux for streaming support.

docs/server/server_spec.md
ramkrishna2910: Done

CMakeLists.txt
ramkrishna2910: Keeping it as is for this PR. Will be handled in a follow up PR.
https://github.com/lemonade-sdk/lemonade/issues/1140

examples/realtime_transcription.py
ramkrishna2910: Done

examples/README.md
ramkrishna2910: Improved the experience with some additional tuning.

examples/realtime_transcription.py
ramkrishna2910: Removed

PR #998, opened 2026-01-31
Adjust behavior of `lemonade-server` with no arguments

superm1:
The most common way to run lemonade-server would be with no arguments. Especially with moving more options to be in the UI at runtime.

If no argument are passed assume the user meant the `serve` command.

### DIFF
--- a/docs/server/lemonade-server-cli.md
+++ b/docs/server/lemonade-server-cli.md
@@ -16,6 +16,7 @@ The `lemonade-server` command-line interface (CLI) provides a set of utility com
 
 | Option/Command      | Description                         |
 |---------------------|-------------------------------------|
+| *(no command)*      | Start the server process (equivalent to `serve`). See command options [below](#options-for-serve-and-run). |
 | `-v`, `--version`   | Print the `lemonade-sdk` package version used to install Lemonade Server. |
 | `serve`             | Start the server process in the current terminal. See command options [below](#options-for-serve-and-run). |
 | `status`            | Check if server is running. If it is, print the port number. |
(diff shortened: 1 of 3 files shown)


bitgamma:
this only works if providing no arguments and no env variables, because the way CLI11 works parses flags and env vars only if they match a subcommand.

Example with your patch:

```
sh$ ./lemonade-server --port 8088
The following arguments were not expected: --port 8088
Run with --help or --help-all for more information.
```

while we could probably force CLI11 to make this scenario work (by defining the switch at app-level) this would pollute all subcommands with options they don't actually handle

jeremyfowers:
I did this once and @danielholanda asked me to put it back.

Which is why `lemonade-tray.exe` exists on windows. I needed a command I could call with no args from the Start Menu.

superm1:
I saw this limitation too, but for the common case I feel it makes a lot of sense.

If you need args just actually use the sub command.

docs/server/lemonade-server-cli.md
```diff
@@ -38,9 +42,10 @@ lemonade-server run Qwen3-0.6B-GGUF --port 8080 --log-level debug --llamacpp roc
 
 ## Options for serve and run
 
-When using the `serve` command, you can configure the server with these additional options. The `run` command supports the same options but also requires a `MODEL_NAME` parameter:
+When using the `serve` command (or when running `lemonade-server` with no command), you can configure the server with these additional options. The `run` command supports the same options but also requires a `MODEL_NAME` parameter:
```
superm1, line 45: If this PR stays this documentation update needs reverting.

superm1:
If the main focus is discoverability, I'd argue that a better default behavior for no argument is a fully expanded `--help` directly instead of the friction of run `lemonade-server`, then run `lemonade-server --help` and then run your command.

jeremyfowers:
> If the main focus is discoverability, I'd argue that a better default behavior for no argument is a fully expanded `--help` directly instead of the friction of run `lemonade-server`, then run `lemonade-server --help` and then run your command.

Agreed, should fix this https://github.com/lemonade-sdk/lemonade/issues/1016

PR #1001, opened 2026-01-31
Windows - fix some issues with buildscripts

LaurentChardin:
### Assumptions:
- windows system with VS 2022
- using powershell prompt for VS 2022
- node for windows installed and in path
- python for windows installed and in path

### Issue 1: Powershell setup script during `pre-commit` installation: 
by default pip or pip3 are not in path nor aliased. Recommended usage is using py. 
See https://packaging.python.org/en/latest/tutorials/installing-packages/#use-pip-for-installing

~~### Issue 2: Cmake command chdir seems to fail with scripts:
Even if `find_program(NPM_EXECUTABLE npm)` is actually finding the path to `npm`, the `chdir` command still fails to invoke it.
I suspect that this is because `npm` is actually a script and not an exe. Calling it with `cmd` did fix the situation.~~
This was breaking the CI build

CC @jeremyfowers


jeremyfowers (approved):
LGTM. Thanks @LaurentChardin for helping to improve the build scripts!

PR #965, opened 2026-01-28
Add support for lemonade-server hosting a Lemonade Web-App

superm1:
Right now, `lemonade-server`/`lemonade-router` require a client to connect to them to do anything.  This can be lemonade-eval or it can be the lemonade electron app.

The electron app is kinda heavy though, at least when I built it was like 300MB.  It also limits you to running lemonade from machines that you can run the electron app.  So for example it will preclude mobile apps or tablets even if you have lemonade-server running headless somewhere else.

So this PR adds support to build the react js app used by the electron app and let it be served from `lemonade-server`/`lemonade-router`.  It will be served at the end point `/web-app`.

The web app is generated by default, but can be configured to be turned off if it's not desired for some reason.

Here's what it looks like:

image

NOTE: this builds on top of https://github.com/lemonade-sdk/lemonade/pull/957, that should merge first.

PR #1002, opened 2026-02-01
Create and distribute systemd service

superm1:
This is leveraged from the [Arch AUR](https://aur.archlinux.org/cgit/aur.git/?h=lemonade-server) (thanks @sofiageo !).

Install the systemd service and default configuration file that it uses.

Changes from Arch implementation:
- Create a dedicated user and group for lemonade, but put it in the render group to ensure it can access the GPU.
- Add some systemd confinement rules
- Install using cmake
- Add postinst rules for the debian cpack package

When paired with https://github.com/lemonade-sdk/lemonade/pull/965 this will allow the user to install `lemonade-server-minimal` and the package will start up the lemonade service on it's own.

### DIFF
--- a/docs/server/server_integration.md
+++ b/docs/server/server_integration.md
@@ -187,12 +187,7 @@ msiexec /i lemonade.msi /qn ALLUSERS=1 INSTALLDIR="C:\Program Files (x86)\Lemona
 
 The Debian package installer handles all system configuration automatically, including setting up a systemd service for managing the Lemonade Server.
 
-**Installation:**
-
-```bash
-sudo apt update
-sudo apt install lemonade-server
-```
+If you would prefer to manage the lifecycle of the server process manually, the service can be disabled and manually run as well.
 
 ### Systemd Service Management
 


src/cpp/postinst-full
```diff
@@ -2,19 +2,25 @@
...
 fi
 
+# Add lemonade user to render group for GPU access
+usermod -a -G render lemonade
+
+# Create lemonade working directory
+mkdir -p /var/lib/lemonade
+chown -R lemonade:lemonade /var/lib/lemonade
+chmod 755 /var/lib/lemonade
+APP_DIR="${INSTALL_DIR}/app"
```
superm1: might not need this actually since useradd will set it up as home directory

src/cpp/postinst
```diff
@@ -1,15 +1,38 @@
...
-mkdir -p "$LLAMA_DIR"
+# Add lemonade user to render group for GPU access
+usermod -a -G render lemonade
 
-# Always set correct permissions (even if directory already existed)
-chmod 777 "$LLAMA_DIR"
+# Create lemonade working directory
+mkdir -p /var/lib/lemonade
+chown -R lemonade:lemonade /var/lib/lemonade
+chmod 755 /var/lib/lemonade
```
superm1: same as `postinst-full`, might not need this.

sofiageo (approved):
I don't have a Debian installation to test at the moment. But it looks good. We can try and transfer some of these improvements to the Arch Linux package as well :)

superm1:
Thanks for reviewing! I'm hoping this means you can totally drop these two parts from the AUR since the upstream cmake now installs it. Win win for everyone.

sofiageo:
> Thanks for reviewing! I'm hoping this means you can totally drop these two parts from the AUR since the upstream cmake now installs it. Win win for everyone.

Oh, I thought that was debian specific. The comments on top of that block confused me. I think it is better if AUR can explicitly handle the systemd unit file in case it needs any customizations. @bitgamma what do you think?

In any case I think we can replace the file before cmake starts the process, so I don't think there will be any issue.

bitgamma:
well I see the `postinst` and `postinst-full` are for CPack (.deb/.rpm) file aren't they? This means that for AUR all would remain as is. However if the systemd unit turns out to be exactly the same, then maybe we could install the upstream one directly instead of maintaining a copy in AUR? But either way works for me @sofiageo

sofiageo:
> well I see the `postinst` and `postinst-full` are for CPack (.deb/.rpm) file aren't they? This means that for AUR all would remain as is. However if the systemd unit turns out to be exactly the same, then maybe we could install the upstream one directly instead of maintaining a copy in AUR? But either way works for me @sofiageo

It's definitely possible, but, I prefer some things to be visible in the AUR page and in the AUR helpers. This is how I would like to use a similar package. But even if we drop it from AUR, it's not the end of the world :)

superm1:
I feel if possible we should use the same systemd service for all distros; that was at least my intent!

sofiageo:
> I feel if possible we should use the same systemd service for all distros; that was at least my intent!

In that case I will need to test it in Arch Linux before merging. I see it's using /lib but in Arch /lib is a symlink to /usr/lib. I will test and report asap

sofiageo:
I've created a git aur package in my PC to be able to test PRs more easily from now on. My suspicions were confirmed.

```
error: failed to commit transaction (conflicting files)
lemonade-server: /lib exists in filesystem (owned by filesystem)
Errors occurred, no packages were upgraded.
```

It seems other packages had similar issues before e.g

https://github.com/fluent/fluent-bit/pull/3394/changes

sofiageo:
If you don't want to mess with the CMakeLists.txt any more, we can always move the /lib directory in the AUR package before creating the .zstd file. So it's up to you how you want to deal with it.

sofiageo:
Also, after I fixed the /lib directory issue, I get another error

`Feb 01 16:16:06 home (lemonade-server)[121474]: lemonade-server.service: Failed at step EXEC spawning /usr/local/bin/lemonade-server: No such file or directory`

in Arch Linux the binary is located in /usr/bin

superm1:
> Also, after I fixed the /lib directory issue, I get another error
> 
> `Feb 01 16:16:06 home (lemonade-server)[121474]: lemonade-server.service: Failed at step EXEC spawning /usr/local/bin/lemonade-server: No such file or directory`
> 
> in Arch Linux the binary is located in /usr/bin
> 
> 

Thanks for testing. I'll fix it to be dynamic.

superm1:
> If you don't want to mess with the CMakeLists.txt any more, we can always move the /lib directory in the AUR package before creating the .zstd file. So it's up to you how you want to deal with it.

Would installing into /usr/lib be correct then?

sofiageo:
> Would installing into /usr/lib be correct then?

Yes, I just tested it with moving the /lib directory to /usr/lib and changed /usr/local/bin to /usr/bin in the systemd unit file and it works. So just those two issues.

superm1:
> > Would installing into /usr/lib be correct then?
> 
> Yes, I just tested it with moving the /lib directory to /usr/lib and changed /usr/local/bin to /usr/bin in the systemd unit file and it works. So just those two issues.

Fixed them both.

sofiageo:
Works great!

docs/server/server_integration.md
```diff
@@ -181,6 +181,85 @@ msiexec /i lemonade.msi /qn ALLUSERS=1 INSTALLDIR="C:\Program Files (x86)\Lemona
...
+
+### Debian/Ubuntu Package
+
+The Debian package installer handles all system configuration automatically, including setting up a systemd service for managing the Lemonade Server.
+
+**Installation:**
+
+```bash
+sudo apt update
+sudo apt install lemonade-server
```
jeremyfowers: @superm1 I’m not too familiar with systemd, can you help me understand a bit more whats happening here?

does this command need to be called in the same directory as a lemonade .deb file? Or after the .deb has been `dpkg -i` installed?

Does this replace the old way that lemonade ran on Linux (as a regular process) or adds a new way?

superm1: This documentation got a little bit ahead of itself for a future PR of mine.  I am planning to _eventually_ replace the `cpack` building method to using the "proper" Debian packaging.

I'll fix the documentation to reflect reality of this PR.

But to your question:

1. Once you install the lemonade debian package the service gets registered with systemd and will run on it's own in the background in a confined environment (can't access anything but it's own directory) under the `lemonade` user at startup
2. If you don't want to do it this way, the service can be disabled and it can launched just the same as before.

superm1: clarified it

jeremyfowers: Is 1. going to have the side effect that models people downloaded to the old default models location are no longer accessible by the server?

superm1: Yes, upgrading would have that side effect.

Options are:
1. Script to migrate people 
2. Disable service by default 
3. Do nothing; message it, growing pains.

jeremyfowers: The `--extra-models-dir DIR` feature that allows LM Studio models to be imported into Lemonade is pretty popular, is there a way to keep that working?

superm1: Well two challenges.
1. Systems can be multi user and you won't know what user has the models in their directory.
2. Due to the security confinement rules the service can't access outside of its directory.

How about this -

If we detect it's an upgrade in the postinst script (which is doable) we don't enable service by default.

jeremyfowers: Ran a poll on discord that unanimously voted for Option 3 above. The next release notes will have a notice and migration advice.

PR #1004, opened 2026-02-01
Add connection settings in Electron app

### LINKED ISSUE #976: Add API Key to the Lemonade App
If the user starts the lemonade-server with an API key, the Lemonade App doesn't know how to talk to it.

The easiest way would be to add a `--api-key KEY` arg to the app.

Better would be to extend the settings menu in the app to let people configure the base url and api key in the GUI.

choppu:
Closes #976.

Adds base url and api key, re-connects on save. Applies API key to all requests. On app launch uses base url and api key, saved in settings file.

### DIFF
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -322,6 +322,25 @@ const broadcastServerPortUpdated = (port) => {
   }
 };
 
+const fetchWithApiKey = async (entpoint) => {
+  let serverUrl = await getBaseURLFromConfig();
+  let apiKey = (await readAppSettingsFile()).apiKey.value;
+  
+  if (!serverUrl) {
+    serverUrl = cachedServerPort ? 'http://localhost:8000' : `http://localhost:${cachedServerPort}`;
+  }
+  
+  const options = {timeout: 3000};
+  
+  if(apiKey != null && apiKey != '') {
+    options.headers = {
+      Authorization: `Bearer ${apiKey}`,
+    }
+  } 
+
+  return await fetch(`${serverUrl}${entpoint}`, options);
+}
+
 const discoverServerPort = () => {
   return new Promise((resolve) => {
     exec('lemonade-server status', { timeout: 5000 }, (error, stdout, stderr) => {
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -215,9 +215,9 @@
       ```bash
```
jeremyfowers, line 215: https://github.com/user-attachments/assets/c1038ebd-5a05-4190-865a-1ff197c1656f

There's some kind of bug where it can say "Error (Reconnecting...)" even though its actually connected. See proof in the video where I can load and talk to models even though it says Error.

I created this situation by doing the following:

1. In terminal A, launch `lemonade-app`
2. In terminal B:

```
$env:LEMONADE_API_KEY="bob"
lemonade-server serve --port 9111
```

Go into the app and set the base url:

image

and push save.

Problem persists even if I restart the app or the server.

docs/faq.md
```diff
@@ -215,9 +215,9 @@
...
       lemonade-server serve --host 0.0.0.0 --port 8000
       ```
```
jeremyfowers, line 217: I think I found a bug: when I connect to a new base url, the models list in the app is not refreshed.

So if I connect to a server with a different models list, I am still looking at the old (or no) models list in the app.

Restarting the app fixes this because the app checks `/models` on startup.

Saving a new base address should also check `/models`.

jeremyfowers (changes requested):
Overall it works nicely! I tested:

1. App and server on the same PC, using an API key
2. App and server on different PCs, both with and without an API key

I found two bugs (see comments). Once those are solved this awesome PR can merge!

docs/faq.md
choppu: Fixed. Sorry, I missed the log window before. Now it is working as expected.

docs/faq.md
choppu: I cannot reproduce this, the models are updating as expected on my machine and checking the code I see all event handlers are in place. The only thing that takes some seconds to update is the currently loaded model because the code checking it runs on a timer.

jeremyfowers (approved):
Everything is working brilliantly now! 

Thanks @choppu I think this will be a popular feature :)

PR #1005, opened 2026-02-01
Add more system information into about modal

superm1:
This information is actually already in the system info API, render it in the about modal for when people need help checking the effective configuration for their system.

image

### DIFF
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -458,6 +458,86 @@ ipcMain.handle('get-system-stats', async () => {
   };
 });
 
+ipcMain.handle('get-system-info', async () => {
+  try {
+    let serverUrl = 'http://localhost:8000';
+
+    // Use explicit server URL if configured
+    if (configuredServerBaseUrl) {
+      serverUrl = configuredServerBaseUrl;
+    } else if (cachedServerPort) {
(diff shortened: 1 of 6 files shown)


src/app/main.js
```diff
@@ -458,6 +458,86 @@ ipcMain.handle('get-system-stats', async () => {
...
+      };
+
+      // Extract GPU information from devices
+      if (data.devices) {
+        if (data.devices.amd_igpu?.name) {
+          systemInfo.gpus.push(data.devices.amd_igpu.name);
+        }
+        if (data.devices.nvidia_igpu?.name) {
+          systemInfo.gpus.push(data.devices.nvidia_igpu.name);
+        }
```
jeremyfowers, line 511: I don't think nvidia makes iGPUs?

superm1: I thought that's what DGX spark is isn't it?

jeremyfowers: Oh, good point. I'm sure we don't properly detect that in the server, but no harm having the field in the app.

superm1: Yeah just looked it up.
> NVIDIA Grace Blackwell architecture with integrated GPU and CPU

PR #1008, opened 2026-02-02
Test real deb install on github hosted runners

jeremyfowers:
This PR changes the Ubuntu CLI and endpoints tests to use real package and service installation.


superm1:
I merged it into my PR's, don't need a separate job

PR #1018, opened 2026-02-03
Look for `snap.lemonade-server.daemon.service` for systemd too

superm1:
When running under a snap the process will use
`snap.lemonade-server.daemon.service`.  Allow both to match!

CC @kenvandine


kenvandine:
This looks good! I tested it with the snap and worked as expected.

jeremyfowers (approved):
LGTM! Tested locally.

PR #1021, opened 2026-02-04
feat: Suppress logging system stats

kenvandine:
Suppress logging for system-stats and stats endpoints, which are polled by the desktop app
    
Filter out log messages for system-stats and stats endpoints to reduce log noise, matching the existing behavior for health check endpoints.


jeremyfowers (approved):
✌️ peace and quiet ✌️

PR #1022, opened 2026-02-04
Reduce down to a single Linux package

superm1:
As the electron runtime isn't needed anymore, rather than having a minimal package and a full package just have a single package.

Add "Recommends" to pull in chromium for a better experience, but this isn't mandatory.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -562,7 +562,7 @@ jobs:
 
   test-deb-inference:
     name: Test .deb - ${{ matrix.name }}
-    runs-on: ${{ matrix.runner }}
+    runs-on: [rai300_400, Linux]
     needs: build-lemonade-deb
     # Skip inference tests when signing is enabled (tag pushes or manual workflow_dispatch)
     if: ${{ !startsWith(github.ref, 'refs/tags/') && inputs.enable_signing != true }}


superm1:
For people wanting the electron app, I've got a solution!  An AppImage.  It's like a "portable EXE" file.

https://github.com/lemonade-sdk/lemonade/pull/1079

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -520,9 +493,13 @@ jobs:
   test-deb-inference:
     name: Test .deb - ${{ matrix.name }}
     runs-on: ${{ matrix.runner }}
+<<<<<<< HEAD
     needs: build-lemonade-server-deb
     # Skip inference tests when signing is enabled (tag pushes or manual workflow_dispatch)
     if: ${{ !startsWith(github.ref, 'refs/tags/') && inputs.enable_signing != true }}
+=======
+    needs: build-lemonade-deb
+>>>>>>> cf803616 (Reduce down to a single Linux package)
```
jeremyfowers: Danger, will robinson

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -747,13 +724,20 @@ jobs:
...
+<<<<<<< HEAD
       - sign-msi-installers
       - build-lemonade-server-deb
       - build-lemonade-deb-full
+=======
+      - build-lemonade-server-installer
+      - build-lemonade-deb
+      - test-exe-inference
+      - test-deb-inference
+>>>>>>> cf803616 (Reduce down to a single Linux package)
```
jeremyfowers: danger danger

superm1: ah rebase messes.  thanks, will fix.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -802,6 +779,5 @@ jobs:
           files: |
             lemonade-server-minimal.msi
             lemonade.msi
-            lemonade-server-minimal_${{ env.LEMONADE_VERSION }}_amd64.deb
             lemonade_${{ env.LEMONADE_VERSION }}_amd64.deb
```
jeremyfowers: Ok so the main user facing change here is for a single installer to replace both `lemonade` and `lemonade-server-minimal`.

I'm thinking about the user journey / growing pains here. Right now, if someone has `lemonade-server-minimal.deb` installed and they want to move to `lemonade.deb` they have to `sudo apt remove lemonade-server-minimal` first.

I'm sure this is original sin on my part, but still, is there any way to ease this pain for people?

Also, if there's only going to be one installer, should it be named `lemonade-server_VERSION_amd64.deb`? This would match the name of the command, such that `sudo apt remove lemonade-server` actually removes `lemoande-server`.

I will go with whatever you think is best here, but I wanted to raise this discussion.

superm1: > I'm sure this is original sin on my part, but still, is there any way to ease this pain for people?

Yeah there are ways to do package name transitions.  How possible it is with cpack will remain to be determined.  Moving to "real native packaging" will make it immensely easier.

I like moving to `lemonade-server` as a name.  That matches what we did in Ubuntu native repo packages too!

CMakeLists.txt
```diff
@@ -1253,16 +1252,10 @@ if(UNIX AND NOT APPLE)
         endif()
 
     else()
-        # Minimal package (server only)
-        set(CPACK_PACKAGE_NAME "lemonade-server-minimal")
-
+        set(CPACK_PACKAGE_NAME "lemonade")
         set(CPACK_DEBIAN_PACKAGE_DEPENDS "libcurl4, libssl3, libz1, unzip, fonts-katex")
+        set(CPACK_DEBIAN_PACKAGE_RECOMMENDS, "chromium-browser | google-chrome-stable | chromium")
```
jeremyfowers:
```suggestion
        set(CPACK_DEBIAN_PACKAGE_RECOMMENDS "chromium-browser | google-chrome-stable | chromium")
```

According to claude your comma is an issue, please review:

```
2. CMake syntax error — spurious comma

  CMakeLists.txt:1257:
  set(CPACK_DEBIAN_PACKAGE_RECOMMENDS, "chromium-browser | google-chrome-stable | chromium")

  There's a comma after CPACK_DEBIAN_PACKAGE_RECOMMENDS which is not valid CMake set() syntax. This will silently create a variable     
  named CPACK_DEBIAN_PACKAGE_RECOMMENDS, (with the comma in the name), so the Recommends field will never appear in the generated .deb. 
  Should be:

  set(CPACK_DEBIAN_PACKAGE_RECOMMENDS "chromium-browser | google-chrome-stable | chromium")
```

docs/assets/install-selector.js
```diff
@@ -196,7 +196,7 @@ function renderDownload() {
         snapCommands = [
           'sudo snap install lemonade-server',
         ];
-        debFile = `lemonade-server-minimal_${version}_amd64.deb`;
+        debFile = `lemonade_${version}_amd64.deb`;
```
jeremyfowers: Image

Missing AppImage instructions to get the app?

docs/assets/install-selector.js
```diff
@@ -196,7 +196,7 @@ function renderDownload() {
...
       }
       const downloadUrl = `https://github.com/lemonade-sdk/lemonade/releases/latest/download/${debFile}`;
```
jeremyfowers, line 187: Image

Does "server only" still make sense as a category?

Would this be smoother if Ubuntu and Snap were separate categories, and they both had

To install the server: INSTRUCTION
[Optional] to install the desktop app: INSTRUCTION
Note: a web ui is provided either way

.github/actions/build-linux-deb/action.yml
```diff
@@ -1,8 +1,8 @@
 name: 'Build Linux .deb Package'
-description: 'Build Lemonade .deb package for Linux (minimal or full with Electron)'
+description: 'Build Lemonade .deb package for Linux'
 inputs:
   include-electron:
-    description: 'Whether to include the Electron app (full build)'
+    description: 'Whether to include the Electron app'
     required: false
     default: 'false'
```
jeremyfowers: Dead code?

.github/workflows/cpp_server_build_test_release.yml
superm1: I changed the name to `lemonade-server`.  After this merges we can see how we can smooth out migrations.

.github/actions/build-linux-deb/action.yml
superm1: fixed

docs/assets/install-selector.js
superm1: fixed

docs/assets/install-selector.js
superm1: fixed

CMakeLists.txt
superm1: fixed

superm1:
image

PR #1024, opened 2026-02-04
Add a limited mobile view for the web-app

superm1:
Rather than showing all panels, it just shows a panel for model manager and for chat.

Also the status bar is disabled.  Here are some screenshots of what it looks like (using Chrome devtools to emulate a mobile device)

Screenshot_20260203_234416
Screenshot_20260203_234354

### DIFF
--- a/src/app/src/renderer/App.tsx
+++ b/src/app/src/renderer/App.tsx
@@ -257,6 +257,18 @@ const App: React.FC = () => {
           onToggleLogs={() => setIsLogsVisible(!isLogsVisible)}
           isDownloadManagerVisible={isDownloadManagerVisible}
           onToggleDownloadManager={() => setIsDownloadManagerVisible(!isDownloadManagerVisible)}
+          onSelectMobilePanel={(panel) => {
+            if (panel === 'model-manager') {
+              setIsModelManagerVisible(true);
+              setIsChatVisible(false);
+            }
+            if (panel === 'chat') {
+              setIsChatVisible(true);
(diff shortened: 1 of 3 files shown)


jeremyfowers:
Cool! Will review this one after the release.

jeremyfowers:
@danielholanda please review

danielholanda:
This is in theory a great idea, and I was really excited to try it. Got it working on my phone, but after playing with it a bit the UX was just not there. Scrolling was a pain due to the way in which the app is structured and I got stuck there having to refresh multiple times to get to where I needed.

My recommendation here is to have a really good UX on the phone, but to do this I think we should probably focus on a single path, and that path should probably be the native app.

"This isn't supported on phones, please download the native app for your mobile OS: LINK" is probably the message we want to show on phones when opening the app's webui.

B0BB7C20-63B0-4BD3-8840-71E98EECB2CA
9C4F75C9-F491-4A33-95CF-5B35942FB73E

danielholanda (changes requested):
See comment above

superm1:
I'm surprised you were able to get into such a bad state on a real phone (although admittedly I only used devtools).

The only panels that were ever supposed to show up were chat and models, and only one at a time full screen.

How about if we just forbid all panels except the chat panel when on mobile?  Then we can make sure it's always full screen, and it's just a matter of working out scrolling.

But if you would rather redirect people to getting the native apps then I think we should just do UA detection and redirect people to google play or apple store.

PR #980, opened 2026-01-29
Add ROCm GPU backend support for stable-diffusion.cpp

makn87amd:
This PR adds ROCm support to sd.cpp. 

Only one GPU configuration (stx-halo) has been tested. Help is needed testing other GPU configurations

PR #1036, opened 2026-02-04
Refactor backends

### LINKED ISSUE #1035: Refactor shared backend logic to reduce duplication
Most backend server share logic to:

- Find binaries
- Check if the binary is outdated
- Download and install binaries

The logic is copied in each backend (and now also in system_info.cpp) but it can be unified by moving them into static methods. This approach keeps the backend writer in control (in case some backends have a different behavior)

The goal is to keep exactly the same behavior while removing lots of lines of code.

bitgamma:
Closes #1035 

Unifies binary installation, identification and finding. Also unifies wait_for_ready by making it more parameterized. This also touches system_info by removing the duplicated logic for binary finding. Remove ~600 lines of code.

I said that I wouldn't change behavior at all, but since I know sd-cpp is going to have a rocm backend soon, I went ahead and marked the current one as `cpu` to simplify things.

Another thing that changed is that support for `LEMONADE_RECIPE_BACKEND_BIN` (e.g: `LEMONADE_LLAMACPP_ROCM_BIN`) environment variables is automatic if using the methods from `BackendUtils`, meaning that now also sd-cpp has it (it was the only one missing it among the refactored ones)

I didn't touch RyzenAI and FLM.

FLM handles installation differently so it wouldn't fit.

RyzenAI could benefit from the utils here but I don't have a Windows machine to test and the code is different enough that I am sure I'll break something in the process. Follow up PR by a Windows dev is very welcome.

### DIFF
--- a/src/cpp/include/lemon/backends/kokoro_server.h
+++ b/src/cpp/include/lemon/backends/kokoro_server.h
@@ -12,18 +12,14 @@ namespace backends {
 class KokoroServer : public WrappedServer, public ITextToSpeechServer {
 public:
     inline static const BackendSpec SPEC = BackendSpec(
-        // log name
-            "KokoroServer",
         // recipe
             "kokoro",
         // executable
     #ifdef _WIN32
-            "koko.exe",
+            "koko.exe"
     #else
-            "koko",
+            "koko"
     #endif
-        // installation dir
-            "kokoro"
     );
 
     explicit KokoroServer(const std::string& log_level,
--- a/src/cpp/server/backends/backend_utils.cpp
+++ b/src/cpp/server/backends/backend_utils.cpp
@@ -164,7 +164,7 @@ namespace lemon::backends {
             return exe_path;
         }
 
-        std::string install_dir = get_install_directory(spec.dir_name, backend);
+        std::string install_dir = get_install_directory(spec.recipe, backend);
         exe_path = find_executable_in_install_dir(install_dir, spec.binary);
 
         if (!exe_path.empty()) {
@@ -180,7 +180,7 @@ namespace lemon::backends {
     }
 
     std::string BackendUtils::get_installed_version_file(const BackendSpec& spec, const std::string& backend) {
-        std::string install_dir = get_install_directory(spec.dir_name, backend);
+        std::string install_dir = get_install_directory(spec.recipe, backend);
         return get_version_file(install_dir);
     }
 
@@ -213,7 +213,7 @@ namespace lemon::backends {
         bool needs_install = exe_path.empty();
 
         if (needs_install) {
-            install_dir = get_install_directory(spec.dir_name, backend);
+            install_dir = get_install_directory(spec.recipe, backend);
             version_file = get_version_file(install_dir);
 
             // Check if already installed with correct version
@@ -228,7 +228,7 @@ namespace lemon::backends {
                 vf.close();
 
                 if (installed_version != expected_version) {
-                    std::cout << "[" << spec.log_name << "] Upgrading " << spec.binary << " from " << installed_version
+                    std::cout << "[" << spec.log_name() << "] Upgrading " << spec.binary << " from " << installed_version
                             << " to " << expected_version << std::endl;
                     needs_install = true;
                     fs::remove_all(install_dir);
@@ -237,7 +237,7 @@ namespace lemon::backends {
         }
 
         if (needs_install) {
-            std::cout << "[" << spec.log_name << "] Installing " << spec.binary << " (version: "
+            std::cout << "[" << spec.log_name() << "] Installing " << spec.binary << " (version: "
                     << expected_version << ")" << std::endl;
 
             // Create install directory
@@ -253,8 +253,8 @@ namespace lemon::backends {
             std::string zip_ext = is_tarball(filename) ? ".tar.gz" : ".zip";
             std::string zip_path = (cache_dir / (zip_name + "_" + expected_version + zip_ext)).string();
 
-            std::cout << "[" << spec.log_name << "] Downloading from: " << url << std::endl;
-            std::cout << "[" << spec.log_name << "] Downloading to: " << zip_path << std::endl;
+            std::cout << "[" << spec.log_name() << "] Downloading from: " << url << std::endl;
+            std::cout << "[" << spec.log_name() << "] Downloading to: " << zip_path << std::endl;
 
             // Download the file
             auto download_result = utils::HttpClient::download_file(
@@ -268,7 +268,7 @@ namespace lemon::backends {
                                         " - " + download_result.error_message);
             }
 
-            std::cout << std::endl << "[" << spec.log_name << "] Download complete!" << std::endl;
+            std::cout << std::endl << "[" << spec.log_name() << "] Download complete!" << std::endl;
 
             // Verify the downloaded file
             if (!fs::exists(zip_path)) {
@@ -276,11 +276,11 @@ namespace lemon::backends {
             }
 
             std::uintmax_t file_size = fs::file_size(zip_path);
-            std::cout << "[" << spec.log_name << "] Downloaded archive file size: "
+            std::cout << "[" << spec.log_name() << "] Downloaded archive file size: "
                     << (file_size / 1024 / 1024) << " MB" << std::endl;
 
             // Extract
-            if (!extract_archive(zip_path, install_dir, spec.log_name)) {
+            if (!extract_archive(zip_path, install_dir, spec.log_name())) {
                 fs::remove(zip_path);
                 fs::remove_all(install_dir);
                 throw std::runtime_error("Failed to extract archive: " + zip_path);
@@ -289,13 +289,13 @@ namespace lemon::backends {
             // Verify extraction
             exe_path = find_executable_in_install_dir(install_dir, spec.binary);
             if (exe_path.empty()) {
-                std::cerr << "[" << spec.log_name << "] ERROR: Extraction completed but executable not found" << std::endl;
+                std::cerr << "[" << spec.log_name() << "] ERROR: Extraction completed but executable not found" << std::endl;
                 fs::remove(zip_path);
                 fs::remove_all(install_dir);
                 throw std::runtime_error("Extraction failed: executable not found");
             }
 
-            std::cout << "[" << spec.log_name << "] Executable verified at: " << exe_path << std::endl;
+            std::cout << "[" << spec.log_name() << "] Executable verified at: " << exe_path << std::endl;
 
             // Save version info
             std::ofstream vf(version_file);
@@ -310,9 +310,9 @@ namespace lemon::backends {
             // Delete ZIP file
             fs::remove(zip_path);
 
-            std::cout << "[" << spec.log_name << "] Installation complete!" << std::endl;
+            std::cout << "[" << spec.log_name() << "] Installation complete!" << std::endl;
         } else {
-            std::cout << "[" << spec.log_name << "] Found executable at: " << exe_path << std::endl;
+            std::cout << "[" << spec.log_name() << "] Found executable at: " << exe_path << std::endl;
         }
     }
 #endif
--- a/src/cpp/include/lemon/backends/backend_utils.h
+++ b/src/cpp/include/lemon/backends/backend_utils.h
@@ -7,14 +7,11 @@ namespace fs = std::filesystem;
 
 namespace lemon::backends {
     struct BackendSpec {
-        const std::string log_name;
         const std::string recipe;
         const std::string binary;
-        const std::string dir_name;
-        BackendSpec(const std::string l,
-            const std::string r,
-            const std::string b,
-            const std::string d): log_name(l), recipe(r), binary(b), dir_name(d) {}
+        BackendSpec(const std::string r, const std::string b): recipe(r), binary(b) {}
+
+        std::string log_name() const { return recipe + " Server"; };
     };
 
     /**
--- a/src/cpp/include/lemon/backends/llamacpp_server.h
+++ b/src/cpp/include/lemon/backends/llamacpp_server.h
@@ -10,18 +10,13 @@ namespace backends {
 class LlamaCppServer : public WrappedServer, public IEmbeddingsServer, public IRerankingServer {
 public:
     inline static const BackendSpec SPEC = BackendSpec(
-        // log name
-            "LlamaCpp",
-        // recipe
             "llamacpp",
         // executable
     #ifdef _WIN32
-            "llama-server.exe",
+            "llama-server.exe"
     #else
-            "llama-server",
+            "llama-server"
     #endif
-        // installation dir
-            "llama"
     );
 
     LlamaCppServer(const std::string& log_level = "info",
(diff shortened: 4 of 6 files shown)


src/cpp/include/lemon/backends/kokoro_server.h
```diff
@@ -10,6 +11,21 @@ namespace backends {
...
+            "kokoro",
+        // executable
+    #ifdef _WIN32
+            "koko.exe",
+    #else
+            "koko",
+    #endif
+        // installation dir
+            "kokoro"
+    );
```
jeremyfowers, line 23: Is there a good reason to tolerate so many degrees of freedom here?

What if the only inputs were recipe and executable?

```
recipe = kokoro
log_name = capitalize_first_letter(recipe) + "Server"
installation_dir = recipe
```

We've already seen that people can submit inconsistent naming and then it gets missed in PR review.

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -104,15 +111,209 @@ namespace lemon::backends {
         return true;
     }
 
+    static bool is_tarball(const std::string& filename) {
+        return (filename.size() > 7) && (filename.substr(filename.size() - 7) == ".tar.gz");
+    }
```
jeremyfowers, line 116: Nice. llamacpp-rocm is moving to tarballs on linux, and now Lemonade wont need a code change to handle that.

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -104,15 +111,209 @@ namespace lemon::backends {
...
+        return (fs::path(install_dir) / "version.txt").string();
+    }
+
+    std::string BackendUtils::get_installed_version_file(const BackendSpec& spec, const std::string& variant) {
+        std::string install_dir = get_install_directory(spec.dir_name, variant);
+        return get_version_file(install_dir);
+    }
+
+#ifndef LEMONADE_TRAY
+    std::string BackendUtils::get_backend_version(const std::string& recipe, const std::string& variant) {
```
jeremyfowers: Is `variant` here what we would typically refer to as a `backend`?  Any reason to not name the arg `backend`?

src/cpp/server/backends/kokoro_server.cpp
```diff
@@ -56,133 +36,21 @@ KokoroServer::~KokoroServer() {
...
-#endif
-
-        // Delete ZIP file
-        fs::remove(zip_path);
-
-        std::cout << "[KokoroServer] Installation complete!" << std::endl;
-    } else {
-        std::cout << "[KokoroServer] Found koko at: " << exe_path << std::endl;
-    }
+    BackendUtils::install_from_github(SPEC, expected_version, repo, filename, backend);
```
jeremyfowers, line 53: `expected_version` is the result of `std::string BackendUtils::get_backend_version(const std::string& recipe, const std::string& variant)`

The args to `get_backend_version` are also available to `install_from_github()`.

Any reason that `install_from_github()` shouldn't call `get_backend_version()` internally? That would save some code in all *_server.cpp files and reduce the error surface.

src/cpp/server/backends/kokoro_server.cpp
```diff
@@ -294,10 +113,7 @@ void KokoroServer::load(const std::string& model_name, const ModelInfo& model_in
     std::cout << "[KokoroServer] Using model: " << model_index["model"] << std::endl;
 
     // Get koko executable path
-    std::string exe_path = get_kokoro_server_path();
-    if (exe_path.empty()) {
-        throw std::runtime_error("koko executable not found");
-    }
+    std::string exe_path = BackendUtils::get_backend_binary_path(SPEC, "cpu");
```
jeremyfowers, line 116: Why the hardcoded "cpu"? Shouldn't we be getting a backend from RecipeOptions at this point?

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -204,200 +169,65 @@ LlamaCppServer::~LlamaCppServer() {
...
-
-        // Delete archive file
-        fs::remove(archive_path);
-
-        std::cout << "[LlamaCpp] Installation complete!" << std::endl;
-    } else {
-        std::cout << "[LlamaCpp] Found llama-server at: " << exe_path << std::endl;
     }
+
+    BackendUtils::install_from_github(SPEC, expected_version, repo, filename, backend);
```
jeremyfowers, line 230: Architectural question for your consideration: one of the big next steps for Lemonade will be to add recipe/backend management into the app.

The first step will be to introduce an `/install` endpoint that can install any recipe/backend supported by the system. Then `/system-info` + `/install` enables the GUI.

 Is `BackendUtils::install_from_github` the right place for the installation code for when we want to support `/install`? Or do we want something more akin to `ModelManager` but for backends/recipes?

I haven't thought deeply about this yet, but wanted to get your take.

src/cpp/include/lemon/backends/kokoro_server.h
bitgamma: I didn't want to change the current names but we can make it as you propose

src/cpp/server/backends/backend_utils.cpp
bitgamma: because the server itself is also called backend, so this is a variant of the backend? I was wondering how to name it, I accept suggestions

src/cpp/server/backends/sd_server.cpp
```diff
@@ -270,7 +104,7 @@ void SDServer::load(const std::string& model_name,
     std::cout << "[SDServer] Loading model: " << model_name << std::endl;
 
     // Install sd-server if needed
-    install("");
+    install("cpu");
```
jeremyfowers, line 107: Can the `"cpu"` be inferred here? The SDServer has already been initialized to some backend, right?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -331,10 +165,7 @@ void SDServer::load(const std::string& model_name,
     model_path_ = model_path;
 
     // Get sd-server executable path
-    std::string exe_path = find_executable_in_install_dir(get_sd_install_dir());
-    if (exe_path.empty()) {
-        throw std::runtime_error("sd-server executable not found");
-    }
+    std::string exe_path = BackendUtils::get_backend_binary_path(SPEC, "cpu");
```
jeremyfowers, line 168: Same comment about hardcoded "cpu"

src/cpp/server/backends/kokoro_server.cpp
bitgamma: because often the version is needed to form the filename so to avoid reading the file twice I take it as an argument

src/cpp/server/backends/whisper_server.cpp
```diff
@@ -91,227 +51,52 @@ WhisperServer::~WhisperServer() {
...
-        chmod(exe_path.c_str(), 0755);
-#endif
-
-        // Delete ZIP file
-        fs::remove(zip_path);
-
-        std::cout << "[WhisperServer] Installation complete!" << std::endl;
     } else {
-        std::cout << "[WhisperServer] Found whisper-server at: " << exe_path << std::endl;
+        throw std::runtime_error("[WhisperServer] Only CPU and NPU backends are supported for automatic installation");
```
jeremyfowers: Can the hardcoded "CPU and NPU" be avoided here based on system-info? We'll add more backends soon.

src/cpp/server/backends/kokoro_server.cpp
bitgamma: Kokoro doesn't have any other backend atm and Is unlikely to get one in the near future, so we didn't define a recipe option for it. Do you think we should add it?

src/cpp/server/system_info.cpp
```diff
@@ -109,7 +115,7 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
     }},
 
     // stable-diffusion.cpp - Windows/Linux x86_64
-    {"sd-cpp", "default", {"windows", "linux"}, {
+    {"sd-cpp", "cpu", {"windows", "linux"}, {
```
jeremyfowers, line 118: Ok so we're banishing "default" after all! Sorry for the confusion @makn87amd...

src/cpp/server/backends/llamacpp_server.cpp
bitgamma: mmm the install method from backend class is probably the best entry point. It is currently an instance method but it could become static if it makes things easier

src/cpp/server/system_info.cpp
```diff
@@ -923,101 +925,39 @@ std::string SystemInfo::get_oga_version() {
...
-#endif
-
-    return fs::exists(exe_path);
+bool SystemInfo::is_sdcpp_installed(const std::string& backend) {
+    try {
+        BackendUtils::get_backend_binary_path(SDServer::SPEC, backend);
+        return true;
+    } catch (const std::exception& e) {
+        return false;
+    }
```
jeremyfowers, line 960: Maybe I'm missing something but this seems like a task that could be abstracted away? Like do we still need per-WrappedServer "is_X_installed" if they're all just calling get_backend_binary_path(X::SPEC, backend)?

src/cpp/server/backends/sd_server.cpp
bitgamma: this is a placeholder for when the rocm patch comes. But the backend, when it exists, should be read from recipeoptions and never stored/passed in constructors so there isn't the concept of initializing to a specific backend. The reason being that different models can prefer different backbends, so load is the only place where it makes sense to determine this

src/cpp/server/backends/sd_server.cpp
bitgamma: yes, placeholder for the upcoming rocm support

src/cpp/server/backends/whisper_server.cpp
bitgamma: I'll remove the references, this line was present before so I left it there

test/server_system_info.py
```diff
@@ -137,7 +137,7 @@ def get_server_version(server_binary: str) -> str:
         "expected_supported": {
             "llamacpp": ["vulkan", "rocm", "cpu"],
             "whispercpp": ["cpu"],  # cpu backend available on x86_64
-            "sd-cpp": ["default"],
+            "sd-cpp": ["cpu"],
```
jeremyfowers, line 140: I've been thinking about the maintenance burden of this file.

It's supposed to verify that, for a given mock hardware_info.json, the right `system-info` gets reported. Right now there are two sources of truth for "right `system-info`": this file and system_info.cpp.

Thoughts about moving the support matrix table out of system_info.cpp and into a json file, and then both this file and system_info.cpp read the JSON? It would make it automatic to maintain this file.

src/cpp/server/system_info.cpp
bitgamma: except FLM, yes. I didn't do it here because I wanted to focus only on the backend class for now

src/cpp/include/lemon/backends/kokoro_server.h
jeremyfowers: I would love to get things totally uniform. It'll make our lives easier forever.

Here's my thought... 

1. Make the names uniform here
2. Rev all the binary versions (we were probably doing this anyways) - this means all the old binaries would be deleted anyways
3. For anyone moving from Lemonade 9.3.0 to 9.3.1 we delete their `bin` folder (clearing out all the backends with inconsistent names) and then they get the nice clean templated names forever after that

test/server_system_info.py
bitgamma: That'd be easier to maintain yes. Makes sense to do in a next PR

src/cpp/server/backends/backend_utils.cpp
jeremyfowers: Here's the terminology I use in my head. Could it be applied to the code?

A `recipe` is the instructions for installing and using an `engine` that supports one or more `backends`.

`engine` examples: llamacpp, Ryzen AI SW, FastFlowLM, MLX
`backend` examples: `vulkan`, `rocm`, `cuda`, etc.

A `WrappedServer` child class is the implementation of a `recipe`.

src/cpp/server/backends/kokoro_server.cpp
jeremyfowers: fair enough!

src/cpp/server/backends/kokoro_server.cpp
jeremyfowers: I can see how this is a reasonable exception for Kokoro.

The main thing I wanted to avoid was when we introduce a new WrappedServer like SD (or maybe MLX in the future) that will inevitably have multiple backends, that there is some templated way to avoid hardcoding stuff that will need to be torn out later.

src/cpp/server/backends/llamacpp_server.cpp
jeremyfowers: Hypothetically speaking, if you wanted to implement the `/install` endpoint later (wink wink) what choice would you make now?

src/cpp/server/backends/sd_server.cpp
jeremyfowers: To rephrase this for my understanding: there is no RecipeOptions for SD right now, and we are letting #980 add it, so this can't be inferred right now, but will be later.

Did I get that right?

src/cpp/server/backends/whisper_server.cpp
jeremyfowers: I'm trying to let system-info be responsible for these kinds of prints because it has the single source of truth for what the content of the prints ought to be.

src/cpp/server/system_info.cpp
jeremyfowers: Gotcha, gotta draw the line for scoping the PR somewhere :)

test/server_system_info.py
jeremyfowers: Gotcha, gotta draw the line for scoping the PR somewhere :)

src/cpp/include/lemon/backends/kokoro_server.h
bitgamma: that sounds like an evil plan. I like it! I'll proceed with removing the parameters. That will actually reduce BackendSpec to 2 properties: recipe and binary. I'll keep the log_name and installation_dir as functions of BackendSpec, so the logic of how they are generated is crystallized in a single place

jeremyfowers: 😈 good!

jeremyfowers (changes requested):
Reviewed the code and discussed changes. Let me know when to take another look! I'm excited for this one.

src/cpp/server/backends/llamacpp_server.cpp
bitgamma: mm I am not sure, even if we had to change that function to static it wouldn't be a problem to do it when the times comes. I'd have to think more about it

src/cpp/server/backends/kokoro_server.cpp
bitgamma: on that note, I am starting to think how we could concentrate all knowledge about the backend inside the backend itself. The idea would be that when coding it, all the info is in one single place and the plumbing is handled at runtime with a sort of registry (basically, a plug-in architecture without the dynamic loading part). But one change at a time

src/cpp/server/backends/sd_server.cpp
bitgamma: exactly

Issue #1038, opened 2026-02-04
systemd service isn't going into the right place with prefix of `/usr`


superm1:
I just did a test build for Debian with the 9.3.0 RC.  The resultant systemd service ended up here (which isn't a valid search path).

```
./usr/lib/x86_64-linux-gnu/systemd/system/lemonade-server.service
```

The `debian/rules` ran:
```sh
dh_auto_configure -- -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_WEB_APP=OFF
```

This expanded out from the build system to:
```sh
	cd obj-x86_64-linux-gnu && DEB_PYTHON_INSTALL_LAYOUT=deb PKG_CONFIG=/usr/bin/pkg-config cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DFETCHCONTENT_FULLY_DISCONNECTED=ON -DCMAKE_INSTALL_RUNSTATEDIR=/run -DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=ON -GNinja -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_WEB_APP=OFF ..
```

I found that when the systemd service was installing this installed to the wrong place.
```
-- Installing: /home/supermario/src/lemonade/debian/tmp/usr/lib/x86_64-linux-gnu/systemd/system/lemonade-server.service
```

This isn't correct though; it should be ending up in `debian/tmp/usr/lib/systemd/systemd/lemonade-server.service`.

I expect this also breaks Arch (or any other distro that uses `/usr`).

superm1:
I'm guessing we shouldn't be using `CMAKE_INSTALL_LIBDIR` for the systemd install, but rather something different.

```
    # Install systemd service and configuration files
    install(FILES ${CMAKE_BINARY_DIR}/lemonade-server.service
[...]

PR #1039, opened 2026-02-04
Hardcode `lib` for systemd service under $PREFIX

superm1:
Closes: https://github.com/lemonade-sdk/lemonade/issues/1038


superm1:
(tested this on Debian - it works)

kenvandine:
Hardcoding lib here isn't a bad thing. I think there is a way to get the systemd unit dir from pkg-config, but I don't recall, and hardcoding this is actually common, as long as we use CMAKE_INSTALL_PREFIX, it's pretty packager-friendly.

PR #1040, opened 2026-02-04
Preserve chat message width when editing

danielholanda:
# Description
When editing a user message, the input was jumping to full width (very narrow) instead of staying about the same size as the original message. This made editing big messages very annoying.

Now the text area sizes itself based on content.

## Before
image

## After
image


jeremyfowers (approved):
lgtm!

PR #1033, opened 2026-02-04
Add badege from the snap store to show the latest version published in stable

kenvandine:
This adds a snap store badge which looks like:

image

PR #1043, opened 2026-02-05
Re-organize badges

superm1:
Follow up from https://github.com/lemonade-sdk/lemonade/pull/1033

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,35 +1,20 @@
 ## 🍋 Lemonade: Local LLMs with GPU and NPU acceleration
 
-<p align="center">
-  <a href="https://discord.gg/5xXzkMu8Zk">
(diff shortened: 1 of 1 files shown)


superm1:
image

kenvandine:
@superm1 this is a really nice improvement, much more readable

jeremyfowers:
I need to noodle on this some more. It's absolutely a better organization of the badges, but it also takes up a lot more vertical space. That raises the question of where it should go, so I tried moving it below the main graphic:

image

But now there's some obvious redundancy with install options and the big discord link. 

It also makes the badge situation stand out a bit more: `black` is only for examples and tests, not everything. There's no mention of the Arch tests even though there's an `aur` badge; debian, docker, and fedora are supported but not in platform support. 

@superm1 do you mind if I take the action to just totally redo the top half of the readme?

superm1:
No problem to me. It was 5 minutes telling Claude what to do.

Issue #1046, opened 2026-02-05
Web app doesn't know the server version


jeremyfowers:
<img width="3452" height="2176" alt="Image" src="https://github.com/user-attachments/assets/82f96556-a8aa-4b58-8b1a-9eeb0f5a8133" />

the electron app is fine though

superm1:
Huh - is this a Windows only issue?  I didn't see that on Linux during development.

superm1:
It was a regression from the later API key changes.  That PR should fix it.

PR #1050, opened 2026-02-06
Fix web-app about dialog

superm1:
The about dialog api key needs a default value set in `window.api` emulation to prevent undefined errors in the web-app.


superm1:
Should fix https://github.com/lemonade-sdk/lemonade/issues/1046

PR #1057, opened 2026-02-06
Adds LEMONADE_API_KEY in secrets file

sofiageo:
Added secrets to align with AUR package. If you think it's too much I can forget about it :) 

The file should probably have 660 permissions while lemonade.conf should probably have 644 permissions. But I'm not sure where we set this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1095,6 +1095,12 @@ if(UNIX AND NOT APPLE)
     endif()
     install(FILES ${CMAKE_SOURCE_DIR}/data/lemonade.conf
         DESTINATION ${CMAKE_INSTALL_FULL_SYSCONFDIR}/lemonade
+        PERMISSIONS OWNER_READ OWNER_WRITE GROUP_READ GROUP_WRITE WORLD_READ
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1093,7 +1093,9 @@ if(UNIX AND NOT APPLE)
             )
         ")
     endif()
-    install(FILES ${CMAKE_SOURCE_DIR}/data/lemonade.conf
+    install(FILES
+        ${CMAKE_SOURCE_DIR}/data/lemonade.conf
+        ${CMAKE_SOURCE_DIR}/data/secrets.conf
```
superm1: can you do your proposed permissions here?

https://stackoverflow.com/questions/19399705/cmake-how-to-change-file-permissions-when-installing

superm1 (commented):
Mostly looks good to me, just fix permissions during install.

CMakeLists.txt
sofiageo: Added and tested in Arch Linux

superm1 (approved):

Issue #1049, opened 2026-02-05
Improve Canter panel look and Feel


danielholanda:
# Description

Improve center panel look and feel, which was changed here
https://github.com/lemonade-sdk/lemonade/pull/1045

danielholanda:
Solved as described here: https://github.com/lemonade-sdk/lemonade/pull/1058

PR #1058, opened 2026-02-07
Cleanup Marketplace Presentation

danielholanda:
# Description

This PR brings back the models look of the center panel while now clearly pointing to the shiny new marketplace.
image

If we learned one thing, was to not leave a "coming soon" banner for something that is not currently being activelly developed XD.

Closes https://github.com/lemonade-sdk/lemonade/issues/1049

### DIFF
--- a/docs/server/lemonade-server-cli.md
+++ b/docs/server/lemonade-server-cli.md
@@ -123,7 +123,7 @@ lemonade-server pull <model_name> [options]
 | Option | Description | Required |
 |--------|-------------|----------|
 | `--checkpoint CHECKPOINT` | Hugging Face checkpoint in the format `org/model:variant`. For GGUF models, the variant (after the colon) is required. Examples: `unsloth/Qwen3-8B-GGUF:Q4_0`, `amd/Qwen3-4B-awq-quant-onnx-hybrid` | For custom models |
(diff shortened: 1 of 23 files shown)


src/app/src/renderer/CenterPanel.tsx
```diff
@@ -41,143 +167,245 @@ const CenterPanel: React.FC<CenterPanelProps> = memo(({ isVisible, currentView,
...
-          console.log('Using remote marketplace URL');
           setMarketplaceUrl(REMOTE_MARKETPLACE_URL);
         } else {
           throw new Error(`HTTP ${response.status}`);
         }
       } catch (error) {
         console.warn('Remote marketplace unavailable, falling back to local:', error);
 
-        // Get local marketplace URL via Electron API
         if (window.api?.getLocalMarketplaceUrl) {
```
superm1, line 177: doesn't this need an emulation in server.cpp?
We shouldn't have new functions using window.api unless it's done in both server (for web-app) and electron.

jeremyfowers: This is just for local development using `npm start` and isn't active in production.

jeremyfowers:
May revisit later

PR #1059, opened 2026-02-07
Remove some unneeded messages from window.api emulation

superm1:
I noticed these aren't actually used.

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -476,8 +476,6 @@ window.api = {
         }
         return 'Unknown';
     },
-    pauseDownload: () => {},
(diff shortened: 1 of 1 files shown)


bitgamma:
looks good to me, at first glance `pauseDownload` and `cancelDownload` also seem to be unused. They seem to be implemented in `DownloadTracker` instead of `window.api`. Did you try removing those as well?

superm1:
> looks good to me, at first glance pauseDownload and cancelDownload also seem to be unused. They seem to be implemented in DownloadTracker instead of window.api. Did you try removing those as well?

Thanks, agreed.

PR #1060, opened 2026-02-07
Add a redirection to web app for mobile users

superm1:
Alternative to https://github.com/lemonade-sdk/lemonade/pull/1024


jeremyfowers (approved):
Works great on my iphone!

PR #1061, opened 2026-02-07
Polish the app center panel

jeremyfowers:
image

image

### DIFF
--- a/src/app/src/renderer/CenterPanel.tsx
+++ b/src/app/src/renderer/CenterPanel.tsx
@@ -28,8 +28,18 @@ const CenterPanel: React.FC<CenterPanelProps> = memo(({ isVisible, currentView, 
       setIsLoading(true);
       setHasError(false);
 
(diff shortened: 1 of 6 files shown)


src/app/src/renderer/CenterPanel.tsx
```diff
@@ -28,8 +28,18 @@ const CenterPanel: React.FC<CenterPanelProps> = memo(({ isVisible, currentView,
       setIsLoading(true);
       setHasError(false);
 
+      // Try local marketplace first, fall back to remote
+      if (window.api?.getLocalMarketplaceUrl) {
```
superm1, line 32: You're sure it's OK that the server doesn't have this for `window.api` emulation right?

superm1 (approved):
It certainly looks fine to me.  I just want to make sure the `window.api` emulation isn't causing a problem from the change for web-app.

Issue #1047, opened 2026-02-05
add image edit endpoint


bitgamma:
Support 
- https://platform.openai.com/docs/api-reference/images/createEdit and 
- https://api.openai.com/v1/images/variations

probably not all properties will be supported, depends on sd-cpp. Also we'll need to add an editing model like Klein 4b/9b

kenvandine:
I've published this on the edge/sdd-image-edit channel for the snap for testing. 

```
sudo snap refresh --channel=edge/sdd-image-edit lemonade-server
```

PR #1063, opened 2026-02-07
Add support for multiple checkpoints in models

### LINKED ISSUE #1009: Add more AI Image Models
I think there are a lot of great open-source AI image models which are better than the current stable diffusion models like for example unsloth/Z-Image-GGUF is the best one right now.

Can there be more added? Because I tried for example to add unsloth/Z-Image-GGUF from HuggingFace but it wasn't possible.

Thanks in advance!

bitgamma:
Closes #1009 
Prerequisite for #1047.

Proposal:

define a new field called `checkpoints` (plural) in `server_models.json` and `user_models.json` file. If `checkpoints` is available, it is expected to be an object with the following format: `{ type1: checkpoint1, type2: checkpoint2}`. Eg:

```
"Flux-2-Klein-9B-GGUF": {
    "checkpoints": {
        "main": "unsloth/FLUX.2-klein-9B-GGUF:flux-2-klein-9b-Q8_0.gguf",
        "text_encoder": "unsloth/Qwen3-8B-GGUF:Qwen3-8B-Q8_0.gguf",
        "vae": "Comfy-Org/vae-text-encorder-for-flux-klein-9b:split_files/vae/flux2-vae.safetensors"
    },
    "recipe": "sd-cpp",
    ...
}
```

Type can be anything, but currently used values would be:

* `main`: the main model, will be used in logs, server responses and all other places where `checkpoint` is used today
* `mmproj`: the mmproj for vision models.
* `vae`: vae for image models
* `text_encoder`: text encoder for image models
* `npu_cache`: NPU cache files (formatted as usual hf checkpoint instead of being on two separate fields)

To retain backwards compatibility, `checkpoint` and `mmproj` will continue to exist but will be automatically mapped to/from `checkpoints`. In particular using `checkpoint` directly in the json description would remain the recommended approach for models with a single file.  

The `mmproj` property will be supported. Up for the debate whether its usage should be considered legacy or if it remains preferred for conciseness. The object-based `checkpoints` approach is more flexible as it allows downloading the mmproj from a different repo (which is rare, but today I saw an example where this would have been useful).

The `npu_cache_repo` and `npu_cache_filename` haven't been yet exposed outside of `server_models.json` so there is no need to retain backwards compatibility, it can be handled with a refactoring.

In scope:

- Applying the described changes to the `ModelManager` and the recipes
- Modify the sd_server backend to accept `vae` and `text_encoder`
- Also accept GGUF files for image generation (they are supported by sdcpp)
- Add Z-Image-Turbo, Flux.2 Klein 4B, Flux.2 Klein 9B

Maybe:

- Move NPU cache downloading/path resolving in model_manager. This can be done trivially by removing some code. The reason I haven't done it is that the current logic only downloads it when the NPU backend is selected and I retained it. If we can live with the NPU cache being always downloaded then it is a nice simplification in the code

Out of scope, but will needed in a future PR:

- Modify the CLI/HTTP pull parameters to support the additional checkpoints
- Modify the UI to support the additional parameters when adding a model

**TESTING NOTES**

- Klein-4B/9B require a newer build of sd-cpp to run. I didn't update it here cause I know stuff is coming on that front
- Z-Image-Turbo works with the current version


bitgamma:
cc @jeremyfowers @ramkrishna2910

jeremyfowers:
@makn87amd @ramkrishna2910 please have the first look and determine if this implementation enables the scenarios you have envisioned.

jeremyfowers:
Large surface area that needs to be tested:
- [x] Legacy `user.` VLMs still work
- [x] Register new `user.` VLMs download and work
- [x] New Flux and Z-Image models
- Downloads and their progress callbacks still work
    - [x] Normal models
    - [x] FLM models

bitgamma:
I tested these except FLM. Tested only on Linux

ramkrishna2910:
Feature works well:
image

ramkrishna2910 (approved):
Works great! Thanks for the great contribution @bitgamma!

PR #1064, opened 2026-02-07
add ENV for LEMONADE_MAX_LOADED_MODELS

bitgamma:
add ENV for LEMONADE_MAX_LOADED_MODELS to make it easy to configure on Linux by editing the config files


bitgamma:
@sofiageo 

1. Yes, this is correct
2. Done!
3. Yes, this is the expected behavior

PR #1065, opened 2026-02-07
handle model start event in chat

### LINKED ISSUE #960: App bug: send during load reverses the load
This behavior really annoys me:

1. Model A is loaded
2. I push load on Model B, which takes a while to load
3. I type and send a message. The model dropdown in LLM Chat still says Model A.
4. This causes Model B's loading to be canceled. Model A is reloaded and gets the message.

Solution: disable the send button during model loading.

cc @danielholanda

choppu:
Closes #960. 

Instead of disabling the send button, I made it so that as soon as you click the model load button in ModelManager, the selected model in chat window is updated immediately. This should fix the problem you describe.

Trying to disable the send button, I encountered a lot of edge cases, that would make the window permanently blocked.


jeremyfowers (approved):
This is a perfect, elegant solution! Much better than my original suggestion. Thanks very much @choppu, this will save me a lot of time and headache!

PR #1067, opened 2026-02-07
Update systemd dependency for Arch Linux builds

sofiageo:


superm1:
Try using:

```
pacman -S --needed <packages...>
```

sofiageo:
sorry I missed one, should be ok now

PR #1071, opened 2026-02-08
Kokoro UI  integration

choppu:
Started implementing text to speech UI.

### DIFF
--- a/src/app/src/renderer/ChatWindow.tsx
+++ b/src/app/src/renderer/ChatWindow.tsx
@@ -1033,7 +1033,7 @@ const sendMessage = async () => {
     e.stopPropagation(); // Prevent closing when clicking inside the edit area
   };
 
-  // Handle text to sreech conversion using Kokoro
+  // Handle text to speech conversion using Kokoro
   useEffect(() => {
     if(currentAudio) {
       currentAudio.addEventListener('ended', stopAudio);
--- a/src/app/src/renderer/AudioButton.tsx
+++ b/src/app/src/renderer/AudioButton.tsx
@@ -11,7 +11,7 @@ interface  AudioButtonProps {
   buttonContext: {buttonId: number, audioState: number};
 }
 
-const AudioButton: React.FC<AudioButtonProps> = ({ textMessage, buttonIndex, onClickFunction, buttonContext}) => {
+const AudioButton: React.FC<AudioButtonProps> = React.memo(function AudioButton({ textMessage, buttonIndex, onClickFunction, buttonContext}) {
   const LoadingIcon = () => {
     return (
       <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18" width="18px" height="18px"  style={{opacity:1}}>
@@ -64,6 +64,8 @@ const AudioButton: React.FC<AudioButtonProps> = ({ textMessage, buttonIndex, onC
       <button className="message-play-button" onClick={() => onClickFunction(textMessage, buttonIndex)} disabled={buttonContext.audioState == LOADING}>{renderButtonIcon()}</button>
     </div>
   );
-}
+}, (prevProps, nextProps) => {
+  return (nextProps.buttonContext.buttonId === prevProps.buttonContext.buttonId) && (nextProps.buttonContext.audioState === prevProps.buttonContext.audioState) && (nextProps.textMessage == prevProps.textMessage) && (nextProps.onClickFunction == prevProps.onClickFunction);
+});
 
 export default AudioButton;


src/app/src/renderer/ChatWindow.tsx
```diff
@@ -999,6 +1043,129 @@ const sendMessage = async () => {
...
+        let audioState = stopAudio();
+        setAudioButtonContext({
+          ...audioButtonContent,
+          isAudioPlaying: audioState
+      });
+      });
+      currentAudio.addEventListener('error', () => {
+        console.log("Error playing audio");
+        stopAudio();
+      });
```
ramkrishna2910: The cleanup function passes new anonymous arrow functions to removeEventListener, which don't match the ones passed to addEventListener. These are different references, so the listeners are never removed they accumulate on every re-render causing an event listener leak.

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -999,6 +1043,129 @@ const sendMessage = async () => {
...
+      }
+
+      const respBlob = await response.blob();
+      const audioUrl = URL.createObjectURL(respBlob);
+      playAudio(audioUrl);
+      return PLAYING;
+    } catch(err: any) {
+      console.log(`Error: ${err}`);
+      setPressedAudioButton(-1);
+      return PAUSED;
```
ramkrishna2910: playAudio() calls URL.createObjectURL() but never revokes the previous URL if one already exists. Rapid clicks on different messages will leak blob URLs. Call stopAudio() at the start of playAudio, or revoke the old URL before creating a new one.

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -999,6 +1043,129 @@ const sendMessage = async () => {
...
+      currentAudio?.removeEventListener('error', () => {
+        console.log("Error playing audio");
+        stopAudio();
+      });
+    }
+    
+  }, [currentAudio, audioButtonContent]);
+
+  const handleTextToSpeech = async (message: MessageContent) => {
+    const textToSpeechModel = 'kokoro-v1';
```
ramkrishna2910, line 1071: Should the model name be hard coded here? What if we add support for other TTS models?

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -281,10 +297,29 @@ useEffect(() => {
     setIsUserAtBottom(true);
   };
 
+const excludeTextToSpeechModel = (data: any) => {
+  const loadedModels = data.all_models_loaded;
+  const loadedModel = loadedModels.find((model: any) => model.model_name === data?.model_loaded);
+
+  let isTextToSpeechModel = (loadedModel.recipe === 'kokoro');
```
ramkrishna2910: loadedModels.find() can return undefined if no model matches, but the next line accesses loadedModel.recipe unconditionally. This will throw a TypeError at runtime. Use optional chaining: `loadedModel?.recipe === 'kokoro'`

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -1946,7 +2145,8 @@ const sendMessage = async () => {
                   className={`chat-message ${message.role === 'user' ? 'user-message' : 'assistant-message'} ${
                     message.role === 'user' && !isLoading ? 'editable' : ''
                   } ${isGrayedOut ? 'grayed-out' : ''} ${editingIndex === index ? 'editing' : ''}`}
-                >
+                  >
+                  <AudioButton textMessage={message.content} index={index} />
```
ramkrishna2910: AudioButton is rendered inside every message div, including user messages. Do we need it on user messages or only the LLM responses.

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -1418,6 +1586,37 @@ const sendMessage = async () => {
     </div>
   );
 
+  const AudioButton = ({ textMessage, index }: {textMessage: MessageContent, index: number}) => {
```
ramkrishna2910: AudioButton is defined as a component inside the parent render body. It gets re-created on every render, which defeats React reconciliation and causes unnecessary DOM churn. Move it outside ChatWindow or wrap it with React.memo.

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -109,6 +114,17 @@ const ChatWindow: React.FC<ChatWindowProps> = ({ isVisible, width }) => {
   }>>([]);
   const [isGeneratingImage, setIsGeneratingImage] = useState(false);
 
+  // Text to speech state
```
ramkrishna2910, line 113: PAUSED, LOADING, PLAYING are defined as local variables inside the component body, so they're re-created on every render. Move them to module scope (above the component) or use as const enum.

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -1418,6 +1586,37 @@ const sendMessage = async () => {
     </div>
   );
 
+  const AudioButton = ({ textMessage, index }: {textMessage: MessageContent, index: number}) => {
+    return (
+      <div key={index} className="message-play-button-container">
```
ramkrishna2910: The three SVGs (loading dots, pause, speaker) are large and make the AudioButton JSX hard to read. The existing codebase also uses inline SVGs, but they're simpler (16x16, short paths). Consider extracting these into const variables above the component (e.g. const SpeakerIcon = <svg>...</svg>) to keep the JSX clean — consistent with how the rest of the file handles more complex markup.

src/app/src/renderer/ChatWindow.tsx
choppu: I was thinking to add TTS model and TTS voice setting to App Settings (in a separate pull request). If the TTS functionality is used for both user and LLM messages, a separate setting for LLM messages voice and user messages voice can be added, instead of a single TTS voice setting. If it's an option, I will be happy to add those setting to App Settings.

choppu:
@ramkrishna2910 , thank you for reviewing my pr. I have addressed all your comments.

src/app/src/renderer/ChatWindow.tsx
ramkrishna2910: Makes sense. Can u open an issue as a follow up to this?
How are u planning to expose the settings button in the app UI?

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -999,6 +1033,90 @@ const sendMessage = async () => {
     e.stopPropagation(); // Prevent closing when clicking inside the edit area
   };
 
+  // Handle text to sreech conversion using Kokoro
```
ramkrishna2910: Typo: speech

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -1946,7 +2065,8 @@ const sendMessage = async () => {
                   className={`chat-message ${message.role === 'user' ? 'user-message' : 'assistant-message'} ${
                     message.role === 'user' && !isLoading ? 'editable' : ''
                   } ${isGrayedOut ? 'grayed-out' : ''} ${editingIndex === index ? 'editing' : ''}`}
-                >
+                  >
+                  <AudioButton textMessage={message.content} buttonIndex={index} onClickFunction={handleAudioButtonClick} buttonContext={{buttonId: pressedAudioButton, audioState: audioState}} />
```
ramkrishna2910, line 2069: This creates a new object literal on each render, causing AudioButton to re-render even if values haven’t changed. Consider memoizing this or using React.memo with a custom comparator. Can be handled in the follow up PR.

ramkrishna2910 (approved):
Looks great! Added a minor comments that can be solved in the follow up PR. Thanks for the contribution. Approved.

src/app/src/renderer/ChatWindow.tsx
choppu: Thank you again for reviewing my code 🙏 I have added React.memo to AudioButton component and fixed typo.

PR #1078, opened 2026-02-09
Add a systemd migration script

superm1:
I haven't tested this, it was originally written by claude around my guidance, but I made some modifications on things I didn't agree with initially.

I'm not sure this is the right location in the source tree for it, also I'm not sure if it should be source tree only or installed somewhere.  All points up for discussion.


sofiageo (commented):
I tested it with an empty /var/lib/lemonade/.cache and it works fine. However, I now have 20GB less in my drive because it is copying the files instead of moving them.

I didn't test it with overwriting my /var/lib/lemonade/.cache directories, because I've already added new models and I didn't want to lose them.

I think it would be better if the script tries to move the models from the home/.cache/lemonade/* and home/.cache/huggingface/hub/* to the /var/lib/lemonade/.cache equivalent directories instead of copying them. And not to try and remove the existing directories in the target /var/lib/lemonade/.cache

superm1:
> I tested it with an empty /var/lib/lemonade/.cache and it works fine. However, I now have 20GB less in my drive because it is copying the files instead of moving them.
> 
> I didn't test it with overwriting my /var/lib/lemonade/.cache directories, because I've already added new models and I didn't want to lose them.
> 
> I think it would be better if the script tries to move the models from the home/.cache/lemonade/* and home/.cache/huggingface/hub/* to the /var/lib/lemonade/.cache equivalent directories instead of copying them. And not to try and remove the existing directories in the target /var/lib/lemonade/.cache

Thanks!  Adjusted.

sofiageo:
I will test it again later tonight

jeremyfowers:
@sofiageo did you have a chance to test again?

superm1:
I'm not sure it really needs a retest - it was just a change from `cp` to `mv`

jeremyfowers:
> I'm not sure it really needs a retest - it was just a change from `cp` to `mv`



> I'm not sure it really needs a retest - it was just a change from `cp` to `mv`

Ok I'll just merge it then - waiting a little for CI to settle though.

sofiageo:
I'm not sure why I didn't get these messages :( sorry everyone, must be something with my notifications

jeremyfowers (approved):
LGTM

sofiageo (approved):
Tested it now, I had a perfect case for it.

```
Found: /home/sofiageo/.cache/lemonade (119M)
  -> Will migrate to: /var/lib/lemonade/.cache/lemonade
Found: /home/sofiageo/.cache/huggingface (17G)
  -> Will migrate to: /var/lib/lemonade/.cache/huggingface

Migrating: /home/sofiageo/.cache/lemonade
       to: /var/lib/lemonade/.cache/lemonade
  Warning: Destination already exists - merging files
  Merging into existing destination...

sent 17,526,020,187 bytes  received 1,110 bytes  432,741,266.59 bytes/sec
total size is 17,521,738,812  speedup is 1.00
  Updating ownership to lemonade:lemonade...
  Done

Migration completed successfully!
```

I only had to remove the empty directories after run from the original ~/.cache/huggingface/hub directory.

PR #1084, opened 2026-02-09
Sdcpp vulkan integration

makn87amd:
Add Vulkan support to sd-cpp.

Vulkan seems to have more robust for SOTA models and may offer different performance compared with ROCm

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -120,6 +120,13 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"amd_dgpu", {"gfx110X", "gfx120X"}},
     }},
 
+    // stable-diffusion.cpp - Vulkan backend (cross-platform GPU)
+    {"sd-cpp", "vulkan", {"windows", "linux"}, {
+        {"cpu", {"x86_64"}},
+        {"amd_igpu", {}},
+        {"amd_dgpu", {}},
+    }},
+
     // stable-diffusion.cpp - CPU backend (Windows/Linux x86_64)
     {"sd-cpp", "cpu", {"windows", "linux"}, {
         {"cpu", {"x86_64"}},
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -121,10 +121,12 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
     }},
 
     // stable-diffusion.cpp - Vulkan backend (cross-platform GPU)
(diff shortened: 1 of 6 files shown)


makn87amd:
Forgot to add vulkan checks to CI workflow

src/cpp/server/system_info.cpp
```diff
@@ -120,6 +120,15 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"amd_dgpu", {"gfx110X", "gfx120X"}},
     }},
 
+    // stable-diffusion.cpp - Vulkan backend (cross-platform GPU)
```
jeremyfowers, line 123: Should vulkan be the default over ROCm if its required for complex models to work?

jeremyfowers:
FYI: https://github.com/lemonade-sdk/lemonade/pull/1089 has a bunch of general fixes/improvements for SD support that will make this PR easier to test.

src/cpp/server/system_info.cpp
makn87amd: That makes sense.

jeremyfowers:
@ramkrishna2910 when testing this PR make sure to use `lemonade-server serve --sdcpp vulkan` 

Everything else will fail silently back to CPU, regardless of defaults. #1089 fixes that.

ramkrishna2910:
@makn87amd is there a minimum driver requirement? 
SD-CPU works for me, I am getting this error on vulkan
```
[ERROR] sd-server process has terminated with exit code: -1073740791
[ERROR] This usually means:
  - Missing required drivers or dependencies
  - Incompatible model file
  - Try running the server manually to see the actual error
 ```

File: src/cpp/server/system_info.cpp [hunk 2]
```diff
@@ -120,6 +120,15 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
...
+    {"sd-cpp", "vulkan", {"windows"}, {
```
jeremyfowers, line 124: Please add Linux Vulkan: sd-master-636d3cb-bin-Linux-Ubuntu-24.04-x86_64-vulkan.zip

ramkrishna2910:
@makn87amd given that we are still seeing issues for this on gfx1150 should we wait for that to be resolved? On Halos in any case we would prefer ROCm over vulkan.

makn87amd:
I think that's perfectly reasonable.

jeremyfowers:
Resolved by https://github.com/lemonade-sdk/lemonade/pull/2039

PR #1086, opened 2026-02-10
Rework --max-loaded-models

jeremyfowers:
Before: 1 integer per model type
Now: 1 integer total, which applies to all model types

This PR also includes 3 new models in server_models.json that will be used in upcoming work.

### DIFF
--- a/src/cpp/server/cli_parser.cpp
+++ b/src/cpp/server/cli_parser.cpp
@@ -28,7 +28,7 @@
 
 namespace lemon {
 
-static void add_serve_options(CLI::App* serve, ServerConfig& config, int& max_models_arg) {
+static void add_serve_options(CLI::App* serve, ServerConfig& config) {
     serve->add_option("--port", config.port, "Port number to serve on")
         ->envname("LEMONADE_PORT")
         ->type_name("PORT")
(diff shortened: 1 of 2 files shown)


src/cpp/server/cli_parser.cpp
```diff
@@ -57,33 +57,22 @@ static void add_serve_options(CLI::App* serve, ServerConfig& config, std::vector
...
+    // Multi-model support: Max loaded models per type slot
+    serve->add_option("--max-loaded-models", max_models_arg,
+                   "Max models per type slot (LLMs, audio, image, etc.). Use -1 for unlimited.")
         ->envname("LEMONADE_MAX_LOADED_MODELS")
-        ->type_name("N [E] [R] [A] [I]")
-        ->expected(1, 5)
-        ->default_val(std::vector<int>{config.max_llm_models, config.max_embedding_models, config.max_reranking_models, config.max_audio_models, config.max_image_models})
+        ->type_name("N")
+        ->default_val(config.max_loaded_models)
         ->check([](const std::string& val) -> std::string {
```
bitgamma, line 66: you can remove this check entirely, CLI11 does this validation/conversion by default for integers

bitgamma: ah got it, -1 is allowed but not 0

src/cpp/server/cli_parser.cpp
```diff
@@ -160,24 +149,9 @@ int CLIParser::parse(int argc, char** argv) {
...
-                config_.max_embedding_models = max_models_vec_[1];
-                config_.max_reranking_models = max_models_vec_[2];
-            }
-            if (max_models_vec_.size() > 3) {
-                config_.max_audio_models = max_models_vec_[3];
-            }
-            if (max_models_vec_.size() > 4) {
-                config_.max_image_models = max_models_vec_[4];
-            }
+        // Process --max-loaded-models value (-1 = unlimited, or positive integer)
```
bitgamma: I think you can let CLI11 write directly to config_.max_loaded_models instead of using a temporary var?

jeremyfowers: You're right, that simplifies it further!

PR #1089, opened 2026-02-10
Improve SD support in the CLI and app

jeremyfowers:
- Display the Load Options menu and backend dropdown for SD in the app
- Remove hardcoded backend selection in the C++ that was blocking the rocm default from working
- Fix for all recipes: pass the default backend from CLI to lemonade-router on creation if no backend is selected

### DIFF
--- a/test/server_system_info.py
+++ b/test/server_system_info.py
@@ -184,12 +184,13 @@ MOCK_HARDWARE_CONFIGS = {
         "expected_supported": {
             "llamacpp": ["vulkan", "rocm", "cpu"],
             "whispercpp": ["npu", "cpu"],  # npu supported on XDNA2, cpu on x86_64
-            "sd-cpp": ["cpu", "rocm"],
+            "sd-cpp": ["cpu"],
             "flm": ["default"],
             "ryzenai-llm": ["default"],
         },
         "expected_unsupported": {
             "llamacpp": ["metal"],
+            "sd-cpp": ["rocm"],
         },
     },
     # Windows x86_64 with no AMD GPU (Intel/NVIDIA only)


bitgamma:
Part of the changes to the cpp code are already in #1063 in the last commit. sd_server becomes even simpler because the class variable is not needed at all (it is only ever used in load). Works the same as llamacpp

src/cpp/server/recipe_options.cpp
```diff
@@ -133,6 +133,14 @@ void RecipeOptions::add_cli_options(CLI::App& app, json& storage) {
             const auto& result = backend_cache[recipe];
             std::string default_backend = result.backends.empty() ? "" : result.backends[0];
 
+            // Pre-populate storage with the dynamically detected default so it's
```
bitgamma, line 136: while it is true that default_val is only for the help, the whole inherit system of the recipe options automatically add the default. Otherwise it wouldn't have worked for llamacpp or whisper either

bitgamma: oh this is dynamic vs static got it. 👍

superm1:
Given those CI failures; I think you "have" to land my other one first even if it's not working due to the missing libomp on your system.  The paths are wrong for the Linux binaries.

jeremyfowers:
> Given those CI failures; I think you "have" to land my other one first even if it's not working due to the missing libomp on your system. The paths are wrong for the Linux binaries.

Yeah mine can merge after @superm1 and @bitgamma PRs... we just need to retest everything before release since there's a lot of churn.

fyi @ramkrishna2910

jeremyfowers:
> Given those CI failures; I think you "have" to land my other one first even if it's not working due to the missing libomp on your system. The paths are wrong for the Linux binaries.

the test failure was actually that the CI check was putting `""` for the CPU backend test instead of `cpu`. In turn, it defaulted to ROCm, and then failed because we don't have your branch.

PR #1091, opened 2026-02-11
Add support for NPU detection on Linux

superm1:
The baseline detection will be based upon kernel version, and based upon stability measurements has been decided that will run on Linux 7.0 or later.

This doesn't actually glue any `flm` launching code and it certainly doesn't mark `flm` as supported on Linux, but baby steps.


jeremyfowers:
Is there a way for me to test that this works as expected on a Strix Point? Or do you want to just merge and then refine later?

superm1:
This is intended to merge now and refine later.
It's baby steps working towards flm working.

PR #1102, opened 2026-02-11
Clean the lemonade/bin folder on upgrade from 9.3.1 or below

### LINKED ISSUE #1072: Hybrid models are failed to load on Windows
I got AMD Strix Halo desktop (Beelink GTR9 Pro) with Windows 11 Pro. It uses AMD RYZEN AI MAX+ 395 w/ Radeon 8060S. It is configured to use 96 GB video memory.

I just downloaded **Lemonade Server 9.3.1** and tried to load several Hybrid models and got error about `onnxruntime_providers_ryzenai.dll` is missing.
Then I also tried to test the command from the homepage:

Image

```
lemonade-server run Llama-3.2-1B-Instruct-Hybrid
```

This command fails the same way. See attached log file: [HybrydLLM_Loading_Failed_Log.txt](https://github.com/user-attachments/files/25162558/HybrydLLM_Loading_Failed_Log.txt).

 I noticed the warning under the installation instructions: "Download and install the [NPU Driver](https://account.amd.com/en/forms/downloads/ryzenai-eula-public-xef.html?filename=NPU_RAI1.5_280_WHQL.zip) before proceeding." I did download and install the driver but it didn't change anything.

Seems like the `onnxruntime_providers_ryzenai.dll` is expected to be present in the installation package for the  **Lemonade Server 9.3.1** but it is missing. 

Please provide a workaround or fix the  **Lemonade Server ** Windows installer so it includes all required DLLs to run Hybrid models.

jeremyfowers:
Closes #1072 

We had some fun renaming folders in prior PRs, and this is causing errors in production. See #1072 for an example.

This PR gives us a clean slate in 9.3.2+

How it works
- Upgrade from <= 9.3.1: hardware_info.json exists with old version → load_hardware_info() returns empty (version mismatch) → enters upgrade path → perform_upgrade_cleanup() reads old version, sees it's < 9.3.2, deletes the bin directory via fs::remove_all with error_code (no-throw)
- New install: No hardware_info.json → cache_exists is false → skips cleanup entirely
- Upgrade from > 9.3.1: Old version is >= 9.3.2 → is_version_less_than check fails → no deletion
- Errors: All wrapped in try/catch(...) and std::error_code — silently passes on any failure

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -2431,10 +2431,16 @@ void SystemInfoCache::perform_upgrade_cleanup() {
 
         std::string old_version = cache_data["version"];
 
-        // When upgrading from 9.3.1 or earlier, delete the backend bin directory.
-        // Backend binaries from older versions may be incompatible and need to be
-        // re-downloaded. Using < 9.3.2 is equivalent to <= 9.3.1.
-        if (is_version_less_than(old_version, "9.3.2")) {
+        // Delete the backend bin directory when upgrading from a version older
+        // than clear_bin_if_lemonade_below (defined in backend_versions.json).
+        // Backend binaries from older versions may be incompatible and need to
+        // be re-downloaded.
+        std::string config_path = utils::get_resource_path("resources/backend_versions.json");
+        std::ifstream config_file(config_path);
+        json config = json::parse(config_file);
+        std::string cleanup_version = config.value("clear_bin_if_lemonade_below", "0.0.0");
+
+        if (is_version_less_than(old_version, cleanup_version)) {
             std::string bin_dir = get_cache_dir() + "/bin";
             std::error_code ec;
             fs::remove_all(bin_dir, ec);
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -23,5 +23,6 @@
   },
   "kokoro": {
     "cpu": "b14"
-  }
+  },
+  "clear_bin_if_lemonade_below": "9.3.2"
 }


src/cpp/server/system_info.cpp
```diff
@@ -2413,6 +2413,38 @@ void SystemInfoCache::clear() {
...
+        if (!cache_data.contains("version")) {
+            return;
+        }
+
+        std::string old_version = cache_data["version"];
+
+        // When upgrading from 9.3.1 or earlier, delete the backend bin directory.
+        // Backend binaries from older versions may be incompatible and need to be
+        // re-downloaded. Using < 9.3.2 is equivalent to <= 9.3.1.
+        if (is_version_less_than(old_version, "9.3.2")) {
```
superm1: IMO this isn't the last time - how about putting this in the json to set the minim version?

jeremyfowers: Good idea.

vgarmash:
Is it possible for you to add some integration test to your CI/CD pipeline so it will imitate the upgrade scenario? I see that you already have tests on fresh install, so you can add the test where you have the older Lemonade server installed and initialized with some models and then you need to upgrade your installation - that's what we all do as users.

src/cpp/server/system_info.cpp
jeremyfowers: Implemented here: https://github.com/lemonade-sdk/lemonade/pull/1102/commits/a52d3e005e9c483d1168326503461a8783d7c6d4

And tested locally (RIP my bin folder).

jeremyfowers:
> Is it possible for you to add some integration test to your CI/CD pipeline so it will imitate the upgrade scenario? I see that you already have tests on fresh install, so you can add the test where you have the older Lemonade server installed and initialized with some models and then you need to upgrade your installation - that's what we all do as users.

This is a good idea but we don't have the Windows CI resources right now.

To do this thoroughly, we'd need to install the current release, run tests to make sure its initialized, then install from the CI checkout, and run the tests again. And do that for all backends. This would about double total Windows CI time on self-hosted runners, which is already a problem.

If upgrade pain is a recurring challenge I could add something like this as a nightly CI. But hopefully we don't see this too much. 

Sorry about the inconvenience with the last upgrade @vgarmash.

PR #1106, opened 2026-02-12
Add a message for older kernel on gfx1151

superm1:
If running on an older kernel that is missing VGPR fixes, notify the user by opening a web page to tell them to update.

Implemented with a new system-checks API that the frontend calls. Can be extended for other purposes (such as NPU driver on Windows).

Here's how it looks on Linux for an affected system.

image

### DIFF
--- a/src/app/src/renderer/App.tsx
+++ b/src/app/src/renderer/App.tsx
@@ -7,8 +7,9 @@ import CenterPanel, { CenterPanelView } from './CenterPanel';
 import ResizableDivider from './ResizableDivider';
 import DownloadManager from './DownloadManager';
 import StatusBar from './StatusBar';
+import SystemChecksModal from './SystemChecksModal';
 import { ModelsProvider } from './hooks/useModels';
-import { SystemProvider } from './hooks/useSystem';
+import { SystemProvider, useSystem } from './hooks/useSystem';
 import { DEFAULT_LAYOUT_SETTINGS } from './utils/appSettings';
 import '../../styles.css';
 
@@ -20,13 +21,16 @@ const LAYOUT_CONSTANTS = {
   absoluteMinWidth: 400,
 };
 
-const App: React.FC = () => {
+// Inner component that can use SystemProvider context
+const AppContent: React.FC = () => {
+  const { systemChecks, shouldShowSystemChecks, dismissSystemChecks } = useSystem();
   const [isChatVisible, setIsChatVisible] = useState(DEFAULT_LAYOUT_SETTINGS.isChatVisible);
   const [isModelManagerVisible, setIsModelManagerVisible] = useState(DEFAULT_LAYOUT_SETTINGS.isModelManagerVisible);
   const [isCenterPanelVisible, setIsCenterPanelVisible] = useState(DEFAULT_LAYOUT_SETTINGS.isCenterPanelVisible_v2);
   const [centerPanelView, setCenterPanelView] = useState<CenterPanelView>('menu');
   const [isLogsVisible, setIsLogsVisible] = useState(DEFAULT_LAYOUT_SETTINGS.isLogsVisible);
   const [isDownloadManagerVisible, setIsDownloadManagerVisible] = useState(false);
+  const [isSystemChecksModalOpen, setIsSystemChecksModalOpen] = useState(false);
   const [modelManagerWidth, setModelManagerWidth] = useState(DEFAULT_LAYOUT_SETTINGS.modelManagerWidth);
   const [chatWidth, setChatWidth] = useState(DEFAULT_LAYOUT_SETTINGS.chatWidth);
   const [logsHeight, setLogsHeight] = useState(DEFAULT_LAYOUT_SETTINGS.logsHeight);
@@ -115,6 +119,17 @@ const App: React.FC = () => {
       window.removeEventListener('download:chatComplete' as any, handleChatDownloadComplete);
     };
   }, []);
+
+  // Show system checks modal when ROCm usage is detected
+  useEffect(() => {
+    if (shouldShowSystemChecks) {
+      setIsSystemChecksModalOpen(true);
+      // Log issues to console for debugging
+      systemChecks.forEach(check => {
+        console.warn(`System check [${check.id}]:`, check.message);
+      });
+    }
+  }, [shouldShowSystemChecks, systemChecks]);
 
   useEffect(() => {
     const hasMainColumn = isCenterPanelVisible || isLogsVisible;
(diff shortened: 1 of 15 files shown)


superm1:
> I see what you’re going for here, but what happens if someone is using lemonade without having the app open? The current Windows NPU driver check runs as a blocking part of WrappedServer load/install for NPU engines, so it guarantees the user wont run into a hard-to-understand error message under any circumstances.

Well my thought is that anyone who sets up lemonade will have first used the electron or web app **at least once**.

Some other ideas:
* If you think it's important the same messaging could also be added into the server (IE run system check once and log something into lemonade logs).
* Could totally block support if the kernel check fails (big hammer)

jeremyfowers:
> Could totally block support if the kernel check fails (big hammer)

This is appropriate specifically for SD+ROCm+Linux, right?

But not for LLM. So IMO its better to only show the scary message when its relevant (user wants to activate SD+ROCm).

superm1:
> > Could totally block support if the kernel check fails (big hammer)
> 
> This is appropriate specifically for SD+ROCm+Linux, right?
> 
> But not for LLM. So IMO its better to only show the scary message when its relevant (user wants to activate SD+ROCm). 

Actually the exact same stability issues happen with llm as well. 
It's anything that needs to do a context switch can cause problems randomly. SD hits it quicker.

jeremyfowers:
> > > Could totally block support if the kernel check fails (big hammer)
> > 
> > 
> > This is appropriate specifically for SD+ROCm+Linux, right?
> > But not for LLM. So IMO its better to only show the scary message when its relevant (user wants to activate SD+ROCm).
> 
> Actually the exact same stability issues happen with llm as well. It's anything that needs to do a context switch can cause problems randomly. SD hits it quicker.

Good to know!

Ok I will try to test this by tricking one of my ubuntu PCs into thinking its a strix halo.

src/cpp/server/system_info.cpp
```diff
@@ -1032,6 +1032,52 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
...
+        if (!node_entry.is_directory()) continue;
+
+        std::string properties_file = node_entry.path().string() + "/properties";
+        if (!fs::exists(properties_file)) continue;
+
+        std::ifstream props(properties_file);
+        if (!props.is_open())
+            continue;
+
+        bool is_gfx1151 = false;
```
jeremyfowers, line 1087: Please use the cached system info struct already available instead of re-detecting the GPU version (DRY please).

superm1:
> Tested, my experience was that the first time I launched the app I got this popup. IMO this is not a good user experience because it immediately scares people who may not have any interest in running ROCm.
> I would much prefer for this popup to only show in the event that the user tries to install rocm or load a model onto rocm.

I think pushing to the first time ROCm is loaded makes sense.  I'll experiment a bit.

> To that end, could the load/install functions return a special http error code that the app catches and uses to display the popup?

I guess we could do that; but I saw this as being extensible to other error types about the stack (see the TODO for NPU).

I also want to eventually get people using the "system" llama.cpp if it's available, and was thinking this infrastructure could be used to message that too.

> If we go that route, I could see still keeping a way to poll for such information, but putting it in a notification drawer in the app instead of covering the full screen on first use. However, please consider: does this need to be a new endpoint, or can it just be a new field in system-info?

superm1:
OK, take a look again now.  I've pushed it so that the modal only comes up when user first tries to chat.

src/cpp/server/system_info.cpp
superm1: done

jeremyfowers:
> OK, take a look again now. I've pushed it so that the modal only comes up when user first tries to chat.

I don't have a Linux Strix Halo. How would you recommend I test this?

superm1:
> I don't have a Linux Strix Halo. How would you recommend I test this?

You can just hack up locally to make this match another system you do have to make the message show up.

src/app/src/renderer/SystemChecksModal.tsx
```diff
@@ -0,0 +1,137 @@
+import React, { useEffect, useRef } from 'react';
```
jeremyfowers: This is only showing for me on llamacpp, should it also show on sd-cpp?

src/app/src/renderer/SystemChecksModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { SystemCheck } from './utils/systemData';
```
jeremyfowers, line 2: This is showing for me when I send a message to an LLM I loaded with Vulkan. Should it only show on ROCm?

src/app/src/renderer/SystemChecksModal.tsx
superm1: Yes for now it only shows on llamacpp.

But yes we should show it on sd-cpp too.  I'll make some changes.

src/app/src/renderer/SystemChecksModal.tsx
superm1: Should only show with ROCm, will need to see how I can change this.

src/app/src/renderer/SystemChecksModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+interface SystemChecksModalProps {
```
jeremyfowers, line 4: The message shows on the first LLM Chat I send every time I open the app. Should it have a "dont show this message again" checkbox? I don't think we should be harassing people about changing their kernel, especially if their use-case is working ok.

superm1: OK, will see what I can do.

src/app/src/renderer/SystemChecksModal.tsx
jeremyfowers: The app regularly queries /health, which includes this kind of information:

```
{"all_models_loaded":[{"backend_url":"http://127.0.0.1:8001/v1","checkpoint":"LiquidAI/LFM2-1.2B-GGUF:LFM2-1.2B-Q4_K_M.gguf","device":"gpu","last_use":61796987,"model_name":"LFM2-1.2B-GGUF","recipe":"llamacpp","recipe_options":{"llamacpp_backend":"vulkan"},"type":"llm"},{"backend_url":"http://127.0.0.1:8002/v1","checkpoint":"stabilityai/sdxl-turbo:sd_xl_turbo_1.0_fp16.safetensors","device":"cpu","last_use":61780412,"model_name":"SDXL-Turbo","recipe":"sd-cpp","recipe_options":{"cfg_scale":1.0,"height":512,"sd-cpp_backend":"rocm","steps":4,"width":512},"type":"image"}]
```

Note the `"recipe_options":{"llamacpp_backend":"vulkan"},` and `"recipe":"sd-cpp","recipe_options":{"cfg_scale":1.0,"height":512,"sd-cpp_backend":"rocm","steps":4,"width":512}`

superm1:
Here it is with the checkbox.

image

src/app/src/renderer/SystemChecksModal.tsx
jeremyfowers: I just saw the error message for llamacpp+vulkan but not for sdcpp+rocm

src/app/src/renderer/hooks/useSystem.tsx
```diff
@@ -52,26 +62,126 @@ export const SystemProvider: React.FC<{ children: React.ReactNode }> = ({childre
...
+          const recipeOptions = model.recipe_options || {};
+          const recipe = model.recipe || '';
+
+          if (recipe === 'llamacpp' && recipeOptions.llamacpp_backend === 'rocm') {
+            return true;
+          }
+
+          if (recipe === 'sd-cpp' && recipeOptions['sd-cpp_backend'] === 'rocm') {
+            return true;
+          }
```
superm1, line 138: @jeremyfowers are you sure about your test?  It's supposed to explicitly be showing only for these two cases

src/app/src/renderer/hooks/useSystem.tsx
```diff
@@ -52,26 +62,126 @@ export const SystemProvider: React.FC<{ children: React.ReactNode }> = ({childre
...
+    } catch {
+      return false;
+    }
+  }, []);
+
+  const checkForRocmUsage = useCallback(async () => {
+    // Fetch fresh system checks to avoid stale state from closures
+    let checks: SystemCheck[] = [];
+    try {
+      checks = await fetchSystemChecks();
```
jeremyfowers, line 101: This is checking system-checks on every single inference request? How much overhead is there for that?

src/app/src/renderer/hooks/useSystem.tsx
jeremyfowers: I'm going to assume that the problem was in my enablement hack and your code is fine.

src/app/src/renderer/App.tsx
```diff
@@ -270,73 +285,88 @@ const App: React.FC = () => {
   }, []);
 
   return (
-    <SystemProvider>
```
jeremyfowers, line 273: All of these changes just to support showing a popup? Or what?

jeremyfowers (approved):
Two problems jumped out to me during review. I opened comments, please see them and decide whether they are a big deal or not. If they're a big deal, please fix and then merge. If they're not a big deal, please just merge.

src/app/src/renderer/hooks/useSystem.tsx
jeremyfowers: I am worried about overhead since none of this is cached (or else whats the point of doing the check on every inference?) and rooting around in the system can take non-negligable time. Especially on windows.

src/app/src/renderer/App.tsx
jeremyfowers: It was just a red flag to me that total LoC changed seems high for introducing a popup, considering that we already have a notion of popups in the app.

src/app/src/renderer/hooks/useSystem.tsx
superm1: The main reason to do it on every request is that you can unload and load models.  Right now we're worried about anything ROCm, but later I want to do things with NPU too.
Maybe it's better to do at model unload/load time instead of inference request time.

jeremyfowers: `load` feels like a better compromise. it gives us a chance to show the warning before a runtime crash is provoked, and its a far less frequent and far higher-latency event than an inference.

src/app/src/renderer/App.tsx
superm1: In manually auditing I did find a rebase error that brought in some dead code, but everything else is reasonable.

src/app/src/renderer/hooks/useSystem.tsx
superm1: Yup, this seems to work pretty well.  The only problem was that the load event doesn't happen if you load from chat.  So instead I need to trigger off "model != old_model".  But it works and I expect scales to NPU use case just as well.

Issue #1093, opened 2026-02-11
[enhancement] Prompt to overwrite existing lemonade config and service files if they exist


winmutt:
The current debian package will overwrite any existing config or service files. It should prompt to do so.

jeremyfowers:
@superm1 thoughts?

superm1:
Yeah it's because we're storing in `/opt/etc`.  
Once we do https://github.com/lemonade-sdk/lemonade/pull/1079 and https://github.com/lemonade-sdk/lemonade/pull/1022 will move it to `/etc`.

winmutt:
SGTM!

PR #1107, opened 2026-02-12
Enhancement: Prompt to overwrite existing lemonade config and service files if they exist

winmutt:
This change modifies the Debian package post-install script to prompt users before overwriting existing configuration and service files during package upgrades. This addresses issue #1093 by preventing accidental loss of user customizations.

### DIFF
--- a/src/cpp/postinst
+++ b/src/cpp/postinst
@@ -14,6 +14,45 @@ usermod -a -G render lemonade
 
 # Set up the service
 if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ] ; then
+        # Check if service file already exists and prompt user
+        service_file="/lib/systemd/system/lemonade-server.service"
+        config_dir="/etc/lemonade"
+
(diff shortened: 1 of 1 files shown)


superm1:
This seems like more than one thing committed to the branch than intended.

src/cpp/postinst
```diff
@@ -14,6 +14,45 @@ usermod -a -G render lemonade
...
+            echo -n "Do you want to overwrite it? (y/N): "
+            read -r response
+            if [[ ! "$response" =~ ^[Yy]$ ]]; then
+                echo "Skipping service file installation."
+                # Skip the service enablement and start
+                exit 0
+            else
+                echo "Overwriting service file..."
+            fi
+        fi
```
superm1, line 34: unfortunately doing this without debconf will mean that it doesn't work with all frontends.  For example if someone was to install an update using a GUI tool it won't be guaranteed to work properly.

I do think we need to migrate the conffile to /etc, but I want to solve those other two issues first.

winmutt:
Ya, sorry about the extra commits, dont have my at home agentic workflows
setup quite yet. I'll clean up the PR. Let me look at the other related
issues you mentioned.

PR #1120, opened 2026-02-14
Issue #1082 - Support 256K Context Size in tray app context menu

### LINKED ISSUE #1082: tray app does not display 256K context size
Working with models supporting 256k context size, and I notice that there is no option for 256K context in the tray app.

I can set the environment variable LEMONADE_CTX_SIZE to 262144 and it works, but the UI does not reflect the change.

vugenti:
Closes #1082 - Support 256K Context Size in tray app context menu


jeremyfowers (approved):
Thanks for the contribution @vugenti!

PR #1121, opened 2026-02-14
feat: Add image edit and variation endpoints

### LINKED ISSUE #1047: add image edit endpoint

kenvandine:
Add image edit and variation endpoints

  Implements OpenAI-compatible image editing and variation endpoints:
  - POST /v1/images/edits - Edit images based on prompt and optional mask
  - POST /v1/images/variations - Create variations of existing images

  Both endpoints support multipart/form-data for file uploads and follow
  the OpenAI API specification.

  Changes:
  - Added image_edits() and image_variations() to IImageServer interface
  - Implemented Router methods for routing to appropriate backends
  - Implemented SDServer methods to forward requests to sd-cpp backend
  - Added Server HTTP handlers with multipart form data support
  - Added base64_encode helper function for file data transport
  - Registered new routes in Server constructor
  - Created example scripts demonstrating both endpoints
  - Uses response_format parameter name per OpenAI API spec
  - Sets denoising_strength=0.75 for img2img edits requests
  - Validates n parameter (1-10 range) with descriptive error messages
  - Handles JSON parse exceptions with proper 400 responses
  - Uses server_error type for 500 responses per OpenAI spec

  Fixes #1047

There is a build of this in the snap store under the edge/sdd-image-edit branch for testing, which can be installed with:
```
sudo snap install --channel=edge/sdd-image-edit lemonade-server
```

There is also an app that can be used to test it as well at https://github.com/Phqen1x/lemon-zest

### DIFF
--- a/docs/server/server_spec.md
+++ b/docs/server/server_spec.md
@@ -1136,7 +1136,7 @@ Explicitly load a registered model into memory. This is useful to ensure that th
 | `save_options` | No | All | Boolean. If true, saves recipe options to `recipe_options.json`. Any previously stored value for `model_name` is replaced. |
 | `ctx_size` | No | llamacpp, flm, ryzenai-llm | Context size for the model. Overrides the default value. |
 | `llamacpp_backend` | No | llamacpp | LlamaCpp backend to use (`vulkan`, `rocm`, `metal` or `cpu`). |
-| `llamacpp_args` | No | llamacpp | Custom arguments to pass to llama-server. The following are NOT allowed: `-m`, `--port`, `--ctx-size`, `-ngl`. |
+| `llamacpp_args` | No | llamacpp | Custom arguments to pass to llama-server. The following are NOT allowed: `-m`, `--port`, `--ctx-size`, `-ngl`, `--jinja`, `--mmproj`, `--embeddings`, `--reranking`. |
 | `whispercpp_backend` | No | whispercpp | WhisperCpp backend to use (`npu` or `cpu`). Default is `npu` if supported. |
 | `steps` | No | sd-cpp | Number of inference steps for image generation. Default: 20. |
 | `cfg_scale` | No | sd-cpp | Classifier-free guidance scale for image generation. Default: 7.0. |
--- a/src/cpp/tray/tray_app.cpp
+++ b/src/cpp/tray/tray_app.cpp
@@ -415,6 +415,11 @@ static bool is_process_alive_not_zombie(pid_t pid) {
         return false;  // Process doesn't exist
     }
 
+#ifdef __APPLE__
+    // macOS doesn't have /proc — if kill(pid, 0) succeeded, process is alive
+    // macOS automatically reaps zombies more aggressively, so just trust kill()
+    return true;
+#else
     // Check if it's a zombie by reading /proc/PID/stat
     std::string stat_path = "/proc/" + std::to_string(pid) + "/stat";
     std::ifstream stat_file(stat_path);
@@ -435,6 +440,7 @@ static bool is_process_alive_not_zombie(pid_t pid) {
 
     // If we can't parse the state, assume alive to be safe
     return true;
+#endif
 }
 #endif
 
@@ -1692,6 +1698,16 @@ int TrayApp::execute_recipes_command() {
     return 0;
 }
 
+// Check if a process is alive (cross-platform)
+static bool is_process_alive(int pid) {
+#ifdef __APPLE__
+    // macOS doesn't have /proc — use kill(pid, 0) which checks existence without signaling
+    return (kill(pid, 0) == 0 || errno == EPERM);
+#else
+    return std::filesystem::exists("/proc/" + std::to_string(pid));
+#endif
+}
+
 // Command: stop
 int TrayApp::execute_stop_command() {
     auto [pid, port] = get_server_info();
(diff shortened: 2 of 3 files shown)


src/cpp/server/backends/sd_server.cpp
```diff
@@ -236,25 +237,15 @@ json SDServer::image_generations(const json& request) {
     // sd-server requires extra params (steps, sample_method, scheduler) to be
     // embedded in the prompt as <sd_cpp_extra_args>JSON</sd_cpp_extra_args>
     // See PR #1173: https://github.com/leejet/stable-diffusion.cpp/pull/1173
+    // Always inject model defaults from recipe_options so sd-server uses the
+    // correct values even when clients send OpenAI-compliant requests without
+    // these backend-specific parameters.
     json extra_args;
-    if (request.contains("steps")) {
```
bitgamma, line 240: these params are used by the webapp. Let's not change the current behavior but let's use the values from recipe_options if the request doesn't contain the params

src/cpp/server/backends/sd_server.cpp
```diff
@@ -269,5 +260,143 @@ json SDServer::image_generations(const json& request) {
     return forward_request("/v1/images/generations", sd_request, 600);
 }
 
+json SDServer::image_edits(const json& request) {
+    // sd-server's /v1/images/edits puts images into ref_images (EDIT mode),
```
bitgamma: having the images into ref_images is the desired behavior and works well with editing models like Qwen-Edit and Flux Klein 4b/9b. img2img with denoise can be added as an additional endpoint but the behavior of the default edit endpoint needs to actually invoke /v1/images/edits

src/cpp/server/server.cpp
```diff
@@ -42,6 +42,35 @@ namespace fs = std::filesystem;
 
 namespace lemon {
 
+// Helper function to encode binary data to base64
+static std::string base64_encode(const std::string& input) {
```
bitgamma: this function looks like it could be useful in general, let's put it in an utility class

src/cpp/server/server.cpp
```diff
@@ -1985,6 +1992,324 @@ void Server::handle_image_generations(const httplib::Request& req, httplib::Resp
     }
 }
 
+void Server::handle_image_edits(const httplib::Request& req, httplib::Response& res) {
```
ramkrishna2910, line 2086: handle_image_edits and handle_image_variations have some code duplication. Can we extract some helper functions to reduce duplication.

src/cpp/server/utils/json_utils.cpp
```diff
@@ -70,5 +71,59 @@ bool JsonUtils::has_key(const json& j, const std::string& key) {
...
+    return output;
+}
+
+std::string JsonUtils::base64_decode(const std::string& input) {
+    static const std::string base64_chars =
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz"
+        "0123456789+/";
+
+    std::vector<int> lookup(256, -1);
```
ramkrishna2910: base64_decode constructs a 256-element lookup vector on every invocation. For large image payloads called in a loop (n>1), this should be static or a compile-time constant.

src/cpp/server/backends/sd_server.cpp
```diff
@@ -11,6 +11,7 @@
 #include <filesystem>
 #include <thread>
 #include <chrono>
+#include <ctime>
```
ramkrishna2910, line 14: Can we add some tests for edits and variations in tests\server_sd.py?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -11,6 +11,7 @@
...
```
ramkrishna2910: Is this import used?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -236,25 +237,21 @@ json SDServer::image_generations(const json& request) {
...
     if (request.contains("steps")) {
-        extra_args["steps"] = request["steps"];
+        extra_args["steps"] = request["steps"].get<int>();
+    } else {
+        extra_args["steps"] = static_cast<int>(recipe_options_.get_option("steps"));
     }
     if (request.contains("cfg_scale")) {
-        extra_args["cfg_scale"] = request["cfg_scale"];
-    }
-    if (request.contains("seed")) {
```
ramkrishna2910, line 246: Was the seed intentionally removed?

ramkrishna2910:
Is the electron app integration planned to be a follow up PR?

kenvandine:
> Is the electron app integration planned to be a follow up PR?

Yeah, I can do that in a follow-up PR

src/cpp/server/server.cpp
kenvandine: Good catch\! I've extracted three helpers — `parse_n_from_form`, `extract_image_from_form`, and `load_image_model` — which eliminates the duplicated `n` validation, image extraction, and model-loading logic. Both handlers are now considerably shorter.

src/cpp/server/utils/json_utils.cpp
kenvandine: Fixed — the lookup vector is now `static` so it's only constructed once.

src/cpp/server/backends/sd_server.cpp
kenvandine: Added tests 009–016 to `test/server_sd.py`: validation error cases for both endpoints (missing image, missing prompt, invalid `n`, non-multipart) plus basic inference tests for edits and variations.

src/cpp/server/backends/sd_server.cpp
kenvandine: Removed — it wasn't used anywhere.

src/cpp/server/backends/sd_server.cpp
kenvandine: No, it was unintentionally dropped during the refactor in f8a33da. I've added it back to both `image_generations()` and `image_edits()`.

kenvandine:
@ramkrishna2910 @bitgamma I think I've handled all the feedback, thanks!

ramkrishna2910:
@bitgamma please approve when you get a chance :)

PR #1122, opened 2026-02-14
Add support for using 'system' provided llamacpp binary

superm1:
Some distros provide llamacpp binaries in their repositories. Users may want to use these instead of downloading a full stack.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -999,9 +999,27 @@ SystemInfo::SupportedBackendsResult SystemInfo::get_supported_backends(const std
         return result;
     }
 
-    // Collect supported backends and capture first error (in preference order from RECIPE_DEFS)
+    // If a default_backend is specified, add it first (if supported)
+    std::string default_backend;
+    if (recipe_info.contains("default_backend")) {
+        default_backend = recipe_info["default_backend"].get<std::string>();
+        if (recipe_info["backends"].contains(default_backend)) {
+            const auto& backend = recipe_info["backends"][default_backend];
+            std::string state = backend.value("state", "unsupported");
+            if (state != "unsupported") {
+                result.backends.push_back(default_backend);
+            }
+        }
+    }
+
+    // Collect remaining supported backends and capture first error (in preference order from RECIPE_DEFS)
     for (const auto& def : RECIPE_DEFS) {
         if (def.recipe == recipe) {
+            // Skip the default_backend since we already added it
+            if (def.backend == default_backend) {
+                continue;
+            }
+
             if (recipe_info["backends"].contains(def.backend)) {
                 const auto& backend = recipe_info["backends"][def.backend];
                 std::string state = backend.value("state", "unsupported");


superm1:
I validated this branch on Ubuntu 26.04 with the inbox llama.cpp and it's GGML HIP backend.  I also confirmed that removing it the artifact gets downloaded instead.

src/cpp/server/recipe_options.cpp
```diff
@@ -130,7 +126,26 @@ void RecipeOptions::add_cli_options(CLI::App& app, json& storage) {
...
+                }
+
+                if (prefer_system) {
+                    auto it = std::find(result.backends.begin(), result.backends.end(), "system");
+                    if (it != result.backends.end()) {
+                        result.backends.erase(it);
+                        result.backends.insert(result.backends.begin(), "system");
+                    }
+                }
+            }
```
jeremyfowers: This could get real confusing. We heavily advertise that Lemonade gives you the AMD-optimized llamacpp binaries. But this changes the default to whatever happens to be on the system instead of what we are shipping as the default.

What happens if a system ships CPU-only llamacpp? Users will think Lemonade sucks and only works with CPU.

src/cpp/server/system_info.cpp
```diff
@@ -100,6 +100,10 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
     {"llamacpp", "cpu", {"windows", "linux"}, {
         {"cpu", {"x86_64"}},
     }},
+    // llamacpp system (uses system-installed llama-server from PATH)
+    {"llamacpp", "system", {"linux"}, {
+        {"cpu", {"x86_64"}}, // Placeholder, actual check is PATH-based
+    }},
```
jeremyfowers: I'm confused between this code and the block in recipe_options.cpp (https://github.com/lemonade-sdk/lemonade/pull/1122/changes#r2828566918)

This table is ordered. Placing `system` last here means "it is the furthest from the default". But in recipe_options.cpp there is code saying "system should be the default."

src/cpp/server/system_info.cpp
```diff
@@ -911,11 +932,18 @@ std::string SystemInfo::get_oga_version() {
...
+        std::string llama_server_path = utils::find_executable_in_path("llama-server");
+        return !llama_server_path.empty();
+    } else {
+        // For other backends, use the existing logic
+        try {
+            BackendUtils::get_backend_binary_path(LlamaCppServer::SPEC, backend);
+            return true;
+        } catch (const std::exception& e) {
+            return false;
+        }
```
jeremyfowers: Is this the right place for this code? Or should `get_backend_binary_path()` know how to check for system binaries for important engines?

I'm picturing the scenario where system whispercpp, sd-cpp, etc. becomes common too.

src/cpp/server/recipe_options.cpp
superm1: > This could get real confusing. We heavily advertise that Lemonade gives you the AMD-optimized llamacpp binaries. But this changes the default to whatever happens to be on the system instead of what we are shipping as the default.
> What happens if a system ships CPU-only llamacpp? Users will think Lemonade sucks and only works with CPU.

I guess this means we need a way to introspect if there is a HIP backend available with llamacpp.  Because llamacpp loads the backends dynamically I feel this is totally doable to detect.

src/cpp/server/system_info.cpp
superm1: > Is this the right place for this code? Or should get_backend_binary_path() know how to check for system binaries for important engines?

Well I think it's related to your question above about detecting that the engine is available AND it has GPU support.  It seems to me we're going to need two kinds of checks - if binaries are there and if they're good enough.

src/cpp/server/system_info.cpp
superm1: I guess it really should be the top of the list then, right?

src/cpp/server/system_info.cpp
```diff
@@ -2622,6 +2650,53 @@ json SystemInfoCache::get_system_info_with_cache() {
         // Add recipes section (always fresh, never cached)
         system_info["recipes"] = sys_info->build_recipes_info(system_info["devices"]);
 
+        // Add supported_recipes section with preferred backends
```
jeremyfowers: Doesn't system-info already have most of this? Seems like a lot of code for just attaching the backend default to the existing recipe info.

Also is there a way to avoid hardcoding "prefer this or that for llamacpp" in so many code blocks? We already have a centralized place for assessing the default.

test/test_llamacpp_system_backend.py
```diff
@@ -0,0 +1,194 @@
+import os
```
jeremyfowers, line 1: Is it intentional that you're committing this, but it is not part of any CI workflow?

src/cpp/server/system_info.cpp
jeremyfowers: Yes - if it should be the default (I'm not convinced).

test/test_llamacpp_system_backend.py
superm1: It's how I tested it locally, but Yeah let me get it glued to CI.  My thought here was to make sure that our logic stays solid.

src/cpp/server/system_info.cpp
superm1: I'll see what I can do.

jeremyfowers:
Marking as draft while we nail down the details.

src/cpp/server/system_info.cpp
superm1: moved to top of list

src/cpp/server/recipe_options.cpp
superm1: Added HIP backend detection (in common paths).  It will reject system backend when missing now.
```
Recipe              Backend     Status          Message/Version                               Action
----------------------------------------------------------------------------------------------------------------------------------------------------
llamacpp            system      unsupported     HIP plugin libggml-hip.so not installed        -
```

superm1:
OK so on most systems this is what it will look like.
```
$ ./build/lemonade-server  recipes 
Recipe              Backend     Status          Message/Version                               Action
----------------------------------------------------------------------------------------------------------------------------------------------------
llamacpp            system      unsupported     llama-server not found in PATH                 -
                    metal       unsupported     Requires macOS                                 -
                    vulkan      installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:vulkan
                    rocm        unsupported     Requires Radeon RX 7000 series (RDNA3) AMD dGPU -
                    cpu         installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:cpu
```

Then if you install llama.cpp:
```
$ ./build/lemonade-server  recipes 
Recipe              Backend     Status          Message/Version                               Action
----------------------------------------------------------------------------------------------------------------------------------------------------
llamacpp            system      unsupported     HIP plugin libggml-hip.so not installed        -
                    metal       unsupported     Requires macOS                                 -
                    vulkan      installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:vulkan
                    rocm        unsupported     Requires Radeon RX 7000 series (RDNA3) AMD dGPU -
                    cpu         installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:cpu
```

Then if you install `libggml0-backend-hip`:
```
$ ./build/lemonade-server  recipes 
Recipe              Backend     Status          Message/Version                               Action
----------------------------------------------------------------------------------------------------------------------------------------------------
llamacpp            system      installed       -                                              -
                    metal       unsupported     Requires macOS                                 -
                    vulkan      installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:vulkan
                    rocm        unsupported     Requires Radeon RX 7000 series (RDNA3) AMD dGPU -
                    cpu         installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:cpu
```

`system` is the top of the priority list.  This should mean it only uses system for `auto` if you have a fully set up system with all those packages installed.  Otherwise you can pick a backend (IE override to vulkan, or rocm or cpu if you want to).

jeremyfowers:
> OK so on most systems this is what it will look like.
> 
> ```
> $ ./build/lemonade-server  recipes 
> Recipe              Backend     Status          Message/Version                               Action
> ----------------------------------------------------------------------------------------------------------------------------------------------------
> llamacpp            system      unsupported     llama-server not found in PATH                 -
>                     metal       unsupported     Requires macOS                                 -
>                     vulkan      installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:vulkan
>                     rocm        unsupported     Requires Radeon RX 7000 series (RDNA3) AMD dGPU -
>                     cpu         installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:cpu
> ```
> 

This makes me wonder if filtering `unsupported` items out of here is wise, and providing a `--show-all` or something to show them all. Wondering how many people will randomly add `llama-server` to their `PATH` just because we mention it here.

This can be a follow up based on user feedback though.

jeremyfowers (approved):
LGTM but consider taking the code simplification in https://github.com/lemonade-sdk/lemonade/pull/1285

superm1:
Either that or other work on the branch from a rebase seems to have caused a problem in my testing.  I'll figure it out.

PR #1128, opened 2026-02-14
fix: show user-specified host in API endpoint log message

### LINKED ISSUE #1092: [bug] lemonade reports listening on 127.0.0.1 when its 0.0.0.0
Minor nit but this took me a few minutes to sort out after a fresh reinstall.
```
\_ /opt/bin/lemonade-router --port 11435 --host 0.0.0.0 --log-level debug --ctx-size 200000 --llamacpp rocm --llamacpp-args -fa on --no-mmap --max-loaded-models 3 2 1 1 1
```
```
lemonade-server[655520]: API endpoint: http://127.0.0.1:11435/api/v1
```

winmutt:
## Summary
Fixes #1092: When users specify --host 0.0.0.0, the tray was converting it to 127.0.0.1 in the API endpoint log message, which was misleading.

## Changes
- Added get_display_host() method to return the original host value for logging
- Updated tray logging to use get_display_host() instead of get_connection_host()
- Kept get_connection_host() for HTTP connections (still uses 127.0.0.1 for valid connections)

When users run 'lemonade-router --host 0.0.0.0', they will now see:
  API endpoint: http://0.0.0.0:11435/api/v1

The tray still correctly converts to 127.0.0.1 for actual HTTP connections, but the log message now accurately reflects what the user specified.

### DIFF
--- a/.github/actions/build-macos-dmg/action.yml
+++ b/.github/actions/build-macos-dmg/action.yml
@@ -1,0 +1,171 @@
+name: 'Build macOS .dmg Package'
+description: 'Build Lemonade .dmg for macOS (signed and notarized via CMake notarize target)'
+inputs:
+  include-electron:
+    description: 'Whether to include the Electron app (full build)'
+    required: false
+    default: 'true'
+outputs:
+  artifact-name:
+    description: 'The name of the generated .dmg file'
+    value: ${{ steps.build-dmg.outputs.artifact_name }}
+runs:
+  using: "composite"
+  steps:
+    - name: Run setup.sh to configure environment
+      shell: bash
+      run: |
+        set -e
+
+        echo "Running setup.sh to configure build environment..."
+        bash setup.sh
+
+        echo "Build environment configured successfully!"
+
+    - name: Setup Node.js
+      if: inputs.include-electron == 'true'
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20'
+
+    - name: Set Electron signing identity
+      shell: bash
+      run: |
+        # Tell electron-builder to use the Developer ID Application cert from keychain
+        if [ -n "$DEVELOPER_ID_APPLICATION_IDENTITY" ]; then
+            echo "CSC_NAME=$DEVELOPER_ID_APPLICATION_IDENTITY" >> $GITHUB_ENV
+            # electron-builder skips signing on PR branches by default
+            echo "CSC_FOR_PULL_REQUEST=true" >> $GITHUB_ENV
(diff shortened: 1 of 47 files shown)


sofiageo:
The only point of logging this URL is for people and applications to be able to use it, so with that in mind:

There are a couple of issues with this approach. First, we need to be sure http://0.0.0.0:8000/ works in every OS. In my limited experience it does for Windows and Linux, but another maintainer expressed some concerns in the discord thread.

Second, this needs to take care of ipv6 binding. Setting :: in the lemonade.conf will produce a wrong URL for the logging. The correct one should be http://[::]:8000/ or http://[::1]:8000/. I'm not sure which one is better because a browser works with both, another application might not.

winmutt:
Thanks for the feedback @sofiageo. I updated the PR so that IF the listening IP is different than the connection IP, it will log it out.

sofiageo:
I didn't have much time today, I'll test it tomorrow, and give feedback.

src/cpp/include/lemon_tray/server_manager.h
```diff
@@ -73,6 +73,12 @@ class ServerManager {
         return (host_ == "0.0.0.0" || host_ == "localhost") ? "127.0.0.1" : host_;
     }
 
+    // Get the host to display in logs (original user-specified value).
+    // Unlike get_connection_host(), this doesn't convert 0.0.0.0 or localhost.
+    std::string get_display_host() const {
```
sofiageo, line 78: This isn't used at the moment

src/cpp/tray/server_manager.cpp
```diff
@@ -150,9 +150,16 @@ bool ServerManager::start_server(
...
         if (!is_ephemeral) {
             std::cout << "Lemonade Server v" << LEMON_VERSION_STRING << " started on port " << port_ << std::endl;
-            // Display localhost for 0.0.0.0 since that's what users can actually visit in a browser
-            std::string display_host = get_connection_host();
-            std::cout << "API endpoint: http://" << display_host << ":" << port_ << "/api/v1" << std::endl;
+            std::cout << "API listening on " << host_ << ":" << port_ << std::endl;
+
+            std::string accessible_host = host_;
+            if (host_ == "0.0.0.0") {
+                accessible_host = "localhost";
```
sofiageo, line 157: I don't like this. I can find some examples to compare but I think most servers just show http://0.0.0.0:8000 and let the browser figure it out. I tested the lemonade-desktop electron app and it connects to 0.0.0.0 just fine.

winmutt:
I found that this is included in DEBUG logs but that the log level is case sensitive. Closing this in favor of making log level case insensitive as upper case log level is more of a standard.

PR #1131, opened 2026-02-16
Define TTS model type separately from audio

bitgamma:
needed because otherwise by default (max_loaded_models=1) loading whisper would evict kokoro and viceversa, while using both at the same time is probably a common use case


antmikinka:
YES! Ill work on testing this. 
Curious when I'm done testing this, should I just merge it if it all works out?

bitgamma:
> YES! Ill work on testing this. Curious when I'm done testing this, should I just merge it if it all works out?

You can leave a review when you are done testing

Issue #1134, opened 2026-02-17
TTS buttons show up on all types of messages


ramkrishna2910:
<img width="549" height="437" alt="Image" src="https://github.com/user-attachments/assets/1936508b-7c83-495d-8967-15d6e9a8ec3c" />

The speaker icon should not show up on images.

ramkrishna2910:
Resolved in #1135

PR #1135, opened 2026-02-17
Tabbed settings

choppu:
* added TTS settings (model, user voice, assistant voice, enable TTS, enable user TTS)
* separated  LLMChat, Connection and TTS Setings in separate components
* implemented tab layout of SettingsModal
* closes issue #1134

### DIFF
--- a/src/app/src/renderer/tabs/TTSSettings.tsx
+++ b/src/app/src/renderer/tabs/TTSSettings.tsx
@@ -49,7 +49,7 @@ const TTSSettings: React.FC<TTSSettingsProps> = ({settings, onValueChangeFunc, o
             Reset
           </button>
         </div>
-        <select className="form-input form-select" defaultValue={settings.tts['userVoice'].value} onChange={(e) => onValueChangeFunc('userVoice', e.target.value)}>
+        <select className="form-input form-select" value={settings.tts['userVoice'].value} onChange={(e) => onValueChangeFunc('userVoice', e.target.value)}>
           {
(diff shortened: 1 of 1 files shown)


choppu:
This PR adds TTS settings,  introduces tab layout for Settings Modal and closes issue #1134. The audio button doesn't show up if the message has only images as its content. In addition, user can disable TTS completely or only for user messages in TTS Settings.

jeremyfowers (changes requested):
Hi @choppu, apologies I am coming in late to reviewing your TTS PRs. 

I would prefer that the UI/UX worked differently. I am open to having my mind changed, and I am also open to doing the coding for this change myself since I am coming in with the feedback late here.

Here's how I want it to work:

1. `kokoro-v1` is visible in the Model Manager and users can pull and load it.
2. When `kokoro-v1` is loaded and selected, a Text-to-Speech view opens in the far-right pane. This view is dedicated to TTS, just like LLM Chat, Image Gen, and STT have their own dedicated views:
    - User enters a message and push send
    - The transcript of that message goes into the chat
    - The message is automatically read aloud in the selected voice
    - A speaker button is available to repeat the speech
    - User can send more messages to get more speeches; these would populate the message history in the view similar to how message/image pairs populate the view in text-to-image mode
3. Speech and the speaker button are not provided in LLM chat.

What do you think? 

My main consideration is that I want to get `main` branch release ready ASAP tomorrow (February 18th), which is why I don't mind taking on the coding changes if you are busy tomorrow.

PS. the tabbed settings are great, thanks for doing that!

PR #1142, opened 2026-02-17
Fix Electron app port discovery always being skipped

### LINKED ISSUE #1068: Tray app isn’t telling electron app the port number
Reproduction:
1. Start tray app
2. Change the port number from the tray's dropdown
3. use tray app to open electron app
4. server not found in  the electron app!

jeremyfowers:
## Summary
- `getBaseURLFromConfig()` is `async` but was called without `await` on line 555 of `main.js`
- The unawaited Promise was always truthy, so port discovery was unconditionally skipped and the handler returned `null`
- The renderer interpreted `null` as "explicit URL configured" and stayed on the default port 8000, never discovering the actual server port
- **One-character fix**: add the missing `await`

Closes #1068

### DIFF
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -552,7 +552,7 @@ ipcMain.handle('get-version', async () => {
 });
 
 ipcMain.handle('discover-server-port', async () => {
-  let baseURL = getBaseURLFromConfig();
+  let baseURL = await getBaseURLFromConfig();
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@Tetramatrix confirmed this fixes the problem

ramkrishna2910 (approved):

PR #1147, opened 2026-02-18
Kokoro UI integration

jeremyfowers:
## Summary
- Adds `AudioButton.tsx` component for text-to-speech playback
- Integrates TTS audio playback into `ChatWindow.tsx`
- Adds related CSS styles and audio library script
- added TTS settings (model, user voice, assistant voice, enable TTS, enable user TTS)
- separated LLMChat, Connection and TTS Setings in separate components
- implemented tab layout of SettingsModal

### DIFF
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -132,8 +132,8 @@ const DEFAULT_TTS_SETTINGS = Object.freeze({
   model: { value: 'kokoro-v1', useDefault: true },
   userVoice: { value: 'fable', useDefault: true },
   assistantVoice: { value: 'alloy', useDefault: true },
-  enableTTS: { value: true, useDefault: true },
-  enableUserTTS: { value: true, useDefault: true }
+  enableTTS: { value: false, useDefault: true },
+  enableUserTTS: { value: false, useDefault: true }
 });
 
 const createDefaultAppSettings = () => ({
--- a/src/app/src/renderer/AudioButton.tsx
+++ b/src/app/src/renderer/AudioButton.tsx
@@ -6,9 +6,9 @@ export const PLAYING = 2;
 
 interface  AudioButtonProps {
   textMessage: any;
-  role: string;
+  role?: string;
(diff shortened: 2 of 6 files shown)


jeremyfowers (commented):
Hi @choppu, apologies I am coming in late to reviewing your TTS PRs. 

I would prefer that the UI/UX worked differently. I am open to having my mind changed, and I am also open to doing the coding for this change myself since I am coming in with the feedback late here.

Here's how I want it to work:

1. `kokoro-v1` is visible in the Model Manager and users can pull and load it.
2. When `kokoro-v1` is loaded and selected, a Text-to-Speech view opens in the far-right pane. This view is dedicated to TTS, just like LLM Chat, Image Gen, and STT have their own dedicated views:
    - User enters a message and push send
    - The transcript of that message goes into the chat
    - The message is automatically read aloud in the selected voice
    - A speaker button is available to repeat the speech
    - User can send more messages to get more speeches; these would populate the message history in the view similar to how message/image pairs populate the view in text-to-image mode
3. Speech and the speaker button are not provided in LLM chat.

What do you think? 

PS. the tabbed settings are great, thanks for doing that!

choppu:
Hi @jeremyfowers, I indeed misunderstood the way tts should be integrated. What I thought was that normally the most common case when a user wants the messages to be spoken out is in an llm chat. So, I implemented it in LLM chat. If you check my last pr with "tabbed settings", you can see that I added a separate tab for TTS settings, where user can completely disable TTS if they find it in some way invasive. Also, I have addressed the issue with audio button on messages with only images (audio button doesn't show up on images anymore). 

However, if the lemonade app is meant more like a showcase of different features in isolation, your approach probably makes more sense. It's OK for me to implement it as you described above, but I'm not sure if I can do it everything this week as it takes some time to implement and test if everything works correctly. I also think I don't have write access to this repo so I'd need to make a new PR from a fork.

PR #1150, opened 2026-02-18
Fix: webapp not reachable at its normal address

### LINKED ISSUE #1148: webapp not reachable at its normal address
Image

going to http://localhost:8000/ just says "Ollama is running"

Webapp is only reachable from http://localhost:8000/webapp.html

jeremyfowers:
Closes #1148


bitgamma (approved):
tested, works fine

jeremyfowers:
I need to fix a minor test issue

PR #1167, opened 2026-02-19
Add connection status to the status bar

superm1:
As there is now a lot of ways to run lemonade without necessarily having lemonade server installed locally, add information into the electron app (and implicitly web-app) about status of the connection.

It will poll frequently until connected.

### DIFF
--- a/src/app/src/renderer/StatusBar.tsx
+++ b/src/app/src/renderer/StatusBar.tsx
@@ -28,6 +28,9 @@ const StatusBar: React.FC = () => {
     gpu_percent: null,
     vram_gb: null,
   });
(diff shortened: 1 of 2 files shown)


superm1:
Screenshot_20260219_100442
Screenshot_20260219_100513

jeremyfowers:
I like this feature! It goes with our “lemonade ui should be more like vs code ui” theme of the month.

in the screenshot it looks like the connected/connecting dynamic text is not vertically aligned to the static “status” text. Can you fix that?

superm1:
Fixed it

image

jeremyfowers:
> Fixed it
> image

looking good

I also just noticed the green dot is out of alignment too

PR #1168, opened 2026-02-19
Enable claude code integration via "ollama launch claude"

### LINKED ISSUE #1162: ollama launch claude raises a bunch of 400 and 404 errors
ollama-cli:

```
ollama launch claude --model GLM-4.7-Flash-GGUF
```

Lemonade:
```
llama-server is ready!
[LlamaCpp] Model loaded on port 8001
[Router] Backend started successfully
[Router] Model loaded successfully. Total loaded: 1
[Server] POST /api/v1/load - 200
[Server PRE-ROUTE] GET /api/v1/models
[Server] GET /api/v1/models - 200
[Server PRE-ROUTE] HEAD /
[Server] HEAD / - 200
[Server PRE-ROUTE] GET /api/status
[Server] Error 404: GET /api/status
[Server] GET /api/status - 404
[Server PRE-ROUTE] POST /api/show
[Server] Error 400: POST /api/show
[Server] 400 Bad Request details - Body length: 97, Content-Type: application/json
[Server] POST /api/show - 400
```

sawansri:
Resolves #1162


docs/server/server_spec.md
```diff
@@ -87,6 +87,16 @@ lemonade-server serve --port 11434
 | `POST /api/copy` | Not supported | Returns 501 |
 | `POST /api/push` | Not supported | Returns 501 |
 
+### Anthropic-Compatible API (Initial)
```
jeremyfowers, line 89: Looking at your implementation, this Anthropic API is only supported as a subset of Ollama API?

If yes, this should be a subsection of the Ollama API section, less people think its more general.

jeremyfowers: On a related note, was there a technical reason to only support it via Ollama API? Or can the Anthropic support be generalized?

src/cpp/server/ollama_api.cpp
```diff
@@ -121,6 +122,144 @@ static void map_ollama_options(const json& ollama_request, json& openai_req) {
     }
 }
 
+static void add_warning(std::vector<std::string>& warnings, const std::string& warning) {
```
jeremyfowers: Since this change is +1000 LoC I would prefer it go into a new file like `ollama_to_anthropic.cpp`.

docs/server/server_spec.md
jeremyfowers, line 89: I'm having trouble getting CC to do anything. Can you suggest a model and some prompts that produce results?

sawansri: > On a related note, was there a technical reason to only support it via Ollama API? Or can the Anthropic support be generalized?

So technically, the Anthropic API can exist standalone from Ollama. The reason I chose to implement Anthropic translation as part of Ollama was to reuse the plumbing to convert into OpenAI API. 

If you think it's a better design choice, maybe I can implement  `api_translation.cpp` which allows reuse of some of the translation code between both the Ollama and Anthropic implementations while keeping them separate.

src/cpp/server/ollama_api.cpp
sawansri: good idea! In this case, it would probably be something like `anthropic_api.cpp` since we can decouple anthropic and ollama.

docs/server/server_spec.md
sawansri: I was also encountering some performance issues on my machine, but I attributed it to the model processing Claude Code's 14k token system prompt + my hardware. Doing a bit more research, it seems like there is definitely a performance gap between hooking up local models to claude code vs cloud.

One thing I was thinking about while looking at Claude Code's system prompt was that there is a lot of anthropic specific content that we could tweak for better performance. An example being "The most recent Claude model family is Claude 4.5/4.6. Model IDs — Opus 4.6: 'claude-opus-4-6', Sonnet 4.6: 'claude-sonnet-4-6', Haiku 4.5: 'claude-haiku-4-5-20251001'. When building AI applications, default to the latest and most capable Claude models." My only concern with this approach would be the added complexity.

jeremyfowers: Code re-use is a good enough reason for me!

jeremyfowers: I think the "performance issue" may actually be a bug that causes claude to livelock.

```
slot update_slots: id  0 | task 218 | prompt done, n_tokens = 18630, batch.n_tokens = 198
slot init_sampler: id  0 | task 218 | init sampler, took 2.97 ms, tokens: text = 18630, total = 18630
slot update_slots: id  3 | task 222 | new prompt, n_ctx_slot = 128000, n_keep = 16, task.n_tokens = 18367
slot update_slots: id  3 | task 222 | n_tokens = 3, memory_seq_rm [3, end)
slot update_slots: id  3 | task 222 | prompt processing progress, n_tokens = 1853, batch.n_tokens = 2048, progress = 0.100887
slot update_slots: id  3 | task 222 | n_tokens = 1853, memory_seq_rm [1853, end)
slot update_slots: id  3 | task 222 | prompt processing progress, n_tokens = 3900, batch.n_tokens = 2048, progress = 0.212337
[HttpClient ERROR] CURL error: Failed writing received data to disk/application
[WrappedServer ERROR] Streaming request failed: CURL error: Failed writing received data to disk/application
```

image

jeremyfowers: Can you investigate that `CURL error`, using the simple prompts I put into CC there?

sawansri: Yeah, let me take a look

PR #1169, opened 2026-02-20
Use upstream debian packaging instead of cpack

superm1:


docs/man/man1/lemonade-router.1
kenvandine, line 1: Wouldn't it be better to move these to src/cpp and install the man pages by default rather than only in the deb?

docs/man/man1/lemonade-server.1
kenvandine, line 1: Wouldn't it be better to move these to src/cpp and install the man pages by default rather than only in the deb?

kenvandine (changes requested):
Since they all come from the same source, I would rather see the man pages be installed regardless of the package format.

docs/man/man1/lemonade-router.1
superm1: sure seems reasonable

superm1:
This is going to require a lot of steps and I will do them over multiple PRs.  The first is here:
https://github.com/lemonade-sdk/lemonade/pull/1366

superm1:
Here is the split out for the container that will do "PPA" test builds.

https://github.com/lemonade-sdk/lemonade/pull/1367

superm1:
Here is the split out for the system nodejs target.

https://github.com/lemonade-sdk/lemonade/pull/1369

PR #1171, opened 2026-02-20
App redesign and centralized Backend Manager

jeremyfowers:
image

image

image

image

## Summary

- *Full app redesign*
- Introduces a `BackendManager` C++ class that centralizes backend install/uninstall logic previously scattered across individual `WrappedServer` subclasses
- Adds `POST /api/v1/install` endpoint with SSE streaming progress and `POST /api/v1/uninstall` endpoint

Details:
- Adds CLI commands: `lemonade-server recipes --install/--uninstall recipe:backend`
- Adds a React Backend Manager UI accessible from the center panel menu and View menu
- Backend installs triggered before model load now show progress in the Download Manager
- Enriches `/system-info` with `release_url`, `download_filename`, and version fallbacks from `BackendManager` (single source of truth)
- React app fetches download sizes from the GitHub Releases API using the exact artifact filename
- Adds integration tests for install/uninstall endpoints (8 new tests) and CLI commands (3 new tests)

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -1180,7 +1180,8 @@ std::string SystemInfo::get_flm_version() {
     // Parse version from output like "FLM v0.9.4"
     if (output.find("FLM v") != std::string::npos) {
         size_t pos = output.find("FLM v");
-        std::string version = output.substr(pos + 5);
+        // Keep the 'v' prefix so it matches backend_versions.json (e.g. "v0.9.34").
+        std::string version = output.substr(pos + 4);
         // Trim whitespace and newlines
         size_t end = version.find_first_of(" \t\n\r");
         if (end != std::string::npos) {


bitgamma:
the screenshots look glorious

jeremyfowers:
Dealing with bugs introduced by logical merge conflicts now. I want to get this merged asap as there's been a lot of these conflicts.

jeremyfowers:
Should be better now.

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -0,0 +1,134 @@
...
+): Promise<void> {
+  if (!recipes || !recipes[recipe]) return;
+
+  const recipeInfo = recipes[recipe];
+  const backends = Object.entries(recipeInfo.backends);
+
+  // Find first supported backend that needs installation
+  const needsInstall = backends.find(
+    ([, info]) => info.supported && !info.available
+  );
```
superm1: do we perhaps need a separate `installable` flag?  Because for example some backends might not be directly installable from the lemonade-server (which is intentionally limited permissions for security)

CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ if(MSVC)
     set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 endif()
 
-project(lemon_cpp VERSION 9.3.4)
+project(lemon_cpp VERSION 9.3.5)
```
superm1: shouldn't this happen when you're closer to tagging?

src/app/src/renderer/utils/backendInstaller.ts
jeremyfowers: Are you thinking of FLM? On Windows we have a completely special codepath for FLM install that downloads their .msi installer and starts it on behalf of the user, which triggers UAC and their whole install wizard GUI.

I imagine for Linux FLM our hands are pretty tied, and we just need to display some instructions for `sudo install_flm_stuff`?

src/cpp/server/backend_manager.cpp
```diff
@@ -0,0 +1,268 @@
...
+        recipe_json["available"] = recipe.available;
+        if (!recipe.error.empty()) {
+            recipe_json["error"] = recipe.error;
+        }
+
+        json backends_json = json::array();
+        for (const auto& backend : recipe.backends) {
+            json b;
+            b["name"] = backend.name;
+            b["supported"] = backend.supported;
```
superm1: how would you feel about renaming `supported` to `hardware_present`?  I got really tripped up on this when working on the FLM/Linux PR.

Then `available` can be `software_present`.

And if you take my suggested for `installable` then you have a comprehensive story for telling people what everything means without a decoder ring.

CMakeLists.txt
jeremyfowers: What difference does it make?

fwiw the reason I'm doing this here is because I set a new minimum bin version in backend_verisons.json (your prediction was correct! it was needed again soon!)

src/cpp/server/backend_manager.cpp
```diff
@@ -0,0 +1,268 @@
...
+            }
+            return "";
+        }
+
+        auto params = get_install_params(recipe, backend);
+        return "https://github.com/" + params.repo + "/releases/tag/" + params.version;
+    } catch (...) {
+        return "";
+    }
+}
```
superm1, line 188: does this really need a `try`/`catch`?  Where are you going to get exceptions from?

src/cpp/server/backend_manager.cpp
jeremyfowers: I'm open to renaming, for sure!

But `hardware_present != supported`. A backend can still be unsupported due to having the wrong OS, for example.

src/cpp/server/system_info.cpp
```diff
@@ -256,30 +255,34 @@ static bool device_matches_constraint(const std::string& device_family,
     return allowed_families.count(device_family) > 0;
 }
 
+// Spec lookup for recipes that follow the standard BackendSpec pattern
+static const BackendSpec* get_spec_for_recipe(const std::string& recipe) {
+    if (recipe == "llamacpp") return &LlamaCppServer::SPEC;
```
bitgamma: BackendManager has a copy of this function. I'd rather call that one here and simply invert the order in which recipes are checked in a if/else if/else chain when this function is only invoked if no special handling was triggered

src/cpp/server/backend_manager.cpp
jeremyfowers: `flm --version` could fail for a variety of reasons? the other backends all have a `version.txt` file that we can read, and that is pretty reliable. OTOH for `flm` we can fail to find it on PATH since we don't control it like we do the others. 

This is part of why I was saying a portable FLM binary would be really nice. We could then be a lot more confident of where it is and whether it's actually installed.

src/cpp/server/system_info.cpp
```diff
@@ -291,30 +294,32 @@ static bool is_recipe_installed(const std::string& recipe, const std::string& ba
         return false;
     }
     if (recipe == "ryzenai-llm") {
-        return SystemInfo::is_ryzenai_serve_available();
+        const char* ryzenai_bin_env = std::getenv("LEMONADE_RYZENAI_SERVER_BIN");
+        if (ryzenai_bin_env && fs::exists(ryzenai_bin_env)) {
```
bitgamma: is the special handling for ryzenai still really needed? in the BackendManager only flm seems to have special handling, but probably I am missing something

src/cpp/server/system_info.cpp
jeremyfowers: Thanks I will work on deduplication.

PS. the system-info vs. backend-manager split got kind of weird during development because the tray/cli builds system-info to directly access it for things like populating CLI help. I definitely want to figure out a cleaner separation later. Ideally the tray/cli are just http clients and dont build chunks of the router for their own purposes.

src/cpp/server/system_info.cpp
jeremyfowers: I tore out a lot of ryzenai special handling but I might have missed some! Will take a look.

CMakeLists.txt
bitgamma: my comment is rather if we shouldn't bump it to 9.4.0 or even 10.0.0 considering the extent of the changes

src/app/src/renderer/utils/backendInstaller.ts
superm1: Well even for Linux ROCm - it will allow us to offer either ROCm 7.2 based llama.cpp or TheROck based llama.cpp.
For Linux FLM I have a modal in my PR that tells people how to do it.

CMakeLists.txt
superm1: Ah I see.  OK no concerns.

src/cpp/server/backend_manager.cpp
superm1: how about another field for `os_supported`?  Then we have very clear delineation without overloading the field meaning.

src/cpp/server/backend_manager.cpp
superm1: Ahhh I see.  OK. leave as is then.

jeremyfowers:
Seems there's a lot of corner cases and code paths in the app for how backends are downloaded. Working through that now - will push a unified code path soon.

src/cpp/server/system_info.cpp
superm1: > I definitely want to figure out a cleaner separation later

As it's a C++ project I also want to move away from having so much preprocessor directives with `#ifdef __linux` etc.  We should just use polymorphism to implement a Linux or Windows or Mac implementation where relevant and then only include the source files relevant for the OS builds relevant.

It's wild west right now.

src/app/src/renderer/utils/backendInstaller.ts
jeremyfowers: Can we handle this in a separate follow up PR? I'd like to take inventory of every scenario we want to support.

CMakeLists.txt
jeremyfowers: @bitgamma I'm open to v9.4! :) we're getting image editing in the next release as well so it'll be a fun one.

Lets save v10 for the full NPU enablement on Linux - that is going to be an epic wave.

bitgamma: sounds good!

src/cpp/server/backend_manager.cpp
jeremyfowers: We already have a system for this: `supported` is a synonym of "installable". If `supported/installable == False`, we give an `error` string that says why, which can be the HW, the OS, or any other condition. 

`supported == True` has a specific meaning: if you try to `/install BACKEND`, we expect it to work. At which point the status becomes `available` (which I would be fine renaming to `installed`.)

current: `unsupported (error)` | `supported` | `available`
change to: `unsupported (error)` | `installable` | `installed` ?

src/app/src/renderer/utils/backendInstaller.ts
superm1: OK.  Let's do it after this + flm groundwork.

src/cpp/server/backend_manager.cpp
superm1: I guess the problem I see is that the error string could have a lot of reasons.  Look at the FLM/Linux case.

If we have the hardware in place, we have the kernel driver, but the backend is missing what would you say?
I feel like it should be "unsupported" but we still want to show that it can be installed by following some directions.

That's why I think lots of fields to explain the status and then you can populate the error string with more info.

jeremyfowers: This is why I was unsure about system-info vs. system-checks being separate endpoints. The information is often needed together.

There's a fourth state, which only system-checks currently knows about: `action required`.

1. unsupported (reason, e.g., Windows 10 vs. 11, NPU not present, etc.)
2. action required (action, e.g., update kernel)
3. installable (automatic vs. manual, e.g., `sudo apt-install flm`)
4. installed

superm1: Yes, bingo.  I'm fine with system checks being folded in and ditching it's API (we're not released yet with it after-all!).

jeremyfowers: ok lets re-do system-info/check and the way that backend status is reported in a follow-up PR?

afaik this PR is still a step in the right direction.

superm1: Up to you if you want to roll it into this PR or do it as a follow up, I don't feel strongly.

jeremyfowers: I want to merge this one asap since its merge conflict city in here haha

also I realized there's actually 5 states

1. unsupported (reason, e.g., Windows 10 vs. 11, NPU not present, etc.)
2. action required (action, e.g., update kernel)
3. installable (automatic vs. manual, e.g., `sudo apt-install flm`)
4. installed, but update required
5. installed (ready)

Right now we dont show that "update required" state explicitly and then there's a surprise download when the user tries to load a model.

`Update required` is important for this work because otherwise the GUI will still have surprise silent downloads. I will implement that now, and rename the two pre-existing states in the same shot.

"action required" will be a follow up.

superm1: OK.

jeremyfowers: image

src/cpp/server/system_info.cpp
Geramy: something like this wouldn't be bad, CRTP with C++20 and "contract enforcement" for polymorphism?
https://medium.com/@rogerbooth/using-the-crtp-and-c-20-concepts-to-enforce-contracts-for-static-polymorphism-a27d93111a75

src/cpp/server/backend_manager.cpp
```diff
@@ -0,0 +1,274 @@
...
+        throw std::runtime_error("backend_versions.json is missing version for: " + recipe + ":" + backend);
+    }
+    return recipe_config[backend].get<std::string>();
+}
+
+// ============================================================================
+// Core operations
+// ============================================================================
+
+static const backends::BackendSpec& get_spec_for_recipe(const std::string& recipe) {
```
superm1: is this not missing `flm` recipes?

jeremyfowers: There is no SPEC defined for FLM at the moment, its a total special case.

jeremyfowers: Actually this is highly related to https://github.com/lemonade-sdk/lemonade/pull/1171#discussion_r2853881143

and will get cleaned up a bit

jeremyfowers: https://github.com/lemonade-sdk/lemonade/pull/1171/commits/93650600bfccebbf6df1b0a83fae98ccee5e64ad

src/cpp/server/system_info.cpp
jeremyfowers: https://github.com/lemonade-sdk/lemonade/pull/1171/commits/93650600bfccebbf6df1b0a83fae98ccee5e64ad

src/cpp/server/system_info.cpp
jeremyfowers: https://github.com/lemonade-sdk/lemonade/pull/1171/commits/93650600bfccebbf6df1b0a83fae98ccee5e64ad

src/cpp/server/backend_manager.cpp
jeremyfowers: done!

PR #1176, opened 2026-02-23
Adding support in backend and frontend UI for super-resolution (Real-ESRGAN models)

bconsolvo:
Adding support for stable-diffusion.cpp super-resolution models (Real-ESRGAN):
- CPU, ROCm.

Removing attempted support for Vulkan. Keeping the stable-diffusion.cpp server the same for now.

The way that stable-diffusion.cpp upscaling works is that it is a part of the stable diffusion pipeline. It is not its "own" feature. It is just a part of the image generation pipeline. In order to make a "generated" image, and then an "upscaled image" separable, I had to add aa `/api/v1/images/upscale` API call to the code.

Adding original and upscaled elements to the UI. See screenshot here:
image

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -2482,9 +2482,10 @@ void Server::handle_image_upscale(const httplib::Request& req, httplib::Response
             // Use the server's --sdcpp CLI setting so we pick up the same
             // backend binary that was installed at startup, not whatever
             // the system auto-detects (which may be a stale installation).
-            if (default_options_.contains("sd-cpp_backend") &&
-                default_options_["sd-cpp_backend"].is_string()) {
-                std::string b = default_options_["sd-cpp_backend"];
+            auto recipe_opts = config_->recipe_options();
+            if (recipe_opts.contains("sd-cpp_backend") &&
+                recipe_opts["sd-cpp_backend"].is_string()) {
+                std::string b = recipe_opts["sd-cpp_backend"];
                 if (!b.empty()) backend = b;
             }
         } catch (const std::exception& e) {


src/app/package-lock.json
```diff
@@ -4379,6 +4379,17 @@
         "node": ">= 0.8"
       }
     },
+    "node_modules/encoding": {
```
ramkrishna2910, line 4380: Is this encoding package necessary for this PR?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -338,5 +338,57 @@ json SDServer::image_variations(const json& request) {
     return forward_multipart_request("/v1/images/edits", fields, 600);
 }
 
+std::string SDServer::upscale_via_cli(
```
ramkrishna2910, line 343: The endpoint request/response schema isn't documented anywhere. Can we add the necessary documentation for this addition?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -338,5 +338,57 @@ json SDServer::image_variations(const json& request) {
...
+    const std::string& b64_image,
```
ramkrishna2910, line 344: Please add necessary tests in test/server_sd.py

src/app/src/renderer/components/panels/ImageGenerationPanel.tsx
```diff
@@ -219,6 +342,16 @@ const ImageGenerationPanel: React.FC<ImageGenerationPanelProps> = ({
...
             disabled={isBusy} placeholder="-1 = random" />
         </div>
+        <div className="image-setting">
+          <label>Upscale</label>
+          <select value={imageSettings.upscaleModel}
+            onChange={(e) => handleUpscaleChange(e.target.value)}
+            disabled={isBusy}>
+            <option value="">Off</option>
+            <option value="RealESRGAN-x4plus">RealESRGAN x4</option>
+            <option value="RealESRGAN-x4plus-anime">RealESRGAN x4 Anime</option>
```
ramkrishna2910: Model names hard coded in the front end.
These should be dynamically queried from the models endpoint (filtered by esrgan label) so adding new ESRGAN models doesn't require frontend changes.

src/cpp/server/backends/sd_server.cpp
```diff
@@ -226,7 +228,6 @@ json SDServer::image_generations(const json& request) {
     // sd-server requires extra params (steps, sample_method, scheduler) to be
     // embedded in the prompt as <sd_cpp_extra_args>JSON</sd_cpp_extra_args>
     // See PR #1173: https://github.com/leejet/stable-diffusion.cpp/pull/1173
-    // Use request values if present (e.g. from webapp), fall back to recipe_options defaults.
```
ramkrishna2910, line 229: Why is this comment removed?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -252,7 +253,6 @@ json SDServer::image_generations(const json& request) {
     LOG(DEBUG, "SDServer") << "Forwarding request to sd-server: "
                   << sd_request.dump(2) << std::endl;
 
-    // Use base class forward_request with 10 minute timeout for image generation
```
ramkrishna2910: Why is this comment removed?

src/cpp/server/backends/sd_server.cpp
```diff
@@ -338,5 +338,57 @@ json SDServer::image_variations(const json& request) {
...
+            std::istreambuf_iterator<char>());
+        result = JsonUtils::base64_encode(upscaled_data);
+        LOG(INFO, "SDServer") << "ESRGAN upscale complete ("
+            << raw.size() << " -> " << upscaled_data.size() << " bytes)" << std::endl;
+    } else {
+        LOG(WARNING, "SDServer") << "ESRGAN upscale failed (exit code: "
+            << exit_code << ")" << std::endl;
+    }
+
+    fs::remove(input_path);
```
ramkrishna2910: No temp file cleanup on exception. Make sure the temp files are removed even when there is a failure.

src/cpp/server/backends/sd_server.cpp
```diff
@@ -338,5 +338,57 @@ json SDServer::image_variations(const json& request) {
...
+    if (exit_code == 0 && fs::exists(output_path)) {
+        std::ifstream in(output_path, std::ios::binary);
+        std::string upscaled_data(
+            (std::istreambuf_iterator<char>(in)),
+            std::istreambuf_iterator<char>());
+        result = JsonUtils::base64_encode(upscaled_data);
+        LOG(INFO, "SDServer") << "ESRGAN upscale complete ("
+            << raw.size() << " -> " << upscaled_data.size() << " bytes)" << std::endl;
+    } else {
+        LOG(WARNING, "SDServer") << "ESRGAN upscale failed (exit code: "
```
ramkrishna2910, line 401: When upscale_via_cli fails, it returns an empty string with only a LOG warning. The process's stdout/stderr is lost because inherit_output = false and there's no output capture. This makes debugging very difficult. Consider capturing stderr from the subprocess.

src/cpp/server/backends/sd_server.cpp
```diff
@@ -338,5 +338,57 @@ json SDServer::image_variations(const json& request) {
...
+    }
+
+    std::vector<std::string> cli_args = {
+        "-M", "upscale",
+        "--upscale-model", upscale_model_path,
+        "-i", input_path.string(),
+        "-o", output_path.string()
+    };
+
+    auto proc = ProcessManager::start_process(
```
ramkrishna2910, line 386: The code constructs the sd-cli path but never checks if the file exists before spawning the process. If the sd-cpp backend isn't installed, the error will be a cryptic "ESRGAN upscale failed" instead of a helpful "sd-cpp backend not installed."

src/cpp/server/server.cpp
```diff
@@ -2320,6 +2323,112 @@ void Server::handle_image_variations(const httplib::Request& req, httplib::Respo
...
+            res.status = 404;
+            nlohmann::json error = {{"error", {
+                {"message", "Upscale model not found: " + upscale_model_name},
+                {"type", "invalid_request_error"}
+            }}};
+            res.set_content(error.dump(), "application/json");
+            return;
+        }
+
+        // Use sd-cli subprocess for upscaling (matches official sd-cli -M upscale)
```
ramkrishna2910: Please document why we are using sd-cli and not the already loaded sd-server process.

src/cpp/server/server.cpp
```diff
@@ -2320,6 +2323,112 @@ void Server::handle_image_variations(const httplib::Request& req, httplib::Respo
...
+        std::string backend = "cpu";
```
ramkrishna2910: The backend detection logic takes the sd-cpp backend from whatever model happens to be loaded. 
  - If no model is loaded, silently defaults to cpu even on a ROCm system
  - If the loaded model is an LLM (not SD), it won't have sd-cpp_backend also defaults to cpu
  - The ESRGAN model's own recipe options or the user's --sdcpp config should determine the backend, not the currently loaded model

src/app/src/renderer/components/panels/ImageGenerationPanel.tsx
```diff
@@ -22,8 +24,19 @@ const DEFAULT_IMAGE_SETTINGS: ImageSettings = {
   width: 512,
   height: 512,
   seed: -1,
+  upscaleModel: '',
 };
 
+interface ImageHistoryItem {
```
ramkrishna2910, line 30: Please add screen shots of the current user experience to the PR description.

src/app/package-lock.json
bconsolvo: encoding is an optional dependency of minipass-fetch (npm's internal dep tree) -- it was auto-resolved by npm install, not intentionally added. But I can just revert this back to the original package-lock.json for now.

src/cpp/server/backends/sd_server.cpp
bconsolvo: Adding endpoint documentation (request/response/errors) as a comment above 
```cpp
static std::string upscale_via_cli(
```
in sd_server.h.

src/cpp/server/backends/sd_server.cpp
bconsolvo: Adding tests 017-020, and making generic ESRGAN_MODEL so we can swap models.

src/app/src/renderer/components/panels/ImageGenerationPanel.tsx
bconsolvo: Replacing with dynamic list filtered from modelsData by the "esrgan" label. New ESRGAN models only need a server_models.json entry.

src/cpp/server/backends/sd_server.cpp
bconsolvo: Restoring comments -- they were accidentally dropped during a rebase.

src/cpp/server/backends/sd_server.cpp
bconsolvo: Restoring comments -- they were accidentally dropped during a rebase.

src/cpp/server/backends/sd_server.cpp
bconsolvo: Changing inherit_output to true so subprocess stdout/stderr is visible in server logs. ProcessManager::read_output is currently a stub, so direct capture isn't possible without extending ProcessManager. Using inherit_output = true ensures the output is at least visible for debugging. Also enhanced the failure log message to include the model path and CLI path. Let me know if this works.

src/cpp/server/backends/sd_server.cpp
bconsolvo: Adding fs::exists check in both upscale_via_cli (early return with LOG error) and handle_image_upscale (HTTP 500 with descriptive message).

src/cpp/server/server.cpp
bconsolvo: Replacing the short comment with: "sd-server's HTTP API does not expose an upscaling endpoint -- upscaling is only available via sd-cli -M upscale. We shell out to sd-cli as a subprocess, which also keeps generation and upscaling as two separate requests. This lets the frontend show the original image immediately while upscaling runs, and then display both images side by side with independent timing."

src/cpp/server/server.cpp
bconsolvo: Fixing. Backend is now resolved from the ESRGAN model's own recipe_options (via model_manager_->get_model_info(upscale_model_name).recipe_options), not from whatever model happens to be loaded. This correctly picks up the user's --sdcpp config for the ESRGAN model.

src/app/src/renderer/components/panels/ImageGenerationPanel.tsx
bconsolvo: Done :)

src/cpp/server/backends/sd_server.cpp
bconsolvo: Adding RAII TempFileGuard that cleans up temp files in its destructor, so they are removed even on exceptions.

docs/server/server_spec.md
```diff
@@ -927,6 +928,139 @@ Image Variations API. You provide a source image and receive a variation of it.
...
+      -H "Content-Type: application/json" \
+      -d '{
+            "model": "SD-Turbo",
+            "prompt": "A serene mountain landscape at sunset",
+            "size": "512x512",
+            "steps": 4,
+            "response_format": "b64_json"
+          }')
+
+    # Step 2: Extract the base64 image from the response
```
ramkrishna2910: The curl command on line 968-973 interpolates $IMAGE_B64 (a ~200K-860K char string) directly into the command line. This exceeds ARG_MAX on many systems (we hit it on Windows/MSYS2). Recommend using curl -d @<file> or piping via stdin instead.

bconsolvo: Fixed with `curl -d @`. Thank you for the review.

PR #1177, opened 2026-02-24
Add more ground work for flm/linux

superm1:
I want to make sure this continues to work on Windows.  I'm a bit concerned that the new `flm validate` flow won't work in Windows, so let's see how it goes.

**Note**: This does not turn on the flm Linux recipe yet.  But it is is **VERY** close to doing so.

Screenshot_20260223_205457

### DIFF
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -21,7 +21,7 @@
   },
   "flm": {
     "npu": "v0.9.34",
-    "min_npu_driver": "42.0.203.304",
+    "min_npu_driver": "32.0.203.304",
     "min_kernel_version": "7.0"
   },
   "kokoro": {


superm1:
Strix halo system check is a casualty (I killed the modal in favor of opening a tab when applicable - like trying to install)- it will come back after landing this PR.

jeremyfowers:
> Strix halo system check is a casualty (I killed the modal in favor of opening a tab when applicable - like trying to install)- it will come back after landing this PR.

This has me thinking... why are we opening a tab when we're already in a HTML app? Should we just open an iframe?

superm1:
> > Strix halo system check is a casualty (I killed the modal in favor of opening a tab when applicable - like trying to install)- it will come back after landing this PR.
> 
> This has me thinking... why are we opening a tab when we're already in a HTML app? Should we just open an iframe?

Let me noodle on that.  If the branch otherwise looks good, can we leave it as a follow up?

jeremyfowers:
> > > Strix halo system check is a casualty (I killed the modal in favor of opening a tab when applicable - like trying to install)- it will come back after landing this PR.
> > 
> > 
> > This has me thinking... why are we opening a tab when we're already in a HTML app? Should we just open an iframe?
> 
> Let me noodle on that. If the branch otherwise looks good, can we leave it as a follow up?

Yeah definitely, its an optimization/polish that can wait.

superm1:
> > > > Strix halo system check is a casualty (I killed the modal in favor of opening a tab when applicable - like trying to install)- it will come back after landing this PR.
> > > 
> > > 
> > > This has me thinking... why are we opening a tab when we're already in a HTML app? Should we just open an iframe?
> > 
> > 
> > Let me noodle on that. If the branch otherwise looks good, can we leave it as a follow up?
> 
> Yeah definitely, its an optimization/polish that can wait.

I played with it, I got an iframe working!  It should work for both gfx1151 and NPU messaging.

superm1:
Are some of the GFX1151 runners not updated to an applicable newest kernel?

superm1:
> Are some of the GFX1151 runners not updated to an applicable newest kernel?

If so; i'll push the gfx1151 messaging part into another PR.

docs/npu_linux.html
```diff
@@ -0,0 +1,347 @@
...
+<body>
+  <div class="navbar-placeholder"></div>
+
+  <main class="main">
+    <div class="hero-section">
+      <div class="main-heading">
+        NPU Software Stack Setup Required
+      </div>
+      <div class="subtitle">
+        Your NPU software stack is not properly configured
```
jeremyfowers: This guide is so long and complex that its hard for me to imagine anyone will have their linux NPU software stack properly configured in the first place. (ie, 100% of users will end up on this page)

Rather than declaring "Your NPU software stack is not properly configured" can this document be framed as a step by step guide for setup? 

And we should probably link to it right from the installation instructions site (when the time comes).

docs/npu_linux.html
```diff
@@ -0,0 +1,347 @@
...
+        <!-- Check NPU firmware and kernel -->
+        <div class="option-box option-primary">
+          <div class="option-header">
+            <span class="option-label">Step 1: Verify System Requirements</span>
+          </div>
+
+          <h3>Check NPU Firmware Version</h3>
+          <p>Verify your NPU firmware is version 1.1.0.0 or later:</p>
+          <div class="code-block">
+            <code>cat /sys/bus/pci/drivers/amdxdna/*/fw_version</code>
```
jeremyfowers: If some of these steps dont require `sudo` can they be automated? 

With so many steps in this guide there's a significant surface area for the user to screw things up and then need support from us.

docs/npu_linux.html
superm1: The problem is the guide will be very different for different distros.  I'm moving chess pieces in different distros to make everything easier.  I don't want to start offering it (IE turn it on until it's easy in Ubuntu and Arch.).

So treat this as generic placeholder guide for now that we'll improve.

docs/npu_linux.html
superm1: Nothing can be automated - it's all stuff the lemonade service user won't be able to do.  Ubuntu 26.04 will have nearly everything ready to go (you just need to download and install FLM deb).  Everything else will have some other work to do.

src/app/src/renderer/ModelManager.tsx
```diff
@@ -367,21 +371,180 @@ const ModelManager: React.FC<ModelManagerProps> = ({ isVisible, width = 280, cur
...
+      const abortController = new AbortController();
+      downloadId = downloadTracker.startDownload(modelName, abortController);
 
-      await fetchCurrentLoadedModel();
-      showSuccess(`Model "${modelName}" downloaded successfully.`);
-    } catch (error) {
-      if (error instanceof DownloadAbortError) {
-        if (error.reason === 'paused') {
-          showWarning(`Download paused: ${modelName}`);
+      // Dispatch event to open download manager
```
jeremyfowers: Why is the download manager in scope for this PR?

src/cpp/server/system_info.cpp
```diff
@@ -256,33 +260,21 @@ static bool device_matches_constraint(const std::string& device_family,
...
+        return SystemInfo::is_llamacpp_installed(backend);
+    }
+    if (recipe == "whispercpp") {
+        return SystemInfo::is_whispercpp_installed(backend);
+    }
+    if (recipe == "kokoro") {
+        return SystemInfo::is_kokoro_installed(backend);
+    }
+    if (recipe == "sd-cpp") {
+        return SystemInfo::is_sdcpp_installed(backend);
```
jeremyfowers: I just got rid of all these redundant backend-specific functions and replaced them with a generic one. Why re-introduce them?

jeremyfowers (changes requested):
There may have been some significant rebasing issues against the big backend manager PR. I'm seeing some big changes that seem like they should be out of scope and/or reverting prior changes.

Please see comments for examples.

docs/npu_linux.html
jeremyfowers: `cat /sys/bus/pci/drivers/amdxdna/*/fw_version` can't be run in a subprocess by lemonade?

docs/npu_linux.html
jeremyfowers: ok

jeremyfowers:
> Are some of the GFX1151 runners not updated to an applicable newest kernel?

I did some patches to enable sdcpp ROCm before

Are there even newer updates needed?

src/app/src/renderer/ModelManager.tsx
superm1: It gets dismissed when the iframe opens

superm1:
> > Are some of the GFX1151 runners not updated to an applicable newest kernel?
> 
> I did some patches to enable sdcpp ROCm before
> 
> Are there even newer updates needed?

Nope; that should have been it.  So there is a problem then still with this part of the code.  I'll split it to another PR.

src/cpp/server/system_info.cpp
superm1: Will see what I can do.

superm1: Should be cleaned up

docs/npu_linux.html
superm1: `flm validate` does this already and returns it in the json output..

superm1: Maybe the page needs to be rewritten as "install flm, run flm validate, if you run into problems then here are the things that you should fix".

jeremyfowers: > Maybe the page needs to be rewritten as "install flm, run flm validate, if you run into problems then here are the things that you should fix".

This would definitely make it less intimidating!

src/cpp/server/system_info.cpp
```diff
@@ -914,6 +922,54 @@ static std::string read_version_file(const fs::path& version_file) {
...
+bool SystemInfo::is_sdcpp_installed(const std::string& backend) {
+    auto* spec = try_get_spec_for_recipe("sd-cpp");
+    if (!spec) return false;
+    try {
+        BackendUtils::get_backend_binary_path(*spec, backend);
+        return true;
+    } catch (...) {
+        return false;
+    }
+}
```
jeremyfowers: All of these should not be here?

superm1: axed

docs/npu_linux.html
superm1: I've reworked the generic content like that.

src/cpp/server/utils/path_utils.cpp
```diff
@@ -216,5 +217,93 @@ std::string get_downloaded_bin_dir() {
...
+    }
+
+    char buffer[1024];
+    std::string output;
+    while (fgets(buffer, sizeof(buffer), pipe) != nullptr) {
+        output += buffer;
+    }
+
+#ifdef _WIN32
+    int exit_code = _pclose(pipe);
```
jeremyfowers, line 250: Unreachable code since Windows has an unconditional `return true` above this?

superm1: 100% true. I wanted to wait until next flm release for windows to use it or will break CI.

src/app/src/renderer/ChatWindow.tsx
```diff
@@ -676,11 +675,26 @@ const sendMessage = async () => {
         });
       } else {
         console.error('Failed to send message:', error);
+        const errorMessage = error.message || 'Failed to get response from the model.';
```
jeremyfowers: Can you move this out to a helper function? Its duplicated below.

If you look at the refactor in https://github.com/lemonade-sdk/lemonade/pull/1195 I changed ChatWindow.tsx so that it is only shared functions, and the definition of each individual inference modality is its own file.

superm1: That explains the merge conflict.  Yes, will fix it after you finish review.

src/app/src/renderer/BackendManager.tsx
```diff
@@ -152,19 +153,41 @@ const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError,
...
+      showError(`Failed: ${errorMessage}`);
+
+      // If the error has a help URL in the action, open it
+      const backendInfo = systemInfo?.recipes?.[recipe]?.backends?.[backend];
+      const action = backendInfo?.action;
+
+      // Extract URL from action
+      if (action) {
+        const urlMatch = action.match(/https?:\/\/[^\s]+/);
+        if (urlMatch && urlMatch[0].endsWith('.html')) {
```
jeremyfowers: Fine for now, I suspect, but FYI the .html filtering will make it so that links like this one wont work https://lemonade-server.ai/docs/

src/app/src/renderer/BackendManager.tsx
```diff
@@ -152,19 +153,41 @@ const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError,
...
       });
     }
-  }, [showError, showSuccess]);
+  }, [showError, showSuccess, systemInfo]);
 
   const handleCopyAction = useCallback(async (recipe: string, backend: string, action?: string) => {
     if (!action) return;
     try {
+      // If the action contains a documentation URL, open it in-app
+      if (action.match(/https?:\/\/[^\s]+\.html/)) {
```
jeremyfowers: This got flagged as a security risk. Someone using a forked or lemonade-compatible backend with our app could supply any URL they like as an `action`. Suggest filtering to lemonade-server.ai links only, like you did with the markdown.

docs/server/server_spec.md
```diff
@@ -58,7 +58,6 @@ The additional endpoints are:
 - GET `/api/v1/health` - Check server status, such as models loaded
```
jeremyfowers, line 58: There is a major regression on Windows.

On `main`, if FLM is not installed and I push download on an FLM model, the FLM installer auto-launches:

Image

On this branch, I just get an error:

Image

src/app/src/renderer/BackendManager.tsx
superm1: I think we can probably just drop the second half of the logic - `urlMatch[0].endsWith('.html')`

src/app/src/renderer/App.tsx
```diff
@@ -298,22 +296,29 @@ const AppContent: React.FC = () => {
...
-            />
+            {externalContentUrl ? (
+              <div className="chat-window" style={isLogsVisible ? { width: `${chatWidth}px` } : undefined}>
+                <div className="external-content-container">
+                  <div className="external-content-header">
+                    <button className="external-content-back-btn" onClick={() => setExternalContentUrl(null)}>
+                      <ChevronLeft size={14} />
+                      Back to Chat
+                    </button>
+                    <span className="external-content-title">Guide</span>
```
jeremyfowers: Image

IMO it would look better without the "GUIDE"

If you want to keep GUIDE I suggest maybe centering it and having the font color/size match the BACKEND MANAGER title

jeremyfowers: Overall the iframe is really slick though, WAY better than popping open a modal and/or browser.

docs/server/server_spec.md
```diff
@@ -58,7 +58,6 @@ The additional endpoints are:
...
 - GET `/api/v1/stats` - Performance statistics from the last request
 - GET `/api/v1/system-info` - System information and device enumeration
-- GET `/api/v1/system-checks` - System health checks for platform-specific issues
 - GET `/live` - Check server liveness for load balancers and orchestrators
```
jeremyfowers, line 62: After I had the iframe open, I was curious what would happen if I clicked links.

The first link on the stx halo page opened a browser window. Good.

The next link I tried, 

Image

Gave me this black iframe.

Image

Can you fix it so that links always open a browser window/tab?

docs/server/server_spec.md
jeremyfowers: Fixed!

jeremyfowers:
Just pushed https://github.com/lemonade-sdk/lemonade/pull/1177/commits/dfb7fb047e91124b2cfe36166bc0eb94b37399ac:

```
 cd /c/work/lsdk/lemonade && git commit -m "$(cat <<'EOF'                                                                                                                                   Fix FLM installation flow, NPU driver checks, and iframe integration                                                                                                                                                                                                                                                                                                                  - Fix unreachable install() calls: check() throws FLMCheckException for all                                                                                                                  failures but callers used if (!check()) install(). Replace with try/catch                                                                                                                  that catches NOT_INSTALLED to call install() and re-throws DRIVER_TOO_OLD                                                                                                             
     and VALIDATION_FAILED. Applied in backend_manager, model_manager, and   
     fastflowlm_server load().                                         
   - Add NPU driver version check to system-info endpoint so the frontend can
     detect driver issues before attempting install/download.               
   - Distinguish driver errors from missing FLM in system-info: driver errors
     get a .html guide URL action, missing FLM gets an install command.
   - Add .html extension to DRIVER_INSTALL_URL for iframe detection.
   - Add backend pre-flight in ModelManager.tsx: fetch fresh system-info and
     call ensureBackendForRecipe before pullModel to install FLM or show   
     driver guide iframe.                              
   - Handle .html action URLs in ensureBackendForRecipe: dispatch           
     open-external-content event to open iframe instead of calling /install.
   - Add fallback driver error detection in handleDownloadModel catch block
     to open driver guide iframe even if pre-flight is bypassed.
   - Remove dead code: unreachable if (!check_npu_driver_version()) guard in
     download_model() since the method now throws.
```

docs/server/server_spec.md
jeremyfowers: https://github.com/lemonade-sdk/lemonade/pull/1177/commits/774d96b26e5f5d3949e1b426fb5510a7d3e0b08b

src/app/src/renderer/App.tsx
jeremyfowers: https://github.com/lemonade-sdk/lemonade/pull/1177/commits/774d96b26e5f5d3949e1b426fb5510a7d3e0b08b

src/app/src/renderer/BackendManager.tsx
jeremyfowers: https://github.com/lemonade-sdk/lemonade/pull/1177/commits/774d96b26e5f5d3949e1b426fb5510a7d3e0b08b

PR #1179, opened 2026-02-24
Stop logging to tmpfs when running under systemd

superm1:
This means the lemonade user needs to have access to the journal.

### DIFF
--- a/src/cpp/include/lemon/single_instance.h
+++ b/src/cpp/include/lemon/single_instance.h
@@ -39,7 +39,13 @@ public:
 
         return false;
 #else
-        // Unix/Linux: Use file locking
+        // Skip lock file when running under systemd, unless explicitly disabled
+        const char* journal_stream = std::getenv("JOURNAL_STREAM");
+        const char* disable_journal = std::getenv("LEMONADE_DISABLE_SYSTEMD_JOURNAL");
+        if (journal_stream && !disable_journal) {
+            return false;  // Systemd ensures single instance
+        }
+
         std::string lock_file = "/tmp/lemonade_" + app_name + ".lock";
         int fd = open(lock_file.c_str(), O_CREAT | O_RDWR | O_CLOEXEC, 0666);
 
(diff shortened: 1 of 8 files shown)


src/cpp/tray/tray_app.cpp
```diff
@@ -2063,10 +2063,15 @@ bool TrayApp::start_server() {
             log_file_ = "lemonade-server.log";
         }
         #else
-        // Unix: /tmp/lemonade-server.log or ~/.lemonade/server.log
-        log_file_ = "/tmp/lemonade-server.log";
+        const char* journal_stream = std::getenv("JOURNAL_STREAM");
```
Geramy, line 2068: Same as above in "single_instance.h" I would think we would want to use a preprocessor directive here too.

Geramy (commented):
I made a few suggestions let me know what you think.

src/cpp/include/lemon/single_instance.h
```diff
@@ -39,7 +39,10 @@ class SingleInstance {
 
         return false;
 #else
-        // Unix/Linux: Use file locking
+        if (std::getenv("JOURNAL_STREAM")) {
```
Geramy: Please surround by #ifdef linux or "HAVE_SYSTEMD" so that this applies to linux only, homebrew supports all linux based systems and programs even systemd "apparently" also mac has introduced a systemd variant so just to future proof and reduce risk of breaking I think we should do as suggested. Let me know your thoughts.

superm1: Huh; it's possible to have systemd on a mac?  TIL.

The way this PR works - it should only run the new codepaths if it's running "under" systemd.  So just compiling in the code for `HAVE_SYSTEMD` on a mac wouldn't change anything.  Now if we start running lemonade under systemd on a mac later, I think we need to look at everything with a fine toothed comb ANYWAY because everything I've done thus far for systemd is based upon how systemd on Linux works.

Geramy: Yeah I didnt know mac could have systemd until a few days ago when I came across some packages I never wanted to learn about haha.  Okay so thats fine then, everything else looks good to me!

superm1:
> Comments resolved.

Thanks!  Now to figure out WTH I broke on CI.  The failures make zero sense to me for the code changes.

Issue #1182, opened 2026-02-24
macOS testing race condition


jeremyfowers:
Generated by claude:

<html>
<body>
<html><head></head><body><p>Here's what I found — <strong>it's not a timeout at all.</strong> The process is being actively killed by SIGTERM after ~1.3 seconds, well before any timeout could fire.</p>
<h2>Precise Timeline (from run 22360009014)</h2>

Timestamp | Event
-- | --
16:41:49.276 | test_macos_arm64 starts
16:41:50.545 | Terminated: 15 — killed after 1.27 seconds
16:41:52.781 | Runner cleans up orphan lemonade-router (pid 2145)

<h2>What IS happening</h2>
<p>The test starts <code>lemonade-server serve --no-tray --log-level debug</code>, which fork()+execv()s a <code>lemonade-router</code> child. All three processes (Python test, <code>lemonade-server</code>, <code>lemonade-router</code>) share the same <strong>process group</strong> — the codebase has no <code>setsid()</code>/<code>setpgid()</code> calls anywhere.</p>
<p>~1 second later, something external sends <strong>SIGTERM to the Python test process</strong> (exit code 143 = 128 + 15). The orphaned <code>lemonade-router</code> (PID 2145) confirms the server DID start but wasn't cleaned up.</p>
<h2>Most likely root cause: LaunchDaemon process group interaction</h2>
<p>The CI flow has a <strong>race condition</strong> involving the macOS LaunchDaemon:</p>
<ol>
<li><code>.pkg</code> install runs <code>postinst-full-mac</code> → <code>launchctl bootstrap + kickstart</code> → starts <code>lemonade-server serve --no-tray</code> as a system daemon → spawns <code>lemonade-router</code></li>
<li>First <code>launchctl bootout</code> (install action, line 62) sends SIGTERM to the service — by default, macOS launchd <strong>kills the entire process group</strong> (no <code>AbandonProcessGroup</code> key in the plist)</li>
<li><code>sleep 2</code>, clean up lock/PID files</li>
<li>Second <code>launchctl bootout</code> (workflow line 837, ~36 seconds later) — should be no-op</li>
<li><code>sleep 2</code>, clean up locks again</li>
<li>Test starts, <code>setUpClass.stop_server()</code> kills any remaining router, sleeps 2s</li>
<li><code>test_macos_arm64</code> starts fresh <code>lemonade-server serve</code> → spawns new <code>lemonade-router</code></li>
[...]

PR #1183, opened 2026-02-24
Fix macOS race condition in execute_stop_command process checks

Geramy:
macOS doesn't have /proc, so std::filesystem::exists("/proc/PID") always returns false — making the stop command think processes exited immediately while they may still be alive. Replace all /proc/PID checks with a cross-platform is_process_alive() helper that uses kill(pid, 0) on macOS. Also guard the zombie detection in is_process_alive_not_zombie to avoid the same false-negative on macOS.

Regarding issue https://github.com/lemonade-sdk/lemonade/issues/1182

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -496,19 +496,6 @@ jobs:
           .venv/bin/python test/server_endpoints.py --server-binary /usr/local/bin/lemonade-server --server-per-test
           echo "Endpoint tests PASSED!"
 
-      - name: Run system-info tests (unsigned path)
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@copilot was there actually a race condition? if so, does this fix it?

jeremyfowers:
Alternative attempt: https://github.com/lemonade-sdk/lemonade/pull/1188

Geramy:
FYI I do believe there was an actual bug as far as /proc/ not existing on macos but my original "port" or "adaption" might have routed around most of these calls, thus not showing up as an actual bug, if someone where to later make changes to the macos specific code they could end up creating a regression which would take these "functional" routes. just some thoughts.

Issue #1181, opened 2026-02-24
MacOS Tests running out of memory


Geramy:
During the tests the MacOS Runners kill the test processes because it runs out of memory. This is not all the time it is periodic, I believe due to the runners being shared instances. The idea would be to replace it with a dedicated runner, of our own or reduce the model sizes being loaded for the tests. The second solution to support forks of lemonade also would be to get a or some macmini's to be runners and we load the secrets into the machine thus not requiring access to github secrets for external contributors to run tests. @jeremyfowers

Geramy:
@jeremyfowers @ramkrishna2910 FYI

jeremyfowers:
Pretty sure we root caused this to a process-killing race condition. But also mitigated by disabling an unintentional stress test.

Geramy:
Sounds good.

PR #1184, opened 2026-02-24
Add unsigned local build+test path for macOS CI when signing secrets …

Geramy:
…are unavailable

When MACOS_APP_CONNECT_KEY_GERAMY is not available (e.g., PRs from forks), skip signing/packaging/notarization and instead build, install binaries locally, and run CLI/endpoint/system-info/ollama tests inline in the build job. Existing signed flow is unchanged when secrets exist.

https://github.com/lemonade-sdk/lemonade/issues/1181
https://github.com/lemonade-sdk/lemonade/issues/1182


superm1:
dup of #1186

PR #1185, opened 2026-02-24
Don't hardcode lemonade-router to debug

superm1:
This has ripple effects and ignores user intent.  Pass the log level accordingly.


kenvandine:
Oh, thank you! LGTM and a much appreciated improvement.

superm1:
> Need to verify that debug logs still show up be default in the electron app on windows

shouldn't change electron app at all.  It's for the router process.

jeremyfowers:
> > Need to verify that debug logs still show up be default in the electron app on windows
> 
> shouldn't change electron app at all. It's for the router process.

Right, but the electron app displays the router's log. I need to ensure the UX hasn't been altered.

superm1:
OK - so no actions on me, this is a @jeremyfowers needs to test and make sure @superm1 didn't break something non-obvious.

PR #1186, opened 2026-02-24
Add unsigned local build+test path for macOS CI when signing secrets …

Geramy:
…are unavailable

When MACOS_APP_CONNECT_KEY_GERAMY is not available (e.g., PRs from forks), skip signing/packaging/notarization and instead build, install binaries locally, and run CLI/endpoint/system-info/ollama tests inline in the build job. Existing signed flow is unchanged when secrets exist.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -19,6 +19,9 @@ permissions:
   contents: write
   actions: read  # Required for SignPath to read workflow/job details
 
+env:
+  LEMONADE_DISABLE_SYSTEMD_JOURNAL: "1"
+
 jobs:
   # ========================================================================
   # BUILD JOBS - Run on rai-160-sdk workers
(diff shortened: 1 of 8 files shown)


.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -366,7 +367,22 @@ jobs:
...
+          APP_CONNECT_KEY: ${{ secrets.MACOS_APP_CONNECT_KEY_GERAMY }}
+        run: |
+          if [ -n "$APP_CONNECT_KEY" ]; then
+            echo "has_signing=true" >> $GITHUB_OUTPUT
+            echo "Signing secrets available - will build signed .pkg"
+          else
+            echo "has_signing=false" >> $GITHUB_OUTPUT
+            echo "No signing secrets - will build and test locally"
+          fi
+
```
superm1, line 383: I tend to think you should only pass the secrets during tagging.  You already have some parts of this conditional on `startsWith(github.ref, 'refs/tags`.  How about just only expose the secrets when that's true?

Geramy: Yeah I can change that to use signing only on release tag. I do like the rule or methodology you said about do you trust a dev build to be signed for release.

jeremyfowers: +1

Issue #876, opened 2026-01-20
Minor app bug: LLM Chat text entry box loses focus after each message is sent


jeremyfowers:
Repro:
1. Type a message in LLM Chat and push enter
2. Wait for LLM's response
3. Try typing another message - nothing happens
4. You must click the text entry box to begin typing again

PR #1195, opened 2026-02-26
Refactor, fix, and polish the inference UI in the app

### LINKED ISSUE #1194: App bug: upload image button doesnt work twice
Repro:
1. Use the Upload Image button to add an image to LLM Chat
2. Hit send
3. Hit the clear chat button
4. Try to repeat step 1, it does nothing

Workaround: restart the app

jeremyfowers:
Closes #1194
Closes https://github.com/lemonade-sdk/lemonade/issues/876

1. Decompose ChatWindow.tsx (2650 → 197 lines) into 6 self-contained modality panels (`LLMChatPanel`, `EmbeddingPanel`, `RerankingPanel`, `TranscriptionPanel`, `ImageGenerationPanel`, `TTSPanel`) plus shared components (`ModelSelector`, `InferenceControls`, `Icons`, etc.)
2. Replace 7 independent boolean state flags with a single `useInferenceState` hook (`idle → pre_flight → inferring`), fixing a double-send bug where rapid clicks could fire multiple requests during model loading
3. Lock the active panel during inference so loading a different-modality model via Model Manager doesn't yank the UI away mid-stream
4. Fix auto-scroll for thinking LLMs — restore phase-change dependency so streaming thinking content scrolls into view
5. Keep text input focused and editable during inference so users can type ahead while waiting for a response
6. Makes the LLM chat view look nicer with a gradient background
7. Improve auto-scroll performance (it was pretty jittery before)
8. Replace `alert()` calls with toast notifications
9. Fixes an image upload bug
10. Add a responsive breakpoint for the stats bar

### DIFF
--- a/src/app/src/renderer/components/panels/EmbeddingPanel.tsx
+++ b/src/app/src/renderer/components/panels/EmbeddingPanel.tsx
@@ -148,7 +148,6 @@ const EmbeddingPanel: React.FC<EmbeddingPanelProps> = ({
             }}
             placeholder="Enter text to generate embeddings..."
             rows={1}
-            disabled={isBusy}
           />
           <InferenceControls
             isBusy={isBusy}
--- a/src/app/src/renderer/components/panels/ImageGenerationPanel.tsx
+++ b/src/app/src/renderer/components/panels/ImageGenerationPanel.tsx
@@ -239,7 +239,6 @@ const ImageGenerationPanel: React.FC<ImageGenerationPanelProps> = ({
             }}
             placeholder="Describe the image you want to generate..."
             rows={1}
-            disabled={isBusy}
           />
           <InferenceControls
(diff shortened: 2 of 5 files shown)


ramkrishna2910:
The context size field's first digit cannot be removed, if I want to go from 4096 to 8192 I have to move the cursor to the front then type the new first digit.
image

ramkrishna2910:
Minor nit: 
Once a model is loaded the gear icon disappears.
If I want to change the context size of a loaded model, I have to unload the model, change the ctx size and reload again.

bitgamma:
looks great! I only found two small bugs, I can file a separate issue for those if you wish, but here they are:

* The CPU activity indicator jumps around frantically as if possessed. (stats and system-stats are called without any timeout)
* The server-logs pane cannot be resized

PR #1197, opened 2026-02-26
Migrate conffile to /etc not /opt/etc

superm1:


jeremyfowers:
@sofiageo @bitgamma @Geramy would one of you mind reviewing? This is outside my depth.

Geramy:
Yeah in 15 I can take a look

Geramy (approved):
Looks good to me, setting the env var to the config inside the service since its being relocated and directories are being created during install, and cleanup done during removal.

I assume User and Group being lemonade is okay in this case for having a unified directory/cache system that works with existing engines and hugging face?

Looks good.

PR #1200, opened 2026-02-26
Introduces Aixlog logging framework

superm1:
1. replaces all std::cout/std::cerr with structured LOG(SEVERITY, "Tag") calls
2. Cleans up noisy messages — downgrades many INFO messages to DEBUG, removes decorative separators
4. Adds new CLI features — --host/--port on 5 subcommands (list, pull, delete, status, stop) 
4. Rewrites Linux server discovery — uses pgrep/lsof instead of PID file
5. Rewrites systemd detection — adds INVOCATION_ID check and isatty() guard
7. Suppresses subprocess output — redirects to /dev/null or NUL when not in debug mode 
9. Fixese a bug in crouter.cpp (log_level_→blog_level_)

### DIFF
--- a/src/cpp/tray/platform/linux_tray.cpp
+++ b/src/cpp/tray/platform/linux_tray.cpp
@@ -2,6 +2,7 @@
 
 #include "lemon_tray/platform/linux_tray.h"
 #include <iostream>
+#include <lemon/utils/aixlog.hpp>
 
 // Headless stub implementation for Linux
 // This avoids LGPL dependencies (GTK3, libappindicator3, libnotify)
@@ -11,7 +12,6 @@ namespace lemon_tray {
 
 LinuxTray::LinuxTray()
     : should_exit_(false)
-    , log_level_("info")
 {
     // Headless mode - no initialization needed
 }
@@ -24,10 +24,8 @@ bool LinuxTray::initialize(const std::string& app_name, const std::string& icon_
     app_name_ = app_name;
     icon_path_ = icon_path;
 
-    if (log_level_ == "debug") {
-        std::cout << "[Linux Tray] Headless mode - tray not supported on Linux" << std::endl;
-        std::cout << "[Linux Tray] Please use --no-tray flag to run in headless mode" << std::endl;
-    }
+    LOG(DEBUG, "LinuxTray") << "Headless mode - tray not supported on Linux" << std::endl;
+    LOG(DEBUG, "LinuxTray") << "Please use --no-tray flag to run in headless mode" << std::endl;
 
     // Call ready callback immediately since there's no UI to initialize
     if (ready_callback_) {
@@ -40,31 +38,23 @@ bool LinuxTray::initialize(const std::string& app_name, const std::string& icon_
 }
 
 void LinuxTray::run() {
-    if (log_level_ == "debug") {
-        std::cout << "[Linux Tray] Headless mode - no event loop to run" << std::endl;
-    }
+    LOG(DEBUG, "LinuxTray") << "Headless mode - no event loop to run" << std::endl;
     // No-op in headless mode
 }
 
 void LinuxTray::stop() {
-    if (log_level_ == "debug") {
-        std::cout << "[Linux Tray] Headless mode - stopping" << std::endl;
-    }
+    LOG(DEBUG, "LinuxTray") << "Headless mode - stopping" << std::endl;
     should_exit_ = true;
 }
 
 void LinuxTray::set_menu(const Menu& menu) {
-    if (log_level_ == "debug") {
-        std::cout << "[Linux Tray] Headless mode - ignoring menu with "
-                  << menu.items.size() << " items" << std::endl;
-    }
+    LOG(DEBUG, "LinuxTray") << "Headless mode - ignoring menu with "
+    << menu.items.size() << " items" << std::endl;
     // No-op in headless mode
 }
 
 void LinuxTray::update_menu() {
-    if (log_level_ == "debug") {
-        std::cout << "[Linux Tray] Headless mode - ignoring menu update" << std::endl;
-    }
+    LOG(DEBUG, "LinuxTray") << "Headless mode - ignoring menu update" << std::endl;
     // No-op in headless mode
 }
 
@@ -74,30 +64,22 @@ void LinuxTray::show_notification(
     NotificationType type)
 {
     // Print to console instead of showing a GUI notification
-    std::cout << "[Notification] " << title << ": " << message << std::endl;
+    LOG(INFO, "Notification") << title << ": " << message << std::endl;
 }
 
 void LinuxTray::set_icon(const std::string& icon_path) {
     icon_path_ = icon_path;
-    if (log_level_ == "debug") {
-        std::cout << "[Linux Tray] Headless mode - ignoring icon: " << icon_path << std::endl;
-    }
+    LOG(DEBUG, "LinuxTray") << "Headless mode - ignoring icon: " << icon_path << std::endl;
     // No-op in headless mode
 }
 
 void LinuxTray::set_tooltip(const std::string& tooltip) {
-    if (log_level_ == "debug") {
-        std::cout << "[Linux Tray] Headless mode - ignoring tooltip: " << tooltip << std::endl;
-    }
+    LOG(DEBUG, "LinuxTray") << "Headless mode - ignoring tooltip: " << tooltip << std::endl;
     // No-op in headless mode
 }
 
 void LinuxTray::set_ready_callback(std::function<void()> callback) {
     ready_callback_ = callback;
-}
-
-void LinuxTray::set_log_level(const std::string& log_level) {
-    log_level_ = log_level;
 }
 
 } // namespace lemon_tray
--- a/src/cpp/tray/platform/macos_tray.mm
+++ b/src/cpp/tray/platform/macos_tray.mm
@@ -4,6 +4,7 @@
 #include <iostream>
 #include <memory>
 #include <string>
+#include <lemon/utils/aixlog.hpp>
 
 // macOS imports for system tray
 #import <Cocoa/Cocoa.h>
@@ -156,7 +157,7 @@ MacOSTray::MacOSTray() : impl_(nullptr), tooltip_("") {}
 
 MacOSTray::~MacOSTray() {
     if (impl_) {
-        CFRelease(impl_); 
+        CFRelease(impl_);
         impl_ = nullptr;
     }
 }
@@ -169,13 +170,13 @@ bool MacOSTray::initialize(const std::string& app_name, const std::string& icon_
     [NSApp setActivationPolicy:NSApplicationActivationPolicyAccessory];
 
     MacOSTrayImpl* trayImpl = [[MacOSTrayImpl alloc] init];
-    
+
     if (ready_callback_) {
         trayImpl.readyCallback = [[CallbackWrapper alloc] initWithCallback:ready_callback_];
     }
-    
+
     // ARC Retain
-    impl_ = (__bridge_retained void*)trayImpl; 
+    impl_ = (__bridge_retained void*)trayImpl;
 
     NSStatusBar *statusBar = [NSStatusBar systemStatusBar];
     trayImpl.statusItem = [statusBar statusItemWithLength:NSVariableStatusItemLength];
@@ -214,13 +215,9 @@ bool MacOSTray::initialize(const std::string& app_name, const std::string& icon_
 }
 
 void MacOSTray::run() {
-    if (log_level_ == "debug") {
-        std::cout << "[macOS Tray] Entering Run Loop..." << std::endl;
-        
-    }
+    LOG(DEBUG, "macOS Tray") << "Entering Run Loop..." << std::endl;
     [NSApp run];
 }
-
 void MacOSTray::stop() {
     if (impl_) {
         dispatch_async(dispatch_get_main_queue(), ^{
@@ -275,31 +272,32 @@ void MacOSTray::show_notification(const std::string& title, const std::string& m
                     UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
                     content.title = nsTitle;
                     content.body = nsMessage;
-                    
+
                     NSString *uuid = [[NSUUID UUID] UUIDString];
                     UNNotificationRequest *request = [UNNotificationRequest requestWithIdentifier:uuid content:content trigger:nil];
                     [center addNotificationRequest:request withCompletionHandler:nil];
                 }
             }];
         }
-    } 
-    // 2. CLI/Debug Mode (No Bundle ID) - Uses Deprecated API safely
-    else {
-        if (current_log_level == "debug") std::cout << "[macOS Tray] Using legacy notification fallback" << std::endl;
-        
-        // Silence warnings for this specific block
+    }
+        // 2. CLI/Debug Mode (No Bundle ID) - Uses Deprecated API safely
+        else {
+            LOG(DEBUG, "macOS Tray") << "Using legacy notification fallback" << std::endl;
+
+            // Silence warnings for this specific block
+
         #pragma clang diagnostic push
         #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-        
+
         NSUserNotification *notification = [[NSUserNotification alloc] init];
         notification.title = nsTitle;
         notification.informativeText = nsMessage;
         notification.soundName = NSUserNotificationDefaultSoundName;
-        
+
         MacOSTrayImpl* trayImpl = (__bridge MacOSTrayImpl*)impl_;
         [[NSUserNotificationCenter defaultUserNotificationCenter] setDelegate:trayImpl];
         [[NSUserNotificationCenter defaultUserNotificationCenter] deliverNotification:notification];
-        
+
         #pragma clang diagnostic pop
     }
 }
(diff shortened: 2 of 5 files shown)


src/cpp/include/lemon/backends/backend_utils.h
```diff
@@ -44,38 +44,41 @@ namespace lemon::backends {
         * Extract ZIP files (Windows/Linux built-in tools)
         * @param zip_path Path to the ZIP file
         * @param dest_dir Destination directory to extract to
+        * @param log_level Output verbosity
         * @return true if extraction was successful, false otherwise
         */
-        static bool extract_zip(const std::string& zip_path, const std::string& dest_dir, const std::string& backend_name);
+        static bool extract_zip(const std::string& zip_path, const std::string& dest_dir, const std::string& backend_name, const std::string& log_level = "info");
```
bitgamma: I'd avoid adding a default value since forgetting to pass `log_level` to the function would go unnoticed and would create a discrepancy between what the user has set (for example debug) and what is being applied by some functions (since they default to info).

superm1: Thanks, this was a design pattern that existed elsewhere and needed to get fixed in more places.  adjusted.

superm1:
Here is an updated view of startup (before you launch a model).

```
╰─❮ ./build/lemonade-server serve
[ServerManager] PID file created: /tmp/lemonade-router.pid
Lemonade Server v9.4.0 started on port 8000
API endpoint: http://127.0.0.1:8000/api/v1
Connect your apps to the endpoint above.
Documentation: https://lemonade-server.ai/
[ServerManager] PID file created: /tmp/lemonade-router.pid
Press Ctrl+C to stop
```

Here is under systemd:
```
Feb 26 08:56:35 lilac-hi systemd[1]: Started Lemonade Server.
Feb 26 08:56:35 lilac-hi lemonade-server[194058]: Starting Lemonade Server...
Feb 26 08:56:35 lilac-hi lemonade-server[194058]:   Version: 9.4.0
Feb 26 08:56:35 lilac-hi lemonade-server[194058]:   Port: 8000
Feb 26 08:56:35 lilac-hi lemonade-server[194058]:   Host: localhost
Feb 26 08:56:35 lilac-hi lemonade-server[194058]:   Log level: info
Feb 26 08:56:35 lilac-hi lemonade-server[194058]: [Server] Detected systemd environment - will use journald for log streaming
Feb 26 08:56:35 lilac-hi lemonade-server[194058]: [Server] [Net Broadcast] Unable to broadcast my existance please use a RFC1918 IPv4,
Feb 26 08:56:35 lilac-hi lemonade-server[194058]: [Server] [Net Broadcast] or hostname that resolves to RFC1918 IPv4.
Feb 26 08:56:36 lilac-hi lemonade-server[194058]: [ModelManager] Building models cache...
Feb 26 08:56:36 lilac-hi lemonade-server[194058]: [ModelManager] Backend availability:
Feb 26 08:56:36 lilac-hi lemonade-server[194058]:   - NPU hardware: Yes
Feb 26 08:56:36 lilac-hi lemonade-server[194058]:   - System RAM: 26.7 GB (max model size: 21.3 GB)
Feb 26 08:56:36 lilac-hi lemonade-server[194058]:   - Largest memory pool: 13.3
Feb 26 08:56:36 lilac-hi lemonade-server[194058]: [ModelManager] Cache built: 52 total, 0 downloaded
Feb 26 08:56:47 lilac-hi lemonade-server[194058]: [Server] Shutdown signal received, exiting...
Feb 26 08:56:47 lilac-hi lemonade-server[194022]: Received interrupt signal, shutting down...
Feb 26 08:56:47 lilac-hi lemonade-server[194022]: Shutting down server...
Feb 26 08:56:47 lilac-hi systemd[1]: Stopping Lemonade Server...
```

superm1:
And here is what model load/unload looks like now.

```
[Server] Loading model: Gemma-3-4b-it-GGUF 
[Router] Loading model: Gemma-3-4b-it-GGUF...
llama-server is ready!
[Router] Model loaded successfully. Total loaded: 1
[Server] POST /api/v1/load - 200
[Server] Unload request received
[Server] Request method: POST, body length: 35
[Server] Content-Type: application/json
[Router] Unload model called: Gemma-3-4b-it-GGUF
[Router] Evicting model: Gemma-3-4b-it-GGUF
[ProcessManager] Process terminated, waiting for GPU driver cleanup...
[Router] Evicted model: Gemma-3-4b-it-GGUF
[Server] Model 'Gemma-3-4b-it-GGUF' unloaded successfully
[Server] POST /api/v1/unload - 200
```

jeremyfowers:
> we will drown in log-level checks and log-level parameter passed in every function

reading the change set I feel this

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -158,7 +158,7 @@ InstallParams LlamaCppServer::get_install_params(const std::string& backend, con
 
     if (backend == "rocm") {
         params.repo = "lemonade-sdk/llamacpp-rocm";
-        std::string target_arch = SystemInfo::get_rocm_arch();
+        std::string target_arch = SystemInfo::get_rocm_arch("info");
```
jeremyfowers: I thought we were against hardcoding log levels

superm1: Yup. this one snuck in, good catch.  Will fix.

superm1:
> looks good to me, but we really need to consider moving to a proper logging framework or we will drown in log-level checks and log-level parameter passed in every function

Any suggestions for a particular cross platform framework?

bitgamma:
haven't used any recently, but this looks pretty straightforward: https://github.com/badaix/aixlog

bitgamma:
while I agree with you in principle, for a long running server instance the log file can hog the system (and kill SSDs long term) so by default the logs should be minimal.

I feel this shows we need a logging framework that makes this much more manageable though

> My design philosophy for logging is that the user can't be trusted to set the right log level before they have a problem, so the _only_ supported log level should be debug.
> 
> I'm wary of any logging choice that would allow users to hide debug info.
> 
> I'm also wary of the sheer amount of changes here: devs now need to remember to add a `if (log_level_ == "debug" || log_level_ == "trace")` virtually every time they add a `cout`? Is there not a more streamlined way to accomplish the same goal?

superm1:
> My design philosophy for logging is that the user can't be trusted to set the right log level before they have a problem, so the only supported log level should be debug.

As @bitgamma points out this will fill up SSDs really quick.  Most software runs at INFO by default and has a mechanism to increase it if a problem comes up.  I don't think we should be straying from that.

> I'm also wary of the sheer amount of changes here

Let me look at switching to a logging framework.  It's unfortunate I found bugs in the process of doing this that are fixed and we should get in, but hopefully switching to a framework isn't too much difficulty.

superm1:
I integrated that framework suggested by @bitgamma .  Have a look again.  I personally like this a lot better.

superm1:
I need to double check all the places that `log_level` was plumbed earlier, so I will double check those next.

bitgamma:
I'll wait until you are done with removing the `log_level` argument from all functions before reviewing, but this is looking good.

superm1:
> I'll wait until you are done with removing the `log_level` argument from all functions before reviewing, but this is looking good.

Should be good to go now, it's rebased on the other big one that just merged (flm-groundwork)

jeremyfowers:
If this PR does all of these things please update the title and description:

1. Introduces the aixlog logging framework — replaces all std::cout/std::cerr with structured LOG(SEVERITY, "Tag") calls                                                                   2. Cleans up noisy messages — downgrades many INFO messages to DEBUG, removes decorative separators                                                                                        3. Adds new CLI features — --host/--port on 5 subcommands (list, pull, delete, status, stop)                                                                                               4. Rewrites Linux server discovery — uses pgrep/lsof instead of PID file                                                                                                                   5. Rewrites systemd detection — adds INVOCATION_ID check and isatty() guard                                                                                                                6. Suppresses subprocess output — redirects to /dev/null or NUL when not in debug mode                                                                                                     8. Fixesea bugaincrouter.cppv(log_level_→blog_level_)ormat change

jeremyfowers:
`claude /review` found some critical-level and moderate-level issues, thoughts?

```
Critical Issues

  1. ~19 instances of missing spaces in error messages (server.cpp)

  The migration introduced broken error strings like:
  "ERROR inhandle_completions"    // should be "ERROR in handle_completions"
  "JSON parse error inhandle_image_edits"
  This affects nearly every error handler in server.cpp. Clearly a find-and-replace artifact.

  2. #pragma once removed from single_instance.h

  The include guard was removed with no #ifndef replacement. This will cause multiple-definition errors if included from more than one translation unit.

  3. LOG() in signal handler (main.cpp)

  void signal_handler(int signal) {
      LOG(INFO) << "Shutdown signal received, exiting..." << std::endl;

  LOG() involves mutexes and memory allocation — not async-signal-safe. Use write(STDOUT_FILENO, ...) or keep the old std::cout (also technically unsafe but simpler).

  4. LOG() after fork() in child process (process_manager.cpp)

  After fork(), using LOG() (which holds mutexes cloned from the parent) is unsafe in multi-threaded programs.

  5. Doubled tags in log messages

  Several log messages now have redundant tags:
  LOG(ERROR, "FastFlowLM") << "[FastFlowLM ERROR] Failed to download..."
  LOG(DEBUG, "LlamaCpp") << "[LlamaCpp] Model loaded on port..."
  LOG(ERROR, "WrappedServer") << "[WrappedServer] Backend returned HTTP..."

  ---
  Moderate Issues

  6. Inconsistent timestamp format
  - main.cpp: "%Y-%m-%d %H:%M:%S.#ms" (colons)
  - tray/main.cpp: "%Y-%m-%d %H-%M-%S.#ms" (dashes)

  7. is_quiet_get — dead code (server.cpp)

  A variable is computed to detect low-signal GET requests but is never used. Looks like an incomplete implementation.

  8. Leftover std::cerr in fastflowlm_server.cpp

  Two std::cerr calls were not converted to LOG() right after a LOG(ERROR) call.

  9. Redundant "Warning:" prefix

  Several messages now read [WARNING] (ModelManager) Warning: Could not load... — double "warning".

  10. LOG() before AixLog::Log::init() (main.cpp)

  The "Another instance already running" error is logged before the logging system is initialized.

  11. Indentation issues in router.cpp and tray_app.cpp

  Extra indentation (8 spaces where 4 expected) introduced during the conversion.

  12. dynamic_mem_gb → virtual_mem_gb rename

  This is a data format breaking change. If any cached hardware JSON still uses "dynamic_mem_gb", the field will be silently ignored. Needs coordination.

  13. Removed default parameters (backend_utils.h, wrapped_server.h, fastflowlm_server.h)

  Default values removed from install_from_github(..., progress_cb = nullptr) and constructor log_level = "info". Any external callers relying on defaults will break at compile time.     
```

superm1:
> `claude /review` found some critical-level and moderate-level issues, thoughts?
> 
> ```
> Critical Issues
> 
>   1. ~19 instances of missing spaces in error messages (server.cpp)
> 
>   The migration introduced broken error strings like:
>   "ERROR inhandle_completions"    // should be "ERROR in handle_completions"
>   "JSON parse error inhandle_image_edits"
>   This affects nearly every error handler in server.cpp. Clearly a find-and-replace artifact.
> 
>   2. #pragma once removed from single_instance.h
> 
>   The include guard was removed with no #ifndef replacement. This will cause multiple-definition errors if included from more than one translation unit.
> 
>   3. LOG() in signal handler (main.cpp)
> 
>   void signal_handler(int signal) {
>       LOG(INFO) << "Shutdown signal received, exiting..." << std::endl;
> 
>   LOG() involves mutexes and memory allocation — not async-signal-safe. Use write(STDOUT_FILENO, ...) or keep the old std::cout (also technically unsafe but simpler).
> 
>   4. LOG() after fork() in child process (process_manager.cpp)
> 
>   After fork(), using LOG() (which holds mutexes cloned from the parent) is unsafe in multi-threaded programs.
> 
>   5. Doubled tags in log messages
> 
>   Several log messages now have redundant tags:
>   LOG(ERROR, "FastFlowLM") << "[FastFlowLM ERROR] Failed to download..."
>   LOG(DEBUG, "LlamaCpp") << "[LlamaCpp] Model loaded on port..."
>   LOG(ERROR, "WrappedServer") << "[WrappedServer] Backend returned HTTP..."
> 
>   ---
>   Moderate Issues
> 
>   6. Inconsistent timestamp format
>   - main.cpp: "%Y-%m-%d %H:%M:%S.#ms" (colons)
>   - tray/main.cpp: "%Y-%m-%d %H-%M-%S.#ms" (dashes)
> 
>   7. is_quiet_get — dead code (server.cpp)
> 
>   A variable is computed to detect low-signal GET requests but is never used. Looks like an incomplete implementation.
> 
>   8. Leftover std::cerr in fastflowlm_server.cpp
> 
>   Two std::cerr calls were not converted to LOG() right after a LOG(ERROR) call.
> 
>   9. Redundant "Warning:" prefix
> 
>   Several messages now read [WARNING] (ModelManager) Warning: Could not load... — double "warning".
> 
>   10. LOG() before AixLog::Log::init() (main.cpp)
> 
>   The "Another instance already running" error is logged before the logging system is initialized.
> 
>   11. Indentation issues in router.cpp and tray_app.cpp
> 
>   Extra indentation (8 spaces where 4 expected) introduced during the conversion.
> 
>   12. dynamic_mem_gb → virtual_mem_gb rename
> 
>   This is a data format breaking change. If any cached hardware JSON still uses "dynamic_mem_gb", the field will be silently ignored. Needs coordination.
> 
>   13. Removed default parameters (backend_utils.h, wrapped_server.h, fastflowlm_server.h)
> 
>   Default values removed from install_from_github(..., progress_cb = nullptr) and constructor log_level = "info". Any external callers relying on defaults will break at compile time.     
> ```

The data format I'm not worried about - we're going to axe cache on upgrade. A lot of the others make sense. Since you have it on /review already maybe just have it fix em and send it and we can review again.

superm1:
> > `claude /review` found some critical-level and moderate-level issues, thoughts?
> > ```
> > Critical Issues
> > 
> >   1. ~19 instances of missing spaces in error messages (server.cpp)
> > 
> >   The migration introduced broken error strings like:
> >   "ERROR inhandle_completions"    // should be "ERROR in handle_completions"
> >   "JSON parse error inhandle_image_edits"
> >   This affects nearly every error handler in server.cpp. Clearly a find-and-replace artifact.
> > 
> >   2. #pragma once removed from single_instance.h
> > 
> >   The include guard was removed with no #ifndef replacement. This will cause multiple-definition errors if included from more than one translation unit.
> > 
> >   3. LOG() in signal handler (main.cpp)
> > 
> >   void signal_handler(int signal) {
> >       LOG(INFO) << "Shutdown signal received, exiting..." << std::endl;
> > 
> >   LOG() involves mutexes and memory allocation — not async-signal-safe. Use write(STDOUT_FILENO, ...) or keep the old std::cout (also technically unsafe but simpler).
> > 
> >   4. LOG() after fork() in child process (process_manager.cpp)
> > 
> >   After fork(), using LOG() (which holds mutexes cloned from the parent) is unsafe in multi-threaded programs.
> > 
> >   5. Doubled tags in log messages
> > 
> >   Several log messages now have redundant tags:
> >   LOG(ERROR, "FastFlowLM") << "[FastFlowLM ERROR] Failed to download..."
> >   LOG(DEBUG, "LlamaCpp") << "[LlamaCpp] Model loaded on port..."
> >   LOG(ERROR, "WrappedServer") << "[WrappedServer] Backend returned HTTP..."
> > 
> >   ---
> >   Moderate Issues
> > 
> >   6. Inconsistent timestamp format
> >   - main.cpp: "%Y-%m-%d %H:%M:%S.#ms" (colons)
> >   - tray/main.cpp: "%Y-%m-%d %H-%M-%S.#ms" (dashes)
> > 
> >   7. is_quiet_get — dead code (server.cpp)
> > 
> >   A variable is computed to detect low-signal GET requests but is never used. Looks like an incomplete implementation.
> > 
> >   8. Leftover std::cerr in fastflowlm_server.cpp
> > 
> >   Two std::cerr calls were not converted to LOG() right after a LOG(ERROR) call.
> > 
> >   9. Redundant "Warning:" prefix
> > 
> >   Several messages now read [WARNING] (ModelManager) Warning: Could not load... — double "warning".
> > 
> >   10. LOG() before AixLog::Log::init() (main.cpp)
> > 
> >   The "Another instance already running" error is logged before the logging system is initialized.
> > 
> >   11. Indentation issues in router.cpp and tray_app.cpp
> > 
> >   Extra indentation (8 spaces where 4 expected) introduced during the conversion.
> > 
> >   12. dynamic_mem_gb → virtual_mem_gb rename
> > 
> >   This is a data format breaking change. If any cached hardware JSON still uses "dynamic_mem_gb", the field will be silently ignored. Needs coordination.
> > 
> >   13. Removed default parameters (backend_utils.h, wrapped_server.h, fastflowlm_server.h)
> > 
> >   Default values removed from install_from_github(..., progress_cb = nullptr) and constructor log_level = "info". Any external callers relying on defaults will break at compile time.     
> > ```
> 
> The data format I'm not worried about - we're going to axe cache on upgrade. A lot of the others make sense. Since you have it on /review already maybe just have it fix em and send it and we can review again.

I've incorporated solutions for most of this and rebased.

superm1:
@jeremyfowers Since this touches nearly all of the code base, can you review this again for merging soon?  It needs a rebase hourly otherwise..

jeremyfowers:
> @jeremyfowers Since this touches nearly all of the code base, can you review this again for merging soon? It needs a rebase hourly otherwise..

Sorry, doing it now.

PR #1201, opened 2026-02-26
Enable clangd by default

superm1:
I've not been able to use clangd with this project and finally got it working.  Yay call hierarchies now work.

image


bitgamma (approved):

bitgamma:
that's useful!

PR #1206, opened 2026-02-26
Fixes StatusBar loop - split useEffects and updates dependencies

sofiageo:
Quick fix for endless loop fetch call. It fetches stats every 15 seconds when connected. In a random test this only took 6 seconds to update after I stopped the lemonade server


jeremyfowers:
@superm1 modifications to your code, can you review?

superm1 (approved):

PR #1212, opened 2026-02-27
Added import / export model 

choppu:
* Added export model as JSON file functionality
* Added import model from JSON functionality

Screenshot 2026-02-27 alle 16 15 53
Screenshot 2026-02-27 alle 16 24 29

### DIFF
--- a/src/app/src/renderer/ModelManager.tsx
+++ b/src/app/src/renderer/ModelManager.tsx
@@ -26,7 +26,7 @@ interface ModelManagerProps {
 
 interface ModelJSON {
   id?: string,
-  model?: string,
+  model_name?: string,
   recipe: string,
   recipe_options?: object,
   checkpoint?: string,
@@ -371,7 +371,7 @@ const ModelManager: React.FC<ModelManagerProps> = ({ isVisible, width = 280, cur
     }));
   };
 
-  const handleDownloadModel = useCallback(async (modelName: string, registrationData?: ModelRegistrationData, isExportedModel?: boolean) => {
+  const handleDownloadModel = useCallback(async (modelName: string, registrationData?: ModelRegistrationData) => {
     let downloadId: string | null = null;
 
     try {
@@ -396,7 +396,7 @@ const ModelManager: React.FC<ModelManagerProps> = ({ isVisible, width = 280, cur
       setLoadingModels(prev => new Set(prev).add(modelName));
 
       // Use the single consolidated download function
-      await pullModel(modelName, { registrationData: registrationData, isExportedModel: isExportedModel });
+      await pullModel(modelName, { registrationData: registrationData });
 
       await fetchCurrentLoadedModel();
       showSuccess(`Model "${modelName}" downloaded successfully.`);
@@ -561,23 +561,37 @@ const ModelManager: React.FC<ModelManagerProps> = ({ isVisible, width = 280, cur
     const file = files[0];
     const reader = new FileReader();
     reader.onload = (e) => {
-      const result = JSON.parse(e.target?.result as string);
-      uploadModelJSON(result);
+      try {
+        const result = JSON.parse(e.target?.result as string);
+        uploadModelJSON(result);
+      } catch(err: any) {
+        showError(`Failed to parse JSON file. ${err}`);
+      }
     };
     reader.readAsText(file);
     event.target.value = '';
   }
 
   const uploadModelJSON = (json: ModelJSON) => {
-    if (json.checkpoint && json.checkpoints) delete json.checkpoint;
-    if (json.id && json.model) delete json.id;
+    let modelName: string;
 
-    if(json.id && !json.model) {
-      json.model = json.id;
-      delete json.id;
+    if (!json.recipe) {
+      showError("Invalid model JSON. Recipe is missing");
+      return;
     }
 
-    handleDownloadModel(json.model as string, json as ModelRegistrationData, true);
+    if(!json.model_name && !json.id) {
+      showError("Invalid model JSON. Either model or id must be present.");
+      return;
+    }
+
+    modelName = json.model_name ? json.model_name : json.id as string;
+
+    if (json.checkpoint && json.checkpoints) delete json.checkpoint;
+    if (json.model_name) delete json.model_name;
+    if(json.id) delete json.id;
+
+    handleDownloadModel(modelName as string, json as ModelRegistrationData);
   }
 
   const viewTitle = currentView === 'models'
--- a/src/app/src/renderer/ModelOptionsModal.tsx
+++ b/src/app/src/renderer/ModelOptionsModal.tsx
@@ -211,11 +211,10 @@ const ModelOptionsModal: React.FC<SettingsModalProps> = ({ isOpen, onCancel, onS
   };
 
   const handleModelExport = () => {
-    URL.revokeObjectURL(exportModelBtn!.current!.href);
     let modelName = (modelInfo?.id as string).startsWith("user.") ? modelInfo?.id : `user.${modelInfo?.id}`;
 
     let modelToExport = {
-      "model": modelName,
+      "model_name": modelName,
       "downloaded": modelInfo?.downloaded,
       "labels": modelInfo?.labels,
       "recipe": modelInfo?.recipe,
@@ -233,7 +232,8 @@ const ModelOptionsModal: React.FC<SettingsModalProps> = ({ isOpen, onCancel, onS
     const blob = new Blob([model], { type: "application/json" });
     const url = URL.createObjectURL(blob);
     exportModelBtn!.current!.href = url;
-    exportModelBtn!.current!.download = modelToExport?.model ? `${modelToExport?.model}.json` as string: 'model.json';
+    exportModelBtn!.current!.download = modelToExport?.model_name ? `${modelToExport?.model_name}.json` as string: 'model.json';
+    setTimeout(() => URL.revokeObjectURL(url), 1000);
   }
 
   const handleCancel = () => {
--- a/src/app/src/renderer/utils/backendInstaller.ts
+++ b/src/app/src/renderer/utils/backendInstaller.ts
@@ -338,23 +338,10 @@ export async function pullModel(
 
   try {
     // Build request body — include registration data for custom models
-    let requestBody: Record<string, unknown>;
+    let requestBody: Record<string, unknown> = { model_name: modelName, stream: true };
 
-    if(options?.isExportedModel && options?.registrationData) {
-      requestBody = {...options.registrationData} as Record<string, any>;
-      requestBody.stream = true;
-    } else {
-      requestBody = { model_name: modelName, stream: true }
-      if (options?.registrationData) {
-        const rd = options.registrationData;
-        requestBody.checkpoint = rd.checkpoint;
-        requestBody.recipe = rd.recipe;
-        if (rd.mmproj) requestBody.mmproj = rd.mmproj;
-        if (rd.reasoning) requestBody.reasoning = rd.reasoning;
-        if (rd.vision) requestBody.vision = rd.vision;
-        if (rd.embedding) requestBody.embedding = rd.embedding;
-        if (rd.reranking) requestBody.reranking = rd.reranking;
-      }
+    if(options?.registrationData) {
+      Object.assign(requestBody, options.registrationData);
     }
 
     const response = await serverFetch('/pull', {


bitgamma:
I am really looking forward getting this included. Makes moving user models from one instance to the other effortless (as well as making sharing them easy).

src/app/src/renderer/ModelManager.tsx
```diff
@@ -536,6 +551,35 @@ const ModelManager: React.FC<ModelManagerProps> = ({ isVisible, width = 280, cur
...
+
+    if (!files || files.length === 0) {
+      return;
+    }
+
+    const file = files[0];
+    const reader = new FileReader();
+    reader.onload = (e) => {
+      const result = JSON.parse(e.target?.result as string);
+      uploadModelJSON(result);
```
jeremyfowers: **Bug: No error handling on `JSON.parse`**

`JSON.parse` can throw on malformed input, and since this is reading a user-supplied file, that's a realistic scenario. An invalid JSON file will throw an uncaught exception here.

Suggestion: wrap in try/catch and show an error toast:
```tsx
reader.onload = (e) => {
  try {
    const result = JSON.parse(e.target?.result as string);
    uploadModelJSON(result);
  } catch (err) {
    showError("Failed to parse JSON file. Please check the file format.");
  }
};
```

src/app/src/renderer/ModelManager.tsx
```diff
@@ -536,6 +551,35 @@ const ModelManager: React.FC<ModelManagerProps> = ({ isVisible, width = 280, cur
...
+    if (json.checkpoint && json.checkpoints) delete json.checkpoint;
+    if (json.id && json.model) delete json.id;
+
+    if(json.id && !json.model) {
+      json.model = json.id;
+      delete json.id;
+    }
+
+    handleDownloadModel(json.model as string, json as ModelRegistrationData, true);
+  }
```
jeremyfowers, line 595: **Bug: No validation of imported JSON**

The parsed JSON is cast to `ModelRegistrationData` with no validation. If required fields like `recipe` or `model`/`id` are missing, `handleDownloadModel` will be called with `undefined` as the model name, which will fail in confusing ways.

Suggestion: validate that at least `recipe` and one of `model`/`id` exist and are strings before proceeding, and show a user-facing error for invalid files:
```tsx
const uploadModelJSON = (json: ModelJSON) => {
  if (!json.recipe || (!json.model && !json.id)) {
    showError("Invalid model JSON: missing required fields (recipe, model/id).");
    return;
  }
  // ... rest of normalization
};
```

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -336,16 +338,23 @@ export async function pullModel(
...
-      if (rd.mmproj) requestBody.mmproj = rd.mmproj;
-      if (rd.reasoning) requestBody.reasoning = rd.reasoning;
-      if (rd.vision) requestBody.vision = rd.vision;
-      if (rd.embedding) requestBody.embedding = rd.embedding;
-      if (rd.reranking) requestBody.reranking = rd.reranking;
+    let requestBody: Record<string, unknown>;
+
+    if(options?.isExportedModel && options?.registrationData) {
+      requestBody = {...options.registrationData} as Record<string, any>;
+      requestBody.stream = true;
```
jeremyfowers: **Bug: `model_name` missing from request body**

When `isExportedModel` is true, the request body is spread from `registrationData` which contains a `model` field, but the `/pull` endpoint expects `model_name`. The non-exported path at line 347 sets `model_name` correctly. This will likely cause server-side failures when importing a model.

Fix:
```ts
requestBody.model_name = modelName;
```

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -336,16 +338,23 @@ export async function pullModel(
...
+        const rd = options.registrationData;
+        requestBody.checkpoint = rd.checkpoint;
+        requestBody.recipe = rd.recipe;
+        if (rd.mmproj) requestBody.mmproj = rd.mmproj;
+        if (rd.reasoning) requestBody.reasoning = rd.reasoning;
+        if (rd.vision) requestBody.vision = rd.vision;
+        if (rd.embedding) requestBody.embedding = rd.embedding;
+        if (rd.reranking) requestBody.reranking = rd.reranking;
+      }
     }
```
jeremyfowers: **Suggestion: remove `isExportedModel` flag, unify both paths**

The two branches do nearly the same thing — the only difference is that exported models carry extra fields (`checkpoints`, `labels`, `size`, etc.). You could eliminate the flag entirely by always building the base request body with `model_name` + `stream`, then spreading all `registrationData` fields on top:

```ts
const requestBody: Record<string, unknown> = { model_name: modelName, stream: true };
if (options?.registrationData) {
  Object.assign(requestBody, options.registrationData);
}
```

This is simpler, avoids the duplicated field-mapping logic, and automatically supports any new fields added to `ModelRegistrationData` in the future.

src/app/src/renderer/ModelOptionsModal.tsx
```diff
@@ -209,6 +210,32 @@ const ModelOptionsModal: React.FC<SettingsModalProps> = ({ isOpen, onCancel, onS
     setOptions(createDefaultOptions(options.recipe));
   };
 
+  const handleModelExport = () => {
+    URL.revokeObjectURL(exportModelBtn!.current!.href);
```
jeremyfowers: **Minor: Object URL memory leak**

`URL.revokeObjectURL` is called on the previous URL at the start of this function, but the object URL created at line 234 is never revoked after the download completes. Each export click leaks a blob URL.

Consider revoking after a short delay to let the browser initiate the download:
```ts
const url = URL.createObjectURL(blob);
exportModelBtn.current!.href = url;
exportModelBtn.current!.download = ...;
exportModelBtn.current!.click(); // trigger download programmatically
setTimeout(() => URL.revokeObjectURL(url), 1000);
```
Or revoke in the modal's cleanup/close handler.

jeremyfowers (commented):
Review comments on a few issues found in the import/export implementation.

src/app/src/renderer/ModelManager.tsx
jeremyfowers: _Written by Claude Code_

src/app/src/renderer/ModelManager.tsx
jeremyfowers: _Written by Claude Code_

src/app/src/renderer/utils/backendInstaller.ts
jeremyfowers: _Written by Claude Code_

src/app/src/renderer/utils/backendInstaller.ts
jeremyfowers: _Written by Claude Code_

src/app/src/renderer/ModelOptionsModal.tsx
jeremyfowers: _Written by Claude Code_

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -23,7 +23,8 @@ function extractServerErrorMessage(errorText: string, fallback: string): string
  * Registration data for custom user models sent with /pull requests.
```
jeremyfowers, line 23: I tried this first by exporting my own model, and then importing it. That worked great!

Then I tried this json from @bitgamma:

`Qwen3.5-35B-A3B-NoThinking.json`

```
{
    "checkpoints": {
        "main": "unsloth/Qwen3.5-35B-A3B-GGUF:Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf",
        "mmproj": "unsloth/Qwen3.5-35B-A3B-GGUF:mmproj-F16.gguf"
    },
    "model": "user.Qwen3.5-35B-A3B-NoThinking",
    "labels": [
        "vision",
        "hot",
        "tool-calling"
    ],
    "recipe": "llamacpp",
    "recipe_options": {
        "ctx_size": 0,
        "llamacpp_args": "--temp 1.0 --top-p 0.95 --top-k 20 --min-p 0.00 --no-context-shift --keep 0 -b 1024 -ub 512 --chat-template-kwargs '{\"enable_thinking\": false}'",
        "llamacpp_backend": "vulkan"
    },
    "size": 19.7
}

First I noticed `"ctx_size": 0`: typo?

Second, when I tried to load it I got this:

```
[ProcessManager] Starting process with filtered output: "C:\Users\Jeremy\.cache\lemonade/bin\llamacpp\vulkan\llama-server.exe" "-m" "C:\Users\Jeremy\.cache\huggingface\hub/models--unsloth--Qwen3.5-35B-A3B-GGUF\snapshots\48b9db213ecf5e4945393ac903dbc52ca852e081\Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf" "--ctx-size" "4096" "--port" "8001" "--jinja" "--mmproj" "C:\Users\Jeremy\.cache\huggingface\hub/models--unsloth--Qwen3.5-35B-A3B-GGUF\snapshots\48b9db213ecf5e4945393ac903dbc52ca852e081\mmproj-F16.gguf" "--reasoning-format" "auto" "--no-webui" "-ngl" "99" "--temp" "1.0" "--top-p" "0.95" "--top-k" "20" "--min-p" "0.00" "--no-context-shift" "--keep" "0" "-b" "1024" "-ub" "512" "--chat-template-kwargs" "{"enable_thinking": false}"
[ProcessManager] Process started successfully, PID: 22100
Waiting for llama-server to be ready...
load_backend: loaded RPC backend from C:\Users\Jeremy\.cache\lemonade\bin\llamacpp\vulkan\ggml-rpc.dll
load_backend: loaded CPU backend from C:\Users\Jeremy\.cache\lemonade\bin\llamacpp\vulkan\ggml-cpu-zen4.dll
error while handling argument "--chat-template-kwargs": [json.exception.parse_error.101] parse error at line 1, column 2: syntax error while parsing object key - invalid literal; last read: '{e'; expected string literal

usage:
--chat-template-kwargs STRING           sets additional params for the json template parser, must be a valid
                                        json object string, e.g. '{"key1":"value1","key2":"value2"}'
                                        (env: LLAMA_CHAT_TEMPLATE_KWARGS)


to show complete usage, run with -h
```

Do I need a new llamacpp binary? Is the json wrong? Or is this a bug in this PR?

jeremyfowers (changes requested):
Please see comments. I really like the feature and think it will help us build an awesome recipe wiki!

`claude /review` found a few potential bugs I would like you to consider, particularly the json validation suggestion.

Please also see my experience with `Qwen3.5-35B-A3B-NoThinking.json` and let me know your thoughts.

src/app/src/renderer/utils/backendInstaller.ts
bitgamma: ctx_size 0 is valid and what I always use, it means "read CTX size from model" this is a feature from llamacpp.

Both files import well through the CLI so there must be a problem with the way it is handled in the PR or something windows specific, I see the single quotes around the embedded json are lost in the log.

@choppu please test using my files

bitgamma: ok I just checked the code, indeed the way the command line is built for Windows (turning it into a string) is not compatible with arguments that contain so `--chat-template-kwargs '{"enable_thinking": false}'` becomes `"--chat-template-kwargs" "'{"enable_thinking": false}'"` which breaks. In Unix-like the argument array is passed as is (as an array) and thus all this quoting issues are automatically solved.

Since the problem is unrelated to this PR and the changes needed are in `ProcessManager::start_process` I'll open a separate PR to handle this.

src/app/src/renderer/ModelManager.tsx
choppu: I have wrapped JSON.parse in try/catch as suggested.

src/app/src/renderer/ModelManager.tsx
choppu: Added the validation of imported JSON

src/app/src/renderer/utils/backendInstaller.ts
choppu: Adopted this suggestion in my code.

src/app/src/renderer/ModelOptionsModal.tsx
choppu: Added URL.revokeObjectURL

choppu:
Thank you for the review 🙏 I addressed all your comments and updated my PR accordingly.

src/app/src/renderer/utils/backendInstaller.ts
jeremyfowers: Thanks for checking @bitgamma!

TIL about ctx size 0!

jeremyfowers:
@claude please provide a final review for this PR

PR #1213, opened 2026-02-27
Revert "Use mDNS beacon to discover server instead of `lemonade-server status`"

jeremyfowers:
Reverts lemonade-sdk/lemonade#1158


superm1:
close in favor of #1215 and #1216

PR #1215, opened 2026-02-27
Add loopback unicast to NetworkBeacon for reliable same-machine disco…

Geramy:
INADDR_BROADCAST does not reliably deliver to loopback listeners, causing same-machine discovery to fail without an active NIC or router. This adds a second sendto to INADDR_LOOPBACK each interval and includes 127.0.0.0/8 in isRFC1918() so the beacon starts even when the server IP is loopback.


Geramy:
This may not be the fix for the problem with electron, but it is a edge case bug. I am now looking at electron now that there are no more edge cases in the network announcement.

PR #1216, opened 2026-02-27
Add background beacon listener for automatic server port discovery

Geramy:
- Add continuous UDP beacon listener in main process that detects server port changes and broadcasts updates to the renderer
- Fix LogsWindow SSE reconnection to avoid race conditions between discovery callbacks and useEffect re-runs
- Route websocketClient health check through serverFetch for automatic port discovery on failure
- Clean up TranscriptionPanel to use serverFetch instead of raw fetch

### DIFF
--- a/src/app/main.js
+++ b/src/app/main.js
@@ -556,6 +556,76 @@ const discoverServerPort = () => {
   });
 };
 
+/**
+ * Background beacon listener that continuously monitors for lemonade server
+ * UDP beacons on localhost. When a beacon is received with a different port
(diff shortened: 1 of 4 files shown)


superm1:
Does this run in web app?  I think it only is needed for electron.

Geramy:
> Does this run in web app? I think it only is needed for electron.

well I did add the start beacon to app.on 'ready' which is a electron event so it technically shouldn't?

superm1:
Yeah that should be fine then, thanks.

PR #1225, opened 2026-02-27
Enumerate all RFC1918 interfaces for network beacon broadcasting

Geramy:
Previously the beacon relied on resolving the server's --host argument to determine the broadcast IP and payload URL. When the host was "localhost" or "0.0.0.0", this resolved to 127.0.0.1 which either failed the RFC1918 gate or produced a useless payload for remote clients.

The beacon now enumerates all local network interfaces at each broadcast cycle using getifaddrs (POSIX) / GetAdaptersAddresses (Windows). For each interface with an RFC1918 address, it sends a beacon to that interface's directed broadcast address with the correct IP in the payload. This ensures clients on any local subnet can discover the server regardless of how --host is configured.

Additional changes:
- Added NetworkInterfaceInfo struct and getLocalRFC1918Interfaces()
- Loopback (127.0.0.1) always receives a beacon for same-machine discovery
- Server gate now checks for any RFC1918 interface instead of the resolved host IP
- Added 127.0.0.0/8 to isRFC1918() so loopback-only configs can still beacon
- Added per-interface logging with source IP and broadcast destination

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -896,6 +896,12 @@ void Server::run() {
     //The beacon will send per-interface with the correct IP in the payload.
     auto rfc1918Interfaces = udp_beacon_.getLocalRFC1918Interfaces();
     if(!rfc1918Interfaces.empty() && !no_broadcast_) {
+        std::cout << "[Server] [Net Broadcast] Broadcasting on " << rfc1918Interfaces.size()
+                  << " RFC1918 interface(s):";
+        for (const auto& iface : rfc1918Interfaces) {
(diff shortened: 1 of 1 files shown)


jeremyfowers:
Please have a Windows Electron app user like @Tetramatrix review this PR to ensure there are no regressions.

Geramy:
> Please have a Windows Electron app user like @Tetramatrix review this PR to ensure there are no regressions.

will do.

Geramy:
dont merge yet.

Geramy:
should be good to merge now.

PR #1231, opened 2026-02-28
Fix image generation request timeout

### LINKED ISSUE #1222: Timeout while waiting for image generation
Using Lemonade Server v9.3.4 on Windows 11, with an image model and a prompt to generate an image, webapp shows an error `Failed to generate image: Network error: CURL error: Timeout was reached` after maybe 10 minutes, while the image generation is still in progress.
Here are a few lines from the output of `lemonade-server serve --log-level trace`:

```
[INFO ] stable-diffusion.cpp:3390 - generating image: 1/1 - seed 42
  |=====>                                            | 2/20 - 44.25s/it
  |=======>                                          | 3/20 - 44.38s/it
  |============>                                     | 5/20 - 44.30s/it
  |================================>                 | 13/20 - 44.37s/it
[Server] Error 500: POST /api/v1/images/generations
[Server] POST /api/v1/images/generations - 500
  |==========================================>       | 17/20 - 44.36s/it
  |==================================================| 20/20 - 44.35s/it
[INFO ] stable-diffusion.cpp:3771 - generate_image completed in 904.06s
```

szkocot:
## Summary

  Fixes #1222.

  Image generation requests for `sd-cpp` could time out after 10 minutes even while the backend was still actively generating the image.
  This change removes the hardcoded timeout behavior and adds a configurable server-side setting for image request timeouts.

  ## Changes

  - replace the hardcoded 600 second Stable Diffusion request timeout
  - add a new server config option:
    - CLI: `--image-request-timeout <seconds>`
    - env var: `LEMONADE_IMAGE_REQUEST_TIMEOUT=<seconds>`
  - default `image_request_timeout` to `0` meaning no overall timeout
  - apply the setting consistently to:
    - image generations
    - image edits
    - image variations
  - update the frontend recipe-options mirror so config metadata stays in sync with the server

  ## Behavior Before

  Stable Diffusion requests used a fixed 10 minute timeout in the backend forwarding layer. Long-running image generation jobs could fail
  with a network timeout even though `sd-server` was still processing the request.

  ## Behavior After

  Stable Diffusion image requests now use a configurable timeout, with the default set to no overall timeout.

  Examples:

  ```bash
  lemonade-server serve --image-request-timeout 1800

  LEMONADE_IMAGE_REQUEST_TIMEOUT=1800 lemonade-server serve

  0 disables the timeout entirely.

  ## Testing

  - built lemonade-router successfully:
      - cmake --build --preset default --target lemonade-router

  ## Notes

  - this setting is implemented in the existing server config / recipe options path
  - the fix is server-side, so it applies to CLI, web app, and other clients using the same server

### DIFF
--- a/src/app/src/renderer/recipes/recipeOptionsConfig.ts
+++ b/src/app/src/renderer/recipes/recipeOptionsConfig.ts
@@ -213,7 +213,6 @@ export const OPTION_DEFINITIONS: Record<string, OptionDef> = {
     label: 'Height',
     description: 'Image height in pixels',
   },
-
   // Common option - save settings
   saveOptions: {
     type: 'boolean',
--- a/src/cpp/include/lemon/backends/fastflowlm_server.h
+++ b/src/cpp/include/lemon/backends/fastflowlm_server.h
@@ -59,7 +59,7 @@ public:
                                    const std::string& request_body,
                                    httplib::DataSink& sink,
                                    bool sse = true,
-                                   long timeout_seconds = 0) override;
+                                   long timeout_seconds = -1) override;
 
 private:
     // Static helpers for install logic (no instance state needed)
--- a/src/cpp/include/lemon/utils/http_client.h
+++ b/src/cpp/include/lemon/utils/http_client.h
@@ -72,19 +72,19 @@ public:
     static HttpResponse post(const std::string& url,
                             const std::string& body,
                             const std::map<std::string, std::string>& headers = {},
-                            long timeout_seconds = 300);
+                            long timeout_seconds = -1);
 
     // Multipart form data POST request
     static HttpResponse post_multipart(const std::string& url,
                                        const std::vector<MultipartField>& fields,
-                                       long timeout_seconds = 300);
+                                       long timeout_seconds = -1);
 
     // Streaming POST request (calls callback for each chunk as it arrives)
     static HttpResponse post_stream(const std::string& url,
                                    const std::string& body,
                                    StreamCallback stream_callback,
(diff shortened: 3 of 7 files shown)


jeremyfowers:
Thanks @szkocot!

im wondering though do we really need the timeout at all? Why not just set it to 0 and skip the env var or CLI option?

szkocot:
@jeremyfowers Defaulting to 0 avoids surprising long-running jobs, but keeping the option is still useful for production or shared deployments where operators may want a safety limit (hung backends, resource control, load balancing, etc.). It’s basically flexibility at no cost for users who don’t need it.

jeremyfowers:
Looks like @superm1 is merging #1272 which is highly related.

@szkocot is it ok for you to just set the new global timeout to 0 for you use case, or do you still want a specific change for image gen?

If you want something specific for image gen, please change this PR so that it has just the env var, not the CLI arg, and the imagegen timeout var supercedes the global timeout for image gen.

szkocot:
Yes, sure.

jeremyfowers:
> Looks like @superm1 is merging #1272 which is highly related.
> 
> @szkocot is it ok for you to just set the new global timeout to 0 for you use case, or do you still want a specific change for image gen?
> 
> If you want something specific for image gen, please change this PR so that it has just the env var, not the CLI arg, and the imagegen timeout var supercedes the global timeout for image gen.

https://github.com/lemonade-sdk/lemonade/pull/1422 makes sure that the global timeout applies to image generation, which solves the core issue.

PR #1233, opened 2026-02-28
Switch to FLM 0.9.35

superm1:
A variety of things need to happen when FLM 0.9.35 releases.  This branch tracks them all.

1. Drop the Linux beta flag
2. Use `--json` output from FLM
3. Set new FLM minimum version to 0.9.35
4. Update Lemonade release to 10.0.0.

Remaining things that we're waiting on to add before merging it.

- [x] `flm validate --json` needs to return values for Windows with NPU driver.  The Lemonade code that looks for driver version needs to be dropped and use the json on Windows
- [ ] Need to make sure that all labels are carried over properly
- [ ] Need to test the upgrade paths (what happens with the branch on older FLM version installed)
- [ ] Need to update Windows CI runners to newer FLM version (maybe leave one at old version and make sure we gracefully fail?)
- [ ] Need to add FLM to Linux CI runners

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -721,6 +721,16 @@ jobs:
           python-version: '3.10'
           requirements-file: 'test/requirements.txt'
 
+      - name: Install FLM backend for FLM wrapped-server tests
+        if: ${{ runner.os == 'Windows' && contains(matrix.extra_args, '--wrapped-server flm') }}
+        shell: PowerShell
+        run: |
+          $ErrorActionPreference = "Stop"
+          $serverExe = Join-Path $env:LEMONADE_INSTALL_PATH "bin\lemonade-server.exe"
+          Write-Host "Installing FLM backend for CI inference tests..." -ForegroundColor Cyan
+          & $serverExe recipes --install flm:npu
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+
       - name: Run tests
         shell: PowerShell
         env:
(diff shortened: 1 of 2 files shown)


jeremyfowers (commented):
Does this change current set the names in Lemonade to be the exact flm name? Like `qwen3:8b`?

Need to think about how users will know this is an NPU model. In the past we had `-NPU`, `-Hybrid`, and `-FLM` suffixes so they were aware.

This is important because:
1. People assume ollama-style model names mean llamacpp GGUFs on GPUs, which is not the case here
2. The NPU has unique restrictions (only 1 LLM on NPU at a time) and performance limitations (less memory bandwidth than GPU)

So people should be aware of what they're loading. `lemoande-server run qwen3:8b` doesn't carry that information.

jeremyfowers:
Also if we were to ever add a translation table for ollama model names for use with ollama API that would collide with the FLM model names.

jeremyfowers:
Another musing... it might *finally* be time to drop "recipes" at the top level.

If we support Qwen3-8B on cpu, gpu, and npu, why not just show the model in the model list ONCE and just offer the user a choice of device?

edit: one limitation of this strategy is that GPU vs. NPU is different model files. We dont want to trigger a lot of random giant downloads just because someone played with the backend menu.

superm1:
The problem with offering a list of devices is third party clients. Like think running in openwebui.

So I think we really do want to keep suffixes.

jeremyfowers:
> The problem with offering a list of devices is third party clients. Like think running in openwebui.
> 
> So I think we really do want to keep suffixes.

So convert their colons to dashes and append -flm?

qwen3:8b becomes qwen3-8b-flm

superm1:
> > The problem with offering a list of devices is third party clients. Like think running in openwebui.
> > So I think we really do want to keep suffixes.
> 
> So convert their colons to dashes and append -flm?
> 
> qwen3:8b becomes qwen3-8b-flm

ya

jeremyfowers:
Reminder to self: ensure this doesn’t accidentally break the gpt-oss-20b filtering on windows.

superm1:
I've updated this branch OP to reflect an ongoing TODO.  CI failures are expected until that list is all covered.

PR #1235, opened 2026-02-28
Add support for passing custom arguments to FLM backend through --flm-args

Jays-1111:
Motivation:
- FLM supports -s (socket count) and -q (queue length) for concurrency control.
- Certain AI agent scenarios require higher concurrency than the default FLM settings, but there was previously no way to configure these options through lemonade-server, causing agents to fail under load.

Changes:
- recipe_options.cpp: Add flm_args default, CLI option (LEMONADE_FLM_ARGS), recipe key
- fastflowlm_server.cpp: Read, parse and append flm_args to FLM command

Usage: lemonade-server serve --flm-args "-s 20 -q 15"


src/cpp/server/backends/fastflowlm_server.cpp
```diff
@@ -165,28 +142,28 @@ void FastFlowLMServer::load(const std::string& model_name,
                            const RecipeOptions& options,
```
jeremyfowers, line 142: Please add documentation to https://github.com/lemonade-sdk/lemonade/blob/main/docs/server/lemonade-server-cli.md for the new CLI and env var options.

jeremyfowers (changes requested):
Tested locally and it works for me!

Please see comment about documentation.

Also we could support passing FLM arguments via the GUI like we do for llamacpp:

Image

Please open an issue for a followup if you think that is a good idea!

src/cpp/server/backends/fastflowlm_server.cpp
Jays-1111: Done.

Jays-1111:
@jeremyfowers Great! I've already updated the doc. For the function of the GUI, I haven't used this feature before, but I think it's a good idea. I'll look into it separately later.

jeremyfowers (approved):
Thank you for the improvement @Jays-1111 !

PR #1239, opened 2026-02-28
Add support for NPU Total Operations Per Second (TOPs) in status bar

superm1:
This is not utilization; that's going to be a lot more work and need a newer kernel driver; it will need to come a lot later.

image


superm1:
It would go down under some conditions but yeah tend to agree. Will move it.

superm1:
closed in favor of https://github.com/lemonade-sdk/lemonade/pull/1261

PR #1243, opened 2026-02-28
Enable NPU for Linux

superm1:
For whenever we decide is the right time to include it for people to test with.


jeremyfowers (approved):
lets go!

PR #1244, opened 2026-02-28
Add an NPU linux beta flag

superm1:


superm1:
Intentionally done separately from https://github.com/lemonade-sdk/lemonade/pull/1243 so that this one can be reverted when we're ready to stop calling it beta.

jeremyfowers:
> Intentionally done separately from #1243 so that this one can be reverted when we're ready to stop calling it beta.

Not working for me. Does this require the change from https://github.com/lemonade-sdk/lemonade/pull/1243/changes in order to work?

jeremyfowers:
> > Intentionally done separately from #1243 so that this one can be reverted when we're ready to stop calling it beta.
> 
> Not working for me. Does this require the change from https://github.com/lemonade-sdk/lemonade/pull/1243/changes in order to work?

tried it both in my .conf as well as killing the systemd service, setting the env var, and starting `lemonade-server serve`

jeremyfowers:
cherry-picked that in but now get

`flm                 npu         unsupported     Requires AMD XDNA 2 AMD NPU `

superm1:
You need both to test; I want to merge them separately so it is a clean simple revert.

You need to set it in the systemd service conffile manually for it to work.

jeremyfowers:
> You need both to test; I want to merge them separately so it is a clean simple revert.
> 
> You need to set it in the systemd service conffile manually for it to work.

I see, so the PRs themselves cant be tested directly

superm1:
If you want me to squash them I can; but then it's not a clean revert when we're ready to end the beta.

PR #1246, opened 2026-03-01
Fix incorrect skipping of PID file and instance locks

abn:
JOURNAL_STREAM is inherited by all processes in a systemd login session, not just actual systemd service processes. Guards in write_pid_file(), remove_pid_file(), and SingleInstance::IsAnotherInstanceRunning() were using this env var to detect a systemd service context, but this caused those mechanisms to be bypassed whenever lemonade was started from a terminal on a modern systemd-based Linux system.

Consequences:
- `lemonade-server status` always reported "Server is not running" even when the server was running (PID file was never written)
- Multiple instances of lemonade-router, lemonade-server, or lemonade-tray could run simultaneously (lock file was never created or checked)

get_server_info() already handles the true systemd service case via is_any_systemd_service_active() before falling through to the PID file, so writing the PID file and lock files unconditionally is safe.

### DIFF
--- a/src/cpp/include/lemon/single_instance.h
+++ b/src/cpp/include/lemon/single_instance.h
@@ -39,13 +39,6 @@ public:
 
         return false;
 #else
-        // Skip lock file when running under systemd, unless explicitly disabled
-        const char* journal_stream = std::getenv("JOURNAL_STREAM");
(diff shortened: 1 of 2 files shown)


abn:
Just for completeness here is a screenshot from a new terminal session that shows the JOURNAL_STREAM var set.

image

superm1 (approved):

PR #1249, opened 2026-03-01
Fix false 'not running' status in sandboxed Linux environments

abn:
In Flatpak (without --share=pid) and similar PID-namespace-isolated environments, getpgid() returns -1 for a live server process because the PID in the file is sandbox-internal and invisible to the caller. This caused the PID file to be deleted and status to falsely report the server as not running.

Add an HTTP liveness check (GET /api/version) before treating a negative getpgid() result as stale. If the server responds, the PID file is kept and the server is reported as running.

Also extract a shared http_server_alive() helper and apply both the pre-stale HTTP check and the lsof/pgrep port-based fallback to all Unix platforms — both tools are available on Linux and macOS, and the sandbox isolation concern is not Linux-exclusive.

### DIFF
--- a/src/cpp/tray/tray_app.cpp
+++ b/src/cpp/tray/tray_app.cpp
@@ -1015,6 +1015,17 @@ bool TrayApp::is_server_running_on_port(int port) {
     }
 }
 
+// Helper: returns true if the router is reachable via HTTP on the given port.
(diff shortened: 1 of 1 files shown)


superm1 (approved):
I have to admit I don't like the fallback path with calling other processes, but it's nothing new.

Thanks!

abn:
I concur, not a fan of that either :smile:

PR #1252, opened 2026-03-01
Make Vulkan take priority over CPU for whisper.cpp

jeremyfowers:
Unless there was some reason CPU should be prioritized over Vulkan?


ramkrishna2910 (approved):

iswaryaalex:
Sounds good!

PR #1255, opened 2026-03-01
Try ptyxis before gnome-terminal when opening logs on Linux

abn:
Ptyxis is the default GNOME terminal since GNOME 47, so try it first before falling back to gnome-terminal and then xterm.

### DIFF
--- a/src/cpp/tray/tray_app.cpp
+++ b/src/cpp/tray/tray_app.cpp
@@ -2707,7 +2707,10 @@ void TrayApp::on_show_logs() {
     pid_t pid = fork();
     if (pid == 0) {
         // Child process
-        std::string cmd = "gnome-terminal -- tail -f '" + log_file_ + "' || xterm -e tail -f '" + log_file_ + "'";
+        std::string tail_cmd = "tail -f '" + log_file_ + "'";
+        std::string cmd = "ptyxis -- " + tail_cmd
+            + " || gnome-terminal -- " + tail_cmd
+            + " || xterm -e " + tail_cmd;
         execl("/bin/sh", "sh", "-c", cmd.c_str(), nullptr);
         exit(0);
     } else if (pid > 0) {


src/cpp/tray/tray_app.cpp
```diff
@@ -2707,7 +2707,10 @@ void TrayApp::on_show_logs() {
     pid_t pid = fork();
     if (pid == 0) {
         // Child process
-        std::string cmd = "gnome-terminal -- tail -f '" + log_file_ + "' || xterm -e tail -f '" + log_file_ + "'";
+        std::string tail_cmd = "tail -f '" + log_file_ + "'";
```
jeremyfowers, line 2710: Should we wait for you to add the tray on Linux before doing this? Not sure how to test it.

superm1: I also have a general problem with opening lists of applications and setting priorities.  My system I happen to have `konsole`, `ptyxis`, `warp`, `cosmic-terminal`, `gnome-terminal` all installed.  

My preference isn't going to be the same as yours. 

---

That being said - why is a terminal being opened to look at logs?  Why not use `xdg-open` with the log which open it in the user's preferred file viewing application.

abn: > Should we wait for you to add the tray on Linux before doing this? Not sure how to test it.

Good question, I use the patch in the [rpm](https://github.com/abn/lemonade-rpm/blob/main/003-fix-terminal-call.patch). But that won't be much help for you unless you are installing the rpm :smile: 

If you are okay with the current [tray patch](https://github.com/lemonade-sdk/lemonade/compare/main...abn:lemonade:support-linux-tray), I can create a PR for that for now and deal with a longer term strategy for tray later.

> I also have a general problem with opening lists of applications and setting priorities. My system I happen to have konsole, ptyxis, warp, cosmic-terminal, gnome-terminal all installed.

I agree with this, I just did not want to change any of the existing logic. I added `ptyxis` to the top of the list as this is the default since GNOME 47. `gnome-terminal` and `xterm` is not available on default installs afaik.

The `xdg-open` option can work, but that means there is no live logs - which is the current behavior. There is also `x-terminal-exec` but that is not commonly adopted across distros. I do not think there is a "one" great option here unfortunately.

superm1: I have another suggestion on what to do if you want live logs.  How about if you open up the web app and focus the logs tab?  We have all that log streaming infrastructure already built and that makes it invisible whether you use systemd or any other means to the client.

abn: If that is already available, then I think that probably a better approach. Can just use `xdg-open` in that case I suppose. Happy to make that change if it is preferable.

superm1: I think you'll need to plumb some way to make the log panel show up by default.  But yeah otherwise yeah you can just launch the web app with `lemonade-web-app` from `$PATH`

superm1: Maybe we should add support for the args passed to that binary as something passed to URL.

so you could call
`lemonade-web-app log`

And that turns into something like 

http://localhost:8000/?param=log

Which causes the log panel to open.

jeremyfowers: > If you are okay with the current [tray patch](https://github.com/lemonade-sdk/lemonade/compare/main...abn:lemonade:support-linux-tray), I can create a PR for that for now and deal with a longer term strategy for tray later.If you are okay with the current [tray patch](https://github.com/lemonade-sdk/lemonade/compare/main...abn:lemonade:support-linux-tray), I can create a PR for that for now and deal with a longer term strategy for tray later.

@abn I'm curious to try it out! Please do make the PR.

PS. This is why I thought you were volunteering to add a Linux tray haha. Guess that's not the same as adding a linux tray and overhauling the whole tray codebase? :D

PR #1258, opened 2026-03-01
Add initial STT transcription UX for chat

Rebreda:
I wanted to be able to "talk" with the chatbot in an intuitive way but without reinventing everything (there's already the existing transcription flow). First PR so apologies if i've missed some conventions! There's still more to be done on the UI side of things but figure this is a good enough starting point to see if the feature is interesting! Ideally there'd be some tests and probably more code reusae across the different panels.

https://github.com/user-attachments/assets/7c7dfb3d-fc9a-45c3-839d-0f313bc71416

### DIFF
--- a/src/app/src/renderer/hooks/useVoiceTranscription.ts
+++ b/src/app/src/renderer/hooks/useVoiceTranscription.ts
@@ -18,23 +18,23 @@ interface UseVoiceTranscriptionOptions {
 }
 
 interface UseVoiceTranscriptionResult {
-  whisperModel: string | undefined;
+  activeModel: string | undefined;
   isRecording: boolean;
   start: () => Promise<void>;
   stop: () => void;
 }
 
 /**
- * Returns the name of an already-loaded whispercpp model from the server, or
+ * Returns the name of an already-loaded audio model from the server, or
  * `null` if none is loaded or the health check fails.
  */
-async function fetchLoadedWhisperModel(modelsData: ModelsData): Promise<string | null> {
+async function fetchLoadedAudioModel(modelsData: ModelsData): Promise<string | null> {
   try {
     const res = await serverFetch('/health');
     if (!res.ok) return null;
     const health = await res.json();
     const allLoaded: { model_name: string }[] = health.all_models_loaded || [];
-    const loaded = allLoaded.find((m) => modelsData[m.model_name]?.recipe === 'whispercpp');
+    const loaded = allLoaded.find((m) => modelsData[m.model_name]?.labels?.includes('audio'));
     return loaded?.model_name ?? null;
   } catch {
     return null;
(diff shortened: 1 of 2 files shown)


Rebreda:
ah, i think i narrowed the issue down. If the transcription takes "too long" (3 seconds was the default) the ws connection closes. I was testing with whisper-base/tiny meaning output was quick before. I'll play around with some better defaults, maybe 30 seconds? On my setup whiser-large-v3 takes a long time to output a completed sentence. 

I also added a check to use whatever whisper model is loaded (if there is one)

jeremyfowers:
> ah, i think i narrowed the issue down. If the transcription takes "too long" (3 seconds was the default) the ws connection closes. I was testing with whisper-base/tiny meaning output was quick before. I'll play around with some better defaults, maybe 30 seconds? On my setup whiser-large-v3 takes a long time to output a completed sentence.
> 
> I also added a check to use whatever whisper model is loaded (if there is one)

Do we need to have a timeout for closing the ws connection? On my iPhone, if I press the dictation/mic button it keeps dictating forever until I click that mic button again to manually disable it. I think that is the experience most people are familiar with.

Its also what we currently support on the Transcription UI, so probably less unique code to enable it in LLM Chat that way?

src/app/src/renderer/hooks/useVoiceTranscription.ts
```diff
@@ -0,0 +1,214 @@
...
+ * Returns the name of an already-loaded whispercpp model from the server, or
+ * `null` if none is loaded or the health check fails.
+ */
+async function fetchLoadedWhisperModel(modelsData: ModelsData): Promise<string | null> {
+  try {
+    const res = await serverFetch('/health');
+    if (!res.ok) return null;
+    const health = await res.json();
+    const allLoaded: { model_name: string }[] = health.all_models_loaded || [];
+    const loaded = allLoaded.find((m) => modelsData[m.model_name]?.recipe === 'whispercpp');
```
jeremyfowers: Is it possible to filter by the `audio` label instead? Models with `recipe: flm` and `label: audio` should be allowed too.

src/app/src/renderer/hooks/useVoiceTranscription.ts
```diff
@@ -0,0 +1,214 @@
...
+  textareaRef,
+  runPreFlight,
+  reset,
+  onAutoSubmit,
+  onError,
+}: UseVoiceTranscriptionOptions): UseVoiceTranscriptionResult {
+  const { modelsData } = useModels();
+  const whisperModel = Object.entries(modelsData).find(
+    ([, info]) => info.recipe === 'whispercpp' && info.downloaded
+  )?.[0] ?? Object.keys(modelsData).find(name => modelsData[name].recipe === 'whispercpp');
```
jeremyfowers: Same comment as https://github.com/lemonade-sdk/lemonade/pull/1258/changes#r2881407905 may apply in a few places.

jeremyfowers: I had a bad initial experience with this PR because I had downloaded Whisper-Large-v3 but not Whisper-Base, so Whisper-Large-v3 was auto-selected. It than ran too slow for realtime use.

I suggest getting the smallest downloaded model instead of the 0th to ensure it is speedy.

jeremyfowers: PS. This works *great* with Whisper Base! Love it! I only had a bad experience because it happened to default to Large through coincidence.

src/app/src/renderer/hooks/useVoiceTranscription.ts
jeremyfowers: Yeah I just confirmed that `whisper-v3-turbo-FLM` gets evicted if I push the mic button. I would like to use it.

jeremyfowers (changes requested):
This PR is nearly perfect! Please just adjust the default model and enable FLM transcription and we are good to merge :)

src/app/src/renderer/hooks/useVoiceTranscription.ts
Rebreda: Nice, figured there was a better filter than "whispercpp"

src/app/src/renderer/hooks/useVoiceTranscription.ts
Rebreda: > I suggest getting the smallest downloaded model instead of the 0th to ensure it is speedy.

Good idea, it's a lot more deterministic/best effort than whatever happened to be sitting at index 0.

Rebreda:
Should be good now @jeremyfowers

PR #1259, opened 2026-03-01
Reduce FLM beta only to `flm validate`

superm1:
This helps to avoid hardware detection getting messed up based upon whether beta flag is set.


superm1:
> Approving... but what about your single revertable commit?

Decided this is more important that beta is usable :(

I found that without this it's so much harder if you don't set variable right the first time.

PR #1261, opened 2026-03-01
Add support for NPU Utilization in status bar

superm1:
Requires this kernel patch to work.  Should be a no-op otherwise.

https://lore.kernel.org/dri-devel/20260228061109.361239-3-superm1@kernel.org/

image

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -1,5 +1,11 @@
 PORTIONS LICENSED AS FOLLOWS
 
+## amdxdna_accel.h
+
+The file `src/cpp/include/lemon/amdxdna_accel.h` is licensed under GPL-2.0 WITH Linux-syscall-note.
+
+The Linux-syscall-note exception allows this GPL-licensed kernel UAPI header to be included in userspace applications without imposing GPL requirements on the application itself. This exception is standard for Linux kernel headers that define the userspace API for system calls and ioctl operations.
+
 ## llama.cpp
 
 Binaries for llama.cpp are downloaded under the MIT license from https://github.com/ggml-org/llama.cpp, as well as https://github.com/lemonade-sdk/llamacpp-rocm (which uses https://github.com/ggml-org/llama.cpp to build them.)


jeremyfowers:
Not showing up for me even though FLM models work in the web app.

```
jfowers@hp-omni-1:~/lemonade$ curl http://localhost:8000/v1/system-stats
{"cpu_percent":2.2135856708347417,"gpu_percent":1.0,"memory_gb":4.1,"npu_percent":null,"vram_gb":0.6321678161621094}
```

Seems like a problem under the hood since npu_percent is null.

superm1:
> Not showing up for me even though FLM models work in the web app.
> 
> ```
> jfowers@hp-omni-1:~/lemonade$ curl http://localhost:8000/v1/system-stats
> {"cpu_percent":2.2135856708347417,"gpu_percent":1.0,"memory_gb":4.1,"npu_percent":null,"vram_gb":0.6321678161621094}
> ```
> 
> Seems like a problem under the hood since npu_percent is null.

It's working as designed.  You need a kernel patch that will merge for kernel 7.1.

jeremyfowers:
> > Not showing up for me even though FLM models work in the web app.
> > ```
> > jfowers@hp-omni-1:~/lemonade$ curl http://localhost:8000/v1/system-stats
> > {"cpu_percent":2.2135856708347417,"gpu_percent":1.0,"memory_gb":4.1,"npu_percent":null,"vram_gb":0.6321678161621094}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Seems like a problem under the hood since npu_percent is null.
> 
> It's working as designed. You need a kernel patch that will merge for kernel 7.1.

oh a new new kernel patch?

src/cpp/include/lemon/amdxdna_accel.h
```diff
@@ -0,0 +1,527 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
```
jeremyfowers, line 1: GPL-2.0

Doesn't that mean we can't paste it here?

superm1: Hmm.  It's a copy of the header from the kernel.  We can't relicense it.

https://github.com/torvalds/linux/blob/master/include/uapi/drm/amdxdna_accel.h

The version in Ubuntu is too old, so we can't add a build dependency on it.  How do handle it then?

superm1:
> > It's working as designed. You need a kernel patch that will merge for kernel 7.1.
> 
> oh a new new kernel patch?

Correct.  Details in OP.

src/cpp/include/lemon/amdxdna_accel.h
superm1: I believe the syscall exception **explicitly** allows for this.

https://www.kernel.org/doc/html/latest/process/license-rules.html

Here's what an LLM told me when I was asking about it.
```
Including GPL-licensed Linux kernel UAPI headers in an Apache-licensed project is generally allowed, thanks to a specific licensing exception in the Linux kernel.

Here’s the key point:

1. The "Linux-syscall-note" exception

Many kernel UAPI headers (found under include/uapi/) are marked with a special note:
"This header is part of the Linux UAPI; including it in userspace code does not make that code subject to the GPL."

This means you can include these headers in your project — even if your project is under a permissive license like Apache 2.0 — without your code becoming GPL-licensed.
The rationale is that these headers define the interface between user space and the kernel, and the kernel developers explicitly allow their use in non-GPL software.
2. What to watch out for

Only UAPI headers have this exception. If you include non-UAPI GPL kernel headers, your code could be considered a derivative work and thus subject to GPL requirements.
Check the top of each header file for the license notice. If it contains the "Linux-syscall-note" or similar wording, you’re safe.
Avoid copying large chunks of implementation code from GPL files — stick to including headers for system calls and constants.
3. Practical workflow

Include only headers from include/uapi/ or arch/*/include/uapi/.
Keep your Apache 2.0 license intact; no need to relicense.
Document in your README that you rely on Linux UAPI headers under the syscall exception.
```

superm1: So I believe the action is to update `NOTICES`

superm1: I've clarified the `NOTICE.md` that this is allowed.

PR #1262, opened 2026-03-02
Drop "default" args for `lemonade-router`

superm1:
As the backend manager has been introduced it doesn't really make sense to pass arguments from lemonade-server to lemonade-router unless a user has changed the defaults.

### DIFF
--- a/src/cpp/server/recipe_options.cpp
+++ b/src/cpp/server/recipe_options.cpp
@@ -144,14 +144,6 @@ void RecipeOptions::add_cli_options(CLI::App& app, json& storage) {
             const auto& result = backend_cache[recipe];
             std::string default_backend = result.backends.empty() ? "" : result.backends[0];
 
-            // Pre-populate storage with the dynamically detected default so it's
-            // available even when the user doesn't explicitly pass the flag.
-            // (add_option_function's callback only fires on explicit CLI input,
-            // and default_val only affects help text display.)
-            if (!default_backend.empty()) {
-                storage[opt_name] = default_backend;
-            }
-
             o = app.add_option_function<std::string>(key, [opt_name, &storage = storage](const std::string& val) { storage[opt_name] = val; }, opt["help"]);
             o->default_val(default_backend);
             o->check(CLI::IsMember(result.backends));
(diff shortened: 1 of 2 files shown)


src/cpp/server/recipe_options.cpp
```diff
@@ -245,6 +237,22 @@ RecipeOptions RecipeOptions::inherit(const RecipeOptions& options) const {
 }
 
 json RecipeOptions::get_option(const std::string& opt) const {
-    return options_.contains(opt) ? options_[opt] : DEFAULTS[opt];
+    if (options_.contains(opt)) {
+        return options_[opt];
+    }
+
+    // Dynamic defaults for backends if not explicitly set
```
bitgamma, line 255: what is the advantage of doing it this way? Also this code is a duplicate of the one `add_cli_options` so I think it'd be better to extract it into its own function

superm1: Well I don't know that we need to do this for all arguments. For example maybe we should make host and port "mandatory" arguments.

The biggest reason I want to do this is some arguments are confusing with backend management.

For example specifying the llamacpp doesnt make sense to me.

bitgamma: I am not sure I am following you, what do you mean?

BTW: my comment was relative to the changes to `recipe_options.cpp` only. Also the suggestion to extract the duplicated code was concerning the code that detects if the option is _backend and reads from system info, since it is basically a copy of what is already in the `add_cli_options` function

superm1: I've made some changes to require passing the port and host and rebased on all the other work the last few days.

My point here is that it's really confusing that currently lemonade-router is launched with a `--llama-cpp=vulkan` and `--sdcpp=rocm`  argument when it's actually an "auto" behavior for both.

With the branch as it is now the process looks like this:

```
superma+ 1015831  0.1  0.0  95732  8380 pts/5    Sl+  21:21   0:00  |   |           \_ ./build/lemonade-server serve
superma+ 1015832  0.2  0.0 530260 15152 pts/5    Sl+  21:21   0:00  |   |               \_ /home/supermario/src/lemonade/build/lemonade-router --port 8000 --host localhost
```

superm1:
like so many of my PRs this is overlapping with the system one.  I will land the relevant pieces in that and we can revisit later if needed.

PR #1271, opened 2026-03-02
Clarify system requirement on FLM Linux setup

### LINKED ISSUE #1241: FLM Linux docs should make it clear only XDNA 2 is supported
Cc @superm1 

To avoid a lot of confusion at launch

jeremyfowers:
Closes #1241 

image

### DIFF
--- a/docs/flm_npu_linux.html
+++ b/docs/flm_npu_linux.html
@@ -22,15 +22,16 @@
       <div class="install-section">
         <h2>System requirements</h2>
         <p>
-          FLM on Linux currently requires an <strong>AMD XDNA 2</strong> NPU.
+          FLM on Linux requires an <strong>AMD XDNA 2</strong> NPU.
         </p>
         <ul>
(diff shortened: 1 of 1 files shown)


docs/flm_npu_linux.html
```diff
@@ -19,6 +19,22 @@
 
     <!-- NPU Setup Content -->
     <div class="install-content">
+      <div class="install-section">
+        <h2>System requirements</h2>
+        <p>
+          FLM on Linux currently requires an <strong>AMD XDNA 2</strong> NPU.
+        </p>
+        <ul>
+          <li><strong>Ryzen AI 300-series:</strong> Kraken Point, Strix Point, Strix Halo</li>
```
superm1: Strix Halo is not Ryzen AI.  Isn't it Ryzen AI Max? It should be it's own line item.

jeremyfowers: Updated

PR #1283, opened 2026-03-04
Lay the groundwork for Lemonade Ultra

jeremyfowers:
This PR lays the groundwork for `Lemonade Ultra` by introduces the concept of Experiences, which are bundles of models with a special UI, into server_models.json and the app,

Experiences are not fully enabled E2E yet, but the UI is in place in the app. When an Experience is loaded from the model manager, the UI shifts into Experience Mode. To review the UI, set the `label: experience` models in `server_models.json` to `suggested: true` rebuild and relaunch the server, then launch the app.

In non-Experience mode, the UI should by 99% the same in this PR and in `main`.

The one difference is that the left panel behaves a little differently:
- "Hide left panel" doesn't hide the icon rail
- Clicking the icon rail while the left panel is hidden reveals the left panel
- Clicking the selected icon in the icon rail hides the left panel
- AKA how VS Code works

Out of scope for this PR:
- The exact models in the Experiences
- Whether to support VLMs as a 5th model type in the Experiences, or if the main LLM must be a VLM (this has an implication on --max-loaded-models, which would need to be `2`, which would be a pain...)
- The names of the Experiences
- Finalizing the CSS styles for Experience Mode
- Wiring up the Experience Mode buttons (image input, mic button) to work correctly
- The backend endpoint and tool calling necessary to actually enable the experiences


src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -499,111 +500,147 @@ export async function ensureModelReady(
...
+        const modelsResponse = await serverFetch('/models?show_all=true');
+        if (modelsResponse.ok) {
+          const data = await modelsResponse.json();
+          const modelList = Array.isArray(data) ? data : data.data || [];
+          const freshModel = modelList.find((m: any) => m.id === modelName);
+          isDownloaded = freshModel?.downloaded === true;
         }
-        return;
+      } catch {
+        // If re-check fails, proceed — pull will be a no-op if already downloaded
```
Geramy, line 596: Is this comment correct? we try/catch inside if !isDownloaded but then the comment says if re-check fails proceed - pull will be a no-op is already downloaded, why would this throw? It just seems like a strange comment. I guess maybe a better comment is whats the expected error we are catching with the comment, or no comment.

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -499,111 +500,147 @@ export async function ensureModelReady(
...
       }
+    }
 
-      throw new Error(errorMsg);
+    // Step 5: Pull model if not downloaded (shows in Download Manager)
```
Geramy, line 600: Why is this not being done in step 4 below the first if statement? Why are we doing two if statements?

Geramy: I assume this has something to do with the try/catch but thats inside the ! statement and this can be moved after the catch reducing one op, which is probably compiled out anyways by the JS engine I would hope. :P

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -499,111 +500,147 @@ export async function ensureModelReady(
...
+    if (!isDownloaded) {
+      await pullModel(modelName);
```
Geramy, line 602: Why not just use loadModel() instead directly inside if !isDownloaded ? move const loadmodel above this entire function or area and use it in the above try/catch ?

Geramy (commented):
Please see comments and let me know what you think.

PR #1285, opened 2026-03-04
Simplify default backend selection for system llamacpp

jeremyfowers:
## Summary
- Replace 45-line post-loop `default_backend` selection with a 3-line inline check during the existing `RECIPE_DEFS` iteration
- Since "system" is already first in `RECIPE_DEFS`, preference ordering is handled naturally by iteration order — just skip it when `LEMONADE_LLAMACPP_PREFER_SYSTEM` isn't set
- Net result: -47 lines, +15 lines, same behavior

## Test plan
- [ ] Existing `llamacpp-system` CI tests pass (covers all env var combinations)
- [ ] Other recipe default_backend selection unaffected

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -762,6 +762,16 @@ json SystemInfo::build_recipes_info(const json& devices) {
         detected_devices.push_back({"metal", "Apple Metal", "metal", true});
     }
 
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@superm1 if the tests pass please consider taking this simplification into #1122

superm1:
Thanks. Will retest with system llamacpp, but at least by review looks good.

PR #1286, opened 2026-03-04
escape arguments when building win32 command line calls

bitgamma:
Allows to use complex llamacpp args like `--temp 1.0 --top-p 0.95 --top-k 20 --min-p 0.00 --no-context-shift --keep 0 -b 1024 -ub 512 --chat-template-kwargs '{"enable_thinking": false}'` to work on windows by properly handling quotes


bitgamma:
@jeremyfowers this should address the problem you experienced when testing #1212

jeremyfowers:
@claude review

src/cpp/server/utils/process_manager.cpp
```diff
@@ -68,6 +68,34 @@ static void log_process_line(const std::string& line) {
 }
```
jeremyfowers, line 68: I merged @choppu's PR and tried Import Model again.

Got this error popup (but there is no error in the console):

Image

I'm not sure why it thinks the json is invalid since there is clearly a "model" key:

```
{
    "checkpoints": {
        "main": "unsloth/Qwen3.5-35B-A3B-GGUF:Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf",
        "mmproj": "unsloth/Qwen3.5-35B-A3B-GGUF:mmproj-F16.gguf"
    },
    "model": "user.Qwen3.5-35B-A3B-NoThinking",
    "labels": [
        "vision",
        "hot",
        "tool-calling"
    ],
    "recipe": "llamacpp",
    "recipe_options": {
        "ctx_size": 0,
        "llamacpp_args": "--temp 1.0 --top-p 0.95 --top-k 20 --min-p 0.00 --no-context-shift --keep 0 -b 1024 -ub 512 --chat-template-kwargs '{\"enable_thinking\": false}'",
        "llamacpp_backend": "vulkan"
    },
    "size": 19.7
}
```

@bitgamma or @choppu would you mind investigating? This PR should be the PR that gets `Qwen3.5-35B-A3B-NoThinking.json` working end-to-end.

jeremyfowers (changes requested):
Please see the error message I hit in the app. I am not sure if what claude is pointing out is a real bug or not, but the app error is preventing me from checking.

bitgamma:
turns out my file uses "model" instead of the expected "model_name" or "id". This is because I created the JSON manually.

@choppu please add "model" to the list of valid fields for model since the backend accepts that as well.

@jeremyfowers here are the corrected files that use "model_name" instead of "model" so that this can be properly tested

[Qwen3.5-35B-A3B-NoThinking.json](https://github.com/user-attachments/files/25752685/Qwen3.5-35B-A3B-NoThinking.json)
[Qwen3.5-35B-A3B-ThinkingCoder.json](https://github.com/user-attachments/files/25752686/Qwen3.5-35B-A3B-ThinkingCoder.json)

bitgamma:
actually I'll just correct the files in the wiki since nothing generates the "model" field, it is just cause I created it manually that it turned out this way. It is probably better rather than allowing 3 different names for the same field.

bitgamma:
regarding the review from Claude: the thing it says are right but not in our context I think. The reason is that we want to make the saner bash-like syntax usable on both linux and windows systems, so that the user do not have to manually enter escape sequences but can use the single quotes `'` to embed json in arguments (since this became a thing for Qwen3.5). 

If we add too many windows-specific quoting rules (try to read [this](https://learn.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments?view=msvc-170) without crying) we will make the command line diverge between windows and linux (well only if the user actively make use of those special rules), which is not the goal.

PR #1289, opened 2026-03-04
Add model family grouping in Model Manager

jeremyfowers:
image

## Summary
- Collapse models sharing a family (e.g. Qwen3-0.6B/1.7B/4B/8B/14B-GGUF) into a single row with selectable size-variant pills
- Family row shows aggregate status dot (green if any member loaded, white if any downloaded, gray otherwise) and the selected member's size/actions
- Each pill is color-coded by its member's status (dim=not downloaded, bright=available, green=loaded, blue highlight=selected)
- Non-matching models (e.g. Qwen3-Coder-*, Qwen3-Embedding-*) remain as individual rows
- Family definitions are data-driven via `MODEL_FAMILIES` array — easy to add more families

### DIFF
--- a/src/app/package-lock.json
+++ b/src/app/package-lock.json
@@ -213,16 +213,6 @@
         "global-agent": "^3.0.0"
       }
     },
-    "node_modules/@electron/get/node_modules/semver": {
-      "version": "6.3.1",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.1.tgz",
-      "integrity": "sha512-BR7VvDCVHO+q2xBEWskxS6DJE1qRnb7DxzUrogb71CWoSficBxYsiAGd+Kl0mmq/MprG9yArRkyrQxTO6XjMzA==",
-      "dev": true,
-      "license": "ISC",
-      "bin": {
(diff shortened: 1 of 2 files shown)


jeremyfowers:
I will finish the design if people vote for it here: https://discord.com/channels/1392562559122407535/1394414791593820252/1478771578690080938

jeremyfowers:
Alternate design:

image

I am not totally happy with the design at the head of the branch, but I like this one less. @ramkrishna2910 likes it more.

choppu:
Really nice feature! It makes the model manager more organized. However I am not sure if regex is the best solution. What do you think of just adding family to the server model response?

ramkrishna2910 (approved):
Looks really good! Reduces scrolling.
Should we do the same for FLM models as well?

src/app/package-lock.json
superm1, line 1: this shouldn't need to change for this change should it?

superm1: I'm worried it undoes some of 30f500cbbeb67671417d0e5021c8fcda449cb743

jeremyfowers: Putting it back

PR #1299, opened 2026-03-04
Feature: Add `logs` CLI subcommand and open webapp logs in Tray

abn:
This PR replaces the OS-specific terminal log viewer implementation with the Web UI Logs pane. This change follows from https://github.com/lemonade-sdk/lemonade/pull/1255#discussion_r2875594809.

1.  Added a new CLI subcommand `lemonade-server logs` that launches a browser directly to the web app with the `?logs=true` query parameter to display the logs panel.
2.  Updated `TrayApp::on_show_logs()` to use the new browser-based log viewer approach instead of launching native terminal windows, reducing cross-platform complexity and dependency on apps like `gnome-terminal`.
3.  Updated the React frontend (`App.tsx`) to detect the `?logs=true` query parameter on mount and immediately force the logs panel to be visible.

Closes: #1255

### DIFF
--- a/src/app/src/renderer/App.tsx
+++ b/src/app/src/renderer/App.tsx
@@ -60,6 +60,11 @@ const AppContent: React.FC = () => {
       } catch (error) {
         console.error('Failed to load layout settings:', error);
       } finally {
+        // Override with URL parameters if present
+        const urlParams = new URLSearchParams(window.location.search);
+        if (urlParams.get('logs') === 'true') {
+          setIsLogsVisible(true);
+        }
         setLayoutLoaded(true);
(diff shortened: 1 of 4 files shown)


superm1 (approved):

src/app/package-lock.json
superm1, line 1: Can you drop changes to this file? Or were they needed?

superm1 (approved):
One minor change otherwise lgtm!

src/app/package-lock.json
abn: Fair point, I am not sure if that is directly required. Dropped.

PR #1300, opened 2026-03-04
Add Linux system tray support via AppIndicator3

abn:
`lemonade-tray` now builds on Linux when AppIndicator3 libraries are present, auto-detected at CMake configure time. The GTK-free `ayatana-appindicator-glib` variant is preferred; GTK3 + Ayatana/upstream `libappindicator3` are the fallback. Use `-DREQUIRE_LINUX_TRAY=ON` to require the dependencies and fail if absent. `lemonade-server` on Linux remains always headless (GTK-free, daemon-friendly).

The tray connects to an already-running server, starts one via systemd if a known unit file is installed, or spawns `lemonade-router` directly. A single `fetch_server_state()` call per cycle drives both menu rendering and refresh detection, with all shared menu state guarded by state_mutex_. System-scope systemd actions are labeled "(requires admin)" when polkit elevation is needed. `normalize_connect_host()` is called before constructing `ServerManager` at all Linux `run()` sites to avoid IPv6 resolution issues on dual-stack systems.

setup.sh adds optional tray dependency detection; `BUILD_OPTIONS.md` documents the new CMake option.

### DIFF
--- a/docs/dev-getting-started.md
+++ b/docs/dev-getting-started.md
@@ -24,7 +24,7 @@ This guide covers everything you need to build, test, and contribute to Lemonade
 - [Usage](#usage)
   - [lemonade-router (Server Only)](#lemonade-router-server-only)
   - [lemonade-server.exe (Console CLI Client)](#lemonade-serverexe-console-cli-client)
-  - [lemonade-tray.exe (GUI Tray Launcher)](#lemonade-trayexe-gui-tray-launcher---windows-only)
+  - [lemonade-tray (GUI Tray Application)](#lemonade-tray-gui-tray-application---windows-and-linux)
   - [Logging and Console Output](#logging-and-console-output)
 - [Testing](#testing)
   - [Basic Functionality Tests](#basic-functionality-tests)
@@ -40,6 +40,7 @@ Lemonade consists of these main executables:
 - **lemonade-router.exe** - Core HTTP server executable that handles requests and LLM backend orchestration
 - **lemonade-server.exe** - Console CLI client for terminal users that manages server lifecycle, executes commands via HTTP API
 - **lemonade-tray.exe** (Windows only) - GUI tray launcher for desktop users, automatically starts `lemonade-server.exe serve`
+- **lemonade-tray** (Linux, when compiled with GTK3+AppIndicator3) - System tray launcher using AppIndicator3
 - **lemonade-log-viewer.exe** (Windows only) - Log file viewer with live tail support and installer-friendly file sharing
 
 ## Building from Source
@@ -100,7 +101,8 @@ cmake --build --preset vs18
   - `build/Release/lemonade-log-viewer.exe` - Log file viewer
 - **Linux/macOS:**
   - `build/lemonade-router` - HTTP server
-  - `build/lemonade-server` - Console CLI client
+  - `build/lemonade-server` - Console CLI client (always headless on Linux)
+  - `build/lemonade-tray` - GUI tray application (Linux only, built when AppIndicator3 libraries are found)
 - **Resources:** Automatically copied to `build/Release/resources/` on Windows, `build/resources/` on Linux/macOS (web UI files, model registry, backend version configuration)
 
 ### Building the Electron Desktop App (Optional)
@@ -170,9 +172,10 @@ chmod +x build/app-appimage/Lemonade-*.AppImage
 - Security features enabled: Control Flow Guard, ASLR, DEP
 
 **Linux:**
-- Linux builds are headless-only (no tray application) by default
-- This avoids LGPL dependencies (GTK3, libappindicator3, libnotify)
-- Run server using: `lemonade-server serve` (headless mode is automatic)
+- `lemonade-server` is always headless on Linux (GTK-free, daemon-friendly); use `lemonade-server serve` to start the server
+- `lemonade-tray` is a separate binary for the system tray, auto-detected at build time: built if AppIndicator3 libraries are found (GTK3 only needed for non-glib variants)
+- To require tray support (fail if deps missing): `-DREQUIRE_LINUX_TRAY=ON`
+- Optional tray dependencies: one of `ayatana-appindicator-glib-devel` (preferred, no GTK3 needed), `ayatana-appindicator3-devel`, or `libappindicator-gtk3-devel` (the latter two also require `gtk3-devel`)
 - Fully functional for server operations and model management
 - Uses permissively licensed dependencies only (MIT, Apache 2.0, BSD, curl license)
 - Clean .deb package with only runtime files (no development headers)
@@ -595,19 +598,20 @@ A console application for terminal users:
 - Manages server lifecycle (start/stop persistent or ephemeral servers)
 - Communicates with `lemonade-router` via HTTP endpoints
 - Starts `lemonade-router` with appropriate options
-- Provides optional system tray interface via `serve` command
+- Provides optional system tray interface via `serve` command (Windows/macOS; on Linux the tray is provided by the separate `lemonade-tray` binary)
 
 **Command Types:**
 - **serve:** Starts a persistent server (with optional tray interface)
 - **run:** Starts persistent server, loads model, opens browser
 - **Other commands:** Use existing server or start ephemeral server, execute command via API, auto-cleanup
 
-#### lemonade-tray (GUI Launcher - Windows Only)
+#### lemonade-tray (GUI Tray Application - Windows and Linux)
 
-A minimal WIN32 GUI application for desktop users:
-- Simple launcher that starts `lemonade-server.exe serve`
+A GUI application for desktop users that exposes the server via a system tray icon:
+- **Windows:** Minimal launcher — finds `lemonade-server.exe`, launches it with the `serve` command, then exits. The server process owns the tray icon.
+- **Linux:** Tray application (requires GTK3 + AppIndicator3). Connects to an already-running server if one is found; otherwise starts one (via systemd if a unit is installed, or by spawning `lemonade-router` directly).
 - Zero console output or CLI interface
-- Used by Start Menu, Desktop shortcuts, and autostart
+- Used by application launchers, desktop shortcuts, and autostart entries
 - Provides seamless GUI experience for non-technical users
 
 ### Client-Server Communication
@@ -705,7 +709,7 @@ The `lemonade-server` executable is the command-line interface for terminal user
 # Run a model (starts persistent server with tray and opens browser)
 ./lemonade-server run Llama-3.2-1B-Instruct-CPU
 
-# Start persistent server (with tray on Windows/macOS, headless on Linux)
+# Start persistent server (with tray on Windows/macOS; always headless on Linux — use lemonade-tray for tray)
 ./lemonade-server serve
 
 # Start persistent server without tray (headless mode, explicit on all platforms)
@@ -727,24 +731,27 @@ The `lemonade-server` executable is the command-line interface for terminal user
 
 **Note:** `lemonade-router` is always launched with `--log-level debug` for optimal troubleshooting. Use `--log-level debug` on `lemonade-server` commands to see client-side debug output.
 
-### lemonade-tray.exe (GUI Tray Launcher - Windows Only)
+### lemonade-tray (GUI Tray Application - Windows and Linux)
 
-The `lemonade-tray` executable is a simple GUI launcher for desktop users:
-- Double-click from Start Menu or Desktop to start server
-- Automatically runs `lemonade-server.exe serve` in tray mode
-- Zero console windows or CLI interface
+The `lemonade-tray` executable provides a system tray icon for desktop users:
+- Double-click from Start Menu, application launcher, or Desktop to start server
+- Zero console windows or CLI interface — always starts the tray directly
 - Perfect for non-technical users
 - Single-instance protection: shows friendly message if already running
 
-**What it does:**
-1. Finds `lemonade-server.exe` in the same directory
-2. Launches it with the `serve` command
-3. Exits immediately (server continues running with tray icon)
+**Platform support:**
+- **Windows:** Always available; uses Win32 notification area APIs. Acts as a minimal launcher: finds `lemonade-server.exe` in the same directory, launches it with the `serve` command, then exits (the server process owns the tray icon).
+- **Linux:** Available when compiled with GTK3 + AppIndicator3 support (auto-detected at build time). Connects to an already-running server if one is found; otherwise starts one (via systemd if a unit is installed, or by spawning `lemonade-router` directly).
+
+**What it does (Linux):**
+1. Starts immediately in tray mode (no subcommand needed)
+2. Connects to an already-running server via the PID file, or starts one (via systemd if a unit is installed, otherwise spawns `lemonade-router` directly)
+3. Shows a system tray icon connected to the server
 
 **When to use:**
-- Launching from Start Menu
+- Launching from Start Menu (Windows) or application launcher (Linux)
 - Desktop shortcuts
-- Windows startup
+- Windows startup / Linux autostart
 - Any GUI/point-and-click scenario
 
 **System Tray Features (when running):**
--- a/docs/server/server_integration.md
+++ b/docs/server/server_integration.md
@@ -101,7 +101,7 @@ By default, the server runs on port 8000. Optionally, you can specify a custom p
 lemonade-server serve --port 8123
 ```
 
-You can also prevent the server from showing a system tray icon by using the `--no-tray` flag (Windows and macOS):
+You can also prevent the server from showing a system tray icon by using the `--no-tray` flag:
 
 ```bash
 lemonade-server serve --no-tray
--- a/setup.sh
+++ b/setup.sh
@@ -200,6 +200,127 @@ else
     fi
 fi
 
+# Check optional Linux tray dependencies (AppIndicator3 + libnotify [+ GTK3 if not using glib variant])
+# These are optional - lemonade-tray is only built when they are present.
+# lemonade-server always builds without them (headless, daemon-friendly).
+if [ "$OS" = "linux" ] && command_exists pkg-config; then
+    print_info "Checking optional Linux tray dependencies (AppIndicator3)..."
+    missing_tray_packages=()
+    appindicator_glib_found=false
+
+    # Check AppIndicator first — the glib variant is GTK-free and preferred.
+    if pkg-config --exists ayatana-appindicator-glib-0.1 2>/dev/null || \
+       pkg-config --exists ayatana-appindicator-glib 2>/dev/null; then
+        print_success "AppIndicator3 glib variant is installed (tray support, GTK-free)"
+        appindicator_glib_found=true
+    elif pkg-config --exists ayatana-appindicator3-0.1 2>/dev/null || \
+         pkg-config --exists appindicator3-0.1 2>/dev/null; then
+        print_success "AppIndicator3 is installed (tray support)"
+    else
+        print_warning "AppIndicator3 not found (optional, needed for lemonade-tray)"
+        if command_exists apt; then
+            missing_tray_packages+=("libayatana-appindicator3-dev")
+        elif command_exists pacman; then
+            missing_tray_packages+=("libayatana-appindicator")
+        elif command_exists dnf; then
+            missing_tray_packages+=("libayatana-appindicator-gtk3-devel")
+        fi
+    fi
+
+    # dbusmenu-glib is required alongside the glib appindicator variant so that
+    # GNOME Shell can find com.canonical.dbusmenu (it does not speak org.gtk.Menus).
+    if [ "$appindicator_glib_found" = true ]; then
+        if pkg-config --exists dbusmenu-glib-0.4 2>/dev/null; then
+            print_success "dbusmenu-glib is installed (GNOME Shell tray menu support)"
+        else
+            print_warning "dbusmenu-glib not found (optional, needed for tray menus on GNOME Shell)"
+            if command_exists apt; then
+                missing_tray_packages+=("libdbusmenu-glib-dev")
+            elif command_exists pacman; then
+                missing_tray_packages+=("libdbusmenu-glib")
+            elif command_exists dnf; then
+                missing_tray_packages+=("dbusmenu-glib-devel")
+            fi
+        fi
+    fi
+
+    # GTK3 is only required when NOT using the glib appindicator variant.
+    if [ "$appindicator_glib_found" = false ]; then
+        if pkg-config --exists gtk+-3.0 2>/dev/null; then
+            print_success "gtk3 is installed (tray support)"
+        else
+            print_warning "gtk3 not found (optional, needed for lemonade-tray)"
+            if command_exists apt; then
+                missing_tray_packages+=("libgtk-3-dev")
+            elif command_exists pacman; then
+                missing_tray_packages+=("gtk3")
+            elif command_exists dnf; then
+                missing_tray_packages+=("gtk3-devel")
+            fi
+        fi
+    fi
+
+    if pkg-config --exists libnotify 2>/dev/null; then
+        print_success "libnotify is installed (tray notifications)"
+    else
+        print_warning "libnotify not found (optional, enables tray notifications)"
+        if command_exists apt; then
+            missing_tray_packages+=("libnotify-dev")
+        elif command_exists pacman; then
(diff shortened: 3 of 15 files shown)


docs/server/server_integration.md
```diff
@@ -101,7 +101,7 @@ By default, the server runs on port 8000. Optionally, you can specify a custom p
 lemonade-server serve --port 8123
 ```
 
-You can also prevent the server from showing a system tray icon by using the `--no-tray` flag (Windows and macOS):
+You can also prevent the server from showing a system tray icon by using the `--no-tray` flag (Windows, macOS, and Linux):
```
superm1: maybe just drop the OS list as it's everything now

docs/dev-getting-started.md
superm1, line 1: I think you'll want to update setup.sh to install all relevant dependencies needed for this feature.

src/cpp/include/lemon_tray/platform/linux_tray.h
```diff
@@ -4,6 +4,20 @@
...
+#include <libayatana-appindicator/app-indicator.h>   // glib variant
+#elif defined(HAVE_AYATANA_APPINDICATOR)
+#include <libayatana-appindicator/app-indicator.h>   // GTK3 variant
+#else
+#include <libappindicator/app-indicator.h>           // upstream fallback
+#endif
+#ifdef HAVE_LIBNOTIFY
+#include <libnotify/notify.h>
+#endif
+#endif
```
superm1: with so many nested `#ifdef` can you please add comments to the `#endif` to disambiguate which goes with which?

src/cpp/tray/CMakeLists.txt
```diff
@@ -55,16 +55,71 @@ elseif(APPLE)
...
+
+    # Allow explicitly enabling/disabling tray support from command line.
+    # Default is OFF (auto-detect): tray is enabled if dependencies are found.
+    # Set -DENABLE_LINUX_TRAY=ON to require tray support (fail if deps missing).
+    option(ENABLE_LINUX_TRAY "Enable Linux system tray support (requires GTK3 + AppIndicator3)" OFF)
+
+    pkg_check_modules(GTK3 gtk+-3.0)
+
+    # Prefer the newer GLib variant (recommended successor, no GTK3 dep for the indicator itself).
+    # Fall back to the Ayatana GTK3 variant, then upstream libappindicator3.
```
superm1, line 71: Why have the GTK3 code in the first place?  GTK4 release 5 years ago.  Isn't GTK3 pretty much dead?

docs/dev-getting-started.md
```diff
@@ -591,12 +593,13 @@ A console application for terminal users:
...
 - **Other commands:** Use existing server or start ephemeral server, execute command via API, auto-cleanup
 
-#### lemonade-tray (GUI Launcher - Windows Only)
+#### lemonade-tray (GUI Tray Application - Windows and Linux)
 
-A minimal WIN32 GUI application for desktop users:
-- Simple launcher that starts `lemonade-server.exe serve`
+A GUI application for desktop users that exposes the server via a system tray icon:
+- **Windows:** Minimal launcher — finds `lemonade-server.exe`, launches it with the `serve` command, then exits. The server process owns the tray icon.
+- **Linux:** Full embedded application — starts the tray icon and server stack in-process (requires GTK3 + AppIndicator3). Runs directly without launching a separate binary.
```
superm1: could we also have it control the systemd service if present?

src/cpp/tray/CMakeLists.txt
abn: Iirc the glib variant is not available in all major distros yet.

docs/dev-getting-started.md
abn: I need to update the doc, I initially made the tray a "server + gtk" binary but figured it was overkill, now it only does the equivalent of `lemonade-server tray` and nops our if no server is running.

abn: Maybe in retrospect it would be good to have a server start/stop menu item. But I'm not sure how well it will play with the systemd unit.

src/cpp/tray/CMakeLists.txt
superm1: What distros is it missing in?

docs/dev-getting-started.md
superm1: There is always wrapping `pkexec systemctl stop lemonade-server` for something simple.

superm1: If I'm following right - you don't even need to wrap pkexec.

https://github.com/hardpixel/systemd-manager/blob/master/systemd-policies/org.freedesktop.policykit.pkexec.systemctl.policy

Just calling `systemctl stop lemonade-server.service` and `systemctl start lemonade-server.service` should show the pkexec popup on modern distros.

src/cpp/tray/CMakeLists.txt
abn: Fedora for example, which is odd because these things often land there early. And Ubuntu only started having it recently. 

I think https://repology.org/project/libayatana-appindicator-glib/packages is a good representation of who has it.

superm1: I'm quite surprised:I guess you're right and the fallback makes sense.

setup.sh
```diff
@@ -170,6 +170,98 @@ else
...
+    if pkg-config --exists ayatana-appindicator-glib-0.1 2>/dev/null || \
+       pkg-config --exists ayatana-appindicator3-0.1 2>/dev/null || \
+       pkg-config --exists appindicator3-0.1 2>/dev/null; then
+        print_success "AppIndicator3 is installed (tray support)"
+    else
+        print_warning "AppIndicator3 not found (optional, needed for lemonade-tray)"
+        if command_exists apt; then
+            missing_tray_packages+=("libayatana-appindicator3-dev")
+        elif command_exists pacman; then
+            missing_tray_packages+=("libayatana-appindicator")
```
superm1: maybe these should be reversed.  If you have ayatana, you don't need all the gtk right?

src/cpp/tray/tray_app.cpp
```diff
@@ -2521,6 +2841,62 @@ Menu TrayApp::create_menu() {
...
+        stop_server();
+        process_owns_server_ = false;
+        server_manager_.reset();
+        show_notification("Server Stopped", "Lemonade Server has been stopped");
+        build_menu();
+    }).detach();
+}
+
+void TrayApp::on_stop_external_server() {
+    int pid = external_server_pid_;
```
superm1: I guess this use is what is breaking Windows?

setup.sh
abn: If the glib flavour is available, I'd think so.

superm1: Yeah; I think re-order them then.

src/cpp/include/lemon_tray/platform/linux_tray.h
```diff
@@ -4,6 +4,26 @@
 
 #include "tray_interface.h"
 
+#ifdef HAVE_APPINDICATOR
+#ifdef HAVE_AYATANA_APPINDICATOR_GLIB
+#include <libayatana-appindicator-glib/ayatana-appindicator.h>  // glib variant (GTK-free, GIO only)
+#ifdef HAVE_DBUSMENU_GLIB
+#include <libdbusmenu-glib/server.h>
+#include <libdbusmenu-glib/menuitem.h>
+#endif
```
superm1: nit:
```suggestion
#endif //HAVE_DBUSMENU_GLIB
```

src/cpp/include/lemon_tray/platform/linux_tray.h
```diff
@@ -27,8 +47,26 @@ class LinuxTray : public TrayInterface {
...
+    GMenu* g_menu_;
+    GSimpleActionGroup* action_group_;
+    std::vector<std::function<void()>*> callbacks_;
+    void build_g_menu(const Menu& menu, GMenu* parent, GSimpleActionGroup* actions,
+                      int& action_id, std::vector<std::function<void()>*>& callbacks);
+#ifdef HAVE_DBUSMENU_GLIB
+    DbusmenuServer* dbusmenu_server_;
+    void build_dbusmenu(const Menu& menu, DbusmenuMenuitem* parent,
+                        std::vector<std::function<void()>*>& callbacks);
+#endif
```
superm1:
```suggestion
#endif // HAVE_DBUSMENU_GLIB
```

src/cpp/tray/platform/linux_tray.cpp
```diff
@@ -1,15 +1,521 @@
...
+
+            // g_menu_item_new_submenu() creates a pure submenu-parent item with no
+            // activatable action. Using g_menu_item_new() + g_menu_item_set_submenu()
+            // sets both "action" and "submenu" attributes, which causes the DBUSMENU
+            // serialiser in ayatana-appindicator-glib to discard the submenu link.
+            GMenuItem* menu_item = g_menu_item_new_submenu(item.text.c_str(),
+                                                           G_MENU_MODEL(submenu));
+            g_menu_append_item(section, menu_item);
+            g_object_unref(menu_item);
+            g_object_unref(submenu);
```
superm1: You should be using `g_autoptr()` instead of manually managing the ref count IMO.

src/cpp/tray/platform/linux_tray.cpp
```diff
@@ -1,15 +1,521 @@
...
+#endif
+}
+
+// ── build_g_menu (glib variant) ───────────────────────────────────────────────────
+
+#ifdef HAVE_AYATANA_APPINDICATOR_GLIB
+
+void LinuxTray::build_g_menu(const Menu& menu, GMenu* parent, GSimpleActionGroup* actions,
+                              int& action_id, std::vector<std::function<void()>*>& callbacks) {
+    GMenu* section = g_menu_new();
```
superm1, line 597: `g_autoptr()`?  Then you don't need so many cleanup paths.  It will manage clearing it when it goes out of scope for you.

superm1 (commented):
Having so many nested ifdefs is making this code really hard to follow for me.  I left a few comments, but I am wondering how you would feel about using polymorphism to create multiple "tray backends" for Linux instead.  Then you can have a clean segregation.  Things that are common to each implementation can be implemented in the parent class.  Things that are expected for each implementation to create can be virtual functions.

src/cpp/tray/tray_app.cpp
```diff
@@ -2605,7 +2857,79 @@ Menu TrayApp::create_menu() {
...
+                menu.add_item(MenuItem::Action("Start Service", [this, known_unit]() {
+                    std::thread([this, known_unit]() {
+                        if (systemd_control_service(known_unit, /*start=*/true)) {
+                            for (int i = 0; i < 10; ++i) {
+                                std::this_thread::sleep_for(std::chrono::seconds(1));
+                                auto [pid, port] = get_server_info();
+                                if (port != 0) {
+                                    server_manager_ = std::make_unique<ServerManager>(server_config_.host, port);
+                                    server_manager_->set_port(port);
+                                    server_config_.port = port;
```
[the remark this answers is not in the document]
abn: Fixed. The three assignments to `server_manager_`, `server_manager_->set_port()`, and `server_config_.port` are now wrapped in a `std::lock_guard<std::mutex> lock(state_mutex_)` block, consistent with the pattern used in `on_start_server()`, `on_stop_server()`, and `on_stop_external_server()`.

src/cpp/tray/CMakeLists.txt
```diff
@@ -107,21 +180,33 @@ if(WIN32)
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/../include
 )
 
 # Link httplib based on what's available (set by parent CMakeLists.txt)
 if(USE_SYSTEM_HTTPLIB)
-    target_link_libraries(lemonade-server PRIVATE cpp-httplib)
+    target_link_libraries(lemonade-server PRIVATE httplib::httplib)
 else()
     target_link_libraries(lemonade-server PRIVATE httplib::httplib)
 endif()
```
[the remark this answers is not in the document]
abn: Fixed. Collapsed the redundant if(`USE_SYSTEM_HTTPLIB`)/else block (both branches linked `httplib::httplib` identically) into a single unconditional `target_link_libraries(lemonade-server PRIVATE httplib::httplib)`. The meaningful `USE_SYSTEM_HTTPLIB` conditional for `HTTPLIB_INCLUDE_DIRS` is unchanged.

src/cpp/tray/CMakeLists.txt
```diff
@@ -55,16 +55,89 @@ elseif(APPLE)
...
+
+    if(ENABLE_LINUX_TRAY AND NOT (APP_INDICATOR_GLIB_FOUND OR (GTK3_FOUND AND APP_INDICATOR3_FOUND)))
+        message(FATAL_ERROR "ENABLE_LINUX_TRAY=ON but required dependencies not found. "
+            "Install one of: ayatana-appindicator-glib-devel (GTK-free), "
+            "or gtk3-devel + ayatana-appindicator3-devel / libappindicator-gtk3-devel.")
+    endif()
+
+    if(APP_INDICATOR_GLIB_FOUND OR (GTK3_FOUND AND APP_INDICATOR3_FOUND))
+        set(ENABLE_LINUX_TRAY ON)
+
```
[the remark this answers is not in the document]
abn: Fixed. Renamed to REQUIRE_LINUX_TRAY to make the "require or fail" intent explicit. Auto-detected availability is tracked in the internal _LINUX_TRAY_BUILD variable, which is never a CMake cache entry, so the user-facing option value is never overwritten by detection.

src/cpp/server/cli_parser.cpp
```diff
@@ -214,19 +218,22 @@ CLIParser::CLIParser()
...
-        if (argc == 1) {
-            throw CLI::CallForHelp();
-        }
+#ifdef LEMONADE_TRAY_GUI
+        // lemonade-tray is always a tray launcher — no subcommand selection needed
+        tray_config_.command = "tray";
+        should_continue_ = true;
+        exit_code_ = 0;
+        return 0;
 #endif
```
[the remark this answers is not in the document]
abn: This doesn't apply to the current implementation — app_.parse(argc, argv) is the first call in CLIParser::parse(), so --help, --version, and argument validation work normally. The LEMONADE_TRAY_GUI guard only affects how the subcommand name is resolved after parsing, defaulting to "tray" when none was given (which app_.require_subcommand(0, 1) allows).

src/cpp/BUILD_OPTIONS.md
```diff
@@ -1,5 +1,32 @@
 # Lemonade Build Options
 
+## Linux Tray Configuration
+
+### Build Options
+
+#### `ENABLE_LINUX_TRAY` (Default: **OFF** / auto-detect)
+Enable system tray support on Linux via GTK3 + AppIndicator3.
+
+- When **OFF** (default): Tray support is auto-detected at configure time. If GTK3 and one of the AppIndicator3 libraries are found, `lemonade-tray` is built as a separate executable. `lemonade-server` on Linux is always headless regardless. If dependencies are missing, only `lemonade-server` is built (headless mode).
```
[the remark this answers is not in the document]
abn: Fixed — updated to clarify that the preferred glib variant only requires GLib/GIO (no GTK3), and that GTK3 is only needed when falling back to the Ayatana or upstream AppIndicator3 variants.

src/cpp/tray/tray_app.cpp
```diff
@@ -2590,7 +2829,93 @@ Menu TrayApp::create_menu() {
...
+        bool server_reachable = false;
+        bool owns_server = false;
+        {
+            std::lock_guard<std::mutex> lock(state_mutex_);
+            if (server_manager_) {
+                try { server_manager_->get_health(); server_reachable = true; } catch (...) {}
+            }
+            owns_server = process_owns_server_;
+        }
+
```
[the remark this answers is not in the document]
abn: Fixed in the latest commit. create_menu() now takes (bool server_reachable, bool owns_server) as parameters. build_menu() reads process_owns_server_ under a brief, I/O-free lock and passes both values in. The redundant get_health() call and its enclosing state_mutex_ lock inside create_menu() are removed.

src/cpp/tray/tray_app.cpp
```diff
@@ -305,25 +335,115 @@ static const char* kSystemdUnitNames[] = {
...
+    return system(cmd.c_str()) == 0;
+}
+#else
+static bool systemd_control_service(const char* unit_name, bool start) {
+    if (!unit_name) return false;
+    std::string action = start ? "start" : "stop";
+    std::string cmd = "systemctl " + action + " " + std::string(unit_name) + " 2>/dev/null";
+    if (system(cmd.c_str()) == 0) return true;
+    // Fall back to pkexec (graphical polkit dialog for system services)
+    cmd = "pkexec systemctl " + action + " " + std::string(unit_name);
```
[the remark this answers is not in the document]
abn: leaving as-is for now. But if we really need to fix this, can,

src/cpp/tray/tray_app.cpp
```diff
@@ -2438,20 +2671,26 @@ void TrayApp::refresh_menu() {
...
+    // Snapshot last_menu_available_models_ here too so the comparison below is race-free.
+    decltype(last_menu_available_models_) last_available;
+    bool state_changed = false;
+    {
+        std::lock_guard<std::mutex> lock(state_mutex_);
+        auto [reachable, loaded] = fetch_server_state();
+        if (reachable != last_menu_server_reachable_) state_changed = true;
+        if (!state_changed && loaded != last_menu_loaded_models_) state_changed = true;
+        last_available = last_menu_available_models_;
     }
```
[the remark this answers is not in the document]
abn: The comment in the code explains the intent: the lock is needed so the reachability comparison against last_menu_server_reachable_ is race-free with build_menu(). Moving the call outside would require a different synchronization strategy (e.g. a separate snapshot variable) and the complexity isn't warranted here.

src/cpp/tray/tray_app.cpp
```diff
@@ -305,25 +335,115 @@ static const char* kSystemdUnitNames[] = {
...
+    if (unit.scope == SystemdBusScope::User) {
+        std::string cmd = "systemctl --user " + action + " " + unit.name + " 2>/dev/null";
+        return system(cmd.c_str()) == 0;
+    }
+    // System service: try direct, fall back to pkexec
+    std::string cmd = "systemctl " + action + " " + unit.name + " 2>/dev/null";
+    if (system(cmd.c_str()) == 0) return true;
+    cmd = "pkexec systemctl " + action + " " + unit.name;
+    return system(cmd.c_str()) == 0;
+}
```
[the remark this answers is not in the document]
abn: All arguments passed to system() come from kSystemdUnitNames, a compile-time constant array. There is no user-controlled input on this path. Switching to execvp/fork would be the right call if we ever pass dynamic input here, but that's not the case today.

jeremyfowers:
@superm1 does the code meet your expectations now?

src/cpp/tray/platform/linux_tray.cpp
```diff
@@ -1,15 +1,520 @@
...
+        NotifyNotification* n = notify_notification_new(d->title.c_str(), d->message.c_str(), nullptr);
+        notify_notification_set_timeout(n, 3000);
+        NotifyUrgency urgency = (d->type == NotificationType::ERROR)   ? NOTIFY_URGENCY_CRITICAL :
+                                (d->type == NotificationType::WARNING) ? NOTIFY_URGENCY_NORMAL   :
+                                                                         NOTIFY_URGENCY_LOW;
+        notify_notification_set_urgency(n, urgency);
+        notify_notification_show(n, nullptr);
+        g_object_unref(G_OBJECT(n));
+#else // !HAVE_LIBNOTIFY
+        std::cout << "[Notification] " << d->title << ": " << d->message << std::endl;
```
superm1: should be `LOG(DEBUG, "Notification")`

src/cpp/tray/platform/linux_tray.cpp
```diff
@@ -82,6 +587,9 @@ void LinuxTray::set_ready_callback(std::function<void()> callback) {
     ready_callback_ = callback;
 }
 
+#endif // HAVE_APPINDICATOR
+
+
 } // namespace lemon_tray
```
superm1, line 922: nit; why so much whitespace

src/cpp/tray/tray_app.cpp
```diff
@@ -119,9 +120,17 @@ static nlohmann::json build_launch_recipe_options(const lemon::TrayConfig& tray_
...
+enum class SystemdBusScope { System, User };
+
+struct SystemdUnitInfo {
+    std::string name;
+    SystemdBusScope scope = SystemdBusScope::System;
+    explicit operator bool() const { return !name.empty(); }
+};
+
+// Internal: check if a unit is active on a specific D-Bus (system or user session)
+static bool is_systemd_service_active_on_bus(const char* unit_name, SystemdBusScope scope) {
```
superm1: as a general comment - I feel like all this Linux specific tray handling code should now be moved into `linux_tray.cpp`.  We shouldn't be having a ton of code that is Linux specific in "tray_app.cpp"

src/cpp/tray/tray_app.cpp
```diff
@@ -883,27 +1033,87 @@ int TrayApp::run() {
...
+        // Then XDG_DATA_DIRS (system-wide locations)
+        const char* xdg_data_dirs = getenv("XDG_DATA_DIRS");
+        if (xdg_data_dirs && xdg_data_dirs[0]) {
+            std::istringstream ss(xdg_data_dirs);
+            std::string d;
+            while (std::getline(ss, d, ':')) {
+                if (!d.empty()) data_dirs.push_back(d);
+            }
+        } else {
+            data_dirs.push_back("/usr/local/share");
```
superm1, line 713: since we install in `/opt` by default for Debian packages today (although that's changin) should `/opt` be in this list?

src/cpp/BUILD_OPTIONS.md
```diff
@@ -1,5 +1,31 @@
...
+Enable system tray support on Linux via AppIndicator3 (GTK3 not required when using the GLib variant).
+
+- When **OFF** (default): Tray support is auto-detected at configure time. If AppIndicator3 libraries are found, `lemonade-tray` is built as a separate executable. `lemonade-server` on Linux is always headless regardless. If dependencies are missing, only `lemonade-server` is built (headless mode).
+- When **ON**: Tray support is required — the build will fail if the dependencies are not found.
+
+Optional runtime dependencies (for tray support):
+- One of (preferred first):
+  - `ayatana-appindicator-glib-devel` (recommended, GTK-free — only GLib/GIO required)
+  - `ayatana-appindicator3-devel` + `gtk3-devel` (Ayatana GTK3 variant)
+  - `libappindicator-gtk3-devel` + `gtk3-devel` (upstream libappindicator3)
```
superm1, line 17: shouldn't you be mentioning libnotify here too?

src/cpp/tray/tray_app.cpp
```diff
@@ -735,28 +508,60 @@ int TrayApp::run() {
...
+                // Server is already running — connect to it
+                // Track PID only for non-systemd processes (systemd ones are tracked by unit name)
+                if (!is_any_systemd_service_active()) {
+                    external_server_pid_ = pid;
+                }
+                server_config_.host = normalize_connect_host(server_config_.host);
+                server_manager_ = std::make_unique<ServerManager>(server_config_.host, running_port);
+                server_manager_->set_port(running_port);
+                server_config_.port = running_port;
+                std::cout << "Connected to Lemonade Server on port " << running_port << std::endl;
```
superm1, line 525: `LOG()`

src/cpp/tray/tray_app.cpp
```diff
@@ -735,28 +508,60 @@ int TrayApp::run() {
...
+#ifdef HAVE_SYSTEMD
+                auto unit = get_first_known_systemd_unit();
+                if (unit) {
+                    std::cout << "Starting server via systemd: " << unit.name << std::endl;
+                    if (systemd_control_service(unit, /*start=*/true)) {
+                        // Wait up to 10 seconds for server to appear
+                        for (int i = 0; i < 10 && !started; ++i) {
+                            std::this_thread::sleep_for(std::chrono::seconds(1));
+                            auto [spid, sport] = get_server_info();
+                            if (sport != 0) {
```
superm1, line 539: this is still a lot of linux specified systemd code in tray_app.cpp.  Why can't all of this be in a Linux implementation of a run function?

jeremyfowers (approved):
@superm1 sez to merge it

PR #1301, opened 2026-03-05
Draw SVG images without lucide-react

superm1:
The extra dependency has two issues:

1. It makes landing in Debian and Ubuntu much harder
2. Adds extra size to the JS for what is just some glyphs

### DIFF
--- a/src/web-app/package.json
+++ b/src/web-app/package.json
@@ -32,7 +32,6 @@
     "ts-loader": "^9.5.4",
     "typescript": "^5.3.3",
     "webpack": "^5.102.1",
-    "lucide-react": "^0.575.0",
     "webpack-cli": "^6.0.1"
   }
 }
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -1,5 +1,25 @@
 PORTIONS LICENSED AS FOLLOWS
 
+## Lucide Icons
+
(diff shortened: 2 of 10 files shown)


src/web-app/package.json
```diff
@@ -32,7 +32,7 @@
     "ts-loader": "^9.5.4",
     "typescript": "^5.3.3",
     "webpack": "^5.102.1",
-    "lucide-react": "^0.575.0",
```
jeremyfowers, line 35: shouldnt it be removed from src/app/package.json too?

jeremyfowers:
Before:

image


After:

image

Me:

image

src/web-app/package.json
superm1: Yes.  Will remove it from there too.

jeremyfowers:
Lucide is OSS https://github.com/lucide-icons/lucide?tab=License-1-ov-file#readme

What about just copying the exact SVGs and mentioning in notice.md?

superm1:
> Lucide is OSS https://github.com/lucide-icons/lucide?tab=License-1-ov-file#readme
> 
> What about just copying the exact SVGs and mentioning in notice.md?

Yeah will just do that, thanks for suggestion.

superm1:
Hopefully I got it right this time.

superm1:
Updated images should match the original now more closely.
image

PR #1305, opened 2026-03-05
Disable microphone when running on insecure Window

superm1:
If connecting to a machine via IP and HTTP instead of localhost the web browser will not allow use of local media devices for security reasons.

Link: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaDevices

### DIFF
--- a/src/app/src/renderer/components/RecordButton.tsx
+++ b/src/app/src/renderer/components/RecordButton.tsx
@@ -19,6 +19,12 @@ const RecordButton: React.FC<RecordButtonProps> = (props) => {
   const { activeModel, isRecording, start, stop } =
     useVoiceTranscription(props);
 
+  // Hide microphone button when not in a secure context (HTTPS or localhost)
+  // getUserMedia() requires a secure context to function
+  if (typeof window !== 'undefined' && !window.isSecureContext) {
+    return null;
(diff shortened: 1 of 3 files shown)


PhilWheat:
Should there be a way to disable this disabling?
I'm thinking about my configuration where I use multiple boxes on my LAN but don't have certificates for them to do SSL because they are internal only and not exposed externally.
This seems to indicate I would be blocked from using a microphone across my lan unless I set up an SSL infrastructure internally.  (I'm not considering doing this now and if I was using audio streaming, I'd likely use a different client, but I just wanted to point out the scenario.)

superm1:
> Should there be a way to disable this disabling?
> I'm thinking about my configuration where I use multiple boxes on my LAN but don't have certificates for them to do SSL because they are internal only and not exposed externally.
> This seems to indicate I would be blocked from using a microphone across my lan unless I set up an SSL infrastructure internally.  (I'm not considering doing this now and if I was using audio streaming, I'd likely use a different client, but I just wanted to point out the scenario.)

It doesn't work - browsers block it...

ramkrishna2910:
@claude review

ramkrishna2910 (approved):

PR #1306, opened 2026-03-05
Fix the copy-to-clipboard button in electron

jeremyfowers:
Turns out the copy-to-clipboard button in the backend manager was only working in the web app.


jeremyfowers:
annnnd this breaks the web app... fixing...

superm1 (approved):

superm1:
> annnnd this breaks the web app... fixing...

lol ofc.

jeremyfowers:
fixed

PR #1307, opened 2026-03-05
Add "launch" subcommand to launch coding agents

sawansri:
Implements `lemonade-server launch` for easy integration of lemonade with coding agents like Claude Code and Codex.

### Clarity

 - [ ] Does the pr description and title explain the change fully?
 - [ ] Are UI/UX changes displayed with screenshots/videos/terminal blocks?

### Completeness

- [x] Are changes documented in markdown and help menus (if appropriate)?
    - For example: cli.md, spec.md, dev-getting-started.md, `lemonade-server --help`
- [x] Do new endpoints have example requests/scripts so others can reproduce?

### UI/UX review

Are the UI/UX changes (if any)... 
- [ ] visually cohesive with the rest of the experience?
- [ ] a cohesive part of existing workflows, or do they introduce a new workflow (justify)?
- [ ] discoverable by new users?
- [ ] usable without documentation, or is new documentation needed (justify)?
- [x] tested on both Windows on Linux (if needed)?

Note: UI/UX includes installers, app, tray, and CLI.

### Code quality

- [x] Is the change maintainable (does it introduce redundancy)?
- [x] Are there any security or licensing issues?

### DIFF
--- a/src/cpp/server/utils/process_manager.cpp
+++ b/src/cpp/server/utils/process_manager.cpp
@@ -138,12 +138,18 @@ static std::vector<char> build_windows_environment_block(
     const std::vector<std::pair<std::string, std::string>>& env_vars) {
     std::vector<std::string> merged_entries;
 
-    LPCH environment = GetEnvironmentStringsA();
+    LPWCH environment = GetEnvironmentStringsW();
     if (environment) {
-        for (const char* entry = environment; *entry != '\0'; entry += std::strlen(entry) + 1) {
-            merged_entries.emplace_back(entry);
+        for (const wchar_t* entry = environment; *entry != L'\0';
+             entry += std::wcslen(entry) + 1) {
+            int size_needed = WideCharToMultiByte(CP_UTF8, 0, entry, -1, nullptr, 0, nullptr, nullptr);
+            if (size_needed > 0) {
+                std::string narrow(size_needed - 1, '\0');
+                WideCharToMultiByte(CP_UTF8, 0, entry, -1, &narrow[0], size_needed, nullptr, nullptr);
+                merged_entries.emplace_back(std::move(narrow));
+            }
         }
-        FreeEnvironmentStringsA(environment);
+        FreeEnvironmentStringsW(environment);
     }
 
     for (const auto& env : env_vars) {


superm1 (commented):
General comment - I am wondering if we actually want a compatibility symlink for `/usr/bin/lemonade` and `/usr/bin/lemonade-server` and then in docs tell people to use `lemonade` instead.

src/cpp/tray/tray_app.cpp
```diff
@@ -1646,6 +1651,92 @@ int TrayApp::execute_status_command() {
...
+        try {
+            auto load_manager = std::make_unique<ServerManager>(load_host, load_port);
+            load_manager->load_model(load_model, nlohmann::json::object(), false);
+        } catch (...) {
+            // Silently ignore — the agent TUI owns stdout/stderr now.
+        }
+    }).detach();
+
+    std::cout << "Loading model in background: " << tray_config_.launch_model << std::endl;
+    std::cout << "Launching " << tray_config_.launch_agent << "..." << std::endl;
```
superm1: should this be `LOG()`?  If intentionally not, no problem

src/cpp/tray/tray_app.cpp
```diff
@@ -1646,6 +1651,92 @@ int TrayApp::execute_status_command() {
...
+        return 1;
+    }
+
+    httplib::Client cli(host, port);
+    cli.set_connection_timeout(1);
+    cli.set_read_timeout(1);
+    auto health = cli.Get("/api/version");
+    if (!health) {
+        std::cerr << "Error: Lemonade server is not reachable at http://" << host << ":" << port << "." << std::endl;
+        std::cerr << "Start the server first with: lemonade-server serve --no-tray" << std::endl;
```
superm1: should this be `LOG()`?  If intentionally not, no problem

src/cpp/tray/tray_app.cpp
```diff
@@ -1646,6 +1651,92 @@ int TrayApp::execute_status_command() {
...
+        }
+    }
+
+    std::string host = server_config_.host;
+    if (host.empty() || host == "0.0.0.0") {
+        host = "127.0.0.1";
+    }
+    if (host == "localhost") {
+        host = "127.0.0.1";
+    }
```
superm1: can you check the codebase for another use like this?  I feel like I've seen this elsewhere.  If it's still around, split it out to a helper.

src/cpp/server/cli_parser.cpp
```diff
@@ -190,6 +190,23 @@ CLIParser::CLIParser()
     recipes->add_option("--uninstall", tray_config_.uninstall_backend,
                         "Uninstall a backend (format: recipe:backend)");
 
+    // Launch
+    CLI::App* launch = app_.add_subcommand("launch", "Launch a coding agent connected to a running Lemonade server");
```
superm1, line 194: probably should have some docs that go with this change

sawansri:
> General comment - I am wondering if we actually want a compatibility symlink for `/usr/bin/lemonade` and `/usr/bin/lemonade-server` and then in docs tell people to use `lemonade` instead.

I think that makes sense. In my mind, launch is less of a "server" feature and fits better with the "lemonade" binary.

src/cpp/tray/tray_app.cpp
sawansri: Good catch!

src/cpp/tray/tray_app.cpp
sawansri: Yup, should use `LOG`

src/cpp/tray/tray_app.cpp
sawansri: will do

superm1 (approved):

sawansri:
Hey guys, something I want to get thoughts on is when the model should be loaded in llama-server. The current implementation loads the model asynchronously while the agent is launched (versus an earlier implementation when it was blocking agent startup). Is this a good approach?

superm1:
I guess what happens if the model load isn't done by the time the agent starts up? Race? Chaos? If so synchronous is the right way.

sawansri:
> I guess what happens if the model load isn't done by the time the agent starts up? Race? Chaos? If so synchronous is the right way.

Based on the testing I did, seems like the processing of the query is just delayed until the model is finished loading. Imo, async leads to a smoother experience since the user can potentially start typing out their prompt while the model is simultaneously loading in the backend.

superm1:
> > I guess what happens if the model load isn't done by the time the agent starts up? Race? Chaos? If so synchronous is the right way.
> 
> Based on the testing I did, seems like the processing of the query is just delayed until the model is finished loading. Imo, async leads to a smoother experience since the user can potentially start typing out their prompt while the model is simultaneously loading in the backend.

Yeah I tend to agree with you from that description. People expect the first query is slow anyways.

sawansri:
Failing test seems to be related to a missing windows import (`processenv.h` which is properly located in the header guards). 
image

jeremyfowers:
> General comment - I am wondering if we actually want a compatibility symlink for `/usr/bin/lemonade` and `/usr/bin/lemonade-server` and then in docs tell people to use `lemonade` instead.

@bitgamma is looking into this!

superm1:
@jeremyfowers can you help figure out the Windows issue?  If we can I'd like to see this in 10.0.0!

docs/server/lemonade-server-cli.md
```diff
@@ -177,6 +179,49 @@ lemonade-server pull user.nomic-embed \
 
 For more information about model formats and recipes, see the [API documentation](../lemonade_api.md) and the [server models guide](https://lemonade-server.ai/models.html). For details on the underlying JSON files (`user_models.json` and `recipe_options.json`), see the [Custom Model Configuration Guide](./custom-models.md).
 
+## Options for launch
```
jeremyfowers, line 182: Image

What is the correct answer to this question? Can you document it, or better yet solve this programmatically by either:
1. printing a big hint message in the terminal right before launching claude
2. setting things up so claude doesnt ask the question at all

jeremyfowers: I answered "No recommended) and it seems to be working!

jeremyfowers:
@claude review

docs/server/lemonade-server-cli.md
sawansri: It's possible to remove `ANTHROPIC_API_KEY` as an env var entirely (functionality works on my end without it set) but several guides online like [unsloth's](https://unsloth.ai/docs/basics/claude-code) suggest keeping it.
image

It might be fine to hint the user to select "No" here.

jeremyfowers (approved):
Image

Works great for me! Agreed to two followups. Thanks @sawansri!

PR #1311, opened 2026-03-06
fix(tray): use get_environment_variable_utf8 for HF cache paths

### LINKED ISSUE #1277: Apply windows path fix to tray_app.cpp
After #1232 merges, apply the new helper function here as well:

```
  1. Missed duplicate: tray_app.cpp:1308-1322 (Medium)

  There's a duplicate copy of the HF cache directory lookup in tray_app.cpp that still uses std::getenv:

  // tray_app.cpp:1306-1322 — NOT updated by this PR
  if (const char* env = std::getenv("HF_HUB_CACHE")) {
      hf_cache = env;
  } else if (const char* env = std::getenv("HF_HOME")) {
      ...
  } else {
      if (const char* userprofile = std::getenv("USERPROFILE")) {
```

fyi @szkocot

kpoineal:
## Summary

Replaces \std::getenv\ calls in \	ray_app.cpp\ with \lemon::utils::get_environment_variable_utf8()\ for proper Windows Unicode path support when importing local models.

## Problem

The HF cache directory lookup in \	ray_app.cpp\ (lines ~1319-1335) was still using raw \std::getenv\ calls, which don't handle Unicode paths on Windows. PR #1232 introduced \get_environment_variable_utf8()\ using the Windows wide-string API (\GetEnvironmentVariableW\) but missed this duplicate in the tray app.

## Changes

- Replaced 3 \std::getenv\ calls (\HF_HUB_CACHE\, \HF_HOME\, \USERPROFILE\/\HOME\) with \lemon::utils::get_environment_variable_utf8()\
- Added proper fallback defaults (\C:\\.cache\\huggingface\\hub\ on Windows, \/tmp/.cache/huggingface/hub\ on Linux/macOS) that were previously missing
- Matches the pattern established in \ModelManager::get_hf_cache_dir()\

## Testing

- No new \std::getenv\ calls remain for path-sensitive env vars in \	ray_app.cpp\
- The two remaining \std::getenv\ calls (\XDG_CONFIG_HOME\, \LEMONADE_SYSTEM_SERVICE\) are Linux-only systemd config flags, not filesystem paths

Closes #1277


ramkrishna2910:
@claude review

kpoineal:
> Please take a look at the comments and fix the code duplication logic.

Fixed. Please review again :D

PR #1312, opened 2026-03-06
Per-backend log filtering

danielholanda:
# Description

Backend subprocess logs are now tagged with their source name (llama-server, whisper-server, etc.) instead of a generic Process tag

## Details
* New ?source= query param on /logs/stream lets you filter the SSE stream to a single backend
* New /logs/sources endpoint returns what backends are currently loaded
* Log viewer UI gets a dropdown to pick "All Sources" or a specific backend
image


src/cpp/server/utils/process_manager.cpp
```diff
@@ -55,15 +56,15 @@ static bool is_error_line(const std::string& line) {
     return lowered.find("error") != std::string::npos;
 }
 
-static void log_process_line(const std::string& line) {
+static void log_process_line(const std::string& line, const std::string& source = "Process") {
```
superm1: I don't really like hardcoded default arguments; this can lead to subtle difficult to find bugs. Can you instead make sure that any caller that should be getting `Process` sends it?

superm1 (approved):

src/cpp/server/utils/process_manager.cpp
danielholanda: Done! All call sites already pass the second argument explicitly,  so no other changes are needed. The default is now gone.

danielholanda:
Waiting a few days to see if @jeremyfowers  also wants to review this

jeremyfowers (changes requested):
Please resolve conflicts and then yes I'd like to test drive it before merging

danielholanda:
Resolved conflicts. @jeremyfowers Please take a look.
There are clearly some tradeoffs here and I would like to know what you think.

src/app/src/renderer/LogsWindow.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
```
jeremyfowers, line 1: Image

What's up with all the `WebSocketServer::handleConnection() HTTP status: 400 error: Missing Sec-WebSocket-Key value`? Never seen that before.

src/app/src/renderer/LogsWindow.tsx
```diff
@@ -1,4 +1,4 @@
...
 import { getAPIKey, getServerBaseUrl, onServerUrlChange, serverConfig } from './utils/serverConfig';
 import {EventSource} from 'eventsource';
```
jeremyfowers, line 3: Doesn't seem to work for Kokoro, SD, Whisper, or RyzenAI-Server?

Image

Image

The KokoroServer logs are ending up under Lemonade Router not Kokoro Server.

src/app/src/renderer/LogsWindow.tsx
```diff
@@ -7,6 +7,11 @@ interface LogsWindowProps {
   height?: number;
 }
```
jeremyfowers, line 8: Image

Two instances of llama-server (under `--max-loaded-models 2`) share the same log view. Should it be clearly separated and labeled into `llama-server[0]` and `[1]`? Or better yet, filter by model name instead of process name?

jeremyfowers (changes requested):
Sorry about the very long wait!

I like this PR conceptually @danielholanda but there seems to be a few bugs. The overall design is sound and I'd like this to be an app feature.

danielholanda:
Awesome! Thanks for the review. Will take a second stab at it in the next few days

jeremyfowers:
Setting to draft until the requested changes are made.

danielholanda:
Closing as this is not stale. That said, I would still love to see this feature here in the future.

PR #1317, opened 2026-03-06
Consolidate "Add a model" and "Import a model" into a single entry point

danielholanda:
# Description

The Model Manager footer had two separate buttons: “Add a model” and “Import a model.” The names are very similar, making it difficult to immediately understand the difference between them. This should be clear and require no explanation.

This also wasted valuable vertical space in a sidebar panel where every pixel counts: two full-width stacked buttons for what is fundamentally one action with two input methods.

## Fix
* Single "Add a model" button replaces both. One entry point, no ambiguity.
* JSON import lives inside the form as a subtle inline hint ("Have a config file? Import from JSON") 
* Reclaimed lost vertical space

## Before
image

## After
image
image


jeremyfowers:
@meghsat has taken over solving this in a different branch, closing.

PR #1323, opened 2026-03-07
Move network beacon to 8080

superm1:


superm1:
This only makes sense to unify ports - we should match whatever port we move web service to by default

PR #1327, opened 2026-03-08
Add query parameter handling to enable firefox chatbot integration

Rebreda:
Firefox (and probably other software/services) have a simple way to pass prompts back from their software to LLM providers using a `/?q=<prompt>` URL query parameter. Kept it as simple as possible, but in theory the `q` parameter name could be customizable etc. 

I noticed there's a flash when first loading lemonade as a chat provider/on-first-load and when handlign the query. I think there's a render issue deeper on the app side that's causing the app to completely re-render in some suboptimal way (just speculating) but that's not related to this implementation. 

https://github.com/user-attachments/assets/5111e99f-dea2-4d55-a48c-1fb6edfd40e8


jeremyfowers:
Building the installer to test with.

PR #1334, opened 2026-03-10
Add a user guide for Claude Code integration

### LINKED ISSUE #1330: Launch command followup: user guide
(Take https://github.com/lemonade-sdk/lemonade/issues/1331 into account)

Please add a guide for `launch` under server apps documentation: https://github.com/lemonade-sdk/lemonade/tree/main/docs/server/apps

This guide should cover:
1. Setting up the pre-reqs: installing claude code, downloading models, starting the server, etc.
2. Best practices in terms of llamacpp-args, model choice, etc. assuming the user has a Strix Halo 128 GB
3. Best practices in terms of what coding tasks are achievable vs. not.

sawansri:
Resolves #1330, follow up PR to #1307

### DIFF
--- a/docs/server/apps/claude-code.md
+++ b/docs/server/apps/claude-code.md
@@ -1,147 +1,87 @@
 # Claude Code
 
-Claude Code is a high-performance agentic coding CLI from Anthropic that can reason about your codebase, execute commands, and edit files. While it is natively designed for Anthropic's hosted models, **Lemonade Server** allows you to use Claude Code with **local models** by emulating the Anthropic API.
+Claude Code is Anthropic's coding agent CLI. With Lemonade, you can run Claude Code against local models through Lemonade's Anthropic-compatible API.
 
-This setup provides a private, offline-capable coding assistant that lives in your terminal and has full access to your local development environment.
+This guide focuses on the most common launch flows.
 
 ## Prerequisites
 
-### 1. Install Claude Code
-Claude Code can be installed via npm or using the official installation script:
+1. Install Claude Code:
 
-**Using curl:**
 ```bash
 curl -fsSL https://claude.ai/install.sh | bash
 ```
 
-**Using npm:**
+or:
+
 ```bash
 npm install -g @anthropic-ai/claude-code
 ```
 
-### 2. Lemonade Server
-Ensure you have Lemonade Server installed. If you haven't set it up yet, refer to the [Getting Started guide](https://lemonade-server.ai/install_options.html).
+2. Make sure Lemonade Server is running (`lemond`).
 
-### 3. Download a Coding Model
-Claude Code requires a model with strong instruction-following and tool-use capabilities. We recommend **Qwen3.5-35B-A3B-GGUF** or **GLM-4.7-Flash-GGUF**. These models require systems with 64 GB of RAM or more.
+## Launch Claude Code with Lemonade
 
-```bash
-# Recommended for most coding tasks
-lemonade-server pull Qwen3.5-35B-A3B-GGUF
-
-# High-performance alternative
-lemonade-server pull GLM-4.7-Flash-GGUF
-```
-
-## Launching Claude Code
-
-Lemonade provides a specialized `launch` command that streamlines the connection between the Claude Code CLI and your local server. It handles all necessary environment variables, including API redirection and performance optimizations.
-
-### Step 1: Start Lemonade Server
-In one terminal window, ensure the server is running:
-```bash
-lemonade-server serve --ctx-size 32768
-```
-
-We recommend starting the server with a context window size starting at 32768 tokens to accomodate for Claude Code's system prompt (20k+ tokens). Note that you might need to change this value depending on your hardware and project size.
-
-### Step 2: Launch the Agent
-Navigate to your project directory in another terminal and run:
-```bash
-lemonade-server launch claude -m Qwen3.5-35B-A3B-GGUF
-```
-
-**What happens under the hood?**
-When you execute the `launch` command, Lemonade Server initiates a **concurrent load** of the specified model on the backend. This means the server starts loading the model into memory in a background thread, allowing the Claude Code CLI interface to start instantly without blocking.
-
-Additionally, the command automatically configures several critical environment variables:
-- `ANTHROPIC_BASE_URL`: Redirects Claude Code to your local Lemonade instance.
-- `ANTHROPIC_API_KEY`: Sets a local placeholder key.
-- `ANTHROPIC_AUTH_TOKEN`: Sets a local placeholder token.
-- `CLAUDE_CODE_ATTRIBUTION_HEADER`: Set to `0` to prevent KV cache invalidation, ensuring maximum inference speed.
-- `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`: Set to `1` to minimize telemetry and unnecessary external calls.
-
-## Performance Expectations
-
-### Concurrent Loading and Initial Delay
-Because the model is loaded concurrently on the backend, if you send a query immediately after launching the CLI, you may experience an initial delay while the model finishes loading into memory.
-
-### System Prompt Processing
-The first query you send to Claude Code will take longer than subsequent ones, typically **30-40 seconds on a Strix Halo system**, though this varies based on your hardware and model selected.
-
-This delay occurs because Claude Code sends a massive system prompt (20,000+ tokens) that defines its agentic behavior and tool-use capabilities. Once Lemonade processes this prompt, it is cached in the KV (Key-Value) cache, and subsequent queries will respond significantly faster as long as the session remains active.
-
-## Best Practices for Strix Halo (128GB)
-
-If you are using a **Strix Halo** system with **128GB of RAM**, you have a top-tier environment for local AI! You can run larger models with significantly expanded context windows.
-
-### Recommended Model Choices - MoE models with strong agentic capabilities
-*   **Qwen3.5-35B-A3B-GGUF**: A Mixture of Experts (MoE) model that excels at agentic tasks. It has 35B parameters, 3B of which are active at a time.
-*   **Qwen3-Coder-Next**: An MoE model designed specifically for coding agents and local development with 80B total parameters, 3B of which are activated
-*   **GLM-4.7-Flash-GGUF**: An excellent alternative for rapid iterations and complex instruction following, a 30B-A3B MoE model
-
-### Custom Tuning with `llamacpp-args`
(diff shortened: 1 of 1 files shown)


docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+curl -fsSL https://claude.ai/install.sh | bash
+```
+
+**Using npm:**
+```bash
+npm install -g @anthropic-ai/claude-code
+```
+
+### 2. Lemonade Server
+Ensure you have Lemonade Server installed. If you haven't set it up yet, refer to the [Getting Started guide](../../dev-getting-started.md).
```
jeremyfowers:
```suggestion
Ensure you have Lemonade Server installed. If you haven't set it up yet, refer to the [Getting Started guide](https://lemonade-server.ai/install_options.html).
```

For installation-only workflows (no dev needed)

docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+
+### 3. Download a Coding Model
+Claude Code requires a model with strong instruction-following and tool-use capabilities. We recommend **Qwen3.5-35B-A3B-GGUF** or **GLM-4.7-Flash-GGUF**.
```
jeremyfowers: Mention that these models require systems with 64 GB or more RAM.

docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+We recommend starting the server with a context window size starting at 32768 tokens to accomodate for Claude Code's system prompt (20k+ tokens). Note that you might need to change this value depending on your hardware and project size.
+
+### Step 2: Launch the Agent
+Navigate to your project directory in another terminal and run:
+```bash
+lemonade-server launch claude -m Qwen3.5-35B-A3B-GGUF
+```
+
+**What happens under the hood?**
+When you execute the `launch` command, Lemonade Server initiates a **concurrent load** of the specified model on the backend. This means the server starts loading the model into memory in a background thread, allowing the Claude Code CLI interface to start instantly without blocking.
```
jeremyfowers: Brainstorming here...

### Option 1

Would this workflow be simpler if we used `lemonade-server run Qwen3.5-35B-A3B-GGUF --ctx-size 32768` ?

That way:
- No need for a separate pull command
- No need to explain "Lemonade Server initiates a **concurrent load**" since `run` makes the load intentional.
- Downside: right now `run` always pops open the Lemonade app, which is not desirable here.

### Option 2

Should `lemonade-server launch claude -m Qwen3.5-35B-A3B-GGUF`:
1. Start the lemonade server, pull, and load the model (just like `run` does)
2. Accept any argument that `run` does: that way `--ctx-size`, `--llamapp-args`, etc. all come from the same place. (note: there is helper code already for generating the CLI for `run`-like commands) 

Then the whole thing is one command.

docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+
+### 2. Lemonade Server
+Ensure you have Lemonade Server installed. If you haven't set it up yet, refer to the [Getting Started guide](../../dev-getting-started.md).
+
+### 3. Download a Coding Model
+Claude Code requires a model with strong instruction-following and tool-use capabilities. We recommend **Qwen3.5-35B-A3B-GGUF** or **GLM-4.7-Flash-GGUF**.
+
+```bash
+# Recommended for most coding tasks
+lemonade-server pull Qwen3.5-35B-A3B-GGUF
```
jeremyfowers: Does this model work well on its own, or is it best to use the ThinkingCoder.json recipe? That will impact whether we need to introduce the concept of custom model recipes in this guide, and how they should interact with the `launch` command.

docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+We recommend starting the server with a context window size starting at 32768 tokens to accomodate for Claude Code's system prompt (20k+ tokens). Note that you might need to change this value depending on your hardware and project size.
+
+### Step 2: Launch the Agent
+Navigate to your project directory in another terminal and run:
+```bash
+lemonade-server launch claude -m Qwen3.5-35B-A3B-GGUF
+```
+
```
jeremyfowers, line 40: This would be a good spot for a screenshot. We can upload screenshots to https://github.com/lemonade-sdk/assets to avoid bloating this repo (see how that is used in the open webui guide in this same folder).

docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+- `-b` and `-ub`: These control the batch size and physical batch size. While the llama.cpp default is `512`, increasing these (e.g., to `1024` or `2048`) helps saturate memory bandwidth on high-end hardware like Strix Halo, though it increases RAM/VRAM consumption at the same time.
+- `-fa on`: Enables Flash Attention for optimized performance
+
+### Using Model Recipes
+While you can manually pass arguments with `--llamacpp-args`, a more scalable approach is to use the model's saved configuration by passing `--use-recipe`.
+
+```bash
+lemonade-server launch claude -m Qwen3.5-35B-A3B-GGUF --use-recipe
```
jeremyfowers: Here's my quandry with the workflow: [Qwen3.5-35B-A3B-ThinkingCoder.json](https://github.com/user-attachments/files/25752877/Qwen3.5-35B-A3B-ThinkingCoder.json) uses some different settings than what you recommend above. 

I need a way to combine your recommendations with the general Qwen3.5 Thinking suggestions from that recipe.

### Option 1

The simplest way would be to populate the Lemonade Wiki with some specific recipe.json files for Claude Code. That way it is easy to use those as a starting point with --use-recipe. In that case, this guide should reference those recipe files and not the base models like `Qwen3.5-35B-A3B-GGUF` throughout.

### Option 2

`lemonade-server launch` itself needs a way to programmatically do graceful blends of Claude Code specific args with generic args from recipes like `Qwen3.5-35B-A3B-ThinkingCoder.json`.

But Option 1 seems way easier.

wdyt?

docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+  - **Linux/macOS:** `~/.cache/lemonade/recipe_options.json`
+  - **Windows:** `%USERPROFILE%\.cache\lemonade\recipe_options.json`
+- **Keys:** Entries in this JSON file use the full prefixed model name (e.g., `"user.Qwen3.5-35B-A3B-GGUF"`).
+
+**Importing via Web UI**
+Instead of manually editing the JSON file, you can also easily add recipes using the Lemonade Web Interface:
+1. Open the Lemonade Web Interface (usually `http://localhost:8000`).
+2. Navigate to the model management section.
+3. Click on **"Import a model"**.
+4. Upload the recipe configuration.
```
jeremyfowers: @bitgamma is there a way to import a recipe.json file using the CLI? That would blend a lot smoother into this guide.

docs/server/apps/claude-code.md
```diff
@@ -0,0 +1,147 @@
...
+
+**Settings Priority**
+When loading a model for a launched agent, Lemonade Server resolves settings in this order (highest priority first):
+1. Explicit values passed in the load request (e.g., using `--llamacpp-args` via CLI).
+2. Per-model values defined in `recipe_options.json` (used when `--use-recipe` is active).
+3. Global environment variables (e.g., `LEMONADE_MAX_LOADED_MODELS`).
+4. Hardcoded system defaults.
```
jeremyfowers: I hope that by streamlining the workflow above we can avoid the need for documentation like this. Let's simplify!

sawansri:
Thanks for the review! I guess I'll explain my vision for what the `launch` subcommand should look like.

The user starts out with `lemonade-server launch claude`. 
1. This queries the `models` endpoint which returns a list of valid models for the user to choose from. (If they have no models installed, possibly query with show all and let them pull any model as well?)
2. After the user selects the model that they want to load, they are presented with the option of using model recipe or not (this option is skipped altogether if they use the `--use-recipe` flag). The recipe is automatically downloaded and imported (if this option exists through CLI).
3. Launch claude

docs/server/apps/claude-code.md
sawansri: I'm leaning more towards option 2 here, using the API gives us more flexibility.

docs/server/apps/claude-code.md
sawansri: So the ThinkingCoder recipe does have better performance for our agentic coding use case (the options are recommended by unsloth: https://unsloth.ai/docs/models/qwen3.5#recommended-settings). I do reference the wiki as a source of model recipes, so if we update that location, I think we should be ok. 

I also elaborate on how `launch` should interact with recipes in my comment below,  let me know what you think.

docs/server/apps/claude-code.md
jeremyfowers: Sounds good!

jeremyfowers:
> Thanks for the review! I guess I'll explain my vision for what the `launch` subcommand should look like.

>     1. This queries the `models` endpoint which returns a list of valid models for the user to choose from. (If they have no models installed, possibly query with show all and let them pull any model as well?)

Nice, I like this interactivity. I wouldn't mind if `run` inherited it as well (since `launch` may be borrowing from `run`'s code path as it is, perhaps this can all be shared?)

I _always_ end up calling `list` before `run` so this will be great!

>     2. After the user selects the model that they want to load, they are presented with the option of using model recipe or not (this option is skipped altogether if they use the `--use-recipe` flag). The recipe is automatically downloaded and imported (if this option exists through CLI).

Programmatically downloading recipes probably requires some kind of recipe repo, right? I doubt the wiki will scale to this.

We could start such a repo now, for use specifically with Claude code, and see how it goes? If you think it would add value. If the base Qwen3.5-35B model is sufficiently strong, then we can skip this for now and just set reasonable default `--llamacpp-args`. __This seems like the pivotal decision here.__

jeremyfowers:
Setting to draft until we merge Spring Cleaning 1 and come up with a final plan.

sawansri:
I was considering adding two or three more features/flags to `launch` on top of #1454, would love to get some thoughts. They are:

1. support for `--resume session-id` flag and maybe a general `--args`: Claude Code offers session resuming as a feature, we currently don't have a way to leverage this. 

There are two routes we could go with, the first being only enabling an `--arg` flag in which the user can pass in `--resume session-id` plus additional arguments (similar to the `--llamacpp-args` arg we already have but for agents).

The second approach is to explicitly define `--resume session-id` as a flag while also enabling `--arg`, imo this makes the command a lot cleaner and easier to type out. It also still exposes `--arg` for any additional arguments to `claude` the user might have. 

One thing to consider is that this would be a `lemonade launch claude` only flag since agents like codex implement `resume` as an explicit subcommand and not a cli flag. 


4. `--env` flag - give user an option to pass in env variables as key-value pairs (this one's less crucial because users can prepend the launch command with their env variable or just set it through their shell)

PR #1336, opened 2026-03-10
add standalone CLI

bitgamma:
Add a standalone CLI which can connect to remote servers and does not manage the lifecycle of the server. It only relies on the HTTP interface to work. It works even if `lemonade-router` was started directly.

Supported commands:

- [x] status
- [x] list
- [x] pull
- [x] delete
- [x] load
- [x] unload
- [x] recipes
- [x] export 
- [x] launch
- [x] run
- [x] scan
- [x] import

Main differences with existing server/tray/cli:

- cannot manage lifecycle state of the server
- can connect to server on remote machines
- expects host and port to be provided explicitly if not set to default values (can read the ENV though)
- supports exporting models as JSON
- supports adding any of the supported labels during pull (--label can be given multiple times)
- supports multiple checkpoints even without JSON (--checkpoint can be given multiple times)

Made with VSCode + RooCode + Qwen3.5-35B-A3B-ThinkingCoder.

### DIFF
--- a/docs/server/lemonade-cli-experimental.md
+++ b/docs/server/lemonade-cli-experimental.md
@@ -14,6 +14,7 @@ The `lemonade` command-line interface (CLI) provides an HTTP client for interact
 - [Options for export](#options-for-export)
 - [Options for recipes](#options-for-recipes)
 - [Options for launch](#options-for-launch)
+- [Options for scan](#options-for-scan)
 
 ## Commands
 
@@ -35,6 +36,7 @@ The `lemonade` command-line interface (CLI) provides an HTTP client for interact
 | `recipes`           | List available recipes and backends. Use `--install` or `--uninstall` to manage backends. |
 | `export MODEL_NAME` | Export model information to JSON format. See command options [below](#options-for-export). |
 | `launch AGENT`      | Launch an agent with a model. See command options [below](#options-for-launch). |
+| `scan`              | Scan for network beacons on the local network. See command options [below](#options-for-scan). |
 
 ## Global Options
 
@@ -430,6 +432,34 @@ lemonade launch codex --model Qwen3-0.6B-GGUF --llamacpp vulkan
 
 # Launch an agent with custom llama.cpp arguments
 lemonade launch claude --model Qwen3-0.6B-GGUF --ctx-size 4096 --llamacpp-args "--flash-attn on --no-mmap"
+```
+
+## Options for scan
+
+The `scan` command scans for network beacons on the local network. Beacons are UDP broadcasts sent by Lemonade Server instances to announce their presence. This command listens for these beacons and displays any discovered servers:
+
+```bash
+lemonade scan [options]
+```
+
+| Option | Description | Default |
+|--------|-------------|---------|
+| `--duration SECONDS` | Scan duration in seconds | `30` |
+
+**Notes:**
+- The scan listens on UDP port 8000 for beacon broadcasts
+- Each beacon must contain `service`, `hostname`, and `url` fields in JSON format
+- Duplicate beacons (same URL) are automatically filtered out
+- The scan runs for the specified duration, collecting all beacons during that time
+
+**Examples:**
+
+```bash
+# Scan for beacons for the default duration (30 seconds)
+lemonade scan
+
+# Scan for beacons for a custom duration
+lemonade scan --duration 5
 ```
 
 ## Next Steps
--- a/src/cpp/cli/main.cpp
+++ b/src/cpp/cli/main.cpp
@@ -4,11 +4,25 @@
 #include <lemon_tray/agent_launcher.h>
 #include <lemon/utils/process_manager.h>
 #include <lemon/utils/path_utils.h>
+#include <lemon/utils/network_beacon.h>
 #include <CLI/CLI.hpp>
 #include <iostream>
 #include <string>
 #include <fstream>
 #include <nlohmann/json.hpp>
+#include <chrono>
+#include <unordered_set>
+
+#ifdef _WIN32
+    #include <winsock2.h>
+    #include <ws2tcpip.h>
+    typedef int socklen_t;
+#else
+    #include <arpa/inet.h>
+    #include <netinet/in.h>
+    #include <sys/socket.h>
+    #include <unistd.h>
+#endif
 
 static const std::vector<std::string> VALID_LABELS = {
     "coding",
@@ -52,6 +66,7 @@ struct CliConfig {
     std::string output_file;
     bool downloaded = false;
     std::string agent;
+    int scan_duration = 30;
 };
 
 static bool validate_and_transform_model_json(nlohmann::json& model_data) {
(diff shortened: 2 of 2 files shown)


bitgamma:
@jeremyfowers I fully expect that changes will be needed, but I need a second pair of eyes to see what is missing to consider this version of the CLI full-featured

jeremyfowers:
@claude review. Please note that this is an experimental new client and doesn't need to be perfect yet, nor does it need to have a complete feature set yet. I mainly want to know if it is architecturally sound.

jeremyfowers:
> 1. Tray dependency undermines "standalone" goal
> main.cpp includes <lemon_tray/agent_launcher.h> and <lemon/utils/process_manager.h> for the launch command. This couples the CLI to the tray subsystem and means the binary can't be built in isolation. This is probably unavoidable for launch, but it's worth being deliberate about it — the launch command is qualitatively different from all other commands (it runs local processes rather than just making HTTP calls). Consider either documenting this coupling clearly, or keeping launch as a separate subcommand that's only compiled when BUILD_TRAY=ON.

@bitgamma should agent_launcher move out of lemon_tray entirely then?

bitgamma:
@jeremyfowers for sure, when we phase-out the old one I'll carry the file over. I didn't touch it for now, but the file is self-contained and doesn't add further dependencies on the tray.

bitgamma:
addressed the relevant comments from claude's review

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -537,6 +537,7 @@ jobs:
           .venv/bin/python test/server_cli.py --server-binary /usr/local/bin/lemonade-server --ephemeral
```
jeremyfowers, line 537: I started the server on a non-default port:

New cli:
```
(base) PS C:\work\forks\origin\lemonade> .\build\Release\lemonade.exe status   
Error fetching health status: Connection failed: Could not establish connection
```

old cli:
```
(base) PS C:\work\forks\origin\lemonade> .\build\Release\lemonade-server.exe status
Server is running on port 8080
```

For me, the main purpose of `status` is to figure out where the lemonade server is, regardless of port.

Thoughts about using the Beacon stuff to enable the new cli to do this without scanning the process list?

jeremyfowers: Also, it would be great if all the other commands (like `list`) worked without making the user type `--port`: can the Beacon help with that too?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -537,6 +537,7 @@ jobs:
...
           .venv/bin/python test/server_cli.py --server-binary /usr/local/bin/lemonade-server --listen-all
           .venv/bin/python test/server_cli.py --server-binary /usr/local/bin/lemonade-server --api-key
```
jeremyfowers, line 539: I see that `list` added the checkpoint for the model to the table, like 

```
user.Qwen3.5-35B-A3B-ThinkingCoder      Yes         llamacpp (unsloth/Qwen3.5-35B-A3B-GGUF:Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf)
```

I would prefer if checkpoint was another table column to keep things clean, instead of having the checkpoint in parenthesis.

Also, thoughts about if its getting too wide? Should there be a `-v` mode that reveals the checkpoint (and potentially other info?)

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -537,6 +537,7 @@ jobs:
...
+          .venv/bin/python test/server_cli2.py --server-binary /usr/local/bin/lemonade-server
           echo "CLI tests PASSED!"
```
jeremyfowers, line 541: There's now an `export` command, but no `import` command. Should there be an `import` command?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -537,6 +537,7 @@ jobs:
...
 
       - name: Run endpoint tests (unsigned path)
```
jeremyfowers, line 543: I tried the `load` command on a model I hadn't downloaded yet, and it hung like this:

```
(base) PS C:\work\forks\origin\lemonade> .\build\Release\lemonade.exe load Qwen3-0.6B-GGUF
Loading model: Qwen3-0.6B-GGUF

```

This CLI should work the same way the Electron app does: 
1. check `/models` to see if the model is downloaded
2. run `/pull MODEL stream: true` and display download progress
3. run `/load MODEL`

Or we can say "you need to `lemonade pull MODEL` first", but the above sounds smoother.

.github/workflows/cpp_server_build_test_release.yml
bitgamma: I really wouldn't like to open that can of worms, it brings a lot of platform-dependent code, edge cases, error handling for very little gain. Using the network beacon works until you only run on one computer. What if you have two on your network? you run the command on a random one?

After all the server is not picking a random port on each start, why does it need to be discovered?

btw the `status` command here outputs the result of the /health endpoint, so it gives more useful info than just the port

To avoid typing, one can set the LEMONADE_HOST, LEMONADE_PORT, LEMONADE_API_KEY env variables (on all OS they can be set permanently per-user)

.github/workflows/cpp_server_build_test_release.yml
bitgamma: I could remove the checkpoint, Qwen added it unasked, but I see the current version doesn't list it

.github/workflows/cpp_server_build_test_release.yml
bitgamma: pull does that. Should I refactor pull to stop doing that and a dedicated import? could make sense indeed

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1126,6 +1127,7 @@ jobs:
               $VENV_PYTHON test/server_cli.py --server-binary "$SERVER_BINARY" --ephemeral
               $VENV_PYTHON test/server_cli.py --server-binary "$SERVER_BINARY" --listen-all
```
jeremyfowers, line 1128:
```
(base) PS C:\work\forks\origin\lemonade> .\build\Release\lemonade.exe status     
Lemonade Server Status
==================================================
Status: ok
Version: 9.4.1
WebSocket Port: 9000
Max Models Per Type: 1
All Loaded Models:
==================================================

  Model: Qwen3-0.6B-GGUF
  Checkpoint: unsloth/Qwen3-0.6B-GGUF:Q4_0
  Type: llm
  Device: gpu
  Recipe: llamacpp
  Backend URL: http://127.0.0.1:8001/v1
  Last Used: 256826512
  Recipe Options:
    llamacpp_backend: "vulkan"

==================================================
(base) PS C:\work\forks\origin\lemonade> .\build\Release\lemonade.exe unload
Unloading model: 
Model unloaded successfully!
(base) PS C:\work\forks\origin\lemonade> .\build\Release\lemonade.exe status
Lemonade Server Status
==================================================
Status: ok
Version: 9.4.1
WebSocket Port: 9000
Max Models Per Type: 1
==================================================
```

I would have expected the `unload` command there to print "Unloading all models"

.github/workflows/cpp_server_build_test_release.yml
bitgamma: ok, can do

.github/workflows/cpp_server_build_test_release.yml
bitgamma: you are right, will do

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: In the electron app we locked it down so that we only look at beacons on localhost

`lemonade-router` has a PID/mutex (depending on OS) that ensures there cant be 2 on the same localhost

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Either way! Mainly just want to make sure it keeps looking clean.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: See I didn't know pull could do that :D

```
pull                        Pull/download a model
```
I see two options, neither is perfect:

### Option A
Put an explicit `import` command next to `export` 

Pro: way more discoverable
Con: high redundancy with `pull` command; bloats the command list

### Option B

Change the `pull` help to say "Pull/download/import a model"

Pro: less verbose, no redundancy
Con: now it's confusing/overloaded since "download" and "import" are very different things for 1 command to do

jeremyfowers: Oh one more wrinkle for `import`: we're going to have https://github.com/lemonade-sdk/recipes !

So ideally we can `import` by a name reference in the recipes repo *or* a local JSON.

At that point pull would be extremely overloaded... maybe that answers the question right there.
 
@sawansri if you want to weigh in?

bitgamma: probably option A is better

bitgamma: I went for option A *and* am quite excited to add support for remote recipes when the repo is ready

.github/workflows/cpp_server_build_test_release.yml
bitgamma: added a scan command that uses the beacon

```
[user@host lemonade]$ build/lemonade scan --duration 2
Scanning for network beacons on port 8000 for 2 seconds...
Listening for beacons...
  Discovered: some_host at http://127.0.0.1:8000/api/v1/

Scan complete. Found 1 beacon(s):
  - some_host at http://127.0.0.1:8000/api/v1/
```

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers, line 543: Thoughts about this here?

New CLI:
```
--llamacpp-args ARGS (Env:LEMONADE_LLAMACPP_ARGS)
                              Custom arguments to pass to llama-server (must not conflict with managed args)
--whispercpp BACKEND (Env:LEMONADE_WHISPERCPP)
                              WhisperCpp backend to use
```

Old CLI:
```
  --llamacpp BACKEND:{vulkan,rocm,cpu} [vulkan]  (Env:LEMONADE_LLAMACPP)
                              LlamaCpp backend to use
  --whispercpp BACKEND:{npu,cpu} [npu]  (Env:LEMONADE_WHISPERCPP)
                              WhisperCpp backend to use
```

the specific backends are populated

technically the new CLI could call the system-info endpoint to provide this info, and it should be pretty fast since we cache system-info on the server side.

bitgamma: It would be very nice, but argument parsing happens only *after* this message is generated. Without those, I don't know where to connect (assuming localhost will be wrong for a lot of people). Also connection could fail (even with a long timeout) and that's unexpected for an help command. I'd say we can add this later if we find a great way to do it

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Nice!

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: yeah lets look into it later as post-merge polishing

PR #1337, opened 2026-03-10
Add CI workflow to validate new llama.cpp releases

jeremyfowers:
## Summary
- Adds `.github/workflows/validate_llamacpp.yml` — a new workflow that fetches the latest llama.cpp and llamacpp-rocm releases, builds lemonade with the updated versions, and tests all "hot" llamacpp models
- Adds `test/validate_llamacpp.py` — a standalone validation script that loads each hot model, sends chat/completions requests, collects /stats, and reports pass/fail
- On cron (Sunday 12 PM ET) or manual dispatch, automatically creates a PR with backend_versions.json updates if all models pass both vulkan and rocm

### DIFF
--- a/.github/workflows/validate_llamacpp.yml
+++ b/.github/workflows/validate_llamacpp.yml
@@ -1,7 +1,6 @@
 name: Validate New llama.cpp Release
 
 on:
-  pull_request:
   workflow_dispatch:
   schedule:
     # Sunday at 12:00 PM ET (17:00 UTC, 16:00 UTC during EDT)


jeremyfowers:
@meghsat to check out the 128 GB STX Halo runner IRL and see why its not loading Qwen3.5-122b

.github/workflows/validate_llamacpp.yml
```diff
@@ -0,0 +1,387 @@
+name: Validate New llama.cpp Release
+
+on:
+  pull_request:
```
jeremyfowers:
```suggestion
```

Remove before merging

Issue #1303, opened 2026-03-05
FLM Linux Reddit Post


jeremyfowers:
## You can run LLMs on your AMD NPU on Linux!

If you have a Ryzen™ AI 300/400-series PC and run Linux, we have good news!

You can now run **LLMs directly on the AMD NPU** in Linux at **high speed**, **very low power**, and **quietly on-device**.

Not just small demos, but **real local inference**.

---

#### 🍋 Lemonade Server
Lightweight Local server for running models on the AMD NPU.

Guide:  https://lemonade-server.ai/flm_npu_linux.html 
GitHub: https://github.com/lemonade-sdk/lemonade

#### ⚡ FastFlowLM (FLM)
Lightweight runtime optimized for AMD NPUs.

GitHub:  
https://github.com/FastFlowLM/FastFlowLM  
https://github.com/FastFlowLM/FastFlowLM/blob/main/docs/linux-getting-started.md

---

This stack brings together:

- Upstream NPU driver in the Linux 7.0+ kernel (with backports for 6.xx kernels)
- AMD IRON compiler for XDNA NPUs
- FLM runtime  
- Lemonade Server

We'd love for you to try it and let us know what you build with it on Discord: https://discord.gg/5xXzkMu8Zk

ramkrishna2910:
Looks crisp and to the point!
Can we add a GIF of lemonade app showing an FLM model running on linux desktop?

jeremyfowers:
https://github.com/user-attachments/assets/da1bb216-ee6c-454c-91eb-09d8ad3330d9

yep @ramkrishna2910 here's the gif

PR #1338, opened 2026-03-10
Update the FLM Linux guide to work as a blog

jeremyfowers:
image

Companion reddit post draft: https://github.com/lemonade-sdk/lemonade/issues/1303


ramkrishna2910:
Can we mention `NPU` on the title?

superm1:
I think we should drop the `flm validate` step and move it to FAQ/troubleshooting.  The reason is; most users will not need to do it if they followed above steps and the memlock stuff will just cause confusion.

superm1:
As an alternative how about we tell them to restart the systemd service (which is a missing step!)

jeremyfowers:
> I think we should drop the `flm validate` step and move it to FAQ/troubleshooting. The reason is; most users will not need to do it if they followed above steps and the memlock stuff will just cause confusion.

It was important for me to get memlock set correctly to have the `flm validate` step right where it is.

ZaneNi (approved):

ZaneNi:
@jeremyfowers I found a typo here.

`XNDA` -> `XDNA`

image

PR #1339, opened 2026-03-10
Attempt to preserve capabilities

superm1:
Rather than giving lemonade-server/lemonade-router free reign on memory resources, instead give them the ability to control resource limits.  Then pass the resource limits into child processes using libcap.


superm1:
This builds (relies) upon @kenvandine's idea here: https://github.com/FastFlowLM/FastFlowLM/pull/397

PR #1340, opened 2026-03-10
Fix streaming: call sink.done() on all error paths to properly terminate chunked responses

abn:
Without sink.done(), clients see 'incomplete chunked read' when the backend returns non-200 (e.g. context overflow) or an exception is thrown. Add sink.done() in the error branches of forward_sse_stream, forward_byte_stream, and forward_streaming_request. Add integration tests that reproduce the bug via context overflow and verify clean termination across all paths.

Relates-to: https://github.com/nowledge-co/community/issues/105

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -475,8 +475,10 @@ jobs:
           # Copy binaries to /usr/local/bin
           sudo cp build/lemonade-router /usr/local/bin/
           sudo cp build/lemonade-server /usr/local/bin/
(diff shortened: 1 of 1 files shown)


bitgamma:
Looks good to me, thanks for your contribution! Please add the tests to CI. Probably is a good idea to just add them to the existing endpoint tests

test/server_streaming_errors.py
```diff
@@ -0,0 +1,163 @@
+"""
```
bitgamma, line 1: these tests need to be added to CI.

bitgamma (approved):

PR #1342, opened 2026-03-11
Fix model size filtering

superm1:
For some reason the logic was looking at the system memory rather than the memory pool available by the GPU.  This exposes models that can't possibly load on a 32GB config to such a config.

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1229,6 +1229,10 @@ bool parse_TF_env_var(const char* env_var_name) {
                    std::string(env) == "yes");
 }
 
+static bool is_gpu_device_key(const std::string& dev_type) {
+    return dev_type.find("gpu") != std::string::npos;
+}
+
 std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
     const std::map<std::string, ModelInfo>& models) {
 
@@ -1270,10 +1274,14 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
                          hardware["amd_npu"].is_object() &&
                          hardware["amd_npu"].value("available", false);
 
-    double largest_mem_pool_gb = 0.0;
+    double largest_gpu_mem_pool_gb = 0.0;
     double curr_mem_pool_gb = 0.0;
 
     for (const auto& [dev_type, devices] : hardware.items()) {
+        if (!is_gpu_device_key(dev_type)) {
+            continue;
+        }
+
         // Because we have mixed types this just makes every device_type an array.
         nlohmann::json dev_list = devices.is_array() ? devices : nlohmann::json{devices};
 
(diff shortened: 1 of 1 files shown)


ramkrishna2910:
@claude review

ramkrishna2910 (changes requested):
Please handle the 2 minor bugs identified.

src/cpp/server/model_manager.cpp
```diff
@@ -1349,18 +1358,26 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
...
-                filter_out = true;
-                std::ostringstream oss;
-                oss << std::fixed << std::setprecision(1);
-                oss << "This model requires approximately " << info.size << " GB of memory, "
-                    << "but your system only has " << system_ram_gb << " GB of RAM. "
-                    << "Models larger than " << max_model_size_gb << " GB (80% of system RAM) are filtered out.";
-                filter_reason = oss.str();
-            }
+        // Filter out GPU-targeted models that are too large for the detected GPU memory pool.
+        if (!filter_out && uses_gpu && largest_gpu_mem_pool_gb > 0.0 && info.size > largest_gpu_mem_pool_gb) {
```
ramkrishna2910, line 1362: Should we `info.size > largest_gpu_mem_pool_gb * some_factor` so we account for kv cache and other overhead?

src/cpp/server/model_manager.cpp
```diff
@@ -1229,6 +1229,10 @@ bool parse_TF_env_var(const char* env_var_name) {
                    std::string(env) == "yes");
 }
 
+static bool is_gpu_device_key(const std::string& dev_type) {
```
ramkrishna2910, line 1232: /api/v1/models now has 4 code paths (GPU-only too big / CPU-only too big / both / neither) can we add a unit test covering the matrix?

src/cpp/server/model_manager.cpp
```diff
@@ -1311,18 +1318,20 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
...
+        if (largest_gpu_mem_pool_gb > 0.0) {
+            LOG(INFO, "ModelManager") << "  - Largest GPU memory pool: " << std::fixed << std::setprecision(1) << largest_gpu_mem_pool_gb << std::endl;
         }
         debug_printed = true;
     }
 
-    int filtered_count = 0;
     for (const auto& [name, info] : models) {
         const std::string& recipe = info.recipe;
+        DeviceType device = info.device != DEVICE_NONE ? info.device : get_device_type_from_recipe(recipe);
```
ramkrishna2910, line 1332: NPU-targeted models lost their size filters - flm and ryzenai-llm are DEVICE_NPU. 
Neither uses_cpu nor uses_gpu fires, we used to filter them against max_model_size_gb. NPU models should also have their own memory constraints that should gate listing.

src/cpp/server/model_manager.cpp
ramkrishna2910, line 1232: his misses macOS. On Mac builds the GPU entry is populated under the key "metal" which doesn't contain the substring "gpu". So on all macs largest_gpu_mem_pool_gb comes out as 0. Can we move to an explicit allow list instead of substring matches?

src/cpp/server/model_manager.cpp
ramkrishna2910, line 1332: When largest_gpu_mem_pool_gb == 0 (Docker/VM without passthrough, or any GPU-detection failure), the > 0.0 guard skips the GPU filter. So a 70B model shows up in /api/v1/models on a 16 GB laptop and OOMs at load. We used to catch this via 0.8 * system_ram.
In model_types.h should we set llamacpp → DEVICE_CPU | DEVICE_GPU?

superm1:
Closing in favor of https://github.com/lemonade-sdk/lemonade/pull/2638

PR #1343, opened 2026-03-11
fix: FLM may emit log lines to stdout before the JSON, so handle it

kenvandine:
The output of flm for checking version, can get polluted with other output, so handle it.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -1489,8 +1489,11 @@ std::string SystemInfo::get_flm_version() {
     #endif
 
     // Parse JSON output: { "version": "0.9.34" }
+    // FLM may emit log lines to stdout before the JSON (e.g. "[FLM]  Using custom model list path: ...")
+    // so find the first '{' and parse from there.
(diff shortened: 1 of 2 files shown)


joeldushouyu:
Any chance to see the stdout or stderr in this case? Because when printing to std:cerr, it means something fatal happened and is unexpected in FLM.  Or it could be something that was unintentionally commited to main while debugging.

src/cpp/server/system_info.cpp
```diff
@@ -1489,8 +1489,11 @@ std::string SystemInfo::get_flm_version() {
     #endif
 
     // Parse JSON output: { "version": "0.9.34" }
+    // FLM may emit log lines to stdout before the JSON (e.g. "[FLM]  Using custom model list path: ...")
```
superm1, line 1492: this seems like an FLM bug, that message shouldn't come out in --json mode

kenvandine:
I've submitted a PR to flm to fix it there. https://github.com/FastFlowLM/FastFlowLM/pull/402

PR #1344, opened 2026-03-11
Update flm_npu_linux.html

de-wim:
Fix git clone instructions.

`cmake` fails without cloning the submodules.


sofiageo:
There is also https://aur.archlinux.org/packages/fastflowlm available. I don't think installing it manually is a good idea for Arch Linux. The package might need some improvements but it's available

superm1:
I say let's put this for now to unblock directions, and any follow ups to use the AUR can be a follow up.

PR #1346, opened 2026-03-12
update backend llamacpp versions in backend_versions.json

james-martinez:
Updated version numbers for llama.cpp backends to the latest releases to support Nemotron-3-Super-120B-A12

llama_model_load: error loading model: check_tensor_dims: tensor 'blk.1.ffn_down_exps.weight' has wrong shape; expected   2688,   4096,    512, got   2688,   1024,    512,      1

on vulkan b8175

https://github.com/ggml-org/llama.cpp/pull/20411


james-martinez:
Closing. This should be an automated pr.

PR #1347, opened 2026-03-12
Install FLM in the docker file

superm1:


Dockerfile
```diff
@@ -48,8 +48,16 @@ RUN apt-get update && apt-get install -y \
     libvulkan1 \
     unzip \
     libgomp1 \
+    software-properties-common \
     && rm -rf /var/lib/apt/lists/*
 
+RUN add-apt-repository -y ppa:amd-team/xrt
+
+# Download FLM and install it
+RUN curl -L -o fastflowlm.deb https://github.com/FastFlowLM/FastFlowLM/releases/download/v0.9.35/fastflowlm_0.9.35_ubuntu24.04_amd64.deb && \
```
jeremyfowers: Oh actually are you sure you want to hardcode the 0.9.35 here? Seems like a maintenance headache. You could do `latest`, or try to grab the value for `backend_versions.json`

superm1: Good idea to grep that. Will change!

PR #1349, opened 2026-03-12
Add libatomic1 as Docker runtime dependency for ROCm backend

jaeiclee:
- Add libatomic1 to runtime dependencies in Dockerfile
   - Update DOCKER_GUIDE.md to document ROCM backend support
   - Include libatomic1 in example Dockerfile configuration

From:
```[Error] (Process) /root/.cache/lemonade/bin/llamacpp/rocm/llama-server: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory```

To:
```[Info] (Process) ggml_cuda_init: found 1 ROCm devices: ```

Confirmed working on Ubuntu 26.04 (6.19.0-9-generic) with Strix Halo (AMD RYZEN AI MAX+ 395 w/ Radeon 8060S) with the usual configuration needed for running GPU inference [with containerized llama.cpp on Strix Halo](https://github.com/kyuz0/amd-strix-halo-toolboxes)

While the user would still need to fulfill the prerequisites on the host and docker configuration, this will at least allow the user to launch the backend for further diagnosis.

### DIFF
--- a/src/cpp/DOCKER_GUIDE.md
+++ b/src/cpp/DOCKER_GUIDE.md
@@ -12,6 +12,7 @@ docker run -d \
   -p 8000:8000 \
   -v lemonade-cache:/root/.cache/huggingface \
   -v lemonade-llama:/opt/lemonade/llama \
+  -v lemonade-recipe:/root/.cache/lemonade \
   -e LEMONADE_LLAMACPP_BACKEND=cpu \
   ghcr.io/lemonade-sdk/lemonade-server:latest
 ```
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jaeiclee** [comment]
To be clear, the configuration needed for docker is just adding `--device /dev/dri --device /dev/kfd` as parameters, but this is only tested on Strix Halo - I cannot speak for other ROCm-based GPUs on whether this will be enough.

The purpose of this PR is to allow the user to use this image without having to build a new one for running the ROCm backend by adding a small package that shouldn't affect the image size too much.

**Comment by jeremyfowers** [comment]
Cool! There have been quite a few requests for this. You should announce it on the discord when it merges @jaeiclee !

**Comment by jaeiclee** [comment]
Ahh- sorry about the trouble. I was going to file a separate PR with better documentation for Dockerfile and Docker compose (that also adds an additional persistent volume for recipes), but I accidentally pushed it to this repo by accident, and trying to only remove those new commits didn't go so well...as you can see here.😢 

I'll reopen the request after they are done.

**Comment by jaeiclee** [comment]
OK, it should be better now. Sorry again for the trouble.

About the new persistent volume "lemonade-recipe:/root/.cache/lemonade" - 

I'm starting to think that the previous `lemonade-cache` volume might no longer be used, as backends were actually stored in `/root/.cache/lemonade` during my trial run, and no directories added to `lemonade-cache` had anything added to them.

However, I kept that volume as-is and named the new volume as `lemonade-recipe` so that it doesn't introduce breaking effects to any existing deployments while allowing the users to keep the backend binaries persistent as intended with the other volume.

Please remove the other volume if this is indeed the case and you feel that it's safe to do so.

File: src/cpp/DOCKER_GUIDE.md [hunk 1]
```diff
@@ -24,13 +25,31 @@ docker run -d \
...
+docker run -d \
+  --name lemonade-server \
+  -p 8000:8000 \
+  -v lemonade-cache:/root/.cache/huggingface \
+  -v lemonade-llama:/opt/lemonade/llama \
+  -v lemonade-recipe:/root/.cache/lemonade \
+  -e LEMONADE_LLAMACPP_BACKEND=rocm \
+  --device=/dev/kfd \
+  --device=/dev/dri \
+  ghcr.io/lemonade-sdk/lemonade-server:latest
```
superm1, line 48: I think it would be a good idea to add the documentation for passing through NPU too.

superm1: As a future follow up!

jaeiclee: Sure thing. I plan to do it when it's merged into the main branch and no one else does it before I do. :)

PR #1351, opened 2026-03-12
fix: ignore stdout from flm version --json to ensure we only get valid json

kenvandine:
When  FLM_CONFIG_PATH is set, flm is mixing a log message into the output of "flm version --json". Now that https://github.com/FastFlowLM/FastFlowLM/pull/402 has been merged, that is now sent to stderr instead, so we should only capture stdout and assume it's json.


superm1:
Your commit message is wrong, but the intent is clear.  Ignore stderr!

PR #1354, opened 2026-03-12
Disable hardware caching for Linux

superm1:
It can be force enabled (like done in CI) by env variable

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1031,6 +1031,7 @@ jobs:
         run: |
           DEB_FILE="lemonade-server_${{ env.LEMONADE_VERSION }}_amd64.deb"
           echo "Installing .deb package: $DEB_FILE"
+          sudo apt-get update
           sudo apt install ./"$DEB_FILE"
 
           # Verify binaries installed


src/cpp/server/system_info.cpp
```diff
@@ -2865,6 +2865,15 @@ bool SystemInfoCache::is_valid() const {
         return false;
     }
 
+#ifdef __linux__
+    // On Linux, cache is disabled by default (always re-detect hardware) unless
+    // LEMONADE_CACHE_DIR is explicitly set (for testing purposes)
```
jeremyfowers, line 2870: Why not enable on LEMONADE_CI_MODE instead?

Otherwise there is a corner case where someone is intentionally relocating their lemonade cache dir (which contains other useful files) and accidentally ends up with hardware info caching they didnt expect.

superm1: I guess LEMONADE_CI_MODE would work too.  I'll change and confirm CI really does still pass.

superm1: Oh duh; we already have `is_ci_mode()`

jeremyfowers:
@claude does this have any unaddressed corner cases? For example, what happens if we release a new lemonade version, set `clear_bin_if_lemonade_below` to the new version number, but there is no cache file on the user's linux PC?

superm1:
> The corner case you identified does exist, but it is not introduced by this PR and the impact is narrow. Here's the full trace:

I say we split this off to a low priority task and assign a robot to fix it.

superm1:
Due to the CI failure I decided to restore the `LEMONADE_CACHE_DIR` check.  Since we're using cache on all other platforms I don't expect a failure from someone setting `LEMONADE_CACHE_DIR` like that.

src/cpp/server/system_info.cpp
```diff
@@ -2998,15 +3007,8 @@ json SystemInfoCache::get_system_info_with_cache() {
...
             } else {
-                // Provide friendly message about why we're detecting hardware
-                if (cache_exists) {
-                    std::cout << "Collecting system info (Lemonade was updated)" << std::endl;
-
-                    // Perform version-specific cleanup (e.g., removing stale backend binaries)
+                if (cache_exists)
                     cache.perform_upgrade_cleanup();
-                } else {
-                    std::cout << "Collecting system info" << std::endl;
```
jeremyfowers, line 3008: Did you remove these messages on Windows too? It's nice to show them during the 15 second wait.

superm1: It's removed for all operating systems. So you would prefer them there for windows?
How about mac?

superm1: > Did you remove these messages on Windows too? It's nice to show them during the 15 second wait.

Also - if we're going to a service on Windows does it _matter_?

jeremyfowers: lets just merge

PR #1366, opened 2026-03-14
Changes to prepare for native packaging

superm1:
I have a goal that I want to move away from cpack into native built packages for Ubuntu that will be built on a PPA.

In order to get ready for that here are some (hopefully) non-controversial changes to start.

* Update minimum version for zstd to 1.5.5
* Update examples not to pull fonts from the web
* Create vendored copy of ixwebsocket
* Add man pages for all command line tools


superm1:
As @ramkrishna2910 is going to switch from ixwebsockets to libwebsockets, let's wait on merging this one until that happens.  The other ones can go sooner.

superm1:
I have dropped the ixwebsocket vendoring since we're replacing with libwebsockets (https://github.com/lemonade-sdk/lemonade/pull/1384).  Everything else should be ready to go.

PR #1369, opened 2026-03-14
Add a new target to build the web app using system nodejs

superm1:
This will be used by the upcoming native Debian packaging target.


kenvandine:
I think this looks good, only question I have is there any changes to the workflows needed by this?

superm1:
> I think this looks good, only question I have is there any changes to the workflows needed by this?

Not immediately. I want to land all the pieces and then the natively built Debian package will be what tested it.

PR #1372, opened 2026-03-15
fix(server): support OpenCode thinking flag

szkocot:
## Summary

  Fix OpenAI-compatible chat requests from OpenCode when they disable reasoning with `thinking: false` instead of
  Lemonade's existing `enable_thinking: false`.

  ## What changed

  - accept `thinking: false` as an alias for `enable_thinking: false`
  - also handle `thinking: { "type": "disabled" }`
  - keep the existing `/no_think` forwarding behavior by prepending it to the last user message before sending the request
  to the backend
  - add a regression test for streamed chat completions on the `llamacpp` `system` backend

  ## Why

  OpenCode sends reasoning controls using `thinking`, but Lemonade only handled `enable_thinking`. That caused
  compatibility issues even though the backend-side `/no_think` mechanism already existed.

  ## Testing

  - `python3 -m py_compile test/test_llamacpp_system_backend.py`
  - `cmake --build build --target lemonade-router`


src/cpp/server/server.cpp
```diff
@@ -54,6 +54,58 @@ namespace fs = std::filesystem;
 
 namespace lemon {
 
+namespace {
+
+bool should_disable_thinking(const json& request_json) {
```
ramkrishna2910, line 59: **Precedence when both fields are present:** If a request contains *both* `enable_thinking: true` and `thinking: false`, `enable_thinking` wins because it's checked first. This may be intentional but is undocumented. A brief comment noting the priority order (e.g. `// enable_thinking takes precedence over thinking`) would help future readers.

src/cpp/server/server.cpp
```diff
@@ -54,6 +54,58 @@ namespace fs = std::filesystem;
...
+            messages[i].contains("content") &&
+            messages[i]["content"].is_string()) {
+
+            std::string original_content = messages[i]["content"].get<std::string>();
+            messages[i]["content"] = "/no_think\n" + original_content;
+            return true;
+        }
+    }
+
+    return false;
```
ramkrishna2910, line 114: **Non-string content is silently skipped (pre-existing, worth noting):** This function only handles `content` as a string. If the last user message uses array-format content (e.g. multimodal messages with images), the function returns `false` and no `/no_think` is prepended. This was the same before this PR, but now that the function is standalone it's a natural place to add a debug log:
```cpp
LOG(DEBUG, "Server") << "No string-content user message found for /no_think injection" << std::endl;
```

src/cpp/server/server.cpp
```diff
@@ -1337,31 +1389,10 @@ void Server::handle_chat_completions(const httplib::Request& req, httplib::Respo
...
-                            request_modified = true;
-                            break;
-                        }
-                    }
-                }
-            }
+        // OpenCode and other OpenAI-compatible clients may send thinking=false
+        // instead of Lemonade's enable_thinking=false.
+        if (should_disable_thinking(request_json)) {
+            request_modified = prepend_no_think_to_last_user_message(request_json);
```
ramkrishna2910, line 1405: **`thinking` / `enable_thinking` fields leak to the backend:** After prepending `/no_think`, the serialized `request_body` still contains the `thinking` (or `enable_thinking`) field. llama.cpp likely ignores unknown fields, but stripping them before forwarding would be cleaner and prevent potential issues with backends that reject unknown fields. Low priority since this is also pre-existing behavior for `enable_thinking`.

test/test_llamacpp_system_backend.py
```diff
@@ -22,30 +24,126 @@
 exit 0
 """
 
+MOCK_LLAMA_SERVER_LINUX = """#!/usr/bin/env python3
```
ramkrishna2910: **Naming / portability:** `MOCK_LLAMA_SERVER_LINUX` is a Python HTTP server that could technically run on any platform — the name implies a Linux-only implementation (like the bash-based `DUMMY_LLAMA_SERVER_LINUX_MAC`). Consider renaming to something like `MOCK_LLAMA_SERVER_PYTHON` to reflect that it's Python-based rather than platform-specific.

Also, the `@skipUnless(linux)` guard on the test means there's no coverage for the `thinking` flag on Windows/macOS. If that matters, a more portable test approach (or a separate unit-style test) could fill the gap.

test/test_llamacpp_system_backend.py
```diff
@@ -82,6 +194,18 @@ def _get_llamacpp_backends(self):
...
+        if self._model_pulled:
+            return
+
+        response = requests.post(
+            f"http://localhost:{PORT}/api/v1/pull",
+            json={"model_name": ENDPOINT_TEST_MODEL},
+            timeout=TIMEOUT_MODEL_OPERATION,
+        )
+        self.assertEqual(response.status_code, 200)
+        type(self)._model_pulled = True
```
ramkrishna2910: **`type(self)._model_pulled` pattern:** This works but is unconventional. A more idiomatic approach would be to use a `@classmethod` or move the pull into `setUpClass`. For example:
```python
@classmethod
def _ensure_model_pulled(cls):
    if cls._model_pulled:
        return
    # ... pull logic ...
    cls._model_pulled = True
```

test/test_llamacpp_system_backend.py
```diff
@@ -208,6 +332,56 @@ def test_005_prefer_system_llamacpp_disabled_or_unset(self):
...
 
+    @unittest.skipUnless(
+        sys.platform.startswith("linux"), "System backend only supported on Linux"
+    )
+    def test_006_thinking_false_maps_to_no_think_for_chat_streams(self):
+        self._write_llama_server(MOCK_LLAMA_SERVER_LINUX)
+        self._add_dummy_llama_server_to_path()
+
+        capture_path = os.path.join(self.temp_bin_dir, "captured_chat_request.json")
+        os.environ["MOCK_LLAMA_REQUEST_PATH"] = capture_path
```
ramkrishna2910, line 345: **Missing cleanup of `MOCK_LLAMA_REQUEST_PATH`:** The env var is set here but never explicitly cleaned up after the test. `setUp` does pop it, so *subsequent* tests are fine, but if this test runs last (or the suite is interrupted), the env var persists in the process. The capture file on disk is cleaned up via `shutil.rmtree(cls.temp_bin_dir)` in `tearDownClass`, so that part is fine — just the env var is a minor loose end. Consider adding cleanup in `tearDown` or using `self.addCleanup(os.environ.pop, "MOCK_LLAMA_REQUEST_PATH", None)`.

PR #1376, opened 2026-03-15
Use native debian packaging

superm1:
This is imported from Debian, but slightly modified for upstream use.  There will always be a  delta between Debian and upstream but it is expected it will decrease over time.

These native Debian packages will upload directly to PPAs owned by launchpad.net/~lemonade-team.

The CI jobs are split up as follows:
 * Single source of truth matrix (prepared for jobs that need it)
 * Pull requests will build binary packages
 * Pushes will build source packages
 * Pushes will upload to lemonade team PPA
   - No tag: bleeding-edge
   - Tag: stable
---
NOTE: This is dependent upon all of these other PRs merging to work:
 * https://github.com/lemonade-sdk/lemonade/pull/1367
 * https://github.com/lemonade-sdk/lemonade/pull/1366
 * https://github.com/lemonade-sdk/lemonade/pull/1369


superm1:
Status update (3/17)
* containers are ready: https://github.com/lemonade-sdk/lemonade/pull/1367
* new target to use system node js is ready: https://github.com/lemonade-sdk/lemonade/pull/1369
* blocked by move from ixwebsockets to libwebsockets: https://github.com/lemonade-sdk/lemonade/pull/1384
* blocked by decreasing zstd version to 1.5.5: https://github.com/lemonade-sdk/lemonade/pull/1366

superm1:
Status update (3/18):

Last blocker is #1384

.github/workflows/launchpad-ppa.yml
```diff
@@ -0,0 +1,164 @@
...
+        run: |
+          mv contrib/debian .
+          sed -e "s|@@DEB_VERSION@@|${{ steps.get_version.outputs.deb_version }}~${{ matrix.ubuntu_version }}|g" \
+              -e "s|@@DEB_CODENAME@@|${{ matrix.deb_codename }}|g" \
+              -e "s|@@DEB_DATE@@|$(date -R)|g" \
+              debian/changelog.in > debian/changelog
+
+      - name: Catch missing build-deps
+        run: |
+          apt update && apt build-dep . -y
```
kenvandine: apt-get would be better here as it has a stable cli, which isn't guaranteed with apt

contrib/debian/copyright
```diff
@@ -0,0 +1,208 @@
...
+
+Files: src/cpp/include/lemon/amdxdna_accel.h
+Copyright: 2022-2024, Advanced Micro Devices, Inc.
+License: GPL-2.0
+
+Files: debian/*
+Copyright: 2025 Mario Limonciello <superm1@debian.org>
+License: GPL-2+
+
+Files: debian/ixwebsocket/ixwebsocket/IXBase64.h
```
kenvandine: With the move away from ixwebsocket, can we drop this from debian/copyright yet?

contrib/debian/README.debian
```diff
@@ -0,0 +1,11 @@
+ixwebsocket has been vendored from https://github.com/machinezone/IXWebSocket.
```
kenvandine: Can we drop this too?

.github/workflows/launchpad-ppa.yml
superm1: ok

contrib/debian/copyright
superm1: yes, good catch!

contrib/debian/README.debian
superm1: yes, good catch!

PR #1384, opened 2026-03-16
Move to libwebsockets

ramkrishna2910:
## Summary

IXWebSocket is not available in Debian/Ubuntu package repositories and is not well maintained. This PR replaces it with https://github.com/warmcat/libwebsockets, which is available as libwebsockets-dev in Debian/Ubuntu, actively maintained (daily commits, multi-platform CI), and supports Windows/Linux/macOS.

## What changed

  - CMakeLists.txt — Replaced IXWebSocket FetchContent with libwebsockets v4.3.3 (matches Ubuntu 24.04 LTS). Removed the Windows/Linux-only platform guard, enabling WebSocket support on macOS for the first time.
  - websocket_server.h/cpp — Rewrote the WebSocket server internals using the libwebsockets callback-driven API. Introduced a message queue + lws_callback_on_writable pattern for thread-safe deferred writes.
  - streaming_audio_buffer.cpp — Replaced macaron::Base64::Decode (IXWebSocket) with lws_b64_decode_string_len (libwebsockets).

## What didn't change

  - No public API changes - Server, RealtimeSessionManager, and all callers are unaffected
  - No test changes - integration tests use the WebSocket protocol, not the library directly
  - No user-facing behavior changes - same endpoints, same message format

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1252,6 +1252,8 @@ if(UNIX AND NOT APPLE)
     set(CPACK_DEBIAN_PACKAGE_SECTION "utils")
     set(CPACK_DEBIAN_PACKAGE_PRIORITY "optional")
     set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE "amd64")
+    set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
+    set(CPACK_DEBIAN_PACKAGE_DEBUG ON)
 
     # Installation paths - use ~/.local for user install (no sudo needed)
     # Set via environment: cmake -DCPACK_PACKAGING_INSTALL_PREFIX=$HOME/.local ..


CMakeLists.txt
```diff
@@ -282,22 +282,31 @@ if(NOT USE_SYSTEM_HTTPLIB)
...
+set(LWS_WITHOUT_TEST_SERVER ON CACHE BOOL "" FORCE)
+set(LWS_WITHOUT_TEST_PING ON CACHE BOOL "" FORCE)
+set(LWS_WITHOUT_TEST_CLIENT ON CACHE BOOL "" FORCE)
+set(LWS_WITH_MINIMAL_EXAMPLES OFF CACHE BOOL "" FORCE)
+# Disable TLS for now (can be enabled later for WSS/network access)
+set(LWS_WITH_SSL OFF CACHE BOOL "" FORCE)
+set(LWS_WITH_ZLIB OFF CACHE BOOL "" FORCE)
+set(LWS_WITH_ZIP_FOPS OFF CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(libwebsockets)
+message(STATUS "libwebsockets will be fetched (v${MIN_LWS_VERSION})")
```
superm1: can you please also add it to `setup.sh` and make sure that Debian/Ubuntu builds pull from the distro?

CMakeLists.txt
```diff
@@ -658,11 +667,10 @@ if(USE_SYSTEM_HTTPLIB AND HTTPLIB_INCLUDE_DIRS)
...
 
-# IXWebSocket for WebSocket support (Windows and Linux)
-if(WIN32 OR CMAKE_SYSTEM_NAME STREQUAL "Linux")
-    target_link_libraries(${EXECUTABLE_NAME} PRIVATE ixwebsocket)
-    target_compile_definitions(${EXECUTABLE_NAME} PRIVATE LEMON_HAS_WEBSOCKET)
-endif()
+# libwebsockets for WebSocket support
+target_link_libraries(${EXECUTABLE_NAME} PRIVATE websockets)
+target_include_directories(${EXECUTABLE_NAME} PRIVATE ${libwebsockets_BINARY_DIR}/include ${libwebsockets_SOURCE_DIR}/include)
+target_compile_definitions(${EXECUTABLE_NAME} PRIVATE LEMON_HAS_WEBSOCKET)
```
superm1, line 717: make sure to add support for system provided websockets too.

superm1:
The remaining issue is that the runners don't have the distro provided libwebsockets library.  I _think_ the runners need these manually installed (the new debian package advertises them being needed).

superm1:
Either runners should be updated or the new Debian package should be installed with "sudo apt install" to pull in deps.

superm1:
I've rebased all of this work on recent changes, fixed the problem flagged with Arch and Jeremy installed the new dep on the runners.

PR #1387, opened 2026-03-16
Make AppImage build opt-in via BUILD_APPIMAGE option

kenvandine:
Add a BUILD_APPIMAGE CMake option (default OFF) that controls whether the AppImage target is part of the default build (ALL) or only available as a manual target, mirroring the BUILD_WEB_APP pattern introduced in #1369.

This avoids triggering an expensive AppImage build on every local `cmake --build` for developers who don't need it. The AppImage target remains available manually and continues to build in CI on every run.


.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -598,6 +603,7 @@ jobs:
   build-lemonade-appimage:
     name: Build Lemonade AppImage
     runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v') || inputs.build_appimage == true
```
superm1: specifically I was thinking `github.event_name != 'pull_request'`

kenvandine: Good call, I've addressed that.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -598,6 +603,7 @@ jobs:
   build-lemonade-appimage:
     name: Build Lemonade AppImage
     runs-on: ubuntu-latest
+    if: startsWith(github.event_name != 'pull_request') || inputs.build_appimage == true
```
superm1: Here's why it's failing:

```
[Invalid workflow file: .github/workflows/cpp_server_build_test_release.yml#L1](https://github.com/lemonade-sdk/lemonade/actions/runs/23228047082/workflow)
(Line: 606, Col: 9): Too few parameters supplied: 'startsWith'. Located at position 1 within expression: startsWith(github.event_name != 'pull_request') || inputs.build_appimage == true
```

kenvandine:
Fair point — if it only takes 2 minutes and catches real problems, keeping it in CI is the right call. I'll revert the CI gating change. The `BUILD_APPIMAGE=OFF` CMake option is still useful so local builds don't accidentally trigger AppImage (which requires Node/npm), so I'd like to keep that part.

setup.sh
```diff
@@ -491,6 +491,6 @@ echo ""
 print_info "Next steps:"
 echo "  Build the project: cmake --build --preset default"
 echo "  Build the electron app: cmake --build --preset default --target electron-app"
-echo "  Build AppImage (Linux): cmake --build --preset default --target appimage"
+echo "  Build AppImage (Linux): cmake --preset default -DBUILD_APPIMAGE=ON && cmake --build --preset default --target appimage"
```
superm1, line 536: Shouldn't this go into the cmake preset now?

kenvandine: The setup.sh message is just a hint/tip, not the canonical docs. Is there value in the preset?

superm1: It's not today but I'm wondering if we want multiple presets instead of telling people to add `-Dfoo`.

Don't feel strongly though.

Issue #1365, opened 2026-03-14
Auto-unload models after a set time of inactivity


wsippel:
It would be very useful if lemonade-server could auto-unload models that haven't been used for a while, similar to what Ollama does. In fact, this feature is the only reason I even still use Ollama at this point. It's just super convenient to run the inference engine as a systemd service on boot and have it available whenever I need it, but having to manually unload the models or stop the service if I want to use other VRAM-heavy applications like Blender or ComfyUI, or run a game gets annoying quickly. Lemonade already keeps track of model inactivity if I understand correctly, but only uses this functionality to unload inactive models if you need VRAM to load another model in Lemonade right now.

It would be even nicer if the server would only unload inactive models if some other application starts filling up VRAM. Best of both worlds - unused RAM is wasted RAM after all.

jaeiclee:
https://github.com/lemonade-sdk/lemonade/pull/1389

I gave this a shot, but there are a few things I'm not sure if I've done this correctly. Feel free to chime in :)

boecks:
I would love to see something like this natively in Lemonade.

Currently I bundled a small Python-based Idle Watchdog with my lemonade server docker image that does exactly this. Having a model idling resident in VRAM gives me 80 Watts power draw for basically nothing — hence the need for this feature!

The watchdog tracks per-model idle time via Lemonade's /api/v1/health endpoint, then unloads models that exceed their configured keepalive. Before unloading, it queries the llama-server's /slots endpoint to verify no inference is in progress—preventing mid-generation kills.

Configuration is flexible: you can set defaults globally via LEMONADE_KEEPALIVE env var, or specify per-model timeouts in either a dedicated keepalive_options.json file or directly inside Lemonade's recipe_options.json. Config is hot-reloadable, so changes take effect without restarting the container.

If anyone is interested how I approached it or even wants to check out my dockerized lemonade server image feel free: https://github.com/boecks/lemonade-docker

kennysamuerto:
It would be wonderful to see this. I currently work with a Strix Halo, and the 128 GB unified RAM for everything I have; it would be ideal if it were managed that way. When a tool calls a model (an embedding model, for example), it loads, but of course, it stays loaded and doesn't unload until I do it manually.

jeremyfowers:
Merged in #2183 and will be in the next release this week!

Please note that auto-unload is opt-in.

kwilkins-82:
[...]

PR #1389, opened 2026-03-17
Unload models on idle through environment variable

jaeiclee:
Disclaimer: The implementation of this feature was done by Qwen 3.5 (running on Lemonade, of course) through Pi Coding Agent.

--

The auto-unload feature automatically unloads idle models after a preset period of inactivity, which can be configured through the "LEMONADE_GLOBAL_AUTO_UNLOAD_TIMER" environment variable.

The timer will run separately for each model once loaded, reset whenever a model becomes no longer busy (=idle), and should be cleared when the model is unloaded manually.

### DIFF
--- a/docs/dev-getting-started.md
+++ b/docs/dev-getting-started.md
@@ -603,7 +603,7 @@ A GUI application for desktop users that exposes the server via a system tray ic
 The `lemonade` client communicates with `lemond` server via HTTP:
 - **Model operations:** `/api/v1/models`, `/api/v1/pull`, `/api/v1/delete`
 - **Model control:** `/api/v1/load`, `/api/v1/unload`
-- **Server management:** `/api/v1/health`, `/internal/shutdown`, `/internal/set`, `/internal/config`
+- **Server management:** `/api/v1/health`, `/internal/shutdown`, `/internal/set`, `/internal/config`, `/internal/cleanup-cache`
 - **Inference:** `/api/v1/chat/completions`, `/api/v1/completions`, `/api/v1/audio/transcriptions`
 
 The client automatically:
@@ -635,6 +635,7 @@ Internal endpoints are restricted to loopback (`127.0.0.1` / `::1`) — requests
 | `POST` | `/internal/shutdown` | Unloads all models and shuts down the server |
 | `POST` | `/internal/set` | Unified config setter (see below) |
 | `GET`  | `/internal/config` | Returns the full runtime config snapshot |
+| `POST` | `/internal/cleanup-cache` | Cleans up orphaned files in the Hugging Face cache |
 
 #### `POST /internal/set`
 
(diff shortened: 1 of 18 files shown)


jaeiclee:
Attempted as a response to https://github.com/lemonade-sdk/lemonade/issues/1365, but I'm not sure if I've done this correctly given that I have zero experience with C++.

~~Specifically this: The timer doesn't get explicitly destroyed when the model is unloaded through other means - So the max-model limit and the NPU exclusivity shouldn't cancel the timer afaik. It shouldn't be a problem since the timer will die gracefully anyway, but it's still not ideal, and might become a problem if the timer is long enough to have a lot of dead timers as models are switched automatically.~~

~~-> Should I make sure to cancel the timer for all cases where the model is unloaded by any means? and did I miss anything other than evict_all_npu_servers() and evict_server(lru)?~~

PS. I wanted to implement the per-model setting option that will be adjusted by the model's recipe file, but this would require frontend changes for the user so it's been put out-of-scope. Would be nice if we could do it as the next step after this PR.

jaeiclee:
Attempts to resolve the hanging timer makes the "Resource deadlock avoided" error return. I think I'll cook this one up a bit more. Sorry for the trouble.

jeremyfowers:
> Attempts to resolve the hanging timer makes the "Resource deadlock avoided" error return. I think I'll cook this one up a bit more. Sorry for the trouble.

no worries! this would be a great feature to have. I'm also really excited that you're using Lemonade to code Lemonade!

kenvandine:
This is something I had worked on last week, and I have a branch but not submitted as a PR yet. Maybe if you want to finish my branch? https://github.com/kenvandine/lemonade/commit/4ce4f6061276c3197e87d8646e6d1fa6c246bfe8

jaeiclee:
@kenvandine This method would have saved me a lot of headache coming from managing multiple timer threads🫠 However, I did want to eventually allow us to override the global timeout variable on a per-model basis through the recipe file (and frontend UI by extension) - that's why I had to make this work with juggling multiple timers.

I think my latest commit works well enough without crashing and behaves similarly to other unloading methods - I just need to test the NPU exclusivity case, then it should be good to go (at least in my eyes).

jaeiclee:
Alright. Works flawlessly for the NPU exclusivity case too. I think the implementation itself is as good as it gets...at least to my eyes :)

I do wonder how (or whether) we should add this feature to the CLI parameter:
-  Managing this together with "--global-timeout" is probably not a good fit since this option is not mandatory like other timeouts are.
- I was slightly cautious of adding a new CLI parameter because I see it as a UI of a sort - I wanted to get the implementation done before thinking about where else this setting can go to.
- And if we are adding this as an option to CLI, it'd be nice to have the same feature on GUI too.

Given that I started this from https://github.com/lemonade-sdk/lemonade/issues/1365, I think we can discuss further there if this gets merged (or is about to).

jaeiclee:
Now that configuration variables are being handled differently, this PR needs a lot of work before it's able to be merged into main again. I'll probably make a new pull request with more ways to configure the timers on a per-model basis. For now, consider this closed :)

PR #1396, opened 2026-03-18
AGENTS.md for non-claude agents

### LINKED ISSUE #1395: AGENTS.md for non-claude agents
Image

[AGENTS.md](https://agents.md/) is a standard similar to CLAUDE.md for agents like codex, cursor, gemini cli, etc. We should adopt this standard for better integration with these tools.

sawansri:
Resolves #1395 

Simply points agents to explore CLAUDE.md for comprehensive information about the project. We avoid duplication this way.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,3 +1,1 @@
-# CLAUDE.md
-
-Please read the instructions located in [AGENTS.md](AGENTS.md). All universal agent guidelines are maintained there to avoid duplication.
+@AGENTS.md


bitgamma:
if it doesn't mess-up claude, I'd rename CLAUDE.md into AGENTS.md and have CLAUDE.md point to AGENTS.md (that is, invert the files) since it is more inline with the openness principle. But I'll let claude users chime in and decide on this. Otherwise, of course, this is good to go. Thanks for that!

superm1:
I agree with @bitgamma, invert them like this:
https://aiengineerguide.com/til/how-to-use-agents-md-in-claude-code/

sawansri:
> if it doesn't mess-up claude, I'd rename CLAUDE.md into AGENTS.md and have CLAUDE.md point to AGENTS.md (that is, invert the files) since it is more inline with the openness principle. But I'll let claude users chime in and decide on this. Otherwise, of course, this is good to go. Thanks for that!

I agree with this, I think Claude Code should be smart enough to pick up instructions from AGENTS.md. Will test to make sure.

sawansri:
image
Works well for me!

sawansri:
> You can just use the tagging like I linked but this should be fine if you want to leave as is.

True, but if we leave a CLAUDE.md file, Claude Code will automatically load it into context vs the user having to tag it manually. I'm fine either way tbh but I think this way is a bit more user friendly.

ramkrishna2910:
I would prefer not having to tag it explicitly in workflows.

bitgamma:
I think the idea was to replace the content of `CLAUDE.md` with just this:

```
@AGENTS.md
```
and it should work exactly as @sawansri already proved, just shorter. But either way works

sawansri:
> I think the idea was to replace the content of `CLAUDE.md` with just this:
> 
> ```
> @AGENTS.md
> ```
> 
> and it should work exactly as @sawansri already proved, just shorter. But either way works

Oh, let me test this too

superm1:
> I think the idea was to replace the content of `CLAUDE.md` with just this:
> 
> ```
> @AGENTS.md
> ```
> 
> and it should work exactly as @sawansri already proved, just shorter. But either way works

Yes; exactly.

sawansri:
image
Seems to work even better! Thanks everyone.

PR #1397, opened 2026-03-18
Device option sep

genrtul:
First off, I'm inexperienced with contributing code, so sorry if there are any mistakes, and I appreciate tips!

The purpose of this simple PR is to allow passing llama-cpp's `--device` option directly to `lemonade-server`. This option selects which accelerator devices should be used by lemonade. The option is renamed `--llamacpp-device`.

The motivation for this change is that currently when using JSON recipes to pass particular parameters to models, the `--llamacpp_args` option is used, which overrides this option if it gets passed on the command line. Now, choosing which particular devices to use for acceleration strikes me as a runtime option and not appropriate for including in a JSON recipe. This is a quick fix in lieu of somehow allowing `--llamacpp_args` to merge arguments from both the command line and recipes, which might be the better fix.

I have only implemented this for the llamacpp backend for now, as that's the only one I'm familiar with. However in principle is seems like it could be generic over backends and useful so long as a backend allows choosing accelerator devices. I don't know if the option would be relevant for the other ones. SD.cpp at least doesn't seem to expose such an option currently (https://github.com/leejet/stable-diffusion.cpp/blob/master/examples/cli/README.md). I'm certainly willing to try and add it for other backends, though I won't be able to test it.

The new option --llamacpp_device is relevant for systems with two or more GPUs (for example an integrated GPU and a discrete one, or one connected over Thunderbolt/USB4)

Example usage for a system with three devices:

`lemonade config set llamacpp.device=` - Reset the option, going back to llamacpp's default device-selection behavior.

`lemonade config set llamacpp.device=Rocm0` - Only the first rocm device (usually a GPU) will be used. Which particular device this is is system-dependent.

`lemonade config set llamacpp.device Vulkan0,Vulkan2` - The two vulkan devices Vulkan0 and Vulkan2 will be used.

In ~/.cache/lemonade/recipe_options.json:
```
}
    "user.TestModel": {
    "llamacpp_device": "Rocm0"
  }
}
```
The above would make user.TestModel run on Rocm0, so long as such a device is available. If it is unavailable llamacpp will crash repeatedly.

Note that on my system, attempting to use two devices causes llama-server to repeatedly crash, but this is a separate bug.


src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -158,6 +158,7 @@ void LlamaCppServer::load(const std::string& model_name,
     LOG(DEBUG, "LlamaCpp") << "Per-model settings: " << options.to_log_string() << std::endl;
 
     int ctx_size = options.get_option("ctx_size");
+    std::string backend_device = options.get_option("device");
```
bitgamma: let's call this llamacpp_device since other backends either will not have this or might have a different format

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -200,6 +201,9 @@ void LlamaCppServer::load(const std::string& model_name,
 
     push_arg(args, reserved_flags, "-m", gguf_path, std::vector<std::string>{"--model"});
     push_arg(args, reserved_flags, "--ctx-size", std::to_string(ctx_size), std::vector<std::string>{"-c"});
+    if (backend_device != "") {
```
bitgamma: also need to add an unconditional `push_reserved` to make sure that `--device` is always added to reserved flags (no problem adding it twice since it is a set) and to also add the `-dev` alias

bitgamma (changes requested):

bitgamma:
Thanks for the contribution! Left a couple of comments. This parameter needs to also be listed in the docs

superm1:
@jeremyfowers is about to overhaul the config system and this will get caught in the cross hairs.  Can you hold off before doing more work and revisit after that's done?  I don't want to see your work thrown away.

jeremyfowers (dismissed):
@genrtul can you help me understand the need for this feature better? if you want CPU, you can do `--llamacpp cpu` already. If you want GPU, you can do `--llamacpp vulkan/rocm`.

Why do we need a new flag?

genrtul:
> @genrtul can you help me understand the need for this feature better? if you want CPU, you can do `--llamacpp cpu` already. If you want GPU, you can do `--llamacpp vulkan/rocm`.
> 
> Why do we need a new flag?

This is for the case where someone has multiple GPUs available on the system, but only wants to use a subset of them, or a different set than is used by default. My personal case is I have a Strix Halo with an eGPU, and I use `--device` to prevent the eGPU from being used.

jeremyfowers:
> > @genrtul can you help me understand the need for this feature better? if you want CPU, you can do `--llamacpp cpu` already. If you want GPU, you can do `--llamacpp vulkan/rocm`.
> > Why do we need a new flag?
> 
> This is for the case where someone has multiple GPUs available on the system, but only wants to use a subset of them, or a different set than is used by default. My personal case is I have a Strix Halo with an eGPU, and I use `--device` to prevent the eGPU from being used.

Thanks, that makes sense! So this is the GPU device ID number? 

Would you kindly update your PR description to include a couple examples of usage, such as 
* default: llamacpp does X
* --device Y: now you see llamacpp running on the iGPU
* --device Z: now you see llamacpp running on the eGPU

genrtul:
I've made the changes suggested by bitgamma and added the option in the docs where it seemed relevant. Taking into account superm1's comment I won't work on this further for now.

I've also added some examples to the OP

bitgamma (approved):
looks good to me!

bitgamma:
@genrtul the PR Mario was talking about has been now been merged

jaminellis:
Would be great to have this!

genrtul:
I've adapted this PR for the new config system. I've tested both using the new llamacpp.device key in the global config file and setting it as a per-recipe option (llamacpp_device).

Possible problem:
Changing this at runtime is a bit strange, and I'm not sure why. In my tests llamacpp crashed a few times complaining that the device it wanted disappeared, before eventually switching over to the new device and trucking on. So it's possible that some additional code is needed to inform llamacpp that it needs to do a full restart on next model load if this is changed at runtime.

Continued need?
I think this option is still needed, as llamacpp_args from the global config file and from recipes cannot combine. The one from the recipe overrides currently. The tension still exists between --llamacpp-device being logically a server-wide option (you often want all models to run on the same accelerator device, and you wouldn't want to change all your recipes to switch this) and --llamacpp-args often being used for per-recipe options.

genrtul:
I've updated the opening post with examples to reflect the new usage.

bitgamma (approved):
looks good to me, just the merge conflict needs to be solved. @jeremyfowers please check if you still have open questions about this one or if the "request changes" from your part can be solved

jeremyfowers:
> looks good to me, just the merge conflict needs to be solved. @jeremyfowers please check if you still have open questions about this one or if the "request changes" from your part can be solved

Shoot im sorry, I missed this. Please feel free to nag me on discord next time if I’m blocking something! 

@bitgamma if you’re satisfied feel free to merge.

bitgamma:
@genrtul thanks for your contribution!

PR #1399, opened 2026-03-18
Fix lemonade CLI linking errors

superm1:


bitgamma:
`try_get_backend_options` is called only code that is already inside #ifdef's so it should be stripped during linking and not trigger a linker error. That's why tests were running correctly etc. Is there a compiler version/flag combination that do no strip the dead code and cause the error?

superm1:
Now OFC with `rm -rf build/` I can't seem to reproduce again.  It was on Arch/CachyOS that I hit it.  I'll close, and if I come back with a reproducer re-open.

PR #1401, opened 2026-03-18
Add code owners for claude and agents md

ramkrishna2910:


.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# Protect AI agent and assistant configuration files from accidental overwrites
+Agents.md @ramkrishna2910 @jeremyfowers
```
sawansri: Shouldn't this be `AGENTS.md`? I believe CODEOWNERS is case sensitive so I imagine `Agents.md` points to a nonexistent file.

jeremyfowers: removing from queue until @ramkrishna2910 has a look. thanks @sawansri !

ramkrishna2910:
Good catch! Fixed it!

PR #1405, opened 2026-03-18
Job reduction revisit

superm1:
Allow the Linux jobs to be cancelled if the PR fails.


superm1:
For this to work `Build on Debian` needs to be dropped from required - it's now a build gate that blocks all other builds.

superm1:
I don't really like how much build gate slows it down though - 4 min and it can't be re-used in two workflows.  I'm gonna back it out and just do concurrency changes.

PR #1406, opened 2026-03-18
Add Linux FLM and Fedora support to install options and release notes

jeremyfowers:
## Summary
- Add FastFlowLM (NPU) as a selectable option on Linux in the install selector, with a link to the [FLM NPU Linux setup guide](https://lemonade-server.ai/flm_npu_linux.html)
- Show "NPU" instead of "NPU, Hybrid" in Device Support when FastFlowLM is the selected engine
- Add Fedora `.rpm` row to the release notes binary download table
- Rename AppImage label from "Standalone App" to "Companion Desktop App"

## Install Options Website

image

image

## Release Notes

image


jeremyfowers:
> We should be prescriptive on what Fedora version we're supporting. I don't want to pretend the RPM runs on Fedora 30 when we build 43 (IIRC?). Likewise when Fedora 44 comes out (or 45 whatever is next) we'll need to adjust our builder to track that etc.

Done

jeremyfowers:
image

docs/assets/install-selector.js
```diff
@@ -306,12 +313,68 @@ function renderDownload() {
         installCmdDiv.style.display = 'none';
       }
     } else if (distro === 'fedora') {
```
superm1, line 315: what do you want to do when fedora 44 comes out?  will 43 disappear and we just start building for 44?

jeremyfowers: Cross that bridge when we get to it!

Issue #1363, opened 2026-03-14
Unable to download rocm backend in v10 via lemonade-server recipes --install


Rebreda:
Running  `lemonade-server recipes`, (truncated) it displayed   

```bash
Recipe              Backend     Status          Message/Version                               Action
----------------------------------------------------------------------------------------------------------------------------------------------------
rocm        installable     Backend is supported but not installed.        lemonade-server recipes --install llamacpp:rocm`
```

after which i then ran `lemonade-server recipes --install llamacpp:rocm` but keep getting a 404 

Parital logs are 

```bash
2026-03-13 20:06:44.818 [Error] (Router) Failed to load model: Failed to download llama-server from: https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1203/llama-b1203-ubuntu-rocm-gfx1036-x64.zip - Download failed after 6 attempts.
2026-03-13 20:06:44.818 [Error] (Router) Last error: HTTP error 404 for URL: https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1203/llama-b1203-ubuntu-rocm-gfx1036-x64.zip
2026-03-13 20:06:44.818 [Error] (Router) Partial file preserved: /tmp/llamacpp_rocm_b1203.zip.partial
2026-03-13 20:06:44.818 [Error] (Router) Partial size: 0.0 MB
2026-03-13 20:06:44.818 [Error] (Router) Run the command again to resume from where it left off.
2026-03-13 20:06:44.818 [Error] (Server) Failed to load model: Failed to download llama-server from: https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1203/llama-b1203-ubuntu-rocm-gfx1036-x64.zip - Download failed after 6 attempts.
2026-03-13 20:06:44.818 [Error] (Server) Last error: HTTP error 404 for URL: https://github.com/lemonade-sdk/llamacpp-rocm/releases/download/b1203/llama-b1203-ubuntu-rocm-gfx1036-x64.zip
2026-03-13 20:06:44.818 [Error] (Server) Partial file preserved: /tmp/llamacpp_rocm_b1203.zip.partial
2026-03-13 20:06:44.818 [Error] (Server) Partial size: 0.0 MB
2026-03-13 20:06:44.818 [Error] (Server) Run the command again to resume from where it left off.
2026-03-13 20:06:44.818 [Error] (Server) Error 500: POST /api/v1/load
```

[...]

PR #1409, opened 2026-03-19
Stricter device check for ROCm (+gfx103X support)

jaeiclee:
Wrote in response to the [latest ROCm repo commit](https://github.com/lemonade-sdk/llamacpp-rocm/commit/0741e19615f1ee7ef40b181494d4b96271c7fdf6) and https://github.com/lemonade-sdk/lemonade/issues/1363

Since Lemonade's ROCm binaries are built with explicit device targets that are converted to a more generalized architecture-based binary name (except for gfx1150 and gfx1151), it is better to be explicit here as well.

This prevents AMD GPUs detected via KFD values from falsely claiming support and attempting to fetch non-existent binaries.

Also, this will allow the gfx103x family to be supported with the same definitions as the latest llamacpp-rocm commit.

I suppose there might be a better way to do this (like managing the supported family in the backend json file), but seems like this is the way to go without dramatically change the structure of this database.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -193,6 +193,7 @@ static const std::map<std::string, std::string> DEVICE_FAMILY_NAMES = {
     // AMD iGPU/dGPU architectures (ROCm)
     {"gfx1150", "Radeon 880M/890M (Strix Point)"},
     {"gfx1151", "Radeon 8050S/8060S (Strix Halo)"},
+    {"gfx103X", "Radeon RX 6000 series (RDNA2)"},
     {"gfx110X", "Radeon RX 7000 series (RDNA3)"},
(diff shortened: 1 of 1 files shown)


danielholanda:
@jaeiclee Can you conform you were able to run this on your device? I personally do not have a gfx103X to test this.

jeremyfowers (approved):
LGTM (as long as it works in real HW! would be nice to have some CI machines eventually...)

jaeiclee:
Whoa, merged already? Thank you!

@danielholanda Yes, but only with a 6600 (gfx1032), and I'm not sure which logic I actually triggered. I assume it's through `device_lower.find("6600")` part because I could see the GPU's full name on Beszel, but I haven't dug too deep into it.

(Also no errors on Strix Halo, but anything otherwise would be a regression.)

Issue #1411, opened 2026-03-19
Qwen-Image 2512 Support - Avoid Basic Text to Image  going to black with default quants


ianbmacdonald:
I was hoping to wire up the new Qwen Image 2512 and Image Edit 2511 building on the Q4 support upstream (sd.cpp release highlights below)  TL;DR it works, but there is a known 'activation dequantization overflow' issue with common quants `Q4_0, Q4_1, Q5_1, Q4_K`, and `Q5_K`.

2025/10/13 🚀 stable-diffusion.cpp now supports Qwen-Image-Edit / Qwen-Image-Edit 2509
👉 Details: https://github.com/leejet/stable-diffusion.cpp/pull/877

2025/10/12 🚀 stable-diffusion.cpp now supports Qwen-Image
👉 Details: https://github.com/leejet/stable-diffusion.cpp/pull/851

My first attempt was following the basic example from https://unsloth.ai/docs/models/tutorials/qwen-image-2512/stable-diffusion.cpp    The steps below allow tesing of image models on the CLI, and might be useful for testing various image prompts ahead of integration into the model manager. 

Grab the quants
```
hf download unsloth/Qwen-Image-2512-GGUF   qwen-image-2512-Q4_K_M.gguf
hf download unsloth/Qwen2.5-VL-7B-Instruct-GGUF   Qwen2.5-VL-7B-Instruct-UD-Q4_K_XL.gguf
```
Export an updated path so you can test agains the shipped sd cli, noting this uses the Ubuntu 26.04 path `/var/lib/lemonade`
```
export LD_LIBRARY_PATH=/var/lib/lemonade/.cache/lemonade/bin/sd-cpp/rocm/build/bin:$LD_LIBRARY_PATH
```
Run the example pattern with the added `--preview proj` so you can see that it starts to work in `preview.png` with each step
```
/var/lib/lemonade/.cache/lemonade/bin/sd-cpp/rocm/build/bin/sd-cli \
--diffusion-model /mnt/shelf/huggingface/hub/models--unsloth--Qwen-Image-2512-GGUF/snapshots/1626d7531f84b4d2ea1cd6d2e69f41ec027dd354/qwen-image-2512-Q4_K_M.gguf \
--vae /mnt/shelf/huggingface/hub/models--Comfy-Org--Qwen-Image_ComfyUI/snapshots/c232bcb51c1523899c62d6dcaa960b2627668de5/split_files/vae/qwen_image_vae.safetensors \
[...]

Issue #1352, opened 2026-03-12
feat: Add UI support for Image Editing and Variations (Image-to-Image)


kenvandine:
The Lemonade backend now supports Image-to-Image capabilities. The this was implemented in [lemonade-server#1121](https://github.com/lemonade-sdk/lemonade/pull/1121) and  documented in [lemonade-server#1278](https://github.com/lemonade-sdk/lemonade/pull/1278).

Currently, the Web and Electron interfaces are limited to Text-to-Image generation. Users have no way to utilize the new /images/edits and /images/variations endpoints through the built-in GUI.

It would be great to see the "Image Gen" tab expanded to support Image-to-Image workflows:

* Image Upload Component: A new UI element to upload a reference image.
* Variations Support: * A "Generate Variations" button that sends the uploaded image to the /images/variations endpoint.
  *  An "Action" button on previously generated images to quickly use them as a seed for variations.
* Edit/In-painting Support: * The ability to send both an image and a text prompt to the /images/edits endpoint.

Phqen1x:
I can work on this!

ianbmacdonald:
I was about to vibe something in for Qwen-Image-Edit in #1412  but saw this .. you need to handle two input images and a prompt, but otherwise probably works fine with sd.cpp right now

ianbmacdonald:
Here's a quick run book I used to test the Qwen-Image-Edit models that might be helpful to tune up the UI if you need some known working test cases.   Good news is they work with Lemonade sd.cpp that is shipping. 

Generate simple images for the edit text replacement prompts (replace X with Y) 
```python 
python3 -c "
from PIL import Image, ImageDraw, ImageFont
img = Image.new('RGB', (512, 512), 'white')
draw = ImageDraw.Draw(img)
try:
    font = ImageFont.truetype('/usr/share/fonts/truetype/dejavu/DejaVuSans-Bold.ttf', 60)
except:
    font = ImageFont.load_default()
draw.text((100, 220), 'flux.cpp', fill='black', font=font)
img.save('/root/test_text.png')
```
The model checkpoints that I have tested;  Q4_K_M may work fine with 2509, but I did not test

```
# 2509 Edit Works perfectly
cat > /tmp/qwen-edit-2509.json << 'EOF'
{
  "model_name": "Qwen-Image-Edit-2509-GGUF",
  "checkpoints": {
[...]

PR #1412, opened 2026-03-19
feat: overhaul multi-checkpoint storage strategy and add Qwen Image with additional sd-cpp params

ianbmacdonald:
## Summary

This PR adds Qwen Image model support to the sd-cpp backend, introduces new recipe options for advanced diffusion parameters, fixes the HuggingFace cache layout for multi-repo models, and adds a `cleanup-cache` CLI command for disk space management.

There is probably a re-usable pattern for when mmproj files from one repo are used with vision models from another (not currently any in suggested models). 

### New models

- **Qwen-Image-GGUF** — Q4_K_M quant, faster inference while preserving quality
- **Qwen-Image-2512-GGUF** — Q5_0 quant, higher fidelity (required by [sd.cpp#851](https://github.com/leejet/stable-diffusion.cpp/pull/851#discussion_r2900951475) for 2512 variant)

Both are multi-repo models with 3 checkpoints:
| Checkpoint | Repo | Purpose |
|---|---|---|
| `main` (diffusion) | `unsloth/Qwen-Image-GGUF` or `unsloth/Qwen-Image-2512-GGUF` | Diffusion model |
| `text_encoder` (LLM) | `unsloth/Qwen2.5-VL-7B-Instruct-GGUF` | Text encoding |
| `vae` | `Comfy-Org/Qwen-Image_ComfyUI` | VAE decoder |

### New sd-cpp recipe options

Four new `image_defaults` parameters for sd-cpp models, plumbed through the full pipeline (`server_models.json` → `ImageDefaults` struct → `RecipeOptions` → sd-server):

| Option | Type | Where applied | Purpose |
|---|---|---|---|
| `sampling_method` | string | Per-request (`<sd_cpp_extra_args>`) | Sampling algorithm (euler, euler_a, dpm2, etc.) |
| `flow_shift` | float | Per-request (`<sd_cpp_extra_args>`) | Flow matching shift parameter |
| `diffusion_fa` | bool→int | sd-server startup arg | Flash attention for diffusion model |
| `offload_to_cpu` | bool→int | sd-server startup arg | Offload weights to RAM, load to VRAM on demand |

**Design rationale**: Context options (`diffusion_fa`, `offload_to_cpu`) affect how the model is loaded into memory, so they're passed as sd-server startup args. Generation options (`sampling_method`, `flow_shift`) can vary per-request, so they're passed via the existing `<sd_cpp_extra_args>` mechanism embedded in the prompt. Boolean values are stored as int `0`/`1` in recipe options to avoid a `json::type_error` in the existing CLI option parser which only handles number/string types.

### HuggingFace cache layout fix (multi-repo models)

**Problem**: Previously, ALL checkpoint files from different repos were downloaded into the main repo's snapshot directory (e.g., `Qwen3-8B-Q8_0.gguf` from `unsloth/Qwen3-8B-GGUF` was saved inside `models--unsloth--FLUX.2-klein-9B-GGUF/`). This breaks the HF cache ecosystem — other tools (huggingface_hub, transformers, diffusers) can't find or share these files.  It is a bit strange considering we put the Q4_1 quant in the correct location when downloaded directly via the Suggested Models. 

**Fix**: Each repo's files now download to their own standard HF cache directory:
```
models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/{hash}/flux-2-klein-9b-Q8_0.gguf
models--unsloth--Qwen3-8B-GGUF/snapshots/{hash}/Qwen3-8B-Q8_0.gguf
models--Comfy-Org--vae-.../snapshots/{hash}/split_files/vae/flux2-vae.safetensors
```

**Backward compatibility**: `resolve_model_path` tries the checkpoint's own repo first, then falls back to searching the main repo dir for models downloaded with the old layout. No re-download required.

**Affected models**: Flux-2-Klein-4B, Flux-2-Klein-9B-GGUF, Qwen-Image-GGUF, Qwen-Image-2512-GGUF (Whisper models use `npu_cache` type which is handled separately).

**Delete behavior**: When deleting a multi-repo model, non-main checkpoint files are only removed if no other model in the registry references the same repo (conservative/safe approach). Shared repos like `unsloth/Qwen3-8B-GGUF` (used by both Flux as text_encoder and as a standalone LLM) are preserved.

### `cleanup-cache` CLI command

New command to reclaim disk space from orphaned files left by the old download layout:

```bash
lemonade-server cleanup-cache --dry-run   # preview what would be cleaned
lemonade-server cleanup-cache             # delete orphaned files
```

Scans all multi-repo models in the registry, finds checkpoint files that exist in the main repo dir but belong to a different repo, and reports/deletes them. Works via `POST /api/v1/cleanup-cache` endpoint.

```
# lemonade cleanup-cache --dry-run 
Previewing cache cleanup...
  /mnt/shelf/huggingface/hub/models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/fde8634245fe6b749a221c25b34672b5b8fbd079/Qwen3-8B-Q8_0.gguf (114.0 MB) [from Flux-2-Klein-9B-GGUF]
  /mnt/shelf/huggingface/hub/models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/fde8634245fe6b749a221c25b34672b5b8fbd079/split_files/vae/flux2-vae.safetensors (320.6 MB) [from Flux-2-Klein-9B-GGUF]

Would free 434.7 MB from 2 file(s). Run without --dry-run to delete.
```

**Migration strategy**: Lazy migration with backward-compatible fallback. Old downloads continue working via the resolution fallback. New downloads use the correct per-repo layout. Users can run `cleanup-cache` to reclaim space from orphaned files at their convenience.  It can be added as an optional migration for distribution packages.

### Bug fixes

- **`[object Object]` error display**: Load errors showed `[object Object]` instead of the actual error message because `errorData.error` is an object (`{message, type}`), not a string. Fixed in `backendInstaller.ts` to extract `.message` from the error object.
- **Boolean crash in recipe_options**: Added `is_boolean()` handling in `format_option_for_logging` and `to_cli_options` — previously booleans in JSON fell through to implicit string conversion which threw `json::type_error.302`.

## Files changed

| File | Change |
|---|---|
| `src/cpp/resources/server_models.json` | Add Qwen-Image-GGUF and Qwen-Image-2512-GGUF model entries |
| `src/cpp/include/lemon/model_manager.h` | Extend `ImageDefaults` struct, add `cleanup_orphaned_cache` |
| `src/cpp/server/model_manager.cpp` | Per-repo download dirs, fallback resolution, cleanup logic, delete for multi-repo models, new `image_defaults` parsing |
| `src/cpp/server/recipe_options.cpp` | 4 new options (DEFAULTS, CLI_OPTIONS, key list), boolean handling |
| `src/cpp/server/backends/sd_server.cpp` | Startup args (`--diffusion-fa`, `--offload-to-cpu`), per-request args (`sample_method`, `flow_shift`) |
| `src/cpp/server/server.cpp` | `cleanup-cache` endpoint registration and handler |
| `src/cpp/include/lemon/server.h` | `handle_cleanup_cache` declaration |
| `src/cpp/cli/main.cpp` | `cleanup-cache` subcommand with `--dry-run` |
| `src/cpp/cli/lemonade_client.cpp` | `cleanup_cache` client method |
| `src/cpp/include/lemon_cli/lemonade_client.h` | `cleanup_cache` declaration |
| `src/app/src/renderer/utils/modelData.ts` | Extend TS `ImageDefaults` with `sampling_method`, `flow_shift` |
| `src/app/src/renderer/utils/backendInstaller.ts` | Fix `[object Object]` in load error messages |

## Test plan

- [x] Verify both Qwen Image models appear in Suggested Models under StableDiffusion.cpp
- [x] Pull and load Qwen-Image-GGUF with `sd-cpp_backend: rocm`
- [x] Confirm sd-server startup args include `--diffusion-fa` and `--offload-to-cpu`
- [x] Generate an image and confirm `sample_method` and `flow_shift` are in `<sd_cpp_extra_args>`
- [x] Verify existing sd-cpp models (Flux, Z-Image-Turbo, SDXL) still load and generate correctly
- [x] Verify load error messages now show actual error text instead of `[object Object]`
- [x] Test `lemonade-server cleanup-cache --dry-run` reports orphaned files
- [x] Test `lemonade-server cleanup-cache` deletes orphaned files and reports space freed
- [] Test delete of multi-repo model preserves shared repos - Lemonade currently does not have this option, it purges the repo on deletes, without an option to leave a model in the HF cache when removed from Suggested that might be shared with another tool/backend; The delete prompt needs to provide better options for multi-repo purge/leave behind. 
- [x] Test delete of Flux removes main repo dir and leaves non-main repos intact (since they could be shared)
- [x] Test delete of Qwen-Image-GGUF, then verify Qwen-Image-2512-GGUF still works (shared text_encoder + vae repos preserved)
- [x] Test new download places files in per-repo HF cache dirs

image
image
image

### DIFF
--- a/test/server_cli.py
+++ b/test/server_cli.py
@@ -20,13 +20,11 @@ Usage:
 """
 
 import argparse
-import glob
 import json
 import os
 import re
 import subprocess
 import sys
-import tempfile
 import time
 import unittest
 
@@ -34,24 +32,11 @@ import requests
 from utils.server_base import wait_for_server, set_server_config, _auth_headers
 from utils.test_models import (
     ENDPOINT_TEST_MODEL,
-    MULTI_REPO_MODEL_A_CACHE_DIR,
-    MULTI_REPO_MODEL_A_MAIN,
-    MULTI_REPO_MODEL_A_NAME,
-    MULTI_REPO_MODEL_B_CACHE_DIR,
-    MULTI_REPO_MODEL_B_MAIN,
-    MULTI_REPO_MODEL_B_NAME,
-    MULTI_REPO_SHARED_CACHE_DIR,
-    MULTI_REPO_SHARED_CHECKPOINT,
     PORT,
-    SHARED_REPO_MODEL_A_CHECKPOINT,
-    SHARED_REPO_MODEL_A_NAME,
-    SHARED_REPO_MODEL_B_CHECKPOINT,
-    SHARED_REPO_MODEL_B_NAME,
     TIMEOUT_MODEL_OPERATION,
     USER_MODEL_MAIN_CHECKPOINT,
     USER_MODEL_NAME,
     get_default_server_binary,
-    get_hf_cache_dir_candidates,
 )
 
 # Global configuration
@@ -112,68 +97,6 @@ def run_cli_command(args, timeout=60, check=False):
         )
 
     return result
-
-
-def _checkpoint_variant_path(checkpoint):
-    """Return the repo-relative variant path for a HF checkpoint string."""
-    parts = checkpoint.split(":", 1)
-    if len(parts) != 2:
-        return ""
-    return os.path.join(*parts[1].split("/"))
-
-
-def _find_cached_checkpoint(cache_root, repo_cache_dir, checkpoint):
-    """Return the on-disk snapshot path for a checkpoint, if present."""
-    variant_path = _checkpoint_variant_path(checkpoint)
-    if not variant_path:
-        return None
-
-    pattern = os.path.join(cache_root, repo_cache_dir, "snapshots", "*", variant_path)
-    matches = glob.glob(pattern)
-    if matches:
-        return matches[0]
-    return None
-
-
-def _resolve_hf_cache_root(repo_cache_dirs, checkpoint_specs=None):
-    """Pick the HF cache root that actually contains the downloaded repo artifacts."""
-    diagnostics = []
-    matches = []
-
-    for hf_cache in get_hf_cache_dir_candidates():
-        missing = [
-            repo_dir
-            for repo_dir in repo_cache_dirs
-            if not os.path.isdir(os.path.join(hf_cache, repo_dir))
-        ]
-        checkpoint_paths = []
-        if not missing and checkpoint_specs:
-            for repo_cache_dir, checkpoint in checkpoint_specs:
-                checkpoint_path = _find_cached_checkpoint(
-                    hf_cache, repo_cache_dir, checkpoint
-                )
-                if checkpoint_path is None:
-                    missing.append(f"{repo_cache_dir}:{checkpoint}")
-                else:
-                    checkpoint_paths.append(checkpoint_path)
-
-        if not missing:
-            probe_paths = [
-                os.path.join(hf_cache, repo_cache_dir)
-                for repo_cache_dir in repo_cache_dirs
-            ] + checkpoint_paths
-            newest_mtime = max(os.path.getmtime(path) for path in probe_paths)
-            matches.append((newest_mtime, hf_cache))
-            continue
-        diagnostics.append(f"{hf_cache} (missing: {', '.join(missing)})")
-
-    if matches:
-        matches.sort(reverse=True)
-        return matches[0][1]
-
-    raise AssertionError(
-        "Could not resolve HF cache root after pull. Checked: " + "; ".join(diagnostics)
-    )
 
 
 class CLITestBase(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


ianbmacdonald:
Oh no, we totally do multi-file pull from different repos for Flux-2-Klein-4B — 2 repos (text_encoder + vae share one repo) Flux-2-Klein-9B-GGUF — 3 repos but then dump them into the main repo.   Going to fix this now, and create some kind of migration option a package update can call to clean up orphans.   A user with Flux downloaded and Qwen3 8B actually has two Qwen3 8B quants .. Q4_1 in the correct repo location, and Q8_0 for flux text encoding not in the Qwen3 repo location.

src/cpp/server/recipe_options.cpp
```diff
@@ -89,6 +93,30 @@ static const json CLI_OPTIONS = {
         {"envname", "LEMONADE_HEIGHT"},
         {"help", "Image height in pixels"}
     }},
+    {"--sampling-method", {
```
bitgamma: I think this is getting rather out of hand with the amount of parameters. I think we should distinguish between:

- Image generation params: cfg, resolution, flow_shift etc. They can be part of the image generation request and can be part of the recipe option but should not be exposed as command line arguments. There is way too many of them and there isn't a good default that encompasses all models.
- Runtime options: `offload_to_cpu`, `diffusion_fa` etc. These are not really model specific and do not affect the final image. These should all be grouped into a --sdcpp-args flag/recipe option like for all other backends.

Runtime options should not be part of the ImageDefaults object since they depend on the machine more than the model itself. Probably the ImageDefaults should be replaced by having default recipe options in a generic matter, but that's for another day.

src/cpp/server/model_manager.cpp
```diff
@@ -2467,6 +2542,56 @@ void ModelManager::delete_model(const std::string& model_name) {
         LOG(INFO, "ModelManager") << "Warning: Model cache directory not found (may already be deleted)" << std::endl;
     }
 
+    // Clean up non-main checkpoint files in their own repo dirs (multi-repo models)
+    // Only delete if no other model in the registry references the same repo
+    std::string main_repo = checkpoint_to_repo_id(info.checkpoint("main"));
```
bitgamma: since you already implemented dependency check, it would be really nice if you could apply this check to  all checkpoints, including main.

The use cases are:

- registering the same model with the same main checkpoints to have different recipe options (thinking/nothinking preset for example). Now if you delete one variant the files are gone.
- using different quants of the same checkpoint: if any file in the directory is referenced by another model, the directory should not be deleted and the other quants should remain where they are.

src/cpp/server/recipe_options.cpp
ianbmacdonald: agreed; back in town;  will resolve these

bitgamma: a PR implementing --sdcpp-args has been merged yesterday, so please rebase before proceeding with the rest :slightly_smiling_face:

ianbmacdonald: Yep, see that.  Image stuff is model_recipe now;  This is going to continue to expand per model and new model features.

src/cpp/server/model_manager.cpp
ianbmacdonald: I now understand another root issue .. in windows you have to enable developer mode to use symlinks, so the entire Lemonade implementation uses direct file blobs rather than the symlink-to-blob HF architecture, which in some cases saves a lot of disk space.   So if you are a linux user, and using any other tool Comfy, etc. and/or standard hf cache download, and the model is present, we have to handle deleting the blob behind the symlink when we cleanup.  For now, this is just done inline, but I noted a TODO as when/if the cache_cleanup capabilities are wired in for non-manual execution, it would logically go there.     #1502 addresses a related issue for directory detection I discovered yesterday testing 10.0.1 and does not conflict, but is complementary.

ianbmacdonald:
Re-tested with the Q5_0 in addition to the unit tests for the tweaks based on comments
image

ianbmacdonald:
Okay, the default curl timeout is going to catch most image generation at 1024x1024 / 20 steps.   512x512 is a much more reasonable default, as image gen is ~90s on strix halo, and users can tweak from there.  timeout issue can be worked around now with the global setting per #1422.  Additionally #1421 may add more granularity, per-endpoint defaults that make more sense for image generation, or use CURLOPT_LOW_SPEED_* to base it on data flow instead. 
image

ramkrishna2910:
@ianbmacdonald Does Qwen-image support image edit? Currently only Flux supports image edit.

test/server_cli.py
```diff
@@ -419,6 +423,88 @@ def test_012_listen_all_via_runtime_config(self):
             # Best-effort restore — don't fail the test
             print(f"Warning: Failed to restore host to localhost: {e}")
 
+    def test_013_removed_image_cli_flags_rejected(self):
```
bitgamma: this test is redundant

test/server_cli.py
```diff
@@ -419,6 +423,88 @@ def test_012_listen_all_via_runtime_config(self):
...
+            result = run_cli_command(args, timeout=TIMEOUT_DEFAULT)
+            combined = result.stdout.lower() + result.stderr.lower()
+            flag = args[1]
+            self.assertTrue(
+                result.returncode != 0 or "error" in combined,
+                f"Flag {flag} should be rejected but was accepted: {combined}",
+            )
+            print(f"[OK] Removed flag {flag} correctly rejected")
+
+    def test_014_delete_preserves_shared_repo(self):
```
bitgamma: this is good, but there is also this scenario to test: 

- Model A depends on checkpoint XY and ZY. 
- Model B depends on KJ and ZY. 
- XY, ZY, KJ are are in different HF repos
- Downloading Model A must download XY and ZY. 
- Downloading Model B afterward must download only KJ and detect ZY as being there. 
- Deleting Model A must only delete XY since ZY is still needed. 
- When finally deleting Model B both KJ and ZY will be deleted

src/cpp/resources/server_models.json
```diff
@@ -1222,6 +1222,54 @@
...
+            "width": 512,
+            "height": 512,
+            "sampling_method": "euler",
+            "flow_shift": 3.0
+        },
+        "recipe_options": {
+            "sdcpp_args": "--diffusion-fa --offload-to-cpu"
+        }
+    },
+    "Qwen-Image-2512-GGUF": {
```
bitgamma, line 1299: could you also add Qwen-Image-Edit-2511?

bitgamma:
Looking good! Would like to see more comprehensive tests and then is good to go for me.

src/cpp/resources/server_models.json
ianbmacdonald: I think adding Image Edit needs to come in behind #1352, where I added a quick run book for testing a simple user case.   There is the potential for multiple file inputs, and when combined with the potential to test different quants if #1411 applies, could lead to a lot of functional test cases for sd alone, even after testing all the parameters.  I would like to see this repo fix in, before more multi-checkpoint cruft builds up from adding multi-modal models in ways not aligned with the on-disk format for HF repos.   An mvp for just getting qwen-image-edit into suggested models could be done with a separate PR easily.   As it stands the suggested Qwen-Image-2512 here will likely change to Q4_K_M when the bug is fixed in sd (I resisted changing to Q4_K_S,  as I realize this is working in the upstream bug referenced in #1412 after thorougly testing the K quants, but it is weaker than the recommended quant, so it might not do everything accurately).

bitgamma: ok, Image Edit can wait

test/server_cli.py
ianbmacdonald: gone, with a related import

test/server_cli.py
ianbmacdonald: added some more scenarios here to cover this off;  I also merged in the warm_cache fix that had its own PR (now closed #1506);  Codex review vibed in a helper to reduce drift as well on a reusable recipe-option merge and another to handle custom model dir; something I do not use, as I am HF_CACHE shared across multiple OS

ianbmacdonald: The test helper follows the HF env/default path resolution, but it does not cover config.json models_dir overrides because there is no API/env surface to query the effective value and adding a config parser to the test suite would be disproportionate. The limitation is now documented in the helper and the test assumes the normal models_dir=auto case.

ianbmacdonald:
vibed summary of the commits since review

feat/refactor (original 4 PR commits) — multi-checkpoint storage, Qwen Image models, sd-cpp recipe options, 512x512 defaults

test: address bitgamma review (726e373d) — removed redundant test_013_removed_image_cli_flags_rejected, added test_014_delete_preserves_cross_repo_dependency covering the full XY/ZY/KJ multi-repo dependency scenario

fix: align add_model_to_cache with build_cache (bb7ae87b) — warm-cache inserts now get the full 3-level recipe options merge (image_defaults -> JSON recipe_options -> user-saved), plus on-disk HF cache verification in the test

refactor: extract shared helpers (f920300e) — parse_image_defaults() and build_recipe_options() eliminate duplicated merge logic; replaced huggingface_hub.constants.HF_HUB_CACHE with get_hf_cache_dir() helper

docs: document models_dir limitation (0b6ee030) — honest docstring about the models_dir config.json gap

Still deferred: Qwen-Image-Edit-2511 (blocked on #1352 / #1411).

bitgamma:
@ianbmacdonald tests are not passing.

ianbmacdonald:
one more tweak that an agent review found 

add_model_to_cache() was already building the full merged option set via build_recipe_options() (image_defaults -> JSON recipe_options -> saved overrides), but download_model() then replaced that with only the incoming recipe_options before calling save_model_options(). That meant the import/warm-cache path could drop image_defaults and JSON-level defaults until the next restart.

bitgamma:
@ianbmacdonald much better now! one test is still failing though

bitgamma (approved):
good to go for me! @jeremyfowers should this be merged before or after the 10.1.0 release?

src/cpp/cli/lemonade_client.cpp
```diff
@@ -508,6 +508,54 @@ int LemonadeClient::delete_model(const std::string& model_name) const {
     }
 }
 
+int LemonadeClient::cleanup_cache(bool dry_run) const {
+    std::cout << (dry_run ? "Previewing" : "Running") << " cache cleanup..." << std::endl;
+
+    try {
+        json request_body = {{"dry_run", dry_run}};
+        std::string response = make_request("/api/v1/cleanup-cache", "POST",
```
jeremyfowers: Is there any reasonable scenario where an arbitrary app should run `/api/v1/cleanup-cache`?

If not, I suggest registering `/internal/cleanup-cache`, which indicates that only 1st party software should use it and protects it under LEMONADE_ADMIN_API_KEY.

Either way, needs to be documented under either server_spec.md (if it stays /api/v1) or dev-getting-started.md (if it moves to /internal).

src/cpp/cli/main.cpp
```diff
@@ -845,6 +846,7 @@ int main(int argc, char* argv[]) {
     CLI::App* unload_cmd = app.add_subcommand("unload", "Unload a model (or all models)")->group("Model management");
     CLI::App* import_cmd = app.add_subcommand("import", "Import a model from JSON file")->group("Model management");
     CLI::App* export_cmd = app.add_subcommand("export", "Export model information to JSON")->group("Model management");
+    CLI::App* cleanup_cmd = app.add_subcommand("cleanup-cache", "Clean up orphaned files in HuggingFace cache")->group("Model management");
```
jeremyfowers, line 1011:
```suggestion
    CLI::App* cleanup_cmd = app.add_subcommand("cleanup-cache", "Clean up orphaned files in Hugging Face cache")->group("Model management");
```

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -625,7 +625,7 @@ async function ensureModelReadyInternal(
```
jeremyfowers, line 625: Image

Nice!

jeremyfowers:
I'm good to merge this @ianbmacdonald except for my one main comment above.

jeremyfowers:
I want to get this in the next release, so I'm taking maintainer privilege and turning clear-cache into an /internal endpoint.

PR #1415, opened 2026-03-19
Fix a PPA FTBFS on Ubuntu noble

superm1:


kenvandine (approved):
LGTM. However it would be better with apt-get, but fine either way.

superm1:
> LGTM. However it would be better with apt-get, but fine either way.

`apt-get` supports the `build-dep .` command?  I didn't think it did.

kenvandine:
> > LGTM. However it would be better with apt-get, but fine either way.
> 
> `apt-get` supports the `build-dep .` command?  I didn't think it did.

It sure does!

PR #1416, opened 2026-03-20
Migrate Debian packages to Launchpad PPA

superm1:
Switch to native package builds

### DIFF
--- a/src/cpp/tray/tray_app.cpp
+++ b/src/cpp/tray/tray_app.cpp
@@ -224,6 +224,7 @@ void signal_handler(int signal) {
 
 // SIGCHLD handler to automatically reap zombie children
 void sigchld_handler(int signal) {
+    (void)signal;
     // Reap all zombie children without blocking
     // This prevents the router process from becoming a zombie
     int status;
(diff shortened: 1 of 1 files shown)


superm1:
Migrating to these will require changes to the runners (to enable the PPA).

```
./deb-extract/usr/bin/lemonade-router: error while loading shared libraries: libcpp-httplib.so.0.26: cannot open shared object file: No such file or directory
```

superm1:
New download page:

image

docs/assets/install-selector.js
```diff
@@ -266,6 +267,13 @@ function renderDownload() {
 
           // Render AppImage commands if App + Server selected
           if (type === 'app') {
+            const desktopPre = document.getElementById('lmn-install-desktop-block');
+            if (desktopPre) {
+              const desktopCmd = 'sudo apt install lemonade-desktop';
```
jeremyfowers, line 272: it's `lemonade-app` everywhere else after Spring Cleaning 1, possible to change it here too?

superm1: Two points.

1. But doesn't that mean electron app outside of Linux?  It's web app in Linux.
2. It's already in Ubuntu archives as `lemonade-desktop`, so it would mean needing to do a package transition if someone starts with that.  Not insurmountable, but effort.

PR #1418, opened 2026-03-20
d/lemonade-server.{postinst,postrm}: Migrate from /opt

superm1:
If a user has existing content in /opt then migrate them to the new intended home for lemonade in /var


superm1:
I view this as a pre-requisite for https://github.com/lemonade-sdk/lemonade/pull/1416

Issue #1364, opened 2026-03-14
v10.0: Large Prompts Timing Out


korgano:
**OS:** Windows 11 Pro
**Lemonade version:** 10.0

When using Lemonade as an LLM provider for another application (in this case, AnythingLLM), large prompts keep failing to generate output due to a Curl timeout:

```
2026-03-13 20:17:49.599 [Info] (Server) Model already loaded: user.Magidonia-v4.2.0
2026-03-13 20:17:49.599 [Info] (Server) POST /api/v1/chat/completions - Streaming
2026-03-13 20:17:49.663 [Info] (Process) srv  params_from_: Chat format: peg-native
2026-03-13 20:17:49.664 [Info] (Process) slot get_availabl: id  3 | task -1 | selected slot by LRU, t_last = -1
2026-03-13 20:17:49.664 [Info] (Process) slot launch_slot_: id  3 | task -1 | sampler chain: logits -> ?penalties -> ?dry -> ?top-n-sigma -> top-k -> ?typical -> top-p -> min-p -> ?xtc -> temp-ext -> dist 
2026-03-13 20:17:49.664 [Info] (Process) slot launch_slot_: id  3 | task 0 | processing task, is_child = 0
2026-03-13 20:17:49.664 [Info] (Process) slot update_slots: id  3 | task 0 | new prompt, n_ctx_slot = 32000, n_keep = 16, task.n_tokens = 15868
2026-03-13 20:17:49.664 [Info] (Process) slot update_slots: id  3 | task 0 | n_tokens = 0, memory_seq_rm [0, end)
2026-03-13 20:17:49.664 [Info] (Process) slot update_slots: id  3 | task 0 | prompt processing progress, n_tokens = 2048, batch.n_tokens = 2048, progress = 0.129065
2026-03-13 20:18:14.192 [Info] (Process) slot update_slots: id  3 | task 0 | n_tokens = 2048, memory_seq_rm [2048, end)
2026-03-13 20:18:14.192 [Info] (Process) slot update_slots: id  3 | task 0 | prompt processing progress, n_tokens = 4096, batch.n_tokens = 2048, progress = 0.258130
2026-03-13 20:18:47.901 [Info] (Process) slot update_slots: id  3 | task 0 | n_tokens = 4096, memory_seq_rm [4096, end)
[...]

PR #1421, opened 2026-03-20
Fix libcurl timeout for large prompts

james-martinez:
Addresses #1364 where large prompts timed out exactly at 5 minutes due to the global timeout overriding the default behavior.
- Changes `HttpClient` to default to `-1` to fallback to default
- Passes `0` to curl directly for infinite
- Constrains `--global-timeout` to be minimum 300 or infinite

### DIFF
--- a/src/cpp/include/lemon/utils/http_client.h
+++ b/src/cpp/include/lemon/utils/http_client.h
@@ -72,19 +72,19 @@ public:
     static HttpResponse post(const std::string& url,
                             const std::string& body,
                             const std::map<std::string, std::string>& headers = {},
-                            long timeout_seconds = 300);
+                            long timeout_seconds = -1);
 
     // Multipart form data POST request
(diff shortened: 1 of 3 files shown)


src/cpp/include/lemon/utils/http_client.h
```diff
@@ -72,19 +72,19 @@ class HttpClient {
     static HttpResponse post(const std::string& url,
                             const std::string& body,
                             const std::map<std::string, std::string>& headers = {},
-                            long timeout_seconds = 300);
+                            long timeout_seconds = -1);
```
winmutt, line 75: Why not add a configuration option per, with global as a default if not set and then a global default of 300s?

Also, under what conditions do we really even want timeouts?

james-martinez: timeouts are configurable for safety and efficiency.

PR #1422, opened 2026-03-20
fix: pass global-timeout to router and use it for sd-server proxy

ianbmacdonald:
## Summary

- **sd_server.cpp**: Replace hardcoded 600s proxy timeouts with `get_default_timeout()` so image generation/edit/variation requests respect `LEMONADE_GLOBAL_TIMEOUT`
- **ServerManager**: Forward `global_timeout` from `lemonade-server` to `lemonade-router` via `--global-timeout` CLI arg (fixes env var not propagating through subprocess spawn)
- **TrayApp**: Pass `server_config_.global_timeout` through both `start_server()` call sites

## Context

Image generation models like Qwen Image Edit can take 30+ minutes on consumer GPUs (e.g. Radeon 8060S). The sd-server proxy had hardcoded 600s timeouts that couldn't be overridden, and `LEMONADE_GLOBAL_TIMEOUT` set in `/etc/lemonade/conf.d/` was never reaching the router because `ServerManager` didn't pass it as an arg.

Complementary to #1421 which fixes the HttpClient timeout semantics — this PR ensures the global timeout value actually reaches the router and is used by the sd-server proxy.

## Test plan

- [x] Set `LEMONADE_GLOBAL_TIMEOUT=3600` in `/etc/lemonade/conf.d/timeout.conf`
- [x] Verify `--global-timeout 3600` appears in router process args
- [x] Image edit request completes after 35m46s without timeout
- [ ] Verify default behavior (no env override) still works with 300s timeout
- [ ] Verify existing image generation models (Flux, SDXL) unaffected

### DIFF
--- a/src/cpp/include/lemon_tray/server_manager.h
+++ b/src/cpp/include/lemon_tray/server_manager.h
@@ -52,7 +52,8 @@ public:
         bool is_ephemeral,
         const std::string& host,
         int max_loaded_models,
-        const std::string& extra_models_dir
+        const std::string& extra_models_dir,
+        long global_timeout = 300
     );
 
     bool stop_server();
@@ -112,6 +113,7 @@ private:
     std::string api_key_;
     int port_;
     int max_loaded_models_;
+    long global_timeout_ = 300;
     nlohmann::json recipe_options_;
     bool show_console_;
     bool is_ephemeral_;  // Suppress output for ephemeral servers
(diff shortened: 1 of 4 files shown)


ianbmacdonald:
Probably a  better approach, but this takes the hardcoded default which breaks long image edit prompts via the router, and wires it up to the existing global timeout ENV when it is defined.

PR #1423, opened 2026-03-20
Use llama.cpp's Anthropic API implementation

sawansri:
This PR switches Anthropic API support from the native Messages API routes I implemented in #1168 to llama.cpp's recently added implementation. Using llama.cpp's Anthropic API ensures that support is upstreamed, reducing maintenance burden.

Also includes tests (mainly covering the translation lemonade does before handing off to llama.cpp plus some basic sanity checks). Llama.cpp already has a comprehensive test suite, so we don't need to replicate that here.

Additionally, it enables Messages API passthrough to vLLM (which similar to llamacpp implements the api already).


sawansri:
@claude review

sawansri:
@claude can you confirm pushed changes address your review and don't create more issues

sawansri:
Closing in favor of #2187

PR #1434, opened 2026-03-23
Add missing deps to build environment container

superm1:
These are used by some test jobs too


superm1:
Here is the test run showing it working. 

https://github.com/lemonade-sdk/lemonade/actions/runs/23454854512/job/68241225501

Issue #1442, opened 2026-03-25
Lemonade UI cannot display logs from remote server


ziouf:
### Description

I'm running lemonade-server on my Strix Halo running ubuntu server. 
And using lemonade UI from snap package on my laptop.

Since 10.0.1 lemonade UI is not able to display server logs anymore in logs panel. Server logs shows that `/api/v1/logs/stream` does not exists.

Http get to this endpoint shows the following :
```
{"error":"Log file not found. Log streaming requires server to be launched via tray application.","note":"When running directly, logs appear in console instead.","path":""}
```

### UI logs

```
Failed to connect to lemonade-server logs.
Make sure the server is running on https://<my-domain>
```

### Server logs

```
Mar 25 09:04:52 llm lemonade-server[4435]: 2026-03-25 09:04:52.898 [Debug] (Server) GET /api/v1/logs/stream
Mar 25 09:04:52 llm lemonade-server[4435]: 2026-03-25 09:04:52.898 [Error] (Server) Log file not found:
Mar 25 09:04:53 llm lemonade-server[4435]: 2026-03-25 09:04:52.921 [Error] (Server) Note: Log streaming only works when server is launched via tray/ServerManager
Mar 25 09:04:53 llm lemonade-server[4435]: 2026-03-25 09:04:52.921 [Error] (Server) Error 404: GET /api/v1/logs/stream
Mar 25 09:04:53 llm lemonade-server[4435]: 2026-03-25 09:04:52.922 [Debug] (Server) GET /api/v1/logs/stream - 404
```

sandivuk:
Just wanted to chime in that I am having the same issue as of this latest version

jeremyfowers:
https://github.com/lemonade-sdk/lemonade/pull/1453 fixes the problem if you are building your own `lemond` from source and run it manually. More work is needed to ensure this works when running `lemond` as a snap or systemd.

jeremyfowers:
Oops didn't mean to close this.

Current status: we're still trying to figure out a way to get this working fully with the web app built into the minimal server on linux.

In the meantime, the workaround is to the use AppImage, which is built with a more robust set of deps.

sandivuk:
> Oops didn't mean to close this.
> 
> Current status: we're still trying to figure out a way to get this working fully with the web app built into the minimal server on linux.
> 
> In the meantime, the workaround is to the use AppImage, which is built with a more robust set of deps.

[...]

PR #1453, opened 2026-03-25
Fix direct-run log streaming for lemonade-router

jeremyfowers:
Partial progress towards #1442 

## Summary
- write direct lemonade-router logs to the runtime log file as well as stdout
- fix /api/v1/logs/stream for direct runs, including --host 0.0.0.0 repros
- remove the incorrect tray-only log streaming message and add endpoint coverage

## Verification
- cmake --build --preset default --target lemonade-router
- run ./build/lemonade-router --host 0.0.0.0 --port 8011
- curl -iN http://127.0.0.1:8011/api/v1/logs/stream and confirm 200 OK with SSE data


src/cpp/server/main.cpp
```diff
@@ -54,9 +55,17 @@ int main(int argc, char** argv) {
...
-        AixLog::Log::init({sink});
+        // Direct router runs should keep console logs while also writing a file
+        // that the SSE log-stream endpoint can tail.
+        auto filter = AixLog::Filter(AixLog::to_severity(config.log_level));
+        auto console_sink = std::make_shared<AixLog::SinkCout>(filter, RuntimeConfig::LOG_FORMAT);
+#ifdef _WIN32
+        AixLog::Log::init({console_sink});
+#else
+        std::string log_file = utils::get_runtime_dir() + "/lemonade-server.log";
+        auto file_sink = std::make_shared<AixLog::SinkFile>(filter, log_file, RuntimeConfig::LOG_FORMAT);
```
superm1: what about systemd?  it will never use this

jeremyfowers: Working on that now (see discord message)

superm1: Besides the missing build dep though, I don't believe `file_sink` should be initialized when launched with systemd

jeremyfowers: You and codex both!

jeremyfowers: Look better now?

src/app/src/renderer/LogsWindow.tsx
```diff
@@ -120,17 +121,29 @@ const LogsWindow: React.FC<LogsWindowProps> = ({ isVisible, height }) => {
...
-
-        const eventSource = new EventSource(`${serverUrl}/api/v1/logs/stream`, options)
+        const streamUrl = `${serverUrl}/api/v1/logs/stream`;
+        const isWebApp = window.api?.isWebApp === true;
+
+        // The browser-served web app should use the native EventSource.
+        // The npm polyfill is for non-browser environments and is not
+        // constructable from the webpacked web bundle.
+        const eventSource = isWebApp
+          ? new window.EventSource(streamUrl)
```
bitgamma: this breaks API key support.

bitgamma: I am also pretty sure that the library used now is not a polyfill but rather an extension to support custom headers (bearer authentication) and it works fine in the webapp. I am not sure the changes to this file are necessary at all

superm1 (approved):

src/app/src/renderer/LogsWindow.tsx
```diff
@@ -120,17 +120,23 @@ const LogsWindow: React.FC<LogsWindowProps> = ({ isVisible, height }) => {
...
-              fetch(input, {
-                ...init,
-                headers: {
-                  ...init.headers,
-                  Authorization: `Bearer ${apiKey}`,
-                },
-            })} : {};
-
-        const eventSource = new EventSource(`${serverUrl}/api/v1/logs/stream`, options)
+        const streamUrl = `${serverUrl}/api/v1/logs/stream`;
```
bitgamma: better but I think the better approach is to just revert all changes to this file since they behaviorually do not change anything at all while renaming the event source into eventsourcepolyfill, which is a misleading name since it is not a polyfill

jeremyfowers:
This PR doesn't fix everything, but I'm going to merge it because it's better than not doing anything.

PR #1460, opened 2026-03-26
Add support for ROCm channels (For llama.cpp and stable diffusion)

superm1:
The ROCm stable runtime utilizes the AMD released  ROCm runtimes (stable and preview) and the upstream llama.cpp artifacts.

### DIFF
--- a/docs/llamacpp.md
+++ b/docs/llamacpp.md
@@ -1,0 +1,174 @@
+# llama.cpp Backend Options
+
+Lemonade uses [llama.cpp](https://github.com/ggerganov/llama.cpp) as its primary LLM inference backend, supporting multiple hardware acceleration options. This document explains the available backends and how to choose between them.
+
+## Available Backends
+
+### CPU
+- **Platform**: Windows, Linux, macOS
+- **Hardware**: All x86_64 processors
+- **Use Case**: Universal fallback, no GPU required
+- **Performance**: Slowest option, suitable for small models or testing
+- **Installation**: Automatically available via upstream llama.cpp releases
+
+### Vulkan
+- **Platform**: Windows, Linux
+- **Hardware**: AMD GPUs (iGPU and dGPU), NVIDIA GPUs, Intel GPUs
+- **Use Case**: Cross-vendor GPU acceleration
+- **Performance**: Good performance across all GPU vendors
+- **Installation**: Automatically available via upstream llama.cpp releases
+- **Notes**: Recommended for most GPU users
+
+### ROCm
+- **Platform**: Windows, Linux
+- **Hardware**: AMD Radeon RX 6000/7000 series (RDNA2/RDNA3/RDNA4), AMD Ryzen AI iGPUs (Strix Point/Halo)
+- **Use Case**: AMD GPU-optimized inference
+- **Performance**: Optimized for AMD hardware, may outperform Vulkan on supported GPUs
+- **Channel Options**:
+  - **Preview** (default): Custom builds with latest optimizations from lemonade-sdk
+  - **Stable**: Upstream llama.cpp releases with AMD ROCm support
+  - **Nightly**: Bleeding-edge builds from lemonade-sdk/llamacpp-rocm (experimental)
+- **Installation**: Varies by channel (see below)
+
+### Metal
+- **Platform**: macOS only
+- **Hardware**: Apple Silicon (M1/M2/M3/M4) and Intel Macs with Metal support
+- **Use Case**: macOS GPU acceleration
+- **Performance**: Optimized for Apple Silicon
+- **Installation**: Automatically available via upstream llama.cpp releases
+
+### System
+- **Platform**: Linux only
+- **Hardware**: Depends on system-installed llama-server binary
+- **Use Case**: Advanced users with custom llama.cpp builds
+- **Performance**: Depends on build configuration
+- **Installation**: Requires manual installation of `llama-server` in system PATH
+- **Notes**: Not enabled by default; set `LEMONADE_LLAMACPP_PREFER_SYSTEM=true` in config
+
+## ROCm Channel Configuration
+
+The ROCm backend supports three channels to balance stability, performance, and access to latest features:
+
+### Preview Channel (Default)
+```json
+{
+  "rocm_channel": "preview"
+}
+```
+- **Source**: Custom builds from [lemonade-sdk/llama.cpp](https://github.com/lemonade-sdk/llama.cpp)
+- **Binaries**: Common builds for supported architectures
+- **Updates**: Frequent updates with latest optimizations and fixes
+- **Platform**: Windows and Linux
+- **Runtime**: Requires runtime for both Windows and Linux to be installed separately.
+- **Best For**: Users who want the latest performance optimizations
+
+### Stable Channel
+```json
+{
+  "rocm_channel": "stable"
+}
+```
+- **Source**: Upstream [llama.cpp](https://github.com/ggerganov/llama.cpp) releases
+- **Binaries**:
+  - **Windows**: Self-contained HIP binaries (no separate runtime needed)
+  - **Linux**: Binaries built against ROCm 7.2 runtime
+- **Updates**: Follows upstream llama.cpp release cycle
+- **Platform**: Windows and Linux
+- **Runtime**:
+  - Windows: Self-contained, no runtime installation required
+  - Linux: Downloads AMD ROCm 7.2.1 runtime if not present at `/opt/rocm`
+- **Best For**: Users who prefer stable, tested releases aligned with upstream
+
+### Nightly Channel
+```json
+{
+  "rocm_channel": "nightly"
+}
+```
+- **Source**: Nightly builds from [lemonade-sdk/llamacpp-rocm](https://github.com/lemonade-sdk/llamacpp-rocm)
+- **Binaries**: Architecture-specific builds (gfx1150, gfx1151, gfx103X, gfx110X, gfx120X)
+- **Updates**: Nightly builds with experimental features and latest upstream changes
+- **Platform**: Windows and Linux
+- **Runtime**: Bundled runtime on Linux, TheRock ROCm dependencies
+- **Best For**: Developers and testers who want bleeding-edge features and are comfortable with potential instability
+
+### Changing Channels
+
+To switch between channels, update your `config.json`:
+
+```json
+{
+  "rocm_channel": "stable"
+}
+```
+
+Or use the Lemonade CLI:
+```bash
+# Switch to stable channel
+lemonade config set rocm_channel stable
+
+# Switch to preview channel (default)
+lemonade config set rocm_channel preview
+
+# Switch to nightly channel (experimental)
+lemonade config set rocm_channel nightly
+```
+
+After changing channels, you'll need to reinstall the ROCm backend:
+```bash
+lemonade backend install llamacpp rocm
+```
+
+## Choosing the Right Backend
+
+### Decision Tree
+
+1. **Do you have an NVIDIA or Intel GPU?**
+   - Use **Vulkan**
+
+2. **Do you have an AMD GPU?**
+   - **For Radeon RX 6000/7000 or Ryzen AI iGPU**:
+     - Try **ROCm** first for best performance
+     - Fall back to **Vulkan** if you encounter issues
+   - **For older AMD GPUs (RX 5000 and earlier)**:
+     - Use **Vulkan** (ROCm not supported)
+
+3. **Do you have Apple Silicon?**
+   - Use **Metal**
+
+4. **No GPU or unsupported GPU?**
+   - Use **CPU**
+
+### ROCm Channel Selection
+
+- **Use Preview** if you:
+  - Want the best performance on AMD hardware
+  - Are comfortable with frequent updates
+  - Are testing new models or features
+
+- **Use Stable** if you:
+  - Prefer stability over latest features
+  - Want upstream llama.cpp compatibility
+  - Are deploying in production
+
+- **Use Nightly** if you:
+  - Want bleeding-edge experimental features
+  - Are testing unreleased llama.cpp functionality
+  - Are comfortable with potential bugs and instability
+  - Are a developer contributing to lemonade or llama.cpp
+
+## Platform Specifics
+
+### Linux
+- All backends supported (CPU, Vulkan, ROCm, System)
+- ROCm requires compatible AMD GPU (see above)
+- System backend requires manual llama-server installation
+
+### Windows
+- Supported: CPU, Vulkan, ROCm
+- ROCm requires compatible AMD GPU
+- No system backend support
+
+### macOS
+- Supported: CPU, Metal
+- Metal recommended for all Macs with Metal support
--- a/docs/llamacpp.md
+++ b/docs/llamacpp.md
@@ -105,18 +105,18 @@ To switch between channels, update your `config.json`:
 Or use the Lemonade CLI:
 ```bash
 # Switch to stable channel
-lemonade config set rocm_channel stable
+lemonade config set rocm_channel=stable
 
 # Switch to preview channel (default)
-lemonade config set rocm_channel preview
+lemonade config set rocm_channel=preview
 
 # Switch to nightly channel (experimental)
-lemonade config set rocm_channel nightly
+lemonade config set rocm_channel=nightly
 ```
 
 After changing channels, you'll need to reinstall the ROCm backend:
 ```bash
-lemonade backend install llamacpp rocm
+lemonade backend install llamacpp:rocm
 ```
 
 ## Choosing the Right Backend
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -1,8 +1,11 @@
 {
   "comment": "This configuration file controls which llama.cpp, whisper.cpp, sd.cpp, ryzenai-llm, and FLM versions are downloaded for each backend. You can modify these values to pin specific versions without rebuilding the application.",
+  "rocm-stable-runtime": "v7.2.1",
   "llamacpp": {
     "vulkan": "b8766",
-    "rocm": "b1238",
+    "rocm-stable": "b8653",
+    "rocm-preview": "b8705",
+    "rocm-nightly": "b1238",
     "metal": "b8460",
     "cpu": "b8766"
   },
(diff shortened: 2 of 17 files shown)


src/cpp/resources/backend_versions.json
```diff
@@ -2,7 +2,9 @@
   "comment": "This configuration file controls which llama.cpp, whisper.cpp, sd.cpp, ryzenai-llm, and FLM versions are downloaded for each backend. You can modify these values to pin specific versions without rebuilding the application.",
   "llamacpp": {
     "vulkan": "b8460",
-    "rocm": "b1217",
+    "rocm-stable": "b8532",
+    "rocm-preview": "b1217",
+    "rocm-stable-runtime": "v7.2",
```
jeremyfowers: This is going to get confusing, since many people previously running rocm didnt know it was a preview. 

What about `lemonade-rocm` vs. `ggml-org-rocm`? Or any other way of making it more clear what they are?

docs/server/configuration.md
```diff
@@ -69,7 +69,9 @@ You can provide your own `llama-server`, `whisper-server`, or `ryzenai-server` b
 
 | Environment Variable | Description |
 |---------------------|-------------|
-| `LEMONADE_LLAMACPP_ROCM_BIN` | Path to custom `llama-server` binary for ROCm backend |
+| `LEMONADE_LLAMACPP_ROCM_STABLE_BIN` | Path to custom `llama-server` binary for ROCm stable backend |
+| `LEMONADE_LLAMACPP_ROCM_PREVIEW_BIN` | Path to custom `llama-server` binary for ROCm preview backend |
+| `LEMONADE_LLAMACPP_ROCM_BIN` | Legacy alias for `LEMONADE_LLAMACPP_ROCM_PREVIEW_BIN` |
```
jeremyfowers: This is a lot of merge conflicts with Spring Cleaning 2, how did you want to handle that?

I definitely don't want to introduce new env vars in one PR and then take them away in the next PR, that will make migration that much more complex.

docs/lemonade-cli.md
```diff
@@ -280,6 +280,8 @@ The following options are available depending on the recipe being used:
 | `--llamacpp BACKEND` | LlamaCpp backend to use | Auto-detected |
 | `--llamacpp-args ARGS` | Custom arguments to pass to llama-server (must not conflict with managed args) | `""` |
 
+Available `--llamacpp` backends include `system`, `rocm-stable`, `rocm-preview`, `vulkan`, `metal`, and `cpu` (`rocm` remains a compatibility alias for `rocm-preview`).
```
jeremyfowers: I had deliberately left this kind of info out of the doc since it was one more point of maintenance. 

Speaking of which, if we dont think of an elegant way to simplify this we might need a whole doc to help people understand the difference between all these options.

src/cpp/server/backend_manager.cpp
```diff
@@ -2,18 +2,169 @@
...
+void install_rocm_stable_runtime_if_needed(const backends::BackendSpec& spec,
+                                           const json& backend_versions,
+                                           DownloadProgressCallback progress_cb) {
+    if (is_rocm_runtime_available()) {
+        LOG(DEBUG, "BackendManager") << "System ROCm runtime detected; skipping rocm-stable runtime install" << std::endl;
+        return;
+    }
+
+    const std::string version = get_rocm_stable_runtime_version(backend_versions);
+    const std::string repo = "lemonade-sdk/rocm-stable";
```
jeremyfowers, line 156: TIL you're maintaining a portable build of ROCm for distribution!

src/cpp/server/backend_manager.cpp
```diff
@@ -2,18 +2,169 @@
...
+    if (recipe == "llamacpp" && backend == "rocm") {
+        return "rocm-preview";
+    }
+    return backend;
+}
+
+bool is_rocm_runtime_available() {
+#ifdef _WIN32
+    return !utils::find_executable_in_path("amdhip64.dll").empty();
+#else
```
jeremyfowers: I thought these ifdef's were banned ;)

src/cpp/server/backend_manager.cpp
```diff
@@ -2,18 +2,169 @@
...
+
+std::string get_rocm_stable_runtime_version(const json& backend_versions) {
+    if (backend_versions.contains("llamacpp") &&
+        backend_versions["llamacpp"].is_object() &&
+        backend_versions["llamacpp"].contains("rocm-stable-runtime") &&
+        backend_versions["llamacpp"]["rocm-stable-runtime"].is_string()) {
+        return backend_versions["llamacpp"]["rocm-stable-runtime"].get<std::string>();
+    }
+
+    return "v7.2";
```
jeremyfowers: Do not like hardcoded values. Is there really no way to read the rocm version back from the artifacts? Even if its just putting a version.txt into your rocm-stable builds?

src/cpp/server/backend_manager.cpp
```diff
@@ -2,18 +2,169 @@
...
+}
+
+std::string get_rocm_stable_runtime_asset_filename(const std::string& version) {
+#ifdef __linux__
+    return "rocm-7.2-runtime-libs.tar.gz";
+#else
+    throw std::runtime_error("ROCm stable runtime artifacts are currently supported on Linux only");
+#endif
+}
```
jeremyfowers, line 143: Can we avoid hardcoding this?

src/cpp/server/backend_manager.cpp
```diff
@@ -2,18 +2,169 @@
...
+
+void install_rocm_stable_runtime_if_needed(const backends::BackendSpec& spec,
+                                           const json& backend_versions,
+                                           DownloadProgressCallback progress_cb) {
+    if (is_rocm_runtime_available()) {
+        LOG(DEBUG, "BackendManager") << "System ROCm runtime detected; skipping rocm-stable runtime install" << std::endl;
+        return;
+    }
```
jeremyfowers: Do we really support system rocm this broadly? What if someone has system rocm v6.2 installed, and they have a gfx1151? It doesn't look like we're doing any HW compatibility checks against the system rocm, or rocm version checks against the llamacpp artifact.

src/cpp/server/system_info.cpp
```diff
@@ -116,23 +116,27 @@ struct RecipeBackendDef {
...
+    }},
+    {"llamacpp", "rocm-preview", {"windows", "linux"}, {
+        {"amd_igpu", {"gfx1150", "gfx1151"}},                      // STX Point/Halo iGPUs (explicit binaries)
+        {"amd_dgpu", {"gfx103X", "gfx110X", "gfx120X"}},          // RDNA2/3/4 dGPUs (family binaries)
+    }},
     {"llamacpp", "vulkan", {"windows", "linux"}, {
         {"cpu", {"x86_64"}},
         {"amd_igpu", {}},      // all iGPU families
         {"amd_dgpu", {}},      // all dGPU families
     }},
```
jeremyfowers: This changes the default from vulkan to rocm. It also changes the default from TheRock to ROCm 7.2 on Linux, but not on Windows. 

Brain dumping some issues that will come up:

1. A lot of people use lemonade specifically because they prefer TheRock over 7.2 so this change could lead to significant confusion.
2. Putting different ROCm builds as the default on Windows vs. Linux will split the community and break the abstraction we have that Lemonade is Lemonade regardless of the OS.
3. TheRock builds have been under heavy use by Lemonade users for months, and I feel confident that it is serving them well. I don't have any data about the upstream llamacpp builds across the range of models and scenarios that Lemonade users care about. How do we know the upstream is more "stable" than the "preview"?
4. Vulkan artifacts are much smaller than rocm artifacts, so changing the default away from vulkan adds noticeable setup friction for new users.
5. ROCm requires the dkms patch on gfx1151, while Vulkan does not, which is a large amount of setup friction to set as the default.

src/cpp/server/system_info.cpp
```diff
@@ -116,23 +116,27 @@ struct RecipeBackendDef {
...
-    // llamacpp with multiple backends (order = preference: system > metal > vulkan > rocm > cpu)
+    // llamacpp with multiple backends (order = preference)
     {"llamacpp", "system", {"linux"}, {
         {"cpu", {"x86_64"}}, // Placeholder, actual check is PATH-based
     }},
     {"llamacpp", "metal", {"macos"},
     {
         {"metal", {}},
     }},
+    {"llamacpp", "rocm-stable", {"linux"}, {
```
jeremyfowers: It doesn't look like this PR puts rocm-stable into CI explicitly (although I suppose its implicit anywhere a backend is not explicitly specified)?

Also regarding CI: I have no idea whether system rocm is installed on some or all CI machines, nor what version of system rocm is on each machine (they may all be different). This PR, as coded, makes system ROCm the default, which adds significant non-determinism into CI unless care is taken to normalize all the runners.

src/cpp/resources/backend_versions.json
superm1: well eventually "ggml-org-rocm" is going to have ROCm artifacts released that are not against the ROCm stable release, but they're waiting for TheRock to no longer be a tech preview.

TBH - I think it's a mistake that the artifacts used by lemonade are tech preview quality artifacts but that's not clear to users.  They should know what they're getting themselves into by using them vs production quality.

docs/server/configuration.md
superm1: Merge your branch first, I'll rebase.

docs/lemonade-cli.md
superm1: Sure, I can write up some docs about it once we have moar alignment.

src/cpp/server/backend_manager.cpp
superm1: Built specifically for this purpose :yum:

src/cpp/server/backend_manager.cpp
superm1: yes, they should be dropped, thanks for catching one.  Will after I rebase on your other branch.

src/cpp/server/backend_manager.cpp
superm1: Well we need to also think about what happens when ROCm 7.12 (and later) eventually stop being tech preview and are stable releases.  We'll be pointing at releases they provide, and I don't think they have a way to tell what version they came from.

src/cpp/server/backend_manager.cpp
superm1: sure

src/cpp/server/backend_manager.cpp
superm1: I guess I'll need to see if I can come up with a way to sanity check the system ROCm version indeed.

src/cpp/server/system_info.cpp
superm1: > This changes the default from vulkan to rocm. It also changes the default from TheRock to ROCm 7.2 on Linux, but not on Windows.

Ah it hadn't occurred to me that it would change priorities - but you're right.  I'll re-order.  I wanted to get this stuff plumbed so we can have all the facts on policy decisions.

> Putting different ROCm builds as the default on Windows vs. Linux will split the community and break the abstraction we have that Lemonade is Lemonade regardless of the OS.

Well actually I could get Windows lit up, but it works differently and I might need help.  There is an upstream llama.cpp artifact that is created for HIP that works on Windows.

> TheRock builds have been under heavy use by Lemonade users for months, and I feel confident that it is serving them well. I don't have any data about the upstream llamacpp builds across the range of models and scenarios that Lemonade users care about. How do we know the upstream is more "stable" than the "preview"?

We don't.  But implicitly that there is a team testing and signing off the ROCm release gives me more confidence.

> Vulkan artifacts are much smaller than rocm artifacts, so changing the default away from vulkan adds noticeable setup friction for new users.

Right; policy change should be separate and weighed out.

> ROCm requires the dkms patch on gfx1151, while Vulkan does not, which is a large amount of setup friction to set as the default.

DKMS?  No we don't need DKMS on gfx1151.  We do need an updated kernel, but FWIW it's rolled out now in both OEM and HWE kernels in Ubuntu.

src/cpp/server/system_info.cpp
superm1: I'd like to revisit CI after all the other comments are solved and we're generally agreeable.  We do need CI coverage in some fashion.

jeremyfowers (changes requested):
Gently blocking PR until policy questions are resolved.

src/cpp/server/backend_manager.cpp
superm1: Found a way to check it!  Added to PR.

src/cpp/server/system_info.cpp
superm1: Added a CI job, we'll see if it works.

superm1:
I've made adjustments.  

1. We now have 3 ROCm channels.

* stable: latest stable ROCm runtime from AMD (Linux) or latest HIP SDK release from AMD (Windows) + upstream llama.cpp artifacts
* preview: latest preview ROCm runtime from AMD for Windows **and** Linux + llama.cpp artifacts from https://github.com/lemonade-sdk/llama.cpp
* nightly: nightly ROCm release from AMD combined with latest llama.cpp (from https://github.com/lemonade-sdk/llamacpp-rocm)

2. Stable diffusion is now here too.

This needs more testing that I didn't break anything.

Geramy:
> I've made adjustments.
> 
> 1. We now have 3 ROCm channels.
> 
> * stable: latest stable ROCm runtime from AMD (Linux) or latest HIP SDK release from AMD (Windows) + upstream llama.cpp artifacts
> * preview: latest preview ROCm runtime from AMD for Windows **and** Linux + llama.cpp artifacts from https://github.com/lemonade-sdk/llama.cpp
> * nightly: nightly ROCm release from AMD combined with latest llama.cpp (from https://github.com/lemonade-sdk/llamacpp-rocm)
> 
> 2. Stable diffusion is now here too.
> 
> This needs more testing that I didn't break anything.

Perfect, I will try to take the same path for my mlx engine and mlx port then.

ramkrishna2910:
@claude review

jeremyfowers (approved):
Provided the policy/defaults I want to see offline. @ramkrishna2910 to provide thorough review.

Geramy (approved):
Looks good to me, I dont see any inefficiencies 100% optimal to me.

src/cpp/resources/defaults.json
```diff
@@ -13,6 +13,7 @@
   "offline": false,
   "disable_model_filtering": false,
   "enable_dgpu_gtt": false,
+  "rocm_channel": "preview",
```
ramkrishna2910, line 17: @superm1 this should still be `nightly` in this PR right? To preserve the existing defaults as is and move to preview in a follow up PR.

docs/llamacpp.md
```diff
@@ -0,0 +1,174 @@
...
+```json
+{
+  "rocm_channel": "nightly"
+}
+```
+- **Source**: Nightly builds from [lemonade-sdk/llamacpp-rocm](https://github.com/lemonade-sdk/llamacpp-rocm)
+- **Binaries**: Architecture-specific builds (gfx1150, gfx1151, gfx103X, gfx110X, gfx120X)
+- **Updates**: Nightly builds with experimental features and latest upstream changes
+- **Platform**: Windows and Linux
+- **Runtime**: Bbundled runtime on Linux, TheRock ROCm dependencies
```
ramkrishna2910:
```suggestion
- **Runtime**: Bundled runtime on Linux, TheRock ROCm dependencies
```

src/cpp/resources/defaults.json
superm1: I would prefer to make the change atomically and then leave nightly only for performance testing/issue regression etc.

superm1: This is because SD doesn't work on nightly.  Due to how nightly builds I don't see a way to solve that.

docs/llamacpp.md
superm1: fixed

docs/llamacpp.md
```diff
@@ -0,0 +1,174 @@
...
+```json
+{
+  "rocm_channel": "stable"
+}
+```
+
+Or use the Lemonade CLI:
+```bash
+# Switch to stable channel
+lemonade config set rocm_channel stable
```
ramkrishna2910: CLI syntax is key=value

```suggestion
lemonade config set rocm_channel=stable
```

docs/llamacpp.md
```diff
@@ -0,0 +1,174 @@
...
+
+# Switch to preview channel (default)
+lemonade config set rocm_channel preview
```
ramkrishna2910:
```suggestion
lemonade config set rocm_channel=preview
```

docs/llamacpp.md
```diff
@@ -0,0 +1,174 @@
...
+
+# Switch to nightly channel (experimental)
+lemonade config set rocm_channel nightly
```
ramkrishna2910:
```suggestion
lemonade config set rocm_channel=nightly
```

docs/llamacpp.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+After changing channels, you'll need to reinstall the ROCm backend:
+```bash
+lemonade backend install llamacpp rocm
```
ramkrishna2910:
```suggestion
lemonade backend install llamacpp:rocm
```

src/cpp/server/backends/sd_server.cpp
```diff
@@ -37,23 +100,36 @@ InstallParams SDServer::get_install_params(const std::string& backend, const std
...
             throw std::runtime_error(
                 SystemInfo::get_unsupported_backend_error("sd-cpp", "rocm")
             );
         }
 #ifdef _WIN32
-        params.filename = "sd-" + short_version + "-bin-win-rocm-x64.zip";
+        if (resolved_backend == "rocm-stable") {
+            // Windows stable diffusion artifacts are currently built against ROCm 7.1.1
+            // and don't follow the rocm-stable-runtime version from backend_versions.json
+            params.filename = "sd-" + short_version + "-bin-win-rocm-7.1.1-x64.zip";
```
ramkrishna2910, line 115: Should 7.1.1 be hardcoded here?

superm1: Yes until this changes either to newer HIP SDK or to Windows using The Rock for stable. It's not configurable separate run-time.

ramkrishna2910 (approved):

PR #1466, opened 2026-03-26
feat: add openSUSE Tumbleweed support

kurtseebauer:
Add zypper package manager detection and openSUSE Tumbleweed package name mappings to setup.sh for all dependency checks: build tools, required libraries, optional tray dependencies, and the install flow.

Add openSUSE Tumbleweed to the Linux distro CI build matrix so it is tested on every PR alongside Arch, Debian, and Fedora.

Tested on openSUSE Tumbleweed with GCC 15.2.1 — lemonade builds and runs without issues.

### DIFF
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -34,6 +34,9 @@ jobs:
           - distro: Fedora
             image: fedora:latest
             python: python3
+          - distro: openSUSE
+            image: opensuse/tumbleweed:latest
+            python: python3
 
     steps:
       - name: Checkout code
--- a/setup.sh
+++ b/setup.sh
@@ -47,7 +47,7 @@ maybe_sudo() {
 }
 
 # Detect OS
-if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+if [[ "$OSTYPE" == "linux"* ]]; then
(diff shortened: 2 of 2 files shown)


superm1:
Makes sense - please work through the CI failures.  If the cpack generated RPM is usable for OpenSUSE then maybe we should also have documentation pages updated too.

kurtseebauer:
> Makes sense - please work through the CI failures. If the cpack generated RPM is usable for OpenSUSE then maybe we should also have documentation pages updated too.

CI works now, but I'm having trouble with the RPM package build process, and TBH I don't really know how things work. Claude Code fixed a few things. I _think_ I can provide a working version over the weekend, but no guarantees that what I'm doing is the right way.

superm1:
> > Makes sense - please work through the CI failures. If the cpack generated RPM is usable for OpenSUSE then maybe we should also have documentation pages updated too.
> 
> CI works now, but I'm having trouble with the RPM package build process, and TBH I don't really know how things work. Claude Code fixed a few things. I _think_ I can provide a working version over the weekend, but no guarantees that what I'm doing is the right way. 
> 
> 

Well I would say what we have right now is not ideal. Ideally we would have a spec file that generates real distro compliant rpms that are unable on fedora and opensuse.

PR #1467, opened 2026-03-26
Clarify message for unsupported AMD GPUs

superm1:
Currently when a user tries to use ROCm backend on an unsupported system they get a message indicating they need a specific GPU, but there are two problems.

1) This is flat out wrong.  There are lots of GPUs that can work just fine. 2) If they're on a laptop, telling them to go get a dGPU isn't going to be an option.

Instead just KISS - tell them there wasn't a supported GPU on the system.

### DIFF
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -67,7 +67,7 @@ jobs:
         run: |
           set -e
 
-          echo "Building lemond and lemonade-server on ${{ matrix.distro }}..."
+          echo "Building lemond and lemonade on ${{ matrix.distro }}..."
 
           # Build
           echo "Building binaries..."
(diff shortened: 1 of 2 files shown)


superm1:
I encountered this specifically when using a Krackan laptop that we don't currently advertise support (GFX1152).  I was told to go get myself a dGPU:

> Requires Radeon RX 6000 series (RDNA2) AMD dGPU

src/cpp/server/system_info.cpp
```diff
@@ -859,25 +859,22 @@ json SystemInfo::build_recipes_info(const json& devices) {
...
-                    // Show specific family requirement (e.g., "Requires XDNA2 NPU")
+            } else if (!missing_devices.empty() || !wrong_family.empty()) {
+                // Get the first unsupported device (prefer missing over wrong family)
+                const auto& [device_type, required_families] = !missing_devices.empty()
+                    ? missing_devices[0]
+                    : wrong_family[0];
+
+                // For AMD GPUs, just say "Unsupported GPU" regardless of the specific issue
+                if (device_type == "amd_dgpu" || device_type == "amd_igpu") {
+                    message = "Unsupported GPU";
```
jeremyfowers: Can you print the detected family, like "Unsupported GPU: gfx900"

That would help a lot when people copy paste their `lemonade recipes` output for troubleshooting.

PR #1471, opened 2026-03-26
FLM embed binary

ramkrishna2910:


src/cpp/server/backends/fastflowlm_server.cpp
```diff
@@ -42,6 +28,23 @@ FastFlowLMServer::~FastFlowLMServer() {
...
+    // Version in backend_versions.json has 'v' prefix (e.g. "v0.9.37"),
+    // but the release filename uses bare version (e.g. "0.9.37")
+    std::string bare_version = version;
+    if (!bare_version.empty() && bare_version[0] == 'v') {
+        bare_version = bare_version.substr(1);
+    }
+#ifdef _WIN32
+    params.filename = "fastflowlm-windows-" + bare_version + ".zip";
+#elif defined(__linux__)
+    params.filename = "fastflowlm_" + bare_version + "_ubuntu24.04_amd64.deb";
```
superm1, line 43: This is not a good assumption that an Ubuntu 24.04 compiled FLM works everywhere. The XRT library and others that are dynamically linked are NOT ABI compatible.

PR #1475, opened 2026-03-27
Upgrade flm version to 0.9.38

ramkrishna2910:
This PR moves to using FLM through zip files and updates to 0.9.38.

### DIFF
--- a/docs/embeddable/backends.md
+++ b/docs/embeddable/backends.md
@@ -49,7 +49,10 @@ Follow these instructions if you want backends to be bundled into your app's ins
 
 > Note: by default, `lemond backends install` will only install backends that are compatible with your current system. The `--force` option ignores these compatibility checks, which enables you to package on a VM and then deploy to a specific system.
 
-> Note: at the time of this writing, `flm` and `llamacpp:rocm` are not available for packaging-time bundling.
+#### Limitations
+At the time of this writing:
+-  `flm` is not available for packaging-time bundling *on Linux*.
+- `llamacpp:rocm` is not available for packaging-time bundling on any OS.
 
 ### Installing Backends at Install-Time or Runtime
 


ramkrishna2910:
@claude review this PR

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -124,23 +119,51 @@ namespace lemon::backends {
...
+        int result = system(command.c_str());
+        if (result != 0) {
+            LOG(ERROR, backend_name) << "DEB extraction failed. Ensure 'dpkg-deb' is installed. Code: " << result << std::endl;
+            return false;
+        }
+        return true;
+#else
+        LOG(ERROR, backend_name) << "DEB extraction is only supported on Linux" << std::endl;
+        return false;
+#endif
```
superm1: NACK - we shouldn't be doing this.  libxrt is not ABI portable and definitely neither is ffmpeg.

In my mind this needs to be a portable zip file for Linux if we want to be going this direction.   It would need to statically link ffmpeg, and _bundle_ libxrt (because the low level kernels are dynamically linked).

jeremyfowers: @ramkrishna2910 what do you think of the following:
- On Windows, use the new embeddable flm.zip
- On Linux, keep things the same as on `main`: user is responsible for installing flm.deb themselves

That helps us avoid these thorny issues, as well as needing to rewrite the FLM Linux guide.

src/cpp/server/backends/fastflowlm_server.cpp
```diff
@@ -1,81 +1,106 @@
...
+FastFlowLMServer::FastFlowLMServer(const std::string& log_level, ModelManager* model_manager,
+                                   BackendManager* backend_manager)
+    : WrappedServer("FastFlowLM", log_level, model_manager, backend_manager) {
+}
+
+FastFlowLMServer::~FastFlowLMServer() {
+    unload();
 }
 
 std::string FastFlowLMServer::download_model(const std::string& checkpoint, bool do_not_upgrade) {
```
jeremyfowers, line 73: Why is download_model() changing at all in this PR?

src/cpp/server/backends/fastflowlm_server.cpp
```diff
@@ -150,10 +198,15 @@ void FastFlowLMServer::load(const std::string& model_name,
     // Note: checkpoint_ is set by Router via set_model_metadata() before load() is called
     // We use checkpoint_ (base class field) for FLM API calls
 
-    // Check FLM is ready via system-info (single source of truth)
-    auto status = SystemInfoCache::get_flm_status();
-    if (!status.is_ready()) {
-        throw std::runtime_error(status.error_string());
+    // Install FLM binary if needed (downloads zip/deb and extracts)
+    backend_manager_->install_backend(SPEC.recipe, "npu");
```
jeremyfowers, line 170: Needs different paths on windows vs. linux since auto-install is not possible on linux.

src/cpp/server/utils/path_utils.cpp
```diff
@@ -173,9 +173,27 @@ bool is_safe_executable_path(const std::string& path) {
...
+            if (entry.is_regular_file() && entry.path().filename().string() == binary_name) {
+                std::string path = entry.path().string();
+                if (is_safe_executable_path(path)) {
+                    return path;
+                }
+            }
+        }
+    }
+
+    // Fall back to system PATH
```
jeremyfowers: I thought we were only going to use embeddable FLM on windows?

src/cpp/server/backends/fastflowlm_server.cpp
ramkrishna2910: This was fixing a pre-existing waitpid reaping race on Linux where is_running() consumes the exit status before get_exit_code() can read it.
I can split that into a separate PR if you needed, or keep it here since the fix is small.

src/cpp/server/backends/backend_utils.cpp
ramkrishna2910: Yeah agreed. Will keep Linux as is for now.

superm1:
If eventually you want to have a portable flm for Linux, we should probably help them work upstream on creating such an artifact

src/cpp/server/backends/fastflowlm_server.cpp
ramkrishna2910: Reverted Linux to existing approach

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -61,14 +63,7 @@ namespace lemon::backends {
...
-#ifdef _WIN32
-        std::string cmd = "if not exist \"" + dir + "\" mkdir \"" + dir + "\" >nul 2>&1";
-        std::string unused;
-        lemon::utils::ProcessManager::run_command(cmd, unused);
-#else
-        std::string cmd = "mkdir -p \"" + dir + "\"";
-        system(cmd.c_str());
-#endif
+        fs::create_directories(dir);
     }
```
jeremyfowers: could all instances of `ensure_directory(dir)` just be replaced by `fs::create_directories(dir);` then? This function seems unnecessary now.

src/cpp/server/model_manager.cpp
```diff
@@ -2261,8 +2284,19 @@ void ModelManager::download_from_flm(const std::string& checkpoint,
...
+        flm_path = backends::BackendUtils::get_backend_binary_path(
+            backends::FastFlowLMServer::SPEC, "npu");
+    } catch (...) {
+#ifndef _WIN32
+        flm_path = utils::find_flm_executable();
+#endif
+    }
+    if (flm_path.empty()) {
+        throw std::runtime_error("FLM executable not found");
+    }
```
jeremyfowers: I keep seeing this same code block, should this be a helper function?

src/cpp/server/system_info.cpp
```diff
@@ -893,90 +882,36 @@ json SystemInfo::build_recipes_info(const json& devices) {
             backend["message"] = message;
             backend["action"] = "";
         } else if (!available) {
-            // FLM uses 5-state model: installable, update_required, action_required
```
jeremyfowers, line 897: Isn't this still needed for linux? The main idea of these update_required and action_required is to guide the user through the manual steps needed for things like installing xrt, flm .deb, etc.

jeremyfowers: More broadly: have you tested this PR, on both Windows and Linux, on systems with poor readiness? Such as 

- bad npu driver version on windows
- xrt not installed on linux
- flm not installed on linux

and everything still behaves the same as on main?

.github/actions/install-lemonade-server-msi/action.yml
```diff
@@ -150,13 +150,14 @@ runs:
         $installPath = "${{ inputs.install-path }}"
```
jeremyfowers, line 150: Address https://github.com/lemonade-sdk/lemonade/pull/1508#discussion_r3045546178 in this PR if #1449 merges first.

src/cpp/server/system_info.cpp
jeremyfowers: Reviewed by @ZaneNi

jeremyfowers:
Testing with emebeddable and then I'll merge.

jeremyfowers:
```
(base) PS C:\Users\Jeremy\Downloads\lemonade-embeddable-10.1.0-windows-x64\lemonade-embeddable-10.1.0-windows-x64> .\lemonade.exe backends install flm:npu
Installing backend: flm:npu
[1/1] fastflowlm_0.9.38_windows_amd64.zip (31.7 MB)
  Progress: 100% (31.7/31.7 MB) 61.7 MB/s
Backend installed successfully: flm:npu
(base) PS C:\Users\Jeremy\Downloads\lemonade-embeddable-10.1.0-windows-x64\lemonade-embeddable-10.1.0-windows-x64> ls bin/


    Directory:
    C:\Users\Jeremy\Downloads\lemonade-embeddable-10.1.0-windows-x64\lemonade-embeddable-10.1.0-windows-x64\bin


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----          4/7/2026  12:29 PM                flm


(base) PS C:\Users\Jeremy\Downloads\lemonade-embeddable-10.1.0-windows-x64\lemonade-embeddable-10.1.0-windows-x64>
```

boom!

.github/actions/install-lemonade-server-msi/action.yml
jeremyfowers: Done

PR #1485, opened 2026-03-29
Adding args to llamacpp to use -np 4 and disable the kv unified cache which is hurting performance on Strix Halo iGPUs

eddierichter-amd:
I am seeing some non-trivial performance degradation on my Strix Halo when using the ROCm backend when the unified kv cache is used, which is on by default. Updating the llamacpp launch args to remove that.

### DIFF
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -231,6 +231,10 @@ void LlamaCppServer::load(const std::string& model_name,
     // Disable llamacpp webui by default
     push_overridable_arg(args, llamacpp_args, "--no-webui");
 
+    // Use 4 parallel slots and disable unified KV cache
+    push_overridable_arg(args, llamacpp_args, "-np", "4");
+    push_overridable_arg(args, llamacpp_args, "--no-kv-unified");
+
     // Disable mmap on iGPU
     if (SystemInfo::get_has_igpu()) {
         push_overridable_arg(args, llamacpp_args, "--no-mmap");
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -231,9 +231,11 @@ void LlamaCppServer::load(const std::string& model_name,
     // Disable llamacpp webui by default
     push_overridable_arg(args, llamacpp_args, "--no-webui");
 
-    // Use 4 parallel slots and disable unified KV cache
-    push_overridable_arg(args, llamacpp_args, "-np", "4");
-    push_overridable_arg(args, llamacpp_args, "--no-kv-unified");
+    // Use 4 parallel slots and disable unified KV cache for ROCm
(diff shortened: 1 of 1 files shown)


slojosic-amd:
@eddierichter-amd This is conclusion after several conducted experiments and deep analysis:
 
image

 
**All three configurations allocate the same total KV cache (24576 MiB), but manage it very differently.**
`-np -1 (auto)` uses a unified KV buffer (`kv_unified=true`) shared across 4 sequences, each with a full 262144-token context window. Because the buffer is unified and prompt caching is enabled by default, KV entries from completed requests are retained for potential reuse. In a sequential benchmark with distinct prompts, this causes the buffer to fill up progressively — resulting in latency degradation from ~6s to ~11s per request.
`-np 1` and` -np 4` both use `kv_unified=false` with independent per-slot buffers. The critical difference is that `-np 4` divides the context window per slot (`n_ctx_seq = 65536`), while `-np 1` keeps the full window. Despite this, both deliver identical throughput (~950 tok/s) because each slot's buffer is fully cleared after every request — no accumulation, no degradation.

**Practical recommendation**: For single-user benchmarking, use `-np 1` — it gives the full context window per request, clean slot resets, and maximum prefill performance. Use `-np 4` only if you need multi-user concurrency and can accept the reduced per-slot context of 65536 tokens.
 
`kv_unified=true` with `-np auto` is designed for real production scenarios where users share a common system prompt or engage in multi-turn conversations. The unified buffer allows the KV cache from one session to be reused by another if they share a common prefix — this is a feature, not a mistake (see `--cache-reuse` and `--cache-prompt` options).

**Why it looks like a bug in a benchmark**:
The benchmark sends 10 completely distinct prompts sequentially — no shared prefix, no opportunity for cache reuse. The unified buffer simply accumulates "dead" KV entries that will never be reused, and the server cannot immediately evict them because it does not know in advance whether the next request will want to reuse them.

**Where the real issue lies**:
The help text for `kv-unified` states "**enabled if number of slots is auto**" — **this is documented, but not intuitive**. A user who starts the server without `-np` and immediately runs a benchmark gets degraded results with no clear warning.

**Conclusion**: Good feature for production, bad default for benchmarking. This is maybe worth opening an issue or PR on llama.cpp repo that adds a warning to the logs when the unified cache grows without a single cache hit, or at minimum documents that benchmarks should use `-np 1` or `--no-cache-prompt`.

osimarr:
Can you share how to reproduce your experiment?
I'm running a test with and without kv unified but I'm not seeing difference.
It could be different llamacpp versions. I'm using this one: https://github.com/lemonade-sdk/llama.cpp/releases/tag/b8895

eddierichter-amd:
> Can you share how to reproduce your experiment? I'm running a test with and without kv unified but I'm not seeing difference. It could be different llamacpp versions. I'm using this one: https://github.com/lemonade-sdk/llama.cpp/releases/tag/b8895

Thanks for reaching out! This is a stale PR and the performance difference was more of an artifact of the microbenchmark. I will close this.

PR #1490, opened 2026-03-30
Vendor-in EventSource

bitgamma:
attempt to fix the EventSource issue in the log window

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -125,3 +125,32 @@ license.  It's purpose is a logging system.
 > LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 > OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 > SOFTWARE.
+
+## EventSource
+
+EventSource is an extended implementation of the standard EventSource API, licensed under the MIT license
+from https://github.com/EventSource/eventsource. It has been imported in the form of a prebuilt NPM archive
+with no modifications to the source code in the lemonade tree under the MIT license. Its purpose is adding
+support for Bearer Token Authentication to the EventSource API.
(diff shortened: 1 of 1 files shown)


src/web-app/vendor/eventsource-4.1.0.tgz
superm1, line 1: Does this _have_ to be a tgz, or could it be an extracted directory?  If we're vendoring the dep I would rather extract it for transparency.

We would also need to update NOTICE.md

bitgamma: only seems to work in .tgz format. I updated the NOTICE.md file

superm1:
@jeremyfowers can you test it works now with this?

src/web-app/vendor/eventsource-4.1.0.tgz
kenvandine: I agree with @superm1 on this, referencing the tgz is pretty icky but I guess in the grand scheme of things it's really not worse than things like npm packages being pulled in. But we should try to avoid a libxz type of situation.

kenvandine: Also note, every time that file changes we need to have someone extract that file and review the contents.

bitgamma: @kenvandine this is not meant to be a long term solution. Long term either the node-eventsource package in Ubuntu needs to be updated or we need to ditch this library altogether. This solution is just here to give ubuntu/debian users working log window now

superm1: I tried up-reving it in a Debian package and ran into major problems, but yes updated debian package is the long term intended solution.

superm1:
Let's leave this a backup only if https://github.com/lemonade-sdk/lemonade/pull/1494 doesn't work.

PR #1494, opened 2026-03-30
Replace log SSE streaming with websockets

jeremyfowers:
Fixes https://github.com/lemonade-sdk/lemonade/issues/1442

## Summary
This PR replaces the server's log SSE endpoint with websocket log streaming and updates the desktop/web logs UI to use that websocket path. It also adds retained log history for reconnect/snapshot behavior, advertises log-streaming capabilities via `/health`, and keeps logging behavior consistent across direct runs, the embedded tray server, and the systemd service.

## Main Behavior Changes
- remove the legacy SSE log stream endpoint
- add websocket log streaming at `/logs/stream` on the shared websocket server
- send an initial retained snapshot plus live `logs.entry` messages after `logs.subscribe`
- advertise websocket log streaming in `/health`, including the websocket port and log stream path
- update the logs UI to connect over websocket and resume with `after_seq` on reconnect
- when `lemond` runs as the systemd service, keep logs in the journal instead of writing a runtime log file

### Server API
- `server.cpp` removes SSE exposure and reports websocket log-streaming support in `/health`
- `websocket_server.cpp` adds `/logs/stream` handling alongside `/realtime`
- `log_stream.*` and `logging_config.*` provide retained log history, subscriber fanout, and logging-target selection

### Client/UI
- `LogsWindow.tsx` and `logWebSocketClient.ts` switch the logs view from `EventSource` to websocket streaming
- app/web package metadata drops the unused `eventsource` dependency

### Runtime / startup behavior
- CLI/server and embedded tray startup now wire websocket log streaming through the shared logging setup
- direct `lemond` runs still keep console logging and a runtime log file
- the embedded tray server keeps the runtime log file
- the systemd service uses stdout/journald plus websocket log streaming, without the runtime log file

### Docs
- `docs/server/server_spec.md` documents the websocket log-streaming protocol

### DIFF
--- a/contrib/debian/control
+++ b/contrib/debian/control
@@ -22,7 +22,6 @@ Build-Depends:
  nlohmann-json3-dev,
  node-buffer,
  node-css-loader,
- node-eventsource,
  node-highlight.js,
  node-html-webpack-plugin,
  node-markdown-it,


src/cpp/server/log_stream.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+private:
+    std::ofstream file_;
+};
+
+std::string log_file_path() {
+#ifdef _WIN32
+    return utils::get_runtime_dir() + "lemonade-server.log";
+#else
+    return utils::get_runtime_dir() + "/lemonade-server.log";
```
superm1: how does this work with systemd though?

jeremyfowers: Generally: the log stream now comes directly from AIXLOG instead of from reading journal or .log files.

Specifically to these lines of code, I need to take a look. Not sure why we’re referencing those files anymore.

jeremyfowers: Refactored and fixed a systemd-related bug.

jeremyfowers:
fyi @ramkrishna2910 you're on here due to minor fixes to AGENTS.md, no need to review the whole thing.

bitgamma:
works perfectly. Tested with and without API keys, all is good.

Only one thing: `node-eventsource` is also listed in `contrib/debian/control‎`. I think it needs to be removed from there as well

PR #1502, opened 2026-04-01
fix(server): return empty string when exact GGUF variant not found

ianbmacdonald:
## Summary
- When `resolve_model_path` couldn't find the exact `.gguf` variant specified in `server_models.json`, it returned the model cache **directory** path instead of an empty string
- This caused the UI to incorrectly show models as "downloaded" (play icon) when only a different variant existed in the cache, and llama.cpp to fail with "failed to read magic" when attempting to load the directory as a model file
- Returns `""` instead of the directory path so the model correctly reports as not downloaded

## Test plan
- [x] Verified on a live system (Debian Trixie) where `Qwen3-Coder-Next-Q2_K.gguf` was present but `Qwen3-Coder-Next-MXFP4_MOE.gguf` (the registered variant) was not
- [x] After fix, model correctly shows `downloaded: false` via `/v1/models?show_all=true`
- [x] Model no longer appears in the default downloaded models list
- [x] Verify downloading the correct variant restores normal operation

### DIFF
--- a/test/server_env_vars.py
+++ b/test/server_env_vars.py
@@ -283,6 +283,73 @@ class TestDefaults(unittest.TestCase):
         self.assertEqual(self.snapshot["extra_models_dir"], "")
 
 
+# ---------------------------------------------------------------------------
+# Test: wrong .gguf variant must not mark model as downloaded
+# ---------------------------------------------------------------------------
+
+
+class TestWrongGgufVariantNotDownloaded(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


ianbmacdonald:
Added a regression test for #1502 that seeds a fake HF cache with the wrong .gguf sibling for Tiny-Test-Model-GGUF and verifies the model is not treated as downloaded. It checks both show_all (downloaded=false) and the default downloaded-only /models list, and it runs in an isolated temp cache/server so it should not affect downstream tests.

PR #1507, opened 2026-04-01
Spring Cleaning 3: Change the default port to 13305

### LINKED ISSUE #1458: Change the default port to 13305
cc @danielholanda @superm1 @ramkrishna2910 this will be part of v10.1.0, which also deprecates `lemonade-server serve`.

We will need a coordinated effort to update all downstream software such as n8n, copilot, HF, etc.

jeremyfowers:
Closes #1458 

## Summary
- change Lemonade's default HTTP port from `8000` to `13305`
- update UDP beacon broadcast and discovery to use `13305`
- keep `8000` available as an explicit option in the tray port menu
- propagate the new default through docs, examples, tests, Docker assets, and GitHub workflows


ramkrishna2910:
@claude review

PR #1508, opened 2026-04-01
Add embeddable lemonade release artifacts

jeremyfowers:
## Reviewer Instructions

1. Make sure you are signed into GitHub.
2. Scroll down to any of the "C++ Server Build, Test, and Release" Checks on this PR, click, and then click Summary in the top left corner.
3. Download the \`lemonade-embeddable-linux\` or \`lemonade-embeddable-windows\` and untar/unzip.

image

4. The instructions are in \`docs/embeddable/README.md\`. Link: https://github.com/lemonade-sdk/lemonade/tree/1449-implement-embeddable-lemonade/docs/embeddable

## Summary
- Adds portable \`.tar.gz\` (Ubuntu) and \`.zip\` (Windows) release artifacts for developers to bundle lemonade into their own installers
    - Contains only \`lemond\`, \`lemonade\` CLI, and resource files — no web app, electron, tray, or legacy CLI
    - Builds with \`BUILD_WEB_APP=OFF\` and targets only \`lemond lemonade\` for fast, lightweight CI jobs
    - Light smoke tests validate archive structure, forbidden file exclusion, version commands, and \`lemond ./\` health check
    - Updates release job and release notes with a separate "Embeddable Lemonade" download table
- Adds embeddable Lemonade guide under \`docs/embeddable/\` covering runtime integration, backends, models, and building from source
- Fixes \`models_dir\` config change not invalidating \`ModelManager\`'s download-status cache, preventing stale download flags after \`/internal/set\` redirects to a new models directory

## Requires

- [x] Write an embeddable lemonade guide
    - [x] Include both Linux and Windows instructions for all steps
- [x] \`models_dir="./models"\` config is not respected (fixed in this PR)
- [x] https://github.com/lemonade-sdk/lemonade/pull/1527
- [x] https://github.com/lemonade-sdk/lemonade/issues/1539
- [ ] https://github.com/lemonade-sdk/lemonade/pull/1475

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -33,7 +33,7 @@ jobs:
     outputs:
       unsigned-artifact-id: ${{ steps.upload-unsigned-msi.outputs.artifact-id }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           clean: true
           fetch-depth: 0
@@ -130,12 +130,12 @@ jobs:
           & $wixExe --version
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: '20'
 
       - name: Cache FetchContent dependencies
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: build/_deps
           key: fetchcontent-windows-${{ hashFiles('CMakeLists.txt') }}
@@ -189,7 +189,7 @@ jobs:
 
       - name: Upload Lemonade Server Installers
         id: upload-unsigned-msi
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: Lemonade_Server_MSI
           path: |
@@ -238,7 +238,7 @@ jobs:
           Get-ChildItem -Path "signed-msi" -Recurse | Format-Table Name, Length
 
       - name: Upload Signed MSI Installers
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: Lemonade_Server_MSI_Signed
           path: |
@@ -257,7 +257,7 @@ jobs:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           clean: true
           fetch-depth: 0
@@ -280,7 +280,7 @@ jobs:
           deb-version: ${{ steps.get_version.outputs.version }}
 
       - name: Upload .deb package
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: lemonade-deb
           path: ${{ steps.build_deb.outputs.output-dir }}/*.deb
@@ -294,7 +294,7 @@ jobs:
     outputs:
       version: ${{ steps.get_version.outputs.version }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           clean: true
           fetch-depth: 0
@@ -338,7 +338,7 @@ jobs:
           rpm -qip "$RPM_FILE"
 
       - name: Upload .rpm package
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: lemonade-rpm
           path: build/lemonade-server-${{ env.LEMONADE_VERSION }}.x86_64.rpm
@@ -351,7 +351,7 @@ jobs:
       version: ${{ steps.get_version.outputs.version }}
       has_signing: ${{ steps.check_signing.outputs.has_signing }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           clean: true
           fetch-depth: 0
@@ -393,7 +393,7 @@ jobs:
 
       - name: Upload .pkg package
         if: steps.check_signing.outputs.has_signing == 'true'
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: lemonade-macos-pkg
           path: build/*.pkg
@@ -558,7 +558,7 @@ jobs:
     outputs:
       version: ${{ steps.get_version.outputs.version }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           clean: true
           fetch-depth: 0
@@ -576,7 +576,7 @@ jobs:
           echo "Build environment configured successfully!"
 
       - name: Setup Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: '20'
 
@@ -605,7 +605,7 @@ jobs:
           "$APPIMAGE_FILE" --appimage-help | head -5
 
       - name: Upload AppImage artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: lemonade-appimage
           path: build/app-appimage/lemonade-app-${{ env.LEMONADE_VERSION }}-x86_64.AppImage
@@ -794,7 +794,7 @@ jobs:
       PYTHONIOENCODING: utf-8
       GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Cleanup processes
         uses: ./.github/actions/cleanup-processes-windows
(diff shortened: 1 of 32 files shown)


.github/actions/generate-release-notes/action.yml
```diff
@@ -99,7 +99,7 @@ runs:
         RELEASE_NOTES_FILE="release_notes.md"
```
jeremyfowers, line 99:
```
(base) PS C:\Users\jefowers\Downloads\lemonade-embeddable-windows\lemonade-embeddable-10.0.1-windows-x64\lemonade-embeddable-10.0.1-windows-x64> .\lemonade.exe pull Qwen3-0.6B-GGUF
Pulling model: Qwen3-0.6B-GGUF
Error pulling model: Could not connect to Lemonade server (Failed to read connection).
Make sure the server is running and try again.
```

Timeout on first model pull?

.github/actions/generate-release-notes/action.yml
```diff
@@ -99,7 +99,7 @@ runs:
...
 
         {
```
jeremyfowers, line 101: Bug: hardware_info.json gets cached on the system doing the customization/packaging, which will be incorrect on the deployed system.

.github/actions/generate-release-notes/action.yml
jeremyfowers: Haven't seen it again

docs/embeddable/runtime.md
```diff
@@ -0,0 +1,185 @@
...
+=== "Linux (bash)"
+
+    ```bash
+    LEMONADE_API_KEY=KEY lemond ./ --port PORT
+    ```
+
+Breaking this down:
+- `LEMONADE_API_KEY=KEY` sets an API key for `lemond` known only to your app. This locks out other apps, as well as users, from interfacing directly with `lemond`'s endpoints.
+- The positional `./` is the working directory for `lemond`, where it will look for `config.json`, `bin/`, etc.
+- `--port PORT` ensures that `lemond` launches on a specific port where your app will find it.
```
superm1, line 33: can you make this positional argument optional instead and assume it's `pwd` if not set?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -611,6 +611,196 @@ jobs:
...
+          set -e
+
+          echo "Configuring build with web app disabled..."
+          # Do not use setup.sh — we intentionally avoid installing system
+          # libraries so that FetchContent statically links all deps, making
+          # the embeddable binary portable across Ubuntu versions.
+          cmake --preset default -DBUILD_WEB_APP=OFF
+
+          echo "Building lemond and lemonade..."
+          cmake --build --preset default --target lemond lemonade
```
superm1: how about making a new CMake preset for embedded lemonade that just builds these two targets

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -611,6 +611,196 @@ jobs:
...
+          cp build/resources/server_models.json "${ARCHIVE_DIR}/resources/"
+          cp build/resources/backend_versions.json "${ARCHIVE_DIR}/resources/"
+          cp build/resources/defaults.json "${ARCHIVE_DIR}/resources/"
+          chmod 755 "${ARCHIVE_DIR}/lemond" "${ARCHIVE_DIR}/lemonade"
+
+          tar czf "${ARCHIVE_DIR}.tar.gz" "${ARCHIVE_DIR}"
+          echo "Archive contents:"
+          tar tzf "${ARCHIVE_DIR}.tar.gz"
+          echo ""
+          ls -lh "${ARCHIVE_DIR}.tar.gz"
```
superm1: with a new cmake preset this can all be done as part of the cmake install step

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -611,6 +611,196 @@ jobs:
...
+
+      - name: Upload embeddable archive
+        uses: actions/upload-artifact@v4
+        with:
+          name: lemonade-embeddable-linux
+          path: lemonade-embeddable-${{ env.LEMONADE_VERSION }}-ubuntu-x64.tar.gz
```
superm1: Ideally you want this to be able to run on non-Ubuntu too; right?  So this shouldn't have `ubuntu` in the filename I'd say.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -611,6 +611,196 @@ jobs:
...
+          test -f build/lemond
+          test -f build/lemonade
+          test -f build/resources/server_models.json
+          test -f build/resources/backend_versions.json
+          test -f build/resources/defaults.json
+          echo "Build outputs verified!"
+
+          # Verify no unexpected dynamic dependencies
+          echo "Dynamic dependencies:"
+          ldd build/lemond | grep -v "linux-vdso\|ld-linux" || true
```
superm1: You should drop the `|| true` here to make sure nothing sneaks in.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -880,6 +1070,199 @@ jobs:
...
+          "$DIR/lemonade" --version
+
+      - name: Test lemond startup and health check
+        shell: bash
+        run: |
+          set -e
+          DIR="lemonade-embeddable-${LEMONADE_VERSION}-ubuntu-x64"
+          cd "$DIR"
+
+          ./lemond ./ --port 13305 &
```
superm1: isn't this port implied default?  just wondering if we need to hardcode in another place

docs/embeddable/backends.md
```diff
@@ -0,0 +1,171 @@
...
+    # Start lemond to download backends to ./bin/
+    ./lemond ./
+
+    # Download llama.cpp with the Vulkan backend to ./bin/llamacpp/vulkan
+    ./lemonade backends install llamacpp:vulkan --force
+    ```
+
+> Note: by default, `lemond backends install` will only install backends that are compatible with your current system. The `--force` option ignores these compatibility checks, which enables you to package on a VM and then deploy to a specific system.
+
+> Note: at the time of this writing, `flm` and `llamacpp:rocm` are not available for packaging-time bundling.
```
superm1, line 52: what's the blocker on flm?  Just the linux side?

docs/embeddable/runtime.md
jeremyfowers: It is optional, but the default is `~/.cache/lemonade`. That's the right default for Lemonade Server and not totally unreasonable here, so I'm not sure I should change it.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Great call

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Need your advice here. Is it possible to build a single portable binary that is portable across many distros? I am following llamacpp's lead here: they release an `ubuntu` binary and you have to build from source if that doesnt work for you.

docs/embeddable/backends.md
jeremyfowers: #1475 hasn't merged yet and I don't necessarily want to block this PR on that PR.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: 33d09e7 - nice simplification!

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: 33d09e7 - nice simplification!

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Dropped the whole block. The clean runner job will be the test of portability.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Removed

docs/embeddable/backends.md
superm1: OK, just make sure a comment is added to that PR that this should be changed when it's rebased.

.github/workflows/cpp_server_build_test_release.yml
superm1: It depends on what it links with.   You can look at `ldd ./lemond` to tell this.

The easiest check is to take an artifact and try to run it in some trivial containers for other distros.  Using distrobox/toolbox makes this very easy.
1. Download artifact
2. Extract it to home directory somewhere
3. Make sure it works in host distro (ubuntu presumably)
4. Spin up a toolbox/distrobox container for each of Debian stable, Fedora latest release, older Ubuntu.
5. Try to run it.

If it fails to run then it's not really portable and we should figure out why.

docs/embeddable/runtime.md
superm1: But since you're doing a different binary build you can potentially have a _compile time_ difference

docs/embeddable/backends.md
jeremyfowers: Added

docs/embeddable/runtime.md
jeremyfowers: When in doubt I am trying to minimize surface area. Make things work the same in server vs. emeddable, windows vs. linux, etc.

If customers have trouble remembering to write `./` let's adjust.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: I am going to leave the name as-is right now to indicate that it is only tested on Ubuntu. If customers want cross-distro support in the release we will do this work as you suggest.

PR #1513, opened 2026-04-02
Add a hack for a problem with node-commander migration in Ubuntu 26.04

superm1:
This should be sorted there eventually, but pull the proposed package manually for the build container for now.

### DIFF
--- a/.ubuntu/Dockerfile
+++ b/.ubuntu/Dockerfile
@@ -28,6 +28,27 @@ RUN apt update && apt install -y \
     cd /contrib && \
     apt build-dep . -y && \
     rm -rf /var/lib/apt/lists/*
+
+# Temporary Ubuntu 26.04 workaround for node-commander version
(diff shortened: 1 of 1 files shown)


superm1:
Here is the test workflow for it.

https://github.com/lemonade-sdk/lemonade/actions/runs/23907638347

kenvandine:
Adding a comment here even though the PR has already merged :)

@superm1 ewww proposed... but yeah, during this current transition in 26.04 is this is a reasonable workaround, but let's try to remember to drop the workaround when the transition is complete or it might bite us later.

superm1:
> Adding a comment here even though the PR has already merged :)
> 
> @superm1 ewww proposed... but yeah, during this current transition in 26.04 is this is a reasonable workaround, but let's try to remember to drop the workaround when the transition is complete or it might bite us later.

Well once the proposed transition is done it will automatically do nothing from the node commander version check even if we forget. That's why I did it that way.

PR #1516, opened 2026-04-03
Add OpenCode integration with `launch`

sawansri:
Adds OpenCode as an integration into `launch` similar to Claude Code and Codex.

Since OpenCode's source of truth is a configuration file like `opencode.json`, this PR introduces a file based agent configuration abstraction so that future agents like pi and openclaw can be added with minimal boilerplate.

There are multiple scenarios in the agent configuration process that `launch` covers:

1. Creates the file if it doesn't exist (also populates config with downloaded models)
image
2. If file exists, reads and parses the JSON. `launch` never removes user-defined providers or models under other providers. It updates connection settings (port, host) and merges models (adds new, updates existing).


ramkrishna2910:
@claude review

iswaryaalex:
Works great on my end on both windows and Linux

PR #1524, opened 2026-04-03
Forward GGML_METAL_NO_RESIDENCY env var to llama-server on macOS

Geramy:
## Summary
- Explicitly forwards `GGML_METAL_NO_RESIDENCY` from lemond's environment to the llama-server subprocess on macOS
- When lemond is started by launchd (e.g. after `.pkg` install), it does not inherit the shell environment, so the env var never reaches llama-server
- This fixes llama-server b8648 crashing on macOS CI runners (MTLGPUFamilyApple5 paravirtualized GPU) due to unsupported Metal residency sets

## Test plan
- [ ] Verify macOS CI tests pass with `GGML_METAL_NO_RESIDENCY=1` set in workflow
- [ ] Verify no regression on macOS when the env var is not set (normal user machines)

### DIFF
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -303,6 +303,17 @@ void LlamaCppServer::load(const std::string& model_name,
     }
 #endif
 
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


Geramy:
transfered to other PR

PR #1527, opened 2026-04-03
Spring Cleaning 4: Eliminate hardware_info.json

jeremyfowers:
## Summary

- Removes on-disk `hardware_info.json` cache entirely — `lemond` now scans hardware at startup and holds the result in memory for the process lifetime
- `SystemInfoCache` stripped from a file-backed singleton to 3 static methods: `get_system_info_with_cache()`, `invalidate_recipes()`, `get_flm_status()`
- Replaces Win32_PnPSignedDriver WMI query with SetupAPI enumeration for driver version lookup (~10 s → ~150 ms)
- Replaces WMI for CPU info with CPUID + `GetLogicalProcessorInformation`, and for OS version with direct registry reads (~7 s → ~0 ms)
- Removes unused system-info fields: OEM System model, BIOS version, CPU max clock speed, Windows power setting, NPU power mode
- Fixes `get_flm_version()` double-invocation per scan by caching the result (only non-"unknown" values cached to preserve post-install verification)
- Extracts `read_cpu_hardware()` helper to eliminate duplication between `get_cpu_device()` and `get_processor_name()`
- Deletes `test/server_system_info.py` (relied on injecting mock `hardware_info.json`) and removes the `system-info` CI job

Net result: cold-start system-info scan reduced from ~34 s to ~3.7 s (hardware ~860 ms + FLM NPU runtime load ~2.7 s, which is irreducible).

## Test plan

- [ ] Build on Windows and verify `lemond` starts cleanly with no `hardware_info.json` written to disk
- [ ] `GET /api/v1/system-info` returns correct hardware and recipe data
- [ ] After installing/upgrading a backend, `invalidate_recipes()` causes the next `system-info` call to re-scan recipes
- [ ] `GET /api/v1/system-info` on Linux and macOS still works correctly
- [ ] CI passes (cli, endpoints, ollama, llamacpp-system, streaming-errors, env-vars jobs)


jeremyfowers:
Tested on a few computers, seems good!

PR #1535, opened 2026-04-04
fix: use sliding window for download speed display

ianbmacdonald:
## Summary
- Replaces the all-time average speed calculation with a **5-second sliding window**, showing actual recent throughput instead of a dragged-down session average.
- Records timestamped byte samples on each progress event, prunes samples older than 5 seconds, and computes speed from the window edges.
- Cleans up sample data on download completion and error.

## Problem
The Download Manager computed speed as `totalBytes / totalElapsedTime` — the average over the entire session. If early portions of the download were slow (API negotiation, initial buffering, HF rate limiting ramp-up), the displayed speed remained far below actual throughput. In testing, btop showed ~925 MiB/s network throughput while the app displayed only 15 MB/s.

## Changes
- `src/app/src/renderer/DownloadManager.tsx` — Add `useRef`-based sliding window sample storage, record samples in `handleDownloadUpdate`, clean up in completion/error handlers, rewrite `calculateSpeed` to use window.

## Test plan
- [x] Verified speed display reflects actual throughput during active download
- [x] Verified ETA calculation remains accurate (uses same `calculateSpeed`)
- [x] Verified speed shows `--` initially (before 2 samples collected)
- [x] No memory leak from sample accumulation — samples pruned to 5s window on every event, map cleaned up on completion/error

### DIFF
--- a/src/app/src/renderer/DownloadManager.tsx
+++ b/src/app/src/renderer/DownloadManager.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { deleteModel, uninstallBackend } from './utils/backendInstaller';
 
 export interface DownloadItem {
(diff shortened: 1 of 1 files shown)


ianbmacdonald:
This one is cosmetic, and a good test case for a new (web) UI test plan execution harness I am working on;  going to put in draft (although mostly ready) ;  ping me if there is appetite to merge right away,  before I undraft it after a 2nd round of testing, probably in an upstream branch, since there are no comments or PR history to save here.

jeremyfowers:
looks like something similar was merged by @fl0rianr here https://github.com/lemonade-sdk/lemonade/pull/1995

PR #1536, opened 2026-04-04
fix: resume downloads across HuggingFace commit hash changes

ianbmacdonald:
Resume an *interrupted* HuggingFace download across a commit-hash change instead of restarting from zero. Complements #1950 (SHA-on-download) and #2066 (sticky refs / completed-snapshot reuse), neither of which covers the interrupted case.

The HF API returns the latest commit SHA on every call, so if a download is interrupted and the repo advances before it resumes, a fresh `snapshots/<new_sha>/` is created and everything re-downloads, orphaning the partial one.

- `src/cpp/server/model_manager.cpp` — before creating the new snapshot dir, find an interrupted prior attempt (a snapshot under a different hash carrying an in-progress `.download_manifest.json`) and rename it forward so `download_from_manifest` resumes in place. It is located by that manifest, not `refs/main`, which is sticky and never points at an in-progress snapshot. Carrying bytes forward is safe — #1950 verifies each file against the new commit's hash and re-fetches whatever changed.
- `test/server_endpoints.py` — `test_037` (interrupted snapshot migrated forward and resumed in place, proven by unchanged mtime) and `test_038` (a completed, manifest-less snapshot is *not* migrated).

Scope: main repo only. Auxiliary repos (text encoder, VAE, mmproj) carry no manifest of their own, so an interrupted auxiliary snapshot re-downloads fresh under the new hash — safe, since #1950 still verifies what is fetched. Resuming those would need a recursive `.partial` scan; left as a follow-up.

**Tested:** `test_037`/`test_038` pass on Ubuntu 26.04, RX 7900 XT (gfx1100), backend `llamacpp`/`vulkan`. Not tested: Windows, macOS.


ianbmacdonald:
added some testing and addressed some agent review feedback 

19211829 makes orphan detection and directory-based download-status checks recursive for .partial files, which is important for nested paths under a snapshot.

226cfb8c finishes that off by updating the parallel add_model_to_cache() path and fixing the nested-partial regression test so the manifest filename actually matches the nested partial path being resumed.

The new test shape uses a real HF-backed /pull, but it does not depend on a live upstream repo changing commit hash during the test window. Instead it seeds orphaned snapshot state locally and verifies selection / cleanup behavior from there.  The main change is extra runtime/network.

One integration note for later: if this branch is rebased onto or merged after #1412, the orphan-resume logic in download_from_huggingface() will need a follow-up pass for #1412’s per-file download_path manifest format and multi-repo snapshot layout. The recursive .partial scan changes should carry over directly, but the resume/rename path handling is still written around a single main-repo snapshot to be compatible with current main

ianbmacdonald:
Now that #1412 has merged and this branch is rebased onto it, the integration notes from the earlier comment have been addressed across several commits:

**Multi-repo manifest support (e8e71cfa, 073520e2):**
- Orphan-resume path now preserves per-file `download_path` entries for non-main repos (text_encoder, vae, etc.) — only updates entries that match the old main snapshot path
- Orphan detection heuristic counts completed file sizes + partial sizes (not just partials) using manifest file entries with their per-repo paths

**Stale manifest and content verification (24ae02ef):**
- Removed `download_from_manifest(existing_manifest)` call from orphan-resume entirely — the stale manifest's URLs or file list may not match upstream after a commit change
- Instead, just relocate salvageable files (completed + partials) and fall through to the normal download path, which builds a fresh manifest from the current HF API response
- Added size verification in `download_from_manifest`: when a completed file exists on disk, its size is compared against the manifest and re-downloaded if mismatched

**Snapshot collision handling (a282ba77):**
- Merge policy when relocating into an existing current-hash snapshot: keep complete files over partials, keep the larger `.partial` when both exist, only move what improves on dest

**Test coverage (073520e2, a282ba77, 70c58d0c):**
- `test_030`: single-repo orphan-resume using refs/main for authoritative hash
- `test_031`: multi-repo orphan-resume with per-file download_path entries
- `test_032`: collision path where both orphan and current-hash snapshots coexist
- Tests manipulate disk state and re-pull directly (no `/delete` calls, which would wipe the seeded orphan state)

src/cpp/server/model_manager.cpp
```diff
@@ -2430,6 +2446,152 @@ void ModelManager::download_from_huggingface(const ModelInfo& info,
...
+    // current HF API response, so any files added or removed upstream
+    // are handled correctly (download_from_manifest skips files already
+    // on disk).
+    fs::path snapshots_dir = model_cache_path / "snapshots";
+    if (fs::exists(snapshots_dir)) {
+        fs::path best_snapshot;
+        size_t best_downloaded_bytes = 0;
+        std::vector<fs::path> orphaned_snapshots;
+
+        for (const auto& entry : fs::directory_iterator(snapshots_dir)) {
```
jeremyfowers: _Drafted by Claude on Jeremy's behalf._

This `fs::directory_iterator` call bypasses the `safe_dir_options` / `safe_exists` helpers defined at the top of this file (lines 35–49). The rest of `model_manager.cpp` uses them religiously to avoid `fs::filesystem_error` exceptions on Windows reparse points and permission-denied entries — a raw `fs::directory_iterator` without `skip_permission_denied` will throw uncaught and abort `download_from_huggingface` if the HF cache contains one.

Suggested fixes in this block:
- Wrap with `safe_dir_options` and an `std::error_code`: `fs::directory_iterator(snapshots_dir, safe_dir_options, ec)`
- Replace the `fs::exists(snapshots_dir)` check on line 2463 with `safe_exists(snapshots_dir)`
- Same treatment for the `fs::recursive_directory_iterator` on line 2542

Separately, on line 2493: `fs::file_size(partial, ec)` returns `static_cast<uintmax_t>(-1)` when `ec` is set, which would massively inflate `downloaded_bytes` and cause that snapshot to win the "best progress" comparison. Worth checking `ec` before adding to the running total.

src/cpp/server/model_manager.cpp
```diff
@@ -2430,6 +2446,152 @@ void ModelManager::download_from_huggingface(const ModelInfo& info,
         LOG(INFO, "ModelManager") << "Warning: No commit hash found in API response, using 'main'" << std::endl;
     }
 
+    // Check for an existing in-progress download in a different snapshot.
```
jeremyfowers: _Drafted by Claude on Jeremy's behalf._

This is ~150 lines of scan-score-walk-merge to clean up a mess the cache layout creates in the first place. A simpler approach: when the new commit hash differs from what's in `refs/main` and the old snapshot exists, just rename the old snapshot directory to the new hash. The per-file size-mismatch check this PR already adds to `download_from_manifest` handles upstream-changed files at download time, so the merge logic isn't needed.

Sketch (replaces this entire block):
```cpp
std::string old_hash;
fs::path refs_main = model_cache_path / "refs" / "main";
if (safe_exists(refs_main)) {
    std::ifstream f(refs_main);
    std::getline(f, old_hash);
}
fs::path old_snap = model_cache_path / "snapshots" / old_hash;
fs::path new_snap = model_cache_path / "snapshots" / commit_hash;
if (!old_hash.empty() && old_hash != commit_hash
    && safe_exists(old_snap) && !safe_exists(new_snap)) {
    std::error_code ec;
    fs::rename(old_snap, new_snap, ec);
}
```

The only thing this doesn't handle is *multiple* accumulated orphans. But those only exist because the current code orphans on every pause/resume — fix it once and they stop piling up. For users who already have orphans from prior versions, a one-time cleanup pass on startup would address it without keeping the merge machinery in the hot download path.

Defensible to keep the current approach if graceful handling of legacy orphan accumulation matters, but worth weighing against the LOC.

jeremyfowers (changes requested):
@ianbmacdonald please see the two comments, other than that good to merge.

ianbmacdonald:
Rebased onto current `main`. In the meantime @fl0rianr's #1950 (SHA-on-download) and #2066 (sticky refs / snapshot reuse) landed and rewrote much of `download_from_huggingface`, so I reworked this down to a much smaller slice that builds on them.

The gap they don't cover: an **interrupted** download resumed after the repo advanced to a new commit. #2066 reuses a *completed* previous snapshot when artifacts are unchanged, and #1950 verifies content on download — but an interrupted attempt (with a `.partial`) under the old hash still gets orphaned and the new commit's snapshot re-downloads from zero.

This adds a small `resume_snapshot_across_commit()`: before creating the new snapshot dir, it finds an interrupted prior attempt (a snapshot under a different hash carrying an in-progress `.download_manifest.json`) and renames it forward to the new hash, so `download_from_manifest` resumes in place. It locates the interrupted snapshot by the manifest marker rather than `refs/main`, since #2066 made `refs/main` sticky (advanced only on completion) so it no longer points at an in-progress snapshot. Carrying bytes forward is safe because #1950's per-file SHA check re-fetches anything whose content changed across the commit — including a `.partial` whose blob changed.

Scoped to the main repo, where the single download manifest lives — I noted the auxiliary-checkpoint case as a follow-up in the PR description (an interrupted aux snapshot is simply re-downloaded fresh; still safe). +73 lines + two tests.

src/cpp/server/model_manager.cpp
ianbmacdonald: comment below.. parts of this merged already in my absence

src/cpp/server/model_manager.cpp
ianbmacdonald: reworked, thin additions, comments below

ianbmacdonald:
Rebased onto current `main`. Both of your review points are addressed in the reworked slice:

- **Simplification** — the ~150-line scan-score-walk-merge is gone, replaced with the rename-forward approach you sketched: `resume_snapshot_across_commit()` locates the interrupted snapshot by its in-progress `.download_manifest.json` marker and `fs::rename`s it onto the new hash, letting `download_from_manifest` resume in place. Per-file hash verification (#1950) re-fetches anything the new commit changed, so no merge logic is needed.
- **Safe FS helpers** — the directory walk now uses `safe_exists` / `safe_dir_options` / `safe_is_directory` with `ec` checks, consistent with the rest of the file. The old `fs::file_size` accumulation that could inflate on a set `ec` is gone entirely with the scan removed.

Scope is now 2 files (+245/-0): the migration helper plus two regression tests (resume-across-commit, and not-migrating a completed stale snapshot). Auxiliary-repo (text-encoder/VAE) interrupted resume is documented as a follow-up.

ianbmacdonald:
Rebased onto current `main` (resolves the merge with the recently-merged `/pull/variants` tests). The two resume regression tests were renumbered **034/035** to avoid colliding with the new `test_030`–`test_033` pull/variants tests; the `model_manager.cpp` change merged cleanly past the `auto re-download missing files` work (#2239).

Verified on a fresh build (Ubuntu 26.04, glibc 2.43): `test_034` (interrupted snapshot migrated forward and resumed in place) and `test_035` (completed stale snapshot left intact) both pass against a live `/pull`, alongside the existing pull/variants tests.

ianbmacdonald:
Rebased onto current `main` (`31cdc925a`) — the only conflict was an import-list addition in `test/server_endpoints.py` (`get_hf_cache_dir_candidates`), resolved by keeping upstream's import; `model_manager.cpp` auto-merged. The PR is green/mergeable again.

The `CHANGES_REQUESTED` review from April predates several rounds of rework and has been addressed through the subsequent rebases. @jeremyfowers @ramkrishna2910 — would appreciate a re-review when you have a moment.

bong-water-water-bong (changes requested):
## Multi-agent review found 3 issues — fixes below

### 1. 🔒 Reject symlinks in interrupted-snapshot scan

\`\`\`diff
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -4339,6 +4339,7 @@
     std::error_code ec;
     for (const auto& entry : fs::directory_iterator(snapshots_dir, safe_dir_options, ec)) {
         if (ec) break;
+        if (fs::is_symlink(entry.path())) continue;
         if (!safe_is_directory(entry.path())) continue;
         if (entry.path().filename() == new_hash) continue;
         if (safe_exists(entry.path() / ".download_manifest.json")) {
\`\`\`

### 2. Gate integration tests

\`\`\`diff
+    @unittest.skipUnless(
+        os.environ.get("LEMONADE_INTEGRATION_TESTS") == "1",
+        "Skipped: set LEMONADE_INTEGRATION_TESTS=1 to run live HuggingFace tests",
+    )
     def test_034_pull_resumes_interrupted_snapshot_across_commit_change(self):
...
+    @unittest.skipUnless(
+        os.environ.get("LEMONADE_INTEGRATION_TESTS") == "1",
+        "Skipped: set LEMONADE_INTEGRATION_TESTS=1 to run live HuggingFace tests",
+    )
     def test_035_pull_does_not_migrate_completed_stale_snapshot(self):
\`\`\`

### 3. addCleanup for test teardown

\`\`\`diff
     fake_old = "0" * 40
     old_snapshot = os.path.join(snapshots_dir, fake_old)
+    self.addCleanup(shutil.rmtree, old_snapshot, ignore_errors=True)
...
     fake_old = "1" * 40
     old_snapshot = os.path.join(snapshots_dir, fake_old)
+    self.addCleanup(shutil.rmtree, old_snapshot, ignore_errors=True)
\`\`\`

ianbmacdonald:
Closing this as stale — the download path has since been refactored behind the
registry abstraction (`download_from_registry`, per-repo `snapshot_id`), so this
patch no longer applies in anything like its original shape.

Worth noting the underlying gap looks like it's still there: an interrupted download's
partial snapshot is orphaned if the repo advances mid-download, and
`repos_reusing_previous_snapshot` only covers the byte-identical *complete* case. Happy
for someone to pick it up against the new structure; I'm not holding the branch open.

PR #1542, opened 2026-04-05
Use vendor directory in /opt

d10n:
According to Linux packaging conventions, each item in /opt should be a self-contained folder. This allows packages to co-exist without stepping on each other.

Items to be considered for a different PR:
TODO just use the regular FHS directories instead of /opt
TODO use systemd-sysusers and systemd-tmpfiles instead of postinst-rpm

### DIFF
--- a/src/cpp/postinst-rpm
+++ b/src/cpp/postinst-rpm
@@ -6,11 +6,11 @@ if ! getent group lemonade > /dev/null 2>&1; then
 fi
 
 if ! getent passwd lemonade > /dev/null 2>&1; then
-    useradd -m -r -g lemonade -d /opt/var/lib/lemonade -s /usr/sbin/nologin lemonade > /dev/null 2>&1 || true
+    useradd -m -r -g lemonade -d /opt/lemonade/var/lib/lemonade -s /usr/sbin/nologin lemonade > /dev/null 2>&1 || true
 fi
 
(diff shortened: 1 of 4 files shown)


src/cpp/postinst-rpm
superm1, line 1: Perhaps there should be some migration for the home directory in this new structure

superm1:
> TODO just use the regular FHS directories instead of /opt
> TODO use systemd-sysusers and systemd-tmpfiles instead of postinst-rpm

Another alternative is to have spec file in the tree that matches the Fedora review, which will use FHS.
https://bugzilla.redhat.com/show_bug.cgi?id=2453414

This is what we do for Debian/Ubuntu - the packaging is nearly identical.

d10n:
@superm1 to make sure I understand, you suggest removing cpack and adopting that Fedora COPR spec file into this repository, right?

superm1:
Yes!

superm1:
I would propose just put it in contrib/ and use the rpm ci job to build it instead of cpack.

d10n:
Planning to make the suggested change later today.

superm1:
FWIW - I think this change makes sense "generally"; but I do still thinkg that RPM should move to a spec file.

jeremyfowers:
Closing abandoned PR.

PR #1543, opened 2026-04-05
Npm security

d10n:
* Avoid axios different-version and lifecycle-script surprises
* `npm ci --ignore-scripts` should always be used instead of `npm install` or `npm ci`
* `npm ci` requires package-lock.json to exist
* Maintainer please verify or recreate package-lock.json before merging
* Maybe developer docs should have a note about updating package-lock.json for releases


Geramy:
@claude Can you double check the packages-lock.json please and verify these are correct and have the same main inclusions from packages.json and that none of the versions selected are at risk or infected?

d10n:
@Geramy I checked the "Allow edits by maintainers" box, so you can make commits in my branch - it's probably easiest to just delete the lock file and manually recreate it yourself with `cd src/web-app; npm install --ignore-scripts`, then push the commit to my branch. I might not trust AI to verify lock files anyway :smile:

d10n:
It seems like Claude is failing because it's unable to check out branches from forks.

superm1:
I couldn't force push your branch, but I rebased it and regenerated the package lock json file locally and then opened https://github.com/lemonade-sdk/lemonade/pull/1543 for us to merge from.

PR #1551, opened 2026-04-06
Fix flakey Ollama vision test

### LINKED ISSUE #1500: Fix the flakey ollama test

jeremyfowers:
Closes #1500 

Attempting to stop the Ollama Vision test from frequently flaking in CI, or at least provide us more actionable info when it does flake.

Making the test model much smaller can only be good for robot health, anyways.

## Summary
- **Switch vision test model** from `Gemma-3-4b-it-GGUF` (3.6 GB) to `Qwen3.5-0.8B-GGUF` (0.56 GB). The 4B model intermittently exceeded the 300s internal curl timeout on CPU-only CI runners, causing the test to fail.
- **Add error handling to Ollama API handlers**. When backend errors (e.g. timeouts) occurred, the Ollama translation layer silently converted them into malformed HTTP 200 responses with missing fields (no `message` in chat responses). Now all non-streaming handlers (`handle_chat`, `handle_generate`, `handle_generate_image`, `handle_embed`, `handle_embeddings`) check for error responses and return proper HTTP 500 — matching the pattern already used in `server.cpp`.

## Root cause
`test_021_chat_with_image_input` loads a 4B vision model on CPU-only GitHub Actions runners. Inference sometimes takes >300s, hitting the internal curl timeout. The resulting `NetworkException` error was returned as JSON `{"error": {...}}` by the router, but `handle_chat` passed it directly to `convert_openai_chat_to_ollama` without checking. That function found no `choices` key, so it produced `{"model": "...", "created_at": "...", "done": true}` — a valid-looking HTTP 200 response with no `message` field, causing the assertion failure.

### DIFF
--- a/src/cpp/server/ollama_api.cpp
+++ b/src/cpp/server/ollama_api.cpp
@@ -86,6 +86,19 @@ static std::string extract_quantization_level(const std::string& checkpoint) {
 }
 
 // ============================================================================
+// Check if a router/backend response is an error and send HTTP 500 if so.
+// Returns true if an error was handled (caller should return immediately).
+// ============================================================================
+static bool send_backend_error(const json& response, httplib::Response& res) {
+    if (!response.contains("error")) return false;
+    LOG(ERROR, "OllamaApi") << "Backend returned error: " << response["error"].dump() << std::endl;
+    res.status = 500;
+    json error = {{"error", response["error"].value("message", "backend error")}};
+    res.set_content(error.dump(), "application/json");
+    return true;
+}
+
+// ============================================================================
 // Ollama → OpenAI option name mapping (ollama_key → openai_key)
 // Options where both names are the same use identical strings.
 // ============================================================================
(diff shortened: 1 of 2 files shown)


jeremyfowers:
The entire Ollama test job is running 4x faster now!

src/cpp/server/ollama_api.cpp
```diff
@@ -635,6 +635,16 @@ void OllamaApi::handle_chat(const httplib::Request& req, httplib::Response& res)
             LOG(INFO, "OllamaApi") << "POST /api/chat - Non-streaming (model: " << model << ")" << std::endl;
 
             auto openai_response = router_->chat_completion(openai_req);
+
```
sawansri, line 651: This error handling is repeated ~5 times, can we use a function to prevent repeating ourselves?

sawansri: Selected the wrong line oops

jeremyfowers: Implemented and pushed!

PR #1559, opened 2026-04-06
Add OmniRouter: multimodal tools + model collections for agentic workflows

ramkrishna2910:
## OmniRouter

Introduces **OmniRouter**: Lemonade exposes its multimodal capabilities (image generation, image editing, TTS, transcription, vision) as OpenAI-compatible tools.

See [docs/omni-router.md](docs/omni-router.md) for the concept overview and [examples/lemonade_tools.py](examples/lemonade_tools.py) for a runnable reference.

## Collections

**Collections** are preconfigured bundles of models that load together in the desktop app — one click gets you an LLM + image + ASR + TTS, sized for a hardware tier:

- **Ultra Collection** — Qwen3.5-35B + Flux-2-Klein-9B (gen+edit) + Whisper-Large + kokoro-v1
- **Lite Collection** — Qwen3.5-4B + SD-Turbo + Whisper-Tiny + kokoro-v1

Collections are marked `hidden: true` so they don't leak into OpenAI-compat clients like AnythingLLM; they surface with `?show_all=true`.


ramkrishna2910:
@claude review

.claude/settings.local.json
sawansri, line 1: Should this be gitignored?

CLAUDE.md
sawansri, line 1: Looks like this file got overwritten by Claude. We want to have general agent instructions in AGENTS.md and maintain the reference in CLAUDE.md.

docs/tools.md
```diff
@@ -0,0 +1,177 @@
...
+| `text_to_speech` | `POST /v1/audio/speech` | JSON: `model`, `input`, `voice` | Binary audio (WAV) |
+| `transcribe_audio` | `POST /v1/audio/transcriptions` | Multipart: `file`, `model`, `language` | `{"text": "..."}` |
+| `analyze_image` | `POST /v1/chat/completions` | JSON: `model`, `messages` with image content | `choices[0].message.content` |
+
+## Quick Start
+
+### 1. Find your models
+
+```bash
+curl http://localhost:8000/v1/models
```
jeremyfowers:
```suggestion
curl http://localhost:13305/v1/models
```

jeremyfowers: Lots of references to old port below as well

docs/tools.md
```diff
@@ -0,0 +1,177 @@
...
+```python
+tools = [
+    {"type": "function", "function": tool["function"]}
+    for tool in tool_definitions["tools"]
+]
+```
+
+### 3. Run your agentic loop
+
+```python
```
jeremyfowers: Should this be a python file so that people can try it?

CLAUDE.md
jeremyfowers: Yep and its also stale content

src/cpp/resources/server_models.json
jeremyfowers: Is there any point in having a Medium mode considering it has basically the same RAM requirement as the Ultra mode? Suggest just removing it.

docs/tools.md
```diff
@@ -0,0 +1,177 @@
+# Lemonade Tools
```
jeremyfowers: Image

When I'm first getting started, it says "Size N/A" and doesn't tell me what the download button will actually do.

Then when I click download, it says its downloading "Lemonade Ultra", but the file is actually a flux.gguf.

Can you refine the onboarding experience so that people understand they are signing up for many-gigabyte downloads and can track them appropriately?

Also it would be nice if it was clear they were downloading the models, not downloading "Lemonade Ultra" (since people might think we made our own model).

jeremyfowers: It also showed in the UI that the download was complete, even though only Flux had downloaded and I still needed Kokoro. Iterating over each model and pulling it if needed should solve this.

Maybe it was an architectural mistake to put the experiences into server_models.json? For example they also show up in the models endpoint in AnythingLLM.

jeremyfowers: I also had to download Ultra, then restart `lemond` to get Ultra to show up as available.

docs/tools.md
```diff
@@ -0,0 +1,177 @@
...
+
```
jeremyfowers: Image

Two mic buttons, one overlapping the text entry

docs/tools.md
```diff
@@ -0,0 +1,177 @@
...
+Lemonade exposes multimodal capabilities through standard OpenAI-compatible endpoints. You can use these as tools in any LLM agentic loop — pass the tool definitions to your LLM, execute the tool calls against Lemonade's endpoints, and feed the results back.
+
```
jeremyfowers: Image

What am I doing wrong?

jeremyfowers: Hmm its not generating speech either, and wont admit that it has any tools available.

docs/tools.md
```diff
@@ -0,0 +1,115 @@
...
+## Tool Definitions
```
jeremyfowers: Image

It needs to be clear here what I would download if I push the button. 

Similarly, what is the behavior if I push delete? All models in the bundle are deleted? Is that desirable?

docs/tools.md
```diff
@@ -0,0 +1,115 @@
+# Lemonade Tools
```
jeremyfowers: Image

hehe

docs/tools.md
```diff
@@ -0,0 +1,115 @@
...
+
+Lemonade exposes multimodal capabilities through standard OpenAI-compatible endpoints. You can use these as tools in any LLM agentic loop — pass the tool definitions to your LLM, execute the tool calls against Lemonade's endpoints, and feed the results back.
```
jeremyfowers: Image

Ultra Bundle should a "load model" button even though its already loaded.

jeremyfowers: Image

Model Options for ultra bundle shows nothing. Should it at least show the list of checkpoints?

docs/tools.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Tool Definitions
+
+The canonical tool definitions live in [`src/app/src/renderer/utils/toolDefinitions.json`](../src/app/src/renderer/utils/toolDefinitions.json). This is the single source of truth used by both the desktop app and this documentation.
```
jeremyfowers: https://github.com/user-attachments/assets/2909185f-289f-4a20-aab3-c12810e9f7a1

bug - clicking download doesnt check whether the bundle is ready

jeremyfowers: https://github.com/user-attachments/assets/a53e7d7e-2473-425f-8628-b694ed63619b

bug: models should show as "loading" in Active Models after we click to load the bundle. But they only pop in once health sees them as fully loaded.

docs/tools.md
```diff
@@ -0,0 +1,115 @@
...
+
+Each entry has a `function` object in standard OpenAI tool-calling format (pass directly to any LLM), plus `requires_labels` or `requires_llm_labels` indicating which model capabilities are needed.
```
jeremyfowers: What's the size for flux generations here? I feel like I'm waiting a long time. Can we make the image smaller to speed things up?

docs/tools.md
```diff
@@ -0,0 +1,115 @@
...
+
```
jeremyfowers: Image

It will be pretty difficult to demo Ultra on social media because most responses take up a lot of screen space, and its basically impossible to have two images on screen at a usable resolution. Anything we can do about that? Would be great to have a really nice mobile-formatted demo video for this.

docs/tools.md
jeremyfowers: Also the black dot next to Ultra means "downloaded" but should be green for "loaded"

docs/tools.md
jeremyfowers: I think just list the model names under the bundle, on a single line with word wrap?

docs/tools.md
jeremyfowers: https://github.com/user-attachments/assets/7196ae46-5294-400b-a115-a319da266c03

Bug: playback progress stuck at the end. 

(also, it would be nice if the speech auto-played, ideally streamed!)

docs/tools.md
ramkrishna2910: Wont that be too verbose in the side panel?

docs/tools.md
ramkrishna2910: Done.

docs/tools.md
ramkrishna2910: I hid the gear icon for now.

docs/tools.md
ramkrishna2910: Went from 1024x1024 to 512x512, makes it reasonably faster.

docs/tools.md
ramkrishna2910: image
Hovering over the bundle shows the models in the buncle

docs/tools.md
ramkrishna2910: Fixed

docs/tools.md
ramkrishna2910: Fixed

Image

docs/tools.md
ramkrishna2910: Fixed and the generated audio is set to auto play.

docs/tools.md
ramkrishna2910: Reduced the height of the image to 280px and clicking the image would open it in a bigger size withing the app.

Image

ramkrishna2910: Let me know if this size works.

.claude/settings.local.json
jeremyfowers: @ramkrishna2910 can you answer the question? I want to know too.

docs/tools.md
jeremyfowers: Should this doc be renamed omni-router.md, and address what omni router is and how the tools fit into it?

docs/tools.md
```diff
@@ -0,0 +1,115 @@
...
+```bash
+pip install openai
+python examples/lemonade_tools.py "Generate an image of a sunset"
+python examples/lemonade_tools.py "Say hello world out loud"
+```
+
+The example sends tools to the LLM, executes the tool calls against Lemonade's endpoints, and feeds the results back in a loop.
+
+## Endpoint Details
```
jeremyfowers: Is all the detail in this section necessary? Seems replicated from the endpoints spec.

docs/tools.md
jeremyfowers: Can you change the PR title and description now that the "experiences" name is not in use?

docs/tools.md
jeremyfowers: Seems like no other documents link to this document?

docs/tools.md
ramkrishna2910: Removed duplication.

docs/tools.md
ramkrishna2910: Moved to omni router title and description.

.claude/settings.local.json
ramkrishna2910: Missed this comment. Added to gitignore

docs/tools.md
ramkrishna2910: Added to mkdocs nav and linked from docs/server/README.md table.

.claude/settings.local.json
jeremyfowers: Ok but the file is still in the PR?

docs/omni-router.md
```diff
@@ -0,0 +1,61 @@
...
+A **Collection** is a preconfigured bundle of models sized for a hardware tier. Selecting a collection in the Lemonade desktop app loads one LLM + one image model + one ASR + one TTS — all the pieces OmniRouter's tools need in a single click.
+
+| Collection | LLM | Image | ASR | TTS |
+|-----------|-----|-------|-----|-----|
+| **Ultra Collection** | Qwen3.5-35B-A3B-GGUF | Flux-2-Klein-9B-GGUF (gen + edit) | Whisper-Large-v3-Turbo | kokoro-v1 |
+| **Lite Collection** | Qwen3.5-4B-GGUF | SD-Turbo (gen only) | Whisper-Tiny | kokoro-v1 |
+
+Collections are hidden from the default `/v1/models` listing so OpenAI-compatible clients don't see "Ultra Collection" as if it were a real model. They surface with `?show_all=true` and appear in the desktop app's model list.
+
+You don't have to use a Collection — any LLM with the `tool-calling` label plus any image / TTS / ASR model works.
```
jeremyfowers: "You" being app developers? Keep in mind that users will read this doc as well.

I also don't know how I could have image/TTS/ASR models work without a collection.

Can you be much more specific here?

docs/omni-router.md
```diff
@@ -0,0 +1,61 @@
...
+You bring the LLM loop. Lemonade brings the tools.
+
+## How it works
+
+1. You describe the tools to your LLM in OpenAI tool-calling format.
+2. The LLM decides which tool to call and with what arguments.
+3. Your client code executes each `tool_call` against the corresponding Lemonade endpoint (`/v1/images/generations`, `/v1/audio/speech`, etc.) and feeds the result back as a `tool` message.
+4. The LLM continues, calling more tools or producing a final response.
+
+This is the same pattern every OpenAI-compatible agent already uses. No custom SDK, no lock-in.
```
jeremyfowers: > No custom SDK, no lock-in.

Can you bend the tone here to be more technical, less marketing?

.claude/settings.local.json
sawansri: Should just be able to `git rm` the file to prevent it from being tracked.

docs/omni-router.md
```diff
@@ -0,0 +1,61 @@
...
+pip install openai
+python examples/lemonade_tools.py "Generate an image of a sunset"
+python examples/lemonade_tools.py "Say hello world out loud"
+```
+
+[`examples/lemonade_tools.py`](../examples/lemonade_tools.py) shows the full agentic loop — tool definitions, LLM call with `tools=[...]`, executing each `tool_call`, and feeding the result back. Fewer than 150 lines of Python.
+
+## Using your own agent
+
+Point your OpenAI-compatible client at `http://localhost:13305/v1`, list models, discover capabilities by label, and wire up the tools you care about. The discovery flow:
```
jeremyfowers: > and wire up the tools you care about

Should this be documented? Or at least "wire up in the style demonstrated in `examples/lemonade_tools.py`"?

src/cpp/resources/server_models.json
```diff
@@ -1351,34 +1351,25 @@
...
-            "gpt-oss-120b-mxfp-GGUF",
-            "Z-Image-Turbo",
-            "Whisper-Large-v3-Turbo",
-            "kokoro-v1"
-        ]
-    },
-    "Lemonade Medium": {
+    "Ultra Collection": {
         "checkpoint": "",
         "recipe": "experience",
```
jeremyfowers: Displays as an "experience" in `lemonade list`:

```
Ultra Collection                        Yes         experience
```

Recipe should be `collection`?

jeremyfowers: Also please `grep` for "experience", I dont think that term should occur anymore?

examples/lemonade_tools.py
```diff
@@ -0,0 +1,152 @@
+"""
+Lemonade Tools — agentic loop example.
```
jeremyfowers: Mixed use of "Lemonade Tools" and "OmniRouter" in docs/scripts is throwing me off. Should the heading here be: "OmniRouter tool calling: agentic loop example"?

examples/lemonade_tools.py
```diff
@@ -0,0 +1,152 @@
...
+
+Demonstrates how to use Lemonade's multimodal endpoints as tools in an
+LLM agentic loop. The LLM decides which tool to call; this script
+executes the tool against Lemonade's API and feeds the result back.
+
+Prerequisites:
+    pip install openai
+    # Lemonade server running with an LLM + image/TTS models loaded
```
jeremyfowers:
```suggestion
    # Lemonade server running on port 13305 with an LLM + image/TTS models loaded
```

hardcoded port in the code

jeremyfowers: Real story: I read this instruction and loaded whatever models I felt like, then read more of the code and saw that models names were hardcoded. Can you make the dev experience smoother?

examples/lemonade_tools.py
```diff
@@ -0,0 +1,152 @@
...
+
+LEMONADE_URL = "http://localhost:13305/v1"
+
+# Edit these to match models you have installed. Defaults are small so
+# they fit on most hardware.
+LLM_MODEL = "Qwen3-4B-Instruct-2507-GGUF"   # any model with the "tool-calling" label
+IMAGE_MODEL = "SD-Turbo"                    # any model with the "image" label
+TTS_MODEL = "kokoro-v1"                     # any model with the "tts" label
+
+# Tool definitions — same format the app uses (from toolDefinitions.json)
```
jeremyfowers: "the app" is ambiguous

provide a path to the file? omit?

examples/lemonade_tools.py
```diff
@@ -0,0 +1,152 @@
+"""
```
jeremyfowers, line 1: Image

Awesome!

src/app/src/renderer/components/panels/LLMChatPanel.tsx
```diff
@@ -193,32 +200,14 @@ const LLMChatPanel: React.FC<LLMChatPanelProps> = ({
...
-    }
-
-    if (experienceMode) {
-      setIsExperienceLayoutActive(true);
-      setModeTransitionClass('mode-transition-to-experience');
-      modeTransitionTimerRef.current = window.setTimeout(() => {
-        setModeTransitionClass('');
-        modeTransitionTimerRef.current = null;
-      }, TRANSITION_MS);
+    if (!experienceMode || !modelsData[selectedModel]) {
```
jeremyfowers: do we still have an experience mode?

src/app/src/renderer/components/panels/LLMChatPanel.tsx
```diff
@@ -847,9 +1047,55 @@ const LLMChatPanel: React.FC<LLMChatPanelProps> = ({
...
+              // uses StreamingAudio for autoplay + scrubbing, while user input
+              // audio uses the simpler MessageAudio playback.
+              if (role === 'assistant') {
+                const mimeByFmt: Record<string, string> = {
+                  wav: 'audio/wav',
+                  mp3: 'audio/mpeg',
+                  m4a: 'audio/mp4',
+                  flac: 'audio/flac',
+                  ogg: 'audio/ogg',
+                  webm: 'audio/webm',
```
jeremyfowers: single source this?

examples/lemonade_tools.py
```diff
@@ -0,0 +1,152 @@
...
+        message = response.choices[0].message
+
+        if not message.tool_calls:
+            print(f"Assistant: {message.content}")
+            break
+
+        messages.append(message)
+
+        for tool_call in message.tool_calls:
+            print(f"  [Tool] {tool_call.function.name}({tool_call.function.arguments})")
```
jeremyfowers, line 167: Future work: I actually really liked that the python script showed the tool calls, and wish that the GUI did too.

src/app/src/renderer/utils/lemonadeTools.ts
```diff
@@ -0,0 +1,314 @@
...
+  const toolList = tools.map(t => `- ${t.function.name}: ${t.function.description}`).join('\n');
+  const systemPrompt = toolDefinitions.system_prompt.replace('{tool_list}', toolList);
+
+  return { tools, systemPrompt, models };
+}
+
+/**
+ * Execute a single Lemonade tool call.
+ */
+export async function executeLemonadeTool(
```
jeremyfowers, line 94: **Claude Code review:** Abort signal isn't wired into any of the tool-call fetches. `handleExperienceChat` forwards `abortControllerRef.current?.signal` only to the `/chat/completions` call; each `serverFetch` below (`/images/edits`, `/images/generations`, `/audio/speech`, `/audio/transcriptions`, `/chat/completions` for vision) runs without a signal. Pressing Stop mid-turn won't cancel an in-flight image or TTS generation, and no `AbortError` will fire the existing catch path. Thread the signal through `ToolExecutionContext` (or a new parameter) and attach it to every `serverFetch` in this file.

src/app/styles.css
```diff
@@ -1719,6 +1719,14 @@ footer {
     animation: pulse-glow 2s ease-in-out infinite;
 }
 
+/* Same dot, but amber + faster pulse to signal "loading, not ready yet".
+   Matches the visual language used for loading in the model list rows. */
+.loaded-model-indicator.loading {
+    background-color: #facc15;
+    animation: pulse 1.5s ease-in-out infinite;
+    filter: drop-shadow(0 0 2px #facc15);
+}
```
jeremyfowers, line 1887: Thank goodness, I actually really wanted this feature in general.

src/cpp/server/model_manager.cpp
```diff
@@ -1968,6 +1990,51 @@ void ModelManager::download_model(const std::string& model_name,
...
+        }
+        LOG(INFO, "ModelManager") << "Downloading " << info.composite_models.size()
+                                  << " component(s) for collection: " << model_name << std::endl;
+
+        // Wrap the callback so recursive per-component downloads don't each
+        // emit a "complete" event — the SSE stream should only see one final
+        // completion after every component finishes.
+        DownloadProgressCallback forward = nullptr;
+        if (progress_callback) {
+            forward = [&progress_callback](const DownloadProgress& p) -> bool {
```
jeremyfowers: **Claude Code review:** Capturing `progress_callback` by reference here is load-bearing on the assumption that this lambda never outlives the enclosing stack frame. It happens to be safe today because `forward` is only consumed by the synchronous `download_model()` recursion below, but `DownloadProgressCallback` is a `std::function` — copying it in (`[progress_callback]`) is cheap and removes a dangling-reference footgun if someone later stores or defers the callback. Recommend capturing by value.

src/app/src/renderer/utils/serverConfig.ts
```diff
@@ -314,9 +314,11 @@ class ServerConfig {
     }
 
     try {
+      console.log('[serverFetch]', options?.method || 'GET', fullUrl);
```
jeremyfowers: **Claude Code review:** Debug logging left in a hot path. `serverFetch` runs on every HTTP call the app makes; this emits a `console.log` for every request and a `console.error` on every failure for the lifetime of the app. The tagged `[LLMChat]` logs in `LLMChatPanel.tsx` (lines 753, 766, 834, 843, 847, 897) are similar. Strip before merge, or gate them behind a debug flag.

src/cpp/server/router.cpp
```diff
@@ -1,4 +1,6 @@
 #include "lemon/router.h"
+#include "lemon/utils/http_client.h"
+#include "lemon/utils/json_utils.h"
```
jeremyfowers: **Claude Code review:** Neither of these includes is referenced anywhere in `router.cpp` — `HttpClient` and `JsonUtils` symbols don't appear in the file. Looks like leftover from an earlier iteration; remove both lines.

src/app/src/renderer/ModelManager.tsx
```diff
@@ -1037,8 +1094,14 @@ const [searchQuery, setSearchQuery] = useState('');
...
-    const isDownloaded = modelsData[modelName]?.downloaded ?? false;
-    const isLoaded = loadedModels.has(modelName);
+    const info = modelsData[modelName];
+    // collections: downloaded when all components are downloaded,
+    // loaded when all components are loaded. Fall back to naive checks
+    // for regular models.
+    const isDownloaded = isModelEffectivelyDownloaded(modelName, info, modelsData);
+    const isLoaded = isCollectionModel(info)
+      ? isCollectionFullyLoaded(modelName, modelsData, loadedModels)
+      : loadedModels.has(modelName);
```
jeremyfowers: **Claude Code review:** Redundant manual branch. `isModelEffectivelyDownloaded` above already dispatches on `isCollectionModel`; the load side does the same check inline. Consider a parallel `isModelEffectivelyLoaded(modelName, info, modelsData, loadedModels)` helper in `collectionModels.ts` so `getModelStatus` doesn't need to know about the collection branching at all.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
```diff
@@ -444,6 +429,273 @@ const LLMChatPanel: React.FC<LLMChatPanelProps> = ({
...
+
+    // artifacts: generated this turn — rendered in the assistant response.
+    const artifacts: Artifact[] = [];
+
+    // Agentic loop
+    const llmMessages = [...processedMessages];
+
+    for (let iteration = 0; iteration < MAX_ITERATIONS; iteration++) {
+      // Show thinking indicator while waiting for LLM response
+      const thinkingText = iteration === 0 ? 'Thinking...' : 'Thinking...';
```
jeremyfowers: **Claude Code review:** Identical branches — both arms of the ternary produce `'Thinking...'`. Collapse to `const thinkingText = 'Thinking...';` (or drop the variable entirely and inline the literal).

src/cpp/server/model_manager.cpp
```diff
@@ -1322,6 +1323,23 @@ void ModelManager::update_model_in_cache(const std::string& model_name, bool dow
...
+                          model_name) == entry.composite_models.end()) {
+                continue;
+            }
+            bool new_state = check_composite_downloaded(entry, models_cache_);
+            if (entry.downloaded != new_state) {
+                entry.downloaded = new_state;
+                LOG(INFO, "ModelManager") << "Experience '" << name
+                          << "' downloaded=" << new_state << " (dependent on " << model_name << ")" << std::endl;
+            }
+        }
```
jeremyfowers: Yeah there's a lot of "Experience" references that are kinda headless now.

src/cpp/server/model_manager.cpp
```diff
@@ -1356,11 +1374,15 @@ std::map<std::string, ModelInfo> ModelManager::get_downloaded_models() {
     // Build cache if needed
     build_cache();
 
-    // Filter and return only downloaded models
+    // Filter and return only downloaded, non-hidden models. Hidden models
+    // are things like experience bundles which are Lemonade-specific UX
```
jeremyfowers: "experience bundles" two dead terms back to back

src/cpp/server/model_manager.cpp
```diff
@@ -1356,11 +1374,15 @@ std::map<std::string, ModelInfo> ModelManager::get_downloaded_models() {
...
+    // abstractions — OpenAI-compatible clients (AnythingLLM, Continue,
+    // etc.) shouldn't see them as if they were regular LLMs. The desktop
```
jeremyfowers: do we need to name specific apps/companies in our comments?

src/app/src/renderer/components/panels/LLMChatPanel.tsx
```diff
@@ -444,6 +429,273 @@ const LLMChatPanel: React.FC<LLMChatPanelProps> = ({
...
+      // Content is an array — strip binary data (images/audio) from all messages
+      // For user messages: extract into context for tool use
+      // For assistant messages: replace with placeholders so LLM doesn't choke
+      const isUser = msg.role === 'user';
+      const newContent: any[] = [];
+      for (const item of msg.content) {
+        if (item.type === 'input_audio' && 'input_audio' in item) {
+          if (isUser) {
+            const fmt = item.input_audio.format || 'wav';
+            const fmtToMime: Record<string, string> = {
```
jeremyfowers: **Claude Code review:** This mime→format map is defined three times in this file — here (`fmtToMime` in `handleExperienceChat`), again as `fmtByMime` at line ~675 in `buildFinalContent`, and again as `mimeByFmt` at line ~1084 in `renderMessageContent`. There's also a parallel `AUDIO_FORMAT_BY_MIME` constant up at line 45. Hoist a single module-level lookup (both directions) and reuse — the current duplication is a recipe for the lists drifting out of sync the next time someone adds a codec.

src/cpp/server/model_manager.cpp
```diff
@@ -1003,6 +1003,7 @@ void ModelManager::build_cache() {
         info.recipe = JsonUtils::get_or_default<std::string>(value, "recipe", "");
         info.suggested = JsonUtils::get_or_default<bool>(value, "suggested", false);
         info.size = JsonUtils::get_or_default<double>(value, "size", 0.0);
+        info.hidden = JsonUtils::get_or_default<bool>(value, "hidden", false);
```
jeremyfowers: Do we need an entirely new property for this? Why not just always hide `recipe=collection`?

src/app/src/renderer/utils/lemonadeTools.ts
```diff
@@ -0,0 +1,314 @@
...
+  }
+  throw new Error(data.error?.message || 'Transcription failed');
+}
+
+async function executeVisionTool(
+  args: Record<string, any>,
+  model: string,
+  context: ToolExecutionContext,
+): Promise<ToolExecutionResult> {
+  let imageUrl = args.image_url || '';
```
jeremyfowers: **Claude Code review:** `args.image_url` is passed straight into the chat completion without any validation. If the LLM hallucinates an `http://` or `file://` URL, whether that turns into a network fetch depends on backend leniency. `renderMessageContent` now explicitly gates assistant image URLs on `data:image/` for exactly this reason — mirror that guard here, and fall back to `context.extractedImages` if the LLM-provided value isn't a `data:image/` URL.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
```diff
@@ -1120,6 +1349,7 @@ const LLMChatPanel: React.FC<LLMChatPanelProps> = ({
             audio={uploadedAudio}
             onRemove={uploadedAudioHandlers.remove}
           />
+
```
jeremyfowers: **Claude Code review:** Cosmetic — stray blank line between `AudioPreviewList` and the `<textarea>`. Doesn't match the formatting used elsewhere in this JSX block; drop the empty line.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
jeremyfowers, line 1: **Claude Code review:** `toggleMicDictation` and `stopMicDictation` were deleted earlier in this PR, but the state they drove is still here — `isMicRecording` / `setIsMicRecording` at line 123 and `speechRecognitionRef` at line 143 are now unreferenced anywhere in the file. Remove both declarations.

src/app/styles.css
jeremyfowers, line 1: **Claude Code review:** Dead CSS. This PR removed the `chat-window-experience-mode` / `chat-window-experience` / `experience-topbar*` / `experience-empty-*` / `experience-refresh-*` / `experience-model-name` / `mode-transition-*` class applications from JSX (ChatWindow.tsx and LLMChatPanel.tsx), but the rule blocks in this file — including the `--experience-composer-*` custom-property token block at line 3527 and the whole `.experience-topbar*` stack starting around line 3057 — are still here. That's ~100 lines of orphaned rules whose selectors will never match anymore. Sweep them in this PR so the cleanup lands together.

jeremyfowers (changes requested):
It's working beautifully! My comments are all about cohesion and simplicity now. Let me know when its time to do one last code/doc scan for cohesion.

jeremyfowers:
image
image

Almost a perfect demo! Except that it forgets to narrate the story after the first turn. Wonder if this works better as a custom app where the dungeon master role is encoded in the system prompt?

src/cpp/resources/server_models.json
```diff
@@ -1351,34 +1351,25 @@
...
-        "recipe": "experience",
-        "suggested": false,
+        "recipe": "collection",
+        "suggested": true,
+        "hidden": true,
         "composite_models": [
-            "gpt-oss-20b-mxfp4-GGUF",
-            "SDXL-Turbo",
+            "Qwen3.5-35B-A3B-GGUF",
+            "Flux-2-Klein-9B-GGUF",
```
jeremyfowers, line 1358:
```suggestion
            "Flux-2-Klein-4B",
```

Maybe something is wrong with my system, but the 9B is excruciatingly slow.

jeremyfowers:
image

Better! This is sick.

.claude/settings.local.json
ramkrishna2910: Fixed in c2f18046 — untracked via `git rm --cached`. Also added `.claude/settings.local.json` to `.gitignore` in eb6d5d97 (the prior `.settings.local.json` pattern was too narrow and didn't match the nested path).

docs/omni-router.md
ramkrishna2910: Rewrote this section in f0b9309a. Collections are now the default path in the docs, with the substitution-without-a-Collection option scoped explicitly as a developer note with the specific label requirements.

docs/omni-router.md
ramkrishna2910: Expanded into a four-step recipe in d56cb974 with an explicit reference to `examples/lemonade_tools.py`.

src/cpp/resources/server_models.json
ramkrishna2910: Fixed in 610d47a5 — renamed the recipe identifier to `collection` across `server_models.json`, C++ model_manager/server/model_types, and the frontend. `lemonade list` now shows `collection`.

examples/lemonade_tools.py
ramkrishna2910: Renamed in f0b9309a to `OmniRouter tool calling: agentic loop example`.

examples/lemonade_tools.py
ramkrishna2910: Fixed in f0b9309a. Added `preflight_models()` that queries `/v1/models?show_all=true` and fails loudly with a specific fix if any of `LLM_MODEL` / `IMAGE_MODEL` / `TTS_MODEL` isn't installed — the error message tells you to either install the Lite Collection from the desktop app or edit the constants at the top of the file.

examples/lemonade_tools.py
ramkrishna2910: Replaced with the explicit path `src/app/src/renderer/utils/toolDefinitions.json` in f0b9309a.

examples/lemonade_tools.py
ramkrishna2910: 🙌

src/app/src/renderer/components/panels/LLMChatPanel.tsx
ramkrishna2910: Renamed to `collectionMode` across the frontend in 610d47a5 (and the recipe string to `collection`). No more `experience` anywhere except the `experienceRailWidth` UI layout constant, which is natural-English usage unrelated to the model recipe.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
ramkrishna2910: Single-sourced in f0b9309a via `mimeForFormat` / `formatForMime` helpers on top of `AUDIO_FORMAT_BY_MIME` / `AUDIO_MIME_BY_FORMAT`. All three inline copies (in `handleCollectionChat`, `buildFinalContent`, `renderMessageContent`) now go through the helpers.

examples/lemonade_tools.py
ramkrishna2910: Noted as future work — agreed, showing tool calls in the GUI chat timeline would make the flow more discoverable. Leaving this thread unresolved as a tracking note is fine; want me to open a follow-up issue instead?

src/app/src/renderer/utils/lemonadeTools.ts
ramkrishna2910: Fixed in f0b9309a. `executeLemonadeTool` now takes an `AbortSignal` parameter and threads it through to every tool fetch (`/images/generations`, `/images/edits`, `/audio/speech`, `/audio/transcriptions`, the vision `/chat/completions`). `handleCollectionChat` passes `abortControllerRef.current?.signal` at the call site.

src/app/styles.css
ramkrishna2910: Thanks! Green/amber status dot with pulse animation for loaded vs loading — will keep that style consistent elsewhere.

src/cpp/server/model_manager.cpp
ramkrishna2910: Fixed in f0b9309a — capture by value instead. `std::function` is copyable, so the copy is free, and it removes the latent dangling-reference hazard if the recursion pattern ever changes.

src/app/src/renderer/utils/serverConfig.ts
ramkrishna2910: Removed in f0b9309a. The `[serverFetch]` console.log in the hot path is gone; the `[LLMChat]` per-request log and pre-flight log in `LLMChatPanel` are also gone. Error paths still log where genuinely useful.

src/cpp/server/router.cpp
ramkrishna2910: Removed both in f0b9309a.

src/app/src/renderer/ModelManager.tsx
ramkrishna2910: Extracted `isModelEffectivelyLoaded` in f0b9309a mirroring `isModelEffectivelyDownloaded`. Both `getModelStatus` and `getModelLoadedState` now go through it — collection fully-loaded logic lives in one place.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
ramkrishna2910: Collapsed in f0b9309a to `const thinkingText = 'Thinking...';`.

src/cpp/server/model_manager.cpp
ramkrishna2910: Scrubbed in f0b9309a. All leftover `Experience` / `bundle` references in comments and log messages are now `Collection` / nothing.

src/cpp/server/model_manager.cpp
ramkrishna2910: Scrubbed in f0b9309a — comment now reads `Lemonade-specific (a virtual entry that loads multiple...)`.

src/cpp/server/model_manager.cpp
ramkrishna2910: Good call — removed the AnythingLLM/Continue name-drop from the comment in f0b9309a.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
ramkrishna2910: Deduped in f0b9309a. Added `mimeForFormat(fmt)` / `formatForMime(mime)` helpers on top of the existing `AUDIO_FORMAT_BY_MIME` / new `AUDIO_MIME_BY_FORMAT` tables. The three inline copies are all replaced with helper calls.

src/cpp/server/model_manager.cpp
ramkrishna2910: Agreed, simpler. Dropped the `hidden` flag entirely in f0b9309a — `ModelInfo.hidden` gone, no more `"hidden": true` in `server_models.json`, `/v1/models` just filters out `recipe == "collection"` directly.

src/app/src/renderer/utils/lemonadeTools.ts
ramkrishna2910: Fixed in f0b9309a. `executeVisionTool` now rejects any `args.image_url` that doesn't start with `data:image/`, falling back to the most recent user-uploaded data URL. If neither is available, returns `'No image available to analyze.'` without calling the backend.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
ramkrishna2910: Fixed in f0b9309a.

src/app/src/renderer/components/panels/LLMChatPanel.tsx
ramkrishna2910: Removed in f0b9309a — `isMicRecording` / `setIsMicRecording` and `speechRecognitionRef` are gone.

src/app/styles.css
ramkrishna2910: Removed in f0b9309a — 36 dead rules stripped (`chat-window-experience*`, `experience-topbar*`, `experience-empty*`, `experience-refresh*`, `experience-model-name`, `experience-unload-icon-button`, `mode-transition-*`, and the `experience-composer` custom-property block).

src/cpp/resources/server_models.json
ramkrishna2910: Noted — 9B is the gen+edit pair we picked for Ultra because it produces noticeably better results than 4B on the test prompts we ran, so I'd rather keep Ultra at 9B and make sure Lite (SD-Turbo) is the fast path. Happy to revisit if we get more reports of the 9B being slow across systems.

ramkrishna2910:
> image image
> Almost a perfect demo! Except that it forgets to narrate the story after the first turn. Wonder if this works better as a custom app where the dungeon master role is encoded in the system prompt?

Yes, ideally the user builds custom apps with their own prompts.

src/cpp/resources/server_models.json
ckuethe: Using RX7900XT
4B, 1024x1024, 8.8s
4B, 512x512, 2.8s
9B, 1024x1024, 25.3s
9B, 512x512, 8.4s
SD, 1024x1024, 3.9s
SD, 512x512, 1.2s

Flux image quality doesn't seem noticeably different to me

lemonade_red_pandas_1776895917378
lemonade_red_panda_1776895913653
lemonade_red_pandas_1776895909289
lemonade_red_panda_1776895897771
lemonade_red_panda_1776896031091
lemonade_red_pandas_1776896035616

examples/lemonade_tools.py
jeremyfowers: Hi robot lmao

src/cpp/resources/server_models.json
jeremyfowers: Thanks @ckuethe ! I'm not happy with the Flux 9B speed on my stx halo, but its hard to argue with the quality. Let's see what the feedback is. I know I'm gonna want to customize my omnirouter models later...

jeremyfowers (approved):
Lets gooooo!

PR #1581, opened 2026-04-08
feat: integrate Qwen2.5-Omni models with native audio input support

ramkrishna2910:
## Summary

Adds Qwen2.5-Omni-7B-GGUF and Qwen2.5-Omni-3B-GGUF to the model registry with text + vision + audio input support on llama.cpp.

## Why the hf_load flag

llama-server's manual-load path (`-m <gguf>` + `--mmproj <mmproj>`) rejects `input_audio` content parts on Qwen2.5-Omni with "unsupported content[].type". The `-hf` path drives the dual-clip (vision+audio) context correctly. To opt in without disturbing other models:

- New `hf_load: true` field on `ModelInfo` (parsed from `server_models.json`).
- `LlamaCppServer::load()` branches on that flag and passes `-hf <checkpoint>` instead of `-m` + `--mmproj`. Everything else is unchanged.
- Only the two Qwen2.5-Omni entries set `hf_load: true`; every other model takes the original code path.

Lemonade's pull still populates `~/.cache/huggingface/hub/`; llama-server with `-hf` reads from that same HF hub cache, so there is no duplicate download.

image


ramkrishna2910:
@claude review

jeremyfowers (approved):
Image

Works for me!

PR #1583, opened 2026-04-08
fix: use wss:// for WebSocket connections when served over HTTPS

ianbmacdonald:
## Summary

- WebSocket URLs in the web app were hardcoded to `ws://`, causing mixed-content errors when Lemonade is served behind an HTTPS reverse proxy
- Adds a `getWebSocketProtocol()` helper that derives `ws` or `wss` from the server base URL protocol
- Both WebSocket clients (log streaming and realtime transcription) now use the correct protocol automatically

This is the narrow protocol-only fix. A broader `external_url` config for full reverse proxy support (including same-origin WebSocket routing) is available on [`feat/external-url-config`](https://github.com/ianbmacdonald/lemonade/tree/feat/external-url-config) and described in [#472 (comment)](https://github.com/lemonade-sdk/lemonade/issues/472#issuecomment-4207315161).

## Test plan

- [ ] Load web app over `http://` — WebSocket connections should use `ws://` (unchanged behavior)
- [ ] Load web app over `https://` via reverse proxy — WebSocket connections should use `wss://`, no mixed-content errors in browser console
- [ ] Verify log streaming and realtime transcription connect successfully in both modes

Fixes #472

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #631, opened 2025-12-01
[Bug] Web interface unusable through reverse proxy


mr-valente:
Set up my lemonade server and all is working well, except the web interface isn't accessible through a reverse proxy like caddy. 

I have a very simple entry in my Caddyfile for lemonade, and the interesting thing is that it works just fine for API calls (e.g. can point open-webui at lemonade through the reverse proxy and all is well). However, when I try to access the lemonade web interface, the page loads, but I get this error:

"Error loading models: Failed to fetch
Check the Lemonade Server logs on the browser or via the system tray app for more information."

When I check the logs, there's really nothing interesting to report:

```text
Starting Lemonade Server...
  Version: 9.0.4
  Port: 32910
  Host: 0.0.0.0
  Log level: debug
  Context size: 4096
[Server] HTTP server initialized with thread pool (8 threads)
[Server] Debug logging enabled - subprocess output will be visible
[Server] Static files mounted from: /usr/local/share/lemonade-server/resources/static
[Server] Routes setup complete
[Server] Starting on 0.0.0.0:32910
[Server DEBUG] Creating new thread pool with 8 threads
[Server PRE-ROUTE] GET /api/v1/health
[Server] GET /api/v1/health - 200
[Server PRE-ROUTE] GET /api/v1/models
[ModelManager] Building models cache...
[Server] Collecting system info
[Server PRE-ROUTE] GET /api/v1/models
[ModelManager] Backend availability:
  - NPU hardware: No
  - FLM available: No
  - OGA available: No
[ModelManager] Cache built: 33 total, 3 downloaded
[Server] GET /api/v1/models - 200
[Server] GET /api/v1/models - 200
[Server PRE-ROUTE] GET /api/v1/logs/stream
[Server] Starting log stream for: /tmp/lemonade-server.log
[Server] Log stream connection opened
[Server PRE-ROUTE] GET /webapp.html
[Server] GET /webapp.html - 200
[Server PRE-ROUTE] GET /static/styles.css
[Server] GET /static/styles.css - 200
[Server PRE-ROUTE] GET /static/favicon.ico
[Server] GET /static/favicon.ico - 200
[Server PRE-ROUTE] GET /static/js/shared.js
[Server] GET /static/js/shared.js - 200
[Server PRE-ROUTE] GET /static/js/chat.js
[Server PRE-ROUTE] GET /static/js/models.js
[Server] GET /static/js/chat.js - 200
[Server] GET /static/js/models.js - 200
[...]

PR #1584, opened 2026-04-08
feat: add external_url config for reverse proxy deployments

### LINKED ISSUE #472: Frontend does not follow url that loaded the page
I tried running lemonade server through a reverse proxy to add ssl and it also sets up domain routing. However, the frontend seems to always use http and append the port, no matter what url is used to load the page. For instance, if I load the page with `https://name.com` the frontend makes a request to `http://name.com:8000/api/v1/health` using whatever port the server is listening on. This causes a cross-origin error and the frontend does not function.

ianbmacdonald:
## Summary

- Adds an `external_url` server config key for specifying the public browser-facing URL when Lemonade runs behind a reverse proxy
- When set, the web app derives all REST and WebSocket URLs from `external_url` instead of internal hostnames and ports
- WebSocket clients route through the proxy origin (e.g. `wss://example.com/realtime`) rather than connecting directly to `:9000`
- Direct-connect and Electron modes are unaffected — `isExternalUrl()` is scoped to web-app mode only
- Includes `ws://` vs `wss://` protocol detection for direct-connect mode
- Emits a startup warning when `external_url` is set but `websocket_port` is still `"auto"`
- Documents the reverse proxy deployment contract with example nginx config

Full context in [#472 (comment)](https://github.com/lemonade-sdk/lemonade/issues/472#issuecomment-4207315161).

### Changes

| File | What |
|------|------|
| `defaults.json` | Adds `external_url: ""` |
| `runtime_config.h` / `.cpp` | Getter, regex validation (http/https, no query/fragment), runtime-changeable |
| `server.cpp` | Injects `external_url` via JSON-serialized `<script>` tag; proxy misconfiguration warning; side-effect handling |
| `serverConfig.ts` | Resolution: `external_url` > `window.location.origin` > `localhost`; `getWebSocketUrl()` helper; `isExternalUrl()` scoped to web-app mode |
| `websocketClient.ts` / `logWebSocketClient.ts` | Use centralized `getWebSocketUrl()` |
| `configuration.md` | Documents `external_url`, reverse proxy section with nginx example |
| `server_spec.md` | Updates WebSocket connection docs for proxied mode |

### Deployment

```bash
lemonade config set external_url=https://lemonade.example.com websocket_port=9000
```

```nginx
location / {
    proxy_pass http://127.0.0.1:13305;
}
location /realtime {
    proxy_pass http://127.0.0.1:9000;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
}
location /logs/stream {
    proxy_pass http://127.0.0.1:9000;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
}
```

## Test plan

- [ ] No `external_url` set — verify local, Electron, and direct-connect modes work exactly as before
- [ ] Set `external_url=http://localhost:13305` — web app should use that as base URL, WS connects via origin
- [ ] Set `external_url=https://example.com` behind HTTPS proxy — REST uses `https://`, WS uses `wss://`, no mixed-content errors
- [ ] Set `external_url=https://example.com/lemonade` with path prefix — verify `/lemonade/realtime` and `/lemonade/logs/stream` paths
- [ ] Set `external_url` with `websocket_port=auto` — verify startup warning is logged
- [ ] `lemonade config set external_url=...` at runtime — verify side-effect log message and warning re-check
- [ ] Electron app with `--base-url` — verify it still uses `websocket_port` from `/health` (no regression)

Fixes #472

### DIFF
--- a/src/app/src/renderer/utils/serverConfig.ts
+++ b/src/app/src/renderer/utils/serverConfig.ts
@@ -14,6 +14,45 @@ type UrlChangeListener = (url: string, apiKey: string) => void;
 
 const trimTrailingSlashes = (value: string): string => value.replace(/\/+$/, '');
 const ensureLeadingSlash = (value: string): string => value.startsWith('/') ? value : `/${value}`;
+const normalizeWebAppBasePath = (pathname: string): string => {
+  let normalizedPath = trimTrailingSlashes(pathname || '/');
+
+  if (!normalizedPath || normalizedPath === '/') {
+    return '';
+  }
+
+  if (normalizedPath.endsWith('/index.html')) {
+    normalizedPath = trimTrailingSlashes(
+      normalizedPath.slice(0, -'/index.html'.length),
+    );
+  }
+
+  if (normalizedPath === '/web-app') {
+    return '';
+  }
+
+  if (normalizedPath.endsWith('/web-app')) {
+    normalizedPath = trimTrailingSlashes(
+      normalizedPath.slice(0, -'/web-app'.length),
+    );
+  }
+
+  return normalizedPath === '/' ? '' : normalizedPath;
+};
+
+const deriveWebAppBaseUrl = (): string | null => {
+  if (typeof window === 'undefined') {
+    return null;
+  }
+
+  const origin = window.location?.origin;
+  if (!origin || origin === 'null') {
+    return null;
+  }
+
+  const basePath = normalizeWebAppBasePath(window.location?.pathname ?? '/');
+  return `${trimTrailingSlashes(origin)}${basePath}`;
+};
 
 class ServerConfig {
   private port: number = 13305;
@@ -46,51 +85,98 @@ class ServerConfig {
 
     try {
       // Get API Key if available
-      if (typeof window !== 'undefined'&& window.api?.getServerAPIKey) {
+      if (typeof window !== 'undefined' && window.api?.getServerAPIKey) {
         this.apiKey = await window.api.getServerAPIKey();
       }
 
-      // Check if an explicit base URL was configured (external_url, --base-url, or env var)
-      if (typeof window !== 'undefined' && window.api?.getServerBaseUrl) {
-        const baseUrl = await window.api.getServerBaseUrl();
-        if (baseUrl) {
-          console.log('Using explicit server base URL:', baseUrl);
-          this.explicitBaseUrl = baseUrl;
-          // Only treat as external URL in web-app mode (server-injected external_url).
-          // Electron's --base-url still needs the separate websocket_port.
-          this.hasExternalUrl = !!(window.api?.isWebApp);
-          this.initialized = true;
-          return;
-        }
+      let configured = false;
+
+      if (typeof window !== 'undefined' && window.api?.isWebApp) {
+        configured = await this.initializeWebAppBaseUrl();
       }
 
-      // In web app mode, fall back to the current origin as the server base URL
-      if (typeof window !== 'undefined' && window.api?.isWebApp) {
-        const origin = window.location?.origin;
-        if (origin && origin !== 'null') {
-          const trimmedOrigin = trimTrailingSlashes(origin);
-          console.log('Using web app origin as server base URL:', trimmedOrigin);
-          this.explicitBaseUrl = trimmedOrigin;
-          this.initialized = true;
-          return;
+      // In Tauri mode, use an explicit base URL when configured via --base-url
+      // or the environment. The separate websocket_port still applies there.
+      if (!configured && typeof window !== 'undefined' && window.api?.getServerBaseUrl) {
+        const baseUrl = await window.api.getServerBaseUrl();
+        if (baseUrl) {
+          const normalizedBaseUrl = trimTrailingSlashes(baseUrl);
+          console.log('Using explicit server base URL:', normalizedBaseUrl);
+          this.explicitBaseUrl = normalizedBaseUrl;
+          this.hasExternalUrl = false;
+          configured = true;
         }
       }
 
       // No explicit URL - use localhost with port discovery
-      if (typeof window !== 'undefined' && window.api?.getServerPort) {
+      if (!configured && typeof window !== 'undefined' && window.api?.getServerPort) {
         const port = await window.api.getServerPort();
         if (port) {
           this.port = port;
         }
+
+        console.log('Using localhost mode with port:', this.port);
       }
 
-      console.log('Using localhost mode with port:', this.port);
       this.initialized = true;
     } catch (error) {
       console.error('Failed to initialize server config:', error);
       this.initialized = true;
     }
 
+    this.registerEventListeners();
+  }
+
+  private async initializeWebAppBaseUrl(): Promise<boolean> {
+    const browserBaseUrl = deriveWebAppBaseUrl();
+    if (!browserBaseUrl) {
+      return false;
+    }
+
+    console.log('Using web app request base URL:', browserBaseUrl);
+    this.explicitBaseUrl = browserBaseUrl;
+    this.hasExternalUrl = false;
+
+    await this.loadExternalUrlFromSystemInfo(browserBaseUrl);
+    return true;
+  }
+
+  private async loadExternalUrlFromSystemInfo(browserBaseUrl: string): Promise<void> {
+    try {
+      const headers = this.apiKey
+        ? { Authorization: `Bearer ${this.apiKey}` }
+        : undefined;
+      const response = await fetch(
+        `${browserBaseUrl}/api/v1/system-info`,
+        headers ? { headers } : undefined,
+      );
+
+      if (!response.ok) {
+        console.warn(
+          'Failed to fetch /system-info for external_url:',
+          response.status,
+        );
+        return;
+      }
+
+      const systemInfo = await response.json();
+      const externalUrl = typeof systemInfo.external_url === 'string'
+        ? trimTrailingSlashes(systemInfo.external_url)
+        : '';
+
+      if (!externalUrl) {
+        return;
+      }
+
+      console.log('Using external_url from /system-info:', externalUrl);
+      this.explicitBaseUrl = externalUrl;
+      this.hasExternalUrl = true;
+    } catch (error) {
+      console.warn('Failed to fetch /system-info for external_url:', error);
+    }
+  }
+
+  private registerEventListeners() {
     // Register event listeners AFTER the first await-cycle so window.api is
     // guaranteed to be installed. tauriShim.ts installs window.api via a
     // fire-and-forget async call that completes on the microtask queue; the
@@ -163,9 +249,10 @@ class ServerConfig {
   }
 
   /**
-   * Whether the server base URL came from the web-app's server-injected
-   * external_url config. When true, browser WebSocket clients should derive
-   * their public URLs from that base URL instead of using websocket_port.
+   * Whether the active web-app base URL came from the server's external_url
+   * config surfaced through /system-info. When true, browser WebSocket
+   * clients should derive public URLs from that base URL instead of using
+   * websocket_port directly.
    */
   isExternalUrl(): boolean {
     return this.hasExternalUrl;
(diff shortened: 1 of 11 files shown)


ianbmacdonald:
@Geramy here you go. @superm1 it looks like there is the proper guard to prevent regressing Electron users.  YMMV; Test plan not executed

src/app/src/renderer/utils/serverConfig.ts
```diff
@@ -333,6 +348,27 @@ export const getServerHost = () => serverConfig.getServerHost();
...
+    const baseUrl = new URL(serverConfig.getServerBaseUrl());
+    baseUrl.protocol = baseUrl.protocol === 'https:' ? 'wss:' : 'ws:';
+    const basePath = trimTrailingSlashes(baseUrl.pathname);
+    baseUrl.pathname = `${basePath}${normalizedPath}` || normalizedPath;
+    baseUrl.search = queryString ? `?${queryString}` : '';
+    baseUrl.hash = '';
+    return baseUrl.toString();
+  }
+
+  const wsUrl = new URL(`${getWebSocketProtocol()}://${serverConfig.getServerHost()}:${wsPort}`);
```
Geramy: we should only wsPort when its pinned in the configuration and by default it shouldn't be. My logic is on the web app the server hosting the web app is the same as the server hosting the api, so in this case we already have the port based on the requestUri so we don't need to specify the wsPort at all for web app. If you could get this fixed we can merge.

src/cpp/resources/defaults.json
```diff
@@ -2,6 +2,7 @@
   "config_version": 1,
   "port": 13305,
   "host": "localhost",
+  "external_url": "",
```
Geramy, line 5: if this is specified then we can pull from this instead of doing the web app assumption. Same concept as above.

ianbmacdonald:
add @sofiageo's comment

src/cpp/server/server.cpp
```diff
@@ -554,7 +572,7 @@ void Server::setup_static_files(httplib::Server &web_server) {
     // Check if web app directory exists
     if (fs::exists(web_app_dir) && fs::is_directory(web_app_dir)) {
         // Create a handler for serving web app index.html for SPA routing
-        auto serve_web_app_html = [web_app_dir](const httplib::Request&, httplib::Response& res) {
+        auto serve_web_app_html = [this, web_app_dir](const httplib::Request&, httplib::Response& res) {
```
Geramy: instead of passing this we should just pass the required parameter unless we need "this" ? thoughts?

Geramy (commented):
I like it generally. Please see changes.

src/app/src/renderer/utils/serverConfig.ts
ianbmacdonald: Better for proxied deployments where the proxy already fronts both HTTP and WS paths and external_url is unset

ianbmacdonald: Not sure I completely understand, but if we don't set wsPort doesn't that mean WS URL becomes ws://localhost:13305/realtime   and then it breaks the locally hosted web app without a proxy.  “don’t use wsPort” means “derive WS from the browser origin/path”?

src/cpp/server/server.cpp
ianbmacdonald: this looks like hygiene; in we go

src/cpp/resources/defaults.json
ianbmacdonald: this is the effective behavior with this change

ianbmacdonald:
Okay, I was able to verify this works with caddy 
image

Then remove settings, and reverse proxy, and default path with the webapp still works
image

ianbmacdonald:
## Testing this branch

### Prerequisites

- A reverse proxy with TLS (nginx, Caddy, etc.)
- Lemonade built from this branch

### Quick test with Caddy

**1. Configure lemonade:**

```bash
lemonade config set \
  external_url=https://<YOUR_IP>:8443 \
  host=127.0.0.1 \
  websocket_port=9000
```

- `external_url` — the public HTTPS URL browsers will use
- `host=127.0.0.1` — bind lemonade to loopback so only the proxy serves external traffic
- `websocket_port=9000` — pin the WebSocket port so the proxy target is stable

**2. Install and configure Caddy:**

```bash
sudo apt install caddy
```

`/etc/caddy/Caddyfile`:

```
https://<YOUR_IP>:8443 {
	tls internal

	reverse_proxy /realtime http://127.0.0.1:9000 {
		header_up Connection {http.request.header.Connection}
		header_up Upgrade {http.request.header.Upgrade}
	}

	reverse_proxy /logs/stream http://127.0.0.1:9000 {
		header_up Connection {http.request.header.Connection}
		header_up Upgrade {http.request.header.Upgrade}
	}

	reverse_proxy http://127.0.0.1:13305
}
```

```bash
sudo systemctl restart caddy
```

**3. Verify:**

Open `https://<YOUR_IP>:8443/` in a browser (accept the self-signed cert warning), then:

- Open DevTools → Console: should show `Using explicit server base URL: https://<YOUR_IP>:8443`
- Open the logs panel — should connect and stream
- DevTools → Network → filter `WS`: connection URL should be `wss://<YOUR_IP>:8443/logs/stream` (routed through the proxy, no separate port 9000 exposed)
- No mixed-content errors in the console

**4. Cleanup** (to restore default behavior):

```bash
lemonade config set external_url= host=0.0.0.0 websocket_port=auto
sudo systemctl stop caddy
```

### What to look for

| Check | Expected |
|-------|----------|
| REST API calls in Network tab | `https://<YOUR_IP>:8443/api/v1/...` |
| WebSocket URL | `wss://<YOUR_IP>:8443/logs/stream` |
| Mixed-content errors | None |
| Log streaming | Connected, entries flowing |
| Without `external_url` set | Behavior identical to current main branch |

src/app/src/renderer/utils/serverConfig.ts
ianbmacdonald: So I tested a branch (https://github.com/ianbmacdonald/lemonade/tree/test-geramy-comments) with this change, and the proxy works, but when I clear the config, and drop external_url, the webapp breaks

`lemonade config set external_url= host=0.0.0.0 websocket_port=auto`

image

src/cpp/server/server.cpp
```diff
@@ -721,10 +749,20 @@ window.api = {
 </script>
 )";
 
-            // Insert mock API before the closing </head> tag
+            // Build a small dynamic script that exposes the external_url config
+            // value (if set) so the frontend can derive its base URL from it.
```
superm1: This feels like a hack.  Shouldn't it be published in /system-info and then used by frontend like any other setting?

ianbmacdonald:
Ran the remaining headless-eligible checks against commit `3faedcae`.

From the PR checklist:
- [x] No `external_url` set: disposable `ai2` server on `25405`/`29000` reported `external_url=""`; direct `ws://192.168.79.18:29000/logs/stream` returned `logs.snapshot`, and direct `ws://192.168.79.18:29000/realtime?model=Whisper-Tiny` returned `session.created` plus `session.updated`.
- [x] Plain HTTP origin / `ws://` origin behavior: using `ai4` as a plain HTTP reverse proxy on `http://192.168.79.20:18080` with `external_url=http://192.168.79.20:18080`, REST resolved through that origin and both `ws://192.168.79.20:18080/logs/stream` and `ws://192.168.79.20:18080/realtime?model=Whisper-Tiny` succeeded.
- [x] HTTPS proxy / `wss://` behavior: previously verified both on-host on `ai2` at `https://192.168.79.18:8443` and cross-host on `ai4` at `https://192.168.79.20:8443`; proxied `/logs/stream` and `/realtime` both succeeded over `wss://`.
- [x] Path prefix: `ai4` served `https://192.168.79.20:18443/lemonade -> ai2`; `/lemonade/app` loaded, `/lemonade/api/v1/system-info` reported `external_url=https://192.168.79.20:18443/lemonade`, and both `wss://192.168.79.20:18443/lemonade/logs/stream` and `.../realtime?model=Whisper-Tiny` succeeded.
- [x] `external_url` with `websocket_port=auto`: captured the startup warning on a disposable `ai2` server: `external_url is set, but websocket_port is "auto"...`.
- [x] Runtime `lemonade config set external_url=...`: captured both `Updated external_url to ...` and the warning re-check when `websocket_port` remained `auto`.
- [ ] Desktop `--base-url` regression check: not run here. All available systems for this round were headless, so I did not have a real desktop UI session to exercise the `window.api.getServerBaseUrl()` path.

Additional testing not described in the PR body:
- Built and installed the Debian package on `ai2` (`Ubuntu 24.04`): `lemonade-server_10.2.0-29-ga1a70290~24.04_amd64.deb`.
- Verified the feature both on-host (`ai2` proxy + server) and cross-host (`ai4` proxy -> `ai2` server).
- Addressed the `/system-info` review point: the old HTML injection approach is gone; `external_url` is now surfaced from `/api/v1/system-info` and consumed by the frontend there.

The disposable listeners have been cleaned up again, and the installed `ai2` service was restored to `external_url=""`, `host="0.0.0.0"`, `websocket_port="auto"` after testing.

If maintainers would prefer a same-commit upstream mirror branch for follow-up pushes, I can mirror this exact tested commit after review.

@superm1 @Geramy re-review appreciated.

src/cpp/server/server.cpp
ianbmacdonald: Addressed in `3faedcae`: the injected `window.__LEMONADE_EXTERNAL_URL__` path is gone. `external_url` is now exposed from `/api/v1/system-info`, and the web renderer reads it there during initialization instead of via HTML injection. I also posted the refreshed headless test results in the main PR thread just above.

ianbmacdonald:
Closing this in favor of the handling that has since landed upstream — the original problem this PR set out to solve is now fixed on `main` without needing the `external_url` config surface.

## Why

This PR was opened against #472 (*"frontend does not follow the URL that loaded the page"* behind a reverse proxy). Since then, `main` has independently resolved that whole problem space at the **root-origin** level:

- **#472** (CLOSED) — the web app now derives its base URL from `window.location.origin` in web-app mode, so it follows the page's scheme/host instead of hardcoding `http://host:port`.
- **#1837** (`buildWebSocketUrl`) — centralized WebSocket URL construction from the server base URL, with `wss://` on HTTPS (#1583/#1666) and IPv6-literal correctness.
- WebSocket upgrades for `/realtime` and `/logs/stream` now ride the **main HTTP port** by default (the dedicated `websocket_port` remains only as a legacy fallback), so a reverse proxy fronts HTTP and WS on the same origin with no extra port forwarding.
- Sibling proxy bugs **#631** (web UI through a proxy) and **#2190** (log stream through a proxy) are both CLOSED by that same work.

That combination covers the deployment everyone actually reported — `https://example.com` → lemond, SSL/vhost at the proxy — out of the box. The `external_url` key this PR adds is no longer necessary for it.

## Topology breakdown

| Deployment | Status on current `main` | Needs `external_url`? |
|---|---|---|
| **Root-origin proxy** (`https://host/` → lemond) — the #472/#631/#2190 case | ✅ works (origin derivation + main-port WS + `wss://`) | No |
| **Sub-path proxy** (`https://host/lemonade/` → lemond) | ⚠️ not handled — derivation uses `window.location.origin` only, dropping the path | No — this needs **pathname-aware derivation**, not a config key |
| **Split-origin** (UI on origin A, API/WS on origin B, browser → B directly) | ❌ not handled | Only case `external_url` uniquely addresses — but delivering it via `/system-info` is self-defeating (the browser must already know origin B to fetch `/system-info` from it) |

So the only gap left is **sub-path mounting**, and the clean fix there is to extend upstream's derivation to `origin + window.location.pathname` so `buildWebSocketUrl` carries the base path — a small change with **no new config key**, complementing #1837. There's currently no open request for sub-path support in the tracker, so rather than carry this 250+-commit-stale branch forward, I'm closing it and will revisit a focused pathname-derivation fix if/when sub-path demand actually appears.

Thanks @Geramy and @superm1 for the reviews — the `/system-info` direction (vs. the original HTML-injection approach) was the right call and informed this analysis.

PR #1596, opened 2026-04-10
fix: skip ROCm download for gfx1033/1035/1036

martinlroth:
gfx1036 is the integrated GPU in AMD Ryzen 9000 desktop processors (e.g. 9950X). When this iGPU was detected on Linux via KFD, get_rocm_arch() returned "gfx1036" and the installer attempted to download a non-existent ROCm artifact (e.g. llama-...-rocm-gfx1036-x64.zip).

ROCM_ARCH_MAPPING now maps gfx1033, gfx1035, and gfx1036 to an empty string. identify_rocm_arch_from_name() returns that empty string, and get_rocm_arch() already treats an empty result as "no supported GPU", so it falls through to any discrete GPU present, or returns "" to signal that ROCm is unsupported on this system.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -71,7 +71,8 @@ const std::vector<std::string> NVIDIA_DISCRETE_GPU_KEYWORDS = {
     "a100", "a40", "a30", "a10", "a6000", "a5000", "a4000", "a2000"
 };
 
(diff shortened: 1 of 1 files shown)


superm1 (approved):

jaminellis:
I was just about to raise this one, so great to see it fixed!

jeremyfowers:
Thank you!

PR #1604, opened 2026-04-10
fix: make /load declarative/idempotent to prevent TOCTOU race

### LINKED ISSUE #1603: fix: make /load idempotent to prevent TOCTOU race causing unnecessary model reloads
## Problem

When a client sends a `/load` request for a model that's already being loaded (or was just loaded) by an inference-triggered auto-load, the current code in `handle_load` does a non-atomic check-unload-reload sequence outside the Router's `load_mutex_`. This causes unnecessary eviction and reload of large models (~90s wasted for a 68GB model).  I tripped on this tinkering with OpenClaw, when my agent responsible for keeping backend uptodate pulls the latest bleeding-edge package, models a are cleared;  A sub-agent triggers an auto-load on next inference, and then when a typical timeout passes (something not expected with a Cloud API) at 90s or so, on a big model like Qwen3.5-122B, the /load is initiated, which just burns another 90s unecessarily by queuing an eviction and reload, even though model and recipe options match. 

**Root cause**: TOCTOU race in `Server::handle_load` (`server.cpp:2792-2815`). `is_model_loaded()`, `unload_model()`, and `load_model()` each acquire/release the mutex independently, allowing another thread to change state between calls.

**Observed scenario**: After a system restart, an inference request triggers `auto_load_model_if_needed()` for a 68GB model. While that load is in progress (or just completed), an explicit `/load` call for the same model arrives, finds it loaded, evicts it, and reloads from scratch — wasting ~90 seconds.

## Proposed Fix

Move the "if loaded with same options, skip; if loaded with different options, evict and reload" decision into `Router::load_model()` under the existing `load_mutex_`, and remove the pre-emptive unload from `Server::handle_load()`.

### Semantic change

This redefines `/load` from "always restart if already loaded" to declarative: **"ensure model is loaded with these options."** If the model is already loaded with matching options, `/load` becomes a no-op. This is a real behavior change — retries and timeout recovery become safe, but there is no longer a way to unconditionally bounce a model. If a true restart is needed in the future, that should be a separate `/reload` endpoint or request flag.

### Implementation

**1. `router.h`** — Add `allow_reload_on_option_change` parameter:

```cpp
// allow_reload_on_option_change: intended for explicit /load callers only.
// Auto-load callers (inference-triggered) should leave this false so they
// don't overturn options set by a prior explicit /load.
void load_model(const std::string& model_name, const ModelInfo& model_info,
                RecipeOptions options, bool do_not_upgrade,
                bool allow_reload_on_option_change = false);
```

**2. `router.cpp`** — Replace the existing "already loaded" no-op with options-aware logic inside `load_mutex_`:

```cpp
WrappedServer* existing = find_server_by_model_name(model_name);
if (existing) {
    if (allow_reload_on_option_change &&
        existing->get_recipe_options().to_json() != effective_options.to_json()) {
        LOG(INFO, "Router") << "Options changed, reloading model: " << model_name << std::endl;
        evict_server(existing);
        // Fall through to create and load with new options
    } else {
        LOG(INFO, "Router") << "Model already loaded, updating access time" << std::endl;
        existing->update_access_time();
        is_loading_ = false;
        load_cv_.notify_all();
        return;
    }
}
```

The comparison uses `to_json()` on both sides. Both are the merged declarative option set after the inherit chain (not every runtime-derived value), so no sentinel normalization issues — `inherit()` filters out `-1`/`""`/`"auto"` values.

**3. `server.cpp`** — Simplify `handle_load()` to delegate entirely to Router:

```cpp
// Remove the is_model_loaded() / unload_model() block (lines 2792-2797)
// Replace with:
LOG(INFO, "Server") << "Ensuring model loaded: " << model_name;
router_->load_model(model_name, info, options, true,
                    /*allow_reload_on_option_change=*/true);
```

## Scenario Coverage

| Scenario | Outcome |
| --- | --- |
| auto-load + `/load` with same options (the observed bug) | Second load waits, then no-ops |
| auto-load + `/load` with different options | Waits for first load, then evicts and reloads atomically |
| `/load` + `/load` (client retry after timeout) | Second waits, then no-ops if same options |
| auto-load + auto-load | Unchanged (already a no-op via `allow_reload_on_option_change=false`) |
| Explicit `/load` to change options on a loaded model | Evicts and reloads (correct) |

## Known Gaps (out of scope)

- **Download-path race**: Two requests racing in the pre-Router download path can both trigger downloads. Separate issue. This fix covers the "already downloaded, slow to initialize" case.
- **Client-side feedback**: The HTTP request still blocks until the mutex frees up. No "load in progress" / 202 response.
- **`/unload` during a load**: Pre-existing behavior, unchanged.
- **True unconditional restart**: If needed later, add a `/reload` endpoint or request flag.

## Files to Modify

- `src/cpp/include/lemon/router.h`
- `src/cpp/server/router.cpp`
- `src/cpp/server/server.cpp`

ianbmacdonald:
## Summary

- Fixes a TOCTOU race in `handle_load` where `is_model_loaded()`, `unload_model()`, and `load_model()` each acquired/released `load_mutex_` independently, allowing concurrent requests to cause unnecessary eviction and reload of large models (~90s wasted for 68GB models)
- Moves the "already loaded?" decision into `Router::load_model()` under the existing `load_mutex_`, with an `allow_reload_on_option_change` parameter for explicit `/load` callers
- Redefines `/load` as declarative ("ensure loaded with these options") — same-options is a no-op, different-options atomically evicts and reloads
- Adds three integration tests: concurrent race regression, sequential idempotency, and option-change reload

Closes #1603

## Test plan

- [x] Build passes (`cmake --build --preset default`)
- [x] Concurrent auto-load + `/load` for the same model: second arrival no-ops, no eviction in logs
- [x] Sequential `/load` with same options: `backend_url` unchanged (same subprocess, no restart)
- [x] Sequential `/load` with different `ctx_size`: evicts and reloads with new options
- [x] Full `server_endpoints.py` suite: 35/35 tests pass on Debian 13 x86_64

### DIFF
--- a/test/server_endpoints.py
+++ b/test/server_endpoints.py
@@ -21,8 +21,8 @@ Usage:
 """
 
 import platform
+import time
 import uuid
-from concurrent.futures import ThreadPoolExecutor
 import requests
 from openai import NotFoundError
 
@@ -474,8 +474,13 @@ class EndpointTests(ServerTestBase):
 
     def test_012a_load_idempotent_same_options(self):
         """Test that /load is idempotent: loading an already-loaded model with
-        the same options is a no-op (no eviction or reload)."""
-        # Ensure model is loaded
+        the same options is a no-op (no eviction or reload).
+
+        Uses wall-clock time as the proof signal: a no-op /load returns in
+        milliseconds, while even a tiny model reload takes several seconds.
+        (backend_url is not a stable identity — WrappedServer::choose_port
+        can pick the same port after a restart.)"""
+        # Ensure model is loaded (this may take seconds for the initial load)
         response = requests.post(
             f"{self.base_url}/load",
             json={"model_name": ENDPOINT_TEST_MODEL},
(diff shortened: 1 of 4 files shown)


test/server_endpoints.py
```diff
@@ -472,6 +472,159 @@ def test_012_load_uses_saved_options(self):
                     print(f"[OK] Load used saved ctx_size={custom_ctx_size}")
                 break
 
+    def test_012a_load_idempotent_same_options(self):
+        """Test that /load is idempotent: loading an already-loaded model with
+        the same options is a no-op (no eviction or reload).
+
+        Uses wall-clock time as the proof signal: a no-op /load returns in
+        milliseconds, while even a tiny model reload takes several seconds.
```
jeremyfowers, line 480: I want to avoid tests that use timers whenever possible as they can turn out to be flakey.

How about monitoring the logs to see if a new process was started? Or tracking the process ID of the first llama-server to see if it got terminated?

jeremyfowers: Opened https://github.com/lemonade-sdk/lemonade/issues/1663 to address this in a follow up!

jeremyfowers (approved):
Thanks @ianbmacdonald for saving everyone some time and headache!

PR #1607, opened 2026-04-10
Disable realtime VAD for manual commit mode

### LINKED ISSUE #1606: Realtime STT Is About 2x Slower Than The Same FLM Model Via HTTP
### Platform

Linux/Arch

### Lemonade Version

10.1.1

### GPU / APU Model

_No response_

### Component

Other / Not sure

### Bug Description

When Lemonade realtime STT is used in manual-commit mode, with the client explicitly sending `turn_detection: null`, final transcription latency is about twice as high as the same FLM Whisper model reached via HTTP.

I narrowed this down to Lemonade's `/realtime` path rather than the FLM model itself:

- Direct FLM HTTP transcription on the same WAV: about `2.167s`
- Lemonade HTTP transcription on the same WAV: about `2.181s`
- Lemonade realtime transcription on the same WAV: about `4.491s`

The realtime session path does not fully honor manual-commit mode: even when the client sends `turn_detection: null`, the server still runs turn detection and interim transcription work. 

### Steps to Reproduce

- Use a fixed WAV file of the spoke text `What is the capital of France?`.
- Measure direct FLM HTTP transcription on that WAV.
- Stop FLM.
- Start Lemonade and measure Lemonade HTTP transcription on the same WAV.
- Without restarting Lemonade, measure Lemonade realtime transcription on the same WAV using manual-commit mode with `turn_detection: null`.
- Compare the timings. In my run: direct FLM `2.167s`, Lemonade HTTP `2.181s`, Lemonade realtime `4.491s`.

### Pseudocode

```text
wav = "realtime_stt_benchmark_france.wav"
expected_transcript = "What is the capital of France?"

# 1. Direct FLM baseline
start flm(asr=True)
t_flm = post_http_transcription(file=wav)
stop flm()

# 2. Lemonade HTTP baseline
start lemond(port=13305)
t_lemonade_http = post_http_transcription(
    file=wav,
    model="whisper-v3-turbo-FLM",
)

# 3. Lemonade realtime, manual-commit mode
ws = connect_realtime(model="whisper-v3-turbo-FLM")
send({
    "type": "session.update",
    "session": {
        "model": "whisper-v3-turbo-FLM",
        "turn_detection": null,
    },
})
for chunk in wav_as_pcm16_chunks(wav):
    send({"type": "input_audio_buffer.append", "audio": base64(chunk)})
send({"type": "input_audio_buffer.commit"})
t_lemonade_realtime = time_until(
    "conversation.item.input_audio_transcription.completed"
)

assert transcript == expected_transcript
assert t_lemonade_http ~= t_flm
assert t_lemonade_realtime >> t_flm
```

This report is specifically about manual-commit mode with `turn_detection: null`. It does not claim that the same latency gap necessarily exists when clients intentionally enable server-side `turn_detection`.

Expected:

- Lemonade realtime STT in manual-commit mode should be close to Lemonade HTTP and direct FLM on the same model and same audio.
- A small wrapper overhead is expected, but not about `2.3s` extra on a `2.4s` WAV.

Actual:

- Direct FLM: about `2.167s`
- Lemonade HTTP: about `2.181s`
- Lemonade realtime: about `4.491s`

This makes Lemonade realtime in manual-commit mode about twice as slow as the same backend reached through HTTP.

### Log Output

Relevant debug log lines from the affected 10.1.1 realtime path:

```shell
[Debug] (RealtimeSession) VAD: SpeechStart detected
[Debug] (RealtimeSession) Firing interim transcription at 1000ms
[Debug] (RealtimeSession) Calling Whisper interim transcription (32044 bytes)...
[Debug] (RealtimeSession) Calling Whisper final transcription (76844 bytes)...
[Info] (Process) [🟢 ]  NPU Locked!
[Info] (Process) [🕒 ]  NPU busy, request queued (1/10): POST /v1/audio/transcriptions
[Debug] (RealtimeSession) Whisper interim response: {"model":"whisper-v3:turbo","text":" What is the cap?"}
[Debug] (RealtimeSession) Sending interim transcript to client: " What is the cap?"
[Debug] (RealtimeSession) Whisper final response: {"model":"whisper-v3:turbo","text":" What is the capital of France?"}
[Debug] (RealtimeSession) Sending final transcript to client: " What is the capital of France?"
```

This is the key evidence for the bug: in a session that was intended to be manual-commit mode, Lemonade still triggers a `1000ms` interim transcription first, and the final transcription is then queued behind it on the single NPU.

Lemonade HTTP and Lemonade realtime both resolve to the same FLM checkpoint:

checkpoint = whisper-v3:turbo
recipe = flm

### Additional Context

_No response_

michaelburch:
Fixes #1606.

Honor `turn_detection: null` in realtime STT sessions by disabling server-side VAD and interim transcription work in manual-commit mode. This keeps audio buffered until `input_audio_buffer.commit`, avoids the extra interim request that was inflating latency, and adds a regression test plus a small spec update for the manual-commit behavior.


ramkrishna2910 (changes requested):
Overall the approach is clean and correct — gating `process_vad()` behind an atomic flag is a minimal, low-risk way to support manual-commit mode. A few items need attention before merging:

### Must fix

1. **Duplicated turn_detection parsing logic** — The `td.is_null()` branching in `create_session()` and `update_session()` is copy-pasted verbatim. Extract a private helper (e.g. `apply_turn_detection_config(session, td)`) so the two copies can't drift apart.

2. **Test replaces VAD coverage instead of adding new coverage** — `_test_007` was the only test exercising the default VAD-enabled WebSocket path. Converting it to manual-commit mode removes that coverage entirely. Please keep `_test_007` as-is and add a new test (e.g. `_test_008_realtime_manual_commit`) for the `turn_detection: null` path.

### Should fix

3. **`session.updated` response doesn't echo `turn_detection` state** — The OpenAI Realtime API echoes the full session config in `session.updated`. Currently the response only includes `id` and `model`. Consider adding `turn_detection` so clients can confirm the mode switch took effect.

4. **Test assertion only checks events after commit** — If a VAD/interim event were emitted during the `append` loop (before `commit()`), the current assertion wouldn't catch it. Consider draining events after sending chunks (with a short timeout) to verify no VAD events leak during the buffering phase.

---
_Generated by [Claude Code](https://claude.ai/code)_

michaelburch:
addressed findings. I did extract a shared helper for tests 7 and 8 and updated test 7 to use it. no functional changes

ramkrishna2910 (approved):
Thanks for the contribution!

PR #1611, opened 2026-04-11
Migrate desktop app from Electron to Tauri

jeremyfowers:
Replaces the Electron 39 desktop app with a Tauri v2 host. The React renderer is unchanged.

### Release artifact sizes

Three release artifacts contain a desktop app and shrink; everything else (server debs, rpms, embeddables, minimal MSI) is unchanged. Numbers from runs [24272200505](https://github.com/lemonade-sdk/lemonade/actions/runs/24272200505) (main) and [24289704993](https://github.com/lemonade-sdk/lemonade/actions/runs/24289704993) (this PR).

| File | main | this PR | Δ |
|---|---:|---:|---:|
| `lemonade.msi` (Windows full) | 112.6 MB | 8.1 MB | **−93%** |
| `Lemonade-Darwin.pkg` (macOS signed) | 106.2 MB | 9.8 MB | **−91%** |
| `lemonade-app-x86_64.AppImage` | 106.1 MB | 77.2 MB | **−27%** |
| `lemonade-server-minimal.msi` (no desktop app, reference) | 5.0 MB | 5.0 MB | — |

Tauri uses the OS-native webview (WebView2 / WKWebView / webkit2gtk) instead of bundling Chromium. AppImage shrinks the least because it still has to bundle webkit2gtk to stay portable across Linux distros.

### Dependency footprint

| Tree | main | this PR |
|---|---:|---:|
| npm packages in `src/app/package-lock.json` | 764 | 431 |
| Rust crates in `src/app/src-tauri/Cargo.lock` | 0 | 591 |

The npm drop is the entire `electron` / `electron-builder` family. The Rust crates are build-time only and never enter the supply chain we ship to users.

### Implementation

The old 1,100-line `main.js` becomes a Rust crate at `src/app/src-tauri/` with the same responsibilities: window controls, settings file at `~/.cache/lemonade/app_settings.json`, UDP beacon discovery, `lemonade://` deep links, and per-platform webview hooks (mic permission, external-link routing). The React renderer reaches it through a TypeScript shim (`tauriShim.ts`) that installs `window.api` against Tauri commands, so the same bundle still runs unchanged in the browser-served `/app` page. `lemond` is untouched.

First cold cargo build compiles ~80 crates with LTO; CI caches via `Swatinem/rust-cache@v2` and renderer-only iteration runs `cd src/app && npm run dev` to skip cargo. The Tauri target is opt-in (`LEMONADE_SETUP_TAURI=1`) so distro CI jobs that don't need it aren't slowed down.


.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -567,6 +575,23 @@ jobs:
...
+            libwebkit2gtk-4.1-dev \
+            libsoup-3.0-dev \
+            libjavascriptcoregtk-4.1-dev \
+            librsvg2-dev \
+            libayatana-appindicator3-dev \
+            libssl-dev \
+            build-essential \
+            curl \
+            wget \
+            file
```
superm1: why doesn't setup.sh handle this?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -580,6 +605,14 @@ jobs:
...
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache Rust build artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src/app/src-tauri
+
       - name: Build AppImage
         shell: bash
         run: |
```
superm1, line 568: probably worth asking - with tauri potentially being in the Ubuntu archives, should lemonade-desktop provide the Tauri app instead of just the web app?

src/app/package.json
```diff
@@ -1,35 +1,33 @@
 {
-  "name": "lemonade",
-  "version": "1.0.0",
-  "description": "Lemonade Electron App",
-  "main": "main.js",
+  "name": "lemonade-app",
+  "version": "10.2.0",
```
superm1: Doesn't need to be this PR; but it would be good to come up with a way to generate this on the fly so it doesn't go out of sync in the future.

src/cpp/server/server.cpp
```diff
@@ -615,108 +615,10 @@ window.api = {
...
-            }
-        } catch (e) {
-            console.warn('Failed to fetch system info:', e);
-        }
-        return { system: 'Unknown', os: 'Unknown', cpu: 'Unknown', gpus: [], gtt_gb: undefined, vram_gb: undefined };
-    }
+    restartApp: () => window.location.reload()
+    // Server version, /system-stats, and /system-info are not exposed via
+    // window.api in either Tauri or web mode. The shared renderer fetches
+    // those endpoints directly via serverConfig.fetch (utils/serverConfig.ts).
```
superm1: I feel like this comment makes sense for explaining the PR, but is going to be confusing in the context after the PR has been merged.  I would suggest dropping it entirely.

setup.sh
```diff
@@ -381,6 +381,96 @@ if command_exists node && ! command_exists npm; then
...
+# install batch) for two reasons:
+#   1. The Tauri desktop app target is OPTIONAL — the C++ server flow
+#      (`cmake --build --preset default`) doesn't need any of this. CI
+#      workflows that only build the C++ server (e.g. linux_distro_builds)
+#      should not be slowed down or broken by Tauri's deps.
+#   2. The mandatory batch must succeed atomically. If a single Tauri
+#      package name is wrong on a given distro, mixing it into the same
+#      install call would abort the whole setup. Keeping them separate
+#      means a Tauri-install failure is contained — the C++ build still
+#      proceeds normally.
```
superm1: I think the deps should just be installed for everyone.  They're harmless if you're not building/using Tauri.  It will just prevent us with a bifurcated build environment based on how someone answered the prompt.

setup.sh
```diff
@@ -505,6 +595,107 @@ fi
...
+        # Install Rust via rustup. Curl is guaranteed to be available by
+        # this point (the mandatory C++ deps batch above already installed
+        # it on distros where it wasn't pre-installed).
+        if [ "$rust_needs_install" = true ]; then
+            if ! command_exists curl; then
+                print_warning "curl is not available; skipping Rust install"
+                print_info "Install curl and re-run setup.sh if you need the Tauri build."
+            else
+                print_info "Downloading and running rustup-init..."
+                if curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable --no-modify-path; then
```
superm1: do the distro rust packages not work?  I would think they're safer if we can use them.

setup.sh
```diff
@@ -530,9 +721,24 @@ echo "=========================================="
...
+# own process. New shells will pick it up automatically because rustup adds
+# itself to ~/.profile / ~/.bashrc / ~/.zshenv.
+if [ "${rust_was_just_installed:-false}" = true ]; then
+    print_warning "Rust was just installed."
+    print_info "To use cargo in your CURRENT shell, run:"
+    echo "  . \"\$HOME/.cargo/env\""
+    print_info "New shells will pick it up automatically."
+    echo ""
+fi
+
```
superm1: if rust from the distro packages is OK this whole block can go

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: That would be cool!

setup.sh
jeremyfowers: This is not something I'm familiar with. Should this just try to install inbox platform rust and use the curl as a fallback?

superm1: If it works, yes use inbox rust. Give it a try. Sometimes software needs newer crates and it's a flop.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: 8715fc97

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: I will leave it up to a separate PR though, this PR is already too big.

setup.sh
jeremyfowers: 8715fc97

setup.sh
jeremyfowers: 8715fc97

setup.sh
jeremyfowers: 8715fc97

setup.sh
jeremyfowers: Looks like the distro builds are all passing

src/app/package.json
jeremyfowers: f8eee10

.github/workflows/cpp_server_build_test_release.yml
superm1: sounds good

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -134,6 +134,14 @@ jobs:
         with:
           node-version: '20'
 
+      - name: Setup Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache Rust build artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src/app/src-tauri
```
superm1: to be cognizant of a supply chain hack - we should probably be using hashes for untrusted 3rd party actions so that "@v2" can't get changed under the covers.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -580,6 +588,14 @@ jobs:
...
+
```
superm1: so I guess this means distro rust didn't work after all that you have this?

jeremyfowers: heh we're also getting node from an action a few lines above... I'm gonna try tearing all of this out.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Got rid of it. The deb builds in 3.5 mins without it.

jeremyfowers: Got rid of it. The deb builds in 3.5 mins without it.

.github/workflows/cpp_server_build_test_release.yml
jeremyfowers: Build passes without this.

sofiageo (commented):
I just tested the current PR with my usual workflow. What I noticed is:

1) I had to use `-D BUILD_TAURI_APP=OFF` to build lemonade-server otherwise I got the error: `CMake Error: File /home/gso/aur/lemonade-server-git/src/lemonade-server/cmake/run_tauri_with_version.cmake.in does not exist.`

2) When I try to build the desktop app, it tries to also build the AppImage, then fails with: `Error failed to bundle project failed to run linuxdeploy`

I used the following commands to build in Arch Linux, maybe they are wrong:

```
  npm ci
  cargo install tauri-cli
  npm run build:renderer:prod
  cargo tauri build
```

3) It also managed to build the desktop app (7.9MB binary) so I can run it fine on my PC.

superm1 (approved):

sofiageo (commented):
As Jeremy said I must have been unlucky when I tested the branch. It works for me now, but it takes 2 minutes and 40 seconds to build the desktop app for the stable rust cargo.

I can try to test the nightly rust toolchain to see if it improves compilation times but I don't think it's realistic to expect all AUR users to have the nightly rust. With that in mind, let's merge it and we figure out how to make the AUR builds faster later.

ramkrishna2910:
Test drove the app on all common use cases across modalities and the experience was identical to the electron app.
The build took ~3mins for me.

ramkrishna2910 (approved):

PR #1638, opened 2026-04-15
Integrate Speculative Decoding into Lemonade

### LINKED ISSUE #1419: Feature Request: Integrate Speculative Decoding for Token Generation Speedup (Default & Configurable)
### Overview
We should integrate and expose `llama.cpp`'s Speculative Decoding features natively within Lemonade. 

At a high level, **Speculative Decoding** accelerates generation by guessing (drafting) multiple upcoming tokens and then having the main model verify/accept them all at once in a single batch. 

On UMA architectures like Strix Point/Halo, models run from shared LPDDR5X memory, making memory bandwidth the primary bottleneck. Speculative decoding is particularly well-suited here since the RDNA 3.5 (GPU) compute is largely idle during standard decoding, and the batched verification step can exploit that spare compute to extract multiple accepted tokens from what is effectively a single memory pass over the weights.

There are two main approaches (more details in the [llama.cpp docs](https://github.com/ggerganov/llama.cpp/blob/master/docs/speculative.md)):
1. **Draft Models:** Uses a secondary, much smaller model to guess the next tokens.
2. **N-gram / Self-Speculative:** Looks for patterns in the current prompt/context window to guess upcoming tokens (e.g., `ngram-simple`, `ngram-mod`, `ngram-map-k4v`). These methods have a very slight overhead and are typically lighter to run compared to draft models, check llama.cpp docs for more details.

### **Early Benchmarks**
* **Model:** GLM 4.7 Flash (No Thinking)
* **Task:** Simple sorting algorithm implementation + code revision
* **Standard Generation:** ~50 tok/s
* **With Speculative Decoding:** **~120 tok/s**
* **Settings Used** `--spec-type ngram-mod --spec-ngram-size-n 24 --draft-min 48 --draft-max 64`

### Implementation Plan

Here is how I think we should integrate this optimization into Lemonade

- [ ] **0. Add Speculative Decoding to Docs**
  If we don't have something similar already, we should create an "Optimizations" section in the docs covering various ways users can configure their setups to get better performance compared to the out of the box defaults. Need to be careful here not to be repetitive and keep it high level (point users back to llama.cpp docs for more details).

- [ ] **1. Enable a Zero-Config Default**
  We should enable a lightweight, speculative method (like `--spec-type ngram-mod` or `ngram-simple` with conservative draft bounds) by default for standard chats. This ensures every Lemonade user gets "free" out-of-the-box speedups without needing to download draft models or tweak settings.

- [ ] **2. Frontend / UI Controls**
  Expose these settings in the Lemonade WebUI model configuration. 
  * Add a dropdown for the user to select the Speculative Mode (`Auto (N-gram)`, `Custom Draft Model`, `Disabled`).
  * Allow attaching a specific GGUF as a draft model.

- [ ] **3. Specialized Recipes for Coding Agents**
  Agentic workflows (like tool-calling, MCP, and code generation) produce highly repetitive text (JSON schemas, <think> blocks, boilerplate code). Since we already maintain model configurations in lemonade-sdk/recipes (like the [claude-code recipes](https://github.com/lemonade-sdk/recipes/tree/main/claude-code)), we simply need to edit these existing recipes. We can inject highly aggressive n-gram mapping flags (e.g., ngram-map-k4v or ngram-mod with a high --draft-max) directly into the agent/coding recipe defaults.

### Qwen Family Support
Speculative decoding support for Qwen 3 Coder Next and the Qwen 3.5 family will be enabled as part of an upcoming llama.cpp PR [ggml-org/llama.cpp#19493](https://github.com/ggml-org/llama.cpp/pull/19493)).

Would love to get everyone's thoughts on this and how we should properly integrate into Lemonade!

sawansri:
Resolves #1419 

Adds UI controls for speculative decoding

image

Also includes logic in backend to resolve model paths if checkpoint/model name is given.
This enables creating recipes with draft models (paths are not user specific and instead get resolved by lemonade)

### DIFF
--- a/src/app/src/renderer/ModelOptionsModal.tsx
+++ b/src/app/src/renderer/ModelOptionsModal.tsx
@@ -70,6 +70,14 @@ const SPEC_DEFAULTS: SpecState = {
   deviceDraft: '',
 };
 
+const SPEC_NUMERIC_MINIMUMS: Record<SpecNumericFieldKey, number> = {
+  draftMax: 0,
+  draftMin: 0,
+  specNgramSizeN: 1,
+  specNgramSizeM: 1,
+  specNgramMinHits: 1,
+};
+
 const SPEC_MANAGED_FLAGS = [
   '--spec-type',
   '--draft-max',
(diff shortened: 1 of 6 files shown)


BenjaminEllisSo:
It would be great to have this. What needs to happen next?

sawansri:
> It would be great to have this. What needs to happen next?

A frontend rewrite is in progress so this feature will most likely be included as part of the refactor.

jeremyfowers:
Is this still needed now that MTP is commonplace?

PR #1645, opened 2026-04-16
Group user models and add log level control

### LINKED ISSUE #1643: [Feature] Dynamic grouping of model recipes by shared name segments
### Feature Description

A recipe named user.gemma4.mysuperduperquant is already dot-delimited. If multiple recipes share a segment at the same level — e.g. user.gemma4.quantA and user.gemma4.quantB — they'd be visually grouped under a gemma4 header in the UI, with each variant showing its own load/delete/settings controls.

_The grouping should be purely dynamic, derived from the existing naming convention rather than requiring a fixed "org" or "model" taxonomy._

If two names happen to share a segment at the same tree level, they group. No semantic meaning is imposed on any particular level.

Scope (intentionally minimal):

This is a quality-of-life improvement, not a deep restructuring. 

Specifically as noted @jeremyfowers:

user. prefixed models should probably remain expanded by default, since users expect to find them immediately after downloading. 

Example:

Image

Would become:

Image

### Use Case / Motivation

When users accumulate multiple model recipes, the flat list becomes hard to navigate. This proposes automatically grouping recipes that share a common string at the same level in their dot-separated name hierarchy.

### Platform Relevance

All platforms

### Additional Context

All UI's

anditherobot:
What it includes:

  - One-level grouping for dot-delimited model names after existing regex family
    grouping.
  - user.* groups expand by default so newly downloaded user models stay visible.
  - Grouped model rows keep the existing load, delete, and settings controls.
  - Loaded and downloaded model rows get clearer visual highlights.
  - Hugging Face search results now show a Hugging Face source pill.        
  - Logs panel now has a log-level dropdown.
  - Log-level dropdown reads current server config, only allows supported levels,
    and posts changes to /log-level.

  Testing:

  - Ran npm run build:renderer
  - Build passed

Closes #1643

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -184,3 +184,6 @@ CLAUDE.local.md
 # WiX Toolset build artifacts
 *.msi
 *.wixpdb
+
+# Local dev folder
+local.dev/
--- a/docs/dev-getting-started.md
+++ b/docs/dev-getting-started.md
@@ -194,6 +194,15 @@ chmod +x build/app-appimage/lemonade-app-*.AppImage
 - The build uses static linking to minimize DLL dependencies
 - All dependencies are built from source (no external DLL requirements)
 - Security features enabled: Control Flow Guard, ASLR, DEP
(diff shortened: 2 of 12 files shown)


jeremyfowers:
Should this PR be closed since it was split into 2 other PRs?

PR #1650, opened 2026-04-16
Add qwen3.6-35b-a3b

sawansri:
Qwen 3.6 35B A3B uses the same architecture as the Qwen 3.5 family of models (qwen35moe). 

Testing locally, this works well on lemonade's currently packaged versions of llamacpp.

image


jeremyfowers (approved):
Thank you @sawansri !

PR #1656, opened 2026-04-16
Feat: Add model grouping infrastructure and modality icons

anditherobot:
Added tree-like nesting infrastructure and standard modality icons to the core.


jeremyfowers (changes requested):
@anditherobot I think this PR is missing some code, as the new code is not wired up to the existing app.

PR #1657, opened 2026-04-16
UI: Refine Model Manager layout and add interactive tooltips

anditherobot:
Consolidated modality icons/size UI, implemented custom interactive tooltips, and visually separated metadata badges from action buttons.

### DIFF
--- a/src/app/src/renderer/components/Icons.tsx
+++ b/src/app/src/renderer/components/Icons.tsx
@@ -156,6 +156,85 @@ export const EjectIcon: React.FC = () => (
   </svg>
 );
 
+export const DownloadIcon: React.FC<IconProps> = ({ size = 24, strokeWidth = 2 }) => (
+  <svg xmlns="http://www.w3.org/2000/svg" width={size} height={size} viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth={strokeWidth} strokeLinecap="round" strokeLinejoin="round">
+    <path d="M21 15v4a2 2 0 0 1-2 2H5a2 2 0 0 1-2-2v-4" />
+    <polyline points="7 10 12 15 17 10" />
+    <line x1="12" y1="15" x2="12" y2="3" />
+  </svg>
+);
+
+// Modality Icons
+export const Brain: React.FC<IconProps> = ({ size = 16, strokeWidth = 2 }) => (
+  <svg xmlns="http://www.w3.org/2000/svg" width={size} height={size} viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth={strokeWidth} strokeLinecap="round" strokeLinejoin="round">
+    <path d="M9.5 2A2.5 2.5 0 0 1 12 4.5v15a2.5 2.5 0 0 1-4.96.44 2.5 2.5 0 0 1-2.96-3.08 3 3 0 0 1-.34-5.58 2.5 2.5 0 0 1 1.32-4.24 2.5 2.5 0 0 1 4.44-2.54Z" />
+    <path d="M14.5 2A2.5 2.5 0 0 0 12 4.5v15a2.5 2.5 0 0 0 4.96.44 2.5 2.5 0 0 0 2.96-3.08 3 3 0 0 0 .34-5.58 2.5 2.5 0 0 0-1.32-4.24 2.5 2.5 0 0 0-4.44-2.54Z" />
+  </svg>
+);
+
+export const SquareCode: React.FC<IconProps> = ({ size = 16, strokeWidth = 2 }) => (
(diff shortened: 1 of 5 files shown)


src/app/src/renderer/components/Icons.tsx
```diff
@@ -156,6 +156,85 @@ export const EjectIcon: React.FC = () => (
   </svg>
```
jeremyfowers, line 156: Image

I expected that all user. models would end up in the same grouping. Would you mind clearly documenting the rules in the PR description to set expectations?

src/app/src/renderer/components/Icons.tsx
```diff
@@ -156,6 +156,85 @@ export const EjectIcon: React.FC = () => (
...
 );
```
jeremyfowers, line 158: Image

Appears to be accidentally picking up the . in Qwen3.5 for grouping?

src/app/src/renderer/components/Icons.tsx
```diff
@@ -156,6 +156,85 @@ export const EjectIcon: React.FC = () => (
...
 
+export const DownloadIcon: React.FC<IconProps> = ({ size = 24, strokeWidth = 2 }) => (
```
jeremyfowers, line 159: Image

The static model label icons should not have on-hover movement or color-change effects.

src/app/src/renderer/components/Icons.tsx
```diff
@@ -156,6 +156,85 @@ export const EjectIcon: React.FC = () => (
...
+  <svg xmlns="http://www.w3.org/2000/svg" width={size} height={size} viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth={strokeWidth} strokeLinecap="round" strokeLinejoin="round">
+    <path d="M21 15v4a2 2 0 0 1-2 2H5a2 2 0 0 1-2-2v-4" />
```
jeremyfowers, line 161: I miss the splash of color from the old model label dots. Can you try applying the old colors to the new icons to see how that looks?

src/app/src/renderer/components/Icons.tsx
```diff
@@ -156,6 +156,85 @@ export const EjectIcon: React.FC = () => (
...
+    <polyline points="7 10 12 15 17 10" />
+    <line x1="12" y1="15" x2="12" y2="3" />
+  </svg>
```
jeremyfowers, line 164: Image

It is confusing to me to have icons that are not buttons right next to icons that are buttons. Thoughts about how to resolve this? Top of mind for me is to move the icons all the way to the right, where the colored dots used to be. Or maybe the buttons could go all the way to the right?

jeremyfowers (changes requested):
Please see UI layout/style requests, as well as grouping issues.

src/app/src/renderer/components/Icons.tsx
anditherobot: ## Model Grouping Rules

  - `user.` is a management prefix, not part of the model family.
  - If a `user.` model clearly matches a known family, it joins that family.
    - Example: `user.Qwen3.5-35B-A3B-ThinkingCoder` joins `Qwen3.5`.
  - If a `user.` model does not clearly match a known family, it stays as its own row with `user.` hidden.
    - Example: `user.Bonsai-4B-requantized` displays as `Bonsai-4B-requantized`.
  - Decimal version dots are part of the model name, not grouping separators.
    - Example: `Qwen3.5` stays `Qwen3.5`; `GLM-4.7` stays `GLM-4.7`.
  - We do not use generic dot-splitting for grouping. Only explicit known-family rules create groups.

leoai-81:
image

I don't find the initial feature working anymore...

1. "user" base tag is simply gone?
2. "." delimiter is not used anymore as level delimiter

If "." cause issue with model versions, I would suggest another delimiter like "," or "+".

jeremyfowers:
Sorry about the slow review cycle, I got overwhelmed.

Here are my suggestions for getting these features merged:

1. Split up into two PRs. The icons work is functionally independent of the grouping work, and so should be a separate PR. Also, the icons are looking good, and I think we could merge it pretty quick.
2. For the grouping work, it seems like we stalled out on some bugs, especially because different people have a lot of differences in their custom models. My suggestion for grouping is to make it much simpler.
    - Rather than broad rules that must apply correctly in every scenario, focus on a few narrow rules that would help the most.
    -  IMO what would help the most are the following two hardcoded rules:
        - All `user.` models go into a `User Added` group with the prefix stripped.
        - All `extra.` models go into a `Extra Models Directory` group with that prefix stripped.

jeremyfowers:
I am going to create branches that implement my suggestions above so that we can get these ideas merged. I'll make sure to give you author credit @anditherobot.

jeremyfowers:
The 2 PRs mentioned above have merged. Thanks very much @anditherobot for working through the UI/UX considerations and building the proof of concept here. The Model Manager is a lot better now thanks to you! Closing the PR.

PR #1665, opened 2026-04-17
Npm security

superm1:
Avoid axios different-version and lifecycle-script surprises
npm ci --ignore-scripts should always be used instead of npm install or npm ci
npm ci requires package-lock.json to exist
Maintainer please verify or recreate package-lock.json before merging
Maybe developer docs should have a note about updating package-lock.json for releases

This is a rebased version of https://github.com/lemonade-sdk/lemonade/pull/1543


Geramy (approved):
Quoting Maybe developer docs should have a note about updating package-lock.json for releases Does not seem to be a huge deal I don't think we are or should be really updating unless its a requirement of a new package or new feature perhaps and/or due to security reasons as well. I don't think its a big deal but I'm not a js/html enthusiast, personally I vote for less tech debt meaning staying on versions unless needed to be changed.

PR #1668, opened 2026-04-18
Add a new configuration option to prevent installation of executables

superm1:
When set users must use a system provided executable instead.

During [review in Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1120707
) there was push back on the maintainability of lemonade for security updates if Debian can't control the rollout of backends.  So I will plan to expand 'system' backend concept into more backends eventually.

But in the short term I'd like a configuration option that will block executable backend installation from Github artifacts.

---

This effectively means that only FLM and llama.cpp system backend "will work" in Debian, but I'll expand it as time goes on.

### DIFF
--- a/docs/server/configuration.md
+++ b/docs/server/configuration.md
@@ -31,6 +31,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
   "models_dir": "auto",
   "ctx_size": 4096,
   "offline": false,
+  "no_fetch_executables": false,
   "disable_model_filtering": false,
   "enable_dgpu_gtt": false,
   "llamacpp": {
@@ -84,6 +85,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
 | `models_dir` | string | "auto" | Directory for cached model files. "auto" follows HF_HUB_CACHE / HF_HOME / platform default |
 | `ctx_size` | int | 4096 | Default context size for LLM models |
 | `offline` | bool | false | Skip model downloads |
+| `no_fetch_executables` | bool | false | Prevent downloading backend executable artifacts; backends must already be installed or use the system backend |
 | `disable_model_filtering` | bool | false | Show all models regardless of hardware capabilities |
 | `enable_dgpu_gtt` | bool | false | Include GTT for hardware-based model filtering |
 
--- a/src/cpp/include/lemon/runtime_config.h
+++ b/src/cpp/include/lemon/runtime_config.h
@@ -33,6 +33,7 @@ public:
 
     // Feature flags
     bool offline() const;
+    bool no_fetch_executables() const;
     bool disable_model_filtering() const;
     bool enable_dgpu_gtt() const;
 
--- a/src/cpp/resources/defaults.json
+++ b/src/cpp/resources/defaults.json
@@ -11,6 +11,7 @@
   "models_dir": "auto",
   "ctx_size": 4096,
   "offline": false,
+  "no_fetch_executables": false,
   "disable_model_filtering": false,
   "enable_dgpu_gtt": false,
   "llamacpp": {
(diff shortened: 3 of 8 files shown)


kenvandine:
What does this look like in the UI? Do those backends still show but with the message that they can't be downloaded?

superm1:
CLI UI:
```
╰─❮ ./build/lemonade recipes
Recipe              Backend     Status          Message/Version                               Action
----------------------------------------------------------------------------------------------------------------------------------------------------
flm                 npu         update_required Backend update is required before use.         -
kokoro              cpu         installable     Backend is not installed.                      -
llamacpp            cpu         installable     Backend is not installed.                      -
                    metal       unsupported     Requires macOS                                 -
                    rocm        installed       b1238                                          -
                    system      unsupported     llama-server not found in PATH                 -
                    vulkan      installed       b8766                                          -
ryzenai-llm         npu         unsupported     Requires Windows                               -
sd-cpp              cpu         installable     Backend is not installed.                      -
                    rocm        installable     Backend is not installed.                      -
whispercpp          cpu         installable     Backend is not installed.                      -
                    npu         unsupported     Requires Windows                               -
                    vulkan      installable     Backend is not installed.                      -
----------------------------------------------------------------------------------------------------------------------------------------------------
```

In the web-app:

image

image

superm1:
image

kenvandine:
Perhaps we should display a hint explaining how to enable it, what do you think?

superm1:
> Perhaps we should display a hint explaining how to enable it, what do you think?

I have mixed feelings.  I don't want to go against the spirit of what the DFSG team wants.  _Eventually_ when more of these backends are available natively in Debian I want to have PackageKit install them for a user.

kenvandine:
Yeah, that's why I asked. I don't like not helping the user do what that want to do, but that does go against the spirit. I guess I'm fine either way, just want to make sure we've had that discussion.

superm1:
In some testing, I think I prefer to mark them unsupported for now.  I think it provides the "better experience" then.

superm1:
image

So now it will show up like this and only expose models that are really available when set.

---

TBH - I think this is how most people using lemond embedded should work too.  They can bundle the software they want with the embedded lemond.

kenvandine:
Great, I prefer that

jeremyfowers:
I prefer this as well. It would have been quite a tease to label things as installable when they are not.

superm1:
OK, so you guys like this now?

docs/server/configuration.md
```diff
@@ -84,6 +85,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
 | `models_dir` | string | "auto" | Directory for cached model files. "auto" follows HF_HUB_CACHE / HF_HOME / platform default |
 | `ctx_size` | int | 4096 | Default context size for LLM models |
 | `offline` | bool | false | Skip model downloads |
+| `no_fetch_executables` | bool | false | Prevent downloading backend executable artifacts; backends must already be installed or use the system backend |
```
jeremyfowers, line 88: What's the behavior if a backend is already installed, but an update is available (installed version is less than backend_versions.json)? From reading this description, it sounds like the update shouldnt pull. Does that deadlock lemonade?

jeremyfowers (approved):
LGTM but please address or disprove the corner case concern before merging.

docs/server/configuration.md
superm1: My expectation is that it shows a message saying that it can't pull.  So yeah you'll effectively be blocked.

This should be a major corner case scenario though - you would have had to go out of your way to download the binary artifacts yourself.

It would be more common for embedded lemond use (IE they downloaded in advance and then tried to upgrade embedded lemond).  Honestly - it's more predictable behavior IMO for that case.

PR #1673, opened 2026-04-19
Theme updates

sofiageo:
This PR extracts more colors into CSS variables. Fixes some issues with the current light theme. The important part is to have the necessary CSS variables in order to be able to override them in different themes. 

When it is merged, I will try to create a different light theme (something similar to vscode github light theme) and convert the current light theme to "lemonade theme".

### DIFF
--- a/src/app/styles.css
+++ b/src/app/styles.css
@@ -1,17 +1,43 @@
 :root {
+    /* Dark theme */
+    --layout-bg: #0a0a0a;
+    --title-surface: #000000;
     --header-bg-color: #000000;
     --bg-primary: #0a0a0a;
     --bg-secondary: #1a1a1a;
     --bg-tertiary: #111111;
     --bg-quaternary: #2a2a2a;
     --bg-hover: #f0f0f0;
+
+    --success-color: #4CAF50;
+    --success-border: #4CAF50;
+    --warning-color: #ffb74d;
+    --info-color: #2196F3;
+    --action-color: #FF9800;
+    --error-background-color: rgba(244, 67, 54, 0.1);
+    --error-border: rgba(244, 67, 54, 0.3);
+    --error-color: #f44336;
+    --error-hover:#dc2626;
+
+    --logs-error: #ff4444;
+    --logs-surface: rgba(255, 68, 68, 0.1);
+
+    --left-panel-surface: linear-gradient(180deg, var(--panel-bg-1) 0%, var(--panel-bg-2) 100%);
+
+    --bg-alpha-01: rgba(255, 255, 255, 0.01);
     --bg-alpha-02: rgba(255, 255, 255, 0.02);
     --bg-alpha-03: rgba(255, 255, 255, 0.03);
+    --bg-alpha-04: rgba(255, 255, 255, 0.04);
     --bg-alpha-05: rgba(255, 255, 255, 0.05);
     --bg-alpha-06: rgba(255, 255, 255, 0.06);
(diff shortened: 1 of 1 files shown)


src/app/styles.css
```diff
@@ -1,17 +1,43 @@
 :root {
```
jeremyfowers, line 1: Image

Can you apply more contrast between the downloaded and not-downloaded markers?

src/app/styles.css
```diff
@@ -1,17 +1,43 @@
...
+    /* Dark theme */
+    --layout-bg: #0a0a0a;
+    --title-surface: #000000;
     --header-bg-color: #000000;
     --bg-primary: #0a0a0a;
```
jeremyfowers, line 6: Image

Should the "Lemonade" title be black in this theme? I don't think it should be so much lighter than the other text.

src/app/styles.css
```diff
@@ -1,17 +1,43 @@
 :root {
+    /* Dark theme */
+    --layout-bg: #0a0a0a;
+    --title-surface: #000000;
     --header-bg-color: #000000;
```
jeremyfowers, line 5: Image

The grayed out Visit button makes me think its disabled, can you pick a different highlight color?

jeremyfowers, line 5: In the dark theme the nav bar and "LLM Chat" header are the same color:

Image

Should that also be the case in the light theme?

Image

Otherwise the LLM Chat header stands out more than it probably should.

src/app/styles.css
jeremyfowers, line 6: Image

The highlight color on the send button could use more contrast.

jeremyfowers:
image

I put a current screenshot into Claude Design and here's what it came back with :point_up: 

I like the 1st and 3rd options on the bottom row pretty well. Curious what you think?

I agree with some of the overall suggestions, like that the stats bar shouldn't be the boldest color in the theme.

jeremyfowers (changes requested):
Please see comments - these are my suggestions for getting the existing theme release-ready.

I'm also intrigued by the Claude Design suggestions that give some more depth to the design. LMK what you think there.

src/app/styles.css
jeremyfowers, line 1: Image

Missing border between logs view and sidebar.

jeremyfowers, line 1: Image

need more contrast on log status

jeremyfowers: Image

There's some issues with the top and right borders of the logs view too

sofiageo:
I'll start reading and fix these tomorrow. Most of the work I managed to do was to transfer the colors to CSS variables and fixing the serious color issues, and I only made the PR because it was a lot of manual work and it will be hard to merge with many conflicts. But I will try to finish this tomorrow so it's working as expected.

sofiageo:
> I like the 1st and 3rd options on the bottom row pretty well. Curious what you think?
> 
> I agree with some of the overall suggestions, like that the stats bar shouldn't be the boldest color in the theme.

Yes, 4 and 5 look great. I will see if we can use these colors in the current theme.

jeremyfowers (approved):
>  I only made the PR because it was a lot of manual work and it will be hard to merge with many conflicts

I see! Feel free to merge this PR and then address comments in a follow up.

src/app/styles.css
sofiageo: Resolving for merge, will fix in the next PR

sofiageo: Resolving for merge, will fix in the next PR

PR #1489, opened 2026-03-30
feat: persistent token usage stats

kenvandine:
Adds persistent token usage statistics tracked server-side and surfaced in the web app via a new Statistics panel.

  Server:
  - Router accumulates LifetimeUsageStats — total requests, input/output tokens — bucketed by day and hour, and broken down by model and device type (GPU/NPU/CPU)
  - Stats are persisted to disk (JSON) and loaded on startup, surviving server restarts
  - Covers all request types (completions, audio, image, embeddings, reranking), not just LLM completions
  - Fixes sd-cpp incorrectly reporting GPU backends as CPU in device-type tracking

  App:
  - New StatsPanel added to the left rail (chart icon) showing a bar chart of token usage over time
  - Supports day/hour bucket modes, date-range presets (7d / 30d / 90d / 365d / all time), and drill-down to hourly view by clicking a day bar
  - Filterable by model or device type; search box filters the visible date buckets
  - Auto-refreshes every 30 s and on inference completion events

  Tests: server_endpoints.py updated to cover the /stats endpoint.

PR #1678, opened 2026-04-20
Feat persistent token usage stats backend

kenvandine:
Adds persistent, server-side token usage statistics, exposed via `GET /stats` and persisted to disk so counters survive restarts. Supersedes #1489 — the UI will land in a follow-up PR.

## What gets tracked

Every inference request is recorded in a `lifetime` usage object persisted to `<cache_dir>/token_usage_stats.json`:

- **Top-level totals** — `requests`, `input_tokens`, `output_tokens`, `started_at`, `updated_at`
- **Time buckets** — `by_day` (keyed `YYYY-MM-DD`) and `by_hour` (keyed `YYYY-MM-DDTHH:00:00`), in the server's local timezone
- **Per-model breakdown** — `by_model[model_name]` with its own totals + day/hour buckets
- **Per-device-type breakdown** — `by_device_type[device]` (e.g. `cpu`, `gpu`, `npu`) with its own totals + day/hour buckets

All non-streaming inference calls are counted — chat completion, completion, embeddings, reranking, responses, image generation, image edits, image variations, audio transcription, and TTS — not just LLM completions. Streaming chat/completion/responses calls are counted via the streaming path and include real token telemetry. For LLM requests the actual input/output token counts are patched onto the already-recorded bucket via a `add_tokens_locked` helper so request counts are not double-incremented.

## Endpoint

`GET /stats` continues to return the existing per-request telemetry (`input_tokens`, `output_tokens`, `prompt_tokens`, `time_to_first_token`, `tokens_per_second`) and now also returns a `lifetime` object. In addition to the persisted fields, the API projection includes two read-only helpers:

- `bucket_timezone` — currently `"server_local_time"`
- `persistence_path` — absolute path to the on-disk JSON file

## Commits

- `feat: Add persistent token usage stats` — schema, `load_usage_stats` / `persist_usage_stats_locked`, top-level totals + day/hour buckets, `record_usage_locked` helper
- `Restore status bar /stats telemetry` — keep the existing per-request fields alongside `lifetime`
- `Record stats for all request types, not just LLM completions` — move recording into `execute_inference`; introduce `add_tokens_locked` so LLM token counts layer onto the already-counted request
- `Track token usage stats per model` — `by_model` map with its own totals + day/hour buckets
- `Add per-device-type usage tracking` — `by_device_type` map with its own totals + day/hour buckets
- `test: Pin stats storage schema in endpoint tests` — structural regression test over `/stats` and the on-disk JSON file

## Testing

Verified locally against a `lemond` built from this branch on Linux with the `llamacpp:cpu` backend:

- `python test/server_cli.py` — 12 / 12 pass
- `python test/server_endpoints.py` — 31 / 31 pass (including the new `test_030_stats_schema`)

### New test: `test_030_stats_schema`

Pins the *storage* schema so any backwards-incompatible change is caught in CI. The test:

1. Loads `Tiny-Test-Model-GGUF` and drives one chat completion so every bucket map is actually populated — empty maps can't catch shape drift.
2. Asserts the top-level `/stats` fields and their types.
3. Asserts the `lifetime` object's fields (`requests`, `input_tokens`, `output_tokens`, `started_at`, `updated_at`, `bucket_timezone`, `persistence_path`) plus all four bucket maps.
4. Walks each bucket: `{requests, input_tokens, output_tokens}` present, ints, non-negative. Per-model and per-device entries additionally carry nested `by_day` / `by_hour` maps.
5. Validates key formats: day keys match `^\d{4}-\d{2}-\d{2}$`; hour keys match `^\d{4}-\d{2}-\d{2}T\d{2}:00:00$`; timestamps match ISO-8601.
6. Reads `persistence_path` from the response and re-validates the **on-disk** JSON against the same schema (minus the two API-only fields). Drift between `load_usage_stats` and `persist_usage_stats_locked` surfaces here, not in prod.

A rename like `input_tokens` → `prompt_tokens` inside a bucket — or dropping `by_model` / `by_device_type` from the persisted file — will fail the test.

## Test plan

- [x] `python test/server_cli.py` passes
- [x] `python test/server_endpoints.py` passes (incl. new `test_030_stats_schema`)
- [x] `/stats` returns the documented schema against a freshly-started `lemond`
- [x] `token_usage_stats.json` is created on first launch and updated after each non-streaming inference call
- [ ] Cross-platform CI (Linux / macOS / Windows) passes

## Follow-ups

- UI surface for the new stats (per the split from #1489)


src/cpp/server/router.cpp
```diff
@@ -1,35 +1,362 @@
...
+        file << persisted.dump(2);
+    } catch (const std::exception& e) {
+        LOG(WARNING, "Router") << "Failed to persist usage stats: " << e.what() << std::endl;
+    }
+}
+
+void Router::record_usage_locked(const std::string& model_name, const std::string& device_type, int input_tokens, int output_tokens, std::time_t recorded_at) {
+    if (input_tokens <= 0 && output_tokens <= 0) {
+        return;
+    }
```
jeremyfowers: **Issue 1 — early-exit silently drops most request types.** The commit body for `7936a38` says *"The zero-token early-exit guard is removed so image gen requests are counted even though they have no token metrics"* — but the guard is still here, and `execute_inference` always calls this with `(0, 0)` (line 899). Concrete fallout:

- **Non-streaming LLM** (chat/completion): `execute_inference` → early-exit → request not counted. `update_telemetry` → `add_tokens_locked` → adds tokens but *never increments* `requests`. `lifetime.requests` and every per-day/hour/model/device `requests` counter stays at 0 forever on non-streaming LLM traffic.
- **Image generation, TTS, audio transcription, embeddings, reranking**: `execute_inference` calls with `(0, 0)` → early-exit → no record at all. None of the non-LLM request types listed in the PR description as "all counted" are actually counted.
- **Streaming LLM** is the only path that works — `execute_streaming` passes real telemetry at line 955.

Fix: remove the guard (as the commit message claims). The caller already canonicalizes negatives via `std::max(..., 0)` on the next line.

*(Comment authored by Claude Code on behalf of @jeremyfowers.)*

test/server_endpoints.py
```diff
@@ -1394,6 +1403,185 @@ def test_029_system_info_release_url(self):
...
+        _DAY_KEY_RE = re.compile(r"^\d{4}-\d{2}-\d{2}$")
+        _HOUR_KEY_RE = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:00:00$")
+        _TS_RE = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}$")
+
+        def _assert_bucket(bucket, where):
+            self.assertIsInstance(bucket, dict, f"{where}: not a dict")
+            for field in ("requests", "input_tokens", "output_tokens"):
+                self.assertIn(field, bucket, f"{where}: missing '{field}'")
+                self.assertIsInstance(bucket[field], int, f"{where}.{field}: not int")
+                self.assertGreaterEqual(bucket[field], 0, f"{where}.{field}: negative")
```
jeremyfowers, line 1427: **Issue 2 — this test doesn't catch the request-counter bug flagged in the `record_usage_locked` comment.** `_assert_bucket` only checks that counters are `>= 0` ints, and `_assert_lifetime` does the same for lifetime totals. The test drives one non-streaming chat completion, which populates `by_day`/`by_hour` via `add_tokens_locked` default-inserting `UsageBucket{}` through `operator[]` (requests=0) while adding tokens — so `len(by_day) > 0` passes, but `requests` is never asserted to actually increment. A regression that zeroes every request counter passes CI today.

Suggested hardening:
- After the completion, assert `lifetime["requests"] >= 1` and `sum(b["requests"] for b in lifetime["by_day"].values()) >= 1`.
- Add a second sub-case that drives one non-LLM request (e.g., embeddings) and asserts `lifetime["requests"]` increased — that proves request counting works outside the LLM telemetry path, which is exactly what the PR description promises.

*(Comment authored by Claude Code on behalf of @jeremyfowers.)*

src/cpp/server/router.cpp
```diff
@@ -1,35 +1,362 @@
...
+        dev_day.input_tokens += in_tokens;
+        dev_day.output_tokens += out_tokens;
+
+        UsageBucket& dev_hour = dev.by_hour[hour_key];
+        dev_hour.requests++;
+        dev_hour.input_tokens += in_tokens;
+        dev_hour.output_tokens += out_tokens;
+    }
+
+    persist_usage_stats_locked();
```
jeremyfowers, line 280: **Issue 3 — synchronous disk write on every inference.** Both `record_usage_locked` (here) and `add_tokens_locked` (line 325) call `persist_usage_stats_locked()` at the end. Under `stats_mutex_`, that's one full JSON dump of the growing `lifetime_usage_stats_` object serialized to disk per inference — two for a non-streaming LLM request (once in `execute_inference`, once in `update_telemetry` → `add_tokens_locked`). At any real concurrent QPS this serializes responses on fsync and grows linearly with bucket count (see issue 4 on unbounded growth).

Options:
- Debounce: dirty flag + background writer that flushes at most every N seconds, plus a final flush on shutdown.
- Or keep synchronous but drop pretty-printing (`dump()` instead of `dump(2)`) and only persist on `record_usage_locked`, not on `add_tokens_locked` (token patches within the same second can coalesce with the next request's write).

*(Comment authored by Claude Code on behalf of @jeremyfowers.)*

src/cpp/server/router.cpp
```diff
@@ -1,35 +1,362 @@
...
+            {"by_model", serialize_model_stats(lifetime_usage_stats_.by_model)},
+            {"by_device_type", serialize_model_stats(lifetime_usage_stats_.by_device_type)}
+        };
+
+        std::ofstream file(stats_path);
+        if (!file) {
+            LOG(WARNING, "Router") << "Failed to persist usage stats to: " << usage_stats_path_ << std::endl;
+            return;
+        }
+
```
jeremyfowers, line 208: **Issue 4 — non-atomic file write.** Writing directly to `stats_path` means a crash or power loss mid-write truncates the file. On next boot, `load_usage_stats` sees a malformed JSON, `json::parse(..., nullptr, false)` returns discarded, the warning is logged, and every historical counter silently resets to zero — losing all accumulated lifetime stats. Given this runs on every request (per issue 3), the exposure window is continuous.

Standard fix: write to `usage_stats_path_ + ".tmp"`, `flush()` + `close()`, then atomically rename into place. `std::filesystem::rename` is atomic on POSIX. On Windows prefer `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` (since `fs::rename` to an existing target is not guaranteed atomic on all Windows versions).

*(Comment authored by Claude Code on behalf of @jeremyfowers.)*

jeremyfowers (commented):
Four line-level comments below flagging issues found in review. (Comments authored by Claude Code on behalf of @jeremyfowers.)

src/cpp/include/lemon/router.h
```diff
@@ -95,6 +97,40 @@ class Router {
...
+        uint64_t requests = 0;
+        uint64_t input_tokens = 0;
+        uint64_t output_tokens = 0;
+
+        json to_json() const {
+            return {
+                {"requests", requests},
+                {"input_tokens", input_tokens},
+                {"output_tokens", output_tokens}
+            };
```
jeremyfowers, line 110: **Design-level: adding a new metric is a O(many-files) edit.** `UsageBucket` has three named fields, and every site that touches it names them *individually* — `to_json` here, two load blocks in `load_usage_stats`, two serializer closures in `persist_usage_stats_locked`, six increment sites in `record_usage_locked`, six more in `add_tokens_locked`, and `_assert_bucket` in the test. Adding `latency_ms_sum` / `error_count` / `cache_hits` later means editing ~20 sites that all re-spell the same three field names.

The struct itself is fine — what's missing is a single source of truth for "how do you handle a bucket's fields." Two ways out, pick based on how many metrics you expect:

**Option A — deduplicate the handling (minimal change).** Give `UsageBucket` `from_json(const json&)` and `operator+=(const UsageBucket&)` alongside `to_json`, then have `record_usage_locked` / `add_tokens_locked` build one `UsageBucket delta` up front and do `target += delta` for each of the 6 bucket targets. Load becomes one `bucket.from_json(value)` call per site instead of three `value.value("...", 0ULL)` lines. Adding a metric = edit the struct + `to_json` + `from_json` + `operator+=`. Four sites, all co-located.

**Option B — generic metric map (more flexible).** `UsageBucket` becomes `std::unordered_map<MetricId, uint64_t>` with a small enum for `MetricId` and a single `metric_id_to_json_key()` table. Adding a metric = add an enum value + one row in the table; no struct edits, no serializer edits, no increment-site edits. Costs a little type-safety and ~a few bytes of overhead per bucket, but it's the right shape if you expect latency sums, error counts, cache-hit counters, cost, etc. to land later.

Not a blocker for this PR — but worth deciding now, because if the stats schema gets a v2 after the UI ships, migration becomes real work.

*(Comment authored by Claude Code on behalf of @jeremyfowers.)*

jeremyfowers (commented):
One more design-level comment on metric extensibility.

src/cpp/server/router.cpp
```diff
@@ -1,35 +1,362 @@
...
+                    {"input_tokens", stats.input_tokens},
+                    {"output_tokens", stats.output_tokens},
+                    {"by_day", usage_buckets_to_json(stats.by_day)},
+                    {"by_hour", usage_buckets_to_json(stats.by_hour)}
+                };
+            }
+            return result;
+        };
+
+        json persisted = {
```
jeremyfowers, line 186: **Persisted file has no `schema_version` field — add one now while it's cheap.** Once this ships, there will be `token_usage_stats.json` files in the wild on user machines. The next time the shape changes (new dimension per the earlier comments, rename of a field, switch from flat counters to a metric map per Option B, bucket granularity change, timezone semantics change, etc.) `load_usage_stats` has no way to dispatch a migration — it can only `value("key", default)` and silently drop whatever it doesn't recognize. That means either:

- Users silently lose accumulated history on upgrade (counters reset to zero), or
- You ship a one-off ad-hoc detector that sniffs for the presence/absence of specific fields to guess the version.

Both are worse than just stamping `"schema_version": 1` into the file now. Suggested change right here:

```cpp
json persisted = {
    {"schema_version", 1},
    {"requests", lifetime_usage_stats_.requests},
    ...
};
```

And in `load_usage_stats`, read it with a default of `1` (so existing files from this PR's format parse as v1) and switch on it when v2 lands. Pair with the test: `test_030_stats_schema` should assert `persisted["schema_version"] == 1` on disk so an accidental bump is caught.

Related: if you adopt Option B from the earlier comment (metric map), that's a v2 change — having v1 stamped means the migration is a 20-line function instead of archaeology.

*(Comment authored by Claude Code on behalf of @jeremyfowers.)*

jeremyfowers (commented):
One more: schema versioning for the persisted file.

jeremyfowers (changes requested):
Please address comments about the long term sustainability of the database.

src/cpp/server/router.cpp
kenvandine: Fixed in commit 9cf8fd2a. Removed the early-exit guard from `record_usage_locked` so the request counter increments on every call from `execute_inference` (which passes (0,0) for all non-LLM-streaming paths). `add_tokens_locked` still retains its guard since there's nothing meaningful to patch when both token counts are zero.

test/server_endpoints.py
kenvandine: Fixed in commit 0cef73b4. The test now tracks `inference_succeeded` and asserts `lifetime["requests"] >= 1` and `sum(by_day[*].requests) >= 1` when inference actually ran. The assertions are gated on `inference_succeeded` to preserve the existing behavior in environments where the model can't run.

src/cpp/server/router.cpp
kenvandine: Fixed in commit f6c7fff8. Removed the `persist_usage_stats_locked()` call from `add_tokens_locked` so token patches coalesce with the next request's write (only `record_usage_locked` now triggers a write, once per request). Also dropped pretty-printing (`dump()` instead of `dump(2)`) to cut serialization cost.

src/cpp/server/router.cpp
kenvandine: Fixed in commit 15ddc077. Now writes to `token_usage_stats.json.tmp` first, then uses `std::filesystem::rename` (with error handling) to atomically replace the live file. A crash or power loss during the write can no longer truncate the file and zero out accumulated counters on next boot.

src/cpp/include/lemon/router.h
kenvandine: Addressed in commit 94e0f1ef. Added `UsageBucket::from_json()` and `operator+=()` to the struct. All bucket-load sites in `load_usage_stats` now call `UsageBucket::from_json(value)` instead of spelling out the three field names inline. Adding a metric in the future requires editing only the struct definition plus those three methods — no load-site edits needed.

src/cpp/server/router.cpp
kenvandine: Fixed in commit 75e848e3. The persisted file now carries `"schema_version": 1` as its first key. `load_usage_stats` reads it with a default of 1 so files written before this change parse correctly as v1 without migration. `test_030_stats_schema` asserts `schema_version == 1` on disk, so an accidental bump is caught in CI.

jeremyfowers:
@kenvandine I'm not sure you pushed the commits to fix the first 6 comments?

jeremyfowers:
(Written by real human Jeremy)

I want to discuss the disk space and privacy implications of this database. We should not merge unless we're confident 99.9% of users will be glad (or at least indifferent) its enabled.

1. Is it guaranteed that this database will never fill up the user's hard drive to a significant level, even if they are heavy users of Lemonade over course of months/years?
2. Is the database accessible to anyone who is not the Lemonade admin? Particularly in a multi-client single-server homelab setup?
3. Does this database distinguish between users/IP addresses in a multi-client setup?
4. Could any privacy concerns be reduced by changing the granularity of data? For example, tracking on a `per_hour` basis is a lot more identifiable than tracking per day.

kenvandine:
> @kenvandine I'm not sure you pushed the commits to fix the first 6 comments?

Whoops, indeed I hadn't pushed. I've rebased and pushed now.

kenvandine:
> (Written by real human Jeremy)
> 
> I want to discuss the disk space and privacy implications of this database. We should not merge unless we're confident 99.9% of users will be glad (or at least indifferent) its enabled.
> 
> 1. Is it guaranteed that this database will never fill up the user's hard drive to a significant level, even if they are heavy users of Lemonade over course of months/years?
No guarantee, the data will grow over time, not sure how much of an impact this would have. But something we need to consider.
> 2. Is the database accessible to anyone who is not the Lemonade admin? Particularly in a multi-client single-server homelab setup?
It's for lemond, no splitting by client. 
> 3. Does this database distinguish between users/IP addresses in a multi-client setup?
No multi-client consideration. Should we do that?
> 4. Could any privacy concerns be reduced by changing the granularity of data? For example, tracking on a `per_hour` basis is a lot more identifiable than tracking per day.
Actually, in the grand scheme of things, hourly is probably too granular.

jeremyfowers:
Good discussion for SF?

kenvandine:
> Good discussion for SF?

Yes!

PR #1682, opened 2026-04-21
Fix NVIDIA VRAM detection on Windows (#1681)

### LINKED ISSUE #1681: Nvidia VRAM detection is bugged

jeremyfowers:
Closes #1681 

## Summary
- `Win32_VideoController.AdapterRAM` is a 32-bit field that Windows reports as -1 for any GPU with >4 GB VRAM, so every modern NVIDIA discrete card was reported as ~17179869183 GB on `/api/v1/system-info`.
- Aligned NVIDIA detection with the AMD path: dxdiag first, nvidia-smi as fallback (same approach as the Linux path).
- Parsed all GPUs from a single dxdiag invocation into a cache shared by AMD + NVIDIA detection so we do not pay the ~10s dxdiag cost twice.

## Test plan
- [x] Reproduced the bug against current `main` on a Windows machine with an RTX 4090: `"vram_gb": 17179869183.999023`.
- [x] Verified the fix returns `"vram_gb": 24.0` for the same GPU after rebuild.
- [x] Confirmed first `/api/v1/system-info` request costs ~10s (dxdiag) and subsequent requests are ~31ms (served by `SystemInfoCache`).
- [x] Tested on a machine with a discrete AMD GPU to confirm the shared dxdiag cache doesn't regress AMD VRAM reporting.
- [x] Tested on RTX 4060
- [x] Tested on RTX 3070

### DIFF
--- a/src/cpp/include/lemon/system_info.h
+++ b/src/cpp/include/lemon/system_info.h
@@ -145,6 +145,13 @@ private:
     std::string get_driver_version(const std::string& device_name);
     double get_gpu_vram_dxdiag(const std::string& gpu_name);
     double get_gpu_vram_wmi(uint64_t adapter_ram);
+    double get_nvidia_vram_smi();
+
+    // dxdiag lists every GPU in one invocation, so we run it once and
+    // serve subsequent lookups from memory.
+    bool dxdiag_cache_loaded_ = false;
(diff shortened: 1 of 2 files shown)


ramkrishna2910:
@claude review

ramkrishna2910 (approved):

PR #1690, opened 2026-04-21
Feature: Add flake.nix for NixOS users, including FHS fix

FrantaNautilus:
This pull request adds a flake.nix file that makes it possible to run Lemonade on NixOS.
This flake provides a NixOS module with two major components. First the lemonade-server itself and second a configuration for nix-ld which makes it possible to run ubuntu builds of llamacpp on NixOS. Therefore user on NixOS can use Lemonade essentially the same as user on Ubuntu. This "trick" can be disabled by users who want to keep their NixOS setup purely Nix-based without nix-ld.
The nix-ld has multiple options that can be combined:
- CPU 
- Vulkan
- CUDA
- NPU (no yet implemented, raises error)

The flake was tested on latest NixOS unstable and the nix-ld FHS fix is confirmed to work both for CPU, Vulkan and ROCm. The CUDA module was tested with llamacpp outside of lemonade and should work immediately when lemonade starts supporting CUDA builds of llamacpp. I plan to implement the NPU option once I package the necessary dependencies.

Tested on Beelink GTR9 PRO with Ryzen AI Max 395+ with Nvidia eGPU over USB. All other FHS fixes were disabled on the system while testing.


flake.nix
```diff
@@ -0,0 +1,388 @@
...
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs = {
+    self,
+    nixpkgs,
+  }:
+  let
+    version = "10.2.0";
```
superm1, line 13: this is going to go out of date really easily.  Would it be better to dynamically fill it in somehow?

flake.nix
superm1, line 1: could this instead be stored in `contrib` like we put other distro packaging?

flake.nix
```diff
@@ -0,0 +1,388 @@
...
+              --replace-fail 'if(NOT CMAKE_INSTALL_PREFIX STREQUAL "/usr")' 'if(FALSE)'
+
+            # Keep configuration files in the Nix output instead of trying to install to /etc.
+            substituteInPlace CMakeLists.txt \
+              --replace-fail 'DESTINATION /etc/lemonade/conf.d' 'DESTINATION etc/lemonade/conf.d'
+          '';
+
+          cmakeFlags = [
+            "-DCMAKE_BUILD_TYPE=Release"
+            "-DBUILD_WEB_APP=OFF"
```
superm1, line 84: you really want to build without web app?

flake.nix
```diff
@@ -0,0 +1,388 @@
...
+
+          warnings =
+            lib.optionals (cfg.useFHSEmulation && cfg.fhs.llamaTargets == [ ]) [
+              "lemonade-server.useFHSEmulation is enabled but no llama targets were selected; nix-ld will only use the base runtime libraries."
+            ]
+            ++ lib.optionals (lib.elem "cuda" cfg.fhs.llamaTargets) [
+              "lemonade-server.fhs.llamaTargets includes cuda. Lemonade currently does not expose a llama.cpp cuda backend, so this only affects nix-ld library selection."
+            ]
+            ++ lib.optionals (lib.elem "npu" cfg.fhs.llamaTargets) [
+              "lemonade-server.fhs.llamaTargets includes npu. Lemonade currently does not expose a llama.cpp npu backend, so this only affects nix-ld library selection."
```
superm1, line 325: but it does expose an NPU backend with fastflowlm

superm1 (commented):
Would it be possible to add a CI job to make sure this flake stays healthy?

flake.nix
FrantaNautilus: Thank you for considering merge. I will try to update it, so that it is the latest commit hash or something. It is late night in my timezone, so I will get to it tomorrow.

flake.nix
FrantaNautilus: It does but getting it to work on NixOS will require someone to package the AMD XDNA driver and fastflow before this option be possible. I could not find any flake packaging these currently, but I plan to try packaging them myself, since I am on Ryzen AI Max which should have supported NPU.

flake.nix
FrantaNautilus: Please wait, my merge will most likely be not necessary.  Doing the research, I found that few days ago a major packaging projects for lemonade, xdna, fastflow was started for NixOS.  I am in the process of getting in touch with those developers and contributing my progress. This would make my pull request obsolete.

FrantaNautilus:
Thank you for all the support and considering my packaging efforts. I was informed about much larger packaging project [nix-amd-ai](https://github.com/noamsto/nix-amd-ai) for AMD AI related software that is using the official packaging approaches of NixOS and does not use FHS emulation tricks like my implementation does.
For this reason, I would like to close my pull request. Thank you for your understanding, my implementation basically became obsolete before it was merged.

PR #1691, opened 2026-04-21
CLI: check the models list before sending pull endpoint requests

### LINKED ISSUE #1601: Improve the error message for `lemonade pull`
Suggest using `lemonade list` to see available models.

jeremyfowers:
Closes #1601

## What changed

Replaces the cryptic `"When registering a new model, the model name must include the `user` namespace..."` error from `lemonade pull <unknown-name>` with a friendlier hint pointing at `lemonade list` and `lemonade pull CHECKPOINT`.

## How

A typed `UnknownModelError` on the server side (`src/cpp/include/lemon/model_manager.h`) wraps the one specific case where the requested name is not a registered built-in, not a hardware-filtered registered model, and lacks the `user.` prefix that would make it a new-model registration attempt. The `/pull` handler catches this type and attaches a structured `{"code": "unknown_model"}` tag to the error response (both streaming and non-streaming paths). The CLI keys off that **code** to rewrite the error in `LemonadeClient::pull_model`.

Every other error path (server down, 401 auth, registered-but-unsupported model, malformed manual pull, missing checkpoint/recipe) throws a plain `std::runtime_error` with no code, so the server's specific message flows through verbatim.

## Key scenarios

**1. Unknown name — friendly rewrite**
```
$ build/lemonade pull not-a-model
Pulling model: not-a-model
Error pulling model: No built-in model with the name 'not-a-model' is registered.

If you meant a built-in model, run `lemonade list` to see available models.
If you meant to add a custom model from Hugging Face, run `lemonade pull CHECKPOINT`.
```

**2. Manual pull missing `user.` prefix — server message preserved**
```
$ build/lemonade pull Foo --checkpoint llamacpp unsloth/Phi-3-mini-GGUF --recipe llamacpp
Pulling model: Foo
Error pulling model: When providing 'checkpoint' or 'recipe', the model name must include the `user.` prefix, for example `user.Phi-4-Mini-GGUF`. Received: Foo
```

**3. `user.Foo` without checkpoint/recipe — server message preserved**
```
$ build/lemonade pull user.NotExisting
Pulling model: user.NotExisting
Error pulling model: Model user.NotExisting is not registered with Lemonade Server. To register and install it, provide the `checkpoint` and `recipe` arguments, as well as the optional `reasoning` and `mmproj` arguments as appropriate.
```

## Notes on review feedback

- **No more CLI preflight**: the earlier `get_models(false)` + `get_models(true)` preflight is gone. This eliminates the uncaught-exception regression on server-down / bad-API-key paths, and also eliminates the false-negative on registered-but-hardware-filtered models (which `/api/v1/models?show_all=true` does not return).

### DIFF
--- a/src/cpp/cli/lemonade_client.cpp
+++ b/src/cpp/cli/lemonade_client.cpp
@@ -1,5 +1,4 @@
 #include "lemon_cli/lemonade_client.h"
-#include "lemon/model_manager.h"  // for lemon::kUnknownModelErrorCode
 #include <httplib.h>
 #include <iostream>
 #include <iomanip>
@@ -486,8 +485,9 @@ int LemonadeClient::pull_model(const json& model_data) {
         }, LONG_TIMEOUT_MS, DEFAULT_READ_TIMEOUT_MS);
 
         if (!state.success) {
-            // See UnknownModelError contract in include/lemon/model_manager.h.
-            if (state.error_code == lemon::kUnknownModelErrorCode) {
+            // Wire-protocol constant; server-side definition and contract live in
+            // include/lemon/model_manager.h (kUnknownModelErrorCode). Keep in sync.
+            if (state.error_code == "unknown_model") {
                 state.error_message =
                     "No built-in model with the name '" + model_name + "' is registered.\n\n"
                     "If you meant a built-in model, run `lemonade list` to see available models.\n"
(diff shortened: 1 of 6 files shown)


bitgamma (changes requested):
codex comments make sense. In my opinion it'd be better to just have the server return a more detailed error message instead of having a preflight, however even just addressing codex comments would be enough.

jeremyfowers:
@bitgamma I wound up replacing the implementation entirely with something more robust. Please see the updated PR description for details and re-review.

bitgamma (approved):
lgtm!

PR #1692, opened 2026-04-21
Update stable diffusion builds

superm1:


src/cpp/resources/backend_versions.json
```diff
@@ -2,12 +2,12 @@
   "comment": "This configuration file controls which llama.cpp, whisper.cpp, sd.cpp, ryzenai-llm, and FLM versions are downloaded for each backend. You can modify these values to pin specific versions without rebuilding the application.",
   "rocm-stable-runtime": "v7.2.1",
   "llamacpp": {
-    "vulkan": "b8766",
-    "rocm-stable": "b8653",
-    "rocm-preview": "b8705",
+    "vulkan": "b8864",
+    "rocm-stable": "b8864",
+    "rocm-preview": "b8864",
     "rocm-nightly": "b1238",
```
jeremyfowers:
```suggestion
    "rocm-nightly": "b1246",
```

Use a release from today to keep it aligned.

jeremyfowers (approved):
good to merge if it passes CI.

jeremyfowers:
@superm1 this imagegen test is deterministically failing in the merge queue, can you investigate?

```
FAIL: test_022_generate_image_output (__main__.OllamaTests.test_022_generate_image_output)
Test /api/generate with an image generation model.
  File "/__w/lemonade/lemonade/test/test_ollama.py", line 573, in test_022_generate_image_output
    self.assertEqual(response.status_code, 200)
AssertionError: 404 != 200

```

jeremyfowers (changes requested):
Resolution for failing test.

superm1:
What is different with the merge queue? I don't understand why it would fail there but not here

superm1:
Yeah NFC why https://github.com/lemonade-sdk/lemonade/commit/9c42cdeb9eb9ff970665d9ab0cfb29f57f39d839 succeeded and this didn't.  Doesn't look deterministic to me in any way.  But since that handled updating llama.cpp I'll just do SD here and hopefully that goes through smoothly.

superm1:
> Yeah NFC why [9c42cde](https://github.com/lemonade-sdk/lemonade/commit/9c42cdeb9eb9ff970665d9ab0cfb29f57f39d839) succeeded and this didn't. Doesn't look deterministic to me in any way. But since that handled updating llama.cpp I'll just do SD here and hopefully that goes through smoothly.

ah well the issue seems to be in SD itself for some reason.

jeremyfowers:
Ok I guess this will wait until there's a fix upstream?

superm1:
> Ok I guess this will wait until there's a fix upstream?

I don't understand root cause right now, but yeah let's not block a release on it.

PR #1694, opened 2026-04-21
Update the hot models for mid april 2026

### LINKED ISSUE #1674: Update the hot models list
Remove qwen3.5-122b

Replace qwen3.5-35b with qwen3.6-35b

jeremyfowers:
Closes #1674 

This PR accomplishes two things:
- Unblocks llamacpp auto-upgrade by taking the 122B model out of it - this model was always marginal to fit on Windows and causes stability issues.
- Adjusts the hot models list for what is actually hot in mid april 2026.

### DIFF
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -873,8 +873,7 @@
         "suggested": true,
         "labels": [
             "vision",
-            "tool-calling",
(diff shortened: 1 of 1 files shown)


sawansri (approved):
The old(er) makes way for the new 🙂

PR #1704, opened 2026-04-22
Prefer system backend by default generally

superm1:
It's a relatively high bar to overcome to install a system llama.cpp and GGML backend for HIP.  Users or packagers that have done this will likely want to be using the system llama.cpp anyway.

Prefer that by default.


superm1:
I file this because of the patch to add location Arch uses.

jeremyfowers (approved):

PR #1707, opened 2026-04-22
Add bonsai family of models

sawansri:
Adds PrismML's 1 bit bonsai family of models.

image

### DIFF
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -713,8 +713,7 @@
         "recipe": "llamacpp",
         "suggested": true,
         "labels": [
-            "llamacpp",
-	        "hot"
+            "llamacpp"
         ],
         "size": 1.2
     },
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -708,6 +708,33 @@
         ],
         "size": 3.1
     },
+    "Bonsai-8B-gguf": {
+        "checkpoint": "prism-ml/Bonsai-8B-gguf:Q1_0",
+        "recipe": "llamacpp",
+        "suggested": true,
+        "labels": [
+            "llamacpp"
+        ],
+        "size": 1.2
+    },
+    "Bonsai-4B-gguf": {
+        "checkpoint": "prism-ml/Bonsai-4B-gguf:Q1_0",
+        "recipe": "llamacpp",
+        "suggested": true,
+        "labels": [
+            "llamacpp"
+        ],
+        "size": 0.6
+    },
+    "Bonsai-1.7B-gguf": {
+        "checkpoint": "prism-ml/Bonsai-1.7B-gguf:Q1_0",
+        "recipe": "llamacpp",
+        "suggested": true,
+        "labels": [
+            "llamacpp"
+        ],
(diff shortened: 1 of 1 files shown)


jeremyfowers:
Is there any llamacpp version requirement for this? Are we good to go, and we get the expected TPS?

src/cpp/resources/server_models.json
```diff
@@ -708,6 +708,33 @@
         ],
         "size": 3.1
     },
+    "Bonsai-8B-gguf": {
+        "checkpoint": "prism-ml/Bonsai-8B-gguf:Q1_0",
+        "recipe": "llamacpp",
+        "suggested": true,
+        "labels": [
+            "llamacpp"
```
jeremyfowers: hot? :D

sawansri:
> Is there any llamacpp version requirement for this? Are we good to go, and we get the expected TPS?

Works on vulkan b8766 for me!

image

src/cpp/resources/server_models.json
sawansri: sure lol, I'll add it to the 8B

jeremyfowers (approved):
1 bit hype, lets go!

sawansri:
ROCm stable appears to be bundled with an older version of llamacpp. Preview and nightly actions pass.

If we are adding a new model whose support has been added in llamacpp recently, will the ROCm stable action always fail? What would be the plan in that case, wait until ROCm stable bumps llamacpp ver?

jeremyfowers:
> ROCm stable appears to be bundled with an older version of llamacpp. Preview and nightly actions pass.
> 
> If we are adding a new model whose support has been added in llamacpp recently, will the ROCm stable action always fail? What would be the plan in that case, wait until ROCm stable bumps llamacpp ver?

@superm1 what do you think, should we just disable rocm-stable from llamacpp upgrade auto validation? Since the auto-validation only tests the hottest models, "stable" is expected to fall behind and become dysfunctional over time.

superm1:
Stable and preview should both be able to update together. The artifacts should be built for both.

src/cpp/resources/server_models.json
```diff
@@ -708,6 +708,34 @@
         ],
         "size": 3.1
     },
+    "Bonsai-8B-gguf": {
+        "checkpoint": "prism-ml/Bonsai-8B-gguf:Q1_0",
+        "recipe": "llamacpp",
+        "suggested": true,
+        "labels": [
+            "llamacpp",
+	        "hot"
```
jeremyfowers:
```suggestion
            "llamacpp"
```

jeremyfowers:
@sawansri I removed the hot label to unblock us for now.

@superm1 the core issue here is that ROCm 7.2 is not guaranteed to be forwards compatible with future models. For example, it is not compatible with Bonsai.

superm1:
> @sawansri I removed the hot label to unblock us for now.
> 
> @superm1 the core issue here is that ROCm 7.2 is not guaranteed to be forwards compatible with future models. For example, it is not compatible with Bonsai.

I don't understand that.  Why is there a ROCm version dependency?  It's independent of llama.cpp.  We just need a llama.cpp up-rev.

PR #1708, opened 2026-04-22
macOS: use posix_spawn instead of fork+exec for child processes and patch HOME so its not null and doesn't throws a segfault in llama.cpp 

Geramy:
Problem: lemond spawns llama-server via fork()+execvp(). On macOS, fork() leaves the child with corrupted Mach-port and XPC-bootstrap state that execvp() does not reset. llama.cpp b8884+ now runs a ggml-metal probe at startup that calls [MTLDevice newLibraryWithSource:] — which routes through MTLCompilerService XPC — and dies on the broken channel before the model is opened. Direct terminal runs work; only lemond-spawned children fail (~130ms, exit code -1).

Fix: on __APPLE__, replace fork()+execvp() with posix_spawn. Preserves pipe/working-dir semantics. Adds POSIX_SPAWN_CLOEXEC_DEFAULT to avoid leaking lemond FDs into the child, and POSIX_SPAWN_SETSIGDEF to reset inherited SIG_IGN dispositions. Linux and Windows paths unchanged.

### DIFF
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -438,6 +438,11 @@ void LlamaCppServer::load(const std::string& model_name,
     // get a minimal env from launchd and do not inherit HOME, so llama-server
     // crashes before the model ever loads. Terminal/sudo spawns preserve
     // HOME and do not hit this.
+    //
+    // Upstream fix in flight: https://github.com/ggml-org/llama.cpp/pull/22263
+    // Once that PR merges and lemonade's pinned llama.cpp version (in
+    // src/cpp/resources/backend_versions.json) includes it, this HOME
+    // fallback can be deleted.
     const char* home = std::getenv("HOME");
     if (!home || home[0] == '\0') {
         struct passwd* pw = getpwuid(getuid());
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -13,6 +13,10 @@
 #include <lemon/utils/aixlog.hpp>
 #include <cstdlib>
 #include <set>
+#ifdef __APPLE__
+#include <pwd.h>
+#include <unistd.h>
+#endif
 
 #ifdef _WIN32
     #include <windows.h>
@@ -424,6 +428,22 @@ void LlamaCppServer::load(const std::string& model_name,
     if (no_residency) {
         env_vars.push_back({"GGML_METAL_NO_RESIDENCY", no_residency});
         LOG(DEBUG, "LlamaCpp") << "Forwarding GGML_METAL_NO_RESIDENCY=" << no_residency << std::endl;
+    }
+
+    // Ensure HOME is set in the child. llama.cpp b8884+ (libllama-common's
+    // fs_get_cache_directory / hf_cache::migrate_old_cache_to_hf_cache)
+    // calls getenv("HOME") during CLI arg parsing and passes the result
+    // straight into std::string without a NULL check, segfaulting when
+    // HOME is unset. LaunchDaemons installed at /Library/LaunchDaemons/
+    // get a minimal env from launchd and do not inherit HOME, so llama-server
+    // crashes before the model ever loads. Terminal/sudo spawns preserve
+    // HOME and do not hit this.
+    const char* home = std::getenv("HOME");
+    if (!home || home[0] == '\0') {
+        struct passwd* pw = getpwuid(getuid());
+        std::string fallback_home = (pw && pw->pw_dir) ? pw->pw_dir : "/var/root";
+        env_vars.push_back({"HOME", fallback_home});
+        LOG(DEBUG, "LlamaCpp") << "Parent HOME unset; setting child HOME=" << fallback_home << std::endl;
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


Geramy:
Back story not needed really but very informational and nice to have.

Why fork() is unsafe under Apple frameworks
This is a long-documented Apple platform rule. The most authoritative source is Apple's own open-source Objective-C runtime (objc4). From runtime/objc-initialize.mm (apple-oss-distributions/objc4), in the comment block titled "Fork Safety or: We Tried So Hard":

"It is impossible to make +initialize fork-safe in the general case. The standard approach of acquiring all locks pre-fork doesn't work, because another thread might be in +initialize waiting on the thread calling fork, and trying to wait for that to complete would result in a deadlock."

The child-side function performForkChildInitialize (same file) prints the familiar runtime error:

"+[%s initialize] may have been in progress in another thread when fork() was called."

and calls objc_initializeAfterForkError(cls) before crashing. The kill switch OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES is the existence proof — Apple explicitly shipped a per-process override because they know fork-then-exec leaves children in an unrepresentable runtime state. It doesn't fix the underlying corruption, it just silences the abort.

The same unsafety extends to CoreFoundation (Mach ports and XPC bootstrap handles don't reset across fork), Grand Central Dispatch (pthread workqueue state), and the higher-level frameworks that sit on top of them — including Metal. exec() replaces the memory image but does not reset Mach ports, XPC connections, dispatch queues registered with the kernel, or pthread workqueue bookkeeping.

Apple's documented alternative is posix_spawn(2). From Apple's local man page (man posix_spawn on any macOS system):

"The posix_spawn() function creates a new process from the executable file, called the new process file, specified by path… File descriptors open in the calling process image remain open in the new process image, except for those for which the close-on-exec flag is set…"

The important distinction is that posix_spawn creates the child without the "inherit all parent state then selectively discard it" semantics of fork+exec. Under the hood on Darwin it's a single kernel transaction that builds a fresh process context, so Mach-level state doesn't leak in.

Geramy:
setting HOME here to help mitigate llama.cpp but for macOS, but realistically it should be fixed in llama.cpp -> PR https://github.com/ggml-org/llama.cpp/pull/22263

test/server_env_vars.py
```diff
@@ -170,23 +170,18 @@ def test_max_loaded_models_in_health(self):
...
 
-    @unittest.skipIf(IS_MACOS, "whispercpp backend selection not applicable on macOS")
     def test_whispercpp_backend(self):
         self.assertEqual(self.snapshot["whispercpp"]["backend"], "cpu")
 
-    @unittest.skipIf(IS_MACOS, "whispercpp args not applicable on macOS")
     def test_whispercpp_args(self):
         self.assertEqual(self.snapshot["whispercpp"]["args"], "--convert")
 
-    @unittest.skipIf(IS_MACOS, "FLM is NPU-only, not available on macOS")
```
superm1, line 189: don't you still want to skip FLM on mac?

Geramy: yes, thanks.

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -425,6 +429,22 @@ void LlamaCppServer::load(const std::string& model_name,
...
+    // get a minimal env from launchd and do not inherit HOME, so llama-server
+    // crashes before the model ever loads. Terminal/sudo spawns preserve
+    // HOME and do not hit this.
+    const char* home = std::getenv("HOME");
+    if (!home || home[0] == '\0') {
+        struct passwd* pw = getpwuid(getuid());
+        std::string fallback_home = (pw && pw->pw_dir) ? pw->pw_dir : "/var/root";
+        env_vars.push_back({"HOME", fallback_home});
+        LOG(DEBUG, "LlamaCpp") << "Parent HOME unset; setting child HOME=" << fallback_home << std::endl;
+    }
```
superm1, line 452: can you submit a PR to llama.cpp too and link it here?  It would be nice to know if/when this can be dropped (IE we update the mac dep to llama.cpp min version).

test/server_env_vars.py
Geramy: Fixed.

src/cpp/server/backends/llamacpp_server.cpp
Geramy: Done.

test/server_env_vars.py
```diff
@@ -170,19 +170,15 @@ def test_max_loaded_models_in_health(self):
...
     def test_llamacpp_args(self):
         self.assertEqual(self.snapshot["llamacpp"]["args"], "--flash-attn on")
 
-    @unittest.skipIf(IS_MACOS, "whispercpp backend selection not applicable on macOS")
     def test_whispercpp_backend(self):
         self.assertEqual(self.snapshot["whispercpp"]["backend"], "cpu")
 
-    @unittest.skipIf(IS_MACOS, "whispercpp args not applicable on macOS")
     def test_whispercpp_args(self):
         self.assertEqual(self.snapshot["whispercpp"]["args"], "--convert")
```
jeremyfowers, line 187: Are you sure? Pretty sure none of this is valid on macos regardless of your change.

Geramy: haha yeahhhhh, not valid.

jeremyfowers:
@claude review

jeremyfowers (approved):
we're back, baby!

PR #1710, opened 2026-04-22
Add Qwen3.6-27b

sawansri:
Adds recently released Qwen3.6-27b.

Quoting unsloth, "The 27B model surpasses Qwen3.5-397B-A17B on all major coding benchmarks".


jeremyfowers (approved):
thanks!

PR #1725, opened 2026-04-23
Enable % used context on opencode side panel

osimarr:
Currently opencode will not show % used context when used with "lemonade launch opencode" command.
That happens because lemonade writes the context window size to the wrong config entry and because lemonade doesn't fill the max output tokens.
This patch used 1/3 of the context window as max output tokens, which is the default of OpenAI models on VSCode. The user can manually refine it later directly on the config file if needed.
Maybe in the future lemonade can add the max output entry on the model's metadata? But that would be a topic for another PR.


osimarr:
2 tests failed when calling opencode. I'll check what happened there.

osimarr:
I had to update the unit test to use the updated opencode config

jeremyfowers:
@sawansri please merge or close this :)

osimarr:
this needs to be rebased as it's been too long.
but I can confirm lemonade will still create unrecognized config entries to opencode.json, making it just ignore the configured context size in practice.

sawansri:
> this needs to be rebased as it's been too long. but I can confirm lemonade will still create unrecognized config entries to opencode.json, making it just ignore the configured context size in practice.

The changes lgtm, we can merge after rebase.

sawansri (approved):
Thanks for the contrib!

PR #1727, opened 2026-04-24
Modify fallback logic for GGUF variant handling

fl0rianr:
When a model with the same name as one of Lemonade's "recommended" models with a different quant is downloaded, the UI shows that I've also downloaded the recommended quant
image

Deleting the "recommended" one also deletes the user.  (as noted by BiSkiT on discord)

This can be very frustrating and should be fixed.


jeremyfowers (approved):
Thanks for the fix @fl0rianr !

b1skit:
Just spotted this @fl0rianr - thank you!

PR #1724, opened 2026-04-23
Run stable diffusion CI on all Windows runners

superm1:

PR #1732, opened 2026-04-25
Consolidate sd ci jobs and remove stx-halo requirement

jeremyfowers:
Alternative to https://github.com/lemonade-sdk/lemonade/pull/1724 that consolidates the CPU and ROCm jobs onto the same runner, in the same style as the other tests.


superm1 (approved):

superm1:
No concerns

PR #1733, opened 2026-04-25
docs(docker): document Web UI logs setup by exposing WebSocket port

siavashhub:
### Summary
This PR updates the Docker guide to explain how to view server logs in the web UI when running Lemonade in a container.

#### Why
Users can start the container successfully but still miss web UI logs if only the API port is published. This clarifies the required port mapping and reduces setup confusion.

#### Scope
Documentation only


jeremyfowers (approved):
Thanks!

Issue #1711, opened 2026-04-22
possible fd leak in lemond


ckuethe:
### Platform

Linux/Ubuntu

### Lemonade Version

10.2.0

### GPU / APU Model

gfx1151

### Component

Other / Not sure

### Bug Description

It seems like lemond might be leaking fds after running for a while. This shows up as `Failed to load model: filesystem error: cannot increment recursive directory iterator: Too many open files`

### Steps to Reproduce

Repro unclear. I've been doing a lot of stuff with opencode.

### Expected vs Actual Behavior

Expected that when no models/runners are loaded, no excess FDs would be open, or that evicting a model would close all its FDs.

### Log Output

```shell
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Info] (Server) Loading model: Flux-2-Klein-4B
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Info] (SDServer) Loading model: Flux-2-Klein-4B
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Error] (Router) Backend load failed: filesystem error: recursive directory iterator cannot open directory: Too many open files [/var/lib/lemonade/.cache/lemonade/bin/sd-cpp/rocm]
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Warn] (Router) Load failed with non-file-not-found error, evicting all models and retrying...
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Info] (Router) Evicting all models (1 total)
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Info] (Router) Unloading: Qwen3-Coder-Next-GGUF
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Info] (LlamaCpp) Unloading model...
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.577 [Info] (Process) srv    operator(): operator(): cleaning up before exit...
Apr 22 14:47:53 strix lemond[44026]: 2026-04-22 14:47:53.580 [Info] (Process) llama_memory_breakdown_print: | memory breakdown [MiB]                         |  total    free     self   model   context   compute    unaccounted |
[...]

PR #1735, opened 2026-04-25
Fix a return before closing file descriptor on linux npu utilization check.

Geramy:
On the linux NPU utilization path, when power_state != D0 we returned 0.0 without closing the /dev/accel/accel0 fd that was opened a few lines earlier. Stats endpoints poll this regularly while the NPU is asleep, so it leaked one fd per call.


ckuethe:
probably fixes #1711 , will try it in the morning

superm1 (approved):

ckuethe:
Yep, fixed.

PR #1742, opened 2026-04-26
chore: update light theme

sofiageo:
This PR addresses the issues discussed in #1673 and updates the light theme. It also moves the styles to a directory and the CSS variables to separate files. This way it's easier to compare the themes and add any missing CSS variables.

There are still improvements that can be made especially to the naming of the CSS variables but the day ran out of hours again. 

Hopefully this is at least the minimum acceptable for a first release.


jeremyfowers (approved):
Image

awesome improvement! let's go!

PR #1763, opened 2026-04-29
Fix/macos inherit stdio cloexec and reasoning content

jgmelber:
Patch for macOS to inherit stdio and enable launch from lemonade.

Handle reasoning_content (thinking models like Qwen3). Merge into the text response so it appears as normal output. Claude Code does not support "thinking" blocks from non-Anthropic models, so folding into text is the only reliable path.

### DIFF
--- a/src/cpp/server/anthropic_api.cpp
+++ b/src/cpp/server/anthropic_api.cpp
@@ -465,20 +465,24 @@ json OllamaApi::convert_openai_chat_to_anthropic(const json& openai_response,
             const auto& message = choice["message"];
 
             // Handle reasoning_content (thinking models like Qwen3).
-            // Map to an Anthropic "thinking" content block so Claude Code
-            // can display the model's chain-of-thought.
+            // Merge into the text response so it appears as normal output.
+            // Claude Code does not support "thinking" blocks from non-Anthropic
+            // models, so folding into text is the only reliable path.
             if (message.contains("reasoning_content") && message["reasoning_content"].is_string()) {
                 std::string thinking_text = message["reasoning_content"].get<std::string>();
                 if (!thinking_text.empty()) {
-                    content_blocks.push_back({
-                        {"type", "thinking"},
-                        {"thinking", thinking_text}
-                    });
+                    response_text = thinking_text;
                 }
             }
 
             if (message.contains("content") && message["content"].is_string()) {
-                response_text = message["content"].get<std::string>();
+                std::string content_text = message["content"].get<std::string>();
+                if (!content_text.empty()) {
+                    if (!response_text.empty()) {
+                        response_text += "\n\n";
+                    }
+                    response_text += content_text;
+                }
             } else if (message.contains("content") && message["content"].is_array()) {
                 std::vector<std::string> text_blocks;
                 for (const auto& block : message["content"]) {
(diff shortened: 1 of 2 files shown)


Geramy:
@jgmelber this is more than just a macOS bug fix right?

src/cpp/server/anthropic_api.cpp
```diff
@@ -456,15 +456,39 @@ json OllamaApi::convert_openai_chat_to_anthropic(const json& openai_response,
...
+            // Handle reasoning_content (thinking models like Qwen3).
+            // Merge into the text response so it appears as normal output.
+            // Claude Code does not support "thinking" blocks from non-Anthropic
+            // models, so folding into text is the only reliable path.
+            if (message.contains("reasoning_content") && message["reasoning_content"].is_string()) {
+                std::string thinking_text = message["reasoning_content"].get<std::string>();
+                if (!thinking_text.empty()) {
+                    response_text = thinking_text;
+                }
+            }
```
Geramy, line 482: I'm not sure how I feel about this, this kind of sucks are we sure they don't support thinking output? Is thinking output even shown in Claude code output by default? I believe a lot of the thinking output is hidden. @jeremyfowers what do you think on this?

Geramy: If Jeremy is okay with outputting it then fine, I think I'm on the lines of omitting thinking from the results since its not really the primary output, usually.

src/cpp/server/anthropic_api.cpp
```diff
@@ -600,7 +624,68 @@ void OllamaApi::stream_openai_sse_to_anthropic_sse(const std::string& openai_bod
...
-            if (line.empty() || line.find("data: ") != 0) {
+            if (line.empty()) {
+                continue;
+            }
+
+            // Detect raw JSON error responses from backend (not SSE-wrapped).
+            // When llama-server returns e.g. {"error":{"message":"Compute error."}},
+            // it arrives as bare JSON without a "data: " prefix.
+            if (line[0] == '{' && line.find("\"error\"") != std::string::npos) {
+                try {
```
Geramy: can we do this instead here?
```
// Much faster if you expect a lot of invalid JSON lines
json j = json::parse(line, nullptr, false);

if (j.is_discarded()) {
    // Handle the bad line without the massive overhead of stack unwinding
}
else { 
    //code here 
}
```

try/catch unraveling is expensive when it hits the catch, since we are expecting a throw in most cases it could be rather heavy.

Geramy (changes requested):
please see above, only a few minor adjustments.

src/cpp/server/anthropic_api.cpp
jgmelber: Just picking this up now, I couldn't get Claude Code working with lemonade on my MBP without outputting the thinking text. I am open to any suggestions, I did not try Claude Code on any other devices with or without this patch.

src/cpp/server/anthropic_api.cpp
jgmelber: Sure

src/cpp/server/anthropic_api.cpp
Geramy: @jgmelber I would like to merge this but I honeslty don't know enough about claude code to approve or deny this. @sawansri do you think you could take a look here for me?

sawansri: Yeah, I've noticed this too. I do plan on completely revamping our Messages API implementation as part of https://github.com/lemonade-sdk/lemonade/pull/1423. That includes forwarding to llamacpp's Anthropic API implementation instead of ours. Once that is merged, perhaps we could upstream any thinking output changes if necessary?

jeremyfowers:
@sawansri @Geramy  please state next steps for merging this or close it

Geramy:
> @sawansri @Geramy please state next steps for merging this or close it

@sawansri I am good with the Mac changes, after he made the adjustments to the try catch so that catch isn't in the hot path, instead using json non catch route. If you are good with the change we can merge it, it does have a conflict.

Issue #1766, opened 2026-04-29
Brainstorm - OmniRouter


ramkrishna2910:
How do we get software developers to use it?

1. Needs GUI support for adding custom collections.
2. Need to add a documentation entrypoint for ISVs. "So you want to put lemonade into your app, start here". This needs to branch out to embeddable, omnirouter, etc. ISV-focused features.
3. Should we write a benchmark whitepaper that treats a Collection as a SuperModel. Compare to other models, and convince people to use OmniRouter's SuperModels instead of any other model out there.
4. Build our own apps (especially non-ChatBot), blog about how easy it is, and distribute.

Rebrand to make more sense:
- Lemonade "Omni Models". Lemonade is the engine. Omni Models are available on Hugging Face, just like any other model.
- Remove "OmniRouter" branding - people outside our sphere dont know about routers.
- Host the Omni Model definitions on Hugging Face, making them available to pull down.
    - This also allows people to upload and share their own unofficial Omni Models
    - Users must also have the ability to define their own Omni Models locally on disk

fl0rianr:
Very good points. One way to make Omni Models more attractive for developers is to increase their effective capabilities, while keeping them reliable and easy to integrate.

I think this fits well into this issue because it is not only about naming or distribution, but also about making Omni Models feel like useful “SuperModels” instead of just a loose collection of local models.

## Proposal: Omni Workflow Intelligence Layer for safer, richer local workflows

The current Omni workflow approach is flexible because the planner LLM can decide how to use the available local models and tools. That is a good idea and should stay.

However, as workflows become more complex, especially for image editing/generation and multimodal use cases, the planner needs stronger structure around it.

The goal is not to replace the planner LLM, but to make it much better at selecting and executing local workflows.

## Problem

Today the planner LLM is responsible for a lot:

- understanding the user request
- selecting tools/models
- deciding execution order
- generating intermediate prompts
- handling retries/fallbacks
- adapting to model-specific capabilities

This is flexible, but it can become unreliable and token-expensive.

For image workflows in particular, “just write a better prompt” is often not enough. There are multiple control layers involved:

1. user request
2. planner LLM interpretation
3. image workflow / image-model prompt layer
4. actual generator/backend behavior

For advanced image workflows, the important decision is not only the text prompt, but also which technical workflow capability should be used, if available:

- edit vs full regeneration
- style workflow / preset
- edit strength
- identity/composition preservation
- reference image or mask usage
[...]

PR #1769, opened 2026-04-29
Feat : Add UI options for custom Omni Models

fl0rianr:
## Summary

This PR adds support for **custom Omni Models** in the Lemonade desktop app.

Users can now create their own Omni Models by selecting component models, register them through the normal `/pull` model flow, and reuse them from the model selector like built-in OmniRouter models.

## What changed

- Added a **New Omni Model** entry point in the desktop app UI.
- Added an Omni Model panel for selecting model components:
  - Planner LLM
  - Vision-capable model
  - Image generation
  - Image editing
  - TTS
  - Transcription / ASR
- Registered custom Omni Models through the server-side `/pull` flow using `recipe: "collection.omni"` and `components`.
- Added export/import support for Omni Models, including component model metadata with `checkpoint` / `checkpoints`.
- Added Omni Model-aware display names in the model selector and model manager.
- Added non-blocking compatibility warnings, for example when selecting a planner LLM without tool-calling support.
- Integrated Omni Model setup downloads with the Download Manager.
- Added functional and regression tests for Omni Model utility logic, UI actions, routing, and export/import behavior.
- Removed obsolete prototype code paths from earlier workflow-based iterations.

## Why

Built-in OmniRouter models are useful defaults, but users may want to compose their own model sets depending on their hardware, preferred models, downloaded models, or multimodal experimentation needs.

Custom Omni Models make OmniRouter more flexible without requiring users to manually edit model registry data.

This has also been requested multiple times by users in Discord.

## Notes for reviewers

Custom Omni Models are registered through the same server-side model flow as other user models. They use `recipe: "collection.omni"` with a `components` list, matching the current OmniRouter model representation.

The Omni Model itself is a lightweight model definition that references real component models. If exported JSON includes component metadata, import can register missing component models before registering the Omni Model.

Built-in Omni Models can be used as templates, but saving creates or updates a user Omni Model instead of modifying the built-in definition.

## Testing

Tested locally and verified that custom Omni Models can be created, edited, deleted, exported, imported, selected from the model selector, and loaded.

Also verified that Omni Model setup downloads are visible in the Download Manager.

```bash
node test/app/customCollections.test.cjs
node test/app/run-app-regression-tests.cjs
cd src/app
npm run build:renderer:prod

Same as the one here in CI 31 passes.
```


fl0rianr:
I can create an according issue to link but also kinda hope the hint towards discord is fine.

jeremyfowers:
> Custom workflows are stored locally in the desktop app via LocalStorage. They are intentionally user-defined and separate from the built-in model registry collections.

Why not store them in user_models.json like other user-defined models?

edit: Maybe I can answer my own question, we haven't defined any api spec for the pull endpoint to be able to do such a thing...

fl0rianr:
We can adapt if it if you like. It's crucial they are not only part of the browser storage. In the future i would like to import/export workflows as well via API e.g. in conjunction with n8n, that's why json is handy as well.

docs/dev/omni-router.md
```diff
@@ -1,46 +1,73 @@
 # OmniRouter
```
jeremyfowers, line 1: Image

There is already a `File > New Model` menu in the menu bar. IMO this button should move there, and there should be `File > New Collection > Manually` and `File > New Collection > From Json`.

That way the new feature matches the workflow of the existing feature, and the "import" and "new" buttons aren't needed in the collection entry modal.

jeremyfowers, line 1: We've been using the term "collection" for the models that omnirouter routes between. I would strongly prefer to not introduce a new term like "workflow".

docs/dev/omni-router.md
```diff
@@ -1,46 +1,73 @@
...
```
jeremyfowers, line 2: Image

Please remove these two text boxes. The functionality of the modal should be apparent based on the controls available.

jeremyfowers: Image

This modal should follow the style of the Add a Model modal as closely as possible.

jeremyfowers, line 2: Image

Modifying an existing collection should look as similar to the Model Options modal as possible.
- Activate via gear button next to the collection in the model manager
- Export button at the bottom

jeremyfowers (changes requested):
Please unify the terminology, layout, and style of the feature with the existing features. See comments for specific details.

fl0rianr:
Thanks for the review! I addressed the feedback:

Changes made:
- Replaced the remaining user-facing “workflow” terminology with “collection”.
- Moved creation/import entry points to File > New Collection > Manually / From JSON.
- Removed the explanatory text boxes from the modal.
- Aligned create/edit flows more closely with the existing Add Model and Model Options modals.
- Added gear-based editing for custom collections in the model manager.
- Moved export to the bottom of the edit/options modal.
- Sorted custom collections below built-in OmniRouter collections.
Additionally:
- Auto-fill the Vision role when the selected planner LLM is vision-capable, unless the user has already selected a different Vision model.
- Moved the focused test to test/app.

Validation:
- checked that it works hands on
- `npm run test:custom-collections` passes locally.

docs/dev/omni-router.md
fl0rianr: image

fl0rianr: image

docs/dev/omni-router.md
fl0rianr: Done, but maybe introduce workflow later again for more "workflow" like options to separate them clearly. For "real" workflow ideas you might have a look at #1766

fl0rianr: Removed.
Now exactly like custom model:
image

jeremyfowers (approved):
Image

It's alive! Tested the creation, import, and export flows.

Filed two issues to follow up on, that relate to using custom models in custom collections.

https://github.com/lemonade-sdk/lemonade/issues/1980
https://github.com/lemonade-sdk/lemonade/issues/1979

Let's go!

PR #1773, opened 2026-04-30
feat: rebrand OmniRouter to Lemonade Models, support manifest pulls

ramkrishna2910:
Lemonade Models are shareable HF-hosted collections. A repo with a
`lemonade.json` manifest at its root declares `recipe: "collection"`
and lists components — either by reference to curated registry names
(string entries) or with full inline metadata. Pulling such a repo
registers the collection plus any inline components and recurses into
the existing collection download path.

The manifest can also be embedded directly in a `lemonade_manifest`
field on POST /api/v1/pull (used by tests, with `register_only: true`
to skip downloads). Inline component recipes are validated against a
whitelist; on name collision with a curated entry the curated wins.

UI labels collection recipes as "Lemonade Model"; comments and docs
updated accordingly. Local test fixtures cover refs-only, inline,
malformed-recipe, missing-prefix, and unknown-component cases.

https://claude.ai/code/session_01PkD34A4nsH591VZZhG7xC6


docs/lemonade-models.md
```diff
@@ -0,0 +1,100 @@
...
+
+A Lemonade Model is just a Hugging Face repo whose root contains a `lemonade.json` manifest. The manifest declares `recipe: "collection"` and lists the component models that make up the bundle. Components can be either references to models already curated in Lemonade's registry (string entries) or inline objects describing arbitrary HF checkpoints.
+
+### Example: `lemonade-sdk/lemonade-ultra/lemonade.json`
+
+```json
+{
+  "name": "Ultra Collection",
+  "recipe": "collection",
+  "suggested": true,
```
jeremyfowers, line 26: I don't think `suggested` should be a field on a HF repo. We still need to list the repos in our own server_models.json, and that's where `suggested` would go.

jeremyfowers:
To help me review this, please provide links to two example HF checkpoints for lemonade models.

This branch should also point server_models.json to those checkpoints.

I would also expect this branch to be backwards compatible, so references to Lemonade Ultra should still be in server_models.json with `suggested: false`. The new models `lemonade-sdk/LMN-Halo-Omni-xxB` (if that's what we're calling them) should be `suggested: true`.

jeremyfowers:
Wound up doing this work iteratively over the course of many PRs.

Issue #11, opened 2025-05-26
Request to you for benchmark report


ZhangWei125521:
I think the Lemonade SDK is really good designed to make it easy to run LLMs with CPU, GPU, and NPU.
Now I am selecting which architecture for our internal apps，so I wonder if you have some benchmark datas about the LLMs.
such as how about the Time to first token and Token generation speed  between the OGA and Llama.cpp in the same hardware.
thank you very much

jeremyfowers:
Hi @ZhangWei125521 thanks for writing the issue!  Can you provide a bit more information please: 
- Do you have the hardware available already? If not, which hardware are you interested in?
- Which specific models are you interested in?
- Are there some specific benchmarking parameters you want, such as input sequence length, output sequence length, etc.?

ZhangWei125521:
> Hi [@ZhangWei125521](https://github.com/ZhangWei125521) thanks for writing the issue! Can you provide a bit more information please:
> 
> * Do you have the hardware available already? If not, which hardware are you interested in?
> * Which specific models are you interested in?
> * Are there some specific benchmarking parameters you want, such as input sequence length, output sequence length, etc.?

Hi @jeremyfowers thank you very much for your promptly reply.
 * Do you have the hardware available already? If not, which hardware are you interested in?
    _I have have no AMD devices on hand currently. Actually, we can face different AMD hardware in my project. So maybe I concern about the latest iGPU data._
 *  Which specific models are you interested in?
     _Llama3 8B int4_
 *  Are there some specific benchmarking parameters you want, such as input sequence length, output sequence length, etc.?
     _3k input and 3k output，batch size 1_
That's all, ​​I await your prompt response，thank you!

jeremyfowers:
@ZhangWei125521 thanks for the details. Unfortunately, LLaMA 3 8B doesn't yet have support for 3k input length on Hybrid execution and is limited to 2k right now (the team is working to improve this).

You can find benchmarking data for LLaMA 8B-class models in the Performance section of this blog: https://www.amd.com/en/developer/resources/technical-articles/deepseek-distilled-models-on-ryzen-ai-processors.html. Looks like TTFT of 5 seconds and 17.6 tokens/second, on Ryzen AI STX.

If you are interested in iGPU-only, I would suggest using llama.cpp+Vulkan instead of OGA+Hybrid. I'm currently working to integrate that fully into Lemonade.

ZhangWei125521:
[...]

PR #1785, opened 2026-05-02
Cloud offload: provider-agnostic OpenAI-compat backend + UI

ramkrishna2910:
## Summary

Adds **cloud offload**: a provider-agnostic backend that lets `lemond` serve chat/completions from any OpenAI-compatible cloud provider (Fireworks, OpenRouter, Together, OpenAI, …) alongside local models. Provider URLs live in `lemond`'s `config.json`; API keys live in env vars or `lemond`'s process memory — **never on disk**.

## Architecture

```mermaid
flowchart LR
    Client["Client (GUI / CLI / SDK / coding agent)"] -->|"chat, models, system-info"| L["lemond"]
    Admin["Admin"] -->|"LEMONADE_&lt;P&gt;_API_KEY"| Env(("env vars"))
    Admin -->|"or POST /v1/cloud/auth"| L

    subgraph LE["lemond"]
        Reg["CloudProviderRegistry<br/>(installed[]: {provider, base_url} persisted<br/>runtime_keys: in-memory only)"]
        MM["ModelManager<br/>refresh_cloud_models per provider"]
        Router["Router"]
        CS["CloudServer<br/>(resolves creds from registry)"]
        Reg --> MM
        MM --> Router
        Router --> CS
    end

    Env -.->|"resolve_key()"| Reg
    CS -->|"HTTPS Bearer"| Up["Fireworks / OpenRouter / Together / OpenAI / …"]
```

## Design decisions

1. **Server-side provider config.** Each cloud provider is registered once via `POST /v1/install {backend:"cloud", provider, base_url}`. The registry persists `{provider, base_url}` to `config.json` under `cloud_providers`. The same provider list is visible to every client (GUI, CLI, third-party agents) and to `lemonade launch`-ed coding agents.

2. **API keys never touch disk.** Two resolution sources, in priority order:
   - `LEMONADE_<PROVIDER>_API_KEY` env var — persistent, operator-provisioned, ideal for Docker / systemd / shared installs.
   - `POST /v1/cloud/auth {provider, api_key}` — ephemeral, lives in `lemond`'s process memory, dies on restart.

   When the env var is set, the runtime POST returns **409 conflict** (env wins) so the operator's "house key" can't be silently overridden.

3. **Auto-discovery on cache build.** `ModelManager::build_cache()` loops every installed provider with a resolvable key and calls `CloudServer::discover_models`. `POST /v1/cloud/auth` and `POST /v1/install` both trigger a focused refresh for the affected provider. Errors are logged and never thrown — one offline provider can't sink cache build.

4. **`/v1/system-info` surfaces auth state.** New `cloud.providers[]` block reports `{name, base_url, env_var, env_var_set, runtime_key_set, models_discovered}` per provider. The key value itself is never returned.

5. **`CloudServer : WrappedServer`, no subprocess.** `load()`/`unload()` are near-no-ops (record the upstream model id, flip a bool). `chat_completion()`, `completion()`, and `forward_streaming_request()` resolve creds via the registry and talk HTTPS Bearer to the upstream.

6. **Provider-agnostic.** Selection is `recipe="cloud"` + the per-model `cloud_provider` field on `ModelInfo`. No per-provider code paths.

7. **Dot-namespaced names.** Public model name is `<provider>.<cleaned_upstream_id>` (e.g. `fireworks.kimi-k2p5`), matching the `user.` / `extra.` namespacing used elsewhere and avoiding collisions between providers exposing the same model.

8. **Chat-only by design.** `is_chat_model()` filters discovery to chat models, trusting provider capability metadata when present (Fireworks `supports_chat`, generic `capabilities`, OpenRouter `architecture.modality`, Together `type`, OpenAI id pattern fallback). Embeddings/audio/reranking/image are intentionally not served — they'd need sibling backends because providers diverge from OpenAI v1 on non-chat shapes.

9. **Capability labels normalized.** Each provider's divergent metadata is mapped into lemonade's shared vocabulary (`vision` / `tool-calling` / `reasoning`) so cloud models gate inputs exactly like local ones (the UI offers image upload iff `vision` is present).

10. **Context window + cost surfaced (display only).** Discovery captures `context_length` and normalizes pricing to USD per 1M tokens (OpenRouter per-token ×1e6; Together per-1M; Fireworks publishes none → unknown). Reported in `/models`, `/system-info`, and `/health`; never affects routing.

11. **LRU exemption.** Cloud servers are skipped in `count_servers_by_type` / `find_lru_server_by_type` — they consume no local memory.

## Endpoints (all under `/api/v0/`, `/api/v1/`, `/v0/`, `/v1/`)

| Endpoint                          | Body / params                                  | Notes                                           |
|-----------------------------------|------------------------------------------------|-------------------------------------------------|
| `POST /install`                   | `{backend:"cloud", provider, base_url, api_key?}` | Registers provider; optional one-shot auth     |
| `POST /uninstall`                 | `{backend:"cloud", provider}`                  | Removes registry record + evicts models         |
| `POST /cloud/auth`                | `{provider, api_key}`                          | 409 if env var set; refreshes on success        |
| `DELETE /cloud/auth/{provider}`   | —                                              | Clears runtime key; env var unaffected          |
| `GET /system-info`                | —                                              | Adds `cloud.providers[]` block                  |

## CLI

```
lemonade cloud install fireworks --base-url https://api.fireworks.ai/inference/v1 [--api-key …]
lemonade cloud uninstall fireworks
lemonade cloud auth fireworks [--api-key …]   # TTY prompt if --api-key omitted
lemonade cloud clear fireworks
lemonade cloud list
```

`lemonade list` already shows cloud models because they appear in `/v1/models` like any other recipe.

## What Jeremy flagged, and how the redesign addresses it

| Concern                                                                  | Resolution                                                                                                       |
|---------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------|
| 1. Logic pushed into the GUI we're about to replace                       | Provider config is server-side; the new GUI inherits it for free                                                  |
| 2. Cloud models not visible to the CLI                                    | `lemonade list` shows them; `lemonade cloud {install,auth,list,…}` manages them                                  |
| 3. Cloud models not available to `lemonade launch`-ed agents              | They appear in `/v1/models` like any other recipe — no per-app workflow                                          |
| 4. SDK friction (apps had to repeat discovery)                            | Set `LEMONADE_<PROVIDER>_API_KEY` and any wrapper just works                                                      |
| 5. Admin couldn't share cloud access with their userbase                  | Admin sets one env var, every client sees the catalog                                                             |

## Files changed

- **New:** `cloud_provider_registry.{h,cpp}`
- **C++ server:** `CloudServer` (drop per-request creds), `ModelManager::refresh_cloud_models`, `/v1/cloud/auth` handlers, `/v1/install`+`/v1/uninstall` cloud branches, `/v1/system-info` cloud section
- **CLI:** `cloud {install,uninstall,auth,clear,list}` subcommands
- **UI:** Replaced `CloudProvidersSection.tsx` with a server-driven version; deleted `CloudProviderModal.tsx`; stripped `cloudProviders` from `appSettings.ts`, header injection from `serverConfig.ts`, client-side cloud merge from `modelData.ts`
- **Tests:** Three new end-to-end cases in `test/server_endpoints.py` using a local mock OpenAI-compat provider (full round-trip; unknown-provider 404; evicted-cloud-model clean error)
- **AGENTS.md:** Updated invariant #11 to distinguish per-client UI state from shared infrastructure config and from cloud keys

## Test results

- `python test/server_endpoints.py` — 53/53 ok (425 s)
- `python test/server_cli2.py` — 58/58 ok, 16 skipped (62 s)


ramkrishna2910:
@claude Review

jeremyfowers (changes requested):
I have big questions about whether the authentication architecture is the right one.

Making the credentials per-client...
1.  pushes a lot of logic into the GUI app. Is this a good idea when we are about to replace the GUI app?
2. means that cloud backends and models are not visible to the CLI, are we disenfranchising non-GUI users?
3. means that cloud models are not available to connected 3rd party apps, even those launched with `lemonade launch`. Do we really mean to exclude coding agents entirely?
4. adds a high amount of friction into the SDK since an app that wraps Lemonade would need to do the cloud discover workflow.
5. precludes a Lemonade admin from conveniently sharing cloud model access with their userbase.

I'm not sure who this is for, aside from single-user Lemonade setups that have most/all of their usage in the Lemonade GUI App.

This choice deserves a robust discussion. I expect users and builders will want their cloud models managed at the server level, manageable via the CLI, visible to 3rd party apps, and, especially, available in coding apps.

src/cpp/server/server.cpp
```diff
@@ -4903,6 +5118,65 @@ void Server::handle_install(const httplib::Request& req, httplib::Response& res)
...
+        // discover its catalog as soon as the matching API key is supplied
+        // (env var or POST /v1/cloud/auth). Shape:
+        //   {backend: "cloud", provider: "fireworks",
+        //    base_url: "https://api.fireworks.ai/inference/v1",
+        //    api_key: "..."}  // optional
+        if (request_json.value("backend", "") == "cloud") {
+            const std::string provider = request_json.value("provider", "");
+            const std::string base_url = request_json.value("base_url", "");
+            const std::string api_key = request_json.value("api_key", "");
+            if (provider.empty() || base_url.empty()) {
```
jeremyfowers, line 4823: Provider names need stricter validation than non-empty. `env_var_name()` uppercases the name, so `Fireworks` and `fireworks` are distinct registry records that resolve the *same* env var, and a name with spaces or `/` produces an unresolvable env-var name plus odd model ids/URLs. The UI comments claim provider names are "constrained to lowercase", but nothing enforces it. Suggest rejecting anything outside `[a-z0-9_-]+` with a 400 here (or lowercasing on install).

src/cpp/server/server.cpp
```diff
@@ -4903,6 +5118,65 @@ void Server::handle_install(const httplib::Request& req, httplib::Response& res)
...
+                res.status = 400;
+                nlohmann::json error = {{"error", {
+                    {"message", "Cloud install requires 'provider' and 'base_url' string fields"},
+                    {"type", "invalid_request_error"}}}};
+                res.set_content(error.dump(), "application/json");
+                return;
+            }
+            LOG(INFO, "Server") << "Installing cloud provider '" << provider
+                                  << "' with base_url " << base_url << std::endl;
+            cloud_registry_->install(provider, base_url);
```
jeremyfowers, line 4849: **Security:** an `http://` base URL is accepted silently, which sends the Bearer API key in plaintext on every forwarded request. A typo'd scheme leaks the user's cloud key to the network. Suggest rejecting non-localhost `http://` URLs (or at minimum logging a prominent warning and flagging it in the install response). Localhost should stay allowed for the mock-provider tests.

src/cpp/server/model_manager.cpp
```diff
@@ -2399,6 +2458,102 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
...
+
+    // Resolve creds outside the cache lock — resolve_key takes a shared
+    // lock on the registry's mutex internally; holding the cache lock here
+    // doesn't matter but keeping it tight is cheaper.
+    const std::string api_key = cloud_registry_->resolve_key(provider);
+    const std::string base_url = cloud_registry_->base_url_for(provider);
+    if (api_key.empty() || base_url.empty()) {
+        // Drop any stale entries for this provider but don't try to discover —
+        // there's nothing to discover with.
+        return evict_cloud_models(provider);
```
jeremyfowers: This returns the number of models *evicted*, but the function contract (and `handle_install`, which reports this value as `models_discovered`) expects the number of models *present after refresh*. Any refresh-without-key on a populated cache reports `models_discovered: N` when the truth is 0. Should be `evict_cloud_models(provider); return 0;` — same issue on the catch path below.

src/cpp/server/model_manager.cpp
```diff
@@ -2346,6 +2396,15 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
             continue;
         }
 
+        // Cloud-offloaded models bypass local backend/RAM checks (the model
+        // executes on a remote provider). Discovery is client-driven now —
```
jeremyfowers: Stale comment: "Discovery is client-driven now — server-side cache normally has none of these" describes the opposite of this PR's design. Discovery is server-side and the cache normally holds *all* discovered cloud models. Worth rewording before it misleads the next reader.

src/cpp/server/system_info.cpp
```diff
@@ -1644,6 +1644,12 @@ SystemInfo::SupportedBackendsResult SystemInfo::get_supported_backends(const std
 }
 
 std::string SystemInfo::check_recipe_supported(const std::string& recipe) {
+    // Cloud offload has no local hardware/OS requirements; availability is
+    // gated by config.json (cloud_offload.enabled) and a provider API key,
```
jeremyfowers: Stale comment: `cloud_offload.enabled` doesn't exist in config.json — availability is gated by the provider registry and key resolution. Update to reference `cloud_providers` / `CloudProviderRegistry`.

src/cpp/server/server.cpp
```diff
@@ -1978,6 +2034,8 @@ void Server::handle_completions(const httplib::Request& req, httplib::Response&
     try {
         auto request_json = nlohmann::json::parse(req.body);
 
+        bool request_modified = false;
```
jeremyfowers: `request_modified` is declared but never set or read in `handle_completions` — unused-variable warning territory. Either wire it up to the intended transformation or drop it.

src/cpp/server/server.cpp
```diff
@@ -3107,6 +3165,8 @@ void Server::handle_responses(const httplib::Request& req, httplib::Response& re
...
```
jeremyfowers: `request_modified` is never set to true in `handle_responses`, so the `request_body = request_modified ? request_json.dump() : req.body` below always takes the `req.body` branch. Dead scaffolding that implies a transformation that never happens — wire it up or remove both.

src/cpp/server/model_manager.cpp
```diff
@@ -2399,6 +2458,102 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
...
+    }
+
+    size_t added = 0;
+    for (const auto& m : models) {
+        if (m.recipe != "cloud" || m.model_name.empty()) continue;
+        ModelInfo info = m;
+        // discover_models() populates name/checkpoint/labels/context/cost but
+        // not recipe_options; Router needs it to construct CloudServer.
+        info.recipe_options = RecipeOptions("cloud", json::object());
+        models_cache_[info.model_name] = std::move(info);
```
jeremyfowers: Name-collision precedence is asymmetric with `build_cache()`: there, discovered cloud models merge with `emplace` (existing entry wins), but here `models_cache_[name] = ...` lets the cloud model overwrite. A user model named e.g. `fireworks.foo` survives startup but is silently replaced by the first `/cloud/auth` refresh. Suggest skipping insertion when a non-cloud entry already holds the name, matching `build_cache()`.

test/server_endpoints.py
```diff
@@ -680,6 +680,291 @@ def test_012c_load_noop_when_already_loaded_by_inference(self):
...
+            f"{self.base_url}/cloud/auth",
+            json={"provider": "never-installed", "api_key": "k"},
+            timeout=TIMEOUT_DEFAULT,
+        )
+        self.assertEqual(resp.status_code, 404, resp.text)
+        body = resp.json()
+        self.assertEqual(body["error"]["type"], "invalid_request_error")
+        print("[OK] /cloud/auth on unknown provider returns 404")
+
+    def test_012f_chat_without_credentials_returns_clean_error(self):
```
jeremyfowers: This test no longer covers what its name/docstring claims: it asserts an evicted-model 404, not the missing-credentials error. The actual `missing_creds_error()` path (model loaded, key then unresolvable) is uncovered — and so is streaming through a cloud model. Suggest renaming this test to match what it asserts and adding a real no-creds case; the meandering comments below ("Manually evict ... Instead: just ...") read as leftover drafting and should be cleaned up either way.

src/cpp/cli/main.cpp
```diff
@@ -1277,6 +1308,49 @@ int main(int argc, char* argv[]) {
...
+#else
+                bool is_tty = isatty(fileno(stdin)) != 0;
+#endif
+                if (!is_tty) {
+                    std::cerr << "Error: --api-key is required when stdin is not a TTY"
+                              << std::endl;
+                    return 1;
+                }
+                std::cout << "API key for " << config.cloud_provider << ": ";
+                std::getline(std::cin, key);
```
jeremyfowers: The API key is echoed to the terminal as the user types. Conventional secret prompts disable echo (termios `ECHO` off on POSIX, `SetConsoleMode` without `ENABLE_ECHO_INPUT` on Windows) — cheap to add and avoids keys lingering in scrollback/screen-share.

jeremyfowers (commented):
Inline comments for the issues found during review: provider-name validation, plaintext-key risk on http:// base URLs, a wrong `models_discovered` count, two stale comments, dead `request_modified` scaffolding, asymmetric name-collision precedence, a test that doesn't cover what its name claims, and the echoing TTY key prompt.

docs/api/lemonade.md
```diff
@@ -19,8 +19,10 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/health`](#get-v1health) | Check server status, such as models loaded |
```
jeremyfowers, line 19: After I added a provider in the GUI I had to manually refresh my web ui tab to get it to show up in the models list. Adding a provider needs to trigger the models refresh helper function.

jeremyfowers (approved):
Works for me! Please see the comments and address as you see fit before merging.

src/cpp/server/server.cpp
ramkrishna2910: Good catch. Added a static `CloudProviderRegistry::validate_provider_name()` that enforces `^[a-z0-9_-]+$` (and rejects > 64 chars). The cloud branch of `handle_install` runs it before `install()` and returns 400 on a bad name. Lowercase-only is enforced specifically because `env_var_name()` uppercases — `Fireworks` and `fireworks` collapsing to the same env var was exactly the registry-level confusion you flagged.

New test `test_012g_install_rejects_bad_provider_name` covers `Fireworks`, `with space`, `vendor/x`, `with.dot`, and empty.

src/cpp/server/server.cpp
ramkrishna2910: Agreed — keys in cleartext on a typo is a real failure mode. Added `CloudProviderRegistry::validate_base_url()` which:

- accepts `https://` to any host
- accepts `http://` only to `localhost` / `127.0.0.1` / `[::1]` (so the mock-provider tests still work)
- rejects everything else with 400 and a message that names the offending host and the plaintext-key reason

Test `test_012h_install_rejects_insecure_http_base_url` covers the rejection path (`http://api.example.com`, `gopher://`) and the localhost allow.

src/cpp/server/model_manager.cpp
ramkrishna2910: Right, contract violation. Both the no-key path and the catch path now `evict_cloud_models(provider); return 0;` so `handle_install` reports the count actually present after refresh.

src/cpp/server/model_manager.cpp
ramkrishna2910: Reworded to describe the current design — server-side discovery runs every cache build, so this branch normally sees the full discovered set plus any user pins.

src/cpp/server/system_info.cpp
ramkrishna2910: Updated. Now reads: gated by the `CloudProviderRegistry` (config.json `cloud_providers`) and a resolvable API key (env var or runtime auth), checked elsewhere in `filter_models_by_backend` / `CloudServer::load`.

src/cpp/server/server.cpp
ramkrishna2910: Dropped — there was never anything reading or assigning it, and `request_body = request_json.dump()` is unconditional. Body downstream is unchanged.

src/cpp/server/server.cpp
ramkrishna2910: Same treatment as `handle_completions` — variable dropped and the ternary collapsed to `request_body = req.body` (which was the only branch ever taken).

src/cpp/server/model_manager.cpp
ramkrishna2910: Right — went one step further than your suggestion and matched `build_cache()` exactly: `emplace` instead of `[]=`. Any pre-existing entry under the same bare key wins, regardless of source (non-cloud user/builtin/extra/FLM, or another cloud provider that already discovered first). On a hit the loop logs `name already held (recipe=…)` and skips.

Hard to construct a non-cloud-bare-key collision from a Python test without restarting lemond, so I added `test_012i_cloud_refresh_is_idempotent_no_duplicates` — POSTs `/cloud/auth` twice with the same key and asserts `models_discovered` stays at the same N and each id appears exactly once in `/v1/models`. The eviction-then-emplace cycle must work for that to hold.

test/server_endpoints.py
ramkrishna2910: Fair. Renamed the existing test to `test_012f_chat_against_evicted_cloud_model_returns_404` (which is what it actually asserts) and dropped the leftover "Manually evict … Instead: just …" drafting comments.

Then added the two cases that were missing:

- `test_012j_chat_with_loaded_model_but_cleared_key_returns_missing_creds` — exercises the real `missing_creds_error()` path. Loads the model so the router holds an active `CloudServer` instance, then DELETE `/cloud/auth/{provider}`. The chat call reuses the loaded server, bypasses model-not-found, and hits `resolve_creds()` at request time. Asserts the response is a structured error envelope (not 200, not an HTML stack-trace) with `error.type`, `error.message` referencing credentials, and the offending provider name.

- `test_012k_streaming_chat_through_cloud_provider` — extended the mock to optionally serve SSE `data:` chunks. Test sends `stream: true`, parses each `data:` line back into a delta, and asserts they assemble to `Hello` followed by `[DONE]`. Covers the SSE forwarding path through `CloudServer::forward_streaming_request`.

src/cpp/cli/main.cpp
ramkrishna2910: Added a `read_secret_line()` helper that disables echo around `std::getline`:

- Windows: `GetConsoleMode` / `SetConsoleMode` clearing `ENABLE_ECHO_INPUT`
- POSIX: `tcgetattr` / `tcsetattr` clearing `ECHO`

Restores the prior mode in both cases (so the user's terminal isn't left silent if they Ctrl-C). Prompt is now `API key for <provider> (input hidden):` and the line we read is never echoed to the terminal, so it doesn't sit in scrollback / show up in a screen-share.

docs/api/lemonade.md
ramkrishna2910: Fixed in `CloudProvidersSection.tsx`. The component now pulls `refresh` from `useModels()` and wraps it together with the local provider-list reload into a single `reloadAll()` helper. Both the install modal's `onInstalled` and the edit modal's `onChanged` invoke it, so after Add/Edit/Uninstall the global models store is re-fetched and the Models tab updates without a manual reload.

Issue #1767, opened 2026-04-29
Brainstorm - Scaling Contributors and Maintainers


jeremyfowers:
1. Need a clear long-term roadmap to determine if potential contributions are aligned.
2. Maintain a list of Good First Issues, and Good Contribution issues that are aligned to the roadmap.
3. Trusted group of reviewers who are committed to thoroughly reviewing and merging contributions.
    - Write down the focus areas of the existing reviewers. Who is the go-to person for certain things. "Maintainer groups".
4. Recurring "maintainers" or office hours meeting where people have a chance to learn what's going on and get to weigh in.
    - Biweekly 30 minutes to start with. Optional for everybody.
    - Launch a poll to figure out a time.
5. Fix the GUI design bottleneck    - Add automations to UI/UX review - drive workflows, capture screenshots, and post on PR reviews.
    - DESIGN.md file specific to the app that AI reviewer can use to validate style guides.

PR #1786, opened 2026-05-02
test(app): add regression guards for model options and collections

fl0rianr:
## Summary

Adds lightweight app regression tests under `test/app` to catch UI and model-routing regressions before they reach review.

The tests are intentionally runnable with plain Node, without a renderer build, Tauri startup, `npm install`, Jest, or jsdom. This makes them suitable for a fast pre-commit or local smoke check.

Covered areas:

- Model Options modal wiring and loading-state regressions
- Model Manager row actions for load, delete, and options
- Collection model helper behavior
- OmniRouter tool mapping and endpoint routing
- Optional custom collection behavior when the feature branch files are present

## Why

A regression in the custom collection changes caused normal model settings to stop opening. These tests add targeted guards so similar issues are caught locally before review.

There a obvious pros and cons with it, I assumed it's the right direction in the lights of #1767 since the add the possibility of pre-commit checks.

The suite is (currently) branch-aware:

- On current `main`, feature-only custom collection tests are skipped.
- On the **custom collection branch** (#1769) , those tests become active.

## Test Plan

Run from the repository root:

```bash
node test/app/run-app-regression-tests.cjs
```

Observed results:

```text
current main:
App regression tests: 26 passed, 11 skipped, 0 failed.

custom collection branch before Model Options fix:
App regression tests: 35 passed, 0 skipped, 2 failed.
ModelOptionsModal regression guards fail as expected.

custom collection branch with fix:
App regression tests: 37 passed, 0 skipped, 0 failed.
```

No app rebuild is required for the test run.

### DIFF
--- a/.github/workflows/docs_and_style.yml
+++ b/.github/workflows/docs_and_style.yml
@@ -9,6 +9,21 @@ permissions:
   contents: read
 
 jobs:
+
+  app-regression-tests:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-app-regression-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - uses: actions/checkout@v5
+      - name: Setup Node.js
+        uses: actions/setup-node@v5
+        with:
+          node-version: '20'
+      - name: Run app regression tests
+        run: node test/app/run-app-regression-tests.cjs
+
   markdown-link-check:
     runs-on: ubuntu-latest
     concurrency:
--- a/test/app/app-regression/README.md
+++ b/test/app/app-regression/README.md
@@ -1,0 +1,16 @@
+# App regression tests
+
+Run from the repository root:
+
+```bash
+node test/app/run-app-regression-tests.cjs
+```
+
+These tests are intentionally dependency-free: no Jest, no jsdom, no React
+runtime, and no TypeScript package are required. They use source-level guards
+for renderer regressions that are easy to introduce while changing Model
+Manager, Model Options, and OmniRouter collection logic.
+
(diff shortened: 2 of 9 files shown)


jeremyfowers (changes requested):
Thanks for making this PR @fl0rianr! I really like the idea of having some CI tests for the app code.

I did an analysis of the test coverage added and would like to request some changes:
1. The test needs to be added to automated CI
2. Please save the 11 dormant tests for a follow-on PR after the collections work merges
3. About 20 of the tests are syntactic restatements of the code under test. Are you sure these are useful?

fl0rianr:
Thanks @jeremyfowers , that makes sense. I’ll update this PR to:

1. Add the Node app regression runner to GitHub Actions so it runs in CI.
2. Remove the dormant custom-collection tests from this PR and save them for a follow-up after the collections PR merges. Even if I did not see the issue since I expected this to merge afterwards AND it's optional.
3. Trim the source-shape assertions and keep the tests focused on regressions that would have caught the Model Options / collection interaction issue.

I’ll rebase on current main first because the branch is now behind and there were workflow/app changes. I hope to find some time for it soon.

fl0rianr:
@jeremyfowers Thanks again! I updated the PR to add the app regression runner to CI via the existing Docs And Style workflow.

I also trimmed and consolidated the source-level guards so they focus more on regression invariants instead of exact implementation shape. The tests remain dependency-free and can still be run locally with plain Node.

I kept the optional custom-collection checks in this PR after clarifying directly with you: they skip cleanly on current main and only become active once the collection files exist, so they should not block main while still protecting the follow-up collection work.

jeremyfowers (approved):
Let's go!

PR #1788, opened 2026-05-03
feat: persistent background download queue

a-huk:
**Background download queue**

Right now if you close the browser tab while a model is downloading, the download just dies. This PR fixes that — downloads now run in the background inside `lemond` and keep going regardless of what the client does.

`POST /pull` returns immediately with a job ID instead of blocking, and there are a few new endpoints to track what's happening:

- `GET /downloads` — see all jobs
- `GET /downloads/{id}` — check a specific one
- `DELETE /downloads/{id}` — cancel it
- `GET /downloads/{id}/stream` — SSE stream for live progress (reconnect-safe, replays recent events)

The queue is also saved to disk, so if the server restarts mid-download it picks up where it left off using the existing `.partial` file resume support.

**Breaking change:** `POST /pull` now returns `202` with a `job_id` instead of either blocking synchronously or streaming SSE directly. If you're using `stream=true` today, switch to `GET /downloads/{id}/stream` after the pull. The CLI (`lemonade pull`) will need a follow-up update too.

Tests are updated to poll for completion via `GET /downloads/{id}` rather than waiting on the pull response.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -561,6 +561,7 @@ set(SOURCES_CORE
     src/cpp/server/cli_parser.cpp
     src/cpp/server/config_file.cpp
     src/cpp/server/model_manager.cpp
+    src/cpp/server/download_queue.cpp
     src/cpp/server/hf_variants.cpp
     src/cpp/server/wrapped_server.cpp
     src/cpp/server/streaming_proxy.cpp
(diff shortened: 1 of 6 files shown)


jeremyfowers:
Sorry about being slow to reply here. I’ve seen plenty of complaints about downloads stopping unexpectedly so this idea is welcome. We can’t do a breaking change on pull though because downstream software is using it. Any way to implement this idea without the breaking change?

jeremyfowers:
Closing in favor of #1948 which does not have a breaking change.

PR #1797, opened 2026-05-05
feat(setup.sh): add rpm-ostree support for Fedora Atomic/Desktop environments

storm1er:
Summary

- Add rpm-ostree package manager detection and support to setup.sh for Fedora Atomic/Desktop environments (Silverblue, Kinoite, Aurora, etc.)
- Covers all dependency installation paths: build tools, libraries, tray dependencies, Tauri dependencies, and Rust toolchain
- Uses rpm-ostree install -A for package installation, consistent with existing DNF/Fedora packaging conventions

Changes

- Detect rpm-ostree command availability alongside existing package managers (apt, pacman, dnf, zypper)
- Add rpm-ostree-specific package lists for: pre-commit, build essentials, pkg-config libraries, tray (libayatana-appindicator, dbusmenu, gtk3), Tauri (webkit2gtk4.1, libsoup3, librsvg2, etc.), and Rust/Cargo
- Include rpm-ostree in Node.js detection guard and all install command branches

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -72,6 +72,8 @@ if ! command_exists pre-commit; then
         missing_packages+=("pre-commit")
     elif [ "$OS" = "linux" ] && command_exists apt; then
         missing_packages+=("pre-commit")
+    elif [ "$OS" = "linux" ] && command_exists rpm-ostree; then
+        missing_packages+=("pre-commit")
     elif [ "$OS" = "linux" ] && command_exists dnf; then
         missing_packages+=("pre-commit")
     elif [ "$OS" = "linux" ] && command_exists zypper; then
@@ -109,6 +111,8 @@ if [ ${#missing_tools[@]} -gt 0 ]; then
             missing_packages+=("git" "cmake" "ninja-build" "build-essential")
         elif command_exists pacman; then
             missing_packages+=("git" "cmake" "ninja" "base-devel")
+        elif command_exists rpm-ostree; then
+            missing_packages+=("git" "cmake" "ninja-build" "gcc" "gcc-c++" "make")
         elif command_exists dnf; then
             missing_packages+=("git" "cmake" "ninja-build" "gcc" "gcc-c++" "make")
         elif command_exists zypper; then
@@ -166,6 +170,19 @@ if command_exists pkg-config; then
                         libwebsockets) ;; # Not available in Arch repos, will use FetchContent
                     esac
                 done
+            elif command_exists rpm-ostree; then
+                # Map pkg-config names to rpm-ostree packages
+                for lib in "${missing_libs[@]}"; do
+                    case "$lib" in
+                        libcurl) missing_packages+=("libcurl-devel") ;;
+                        openssl) missing_packages+=("openssl-devel") ;;
+                        zlib) missing_packages+=("zlib-devel") ;;
+                        libsystemd) missing_packages+=("systemd-devel") ;;
+                        libdrm) missing_packages+=("libdrm-devel") ;;
+                        libcap) missing_packages+=("libcap-devel") ;;
+                        libwebsockets) missing_packages+=("libwebsockets-devel") ;;
+                    esac
+                done
             elif command_exists dnf; then
                 # Map pkg-config names to dnf packages
                 for lib in "${missing_libs[@]}"; do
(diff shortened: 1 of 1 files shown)


superm1 (commented):
I guess I would question - do people usually want to slipstream in the extra packages?  Or would it more logical to guide them to toolbx?

storm1er:
@superm1 I don't think so: why atomic OSes would be different than other options in setup.sh?

But I'm not at ease with the concept of toolbox either so I may miss something: does your suggestions means removing the actual usage of setup script to redirect everyone on a distrobox setup?

superm1:
Here is a small summary generated by a search engine when I searched `rpm-ostree install vs toolbox`.  It tries to guide development environments towards toolbox/distrobox.  That's why I found the rpm-os-tree approach odd for development.  Do you still feel the same way?

---

When running Fedora Kinoite (KDE Plasma on an immutable OSTree base), you have two main ways to add software: rpm-ostree layering and Toolbox/Distrobox containers. Both approaches work, but they serve different purposes.

rpm-ostree is used to modify the host system image. It’s ideal for:

Switching desktop environments transactionally using rpm-ostree rebase (e.g., from Silverblue GNOME to Kinoite KDE) without reinstalling.

Installing packages that need deep system integration, such as kernel modules, hardware drivers, or systemd services.

Ensuring changes are immutable and rollback-capable — if something breaks, rpm-ostree rollback restores the previous state.

Example: Switching from Silverblue to Kinoite KDE

sudo ostree admin pin 0
rpm-ostree rebase fedora:fedora/38/x86_64/kinoite
reboot
Copy
Toolbox/Distrobox creates mutable container environments on top of the immutable host. They are best for:

Development environments where you need frequent package changes without touching the host.

Running CLI tools, compilers, or GUI apps that don’t require kernel-level access.

Keeping the host clean — all extra packages live inside the container.

Exporting GUI apps to the host (distrobox-export --app <pkg>), though systemd services won’t run inside Toolbox.

storm1er:
So finally the decision should be based on ease of development and accessibility. I'm not a maintainer of this project so my response are just wishes.

- Does small developers/contributors will be at ease with local VMs/container?
_I'm not, especially given that I'm doing this on my personal machine, not dedicated to development_
- Will it impact current development workspaces?
_I think not, unless you want to migrate everyone to distrobox/VM_
- Does handling distrobox/vm will be feasible using setup scripts?
_Yes, obviously, but then we go back to my previous message:_
- Why the same script should handle a very completely different process/template for machine unable to use the existing stuff when rpm-ostree exists?
_I think this should rely on the willingness of the developers/contributors as "if X want to use a distrobox/VM, then X should use the existing setup scripts in a distrobox/VM"_

We then end up with a basic reflection: Does adding rpm-ostree option impact negatively this script? Making it's maintainability harder?  
_Since it follows the same template as other installation method (like dnf, apt, ...) I would say no?_

You decide, I'll follow

superm1:
How about we meet in the middle - just show a disclaimer along the lines that installing using rpm-ostree will change the atomic OS image and can affect future upgrades, do you want to proceed?

Then do your PR as is otherwise.

storm1er:
> _show a disclaimer along the lines that installing using rpm-ostree will change the atomic OS image and can affect future upgrades_

Understandable, I'll do this, not at home right now.
I'll update the PR as soon as I can

jeremyfowers:
Love to see the good discussion here! Hope it will merge soon. @storm1er lmk if you want to do any other fedora related work, there’s things people have been asking for like fixing fedora 43 support in the RPM (it broke when 44 came out).

storm1er:
@jeremyfowers Are you talking about libdrm-devel issues? I though it was only on bazzite distro?

storm1er:
@superm1 

Please validate 3fc73fe about CI & gh action. I'm not aware of the why and how about these in this script. Works and ok on my machine, but my machine is not CI nor gh action :p

setup.sh
```diff
@@ -60,6 +60,40 @@ print_info "Lemonade Development Setup"
 print_info "Operating System: $OS"
 echo ""
 
+if [ "$OS" = "linux" ] && command_exists rpm-ostree; then
+    print_warning "============================================"
+    print_warning " rpm-ostree detected (Fedora Atomic/Desktop)"
```
superm1, line 70: Is there any other way to detect this? Because I'm pretty sure you can install this command on Ubuntu or regular Fedora too.

storm1er: _other way to detect this_: probably

_you can install this command on ubuntu or regular fedora_: 
seems exclusive to redhat based OSes https://coreos.github.io/rpm-ostree/#getting-started   
while debian based can use deb-ostree https://github.com/dbnicholson/deb-ostree-builder

Unless I miss something?

storm1er: Also, please note that you can't use dnf on rpm-ostree's enabled system as dnf error like this:

```
$ dnf install ...
ERROR: Fedora Atomic images utilize rpm-ostree instead (and is discouraged to use).
```

superm1: Try using `rpm-ostree status` to confirm it after you've detected `rpm-ostree` in `PATH`. From my searching that should be enough to decide.

storm1er: Done, i'm closing this so you can do your review again =)

superm1 (approved):

PR #1800, opened 2026-05-05
Replace model label dots with icons

jeremyfowers:
## Summary

- Replace the small colored model label dots with colored modality icons.
- Preserve native title tooltips and use a canonical display order for label icons.
- Fixed pre-existing bug: Align family header chevrons and icons to the same rails used by regular model rows while keeping the original chevron size.

image

Idea and SVGs ported from https://github.com/lemonade-sdk/lemonade/pull/1657 by @anditherobot


danielholanda (approved):
I really like the icons here. That said, this will start looking like a "power user" tool if we start adding too many visual elements.

I think we should approve this and also consider adding a "Power user" toggle at the bottom at a future version. Those icons (as well as many other options) should only show up in the power user mode.

TLDR: LGTM, but please wait for @sofiageo 's review

jeremyfowers:
> TLDR: LGTM, but please wait for @sofiageo 's review

Since he's traveling and the release is today I'm going to merge. But I am happy to make any modifications suggested by @sofiageo in the future!

sofiageo:
Looks good to me. I never learned what the dot colors were, but the icons are easier to remember. The custom icon has some contrast issues with the light theme. I will note to check it in the next days (still jet lagged to do anything)

PR #1801, opened 2026-05-05
Dockerfile: libxrt2 & libxrt-npu2 dependencies for fastflowlm

gunnarx:
Docker image build was failing. 
Maybe FLM_VERSION has changed recently and/or fastflowlm.deb changed its dependencies.

Whatever the reason, installing libxrt2 and libxrt2-npu seems to fix the issue.
Using apt-get instead of apt simply avoids an annoying warning.


gunnarx:
Just a heads-up - there are also different failures in publishing the container here but they are of a different nature:  See https://github.com/lemonade-sdk/lemonade/actions/workflows/build-container.yml

The problem I had (building with podman, not docker, but that seems irrelevant), was that fastflowlm.deb required libxrt-npu2 / libxrt2.

superm1 (approved):

PR #1803, opened 2026-05-06
Improve model manager naming and grouping

jeremyfowers:
## Summary

- Hide `user.` and `extra.` prefixes in the Model Manager display, search, sorting, and loaded-model list while preserving canonical model IDs for actions and Load Options.
- Align Hugging Face app installs with CLI-style names by appending the selected GGUF variant to the default `user.*` model name.
- Broaden llama.cpp model family grouping for custom/local variants and add groups for Bonsai, Gemma-4, Qwen2.5-Omni, and Qwen3.6.
- Ensure specific Qwen families claim models before broader Qwen groups so models do not appear in multiple families.

Idea ported from https://github.com/lemonade-sdk/lemonade/pull/1657 by @anditherobot

## Demo

image

Important things to note:
- Qwen3's 0.6B-UD-1Q_M.gguf is an `extra_models_dir` model. This will be a lot more obvious in concert with https://github.com/lemonade-sdk/lemonade/pull/1800 because it will have nice "user" icon.
- Qwen3's 0.6B-GGUF is the old-style custom model name. This is unfortunately not that distinct, but again the user icon will help distinguish it. Qwen3.5's 0.8B-GGUF-UD-IQ2_M is a new-style name, which is obviously distinct.

### DIFF
--- a/src/app/src/renderer/ModelManager.tsx
+++ b/src/app/src/renderer/ModelManager.tsx
@@ -23,6 +23,7 @@ import { getCollectionComponents, isCollectionFullyDownloaded, isCollectionModel
 interface ModelFamily {
   displayName: string;
   regex: RegExp;
+  recipe?: string;
 }
 
 const SIZE_TOKEN = String.raw`(\d+\.?\d*B(?:-A\d+\.?\d*B)?)`;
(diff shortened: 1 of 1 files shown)


danielholanda (approved):
Looks good and it is "backwards compatible" since we are only changing the app, not the actual server.

sofiageo:
This looks great but I didn't manage to fully test it. I don't like how in the old custom style you get the same name and no way to easily identify it (maybe on-hover should show the checkpoint). I will test it more tomorrow.

PR #1823, opened 2026-05-06
Add static max context metadata to /models

### LINKED ISSUE #1794: Report `max_context_window` on `/models` endpoints
### Feature Description

For tools with integrate the Lemonade server it would be helpful to report the maximum token context window supported for a specific model when listing via `/models`. This would allow the user or developer to be aware of the maximum amount of tokens they can potentially have in context before using the model

### Use Case / Motivation

AnythingLLM is very prudent with context windows and being able to auto-detect the context window is a helpful automation we can abstract from the user so they do not have to specify a value that may be too small/large for their system or use case.

It is totally okay in our use case to only have this value available on models that are already downloaded. This would also be great to expand into an Ollama-like `capabilities` array where we can report `tool-calling` or `reasoning`  derived from the chat template for such a model so it does not need to be manually added into the model config.

### Platform Relevance

All platforms

### Additional Context

_No response_

jeremyfowers:
Closes https://github.com/lemonade-sdk/lemonade/issues/1794

image

## Summary

Adds optional `max_context_window` metadata to OpenAI-compatible model responses when Lemonade can discover a model's static text context limit from local files.

- Parses downloaded GGUF headers for `general.architecture` and `<architecture>.context_length`.
- Reads installed FLM model `config.json` and uses `max_position_embeddings` for text-context models.
- Emits the field through the shared model serializer, so it appears in both `/v1/models` and `/v1/models/{model_id}` when known.
- Uses the metadata in the app's Load Options modal to show a context-size slider with power-of-two marks from 2k up to the model maximum, while keeping the editable text box in sync. Entering `0` remains valid and moves the slider to the maximum.
- Documents the field and adds a metadata regression test for llama.cpp and FLM.

This does not change runtime `ctx_size`, saved recipe options, or backend launch behavior.

## Demo

GGUF model:

```bash
curl -sS http://127.0.0.1:13305/v1/models/LFM2-1.2B-GGUF
```

Result:

```json
{"checkpoint":"LiquidAI/LFM2-1.2B-GGUF:LFM2-1.2B-Q4_K_M.gguf","checkpoints":{"main":"LiquidAI/LFM2-1.2B-GGUF:LFM2-1.2B-Q4_K_M.gguf"},"composite_models":[],"created":1234567890,"downloaded":true,"id":"LFM2-1.2B-GGUF","labels":[],"max_context_window":128000,"object":"model","owned_by":"lemonade","recipe":"llamacpp","recipe_options":{},"size":0.731,"suggested":true}
```

FLM model:

```bash
curl -sS http://127.0.0.1:13305/v1/models/gemma4-it-e2b-FLM
```

Result:

```json
{"checkpoint":"gemma4-it:e2b","checkpoints":{"main":"gemma4-it:e2b"},"composite_models":[],"created":1234567890,"downloaded":true,"id":"gemma4-it-e2b-FLM","labels":["vision","reasoning","tool-calling","audio","transcription"],"max_context_window":131072,"object":"model","owned_by":"lemonade","recipe":"flm","recipe_options":{},"size":6.0,"suggested":true}
```

## Validation

- `cmake --build --preset default --target web-app`
- `uv run python test/server_llm.py --wrapped-server llamacpp --backend cpu LLMTests.test_000_model_info_includes_max_context_window` (runner executed the full LLM class: 24 tests OK, 4 skipped)
- Targeted FLM unittest invocation for `LLMTests.test_000_model_info_includes_max_context_window`
- Manual curl checks for `LFM2-1.2B-GGUF` and `gemma4-it-e2b-FLM`

### DIFF
--- a/docs/api/openai.md
+++ b/docs/api/openai.md
@@ -878,7 +878,7 @@ The generated audio file is returned as-is.
 ## `GET /v1/models`
 <sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
 
-Returns a list of models available on the server in an OpenAI-compatible format. Each model object includes extended fields like `checkpoint`, `recipe`, `size`, `downloaded`, and `labels`.
+Returns a list of models available on the server in an OpenAI-compatible format. Each model object includes extended fields like `checkpoint`, `recipe`, `size`, `downloaded`, `labels`, and, when known, `max_context_window`.
 
 By default, only models available locally (downloaded) are shown, matching OpenAI API behavior.
 
@@ -912,6 +912,7 @@ curl http://localhost:13305/v1/models?show_all=true
       "checkpoint": "unsloth/Qwen3-0.6B-GGUF:Q4_0",
       "recipe": "llamacpp",
       "size": 0.38,
+      "max_context_window": 40960,
       "downloaded": true,
       "suggested": true,
       "labels": ["reasoning"]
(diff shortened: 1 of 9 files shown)


ramkrishna2910 (approved):
Looks good!

PR #1829, opened 2026-05-06
Warm model cache asynchronously

jeremyfowers:
## Summary

- Warm the `ModelManager` model list cache asynchronously during `lemond` startup.
- Pass `extra_models_dir` into `ModelManager` construction so discovery configuration is initialized before cache warmup.
- Keep warmup success logs at debug level while preserving warning logs for failures.

## Impact

This removes the first `/v1/models` cache-build pause for normal startup flows while preserving the existing cache mutex behavior for concurrent requests and invalidation.

Benchmark after starting `lemond`, waiting 5 seconds, then timing `curl http://localhost:13305/v1/models`:

- Before: `0.436339s`
- After: `0.000943s`

On Windows the improvement is even more striking, it can reduce a ~10 second wait to instant.

### DIFF
--- a/src/cpp/include/lemon/model_manager.h
+++ b/src/cpp/include/lemon/model_manager.h
@@ -95,7 +95,7 @@ struct ModelInfo {
 
 class ModelManager {
 public:
-    ModelManager();
+    explicit ModelManager(const std::string& extra_models_dir = "");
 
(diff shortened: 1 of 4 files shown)


ramkrishna2910:
@claude review

ramkrishna2910 (approved):

PR #1830, opened 2026-05-06
vllm: stop reporting update_required immediately after install

ramkrishna2910:
Closes #1816.

The install path writes `version.txt = "{base}-{gfx_target}"` (e.g. `vllm0.20.1-rocm7.12.0-gfx1151`) via `version_override`, but `system_info.cpp::build_recipes_info` compared it against the base pin from `backend_versions.json` (`vllm0.20.1-rocm7.12.0`) — never equal, so `vllm:rocm` perpetually showed `update_required` and the desktop app prompted to download every load.

Fix: in the version comparison, treat `"{expected}-{anything}"` as a match for `"{expected}"`. Recipe-agnostic.

## Test plan

```
$ build/lemonade backends 2>&1 | grep '^vllm'
before:  vllm  rocm  update_required  Backend update is required before use.
after:   vllm  rocm  installed        vllm0.20.1-rocm7.12.0-gfx1151
```

- [x] `lemonade backends` no longer reports `update_required` post-install
- [x] Inference round-trip still works

image

PR #1832, opened 2026-05-06
vllm: fix BackendManager download progress for split-archive installs

### LINKED ISSUE #1821: vllm: BackendManager download progress shows '9 bytes' for split-archive install
**From PR #1802 review:** https://github.com/lemonade-sdk/lemonade/pull/1802#discussion_r3195921619

When installing `vllm:rocm` via the BackendManager UI, the download progress incorrectly displays "9 bytes" instead of the actual ~2.5 GB download size.

(Screenshot: https://github.com/user-attachments/assets/77288007-fe80-4401-b0b8-41213b642c92)

## Suspected root cause

The vllm-rocm release uses **split tarballs** (`*-x64.part00.tar.gz` + `*-x64.part01.tar.gz`) because each archive exceeds GitHub's 2 GB per-asset limit. The split-archive code path in `backend_utils.cpp::install_from_github` likely:

- (a) Reports the version.txt write or part00 metadata size to the progress callback before the actual download starts, or
- (b) Sums only one part's `total_bytes` instead of both, or
- (c) Loses the running total when transitioning between parts.

`llamacpp:rocm` and other backends use single-asset releases and don't exercise this code path, which is why the regression is vllm-specific.

## Investigation

1. Add per-callback-event logging in the split-archive path of `BackendUtils::install_from_github`.
2. Check that `DownloadProgress::bytes_total` reflects the *combined* size of both parts, and that the per-part progress callbacks accumulate against that combined total.
3. Verify the BackendManager UI is consuming `bytes_total` and `bytes_downloaded` in the units the backend reports them.

## Acceptance

- Installing `vllm:rocm` from the BackendManager UI shows realistic byte counts as the download progresses (~2.5 GB total).
- The progress bar is monotonically increasing from 0% to 100%.

ramkrishna2910:
Closes #1821.

The vLLM-ROCm release uses split parts (`*-x64.part00.tar.gz`, `*-x64.part01.tar.gz`) because each archive exceeds GitHub's 2 GB per-asset limit. The install path had three progress-reporting bugs that combined to produce the "9 bytes" symptom:

1. Single-file URL 404s for vLLM. Curl wrote the ~9-byte error body and reported it through the progress callback as `bd=9 bt=9 100%` — the stuck "9 bytes" Jeremy saw.
2. Per-part downloads in the split fallback used `create_throttled_progress_callback()` (console-only), so the GUI received nothing during the actual ~2.5 GB transfer.
3. The completion event reported `download_result.bytes_downloaded` — also from the failed single-file attempt (= 9 bytes), not the real archive size.

## Fix

- Filter out tiny progress events (`total < 1 MB`) — real backend archives are tens-to-thousands of MB; error response bodies are bytes-to-KB.
- Reset event before falling through to split-archive loop.
- Per-part progress wrapped to forward to the user's `progress_cb` and report **cumulative** bytes across all parts (so the bar climbs continuously, not per-part).
- Completion event uses `file_size` (verified on-disk archive size) instead of the misleading `download_result.bytes_downloaded`.

## Verified

```
$ lemonade backends uninstall vllm:rocm && lemonade backends install vllm:rocm

before:  GUI flashes "9 bytes 100%" → blank for ~60s → "9 bytes" final
after:   clean ramp 0 → 2.49 GiB → completion at 2.49 GiB
```

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -464,7 +464,6 @@ jobs:
         env:
           LEMONADE_CI_MODE: "True"
           PYTHONIOENCODING: utf-8
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
           HF_HOME: ${{ github.workspace }}/hf-cache
           GGML_METAL_NO_RESIDENCY: "1"
         run: |
@@ -480,7 +479,6 @@ jobs:
         env:
           LEMONADE_CI_MODE: "True"
           PYTHONIOENCODING: utf-8
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
           HF_HOME: ${{ github.workspace }}/hf-cache
           GGML_METAL_NO_RESIDENCY: "1"
         run: |
@@ -495,7 +493,6 @@ jobs:
         env:
           LEMONADE_CI_MODE: "True"
           PYTHONIOENCODING: utf-8
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
           HF_HOME: ${{ github.workspace }}/hf-cache
           GGML_METAL_NO_RESIDENCY: "1"
         run: |
@@ -510,7 +507,6 @@ jobs:
         env:
           LEMONADE_CI_MODE: "True"
           PYTHONIOENCODING: utf-8
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
           HF_HOME: ${{ github.workspace }}/hf-cache
           GGML_METAL_NO_RESIDENCY: "1"
         run: |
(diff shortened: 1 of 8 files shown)


jeremyfowers:
Can you combine this and https://github.com/lemonade-sdk/lemonade/pull/1830 into one PR? It's hard to test one without the other.

ramkrishna2910:
Verification:
Download shows all parts: 
image

Logs dont show 404s:
image


update_required fix:
image

jeremyfowers (approved):
Works locally for me! Pushed a commit to unblock CI - it was tripping on the GH_TOKEN thing.

jeremyfowers (changes requested):
New methodology as discussed on the phone

jeremyfowers (approved):
it's alive!

PR #1842, opened 2026-05-07
Enable custom user collections via /pull

jeremyfowers:
image

image

## Summary

- Extend `POST /pull` to accept `recipe: "collection.omni"` with a `components` array, registering a user-defined collection in `user_models.json`.
- Add `--components` flag to `lemonade pull` so the CLI can compose collections (e.g. `lemonade pull user.MyKit --recipe collection.omni --components Qwen3-0.6B-GGUF Whisper-Tiny SD-Turbo`).
- Validation in both server and `download_model`: `user.*` prefix required, non-empty string array, every component already registered, no self-reference. A visited-set in `download_model` rejects transitive cycles (A → B → A) at download time.
- Components do not need to be downloaded first — the existing collection branch in `download_model` cascades pulls as needed.

## Why

Built-in collections existed in `server_models.json` but couldn't be user-defined. Users now have a registration surface for composing their own multi-model bundles without touching the registry on disk.

## Breaking changes

This PR renames two public-facing names in a coordinated, backwards-incompatible way. There is no transitional alias.

| Old | New | Where it surfaces |
|-----|-----|-------------------|
| `recipe: "collection"` | `recipe: "collection.omni"` | `server_models.json`, `user_models.json`, `/pull` request body, `/models` response, CLI `--recipe` value |
| `composite_models: [...]` | `components: [...]` | `server_models.json`, `user_models.json`, `/pull` request body, `/models` response |

Impact:
- External clients reading `composite_models` from `/v1/models` get an empty/missing field — switch to `components`.
- Distributions that ship a customized `server_models.json` need to rename both fields in the same patch.
- Hand-edited `user_models.json` files containing the old field names will silently lose their components on next load (built-in collections existed prior to this PR; user-authored ones did not).

## Test plan

- [x] New endpoint tests in `test/server_endpoints.py` (`test_021j..q`) cover the success path, missing components, empty array, missing `user.` prefix, self-reference, the load-without-HF-download path for collections, alias canonicalization, and the re-pull overwrite path.
- [x] New CLI tests in `test/server_cli2.py` (`test_055`, `test_056`) cover the `--components` happy path and the missing-`--components` error path.
- [x] Existing pull/delete regressions still pass (`test_007`, `test_016`, `test_021_pull_multi`, `test_021a`, `test_021c..i` naming-spec tests).
- [x] End-to-end smoke: register collection via curl + CLI, verify in `GET /models?show_all=true`, persist across `lemond` restart, delete removes only the collection (components remain).


src/cpp/server/model_manager.cpp
```diff
@@ -2276,33 +2276,40 @@ void ModelManager::download_model(const std::string& model_name,
...
+            if (actual_checkpoint.empty() || actual_recipe.empty()) {
                 throw std::runtime_error(
-                    "You are required to provide a 'variant' in the checkpoint field when "
-                    "registering a GGUF model. The variant is provided as CHECKPOINT:VARIANT. "
-                    "For example: Qwen/Qwen2.5-Coder-3B-Instruct-GGUF:Q4_0 or "
-                    "Qwen/Qwen2.5-Coder-3B-Instruct-GGUF:qwen2.5-coder-3b-instruct-q4_0.gguf"
+                    "Model " + model_name + " is not registered with Lemonade Server. "
+                    "To register and install it, provide the `checkpoint` and `recipe` "
+                    "arguments, as well as the optional `reasoning` and `mmproj` arguments "
+                    "as appropriate."
```
jeremyfowers: Don't mention the optional arguments.

src/cpp/server/server.cpp
fl0rianr, line 3019: This runs before the collection-specific branch below, so a collection that is not yet fully downloaded goes through `download_registered_model(info)`. Collections do not have their own backend/checkpoint, and download_registered_model routes non-FLM recipes to `download_from_huggingface(info)`, so load can fail before it ever reaches the component fan-out.

This also contradicts the new docs that say loading a collection installs/loads each component as needed. Could we move the collection branch before this generic download block and download missing components through the same collection-aware path used by /pull? A regression test for `POST /load` on a collection with an undownloaded component would be helpful.

fl0rianr: The new API docs say per-model options like ctx_size or  backend etc.  are not forwarded to collection components and should instead be set on each component's own recipe_options.json entry (Which is good). This line still builds each component's options from the original collection load request, so request-level options are forwarded to every component.

src/cpp/server/model_manager.cpp
```diff
@@ -1273,23 +1273,23 @@ static void parse_legacy_mmproj(ModelInfo& info, const json& model_json) {
...
-// Check if all component models of a collection model are downloaded.
-static bool check_composite_downloaded(const ModelInfo& info,
+// Check if all components of a collection model are downloaded.
+static bool check_component_downloaded(const ModelInfo& info,
                                         const std::map<std::string, ModelInfo>& model_map) {
-    if (info.composite_models.empty()) return false;
-    for (const auto& component_name : info.composite_models) {
+    if (info.components.empty()) return false;
+    for (const auto& component_name : info.components) {
         auto it = model_map.find(component_name);
```
fl0rianr, line 1293: validate_collection_request() accepts component names via `model_exists(component_name)`, which resolves public aliases, but the collection stores the submitted string as-is. Later this check uses a direct `model_map.find(component_name)`, and `update_model_in_cache()` also compares component strings directly against the canonical model name.

So a client can build a collection from IDs returned by /models — for example a unique `user.X` shown publicly as `X` — and registration succeeds, but downloaded state does not update reliably because the cache key is `user.X`.

Could we canonicalize components at registration time, e.g. via `resolve_model_name()`, or make all component lookups consistently go through alias resolution? A good regression would register `user.ComponentX`, create a collection with components: ["ComponentX"] and verify downloaded status/update behavior.

src/cpp/server/model_manager.cpp
fl0rianr: For already-registered models, this treats an empty actual_checkpoint as "use the existing registry entry". That works for normal models, but collections intentionally have no checkpoint. So if a user calls /pull or lemonade pull again with `recipe: "collection.omni-model"` and a new components array for an existing collection, the request can return success while silently ignoring the new components.

Can we either treat collection.omni-model + components as an overwrite even when no checkpoint is present, or explicitly reject updates to an existing collection with a clear 4xx? Silent success with no update will be hard to diagnose from the CLI/API.

fl0rianr (changes requested):
Hey @jeremyfowers, very good work! A lot of changes, I left a few comments and I don't think we can merge that as-is and still some minor adaptions are needed.

A small nitty question, do you really consider the bots point as resolved? The code still keeps model_name in `visited` after returning from the recursive call, so `visited` tracks “seen anywhere” instead of “currently on the recursion stack”. This still false-positives on valid DAGs where two branches share the same sub-collection, e.g. `A -> B -> D` and `A -> C -> D`. Or did I get this wrong?

src/cpp/server/model_manager.cpp
jeremyfowers: Already addressed in commit 2d61498e7 — the error message no longer mentions the optional `reasoning`/`mmproj` arguments.

src/cpp/server/model_manager.cpp
```diff
@@ -2325,13 +2341,31 @@ void ModelManager::download_model(const std::string& model_name,
...
+    // Collections don't have their own backend — download each component instead
+    if (is_collection_recipe(actual_recipe)) {
+        // Cycle guard: re-entering the same collection on the current call
+        // chain means the user registered a circular reference (e.g. user.A
+        // includes user.B which includes user.A). Without this, the fan-out
+        // would recurse until the stack or HTTP timeout gives out.
+        if (!visited.insert(model_name).second) {
+            throw std::runtime_error(
+                "Cycle detected in collection components: '" + model_name +
+                "' transitively references itself. Edit the offending collection "
```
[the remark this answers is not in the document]
jeremyfowers: Fixed in commit 78416ad1d. `visited.erase(model_name)` now runs immediately before the collection branch returns, so the set tracks "currently on the recursion stack" instead of "seen anywhere". Valid DAGs like `A→{B,C}→D` (with `D` itself a collection) no longer false-positive.

src/cpp/server/server.cpp
jeremyfowers: Fixed in commit aef4b7e19. `handle_load` now guards the generic `download_registered_model(info)` call with `!is_collection_recipe(info.recipe)`, so collections (which have no checkpoint of their own) skip straight to the per-component branch — that branch already cascades downloads via `download_registered_model(comp_info)` for each missing component. Added `test_021o_load_collection_routes_through_component_branch` as a regression in `test/server_endpoints.py`.

jeremyfowers: Fixed in commit e98c02c85. Component load now passes `comp_info.recipe_options` instead of `RecipeOptions(comp_info.recipe, request_json)`, matching the documented contract that per-model options are not forwarded from the collection load request to its components.

src/cpp/server/model_manager.cpp
jeremyfowers: Fixed in commit d90083275. After `validate_collection_request` succeeds in `handle_pull`, each component is run through `resolve_model_name()` and written back into `request_json["components"]` before it reaches `register_user_model`. Persisted components are now in canonical `user.X` / `builtin.X` form, so `check_component_downloaded` and `update_model_in_cache` find them via direct cache-key lookup. Added `test_021p_collection_components_canonicalized` covering exactly the scenario you suggested (register `user.AliasComp-…`, build a collection from the bare alias, assert canonical storage + `downloaded=true`).

src/cpp/server/model_manager.cpp
jeremyfowers: Fixed in commit 1331a4c0e. `download_model` now detects the collection-overwrite case (already-registered + `collection.omni` recipe + `components` present in the request) and: re-runs `validate_collection_request`, sets `model_registered = false`, and lets the existing `register_user_model` call downstream overwrite the entry. The previous behavior of silently dropping the new components is gone — re-pulling a collection now persists the new list. Added `test_021q_collection_repull_overwrites_components` as a regression.

fl0rianr (approved):
Great job! Thanks @jeremyfowers, accepted.

Small nit, non-blocking: after component canonicalization, the second collection validation happens inside `download_model()`. If that validation rejects user input, it looks like the sync /pull path will return 500 via the generic std::exception catch instead of 400. Not a blocker, but it might be cleaner to canonicalize before the first validation or map these validation failures to 400 consistently.

jeremyfowers:
> Great job! Thanks @jeremyfowers, accepted.
> 
> Small nit, non-blocking: after component canonicalization, the second collection validation happens inside `download_model()`. If that validation rejects user input, it looks like the sync /pull path will return 500 via the generic std::exception catch instead of 400. Not a blocker, but it might be cleaner to canonicalize before the first validation or map these validation failures to 400 consistently.

Thanks! I will go ahead and merge to unblock your PR, and if we run into a problem with the validation let's do a follow up.

PR #1843, opened 2026-05-07
Add FLM model catalog snapshot fallback

### LINKED ISSUE #1834: Add FLM model catalog snapshot fallback
## Implementation Plan

- Generate `src/cpp/resources/flm_models.json` from `flm list --json`.
- Normalize FLM catalog entries into Lemonade model registry shape.
- Strip machine-local FLM fields such as installed state from the generated snapshot.
- Load FLM catalog data in `ModelManager`.
- Use live `flm list --json` whenever FLM is installed.
- Use `flm_models.json` only when FLM is not installed.
- Keep local `lemond` model APIs hardware-filtered so unsupported systems do not expose FLM models.
- Include FLM snapshot models in unfiltered model lookup paths for useful unsupported/install/update guidance.
- Update the website's model catalog loading to merge `server_models.json` and optional `flm_models.json`.
- Package `flm_models.json` wherever server resources are explicitly bundled.
- Add an automation workflow to refresh `flm_models.json` when the pinned FLM version changes.

jeremyfowers:
Closes #1834.

## Summary
- `/models?show_all=true` and the lemonade.ai catalog now surface FLM models on supported hardware even before FLM is installed, so users see install/upgrade affordances.
- Live `flm list --json` is still preferred when FLM is installed; the bundled `src/cpp/resources/flm_models.json` snapshot is the fallback. Hardware filtering in `filter_models_by_backend()` keeps FLM out of `/models` on non-NPU hosts (the entries land in `filtered_out_models_` so `get_model_filter_reason()` can serve "this requires NPU" guidance).
- New `tools/generate_flm_models.py` produces the snapshot from `flm list --json` (strips per-host fields like `installed`/`url`).
- New `.github/workflows/refresh_flm_models.yml` regenerates the snapshot when the FLM pin in `backend_versions.json` changes (or via `workflow_dispatch`). On push it opens a PR; on PR it fails with instructions so contributors regenerate locally rather than us auto-committing into their branch.
- Snapshot is bundled in the embeddable archive (CMakeLists.txt) and asserted present by the existing release verification step.

## Test plan
- [x] `cmake --build --preset default` builds clean.
- [x] With FLM on PATH (Ubuntu host with `/usr/bin/flm`): live discovery wins; log emits `FLM binary found at: flm`; `/api/v1/models?show_all=true` returns 35 FLM models.
- [x] With FLM removed from PATH (sandbox PATH): snapshot fallback activates; log emits `Loading FLM snapshot from .../flm_models.json`; same 35 FLM models surface, all `downloaded=false`.
- [x] `/api/v1/models/qwen3-4b-FLM` returns the correct snapshot details.
- [ ] Non-NPU host: FLM models filtered out of `/api/v1/models?show_all=true`; `/api/v1/models/<flm-id>` returns the unsupported-hardware reason. (Not testable on this NPU host — covered by existing `filter_models_by_backend` behavior.)
- [ ] Website: open `docs/models.html` after this branch is the latest tag; confirm FLM models appear under "FastFlowLM NPU" filter.
- [ ] `refresh_flm_models.yml` `workflow_dispatch` run is a no-op against the current pin (committed snapshot matches FLM v0.9.39 output).


jeremyfowers:
Closing stale PR

PR #1845, opened 2026-05-07
Updated to pull llamacpp-rocm builds from the new consolodated llama.cpp-builds repo

kenvandine:
With the goal of consolidating repositories where we build llamacpp, this updates the ROCm backend to fetch from the new repository. Other backends coming.


jeremyfowers (changes requested):
Let’s have a conversation about this, llamacpp-rocm is rather popular.

kenvandine:
We'll keep llamacpp-rocm separate

PR #1857, opened 2026-05-09
Add support for hermes tool calling

superm1:
Hermes has a different schema for tool calling than other agents.

Link: https://markaicode.com/hermes-agent-not-working-fix/

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1180,7 +1180,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        test_type: [cli, endpoints, ollama, llamacpp-system, streaming-errors, env-vars]
+        test_type: [cli, endpoints, ollama, llamacpp-system, streaming-errors, env-vars, hermes-transformation]
     env:
       LEMONADE_CI_MODE: "True"
       PYTHONIOENCODING: utf-8
@@ -1248,6 +1248,9 @@ jobs:
           elif [ "${{ matrix.test_type }}" = "env-vars" ]; then
               echo "Running environment variable tests..."
               $VENV_PYTHON test/server_env_vars.py --lemond-binary ./deb-extract/usr/bin/lemond
+          elif [ "${{ matrix.test_type }}" = "hermes-transformation" ]; then
+              echo "Running Hermes XML transformation tests..."
+              $VENV_PYTHON test/test_hermes_transformation.py
           fi
 
           echo "${{ matrix.test_type }} tests PASSED!"
@@ -1262,7 +1265,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, macos-latest]
-        test_type: [cli, endpoints, ollama, llamacpp-system, streaming-errors]
+        test_type: [cli, endpoints, ollama, llamacpp-system, streaming-errors, hermes-transformation]
         include:
           - os: macos-latest
             test_type: env-vars
@@ -1358,6 +1361,9 @@ jobs:
               echo "Running environment variable tests..."
               LEMOND_BINARY="$(dirname "$SERVER_BINARY")/lemond"
               $VENV_PYTHON test/server_env_vars.py --lemond-binary "$LEMOND_BINARY"
+          elif [ "${{ matrix.test_type }}" = "hermes-transformation" ]; then
+              echo "Running Hermes XML transformation tests..."
+              $VENV_PYTHON test/test_hermes_transformation.py
           fi
 
           echo "${{ matrix.test_type }} tests PASSED!"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -571,6 +571,7 @@ set(SOURCES_CORE
     src/cpp/server/log_stream.cpp
     src/cpp/server/utils/http_client.cpp
     src/cpp/server/utils/json_utils.cpp
+    src/cpp/server/utils/tool_call_parser.cpp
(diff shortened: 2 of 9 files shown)


LaurentChardin:
I find it quite confusing from NousResearch when we talk about:
- Hermes, the Agent : https://hermes-agent.nousresearch.com
- Hermes, the Model(s) : 
  - https://huggingface.co/collections/NousResearch/hermes-4-collection
  - https://huggingface.co/collections/NousResearch/hermes-3

Here you are referring to the model, as markaicode is referring to hermes-3 models in the agentic pipeline.

Genuine question : Was wondering if the format mistmatch should actually be handled down in the stack, at llama.cpp level ? Looks like we are slowly dispatching model specific behavior all over the pipeline stack. If i understand well, you are basically remapping the Hermes model tool output to OpenAI format, right ?

superm1:
Well llama.cpp is what I used here; but conceivably this could be used with Fastflowlm or another backend provider too. So I'm not sure we should be pushing for support all over the backends (although I could see that argument).

I think all the cloud services provided by hermes probably use vLLM.

And yes this was hermes agent not hermes model.

superm1:
@sawansri any plans to review soon?  If you're busy no worry.

I don't think you need to reproduce the whole setup unless you want to - I'm actively using this with Hermes myself.

sawansri:
> @sawansri any plans to review soon?  If you're busy no worry.
> 
> I don't think you need to reproduce the whole setup unless you want to - I'm actively using this with Hermes myself.

I'll take a look today, thx for the reminder

sawansri:
image

@superm1 What tool calling issues did you run into specifically? `hermes chat` with tool calls seems to be working fine for me. The blog you linked seems to be talking about Hermes 3 8B the model not the agent.

I'm building off of main and running into no issues so far.

superm1:
> image
> @superm1 What tool calling issues did you run into specifically? `hermes chat` with tool calls seems to be working fine for me. The blog you linked seems to be talking about Hermes 3 8B the model not the agent.
> 
> I'm building off of main and running into no issues so far.

I ended up with stuff like this in my results rather than calling of tools:

```
╭─ ⚕ Hermes ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
    I'll help you build a Linux kernel from the source code in ~/src/linux. Let me first check if the source code is already present and then proceed with the build process.
    
    <function=terminal>
    <parameter=command>
    ls -la ~/src/linux
    </parameter>
    </function>
    </tool_call>
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

sawansri:
> I ended up with stuff like this in my results rather than calling of tools:
> 
> ```
> ╭─ ⚕ Hermes ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
>     I'll help you build a Linux kernel from the source code in ~/src/linux. Let me first check if the source code is already present and then proceed with the build process.
>     
>     <function=terminal>
>     <parameter=command>
>     ls -la ~/src/linux
>     </parameter>
>     </function>
>     </tool_call>
> ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
> ```

Interesting, this could be an issue with llamacpp not picking up the correct chat template. What model are you using here? Qwen 3.6 MoE works fine for me.

superm1:
This is what I was using:

```
  "user.Qwen3.6-35B-A3B-ThinkingCoder": {
    "ctx_size": 0,
    "llamacpp_args": "--temp 0.6 --top-p 0.95 --top-k 20 --min-p 0.00 --presence-penalty 0 --repeat-penalty 1.0 --chat-template-kwargs '{\"preserve_thinking\": true}' -b 4096 -ub 1024",
    "llamacpp_backend": "vulkan"
  }
```

sawansri:
> This is what I was using:
> 
> ```
>   "user.Qwen3.6-35B-A3B-ThinkingCoder": {
>     "ctx_size": 0,
>     "llamacpp_args": "--temp 0.6 --top-p 0.95 --top-k 20 --min-p 0.00 --presence-penalty 0 --repeat-penalty 1.0 --chat-template-kwargs '{\"preserve_thinking\": true}' -b 4096 -ub 1024",
>     "llamacpp_backend": "vulkan"
>   }
> ```

image

Just tried that recipe out, seems to work on my end...

I'm guessing the autoparser isn't working for some reason? We can debug this over discord if necessary.

superm1:
> > This is what I was using:
> > ```
> >   "user.Qwen3.6-35B-A3B-ThinkingCoder": {
> >     "ctx_size": 0,
> >     "llamacpp_args": "--temp 0.6 --top-p 0.95 --top-k 20 --min-p 0.00 --presence-penalty 0 --repeat-penalty 1.0 --chat-template-kwargs '{\"preserve_thinking\": true}' -b 4096 -ub 1024",
> >     "llamacpp_backend": "vulkan"
> >   }
> > ```
> 
> image
> Just tried that recipe out, seems to work on my end...
> 
> I'm guessing the autoparser isn't working for some reason? We can debug this over discord if necessary.

Since you just did a fresh setup and got everything working let me uprev everything and tear out my PR see if I can repro again on main and then we'll see what next step is.

superm1:
I upreved everything and it's all working now...

Will close for now, and if I figure out why that was happening will come back.

sawansri:
> I upreved everything and it's all working now...
> 
> Will close for now, and if I figure out why that was happening will come back.

Sounds good.

PR #1866, opened 2026-05-11
Fix: Properly terminate child processes on systemd service stop

superm1:
The signal handler was calling _exit(0) immediately on SIGINT/SIGTERM, which terminated the process without running destructors or calling Server::stop(). This orphaned child processes like llama-server that were spawned as subprocesses by the backend servers.

Changes:
- Signal handler now sets shutdown_requested_ flag and returns normally instead of calling _exit(0)
- Server::run() main loop now checks shutdown_requested_ and calls stop() before exiting, ensuring proper cleanup
- Added shutdown_requested_ member to Server class with getter/setter
- wait_until_not_busy() now accepts an optional timeout parameter to prevent infinite blocking if a backend is stuck
- evict_all_servers() uses 30-second timeout for waiting on backends
- SIGHUP thread now checks Server::should_shutdown() instead of the removed global flag
- Removed unused g_shutdown_requested global variable

This ensures that when systemd sends SIGINT to stop the service, lemond properly unloads models and terminates backend child processes (llama-server, whisper-server, etc.) before exiting.

### DIFF
--- a/src/cpp/server/router.cpp
+++ b/src/cpp/server/router.cpp
@@ -164,9 +164,9 @@ void Router::evict_server(WrappedServer* server) {
 void Router::evict_all_servers() {
     LOG(INFO, "Router") << "Evicting all models (" << loaded_servers_.size() << " total)" << std::endl;
 
-    // Wait for all servers to finish
+    // Wait for all servers to finish (with timeout to prevent infinite hang).
     for (const auto& server : loaded_servers_) {
-        server->wait_until_not_busy();
+        server->wait_until_not_busy(EVICTION_TIMEOUT);
     }
 
     // Unload all
--- a/src/cpp/include/lemon/router.h
+++ b/src/cpp/include/lemon/router.h
@@ -12,6 +12,10 @@
 #include "backend_manager.h"
 #include "runtime_config.h"
 
+// 5 seconds is generous enough for inference to complete but prevents
+// indefinite blocking if a backend is stuck.
+#define EVICTION_TIMEOUT 5
+
 namespace lemon {
 
 using json = nlohmann::json;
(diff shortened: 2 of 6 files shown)


superm1:
I'm not sure if 30s is too long for this.  It _feels_ long.  I'm wondering if it should be immediate or less time.

jeremyfowers:
And the winner of the award for spiciest branch name 2026 is…

bitgamma:
this is a very welcome change since sometimes llamacpp remains hanging. I see one failing test which seems to be the server failing to shut down though

superm1:
@bitgamma can you review again now?  I got CI fixed.

superm1:
And comments please about 30s; if it should be made shorter like 5-15.

sofiageo:
> And comments please about 30s; if it should be made shorter like 5-15.

I think it should be made shorter. 10-15 seconds

src/cpp/server/router.cpp
```diff
@@ -164,9 +164,11 @@ void Router::evict_server(WrappedServer* server) {
 void Router::evict_all_servers() {
     LOG(INFO, "Router") << "Evicting all models (" << loaded_servers_.size() << " total)" << std::endl;
 
-    // Wait for all servers to finish
+    // Wait for all servers to finish (with timeout to prevent infinite hang).
+    // 30 seconds is generous enough for inference to complete but prevents
```
bitgamma: it still says 30 seconds here but in the code is 5

src/cpp/server/router.cpp
```diff
@@ -164,9 +164,11 @@ void Router::evict_server(WrappedServer* server) {
...
+    // indefinite blocking if a backend is stuck.
     for (const auto& server : loaded_servers_) {
-        server->wait_until_not_busy();
+        server->wait_until_not_busy(5);
```
bitgamma: would be better to define a constant so it is easier to find/change. 5 secs is ok for me

bitgamma (approved):
lgtm, just two nitpicks

PR #1868, opened 2026-05-11
fix model resolution for sharded models on windows

bitgamma:
Fix bug reported on discord where sharded models wouldn't be properly resolved on Windows.


bitgamma:
user reported that the fix is working and solves the problem

jeremyfowers (approved):

PR #1872, opened 2026-05-12
Drop support for user and external prefixes on models

superm1:
Simplifies model naming by removing the requirement for user. prefix on custom models and extra. prefix on models from --extra-models-dir. Model names are now used directly, making Lemonade easier to integrate with external tools.                              
                                                                                                                                                                                                                                                                              
  Changes                                                                                                                                                                                                                                                                     
                                                                       
  Before:                                                                                                                                                                                                                                                                     
  lemonade pull user.MyModel --checkpoint repo/id:variant --recipe llamacpp
                                                                                                                                                                                                                                                                              
  After:                                                                                                                                                                                                                                                                      
  lemonade pull MyModel --checkpoint repo/id:variant --recipe llamacpp                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                              
  Key Behavior: Configuration Updates                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                              
  When you provide custom parameters (checkpoint/recipe) for a model name, you're updating that model's configuration, not creating a duplicate:                                                                                                                              
                                                                                                                                                                                                                                                                              
  # Update built-in model to use different checkpoint                                                                                                                                                                                                                         
  lemonade pull Qwen3-0.6B-GGUF --checkpoint different/repo:Q8_0 --recipe llamacpp                                                                                                                                                                                            
                                                                                                                                                                                                                                                                              
  What happens:                                                                                                                                                                                                                                                               
  - Saves custom configuration to user_models.json                                                                                                                                                                                                                            
  - Custom config takes precedence over built-in defaults                                                                                                                                                                                                                     
  - All clients using "Qwen3-0.6B-GGUF" now get the updated checkpoint 
  - Clear warning logged: ⚠ Updating configuration for built-in model 'Qwen3-0.6B-GGUF'                                                                                                                                                                                      
                                                                                                                                                                                                                                                                              
  To restore built-in:                                                                                                                                                                                                                                                        
  lemonade delete Qwen3-0.6B-GGUF  # Removes custom override, restores built-in                                                                                                                                                                                               
                                                                                                                                                                                                                                                                              
  Corner Cases Addressed                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                              
  1. Intentional override of built-in models - Allowed and logged with warning showing old/new checkpoint                                                                                                                                                                     
  2. Custom model updates - Logged with info showing configuration change                                                                                                                                                                                                     
  3. Multi-client environments - Warning indicates "This affects ALL clients using this model name"                                                                                                                                                                           
  4. Extra-models-dir collisions - Existing warning behavior preserved (skips conflicting models)                                                                                                                                                                             
                                                                                                                                                                                                                                                                              
  Implementation Details                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                              
  - Precedence: user_models.json overrides server_models.json (built-ins)                                                                                                                                                                                                     
  - Logging: Enhanced warnings when updating existing model configurations
  - Restoration: DELETE /v1/models/{name} removes custom config, restores built-in defaults                                                                                                                                                                                   
  - No prefixes anywhere: CLI, API, frontend, tests all updated                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                              
  Migration                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                              
  Existing user. and extra. prefixed models in user_models.json will need migration (manual or automated script). Consider providing a migration tool or documentation for users with existing custom models.

### DIFF
--- a/src/app/src/renderer/hooks/useModels.tsx
+++ b/src/app/src/renderer/hooks/useModels.tsx
@@ -1,5 +1,5 @@
 import React, { createContext, useContext, useState, useEffect, useCallback, useMemo, useRef } from 'react';
-import { ModelsData, ModelInfo, USER_MODEL_PREFIX, fetchSupportedModelsData } from '../utils/modelData';
+import { ModelsData, ModelInfo, fetchSupportedModelsData } from '../utils/modelData';
 import { onServerPortChange } from '../utils/serverConfig';
 import { isModelEffectivelyDownloaded } from '../utils/collectionModels';
 
@@ -60,10 +60,10 @@ export const ModelsProvider: React.FC<{ children: React.ReactNode }> = ({ childr
       .map(([id, info]) => ({ id, info }));
   }, [modelsData]);
 
-  // Derive suggested models for Model Manager (suggested + user models)
+  // Derive suggested models for Model Manager
   const suggestedModels = useMemo<SuggestedModel[]>(() => {
     return Object.entries(modelsData)
-      .filter(([name, info]) => info.suggested || name.startsWith(USER_MODEL_PREFIX))
+      .filter(([name, info]) => info.suggested)
       .map(([name, info]) => ({ name, info }))
       .sort((a, b) => a.name.localeCompare(b.name));
   }, [modelsData]);
--- a/src/app/src/renderer/AddModelPanel.tsx
+++ b/src/app/src/renderer/AddModelPanel.tsx
@@ -159,16 +159,13 @@ const AddModelPanel: React.FC<AddModelPanelProps> = ({ onClose, onInstall, initi
           <label className="form-label" title="A unique name to identify your model in the catalog">
             Model Name
           </label>
-          <div className="input-with-prefix">
-            <span className="input-prefix">user.</span>
-            <input
-              type="text"
-              className="form-input with-prefix"
-              placeholder="Gemma-3-12b-it-GGUF"
-              value={form.name}
-              onChange={(e) => handleChange('name', e.target.value)}
-            />
-          </div>
+          <input
+            type="text"
+            className="form-input"
+            placeholder="Gemma-3-12b-it-GGUF"
+            value={form.name}
+            onChange={(e) => handleChange('name', e.target.value)}
+          />
         </div>
 
         <div className="form-section">
--- a/src/app/src/renderer/ChatWindow.tsx
+++ b/src/app/src/renderer/ChatWindow.tsx
@@ -208,7 +208,7 @@ const ChatWindow: React.FC<ChatWindowProps> = ({ isVisible, width }) => {
     setAddModelInitialValues(undefined);
     window.dispatchEvent(new CustomEvent('installModel', {
       detail: {
-        name: `user.${data.name}`,
+        name: data.name,
         registrationData: {
           checkpoint: data.checkpoint,
           recipe: data.recipe,
--- a/src/app/src/renderer/ModelManager.tsx
+++ b/src/app/src/renderer/ModelManager.tsx
@@ -210,23 +210,8 @@ interface DetectedBackend {
   mmprojFiles?: string[];
 }
 
-const HIDDEN_MODEL_NAME_PREFIXES = ['user.', 'extra.'];
-
-const getModelDisplayName = (modelName: string): string => {
-  const prefix = HIDDEN_MODEL_NAME_PREFIXES.find(p => modelName.startsWith(p));
-  return prefix ? modelName.slice(prefix.length) : modelName;
-};
-
-const isNamespaceHiddenModel = (modelName: string): boolean =>
-  HIDDEN_MODEL_NAME_PREFIXES.some(prefix => modelName.startsWith(prefix));
-
 const getFamilyMemberLabel = (modelName: string, family: ModelFamily, match: RegExpExecArray): string => {
-  if (!isNamespaceHiddenModel(modelName)) return match[1];
-
-  const displayName = getModelDisplayName(modelName);
-  return displayName.startsWith(`${family.displayName}-`)
-    ? displayName.slice(family.displayName.length + 1)
-    : displayName;
+  return match[1];
 };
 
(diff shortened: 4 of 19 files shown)


jeremyfowers:
Can you provide an analysis of the corner cases in the description?

Like what happens if one software uses `/v1/pull` to register a model under a canonical name, and a different piece of software runs `/v1/load` using the canonical name expecting the canonical model?

jeremyfowers:
Also considering the situation where someone shares a debug log that shows a canonical name, but it doesn’t mean the canonical model.

superm1:
> Also considering the situation where someone shares a debug log that shows a canonical name, but it doesn’t mean the canonical model.

why would this matter? 

> Can you provide an analysis of the corner cases in the description?
> 
> Like what happens if one software uses `/v1/pull` to register a model under a canonical name, and a different piece of software runs `/v1/load` using the canonical name expecting the canonical model?

I'll have an LLM generate that and put in the description.

jeremyfowers:
There are clear pros/cons of this, but people really hate user. and extra. 

so I’m good with trying this and seeing if the cons are a big problem or only hypothetical.

jeremyfowers:
@claude does this PR properly remove the code for the  “appear builtin” model label now that it’s no longer needed?

Geramy:
I'll take a look tonight one question @superm1 what happens when a user has a seperate location or repository location selected? How do we handle those cases like the extra env or config variable?

superm1:
> I'll take a look tonight one question @superm1 what happens when a user has a seperate location or repository location selected? How do we handle those cases like the extra env or config variable?

It subsumes any configuration from the json

jeremyfowers:
@claude I remember there was some code in the GUI app to hide the user and extra prefixes. Has this PR removed that dead code?

src/app/src/renderer/hooks/useModels.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { createContext, useContext, useState, useEffect, useCallback, useMemo, useRef } from 'react';
```
jeremyfowers, line 1: Please grep for "user." and "extra.", there are at least 5 instances left:

  - src/cpp/cli/hf_pull.cpp:122-123 — the interactive prompt still says Press enter to use the default: user.{name} and Or type a name starting with "user." and press enter:. Users who follow this prompt will create a custom model literally named user.MyModel.
  - src/cpp/cli/main.cpp:1037,1042,1046 — pull option help text references "custom user.* model" three times (visible in lemonade pull --help).
  - src/cpp/legacy-cli/main.cpp:80 — deprecated-CLI help still prints Manual pull: lemonade pull user.MyModel --checkpoint main org/repo:Q4_K_M --recipe llamacpp. This example is now wrong.
  - src/cpp/include/lemon/model_manager.h:19 — UnknownModelError comment still describes the user. prefix contract (also: the error message itself in download_model was correctly removed, so the comment is now lying about the contract too).
  - test/server_cli2.py:1752 — test_005_cli_legacy_deprecated_help asserts "Manual pull: lemonade pull user.MyModel ..." is in the deprecated help output. The test passes only because the legacy-cli help was not updated. Update both together.

src/app/src/renderer/hooks/useModels.tsx
```diff
@@ -1,5 +1,5 @@
...
-import { ModelsData, ModelInfo, USER_MODEL_PREFIX, fetchSupportedModelsData } from '../utils/modelData';
+import { ModelsData, ModelInfo, fetchSupportedModelsData } from '../utils/modelData';
 import { onServerPortChange } from '../utils/serverConfig';
```
jeremyfowers, line 3: I dont think this PR is backwards compatible to any software that registered a custom model as `user.MY_MODEL` and then makes API calls against that name, since the name is going to suddenly change to `MY_MODEL`. Can you add a backwards compatibility shim that `user.` prefixes referencing user_model.json models ("custom" label) will still work?

src/app/src/renderer/hooks/useModels.tsx
```diff
@@ -1,5 +1,5 @@
...
 import { isModelEffectivelyDownloaded } from '../utils/collectionModels';
```
jeremyfowers, line 5: My understanding is that build_cache (model_manager.cpp:1309-1318) skips a discovered extra_models_dir model when its name matches a built-in or registered model.

Is that accurate? It doesn't seem like correct behavior if the extra_models_dir model is downloaded but the built-in or registered model is not downloaded.

src/app/src/renderer/hooks/useModels.tsx
```diff
@@ -60,10 +60,10 @@ export const ModelsProvider: React.FC<{ children: React.ReactNode }> = ({ childr
       .map(([id, info]) => ({ id, info }));
```
jeremyfowers, line 60: discover_extra_models (model_manager.cpp:792-797) uses the raw filename as model_name for standalone files. With the extra. prefix gone, the user-visible name is literally Qwen3-Coder-3B-Instruct-Q4_K_M.gguf.

Is that intended? IMO the suffix should be stripped.

src/app/src/renderer/hooks/useModels.tsx
jeremyfowers: Same thing for `extra.` models.

src/app/src/renderer/hooks/useModels.tsx
superm1: should be stripped now, unless I made a mistake.

src/app/src/renderer/hooks/useModels.tsx
superm1: fixed

src/app/src/renderer/hooks/useModels.tsx
superm1: fixed

superm1:
Closed in favor of Jeremy's canonical name proposal.

PR #1873, opened 2026-05-12
Merge arguments instead of replacing them.

bitgamma:
When dealing with *_args parameters in recipe options, the arguments from each "layer" are not replaced but merged. 

Binary flags are handled by following the common `--no-` pattern. Flags with multiple arguments are handled correctly. Aliases are not handled (i.e: if on one side you use the extend name and the other side uses the alias, both get into the final command line) because it would require keeping a list for each backend to handle them.

Example:

Global `llamacpp.args`: `"-b 1024 -ub 1024 --temp 1.0 --mmap"` 
Model `llamacpp_args`: `"--temp 0.6 --top-p 0.95 --top-k 40 --no-mmap"` 
Final params:  `"-b 1024 -ub 1024 --temp 0.6 --top-p 0.95 --top-k 40 --no-mmap"`

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -243,6 +243,7 @@ Explicitly load a registered model into memory. This is useful to ensure that th
 | `cfg_scale` | No | sd-cpp | Classifier-free guidance scale for image generation. Default: 7.0. |
 | `width` | No | sd-cpp | Image width in pixels. Default: 512. |
 | `height` | No | sd-cpp | Image height in pixels. Default: 512. |
+| `merge_args` | No | All | Boolean. If true (default), `*_args` values from global config and per-model config are merged (per-model takes priority). If false, per-model `*_args` replace global `*_args` entirely. |
 
 **Setting Priority:**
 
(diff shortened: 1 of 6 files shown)


jeremyfowers:
@claude review

jeremyfowers:
@sawansri are you available to review?

sawansri:
> @sawansri are you available to review?

Yeah, I'll take a look.

sawansri (commented):
I'm not sure if merging args should be enabled by default...

Imo, this should be put behind a feature flag (something like `--merge-args`) so that existing setups are not impacted and people that want to use the feature can opt in.

sawansri (commented):
The main issue that I see with merging args is a lack of semantic understanding in the merging process. If there are args that play well together set globally and model args that override some but not all globally set arguments, the merging process doesn't consider how well these arguments work together before merging them.

Ex: global llamacpp.args sets `--tensor-split 1,2 --gpu-layers 30` but model only sets `--tensor-split 4,5`. Merged arguments look like `--tensor-split 4,5 --gpu-layers 30` which might conflict with the model's specific tensor split.

This can largely be mitigated if the user opts in to the merging behavior because then they are largely aware that the merging process exists and is applied to their global and model specific settings.

bitgamma:
@sawansri after some discussion in Discord we settled on a per-model opt-out rather than an opt-in. Could you please re-review?

jeremyfowers:
Added to breaking changes list here: https://github.com/lemonade-sdk/lemonade/issues/1887

jeremyfowers (approved):
Worked great for me! Tried locally with a mix of global-only, per-model + global, and per-model with --no-merge-args.

sawansri (approved):
LGTM!

PR #1886, opened 2026-05-13
Fixing opencode unreachable on Windows bug

meghsat:
On Windows, `npm i -g opencode-ai` installs three files in `AppData/Roaming/npm`: opencode, opencode.cmd, and opencode.ps1. Since the default is the Unix opencode file, which is not supported on Windows, opencode.cmd is never tried, causing `lemonade launch opencode` to fail.
The fix is to make opencode.cmd the default on Windows.


jeremyfowers (approved):
LGTM

iswaryaalex:
Works for me!

PR #1888, opened 2026-05-13
fix: missing file size in custom models

### LINKED ISSUE #1641: Custom user models don't display their file size in the Web UI
### Platform

Windows

### Lemonade Version

10.2.0

### GPU / APU Model

_No response_

### Component

llama.cpp

### Bug Description

I've installed some user models, but when I view them in the web UI they have "Size N/A" even though they're installed and functional

Image

### Steps to Reproduce

Install a user model (e.g. from huggingface)
Open the Lemonade App web UI
Observe: All user models have "Size N/A" written next to them

Expected: The file size will be displayed
Actual: "Size N/A"

### Log Output

```shell

```

### Additional Context

_No response_

sofiageo:
Fixes #1641. However I don't like how this works.

1.  The **server models** and the **user models** only report the main model file size. We don't calculate the .mmproj file size.

For example for a model like PaddleOCR it will report 893MB while it is double.

```
-rw-r--r-- 1 lemonade lemonade 893M May 13 23:08 PaddleOCR-VL-1.5.gguf
-rw-r--r-- 1 lemonade lemonade 841M May 13 23:09 PaddleOCR-VL-1.5-mmproj.gguf

```

2. It only works when we download new models. We might need a way to "repair" broken models like this.

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -2357,6 +2357,7 @@ void ModelManager::download_registered_model(const ModelInfo& info, bool do_not_
     std::string canonical_model_name = resolve_model_name(info.model_name);
     if (is_user_model_name(canonical_model_name))
     {
+        std::lock_guard<std::mutex> lock(models_cache_mutex_);
         auto it = models_cache_.find(info.model_name);
         if (it != models_cache_.end())
         {
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1771,6 +1771,26 @@ void ModelManager::update_model_in_cache(const std::string& model_name, bool dow
             LOG(INFO, "ModelManager") << "Updated '" << model_name
                       << "' downloaded=" << downloaded << std::endl;
         }
+        // Calculate size in GB
+        uintmax_t total_size = 0;
+        for (auto& [type, path] : it->second.resolved_paths) {
+            try {
+                total_size += fs::file_size(path);
+            } catch (...) {
+                // skip inaccessible entries
+            }
+        }
+        double file_size_gb = static_cast<double>(total_size) / (1024.0 * 1024.0 * 1024.0);
+        if (file_size_gb < 1.0)
+        {
+            it->second.size = std::round(file_size_gb * 1000) / 1000;
+        } else if (file_size_gb < 10.0)
+        {
+            it->second.size = std::round(file_size_gb * 100) / 100;
+        } else
+        {
+            it->second.size = std::round(file_size_gb * 10) / 10;
+        }
 
         // Recompute downloaded status for any collections that
         // depend on this model, so the collection reflects component changes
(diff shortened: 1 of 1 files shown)


sofiageo:
I realized after pushing it it's only adding it to cache and not saving it. Let me take another look tomorrow and see if I can fix it properly.

sofiageo:
I think I can improve this PR by using 3 decimal points on files less than 1GB, and by adding the file size for all checkpoints. What do you think? It's also my first time writing any C++ so I definitely need someone to review my changes :)

bitgamma:
adding the file size of all checkpoints is definitely a good idea!

src/cpp/server/model_manager.cpp
```diff
@@ -1657,7 +1657,14 @@ void ModelManager::update_model_in_cache(const std::string& model_name, bool dow
             LOG(INFO, "ModelManager") << "Updated '" << model_name
```
jeremyfowers, line 1771: There's a bug here, I downloaded Qwen3.5-0.8B-GGUF:Q8_0 and its reported as 778 MB in the GUI.

Image

However when I list the directory its in, it should be more like 1 GB since its the Q8_0 plus the mmproj:

```
jfowers@jfowers-GMK:~/.cache/huggingface/hub/models--unsloth--Qwen3.5-0.8B-GGUF/snapshots/6ab461498e2023f6e3c1baea90a8f0fe38ab64d0$ ls -l
total 1515360
-rw-rw-r-- 1 jfowers jfowers 532517120 Apr 21 15:10 Qwen3.5-0.8B-Q4_K_M.gguf
-rw-rw-r-- 1 jfowers jfowers 811843840 May 19 09:46 Qwen3.5-0.8B-Q8_0.gguf
-rw-rw-r-- 1 jfowers jfowers 207346528 Apr 21 15:10 mmproj-BF16.gguf
```

sofiageo: I didn't get to do this yet, after I finish with the flatpak submission I will fix this asap.

But I wanted your comments first because the server_models don't take the mmproj size into account (at least the last time I checked).

jeremyfowers: > But I wanted your comments first because the server_models don't take the mmproj size into account (at least the last time I checked).

Well that isn't good! I'm going to fix it now.

jeremyfowers: Please review! https://github.com/lemonade-sdk/lemonade/pull/1962

sofiageo:
@jeremyfowers please review

src/cpp/server/model_manager.cpp
jeremyfowers: Fixed!

jeremyfowers (approved):
Works great! Confirmed that it includes the mmproj file size :)

Pushed a 1 line change to https://github.com/lemonade-sdk/lemonade/pull/1888/commits/9902069674b1a0d3f3f2d11e25d5b36f5455afc9 to ensure that we would have this PR in the 10.6 release.

PR #1894, opened 2026-05-14
fix: rpm builds not working on fedora 43

### LINKED ISSUE #1753: rpm 10.3 installation failed due to missing dependencies
### Platform

Linux/Fedora

### Lemonade Version

10.3

### GPU / APU Model

amd strix halo 

### Component

Other / Not sure

### Bug Description

lemonade-server 10.3 rpm has dependencies which cannot be satisfied on regular fedora 43. 

I can work with 10.2 just fine.

### Steps to Reproduce

$ sudo rpm -Uvh lemonade-server-10.3.0.x86_64.rpm 
error: Failed dependencies:
        libm.so.6(GLIBC_2.43)(64bit) is needed by lemonade-server-10.3.0-1.x86_64
        libwebsockets.so.21()(64bit) is needed by lemonade-server-10.3.0-1.x86_64

I get this on fully updated fedora 43.

_No response_

### Log Output

```shell

```

### Additional Context

_No response_

lucifer-vali:
**Fixes #1753**

## Issue
RPM builds not working on older Fedora release.

## Explanation
- Container images for Arch and Fedora are pinned to latest release.
- Fedora 44 was released a couple of weeks ago, due to which rpm built on the latest image broke compatibility for older releases.
- Fedora, unlike Arch isn't truly a rolling release so image built on a newer release wouldn't always work on older release

## Chages
Pinned image versions to 43 for compatibility.

## Additionally
If even older releases are to be supported the version could be bumped down further since fedora does maintain compatibility for a few releases.

### DIFF
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -32,7 +32,7 @@ jobs:
             image: debian:trixie
             python: python3
           - distro: Fedora
-            image: fedora:latest
+            image: fedora:43
             python: python3
           - distro: openSUSE
             image: opensuse/tumbleweed:latest
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -278,7 +278,7 @@ jobs:
     name: Build Lemonade .rpm Package
     runs-on: ubuntu-latest
     container:
-      image: fedora:latest
+      image: fedora:43
     outputs:
       version: ${{ steps.get_version.outputs.version }}
     steps:
@@ -899,7 +899,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: build-lemonade-rpm
     container:
-      image: fedora:latest
+      image: fedora:43
     env:
       LEMONADE_VERSION: ${{ needs.build-lemonade-rpm.outputs.version }}
     steps:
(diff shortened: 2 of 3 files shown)


.github/workflows/linux_distro_builds.yml
```diff
@@ -32,7 +32,7 @@ jobs:
             image: debian:trixie
             python: python3
           - distro: Fedora
-            image: fedora:latest
+            image: fedora:43
```
jeremyfowers, line 35: Should we leave this one on `latest`, or create a matrix of `[latest, 43]`? I don't want us to accidentally drop Fedora 44 support.

lucifer-vali: I just checked the new rpm from the artifacts, unfortunately it does not work with Fedora 44.
GLIBC does not give an error but libwebsockets still gives a SONAME error.

The libwebsockets could to be linked as static binary for a solution because it would no longer be a shared library.

jeremyfowers: Glad I asked! Please ping on discord when you have a solution and builds are showing that both Fedora 43 and 44 are working.

superm1:
Two other alternatives:

1) Build **both** 43 and 44, have RPM artifacts for each.
2) Get a COPR repo and publish there and build for 43,44,rawhide etc.  This is like what we do for Launchpad.

PR #1897, opened 2026-05-15
Native Prometheus Metrics Endpoint

### LINKED ISSUE #1714: Prometheus-Grafana integration to Lemonade
### Feature Description

Prometheus is a FOSS monitoring and alerting tool widely used by the self-hosting community, commonly paired with Grafana for visualization. We want to add a Prometheus integration to Lemonade to aggregate hardware and Lemonade-specific statistics (ie. CPU, RAM, token/s). 

As an additional feature, we can also look into building a customizable, default Grafana dashboard and integrate it into the current Lemonade dashboard.

We will write documentation/blog post on how to use these features.

[sbates130272](https://github.com/sbates130272) has already done some work on this in #996

### Use Case / Motivation

Running models locally is resource intensive, so good metrics aggregation and alerting mechanisms are must-haves. A Lemonade Prometheus integration would allow users to integrate Lemonade statistics to their existing observability stack seamlessly, greatly benefiting the self-hosting community.

The default Grafana dashboard can also enhance the dashboard user experience, especially for users who don't have their own Grafana instance running.

### Platform Relevance

All platforms

### Additional Context

_No response_

Storce:
Resolves #1714

Native Prometheus-compatible metrics endpoint for Lemonade at `/metrics`, building off @sbates130272's work!

Exposes metrics directly supported by Lemonade, and also forwards llama.cpp's `/metrics` endpoint. A complete list of metrics can be found under my changes to `docs/api/lemonade.md`. 

Tests have been added under /tests directory, along with the `prometheus-grafana/` directory, where I attached the testing Prometheus-Grafana setup, along with a pre-built Grafana dashboard. Test setup on my MacOS:
image

*System utilization metrics are not directly compatible with MacOS. It would be nice if someone can test it on AMD hardware


Storce:
Moved the testing setup and demo Grafana to my personal repo at https://github.com/Storce/lemonade-prometheus-demo

sbates130272:
@Storce I will test this out! Thanks. Assuming this PR gets merged will you publish a Grafana JSON dashboard to the Grafana hosted library?

docs/api/lemonade.md
```diff
@@ -527,6 +528,107 @@ curl http://localhost:13305/v1/stats
...
+lemonade_server_info{version="10.4.0"} 1
+```
+
+Content type:
+
+```text
+text/plain; version=0.0.4; charset=utf-8
+```
+
+### Lemonade Metric Families
```
sbates130272, line 783: @Storce i would suggest just pointing the user at the source code. Trying to keep this markdown in sync with the metrics seems impractical and a maintainability nightmare. Especially as new metrics come online.

test/server_endpoints.py
```diff
@@ -21,6 +21,7 @@
 """
 
 import platform
+import re
```
sbates130272: @Storce should we use the Python client library for
Prometheus exporters here and leverage their tools to validate our schema?

docs/api/lemonade.md
Storce: Okay! Done

test/server_endpoints.py
Storce: Good point, will do

Storce:
Will publish the Grafana Dashboard once merged! Will probably also write a community guide on how to use it.

Storce:
Updates:
1. Add comments over existing endpoint to compensate for the lack of direct documentation
2. Added more detailed existing endpoint list to metrics.md at https://github.com/Storce/lemonade-prometheus-demo to. Modified some grafana dashboard details too

src/cpp/include/lemon/streaming_proxy.h
```diff
@@ -29,6 +31,10 @@ class StreamingProxy {
                           << time_to_first_token << std::endl;
                 LOG(INFO, "Telemetry") << "TPS:           " << std::fixed << std::setprecision(2)
                           << tokens_per_second << std::endl;
+                if (!decode_token_times.empty()) {
+                    LOG(INFO, "Telemetry") << "Decode times:  " << decode_token_times.size()
```
sawansri: The per token decode time metric is a good idea but imo we should leave it out of scope for this PR. 

We can discuss further improvements/useful metrics to log in another issue or PR. There's a lot of good work already done here so let's get the basics done right and then we can brainstorm ideas to improve in more follow ups.

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -303,6 +303,7 @@ void LlamaCppServer::load(const std::string& model_name,
 
     push_arg(args, reserved_flags, "--port", std::to_string(port_));
     push_arg(args, reserved_flags, "--jinja", std::vector<std::string>{"--no-jinja"});
+    push_arg(args, reserved_flags, "--metrics");
```
sawansri, line 330: One of my objectives is to reduce the amount of reserved llamacpp flags lemonade keeps. For now, this change is fine but just a heads up, I will probably introduce a PR removing a lot of these and implementing a better default system.

src/cpp/server/streaming_proxy.cpp
sawansri, line 1: Like I mentioned earlier, let's leave the decode token timing out of this PR, we can open follow ups adding more granular stats.

sawansri (commented):
I love the direction of this feature! Left a few comments, I think adding this to Lemonade would be really useful.

src/cpp/server/router.cpp
```diff
@@ -808,6 +808,52 @@ json Router::get_stats() const {
...
+        {"output_tokens", 0},
+        {"prompt_tokens", 0}
+    };
+
+    uint64_t request_count_total = 0;
+    uint64_t input_tokens_total = 0;
+    uint64_t output_tokens_total = 0;
+    uint64_t prompt_tokens_total = 0;
+
+    for (const auto& server : loaded_servers_) {
```
sawansri: What happens as servers/processes get loaded and unloaded? I think we should introduce some persistency here, maybe keep some limited stats like # of inference requests, output tokens, input, etc in some kind of temp file that you could periodically flush to disk and store in a json file.

sawansri:
@claude review

src/cpp/server/server.cpp
sawansri, line 1: Since this is quite a large `server.cpp` diff, perhaps we could pull out prometheus changes to its own file?

src/cpp/server/backends/llamacpp_server.cpp
Storce: Alright

src/cpp/include/lemon/streaming_proxy.h
Storce: yeah per token decode is kind of convoluted, I will remove it

src/cpp/server/router.cpp
Storce: Thanks for pointing that out! The Prometheus counters currently live in WrappedServer instances and would be lost when we unload models. I will move the counter to Router level so it always persists across models.

However thinking about persistence, we would be looking at ways to persist data across lemond restarts, which ties into more complex state management problems (aka how/when to wipe state? when/where to store state? debug instructions? multiple lemond instances? etc.). For the time being it might be cleaner to just assume that total token is unique across lemond instances and reset upon restarts. What do you think?

Storce:
Addressed the main issues pointed out by Claude:
- src/cpp/server/prometheus_metrics.cpp: stopped emitting lemonade_memory_used_gb on macOS
- src/cpp/include/lemon/wrapped_server.h: made set_telemetry() a pure setter again
- src/cpp/server/router.cpp: changed non-streaming telemetry updates to call record_telemetry(), so counter increments are explicit

Storce:
Updates:
- Move inference telemetry ownership from WrappedServer instances into Router so metrics do not get wiped upon backend unloads. Track both aggregate totals and per-model counters keyed by model.
- Not implementing counter persistence across lemond restart. Similar tools like vLLM also reset Prometheus counter to zero across restarts. The standard is to assume that counter persistence is implemented on the metric aggregation side.

sawansri (approved):
Works for me!

Image

sbates130272:
image

Looks good to me!

sawansri:
> Looks good to me!

Awesome, just waiting on some CI fixes before merging.

sbates130272:
@Storce @sawansri  I closed my original draft PR (#966) to keep things tidy. Can't wait to see this in a released version of lemonade soon!

PR #1899, opened 2026-05-15
Canonical model IDs with friendly-name precedence

jeremyfowers:
## Summary

- Every model now has a canonical ID (`user.NAME` / `extra.NAME` / `builtin.NAME`). The bare name `NAME` is an alias resolved by precedence: registered > imported > built-in.
- `/v1/models`, `/v1/models/{id}`, Ollama `/api/tags`, and `lemonade list` emit the bare name for the precedence-winner and the canonical-prefixed ID for shadowed sources, so no model is ever hidden because another source shares its name.
- `extra.*` filesystem-discovered entries are no longer silently dropped when they collide with a `user.*` entry; the GUI shows them with a `(imported)` suffix, the CLI shows them under their `extra.*` canonical id (copy-paste safe).
- `recipe_options.json` migrates legacy bare keys to `builtin.NAME` on startup (one-shot, logged at INFO).
- The deprecated `appear-builtin` label feature is removed; the new precedence rule subsumes it.
- `/pull` and `lemonade import` now reject `extra.*` / `builtin.*` model names (reserved prefixes).
- Tauri renderer applies a `(registered)` / `(imported)` / `(builtin)` display suffix to shadowed rows.
- Documentation: new "Model naming spec" section in `docs/guide/configuration/custom-models.md` with the precedence rule, emission rule, and five reference cases.

## Demo

GUI shows user, extra, and builtin models in harmony as long as there are no collisions. The `(builtin)` suffix elegant resolves the one collision we have:

image 

Streamlined CLI pull experience:

image

Registered models show in `lemonade list` without prefixes:

image

No `user.` visible in Lemonade Launch:

image

Nor in claude code itself:

image

 

## Test plan

- [x] `cmake --build --preset default` — clean build on macOS arm64
- [x] `python test/server_endpoints.py` — 38/38 pass (1 skipped)
- [x] `python test/server_cli.py` — 8/8 pass
- [x] `python test/server_cli2.py` — 54/54 pass (4 skipped)
- [x] `python test/test_ollama.py` — 27/27 pass (2 skipped)
- [x] New tests cover the four most-important reference cases (unique built-in, user-shadows-builtin, three-way input resolution, unique registered) plus reserved-prefix rejection
- [x] Manual: seeded a `recipe_options.json` with a bare key matching a built-in; restart rewrites it to `builtin.<name>` and the INFO log line is emitted; value applies on load
- [x] Linux build/test (CI)
- [x] Windows build/test (CI)
- [x] Spot-check the desktop app to verify shadowed rows render with the correct suffix

## Notes

- The `appear-builtin` label is removed but existing `"labels": ["appear-builtin"]` entries in users' `user_models.json` are left untouched as dead data — no automatic rewrite of user-owned files.
- The CLI deliberately does **not** apply the display-suffix transform so that copy-paste between `lemonade list` and `lemonade load` / `delete` / `run` always works. The canonical prefix on shadowed rows conveys source.
- Downgrade caveat: an older Lemonade binary won't apply settings keyed under `builtin.NAME` until those keys are reverted to bare names. Logged at INFO during migration.

### DIFF
--- a/docs/guide/configuration/custom-models.md
+++ b/docs/guide/configuration/custom-models.md
@@ -29,6 +29,54 @@ Custom model configuration involves two files, both located in the Lemonade cach
 
 See [configuration.md](./README.md) for more information about finding the cache directory.
 
+## Model naming spec
+
+Lemonade tracks three sources of models. Every model has a **canonical ID** of the form `<source>.<bare-name>`:
+
+| Canonical ID    | Source                                                                   |
+|-----------------|--------------------------------------------------------------------------|
+| `user.NAME`     | Model registered via `lemonade pull` (entry in `user_models.json`)       |
+| `extra.NAME`    | Model imported by dropping a GGUF in `--extra-models-dir`                |
+| `builtin.NAME`  | Model compiled into Lemonade's built-in catalog (`server_models.json`)   |
+
+The **bare name** `NAME` is an alias that always resolves to whichever source wins precedence for that name. Precedence is **registered > imported > built-in**.
+
+### What the API emits
+
+`/v1/models`, `/v1/models/{id}`, `lemonade list`, and the Ollama `/api/tags` endpoint emit each model with an `id` set to either:
+
+- the **bare name** if the model is the precedence-winner for its bare name, or
+- the **canonical-prefixed ID** if another source outranks it on the same bare name.
+
+For each bare name with collisions, the response contains one bare row plus one canonical-prefixed row per shadowed source.
+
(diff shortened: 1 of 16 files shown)


jeremyfowers:
things to check on:
- [x] When I `lemonade pull CHECKPOINT` I shouldn't see "user." in the naming suggestion
- [x] Same in the Manually Add a Model flow in the GUI

bitgamma (approved):
lgtm

PR #1904, opened 2026-05-15
feat: added model name copy button

### LINKED ISSUE #1903: Add a copy button next to the model name in the Model Options menu
Or just have it so that clicking the model name copies it to clipboard.

This would really help if I need to tell external software a model name.

Image

fl0rianr:
## Summary

Adds a small copy button next to the model name in the Model Options modal.

closes #1903.

## Changes

- Add model-name copy action using the existing clipboard utility
- Show a temporary copied state after successful copy
- Add minimal, locally scoped styles for the copy button

## Testing

- Verified the Model Options modal opens correctly
- Verified clicking the copy button copies the model name to the clipboard
- Verified the copied state resets after a short delay
image

Right after copying for a few seconds:

image

### DIFF
--- a/src/app/styles/styles.css
+++ b/src/app/styles/styles.css
@@ -1807,6 +1807,46 @@ footer {
     text-transform: none;
 }
 
+.model-options-name-row {
+    display: inline-flex;
+    align-items: center;
+    gap: 6px;
+}
+
+.model-options-copy-button {
+    width: 18px;
+    height: 18px;
+    padding: 2px;
+    border: none;
+    border-radius: 4px;
+    background: transparent;
+    color: var(--text-secondary);
+    cursor: pointer;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    vertical-align: middle;
+    transition: color 0.15s ease, background-color 0.15s ease;
+}
+
+.model-options-copy-button:hover {
+    color: var(--text-primary);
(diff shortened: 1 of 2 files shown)


src/app/styles/styles.css
```diff
@@ -1807,6 +1807,46 @@ footer {
...
 }
 
+.model-options-name-row {
+    display: inline-flex;
+    align-items: center;
+    gap: 6px;
+}
+
+.model-options-copy-button {
+    width: 18px;
```
sofiageo: Removing width, height, padding, display: inline-flex, align-items: center, justify-content: center, vertical-align-middle

makes this button better aligned

sofiageo (changes requested):
Looks good, but check the comments for better alignment. I know it's better with a padding but having it better aligned and taking less space I think is more important.

sofiageo:
By the way I only made these comments because it looks different on my PC from your screenshot. I'm not sure if you changed something or there is something else going on.

image

src/app/styles/styles.css
fl0rianr: Thanks, makes sense. Done.

jeremyfowers:
FYI Webkit vs. Chromium engines can make a difference in things like padding.

sofiageo (approved):
For me now it works great both in tauri app and in Firefox web app.

PR #1906, opened 2026-05-15
GUI: render canonical display names in LLM Chat model dropdown

jeremyfowers:
## Summary
- The Active Models row in the Model Management panel rendered shadowed canonical ids (`user.*`, `extra.*`, `builtin.*`) as `Foo (registered)` / `(imported)` / `(builtin)`, but the LLM Chat panel's model dropdown showed the raw id. The dropdown trigger and each option now render the same display string as Active Models.
- Extracted `getModelDisplayName` and `CANONICAL_PREFIXES` from `ModelManager.tsx` into a new shared util at `src/app/src/renderer/utils/modelDisplayName.ts`. `ModelManager.tsx` and `components/ModelSelector.tsx` both import from it.
- The dropdown's text search now matches against both the raw id and the rendered display string, so typing e.g. `registered` surfaces user-registered models.
- Hover tooltips on the dropdown trigger and options still show the raw id, matching the Active Models tooltip.

## Demo

image

image


sofiageo (approved):
Tested with 2 collisions and without. It's working fine.

PR #1909, opened 2026-05-15
feat(app): add full custom sd.cpp model options

fl0rianr:
## Summary

Expands manual custom model registration to support backend-reported recipes and optional StableDiffusion.cpp component checkpoints.

## Changes

- Populate the Add Model recipe dropdown from supported backend recipes
- Reuse shared recipe display names
- Add optional text encoder and VAE checkpoint fields for StableDiffusion.cpp models
- Pass named component checkpoints through the existing custom model install flow
- Support resolving sd.cpp GGUF variants such as `CHECKPOINT:Q8_0` to the downloaded `.gguf` file

## Testing

- Verified the Add Model dialog still defaults to Llama.cpp
- Verified StableDiffusion.cpp appears as a selectable recipe when supported
- Verified a custom sd.cpp model can be installed with main, text encoder, and VAE checkpoints
- Verified existing cached component files are reused during install
- Verified sd.cpp GGUF variant checkpoints resolve to the downloaded GGUF file

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1039,17 +1039,8 @@ std::string ModelManager::resolve_model_path(const ModelInfo& info, const std::s
         return all_bin_files[0];
     }
 
-    // For GGUF-backed recipes, find the GGUF file with advanced sharded model support.
-    // sd-cpp supports GGUF checkpoints too; those can be registered with a quant alias
-    // such as CHECKPOINT:Q8_0, and download_from_huggingface() resolves that alias to
-    // the concrete filename. Use the same resolver here so load finds the downloaded file.
-    auto is_direct_non_gguf_variant = [](const std::string& value) {
-        return ends_with_ignore_case(value, ".safetensors");
-    };
-    const bool should_resolve_gguf = type == "main" &&
-        (info.recipe == "llamacpp" ||
-         (info.recipe == "sd-cpp" && !is_direct_non_gguf_variant(variant)));
-    if (should_resolve_gguf) {
+    // For llamacpp, find the GGUF file with advanced sharded model support
+    if (info.recipe == "llamacpp" && type == "main") {
         if (!safe_exists(model_cache_path_fs)) {
             return model_cache_path;  // Return directory path even if not found
         }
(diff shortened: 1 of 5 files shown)


jeremyfowers:
@bitgamma would you like to review this one?

src/cpp/server/model_manager.cpp
```diff
@@ -1039,8 +1039,19 @@ std::string ModelManager::resolve_model_path(const ModelInfo& info, const std::s
...
 
-    // For llamacpp, find the GGUF file with advanced sharded model support
-    if (info.recipe == "llamacpp" && type == "main") {
+    // For GGUF-backed recipes, find the GGUF file with advanced sharded model support.
+    // sd-cpp supports GGUF checkpoints too; those can be registered with a quant alias
+    // such as CHECKPOINT:Q8_0, and download_from_huggingface() resolves that alias to
+    // the concrete filename. Use the same resolver here so load finds the downloaded file.
+    auto is_direct_non_gguf_variant = [](const std::string& value) {
+        return ends_with_ignore_case(value, ".safetensors") ||
+               ends_with_ignore_case(value, ".pth") ||
```
bitgamma: .pth and .ckpt files are outdated and potentially unsafe. While I don't know if stable-diffusion.cpp has measures to mitigate the risks, I would avoid supporting them.

bitgamma (changes requested):

src/cpp/server/model_manager.cpp
fl0rianr: Addressed, thanks. The sd.cpp non-GGUF bypass now only treats `.safetensors` as a direct model file.

**NEW:** Also preserves existing named checkpoints when re-registering the same user model. This prevents partial clients or older flows from dropping sd.cpp component checkpoints such as `text_encoder` and `vae`, which can break image model loading after restart. Potentially related to #1761. Since I had such an issue as well this makes it more stable. I can remove it again if requested.

bitgamma:
Screenshot From 2026-05-16 14-59-23

@fl0rianr nice job! it works! but I see several issues:

- The text-encoder and vae field do not have a separate label each. This is a problem because as soon as I copy something in the text field (for example user/repo:) I don't know anymore which is which and also makes the layout look strange.
- mmproj, embedding, reranking, vision should disappear when selecting StableDiffusion.cpp recipe
- ideally the placeholder text should indicate the correct input format for image checkpoints

Another question, could I have filled out those fields in a shorter way? because I tried just writing the file name after the `:` but with no results. 

Since I think most image models require specifying the full path relative to the repo anyway, I would revert all changes made to `model_manager.cpp` and teach the user that for image models you need the full path, so they don't have to wonder whether the shorthand format will work or not.

fl0rianr:
Thanks @bitgamma! I agree. I simplifyed this by reverting the model_manager.cpp changes and making the sd.cpp custom model flow require the full file path relative to the Hugging Face repo.

That should avoid the ambiguity around shorthand variants like `:Q8_0`, especially since image models often need paths under subfolders such as `split_files/text_encoders/...`.

I’ also updated the UI so each sd.cpp component field has its own label, hide the unrelated mmproj / embedding / reranking / vision fields for StableDiffusion.cpp, and make the placeholders show the expected full `repo:path/to/file` format.

For the image models it looks like this now:
image

But I generally added recipe examples so you don't get confused to add GGUF to vLLM:
image

bitgamma (approved):
looks really good to me now! Thanks for your contribution!

PR #1915, opened 2026-05-16
Introduce per-backend args

bitgamma:
Example:

`lemonade config set llamacpp.rocm_args="-b 4096 -ub 2048" llamacpp.vulkan_args="-b 1024 -ub 512"`

when loading a model with ROCm, the ROCm-specific args will be merged, when loading with Vulkan, those will be merged. If no backend-specific args are defined (etc), fallback to unprefixed "args". Works for all recipes that have multiple backends.

I also fixed indentation of log lines throughout `router.cpp` since many were misindented for some reasons.

### DIFF
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -38,9 +38,9 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
   "llamacpp": {
     "backend": "auto",
     "args": "",
-    "vulkan_args",
-    "rocm_args",
-    "cpu_args",
+    "vulkan_args": "",
+    "rocm_args": "",
+    "cpu_args": "",
 	"device": "",
     "prefer_system": false,
     "rocm_bin": "builtin",
@@ -108,7 +108,7 @@ Backend-specific settings are nested under their backend name:
 | Key | Default | Description |
 |-----|---------|-------------|
 | `backend` | "auto" | Backend to use: "auto" means "choose for me" |
-| `args` | "" | Custom arguments to pass to llama-server (fallback) |
+| `args` | "" | Custom arguments to pass to llama-server (fallback, unused when backend-specific args defined) |
 | `*_args` | "" | Backend-specific custom arguments to pass to llama-server |
 | `device` | "" | Comma-separated list of devices to use for offloading. Empty is auto. |
 | `prefer_system` | false | Prefer system-installed llama.cpp over bundled |
@@ -118,7 +118,7 @@ Backend-specific settings are nested under their backend name:
 | Key | Default | Description |
 |-----|---------|-------------|
 | `backend` | "auto" | Backend to use: "auto" means "choose for me" |
-| `args` | "" | Custom arguments to pass to whisper-server (fallback) |
+| `args` | "" | Custom arguments to pass to whisper-server (fallback, unused when backend-specific args defined) |
 | `*_args` | "" | Backend-specific custom arguments to pass to whisper-server |
 | `*_bin` | "builtin" | Backend binary selection — see [Backend binary selection](#backend-binary-selection) |
 
@@ -126,7 +126,7 @@ Backend-specific settings are nested under their backend name:
 | Key | Default | Description |
 |-----|---------|-------------|
 | `backend` | "auto" | Backend to use: "auto" means "choose for me" |
-| `args` | "" | Custom arguments to pass to `sd-server` (fallback) |
+| `args` | "" | Custom arguments to pass to `sd-server` (fallback, unused when backend-specific args defined) |
 | `*_args` | "" | Backend-specific custom arguments to pass to `sd-server` |
 | `steps` | 20 | Number of inference steps |
 | `cfg_scale` | 7.0 | Classifier-free guidance scale |
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -20,7 +20,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
 
 ```json
 {
-  "config_version": 1,
+  "config_version": 2,
   "port": 13305,
   "host": "localhost",
   "log_level": "info",
@@ -38,6 +38,9 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
   "llamacpp": {
     "backend": "auto",
     "args": "",
+    "vulkan_args",
+    "rocm_args",
+    "cpu_args",
 	"device": "",
     "prefer_system": false,
     "rocm_bin": "builtin",
@@ -47,12 +50,17 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
   "whispercpp": {
     "backend": "auto",
     "args": "",
+    "cpu_args": "",
+    "npu_args": "",
     "cpu_bin": "builtin",
     "npu_bin": "builtin"
   },
   "sdcpp": {
     "backend": "auto",
     "args": "",
+    "cpu_args": "",
+    "rocm_args": "",
+    "vulkan_args": "",
     "steps": 20,
     "cfg_scale": 7.0,
     "width": 512,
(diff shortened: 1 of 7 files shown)


src/cpp/server/runtime_config.cpp
```diff
@@ -286,23 +286,37 @@ json RuntimeConfig::recipe_options() const {
...
+    const std::string backend_args = backend + "_args";
+
     if (config_.contains("llamacpp")) {
         const auto& lc = config_["llamacpp"];
         if (lc.contains("backend")) result["llamacpp_backend"] = resolve_auto(lc["backend"]);
-        if (lc.contains("args")) result["llamacpp_args"] = lc["args"];
+        if (lc.contains(backend_args)) {
+            result["llamacpp_args"] = lc[backend_args];
+        } else if (lc.contains("args")) {
+            result["llamacpp_args"] = lc["args"];
```
fl0rianr, line 297: This sections (three in total) break fallback semantics when defaults contain empty backend-specific args, like it currently is.

After defaults are merged, cpu_args/rocm_args/vulkan_args always exist as `""`, so `contains(backend_args)` is true and the unprefixed `args` fallback is never used.

Backend-specific args should only win when they are non-empty.

src/cpp/server/router.cpp
```diff
@@ -222,14 +222,16 @@ void Router::load_model(const std::string& model_name,
                        bool do_not_upgrade,
                        bool allow_reload_on_option_change) {
     const std::string canonical_model_name = resolve_model_name(model_name);
-    RecipeOptions default_opt = RecipeOptions(model_info.recipe, config_->recipe_options());
+    const std::string backend_option = model_info.recipe + "_backend";
+    std::string backend = options.get_option(backend_option).get<std::string>();
+    backend = backend.empty() ? model_info.recipe_options.get_option(backend_option).get<std::string>() : backend;
+    RecipeOptions default_opt = RecipeOptions(model_info.recipe, config_->recipe_options(backend));
```
fl0rianr: This selects the backend for `config_->recipe_options(backend)` before computing the final effective options.

`options.get_option(backend_option)` can return a dynamic default even when the load request did not explicitly set a backend. In that case, config/model-level backend choices can be skipped when selecting backend-specific args.

I think the backend used for arg selection should be derived from a preliminary effective-options merge, then defaults should be rebuilt with that effective backend.

fl0rianr (changes requested):
Thanks @bitgamma for that PR, I like that feature as mentioned. But I found two critical points that should be improved first.

src/cpp/server/router.cpp
bitgamma: resolving as discussed on discord

src/cpp/server/runtime_config.cpp
bitgamma: done

fl0rianr (approved):

src/cpp/resources/defaults.json
```diff
@@ -1,5 +1,5 @@
 {
-  "config_version": 1,
+  "config_version": 2,
```
jeremyfowers: Why does this need to be version 2? I was only going to rev this number in the event of a schema-breaking change.

bitgamma: because I didn't know :D reverted

docs/guide/configuration/README.md
```diff
@@ -38,6 +38,9 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
   "llamacpp": {
     "backend": "auto",
     "args": "",
+    "vulkan_args",
+    "rocm_args",
+    "cpu_args",
```
jeremyfowers: (posted on behalf of @jeremyfowers)

These three entries are missing both `:` and a value, so this JSON example block is no longer valid. Should be:

```json
"vulkan_args": "",
"rocm_args": "",
"cpu_args": "",
```

The `whispercpp` and `sdcpp` blocks further down in this file were done correctly — only this `llamacpp` block has the syntax issue.

src/cpp/server/router.cpp
jeremyfowers: (posted on behalf of @jeremyfowers)

This resolves the backend from only the **load-time** override and the **model-level** override. A user who pins `llamacpp.backend = "rocm"` in `config.json` (with no per-load or per-model backend) gets `backend = ""` here, so `rocm_args` is silently ignored and the code falls back to `args`.

This is the same gap @fl0rianr raised earlier; the thread was marked "resolving as discussed on discord" but the code on disk still has it.

A two-pass merge resolves it cleanly:

```cpp
// First pass with empty backend so we can read the merged backend choice
RecipeOptions tentative = options.inherit(model_info.recipe_options.inherit(
    RecipeOptions(model_info.recipe, config_->recipe_options(""))));
const std::string backend = tentative.get_option(backend_option).get<std::string>();
// Second pass: rebuild defaults using the resolved backend
RecipeOptions default_opt = RecipeOptions(model_info.recipe, config_->recipe_options(backend));
RecipeOptions effective_options = options.inherit(model_info.recipe_options.inherit(default_opt));
```

If the discord conclusion was to accept this gap, can we at least add a comment here explaining the limitation?

src/cpp/server/runtime_config.cpp
```diff
@@ -286,23 +286,37 @@ json RuntimeConfig::recipe_options() const {
...
+
     if (config_.contains("llamacpp")) {
         const auto& lc = config_["llamacpp"];
         if (lc.contains("backend")) result["llamacpp_backend"] = resolve_auto(lc["backend"]);
-        if (lc.contains("args")) result["llamacpp_args"] = lc["args"];
+        if (lc.contains(backend_args) && lc[backend_args] != "") {
+            result["llamacpp_args"] = lc[backend_args];
+        } else if (lc.contains("args")) {
+            result["llamacpp_args"] = lc["args"];
+        }
```
jeremyfowers, line 298: (posted on behalf of @jeremyfowers)

The PR description says "the ROCm-specific args will be **merged**", but this is a *replacement* — if `rocm_args` is non-empty, the unprefixed `args` is dropped entirely. A user with global `args: "-c 4096"` and `rocm_args: "-fa on"` likely expects both flags on ROCm; they'll only get `-fa on`.

Either:
- Update the PR description (and the docs change in this PR that says `args` is the "fallback") to say "override" rather than "merged", or
- Concatenate `args + " " + backend_args` when both are non-empty.

I'd lean on concatenation — it matches the user mental model and lets `args` carry truly cross-backend flags. Whichever way you go, the same fix applies to the `whispercpp` and `sdcpp` blocks below.

bitgamma: I wouldn't concatenate args and backend args, they are a fallback. What I wanted to say in the documentation is that they are merged with model specific arguments, not between backend_args and plain_args

jeremyfowers: Gotcha. IMO you should add a note on configuration/README.md specifically that backends args replace generic args. So that its 100% clear what the behavior is to readers (and mainly their agents lol).

jeremyfowers (approved):
Good to merge when the comments are addressed. They were written by a robot but sounded reasonable enough to me; use your judgement!

docs/guide/configuration/README.md
bitgamma: fixed

src/cpp/server/router.cpp
bitgamma: addressed as suggested

src/cpp/server/runtime_config.cpp
bitgamma: done!

bitgamma:
@jeremyfowers addressed the comments as suggested. I wanted to avoid having to do two passes but it wasn't worth the additional code/complexity considering this only runs once per load and probably wastes a few milliseconds at most.

PR #1744, opened 2026-04-26
Feature/memory manager

fl0rianr:
## Summary

This PR adds memory-aware context targeting for Lemonade models.

The main goal is to replace the old fixed small context default with better automatic behavior: Lemonade now uses a configurable **Context Target**, performs a base-model memory preflight before loading, and then lets each backend load as much context as it can reasonably support.

### Context Target

- Adds a new global `context_target`, defaulting to `131072`.
- Per-model `ctx_size` is still supported as the stored legacy field, but is now interpreted as that model's individual Context Target.
- The UI presents this as **Context Target** instead of exposing the old fixed `ctx_size = 4096` behavior.
- When the model's maximum context length can be detected, the effective target is capped to that model limit by default.
- Explicit user overrides can request a target above the detected model limit. In that case Lemonade allows the request, lets the backend fit what it can, and reports a warning if the actually loaded context exceeds the detected model support.

### llama.cpp

- Uses llama.cpp dynamic fitting for normal LLM loading.
- Performs a base-model memory preflight before starting the backend. If the base model cannot fit, Lemonade fails before launching `llama-server`.
- Starts llama.cpp with `--ctx-size 0`, `--fit on`, and `--fit-ctx 1024` so llama.cpp can reduce the context before allocation when needed.
- Uses `--override-kv <arch>.context_length=int:<context_target>` to provide llama.cpp with a fit-adjustable starting target.
- Caps the target to the detected model context by default; going above model support is only treated as an explicit user override.
- Avoids probe-load / unload / reload loops.

### RAM limit

- Adds configurable `ram_limit`, default `-1`.
- `-1` means no artificial Lemonade RAM limit and remains the default behavior.
- When `ram_limit` is set, Lemonade treats it as an opt-in constrained-memory mode.
- In this mode Lemonade still performs the base-model preflight before launch.
- For llama.cpp, Lemonade passes an adjusted `--fit-target` so llama.cpp can fit the context inside the requested memory budget before allocation.
- In RAM-limit mode Lemonade disables the llama.cpp prompt cache with `--cache-ram 0` to avoid spending constrained host RAM on cache state. 
- This is intentionally a pre-launch cap because the external `llama-server` process cannot shrink an already-created context. But should be improved on a later point.

### Warnings for restricted context

- Lemonade now distinguishes resource-related context reduction from intentional small-context configurations.
- A warning that the loaded context is **above model support** is emitted only when:
  - the model limit is known,
  - the user explicitly requested a target above that limit,
[...]

PR #1919, opened 2026-05-17
fix: vLLM memory behavior on APUs

fl0rianr:
## Summary
- wire `vllm.backend` / `vllm.args` into runtime config
- default vLLM KV cache to 4G when no memory option is set

## Why
vLLM defaults to `gpu_memory_utilization=0.92`, which is unsafe on shared-memory/UMA systems that have a high RAM for GPU setup e.g. 120 / 128 GB on Strix Halo and can OOM the host even for the smallest models.

## Notes
Draft PR for discussion. Users can still override memory behavior with `--kv-cache-memory-bytes` or `--gpu-memory-utilization` in `vllm.args`.
Let us discuss what is the best option to handle memory for vLLM.

100% safe if you want to try out and had memory issues before. At least a 64 GB Ram system recommended.

### DIFF
--- a/src/cpp/resources/defaults.json
+++ b/src/cpp/resources/defaults.json
@@ -43,6 +43,10 @@
   "flm": {
     "args": ""
   },
+  "vllm": {
+    "backend": "auto",
+    "args": ""
+  },
   "ryzenai": {
     "server_bin": "builtin"
   },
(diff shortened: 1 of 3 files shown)


jeremyfowers:
fyi @ramkrishna2910

sawansri (approved):
LGTM! I think we should consider making the 4G kv cache allocation a dynamic limit rather than hardcoding the value (even though the user can override) especially since we plan on adding larger models (quantized of course). This is a good fix for now though.

fl0rianr:
Sure I will improve that in the future! PR #1744 will be revived once llama.cpp maintainers come around...

PR #1922, opened 2026-05-17
Switch to upstream builds for stable diffusion

superm1:


jeremyfowers:
Woohoo! Let's get proof of life on ROCm on Windows and Linux for a Flux model and then merge.

superm1:
We don't have that from CI?

jeremyfowers:
> We don't have that from CI?

Not for flux models on real image sizes. Maybe we should, I did this recently for llamacpp (run big models).

jeremyfowers (approved):
Runs on GPU on Windows!

Issue #1789, opened 2026-05-03
watchdog that auto-kills internal llama-server when it hangs


Rose22:
### Feature Description

hi! as i suggested in the discord earlier, this would be a very nice feature to have. llamacpp can sometimes be a bit unstable, especially on my system for some reason, and i often have to kill the process when it freezes. a watchdog that auto detects a crash and then force-restarts llamacpp would help a lot!

### Use Case / Motivation

in cases where llamacpp freezes (which may be hardware dependent), it ensures things proceed smoothly

### Platform Relevance

All platforms

leoai-81:
Happened to me a couple time already too.

PR #1924, opened 2026-05-17
fix(sd image): seed UX and drop generation timeout

fl0rianr:
## Summary

Fix long-running Stable Diffusion image requests timing out while sd-server is still actively generating.

SDServer currently forwards image generation/edit/variation requests using the global HTTP timeout.
With the default global timeout of 600 seconds, valid slow image jobs can fail with:

Network error: CURL error: Timeout was reached

even though stable-diffusion.cpp continues generating and the backend process remains alive.
So timeout is dropped here.

Additionally seed parameter UX improved/fixed.

**Changes:**

- Image request timeout to 0, which disables libcurl's overall CURLOPT_TIMEOUT.
- random seed is now possible
- deletion bug in the input panel solved.

Seed behavior

**Low Risk:**

Apply the image-specific timeout consistently to:
- image generations-
- image edits-
- image variations

Preserve existing backend error handling:
- backend HTTP errors still propagate normally
- network/connection errors still return network errors

## Seed behavior

This also clarifies the existing image seed behavior:

- `seed => 0`: pass the seed through for reproducible output
- `seed  -1`: gives a radom seed to sd-server

The UI/default seed remains `42`, preserving the current deterministic default behavior.
Users who want variable output can set the existing seed field to `-1`.
The placeholder -1 = random works now since due to solved deletion bug the input can be empty.

with current main it is unclear that with seed -1 actually 42 is used. The input is buggy.

This fix avoids adding another UI control and avoids making image editing accidentally
random by default, since default remains 42.

## Validation

- Verified the patch applies only to SDServer image request forwarding.
- Confirmed the change does not alter global timeout behavior for non-image requests.
- Seed behavior tested


jeremyfowers:
This timeout thing has been a constant headache. What if we just set the default global timeout to a much larger value like 3600? Or 0?

jeremyfowers:
PS. the reason I leave the comment above is that I'm just wary, in a semantic sense, of having a config named "global timeout" if it is not actually a global timeout.

fl0rianr:
I get you, but Image work is just a special job and does not really fit into a global setting. The 300s is useful if an LLM is hanging but better would be a proper watcher as discussed (if think on discord and we have an issue as well #1789) but not yet implemented.

jeremyfowers:
> I get you, but Image work is just a special job and does not really fit into a global setting. The 300s is useful if an LLM is hanging but better would be a proper watcher as discussed (if think on discord and we have an issue as well #1789) but not yet implemented.

Ok fair enough, I'll let someone who does more image gen than me review this one.

bitgamma:
I also would rather make the global timeout larger instead of making exceptions.  300s is way too little for many LLM especially if non-streaming mode or if reprocessing a huge context. Image generation is actually more predictable in timing.

fl0rianr:
@bitgamma in PR #2004 I'm doubling the timeout. But still 600s is not enough for images, they are a special workload. I can adapt this PR to use sd's prediction plus an margin as timeout.

bitgamma:
> @bitgamma in PR #2004 I'm doubling the timeout. But still 600s is not enough for images, they are a special workload. I can adapt this PR to use sd's prediction plus an margin as timeout.

600s is not enough? are you using cpu backend? Normal image generation should never be over 60 seconds actually (Z-Image, Anima, SDXL, Qwen-Image, Flux-Klein 9b) and normal generation time is 5-20secs for distilled models. I literally printed a book of images I generated using ComfyUI so I know those timing very well (on Strix Halo + ROCm).

In any case, 600s is still too little to reprocess ~200k context with Qwen3.6-27B, so it would make sense to at least double that once more

fl0rianr:
@bitgamma after discussion with @jeremyfowers I went with no timeout at all. Since timeouts are not a welcomed "feature" according to several users. 
While we keep the central change that you can customize seeds better, specifically use random seed.

src/cpp/server/backends/sd_server.cpp
```diff
@@ -471,6 +479,9 @@ json SDServer::image_generations(const json& request) {
     }
 
     json extra_args = build_extra_args(request);
+    if (extra_args.contains("seed")) {
+        sd_request["seed"] = extra_args["seed"];
```
jeremyfowers: _Posted on behalf of @jeremyfowers._

This top-level `sd_request["seed"]` is set only in `image_generations` — `image_edits` and `image_variations` rely solely on the seed embedded in the `<sd_cpp_extra_args>` prompt blob. If the embedded seed is what sd-server actually honors (and it must be, since edits/variations work without this), then this line is redundant. Please either drop it or set it consistently across all three endpoints so the seed handling isn't divergent.

src/app/src/renderer/components/panels/ImageGenerationPanel.tsx
```diff
@@ -576,7 +574,7 @@ const ImageGenerationPanel: React.FC<ImageGenerationPanelProps> = ({
           <div className="image-setting">
             <label>Seed</label>
             <input type="number" min="-1" value={imageSettings.seed}
-              onChange={(e) => setImageSettings(prev => ({ ...prev, seed: parseInt(e.target.value) || -1 }))}
+              onChange={(e) => setImageSettings(prev => ({ ...prev, seed: Math.max(parseInt(e.target.value) || -1, -1) }))}
```
jeremyfowers: _Posted on behalf of @jeremyfowers._

Seed `0` is unreachable here: `parseInt("0") || -1` evaluates to `-1` (0 is falsy), and `Math.max(-1, -1)` keeps it `-1`. So typing `0` silently becomes random, even though the backend now accepts `seed >= 0`. The `Math.max` clamp gives a false impression of handling `0`. Consider `Number.isNaN(parseInt(e.target.value)) ? -1 : Math.max(parseInt(e.target.value), -1)` so `0` is preserved.

fl0rianr:
Significant improvement on the seed UX behavior:
Review adressed and UI shows -1 is the random seed trigger if the field is empty. Additionally fixed the bug in main if -1 was set the entry could not be deleted only overwritten by typing a new number.
Tested and works fine.

src/cpp/server/backends/sd_server.cpp
```diff
@@ -17,6 +17,7 @@
 #include <random>
 #include <sstream>
 #include <set>
+#include <random>
```
jeremyfowers:
```suggestion
```

Duplicate

jeremyfowers (approved):
Let's go! Thanks for iterating through the feedback.

PR #1933, opened 2026-05-17
Preserve backend error status

neubig:
## Summary
- Preserve wrapped-backend JSON error payloads instead of replacing them with generic 500s.
- Return preserved backend HTTP status codes from chat completions, completions, and responses handlers (non-streaming paths only; streaming SSE error propagation is out of scope).
- Convert llama.cpp context-window failures into OpenAI-shaped `invalid_request_error` responses with `code=context_length_exceeded`, while keeping backend details for debugging.
- Add a system-backend regression path that can mock llama.cpp non-2xx error responses.

## Server Validation
- Started `build/lemond --host 127.0.0.1 --port 13305` from this branch.
- Loaded cached `Qwen3.6-35B-A3B-GGUF` successfully through `POST /api/v1/load`.
- Sent an over-context `POST /api/v1/chat/completions`; Lemonade returned HTTP `400` with `type=invalid_request_error`, `code=context_length_exceeded`, `status_code=400`, and preserved llama.cpp backend details including `n_ctx=65536` and `n_prompt_tokens=80010`.

## Tests
- `python3 -m py_compile test/test_llamacpp_system_backend.py`
- `git diff --check`

## Not Completed
- Full `test/test_llamacpp_system_backend.py` with `build/lemonade-server` starts the server but fails before the new regression on existing system-backend availability expectations: this host reports the dummy `llama-server` as `unsupported` because system llama.cpp requires the HIP plugin on AMD GPU hosts.

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -112,6 +112,42 @@ bool prepend_no_think_to_last_user_message(json& request_json) {
     return false;
 }
 
+bool valid_error_status(int status_code) {
+    return status_code >= 400 && status_code <= 599;
+}
+
+int get_error_status_code(const json& response, int default_status_code = 500) {
+    if (!response.contains("error") || !response["error"].is_object()) {
+        return default_status_code;
+    }
+
+    const auto& error = response["error"];
+    if (error.contains("status_code") && error["status_code"].is_number_integer()) {
+        int status_code = error["status_code"].get<int>();
+        if (valid_error_status(status_code)) {
+            return status_code;
+        }
+    }
+
+    if (error.contains("details") && error["details"].is_object()) {
+        const auto& details = error["details"];
+        if (details.contains("status_code") && details["status_code"].is_number_integer()) {
+            int status_code = details["status_code"].get<int>();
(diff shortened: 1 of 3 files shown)


fl0rianr (commented):
Thanks, this looks good for the non-streaming paths.

Two small concerns:

1. The PR description sounds broader than the implementation. `stream: true` still seems unaffected: the handlers enter the SSE path before backend status is known, and StreamingProxy only logs non-200 backend responses and closes the stream. Can you confirm whether streaming error propagation is intentionally out of scope for this PR?

2. Test hygiene nit: please clear `MOCK_LLAMA_ERROR_STATUS` and `MOCK_LLAMA_ERROR_RESPONSE` in `setUp()`, alongside `MOCK_LLAMA_REQUEST_PATH`, so external env or failed tests cannot leak into later server starts.

fl0rianr (approved):
Thanks for slightly adapting it @jeremyfowers and thanks for the nice PR @neubig looks good, we can go now as-is.

jeremyfowers:
Thanks for the review @fl0rianr !

neubig:
Thanks both!

PR #1944, opened 2026-05-18
feat: built-in MTP support

jeremyfowers:
- Adds an `mtp` label that indicates llama-server should be launched with mtp enablement args
- Adds Qwen3.6 MTP models to the registry

### DIFF
--- a/docs/api/openai.md
+++ b/docs/api/openai.md
@@ -1001,6 +1001,12 @@ Labels describe what a model can do. A model may carry multiple labels.
 |-------|-------------|
 | `realtime-transcription` | Supports the WebSocket `/realtime` endpoint for live microphone transcription. |
 
+**Runtime labels** — affect backend launch defaults:
+
+| Label | Description |
+|-------|-------------|
+| `mtp` | Enables llama.cpp MTP draft decoding defaults (`--spec-type draft-mtp --spec-draft-n-max 6`); users can override these with `llamacpp_args`. |
+
 **Characteristic labels** — informational, do not affect routing:
 
 | Label | Description |
--- a/docs/api/openai.md
+++ b/docs/api/openai.md
@@ -1005,7 +1005,7 @@ Labels describe what a model can do. A model may carry multiple labels.
 
 | Label | Description |
 |-------|-------------|
-| `mtp` | Enables llama.cpp MTP draft decoding defaults (`--spec-type draft-mtp --spec-draft-n-max 6`); users can override these with `llamacpp_args`. |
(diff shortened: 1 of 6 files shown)


docs/api/openai.md
```diff
@@ -1001,6 +1001,12 @@ Labels describe what a model can do. A model may carry multiple labels.
 |-------|-------------|
 | `realtime-transcription` | Supports the WebSocket `/realtime` endpoint for live microphone transcription. |
 
+**Runtime labels** — affect backend launch defaults:
+
+| Label | Description |
+|-------|-------------|
+| `mtp` | Enables llama.cpp MTP draft decoding defaults (`--spec-type draft-mtp --spec-draft-n-max 6`); users can override these with `llamacpp_args`. |
```
bitgamma: the exact number to use is really open for the debate. Many say 2 is the good one, others say 3 and finally on my Strix Halo 4 is most consistent (also found some benchmark saying the same). Unsloth says 6.

All this to say that maybe 6 isn't the safest default, maybe 2-3 could be on the safer side.

Also seems to be the consensus that `--spec-draft-p-min 0.75` is also needed

bitgamma (approved):
looks great! Just a comment about the defaults

docs/api/openai.md
jeremyfowers: Pushed updated settings [5b9c90e](https://github.com/lemonade-sdk/lemonade/pull/1944/commits/5b9c90e5c01654e7be750b6edfa23c4e2af73c28)

sofiageo:
Are the checkpoints correct? Shouldn't they be `unsloth/Qwen3.6-35B-A3B-MTP-GGUF:UD-Q4_K_XL` instead? Or maybe it's the same.

jeremyfowers:
> Are the checkpoints correct? Shouldn't they be `unsloth/Qwen3.6-35B-A3B-MTP-GGUF:UD-Q4_K_XL` instead? Or maybe it's the same.

That might be more elegant but exact filename is also allowed.

sofiageo:
OK let me actually test one and I'll be back asap

sofiageo (approved):
Looks good. I only have two comments. Do we still need the non MTP models? and maybe we can add Qwen3.5-122B-A10B-MTP-GGUF too.

jeremyfowers:
> Do we still need the non MTP models?

I thought the non-MTP was better in some scenarios? Either way we can't delete them (that would be a breaking API change) but in the future we can `suggested: false` them.

> and maybe we can add Qwen3.5-122B-A10B-MTP-GGUF too.

ohh i didn't realize there were a whole ton of Qwen3.5 MTPs

PR #1950, opened 2026-05-19
Fix: Adds SHA check on every download

### LINKED ISSUE #1846: SHA check on every download
A huge portion of bug reports are root caused to partial model and/or backend downloads.

fl0rianr:
## Summary

Adds content verification for model/backend downloads so corrupted cached files are not silently reused.

Closes #1846 

## Changes

- Verify downloaded files with SHA256 or Git blob SHA1 metadata when available.
- Re-download existing cache files when hash verification fails.
- Use immutable Hugging Face commit refs for model file URLs and metadata lookup.
- Reject obviously invalid cached GGUF files before they are treated as downloaded.
- Add focused download-registry tests covering real tiny-model download and corrupted-cache redownload behavior.
- Extend the compact CI download suite to exercise the SHA verification path.

## Validation

- `LEMONADE_RUN_REAL_DOWNLOAD_TESTS=1 LEMONADE_DOWNLOAD_TEST_MODEL=Tiny-Test-Model-GGUF python test/server_downloads.py --cli-binary ./build/lemonade`
- Passes registry lifecycle, real tiny model download, and SHA/Git-SHA corrupted-payload redownload coverage.
- Optional GGUF magic-header test remains gated behind `LEMONADE_RUN_GGUF_MAGIC_TESTS=1`.


src/cpp/server/backends/backend_utils.cpp
```diff
@@ -45,6 +46,76 @@ namespace lemon::backends {
...
+                {"checksums", recipe, filename},
+                {"artifact_hashes", "github", repo, version, filename},
+                {"artifact_hashes", repo, version, filename},
+                {"artifact_hashes", recipe, backend, version, filename},
+                {"artifact_hashes", recipe, version, filename},
+                {"artifact_hashes", filename}
+            };
+
+            for (const auto& path : candidate_paths) {
+                // Skip repo-keyed shapes when no repo is available (e.g. TheRock).
```
superm1, line 103: it's worth filing something upstream to https://github.com/rocm/therock requesting this.  It feels totally doable.

src/cpp/server/utils/http_client.cpp
```diff
@@ -17,6 +23,409 @@ namespace utils {
...
+
+class Sha1Context {
+public:
+    Sha1Context() {
+        state_[0] = 0x67452301;
+        state_[1] = 0xefcdab89;
+        state_[2] = 0x98badcfe;
+        state_[3] = 0x10325476;
+        state_[4] = 0xc3d2e1f0;
+    }
```
superm1: there should be some libraries available in the distros that we can use.  I would like to avoid having to maintain something error prone and sensitive like hashing verification when there is already stuff to do this available.

fl0rianr:
@superm1 Thank you very much for this very valuable feedback! So this is the real deal now, if you have some time I would appreciate your review.

src/cpp/cli/CMakeLists.txt
```diff
@@ -118,11 +120,22 @@ if(APPLE)
...
-    target_link_libraries(lemonade PRIVATE
-        Threads::Threads
-    )
+    find_package(OpenSSL REQUIRED)
+    target_link_libraries(lemonade PRIVATE Threads::Threads)
+    if(TARGET OpenSSL::Crypto)
+        target_link_libraries(lemonade PRIVATE OpenSSL::Crypto)
+    elseif(OPENSSL_CRYPTO_LIBRARY)
+        target_link_libraries(lemonade PRIVATE ${OPENSSL_CRYPTO_LIBRARY})
+        target_include_directories(lemonade PRIVATE ${OPENSSL_INCLUDE_DIR})
```
superm1: @jfowers what's the stance on OpenSSL?  Should this be GnuTLS?

jeremyfowers: I really dont know about anything about SHA/SSL libraries so I’m not the right person to weigh in :smile:

fl0rianr: Instead of OpenSSL we are using a solid, very light, multi OS, ARM mentained trusted firmware lib: MbedTLS

fl0rianr:
Clarifying the intended cache behavior for this PR, since this is important for large model downloads:

This change does **not** proactively redownload already installed models just because SHA/Git-SHA verification support exists.

The behavior is:

- If a final cached file already exists and the manifest has an expected hash, Lemonade verifies the local file first.
- If the hash matches, the file is reused and the download path returns success with bytes_downloaded = 0.
- If the hash does not match, the cached file is treated as corrupt, removed, and downloaded again from scratch.
- If no expected hash is available, the existing legacy behavior is preserved and a final file with no `.partial` is treated as complete.
- The extra GGUF magic-header check is only meant to catch obviously invalid cached GGUF payloads such as HTML/error/pointer files before llama.cpp fails later with an opaque startup error.

For Hugging Face models, the PR now uses immutable commit refs once a pull resolves the current checkpoint metadata. That means an upstream update to `main` does not automatically redownload a model that is already installed. A newer upstream revision is only considered when the user/server performs a new pull of the same unpinned checkpoint, at which point the current HF commit/hash metadata may point to a new snapshot.

In other words: valid cache should be reused; corrupted or invalid cache should be repaired.
Therefore I expanded test/server_downloads.py to test that as well.

fl0rianr:
This follow-up commit fixes the remaining README-only update case.

Before:
Lemonade keyed reuse to the HF snapshot path. When upstream pushed a README-only commit, HF produced a new repo SHA, Lemonade created a new snapshots/<sha> directory, and the same multi-GB model artifact could be downloaded again even though the GGUF bytes were unchanged.

After:
For each manifest entry with immutable HF content metadata, Lemonade first checks sibling snapshot directories for the same repo-relative file and size. If found, it materializes it into the new snapshot via hardlink/copy, then lets the existing SHA/Git-SHA verifier decide whether it is valid.

Impact:
README-only or metadata-only HF commits no longer force model redownloads. Real model changes still download normally because the content hash changes and the verifier rejects stale bytes.

Validation:
Tested with a synthetic README-only snapshot scenario: first pull downloaded Tiny-Test-Model-GGUF, the current snapshot was moved to a sibling snapshot and the target snapshot removed, then a second pull reused the unchanged GGUF from the sibling snapshot and verified it by hash without redownloading.

fl0rianr:
Validated. README-only snapshot simulation reuses unchanged GGUF from a sibling snapshot and keeps refs/main pinned. Corrupting that sibling artifact triggers hash verification failure, fresh redownload, successful hash verification, and refs/main advancement. So unchanged artifacts avoid redownloads, but real model changes still download correctly.

fl0rianr:
@claude: review

fl0rianr:
Further unnecessary re-download fixes  and scope change: UI added opt-in for auto update - current default: user loaded model version will be kept. Latest HF version always - with opt-in.

fl0rianr:
@claude: review

fl0rianr:
CI fix was removed from the scope of this PR.

fl0rianr:
Reduced Scope the original SHA feature only, fixes mentioned in the comments like readme change - no re-download are NOT valid anymore.

src/cpp/cli/CMakeLists.txt
```diff
@@ -2,6 +2,84 @@
...
+# Prefer a distro-provided mbedcrypto package, but fall back to FetchContent so
+# local Lemonade builds do not require developers to install a new package.
+if(NOT TARGET lemonade-digest-crypto)
+    find_package(PkgConfig QUIET)
+    if(PkgConfig_FOUND)
+        pkg_check_modules(MBEDCRYPTO QUIET mbedcrypto)
+    endif()
+
+    add_library(lemonade-digest-crypto INTERFACE)
+    target_compile_definitions(lemonade-digest-crypto INTERFACE LEMONADE_USE_MBEDTLS_DIGEST)
```
jeremyfowers: _Drafted by Claude on behalf of @jfowers._

`LEMONADE_USE_MBEDTLS_DIGEST` is never referenced anywhere in the tree — nothing guards on it with `#ifdef`/`#if defined`, and `http_client.cpp` includes `<mbedtls/md.h>` unconditionally. It's a dead compile definition; suggest dropping it.

```suggestion
```

src/cpp/include/lemon/utils/http_client.h
```diff
@@ -38,6 +38,7 @@ struct DownloadResult {
     size_t total_bytes = 0;           // Total file size (if known)
     bool can_resume = false;          // Whether partial download can be resumed
     bool disk_full = false;            // True if download failed due to insufficient disk space
+    std::string verified_hash;        // Hex digest computed during verification, when applicable
```
jeremyfowers: _Drafted by Claude on behalf of @jfowers._

`verified_hash` is written in three places (`http_client.cpp:792`, `:817`, `:924`) but never read anywhere in the codebase or tests. Either consume it (e.g. surface it to the caller or logs) or drop the field along with its three assignments. Not posting a suggestion since the declaration and its write sites are non-contiguous.

jeremyfowers (approved):
LGTM! Thanks for all your efforts on this @fl0rianr, especially thanks for taking the feedback about refactoring into multiple PRs.

fl0rianr:
Just removed the dead code.

PR #1951, opened 2026-05-19
when agent cli is not found print env vars and launch cmd

anishp55:
this changes the behavior of `lemonade launch <agent>` to dump the env vars and launch command needed to start the agent when the agent cli is not found.  usee is responsible for updating the URL if needed.


fl0rianr (changes requested):
Thanks for adding this fallback! I think the idea is useful - but I also don’t think this is ready yet.

The main issue is that the printed `Command:` is not safely shell-escaped. For Codex in particular, `extra_args` includes config strings with spaces/quotes/braces, so the printed command will not round-trip as a copy-pasteable shell command. --agent-args can also intentionally contain values with spaces. Since the goal of this PR is to print the command needed to start the agent manually, I think this needs proper platform-aware quoting before merge.

A second concern is that this prints all env vars verbatim, including API keys/tokens such as OPENAI_API_KEY, ANTHROPIC_AUTH_TOKEN, and LEMONADE_API_KEY. We should not print secrets by default. If the manual-launch UX needs credentials, please make that explicit with a warning and/or gate it behind an opt-in flag.

Also, this changes the missing-agent error path to fire `/api/v1/load` before exiting, while swallowing all errors. That seems like a behavioral change that should be documented and tested, especially because the CLI can now print “Loading model in background” even if the load request failed.

Before merge, could we address these points:

* [ ] Add proper shell quoting for the printed command, especially for Codex args with spaces/quotes/braces.
* [ ] Add a lean automated test for the missing-agent path that verifies the printed command is copy-paste safe.
* [ ] Do not print API keys/tokens by default; add an explicit warning and/or opt-in if credentials must be included for manual launch.
* [ ] Document or clarify why loading should still be fired when the agent CLI is missing, or revert this part.
* [ ] Avoid misleading output if the load request fails or times out.

anishp55:
i would normally agree with you about the secrets, but they are literally non used values that are just generated to keep the cli clients happy.  I'll fix up the escape issues for the codex cli.

anishp55:
i cheated an used qwen to add the escape chars and the test for that.  i've validated that the escaped output works and that the new tests pass.  the API key is still there as its value doesn't matter, but is required to be set for the CLIs.  We can revisit this if API tokens are going to be generated by lemonade.

fl0rianr (changes requested):
Thanks for iterating on this. This is closer, but I still don’t think it is ready to merge yet.

A few remaining issues:

* There is now duplicate unreachable code in handle_launch_command. The first `if (agent_binary.empty())` block prints the manual launch info, calls request_model_load_for_manual_agent_launch(...), and returns. Immediately after that, there is another `if (agent_binary.empty())` block, which can never run. That second block also contains the old unquoted command-printing logic, so it should be removed before merge.

* The secret-printing concern is not fully resolved. I agree that the default "lemonade" token is not sensitive, but this path also prints config.api_key when the user provided --api-key. In that case, OPENAI_API_KEY, ANTHROPIC_AUTH_TOKEN, LEMONADE_API_KEY, and the OpenCode/Pi `apiKey=` provider details can contain a real secret. We should not print user-provided API keys by default.

* The new quoting helper is POSIX-shell quoting. That is useful, and the new `shlex` test is a good start, but it is not really platform-aware. On Windows the agent binaries are .cmd/.exe, and single-quote shell escaping is not generally cmd.exe -safe. Either the output should clearly be labeled as POSIX-shell command output, or the implementation should emit platform-appropriate commands.

* The new test verifies round-tripping through Python `shlex`, which covers the original Codex spacing/quotes issue for POSIX shells. It does not prove the command is copy-paste safe on Windows. That is okay if we scope the feature to POSIX output, but then the UX/message should be explicit.

* I’m still not fully convinced that we should trigger /api/v1/load when the agent binary is missing. The new message is less misleading than before, but this still changes the failure path from a local validation failure into a server-side side effect. If we keep it, I’d like the PR description/docs to explicitly say this is intentional.

Before merge, could we please address:

* [ ] Remove the duplicate unreachable `if (agent_binary.empty())` block.
* [ ] Do not print user-provided API keys/tokens by default; printing the default dummy "lemonade" value is less concerning, but real --api-key values should be redacted or gated.
* [ ] Either make command quoting platform-aware, or explicitly label the printed command as POSIX-shell-only.
* [ ] Adjust/extend tests to match the intended platform behavior.
* [ ] Clarify in the PR description why /api/v1/load should still be triggered when the agent CLI is missing, or revert that side effect.

jeremyfowers:
@fl0rianr this seems like a nice feature to have. You did leave a *lot* of feedback for a first time contributor on 200 LoC PR though. How would you feel about pushing the necessary improvements to the branch yourself, and then merging this?

PR #1953, opened 2026-05-19
Add bench command

bitgamma:
still heavily WIP and missing some important features. Opening a draft PR just so that we can start discussion. 

Note: for now benchmarking will only focus on e2e speed metrics. I have seen that `lemonade-eval` is using PyTorch directly for the other benchmarks so that is out of scope at least for now.

Adds a new lemonade bench <model> subcommand to benchmark chat completion performance across backends, context sizes, and predefined scenarios.                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                              
 ### New Files                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                              
 - src/cpp/cli/bench.cpp — Full benchmark implementation                                                                                                                                                                                                                                                         
 - src/cpp/include/lemon_cli/bench.h — Public API header                                                                                                                                                                                                                                                          
 - src/cpp/resources/bench_scenarios.json  — Built-in benchmark scenarios                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
 ### Modified Files                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                              
 - src/cpp/cli/main.cpp — Registers the bench subcommand with CLI11, wires up options → BenchConfig → handle_bench_command()                                                                                                                                                                                                  
 - src/cpp/cli/CMakeLists.txt — Adds bench.cpp to the build                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                              
 ### Features                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                              
 - Multi-backend comparison — Test the same model across backends (--backend vulkan --backend metal --backend cpu)                                                                                                                                                                                                            
 - Multi context-size testing — Benchmark at different context lengths (--ctx-size 2048 --ctx-size 8192)                                                                                                                                                                                                                      
 - 9 built-in scenarios across 3 categories:                                                                                                                                                                                                                                                                                  
     - Chat: chat-short, chat-medium, chat-long-output                                                                                                                                                                                                                                                                        
     - Coding: code-short, code-explain, code-debug                                                                                                                                                                                                                                                                           
     - Long-context: context-2k, context-8k (with auto-expanded filler text), context-multi-turn                                                                                                                                                                                                                              
 - Metrics collected: TTFT (mean/min/max/p50/p95), tokens/sec (mean/min/max/p50/p95), VRAM usage, RAM usage, input/output token counts                                                                                                                                                                                        
 - Configurable runs: --warmup N and --runs N (default: 1 warmup, 3 measurement)                                                                                                                                                                                                                                              
 - Custom scenarios: Load from a single file (--scenario-file) or directory (--scenario-dir)                                                                                                                                                                                                                                  
 - JSON output: --json flag for programmatic consumption, --output FILE to save results                                                                                                                                                                                                                                       
 - Comparison mode: --compare <file> to diff results against a previous JSON benchmark run (shows % deltas for TTFT, TPS, VRAM)                                                                                                                                                                                               
 - Auto-pull: --auto-pull to download the model if not already present                                                                                                                                                                                                                                                        
 - Memory tracking toggle: --no-memory to skip VRAM/RAM tracking                                                                                                                                                                                                                                                              
 - Formatted table output to stdout with adaptive columns (min/max for few runs, p50/p95 for 10+ runs)                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                              
 ### Usage Examples                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                              
 ```bash                                                                                                                                                                                                                                                                                                                      
   # Quick benchmark with defaults                                                                                                                                                                                                                                                                                            
   lemonade bench Qwen3.6-35B-A3B-MTP-GGUF                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                              
   # Compare backends                                                                                                                                                                                                                                                                                                         
   lemonade bench --backend vulkan --backend cpu  Qwen3.6-35B-A3B-MTP-GGUF                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                              
   # Benchmark at multiple context sizes with JSON output                                                                                                                                                                                                                                                                     
   lemonade bench --backend vulkan --ctx-size 2048 --ctx-size 8192 --json --output results.json  Qwen3.6-35B-A3B-MTP-GGUF                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                              
   # Compare against previous results                                                                                                                                                                                                                                                                                         
   lemonade bench  --backend vulkan --compare results.json Qwen3.6-35B-A3B-MTP-GGUF                                                                                                                                                                                                                             
 ```

### Example output
```
$ build/lemonade bench --ctx-size 16384 lfm2-2.6b-FLM

=== flm/npu (ctx=16384) ===
  Loading model with flm/npu (ctx=16384)... done
  Scenario: chat-short (chat)
    Warmup 1/1... done
    Run 1/3... TTFT=730.8ms TPS=30.4
    Run 2/3... TTFT=696.9ms TPS=30.7
    Run 3/3... TTFT=704.6ms TPS=30.8
  Scenario: chat-medium (chat)
    Warmup 1/1... done
    Run 1/3... TTFT=717.0ms TPS=30.5
    Run 2/3... TTFT=714.9ms TPS=30.4
    Run 3/3... TTFT=746.0ms TPS=30.1
  Scenario: chat-long-output (chat)
    Warmup 1/1... done
    Run 1/3... TTFT=805.8ms TPS=27.2
    Run 2/3... TTFT=765.9ms TPS=27.5
    Run 3/3... TTFT=719.6ms TPS=29.7
  Scenario: code-short (coding)
    Warmup 1/1... done
    Run 1/3... TTFT=707.7ms TPS=29.9
    Run 2/3... TTFT=764.4ms TPS=29.3
    Run 3/3... TTFT=689.5ms TPS=29.8
  Scenario: code-explain (coding)
    Warmup 1/1... done
    Run 1/3... TTFT=730.5ms TPS=29.8
    Run 2/3... TTFT=724.9ms TPS=29.8
    Run 3/3... TTFT=710.8ms TPS=29.4
  Scenario: code-debug (coding)
    Warmup 1/1... done
    Run 1/3... TTFT=897.1ms TPS=29.9
    Run 2/3... TTFT=926.4ms TPS=29.7
    Run 3/3... TTFT=910.9ms TPS=28.9
  Scenario: context-2k (long-context)
    Warmup 1/1... done
    Run 1/3... TTFT=1880.8ms TPS=29.5
    Run 2/3... TTFT=1918.0ms TPS=28.8
    Run 3/3... TTFT=1850.7ms TPS=29.6
  Scenario: context-8k (long-context)
    Warmup 1/1... done
    Run 1/3... TTFT=6174.6ms TPS=27.4
    Run 2/3... TTFT=6350.6ms TPS=27.7
    Run 3/3... TTFT=6221.3ms TPS=27.8
  Scenario: context-multi-turn (long-context)
    Warmup 1/1... done
    Run 1/3... TTFT=1515.8ms TPS=29.2
    Run 2/3... TTFT=1561.1ms TPS=29.3
    Run 3/3... TTFT=1543.8ms TPS=29.4

Benchmark: lfm2-2.6b-FLM
====================================================================================================

Backend: flm/npu (ctx=16384)
chat-short          710.8   696.9   730.8   30.6    30.4    30.8    3.3     
chat-medium         726.0   714.9   746.0   30.3    30.1    30.5    3.3     
chat-long-output    763.8   719.6   805.8   28.2    27.2    29.7    3.6     
code-short          720.5   689.5   764.4   29.7    29.3    29.9    3.5     
code-explain        722.1   710.8   730.5   29.7    29.4    29.8    3.4     
code-debug          911.4   897.1   926.4   29.5    28.9    29.9    3.5     
context-2k          1883.2  1850.7  1918.0  29.3    28.8    29.6    3.4     
context-8k          6248.8  6174.6  6350.6  27.6    27.4    27.8    3.4     
context-multi-turn  1540.2  1515.8  1561.1  29.3    29.2    29.4    3.3
```

cc @jeremyfowers @fl0rianr @amd-pworfolk


src/cpp/cli/bench.cpp
```diff
@@ -0,0 +1,1217 @@
...
+}
+
+std::vector<BenchComparisonDelta> compute_deltas(const std::vector<BenchBackendResult>& current,
+                                                  const json& previous_results) {
+    std::vector<BenchComparisonDelta> deltas;
+
+    // Build lookup map from previous results
+    struct PrevKey {
+        std::string backend;
+        std::string scenario;
```
fl0rianr, line 876: This supports testing multiple backend arg sets, I think the comparison key needs to include the full tested configuration.

Currently the previous/current lookup appears to key only on recipe/backend + scenario. That can collapse different ctx_size and backend_args runs into the same comparison bucket. For example:

```bash
lemonade bench model \
  --backend vulkan \
  --ctx-size 4096 --ctx-size 8192 \
  --llamacpp-args "-b 512 -ub 512" \
  --llamacpp-args "-b 2048 -ub 1024"

Those are materially different configurations, but they would compare as the same backend/scenario.

Could the key include at least:

recipe
backend
ctx_size
backend_args
scenario

or maybe use BenchBackendResult::label() consistently for comparison keys? This matters especially if we want to use this benchmark to decide which llama.cpp args are “best”, “indifferent”, “bad”, or “critical” for a device/backend.

File: src/cpp/cli/bench.cpp [hunk 2]
```diff
@@ -0,0 +1,1217 @@
...
+                            const std::string& backend,
+                            int ctx_size,
+                            const std::string& backend_args) {
+    try {
+        json request_body;
+        request_body["model_name"] = model;
+        request_body["save_options"] = false;
+
+        // Always disable merge_args so benchmark args fully override defaults
+        request_body["merge_args"] = false;
```
fl0rianr: I like the explicit arg-set benchmarking, but I’m not sure `merge_args=false` should always be set.

If no custom backend args are passed, does this still benchmark the same behavior a normal user would get from the current defaults? If disabling merge means recipe/model defaults are not applied, then the “baseline” run may no longer represent the real default configuration.

Maybe only set this when backend_args is non-empty, or expose this as an explicit benchmark mode, e.g.:

request_body["merge_args"] = backend_args.empty();

or a CLI flag like --replace-default-args.

This matters because we want to compare candidate llama.cpp defaults against the actual current defaults, not accidentally compare them against an empty/defaultless configuration.

src/cpp/cli/bench.cpp
```diff
@@ -0,0 +1,1217 @@
...
+
+        if (resp_json.contains("timings") && resp_json["timings"].is_object()) {
+            extract_timings_into_result(resp_json["timings"], result);
+        }
+        if (resp_json.contains("usage") && resp_json["usage"].is_object()) {
+            extract_usage_into_result(resp_json["usage"], result);
+        }
+    } catch (const std::exception& e) {
+        std::cerr << "    Benchmark run failed: " << e.what() << std::endl;
+        return result;
```
fl0rianr, line 577: For the “critical” classification, failed runs should probably be represented explicitly.

Right now a failed run returns the default BenchRunResult, so metrics like ttft_ms=0 etc. can enter the aggregate stats. That can make failures look like valid measurements and makes it hard to distinguish a slow/bad arg set from a broken one.

Could we add something like:

```
bool success = true;
std::string error;
```

to BenchRunResult, set success=false here, and exclude failed runs from mean/min/max/p50/p95? Then load failures, OOMs, invalid metrics, or request failures can be classified as critical.

src/cpp/include/lemon_cli/bench.h
```diff
@@ -0,0 +1,259 @@
...
+// Comparison
+// ============================================================
+
+struct BenchComparisonDelta {
+    std::string backend;
+    std::string scenario;
+    double ttft_pct_change;    // Positive = slower, negative = faster
+    double tps_pct_change;     // Positive = faster, negative = slower
+    double vram_gb_change;     // Positive = more VRAM used
+    std::string status;        // "matched", "new", "removed"
```
fl0rianr, line 241: Could we separate structural comparison status from performance/impact status?

Right now `status` means matched/new/removed, but for backend-arg selection I think we also want an impact classification. Something like:

```
std::string comparison_status; // matched/new/removed
std::string impact_status;     // best/good/indifferent/bad/critical
std::string impact_reason;
```
That would make the output useful for deciding whether a tested arg set should become a default, rather than only showing raw metric deltas, like addressed before. It would also avoid hard-coding “best” llama.cpp args by device/backend without evidence, since the benchmark can classify them from the actual measured results.

src/cpp/cli/bench.cpp
```diff
@@ -0,0 +1,1217 @@
...
+
+static std::string fmt_pct_change(double pct) {
+    std::ostringstream oss;
+    oss << (pct >= 0 ? "+" : "") << std::fixed << std::setprecision(1) << pct << "%";
+    return oss.str();
+}
+
+static std::string fmt_vram_change(double val) {
+    if (val < 0) return "-";
+    return "+" + fmt_double(val) + " GB";
```
fl0rianr: Small topic, negative VRAM deltas are valid and usually good, but this currently prints any `val < 0` as unavailable.

For example, if an arg set saves 0.5 GB VRAM, the comparison would show `-` instead of `-0.5 GB`.

Could we use a separate availability marker instead of overloading negative values? For example:

`std::optional<double> vram_gb_change;`

or a separate has_vram_change flag.

fl0rianr (commented):
Nice update and good work! The repeatable backend arg support makes this much closer to being useful for deciding llama.cpp defaults.

One broader thought: if the goal is to choose default args depending on device/backend, I’d avoid hard-coding “best” parameters directly from assumptions. The best values can depend on exact iGPU/dGPU, driver version, backend version, model/quant, ctx size, and may change over time.

This benchmark command could be the mechanism to validate those defaults instead:

- run multiple candidate arg sets per backend/device
- compare them against the current default behavior
- classify each arg set as `best`, `good`, `indifferent`, `bad`, or `critical`
- include the classification and reason in JSON output so this can be used for CI or periodic revalidation

(With a certain careful consideration a special bug does not kill all args that are in the default before the test)

The current update already adds backend_args, which is great. I think the next step would be to make comparison keys include the full tested config (recipe, backend, ctx_size, backend_args, scenario) and add an explicit impact/severity field separate from matched/new/removed.

That would let us say “these defaults are currently best for this device/backend/model class” based on evidence, while still making it easy to re-run when llama.cpp, drivers, or Lemonade defaults change.

src/cpp/include/lemon_cli/bench.h
bitgamma: I don't think we should try to interpret the data with a "score", since each user might have different ideas about what constitutes a significant change and an insignificant one. Or did I misunderstand your proposal?

bitgamma:
@fl0rianr I think I addressed most of your comments, could you please have a look?

src/cpp/cli/bench.cpp
```diff
@@ -0,0 +1,1263 @@
...
+            s_json["output_tokens"] = scenario.output_tokens();
+
+            json ttft_stats;
+            ttft_stats["mean"] = scenario.ttft_mean_ms();
+            ttft_stats["min"] = scenario.ttft_min_ms();
+            ttft_stats["max"] = scenario.ttft_max_ms();
+            ttft_stats["p50"] = scenario.ttft_p50_ms();
+            ttft_stats["p95"] = scenario.ttft_p95_ms();
+            s_json["ttft_ms"] = ttft_stats;
+
```
fl0rianr: Nice improvement excluding failed runs from the table stats. I thing there is one remaining literal edge case: if all measurement runs for a scenario fail, the table prints dashes, but the JSON still appears to emit numeric stats via `ttft_mean_ms()`, `tps_mean()`, etc.

Since those helpers return 0.0 for empty runs, an all-failed scenario can look like a valid 0ms TTFT / 0 TPS result in JSON and in `--compare`.

Could we emit `null` or omit the metric blocks when `scenario.runs.empty()`?

src/cpp/cli/bench.cpp
```diff
@@ -0,0 +1,1263 @@
...
+                tmp.ctx_size = ctx_size;
+                tmp.backend_args = recipe_args;
+                std::cout << "\n=== " << tmp.label() << " ===" << std::endl;
+
+                // Unload all models before loading
+                unload_all_models(client);
+
+                // Load model for this backend
+                if (!load_model_for_backend(client, config.model, recipe, backend, ctx_size, recipe_args)) {
+                    std::cerr << "Warning: Skipping " << recipe << "/" << backend
```
fl0rianr: For the impact/severity use case, I think load failures should be recorded as results instead of being skipped entirely.

Right now, if a backend/ctx/arg-set fails to load, we `continue`, so the attempted configuration disappears from all_results. In compare mode that can look like the previous result was simply removed, and for a new arg set there may be no JSON record at all.

Could we add a `BenchBackendResult` entry with something like:

```
bool load_success = false;
std::string load_error;
```
or an explicit result status like critical?

That would make failed arg sets visible in both table and JSON output and would directly support the critical classification for OOM/load/backend incompatibility cases.

src/cpp/cli/bench.cpp
```diff
@@ -0,0 +1,1263 @@
...
+            delta.ctx_size = backend_result.ctx_size;
+            delta.backend_args = backend_result.backend_args;
+            delta.scenario = scenario.scenario_name;
+
+            if (it != prev_map.end()) {
+                delta.status = "matched";
+                matched_prev.insert(key);
+
+                double curr_ttft = scenario.ttft_mean_ms();
+                double curr_tps = scenario.tps_mean();
```
fl0rianr: Now that failed runs are tracked, should `compute_deltas()` treat scenarios with no successful runs as a distinct status?

If scenario.runs.empty() and failed_runs > 0, the current mean helpers return 0.0, so comparison can produce misleading deltas against a valid previous baseline. For example, TTFT could look like a huge improvement because current TTFT is 0.

Maybe comparison should short-circuit:

```
if (scenario.runs.empty() && scenario.failed_runs > 0) {
    delta.status = "failed"; // or impact_status = "critical"
    ...
}
```

and avoid computing TTFT/TPS deltas when there is no valid current measurement.

src/cpp/include/lemon_cli/bench.h
fl0rianr: No, I guess you did not misunderstood - was just my mindset regarding that. There should be a certain metric to tell if an -arg is really good performing, problematic/unstable or even certainly crashing. Or why add an -arg that was no effect? I'm sure we will get cases where an certain setting is useful while it degrades another score. If we can't evaluate that impact - what is our decision basis instead? We can continue that discussion on discord as well.

src/cpp/cli/bench.cpp
```diff
@@ -0,0 +1,1263 @@
...
+    auto time_t_now = system_clock::to_time_t(now);
+    struct tm tm_buf;
+#ifdef _WIN32
+    localtime_s(&tm_buf, &time_t_now);
+#else
+    localtime_r(&time_t_now, &tm_buf);
+#endif
+    char buf[64];
+    std::strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%SZ", &tm_buf);
+    return std::string(buf);
```
fl0rianr, line 37: A bit nitty here, not sure how impactfully the timestep will be. But could we use gmtime_s / gmtime_r instead, localtime may lead to variations.

fl0rianr (commented):
Nice update — this is moving in the right direction. Thanks!

I commented an each point I found.
I think the remaining conceptual gap is separating comparison status from impact/severity. For choosing device/backend-specific llama.cpp defaults, raw deltas are helpful, but the tool would be even more actionable if it could classify each tested arg set - doesn't not have to be like this! -  just my raw idea:

- best
- good
- indifferent
- bad
- critical

But we can discuss this elsewhere and update the findings here.

src/cpp/include/lemon_cli/bench.h
bitgamma: I think this tool is for the users to take their decisions, each according to their own criteria. Even if we want to automate decisions to provide machine-specific defaults in the future, the analysis can be done on the raw data rather than having the tool decide by itself what is good and what is bad

src/cpp/cli/bench.cpp
bitgamma: I think the timestamp there is more for the user than anything, so it is nicer to show a value the user recognizes

bitgamma:
@fl0rianr I think I will not do any more changes/improvements to the result comparison functionality in the scope of this PR. I want to set the foundation for doing benchmarks and then we can improve/add functionality later. Since you have some ideas on how to make use of the results for taking some (automated?) decision, I will leave that part to you.

src/cpp/include/lemon_cli/bench.h
fl0rianr: Convinced.

fl0rianr:
@bitgamma Thanks for iterating on this — the current version already addresses most of the things I pointed out. Good work!

For the remaining all-failed/0.0 metrics edge case, I’m fine with handling that in a follow-up automation as long as we consistently treat 0.0 results as invalid there.

Everything else from my side can be follow-up.

fl0rianr (approved):

PR #1956, opened 2026-05-19
fix(app): don't skip port discovery when no explicit base URL is set

abn:
The init early-return fired even when getServerBaseUrl() returned nothing, so the desktop app stayed pinned to the default port 13305 and never registered the beacon/connection-settings listeners. Works by luck when lemond is on 13305, breaks on any other port (e.g. flatpak).

Move the return inside the `if (baseUrl)` branch so the localhost case falls through to port discovery, like the web-app branch above it.


fl0rianr (commented):
Thanks @abn! A good fix! But to be solid I will approve and merge it after the PR mentioned above is merged.

fl0rianr (approved):

PR #1957, opened 2026-05-19
Register lemonade:// scheme handler in app desktop entry

abn:
The Tauri app declares and handles the lemonade:// deep-link scheme, but the installed desktop entry did not advertise it. Add the x-scheme-handler/lemonade MimeType and the %u Exec field code so the desktop environment routes lemonade:// URLs to the app immediately on install, rather than only after the first launch (which is when the runtime fallback registration kicks in).


fl0rianr (commented):
Thank you. The `.desktop` change looks directionally correct and only a tiny change.

Since you did not stated this, before I’d be comfortable merging, can you confirm this was tested from a fresh Linux install/package, not only after launching the app once?

Suggested checks:
- `desktop-file-validate data/lemonade-app.desktop`
- install the deb/rpm/full package
- `xdg-mime query default x-scheme-handler/lemonade`
- `xdg-open 'lemonade://open?view=logs'`

One concern: on Linux/Windows Tauri deep links are delivered as command-line args on a new process. The code handles URLs in the single-instance callback and `on_open_url`, but I don’t see an explicit cold-start get_current/initial-argv handling path. If the app is not already running, does the URL actually get consumed and navigate, or does it only launch the app? Thanks for clarification!

superm1 (approved):

abn:
@fl0rianr I have added a new commit to address your concern. Additionally, I have followed the following steps to verify the fixes. This was performed on a Fedora machine, this should be sufficient to validate the fix.

1. Verified `desktop-file-validate data/lemonade-app.desktop` exits with 0 (only the pre-existing Categories hint).
2. Deleted the per-user runtime handler `~/.local/share/applications/lemonade-app-handler.desktop`. Confirmed `xdg-mime query default x-scheme-handler/lemonade` returned empty (no handler).
3. Installed the simulated new-rpm desktop file over `/usr/share/applications/lemonade.desktop`, ran `update-desktop-database`.
4. Verified `xdg-mime query default x-scheme-handler/lemonade` returns `lemonade.desktop`.
5. Built the Tauri app with the cold-start argv handler. Ran the new binary directly: `lemonade-app 'lemonade://open?view=logs'` with no prior instance. Within 1s of startup, the log shows `Handling lemonade:// URL: lemonade://open?view=logs` — meaning the URL reaches handle_protocol_urls, which then queues a NAVIGATE event (drained by the renderer on renderer_ready).

fl0rianr (approved):
Thanks for the follow-up and for validating this on Fedora! The added cold-start argv handling addresses my concern, and the documented checks cover the relevant desktop-entry/handler path well enough for me.

Very nice work! Approving — thanks for the quick fix!

PR #1962, opened 2026-05-19
Add a tool for updating model sizes

jeremyfowers:
Most model sizes in server_models.json were off by a bit.

For models with mmproj files, the model sizes often did not include the mmproj and could be off by ~1 GB.

The PR fixes the model sizes and adds a script that uses the HF API to get correct sizes.

Usage:
```
  tools/model_sizes.py src/cpp/resources/server_models.json --all
  tools/model_sizes.py src/cpp/resources/server_models.json --models Qwen3.5-4B-GGUF "Ultra Collection"
  tools/model_sizes.py src/cpp/resources/server_models.json --all --update
```


sofiageo (approved):
Tested with --all, with --models, and with --all --update. Also compared a couple of models manually with huggingface. Seems to be working great!

PR #1965, opened 2026-05-20
Introduce better model level vLLM args

### LINKED ISSUE #1964: Better default vLLM arguments (per model family/model)
vLLM requires cli args that are specific to model family/model for tool calling and running quantized models. Currently we have two ways of configuring these, modifying the launch args and passing them through vllm_args.

There should be a middle ground for model specific cli flags to reduce friction and get tool calling and running quantized models working by default. 

I am suggesting `vllm_model_config.json`, an example of which I've included below.

```json
    {
      "schema_version": 1,
      "families": {
        "qwen3": {
          "match": [{"checkpoint_regex": "^Qwen3."}],
          "args": "--enable-auto-tool-choice --tool-call-parser qwen3_coder"
        }
      },
      "models": {
        "Qwen3.5-4B-vLLM": {
          "family": "qwen3"
        },
        "Some-vLLM-Model": {
          "disable_family_match": true,
          "args": "--tool-call-parser hermes"
        }
      }
    }
```

This reduces repetition in server_models.json, makes tool-calling parser and quantization choices explicit per model family, and keeps Lemonade managed launch behavior protected.

sawansri:
Resolves #1964


sawansri:
Branch is rebased off of #1423 hence the # of commits and LoC changed.

PR #1970, opened 2026-05-20
feat: Add Ryzen AI NPU backend for Stable Diffusion image generation

makn87amd:
Add SDNPUServer backend (sd-npu recipe) that manages ryzenai-sd-server as a subprocess for NPU-accelerated image generation on AMD Ryzen AI.

Backend integration:
- SDNPUServer class implementing IImageServer (generations, edits, variations)
- Recipe option defaults for sd-npu_backend, steps, cfg_scale, width, height
- NPU exclusivity enforcement (mutual exclusion with ryzenai-llm, whispercpp)
- Backend version pinned at v0.1.0 (npu and cpu variants)
- Router instantiation gated on sd-npu recipe

Models (7):
- SD-Turbo-NPU, SDXL-Turbo-NPU, SD-1.5-NPU, SDXL-Base-NPU, Segmind-Vega-NPU, SD3-Medium-NPU, SD3.5-Medium-NPU

Tests:
- test/server_sd_npu.py with txt2img, img2img, variations, size, and multi-model coverage
- Test utility updates (capabilities, server_base, test_models)

### DIFF
--- a/test/server_sd_npu.py
+++ b/test/server_sd_npu.py
@@ -1,0 +1,412 @@
+"""
+SD NPU image generation tests for Lemonade Server.
+
+Tests the /images/generations, /images/edits, and /images/variations endpoints
+with SD NPU models (ONNX-based Stable Diffusion on AMD Ryzen AI NPU).
+
+Usage:
+    python server_sd_npu.py
+    python server_sd_npu.py --server-binary /path/to/lemond
+"""
+
+import base64
+import io
+import struct
+import zlib
+import requests
+
+from utils.server_base import (
+    ServerTestBase,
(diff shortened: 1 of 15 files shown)


test/server_sd_npu.py
```diff
@@ -0,0 +1,412 @@
+"""
```
jeremyfowers, line 1: Why do we need a dedicated test file for SD NPU? Ideally we would just pass an npu flag to the existing sd test file.

makn87amd: I thought we needed to test the image "variation" and "edits" server endpoints. I don't believe those are tested currently

PR #1973, opened 2026-05-20
Rebrand OmniRouter to Lemonade Omni

jeremyfowers:
image

This PR rebrands OmniRouter to be more user friendly, instead focusing on the abstraction of a Lemonade Omni Model.

Branding details:
- **OmniRouter** still exists, but it's just the internal mechanism for all of this. Not user facing.
- A **Lemonade Omni Model** is a true all-to-all omni-modal model from the perspective of the user. Under the hood, it is a collection of models.
- **LMX-Omni-52B-Halo** follows the new naming scheme: "halo" means large MoE model, and 52B is the sum of params of the components.

Implementation details:
- "Ultra Collection" and "Lite Collection" have been retired (suggested: false) and replaced with LMX-Omni-52B-Halo and LMX-Omni-5.5B-Lite, respectively.
- They are mostly the same, but use Qwen3.6-35B-MTP and Qwen3.5-4B-MTP models now.
- Added Qwen3.5-4B-MTP-GGUF to the models registry to support.
- Desktop Model Manager category renamed from "OmniRouter" to "Lemonade".
- Merged latest `main` (custom Omni Model UI from #1769) and re-applied the rebrand on top, so the new UI uses "Omni Model" + `user.*` naming consistently.
- Custom Omni Model docs live in `docs/guide/configuration/custom-models.md`; testing checklist in `docs/dev/app.md`; concept overview in `docs/dev/lemonade-omni.md`.

Naming scheme decoder ring (`LMX-Omni-<xB>-<class>`):
- **LMX**: Lemonade Mix
- **Omni**: True all-to-all omni-modal model
- **xB**: Total parameter count of the component models
- **Halo**: Based on a large MoE model (e.g., targeted to Strix Halo)
- **Lite**: Based on small models targeted at 32 GB APUs
- **Dense**: Based on a dense LLM targeted towards 32 GB dGPUs (we don't have one of these yet)

### DIFF
--- a/docs/dev/lemonade-omni.md
+++ b/docs/dev/lemonade-omni.md
@@ -2,22 +2,23 @@
 
 **Lemonade Omni Models** are one-click multimodal model bundles. A single selection in the Lemonade desktop app loads one LLM, one image model, one ASR model, and one TTS model — everything a multimodal agent needs to chat, generate images, transcribe audio, and speak responses out loud.
 
-Under the hood, Lemonade Omni Models are powered by **OmniRouter** — Lemonade's pattern for exposing each modality as an OpenAI-compatible tool that any existing LLM agent (Continue, OpenHands, Claude Code, your own app) can call against Lemonade's endpoints.
+Under the hood, Lemonade Omni Models are powered by **OmniRouter** — Lemonade's pattern for exposing each modality as an OpenAI-compatible tool that an existing LLM agent (Continue, OpenHands, Claude Code, your own app) can call against Lemonade's endpoints.
 
-You bring the LLM loop. Lemonade brings the tools.
+You bring the LLM loop. Lemonade brings the local tools.
 
 ## How OmniRouter works
 
-1. You describe the tools to your LLM in OpenAI tool-calling format.
+1. Describe the tools to your LLM in OpenAI tool-calling format.
 2. The LLM decides which tool to call and with what arguments.
-3. Your client code executes each `tool_call` against the corresponding Lemonade endpoint (`/v1/images/generations`, `/v1/audio/speech`, etc.) and feeds the result back as a `tool` message.
-4. The LLM continues, calling more tools or producing a final response.
+3. Your client executes each `tool_call` against the corresponding Lemonade endpoint, such as `/v1/images/generations` or `/v1/audio/speech`.
+4. The client sends the tool result back to the LLM as a `tool` message.
+5. The LLM continues until it either calls another tool or returns a final response.
 
-This is the standard OpenAI tool-calling loop. The tool schemas OmniRouter provides are plain JSON (no Lemonade-specific client library required), the endpoints they target are OpenAI-compatible, and the server returns standard response shapes.
+The tool schemas OmniRouter provides are plain JSON. They do not require a Lemonade-specific client library, and the endpoints they target use OpenAI-compatible request and response shapes.
 
 ## The omni models
 
-An **omni model** is a meta-model made up of components, registered with `recipe: "collection.omni"`. Selecting one in the Lemonade desktop app loads every component in a single click. Lemonade these:
+An **omni model** is a meta-model made up of components, registered with `recipe: "collection.omni"`. Selecting one in the Lemonade desktop app loads every component in a single click. Lemonade ships these:
 
 | Omni model | LLM | Image | ASR | TTS |
 |-----------|-----|-------|-----|-----|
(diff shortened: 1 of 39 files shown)


src/cpp/resources/server_models.json
```diff
@@ -1466,10 +1478,32 @@
             "height": 1024
         }
     },
-    "Ultra Collection": {
+    "LMN-Halo-Omni-52B": {
         "checkpoint": "",
         "recipe": "collection.omni",
         "suggested": true,
+        "components": [
+            "Qwen3.6-35B-A3B-MTP-GGUF",
```
fl0rianr, line 1486: Great to introduced the better/newer models here! But did you forget to update examples/lemonade_tools.py as well?

tools/model_sizes.py
```diff
@@ -80,6 +122,133 @@ def fetch_repo(repo_id: str, cache: dict):
     return cache[repo_id]
 
 
+def fetch_range(repo_id: str, filename: str, start: int, end: int):
+    quoted_name = urllib.parse.quote(filename)
+    url = f"https://huggingface.co/{repo_id}/resolve/main/{quoted_name}"
+    req = urllib.request.Request(url, headers={"Range": f"bytes={start}-{end}"})
```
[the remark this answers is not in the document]
fl0rianr: Valid! A fix could be?:

```
def hf_headers(extra=None):
    headers = dict(extra or {})
    tok = os.environ.get("HF_TOKEN")
    if tok:
        headers["Authorization"] = f"Bearer {tok}"
    return headers

def fetch_range(repo_id: str, filename: str, start: int, end: int):
    quoted_name = urllib.parse.quote(filename)
    url = f"https://huggingface.co/{repo_id}/resolve/main/{quoted_name}"
    req = urllib.request.Request(
        url,
        headers=hf_headers({"Range": f"bytes={start}-{end}"}),
    )
    with urllib.request.urlopen(req, timeout=30) as resp:
        return resp.read()
```

docs/dev/lemonade-omni.md
```diff
@@ -13,20 +15,35 @@ You bring the LLM loop. Lemonade brings the tools.
 
 This is the standard OpenAI tool-calling loop. The tool schemas OmniRouter provides are plain JSON (no Lemonade-specific client library required), the endpoints they target are OpenAI-compatible, and the server returns standard response shapes.
 
-## Collections
+## The omni models
 
-A **Collection** is a meta-model made up of components. An **omni collection** is the collection type used by OmniRouter, registered with `recipe: "collection.omni"`. Selecting an omni collection collection in the Lemonade desktop app loads one LLM + one image model + one ASR + one TTS — all the pieces OmniRouter's tools need in a single click.
+An **omni model** is a meta-model made up of components, registered with `recipe: "collection.omni"`. Selecting one in the Lemonade desktop app loads every component in a single click. Lemonade these:
```
fl0rianr: Typo: Lemonade these // Maybe ship these?

fl0rianr (changes requested):
Nice job @jeremyfowers! This rebrand looks directionally good and the registry/UI naming changes are mostly consistent. I’d fix two things before merging:

1. examples/lemonade_tools.py updated

2. The new `fetch_range()` path in `tools/model_sizes.py` should forward HF_TOKEN, same as `fetch_repo()`. Otherwise header-based parameter counting downgrades to unknown/lower-bound for private/gated HF repos even when auth is configured.

After those it looks good, but please let me remark this as well:

I think this PR should land after #1769.

#1769 is the larger feature PR and adds the custom OmniRouter / custom workflow UI, including substantial changes in `ModelManager.tsx` and `docs/dev/omni-router.md`. Since #1973 renames the user-facing OmniRouter/Collection concept to Lemonade Omni Models and moves the docs nav to `docs/dev/lemonade-omni.md`, these two PRs will almost certainly conflict after #1769 merges.

Could we merge #1769 first, then rebase/update this PR on top of it? In that rebase, I’d specifically expect:

- Preserve or update accordingly the custom workflow documentation from #1769 and move/adapt it into the new Lemonade Omni docs page.
- Update the new #1769 UI strings from “OmniRouter”, “Omni Model”, “Experience model”, and “custom workflow” where appropriate so they match the new Lemonade Omni naming.

If it was me I would always go for Omni Model to have the shorter version. Only in this case Lemonade is perfect:

Image

- Re-apply the `ModelManager.tsx` category label change on top of the custom workflow row actions and delete/edit behavior from #1769.
- Re-run the app regression/custom workflow tests after the rebase, since the conflict touches Model Manager and collection/workflow behavior.

I’m aligned with the direction of the rebrand, but I’d prefer not to merge this before #1769 because we’d otherwise make the larger feature PR harder to land and risk losing the custom workflow docs/UI wording during conflict resolution. Thanks!

src/cpp/resources/server_models.json
jeremyfowers: good catch!

jeremyfowers: fix in 6e1932c40

tools/model_sizes.py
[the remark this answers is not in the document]
jeremyfowers: fix in eed9c3b04

docs/dev/lemonade-omni.md
jeremyfowers: fix in 3b87b508a

jeremyfowers:
Updates pushed this session:

- Merged latest `main` (PR #1769 custom Omni Model UI), with the rebrand terminology re-applied so the new UI uses "Omni Model" / `user.` prefix instead of the "workflow" naming the PR's docs had drifted into.
- Renamed model prefix: `LMN-*` → `LMX-Omni-<xB>-<class>` (LMX = "Lemonade Mix"). New names: `LMX-Omni-52B-Halo`, `LMX-Omni-5.5B-Lite`.
- Renamed desktop category label: "Lemonade Omni" → "Lemonade".
- Moved the "Custom Omni Model" how-to to `docs/guide/configuration/custom-models.md` (next to the CLI flow) and the testing checklist to `docs/dev/app.md`; `lemonade-omni.md` links to both.
- Addressed all three review comments (HF auth, lemonade_tools.py, docs typo) — see threads for the specific fix commits.

fl0rianr (approved):
Thanks addressing this point, looks good! We can go for it, I see no blocker.

tiny, tiny nit:
Model Manager > Lemonade _Omni_ > LMX-Omni-5.5B-Lite > Download

While it should be Lemonade only now?

PR #1974, opened 2026-05-20
fix(app): store settings under user config dir, not ~/.cache

abn:
The Tauri host wrote `app_settings.json` to
`dirs::home_dir().join(".cache").join("lemonade")` on every platform. That's wrong on two counts:

- `.cache` is the wrong XDG basedir for user configuration (the OS may prune it on Linux), and renderer state like layout sizes, base URL, and API key isn't cache.
- The literal `~/.cache` path is Linux-flavored; macOS and Windows ended up with a hidden dotdir in $HOME instead of the platform's proper config location (`Library/Application Support` / `%APPDATA%`).

Switch to `dirs::config_dir()`, which resolves to `$XDG_CONFIG_HOME` on Linux, `~/Library/Application Support` on macOS, and `%APPDATA%` on Windows. Add a one-shot idempotent migration that moves an existing legacy file forward; when both locations exist the new file wins and the legacy file is left alone.

### DIFF
--- a/src/app/src-tauri/src/settings.rs
+++ b/src/app/src-tauri/src/settings.rs
@@ -1,7 +1,9 @@
-//! App-settings persistence. Reads and writes `~/.cache/lemonade/app_settings.json`,
-//! sanitizing values and applying defaults for missing/invalid fields before
-//! handing the struct back to the renderer. The JSON shape matches what the
-//! existing React renderer expects (each user-tunable field is a
+//! App-settings persistence. Reads and writes `app_settings.json` from the
+//! platform's per-user config directory (XDG_CONFIG_HOME on Linux,
+//! Library/Application Support on macOS, APPDATA on Windows), sanitizing
+//! values and applying defaults for missing/invalid fields before handing the
+//! struct back to the renderer. The JSON shape matches what the existing
+//! React renderer expects (each user-tunable field is a
 //! `{ value, useDefault }` envelope).
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


jeremyfowers:
There is already a `%APPDATA%\lemonade_server` on Windows that gets blown away every time the user upgrades. I would prefer to leave the app settings alone on Windows in the home cache dir, rather than introduce a new `%APPDATA%\lemonade`.

@Geramy should comment on the blast radius on macOS.

abn:
> There is already a `%APPDATA%\lemonade_server` on Windows that gets blown away every time the user upgrades. I would prefer to leave the app settings alone on Windows in the home cache dir, rather than introduce a new `%APPDATA%\lemonade`.

Is this the issue caused by roaming profiles on windows? iirc, using `LocalAppData` should fix that. Using `~/.cache` is not really a common pattern for windows iiuc. But happy to exclude this change for windows and/or reuse the existing directory as you prefer.

jeremyfowers:
> > There is already a `%APPDATA%\lemonade_server` on Windows that gets blown away every time the user upgrades. I would prefer to leave the app settings alone on Windows in the home cache dir, rather than introduce a new `%APPDATA%\lemonade`.
> 
> Is this the issue caused by roaming profiles on windows? iirc, using `LocalAppData` should fix that. Using `~/.cache` is not really a common pattern for windows iiuc. But happy to exclude this change for windows and/or reuse the existing directory as you prefer.

> Is this the issue caused by roaming profiles on windows?

Pretty sure its a problem of our own making, but a problem nonetheless.

> iirc, using LocalAppData should fix that. 

as long as both of the following are true, I'm good

1. there isn't an AppData/Local/lemonade_server and AppData/Local/lemonade sitting right next to each other confusing people
2. config isn't being placed into the AppData/Local/lemonade_server that our MSI installer deletes and re-generates each time it runs

Geramy:
@abn keep in mind macOS gets installed as a service and doesn't have a home so ~/ won't work.

src/app/src-tauri/src/settings.rs
```diff
@@ -177,8 +179,69 @@ impl Default for AppSettings {
 
 // ---------- Path helpers ----------
 
+/// Per-user config directory for the desktop app. Resolves to
+/// `$XDG_CONFIG_HOME/lemonade` (Linux), `~/Library/Application Support/lemonade`
```
Geramy, line 208: what loads the user config the tray icon or the service, or the app? Did you verify this on Mac? I believe the server gets installed as a service and so ~/ doesn't exist and the tray icon / app are also a type of daemon launch items.

Geramy (changes requested):
Please verify this is true

superm1 (approved):

jeremyfowers:
@kpoineal since this is targeting the rust backend is it safe to merge, even though GUI3 is coming?

kpoineal:
GUI3 uses the same Tauri settings.rs and window.api settings contract, so this path/migration change should integrate cleanly without affecting GUI3 behavior. I'm good with it.

jeremyfowers:
> GUI3 uses the same Tauri settings.rs and window.api settings contract, so this path/migration change should integrate cleanly without affecting GUI3 behavior. I'm good with it.

Thanks! @abn please respond to the open reviewer comment or close the PR.

superm1:
I think we have the same problem with the server.  We are storing configuration items in a cache directory.

I don't think any of our JSON files should really be in `~lemonade/.cache`.  They should be in `~lemonade/.config`

abn:
Sorry folks for the delay; will pick this up again soon. I still want the change, but this change might have some broader impact that I might not have considered. In particular migration and existing assumptions that have formed due to the current structure.

superm1:
Take a look at https://github.com/lemonade-sdk/lemonade/pull/3028

This will do it more widely.

abn:
@superm1 great; commented there. I will close this in favor of that.

PR #1995, opened 2026-05-22
Smooth displayed download speed

fl0rianr:
The download manager uses the currently reported download speed to estimate the remaining time. Because this value is calculated from individual progress snapshots, it can be very jumpy.

Users may watch this value while waiting to use an exciting new model, so the unstable speed display makes the download UX feel less reliable than expected.

This PR smooths the displayed speed with a small exponential moving average while keeping the progress calculation unchanged.

- Smooth displayed download speed using a lightweight EMA.
- Keep existing progress, percent, and byte-count logic unchanged.
- Reset speed to `0` for terminal download states as before.
- No backend/API changes.


bitgamma (approved):
lgtm

PR #1997, opened 2026-05-22
Remove more cases of hardcoded /tmp or predictable paths to mitigate …

superm1:
…TOCTOU attacks

### DIFF
--- a/src/cpp/server/utils/path_utils.cpp
+++ b/src/cpp/server/utils/path_utils.cpp
@@ -401,6 +401,18 @@ std::string get_runtime_dir() {
     char temp_path[MAX_PATH];
     GetTempPathA(MAX_PATH, temp_path);
     return std::string(temp_path);
+#elif defined(__APPLE__)
+    std::error_code ec;
+    fs::path base = fs::temp_directory_path(ec);
+    if (!ec && !base.empty()) {
+        fs::path lemon_dir = base / "lemonade";
+        ec.clear();
+        fs::create_directory(lemon_dir, ec);
+        if (!ec || fs::is_directory(lemon_dir)) {
+            return lemon_dir.string();
+        }
+    }
+    throw std::runtime_error("Unable to resolve writable runtime directory on macOS");
 #else
     const char* xdg = std::getenv("XDG_RUNTIME_DIR");
     if (xdg && xdg[0] != '\0') {
--- a/src/cpp/server/utils/path_utils.cpp
+++ b/src/cpp/server/utils/path_utils.cpp
@@ -117,8 +117,6 @@ std::string get_executable_dir() {
         fs::path exe_path(buffer);
         return exe_path.parent_path().string();
     }
-    // Fallback: return current directory
-    return ".";
 #elif defined(__APPLE__)
     char buffer[PATH_MAX];
     uint32_t size = sizeof(buffer);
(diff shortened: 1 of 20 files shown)


src/cpp/server/utils/path_utils.cpp
```diff
@@ -428,8 +433,19 @@ std::string get_runtime_dir() {
...
+    fs::path temp_dir = fs::temp_directory_path(ec);
+    if (!ec && !temp_dir.empty()) {
+        fs::path lemon_dir = temp_dir / "lemonade";
+        ec.clear();
+        fs::create_directory(lemon_dir, ec);
+        std::error_code ec2;
+        if (!ec || fs::is_directory(lemon_dir, ec2)) {
+            return lemon_dir.string();
+        }
+        return temp_dir.string();
```
fl0rianr: Nice direction overall. One question on the fallback behavior here: if creating <temp>/lemonade fails, this returns the temp root itself. Since callers append predictable names like lemonade-server.log or lemonade_<app>.lock, this seems to partially reintroduce the shared-temp-root behavior this PR is trying to avoid.

Was this fallback kept to avoid breaking existing behavior? I’m wondering if we should instead fail here if we can’t create the private runtime dir.

superm1: I think you're right - just dropping fallback path makes more sense.

I don't see any reason that there would be a system without XDG session support we need to also support.

fl0rianr: Thanks, for looking into this co quickly. I think the failed test are helpful because it makes the trade-off very visible. Dropping the fallback means CI now fails in many places where XDG_RUNTIME_DIR is not set, so those jobs were probably relying on the old implicit temp behavior.

I don’t think we need to go back to predictable files in shared /tmp, but we may need one of two things:

1. Set up a safe XDG_RUNTIME_DIR in CI before starting lemond, for example a private directory under `$RUNNER_TEMP` with 0700 permissions.
2. Or keep a code fallback, but only if it creates and verifies a private runtime directory instead of returning the temp root.

For CI, something like this before server startup might be enough:

```bash
mkdir -p "$RUNNER_TEMP/xdg-runtime"
chmod 700 "$RUNNER_TEMP/xdg-runtime"
echo "XDG_RUNTIME_DIR=$RUNNER_TEMP/xdg-runtime" >> "$GITHUB_ENV"
export XDG_RUNTIME_DIR="$RUNNER_TEMP/xdg-runtime"
```

That would keep the security direction while making the CI environment explicit.

superm1: Thanks, totally agree with the direction.

fl0rianr:
Looks much better now. Most of the remaining failures look unrelated or infra-ish to me, especially the checkout failures. The one I’d still double-check is the Docker smoke test: the container starts, but waiting for the server fails. If that container does not set XDG_RUNTIME_DIR, it may need the same private runtime-dir setup inside the container or entrypoint.

fl0rianr:
Thanks for the hard work! The security direction looks good now, but I see one possible macOS regression before approving: get_runtime_dir() now throws whenever XDG_RUNTIME_DIR is missing, and that path also applies to macOS. lemond calls configure_application_logging(..., direct_server) during startup, which uses get_runtime_dir() for file logging on non-systemd platforms.

The macOS server plist does not set XDG_RUNTIME_DIR, so launchd-started lemond may fail before it comes up. The tray path looks similar, since lemonade-tray calls SingleInstance, which also uses get_runtime_dir(), but the tray plist / launcher do not set XDG_RUNTIME_DIR either.

What do you think? Could we either make get_runtime_dir() platform-specific for macOS, or set up a secure runtime dir in the macOS plists/install scripts as well?

And a really tiny one:  The docs still mention falling back to the system temp directory, but the code now throws.

fl0rianr (approved):

PR #1998, opened 2026-05-23
Fix(Omni): image tool options for Omni image generation

fl0rianr:
## Summary

This PR fixes Omni image tool handling in collection chat so image generation/edit requests can pass explicit image options without changing the default OOB image size or letting prompt content accidentally override the canvas size.

## Changes

* Expose optional image parameters for `generate_image` and `edit_image`: `size`, `width`/`height`, `seed`, `steps`, `cfg_scale`, `sample_method`, and `flow_shift`
* Forward supported image options to `/images/generations` and `/images/edits`
* Keep the default collection image size at `512x256` unless explicit, valid canvas dimensions are provided
* Avoid inferring canvas size from prompt text such as `100x100 monitor` or pixel-art content descriptions
* Add guards for duplicate image generation in a single assistant turn
* Make image edit-vs-generate routing **planner-driven**: remove the previous heuristic that auto-switched a `generate_image` call to `edit_image` whenever a prior image existed and the model could edit. The planner now chooses the tool, steered by strengthened `edit_image` prompt guidance, and `edit_image` errors clearly when no source image is available.
* Keep planner tool instructions conditional so the planner only receives tool guidance for capabilities available in the current collection

### Behavior change

Because the automatic `generate_image → edit_image` switch was removed, an "edit this" style request now performs an edit only when the planner actually selects `edit_image`. The strengthened prompt guidance is what drives that choice; a planner that still emits `generate_image` will produce a fresh image rather than silently editing the previous one.

## Testing

Added regression coverage for:

* image option schema limits and default-size behavior
* avoiding prompt-derived canvas size inference
* conditional planner instructions
* duplicate `generate_image` suppression
* planner-driven (not regex-driven) edit tool selection


examples/lemonade_tools.py
```diff
@@ -18,13 +18,22 @@
 
 Usage:
     python examples/lemonade_tools.py "Generate an image of a sunset"
+    python examples/lemonade_tools.py "Generate a 2:1 cyberpunk street with seed 1234 and 20 steps"
```
jeremyfowers: The purpose of lemonade_tools.py is to be a minimal example that shows it is reasonably easy to integrate omnirouter tools into an application. This changeset increases the LoC in the file by more than 2x, which I think makes it no longer a minimal example.

If you think its important to have a python example of an OmniRouter-based image editing workflow it is probably a good idea to start a new script for that specifically, rather than extending the minimal example.

src/app/src/renderer/utils/collectionImageConfig.ts
```diff
@@ -1,14 +1,11 @@
...
-// 2:1 aspect ratio with 64-aligned dimensions for compatibility across
-// SD/SDXL/Flux variants. See SDServer::resolve_size for the server-side
-// passthrough.
+// Keep this aligned with the image tool executor fallback so omitted planner
+// size args do not drift from the default generation canvas. The 640x320
+// default is intentionally 2:1: wide enough for common UI/preview use cases
+// without pushing the local image backend into an expensive resolution.
 
-export const COLLECTION_IMAGE_SIZE = '512x256';
+export const COLLECTION_IMAGE_SIZE = '640x320';
```
jeremyfowers: 512x256 = 131072 pixels
640x320 = 204800 pixels

This change would (presumably) make the OOB LMX image experience 56% slower. IMO it is already too slow, even on high end strix halo hardware, so I am not comfortable with this change.

examples/lemonade_tools.py
jeremyfowers: PS. the GUI app itself is under no constraint to be a minimal reference, and its ok for the python and gui app to diverge in features.

src/app/src/renderer/utils/collectionImageConfig.ts
fl0rianr: It's 512x256 again

examples/lemonade_tools.py
fl0rianr: simplified again.

src/app/src/renderer/utils/toolDefinitions.json
```diff
@@ -1,75 +1,198 @@
...
+            },
+            "size": {
+              "type": "string",
+              "description": "Optional output image size as WIDTHxHEIGHT pixels. Use exact user dimensions when provided; otherwise omit to use the {image_size} default. For aspect-ratio/orientation requests, pass a concrete size only when obvious."
+            },
+            "width": {
+              "type": "integer",
+              "description": "Optional output width in pixels. Use with height when dimensions are provided separately.",
+              "minimum": 64
+            },
```
jeremyfowers: Optional, non-blocking: the `width`/`height` schema sets a `minimum` but no `maximum`, while the executor caps at `MAX_IMAGE_DIMENSION` (2048) and silently drops out-of-range values back to the default. Consider adding a matching `maximum` so the schema contract aligns with the resolver (applies to the `height` block here and both width/height blocks under `edit_image` too):

```suggestion
            "width": {
              "type": "integer",
              "description": "Optional output width in pixels. Use with height when dimensions are provided separately.",
              "minimum": 64,
              "maximum": 2048
            },
```

_— Filed by Jeremy Fowers; comment drafted by Claude Code on his behalf._

src/app/src/renderer/utils/lemonadeTools.ts
```diff
@@ -209,6 +271,122 @@ async function executeImageTool(
...
+    if (SIZE_HINT_TO_SIZE[orientation]) return SIZE_HINT_TO_SIZE[orientation];
+  }
+
+  for (const [hint, size] of Object.entries(SIZE_HINT_TO_SIZE)) {
+    const pattern = new RegExp(`\\b${hint}\\b`);
+    if (pattern.test(text)) return size;
+  }
+
+  return '';
+}
```
jeremyfowers: This reads `args.aspect_ratio` and `args.orientation`, but neither is a declared property and both schemas set `additionalProperties: false` — so the planner can never pass them (the test even asserts they are absent). Those two branches, plus their entries in `textParts`, are dead; only the `size`/`prompt` text scans can ever fire. Suggest dropping them:

```suggestion
function inferSizeFromRatioOrOrientation(args: Record<string, any>): string {
  const text = ['size', 'prompt']
    .map(key => typeof args[key] === 'string' ? args[key] : '')
    .filter(Boolean)
    .join(' ')
    .toLowerCase();

  for (const [ratio, size] of Object.entries(ASPECT_RATIO_TO_SIZE)) {
    const [left, right] = ratio.split(':');
    const pattern = new RegExp(`(?<!\\d)${left}\\s*[:/]\\s*${right}(?!\\d)`);
    if (pattern.test(text)) return size;
  }

  for (const [hint, size] of Object.entries(SIZE_HINT_TO_SIZE)) {
    const pattern = new RegExp(`\\b${hint}\\b`);
    if (pattern.test(text)) return size;
  }

  return '';
}
```

_— Filed by Jeremy Fowers; comment drafted by Claude Code on his behalf._

src/app/src/renderer/utils/lemonadeTools.ts
```diff
@@ -209,6 +271,122 @@ async function executeImageTool(
...
+
+  const inferredSize = inferSizeFromRatioOrOrientation(args);
+  if (inferredSize) return inferredSize;
+
+  return DEFAULT_IMAGE_SIZE;
+}
+
+function parseSizeFromText(text: string): string {
+  if (!text) return '';
+  const match = text.match(/(?<!\d)(\d{2,4})\s*(?:x|×|by)\s*(\d{2,4})(?!\d)/i);
```
jeremyfowers, line 257: Concern: `parseSizeFromText` is also run against `args.prompt`, so a `WxH` token that is part of the *described subject* rather than the canvas can hijack the output size — e.g. a prompt like "a 100x100 pixel-art grid" would set the image to `100x100`. The `>=64` guard in `isValidImageDimension` only filters tiny values, not mid-range false positives. Worth either restricting prompt-derived sizes more tightly, or adding a comment that this is a best-effort heuristic and `size`/`width`/`height` args take precedence.

_— Filed by Jeremy Fowers; comment drafted by Claude Code on his behalf._

jeremyfowers:
Two PR-description corrections (drafted by Claude Code on behalf of Jeremy Fowers):

1. The summary says the fallback size was changed to **square `512x512`**, but the code sets `DEFAULT_IMAGE_SIZE = COLLECTION_IMAGE_SIZE = '512x256'` (2:1 landscape — the first preset). The default is not square; please reconcile the description with the code.
2. The Changes list mentions updating `examples/lemonade_tools.py`, but that file is not in the diff (only `lemonadeTools.ts`, `toolDefinitions.json`, and the test changed). Either re-add the example change or drop it from the description.

src/app/src/renderer/utils/lemonadeTools.ts
```diff
@@ -5,6 +5,51 @@ import { getCollectionComponents } from './collectionModels';
 import { COLLECTION_IMAGE_SIZE } from './collectionImageConfig';
```
jeremyfowers, line 5: Image

The image above shows two issues:
1. I would only expect this prompt to produce one output image
2. I did not ask for the image to be 100x100px, but it is (related to https://github.com/lemonade-sdk/lemonade/pull/1998/changes#r3338167737):

```
2026-06-01 22:00:00.322 [Debug] (SDServer)   "prompt": "A close-up of a retro computer monitor screen displaying a very pixelated, blocky image, simulating a very low 100x100 resolution display. <sd_cpp_extra_args>{\"sample_params\":{\"guidance\":{\"txt_cfg\":1.0},\"sample_steps\":4}}</sd_cpp_extra_args>",
2026-06-01 22:00:00.322 [Debug] (SDServer)   "response_format": "b64_json",
2026-06-01 22:00:00.322 [Debug] (SDServer)   "size": "100x100"
2026-06-01 22:00:00.322 [Debug] (SDServer) }
```

jeremyfowers: Neither issue seems to happen on `main`:

Image

jeremyfowers (changes requested):
The double-image bug came up a few times for me, and that one is a blocker. Curious what you think of the other comments.

fl0rianr:
Thanks again for the review. I pushed a more conservative follow-up:

- width/height schema now matches the 2048 executor cap.
- Prompt text is no longer parsed for WxH canvas sizes, so content like “100x100 monitor” stays in the prompt and does not become the output size.
- Explicit dimensions must be 64-aligned, matching backend expectations.
- Repeated generate_image calls in one assistant turn are skipped to prevent duplicate images.
- Likely edit intents are rerouted to edit_image when an edit-capable tool is available; otherwise they return a clear “editing unavailable” message instead of generating a new image.

Local app regression tests pass: 36 passed, 0 skipped, 0 failed.

I also updated the implementation to keep the existing 512x256 OOB default and avoid prompt/orientation heuristics changing the canvas size.

jeremyfowers:
> Repeated generate_image calls in one assistant turn are skipped to prevent duplicate images.

Wouldn't this block someone from prompting "make me one image of a cat and another image of a dog"?

For something like Halo Tales I could see that being useful, for example producing multiple images representing different "camera angels" in a scene.

jeremyfowers:
I have a broader question as well: doesn't parsing the input and modifying the prompt take agency away from our agent? For example parsing for the word "edit" instead of letting the agent decide when it ought to edit. 

I would expect the planner LLM to typically do a better job than any regular expression we could implement.

As an alternative, is it possible tell the agent it has a bunch of choices of image resolution and aspect ratio, to prefer a default, but to respect the user's wishes if they specifically ask for a certain type of image?

fl0rianr:
Good point @jeremyfowers  — I agree that the UI should not override the planner’s tool choice with regex-based edit detection. Current implementation was just not good enough!

I changed the follow-up so the planner keeps agency over generate_image vs edit_image. The UI no longer rewrites a generate_image tool call into edit_image; instead the planner prompt/tool descriptions explain when to use edit_image, what common canvas sizes are available, and that the default should be preferred unless the user explicitly asks for a canvas size/aspect ratio. Which works fine.

I also narrowed the duplicate-image guard: it now skips only exact repeated generate_image requests with the same prompt/options in the same turn. Distinct requests like “one image of a cat and another image of a dog” or multiple camera angles should still run.

src/app/src/renderer/utils/toolDefinitions.json
```diff
@@ -3,77 +3,205 @@
...
         "name": "generate_image",
-        "description": "Generate a NEW image from scratch based on a text description. Use this ONLY when the user asks you to create an entirely new image. Do NOT use this to modify or change an existing image — use edit_image instead.",
+        "description": "Generate a NEW image from scratch based on a text description. Use this only for a new image, not for modifying an existing image.",
         "parameters": {
           "type": "object",
           "properties": {
-            "prompt": { "type": "string", "description": "A detailed description of the image to generate. Output size is fixed at {image_size}." }
+            "prompt": {
+              "type": "string",
+              "description": "A detailed description of the image to generate. Keep content details such as monitor/screen resolution in the prompt; do not move them into size."
```
jeremyfowers: "Keep content details such as monitor/screen resolution in the prompt; do not move them into size."

That was kind of a facetious example on my part meant to stress test the previous regular expression system. I'm uncertain if this corner case is worth spending system prompt on?

src/app/src/renderer/utils/toolDefinitions.json
```diff
@@ -3,77 +3,205 @@
...
+              "description": "Optional output width in pixels. Use with height when dimensions are provided separately.",
+              "minimum": 64,
+              "maximum": 2048
+            },
+            "height": {
+              "type": "integer",
+              "description": "Optional output height in pixels. Use with width when dimensions are provided separately.",
+              "minimum": 64,
+              "maximum": 2048
+            },
```
jeremyfowers, line 22: Why is there `size` and `width` and `height`? Should it be just `size`, or just `width` and `height`?

src/app/src/renderer/utils/toolDefinitions.json
```diff
@@ -3,77 +3,205 @@
...
           },
-          "required": ["prompt"]
+          "required": [
+            "prompt"
+          ],
+          "additionalProperties": false
         }
-      }
+      },
+      "prompt_guidance": "For image generation, omit size by default to use {image_size}. Only set size or width+height when the user explicitly asks for output/canvas dimensions or aspect ratio. Common safe canvas sizes include {image_size} default/wide, 512x512 square, 512x768 portrait, 768x512 landscape, 1024x576 16:9, and 576x1024 9:16. Keep content details such as screen, monitor, icon, grid, or pixel-art resolution in the prompt; do not move them into size."
```
jeremyfowers: Seems redundant with the string for `size`, can we drop one or the other?

src/app/src/renderer/utils/toolDefinitions.json
```diff
@@ -3,77 +3,205 @@
...
+              "description": "Optional output width in pixels. Use with height when dimensions are provided separately.",
+              "minimum": 64,
+              "maximum": 2048
+            },
+            "height": {
+              "type": "integer",
+              "description": "Optional output height in pixels. Use with width when dimensions are provided separately.",
+              "minimum": 64,
+              "maximum": 2048
+            },
```
jeremyfowers, line 73: I'm a bit confused by this change, can we really change the dimensions with edit now? Is that desirable?

src/app/src/renderer/utils/toolDefinitions.json
```diff
@@ -3,77 +3,205 @@
...
-      "prompt_guidance": "When you see '[User provided audio file #N]' in a message, it means the user sent audio data. Call the transcribe_audio tool to transcribe it — the audio data is handled automatically by the system.",
+      "requires_labels": [
+        "transcription"
+      ],
+      "prompt_guidance": "When you see '[User provided audio file #N]' in a message, it means the user sent audio data. Call the transcribe_audio tool to transcribe it \u2014 the audio data is handled automatically by the system.",
       "function": {
         "name": "transcribe_audio",
-        "description": "Transcribe audio to text (speech-to-text). Use this when the user provides an audio file or when you see '[User provided audio file #N]' placeholders in the conversation. The audio data is automatically provided by the system — just call this tool with the language parameter.",
+        "description": "Transcribe audio to text (speech-to-text). Use this when the user provides an audio file or when you see '[User provided audio file #N]' placeholders in the conversation. The audio data is automatically provided by the system \u2014 just call this tool with the language parameter.",
```
jeremyfowers:
```suggestion
        "description": "Transcribe audio to text (speech-to-text). Use this when the user provides an audio file or when you see '[User provided audio file #N]' placeholders in the conversation. The audio data is automatically provided by the system, just call this tool with the language parameter.",
```

seems simpler

src/app/src/renderer/utils/toolDefinitions.json
```diff
@@ -3,77 +3,205 @@
...
           "type": "object",
           "properties": {
-            "prompt": { "type": "string", "description": "A detailed description of the image to generate. Output size is fixed at {image_size}." }
+            "prompt": {
+              "type": "string",
+              "description": "A detailed description of the image to generate. Keep content details such as monitor/screen resolution in the prompt; do not move them into size."
+            },
+            "size": {
+              "type": "string",
+              "description": "Optional output canvas size as WIDTHxHEIGHT pixels. Omit by default to use {image_size}. Use only when the user explicitly asks for canvas dimensions or aspect ratio. Common choices include 512x512 square, 512x768 portrait, 768x512 landscape, 1024x576 16:9, and 576x1024 9:16. Keep content details like a monitor or pixel-art resolution in the prompt. Dimensions must be multiples of 64."
```
jeremyfowers: > Dimensions must be multiples of 64.

The models only require multiples of 8, is this overly strict?

My understanding is also that we fall back to the default dimensions if the planner doesn't produce a valid dimension. Why not just round up to the nearest multiple of 8?

jeremyfowers (commented):
Image

Love it! Sorry about the delay in merging.

jeremyfowers (approved):

PR #2003, opened 2026-05-23
fix(rocm): repair stable runtime when backend is update_required

fl0rianr:
## Summary

Fix the ROCm Stable repair path when the selected backend is already reported as `update_required`.

Before this change, the frontend/backend preflight could depend on how the model was loaded, and the backend install path could still skip TheRock if the local TheRock `version.txt` marker matched. In that case the install/repair job could become a short no-op even though the selected ROCm Stable backend needed repair.

This PR keeps the normal TheRock version marker fast path for regular installs, but makes the repair path remove the existing Lemonade-managed TheRock runtime directory before reinstalling it.

## Problem

ROCm Stable uses two pieces:

1. the llama.cpp / sd.cpp backend binary
2. the TheRock ROCm runtime used by that backend

The normal install path uses TheRock's local `version.txt` marker to avoid downloading the runtime again. That is fine for the common case.

The issue is the repair case: when the backend is already reported as `update_required`, a stale matching TheRock `version.txt` can still make `install_therock()` return early. That means the repair command may not actually repair the Lemonade-managed TheRock runtime.

## Changes

* Use saved model recipe_options during direct-load preflight so the preflight checks the same selected backend that the model load will use.
* Only include TheRock in the backend install job when the existing `will_install_therock(...)` logic says TheRock is applicable for this OS, ROCm architecture, and system ROCm state.
* In the `update_required` repair path, remove the existing Lemonade-managed TheRock runtime directory before calling the existing `install_therock_if_needed(...)` path.
* Do not add a separate ROCm runtime filesystem marker check in `system-info`.

## Testing

1. Download and installing works fine as before
2. If update is required the install is always triggered no only in special model load scenarios


superm1 (commented):
Is this actually a poor mans consistency check?  If we check the sha at download time and the return code at extract time do we need this much?  As long as we clean up a partial extraction it should be fine I would think.

fl0rianr:
Thanks for having a look!

1. Yes, that is true. But I still think it is helpful if the installation is broken after extraction, for example by user error or a partial/bad state on disk.

2. The SHA check is quite recent and, as far as I understand, not yet fully there for this backend/runtime case.

3. The other update logic is still needed independent of this extra check. Before, it depended on how the model was loaded and we could end up checking/installing the wrong backend state, or not really repairing the broken ROCm stable runtime even though system-info already reported `update_required`.

Possible directions I see:

* remove only this extra runtime check and keep `version.txt` as the only source of truth

* do a simpler check, e.g. require that some expected runtime content is present in the folder

* or go for a ROCm-included test binary and run it. That would probably be the safest check, but also slower and more invasive, while the current check is only a few extra lines and should have basically no runtime impact.

My preference would be to keep the ROCm installation / repair path a bit extra safe here, because version.txt alone is exactly the case that made the broken runtime look installed.

What would be your take?

superm1:
It's a lot of code; and I really worry it falls apart with other scenarios that we support today:
1. `pip` installed
2. `apt` installed
3. Tarball extracted already in the filesystem.

I didn't realize the sha check isn't part of the runtime yet.  I honestly think the right direction is to add it there.  Nearly all of this code falls off if you have consistency checks the moment you extract it.

fl0rianr:
Thanks, that makes sense.

I reworked the PR in that direction. I removed the broad runtime marker / filesystem consistency check from system-info and kept the existing update_required behavior.

The PR is now more focused on the concrete repair/no-op case:

* keep the load/preflight fix, so we check the same backend the model load will actually use
* only consider TheRock when the existing `will_install_therock(...)` logic says it is applicable
* if that selected backend is already `update_required`, remove the Lemonade-managed TheRock dir before going through the existing `install_therock_if_needed(...)` path

So this no longer tries to validate all possible ROCm layouts from system-info. It just prevents a stale TheRock `version.txt` from making the repair path skip the reinstall.

I also agree that the more complete long-term fix should live closer to the runtime download/extract path, with checksum support for the runtime assets and cleanup before version.txt can make an incomplete install look valid.

superm1 (approved):

PR #2004, opened 2026-05-23
fix(vLLM): make startup timeout adjustable, increased global_timeout

fl0rianr:
## Summary

- Use the configurable global timeout for vLLM server readiness checks instead of a hardcoded startup timeout of 600 s
- Set the default `global_timeout` to 600 s.

vLLM model loading can take longer than the hard-coded 600-second default, depending on model size and hardware.
To avoid a timeout removal global_timeout is used to introduce the option for adaption.
A user would expect eliminating a timeout error with increasing global_timeout - this patch ensures that behavior.

A 600-second timeout has been used for vLLM and the default global_timeout is increased to that level - a level that was in use also before and is considered acceptable based on prior discussion, while still keeping a bounded startup timeout.

## Testing

- Verified that vLLM startup/readiness now uses the configured global_timeout.
- Verified `global_timeout=600` can be set and used for longer vLLM load times.

### DIFF
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -24,7 +24,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
   "port": 13305,
   "host": "localhost",
   "log_level": "info",
-  "global_timeout": 300,
+  "global_timeout": 600,
   "max_loaded_models": 1,
   "no_broadcast": false,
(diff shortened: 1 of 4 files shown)


ramkrishna2910 (approved):

fl0rianr:
@ramkrishna2910 Thanks!

PR #2009, opened 2026-05-24
fix(vllm): report streaming TTFT and TPS

### LINKED ISSUE #1911: vllm: TPS reporting still does not work
## Summary

TPS / tokens-per-second reporting for the vLLM backend still does not appear to work.

## Context

Reported by @lenne0815 in Discord on May 16, 2026:

> tps for vllm still doesnt seem to work ;)

Jeremy Fowers asked that this be tracked in the known vLLM issues milestone.

## Expected behavior

When using the vLLM backend, Lemonade should report TPS consistently in the relevant UI/API/status output, matching the behavior users expect from the other backends.

## Actual behavior

TPS reporting for vLLM appears to be missing, incorrect, or otherwise non-functional.

## Notes

Needs repro details added once available: Lemonade version, vLLM backend bundle/version, GPU target, model, launch command, and where TPS was expected to appear.

fl0rianr:
## Summary

Fix vLLM streaming telemetry so `/stats` reports token usage, TTFT, and TPS after streamed chat/completion requests.

The change keeps the fix narrow:
- request vLLM streaming usage via `stream_options.include_usage`
- record TTFT from the first streamed SSE chunk when the backend does not provide it
- compute TPS from streamed completion tokens when vLLM does not report TPS directly

## Testing

Tested locally with the vLLM backend:
- streamed chat/completion request completes successfully
- output_tokens, time_to_first_token and tokens_per_second reported

Fixes #1911


ramkrishna2910:
@claude review

ramkrishna2910:
@fl0rianr PR works as expected. A few comments:

 Behavioral:
  - include_usage is forced on unconditionally, overwriting a client-supplied stream_options.include_usage: false. Consequence: every
  streamed vLLM response now ends with a usage frame in the client's SSE, even if they explicitly opted out. It's harmless and matches
  typical OpenAI behavior, but it's a silent override of a client request — worth a one-line code comment so the next reader knows it's
  intentional.

  Minor (correctness, sub-second, probably fine):
  - The TPS elapsed clock starts in VLLMServer::forward_streaming_request (before the JSON re-parse + base call), while the TTFT clock
  starts later inside StreamingProxy::forward_sse_stream. So decode_seconds = elapsed − ttft slightly over-counts elapsed. Negligible in
  practice (the warm numbers looked sane), just noting the two clocks aren't the same origin.
  - Cold-start TTFT reads as 3.6s because the first post-load request absorbs vLLM's graph-capture/JIT. The value is correct (real
  time-to-first-token) but could mislead someone reading /stats once right after load. If /stats is surfaced in UI, consider a doc note;
  not a code issue.

  Nits:
  - Stray extra blank line before the closing brace in vllm_server.h.
  - The <chrono> include is added in both files — fine.

ramkrishna2910 (approved):
Thanks for the contribution.

PR #2011, opened 2026-05-25
Update URL for CUDA backend for llama.cpp (llamacpp:cuda)

Phqen1x:
## Summary

Updates the URL for downloading the CUDA backend, depends on https://github.com/lemonade-sdk/llama.cpp/pull/5

### DIFF
--- a/docs/guide/configuration/llamacpp.md
+++ b/docs/guide/configuration/llamacpp.md
@@ -34,7 +34,7 @@ Lemonade uses [llama.cpp](https://github.com/ggerganov/llama.cpp) as its primary
 - **Hardware**: NVIDIA GPUs with Compute Capability 7.5+ (Turing, Ampere, Ada, Hopper, Blackwell)
 - **Use Case**: NVIDIA GPU-optimized inference
 - **Performance**: Optimized for NVIDIA hardware, typically outperforms Vulkan on supported GPUs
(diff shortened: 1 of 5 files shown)


kenvandine:

fl0rianr:
Thanks @Phqen1x! But has to be re-based first. Only this action is the proper solution for the merge conflicts.

fl0rianr:
Very good @Phqen1x! But please make sure you don't confuse the name, I think `lemonade-sdk/llama.cpp-builds` might be the right one once your PR is merged there. 

Once that is done and you can set it to ready, please adapt src/cpp/resources/backend_versions.json accordingly, currently it is `"cuda": "b1016",` which might not be the right version by then.

kenvandine:
> Very good @Phqen1x! But please make sure you don't confuse the name, I think `lemonade-sdk/llama.cpp-builds` might be the right one once your PR is merged there. 
> 
> Once that is done and you can set it to ready, please adapt src/cpp/resources/backend_versions.json accordingly, currently it is `"cuda": "b1016",` which might not be the right version by then.

No, we were told to use llama.cpp and llama.cpp-builds needs to get deleted once we migrate away. Too many repositories

kenvandine (approved):
Looks great, thanks!

fl0rianr (approved):
Great! I'm glad we'll have this ready for 10.7! Thanks both of you!

PR #2013, opened 2026-05-25
feat: add mlx-lemon backend  for ROCm (initial)

### LINKED ISSUE #1642: [Feature] MLX Engine ROCm backend — 83% faster than Vulkan on Strix Halo
## Summary

Proposing MLX Engine (lemon-mlx-engine) as a new backend in Lemonade SDK. We have it running on AMD Strix Halo (gfx1151) via ROCm 7.12 with significant performance gains over both Vulkan llama.cpp and vLLM.

### MLX Engine ROCm (b1004-tech-preview, 5 runs, 2 warmup)

| Model | tok/s | ±stddev |
|-------|------:|--------:|
| Qwen3-0.6B-4bit | 151.2 | ±0.1 |
| Qwen3-1.7B-4bit | 66.4 | ±0.0 |
| Qwen3-4B-4bit | 46.9 | ±0.0 |
| Phi-4-mini-instruct-4bit | 38.3 | ±0.0 |
| Qwen3-8B-4bit | 21.7 | ±0.0 |

### Head-to-Head (same hardware, same day)

| Model | MLX | vLLM | Vulkan | Advantage |
|-------|----:|-----:|-------:|-----------|
| Qwen3-0.6B | 151.2 | 116.7 | 82.5 | +83% vs vulkan |
| Qwen3-4B | 46.9 | 25.4 | — | +85% vs vllm |
| Qwen3-8B | 21.7 | 12.3 | — | +76% vs vllm |
| Phi-4-mini | 38.3 | 25.1 | — | +53% vs vllm |

### NPU (FastFlowLM, running simultaneously with GPU)

| Model | tok/s | Hardware |
|-------|------:|----------|
| Qwen3-0.6B-FLM | 94.4 | NPU (xdna2, 50 TOPS) |
| Llama-3.2-1B-FLM | 61.7 | NPU |
| Qwen3-8B-FLM | 10.8 | NPU |

## Why MLX Engine

- **Pure C++ static binary** — no Python, no Torch, no Triton JIT
- **Cold start in seconds** vs minutes for vLLM
- **HuggingFace native** — no GGUF conversion, day-one model support
- **GPU + NPU simultaneously** — NPU handles always-on agents while GPU runs big models

## Integration Path

MLX Engine already follows the server pattern (OpenAI-compatible API on a port). Integration into Lemonade would follow the same WrappedServer approach as the vLLM backend in PR #1537:

1. New backend entry in `src/` wrapping lemon-mlx-engine binary
2. Model router selects MLX for supported architectures (gfx1151, gfx1150, gfx110X, gfx120X)
3. Auto-download of pre-built binaries per GPU target

## Hardware

```
AMD Ryzen AI MAX+ PRO 395
Radeon 8060S (gfx1151)
128GB unified memory
CachyOS (Arch), kernel 7.0.0
ROCm 7.12.0 (TheROCk portable)
```

## References

- Benchmark repo: https://github.com/stampby/bleeding-edge
- Raw JSON results: https://github.com/stampby/bleeding-edge/tree/main/benchmarks
- MLX Engine: https://github.com/lemonade-sdk/lemon-mlx-engine
- vLLM ROCm results also posted to PR #1537
- Reproducible with `bench.sh`: https://github.com/stampby/bleeding-edge/blob/main/bench.sh

fl0rianr:
<html><head></head><body><h2>Summary</h2><p>This PR adds <code>lemon-mlx</code> as an experimental text-generation backend for Lemonade, with supported runtime paths for:</p><ul><li><p>Apple Silicon through Metal</p></li><li><p>Linux on AMD Strix Halo (<code>gfx1151</code>) through ROCm</p></li></ul><p>The integration supports model installation, loading, non-streaming and streaming chat completions, reasoning-output normalization, stop handling, telemetry, cancellation, and backend error propagation.</p><p>The CPU implementation and release artifact are intentionally retained for development and future optimization. Manual testing confirmed that the existing CPU path works, but it is currently not advertised as supported because its inference performance is significantly behind the equivalent <code>llama.cpp</code> path.</p><p>Closes #1642.</p><h2>What changed</h2><h3>Backend integration</h3><ul><li><p>Register <code>lemon-mlx</code> through the backend descriptor and registry infrastructure.</p></li><li><p>Add selectable Metal and ROCm runtime variants.</p></li><li><p>Add installation and binary resolution for:</p><ul><li><p>macOS ARM64 Metal</p></li><li><p>Linux x86-64 ROCm</p></li><li><p>retained Linux/macOS CPU development artifacts</p></li></ul></li><li><p>Resolve and configure the appropriate bundled or system ROCm runtime.</p></li><li><p>Add <code>lemon-mlx</code> configuration defaults and backend version pins.</p></li><li><p>Integrate MLX models into the model registry, CLI, Model Manager, documentation, and generated backend references.</p></li></ul><h3>Included models</h3><ul><li><p><code>Qwen3.5-0.8B-MLX</code></p></li><li><p><code>Qwen3.6-27B-MLX</code></p></li><li><p><code>Qwen3.6-35B-A3B-MLX</code></p></li></ul><h3>Request handling</h3><ul><li><p>Keep the loaded MLX model reference consistent when forwarding requests.</p></li><li><p>Normalize <code>max_completion_tokens</code> to the backend-supported <code>max_tokens</code> field.</p></li><li><p>Preserve user-provided generation settings.</p></li><li><p>Apply narrowly scoped repetition protection for the smallest Qwen variant.</p></li><li><p>Support both chat-completion and completion request transformations.</p></li></ul><h3>Streaming and reasoning</h3><p>The MLX streaming output is normalized to Lemonade's OpenAI-compatible response format:</p><ul><li><p>reasoning is emitted through <code>reasoning_content</code></p></li><li><p>final assistant output remains in <code>content</code></p></li><li><p><code>&lt;think&gt;</code> and <code>&lt;/think&gt;</code> tags are not exposed to clients</p></li><li><p>split reasoning/content events do not duplicate usage or timing metadata</p></li><li><p>backend stop markers and request-provided stop sequences are handled locally</p></li><li><p>a clean <code>[DONE]</code> event is emitted when the backend completes without providing one</p></li></ul><p>The implementation remains local to the MLX backend instead of changing shared streaming behavior for unrelated backends.</p><h3>Streaming transport security</h3><p>After the latest <code>main</code> merge, Lemonade's HTTP client defaults to an external HTTPS-only policy.</p><p>The MLX backend communicates with its managed subprocess through a loopback HTTP URL, so its streaming request now explicitly uses:</p><pre><code class="language-cpp">utils::HttpSecurityPolicy::TrustedLoopback
</code></pre><p>Without this policy, the internal <code>http://127.0.0.1:...</code> stream was rejected by libcurl with:</p><pre><code class="language-text">CURL error: Unsupported protocol
</code></pre><h3>Cancellation and error handling</h3><ul><li><p>Treat client disconnects and stop requests as cancellation rather than backend failure.</p></li><li><p>Avoid reporting transport errors after a locally completed stream.</p></li><li><p>Capture non-200 backend response bodies instead of forwarding them as malformed SSE data.</p></li><li><p>Extract useful messages from JSON backend errors where available.</p></li><li><p>Propagate streaming failures as OpenAI-compatible SSE error objects:</p></li></ul><pre><code class="language-json">{
  "error": {
    "message": "...",
    "type": "backend_error"
  }
}
</code></pre><ul><li><p>Treat an empty or interrupted stream without a valid completion marker as an explicit backend error instead of silently returning an empty <code>200 OK</code> response.</p></li></ul><h3>Telemetry</h3><p>The MLX path records and reports:</p><ul><li><p>prompt/input tokens</p></li><li><p>completion/output tokens</p></li><li><p>time to first token</p></li><li><p>decoding tokens per second</p></li></ul><p>Backend-provided usage and timing values are preferred. Narrow estimates are used when the backend omits individual fields.</p><h2>Supported runtime policy</h2><h3>Publicly supported</h3>
Backend | Platform | Hardware
-- | -- | --
metal | macOS ARM64 | Apple Silicon
rocm | Linux x86-64 | AMD Strix Halo gfx1151

<h3>CPU implementation</h3><p>The CPU server path has been manually tested successfully and remains in the codebase.</p><p>The following are deliberately retained:</p><ul><li><p>CPU binary installation logic</p></li><li><p>CPU version pin</p></li><li><p><code>cpu_bin</code> defaults</p></li><li><p>CPU runtime setup</p></li><li><p>CPU device handling</p></li><li><p>the <code>cpu</code> binary variant</p></li></ul><p>CPU is not currently included in the public support descriptor, automatic backend selection, capability test matrix, or generated support documentation because it was measured at roughly one tenth of the performance of the comparable <code>llama.cpp</code> CPU path.</p><p>This keeps the implementation available for future optimization without presenting an uncompetitive runtime as a recommended backend.</p><h2>Testing</h2><h3>Manual validation</h3><p>Validated with MLX Qwen models:</p><ul><li><p>model installation and startup</p></li><li><p>non-streaming chat completion</p></li><li><p>streaming chat completion</p></li><li><p>reasoning and final-answer separation</p></li><li><p>client stop/cancellation</p></li><li><p>restart and subsequent requests</p></li><li><p>input/output token telemetry</p></li><li><p>TTFT and TPS telemetry</p></li><li><p>existing CPU runtime path</p></li></ul><h3>Automated integration coverage</h3><p>The PR adds ROCm runtime coverage through:</p><pre><code class="language-bash">python test/server_llm.py \
  --wrapped-server lemon-mlx \
  --backend rocm
</code></pre><p>This covers the standard Lemonade LLM integration path, including non-streaming and streaming chat completions.</p><p>Static integration coverage validates:</p><pre><code class="language-bash">pytest -q test/test_lemon_mlx_integration.py
</code></pre><p>The static contract checks include:</p><ul><li><p>backend registration</p></li><li><p>supported/public variants</p></li><li><p>retained CPU binary variant</p></li><li><p>version and default configuration</p></li><li><p>model registry entries</p></li><li><p>request and reasoning normalization</p></li><li><p>streaming error propagation</p></li><li><p>loopback HTTP security policy</p></li><li><p>generated capability declarations</p></li></ul><h2>Acknowledgements</h2><p>This integration builds on the earlier work by @bong-water-water-bong. Thank you for establishing and continuing to improve the initial MLX integration.</p></body></html>


src/cpp/server/backends/mlx/mlx_server.cpp
```diff
@@ -0,0 +1,1694 @@
...
+        LOG(INFO, "Telemetry") << "TPS:           " << std::fixed << std::setprecision(2)
+                                << telemetry.tokens_per_second << std::endl;
+        LOG(INFO, "Telemetry") << "=================" << std::endl;
+
+        sink.done();
+        LOG(INFO, "Server") << "Streaming completed - 200 OK" << std::endl;
+
+    } else {
+        sink.done();
+    }
```
bong-water-water-bong, line 1746: **Surface backend errors to the streaming client.** On `stream_error` this branch only logged server-side and ended the stream, so the UI always shows the generic *"No content received from stream"* — even though `LLMChatPanel.tsx` in this PR already parses `{"error": {...}}` SSE events. Emitting the captured error lets users see the real cause (e.g. `[hip_kernel] No ROCm back-end`), which matches the PR's stated goal of surfacing streaming errors explicitly.

Needs a small captured message string — 3 supporting one-liners are in the review summary.

```suggestion
    } else {
        // Surface the backend failure instead of ending the stream silently.
        // LLMChatPanel already handles {"error": {...}} SSE events.
        if (!client_aborted) {
            const json err = {{"error",
                {{"message", stream_error_message.empty()
                                 ? std::string("lemon-mlx streaming failed")
                                 : stream_error_message},
                 {"type", "backend_error"}}}};
            const std::string ev = sse_data_event(err);
            sink.write(ev.c_str(), ev.size());
        }
        sink.done();
    }
```

bong-water-water-bong (commented):
Did a full read of `mlx_server.cpp` (1,694 lines) — overall solid and close to merge-ready (idempotent `send_final_done`, null-safe `getenv`→`join_paths`, good telemetry fallbacks, the reasoning-stream normalizer + tiny-Qwen repetition guard are nicely done). One actionable gap + two minor notes.

### 🔴 Surface streaming backend errors to the client
See the inline suggestion on the final `else` branch. It needs a captured message string — 3 small supporting edits:

1. Declare next to `bool stream_error = false;` (~line 1405):
   ```cpp
   std::string stream_error_message;
   ```
2. At the non-200 site (~line 1604):
   ```cpp
   stream_error_message = "backend returned HTTP " + std::to_string(result.status_code);
   ```
3. At the exception site (~line 1613):
   ```cpp
   stream_error_message = e.what();
   ```
Without this, the backend error is only logged and the UI falls back to the generic "No content received from stream" — even though `LLMChatPanel.tsx` here is already wired to display `parsed.error.message`.

### 🟡 Minor — duplicate `usage`/`timings` in split chunks
`transform_chat_stream_chunk` does `json split_chunk = chunk;`, which deep-copies `usage`/`timings` into every split piece (and `passthrough` keeps them too). Low impact since the final usage chunk usually has no `delta.content`, but `split_chunk.erase("usage"); split_chunk.erase("timings");` after the copy avoids emitting the same usage across multiple SSE events.

### 🟡 Minor — ctx clamp only matches Qwen3.5
`clamp_qwen35_rocm_ctx_size` keys off `is_qwen35_model` (`qwen3.5`/`qwen3_5`), but the registry also ships `Qwen3.6-27B-MLX` and `Qwen3.6-35B-A3B-MLX`. If those hit the same ROCm large-ctx startup failure they won't be clamped to 4096 — worth confirming 3.6 is unaffected (or widening the match).

### Note on CPU
The wrapper does the right things for CPU (`MLX_DISABLE_COMPILE=1`, `LD_LIBRARY_PATH`); the `[hip_kernel] No ROCm back-end` blocker looks engine-side in `lemon-mlx-engine`. Since the descriptor/docs/`backend_versions.json` already advertise `cpu`, might be worth gating/hiding it until the engine fix lands so it doesn't ship as a broken option.

bong-water-water-bong:
**CPU fix submitted as PR #2740**

The `[hip_kernel] No ROCm back-end` CPU failure is caused by inherited `LD_LIBRARY_PATH` leaking system ROCm paths into the CPU binary.

Fix: restrict LD_LIBRARY_PATH to the executable directory, set ROCM_PATH/HIP_PATH to /dev/null, and HIP_VISIBLE_DEVICES to empty for CPU mode.

Tests pass. Can you review?

Geramy (approved):

src/cpp/server/backends/mlx/mlx_server.cpp
Geramy: @fl0rianr can you check this?

antmikinka:
### Heads-up: parallel tools validation draft (not meant as a takeover)

@fl0rianr @bong-water-water-bong — I spent time on my **own full-stack implementation** of lemon-mlx + OpenAI tools so I could be confident the path works end-to-end and is reviewable, without blocking on or force-pushing anyone else’s branch.

**What I opened (all drafts)**

- Lemonade: https://github.com/lemonade-sdk/lemonade/pull/2751 — backend-shaped stack **plus** tools stream hygiene, capability catalog, test harness policy  
- Engine: https://github.com/lemonade-sdk/lemon-mlx-engine/pull/62 — actual `tools` / `tool_calls` parse & emit (and tools×thinking policy)  
- Engine: https://github.com/lemonade-sdk/lemon-mlx-engine/pull/61 — gfx1150 fatbin (orthogonal to tools)

**Intent**

- **Not** trying to replace or race your PR as the sole backend. #2013 is still the obvious place for the lemon-mlx integration that closes [#1642](https://github.com/lemonade-sdk/lemonade/issues/1642).
- I needed a **private-enough sandbox** to get tools right (engine first, then lemonade stream preserve/dedupe, then test honesty). That’s why it’s a separate draft rather than commits pushed onto `fl0rianr/add_mlx_lemon_backend` without asking.
- Ideal collab from my side: keep **#2013 as the backend vehicle**, and **rebase / stack** the tools-only lemonade delta onto your branch (or open a follow-up PR targeting it) after you/maintainers prefer that. Co-author / credit as you see fit.

**Note on tools caps**

Your tip currently keeps `tool_calls` / `tool_calls_streaming` **False** for the small test model (sensible before engine tools exist). Once [engine#62](https://github.com/lemonade-sdk/lemon-mlx-engine/pull/62) (or a release with tools) is available, the flip + stream hygiene from #2751 is the small delta worth stacking.

If you’d rather I leave backend entirely to #2013 and only maintain a thin tools PR against your head, say the word—I’m good either way as long as we avoid two full competing backends.

antmikinka:
### Tools delta pushed onto this branch (minimal, for review)

Following your OK to put the small tools changes on **#2013** for easy review (and keep the big backend point yours), I **fast-forwarded** four tools-only commits onto `fl0rianr/add_mlx_lemon_backend` (**no force-push**). Your prior tip remains the parent; history is strictly additive.

**Commits (newest last)**

1. `fix(mlx): stream hygiene for tool_calls` — preserve/dedupe engine `tool_calls` on the MLX SSE path; keep `finish_reason=tool_calls`; no free-text tool invent  
2. `feat(mlx): enable tool_calls capabilities (P4)` — caps True; honest registry labels (only **Qwen3.5-4B-MLX** labeled tool-calling; 0.8B/27B/35B stripped of dishonest tool-calling)  
3. `feat(mlx): tools test policy, harness no-think, L-fwd, argv log` — test harness can set `lemon-mlx.args` (defaults `--no-think` **for lemon-mlx tests only**); product `defaults.json` still `args: ""`; 012/013 harden; forward `enable_thinking=true`; INFO argv log  
4. `test(mlx): drop obsolete HIP_VISIBLE assert on 2013 stack` — aligns static contract with this branch’s CPU path  

**Mirror branch (same tip):** `tools/stack-on-pr2013`  
**Dependency:** structured tools need a **tools-capable** lemon-mlx-engine binary ([engine#62](https://github.com/lemonade-sdk/lemon-mlx-engine/pull/62) / local `rocm_bin`). Stock **b1049** pin alone does not emit OpenAI `tool_calls`.  

**CI note:** flipping caps to True may exercise 012/013 against stock pin and fail until a tools-enabled engine is available — happy to adjust (keep caps False in CI until pin, or document skip) if you prefer the “big point” green first.

Feel free to edit/drop any of these commits; happy to iterate.

fl0rianr:
For for clarification regarding the reverting force push clarification here, we discussed the tool path to be a follow up on this already significant change.

fl0rianr:
Updated PR description to also make clear CPU support is disabled for now due to lacking performance.

Geramy:
I am closing this out, I will be deprecating this experiment, it was fun and nice but it got too messy to fast. I have developed a new engine which solves two unique problems, multi-gpu inference and multi-node inference, it also uses HRX and LOOM which is the new ROCm for AMD and the new HIPC for AMD. This new engine also works with my amdgpu driver on Mac and works on linux and probably will work on windows. I will also be able to add NPU support to it fairly easily and this entire engine I built also is AMD GPU Architected, meaning it was built completely around AMD GPUs. https://github.com/Geramy/LSE/releases

Thanks for the help on getting this experiment where it got!

PR #2015, opened 2026-05-26
Drop support for `flm_args`

superm1:
Arguments are passed without validation into `flm` which can cause unintended arguments to override behavior in `flm`.

There are two options to solve this:
1. Drop all flm argument support
2. Whitelist individual arguments that should be passed down.

This aims for the former.

### DIFF
--- a/docs/dev/getting-started.md
+++ b/docs/dev/getting-started.md
@@ -657,7 +657,6 @@ Accepts a JSON object with one or more keys to update atomically. Returns `{"sta
 | `cfg_scale` | number |
 | `width` | int (positive) |
 | `height` | int (positive) |
-| `flm_args` | string |
 
 **Example:**
 ```bash
--- a/docs/embeddable/runtime.md
+++ b/docs/embeddable/runtime.md
@@ -167,7 +167,6 @@ Accepts a JSON object with one or more keys to update atomically. Returns `{"sta
 | `cfg_scale` | number |
 | `width` | int (positive) |
 | `height` | int (positive) |
-| `flm_args` | string |
 
 **Example:**
 === "Windows (cmd.exe)"
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -327,7 +327,6 @@ The following options are available depending on the recipe being used:
 | Option | Description | Default |
 |--------|-------------|---------|
 | `--ctx-size SIZE` | Context size for the model | `4096` |
-| `--flm-args ARGS` | Custom arguments to pass to flm serve (e.g., `"--socket 20 --q-len 15"`) | `""` |
 
 #### RyzenAI LLM (`ryzenai-llm` recipe)
 
@@ -659,7 +658,6 @@ lemonade bench [options] MODEL_NAME [MODEL_NAME ...]
 | `--no-memory` | Disable VRAM/RAM tracking | Tracking enabled |
 | `--no-reload` | Skip model reload between scenarios (faster, but prompt cache may skew results) | Model reloaded |
 | `--llamacpp-args ARGS` | Custom args for llama-server (e.g. `"-b 2048 -ub 1024"`). Repeat for multiple arg sets. | — |
-| `--flm-args ARGS` | Custom args for flm serve. Repeat for multiple. | — |
 | `--vllm-args ARGS` | Custom args for vllm-server. Repeat for multiple. | — |
 
 **Preflight behavior:**
--- a/docs/guide/configuration/custom-models.md
+++ b/docs/guide/configuration/custom-models.md
@@ -435,7 +435,7 @@ When loading a model, settings are resolved in this order (highest to lowest pri
 2. Per-model values from `recipe_options.json`
 3. Global configuration values, see [Server Configuration](./README.md)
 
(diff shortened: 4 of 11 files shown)


fl0rianr:
Given the broader backend-args discussion in #2018, I think this is probably the right security area to tighten.

One thing I’m trying to understand: should `flm_args` be treated differently from the other `*_args` surfaces?

For llama.cpp / whisper.cpp / sd.cpp, we’re discussing whether to preserve expert args with stronger validation / expanded deny-listing because a strict allow-list can become a breaking maintenance burden. This PR takes the stricter route for FLM and removes the surface entirely.

That may be the right call if FLM does not have a meaningful set of safe tuning args, or if its argument semantics make validation too brittle. If so, maybe worth stating that explicitly in the PR description so the security rationale is clear and this doesn’t read as an accidental feature drop.

Otherwise, an alternative consistent with the #2018 direction could be to keep flm_args but block Lemonade-managed / dangerous flags like host, port, mode/model selection, path/file-loading flags, and add regression coverage for override attempts.

superm1:
Well the difference here is that flm doesn't have meaningful tuning args to pass in the first place though.

So our deny list would pretty much cover them all wouldn't it?

So that's why I lean tear it all out.

fl0rianr (approved):
That makes sense, thanks for clarifying. I’m good with removing flm_args for now.

Given that FLM does not currently have many meaningful safe tuning args beyond the options Lemonade already manages, removing the surface seems preferable to maintaining a near-complete deny-list.

I would just prefer that we don’t treat this as a permanent policy decision against FLM-specific tuning. If FLM later has a cleaner integration story, especially on Linux where it can be managed more like the other Lemonade backends, I think it should be fine to reintroduce explicit safe options or a narrower expert surface as needed.

**Longer term**, I’d like FLM not to remain a special case in the backend model. But that’s not a reason to block this PR now.

bitgamma:
from flm help:

```
  --prefill-chunk-len arg (=-1)    Set prefill chunk length
  -r [ --img-pre-resize ] arg (=2) Pre-resize the image, 0: original size, 1: 
                                   height = 480, 2: height = 720, 3: height = 
                                   1080, 4: height = 1440
  -s [ --socket ] arg (=10)        Set the maximum number of socket connections allowed (for serve command)
  -q [ --q-len ] arg (=10)         Set number of max npu queue length (for serve command)
```

are none of these really relevant? especially the first two look like something that could be tuned

fl0rianr:
These are good points. As I said this is not like a final target situation, a quick security fix. And truly an arg as prefill-chunk-len is great, maybe some of it can be hard-coded in the meanwhile.

superm1:
I wasn't aware of some of those; some of them might make sense to leave as tunable.
But I'm starting to feel like the real issue is that we allow tuning and loading new models from the same API interface you use to interact with an LLM.

It's one of those don't 💩 where you eat type of things.  I'm feeling like we should be having separate or at least segregated configuration.  We should be defaulting to a randomly generated API key, not defaulting to open.  There should be a different API key for accessing the LLM from configuring lemond.

So yes; I feel like this is step 1 for security to tear out argument support; but it's part of a bigger story that lemond and lemonade need to be locked down by default.

bitgamma:
using a random API key generated during installation will generate friction, but I also think it is the way to go at least for the admin endpoints. No idea how this can be done without the user just thinking the software is broken though.

>  There should be a different API key for accessing the LLM from configuring lemond.

There is already `LEMONADE_ADMIN_API_KEY` for this, but if it is not set, it defaults to the "normal" API key (or empty).

Another layer of protection is that `lemonade config set` only works if the request comes on the local interface, not remotely (not even LAN)

>  I feel like this is step 1 for security to tear out argument support

Tearing out argument support is throwing the baby out with the bath water imho.

superm1:
> using a random API key generated during installation will generate friction, but I also think it is the way to go at least for the admin endpoints. No idea how this can be done without the user just thinking the software is broken though.

Maybe this needs some sort of of onboarding flow.

> Another layer of protection is that lemonade config set only works if the request comes on the local interface, not remotely (not even LAN)

There is also this: https://github.com/lemonade-sdk/lemonade/pull/2110

> Tearing out argument support is throwing the baby out with the bath water imho.

When you start to chain vulnerabilities due to design decisions of flexibility you sometimes need to.

Jays-1111:
Hi @superm1 @fl0rianr @bitgamma 
I understand the security concern around generic `flm_args`.

I just want to provide some additional context on why this capability was useful for us. We integrated Lemonade + FastFlowLM into an agent-style product similar to MineContext, where the agent continuously sends screenshot / VLM analysis requests to the local FLM server. In this type of workload, when the request volume exceeds FLM server capacity, we observed failures or repeated retries.

We discussed this with the FastFlowLM team here:
https://github.com/FastFlowLM/FastFlowLM/issues/328

Their suggestion was to tune FLM server parameters such as `-s / --socket` and `-q / --q-len` to increase the server connection / queue capacity.

Given that raw `flm_args` has been removed, it would be helpful to have a safer replacement in the future, such as a small allowlisted set of FLM runtime options. This would preserve the security model while still supporting continuous VLM / agent workloads.

fl0rianr:
@Jays-1111 you are right to call this out this is not a proper long-term solution (having those args remove without an substitute). Could you file an issue please?

Jays-1111:
@fl0rianr Yes, agree this should be tracked as a follow-up issue. My comment was mostly to provide a concrete example of why these args can matter in practice.
Would you mind filing the issue since you probably have a better view of the right long-term direction here

superm1:
We should also in that issue discuss with the FastFlow team if there is a downside to just better defaults.  If there is no downside, setting that higher in fastflow code could work too.

PR #2018, opened 2026-05-26
security: replace deny-list with allow-list for backend arguments

superm1:
HTTP-reachable config keys (llamacpp.args, whispercpp.args, sdcpp.args) were validated using only a deny-list of reserved flags. This allowed attackers to inject dangerous flags like --path (serve arbitrary directory), --log-file (write to arbitrary path), -lua-init, -grammar-file, etc. leading to arbitrary file read/write and RCE as the lemonade process UID (root on macOS daemon).

Replace deny-list validation with strict allow-list of safe tuning flags:
- llamacpp.args: allows only --threads, --ctx-size, --batch-size, --n-gpu-layers, tuning parameters (temp, top-k, top-p, etc.), and sampling options
- whispercpp.args: allows only --threads and device selection
- sdcpp.args: allows only --threads and inference tuning

### DIFF
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -389,17 +389,44 @@ void LlamaCppServer::load(const std::string& model_name,
     LOG(DEBUG, "LlamaCpp") << "ngl set to " << gpu_layers << std::endl;
     push_arg(args, reserved_flags, "-ngl", gpu_layers, std::vector<std::string>{"--gpu-layers", "--n-gpu-layers"});
 
-    // Validate and append custom arguments
+    // Validate and append custom arguments using allow-list only
     if (!llamacpp_args.empty()) {
-        std::string validation_error = validate_custom_args(llamacpp_args, reserved_flags);
-        if (!validation_error.empty()) {
-            throw std::invalid_argument(
-                "Invalid custom llama-server arguments:\n" + validation_error
-            );
-        }
+        static const std::set<std::string> kAllowedFlags = {
+            "--threads", "-t",
+            "--ctx-size", "-c",
+            "--batch-size", "-b",
+            "--ubatch-size", "-ub",
+            "--n-gpu-layers", "-ngl",
+            "--flash-attn",
+            "--temp",
+            "--top-k",
+            "--top-p",
+            "--min-p",
+            "--repeat-penalty",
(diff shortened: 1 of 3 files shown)


bitgamma:
while this makes sense purely from a security standpoint, this is a breaking change that will require constant maintenance as flags are constantly added to llamacpp and your allow-list doesn't include a lot of the needed ones (for example draft) while includes -ngl which is actually reserved.

Let's just expand the deny-list to include the potentially dangerous one. Hopefully new potentially dangerous flags are a rare(r) addition. Also, we could add a heuristic to detect paths and block those as well.

Default behavior doesn't allow changing the configuration from non-loopback address and additionally we have API KEY and ADMIN API KEY

fl0rianr:
Good points. I agree that a raw backend-flag allow-list is likely to become a maintenance burden and may break legitimate expert use cases as llama.cpp / whisper.cpp / sd.cpp evolve.

One thing I would still like us to keep explicit is that the risk is broader than only currently known dangerous flags. A deny-list can miss new flags, aliases, or value forms, and a path heuristic may be bypass-prone unless it is very carefully defined. So whichever direction we choose, I think we should make the intended trust boundary clear:

- If these args are reachable from HTTP/admin config, they probably need stricter validation than "not in deny-list".
- Lemonade-managed flags should remain non-overridable (--port, model paths, host/listen options, etc.).
- We should add regression tests for the known dangerous examples from this PR (--path, --log-file, -lua-init, -grammar-file) plus a few edge cases around aliases and --flag=value.

A compromise could be: expand the deny-list for the immediate compatibility-preserving fix, but also add tests and document this as an expert/local-only escape hatch rather than a generally safe remote configuration surface. If we keep the allow-list approach instead, I think it should also validate flag arity/value shape and avoid allowing reserved flags like -ngl.

jeremyfowers (changes requested):
Surely this is not a comprehensive list of llamacpp args? I think we will cause a lot of headaches for ourselves this way when people try to customize and we block them.

Even a comprehensive static list of all llamacpp args will still cause problems because some people will use newer llamacpp builds, and those can introduce new valid args. Consider the situation where everyone was partying with MTP before Lemonade officially supported it?

If we want something like this we may need to dynamically run `llama-server -h` to populate the whitelist. wdyt?

jeremyfowers:
Closing PR with unanswered "request changes" for over 60 days

PR #2019, opened 2026-05-26
security: validate URL scheme in openExternal handler

superm1:
Only allow http:// and https:// schemes in window.api.openExternal(). This prevents attackers from injecting dangerous schemes like file://, \\unc, smb://, search-ms:, ms-msdt:, etc. that Windows ShellExecute would execute.

Attack surface includes: prompt-injected LLM output in markdown links, XSS, compromised npm dependencies, or malicious content from lemonade-server.ai iframe. Validation is applied at the Tauri layer (URL parsing + scheme check) to catch injection before it reaches the platform opener plugin.

### DIFF
--- a/src/app/src/renderer/tauriShim.ts
+++ b/src/app/src/renderer/tauriShim.ts
@@ -113,7 +113,14 @@ async function installTauriApi(): Promise<void> {
       await writeText(String(text));
     },
     openExternal: (url: string) => {
-      openUrl(url).catch((e) => console.warn('openExternal', e));
+      try {
+        const u = new URL(url);
+        if (u.protocol === 'http:' || u.protocol === 'https:') {
+          openUrl(u.href).catch((e) => console.warn('openExternal', e));
+        }
+      } catch {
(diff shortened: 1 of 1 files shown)


fl0rianr:
Good catch. Since this is now a security boundary, What do you think about passing the canonical parsed URL to the opener instead of the original raw string:

```
const u = new URL(url);
if (u.protocol === 'http:' || u.protocol === 'https:') {
  openUrl(u.href).catch((e) => console.warn('openExternal', e));
}
```

That avoids parser/sink discrepancies where the JS URL parser validates one representation but the platform opener receives the original unnormalized input.

kenvandine (approved):
Great catch!

fl0rianr (commented):
@superm1 Could you quickly consider my security point above before merging? Thanks!

superm1:
> @superm1 Could you quickly consider my security point above before merging? Thanks!

Thanks, it's a good point!  I've adjusted accordingly.

PR #2020, opened 2026-05-26
security: fix path traversal in /web-app/ route

superm1:
Prevents directory traversal attacks by:
1. Adding /web-app/ to authenticated routes (requires API key when configured)
2. Canonicalizing filesystem paths to resolve .. segments
3. Validating resolved path remains within web_app_dir

Before: GET /web-app/..%2f..%2fetc%2fpasswd disclosed arbitrary files
After: Path canonicalization and bounds checking prevent escape


fl0rianr:
Good work, this looks like the right direction to me. The canonicalization + bounds check is a good replacement for the previous string concatenation.

One thing that would be useful before merging: a small regression test for the traversal case from the PR description, plus one normal asset request, so we lock down both the deny and allow paths.

Minor note: this adds auth for `/web-app/...`, while root-level assets / the SPA fallback still behave as before. If that’s intentional, maybe worth tightening the PR description slightly.

Issue #1925, opened 2026-05-17
bug: Kokoro koko b17 (10.5) - ESPEAK_DATA_PATH missing path separator causes all non-US voices to produce garbled 0.4s audio


wirehead82:
### Platform

Linux/Ubuntu

### Lemonade Version

10.5.0

### GPU / APU Model

rx7600, ryzen 5 7600.

### Component

Kokoro TTS

## Description

After upgrading to Lemonade Server 10.5 (which installs koko b17), all non-US Kokoro voices (e.g. `bm_lewis`, `bf_emma`, and other British/non-American voices) produce a 0.4-second garbled audio blip instead of full speech. American voices (e.g. `af_sarah`) continue to work correctly.

## Environment

- **OS:** Ubuntu 26.04 (native install, `lemond` systemd service)
- **Lemonade version:** 10.5.0
- **koko version:** b17
- **Hardware:** AMD Ryzen 5 7600 / Radeon RX 7600

## Root Cause

When `lemond` launches the `koko` subprocess, it sets `ESPEAK_DATA_PATH` by concatenating `LD_LIBRARY_PATH` and `"espeak-ng-data"` **without a path separator**. The result is a malformed path:
ESPEAK_DATA_PATH=/var/lib/lemonade/.cache/lemonade/bin/kokoro/cpuespeak-ng-data

The actual espeak-ng-data directory is at:
/var/lib/lemonade/.cache/lemonade/bin/kokoro/cpu/espeak-ng-data

Because `ESPEAK_DATA_PATH` points to a non-existent directory, koko cannot load any language data. American English (`en-US`) appears to have a fallback path or is embedded differently, but British English and other locales fail completely, producing the short garbled output.

This can be verified by inspecting the running koko process:
```bash
cat /proc/$(pgrep -x koko)/environ | tr '\0' '\n' | grep ESPEAK_DATA_PATH
# Output: ESPEAK_DATA_PATH=/var/lib/lemonade/.cache/lemonade/bin/kokoro/cpuespeak-ng-data


Workaround
Create a symlink at the broken path pointing to the real directory:

bash
sudo ln -s \
  /var/lib/lemonade/.cache/lemonade/bin/kokoro/cpu/espeak-ng-data \
  /var/lib/lemonade/.cache/lemonade/bin/kokoro/cpuespeak-ng-data
sudo pkill -x koko
sudo systemctl restart lemond
[...]

PR #2025, opened 2026-05-26
fix(kokoro): correct ESPEAK_DATA_PATH construction

anditherobot:
## Summary

Addresses the malformed `ESPEAK_DATA_PATH` setup reported in #1925.

## Changes

- Construct `ESPEAK_DATA_PATH` with `fs::path` joining instead of string concatenation.
- Prevent paths such as `.../cpuespeak-ng-data`; the expected path is `.../cpu/espeak-ng-data`.

## Testing

- `cmake --build --preset windows --target lemond`
- `python test/server_tts.py` (6 tests passed against the compiled `lemond`)

## Note

This corrects the malformed root data path. The language-alias packaging behavior discussed for British/French voices is separate and is not addressed here.


bitgamma (approved):
thank you for your contribution!

PR #2027, opened 2026-05-26
Add model folders to Settings UI

boclifton-MSFT:
## Summary

Adds a new **Settings → Server** panel in the desktop app for editing two server-wide config keys that previously required the CLI:

- `models_dir` — where the server downloads HF models (default: `auto`).
- `extra_models_dir` — recursive search path for loose `.gguf` files (default: empty / disabled).

Each field has a text input, a native folder picker (**Browse…**), **Apply**, and **Reset** (which restores the field's own default — `auto` for `models_dir`, empty for `extra_models_dir`). Status messages distinguish:

- **401** — admin API key missing/invalid (fixable by setting `LEMONADE_ADMIN_API_KEY`).
- **403** — `/internal/*` is loopback-only; the change must be made on the server machine (the admin key cannot unlock this from a remote client).
- Dialog/plugin failures, surfaced through the shared error row instead of becoming unhandled promise rejections.

image

## Architecture notes

- These keys are **server-wide config**, not per-client preferences, so they are *not* added to `AppSettings` / `app_settings.json` (per AGENTS.md invariant #11). The renderer reads/writes them directly via `GET /internal/config` and `POST /internal/set`, mirroring the existing direct-fetch pattern used for `/health` and `/system-stats`.
- New helper module `src/app/src/renderer/utils/serverRuntimeConfig.ts` encapsulates the HTTP calls and classifies HTTP errors into `unauthorized` (401), `forbidden_remote` (403), and generic `error` variants.
- `ServerSettings.tsx` factors a small `<DirectoryField>` sub-component shared by both rows. The panel stays visible (read-only) in unauthorized / forbidden_remote states so users can still see the configured values and an explanation of why editing is disabled.
- After a successful write, the panel re-reads the config *before* transitioning to "Saved", so a failed refresh never leaves stale values labelled as saved.
- No server-side C++ changes — `handle_config_set` in `src/cpp/server/server.cpp` already handles both keys and invalidates the model cache, so changes take effect immediately.

## Dependencies added

| Layer | Package | Version | Reason |
|---|---|---|---|
| npm (runtime) | `@tauri-apps/plugin-dialog` | `~2.2.2` | Native folder picker for the **Browse…** button |
| Rust (Tauri) | `tauri-plugin-dialog` | `~2.2` | Host-side plugin registration |
| Capability | `dialog:allow-open` | — | Tauri permission allowlist entry |

These versions are deliberately pinned to the last plugin-dialog releases compatible with `@tauri-apps/api ^2.0.0` and `tauri ^2` respectively (plugin-dialog 2.3.0 raised both floors). This keeps `tauri`, `@tauri-apps/api`, `@tauri-apps/cli`, `tao`, `wry`, `tauri-runtime`, `muda`, and `tray-icon` at main's versions; the `Cargo.lock` diff is purely additive (0 removals). A cosmetic Tauri Rust/JS version-mismatch warning is the trade-off; a follow-up dependency PR can bump them all together.

## Docs

- Added a consistent one-line tip ("in the Lemonade desktop app, `models_dir` and `extra_models_dir` can also be set under **Settings → Server**") to `docs/embeddable/models.md`, `docs/guide/configuration/README.md`, and `docs/guide/faq.md`.
- Corrected the description of how imported models appear in the model list in `docs/embeddable/models.md` (plus the in-app description and the helper module's doc comment): imported models are shown under their **bare filename**; the `extra.` prefix is only added when the bare name would conflict with an already-registered model. The previous wording implied the prefix was always applied.

## Verification

- `npm run build:renderer:prod` passes (no new warnings).
- `cargo check` against the Tauri host passes.
- Manually verified: picking a folder via Browse persists to the server, Reset returns to `auto` / empty, and a loose `.gguf` dropped into the extra folder appears under its bare filename in Model Manager.
- Unauthorized path (401) verified by hitting `/internal/*` without admin auth — UI displays the admin-key guidance.

## AI NOTE
AI was used to assist in creating this update, but the code was manually reviewed and verified before submission.

### DIFF
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -41,7 +41,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
     "vulkan_args": "",
     "rocm_args": "",
     "cpu_args": "",
-    "device": "",
+	"device": "",
     "prefer_system": false,
     "rocm_bin": "builtin",
     "vulkan_bin": "builtin",
@@ -70,7 +70,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
     "vulkan_bin": "builtin"
   },
   "flm": {
-    "args": ""
+    "args": "",
   },
   "ryzenai": {
     "server_bin": "builtin"
@@ -83,22 +83,22 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
 
 ### Settings Reference
 
-| Key                       | Type   | Default     | Description                                                                                                    |
-| ------------------------- | ------ | ----------- | -------------------------------------------------------------------------------------------------------------- |
-| `port`                    | int    | 13305       | Port number for the HTTP server                                                                                |
-| `host`                    | string | "localhost" | Address to bind for connections                                                                                |
-| `log_level`               | string | "info"      | Logging level (trace, debug, info, warning, error, fatal, none)                                                |
-| `global_timeout`          | int    | 300         | Timeout in seconds for HTTP, inference, and readiness checks                                                   |
-| `max_loaded_models`       | int    | 1           | Max models per type slot. Use -1 for unlimited                                                                 |
-| `no_broadcast`            | bool   | false       | Disable UDP broadcasting for server discovery                                                                  |
-| `extra_models_dir`        | string | ""          | Secondary directory to scan for GGUF model files                                                               |
-| `models_dir`              | string | "auto"      | Directory for cached model files. "auto" follows HF_HUB_CACHE / HF_HOME / platform default                     |
-| `ctx_size`                | int    | 4096        | Default context size for LLM models                                                                            |
-| `offline`                 | bool   | false       | Skip model downloads                                                                                           |
-| `no_fetch_executables`    | bool   | false       | Prevent downloading backend executable artifacts; backends must already be installed or use the system backend |
-| `disable_model_filtering` | bool   | false       | Show all models regardless of hardware capabilities                                                            |
-| `enable_dgpu_gtt`         | bool   | false       | Include GTT for hardware-based model filtering                                                                 |
-| `rocm_channel`            | string | "stable"    | ROCm backend channel: "stable" (default) or "nightly". See [llama.cpp Backend](./llamacpp.md) for details      |
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| `port` | int | 13305 | Port number for the HTTP server |
+| `host` | string | "localhost" | Address to bind for connections |
+| `log_level` | string | "info" | Logging level (trace, debug, info, warning, error, fatal, none) |
+| `global_timeout` | int | 300 | Timeout in seconds for HTTP, inference, and readiness checks |
+| `max_loaded_models` | int | 1 | Max models per type slot. Use -1 for unlimited |
+| `no_broadcast` | bool | false | Disable UDP broadcasting for server discovery |
+| `extra_models_dir` | string | "" | Secondary directory to scan for GGUF model files |
+| `models_dir` | string | "auto" | Directory for cached model files. "auto" follows HF_HUB_CACHE / HF_HOME / platform default |
+| `ctx_size` | int | 4096 | Default context size for LLM models |
+| `offline` | bool | false | Skip model downloads |
+| `no_fetch_executables` | bool | false | Prevent downloading backend executable artifacts; backends must already be installed or use the system backend |
+| `disable_model_filtering` | bool | false | Show all models regardless of hardware capabilities |
+| `enable_dgpu_gtt` | bool | false | Include GTT for hardware-based model filtering |
+| `rocm_channel` | string | "stable" | ROCm backend channel: "stable" (default) or "nightly". See [llama.cpp Backend](./llamacpp.md) for details |
 
 > Tip: in the Lemonade desktop app, `models_dir` and `extra_models_dir` can also be set under **Settings &rarr; Server**.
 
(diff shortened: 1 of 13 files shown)


docs/guide/configuration/README.md
```diff
@@ -41,7 +41,7 @@ If you are using a standalone `lemond` exectable, the default location is `~/.ca
     "vulkan_args": "",
     "rocm_args": "",
     "cpu_args": "",
-	"device": "",
+    "device": "",
     "prefer_system": false,
```
fl0rianr, line 45: please try avoiding unrelated unrelated churn, thanks. 

Also don't cripple the lists.

src/app/src/renderer/utils/serverRuntimeConfig.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+function authHeaders(): Record<string, string> {
+  const key = serverConfig.getAPIKey();
+  return key ? { Authorization: `Bearer ${key}` } : {};
+}
+
+function classifyHttpError(status: number, body: string): RuntimeConfigStatus & RuntimeConfigUpdateResult {
+  if (status === 401 || status === 403) {
+    // /internal/* may require LEMONADE_ADMIN_API_KEY rather than the regular
```
fl0rianr: This classification seems too broad. `/internal/*` can fail with 403 not because the admin key is missing, but because the request is not from loopback; the server explicitly restricts internal endpoints to localhost. Treating both 401 and 403 as unauthorized makes the UI tell remote users to set LEMONADE_ADMIN_API_KEY, even though that cannot make a remote `/internal/*` request succeed.

Could we distinguish:
- 401: missing/invalid admin API key
- 403: internal endpoint is localhost-only / remote server not supported

Alternatively, disable or hide this Server panel when `serverConfig.isRemoteServer()` is true.

src/app/src/renderer/tabs/ServerSettings.tsx
```diff
@@ -0,0 +1,253 @@
...
+      {status.kind === 'saving' && (
+        <div className="settings-description">Saving…</div>
+      )}
+      {status.kind === 'saved' && (
+        <div className="settings-description">Saved. Server is using the new value.</div>
+      )}
+      {status.kind === 'unauthorized' && (
+        <div className="settings-description">
+          The server rejected the request. Editing server configuration requires the
+          admin API key (<code>LEMONADE_ADMIN_API_KEY</code>). Set it on the server and
```
fl0rianr: The guidance here is incomplete for remote servers. A 403 from `/internal/*`  can mean “internal endpoints are only accessible from localhost”, not an admin-key problem. In that case, setting 
LEMONADE_ADMIN_API_KEY in Connection → API Key will not fix the request.

Could we surface a separate message for remote/localhost-only failures, or disable this panel when the configured server base URL is remote?

src/app/package-lock.json
```diff
@@ -7,11 +7,12 @@
     "": {
       "name": "lemonade-app",
       "version": "1.0.0",
-      "license": "MIT",
+      "license": "Apache-2.0",
       "dependencies": {
-        "@tauri-apps/api": "^2.0.0",
+        "@tauri-apps/api": "^2.11.0",
```
fl0rianr: Do we need to bump @tauri-apps/api and @tauri-apps/cli for this feature, or is adding @tauri-apps/plugin-dialog sufficient?

A Tauri version bump expands the scope of this PR and may introduce unrelated platform/build changes. If the bump is only to silence a warning, I’d prefer either documenting why it is required or moving it to a separate dependency-maintenance PR.

Basically the same kind of question for src/app/src-tauri/Cargo.toml as well.

src/app/src/renderer/tabs/ServerSettings.tsx
```diff
@@ -0,0 +1,253 @@
...
+  // it from `serverValue` whenever the parent reloads, so the input always
+  // reflects the server after a save.
+  const [draft, setDraft] = useState(serverValue);
+  useEffect(() => {
+    setDraft(serverValue);
+  }, [serverValue]);
+
+  const handleBrowse = async () => {
+    // Seed the picker with the current path when it's a real directory; for
+    // the default sentinel (or empty), let the OS pick the start location.
```
fl0rianr, line 72: Could we wrap the dialog call in `try/catch` and surface failures via the existing error status path? Right now a dialog/plugin failure would become an unhandled promise rejection because the click handler intentionally discards the promise.

src/app/src/renderer/tabs/ServerSettings.tsx
```diff
@@ -0,0 +1,253 @@
...
+    }
+  }, []);
+
+  useEffect(() => {
+    void loadConfig();
+  }, [loadConfig]);
+
+  const finishApply = async (result: RuntimeConfigUpdateResult) => {
+    if (result.kind === 'ok') {
+      setStatus({ kind: 'saved' });
```
fl0rianr: After a successful update, this sets saved before re-reading the config, but if the follow-up `getRuntimeConfig()` fails, the UI still shows “Saved” and may display stale values. Could we handle non-ok refresh results explicitly?

fl0rianr (changes requested):
Thanks for adding this — exposing models_dir / extra_models_dir in Settings is useful, and keeping these out of AppSettings is the right direction. I like the UI approach, a bit more near to the current design (e.g. less text) would be nice as well.

But before merge I would like to see the point below being taken care of. Thanks!

src/app/src/renderer/tabs/ServerSettings.tsx
boclifton-MSFT: Fixed in `f82829f8`. `handleBrowse` now wraps `open()` in `try/catch` and surfaces failures via a new `onError` prop, which the parent routes into the shared error status row. User-cancel (null return) is still treated as a no-op, not an error.

src/app/src/renderer/utils/serverRuntimeConfig.ts
boclifton-MSFT: Good catch. Split in `f5f1e1be`: 401 → `unauthorized` (admin key missing/invalid), 403 → new `forbidden_remote` variant (server restricts `/internal/*` to loopback). The two cases are now handled independently in both the helper and the panel.

src/app/src/renderer/tabs/ServerSettings.tsx
boclifton-MSFT: Falls out of #2. The panel now renders distinct copy for `forbidden_remote` explaining that `/internal/*` is loopback-only and that `LEMONADE_ADMIN_API_KEY` won't help — the user has to make the change on the server machine. Kept the panel visible (read-only) rather than hiding it, so the user can still see the configured values and the reason. `f5f1e1be`.

src/app/package-lock.json
boclifton-MSFT: Overaggressive version bump, sorry. Reverted in `bdf819c3`. Pinned `@tauri-apps/plugin-dialog` to `~2.2.2` and the Rust crate to `~2.2` (last releases compatible with `@tauri-apps/api ^2.0.0` and `tauri ^2` respectively; 2.3.0 raised both floors). `tauri`, `api`, `cli`, `tao`, `wry`, `tauri-runtime`, `muda`, `tray-icon` all stay at main's versions; `Cargo.lock` diff is now purely additive (0 removals). Trade-off: we ship plugin-dialog 2.2.2 instead of 2.7.1, and the cosmetic version-mismatch warning returns. Happy to bump in a follow-up dependency PR if you'd prefer.

docs/guide/configuration/README.md
boclifton-MSFT: Agreed. Reset all three doc files (`faq.md`, `configuration/README.md`, `embeddable/models.md`) to main and re-applied only the intended changes: the three "Settings → Server" tips and the corrected imported-models naming note in `models.md`. Fixed in `4bb50f6b`.

src/app/src/renderer/tabs/ServerSettings.tsx
boclifton-MSFT: Fixed in `f82829f8`. `finishApply` now re-reads the config *before* transitioning to `saved`. If the refresh fails (unauthorized, forbidden_remote, or network error), it sets an explicit `"Saved, but could not re-read the configuration: …. Displayed values may be stale."` error instead.

dochoss:
Also, to your point @fl0rianr of reducing text in the components, I added a hoverable help icon and put the expanded text in there to help with clarity.  Two model folders can be confusing, needed to have some more supporting text, and I think this solves that without cluttering up the settings panel.  It's CSS + JS with SVG so no dependencies involved. Let me know if you prefer a different approach.

Issue #1930, opened 2026-05-17
Possible RPM bug: no service auto-start?


jeremyfowers:
Investigate the situation the reddit user described here: https://discordapp.com/channels/@me/1505611655579697302/1505624987950907604

splifingate:
Post-install (10.5.0-x86-64.rpm; Fedora 44 Plasma), I had to manually run "sudo systemctl daemon-reload && sudo systemctl restart lemond" to get the service up.

jeremyfowers:
This is expected behavior on Fedora. We added specific instructions to the install guide for how to start the service.

Issue #1947, opened 2026-05-18
Fedora RPM (10.5.0): `lemond` systemd service is not enabled after install, and Fedora docs still reference the old `lemonade-server` service name


cdamiens:
### Platform

Linux/Fedora

### Lemonade Version

10.5.0 (`lemonade-server-10.5.0-1.x86_64`)

### Component

Other / Not sure

### Problem

After a fresh install, the server is not running and `lemonade status` fails with `Could not connect to Lemonade server`. The docs say the server should start automatically, but it doesn't.

Two issues combined:

1. The [Fedora install page](https://lemonade-server.ai/docs/guide/install/fedora/) and [configuration page](https://lemonade-server.ai/docs/guide/configuration/) still mention the service as `lemonade-server`, but it was renamed to `lemond` in 10.3.
2. The unit file is shipped at `/opt/lib/systemd/system/lemond.service`, which is not in systemd's default search path, and the RPM's `%post` does not enable it. So `systemctl status lemond` returns "not found" until the user runs `systemctl enable` manually.

### Workaround

```
sudo systemctl enable --now lemond
```

### Suggested fixes

- Ship the unit at `/usr/lib/systemd/system/` instead of `/opt/lib/systemd/system/`.
- Enable the service in the RPM `%post` scriptlet (or via preset).
- Update the Fedora docs to use `lemond`.

### Steps to Reproduce

1. Download rpm file for Fedora `lemonade-server-10.5.0.x86_64.rpm`
2. sudo dnf install lemonade-server-10.5.0.x86_64.rpm
3. lemonade status => Error fetching health status: Could not connect to Lemonade server (Could not establish connection). Make sure the server is running and try again.
4. systemctl status lemonade-server => Unit lemonade-server.service could not be found.

### Expected vs Actual Behavior

Expected Behavior:
After `dnf install`, the `lemond` service should be enabled and running, matching the documentation. `lemonade status` should succeed without any extra steps.

Actual Behavior:
The service is installed but not enabled, not running, and not even discoverable by `systemctl` without prior knowledge of its non-standard install path. `lemonade status` fails.

### Log Output

```shell

```

jeremyfowers:
@abn @lucifer-vali please have a look.

jeremyfowers:
Duplicate of https://github.com/lemonade-sdk/lemonade/issues/1930 ?

jeremyfowers:
This is expected behavior on Fedora. We added specific instructions to the install guide for how to start the service.

PR #2029, opened 2026-05-27
add instructions to start/enable lemond on Fedora based distros

lucifer-vali:
## Changes
- added instructions to enable and start lemond.service on the website docs

I'll add another PR that fully addresses #1947 later

### DIFF
--- a/docs/guide/install/fedora.md
+++ b/docs/guide/install/fedora.md
@@ -2,7 +2,34 @@
 
 Lemonade is built for Fedora 43 and 44.
 
-1. Get the RPM for your version from the [latest release](https://github.com/lemonade-sdk/lemonade/releases):
-   - Fedora 43: `lemonade-server-<version>-fc43.x86_64.rpm`
-   - Fedora 44: `lemonade-server-<version>-fc44.x86_64.rpm`
-2. `sudo dnf install ./DOWNLOADED_FILE`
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Thanks for updating the Fedora instructions — this is a useful improvement.

One small thing before merge: could you change the PR description from `Fixes #1930 and #1947 (partially)` to something like `Refs #1930` / `Partially addresses #1947`? Since #1947 also tracks package-level fixes, using `Fixes` might auto-close it if that's not the intention, even though this PR only covers the docs side.

Optional nit: `systemctl --no-pager status lemond` may be slightly better for copy/paste docs, but not a blocker.

lucifer-vali:
> One small thing before merge: could you change the PR description from `Fixes #1930 and #1947 (partially)` to something like `Refs #1930` / `Partially addresses #1947`? Since #1947 also tracks package-level fixes, using `Fixes` might auto-close it if that's not the intention, even though this PR only covers the docs side.

Sorry, I'll be careful from next time. Does the updated description look better?

> Optional nit: `systemctl --no-pager status lemond` may be slightly better for copy/paste docs, but not a blocker.

Added.

fl0rianr (approved):
I would assume you tested it as well, so from my side good to go now. Thanks!

PR #2030, opened 2026-05-27
fix: bump macOS sd.cpp metal version and macOS release number

Geramy:
Bump the macOS sd.cpp metal backend to the new upstream release: `master-616-cde20d5` → `master-655-29ab511`, and update the macOS release number from 15.7.4 to 15.7.7.


fl0rianr (approved):
If it helps, you have my go.

PR #2031, opened 2026-05-28
Fix downloaded status for multi-checkpoint models

### LINKED ISSUE #1958: `pull` skips auxiliary checkpoints (text_encoder/vae) when main file is already cached
## Platform
Linux/Ubuntu 26.04 (x86_64)

## Lemonade Version
10.5.1

## GPU / APU
AMD Radeon 8060S Graphics (gfx1151, Strix Halo)

## Component
stable-diffusion.cpp (and any other model whose recipe declares multiple `checkpoints`)

## Bug Description

`lemonade pull <model>` (and `lemonade pull <collection>` via fan-out) does **not** download auxiliary checkpoints declared under a model's `checkpoints` map (e.g. `text_encoder`, `vae`) when the **main** checkpoint file already exists in the HF cache. The pull command silently reports success even though the model is unusable until `load` is invoked — `load` then downloads the missing files.

This means a user who already has the main GGUF for, say, `Flux-2-Klein-9B-GGUF` cached (for any reason — partial prior download, manual fetch, etc.) and then runs `lemonade pull "Ultra Collection"` ends up with a "successfully pulled" collection that is missing ~8.5 GiB of required auxiliary files. The first attempt to actually use the model triggers a large unexpected download.

## Steps to Reproduce

Confirmed on `main` at 8163b274 with `build/lemonade` v10.5.1 against a running `lemond`.

1. Have only the main file of a multi-checkpoint model cached. Easiest setup: ensure `~/.cache/huggingface/hub/models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/<sha>/flux-2-klein-9b-Q8_0.gguf` exists, but `Qwen3-8B-Q8_0.gguf` and `split_files/vae/flux2-vae.safetensors` do NOT exist anywhere in the HF cache.
2. Run `build/lemonade pull "Ultra Collection"` (Ultra Collection includes `Flux-2-Klein-9B-GGUF`, whose `checkpoints` map declares `main` + `text_encoder` + `vae`).
3. Observe: command exits 0, prints `Model pulled successfully: Ultra Collection`. No download progress for the Flux text_encoder or vae.
4. Verify on disk: only `flux-2-klein-9b-Q8_0.gguf` is present in the FLUX repo's snapshot dir; neither `Qwen3-8B-Q8_0.gguf` nor `split_files/vae/flux2-vae.safetensors` was downloaded.
5. Now `POST /api/v1/load` with `Flux-2-Klein-9B-GGUF` — this triggers an ~8.5 GiB download of the missing aux files into the **main** repo's snapshot dir (see "Co-location quirk" below) before the model becomes usable.

A simple post-pull check that demonstrates the gap:
```
find ~/.cache/huggingface/hub/models--unsloth--FLUX.2-klein-9B-GGUF/snapshots -type f
```
After pull → 1 file. After load → 3 files (main + text_encoder + vae, ~17.9 GiB).

## Expected vs Actual

- **Expected:** `pull` downloads every entry in `checkpoints` (except `npu_cache`, which is intentionally excluded per the existing logic in `download_from_huggingface`). The "downloaded" status of a model is only true once all required checkpoint files are present.
- **Actual:** `pull` short-circuits when the main file exists, so aux checkpoints are never fetched. The model is reported as fully downloaded.

## Root Cause (likely)

Two interacting issues in `src/cpp/server/model_manager.cpp`:

### 1. `downloaded` status only checks the main file

In the cache-build pass (around `model_manager.cpp:1434`), for non-collection, non-flm models:
```cpp
bool file_exists = !info.resolved_path().empty() && safe_exists(info.resolved_path());
...
info.downloaded = !has_manifest && !has_partial;
```
`info.resolved_path()` with no argument resolves the **main** checkpoint only. The `checkpoints` map is never iterated to verify aux files (text_encoder/vae/mmproj/etc.) exist. Compare with `resolve_all_model_paths` (`model_manager.cpp:1183`), which does iterate all entries — the existence check should do the same.

### 2. Collection pull fan-out trusts that flag and short-circuits

In `download_model` (`model_manager.cpp:2389-2402`):
```cpp
for (const auto& component : info.components) {
    ...
    auto comp_info = get_model_info(component);
    if (comp_info.downloaded) {
        LOG(INFO, "ModelManager") << "Component already downloaded: " << component << std::endl;
        continue;
    }
    ...
    download_model(component, comp_data, do_not_upgrade, forward, visited);
}
```
Because (1) sets `downloaded=true` as soon as the main file exists, the collection skips the component, and the per-component `download_from_huggingface` (which **does** know how to enqueue aux checkpoints — see `model_manager.cpp:2881-2894`) is never called.

The single-model `pull` path hits the same problem one layer up: at `model_manager.cpp:2446`,
```cpp
if (do_not_upgrade && is_model_downloaded(model_name)) { ... return; }
```
short-circuits on the same incomplete `downloaded` signal. (For collection fan-out the bypass happens at line 2395 before this.)

### Suggested fix direction

Make `info.downloaded` reflect completeness of the entire `checkpoints` map — iterate all entries (skip `npu_cache`, as `download_from_huggingface` does at line 2887) and require each to exist and be free of `.partial` / `.download_manifest.json` markers. Then the existing fan-out and short-circuit logic will do the right thing automatically.

## Co-location quirk worth noting

When `load` (not `pull`) fetches the aux files, it puts them inside the **main** model's snapshot dir keyed by the file path from the JSON, e.g.:
```
models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/<sha>/Qwen3-8B-Q8_0.gguf
models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/<sha>/split_files/vae/flux2-vae.safetensors
```
…not into `models--unsloth--Qwen3-8B-GGUF/` or `models--Comfy-Org--vae-text-encorder-for-flux-klein-9b/` (the repos declared in `server_models.json`). This is consistent with `download_from_huggingface` writing the main snapshot dir first and then placing aux files relative to it, but it's surprising if you grep the HF cache by source repo name. The fix for the existence check needs to look in the right place — either match the on-disk co-location convention, or make `resolve_model_path("<type>")` authoritative and reuse it for the existence probe.

## Log Output

`pull` output — note no per-file download progress for Flux aux checkpoints:
```
$ build/lemonade pull "Ultra Collection"
Pulling model: Ultra Collection

Model pulled successfully: Ultra Collection
```

Subsequent `load` shows the aux files being downloaded on demand (excerpt):
```
2026-05-19 13:22:27.659 [Debug] (ProcessManager) Starting process with inherited output:
  /home/.../sd-server --listen-port 8002
    --diffusion-model .../models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/.../flux-2-klein-9b-Q8_0.gguf
    --llm           .../models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/.../Qwen3-8B-Q8_0.gguf
    --vae           .../models--unsloth--FLUX.2-klein-9B-GGUF/snapshots/.../split_files/vae/flux2-vae.safetensors
...
[DEBUG] model.cpp:925  - loading tensors from .../Qwen3-8B-Q8_0.gguf
[DEBUG] model.cpp:925  - loading tensors from .../split_files/vae/flux2-vae.safetensors
```

## Additional Context

The same logic should be audited for **every** model that uses the multi-`checkpoints` schema, e.g. `Flux-2-Klein-4B`, `Qwen-Image-GGUF`, `Qwen-Image-2512-GGUF`, `Z-Image-Turbo`, all `Whisper-*` (though those use `npu_cache` which is correctly excluded), and any model that uses `mmproj` (`mmproj` is converted into a synthetic `checkpoints` entry at `model_manager.cpp:1272`, so it has the same exposure).

anditherobot:
## Summary

Fixes #1958.

This PR fixes how Lemonade decides whether a model is downloaded.

Before this change, a model could be marked as downloaded when only the main checkpoint existed. That was wrong for models that also need extra files like `vae` or `mmproj`. In that case, `pull` could skip the model even though some required files were still missing.

## What changed

The changes unify how Lemonade calculates a model's downloaded status:
- **Verified completeness:** Every required checkpoint is now checked (skipping lazy `npu_cache`).
- **Unified logic:** Reused the same check in both cache initialization and single-model updates.
- **Preserved markers:** Kept the existing detection for `.partial` and `.download_manifest.json` files.

Incomplete download markers are checked the same way as before: file checkpoints check their parent directory for `.download_manifest.json` and their own `.partial` marker; directory checkpoints check the directory itself.

## Tests

Added coverage for:

- a multi-checkpoint model with the main checkpoint present but an auxiliary checkpoint missing;
- collection pull fan-out, making sure an incomplete component is reached instead of skipped;
- single-checkpoint models, to make sure existing behavior still works.

## Scope

This PR only changes downloaded-status detection. It does not change the downloader, model deletion, cache cleanup, GGUF shard handling, or path resolution.

<details>
<summary>Review context</summary>

The original issue was that downloaded status only checked the main model file. That made multi-checkpoint models look complete even when auxiliary files were missing.

During review, the scope was narrowed to avoid changing the downloader or cache cleanup paths. The final approach keeps the fix focused on downloaded-status calculation:

- extract the existing completeness check into a small helper;
- apply it to every required checkpoint;
- skip only lazy `npu_cache`;
- use that result in `build_cache()` and `add_model_to_cache()`.

The marker-check wording was also updated. The implementation does not claim to check the Hugging Face snapshot root directly. It keeps the existing local behavior: file checkpoints check their parent directory and specific `.partial` marker; directory checkpoints check the directory itself.

</details>

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1250,7 +1250,7 @@ std::string ModelManager::resolve_model_path(const ModelInfo& info, const std::s
                 return filepath;
             }
         }
-        
+
         // Case 5: Local quant-token fallback.
         //
         // Keep the existing resolver cases above as the primary logic: exact
@@ -1500,6 +1500,50 @@ static bool check_component_downloaded(const ModelInfo& info,
     return true;
 }
 
+static bool has_partial_files(const fs::path& dir) {
+    std::error_code ec;
+    if (!safe_is_directory(dir)) return false;
+    // Non-recursive scan for .partial markers to confirm folder integrity
+    for (const auto& entry : fs::directory_iterator(dir, ec)) {
+        if (entry.path().extension() == ".partial") {
+            return true;
+        }
+    }
+    return false;
+}
+
+static bool is_checkpoint_path_complete(const std::string& path_str) {
+    if (path_str.empty()) return false;
+
+    fs::path resolved(path_str);
+    if (!safe_exists(resolved)) return false;
+
+    // A manifest or .partial file indicates an interrupted multi-file download
+    fs::path marker_dir = safe_is_directory(resolved) ? resolved : resolved.parent_path();
+    if (safe_exists(marker_dir / ".download_manifest.json")) return false;
+
+    if (safe_is_directory(resolved)) {
+        if (has_partial_files(marker_dir)) return false;
+    } else if (safe_exists(path_from_utf8(path_str + ".partial"))) {
+        return false;
+    }
+
+    return true;
+}
+
+/**
+ * Returns true if all files required by the model recipe are present and complete.
+ * Note: npu_cache is skipped as it is managed lazily by the flm-npu backend.
+ */
+static bool are_required_checkpoints_complete(const ModelInfo& info) {
+    for (const auto& [type, _] : info.checkpoints) {
+        if (type == "npu_cache") continue;
+
+        if (!is_checkpoint_path_complete(info.resolved_path(type))) return false;
+    }
+    return true;
+}
+
 void ModelManager::build_cache() {
     std::lock_guard<std::mutex> lock(models_cache_mutex_);
 
@@ -1640,43 +1684,7 @@ void ModelManager::build_cache() {
         } else if (info.recipe == "flm") {
             info.downloaded = flm_set.count(info.checkpoint()) > 0;
         } else {
-            // Check if model file/dir exists
-            bool file_exists = !info.resolved_path().empty() && safe_exists(info.resolved_path());
-
-            if (file_exists) {
-                // Also check for incomplete downloads:
-                // 1. Check for .download_manifest.json in snapshot directory
-                // 2. Check for any .partial files
-                fs::path resolved(info.resolved_path());
-
-                // For directories (OGA models), check within the directory
-                // For files (GGUF models), check in parent directory
-                fs::path snapshot_dir = safe_is_directory(resolved) ? resolved : resolved.parent_path();
-
-                // Check for manifest (indicates incomplete multi-file download)
-                fs::path manifest_path = snapshot_dir / ".download_manifest.json";
-                bool has_manifest = safe_exists(manifest_path);
-
-                // Check for .partial files
-                bool has_partial = false;
-                if (safe_is_directory(resolved)) {
-                    // For directories, scan for any .partial files inside
-                    std::error_code ec;
-                    for (const auto& entry : fs::directory_iterator(snapshot_dir, ec)) {
-                        if (entry.path().extension() == ".partial") {
-                            has_partial = true;
-                            break;
-                        }
-                    }
-                } else {
-                    // For files, check if the specific file has a .partial version
-                    has_partial = safe_exists(info.resolved_path() + ".partial");
-                }
-
-                info.downloaded = !has_manifest && !has_partial;
-            } else {
-                info.downloaded = false;
-            }
+            info.downloaded = are_required_checkpoints_complete(info);
         }
 
         if (info.downloaded) {
@@ -1779,33 +1787,7 @@ void ModelManager::add_model_to_cache(const std::string& model_name) {
         auto flm_models = get_flm_installed_models();
         info.downloaded = std::find(flm_models.begin(), flm_models.end(), info.checkpoint()) != flm_models.end();
     } else {
-        bool file_exists = !info.resolved_path().empty() && safe_exists(info.resolved_path());
-
-        if (file_exists) {
-            // Check for incomplete downloads
-            fs::path resolved(info.resolved_path());
-            fs::path snapshot_dir = safe_is_directory(resolved) ? resolved : resolved.parent_path();
-
-            fs::path manifest_path = snapshot_dir / ".download_manifest.json";
-            bool has_manifest = safe_exists(manifest_path);
-
-            bool has_partial = false;
-            if (safe_is_directory(resolved)) {
-                std::error_code ec;
-                for (const auto& entry : fs::directory_iterator(snapshot_dir, ec)) {
-                    if (entry.path().extension() == ".partial") {
-                        has_partial = true;
-                        break;
-                    }
-                }
-            } else {
-                has_partial = safe_exists(info.resolved_path() + ".partial");
-            }
-
-            info.downloaded = !has_manifest && !has_partial;
-        } else {
-            info.downloaded = false;
-        }
+        info.downloaded = are_required_checkpoints_complete(info);
     }
 
     populate_static_max_context_window(info);
--- a/test/test_multi_checkpoint_completeness.py
+++ b/test/test_multi_checkpoint_completeness.py
@@ -1,0 +1,273 @@
+import json
+import os
+import shutil
+import subprocess
+import tempfile
+import time
+import unittest
+import requests
+
+from utils.test_models import get_default_lemond_binary, get_default_cli_binary
+
+
+class TestMultiCheckpointCompleteness(unittest.TestCase):
+    """
+    Integration tests for multi-checkpoint download integrity.
+
+    These tests verify that models requiring multiple files (e.g., Image Gen with VAEs)
+    are only marked as 'Downloaded' when all required components are 100% present.
+
+    Operation:
+    - Creates a temporary workspace for each test case.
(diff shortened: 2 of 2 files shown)


fl0rianr (changes requested):
Thanks for working on this. I agree with the goal: a model should not be reported as downloaded if any required auxiliary checkpoint is missing.

I think we can make this less risky by narrowing the change to the exact status predicate that is currently wrong.

The downloader path itself is already checkpoint-aware once it is reached: download_from_huggingface() iterates over info.checkpoints and queues non-main / non-npu_cache files. The problematic part seems to be earlier: build_cache() and add_model_to_cache() still derive `downloaded` from `info.resolved_path()` without a checkpoint type, which effectively means “main checkpoint exists”.

A minimal fix I would prefer is:

1. Extract the existing downloaded-status logic into a small helper, e.g. is_checkpoint_path_complete(...).
2. Add a second helper, e.g. are_required_checkpoints_complete(const ModelInfo& info), which:

   * iterates over info.checkpoints;
   * skips npu_cache;
   * uses info.resolved_path(type) for each required checkpoint;
   * returns false if any required path is empty or missing;
   * preserves the existing .partial / .download_manifest.json semantics as much as possible.
3. Use that helper only in the existing `downloaded` calculation sites: `build_cache()` and `add_model_to_cache()`.
4. Leave the actual download path unchanged unless a failing test proves it also needs changes.

For validation, I would like the PR to include a regression test that fails on current `main` (since your issue is based on 10.5.1 if I saw it correctly) and passes with the minimal helper change. The most important case is a collection component where the main checkpoint exists but an auxiliary checkpoint is missing: the component must not be considered downloaded, so collection fan-out should not skip it.

I would also like one guard test for a normal single-checkpoint model to make sure its behavior does not change.

My concern is not with the direction of the fix, but with changing a central cache/download path more broadly than necessary. Thanks!

fl0rianr:
The plan seem reasonable, but ready to merge can only be accessed if the code is accordingly. There is no change, currently. Would be also nice to have the test result of the current main and main + your fix. Thanks!

fl0rianr:
Thanks, this version is much closer to what I had in mind. The source change now looks properly scoped to the downloaded-status predicate, and leaving the actual download path untouched is the right direction.

The new tests are also a clear improvement. The multi-checkpoint and single-checkpoint guard cases are useful, and the collection status test covers the condition that drives the collection skip logic.

I have two remaining small concerns before approving:

1. `test_collection_fan_out` currently verifies the "Downloaded" status via "lemonade list", but it does not actually exercise the collection `pull` fan-out path. Since the production fan-out path branches on `comp_info.downloaded`, this is a reasonable proxy, but the test comment currently says the component “will be included in a pull”, which the test does not directly prove. Could you either adjust the test/comment to make clear that this is a status-regression test, or add a direct pull/fan-out assertion if that is feasible?

2. There seems to be a leftover `find_download_marker_dir()` helper that is no longer used. Also, the PR description says marker checks happen at the HF snapshot root, while the current helper checks the resolved path or its immediate parent. Could you clean up the unused helper and align the wording with the actual behavior?

After that, plus the relevant CI being run, I’m comfortable with this approach. Good work, thanks!

fl0rianr:
Thanks, the two core files now look much better to me.

The `model_manager.cpp` change is scoped in the right way: the downloader path is untouched, and the new logic only changes how `downloaded` is computed in build_cache() and add_model_to_cache(). The helper now checks all required checkpoints and skips only npu_cache, which addresses the original main-only status bug.

The test coverage is also much stronger now. 
But there are still two remaining cleanup requests before I’m comfortable approving:

1. Could you please remove the unrelated changes from this PR? Since your latest push I see `.gitignore`, `setup.ps1`, and `setup.sh` modified. Those are out of scope for a downloaded-status bugfix, I suppose a mistake since I cannot understand why those changes are necessary. 

2. Since you have to touch in anyway, here also a small nit: In test_collection_status_with_incomplete_component, the `pull` subprocess return code is intentionally not asserted, which is probably fine because the fake local-path component may fail later in the download path. But please make that explicit in the test comment: this test is asserting that collection fan-out reaches the incomplete component, not that the full pull succeeds.

Once the PR is reduced to the two relevant files and the test comment is clarified, this approach looks good to me. Thanks!

fl0rianr (approved):
Thanks, this looks good to me now.

The PR is scoped to the downloaded-status calculation, the unrelated file changes were removed, and the tests cover the cases I was concerned about. I also verified the test locally.

Approved from my side.

PR #2034, opened 2026-05-28
Correct websockets not to listen on 0.0.0.0 by default

superm1:


fl0rianr (approved):

fl0rianr:
fyi, the test failure is a current runner issue, don't know why it started now, but does seem to affect all runs, this commit resolves it: c82c70b91c2bd5e10d99522d8189cbf995e1ee2d - you can also advise improving on that e.g. killing ports etc.

**Update:** his commit isn't used anywhere anymore as well as the CI errors are gone. To harden that for the future #2037 was openend.

PR #2037, opened 2026-05-28
ci: harden Linux cleanup for self-hosted runners

fl0rianr:
## Summary

This hardens the Linux cleanup action for self-hosted runners by making process cleanup best-effort, scoped, and less destructive.

The previous cleanup logic used broad `pkill -9 -f` patterns, including a generic Python kill and a final broad llama|lemonade fallback. That can be unsafe on shared or long-lived self-hosted runners because it may terminate unrelated jobs or persistent runner-side processes.

This PR replaces that behavior with a safer cleanup flow:

- never fail the real CI job because cleanup is best-effort
- avoid broad generic Python process kills
- protect GitHub runner processes and the current cleanup process tree
- resolve listener PIDs for known Lemonade CI ports before killing anything
- only kill processes that are scoped to the current workspace, runner temp/runtime dir, or current GitHub Actions run
- keep unscoped kills opt-in via LEMONADE_CLEANUP_ALLOW_UNSCOPED_TARGETS=1 for dedicated CI hosts
- centralize cleanup behavior so backend workflows do not need their own ad-hoc `pkill` blocks (see vLLM)

## Why this is larger than the old cleanup script

Most of the added lines are guardrails, diagnostics, and explicit ownership checks. The old implementation was short because it was intentionally aggressive; the new implementation is longer because it avoids killing unrelated processes on self-hosted runners.

This should be easy to maintain despite the larger diff because the policy is now explicit:

- **safe by default** for shared runners
- **configurable** for dedicated runners
- **backend-agnostic** through process hints and port configuration
- **centralized** in one reusable cleanup action instead of repeated workflow-specific cleanup snippets

## Notes

Cleanup remains best-effort. If a process cannot be safely associated with the current job, the action logs it and leaves it alone rather than risking termination of unrelated work.

### DIFF
--- a/.github/actions/cleanup-processes-linux/action.yml
+++ b/.github/actions/cleanup-processes-linux/action.yml
@@ -377,7 +377,7 @@ runs:
 
             # CI Linux runners are expected to provide ss. Avoid tool probing here;
             # if the runner image is missing ss, the cleanup log should show that directly.
-            ss -H -ltnp "sport = :$port" 2>/dev/null \
+            ss -H -ltnp "sport = :$port" \
                 | sed -n 's/.*pid=\([0-9][0-9]*\).*/\1/p'
         }
 
@@ -572,4 +572,4 @@ runs:
 
         # Always succeed. Cleanup problems should be warnings, not CI blockers.
         exit 0
-        
+
--- a/.github/actions/cleanup-processes-linux/action.yml
+++ b/.github/actions/cleanup-processes-linux/action.yml
@@ -6,7 +6,7 @@ runs:
     - name: Clean up Lemonade-related processes
       shell: bash
       run: |
-        # GitHub runs bash composite steps with error handling enabled.
+        # GitHub composite steps may run with strict error handling.
         # Cleanup must be best-effort and must never fail the real test job.
         set +e
 
@@ -18,8 +18,14 @@ runs:
         TEMP_ROOT="${RUNNER_TEMP:-/tmp}"
         CURRENT_PID="$$"
 
+        # Default: safest behavior for shared self-hosted runners.
+        # Set LEMONADE_CLEANUP_ALLOW_UNSCOPED_TARGETS=1 only on dedicated CI hosts
+        # if you intentionally want to kill unscoped Lemonade-like listeners.
+        ALLOW_UNSCOPED_TARGETS="${LEMONADE_CLEANUP_ALLOW_UNSCOPED_TARGETS:-0}"
+
         echo "Workspace root: $WORKSPACE_ROOT"
         echo "Runner temp:    $TEMP_ROOT"
+        echo "Allow unscoped target kills: $ALLOW_UNSCOPED_TARGETS"
 
         read_pid_cmdline() {
             local pid="$1"
@@ -39,8 +45,10 @@ runs:
                 return 0
             fi
 
+            # Keep this intentionally narrow. Do not match "actions-runner"
+            # broadly because normal job workspaces often live under that path.
             case "$cmdline" in
-                *cleanup-processes-linux*|*actions-runner*|*Runner.Listener*|*Runner.Worker*)
+                *Runner.Listener*|*Runner.Worker*)
                     return 0
                     ;;
             esac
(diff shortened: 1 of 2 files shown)


.github/actions/cleanup-processes-linux/action.yml
```diff
@@ -1,115 +1,309 @@
...
+            esac
+
+            return 1
+        }
+
+        is_target_like_process() {
+            local cmdline="$1"
+
+            case "$cmdline" in
+                *lemonade*|*lemond*|*llama*|*stable-diffusion*|*whisper*|*tts*|*server_llm.py*|*server_sd.py*|*server_tts.py*|*server_whisper.py*)
```
sawansri: Should we include `flm`, `sd-server`, etc. here and in all process pattern matching?

fl0rianr: Good catch, made an overhaul that is more suitable.

.github/actions/cleanup-processes-linux/action.yml
sawansri, line 1: This is looking better to me but let's get a review from @jeremyfowers or @superm1 since they have a better understanding of how our CI is setup.

.github/actions/cleanup-processes-linux/action.yml
```diff
@@ -1,123 +1,589 @@
...
+        cleanup_runtime_files
+        cleanup_installation_dir
 
         echo ""
         echo "========================================"
         echo "Process cleanup complete!"
         echo "========================================"
+
+        # Always succeed. Cleanup problems should be warnings, not CI blockers.
+        exit 0
```
superm1: missing newline

.github/workflows/validate_vllm.yml
```diff
@@ -307,4 +307,4 @@ jobs:
             --title "Update vllm-rocm to ${VLLM_ROCM_RELEASE}" \
             --body-file pr_body.md \
             --base main \
-            --head "$BRANCH"
+            --head "$BRANCH"
```
superm1: missing newline

.github/actions/cleanup-processes-linux/action.yml
```diff
@@ -1,123 +1,589 @@
...
+            fi
+
+            if command -v fuser >/dev/null 2>&1; then
+                fuser -n tcp "$port" 2>/dev/null | tr ' ' '\n' || true
+            fi
+
+            if command -v ss >/dev/null 2>&1; then
+                ss -H -ltnp "sport = :$port" 2>/dev/null \
+                    | sed -n 's/.*pid=\([0-9][0-9]*\).*/\1/p' || true
+            fi
```
superm1: none of these should be `|| true` we also shouldn't be looking for the commands and conditional behavior.  This is CI.  They're there or they're not.

superm1 (commented):
The real problem to me is that it's sending SIGKILL in the first place.  Nothing gets a chance to even try to clean up properly.

This really I really have to wonder if this is better done as user systemd services for CI.  Then sending a SIGTERM should be working just fine and everything _should_ clean up properly.

If there' s a reason that user services can't work, how about instead sending SIGTERM instead of SIGKILL on it's own to lemond should clean the whole tree.

.github/actions/cleanup-processes-linux/action.yml
fl0rianr: Thanks for the review! Good point, agreed. I removed the `|| true` cleanup style and the command-existence probing. Since this is CI, the script now assumes the expected tools are present and logs real failures instead of branching around them.

I also changed the cleanup path to use SIGTERM by default. SIGKILL is no longer part of the normal cleanup flow and is only available behind an explicit opt-in fallback for dedicated runners.

superm1: But what about managing lifecycle with systemd user services instead?

fl0rianr: I agree that systemd user services would be the cleaner long-term lifecycle model for these CI services.

My concern with doing that in this PR is that it changes more than just the cleanup behavior: we would need to update how the workflows start/supervise the services, validate that user units are available/configured correctly on the self-hosted runners, and make sure the backend-specific startup paths still work under that model.

This PR is intended as a smaller safety improvement for the current setup: remove broad pkill -9, use SIGTERM by default, avoid unscoped process kills, and keep cleanup from terminating unrelated runner processes.

I’m happy to work on a follow-up PR to move the service lifecycle to systemd user units. Do you see that as a blocker for this cleanup hardening PR, or would you be okay with landing this first and handling the systemd lifecycle separately?

.github/actions/cleanup-processes-linux/action.yml
```diff
@@ -1,123 +1,575 @@
...
-            echo ""
-            echo "Killing $description processes..."
+        # Safe default for shared self-hosted runners: never kill unscoped processes.
+        # Set LEMONADE_CLEANUP_ALLOW_UNSCOPED_TARGETS=1 only on dedicated CI hosts.
+        ALLOW_UNSCOPED_TARGETS="${LEMONADE_CLEANUP_ALLOW_UNSCOPED_TARGETS:-0}"
+        FORCE_KILL="${LEMONADE_CLEANUP_FORCE_KILL:-0}"
+
+        # Keep the cleanup backend-agnostic. New backends should not need changes here
+        # unless they introduce additional long-lived CI ports or opt-out of normal CI env.
+        CLEANUP_PORTS="${LEMONADE_CLEANUP_PORTS:-13305 9001}"
```
jeremyfowers, line 29: The ws is usually on port 9000

jeremyfowers, line 29: What about CI jobs that start lemond on some other port?

jeremyfowers (changes requested):
Disabling the misbehaving CI runner seems to have resolved the problem in CI for now.

I spent a lot of time getting our self-hosted CI runners stable, and the system is pretty brittle. I would prefer to not change it at all unless there is a sufficient amount of evidence that a change is required.

Marking “changes requested” to gate the PR on that evidence.

jeremyfowers:
Closing PR since we seemed to have resolved the issue. Please ping if this needs to be reopened.

PR #2038, opened 2026-05-29
docs: add LangChain integration guide for Lemonade Server

Kushal1213:
This PR adds a LangChain integration guide for Lemonade Server at docs/integrations/langchain.md.

What's included:

-Setup instructions to connect LangChain to Lemonade Server using ChatOpenAI with /api/v1 as the base URL
-Example 1: Simple chat with system and human messages
-Example 2: Full RAG pipeline using nomic-embed-text-v1-GGUF as the embedding model, with check_embedding_ctx_length=False set for OpenAI-compatible non-OpenAI providers. Requires pulling both Llama-3.2-3B-Instruct-Hybrid and nomic-embed-text-v1-GGUF before running.
-Example 3: Prompt template + chain
-Troubleshooting table

Testing:
All examples were tested end-to-end including both model pulls and the full RAG pipeline against a real PDF.
Navigation:
File placed in docs/integrations/ alongside existing guides. Added full App Integrations section to mkdocs.yml so all integration guides (including this one) are now discoverable in the sidebar.


fl0rianr (changes requested):
Thanks for adding this — a LangChain guide would be very useful. I think this needs a small round of changes before merge. The RAG example currently initializes OpenAIEmbeddings without a model, which means LangChain will use its OpenAI default rather than a Lemonade embedding model. Lemonade’s embeddings endpoint requires a model, and the docs example uses nomic-embed-text-v1-GGUF, so the guide should add an embedding model prerequisite and pass model=... explicitly, ideally with check_embedding_ctx_length=False for OpenAI-compatible non-OpenAI providers.

Also, please align the base URL with the existing app docs (/api/v1) or explicitly mention that both /api/v1 and /v1 are supported, and make sure the new page is linked from the app guides/navigation so users can discover it. Thanks!

Kushal1213:
Hi @fl0rianr, all three points have been addressed in the latest commit (3eb270c):

1 . Embeddings — OpenAIEmbeddings now passes model="nomic-embed-text-v1-GGUF" and check_embedding_ctx_length=False, with a lemonade pull nomic-embed-text-v1-GGUF prerequisite step added before the RAG example.
2. Base URL — all occurrences updated from /v1 to /api/v1 to match the existing app docs.
3. Navigation — file moved to docs/integrations/ and linked in mkdocs.yml under a new App Integrations section alongside all other guides.

Ready for re-review, thanks!

fl0rianr:
@Kushal1213 Thanks for the update. Maybe I'm missing something, but according to what I can see the changes don't reflect what you said. Only point 3 navigation is included.

Kushal1213:
Hi @fl0rianr, sorry for the confusion earlier — the fixes are now included in commit fad75b6. All three points addressed:

OpenAIEmbeddings now passes model="nomic-embed-text-v1-GGUF" and check_embedding_ctx_length=False, with a prerequisite step to pull the embedding model.
All base URLs updated from /v1 to /api/v1.
File moved to docs/integrations/ and linked in mkdocs.yml under App Integrations.

Ready for re-review, thanks!

fl0rianr:
Thanks, this addresses the main blockers from my previous review. The guide now uses /api/v1, configures a dedicated Lemonade embedding model for the RAG example, and sets check_embedding_ctx_length=False for OpenAIEmbeddings.

A few non-blocking follow-ups before merge:

Please update the PR description so it matches the current guide (/api/v1 and the additional nomic-embed-text-v1-GGUF embedding model).
Please confirm the full command sequence was tested end-to-end, including both model pulls and running the RAG example against a real PDF.
Consider softening “Full LangChain ecosystem — everything works” to avoid overpromising beyond the examples covered here.

From my side the original blockers are resolved.



@jeremyfowers could you please provide a second opinion here? Since the **navigation** change is useful but maybe a quite impactful one I’d appreciate your take before we finalize the review.

Kushal1213:
Hi @fl0rianr, @jeremyfowers ,all follow-ups have been addressed:

 1. PR description updated to reflect `/api/v1` base URL, `nomic-embed-text-v1-GGUF` embedding model, and `check_embedding_ctx_length=False`.
 2. Yes, the full command sequence was tested end-to-end — both model pulls (`Llama-3.2-3B-Instruct-Hybrid` and `nomic-embed-text-v1-GGUF`) and the RAG example were run successfully against a real PDF.
 3. Softened "Full LangChain ecosystem — everything works" to "chains, prompt templates, and RAG pipelines work out of the box" in the latest commit.

 Ready for final review, thanks!

fl0rianr (approved):
Alright, thanks, you have my go.

But merging is on hold until @jeremyfowers' verdict.

mkdocs.yml
```diff
@@ -121,4 +134,4 @@ markdown_extensions:
   - admonition
   - pymdownx.superfences  # Better code blocks
   - pymdownx.tabbed:      # Tabbed code blocks
-      alternate_style: true
+      alternate_style: true
```
superm1: can you please fix the missing newline character here?

PR #2039, opened 2026-05-29
Vulkan back-end support for sdcpp

ckuethe:
@sofiageo made me do it.  tested on gfx1152, I'm able to generate and upscale images using vulkan.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -195,6 +195,7 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
     {"sd-cpp", "vulkan", {"windows", "linux"}, {
         {"cpu", {"x86_64"}},
         {"amd_gpu", {}},
+        {"nvidia_gpu", {}},
     }},
 
     // stable-diffusion.cpp - CPU backend (Windows/Linux x86_64)
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -191,6 +191,12 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         }},
     }},
 
+    // stable-diffusion.cpp - Vulkan backend (Windows/Linux x86_64)
+    {"sd-cpp", "vulkan", {"windows", "linux"}, {
+        {"cpu", {"x86_64"}},
+        {"amd_gpu", {}},
+    }},
+
     // stable-diffusion.cpp - CPU backend (Windows/Linux x86_64)
     {"sd-cpp", "cpu", {"windows", "linux"}, {
         {"cpu", {"x86_64"}},
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -825,8 +825,8 @@ jobs:
           - name: stable-diffusion
             script: server_sd.py
             extra_args: ""
-            backends: "cpu rocm"
-            runner: [Windows, rocm]
+            backends: "cpu rocm vulkan"
+            runner: [Windows, rocm, vulkan]
           - name: text-to-speech
             script: server_tts.py
             extra_args: ""
@@ -923,8 +923,8 @@ jobs:
           - name: stable-diffusion
             script: server_sd.py
             extra_args: ""
-            backends: "cpu rocm"
-            runner: [Linux, rocm]
+            backends: "cpu rocm vulkan"
+            runner: [Linux, vulkan, rocm]
           - name: whisper
             script: server_whisper.py
             extra_args: "--wrapped-server whispercpp"
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -953,12 +953,6 @@ jobs:
 
       - name: Set HF_HOME environment variable
         run: echo "HF_HOME=$PWD/hf-cache" >> $GITHUB_ENV
-      
-      - name: Install Vulkan runtime
-        if: runner.os == 'Linux'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libvulkan1 vulkan-tools
 
       - name: Install FLM backend for FLM wrapped-server tests
         if: ${{ contains(matrix.extra_args, '--wrapped-server flm') }}
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -953,6 +953,12 @@ jobs:
 
       - name: Set HF_HOME environment variable
         run: echo "HF_HOME=$PWD/hf-cache" >> $GITHUB_ENV
+      
+      - name: Install Vulkan runtime
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libvulkan1 vulkan-tools
 
       - name: Install FLM backend for FLM wrapped-server tests
         if: ${{ contains(matrix.extra_args, '--wrapped-server flm') }}
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -18,6 +18,7 @@
   "sd-cpp": {
     "cpu": "master-616-cde20d5",
     "rocm-stable": "master-616-cde20d5",
+    "vulkan": "master-616-cde20d5",
     "metal": "master-655-29ab511"
   },
   "therock": {
--- a/src/cpp/server/backends/sd_server.cpp
+++ b/src/cpp/server/backends/sd_server.cpp
@@ -31,10 +31,6 @@ bool is_rocm_backend(const std::string& backend) {
     return backend == "rocm" || backend == "rocm-stable";
 }
 
-bool is_gfx115x_arch(const std::string& arch) {
-    return arch == "gfx1150" || arch == "gfx1151" || arch == "gfx1152";
-}
-
 std::string resolve_sdcpp_backend(const std::string& backend) {
     if (backend == "rocm") {
         std::string channel = "stable";
@@ -224,16 +220,12 @@ void SDServer::load(const std::string& model_name,
     }
 
     if (resolved_backend == "vulkan") {
-        std::string rocm_arch = SystemInfo::get_rocm_arch();
-        if (is_gfx115x_arch(rocm_arch)) {
-            LOG(INFO, "SDServer")
-                << "Applying Vulkan gfx115X VAE workaround: --vae-tiling"
-                << std::endl;
-            args.push_back("--vae-tiling");
-        }
+        LOG(INFO, "SDServer")
+            << "Applying Vulkan SD workaround: --vae-tiling --diffusion-fa"
+            << std::endl;
+        args.push_back("--vae-tiling");
         args.push_back("--diffusion-fa");
     }
-
     std::set<std::string> reserved_flags = {
         "-m",
         "--model",
--- a/src/cpp/server/backends/sd_server.cpp
+++ b/src/cpp/server/backends/sd_server.cpp
@@ -111,7 +111,15 @@ InstallParams SDServer::get_install_params(const std::string& backend, const std
 #else
         throw std::runtime_error("ROCm sd.cpp only supported on Windows and Linux");
 #endif
-    } else {
+        } else if (resolved_backend == "vulkan") {
+    #ifdef _WIN32
+        params.filename = "sd-" + short_version + "-bin-win-vulkan-x64.zip";
+    #elif defined(__linux__)
+        params.filename = "sd-" + short_version + "-bin-Linux-Ubuntu-24.04-x86_64-vulkan.zip";
+    #else
+        throw std::runtime_error("Vulkan sd.cpp only supported on Windows and Linux");
+    #endif
+        } else {
         // CPU build (default)
     #ifdef _WIN32
         params.filename = "sd-" + short_version + "-bin-win-avx2-x64.zip";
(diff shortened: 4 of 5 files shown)


src/cpp/server/system_info.cpp
```diff
@@ -191,6 +191,12 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         }},
     }},
 
+    // stable-diffusion.cpp - Vulkan backend (Windows/Linux x86_64)
+    {"sd-cpp", "vulkan", {"windows", "linux"}, {
+        {"cpu", {"x86_64"}},
```
fl0rianr, line 196: This is confusing because sd-cpp:cpu is already an existing and supported backend. Adding cpu/x86_64 under the new sd-cpp:vulkan backend makes it look like Vulkan is adding CPU support, but CPU support already exists through the dedicated sd-cpp:cp` path.

Today the expected matrix is roughly:

* AMD systems: sd-cpp:rocm where supported, otherwise sd-cpp:cpu
* NVIDIA systems: currently sd-cpp:cpu
* CPU-only systems: sd-cpp:cpu

With this change, sd-cpp:vulkan can become supported on CPU-only x86_64 systems because the device constraints are evaluated as OR. Since the Vulkan entry is inserted before cpu, it may also become the default backend where previously sd-cpp:cpu would have been selected. I guess you understand why this is critical.

I would prefer something like AMD + NVIDIA GPU support would be much clearer, assuming the published Vulkan artifacts are vendor-neutral:

```cpp
{"sd-cpp", "vulkan", {"windows", "linux"}, {
    {"amd_gpu", {}},
    {"nvidia_gpu", {}},
}},
```
I can test later if that would work on NV.

fl0rianr (changes requested):
Thanks for adding Vulkan here! I don’t think we can merge it quite like this yet, though.

I addressed my main concern in the inline comment. Apart from that, the change is fairly small and should be straightforward to integrate if the backend works as intended.

Before merging, I’d like to see tests or validation for:

* CPU-only systems still selecting `sd-cpp:cpu` (for example via a Docker setup or another CPU-only environment)
* AMD GPU systems being able to select and run `sd-cpp:vulkan` (re-testing the scenario already mentioned in the PR description)
* NVIDIA GPU behavior being explicit. I can help test this; if the backend is not supported there, we won't enable it for NVIDIA.

sofiageo:
Just to add my thoughts because I had also started working on this, on my system there are more things required for a good user experience. Things I noticed: 

- Vulkan needs --vae-tiling or --vae-conv-direct for 1024x1024 images
- Upscaler currently defaults to the first available backend (ROCm on my machine)
- When I started my stable diffusion model, ROCm would download even if Vulkan was selected in options

I believe trying to run Vulkan shouldn't affect the ROCm installation in any way. So IMO these should be fixed before merging. (I can take a look tomorrow if I'm not far away from my PC)

fl0rianr:
Very short update here, if done like I proposed, Nvidia works also.

ckuethe:
I'll look into that later this evening.

Can you think of a reason the vulkan back end should't run on a CPU-only host? Mesa supports vulkan on CPU via llvmpipe, and while it may not be fast (possibly slower than using CPU directly) it seems like it should be allowed. Making CPU preferred over Vulkan (for now) is fine by me, as long as people can force vulkan if they want.

Will have to see what happens on my RX7900 box, which has a working rocm back end.

src/cpp/server/system_info.cpp
jeremyfowers: @fl0rianr this is a quirk of how system-info deals with Vulkan. See the llamacpp entry a few lines up:

```
 {"llamacpp", "vulkan", {"windows", "linux"}, {
        {"cpu", {"x86_64"}},
    }},
```

What we are saying here is that virtually all modern PC x84_64 CPUs have an integrated GPU, and that is fair game for the Vulkan backend. 

I know it looks weird, but this allows us to support things like Intel ARC GPUs even though the codebase has no explicit ARC support.

ckuethe: Yep, my old "Intel(R) Core(TM) Ultra 7 155U" laptop had whatever iGPU Intel glued on, and llama.cpp would use it though vulkan (albeit slowly). I expect stable diffusion would do the same.

fl0rianr:
Thanks to Jeremy’s explanation, I now agree that my earlier remark about the cpu/x86_64 entry was too strict.

Before merge, I still see three points we should address:

1. Add NVIDIA to the sd-cpp:vulkan support entry as well, since Vulkan image generation works on NVIDIA in testing.
2. Add a narrow workaround for Strix Halo / Strix Point style AMD Vulkan failures at high resolutions.
3. Make sure the auto-selection behavior is correct.

For **auto-selection**, I think the desired behavior should be:

```text
ROCm-capable AMD GPU -> sd-cpp:rocm
NVIDIA GPU -> sd-cpp:vulkan
older / non-ROCm AMD GPU -> sd-cpp:vulkan
generic CPU-only fallback -> sd-cpp:cpu
```

So effectively:

```text
rocm > vulkan > cpu
```

but Vulkan should only win automatically when there is an actual GPU path where Vulkan is expected to be useful. 

Regarding **2**, @sofiageo pointed me to a useful upstream issue: leejet/stable-diffusion.cpp#1290. It is also Strix Halo / Radeon 8060S, Vulkan, 1024x1024, failing in the final VAE decode with a large Vulkan buffer allocation. The confirmed workaround there is `--vae-tiling`.

Based on that, I think the Lemonade-side workaround should be:

* only for `sd-cpp:vulkan`
* only for AMD gfx1150, gfx1151, gfx1152
* hard-code `--vae-tiling` for now (those 3 only)

A tiny helper:

```cpp
bool is_gfx115x_arch(const std::string& arch) {
    return arch == "gfx1150" || arch == "gfx1151" || arch == "gfx1152";
}
```

Then in `SDServer::load()`:

```cpp
if (resolved_backend == "vulkan") {
    if (is_gfx115x_arch(rocm_arch)) {
        LOG(INFO, "SDServer")
            << "Applying Vulkan gfx115X VAE workaround: --vae-tiling"
            << std::endl;
        args.push_back("--vae-tiling");
    }
}
```

This keeps the workaround narrow and matches the upstream-confirmed fix. I tested this path successfully.

ckuethe:
Strix Halo (gfx1151) test

Logs and `ps` show that the vae tiling flag is being applied: `2026-05-29 22:10:23.827 [Info] (SDServer) Applying Vulkan gfx115X VAE workaround: --vae-tiling` I set the back end to `vulkan` and the vulkan back end ran, both for generation and upscaling.

The vae flag isn't 100% reliable; upscaling to 1024x1024 fails with

```
May 29 22:11:42 strix lemond[15754]: ggml_vulkan: Device memory allocation of size 7272923136 failed.
May 29 22:11:42 strix lemond[15754]: ggml_vulkan: Requested buffer size exceeds device buffer size limit: ErrorOutOfDeviceMemory
May 29 22:11:42 strix lemond[15754]: [ERROR] ggml_extend.hpp:69   - ggml_gallocr_reserve_n_impl: failed to allocate Vulkan0 buffer of size 15530983436
May 29 22:11:42 strix lemond[15754]: [ERROR] ggml_extend.hpp:1897 - flux: failed to allocate the compute buffer
May 29 22:11:42 strix lemond[15754]: [ERROR] ggml_extend.hpp:2805 - flux alloc compute buffer failed
May 29 22:11:42 strix lemond[15754]: [ERROR] stable-diffusion.cpp:1967 - diffusion model compute failed
May 29 22:11:42 strix lemond[15754]: [ERROR] stable-diffusion.cpp:2061 - Diffusion model sampling failed
```

I can upscale from 128x128 to 768x768 though. The suggestion to use [GGML_VK_FORCE_MAX_BUFFER_SIZE](https://github.com/leejet/stable-diffusion.cpp/issues/1290#issuecomment-4354208421) prevents the upscale from aborting, but it doesn't produce a meaningful 1024x1024 image.

Image generation at 1024x1024 works.

Setting the back end to "auto" runs the rocm version.

ckuethe:
RX7900XT (gfx1100) test

"auto" back end runs rocm (`Initializing backend: ROCm0`)
"vulkan" back end runs vulkan (`Initializing backend: Vulkan0`), but runs out of memory unless I also give it "--vae-tiling"

1024x1024 generation works, 768x768 upscale works

ckuethe:
Krackan Point (gfx1152) test

Logs show that the vae flag is applied: `2026-05-29 22:42:56.600 [Info] (SDServer) Applying Vulkan gfx115X VAE workaround: --vae-tiling`

setting auto and vulkan both run the vulkan back end.

1024x1024 image generation works, didn't recheck upscaling this evening.

ckuethe:
Intel Core i5 / Intel(R) Iris(R) Pro Graphics 5200 (HSW GT3) test

Yeah, an 11 year old NUC probably won't work...

```
MESA-INTEL: warning: Haswell Vulkan support is incomplete
[DEBUG] ggml_extend.hpp:60   - ggml_vulkan: Found 1 Vulkan devices:
[DEBUG] ggml_extend.hpp:60   - ggml_vulkan: 0 = Intel(R) Iris(R) Pro Graphics 5200 (HSW GT3) (Intel open-source Mesa driver) | uma: 1 | fp16: 0 | bf16: 0 | warp size: 32 | shared memory: 65536 | int dot: 0 | matrix cores: none
[DEBUG] ggml_extend_backend.cpp:311  - Found 2 backend devices:
[DEBUG] ggml_extend_backend.cpp:314  - #0: Vulkan0
[DEBUG] ggml_extend_backend.cpp:314  - #1: CPU
[DEBUG] ggml_extend_backend.cpp:291  - Initializing backend: Vulkan0
ggml_vulkan: device Vulkan0 does not support 16-bit storage.
terminate called after throwing an instance of 'std::runtime_error'
   what():  Unsupported device
```

So "auto" tried to go for Vulkan. Forcing CPU mode works, though I had to use SD-Turbo since this machine only has 16GB total memory.

fl0rianr:
@ckuethe Great testing, so we are near to a working solution then. 
Just a small note, you checked up-scaling on  RX7900XT (gfx1100) test, but not the problematic 1024 variant.

I made some testing with different args and also the latest upstream does not seem to fix the issue.

But I found a potential solution, adding this as well seems to fix the issue:
`args.push_back("--diffusion-fa");`
Flash Attention changes the compute buffers so we don't see the error.

Finally, would you mind adding nvidia to the vulkan backend as well?

ckuethe:
I did try 1024 upscale on gfx1100, which didn't work. I was reporting the outer limit of what did work. Anyway, with `--diffusion-fa` all three of my test platforms can now upscale to 1024.

I had to manually add `--vae-tiling` to my RX7900, so I'm tempted to make it default, or maybe add a log message to say "if you're running out of memory, add `--vae-tiling` to your `sdcpp.args`"  I already have custom sdcpp.args: the rocm backend requires `--offload-to-cpu` on my RX7900.

fl0rianr:
@ckuethe Thanks for the update. Since you had problems on the RDNA4 dGPU let us set those flags in all cases. The trade off is limited I also tested it on Nvidia.
Function is critical, we can optimize later on if there is an upstream fix.
So I propose the following:

```cpp
if (resolved_backend == "vulkan") {
    LOG(INFO, "SDServer")
        << "Applying Vulkan SD workaround: --vae-tiling --diffusion-fa"
        << std::endl;
    args.push_back("--vae-tiling");
    args.push_back("--diffusion-fa");
}
```
Additionally we have to fix the CI runs, they failed on image gen since the backend there is not set, Vulkan is choosen before CPU but Vulkan is not installed.

So we should modify the tests so the back-ends are clearly separated and we can safely identify  what gets broken also maybe due to an upstream change etc.

**Windows**
```yaml
- name: stable-diffusion-cpu
  script: server_sd.py
  extra_args: ""
  backends: "cpu"
  runner: [Windows, rocm]

- name: stable-diffusion-vulkan
  script: server_sd.py
  extra_args: ""
  backends: "vulkan"
  runner: [Windows, vulkan, rocm]

- name: stable-diffusion-rocm
  script: server_sd.py
  extra_args: ""
  backends: "rocm"
  runner: [Windows, rocm]
```

**Linux**
```yaml
- name: stable-diffusion-cpu
  script: server_sd.py
  extra_args: ""
  backends: "cpu"
  runner: [Linux, rocm]

- name: stable-diffusion-vulkan
  script: server_sd.py
  extra_args: ""
  backends: "vulkan"
  runner: [Linux, vulkan, rocm]

- name: stable-diffusion-rocm
  script: server_sd.py
  extra_args: ""
  backends: "rocm"
  runner: [Linux, rocm]
```

Good work so far, let me know if I can help you get it done.

ckuethe:
For some reason GH isn't showing me the failed test logs, so I just copied the pattern from the llama runner

joshefin:
For Strix Halo, you can try setting environment variables `GGML_VK_FORCE_MAX_BUFFER_SIZE` and `GGML_VK_FORCE_MAX_ALLOCATION_SIZE` to larger than default (2GB) value, for example `8589934591`. Instead of using vae tiling.

ckuethe:
@joshefin I'm open to revisiting the vae tuning in a new diff. That said, is there a way to automatically compute the right values for those variables, especially on vastly differing GPU architectures so that people don't have to try a bunch of different values? If so, shouldn't that be a thing sdcpp should do itself?

fl0rianr:
@ckuethe If you can test those, would be great! The current CI fail is unrelated and will be green after re-run of that test.

ckuethe:
turning up those environment variables on gfx1152 doesn't work. I tried a few different values, which makes the error message go away, but also doesn't produce a meaningful image... just this:

lemonade_upscale_this_1780184476206

fl0rianr:
I'm also trying it, taking an eternity, I think we go with what we have here. Much more stable. Any objections?

fl0rianr (approved):

PR #2044, opened 2026-05-29
Allow LEMONADE_GGML_HIP_PATH to locate the HIP plugin on non-FHS installs

noamsto:
Allow `LEMONADE_GGML_HIP_PATH` to locate `libggml-hip.so` on non-FHS installs

## Problem

`is_ggml_hip_plugin_available()` probes only a fixed set of FHS paths for `libggml-hip.so`:

```
/usr/lib/x86_64-linux-gnu/ggml/backends0/libggml-hip.so   (Debian/Ubuntu)
/usr/lib/libggml-hip.so                                   (Arch AUR)
/usr/lib/ggml/backends0/libggml-hip.so
/usr/lib64/ggml/backends0/libggml-hip.so
```

On distros/packaging that install the plugin elsewhere — NixOS, Guix, custom prefixes, relocatable installs — the `system` llamacpp recipe is reported as `HIP plugin libggml-hip.so not installed` on any machine with an AMD GPU (`/sys/class/kfd` present), even when a perfectly good `libggml-hip.so` exists. There is currently no way to override the lookup.

## Fix

Add a `LEMONADE_GGML_HIP_PATH` environment variable that points directly at the plugin file. When it is set and names a regular file, the plugin is considered available; otherwise the existing FHS probe runs unchanged.

- Default behavior on Debian/Ubuntu/Arch is **identical** (the env var is opt-in).
- `fs::is_regular_file` follows symlinks, so a symlink into a store/prefix resolves correctly, and a stale path or directory does **not** falsely pass the gate.
- Scoped inside the existing `#ifdef __linux__`; non-Linux builds are unaffected.

## Test

Built `lemonade` from source on NixOS where `libggml-hip.so` lives under the Nix store. Before: `system` + ROCm shows as not installed. After: `LEMONADE_GGML_HIP_PATH=/nix/store/.../lib/libggml-hip.so` makes the `system` recipe report installed and ROCm acceleration works.

### DIFF
--- a/docs/guide/configuration/llamacpp.md
+++ b/docs/guide/configuration/llamacpp.md
@@ -53,6 +53,13 @@ Lemonade uses [llama.cpp](https://github.com/ggerganov/llama.cpp) as its primary
 - **Performance**: Depends on build configuration
 - **Installation**: Requires manual installation of `llama-server` in system PATH
 - **Notes**: Not enabled by default; set `LEMONADE_LLAMACPP_PREFER_SYSTEM=true` in config
+- **HIP plugin on non-standard paths**: When an AMD GPU is present, the system backend needs the GGML HIP plugin (`libggml-hip.so`). Lemonade looks for it in the standard system library paths. If your distribution or package manager installs it elsewhere (e.g. NixOS, a custom prefix, or a manual build), set `LEMONADE_GGML_HIP_PATH` to the full path of the plugin so the backend is reported as available:
+
+  ```bash
+  export LEMONADE_GGML_HIP_PATH=/opt/rocm/lib/libggml-hip.so
+  ```
+
+  The filename must look like `libggml-hip*.so*` (versioned sonames such as `libggml-hip.so.0` are accepted). This Linux-only variable is used solely to detect plugin availability; it is not forwarded to the GGML loader, so it does not change where llama.cpp actually loads the plugin from.
 
 ## ROCm Channel Configuration
 
--- a/src/cpp/server/utils/path_utils.cpp
+++ b/src/cpp/server/utils/path_utils.cpp
@@ -269,6 +269,11 @@ std::string find_executable_in_path(const std::string& executable_name) {
 
 bool is_ggml_hip_plugin_available() {
 #ifdef __linux__
+    // Allow distros/packagers that install outside the FHS paths below
+    // (e.g. NixOS, custom prefixes) to point directly at libggml-hip.so.
+    if (const char* env = std::getenv("LEMONADE_GGML_HIP_PATH"); env && *env && fs::is_regular_file(env)) {
+        return true;
+    }
     // On Linux x86_64, check common system library paths for the HIP plugin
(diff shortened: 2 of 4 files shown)


fl0rianr (commented):
Thanks, this looks like a useful escape hatch for non-FHS installs.

Before merging, could you please consider two small follow-ups?

1. Use the non-throwing filesystem overload for the env-var path, e.g. `std::error_code ec; fs::is_regular_file(path, ec)`. Since LEMONADE_GGML_HIP_PATH is user-controlled, this avoids turning odd/bad paths into filesystem exceptions.

2. Add a regression test in `test/test_llamacpp_system_backend.py`, near the existing Linux system-backend availability tests, covering a valid file and an invalid path/directory.

Also, since this becomes user-facing config, it would be helpful to document the new env var in `docs/guide/configuration/llamacpp.md`, probably in the System backend section.


One question: should this intentionally accept any regular file as a trust-me override, or should we require the filename to be `libggml-hip.so` to avoid false-positive availability?

noamsto:
Thanks for the review! Addressed all three follow-ups in d9f78611, and went with requiring the basename for your open question.

**1. Non-throwing overload** — done. The env-var path now uses `std::error_code ec; fs::is_regular_file(env, ec)`, so a malformed/odd path resolves to "not available" instead of raising `filesystem_error`.

**2. Regression tests** — added `test_005a`/`test_005b` in `test_llamacpp_system_backend.py` next to the existing Linux system-backend tests, covering a valid override, a missing path, and a directory. Note these assert against the **`system`** backend's HIP-plugin message and are gated on `/sys/class/kfd` + llama-server in PATH, since that's the only path that actually reaches `is_ggml_hip_plugin_available()` — so they're skipped on non-AMD CI. I also added a `test/cpp/test_ggml_hip_path.cpp` unit test that exercises the basename matching and the non-throwing behavior directly (deterministic, no GPU needed).

**3. Docs** — documented `LEMONADE_GGML_HIP_PATH` in the System backend section of `docs/guide/configuration/llamacpp.md`, including the filename rule and the caveat that it only affects availability detection (it isn't forwarded to ggml's loader).

**On your question (any file vs. require `libggml-hip.so`):** went with requiring the basename to look like `libggml-hip*.so*` (case-insensitive, versioned sonames like `.so.0` allowed). It catches an accidental override pointing at an unrelated file, which is the false-positive you flagged. I kept it as a loose glob rather than an exact `libggml-hip.so` so packagers shipping versioned sonames aren't locked out. It's framed in-code as a sanity check rather than a security boundary, since the path isn't forwarded to the loader and so can't actually be validated as loadable.

For what it's worth, I verified the end-to-end behavior against a built `lemond` on an AMD box (kfd present, no FHS plugin): valid `libggml-hip.so` → system backend reports `installed`; directory / wrong name / missing path → stays blocked with the "HIP plugin… not installed" message.

fl0rianr (approved):
Thanks for the update. All my concerns are addressed.

PR #2047, opened 2026-05-29
Link system cpp-httplib portably (header-only friendly)

### LINKED ISSUE #2046: From-source build cannot use a system header-only cpp-httplib
### Feature Description

Allow a from-source build to link a **system, header-only** `cpp-httplib` instead of falling back to a network `FetchContent`. Two small `CMakeLists` assumptions currently prevent this:

1. **Detection** (`CMakeLists.txt`) only probes the `cpp-httplib` pkg-config module name:
   ```cmake
   pkg_check_modules(HTTPLIB QUIET cpp-httplib>=${MIN_HTTPLIB_VERSION})
   ```
   The upstream project (`yhirose/cpp-httplib`) and several distros install the pkg-config file as **`httplib.pc`**, not `cpp-httplib.pc` (Debian is the one that renames it). So `HTTPLIB_FOUND` is false and the build silently falls through to `FetchContent`.

2. **Linking** (server core, CLI, tray) hardcodes the bare name:
   ```cmake
   target_link_libraries(lemonade-server-core PUBLIC cpp-httplib)
   ```
   This resolves only on Debian, which ships a *compiled* `libcpp-httplib.so`. `cpp-httplib` is header-only upstream, so on a header-only install `-lcpp-httplib` fails to link.

The sibling pkg-config deps (`curl`, `zstd`) already avoid both problems by linking `${X_LIBRARIES}` from `pkg_check_modules`. Making httplib consistent fixes it: fall back to the `httplib` module name, and link `${HTTPLIB_LIBRARIES}` (empty for header-only, the compiled lib on Debian). Debian still matches `cpp-httplib` first, so there is no change there.

### Use Case / Motivation

Packaging Lemonade for distros that ship `cpp-httplib` header-only (e.g. Nixpkgs) currently requires patching `CMakeLists` and synthesizing a fake `cpp-httplib.pc`, because the build otherwise reaches for the network — which fails in sandboxed/offline package builds. This also helps anyone doing a reproducible/offline from-source build with httplib already installed.

### Platform Relevance

Linux only.

### Additional Context

Only the Linux system-httplib path is affected; macOS and Windows use `FetchContent` and are untouched. I have a small patch ready (detection fallback + `${HTTPLIB_LIBRARIES}` at the three link sites) and will open a PR referencing this issue. Verified end-to-end against a header-only system httplib, and behavior on Debian is unchanged.

noamsto:
Link system `cpp-httplib` portably (header-only friendly)

Closes #2046.

## Problem

The system-httplib path assumes Debian's packaging in two places, so a from-source build cannot use a header-only system `cpp-httplib` and silently falls back to a network `FetchContent`:

1. **Detection** only probes the `cpp-httplib` pkg-config name. `yhirose/cpp-httplib` is header-only and several distros (e.g. Nixpkgs) ship it as `httplib.pc`, not `cpp-httplib.pc` — so `HTTPLIB_FOUND` is false there.
2. **Linking** hardcodes the bare name `cpp-httplib`, which only resolves on Debian's *compiled* `libcpp-httplib.so`; `-lcpp-httplib` fails on a header-only install.

## Fix

Both changes mirror how the sibling pkg-config deps (`curl`, `zstd`) are already handled:

- **Detection:** fall back to the `httplib` module name when `cpp-httplib` is not found. Debian still matches `cpp-httplib` first, so no change there.
- **Linking:** use `${HTTPLIB_LIBRARIES}` (from `pkg_check_modules`) instead of the bare `cpp-httplib` name at all three system-httplib link sites (server core, CLI, tray). It is empty for a header-only install (headers come via the existing `${HTTPLIB_INCLUDE_DIRS}` handling) and carries the compiled lib on Debian.

## Scope / testing

- Only the Linux system-httplib path changes; macOS forces `FetchContent` and Windows is unaffected.
- Verified end-to-end: a from-source build against a header-only system `cpp-httplib` now reports `Using system cpp-httplib` and links the server core + CLI with no `FetchContent` and no patching. Behavior on Debian (compiled `cpp-httplib.pc` with `Libs: -lcpp-httplib`) is unchanged — `${HTTPLIB_LIBRARIES}` resolves to the same `-lcpp-httplib`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,11 +86,15 @@ find_package(nlohmann_json ${MIN_NLOHMANN_JSON_VERSION} QUIET)
 if(PkgConfig_FOUND)
     pkg_check_modules(CURL QUIET libcurl>=${MIN_CURL_VERSION})
     pkg_check_modules(ZSTD QUIET libzstd>=${MIN_ZSTD_VERSION})
-    pkg_check_modules(HTTPLIB QUIET cpp-httplib>=${MIN_HTTPLIB_VERSION})
+    # Debian ships the package as `cpp-httplib`, Nixpkgs as `httplib` — search
+    # both. No IMPORTED_TARGET: we build our own interface target below so we
+    # can skip the .pc Cflags (see there for why).
+    pkg_search_module(HTTPLIB QUIET
+        cpp-httplib>=${MIN_HTTPLIB_VERSION}
+        httplib>=${MIN_HTTPLIB_VERSION})
     pkg_check_modules(LWS QUIET libwebsockets>=${MIN_LWS_VERSION})
 endif()
 find_path(CLI11_INCLUDE_DIRS "CLI/CLI.hpp")
-find_path(HTTPLIB_INCLUDE_DIRS "httplib.h")
 # Verify CLI11 version if found
 if(CLI11_INCLUDE_DIRS)
     if(EXISTS "${CLI11_INCLUDE_DIRS}/CLI/Version.hpp")
@@ -316,6 +320,22 @@ if(NOT USE_SYSTEM_HTTPLIB)
             ${zstd_SOURCE_DIR}/lib
             ${zstd_SOURCE_DIR}/lib/common
         )
+    endif()
+else()
+    # System cpp-httplib: a local interface target carrying the include dirs
+    # and link lib (empty for a header-only install like Nixpkgs, the compiled
+    # .so on Debian). We deliberately skip HTTPLIB_CFLAGS_OTHER: the .pc Cflags
+    # carry cpp-httplib's feature macros (-DCPPHTTPLIB_OPENSSL_SUPPORT, etc.),
+    # which would flip on header features needing link deps (libcrypto) we
+    # don't provide. Created at top level, so it is visible to the cli/tray
+    # subdirs added later.
+    add_library(lemonade-httplib INTERFACE)
+    target_include_directories(lemonade-httplib INTERFACE ${HTTPLIB_INCLUDE_DIRS})
+    if(HTTPLIB_LINK_LIBRARIES)
+        target_link_libraries(lemonade-httplib INTERFACE ${HTTPLIB_LINK_LIBRARIES})
+    else()
+        target_link_libraries(lemonade-httplib INTERFACE ${HTTPLIB_LIBRARIES})
+        target_link_directories(lemonade-httplib INTERFACE ${HTTPLIB_LIBRARY_DIRS})
     endif()
 endif()
 
@@ -640,7 +660,7 @@ endif()
 target_link_libraries(lemonade-server-core PUBLIC nlohmann_json::nlohmann_json)
 
 if(USE_SYSTEM_HTTPLIB)
-    target_link_libraries(lemonade-server-core PUBLIC cpp-httplib)
+    target_link_libraries(lemonade-server-core PUBLIC lemonade-httplib)
 else()
     target_link_libraries(lemonade-server-core PUBLIC httplib::httplib)
 endif()
@@ -664,10 +684,6 @@ if(USE_SYSTEM_ZSTD)
     target_include_directories(lemonade-server-core PUBLIC ${ZSTD_INCLUDE_DIRS})
     target_link_directories(lemonade-server-core PUBLIC ${ZSTD_LIBRARY_DIRS})
     target_compile_options(lemonade-server-core PUBLIC ${ZSTD_CFLAGS_OTHER})
-endif()
-
-if(USE_SYSTEM_HTTPLIB AND HTTPLIB_INCLUDE_DIRS)
-    target_include_directories(lemonade-server-core PUBLIC ${HTTPLIB_INCLUDE_DIRS})
 endif()
 
 # libwebsockets for the realtime and log-stream WebSocket server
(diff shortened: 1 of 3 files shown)


fl0rianr (commented):
Nice minimal fix — this should address the reported header-only httplib.pc case.

One small CMake robustness suggestion: instead of doing two pkg_check_modules() calls and then linking only ${HTTPLIB_LIBRARIES}, could we use a pkg-config imported target here?

Something like:

```cmake
pkg_search_module(HTTPLIB QUIET IMPORTED_TARGET
    cpp-httplib>=${MIN_HTTPLIB_VERSION}
    httplib>=${MIN_HTTPLIB_VERSION}
)
```

and then link PkgConfig::HTTPLIB at the three system-httplib call sites:

```cmake
target_link_libraries(lemonade-server-core PUBLIC PkgConfig::HTTPLIB)
target_link_libraries(lemonade PRIVATE PkgConfig::HTTPLIB)
target_link_libraries(${target_name} PRIVATE PkgConfig::HTTPLIB)
```

That keeps the same Debian-first behavior, models the “try either pkg-config module name” intent directly, and avoids consuming only ${HTTPLIB_LIBRARIES} while dropping other pkg-config usage requirements such as library dirs, linker options, include dirs, or compile flags. It should also let us remove the separate `HTTPLIB_INCLUDE_DIRS` handling for the system-httplib path.

I would avoid GLOBAL here since the project still supports CMake 3.12; the target is created in the top-level directory before the tray/CLI subdirectories are added, so normal imported-target visibility should be sufficient.

Not a blocker just some alternatives to consider. What do you think?

noamsto:
@fl0rianr Adopted — switched to `pkg_search_module(... IMPORTED_TARGET cpp-httplib httplib)` and link `PkgConfig::HTTPLIB` at all three sites (server core, CLI, tray). That folds the two `pkg_check_modules` calls into one, sources include dirs/cflags from the `.pc` file, and let me drop the separate `find_path(HTTPLIB_INCLUDE_DIRS)` path.

No `GLOBAL` — the target is created at top level before the cli/tray subdirs are added, so directory-scoped imported-target visibility is enough on 3.12. Pushed in 50baf9b4.

fl0rianr:
@noamsto Thanks for pushing the update. Looking at the CI shape, this seems to be failing in build/package steps rather than in the actual runtime tests, so I don’t think we should treat this as a test expectation issue.

One possible cause is that PkgConfig::HTTPLIB may be a bit too strong for this dependency. It is CMake-native, but for cpp-httplib it can also propagate compile options/definitions from the `.pc` file. Since cpp-httplib’s optional features are controlled by compile definitions, this might accidentally change the header’s feature set versus the previous build.

Maybe the safer fix is to keep the pkg_search_module() module-name fallback, but avoid the imported target and consume only the pieces we actually need.  Let me share a few ideas:

```cmake
pkg_search_module(HTTPLIB QUIET
    cpp-httplib>=${MIN_HTTPLIB_VERSION}
    httplib>=${MIN_HTTPLIB_VERSION})
```

Then define a small local interface target for Lemonade:

```cmake
if(USE_SYSTEM_HTTPLIB)
    add_library(lemonade-httplib INTERFACE)

    target_include_directories(lemonade-httplib INTERFACE
        ${HTTPLIB_INCLUDE_DIRS}
    )

    if(HTTPLIB_LINK_LIBRARIES)
        target_link_libraries(lemonade-httplib INTERFACE
            ${HTTPLIB_LINK_LIBRARIES}
        )
    else()
        target_link_libraries(lemonade-httplib INTERFACE
            ${HTTPLIB_LIBRARIES}
        )
        target_link_directories(lemonade-httplib INTERFACE
            ${HTTPLIB_LIBRARY_DIRS}
        )
    endif()

    # Intentionally do not propagate HTTPLIB_CFLAGS_OTHER here: cpp-httplib
    # feature macros can change header behavior and optional link deps.
endif()
```

And link `lemonade-httplib` at the three existing system-httplib sites.

That should preserve the original compile semantics more closely while still fixing the two original portability issues: alternate pkg-config module name and header-only installs with empty Libs.

If the failing logs show a specific compile/link error, we can tune this further, but this seems like the least risky direction from the CI signal.

noamsto:
@fl0rianr Confirmed from the failure logs — it's a link error, not a test-expectation issue, and your read on the cause was right.

All three failing jobs (Build 24.04, .deb Package, Registry Tests) die the same way:

```
/usr/bin/ld: undefined reference to symbol 'OPENSSL_thread_stop@@OPENSSL_3.0.0'
/usr/bin/ld: /lib/x86_64-linux-gnu/libcrypto.so.3: error adding symbols: DSO missing from command line
```

The compile line shows `-DCPPHTTPLIB_OPENSSL_SUPPORT -DCPPHTTPLIB_ZLIB_SUPPORT -DCPPHTTPLIB_BROTLI_SUPPORT` coming in, and we never define those ourselves — so they're exactly the `.pc` Cflags that `PkgConfig::HTTPLIB` propagated. They flip on header features that reference OpenSSL symbols, but `-lcrypto` is never on our link line. On `main` the bare `cpp-httplib` link name carried no Cflags, so the feature stayed off and it linked.

Went with your interface-target suggestion. Dropped `IMPORTED_TARGET` and build a local `lemonade-httplib` INTERFACE target that carries `${HTTPLIB_INCLUDE_DIRS}` and the link lib (`${HTTPLIB_LINK_LIBRARIES}`, falling back to `${HTTPLIB_LIBRARIES}` + `${HTTPLIB_LIBRARY_DIRS}`), but deliberately **not** `${HTTPLIB_CFLAGS_OTHER}` — so the header keeps its default feature set, matching the previous Debian build. Linked at all three sites (server core, CLI, tray). Empty Libs on a header-only install (Nixpkgs) still works since includes come through the same target.

Pushed as `13e678e1`. Could you approve the workflow run when you get a chance? The force-push re-gated it on `action_required`.

fl0rianr (approved):
Looks good to me now. Thanks for iterating on this.

Issue #2014, opened 2026-05-25
Applications using OpenAI API cannot use `extra_models_dir` models, trigger attempts to download


korgano:
### Platform

Windows

### Lemonade Version

10.6.0

### Component

Other / Not sure

### Bug Description

Using Lemonade as a server for other AI apps (custom scripts, AnythingLLM, POML VS Code extension, etc...) appear to be unable to use any models in `extra_models_dir`. They can see the models in the list, but when they attempt to load, load fails and triggers an attempt to download the model from HuggingFace.

Using `user.` or `extra.` prefixes to the user generated model names has no effect either.

### Steps to Reproduce

1. Configure Lemonade to enable `extra_models_dir`
2. Download models into `extra_models_dir`
3. Ensure models are visible in `user_models.json`
4. Connect to a program using OpenAI API
5. Go to model selection
6. Pick a model in the `extra_models_dir`
7. Lemonade Server receives request
8. Lemonade Server begins auto-loading model
9. Lemonade Server generates `Model not cahched, downloading from Hugging Face` log message
10. Lemonade attempts to download model
11. If model download fails, timeout error in app

### Expected vs Actual Behavior

Expected: Model would load normally, like models stored in `.cache\huggingface`
Actual: Models in `extra_models_dir` fail to load, trigger model download

### Log Output

```shell

```

anditherobot:
Fixed in #2048.

**What was happening:** It was a naming mix-up. The server tracks these local models internally as extra.my-model. When the OpenAI API asked if \my-model\ was ready, the server didn't translate the name, couldn't find it in the cache, and mistakenly tried to pull it from the internet instead.

The PR teaches the download-checker to translate the names correctly so local files are recognized instantly without triggering a download.

PR #2048, opened 2026-05-29
fix(server): translate model names before checking if they are downloaded (#2014)

anditherobot:
# OpenAI API cannot load models from `extra_models_dir`

## Description

When Lemonade is used as an OpenAI-compatible server, client apps can see models from `extra_models_dir`, but they cannot load them.

When a client selects one of these models, Lemonade treats it as missing and tries to download it from Hugging Face instead of loading the local model file.

Using the `extra.` prefix does not fix the issue.

## Platform

Windows

## Lemonade Version

10.6.0

## Steps to Reproduce

1. Enable `extra_models_dir`.
2. Add a model to `extra_models_dir`.
3. Start Lemonade Server.
4. Connect an OpenAI-compatible client, such as AnythingLLM, a custom script, or the POML VS Code extension.
5. Select the model from `extra_models_dir`.
6. Send a chat completion request.

## Expected Behavior

Lemonade loads the local model from `extra_models_dir`.

## Actual Behavior

Lemonade treats the model as missing and tries to download it from Hugging Face.

Example log message:

```text
Model not cached, downloading from Hugging Face
```

If the download fails, the client eventually times out.

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1250,7 +1250,7 @@ std::string ModelManager::resolve_model_path(const ModelInfo& info, const std::s
                 return filepath;
             }
         }
-        
+
         // Case 5: Local quant-token fallback.
         //
         // Keep the existing resolver cases above as the primary logic: exact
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Nice change, thanks!

What is your reasoing for this rather large test code occopaning this?
My take would tend in that direction:
Keep one lean regression test that proves the OpenAI auto-load path no longer treats an extra_models_dir public alias as missing.

fl0rianr (approved):
Thanks, good work, tested it.

PR #2049, opened 2026-05-30
feat: Add WSL support via librocdxg

julianxhokaxhiu:
This PR will add support for [ROCm on WSL](https://rocm.docs.amd.com/projects/radeon-ryzen/en/latest/docs/install/installrad/wsl/howto_wsl.html) which provides a `/dev/dxg` node on the WSL Linux instance.

Lemonade by default to find out if the machine supports ROCm used to query `/sys/class/kfd` which is usually available only on native Linux installations with `AMDGPU` drivers. This patch will add on top the detection of `/dev/dxg` and will read informations about the nodes using the `libhsa-runtime64.so` library, which is fully optional. If not found, lemonade will keep thinking there is no GPU.

The flow to detect ROCm for AMD GPUs is then the following:
1. Test `/dev/kfd`, if not found
2. Test `/dev/dxg`, if not found
3. Let the user know no GPUs were found that support ROCm

Regarding information pulling, the flow is the following:
1. Pull information from `/sys/class/kfd`, if not found
2. Pull information via `libhsa-runtime64.so` using `hsa_amd_memory_pool_get_info`

Finally I added also a docker command to run lemonade on WSL ( with docker installed in it ) to fully be able to pass through the ROCm layer.

Here is the final result screenshot of me testing the whole setup:
- `Windows 11 v10.0.26200.8524`
- `AMD Adrenalin 26.5.2` on AMD Radeon RX 9070 XT
- `WSL 2.7.3.0`
- `ROCm 7.13.0`
- `Docker 29.5.1`
- `Lemonade devel`
- `llamacpp rocm b9247`

immagine

---

I want to make clear that this PR was assited by GPT Codex 5.3 while however trying my best to steer it towards a minimal and as possible lightweight implementation. I'm totally open for feedback if anything needs to be changed for a better fit towards the overall code project. Thank you in advance!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -736,7 +736,7 @@ elseif(UNIX)
 
     # Link libcap for capability management on Linux
     if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
-        target_link_libraries(lemonade-server-core PUBLIC ${CMAKE_DL_LIBS})
+        target_link_libraries(lemonade-server-core PRIVATE ${CMAKE_DL_LIBS})
 
         if(PkgConfig_FOUND)
             pkg_check_modules(LIBCAP QUIET libcap)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -736,6 +736,8 @@ elseif(UNIX)
 
     # Link libcap for capability management on Linux
     if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        target_link_libraries(lemonade-server-core PUBLIC ${CMAKE_DL_LIBS})
+
         if(PkgConfig_FOUND)
             pkg_check_modules(LIBCAP QUIET libcap)
             if(LIBCAP_FOUND)
--- a/docs/guide/install/docker.md
+++ b/docs/guide/install/docker.md
@@ -63,6 +63,30 @@ docker run -d \
 
 > This will run the server using the ROCm backend as the default for llama.cpp.
 
+### Docker Run with AMD GPU Passthrough using ROCm on WSL
+
+Make sure you follow install steps described in [ROCm for WSL](https://rocm.docs.amd.com/projects/radeon-ryzen/en/latest/docs/install/installrad/wsl/howto_wsl.html)
+
+Then:
+
+```bash
+docker run -d \
+  --name lemonade-server \
+  -p 13305:13305 \
+  -v lemonade-cache:/root/.cache/huggingface \
+  -v lemonade-llama:/opt/lemonade/llama \
+  -v lemonade-recipe:/root/.cache/lemonade \
+  -v /usr/lib/wsl/lib:/usr/lib/wsl/lib:ro \
+  -v /opt/rocm/lib:/opt/rocm/lib:ro \
+  -v /opt/rocm/share/rocdxg:/opt/rocm/share/rocdxg:ro \
+  -e LD_LIBRARY_PATH=/opt/rocm/lib:/opt/rocm/lib/rocm_sysdeps/lib:/usr/lib/wsl/lib:/usr/lib \
+  -e LEMONADE_LLAMACPP=rocm \
+  --device=/dev/dxg \
+  ghcr.io/lemonade-sdk/lemonade-server:latest
+```
+
+> This will run the server using the ROCm backend as the default for llama.cpp.
+
 ### Other Docker Methods
 
 #### Docker Compose Setup
--- a/docs/guide/install/docker.md
+++ b/docs/guide/install/docker.md
@@ -78,7 +78,6 @@ docker run -d \
   -v lemonade-recipe:/root/.cache/lemonade \
   -v /usr/lib/wsl/lib:/usr/lib/wsl/lib:ro \
   -v /opt/rocm/lib:/opt/rocm/lib:ro \
-  -v /opt/rocm/share/rocdxg:/opt/rocm/share/rocdxg:ro \
   -e LD_LIBRARY_PATH=/opt/rocm/lib:/opt/rocm/lib/rocm_sysdeps/lib:/usr/lib/wsl/lib:/usr/lib \
   -e LEMONADE_LLAMACPP=rocm \
   --device=/dev/dxg \
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -160,7 +160,7 @@ constexpr hsa_amd_memory_pool_location_t HSA_AMD_MEMORY_POOL_LOCATION_CPU = 0;
 constexpr hsa_amd_memory_pool_location_t HSA_AMD_MEMORY_POOL_LOCATION_GPU = 1;
 
 constexpr uint32_t HSA_AMD_MEMORY_POOL_GLOBAL_FLAG_KERNARG_INIT = 1;
-constexpr uint64_t HSA_AMD_MEMORY_PROPERTY_AGENT_IS_APU = (1ull << 0);
+constexpr uint8_t HSA_AMD_MEMORY_PROPERTY_AGENT_IS_APU = (1u << 0);
 
 using HsaAgentCallback = hsa_status_t (*)(hsa_agent_t, void*);
 using HsaMemoryPoolCallback = hsa_status_t (*)(hsa_amd_memory_pool_t, void*);
@@ -369,13 +369,13 @@ hsa_status_t collect_hsa_agent_info(hsa_agent_t agent, void* data) {
         rocm_agent.display_name = arch;
     }
 
-    uint64_t memory_properties = 0;
+        uint8_t memory_properties[8] = {0};
     if (context->api->agent_get_info(
             agent,
             HSA_AMD_AGENT_INFO_MEMORY_PROPERTIES,
-            &memory_properties) == HSA_STATUS_SUCCESS) {
+            memory_properties) == HSA_STATUS_SUCCESS) {
         rocm_agent.is_integrated =
-            (memory_properties & HSA_AMD_MEMORY_PROPERTY_AGENT_IS_APU) != 0;
+            (memory_properties[0] & HSA_AMD_MEMORY_PROPERTY_AGENT_IS_APU) != 0;
     }
     rocm_agent.vram_gb = pool_context.largest_global_pool_gb;
 
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -201,17 +201,6 @@ bool is_dxg_rocm_environment() {
     return fs::exists("/dev/dxg");
 }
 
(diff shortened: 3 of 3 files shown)


julianxhokaxhiu:
Rebased on top of the latest main

src/cpp/server/system_info.cpp
```diff
@@ -116,6 +117,304 @@ const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
...
+        return HSA_STATUS_SUCCESS;
+    }
+
+    size_t pool_size = 0;
+    if (context->api->amd_memory_pool_get_info(pool, HSA_AMD_MEMORY_POOL_INFO_SIZE, &pool_size) != HSA_STATUS_SUCCESS) {
+        return HSA_STATUS_SUCCESS;
+    }
+
+    const double pool_gb = round_gb(static_cast<double>(pool_size) / (1024.0 * 1024.0 * 1024.0));
+    context->largest_global_pool_gb = std::max(context->largest_global_pool_gb, pool_gb);
```
fl0rianr, line 303: This currently treats the largest global non-kernarg pool as VRAM. That seems risky on WSL/APU/container setups, where a global pool may be host/system memory rather than GPU-local memory. ROCr exposes HSA_AMD_MEMORY_POOL_INFO_LOCATION and HSA_AMD_MEMORY_POOL_LOCATION_GPU; can we filter for GPU-local pools before reporting this as `vram_gb`?

src/cpp/server/system_info.cpp
```diff
@@ -116,6 +117,304 @@ const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
...
+        return HSA_STATUS_SUCCESS;
+    }
+
+    char arch_name[64] = {0};
+    char marketing_name[64] = {0};
+    char vendor_name[64] = {0};
+
+    if (context->api->agent_get_info(agent, HSA_AGENT_INFO_NAME, arch_name) != HSA_STATUS_SUCCESS ||
+        context->api->agent_get_info(agent, HSA_AGENT_INFO_VENDOR_NAME, vendor_name) != HSA_STATUS_SUCCESS ||
+        context->api->agent_get_info(agent, static_cast<hsa_agent_info_t>(HSA_AMD_AGENT_INFO_PRODUCT_NAME), marketing_name) != HSA_STATUS_SUCCESS) {
```
fl0rianr: I don't think HSA_AMD_AGENT_INFO_PRODUCT_NAME should be required for detecting the GPU. If this query fails on a particular ROCm/WSL runtime, we skip an otherwise valid GPU even though DEVICE, NAME, and VENDOR_NAME succeeded. Can we make product name optional and fall back to `arch_name` for the display name?

src/cpp/server/system_info.cpp
```diff
@@ -116,6 +117,304 @@ const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
...
+    } else {
+        rocm_agent.display_name = arch;
+    }
+
+    const std::string classification_name = marketing.empty() ? rocm_agent.display_name : marketing;
+    rocm_agent.is_integrated = !is_amd_discrete_gpu_name(classification_name) &&
+        (to_lower_copy(arch).find("gfx1150") == 0 || to_lower_copy(arch).find("gfx1151") == 0);
+    rocm_agent.vram_gb = pool_context.largest_global_pool_gb;
+
+    const std::string dedupe_key = rocm_agent.display_name + "|" + rocm_agent.arch_name;
```
fl0rianr: This dedupes agents by display_name|arch_name. On a system with two identical GPUs, this would collapse distinct devices into one. Can we either remove this dedupe or include a stable per-agent identifier, e.g. BDFID / driver UID / agent handle, in the key?

src/cpp/server/system_info.cpp
```diff
@@ -2869,9 +3179,16 @@ std::vector<GPUInfo> LinuxSystemInfo::detect_amd_gpus(const std::string& gpu_typ
     std::string kfd_path = "/sys/class/kfd/kfd/topology/nodes";
 
     if (!fs::exists(kfd_path)) {
+        auto dxg_gpus = query_dxg_amd_gpus(gpu_type);
```
fl0rianr, line 3198: The /dev/dxg fallback only runs when the KFD topology path is absent. Should we also try the HSA /dev/dxg path when KFD exists but yields no matching AMD GPUs? That would make this more robust for WSL/container edge cases where the topology path may exist but not provide usable GPU nodes.

src/cpp/server/system_info.cpp
```diff
@@ -116,6 +117,304 @@ const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
...
+    if (!marketing.empty() && marketing != arch) {
+        rocm_agent.display_name = marketing + " (" + arch + ")";
+    } else if (!marketing.empty()) {
+        rocm_agent.display_name = marketing;
+    } else {
+        rocm_agent.display_name = arch;
+    }
+
+    const std::string classification_name = marketing.empty() ? rocm_agent.display_name : marketing;
+    rocm_agent.is_integrated = !is_amd_discrete_gpu_name(classification_name) &&
```
fl0rianr: Can we avoid using the marketing name for integrated/discrete classification here? The existing Linux/KFD path does not classify AMD GPUs from names either: it uses the KFD node's `drm_render_minor` and then checks the DRM sysfs device (`/sys/class/drm/renderD*/device/board_info`) via get_amd_is_igpu(). For the new HSA /dev/dxg path, we should similarly prefer runtime/topology metadata, e.g. HSA_AMD_AGENT_INFO_MEMORY_PROPERTIES / HSA_AMD_MEMORY_PROPERTY_AGENT_IS_APU, instead of marketing-name keywords or gfx1151 etc.

CMakeLists.txt
```diff
@@ -736,6 +736,8 @@ elseif(UNIX)
 
     # Link libcap for capability management on Linux
     if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        target_link_libraries(lemonade-server-core PUBLIC ${CMAKE_DL_LIBS})
```
fl0rianr: Is PUBLIC needed here? dlopen/dlsym are only used internally by `lemonade-server-core`, so this should probably be `PRIVATE ${CMAKE_DL_LIBS}` unless consumers of this target also need to inherit it.

docs/guide/install/docker.md
```diff
@@ -63,6 +63,30 @@ docker run -d \
...
+  -p 13305:13305 \
+  -v lemonade-cache:/root/.cache/huggingface \
+  -v lemonade-llama:/opt/lemonade/llama \
+  -v lemonade-recipe:/root/.cache/lemonade \
+  -v /usr/lib/wsl/lib:/usr/lib/wsl/lib:ro \
+  -v /opt/rocm/lib:/opt/rocm/lib:ro \
+  -v /opt/rocm/share/rocdxg:/opt/rocm/share/rocdxg:ro \
+  -e LD_LIBRARY_PATH=/opt/rocm/lib:/opt/rocm/lib/rocm_sysdeps/lib:/usr/lib/wsl/lib:/usr/lib \
+  -e LEMONADE_LLAMACPP=rocm \
+  --device=/dev/dxg \
```
fl0rianr, line 83: Can we clarify which of these mounts are required for ROCm-on-WSL and which are specific to the tested installation? In particular `/opt/rocm/share/rocdxg` and the full LD_LIBRARY_PATH may not be obvious to users copying this command.

fl0rianr (changes requested):
Thanks for the detailed WSL/ROCm/Docker test report — that is a strong positive signal and makes this much easier to review. I like the minimal runtime-linking approach via dlopen. Great work overall.

But I don’t think this is quite merge-ready yet, mainly because the new HSA detection path has a few robustness issues mentioned below in the specific comments.

I made it very detailed to help you improving and hardening the logic.

docs/guide/install/docker.md
julianxhokaxhiu: `LD_LIBRARY_PATH` is required as I did try to run it without and rocm unfortunately was not available, so I think it's a key part required because of the `ldopen` call inside our code, not having the path in an usuar `/usr/lib` directory, most likely the so is being missed while being searched.

`/opt/rocm/share/rocdxg` probably can be removed, I'll do some tests and let you know :) Good catch, thanks!

CMakeLists.txt
julianxhokaxhiu: Oh yeah, not sure why the LLM put this to PUBLIC but I agree with you, PRIVATE should have been since the beginning. I'll fix it. Cheers!

docs/guide/install/docker.md
julianxhokaxhiu: Patch pushed, tested and works as intended even without that mount :) `LD_LIBRARY_PATH` still required

CMakeLists.txt
julianxhokaxhiu: Patch pushed, tested and works as intended :)

src/cpp/server/system_info.cpp
julianxhokaxhiu: Done, patch pushed

src/cpp/server/system_info.cpp
julianxhokaxhiu: Done, patch pushed

src/cpp/server/system_info.cpp
julianxhokaxhiu: Done as well, patch pushed

src/cpp/server/system_info.cpp
julianxhokaxhiu: I like this one, thanks a lot for the feedback :) Patch added

src/cpp/server/system_info.cpp
julianxhokaxhiu: Done, now I'm using the `agent_get_info` HSA API to find out which kind of GPU we have, does that work? Let me know, patch pushed, ty :)

julianxhokaxhiu:
There are some minor compilation issues, I'll fix them now

julianxhokaxhiu:
Fixed, tested, and everything seems to work as intended

immagine

fl0rianr:
@julianxhokaxhiu Great work! This is much better than the previous name-based classification. One small ABI concern: ROCr documents HSA_AMD_AGENT_INFO_MEMORY_PROPERTIES as `uint8_t[8]`, not `uint64_t`. Can we mirror that ABI surface exactly and test bit 0 from the byte array?

julianxhokaxhiu:
Done :) Compiles successfully and runs fine as before

fl0rianr:
Oh, I still saw something, if you just remove the unused `is_amd_discrete_gpu_name()` I'm gonna approve.

julianxhokaxhiu:
Good catch :) Thanks a lot!

fl0rianr (approved):

PR #2055, opened 2026-06-01
docs: fill coverage gaps across API reference, CLI, Omni, and app dev guide

kenvandine:
API reference:
- openai.md: document enable_thinking/thinking params for reasoning model control;
  add SSE streaming client-side guidance (reconnection, cold-start, error handling)
- lemonade.md: add GET /v1/system-stats, POST /v1/log-level, POST /v1/params
  endpoint sections with request/response examples; add sdcpp_args, flm_args,
  and vllm_args to the /v1/load parameters table; update summary table
- anthropic.md: expand from a 2-line stub to a full API reference with parameter
  table, curl + Python SDK examples, non-streaming and streaming response formats,
  and error response table

CLI reference:
- cli.md: add lemonade config to the Server commands table and ToC; add
  "Options for config" section with examples and link to /v1/params

Omni/collection docs:
- lemonade-omni.md: add REST API section for collection create/query/delete;
  document component loading behavior and LRU eviction implications; add
  chat-transcription model guide covering when to use native multimodal models
  vs OmniRouter, message format, and label-based discovery

New app developer guide:
- building-apps.md: new guide covering connection pattern, model discovery,
  inference examples (chat, embeddings, images, audio), streaming best practices,
  runtime model management, error handling reference table, authentication, and
  performance tips
- dev/README.md: restructure to surface building-apps.md and examples index for
  app developers alongside the existing contributor docs

https://claude.ai/code/session_01RZAtERd26JPV38HTa8Sj9d

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -20,6 +20,9 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/stats`](#get-v1stats) | Performance statistics from the last request |
 | `GET` | [`/v1/system-stats`](#get-v1system-stats) | Current host resource usage |
 | `GET` | [`/v1/system-info`](#get-v1system-info) | System information and device enumeration |
+| `GET` | [`/v1/system-stats`](#get-v1system-stats) | Live CPU, RAM, and VRAM usage |
+| `POST` | [`/v1/log-level`](#post-v1log-level) | Change the server log verbosity at runtime |
+| `POST` | [`/v1/params`](#post-v1params) | Read and update server configuration at runtime |
 | `POST` | [`/v1/install`](#post-v1install) | Install or update a backend, or register a cloud provider |
 | `POST` | [`/v1/uninstall`](#post-v1uninstall) | Remove a backend or cloud provider |
 | `POST` | [`/v1/cloud/auth`](#post-v1cloudauth) | Set an in-memory API key for a cloud provider |
@@ -448,10 +451,13 @@ Explicitly load a registered model into memory. This is useful to ensure that th
 | `llamacpp_args` | No | llamacpp | Custom arguments to pass to llama-server. The following are NOT allowed: `-m`, `--port`, `--ctx-size`, `-ngl`, `--jinja`, `--mmproj`, `--embeddings`, `--reranking`. |
 | `whispercpp_backend` | No | whispercpp | WhisperCpp backend: `npu` or `cpu` on Windows; `cpu` or `vulkan` on Linux. Default is `npu` if supported. |
 | `whispercpp_args` | No | whispercpp | Custom arguments to pass to whisper-server. The following are NOT allowed: `-m`, `--model`, `--port`. Example: `--convert`. |
+| `sdcpp_args` | No | sd-cpp | Custom arguments to pass to `sd-server`. The following are NOT allowed: `-m`, `--port`, `--steps`, `--cfg-scale`, `--width`, `--height`. |
 | `steps` | No | sd-cpp | Number of inference steps for image generation. Default: 20. |
 | `cfg_scale` | No | sd-cpp | Classifier-free guidance scale for image generation. Default: 7.0. |
 | `width` | No | sd-cpp | Image width in pixels. Default: 512. |
 | `height` | No | sd-cpp | Image height in pixels. Default: 512. |
+| `flm_args` | No | flm | Custom arguments to pass to `flm serve` (e.g., `"--socket 20 --q-len 15"`). |
+| `vllm_args` | No | vllm | Custom arguments to pass to `vllm-server` (e.g., `"--max-num-seqs 128 --enable-prefix-caching"`). |
 | `merge_args` | No | All | Boolean. If true (default), `*_args` values from global config and per-model config are merged (per-model takes priority). If false, per-model `*_args` replace global `*_args` entirely. |
 
 **Setting Priority:**
@@ -1371,3 +1377,146 @@ curl http://localhost:13305/live
 ```json
 {"status":"ok"}
 ```
+
+## `GET /v1/system-stats`
+<sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
+
+Live resource usage snapshot: CPU load, RAM, and VRAM. Useful for building resource monitors or deciding whether there is headroom to load an additional model.
+
+### Parameters
+
+This endpoint takes no parameters.
+
+### Example request
+
+```bash
+curl http://localhost:13305/v1/system-stats
+```
+
+### Response format
+
+```json
+{
+  "cpu_percent": 12.4,
+  "ram_used_gb": 18.2,
+  "ram_total_gb": 32.0,
+  "ram_percent": 56.9,
+  "vram": [
+    {
+      "device": "AMD Radeon(TM) 890M Graphics",
+      "used_gb": 3.1,
+      "total_gb": 8.5,
+      "percent": 36.5
+    }
+  ]
+}
+```
+
+**Field descriptions:**
+
+- `cpu_percent` — Current CPU utilization across all cores (0–100).
+- `ram_used_gb` — RAM currently in use, in GB.
+- `ram_total_gb` — Total physical RAM, in GB.
+- `ram_percent` — RAM utilization percentage (0–100).
+- `vram` — Array of GPU VRAM entries, one per detected GPU.
+  - `device` — GPU display name.
+  - `used_gb` — VRAM currently in use, in GB.
+  - `total_gb` — Total VRAM for this GPU, in GB.
+  - `percent` — VRAM utilization percentage (0–100).
+
+If no GPU is detected, `vram` is an empty array. Fields may be absent on platforms where the metric is unavailable.
+
+## `POST /v1/log-level`
+<sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
+
+Change the server's log verbosity at runtime without restarting. The new level applies immediately to all subsequent log output.
+
+### Parameters
+
+| Parameter | Required | Description |
+|-----------|----------|-------------|
+| `level` | Yes | Log level string: `trace`, `debug`, `info`, `warning`, `error`, `fatal`, or `none`. |
+
+### Example request
+
+```bash
+curl -X POST http://localhost:13305/v1/log-level \
+  -H "Content-Type: application/json" \
+  -d '{"level": "debug"}'
+```
+
+### Response format
+
+```json
+{"status": "success", "level": "debug"}
+```
+
+> **Note:** This endpoint requires `LEMONADE_ADMIN_API_KEY` when an admin key is configured. See [API Key and Security](../guide/configuration/README.md#api-key-and-security).
+
+## `POST /v1/params`
+<sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
+
+Read or update server configuration values at runtime. Changes are applied immediately and persisted to `config.json` without restarting the server. This is the HTTP equivalent of `lemonade config set`.
+
+> **Note:** This endpoint requires `LEMONADE_ADMIN_API_KEY` when an admin key is configured.
+
+### Read current config
+
+Send an empty body (or `{}`) to retrieve the full runtime configuration snapshot.
+
+```bash
+curl -X POST http://localhost:13305/v1/params \
+  -H "Content-Type: application/json" \
+  -d '{}'
+```
+
+### Update config values
+
+Send a JSON object with one or more keys to update them atomically.
+
+```bash
+curl -X POST http://localhost:13305/v1/params \
+  -H "Content-Type: application/json" \
+  -d '{"ctx_size": 8192, "max_loaded_models": 3}'
+```
+
+### Supported keys
+
+**Server-level keys** (take effect immediately):
+
+| Key | Type | Side effect |
+|-----|------|-------------|
+| `port` | int (1–65535) | HTTP server rebinds to the new port |
+| `host` | string | HTTP server rebinds to the new address |
+| `log_level` | string | Log filter reconfigured (same as `POST /v1/log-level`) |
+| `global_timeout` | int | Default HTTP client timeout updated |
+| `no_broadcast` | bool | UDP beacon stopped or started |
+| `extra_models_dir` | string | Model manager search path updated |
+
+**Deferred keys** (take effect on the next model load):
+
+| Key | Type |
+|-----|------|
+| `max_loaded_models` | int (-1 or positive) |
+| `ctx_size` | int |
+| `llamacpp_backend` | string |
+| `llamacpp_args` | string |
+| `sdcpp_backend` | string |
+| `sdcpp_args` | string |
+| `whispercpp_backend` | string |
+| `whispercpp_args` | string |
+| `vllm_backend` | string |
+| `vllm_args` | string |
+| `flm_args` | string |
+| `steps` | int |
+| `cfg_scale` | number |
+| `width` | int |
+| `height` | int |
+
+### Response format
+
+```json
+{"status": "success", "updated": {"ctx_size": 8192, "max_loaded_models": 3}}
+```
+
+On validation failure, the server returns `400` with an `error` field describing which key failed.
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -21,7 +21,6 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/system-stats`](#get-v1system-stats) | Current host resource usage |
 | `GET` | [`/v1/system-info`](#get-v1system-info) | System information and device enumeration |
 | `POST` | [`/v1/log-level`](#post-v1log-level) | Change the server log verbosity at runtime |
-| `POST` | [`/v1/params`](#post-v1params) | Read and update server configuration at runtime |
 | `POST` | [`/v1/install`](#post-v1install) | Install or update a backend |
 | `POST` | [`/v1/uninstall`](#post-v1uninstall) | Remove a backend |
 | `POST` | [`/v1/cloud/auth`](#post-v1cloudauth) | Set an in-memory API key for a cloud provider |
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -20,11 +20,10 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/stats`](#get-v1stats) | Performance statistics from the last request |
 | `GET` | [`/v1/system-stats`](#get-v1system-stats) | Current host resource usage |
 | `GET` | [`/v1/system-info`](#get-v1system-info) | System information and device enumeration |
-| `GET` | [`/v1/system-stats`](#get-v1system-stats) | Live CPU, RAM, and VRAM usage |
 | `POST` | [`/v1/log-level`](#post-v1log-level) | Change the server log verbosity at runtime |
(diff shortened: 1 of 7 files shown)


docs/api/lemonade.md
```diff
@@ -19,6 +19,9 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/health`](#get-v1health) | Check server status, such as models loaded |
 | `GET` | [`/v1/stats`](#get-v1stats) | Performance statistics from the last request |
 | `GET` | [`/v1/system-info`](#get-v1system-info) | System information and device enumeration |
+| `GET` | [`/v1/system-stats`](#get-v1system-stats) | Live CPU, RAM, and VRAM usage |
+| `POST` | [`/v1/log-level`](#post-v1log-level) | Change the server log verbosity at runtime |
+| `POST` | [`/v1/params`](#post-v1params) | Read and update server configuration at runtime |
```
jeremyfowers: Intentionally not documented; plan to remove

docs/api/lemonade.md
```diff
@@ -19,6 +19,9 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/health`](#get-v1health) | Check server status, such as models loaded |
 | `GET` | [`/v1/stats`](#get-v1stats) | Performance statistics from the last request |
 | `GET` | [`/v1/system-info`](#get-v1system-info) | System information and device enumeration |
+| `GET` | [`/v1/system-stats`](#get-v1system-stats) | Live CPU, RAM, and VRAM usage |
```
jeremyfowers: This should probably be moved /internal

docs/dev/lemonade-omni.md
```diff
@@ -84,3 +84,114 @@ Integrate OmniRouter into an existing agent by following the pattern in [`exampl
 4. If you want to pick models programmatically rather than rely on an omni model being loaded, query `GET /v1/models?show_all=true` and match the `labels` array against the "Needs a model with label" column above.
 
 The example script implements all four steps end-to-end against the `generate_image` and `text_to_speech` tools.
+
+## Managing collections via the API
```
jeremyfowers: Intentionally single-sourced in custom-models.md

docs/dev/building-apps.md
```diff
@@ -0,0 +1,331 @@
+# Building Apps with Lemonade
```
jeremyfowers, line 1: Yeah, something like this is probably needed. We need a whole "SDK" landing page tbh.

fl0rianr (changes requested):
Thanks for filling these docs gaps — the new app-dev guide is good, but I think this needs a small cleanup before merge:

- `/v1/log-level` is documented as requiring LEMONADE_ADMIN_API_KEY, but it is registered as a regular `/v1/*` route. With only LEMONADE_ADMIN_API_KEY configured, regular API routes are still unauthenticated, so this doc currently overstates the protection. Either move this to `/internal/*` or document the actual auth behavior. I think Jeremy's comment was not addresed.
- The CLI docs say `lemonade config` uses `POST /v1/params`, but the CLI calls `/internal/set`.

kenvandine:

docs/api/lemonade.md
kenvandine: Removed in a previous commit — those sections are no longer in the docs.

docs/api/lemonade.md
kenvandine: The endpoint is registered as a regular `/v1/*` route in `server.cpp` (line 521), so moving it to `/internal` would require a server change beyond this PR's scope. I've corrected the documentation instead — the auth note now accurately states that `/v1/log-level` is protected by `LEMONADE_API_KEY` (the standard API key), not the admin key.

docs/dev/lemonade-omni.md
kenvandine: Addressed — the Custom Omni Models section now redirects to `custom-models.md` rather than duplicating the content.

docs/dev/building-apps.md
kenvandine: Agreed — a proper SDK landing page would be valuable. Happy to contribute to that in a follow-up PR once this lands.

kenvandine:
@fl0rianr Both issues are addressed in the latest commit:

- **`/v1/log-level` auth**: The note previously overstated the protection by saying it required `LEMONADE_ADMIN_API_KEY`. Since the endpoint is registered as a regular `/v1/*` route in `server.cpp` (not under `/internal/`), it is actually protected by the standard `LEMONADE_API_KEY`. The note now documents the actual behavior correctly.
- **`lemonade config` endpoint**: Fixed — the note now references `POST /internal/set`, which is what the CLI actually calls (`src/cpp/cli/main.cpp:909`), rather than the incorrect `POST /v1/params`.

fl0rianr (commented):
Thanks for addressing the earlier two points. I found two remaining docs-accuracy issues before I can approve:

- docs/dev/building-apps.md still lists /v1/log-level under LEMONADE_ADMIN_API_KEY elevated access. That conflicts with the corrected docs/api/lemonade.md note and the server auth behavior: /v1/log-level is a regular /v1/* API route, while admin auth is for /internal/*. Please remove /v1/log-level from the admin-key example or rephrase it as a regular API route.
- docs/api/anthropic.md documents "thinking" as ignored, but the Anthropic converter maps thinking.type: enabled|disabled to "enable_thinking". Please document the supported subset or remove the row.

kenvandine:
> Thanks for addressing the earlier two points. I found two remaining docs-accuracy issues before I can approve:
> 
> - docs/dev/building-apps.md still lists /v1/log-level under LEMONADE_ADMIN_API_KEY elevated access. That conflicts with the corrected docs/api/lemonade.md note and the server auth behavior: /v1/log-level is a regular /v1/* API route, while admin auth is for /internal/*. Please remove /v1/log-level from the admin-key example or rephrase it as a regular API route.
> - docs/api/anthropic.md documents "thinking" as ignored, but the Anthropic converter maps thinking.type: enabled|disabled to "enable_thinking". Please document the supported subset or remove the row.

Both points addressed in the latest commit:
docs/dev/building-apps.md: Removed /v1/log-level from the LEMONADE_ADMIN_API_KEY description — it's a standard /v1/* route, so only /internal/* is listed as requiring the admin key now.
docs/api/anthropic.md: Updated the thinking row to accurately reflect the implementation: thinking.type: "enabled" maps to enable_thinking: true, "disabled" maps to enable_thinking: false, and other type values are ignored with a warning. Badge changed from not_available to partial.

fl0rianr (approved):
Fine for me now.

kenvandine:

docs/api/lemonade.md
```diff
@@ -1371,3 +1375,30 @@ curl http://localhost:13305/live
 ```json
 {"status":"ok"}
 ```
+
+## `POST /v1/log-level`
```
jeremyfowers, line 1379: Need to remove this API, let's not put it in the docs.

docs/api/lemonade.md
```diff
@@ -20,8 +20,9 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/stats`](#get-v1stats) | Performance statistics from the last request |
 | `GET` | [`/v1/system-stats`](#get-v1system-stats) | Current host resource usage |
 | `GET` | [`/v1/system-info`](#get-v1system-info) | System information and device enumeration |
-| `POST` | [`/v1/install`](#post-v1install) | Install or update a backend, or register a cloud provider |
-| `POST` | [`/v1/uninstall`](#post-v1uninstall) | Remove a backend or cloud provider |
+| `POST` | [`/v1/log-level`](#post-v1log-level) | Change the server log verbosity at runtime |
```
jeremyfowers, line 23: Need to remove this API, let's not put it in the docs.

docs/api/openai.md
```diff
@@ -39,6 +39,8 @@ Chat Completions API. You provide a list of messages and receive a completion. T
...
 | `max_completion_tokens` | No | An upper bound for the number of tokens that can be generated for a completion. Mutually exclusive with `max_tokens`. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `enable_thinking` | No | Lemonade extension. Boolean. Set to `false` to suppress chain-of-thought reasoning on models that default to thinking mode (e.g. Qwen3). When `false`, Lemonade prepends `/no_think` to the last user message before forwarding the request. Takes precedence over `thinking` when both are present. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
+| `thinking` | No | Lemonade extension. Boolean or object `{"type": "enabled" \| "disabled"}`. Equivalent to `enable_thinking` for clients that use the object form (e.g. OpenCode). Set to `false` or `{"type": "disabled"}` to suppress thinking. Ignored if `enable_thinking` is also present. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
```
jeremyfowers, line 43: These only work with Qwen3 models afaik, we may want to remove them instead of documenting.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
```
jeremyfowers, line 1: New docs need to be added to mkdocs nav.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Discover the server on the local network
+
+If you don't know which machine is running the server, `lemond` broadcasts a UDP beacon on port 13305. The CLI command `lemonade scan` listens for these beacons. From your own code:
+
+1. Open a UDP socket and listen on port 13305.
+2. Each beacon is a JSON payload with `service`, `hostname`, and `url` fields.
+3. Use `url` as your base URL.
+
+For remote servers you control, set the host explicitly via `lemonade config set host=0.0.0.0` on the server and point your client at the server's IP.
```
jeremyfowers, line 54: I don't know who "you" is supposed to be here. I thought "you" was the app developer, but app developers don't control the servers of their users?

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
+Each model object includes:
+
+| Field | Description |
+|-------|-------------|
+| `id` | Model name used in inference requests |
+| `labels` | Array of capability labels (e.g. `["reasoning", "tool-calling"]`) |
+| `downloaded` | `true` if the model files are present on disk |
+| `recipe` | Backend recipe (e.g. `llamacpp`, `flm`, `sd-cpp`) |
+
+By default, the listing hides collections and some internal entries. Pass `?show_all=true` to see everything:
```
jeremyfowers, line 79: Non-downloaded models are also hidden by default, I don't think there's clear here.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
+```python
+all_models = requests.get(f"{BASE_URL}/v1/models?show_all=true").json()["data"]
+```
+
+### Filter by capability
+
+Use the `labels` array to find models for a specific task:
+
+```python
+llm_models    = [m for m in models if not any(l in m.get("labels", [])
```
jeremyfowers, line 90: We need an LLM label don't we...

docs/api/openai.md
```diff
@@ -113,6 +115,18 @@ Chat Completions API. You provide a list of messages and receive a completion. T
     }
     ```
 
+### Streaming
+
+When `stream: true`, the server returns a sequence of `text/event-stream` chunks. Each chunk is a `data:` line containing a JSON object with `object: "chat.completion.chunk"`. The stream ends with a `data: [DONE]` sentinel.
+
+**Client-side guidance:**
+
+- Use the [OpenAI Python or JavaScript SDK](https://platform.openai.com/docs/libraries) — they handle SSE parsing, reconnection, and the `[DONE]` sentinel automatically.
```
jeremyfowers, line 124: Proposed changes contain 13 em dashes.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
+# Building Apps with Lemonade
+
+This guide is for developers building applications that talk to a Lemonade server over HTTP — web apps, desktop clients, agents, and background services. It covers the connection pattern, model discovery, inference, streaming, error handling, and security.
+
+If you want to **embed** a Lemonade server inside your own installer or process, see the [Embeddable Lemonade](../embeddable/README.md) guide instead.
```
jeremyfowers, line 5: Not sure I agree with this guidance. Only the Connection Pattern section is unique to non-embedded Lemonade, the rest applies equally to embedded and non-embedded.

docs/dev/building-apps.md
jeremyfowers, line 1: FYI... probably out of scope for this PR but needs a cleanup: the `docs/dev` folder is for guidance on developing *this* project.

We probably need a `docs/sdk` folder that has this document, the embedded SDK, and the omni doc in it.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
+
+1. Open a UDP socket and listen on port 13305.
+2. Each beacon is a JSON payload with `service`, `hostname`, and `url` fields.
+3. Use `url` as your base URL.
+
+For remote servers you control, set the host explicitly via `lemonade config set host=0.0.0.0` on the server and point your client at the server's IP.
+
+---
+
+## Discover models
```
jeremyfowers, line 58: Worth a link to the custom_models.md somewhere in this section so people know how to deal with models outside our registry.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
+
+For models that default to extended chain-of-thought reasoning (e.g. Qwen3), you can disable thinking for faster, shorter responses:
+
+```python
+response = client.chat.completions.create(
+    model="Qwen3-0.6B-GGUF",
+    messages=[{"role": "user", "content": "What is 2 + 2?"}],
+    extra_body={"enable_thinking": False},
+)
+```
```
jeremyfowers, line 172: Similar comment to on openai.md, this doesn't work in the general case AFAIK so I don't think its a good idea to suggest it here.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
+
+**Context exceeded:** When the prompt exceeds the model's context window, llama.cpp returns a `400` or truncates silently depending on the `--ctx-shift` flag. Load the model with a larger `ctx_size` via `POST /v1/load` if you need longer contexts.
+
+**Retry guidance:** Transient errors (network blips, cold-start timeout) are best retried with exponential backoff starting at 1 second, capped at 30 seconds, up to 5 attempts.
+
+---
+
+## Authentication
+
+By default, Lemonade runs with no authentication. Set `LEMONADE_API_KEY` on the server to require a bearer token:
```
jeremyfowers, line 287: This sentence doesn't have a noun of direct address, but it's the Lemonade user/admin who can set the API key. The app builder has no idea whether they have set the API key unless they see auth errors. App builders are required to respect auth.

docs/dev/building-apps.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Performance tips
+
+**Pre-load models at app startup.** Use `POST /v1/load` during your app's initialization phase so the first user request gets a warm model. Poll `/v1/health` to confirm the model is ready before showing the UI.
+
+**Use `max_loaded_models` to hold multiple models.** By default only one model per type can be loaded at a time. If your app uses an LLM and an embedding model together, set `max_loaded_models` to 2 or higher so switching between them doesn't evict and reload:
+
+```bash
+lemonade config set max_loaded_models=2
+```
```
jeremyfowers, line 319: `lemonade config` is not available to app builders, that is a user/admin control.

docs/dev/lemonade-omni.md
```diff
@@ -83,3 +83,59 @@ python examples/lemonade_tools.py "Say hello world out loud"
 You can build your own omni model from registered models — see [Register a custom Omni Model from the desktop app](../guide/configuration/custom-models.md#register-a-custom-omni-model-from-the-desktop-app) in the custom models guide. The planner LLM must carry the `tool-calling` label, and each modality must have a downloaded model whose `labels` include the matching entry from the [tools table](#available-tools).
 
 To distribute a custom omni model to other machines or via Hugging Face, see [Share a collection](../guide/configuration/custom-models.md#share-a-collection-export-import-and-hugging-face).
+
+## Component loading behavior
+
+When you load a collection (`POST /v1/load` or the first inference request), Lemonade loads all components eagerly — the LLM, image model, ASR model, and TTS model are all started before the first request returns. This ensures tool calls can be dispatched immediately once the collection is ready, at the cost of higher startup VRAM.
```
jeremyfowers, line 89:
```suggestion
When you load a collection (`POST /v1/load` or the first inference request), Lemonade loads all components eagerly — the LLM, image model, ASR model, and TTS model are all started before the first request returns.
```

Not sure we need to justify design choices like this.

docs/dev/lemonade-omni.md
```diff
@@ -83,3 +83,59 @@ python examples/lemonade_tools.py "Say hello world out loud"
...
+
+**LRU eviction and collections:** Each component occupies its own LRU slot within its model type (one LLM slot, one image slot, one ASR slot, one TTS slot). If another model of the same type is loaded while a component is in its slot, that component will be evicted. After eviction, the next tool call targeting that component will re-load it automatically before the request continues — adding latency. To avoid this, set `max_loaded_models` high enough to hold all components you intend to use concurrently.
```
jeremyfowers, line 91: > To avoid this, set `max_loaded_models` high enough to hold all components you intend to use concurrently.

Depending who the audience of this document is, it should be "advise your users to..." or "set your app's embedded Lemond to..."

docs/dev/lemonade-omni.md
```diff
@@ -83,3 +83,59 @@ python examples/lemonade_tools.py "Say hello world out loud"
...
+
+**Collections do not hold model-type slots** — only their individual components do. Deleting the collection entry does not evict its components from memory.
```
jeremyfowers, line 93: This doesn't seem like loading behavior to me.

docs/dev/lemonade-omni.md
```diff
@@ -83,3 +83,59 @@ python examples/lemonade_tools.py "Say hello world out loud"
...
+
+## Chat-transcription models
```
jeremyfowers, line 95: This seems like it's in the wrong document. Maybe it belongs in building-apps.md instead?

docs/dev/README.md
```diff
@@ -2,11 +2,32 @@
...
+| Example | What it shows |
+|---------|--------------|
+| `lemonade_tools.py` | Full OmniRouter agentic loop (tool definitions, LLM call, tool dispatch) |
+| `realtime_transcription.py` | WebSocket realtime audio transcription |
+| `api_image_generation.py` | Image generation via `/v1/images/generations` |
+| `api_image_edits.py` | Image editing via `/v1/images/edits` |
+| `api_image_variations.py` | Image variations via `/v1/images/variations` |
+| `api_text_to_speech.py` | TTS via `/v1/audio/speech` |
+| `multi-model-tester.html` | Browser demo: test prompts across multiple models |
+| `llm-debate.html` | Browser demo: LLM debate arena |
```
jeremyfowers, line 26: It would be nice to provide links to the files.

docs/dev/README.md
```diff
@@ -2,11 +2,32 @@
 
 Lemonade is a community-driven project organized around the [Lemonade Discord server](https://discord.gg/5xXzkMu8Zk). That should be your first stop to meet the developers, get support, and propose contributions.
 
-This documentation is here to help you set up your environment, start modifying Lemonade to your liking, and join the developer community!
+This documentation covers two audiences: developers building apps that consume the Lemonade API, and contributors who want to modify Lemonade itself.
```
jeremyfowers, line 5: This is confusing to me, they are really separate audiences. See earlier comment about needing an SDK section.

PR #2058, opened 2026-06-01
fix: avoid HF user model name collisions act only when they occur

### LINKED ISSUE #2053: Lemonade does not distinguish between same name models in different namespace
### Platform

Linux/Other (please specify)

### Lemonade Version

10.6.0

### GPU / APU Model

AMD Ryzen 7735U

### Component

Other / Not sure

### Bug Description

Downloading two different models from huggingface with the same name, but in a different namespace, makes it so only the latest downloaded one is available, as seen in server logs when loading the model.

Redownloading the other not-visible one is instant, as it is still there on disk.
Doing so now makes that one available instead for use in lemonade, but of course hides the one that previously was available.

### Steps to Reproduce

1. `lemonade pull LiquidAI/LFM2.5-8B-A1B-GGUF:Q8_0`
2. `lemonade pull unsloth/LFM2.5-8B-A1B-GGUF:Q8_0`
3. `$ ls -d .cache/huggingface/hub/*LFM*`
`.cache/huggingface/hub/models--LiquidAI--LFM2.5-8B-A1B-GGUF/  .cache/huggingface/hub/models--unsloth--LFM2.5-8B-A1B-GGUF/`
=> two results for `ls -d`
4. `$ lemonade list --downloaded | grep -i lfm`
`LFM2.5-8B-A1B-GGUF-Q8_0                 Yes         llamacpp`
=> only one single result for `lemonade list`

Expected: two results for lemonade list, and both variants usable:
```
$ lemonade list --downloaded | grep -i lfm
LiquidAI/LFM2.5-8B-A1B-GGUF-Q8_0                 Yes         llamacpp
unsloth/LFM2.5-8B-A1B-GGUF-Q8_0                 Yes         llamacpp
```

### Log Output

```shell

```

### Additional Context

_No response_

fl0rianr:
## Summary

Fixes Hugging Face user model name collisions without changing the normal happy-path model names.

Previously, two HF pulls could resolve to the same `user.*` model name and the later pull could silently overwrite the existing `user_models.json` registration. This could make the first model or GGUF variant disappear from Lemonade, even if the files were still present in the Hugging Face cache.

This PR keeps the default short model name unchanged, e.g.

`LFM2.5-8B-A1B-GGUF-Q8_0`

and only falls back to an owner-qualified name when that default already exists with a different checkpoint. The owner is appended as a suffix so the family/model name remains at the front for UI grouping, e.g.

`LFM2.5-8B-A1B-GGUF-Q8_0`
`LFM2.5-8B-A1B-GGUF-Q8_0-unsloth`

The server also prevents silent re-registration of an existing `user.*` model with different checkpoint metadata. In that case, the caller must choose a different model name, or delete the existing user model before registering a replacement.

## Why

This keeps the happy path unchanged while fixing the real collision bug. It avoids breaking existing HF “Use this model” flows and prevents users from accidentally losing access to an already installed model or GGUF variant.

## Testing

* Verified that pulling/registering the same generated model name with a different checkpoint no longer silently overwrites the old registration.
* Verified that two HF repositories with the same leaf repo name are registered as separate user models.
* Verified that two GGUF variants from the same repository remain separately visible and usable.
* Verified that normal single-model pulls still use the short default name.

Closes #2053

### DIFF
--- a/src/cpp/server/hf_variants.cpp
+++ b/src/cpp/server/hf_variants.cpp
@@ -258,13 +258,15 @@ nlohmann::json fetch_pull_variants(const std::string& checkpoint, bool& not_foun
     }
     for (const auto& kv : file_sizes) total_repo_size += kv.second;
 
-    // Suggested name = component after the last '/'.
+    // Keep the Hugging Face namespace in the suggested name so repos with the
+    // same leaf name but different owners do not collide in user_models.json.
+    // Example:
+    //   LiquidAI/LFM2.5-8B-A1B-GGUF:Q8_0
+    //   unsloth/LFM2.5-8B-A1B-GGUF:Q8_0
+    // become distinct user model names instead of both becoming
+    // LFM2.5-8B-A1B-GGUF-Q8_0.
     std::string suggested_name = checkpoint;
-    {
-        size_t slash = checkpoint.find_last_of('/');
-        if (slash != std::string::npos) suggested_name = checkpoint.substr(slash + 1);
-    }
-
+    
     // ONNX RyzenAI branch: synthesize a 1-element variant set.
     if (!has_gguf && has_onnx && has_genai_config) {
         nlohmann::json vj;
--- a/src/cpp/server/hf_variants.cpp
+++ b/src/cpp/server/hf_variants.cpp
@@ -258,15 +258,13 @@ nlohmann::json fetch_pull_variants(const std::string& checkpoint, bool& not_foun
     }
     for (const auto& kv : file_sizes) total_repo_size += kv.second;
 
-    // Keep the Hugging Face namespace in the suggested name so repos with the
-    // same leaf name but different owners do not collide in user_models.json.
-    // Example:
-    //   LiquidAI/LFM2.5-8B-A1B-GGUF:Q8_0
-    //   unsloth/LFM2.5-8B-A1B-GGUF:Q8_0
-    // become distinct user model names instead of both becoming
-    // LFM2.5-8B-A1B-GGUF-Q8_0.
+    // Suggested name = component after the last '/'.
     std::string suggested_name = checkpoint;
-    
+    {
+        size_t slash = checkpoint.find_last_of('/');
+        if (slash != std::string::npos) suggested_name = checkpoint.substr(slash + 1);
+    }
+
     // ONNX RyzenAI branch: synthesize a 1-element variant set.
     if (!has_gguf && has_onnx && has_genai_config) {
         nlohmann::json vj;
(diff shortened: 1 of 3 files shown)


src/cpp/server/hf_variants.cpp
```diff
@@ -257,13 +257,15 @@ nlohmann::json fetch_pull_variants(const std::string& checkpoint, bool& not_foun
     }
```
jeremyfowers, line 258: This change seems to regress the happy path (see https://lemonade-server.ai/docs/dev/philosophy/#prioritize-the-happy-path)

Before: `lemonade pull LiquidAI/LFM2.5-8B-A1B-GGUF:Q8_0` would produce `LFM2.5-8B-A1B-GGUF-Q8_0`, which would match the naming convention of the built-in models and get pulled into the GUI category for LFM2.5 (well, it would if we had one, which we should).

After: the collision case is solved, but the happy path (where someone only downloads one GGUF and not a second nearly-identical one) is regressed.

Also, **blocking**, this is a breaking change that would impact things like the HF Use This Model menu:

Image

jeremyfowers (dismissed):
My suggestion for an alternative non-breaking solution is to check for naming collisions and prompt the user to pick a different name if and only if there is a collision.

fl0rianr:
Thanks, agreed — this change is too broad as-is.

I'll rework it so the happy path stays unchanged: the default model name should still use the repo leaf name, e.g. LFM2.5-8B-A1B-GGUF-Q8_0.

Instead of always including the HF namespace, I'll only disambiguate when there is an actual name collision with a different checkpoint. In that case we can offer/use a namespace-qualified fallback such as unsloth`-`LFM2.5-8B-A1B-GGUF-Q8_0, while still allowing the user to choose a custom name.

That would be a good but more complex fix. I hope it's ok if I prioritize it down a bit to bring it on a later point since the bug is real but maybe not the most critical one IMO.

fl0rianr:
@claude review

fl0rianr:
Thanks for the review. I want to clarify the intended behavior here: I do not think this PR needs an overwrite escape hatch.

For this user-model registration path, silently replacing an existing `user.*` entry with different checkpoint metadata is exactly the class of bug this PR is preventing. If a user wants another checkpoint under the same name, they should either:

* update the existing model through the intended update flow, or
* delete the existing user model and register the replacement.

Keeping implicit in-place re-registration blocked makes the behavior safer and more predictable. Adding a generic `overwrite`/`force` flag in this PR would broaden the scope and could reintroduce the accidental replacement behavior we are trying to eliminate.

I updated the PR description so it no longer claims that an explicit overwrite path exists.

fl0rianr:
@claude review again

fl0rianr:
I checked the error path: the server-side guard throws, the server-owned download job stores that exception in the download snapshot, `waitForServerDownloadTerminal()` turns an error snapshot into a thrown Error, and `handleDownloadModel()` catches that and surfaces it through `showError(...)`.

So the collision guard should be visible to the user as a failed download toast rather than failing silently.

bitgamma (approved):

PR #2059, opened 2026-06-01
test: give pull multi endpoint test a unique number

### LINKED ISSUE #2041: test_021_stats_endpoint is silently skipped — shadowed by duplicate test_021_pull_multi in server_endpoints.py
### Platform

Windows

### Lemonade Version

 Any

### GPU / APU Model

N/A (test infrastructure bug)

### Component

Other / Not sure

### Bug Description

Two test methods share the identical prefix test_021_ in server_endpoints.py:

test_021_stats_endpoint (line ~929)
test_021_pull_multi (line ~959)

### Steps to Reproduce

Open test/server_endpoints.py
Search for def test_021_ — two methods are returned
Run the test suite with verbose output: python -m pytest test/server_endpoints.py -v
Observe that test_021_stats_endpoint does not appear in the output — only test_021_pull_multi runs

Expected: Both test_021_stats_endpoint and test_021_pull_multi run independently.
Actual: test_021_stats_endpoint is silently dropped; the /stats endpoint is never tested.

### Log Output

```shell
N/A — no runtime error occurs; the test simply never runs.
```

### Additional Context

_No response_

fl0rianr:
Fixes #2041

Renames `test_021_pull_multi` to `test_021s_pull_multi` so the /stats endpoint test keeps a unique test number.

**Risk:**

Zero. Test-only change; no production code affected.

**Tested:**
```text
PYTHONPATH=test python -m unittest server_endpoints.EndpointTests.test_021_stats_endpoint server_endpoints.EndpointTests.test_021s_pull_multi -v
```

### DIFF
--- a/test/server_endpoints.py
+++ b/test/server_endpoints.py
@@ -956,7 +956,7 @@ class EndpointTests(ServerTestBase):
 
         print(f"[OK] /stats endpoint returned: {list(data.keys())}")
 
-    def test_022_pull_multi(self):
+    def test_021s_pull_multi(self):
         # First delete model if it exists to ensure we're actually testing pull
         delete_response = requests.post(
             f"{self.base_url}/delete",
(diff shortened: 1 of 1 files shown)


jeremyfowers (changes requested):
Wont this just collide with `test_022_install_backend_non_streaming` instead?

fl0rianr:
Good catch, that one shouldn't have happend! I would have put the test o the end of the 021 section, but the alphabetic order is messed up anyway so I kept it as-is rather then refactoring the whole file.

jeremyfowers (approved):

PR #2060, opened 2026-06-01
test: add concurrent chat completion coverage

### LINKED ISSUE #2042: Test suite is entirely sequential — no concurrent request test exists to catch race conditions or deadlocks in the backend
### Platform

Windows

### Lemonade Version

Any

### GPU / APU Model

Any (particularly relevant on multi-queue GPU/NPU backends)

### Component

Other / Not sure

### Bug Description

Every test in server_llm.py issues requests one at a time. There is no test that fires multiple simultaneous requests to the same loaded model. In production, concurrent requests are the most common access pattern and the most likely to expose race conditions, mutex deadlocks, or request-queue corruption in the C++ backend. A concurrency bug could be present and ship undetected because the test suite never exercises this path.

### Steps to Reproduce

Search test/server_llm.py for threading, ThreadPoolExecutor, or concurrent asyncio.gather calls across multiple independent requests
None exist
Introduce a race condition in the C++ request handler (e.g., a missing mutex around shared model state)
Run the full test suite — all tests pass

A test using threading.Thread or asyncio.gather fires at least 5 simultaneous chat completion requests to the same loaded model and asserts all 5 return valid, non-error responses.
Actual: No such test exists. Concurrent access is never exercised.

### Log Output

```shell
this is a test coverage gap, not a runtime crash. Once the test is added, reproducing a real race condition would require debug logs from lemonade config set log_level=debug during a concurrent load run.
```

### Additional Context

_No response_

fl0rianr:
## Summary
Adds a minimal concurrent chat completion test,
closes #2042 

The test sends 5 simultaneous non-streaming chat completion requests against an already loaded model and asserts that all requests return valid responses.

## Testing
- `python -m py_compile test/server_llm.py`
- `python test/server_llm.py --wrapped-server <...> --backend <...>`
- Verified `test_001b_chat_completions_concurrent_requests` passes locally.

### DIFF
--- a/test/server_llm.py
+++ b/test/server_llm.py
@@ -24,6 +24,8 @@ import os
 import time
 import requests
 import numpy as np
+from concurrent.futures import ThreadPoolExecutor, as_completed
+import threading
 
 from utils.server_base import (
     ServerTestBase,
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@fl0rianr considering our ci system is quite slow these days, how are you feeling about this PR? please either close it or request a reviewer.

fl0rianr:
I’d prefer to keep this test. If Lemonade aims to be enterprise-grade, handling concurrent requests reliably is baseline server behavior, especially for backends like vLLM. I think having explicit coverage for that is worth the small CI cost.

test/server_llm.py
```diff
@@ -24,6 +24,8 @@
 import time
 import requests
 import numpy as np
+from concurrent.futures import ThreadPoolExecutor, as_completed
```
abn: `as_completed`  is imported from `concurrent.futures` but is unused.

abn (changes requested):
Nice work on this! Using `threading.Barrier` to fire simultaneous requests alongside pre-loading the model via `POST /load` keeps the test clean and reliable.

One minor cleanup issue, otherwise LGTM.

test/server_llm.py
fl0rianr: Good catch, thanks.

abn (approved):
Apologies for the delay in looking at this. LGTM.

PR #2061, opened 2026-06-01
test: cover embeddings missing model error

### LINKED ISSUE #2040: test_015_embeddings_single_string never tests missing model= — embeddings error path has zero coverage
### Platform

Windows

### Lemonade Version

any 

### GPU / APU Model

test infrastructure bug

### Component

Other / Not sure

### Bug Description

test_015_embeddings_single_string always passes model= to client.embeddings.create(). There is no test that omits model= and asserts the server returns a 400 error with a helpful message. This means the error-handling path for malformed embeddings requests is completely untested — a regression could be introduced and the entire test suite would still pass green.

### Steps to Reproduce

Open test/server_llm.py and find test_015_embeddings_single_string
Note that model=model is always supplied
Run the full test suite — all tests pass
Remove the server-side validation that returns 400 when model= is missing
Run the test suite again — all tests still pass

Expected: A dedicated test sends an embeddings request without model= and asserts a 400 response with a descriptive error message.
Actual: No such test exists. The missing-model= error path has zero test coverage.

### Log Output

```shell
N/A — this is a test coverage gap, not a runtime crash.
```

### Additional Context

_No response_

fl0rianr:
## Summary
- add coverage for embeddings requests without model
- restore HTTP 400 for the structured invalid_request returned by the router
- map Lemon error types to the appropriate HTTP status in the shared error response helper

**Updated:**
The original test coverage exposed a regression on current main: the router correctly returns a structured invalid_request when model is missing, but handle_embeddings was returning that error body without propagating the corresponding HTTP status.

closes #2040 

## Testing

Before the server fix, the regression test fails as expected:

```text
test_015b_embeddings_missing_model_returns_400
AssertionError: 200 != 400
```

After applying the server fix and rebuilding/restarting the server:

```text
test_015b_embeddings_missing_model_returns_400 ... ok
```

The test also verifies that the response contains:

```json
{
  "error": {
    "type": "invalid_request",
    "message": "...No model specified..."
  }
}
```

This negative-path test does not require a model load or inference.

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -132,6 +132,20 @@ int get_error_status_code(const json& response, int default_status_code = 500) {
         }
     }
 
+    if (error.contains("type") && error["type"].is_string()) {
+        const std::string type = error["type"].get<std::string>();
+
+        if (type == ErrorType::INVALID_REQUEST ||
+            type == "invalid_request_error" ||
+            type == ErrorType::UNSUPPORTED_OPERATION) {
+            return 400;
+        }
+
+        if (type == ErrorType::MODEL_NOT_LOADED) {
+            return 404;
+        }
+    }
+
     return default_status_code;
 }
 
@@ -3451,6 +3465,11 @@ void Server::handle_embeddings(const httplib::Request& req, httplib::Response& r
 
         // Call router's embeddings method
         auto response = router_->embeddings(request_json);
+        if (response.contains("error")) {
+            set_error_response(response, res);
+            return;
+        }
+
         res.set_content(response.dump(), "application/json");
 
     } catch (const std::exception& e) {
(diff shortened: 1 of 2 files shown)


jeremyfowers:
@fl0rianr considering our ci system is quite slow these days, how are you feeling about this PR? please either close it or request a reviewer.

fl0rianr:
I’d definitely keep this one. After rebasing it onto current main, the test actually exposed a regression: the router returns the correct structured invalid_request for a missing model, but the embeddings handler wasn’t propagating it as HTTP 400.

It also stays very cheap CI coverage since this path doesn’t load a model or run inference. I’ve updated the PR to include the small server fix together with the regression test.

bitgamma (approved):
lgtm

PR #2062, opened 2026-06-01
feat: add CUDA backend support for stable-diffusion.cpp

Phqen1x:
## Summary

Adds CUDA backend support for `sd-cpp` (stable-diffusion.cpp) on Windows and Linux, matching the architecture-specific download pattern already used for the `llamacpp` CUDA backend.

- **Install**: selects architecture-specific binaries from `lemonade-sdk/stable-diffusion.cpp`, using `SystemInfo::get_cuda_arch()` to pick the right `sm_XX` asset (`.zip` on Windows, `.tar.xz` on Linux).
- **Auto-selection**: if no backend is specified at load time, `sd-cpp` now auto-selects the best supported backend via `SystemInfo::get_supported_backends()` rather than silently falling through to an empty string.
- **Device type**: `cuda` is added alongside `rocm`, `vulkan`, and `metal` as a GPU backend so device tracking is correct.
- **CUDA_VISIBLE_DEVICES**: at load time, restricts visible GPUs to those matching the selected compute architecture; respects any pre-existing override from the environment.
- **PATH (Windows)**: prepends the executable directory so bundled `cudart64_*.dll` / `cublas64_*.dll` are resolved before any system-wide CUDA install.
- **Registry**: adds the `cuda` backend definition to `system_info.cpp` (`RECIPE_DEFS`) and pins the initial version in `backend_versions.json` (`master-666-e3937d9`).

## Test plan

- [ ] Windows: pull and run an sd-cpp model with `--backend cuda`; verify the correct `sm_XX` zip is downloaded and inference runs on the NVIDIA GPU
- [ ] Linux: same with the `.tar.xz` artifact
- [ ] Verify `CUDA_VISIBLE_DEVICES` is set correctly when multiple NVIDIA GPUs with different architectures are present
- [ ] Verify pre-existing `CUDA_VISIBLE_DEVICES` override is respected
- [ ] Verify non-CUDA backends (rocm, vulkan, cpu, metal) are unaffected


src/cpp/server/system_info.cpp
```diff
@@ -513,16 +513,21 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"nvidia_gpu", {}},
     }},
 
-    // stable-diffusion.cpp - CPU backend (Windows/Linux x86_64)
-    {"sd-cpp", "cpu", {"windows", "linux"}, {
-        {"cpu", {"x86_64"}},
```
fl0rianr, line 518: 1. Revert the replacement of the CPU backend.
2. Put the CUDA backend above vulkan so cuda is chosen via auto for NVIDIA

src/cpp/server/backends/sd_server.cpp
fl0rianr, line 1: Lot's of CUDA environment Code in this file.

So can we avoid duplicating CUDA runtime environment logic here? After PR #2070, CUDA behavior now includes subtleties like not setting CUDA_VISIBLE_DEVICES on single-arch systems and setting `__NV_PRIME_RENDER_OFFLOAD=1` on Linux PRIME/Optimus systems. sd-cpp CUDA should reuse the same helper/path as llama.cpp CUDA instead of reimplementing part of it here, otherwise future CUDA fixes can easily apply to one backend but not the other.

fl0rianr (changes requested):
Thanks for adding sd-cpp CUDA support! But I think this needs a small cleanup before merge:

1. Please rebase/merge latest main so the CUDA fixes from PR #2070 are preserved.
2. Move sd-cpp:cuda before sd-cpp:vulkan in RECIPE_DEFS; otherwise CUDA is not actually preferred over Vulkan.
3. Avoid duplicating CUDA environment setup in SDServer::load(). Ideally factor shared CUDA subprocess env handling so both llama.cpp and sd-cpp get the same behavior for CUDA_VISIBLE_DEVICES and Linux PRIME/Optimus.

src/cpp/server/system_info.cpp
Phqen1x: I've updated the locations of both CPU and CUDA. I will push that commit soon. Thank you!

src/cpp/server/backends/sd_server.cpp
Phqen1x: I've applied these changes in the latest commit, let me know what you think.

fl0rianr:
@Phqen1x Does it work for you? For me not, Vulkan works, cuda on llama.cpp but wie sd cuda falls back on CPU.

Phqen1x:
@fl0rianr I made a few changes to the stable-diffusion cuda backend and it definitely works now! Give it a try again and see if it is still having issues please.

fl0rianr (approved):
@Phqen1x Thanks! Works now also on my system. time to generate is roughly 60% in comparison to vulkan, quite useful I would say.

PR #2064, opened 2026-06-01
fix: user model registry writes and stale cache lookups

fl0rianr:
## Summary

Fixes stale user-model registry handling by refreshing `user_models.json` on cache-miss lookups and by making registry read/modify/write paths safer.

## Changes

- Write user JSON files through a unique sibling temp file before replacing the target.
- Reload user_models.json from disk before mutating registered user models.
- Invalidate the model cache after registry mutations instead of incrementally updating stale cache state.
- Refresh user_models.json on user-model lookup misses before returning "Model not found".
- Use the refreshed `ModelInfo` canonical name when deleting models.


jeremyfowers (approved):
LGTM!

PR #2067, opened 2026-06-01
fix: user model RAM filtering

fl0rianr:
## Summary

Keep user-controlled models visible even when their declared size exceeds the automatic RAM filter.

## Changes

* Exempt `user.*`, `extra.*`, and `local_upload` models from the RAM size filter.


src/cpp/server/model_manager.cpp
```diff
@@ -2105,6 +2105,10 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
             continue;
         }
 
+        const bool user_controlled_model = is_user_model_name(name) ||
+                                         name.rfind("extra.", 0) == 0 ||
```
jeremyfowers: It would be nice to have a `is_extra_model_name(name)` helper similar to `is_user_model_name(name)` but I wont hold up the PR over it if you dont want to do it here.

src/cpp/server/model_manager.cpp
```diff
@@ -2105,6 +2105,10 @@ std::map<std::string, ModelInfo> ModelManager::filter_models_by_backend(
...
+                                         info.source == "local_upload";
```
jeremyfowers: local_upload might be completely deprecated / dead code... but again not really your problem here.

jeremyfowers:
the cpp change lgtm

the yml change is out of my depth, @superm1 can you review?

@fl0rianr if you want to split this up into 2 PRs (lol) I am happy to approve and merge the cpp file change.

fl0rianr:
@jeremyfowers Aye aye sir! Splitted and added your request. Thanks for the review!

jeremyfowers (approved):
LGTM thanks!

PR #2071, opened 2026-06-01
feat: server-side Omni orchestration for /chat/completions (Open WebUI)

jeremyfowers:
image

## What & why

Lemonade **Omni "collection" models** (`recipe: collection.omni`) bundle a chat LLM with image and speech models. Until now, all the multimodal orchestration — injecting the tool system prompt, running the tool-calling loop, embedding generated media — lived in the **desktop frontend**, so a plain OpenAI `/chat/completions` to a collection model was rejected and collections were hidden from `/v1/models`.

This ports that loop into the **C++ server** so any OpenAI-compatible client — **Open WebUI** in particular — can drive an Omni model with no plugin and no frontend.

## How it works

A new `CollectionOrchestrator` runs server-side when `/chat/completions` targets a collection model:

- Injects the reference tool system prompt (from `toolDefinitions.json`) and runs an internal tool-calling loop against the chat component.
- Executes `generate_image` / `edit_image` (router `image_generations`/`image_edits`) and `text_to_speech` (router `audio_speech`) by routing to the matching components.
- Embeds results into the assistant message as a markdown image data-URI and an `<audio>` data-URI that Open WebUI renders inline.
- Supports both **streaming (SSE)** and non-streaming.

**Two ways to use Omni models** (documented): the existing client-side OmniRouter path (bring your own loop, target a component LLM) and this new server-side path (target the collection name).

**Middleware semantics:** an app-provided system prompt is merged (omni instructions prepended); app `tools` are merged with omni tools — omni calls are resolved server-side, app/native tool calls are passed back as a `tool_calls` response to execute and resume (so Open WebUI's own tools keep working).

### Other changes
- Collections appear in `/v1/models` and Ollama `/api/tags` once their components are downloaded (they're genuine OpenAI-compatible chat models now).
- Collection component loading extracted into `Server::ensure_collection_loaded`, shared by `/load`, `auto_load_model_if_needed`, and the chat path (auto-load is now collection-aware).
- `toolDefinitions.json` is staged from the frontend tree into `resources/` as the single source of truth.

**Scope:** `generate_image`, `edit_image`, `text_to_speech`. `transcribe_audio` and `analyze_image` remain client-path tools (Open WebUI transcribes audio itself and passes images straight to the model).

## Verification

Validated end-to-end with real inference against both shipped collections:

- **LMX-Omni-52B-Halo** (Flux): image generation, **image editing** (confirmed img2img via `image_edits`), and TTS — plus multi-turn with image history (no context blow-up, no duplicate image) and streaming.
- **LMX-Omni-5.5B-Lite** (SD-Turbo, gen-only): generation + TTS; an "edit" request correctly falls back to a fresh generation since SD-Turbo lacks the `edit` label.
- Image renders as a valid PNG, audio as a valid MP3 with an `<audio>` player (verified against Open WebUI's actual `marked` 9.1.0 + HTMLToken rendering path).

Automated regression test: `test/server_omni.py` drives a `/chat/completions` conversation against `LMX-Omni-5.5B-Lite` and asserts server-side image generation and text-to-speech each return embedded media data-URIs (it pre-pulls the collection in `setUpClass` so inference never triggers a download). It runs in CI on Linux (.deb) and Windows (.exe) with the vulkan backend.

UI/frontend is untouched.


fl0rianr (changes requested):
Thanks, great PR— overall this looks solid. But I’d just like one compatibility issue fixed before approval:

In the streaming collection path, app/native tool calls are emitted as delta.tool_calls by passing through the non-streaming `message.tool_calls` objects directly. Streaming tool-call deltas need per-call `index` fields so OpenAI-compatible clients/SDKs can merge them correctly; without that, clients such as the OpenAI Python SDK reject or fail to reconstruct the returned app tool call. This affects the advertised behavior where Omni tools are handled server-side while client tools are passed back, specifically for the case stream: true.

Could you normalize streamed app tool calls to the OpenAI streaming shape, e.g. add index per call, and add a streaming variant of the mixed Omni + app-tool test?

Minor: collection_orchestrator.h`should include <map> directly (is that our code-standard?)

And finally the server-injected Omni prompt still mentions analyze_image/transcribe_audio even though those tools are intentionally not exposed server-side, which may confuse the planner for uploaded image/audio inputs.

jeremyfowers:
Thanks @fl0rianr for the review! I made the following updates:

1. Added a test for streaming external tool calls, and verified that it failed for the initial implementation.
2. Added OpenAI compliant support for external tool calls, verified the new test passed.
3. Added the requested include.
4. Changed toolDefinitions.json so that each tool can carry its own prompt guidelines, rather than hardcoding all guidelines into the prompt. The server and GUI are each responsible for assembling the system prompt based on the available tools. This way the GUI makes the analyze_image and transcribe_audio tools known to the prompt, while the server does not.

Rebuilt and manually tested the update in both the Lemonade App and Open WebUI.

fl0rianr:
@jeremyfowers thanks for the updates adressing the points.

There seems to be a blocker regarding image inputs in the collection path. The docs/PR text say `analyze_image` remains client-side and frontends pass images straight through to the model, but `run_loop()` replaces incoming `image_url` parts with a text placeholder (`[User provided image #N]`) and only saves the base64 as an edit source. That means an uploaded-image question like “what’s in this image?” against the collection model won’t actually pass the image to the planner; only edit-style flows can use it. The server should preserve/pass through image_url for a vision-capable planner path and add a regression test.

fl0rianr (changes requested):
Stated in the comment above.

jeremyfowers:
> @jeremyfowers thanks for the updates adressing the points.
> 
> There seems to be a blocker regarding image inputs in the collection path. The docs/PR text say `analyze_image` remains client-side and frontends pass images straight through to the model, but `run_loop()` replaces incoming `image_url` parts with a text placeholder (`[User provided image #N]`) and only saves the base64 as an edit source. That means an uploaded-image question like “what’s in this image?” against the collection model won’t actually pass the image to the planner; only edit-style flows can use it. The server should preserve/pass through image_url for a vision-capable planner path and add a regression test.

Reproduced, great catch!

image

jeremyfowers:
image

Fixed! Whew, we would have gotten in a lot of trouble if we released that bug.

fl0rianr:
Thanks for the follow-up fixes — the uploaded image passthrough looks much better now.

Please understand I'm extra critical here since we wanna land this right.
There is one important middleware/resume gap that is not covered by the tests:

We advertise that  app/native tools are returned to the client with `finish_reason: "tool_calls"`, the client executes them, and then resumes the conversation. However, the collection pre-processing rebuilds incoming history messages as only `{role, content}` in the string-content path. That drops `assistant.tool_calls` from the previous assistant turn and also drops `tool_call_id` from subsequent `role: "tool"` messages. Those fields are required for a standards-compatible tool-call resume. As a result, the first request can return the app tool call, but the second request after the client executes that tool is likely to reach the planner with a malformed/incomplete tool-call history.

Could we preserve the non-media fields when sanitizing messages instead of reconstructing only role/content? For example, start from the original message object and replace only `content` after stripping/transforming media, so `tool_calls`, `tool_call_id`, etc. survive.

Wouldn't it make sense to also add an regression test that exercises the full two-request flow, not just returning the tool call?

1. Request image generation + an app tool.
2. Assert the app tool is returned.
3. Append the assistant message with its tool_calls.
4. Append a `role: "tool"` message with matching `tool_call_id`.
5. Re-issue /chat/completions to the collection model.
6. Assert the planner accepts the resumed history and returns a final answer.

I’d test this for non-streaming at minimum; streaming would be ideal too.

fl0rianr (changes requested):
important middleware/resume gap (see above)

jeremyfowers:
@fl0rianr the resume gap should be addressed by 5aa4879: history sanitization now starts from the original message and replaces only `content`, so `tool_calls` / `tool_call_id` (and any other fields) survive pre-processing.

Also added the two-request resume regression tests you suggested — `test_006_app_tool_resume` (non-streaming) and `test_007_app_tool_resume_streaming` — covering: app tool returned → client executes → assistant turn with `tool_calls` + `role:"tool"` result echoed back → final answer incorporates the tool result.

One honest note: the new tests pass even against the pre-fix server, because Qwen's chat template tolerates an orphaned tool result (its text still arrives via `content`). So the dropped fields were a latent protocol bug rather than an observable failure on today's collections — the fix matters for standards compliance and for stricter chat templates (e.g. Llama-3-style, which hard-error on a tool message with no preceding tool call). The tests now lock in the full resume contract either way.

fl0rianr (approved):
Great work! Assuming the new test don't lead to an CI error I say - go for it!

PR #2072, opened 2026-06-01
ci: resolve deb runtime deps in extract action

fl0rianr:
## Summary

Make the extract-only .deb CI action resolve package runtime dependencies before running the staged binary.

## Changes

- Read runtime dependencies from the `.deb` Depends metadata.
- Install them when root or passwordless sudo is available.
- Otherwise download and locally extract missing runtime dependencies.

## Validation

- Scope limited to `.github/actions/install-lemonade-deb/action.yml`.


superm1:
What's the background of this?  All the hosted runners already have runtime deps installed.

fl0rianr:
Thanks, that makes sense. I think this PR is too broad for the current runner setup then, was generated dividing a bigger PR.

The only smaller thing I could reduce it to is an improved diagnostic for the no-root/no-passwordless-sudo path, so that if a self-hosted runner ever lacks a declared .deb runtime dependency, the failure points back to the package `Depends` metadata instead of looking like a generic missing shared library issue.

If that is not useful either, it would make sense to close this.

superm1:
Let's let @jeremyfowers decide on this one.  My opinion is it is unnecessary.

superm1 (commented):
NACK from me.   I don't see the need on self hosted runners.

Issue #2074, opened 2026-06-02
Loading a model on an AMD iGPU + NVIDIA GPU system with RoCM backend crashes llama.cpp


quanchobi:
### Platform

Linux/Ubuntu

### Lemonade Version

10.5.1-snap

### GPU / APU Model

Advanced Micro Devices, Inc. [AMD/ATI] Strix Halo [Radeon Graphics / Radeon 8050S Graphics / Radeon 8060S Graphics] (rev c1), NVIDIA Corporation GA102 [GeForce RTX 3090 Ti] (rev a1)

### Component

llama.cpp

### Bug Description

When attempting to load a model with the RoCM backend on an AMD iGPU with an NVIDIA GPU connected, llama.cpp attempts to allocate memory for the model on the NVIDIA card, causing llama.cpp to crash:
```
Jun 01 20:53:37 yappy-yiqi lemonade-server.daemon[38780]: 2026-06-01 20:53:37.529 [Info] (LlamaCpp) Loading model: user.Qwen3.6-35B-A3B-MTP-GGUF-BF16
Jun 01 20:53:37 yappy-yiqi lemonade-server.daemon[38780]: 2026-06-01 20:53:37.529 [Info] (LlamaCpp) Using LlamaCpp Backend: rocm-stable
Jun 01 20:53:37 yappy-yiqi lemonade-server.daemon[38780]: 2026-06-01 20:53:37.530 [Info] (LlamaCpp) Starting llama-server...
Jun 01 20:53:37 yappy-yiqi lemonade-server.daemon[38780]: 2026-06-01 20:53:37.530 [Info] (ProcessManager) Process started successfully, PID: 39269
Jun 01 20:53:37 yappy-yiqi lemonade-server.daemon[38780]: Waiting for llama-server to be ready (timeout: 600s)...
Jun 01 20:53:37 yappy-yiqi lemonade-server.daemon[38780]: 2026-06-01 20:53:37.651 [Info] (Process) 0.00.116.368 I log_info: verbosity = 3 (adjust with the `-lv N` CLI arg)
Jun 01 20:53:37 yappy-yiqi lemonade-server.daemon[38780]: 2026-06-01 20:53:37.651 [Info] (Process) 0.00.116.371 I device_info:
[...]

Issue #2043, opened 2026-05-29
Agenda Topics - Maintainers Meeting June 5, 2026


jeremyfowers:
Please comment on this issue to submit topics for the agenda. The agenda will be finalized by me the morning of Thursday, June 4, and sent out on the discord.

jeremyfowers:
Agenda topic: propose formal working groups with maintainer-leads to better organize the development process.

kenvandine:
Release cadence and milestone planning

fl0rianr:
Besides the very basic feature roadmap in the README, could we add a more detailed roadmap that separates what we need to get there, what would be nice to have, and what still needs to be fixed?

AI could also help visualize the roadmap and clarify dependencies, so we end up with a highly visual overview of where things are heading.

jeremyfowers:
Careful what you wish for @kenvandine, I will probably federate milestone planning to working group leads ;)

Truthfully this project's direction tends to be set by what people are passionate about, not a precise and/or monolithic long-term set of requirements to be fulfilled. I think the working groups can help us write down, distribute, and cross-reference what people are passionate about developing in a format that is much easier to share with the community.

So I will write down what I want to see for omni models, since that's what I'm passionate about right now. Maybe @kenvandine will write down the requirements for true cross-vendor support, @bitgamma can specify what is needed for a great auto-tuner, etc.

Curious what you and others would like to see for release cadence. I try to release once per week, expect when I travel (or there is some other disruption) in which case its every other week.

jeremyfowers:
https://github.com/lemonade-sdk/lemonade/pull/2079 organizes the roadmap.

PR #2076, opened 2026-06-02
Auto-label issues and PRs by backend and area

### LINKED ISSUE #2091: Add runtime / device label dimension to complement engine::* (follow-up to #2076)
Following up on @kenvandine's [proposal](https://github.com/lemonade-sdk/lemonade/pull/2076#issuecomment-) in PR #2076 to add labels for the underlying runtime / device so an issue can be tagged with both an engine and a runtime — e.g. `engine::llamacpp` + `<vulkan-label>` for Vulkan-specific llama.cpp bugs.

This issue exists to lock down the **naming** before any labels get created, because the project already uses "backend" with a specific meaning that collides with the obvious naming.

## Naming collision

- In `AGENTS.md` and the C++ code (`WrappedServer`), "backend" already means the inference engine itself — llama.cpp, FastFlowLM, RyzenAI, vLLM, whisper.cpp, stable-diffusion.cpp, Kokoro. That is the dimension PR #2076 calls `engine::*`.
- @kenvandine's suggestion uses `backend::cpu`, `backend::vulkan`, `backend::rocm`, `backend::cuda` — but those are really the *runtime* (Vulkan / ROCm / CUDA / Metal) or the *device* (CPU / GPU / NPU).
- There is also an existing `target::npu` label that overlaps with the device dimension.

## Options

1. **`runtime::vulkan`, `runtime::rocm`, `runtime::cuda`, `runtime::metal`, `runtime::cpu`** — cleanly separates from `engine::*`. Each engine can run on multiple runtimes.
2. **`device::cpu`, `device::gpu`, `device::npu`** + **`runtime::*`** — two new dimensions. More expressive, but more labels to maintain. Would also subsume / replace `target::npu`.
3. **Single `backend::*`** as proposed, and rename the existing `engine::*` to something else. High churn, not recommended.

## Decision needed from maintainers

- Which dimension(s) to add?
- What to do with the existing `target::npu` label — keep, merge into `device::npu`, deprecate?
- Set of values for each dimension (need to enumerate Metal, Vulkan, ROCm, CUDA, CPU at minimum; possibly DirectML / OpenCL / SYCL).

Once the taxonomy is agreed, the follow-up PR is small: create the labels, extend the prompt in `.github/scripts/auto_label.py` with the new dimension(s) and an "apply at most one" rule per dimension, and re-spot-test.

cc @kenvandine @jeremyfowers @fl0rianr

ramkrishna2910:
## Summary

`.github/workflows/auto-label.yml` + `.github/scripts/auto_label.py` (stdlib-only). Two passes per item:

1. **LLM classification** (Claude Haiku 4.5) over title + body for `engine::*`, `area::*`, `runtime::*`, component (`cpp` / `app` / `web ui` / `audio`), and type (`bug` / `enhancement` / `documentation` / `question`). Runs once when an item is opened. At-most-one rule enforced per family — `engine::`, `area::`, and `runtime::` — in both the prompt and the parser.
2. **Deterministic community-priority** (`priority::😎warm` / `priority::🔥hot`) based on distinct non-write-access engagement (author + commenters + supporting reactions). Maintainers excluded by both `author_association` and per-user collaborator-permission lookup. Refreshed on every new comment and via a daily cron to catch reaction-only changes.

Add-only for human / LLM labels — never removes a human-applied label. The bot may transition between its own priority labels (drops warm when promoting to hot) so the two never coexist.

The same script runs locally for prompt iteration (`python .github/scripts/auto_label.py <n>... --dry-run`) and in CI; no duplication.

## Triggers

| Trigger | Mode | Behavior |
|---|---|---|
| `issues.opened` | LLM + priority | Single-item label of the new issue |
| `pull_request_target.opened` | LLM + priority | Single-item label of the new PR |
| `issue_comment.created` | priority only | Refresh priority on the commented item (no LLM call) |
| `schedule` (06:00 UTC daily) | priority only | Sweep every open item; catches reaction-only changes |
| `workflow_dispatch` | choice (`all` / `priority-only`) | Manual: spot-test a single item or backfill the unlabeled |

`ANTHROPIC_API_KEY` is only injected on the LLM-mode step. `GITHUB_TOKEN` is scoped to `contents: read`, `issues: write`, `pull-requests: write`.

## Labels referenced by the workflow

All exist on the repo. The script will fail closed (`gh issue edit --add-label`) if any are missing, so this list must stay in sync.

**Created by this PR** (separate, ahead of the workflow shipping):

- **Engines (6 new, 1 pre-existing standardized):** `engine::llamacpp`, `engine::flm` *(pre-existing, description rewritten for consistency)*, `engine::ryzenai`, `engine::vllm`, `engine::whispercpp`, `engine::sd`, `engine::kokoro`
- **Areas (4 new):** `area::cli`, `area::installer`, `area::api`, `area::tray`
- **Runtimes (5 new — Option 1 from #2091, per @kenvandine):** `runtime::vulkan`, `runtime::rocm`, `runtime::cuda`, `runtime::metal`, `runtime::cpu`

**Pre-existing, reused as-is:**

- **Component:** `cpp`, `app`, `web ui`, `audio`
- **Type:** `bug`, `enhancement`, `documentation`, `question`
- **Priority:** `priority::😎warm`, `priority::🔥hot` *(the script does not reference `priority::❄️cold`)*

## Spot-test results (LLM pass)

`python .github/scripts/auto_label.py 2001 1968 1925 1911 1874 1942 1947 --dry-run`:

| # | Title | Decision |
|---|---|---|
| 2001 | failed to read modellist form fastflowlm | `engine::flm` |
| 1968 | Whisper fails to load on NPU | `engine::whispercpp` |
| 1925 | Kokoro koko b17 — ESPEAK_DATA_PATH | `engine::kokoro` |
| 1911 | vllm: TPS reporting still does not work | `engine::vllm, bug` |
| 1874 | Qwen-Image-2512-GGUF fails to load via sd-cpp | `engine::sd, bug` |
| 1942 | Tool calling hallucinated function names (NPU/Hybrid) | `engine::ryzenai, area::api` |
| 1947 | Fedora RPM `lemond` systemd service not enabled | `area::installer, documentation` |

Prompt was iterated twice during spot-testing:
- The model occasionally returned two `engine::` labels — now prevented by both prompt and the at-most-one parser cap (generalized later to also cover `area::` and `runtime::` families with a single `AT_MOST_ONE_PREFIXES` constant).
- The model was attaching `area::cli` / `audio` based on incidental mentions in repro steps — prompt now distinguishes "what the item is about" from "what it mentions".

## Spot-test results (priority pass)

| # | Engagement | Distinct non-write users | Decision |
|---|---|---|---|
| #2043 | 5 comments, all maintainers | 0 | (none) — verifies maintainer exclusion |
| #2082 | 1 contributor (`ckuethe`) | 1 | (none) |
| #2074 | 1 community user (`quanchobi`) | 1 | (none) |

Threshold paths verified by lowering `COMMUNITY_*_THRESHOLD` in-process; promotion logic verified across 6 scenarios including the both-labels-present cleanup path.

## Review hardening (commits subsequent to the initial design)

- `ANTHROPIC_API_KEY` moved from job-level `env` to the classifier step only
- `workflow_dispatch.inputs.target` validated as a positive integer before passing to the script
- Backfill replaced `--limit 500` with `gh api --paginate ...&per_page=100`
- Advisory-only header comment + this PR description state explicitly that labels here must not be used as CI / release / policy gates
- Priority labels are mutually exclusive — `community_priority_labels` returns `(to_add, to_remove, count)` and the bot drops `priority::😎warm` when promoting to `priority::🔥hot`
- `pull-requests: write` retained with an inline comment explaining why label ops on PRs require it even when `gh issue edit` is the surface
- `has_write_access` decorated with `@functools.lru_cache(maxsize=None)` so a scheduled sweep over many items only hits the `collaborators/{user}/permission` endpoint once per distinct user instead of once per item
- Priority is **sticky by design**: once `priority::😎warm` or `priority::🔥hot` is applied, it is never demoted or removed if engagement later drops. Promotion warm→hot is supported. Documented in the script docstring and workflow header

Closes #2091.

## Test plan

- [ ] Confirm `ANTHROPIC_API_KEY` repo secret is present (already used by `claude-review.yml`)
- [ ] After merge, dispatch a dry run against a few real recent issues: `gh workflow run auto-label.yml -f target=<num> -f dry_run=true`
- [ ] Re-dispatch one without dry run to confirm `gh issue edit --add-label` succeeds end-to-end
- [ ] Open a fresh issue/PR to validate the event-triggered LLM path
- [ ] Add a comment on an open issue to validate the priority-only `issue_comment` path
- [ ] Run a one-shot backfill: `gh workflow run auto-label.yml` (no inputs) to label all open items missing `engine::`/`area::`
- [ ] Wait for the next 06:00 UTC scheduled run to confirm the reaction-sweep path


fl0rianr (changes requested):
Thanks for putting this together — this looks like a very useful workflow.

I’d suggest a small hardening pass before merging, mainly because this runs on `pull_request_target` and uses both write permissions and an external API key. I don’t see this checking out the PR head, which is good, but the workflow still feeds untrusted issue/PR title/body text into the model and then applies any valid labels returned by the model. That seems okay for best-effort triage, but it is worth making sure labels cannot become a trust/security/release signal elsewhere.

Suggested changes:
- Reduce permissions if possible: `pull-requests: write` looks unnecessary for `gh issue edit --add-label`; `issues: write` plus maybe `pull-requests: read` should be enough.
- Move ANTHROPIC_API_KEY from job-level `env` to only the step that calls the classifier.
- Consider making PR-triggered runs dry-run only, or at least document that AI-applied labels are advisory and must not be used as policy/CI/release gates.
- Validate workflow_dispatch.inputs.target as numeric input before passing it into the script.
- Make the backfill path paginate through all open issues/PRs instead of relying on --limit 500. That is enough for today, but if the repo grows beyond that (especially issues), the workflow should still label all eligible open items reliably.
- If type labels are meant to be mutually exclusive, add the same “at most one” parser guard for bug / enhancement / documentation / question.

None of these change the overall design, which is already good, but they would make the workflow safer and easier to maintain once it starts running automatically on external issues/PRs.

fl0rianr:
Out of scope for this PR, but worth noting separately: this workflow classifies issues/PRs, it does not prioritize them. It won’t help decide which bugs should be fixed first among many open bugs etc. Might be better to do it elsewhere, just let's not forget about that if we already read through all the data via an AI.

jeremyfowers:
> Out of scope for this PR, but worth noting separately: this workflow classifies issues/PRs, it does not prioritize them. It won’t help decide which bugs should be fixed first among many open bugs etc. Might be better to do it elsewhere, just let's not forget about that if we already read through all the data via an AI.

I would love to see the “warm” label get auto-applied on any issue with 3+ people (author +2 others reactions or comments) and “hot” applied to issues with 6+ people.

fl0rianr:
> I would love to see the “warm” label get auto-applied on any issue with 3+ people (author +2 others reactions or comments) and “hot” applied to issues with 6+ people.

Or an AI-hot label as well.

kenvandine:
> > Out of scope for this PR, but worth noting separately: this workflow classifies issues/PRs, it does not prioritize them. It won’t help decide which bugs should be fixed first among many open bugs etc. Might be better to do it elsewhere, just let's not forget about that if we already read through all the data via an AI.
> 
> I would love to see the “warm” label get auto-applied on any issue with 3+ people (author +2 others reactions or comments) and “hot” applied to issues with 6+ people.

I really like the "warm" and "hot" label idea!

kenvandine:
I would propose adding some backends to KNOWN_LABELS, such as:

```
 - backend::cpu
 - backend::vulkan
 - backend::rocm
 - backend::cuda
```

Where an issue could potentially get tagged with engine::llamacpp and backend::vulkan for a vulkan specific issue to llamacpp.

fl0rianr:
`I would love to see the “warm” label get auto-applied on any issue with 3+ people (author +2 others reactions or comments) and “hot” applied to issues with 6+ people.`

I would exclude people with write access since then too many issues will get warm, if 3 people with write access are inside of an issue it's well enough known anyway. Think about one maintainer helps in an issue another one goes in as well.

Additionally, a common behavior to strengthen issues or PRs is to set a Smiley. If an issue says everything that's needed, no commend but 10 people have given a Smiley  it's definetly a very hot one.

.github/scripts/auto_label.py
```diff
@@ -0,0 +1,209 @@
...
+    "cpp",
+    "app",
+    "web ui",
+    "audio",
+    "bug",
+    "enhancement",
+    "documentation",
+    "question",
+}
+
```
fl0rianr, line 184: Could we keep prioritization separate from the LLM classification and make it deterministic?

This would cover the “warm” / “hot” discussion without letting the model invent priority labels:

- `warm`: 3+ distinct community users involved
- `hot`: 6+ distinct community users involved
- count the issue/PR author, commenters, and positive/interest reactions
- exclude users with write/maintain/admin access so maintainer discussion does not inflate demand
- keep `ai-hot` as a separate optional follow-up label if we later want model-scored urgency

Rough implementation sketch:

```python
COMMUNITY_WARM_THRESHOLD = 3
COMMUNITY_HOT_THRESHOLD = 6
WRITE_PERMISSIONS = {"admin", "write"}  # GitHub maps maintain -> write
WRITE_ASSOCIATIONS = {"OWNER", "MEMBER", "COLLABORATOR"}
SUPPORTING_REACTIONS = {"+1", "heart", "hooray", "rocket", "eyes", "laugh"}


def gh_api(path):
    return json.loads(
        run([
            "gh",
            "api",
            "-H",
            "Accept: application/vnd.github+json",
            path,
        ])
    )


def gh_api_pages(path):
    pages = json.loads(
        run([
            "gh",
            "api",
            "--paginate",
            "--slurp",
            "-H",
            "Accept: application/vnd.github+json",
            path,
        ])
    )
    return [item for page in pages for item in page]


def resolve_repo(repo):
    if repo:
        return repo
    return run([
        "gh",
        "repo",
        "view",
        "--json",
        "nameWithOwner",
        "--jq",
        ".nameWithOwner",
    ]).strip()


def has_write_access(login, repo, cache):
    if login in cache:
        return cache[login]

    try:
        data = gh_api(f"repos/{repo}/collaborators/{login}/permission")
    except subprocess.CalledProcessError:
        cache[login] = False
        return False

    cache[login] = data.get("permission") in WRITE_PERMISSIONS
    return cache[login]


def add_community_user(users, login, author_association, repo, permission_cache):
    if not login:
        return
    if author_association in WRITE_ASSOCIATIONS:
        return
    if has_write_access(login, repo, permission_cache):
        return
    users.add(login)


def community_priority_labels(item_num, existing, repo):
    issue = gh_api(f"repos/{repo}/issues/{item_num}")
    comments = gh_api_pages(f"repos/{repo}/issues/{item_num}/comments?per_page=100")
    reactions = gh_api_pages(f"repos/{repo}/issues/{item_num}/reactions?per_page=100")

    users = set()
    permission_cache = {}

    add_community_user(
        users,
        issue.get("user", {}).get("login"),
        issue.get("author_association"),
        repo,
        permission_cache,
    )

    for comment in comments:
        add_community_user(
            users,
            comment.get("user", {}).get("login"),
            comment.get("author_association"),
            repo,
            permission_cache,
        )

    for reaction in reactions:
        if reaction.get("content") not in SUPPORTING_REACTIONS:
            continue
        login = reaction.get("user", {}).get("login")
        if login and not has_write_access(login, repo, permission_cache):
            users.add(login)

    count = len(users)
    existing = set(existing)

    if count >= COMMUNITY_HOT_THRESHOLD and "hot" not in existing:
        return ["hot"]
    if count >= COMMUNITY_WARM_THRESHOLD and "warm" not in existing and "hot" not in existing:
        return ["warm"]
    return []

File: .github/scripts/auto_label.py [hunk 2]
```diff
@@ -0,0 +1,209 @@
...
+    return out
+
+
+def main():
+    p = argparse.ArgumentParser(description=__doc__.split("\n")[0])
+    p.add_argument("items", nargs="+", type=int, help="Issue or PR numbers")
+    p.add_argument("--dry-run", action="store_true", help="Print decisions; do not apply")
+    p.add_argument("--repo", help="OWNER/REPO; defaults to current repo")
+    args = p.parse_args()
+
```
fl0rianr, line 356:
```
repo = resolve_repo(args.repo)

for num in args.items:
    item = gh_view(num, args.repo)
    existing = [lbl["name"] for lbl in item.get("labels", [])]

    decision = classify(item, num)
    to_add = parse_decision(decision, existing)

    to_add.extend(
        community_priority_labels(num, existing + to_add, repo)
    )
```

fl0rianr (commented):
Thanks again for putting this together — I think this workflow will be very useful for keeping issues and PRs easier to triage.

I left a few inline draft comments, mostly around making the automation safer and easier to maintain before it starts running automatically:

- keep the LLM-based labels advisory and avoid using them as policy / CI / release signals
- reduce the privileged scope where possible, especially around `pull_request_target`, write permissions, and secret exposure
- validate manual workflow inputs and make the backfill path reliable beyond the current --limit 500
- keep prioritization separate from classification: **warm** / **hot** should be deterministic community-signal labels, not model-inferred labels
- when counting community interest, include comments and reactions, but exclude maintainers / users with write access so maintainer discussion does not inflate priority

None of this changes the overall direction of the PR — the design is already useful. These changes would just make the workflow safer, more predictable, and easier to extend later. Due to the extended scope form the comments I thought I could chime in here with a small  proposition.

ramkrishna2910:
Thanks for the careful review — really useful, and the priority sketch saved a bunch of design work. Latest commit (2a0909a5) addresses everything except where noted.

**Addressed:**

- ✅ **Step-scoped `ANTHROPIC_API_KEY`** — moved out of job env onto only the `Run auto-labeler` step.
- ✅ **Numeric validation of `inputs.target`** — guarded with `^[0-9]+$`; errors out cleanly otherwise.
- ✅ **Paginated backfill** — replaced `--limit 500` with `gh api --paginate /repos/.../issues?state=open&per_page=100`. Scales as the repo grows.
- ✅ **Advisory-only docs** — added a header comment to the workflow stating these labels must not gate CI / releases / policy.
- ✅ **Community-priority labels** — implemented per your sketch. `priority::😎warm` at ≥3 distinct non-write-access users (author + commenters + supporting reactions), `priority::🔥hot` at ≥6. Maintainers excluded via both `author_association` and per-user `collaborators/.../permission` lookup with caching. Spot-tested against three real items:

  | item | engagement | community count | label |
  | --- | --- | --- | --- |
  | #2043 | 5 comments, all maintainers | **0** | (none) ✓ verifies exclusion |
  | #2082 | 1 contributor (`ckuethe`) | **1** | (none) ✓ |
  | #2074 | 1 community user (`quanchobi`) | **1** | (none) ✓ |

  Threshold paths verified by lowering `COMMUNITY_*_THRESHOLD` in-process. Output now includes the community user count alongside the label decision for transparency.

**Pushing back gently on two:**

1. **Drop `pull-requests: write` → `read`** — I think this is required. Even though `gh issue edit --add-label` is the underlying command, `GITHUB_TOKEN`'s label-write scope is checked against the resource type — when the resource is a PR the check is against `pull_request`, not `issues`. I documented the reasoning inline in the workflow so the next reader doesn't retry the downgrade. Happy to be wrong here — if you've seen labels successfully applied to a PR from a `pull-requests: read` token, I'll drop it.
2. **At-most-one for type labels** (`bug` / `enhancement` / `documentation` / `question`) — these legitimately co-occur: a documentation bug is genuinely both `documentation` and `bug`; a feature with unresolved design questions is both `enhancement` and `question`. Hard-capping would force the model to drop the secondary label, which I think is the wrong default. Leaving these uncapped for now; happy to revisit if usage shows the model abusing it.

**Out of scope, tracked separately:**

@kenvandine's `backend::*` proposal collides with the existing use of "backend" in `AGENTS.md` (= engine), and there's already a `target::npu` label to reconcile. Opened #2091 to nail down the taxonomy (runtime vs device vs backend) before the follow-up PR creates the labels.

ramkrishna2910:
@kenvandine — went with your Option 1 from #2091 in 0cf01045. The 5 `runtime::*` labels are created on the repo (`vulkan`/`rocm`/`cuda`/`metal`/`cpu`), the prompt teaches the model when each applies, and the at-most-one rule is now generalized to cover `engine::`, `area::`, and `runtime::` families with one constant. A Vulkan-specific llama.cpp issue will now pick up both `engine::llamacpp` and `runtime::vulkan`.

fl0rianr (changes requested):
Thanks for the follow-up changes — this is moving in a good direction, and I like that prioritization is now deterministic rather than model-inferred.

I’d still hold off on merging until a few operational details are tightened up:

1. **Priority labels will not update after engagement changes.**  
   The workflow currently runs on issue/PR creation plus manual dispatch, but **warm** / **hot** are based on comments and reactions. That means an issue opened with only the author will usually never become warm or hot unless someone manually backfills it. Could we add an `issue_comment` trigger and/or a scheduled/manual priority refresh path? Reaction-only changes may still need a scheduled refresh.

2. **warm and hot should probably be mutually exclusive.**  
   Right now the script can add `priority::hot` to an item that already has `priority::warm`, but it never removes warm. I think we should either remove **warm** when promoting to **hot**

3. **Please make sure all newly introduced labels exist and are documented.**  
   The script now emits `runtime::*` and `priority::*` labels, but the PR description still only lists the created `engine::*` and `area::*` labels. If any of the runtime/priority labels do not already exist, `gh issue edit --add-label` will fail at runtime.

4. **Backfill scalability could become expensive.**  
   `community_priority_labels()` fetches issue details, all comments, all reactions, and then does collaborator-permission checks per distinct user per item. During a full backfill this can become a lot of API calls. A repo-wide permission cache, or at least a cache shared across all items in one run, would make this safer.

5. **Permissions are better than before, but still worth keeping minimal.**  
   I appreciate moving ANTHROPIC_API_KEY to the classifier step and documenting that labels are advisory. For `pull-requests: write`, I’d either verify it is truly required in this workflow or keep the explanatory comment, because this still runs under `pull_request_target`.

Overall, the design is useful, but I think the priority-label behavior needs one more pass before this runs automatically.

ramkrishna2910:
@fl0rianr  Addressed points 1 and 2 in b943d073. Leaving 3-5 separately so you can read them in isolation.

**Point 1 — engagement-driven priority refresh:**
- Added `issue_comment.created` trigger that re-runs the priority pass only (LLM stays single-fire at open-time so we don't burn tokens on every comment).
- Added a daily `schedule` cron at 06:00 UTC that sweeps every open item in priority-only mode. This covers the reaction-only-changes case you flagged — reactions don't fire webhook events so a comment-only trigger would miss them.
- Split the run step in two: `ANTHROPIC_API_KEY` is now only injected on the LLM path, never on priority-only runs.
- Skip-on-bot guard so the workflow doesn't self-trigger on its own comments (it doesn't post any today, but cheap insurance).
- `workflow_dispatch` gains a `mode` choice (`all` / `priority-only`) for manual priority refreshes against a specific item.

**Point 2 — warm/hot mutually exclusive:**
- `community_priority_labels` now returns `(to_add, to_remove, count)`. When an item crosses the hot threshold and already carries warm, warm is queued for removal in the same call.
- New `gh_remove_labels` helper. Scoped to bot-managed priority transitions only — explicit docstring + carve-out comment that it must never be called on human- or LLM-applied labels, so the overall "add-only" invariant still holds for everything else.
- Verified 6 promotion scenarios end-to-end including the cleanup path for items that somehow ended up with both labels.


4- Backfill would be fine, I am not worried about the cost, its a 1 time thing.

ramkrishna2910:
Addressed point 3. Verified all four label families exist on the repo:

| Family | Count | Source |
|---|---|---|
| `engine::` | 7 | 6 created by this PR + `engine::flm` pre-existing (description standardized) |
| `area::` | 4 | all created by this PR |
| `runtime::` | 5 | all created by this PR (commit 0cf01045, per @kenvandine's Option 1 on #2091) |
| `priority::` | 3 on the repo, 2 referenced | `priority::😎warm` and `priority::🔥hot` were already present; the script does **not** reference `priority::❄️cold` |

Rewrote the PR description to document the full label surface, the four trigger paths, the script's two-pass behavior (LLM + deterministic priority), and the spot-test results for both passes. If any label is ever renamed or removed the script will fail closed at `gh issue edit --add-label`, which is the loud-failure mode we want.

fl0rianr:
Thanks for the updates — this addresses the main concerns I had.

The split between LLM classification and deterministic priority now looks much cleaner: classification only runs on newly opened items, priority can refresh separately, reactions are covered by the scheduled sweep, and **warm / hot** no longer coexist when promoting to **hot**. I also appreciate the advisory-only wording, the narrower secret exposure, the input validation, and the expanded PR description/test plan.

I only have two small non-blocking follow-ups:

* The permission cache is still scoped per item, so a large scheduled/backfill run may repeat collaborator-permission lookups for the same users across many issues. Sharing that cache across the full run would reduce API calls.
* Priority labels currently promote **warm → hot**, but do not appear to demote/remove priority if engagement later drops below the threshold. That may be intentional to avoid churn, but it would be worth documenting the intended behavior.

Nothing here blocks the current PR from my side. The workflow is much safer and clearer now. Thanks so much!

ramkrishna2910:
Addressed both in 0b65c785.

**Permission cache shared across the run.** Decorated `has_write_access` with `@functools.lru_cache(maxsize=None)`. Dropped the local `cache = {}` from `community_priority_labels` and the `cache` parameter from `_add_community_user`. Verified with the same three test items — counts unchanged, and `cache_info()` after a 3-item run shows `hits=2, misses=2` rather than 4 fresh misses. For a real scheduled sweep over ~100 items where a handful of regulars / maintainers appear repeatedly, this collapses hundreds of duplicate lookups.

**Sticky priority documented.** You guessed right that it's intentional. Reasoning:
- Priority should reflect *peak* engagement — a topic that drew N distinct community users is a backlog signal even if those users have since moved on.
- Comments and reactions are rarely retracted in practice; supporting demotion would mostly only ever react to mass spam moderation, which is a manual situation anyway.
- Toggling labels as counts drift around the threshold (warm → hot → warm → ...) would be churn without signal.

Spelled out in three places: `community_priority_labels` docstring (longest version, with the rationale), the workflow header comment (one-line summary), and the PR description's review-hardening section.

fl0rianr (approved):
Thanks for the follow-up updates — this addresses the concerns I had. Nice!.

Looks good from my side. Approving!

PR #2079, opened 2026-06-02
Create a set of working groups

jeremyfowers:
This PR introduces the idea of Lemonade Working Groups, which define the project's roadmap and focus areas for contribution.

- Define the idea of a working group, lead, and charter
- Draft charters for the Omni Models and Cross Vendor working groups
    - NOTE: these drafts to not need to be finalized for this PR to merge, they can be finalized in a followup PR.
    - The purpose of this content is to make it clear what a working group charter should look like, so we can agree on a template using examples.
    - Working group leads are expected to make their own PRs to introduce/finalize their charters after this PR merges.
- Remove the roadmap table from the top-level README and instead direct people to the working groups.


kenvandine (approved):
I feel like we are growing up... so organized! I certainly approve of my role in the cross-vendor working group and am happy to own that. I think the rest of this looks great too.

Geramy (changes requested):

bitgamma (approved):

Geramy (approved):

PR #2081, opened 2026-06-02
feat: add Linux ARM64 CPU and Vulkan llamacpp backend support

kenvandine:
## Summary

- Extend `RECIPE_DEFS` in `system_info.cpp` to allow `arm64` CPU family for `llamacpp` `cpu`, `vulkan`, and `system` backends, so they appear as supported/installable on ARM64 Linux
- Download ARM64-specific binaries from upstream llama.cpp releases when compiled for `aarch64`:
  - `cpu`: `llama-{version}-bin-ubuntu-arm64.tar.gz`
  - `vulkan`: `llama-{version}-bin-ubuntu-vulkan-arm64.tar.gz`
- Fix `get_device_dict()` catch block to always set the CPU `family` field via compile-time macros — without this, an exception in `get_cpu_device()` left `family` missing from the JSON, causing backend matching to fail even after the `RECIPE_DEFS` change (manifested as "Requires ARM64 processors CPU" despite being on an ARM64 system)
- Update `docs/guide/configuration/llamacpp.md` and `README.md` to document ARM64 Linux support for `cpu` and `vulkan` backends

Validated against llama.cpp release assets at `b9253` and `b9482` — both ship `bin-ubuntu-arm64.tar.gz` and `bin-ubuntu-vulkan-arm64.tar.gz`. No version bump to `backend_versions.json` needed.

On ARM64 Linux (e.g., Qualcomm X Elite), `vulkan` is preferred over `cpu` by the existing `RECIPE_DEFS` preference order.

## Test plan

- [x] Restart `lemond` on an ARM64 Linux system and run `lemonade recipes` — `llamacpp:cpu` and `llamacpp:vulkan` should show as `installable`
- [x] `lemonade backends install llamacpp:vulkan` downloads `bin-ubuntu-vulkan-arm64.tar.gz` and runs inference
- [x] `lemonade backends install llamacpp:cpu` downloads `bin-ubuntu-arm64.tar.gz` and runs inference
- [x] Existing x86_64 Linux behavior unchanged (still downloads `x64` variants)
- [x] macOS and Windows builds unaffected


.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1507,6 +1544,71 @@ jobs:
...
+            echo "Running streaming error termination tests..."
+            $VENV_PYTHON test/server_streaming_errors.py --cli-binary "$CLI_BINARY"
+          fi
+          echo "${{ matrix.test_type }} tests PASSED!"
+
+      - name: Capture and upload server logs
+        if: always()
+        uses: ./.github/actions/capture-server-logs
+        with:
+          artifact-name: server-logs-arm64-${{ matrix.test_type }}
```
jeremyfowers, line 1663: I noticed this PR doesn't add a release artifact, is that intentional?

kenvandine: Yes, I was just getting it working with this PR and wanted to tackle any packaging we want separately.

src/cpp/server/system_info.cpp
```diff
@@ -873,11 +873,19 @@ json SystemInfo::get_device_dict() {
...
+        std::string cpu_family = "arm64";
+        #else
+        std::string cpu_family = "unknown";
+        #endif
         devices["cpu"] = {
             {"name", "Unknown"},
             {"cores", 0},
             {"threads", 0},
             {"available", true},  // Assume available - trust the user
+            {"family", cpu_family},
```
superm1, line 893: is family the right term?  I would think this is architecture

kenvandine: @superm1 you're right, but we use the term "architecture" differently internally, with things like target_arch, rocm_arch, etc. And data structures like "architectures," which are really about the GPU. But perhaps cpu_arch would be more appropriate?

superm1: how about `isa` then?

kenvandine: ```isa``` is certainly accurate, but I don't think people outside of the IHV world will recognize.

superm1: I mean family does not make sense. Family and model has a specific meaning already in the CPU world and it's not this.

kenvandine: I totally agree... just trying to tease out the best term to use here. My preference would be ```cpu_arch```, but I could be convinced to use ```isa```

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1507,6 +1544,71 @@ jobs:
...
+        env:
+          HF_HOME: ${{ env.HF_HOME }}
+        run: |
+          set -e
+          VENV_PYTHON=.venv/bin/python
+          CLI_BINARY=./build/lemonade
+
+          if [ "${{ matrix.test_type }}" = "cli" ]; then
+            echo "Running CLI tests..."
+            $VENV_PYTHON test/server_cli2.py --cli-binary "$CLI_BINARY"
```
fl0rianr, line 1646: I think this ARM64 test job needs to start `lemond` before running these tests.

This matrix downloads the raw build artifacts and sets CLI_BINARY=./build/lemonade, but the CLI/endpoint/Ollama tests expect a running server on port 13305. That seems to be why the ARM64 jobs currently fail with “Server is not running on port 13305”.

Could we add a startup step before this, similar to the unsigned macOS path: start ./build/lemond, set XDG_RUNTIME_DIR, wait for http://localhost:13305/live, and print the server log on timeout?

src/cpp/server/system_info.cpp
superm1: We could rename gpu_arch to gpu_isa and then use cpu_isa here too maybe

fl0rianr (commented):
Thanks for bringing this in @kenvandine! I guessed it might be welcomed if I take a look here as well...

Non-blocking: after the ARM64 server startup issue is fixed, it might be useful to add one small ARM64-specific smoke test for the core change in this PR.

The PR changes the llama.cpp asset names to bin-ubuntu-arm64.tar.gz / bin-ubuntu-vulkan-arm64.tar.gz, but this matrix currently only runs the generic CLI/endpoint/Ollama tests. At least `lemonade backends install llamacpp:cpu` should be testable on the ARM64 runner and would cover the new CPU asset path directly. Vulkan may be harder without GPU access.

src/cpp/server/system_info.cpp
kenvandine: Done — renamed `"family"` to `"cpu_isa"` for CPU, `"gpu_isa"` for GPU devices (AMD, NVIDIA, Metal), and `"npu_isa"` for NPU for consistency. All read/write sites in `system_info.cpp` and `model_manager.cpp` updated in 40deb128.

.github/workflows/cpp_server_build_test_release.yml
kenvandine: Fixed in 0c796bd0 — added a "Start lemond server" step between "Make binaries executable" and the Python venv setup. It sets `XDG_RUNTIME_DIR`, launches `./build/lemond` in the background, and polls `/live` for up to 60 seconds, dumping the server log on timeout.

superm1 (commented):
Doesn't the llama.cpp uprev job need to be changed too?

kenvandine:
No change needed to the uprev job. The ARM64 and x86 Linux builds come from the same upstream llama.cpp release tag — `llamacpp_server.cpp` selects the right archive name at compile time (`bin-ubuntu-arm64.tar.gz` vs `bin-ubuntu-x64.tar.gz`), but both pull from the same release. So when the uprev job bumps `llamacpp.cpu` and `llamacpp.vulkan` in `backend_versions.json`, the updated version applies to both architectures automatically.

The one gap is that the validate job only tests on Windows self-hosted runners and won't exercise the Linux ARM64 download paths, but that requires an ARM64 self-hosted runner and is out of scope here.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1507,6 +1544,92 @@ jobs:
...
+          exit 1
+
+      - name: Setup Python and virtual environment
+        uses: ./.github/actions/setup-venv
+        with:
+          venv-name: '.venv'
+          python-version: '3.10'
+          requirements-file: 'test/requirements.txt'
+
+      - name: Set HF_HOME
```
fl0rianr: Nice, the explicit server startup fixes the earlier ARM64 test issue.

One small follow-up: HF_HOME is currently set after lemond has already been started, so the server process will not inherit it. Should we move the `Set HF_HOME` step before `Start lemond server`? That would match the macOS unsigned test path more closely and keep server-side model/cache behavior under $PWD/hf-cache during these tests.

.github/workflows/cpp_server_build_test_release.yml
fl0rianr: Thanks, this addresses my original concern. I left one small follow-up about HF_HOME ordering, but the missing server startup issue itself looks fixed.

.github/workflows/cpp_server_build_test_release.yml
kenvandine: Good catch — moved `Set HF_HOME` to before `Start lemond server` so the server inherits it. Fixed in 5ef3a941.

src/cpp/server/system_info.cpp
kenvandine: @superm1 are you happy with the ```cpu_isa```, ```gpu_isa```, ```npu_isa``` approach?

fl0rianr:
From my side this is ready for merge, if the super fast tests are running successfully.

src/cpp/server/system_info.cpp
superm1: Yeah

jeremyfowers (changes requested):
Gentlemen, I understand that you found "family" to be unclear before, but changing it now is a breaking API change to `/v1/system-info`.

Please revert the breaking API change so that we can merge the ARM support.

kenvandine:
Done — reverted the `cpu_isa`/`gpu_isa`/`npu_isa` rename back to `family` in 9e524951. The ARM64 support changes are untouched.

jeremyfowers (approved):
LGTM! CI issue is just github being glitchy.

PR #2085, opened 2026-06-03
feat: add llama.cpp OpenVINO backend for Linux

kenvandine:
Adds support for the upstream ggml-org/llama.cpp OpenVINO backend on Linux. OpenVINO enables inference on Intel CPUs, iGPUs, dGPUs, and NPUs via Intel's optimization runtime.

- backend_versions.json: pin openvino to b9253 (same build as cpu/vulkan/metal)
- llamacpp_server.cpp: add is_llamacpp_openvino_backend() helper; handle openvino in get_install_params() (Linux-only, ggml-org/llama.cpp release asset llama-{ver}-bin-ubuntu-openvino-x64.tar.gz); enable context-shift and LD_LIBRARY_PATH setup for OpenVINO like CUDA/Vulkan
- system_info.cpp: register llamacpp/openvino in RECIPE_DEFS for Linux x86_64 (between Vulkan and ROCm in preference order)
- defaults.json: add openvino_args and openvino_bin defaults to llamacpp section
- config_file.cpp: add LEMONADE_LLAMACPP_OPENVINO_ARGS and LEMONADE_LLAMACPP_OPENVINO_BIN env variable mappings

https://claude.ai/code/session_01A3rK6yxjK9h7pe4ikrAvqj

### DIFF
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -54,6 +54,9 @@
     "cpu": "b10241",
     "openvino": "b10241"
   },
+  "openvino": {
+    "runtime_version": "2026.0"
+  },
   "whispercpp": {
     "cpu": "v1.8.4",
     "vulkan": "v1.8.4",
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -51,7 +51,8 @@
     "rocm-nightly": "b1305",
     "cuda": "b10236",
     "metal": "b10241",
-    "cpu": "b10241"
+    "cpu": "b10241",
+    "openvino": "b10241"
   },
   "whispercpp": {
     "cpu": "v1.8.4",
--- a/src/cpp/resources/defaults.json
+++ b/src/cpp/resources/defaults.json
@@ -29,6 +29,8 @@
     "cpu_args": "",
     "cpu_bin": "builtin",
     "cuda_bin": "builtin",
+    "openvino_args": "",
+    "openvino_bin": "builtin",
     "prefer_system": true,
     "rocm_args": "",
     "rocm_bin": "builtin",
--- a/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
@@ -159,6 +159,15 @@ static std::string get_therock_version() {
     return trim_to_major_minor(config["therock"]["version"].get<std::string>());
 }
 
+static std::string get_openvino_runtime_version() {
+    auto config = JsonUtils::load_from_file(utils::get_resource_path("resources/backend_versions.json"));
+    if (!config.contains("openvino") || !config["openvino"].is_object() ||
+        !config["openvino"].contains("runtime_version") || !config["openvino"]["runtime_version"].is_string()) {
+        throw std::runtime_error("backend_versions.json is missing 'openvino.runtime_version'");
+    }
+    return config["openvino"]["runtime_version"].get<std::string>();
+}
+
 InstallParams LlamaCppServer::get_install_params(const std::string& backend, const std::string& version) {
     InstallParams params;
 
@@ -236,7 +245,8 @@ InstallParams LlamaCppServer::get_install_params(const std::string& backend, con
     } else if (resolved_backend == "openvino") {
         params.repo = "ggml-org/llama.cpp";
 #ifdef __linux__
-        params.filename = "llama-" + version + "-bin-ubuntu-openvino-x64.tar.gz";
+        std::string openvino_ver = get_openvino_runtime_version();
+        params.filename = "llama-" + version + "-bin-ubuntu-openvino-" + openvino_ver + "-x64.tar.gz";
 #else
         throw std::runtime_error("OpenVINO llamacpp is currently supported on Linux only");
 #endif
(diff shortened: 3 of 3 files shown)


src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -211,6 +215,13 @@ InstallParams LlamaCppServer::get_install_params(const std::string& backend, con
         params.filename = "llama-" + version + "-bin-macos-arm64.tar.gz";
 #else
         throw std::runtime_error("Metal llamacpp only supported on macOS");
+#endif
+    } else if (resolved_backend == "openvino") {
+        params.repo = "ggml-org/llama.cpp";
+#ifdef __linux__
+        params.filename = "llama-" + version + "-bin-ubuntu-openvino-x64.tar.gz";
```
superm1: this doesn't look to match the structure of the upstream releases ?  I just looked at most recent one:

llama-b9488-bin-ubuntu-openvino-2026.0-x64.tar.gz

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -443,6 +454,19 @@ void LlamaCppServer::load(const std::string& model_name,
...
+        // The OpenVINO tarball bundles OpenVINO runtime libraries alongside
+        // llama-server, so add the executable's directory to LD_LIBRARY_PATH.
+        fs::path exe_dir = fs::path(executable).parent_path();
+        std::string lib_path = exe_dir.string();
+
+        const char* existing_ld_path = std::getenv("LD_LIBRARY_PATH");
+        if (existing_ld_path && strlen(existing_ld_path) > 0) {
+            lib_path = lib_path + ":" + std::string(existing_ld_path);
+        }
+
```
superm1, line 487: isn't this the exact same as the cuda path?  It looks like code duplication, just combine them?

src/cpp/server/config_file.cpp
```diff
@@ -200,12 +200,14 @@ static const EnvMapping env_mappings[] = {
     {"LEMONADE_LLAMACPP_ROCM_ARGS",      "llamacpp",  "rocm_args"},
     {"LEMONADE_LLAMACPP_VULKAN_ARGS",    "llamacpp",  "vulkan_args"},
     {"LEMONADE_LLAMACPP_CPU_ARGS",       "llamacpp",  "cpu_args"},
+    {"LEMONADE_LLAMACPP_OPENVINO_ARGS",  "llamacpp",  "openvino_args"},
```
superm1: isn't this migration code?  We didn't have oepnvino support before so how can you migrate?

src/cpp/server/backends/llamacpp_server.cpp
kenvandine: Good catch. Fixed — the filename now includes the OpenVINO runtime version: llama-{build}-bin-ubuntu-openvino-{runtime}-x64.tar.gz. The runtime version (2026.0) is pinned in a new top-level openvino.runtime_version key in backend_versions.json, following the same pattern as therock.version. Build updated to b9488 to match the first upstream release with this asset.

src/cpp/server/backends/llamacpp_server.cpp
kenvandine: Done — collapsed the CUDA and OpenVINO blocks into a single else if (is_llamacpp_cuda_backend(...) || is_llamacpp_openvino_backend(...)) branch.

src/cpp/server/config_file.cpp
kenvandine: These aren't migration-only mappings — migrate_from_env() is called on every fresh install to bootstrap config.json from env vars. All backends use the same mechanism (see LEMONADE_LLAMACPP_VULKAN_ARGS, LEMONADE_LLAMACPP_CUDA_BIN, etc.). Adding OpenVINO entries here means users who configure via env vars get them picked up on first run, consistent with the existing pattern.

superm1: I could have sworn there was a discussion somewhere about axing them.

superm1: @jfowers comments please

src/cpp/server/system_info.cpp
```diff
@@ -469,6 +469,9 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"cpu", {"x86_64"}},
         {"amd_gpu", {}},      // all AMD GPU families
     }},
+    {"llamacpp", "openvino", {"linux"}, {
+        {"cpu", {"x86_64"}},
```
superm1: How does this only show on Intel platforms?

kenvandine: Good catch — it didn't before this PR was updated. Fixed in the latest commit: introduced a new `intel_cpu` device type that is populated in `build_recipes_info()` only when the CPU brand string contains "Intel" (checked case-insensitively). The OpenVINO recipe is now gated on `{"intel_cpu", {}}` instead of `{"cpu", {"x86_64"}}`, so it will show as unsupported on AMD and other x86_64 systems with the message "Requires Intel CPU".

src/cpp/server/config_file.cpp
superm1: Please see https://github.com/lemonade-sdk/lemonade/pull/2106.  I'm getting rid of cruft, don't add more.

src/cpp/resources/backend_versions.json
```diff
@@ -6,7 +6,11 @@
     "rocm-nightly": "b1274",
     "cuda": "b9436",
     "metal": "b9253",
-    "cpu": "b9253"
+    "cpu": "b9253",
+    "openvino": "b9488"
+  },
+  "openvino": {
+    "runtime_version": "2026.0"
```
superm1, line 58: does the runtime need to get installed somehow?

jeremyfowers:
@kenvandine please rebase or close this PR

kenvandine:

PR #2086, opened 2026-06-03
Separate local and available model list output

### LINKED ISSUE #2080: Separate "local" models from "available" in lemonade list (CLI)
### Feature Description

When using lemonade via CLI, I think we should separate "available for download" & "already downloaded" models from the ```lemonade list```

Currently the two are merged and listed alphabetically which makes for a bit of a messy read when you're looking to see what's already local on the system.
(especially when this list is handy for copying the exact naming scheme required to load a model)

Image

I propose something like this;

Image

### Use Case / Motivation

QOL/UI improvement, ease of reading

### Platform Relevance

All platforms

### Additional Context

_No response_

anditherobot:
## Background

`lemonade list` used to show local and not-yet-downloaded models in one table. That made the output harder to scan, especially for new users.

This change keeps the command simple but separates the output into two sections:

- `Local`
- `Available for Download`

`lemonade list --downloaded` still prints one flat table, so existing scripts and users who rely on that output are not affected.

## Changes

- Added sectioned output for the default `lemonade list` command.
- Kept the original table printing logic in one helper.
- Kept `lemonade list --downloaded` as a single flat table.
- Updated CLI help text for `list`.
- Added test assertions for the new section headers and the old `--downloaded` behavior.
- Updated `docs/guide/cli.md` to explain the new default output.

## Testing

```powershell
python -m py_compile test\server_cli2.py
```

## Documentation

`docs/guide/cli.md` now says that `lemonade list` shows two sections by default:

- `Local` for downloaded models
- `Available for Download` for models that can be pulled

It also keeps the existing guidance that `--downloaded` restricts the output to local models.

Fixes #2080

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -145,7 +145,7 @@ lemonade export Qwen3-0.6B-GGUF --output model-info.json
 
 ## Options for list
 
-The `list` command displays available models. By default, it shows all models. Use the `--downloaded` flag to filter for downloaded models only:
+The `list` command displays all models. By default, the output is partitioned into **Local** (downloaded) and **Available for Download** sections. You can restrict the output to only local models by passing the `--downloaded` flag:
 
 ```bash
 lemonade list [options]
--- a/src/cpp/cli/lemonade_client.cpp
+++ b/src/cpp/cli/lemonade_client.cpp
@@ -387,27 +387,54 @@ int LemonadeClient::list_models(bool show_all, const std::string& name_filter) c
             return 0;
         }
 
-        std::cout << std::left << std::setw(40) << "Model Name"
-                  << std::setw(12) << "Downloaded"
-                  << "Details" << std::endl;
-        std::cout << std::string(100, '-') << std::endl;
+        // Helper lambda to print a formatted table of models.
+        auto print_model_table = [](const std::vector<ModelInfo>& models) {
+            std::cout << std::left << std::setw(40) << "Model Name"
+                      << std::setw(12) << "Downloaded"
+                      << "Details" << std::endl;
+            std::cout << std::string(100, '-') << std::endl;
 
-        // Model Name is the API id emitted verbatim by `/v1/models`. For each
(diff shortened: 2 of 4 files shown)


fl0rianr (changes requested):
Thanks for the focused PR — the direction matches #2080 and keeping lemonade list --downloaded as a flat table is a good compatibility choice.

I would like to see a bit more polish before merge:

1. The default output currently always prints both sections, even when one side is empty. For a new user with no local models, this means **Local** is followed by an empty table before the useful **Available for Download** section. Could we print a clearer message in such case like `No local models downloaded.`?

2. The new tests only assert that the section headers exist and are ordered. They do not verify that downloaded=true models are actually printed only under **Local,** or that downloaded=false models are printed only under **Available for Download**. A regression could put models in the wrong section and the tests would still pass.

3. Please add/adjust coverage for `lemonade list --downloaded` to assert that non-downloaded models are excluded, not just that the section headers are absent.

4. The Windows `sys.stdout.reconfigure(...)` change seems unrelated to this PR. If it is required for this test, please add a short comment or guard it with `hasattr(..., "reconfigure")`; otherwise it may be better as a separate cleanup.

Also, the PR testing currently lists only py_compile, which does not exercise the CLI behavior. Please run the relevant CLI test or document why it cannot be run locally. Thanks!

fl0rianr:
The CI error is related:

```
nomic-embed-text-v1-GGUF                No          llamacpp
nomic-embed-text-v2-moe-GGUF            No          llamacpp

Running: /Users/runner/work/lemonade/lemonade/build/lemonade list --downloaded
stdout: No models available

List --downloaded output: No models available

Error: Process completed with exit code 1.
```

fl0rianr (approved):
LGTM — thanks for iterating on this.

(You known me now: Minor non-blocking note: the table parser in the test is intentionally simple, but given the current CLI table format and model IDs, that seems acceptable, and since I see this uncritical and simple is also something good here you go.)

PR #2088, opened 2026-06-03
Fix #1833: Improve transcription response format compatibility

### LINKED ISSUE #1833: Update OpenAI Compatible `/v1/audio/transcriptions` with more response formats

https://lemonade-server.ai/docs/api/openai/#post-v1audiotranscriptions

### What's Missing or Wrong

https://github.com/lemonade-sdk/lemonade/blob/b5ca24be4561bcf9c6e03a76676b1a65936a63d0/docs/api/openai.md?plain=1#L385

states
```
| `response_format` | No | The format of the response. Currently only `json` is supported. | <sub>![Status](https://img.shields.io/badge/available-green)</sub> |
```

But you can actually pass in `verbose_json` and possibly [a couple more options](https://github.com/lemonade-sdk/lemonade/blob/b5ca24be4561bcf9c6e03a76676b1a65936a63d0/src/cpp/include/lemon/audio_types.h#L21-L29)

### Suggested Improvement

During investigation I found that with no additional arguments I can pass in `verbose_json` and get back a shape like
```json5
{
    "detected_language": "english",
    "detected_language_probability": 0.9998353123664856,
    "duration": 22.794687271118164,
    "language": "english",
    "language_probabilities": {
        "en": 0.9998353123664856
    },
    "segments": [
        {
            "avg_logprob": -0.3009469509124756,
            "end": 6.71,
            "id": 0,
            "no_speech_prob": 1.9279243133496493e-05,
            "start": 0.0,
            "temperature": 0.0,
            "text": " Okay. So this is my string",
            "tokens": [
               // nums
            ],
            "words": [
                {
                    "end": 2.0,
                    "probability": 0.4772109389305115,
                    "start": 0.08,
                    "t_dtw": -1,
                    "word": " Okay"
                },
               // more words
            ]
        }
    ],
    "task": "transcribe",
    "text": " -----text here------"
}
```

anditherobot:
## Summary

This PR fixes #1833 for the OpenAI-compatible transcription API.

Apps can request transcription results in different formats, including JSON, plain text, SRT, and VTT. Lemonade already accepted those requests, but plain text and subtitle formats were still wrapped in JSON instead of being returned directly. That could confuse apps expecting normal text or subtitle output.

This change makes `text`, `srt`, and `vtt` return raw text bodies. `json`, `verbose_json`, and errors still return JSON.

Before, a `vtt` response looked like this:

```json
{
  "text": "WEBVTT\n\n00:00:00.000 --> 00:00:04.000\nListen, read, repeat.\n"
}
```

After this change:

```text
WEBVTT

00:00:00.000 --> 00:00:04.000
Listen, read, repeat.
```

The same applies to `srt`.

## Testing

I reviewed the full diff and kept the change scoped to transcription response formats.

I manually checked the endpoint with a WAV file using all supported response formats: `json`, `verbose_json`, `text`, `srt`, and `vtt`. This confirmed the old behavior: `text`, `srt`, and `vtt` were returned as JSON wrappers.

I ran `git diff --check`. The commit also passed the repo pre-commit hooks, including Black.

## Documentation

Updated the OpenAI API docs to list the supported transcription response formats: `json`, `verbose_json`, `text`, `srt`, and `vtt`.

Fixes #1833


fl0rianr (changes requested):
Thanks, good work! The transcription behavior looks right.
But there is a blocker regarding the scope of the forward_multipart_request() change. This helper is also used by SD image edits/variations, where successful responses are expected to be JSON. With this PR, any successful non-JSON response from those backends would become `{"text": ...}` and be returned to clients as HTTP 200 instead of surfacing as an invalid backend response. Could we make the non-JSON-success fallback opt-in and enable it only for transcription requests with response_format set to `text`, `srt,` or `vtt`?

I’d also add a negative test for an unsupported response_format value so the new documented supported-values list is enforced deterministically.

ramkrishna2910:
@claude review

fl0rianr:
CI error is related:

```
FAIL: test_002b_transcription_response_formats (__main__.WhisperTests) (response_format='verbose_json')
Test audio transcription response_format handling.
  File "C:\Users\user\actions-runner\_work\lemonade\lemonade\test\server_whisper.py", line 246, in test_002b_transcription_response_formats
    self.assertIn("segments", result)
AssertionError: 'segments' not found in {'model': 'whisper-v3:turbo', 'text': ' And so, my fellow Americans, ask not what your country can do for you ask what you can do for your country.'}

Ran 3 tests in 14.913s

FAILED (failures=1)
```

jeremyfowers:
@anditherobot please get CI healthy or close the PR (nothing personal just putting messages like this any PR of this age)

fl0rianr (commented):
Thanks for updating and rebasing this. I re-reviewed the current head. My original two requests are addressed: the shared multipart fallback is now opt-in, and unsupported `response_format` values are rejected with a 400 and covered by a test.

I still see two issues before I can approve:

### 1. Raw text is still detected by JSON parse failure

Both `WrappedServer::forward_multipart_request()` and the Whisper paths first try `json::parse(body)` and only treat the response as raw text if parsing fails.

That is not safe for `response_format=text`. A perfectly valid transcript such as `true`, `null`, or `123` is also valid JSON. In that case parsing succeeds, the expected `{"text": ...}` wrapper is never created, and `handle_audio_transcriptions()` falls back to returning JSON instead of the requested raw text.

The requested format should determine how a successful response is interpreted, e.g.:

```cpp
if (allow_plain_text_success) {
    return json{{"text", response.body}};
}
return json::parse(response.body);
```

The same applies to the Whisper paths. A small regression test with a JSON-looking text body such as `true\n` would cover this nicely.

### 2. FLM does not actually provide SRT/VTT responses

I traced the current FastFlowLM implementation as well. Its `/v1/audio/transcriptions` endpoint only copies `model` and `file` from the multipart request; `response_format` is not passed to the ASR handler. The handler always returns the compact `{model, text}` JSON response.

Lemonade can turn that into plain `text`, but it cannot produce real SRT or VTT from it. The current test masks this by skipping the `-->` / `WEBVTT` assertions when `wrapped_server == "flm"`, so ordinary transcript text can pass as successful `srt` or `vtt`.

I think we should either reject/document those formats as unsupported for FLM, or implement actual format support. The test should not report SRT/VTT support without validating the requested format. The same limitation also explains why FLM's `verbose_json` response is only the compact shape.

One small bonus of fixing Nr. 1 by branching on the requested format before parsing is that the strict multipart path can keep its previous JSON parse-error behavior instead of changing error semantics for unrelated callers.

Once these two points are addressed this looks good from my side.

fl0rianr (approved):
LGTM

PR #2090, opened 2026-06-03
Add Gemma 4 12B

sawansri:
image

Works well for me!

### DIFF
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -688,6 +688,16 @@
         ],
         "size": 19.5
     },
+    "Gemma-4-12B-it-GGUF": {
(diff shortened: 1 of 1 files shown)


fl0rianr (commented):
Great! Does it not work with vision, or audio?

sawansri:
> Great! Does it not work with vision, or audio?

The unsloth quant and the full precision model from google don't include a mmproj encoder so I don't think llamacpp currently has a way of consuming multimodal input. Text capabilities work fine atm.

In the model card, Google emphasizes that "it brings native audio and vision understanding directly to local environments without the need for separate encoders. This unified approach to multimodality makes the model encoder-free".

fl0rianr (approved):

PR #2092, opened 2026-06-03
feat(rocm): enable AMD Instinct MI100 (gfx908) and MI200 (gfx90a) in llama.cpp ROCm

kenvandine:
## Summary

Enables AMD Instinct MI100 (CDNA1/gfx908) and MI200/MI210/MI250 (CDNA2/gfx90a) in the `llamacpp` ROCm backend. The kernels are already shipped — `GPU_TARGETS` in `lemonade-sdk/llama.cpp`'s `release.yml` includes `gfx908;gfx90a` for the Linux ROCm build, and `lemonade-sdk/llamacpp-rocm` publishes per-arch nightly assets. Only the descriptor support list gated them out.

- Add `gfx908`/`gfx90a` to the `llamacpp` rocm support row, each with a **Linux-only** `ArchInstallGate` — the Windows ROCm build's `GPU_TARGETS` omits CDNA entirely.
- Add `DEVICE_FAMILY_NAMES` display names for both.

`backend_versions.json` already carries these arches (`therock.architectures`, `url_mapping` → `gfx90X-dcgpu`), so no registry change is needed.

## Review feedback addressed

- Dropped the name-based CDNA fallback in `identify_rocm_arch_from_name()` (@superm1). Linux detection goes through KFD `gfx_target_version`, which the digit path already converts to `gfx908`/`gfx90a` directly; the WSL/HSA fallback reports the ISA in the device name, caught by the regex. The marketing-name checks were unreachable.
- Dropped `test/test_rocm_arch_mapping.py` (@superm1) — it tested a Python replica of the C++ function, so it could not catch a C++ regression.
- Regenerated the docs via `docs/tools/gen_backend_boilerplate.py`.

## Test plan

- [x] `ctest` — 38/38 pass, including the existing `LlamacppGfx950GatingTest`
- [x] `gen_backend_boilerplate.py --check` clean
- [x] Verified on Linux: `backend_supports_arch("llamacpp", …, "gfx908"/"gfx90a")` is true on stable and nightly; `sd-cpp` correctly excluded; KFD `90008`→`gfx908`, `90010`→`gfx90a`; `"AMD Instinct MI210"` now returns `""` while `"AMD Instinct MI210 (gfx90a)"` still resolves
- [ ] End-to-end `lemonade pull` + `run` on real MI100 / MI210 hardware


src/cpp/server/system_info.cpp
```diff
@@ -91,6 +91,14 @@ const std::set<std::string> CUDA_SUPPORTED_ARCHS = {
 // ROCm architecture mapping - maps specific gfx architectures to their family (download target).
 // Empty string means "no ROCm binary for this ISA" — skip for get_rocm_arch / install filenames.
 const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
+    // CDNA1 - AMD Instinct MI100 (Arcturus)
+    {"gfx908",  "gfx908"},   // Direct arch string (from HSA/WSL path)
```
superm1: what is this

src/cpp/server/system_info.cpp
```diff
@@ -91,6 +91,14 @@ const std::set<std::string> CUDA_SUPPORTED_ARCHS = {
...
+    {"gfx9008", "gfx908"},   // KFD-computed string on native Linux (90008 → gfx9008)
+
+    // CDNA2 - AMD Instinct MI200/MI210 (Aldebaran)
+    {"gfx90a",  "gfx90a"},   // Direct arch string (from HSA/WSL path)
```
superm1: what is this

src/cpp/server/system_info.cpp
```diff
@@ -1895,6 +1905,19 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
...
+        return "gfx908";
+    }
+
+    // CDNA2 GPUs (gfx90a architecture) - AMD Instinct MI200/MI210
+    if (device_lower.find("mi200") != std::string::npos ||
+        device_lower.find("mi210") != std::string::npos ||
+        device_lower.find("aldebaran") != std::string::npos) {
+        return "gfx90a";
+    }
+
```
superm1: where did all this come from?

superm1 (commented):
don't you need lemonade/llama.cpp too?

kenvandine:
Yes, https://github.com/lemonade-sdk/llama.cpp/pull/15

superm1:
> Yes, https://github.com/lemonade-sdk/llama.cpp/pull/15

That's for openvino though. We need rocm change.

kenvandine:
> > Yes, [lemonade-sdk/llama.cpp#15](https://github.com/lemonade-sdk/llama.cpp/pull/15)
> 
> That's for openvino though. We need rocm change.

Sorry, crossed the streams there.  Our llama.cpp already includes these in the gpu_targets.

kenvandine:

src/cpp/server/system_info.cpp
```diff
@@ -439,7 +439,7 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"amd_gpu", {}},      // all AMD GPU families
     }},
     {"llamacpp", "rocm", {"windows", "linux"}, {
-        {"amd_gpu", {"gfx1150", "gfx1151", "gfx1152", "gfx103X", "gfx110X", "gfx120X"}},  // STX iGPUs + RDNA2/3/4 dGPUs
+        {"amd_gpu", {"gfx908", "gfx90a", "gfx1150", "gfx1151", "gfx1152", "gfx103X", "gfx110X", "gfx120X"}},  // CDNA1/2 + STX iGPUs + Krackan + RDNA2/3/4 dGPUs
     }},
```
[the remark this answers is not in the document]
kenvandine: This is a pre-existing issue in main, not introduced by this PR. The new `gfx908` and `gfx90a` entries are exact arch strings (not wildcards), so they match correctly both via the KFD numeric path (`snprintf("gfx%d%x%x", ...)` produces exactly `"gfx908"` or `"gfx90a"`) and via the name-based fallback. The wildcard mismatch for RDNA2/3/4 (`gfx103X` etc.) on Linux KFD is a separate pre-existing issue in main and out of scope here.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/cpp/server/system_info.cpp
```diff
@@ -1897,6 +1900,19 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
         return "gfx103X";
     }
 
+    // CDNA1 GPUs (gfx908 architecture) - AMD Instinct MI100
+    if (device_lower.find("mi100") != std::string::npos ||
+        device_lower.find("arcturus") != std::string::npos) {
+        return "gfx908";
+    }
```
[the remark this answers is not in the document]
kenvandine: Fixed in d3d994bc — the CDNA1/CDNA2 name checks are now moved before the early `radeon`/`amd` guard so bare names like `"MI100"`, `"Arcturus"`, or `"MI250"` are detected correctly regardless of branding prefix.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/cpp/server/system_info.cpp
```diff
@@ -439,7 +439,7 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"amd_gpu", {}},      // all AMD GPU families
     }},
     {"llamacpp", "rocm", {"windows", "linux"}, {
-        {"amd_gpu", {"gfx1150", "gfx1151", "gfx1152", "gfx103X", "gfx110X", "gfx120X"}},  // STX iGPUs + RDNA2/3/4 dGPUs
+        {"amd_gpu", {"gfx908", "gfx90a", "gfx1150", "gfx1151", "gfx1152", "gfx103X", "gfx110X", "gfx120X"}},  // CDNA1/2 + STX iGPUs + Krackan + RDNA2/3/4 dGPUs
```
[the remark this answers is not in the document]
kenvandine: The regex and KFD parsing changes were already present in `main` before this PR — they predated the merge conflict that Copilot SWE resolved in 7157f1a. This PR only adds the backend support list entries, family names, and name-based fallback detection for gfx908/gfx90a. The PR description has been updated to reflect the actual scope of changes.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/cpp/server/system_info.cpp
```diff
@@ -1897,6 +1900,19 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
...
+        device_lower.find("arcturus") != std::string::npos) {
+        return "gfx908";
+    }
+
+    // CDNA2 GPUs (gfx90a architecture) - AMD Instinct MI200/MI210
+    if (device_lower.find("mi200") != std::string::npos ||
+        device_lower.find("mi210") != std::string::npos ||
+        device_lower.find("aldebaran") != std::string::npos) {
+        return "gfx90a";
+    }
```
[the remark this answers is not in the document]
kenvandine: Fixed in d3d994bc — added `mi250` to the CDNA2 detection block (MI250 and MI250X are both gfx90a). The `DEVICE_FAMILY_NAMES` entry has also been updated to `"AMD Instinct MI200/MI210/MI250 (CDNA2)"`.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/cpp/server/system_info.cpp
```diff
@@ -538,6 +539,8 @@ static const std::map<std::string, std::string> DEVICE_FAMILY_NAMES = {
     {"arm64", "ARM64 processors"},
 
     // AMD GPU architectures (ROCm)
+    {"gfx908",  "AMD Instinct MI100 (CDNA1)"},
+    {"gfx90a",  "AMD Instinct MI200/MI210 (CDNA2)"},
```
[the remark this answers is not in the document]
kenvandine: Fixed in d3d994bc — removed the extra space to align with adjacent entries.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/cpp/server/system_info.cpp
```diff
@@ -1856,6 +1859,20 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
...
+        return "gfx908";
+    }
+
+    // CDNA2 GPUs (gfx90a architecture) - AMD Instinct MI200/MI210/MI250
+    if (device_lower.find("mi200") != std::string::npos ||
+        device_lower.find("mi210") != std::string::npos ||
+        device_lower.find("mi250") != std::string::npos ||
+        device_lower.find("aldebaran") != std::string::npos) {
+        return "gfx90a";
+    }
```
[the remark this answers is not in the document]
kenvandine: Done in 2b6db201 — added `test/test_rocm_arch_mapping.py` following the same pattern as `test/test_cuda_arch_mapping.py`. It covers all three detection paths:

- **KFD decimal path**: `"90008"` → `gfx908`, `"90010"` → `gfx90a`, `"100300"` → `gfx1030`
- **Direct gfx token**: `"gfx908"`, `"gfx90a"`, etc. extracted from the device string
- **CDNA name fallback**: full names (`"AMD Instinct MI250X"`) and bare names (`"MI250X"`, `"Arcturus"`) that must work even without the "AMD" prefix
- **RDNA name fallback**: RDNA2/3/4 and iGPU model strings
- **Non-AMD names** return `""`

29 tests, runs in ~0.001s with no hardware required.

---
_Generated by [Claude Code](https://claude.ai/code)_

test/test_rocm_arch_mapping.py
```diff
@@ -0,0 +1,229 @@
...
+ROCM_SUPPORTED_ARCHS = {
+    "gfx908",   # CDNA1 (MI100)
+    "gfx90a",   # CDNA2 (MI200/MI210/MI250)
+    "gfx1150",  # Strix Point iGPU
+    "gfx1151",  # Strix Halo iGPU
+    "gfx1152",  # Krackan Point iGPU
+    "gfx103X",  # RDNA2 dGPUs
+    "gfx110X",  # RDNA3 dGPUs
+    "gfx120X",  # RDNA4 dGPUs
+}
```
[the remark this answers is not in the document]
kenvandine: Fixed in e47f6785 — renamed to `ROCM_SUPPORTED_FAMILIES` and added a comment explaining that the RDNA wildcard tokens (`gfx103X` etc.) are only reachable via the name-based path, not the KFD path which produces exact arch strings like `gfx1030`. Assertion message updated to match.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/cpp/server/system_info.cpp
kenvandine: These are the new CDNA1 (`gfx908`, MI100/Arcturus) and CDNA2 (`gfx90a`, MI200/MI210/MI250/Aldebaran) entries in `ROCM_ARCH_MAPPING` and `ROCM_SUPPORTED_ARCHS` — the core addition of this PR to register ROCm backend support for those GPU families.

src/cpp/server/system_info.cpp
kenvandine: This is new name-based detection for CDNA1/CDNA2 GPUs, added by this PR. It allows `identify_rocm_arch_from_name()` to recognize MI100, MI200, MI210, MI250, Arcturus, and Aldebaran device strings and map them to `gfx908`/`gfx90a` respectively — complementing the KFD numeric path that may not always be available.

src/cpp/server/system_info.cpp
kenvandine: Same as above — the new `gfx908`/`gfx90a` entries in `ROCM_ARCH_MAPPING`. These map the CDNA1/CDNA2 arch strings to themselves (they're downloaded as exact named targets), which is the expected pattern for arches that don't need to be aliased to a family wildcard.

src/cpp/server/system_info.cpp
superm1: But we don't use name based mapping on Linux.  And CDNA doesn't work on Windows (AFAIK).

src/cpp/server/system_info.cpp
superm1: But you can't get CDNA working from WSL via KFD topology.  That makes no sense.

test/test_rocm_arch_mapping.py
superm1, line 1: please don't add a test that doesn't add anything of value like this.

src/cpp/server/system_info.cpp
kenvandine: You were right to question it, and my earlier reply here was wrong — it described `ROCM_ARCH_MAPPING`, which had already been deleted from this branch by the time I wrote it. Sorry for the noise.

The branch is now rebased on current `main` and there is no mapping table left. The whole diff is +10/-4:

- `gfx908`/`gfx90a` added to the `llamacpp` rocm support row in `llamacpp.h`, each with a Linux-only `ArchInstallGate`
- two `DEVICE_FAMILY_NAMES` display strings
- regenerated `README.md` + `docs/dev/backends-reference.md`

`backend_versions.json` already carried both arches (`therock.architectures`, `url_mapping` -> `gfx90X-dcgpu`), so nothing was needed there.

src/cpp/server/system_info.cpp
kenvandine: Agreed, and my earlier answer in this thread was describing code that no longer existed — `ROCM_ARCH_MAPPING` was gone from the branch already. There is no mapping table in the current diff.

To your point directly: nothing here relies on WSL KFD topology for CDNA. On native Linux the arch comes from KFD `gfx_target_version`, which the digit branch of `identify_rocm_arch_from_name()` converts straight to `gfx908`/`gfx90a` (`90008` -> `gfx908`, `90010` -> `gfx90a`, verified against the built binary). In the WSL/container case where KFD yields no usable nodes, the `/dev/dxg` HSA fallback reports the ISA in the device name, so the `gfx[0-9a-f]{3,4}` regex catches it first. Neither path needed the marketing-name mapping, which is why it is gone.

src/cpp/server/system_info.cpp
kenvandine: You were right on both counts — the name-based CDNA block is removed.

On Linux it was unreachable: `get_amd_dgpu_devices()` sets `gpu.name` to the raw KFD `gfx_target_version`, so the digit branch returns `gfx908`/`gfx90a` before the name checks are ever consulted. Checked against the built binary — `"AMD Instinct MI210"` now returns `""`, while `"AMD Instinct MI210 (gfx90a)"` (the HSA/dxg form) still resolves via the regex.

On Windows you are right that CDNA does not work, and it turns out the build already says so: `GPU_TARGETS` in `lemonade-sdk/llama.cpp`'s `release.yml` lists `gfx908;gfx90a` for the `ubuntu-22-rocm` job but the `windows-rocm` job is RDNA-only. So both arches now carry a Linux-only `ArchInstallGate`, same shape as the existing `gfx950` gate — otherwise we would have advertised a Windows ROCm target with no kernels behind it. `llamacpp-rocm` does publish Windows `gfx908`/`gfx90a` nightly assets, but the gate keeps us from pointing at them.

test/test_rocm_arch_mapping.py
kenvandine: Agreed — deleted. It reimplemented `identify_rocm_arch_from_name()` in Python and then asserted against the replica, so it could not have caught a regression in the C++ at all.

No test replaces it. The gate is exercised by the existing `LlamacppGfx950GatingTest` path and by `server_gfx_topology.py`; full `ctest` is 38/38 on this branch.

kenvandine:
Verification on this branch (Linux, `build/lemond` from the rebased tree):

`POST /api/v1/install/dry-run` for `llamacpp` / `rocm-nightly`:

```
gfx908 -> supported=true  llama-b1292-ubuntu-rocm-gfx908-x64.zip
gfx90a -> supported=true  llama-b1292-ubuntu-rocm-gfx90a-x64.zip
```

Both resolved URLs HEAD-probe **200** against `lemonade-sdk/llamacpp-rocm`. `server_gfx_topology.py` passes (3/3, including the live link check, which covers `("llamacpp", "rocm-nightly")` across the full `therock.architectures` matrix — so the new arches are exercised end-to-end by an existing test rather than a new one).

Also: `ctest` 38/38, `gen_backend_boilerplate.py --check` clean.

Still outstanding is the one thing I cannot check here — an actual `pull` + `run` on real MI100 / MI210 silicon.

src/cpp/include/lemon/backends/llamacpp/llamacpp.h
```diff
@@ -37,10 +37,14 @@ inline const BackendDescriptor descriptor = {
          {{"nvidia_gpu", {"sm_75", "sm_80", "sm_86", "sm_89", "sm_90", "sm_100", "sm_120", "sm_121"}}}, "NVIDIA GPUs (Turing or newer)**"},
         {"vulkan", {"windows", "linux"}, {{"cpu", {"x86_64", "arm64"}}, {"amd_gpu", {}}}, "x86_64 CPU, AMD iGPU, AMD dGPU; ARM64 CPU/GPU (Linux)"},
         {"rocm", {"windows", "linux"},
-         {{"amd_gpu", {"gfx1150", "gfx1151", "gfx1152", "gfx103X", "gfx110X", "gfx120X", "gfx942", "gfx950"}}}, "Supported AMD ROCm iGPU/dGPU families, incl. AMD Instinct MI300X (gfx942) and MI350X (gfx950, Linux + stable only)*",
+         {{"amd_gpu", {"gfx908", "gfx90a", "gfx1150", "gfx1151", "gfx1152", "gfx103X", "gfx110X", "gfx120X", "gfx942", "gfx950"}}}, "Supported AMD ROCm iGPU/dGPU families, incl. AMD Instinct MI100/MI200 (gfx908/gfx90a, Linux only), MI300X (gfx942) and MI350X (gfx950, Linux + stable only)*",
```
superm1: I suggest we drop the detailed comment and just say "AMD GPUs supported by ROCm"

superm1: can you please keep the order alphabetical?  we shouldn't have gfx90* at the start and end of the list.

src/cpp/include/lemon/backends/llamacpp/llamacpp.h
```diff
@@ -37,10 +37,14 @@ inline const BackendDescriptor descriptor = {
...
          // gfx950 (MI350) only ships a Linux + stable-channel asset so far; the
          // Windows TheRock dist and the nightly build aren't published yet.
-         {{"gfx950", {/*os*/ {"linux"}, /*channels*/ {"stable"}}}}},
+         // CDNA1/CDNA2 are absent from the Windows ROCm build's GPU_TARGETS, so
+         // they are Linux-only on both channels.
+         {{"gfx950", {/*os*/ {"linux"}, /*channels*/ {"stable"}}},
+          {"gfx908", {/*os*/ {"linux"}, /*channels*/ {}}},
+          {"gfx90a", {/*os*/ {"linux"}, /*channels*/ {}}}}},
```
superm1: If you're going to be complete; I don't expect gfx942 has Windows binaries either.

src/cpp/server/system_info.cpp
superm1: please move these to group with the other gfx9.

Keep in mind it's like "gfx09" so it should come before gfx10.

README.md
```diff
@@ -159,7 +159,7 @@ Lemonade supports multiple inference engines for LLM, speech, TTS, and image gen
     </tr>
     <tr>
       <td><code>rocm</code></td>
-      <td>Supported AMD ROCm iGPU/dGPU families, incl. AMD Instinct MI300X (gfx942) and MI350X (gfx950, Linux + stable only)*</td>
+      <td>Supported AMD ROCm iGPU/dGPU families, incl. AMD Instinct MI100/MI200 (gfx908/gfx90a, Linux only), MI300X (gfx942) and MI350X (gfx950, Linux + stable only)*</td>
```
superm1: I don't want to be in the business keeping this comment up to date.  Maybe a link to https://rocm.ai would be better?

src/cpp/include/lemon/backends/llamacpp/llamacpp.h
kenvandine: Done — simplified to "AMD GPUs supported by ROCm" in 3d2e44f7.

src/cpp/include/lemon/backends/llamacpp/llamacpp.h
kenvandine: Good catch — added a Linux-only gate for `gfx942` in 3d2e44f7.

src/cpp/include/lemon/backends/llamacpp/llamacpp.h
kenvandine: Fixed — arch list is now alphabetical in 3d2e44f7.

src/cpp/server/system_info.cpp
kenvandine: Fixed in 7920f10f — gfx9* entries are now grouped first (gfx09 < gfx10).

superm1 (approved):

superm1:
Plz fix the CI failure for docs drift, otherwise LGTM now.

PR #2093, opened 2026-06-04
fix: enable TheRock ROCm installation on Windows for Radeon RX GPUs

### LINKED ISSUE #2050: therock installation does not work on windows
### Platform

Windows

### Lemonade Version

10.6.0

### GPU / APU Model

AMD Radeon RX 9060 XT

### Component

stable-diffusion.cpp

### Bug Description

during installation of rocm-stable backend for sd-cpp therock is not installed if required

### Steps to Reproduce

1. Go to backend manager
2. Make sure rocm-stable backend of sd-cpp is not installed
3. Install rocm-stable backend of sd-cpp

Expected: therock gets installed after rocm-stable backend of sd-cpp is installed if it is not present system wide
Actual: therock does not get installed, no log entry

### Log Output

```shell

```

### Additional Context

identify_rocm_arch_from_name in system_info.cpp maps Radeon RX 9060 XT to arch "gfx120X".
backend_versions.json does not contain "gfx120X" in architectures so will_install_therock in backend_manager.cpp will return false and therock is not installed.

When "gfx120X" is added to the architectures section in backend_versions.json and also to the url_mapping section, therock will get installed.

anditherobot:
# fix: enable TheRock ROCm installation on Windows for Radeon RX GPUs (#2050)

## Summary
Enables TheRock installation for Radeon RX GPUs when Lemonade detects the GPU by a family architecture code such as `gfx120X`.

In the reported Windows RX 9060 XT case, Lemonade correctly identifies the card as `gfx120X`. That code means the RX 9000 / RDNA4 GPU family. However, `backend_versions.json` did not list `gfx120X`, so Lemonade skipped the TheRock install step.

## Change
Adds the Radeon RX family architecture codes to the TheRock install metadata:

| GPU Series | Architecture Code | TheRock Package |
| :--- | :--- | :--- |
| ~~RX 5000~~ | ~~`gfx101X`~~ | ~~`gfx101X-dgpu`~~ *(removed per review — out of scope for #2050)* |
| RX 6000 | `gfx103X` | `gfx103X-all` |
| RX 7000 | `gfx110X` | `gfx110X-all` |
| RX 9000 | `gfx120X` | `gfx120X-all` |

`gfx1103` is also listed directly, since the removed C++ table no longer translates it to `gfx110X`.

Per review, the duplicate C++ arch table (`ROCM_ARCH_MAPPING`) was removed (2c1e6316) so the JSON file is the single source for arch names.

~~Also adds RX 5000 name detection so cards such as RX 5700 XT can resolve to `gfx101X`.~~ *(removed per review — out of scope for #2050)*

## Testing
- ~~Confirmed locally on an RX 5700 XT that the ROCm backend is marked installable.~~ *(RX 5000 support removed from this PR)*
- Verified the new TheRock package URLs exist on AMD's servers.
- Confirmed the change covers the reported RX 9060 XT case, where detection returns `gfx120X`.

Fixes #2050


superm1 (commented):
I think this needs more matches for Linux too or this will cause an issue when a Linux user tries.

src/cpp/server/system_info.cpp
```diff
@@ -91,6 +91,12 @@ const std::set<std::string> CUDA_SUPPORTED_ARCHS = {
 // ROCm architecture mapping - maps specific gfx architectures to their family (download target).
 // Empty string means "no ROCm binary for this ISA" — skip for get_rocm_arch / install filenames.
 const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
+    // RDNA1 family. Linux KFD reports the exact GPU ISA (gfx1010/gfx1011/gfx1012),
+    // while Lemonade recipes and ROCm backend downloads use the shared family name gfx101X.
+    {"gfx1010", "gfx101X"}, // Navi 10 (e.g., RX 5700 / 5700 XT)
+    {"gfx1011", "gfx101X"}, // Navi 12 (e.g., Radeon Pro 5600M)
+    {"gfx1012", "gfx101X"}, // Navi 14 (e.g., RX 5500 / 5500 XT)
+
```
superm1: you're adding these, but the mappings you added also had more (103x, 110x, 120x).   Either add them all here or drop the mapping from backend_versions.json

src/cpp/server/system_info.cpp
```diff
@@ -470,7 +476,7 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"amd_gpu", {}},      // all AMD GPU families
     }},
     {"llamacpp", "rocm", {"windows", "linux"}, {
-        {"amd_gpu", {"gfx1150", "gfx1151", "gfx103X", "gfx110X", "gfx120X"}},  // STX iGPUs + RDNA2/3/4 dGPUs
+        {"amd_gpu", {"gfx101X", "gfx1150", "gfx1151", "gfx103X", "gfx110X", "gfx120X"}},  // STX iGPUs + RDNA1/2/3/4 dGPUs
```
superm1: same comment as above, missing more mappings

superm1:
All of the targets and how they map are described here: https://github.com/ROCm/TheRock/blob/main/cmake/therock_amdgpu_targets.cmake

You can cross reference this.

src/cpp/server/system_info.cpp
```diff
@@ -470,7 +476,7 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
         {"amd_gpu", {}},      // all AMD GPU families
     }},
     {"llamacpp", "rocm", {"windows", "linux"}, {
-        {"amd_gpu", {"gfx1150", "gfx1151", "gfx103X", "gfx110X", "gfx120X"}},  // STX iGPUs + RDNA2/3/4 dGPUs
+        {"amd_gpu", {"gfx101X", "gfx1150", "gfx1151", "gfx1152", "gfx1153", "gfx103X", "gfx110X", "gfx120X"}},  // STX iGPUs + RDNA1/2/3/4 dGPUs
```
superm1: You should just drop the comments. This is never going to age well. You also added things like Phoenix, hawk point here.

AGENTS.md
```diff
@@ -191,6 +191,16 @@ These MUST be maintained in all changes:
...
 13. **Desktop app is on-demand; `lemond` runs independently** — On Windows, `LemonadeServer.exe` (which embeds `lemond` + tray icon) is the always-on process, auto-started via the Windows startup folder. The Tauri desktop app (`lemonade-app.exe`) is opened on demand when the user wants the UI and must not be added to startup. The desktop app must not embed or manage `lemond`'s lifecycle — it discovers the already-running server (UDP beacon for local, explicit base URL for remote) and speaks to it over HTTP.
+14. **AMD GPU platform parity** — When adding or expanding AMD/ROCm GPU support, keep Windows and Linux detection in sync. Windows commonly detects GPUs from product names like `5700` or `6800`, while Linux may report raw KFD ISA codes like `gfx1010` or `gfx1030`. Update `ROCM_ARCH_MAPPING` in `system_info.cpp` as needed so Linux detections resolve to the backend target expected by recipes and ROCm downloads, or to `""` when that ISA is intentionally unsupported.
+
+## Pending Local Work
```
superm1: Shouldn't this not commit?

src/cpp/resources/backend_versions.json
```diff
@@ -27,12 +27,18 @@
       "gfx908",
       "gfx90a",
       "gfx942",
+      "gfx101X",
```
superm1: I don't think this is correct here

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -471,9 +470,9 @@ void LlamaCppServer::load(const std::string& model_name,
         }
 
         std::string arch = lemon::SystemInfo::get_rocm_arch();
-        if (arch == "gfx1151") {
+        if (arch == "gfx1151" || arch == "gfx1152") {
```
superm1: Why?

anditherobot: Fair point. I don't have a way to test if `gfx1152` actually needs this memory size fix.

superm1:
Option A is the correct approach.  We rely upon AMD binaries, and if AMD doesn't publish them we should fall back to vulkan or CPU.

src/cpp/server/backends/llamacpp_server.cpp
superm1: yes please drop then.

superm1:
> gfx1010–12 	100100–02

HTH do you get 100100-02 from a 32 bit integer?  This is impossible.

https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/amd/amdkfd/kfd_topology.c#L472

superm1:
> > > gfx1010–12 	100100–02
> > 
> > HTH do you get 100100-02 from a 32 bit integer? This is impossible.
> > 
> > https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/amd/amdkfd/kfd_topology.c#L472
> 
> The dash was meant as range shorthand, not as a literal kernel value: `gfx1010` / `gfx1011` / `gfx1012` =
>   `100100` / `100101` / `100102`

Ah Thanks for confirm

src/cpp/resources/backend_versions.json
```diff
@@ -27,9 +27,14 @@
...
+      "gfx103X",
+      "gfx110X",
+      "gfx120X",
```
superm1: these are not correct

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -447,8 +447,7 @@ void LlamaCppServer::load(const std::string& model_name,
         LOG(DEBUG, "LlamaCpp") << "Setting LD_LIBRARY_PATH=" << lib_path << std::endl;
     }
 #else
-    // For ROCm on Windows with gfx1151, set OCL_SET_SVMSIZE
-    // This is a patch to enable loading larger models
+    // For ROCm on Windows, set OCL_SET_SVM_SIZE where needed to load larger models.
```
superm1: why did you change this comment?

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -473,7 +472,7 @@ void LlamaCppServer::load(const std::string& model_name,
         std::string arch = lemon::SystemInfo::get_rocm_arch();
         if (arch == "gfx1151") {
             env_vars.push_back({"OCL_SET_SVM_SIZE", "262144"});
-            LOG(DEBUG, "LlamaCpp") << "Setting OCL_SET_SVM_SIZE=262144 for gfx1151 (enables loading larger models)" << std::endl;
+            LOG(DEBUG, "LlamaCpp") << "Setting OCL_SET_SVM_SIZE=262144 for " << arch << " (enables loading larger models)" << std::endl;
```
superm1: why did you change this log?

src/cpp/server/system_info.cpp
```diff
@@ -91,28 +91,28 @@ const std::set<std::string> CUDA_SUPPORTED_ARCHS = {
...
     {"gfx1102", "gfx110X"},
     {"gfx1103", "gfx110X"},
 
-    // RDNA3.5 iGPUs - explicit binary names (no family mapping)
-    {"gfx1150", "gfx1150"},  // Maps to exact binary name
-    {"gfx1151", "gfx1151"},  // Maps to exact binary name
+    {"gfx1150", "gfx1150"},
+    {"gfx1151", "gfx1151"},
+    {"gfx1152", "gfx1152"},
+    {"gfx1153", ""},
```
superm1: why even have the gfx1153 line?

src/cpp/server/system_info.cpp
```diff
@@ -535,7 +535,7 @@ static const std::vector<RecipeBackendDef> RECIPE_DEFS = {
 
     // vLLM - ROCm backend for AMD GPUs (Linux only)
     {"vllm", "rocm", {"linux"}, {
-        {"amd_gpu", {"gfx1150", "gfx1151", "gfx110X", "gfx120X"}},
+        {"amd_gpu", {"gfx1150", "gfx1151", "gfx1152", "gfx110X", "gfx120X"}},
```
superm1: you don't want 103X here too?

src/cpp/server/backends/llamacpp_server.cpp
anditherobot: Leftover from when the code briefly supported gfx1152. Missed reverting the comment in 3f1f6898. Will fix.

src/cpp/server/backends/llamacpp_server.cpp
anditherobot: Leftover from when the condition covered gfx1152. Missed reverting the log in 3f1f6898. Will fix.

src/cpp/server/system_info.cpp
anditherobot: Leftover from 1d5f3294. Falls back to CPU/Vulkan by default so the empty mapping is redundant. Will remove.

src/cpp/resources/backend_versions.json
anditherobot: Good catch. These are also checked in `backend_manager.cpp` against the alias from `get_rocm_arch()`, so removing them would block install for RDNA users. Would something like this work:
```cpp
if (backend_versions["therock"]["url_mapping"].contains(rocm_arch)) {
    arch_supported = true;
}
```
or do you have a different approach in mind?

src/cpp/server/system_info.cpp
anditherobot: Intentionally left out. No gfx103X package exists in vllm-rocm releases so it would fail at install. Any proposed alternative?

superm1: That makes sense, no concerns.

src/cpp/resources/backend_versions.json
superm1: I think the actual issue is we have `ROCM_ARCH_MAPPING` but also the mapping in the JSON.  Yeah - I think we need to simplify this.  How about we adjust code to drop `ROCM_ARCH_MAPPING`?

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -447,7 +447,7 @@ void LlamaCppServer::load(const std::string& model_name,
         LOG(DEBUG, "LlamaCpp") << "Setting LD_LIBRARY_PATH=" << lib_path << std::endl;
     }
 #else
-    // For ROCm on Windows with gfx1151, set OCL_SET_SVMSIZE
+    // For ROCm on Windows with gfx1151, set OCL_SET_SVM_SIZE
```
superm1: still changing this comment, leave as is please.

anditherobot: Restored in 2c1e6316

src/cpp/resources/backend_versions.json
anditherobot: Removed `ROCM_ARCH_MAPPING` from `system_info.cpp` in 2c1e6316. `url_mapping` in `backend_versions.json` 
 Remove two comment lines that  are potentially stale 
 

// - Returns the mapped family (or exact gfx115x target); map value may be "" to skip ROCm for that ISA.
// - If not in ROCM_ARCH_MAPPING, returns the raw detected arch for other unsupported GPUs.

superm1 (commented):
Please remove the changes to src/cpp/resources/backend_versions.json that aren't needed.
Please drop the unrun tests.  They were useful for your development but either run them in CI or drop them (I'd lean on drop)

src/cpp/server/backends/llamacpp_server.cpp
ckuethe: I do have gfx1152. I haven't exhaustively tested rocm yet, but #2186 allows me to consume at least 48GB of VRAM with qwen3-coder-next. I'll try push it harder in the next few days.

src/cpp/resources/backend_versions.json
```diff
@@ -27,9 +27,13 @@
       "gfx908",
       "gfx90a",
       "gfx942",
+      "gfx103X",
```
superm1: these still aren't correct here.

gfx***X is not an architecture.  This is for list of ISAs.

superm1 (approved):
I've rebased the branch to fix the CI and dropped the spurious names in architecture list.

PR #2096, opened 2026-06-04
test: add coverage for GET /v1/pull/variants (4 paths)

### LINKED ISSUE #2095: GET /v1/pull/variants has zero test coverage — missing, malformed, and 404 error paths are completely untested
### Platform

Windows

### Lemonade Version

any

### GPU / APU Model

this is a test infrastructure bug, no GPU involved

### Component

Other / Not sure

### Bug Description

The GET /v1/pull/variants endpoint is fully implemented in server.cpp (handle_pull_variants, line ~3032) and documented in docs/api/lemonade.md, but has zero test coverage across the entire test suite. No file in test/ contains a single call to this endpoint. Three distinct error-handling code paths exist in the C++ implementation and none are exercised by any test, meaning a regression could be introduced and the full test suite would still pass green.

### Steps to Reproduce

1. Search all files under test/ for "pull/variants" — zero results
2. Open src/cpp/server/server.cpp and find handle_pull_variants() (~line 3032)
3. Note three implemented error paths:
   - Missing checkpoint param → 400
   - Malformed checkpoint (no slash) → 400
   - Non-existent HF repo → 404
4. Run the full test suite — all tests pass
5. Delete or break any of those three if/return blocks in server.cpp
6. Run the test suite again — all tests still pass

Expected: test/server_endpoints.py contains tests that call GET /v1/pull/variants with a missing checkpoint param (assert 400), a malformed checkpoint without / (assert 400), a non-existent HuggingFace repo (assert 404), and a valid checkpoint (assert 200 with variants, checkpoint, recipe fields in the response).
Actual: No such tests exist. The endpoint has zero test coverage despite being a documented, production-facing endpoint used by the desktop app and CLI install flow.

### Log Output

```shell
this is a test coverage gap, not a runtime crash.
```

### Additional Context

The fix is purely additive — no server-side code changes needed. Four tests covering all code paths can be added to test/server_endpoints.py. The happy-path test can use TheBloke/TinyLlama-1.1B-Chat-v1.0-GGUF as a reliable, permanently-public small HuggingFace repo. The three error-path tests require no network access and run instantly.

Kushal1213:
Summary
`GET /v1/pull/variants` is a documented, production endpoint used by the 
desktop app and CLI to enumerate GGUF quantizations from HuggingFace. 
The three error paths are fully implemented in `handle_pull_variants()` 
(server.cpp ~line 3032) but have zero test coverage across the entire 
test suite.

 Changes
Adds 4 tests to `test/server_endpoints.py`:
- `test_030` — missing `checkpoint` param → 400
- `test_031` — malformed checkpoint (no `/`) → 400  
- `test_032` — nonexistent HuggingFace repo → 404
- `test_033` — valid checkpoint → 200 with correct response shape

Closes #2095


fl0rianr (changes requested):
Thanks for adding coverage here — this is the right area to test, and the missing/malformed/not-found paths map directly to `handle_pull_variants()`.

I’d request a few changes before merging:

1. Please avoid making the default endpoint suite depend on live Hugging Face availability. Both the nonexistent-repo and valid-checkpoint tests can hit the network through `fetch_pull_variants()`. That makes CI vulnerable to rate limits, network policy, HF outages, repo changes, or first-party API behavior changes. Prefer mocking/faking the HF response, or gate the live happy-path behind an explicit integration-test env var.

2. Please add `pull/variants` to the existing endpoint-registration list as well, so route registration is covered consistently with the other endpoints.

3. Please strengthen the assertions:
   - missing checkpoint: assert the specific “Missing required query parameter” message
   - malformed checkpoint: assert the specific malformed-checkpoint / `owner/name` message
   - valid response: assert `checkpoint == input`, expected `recipe`, non-empty variants, and all documented variant fields including `size_bytes`

The PR is directionally good, but I wouldn’t merge it as-is because the happy path is live-network-dependent and the response-shape assertions are weaker than the documented contract.

Kushal1213:
Done — addressed all three points in the latest commit:

test_032 and test_033 are now gated behind LEMONADE_INTEGRATION_TESTS=1
pull/variants added to the test_000 registration list
Error messages now assert exact strings from handle_pull_variants(), and test_033 validates the full response shape including size_bytes

fl0rianr:
Thanks for the update — this addresses my previous blockers.

The default endpoint suite no longer depends on live Hugging Face availability, pull/variants is now covered by the endpoint-registration check, and the assertions are much stronger.

Only two non-blocking nits:
- test_031 says it checks the exact malformed-checkpoint error, but currently only asserts the `owner/name` hint. Consider also asserting Malformed checkpoint, or loosen the docstring.
- Minor style nit: `unittest` could be moved into sorted stdlib import order.

Code-wise this looks good to me now. I’d merge once we have this CI issues solved that are troubling us. The final potential solutions are PR'ed.

fl0rianr (approved):
Thanks and we can go with this.

PR #2098, opened 2026-06-04
docs: specify DEB generator for Linux package

ynott:
## Summary

This PR updates the Linux .deb packaging instructions to explicitly use:

```bash
cpack -G DEB
```

instead of:

```bash
cpack
```

## Background

When I tried to build a Debian package by following the current documentation, `cpack` alone did not generate a `.deb` package in my environment.

Because this section is specifically for Debian/Ubuntu `.deb` packages, I think explicitly specifying the DEB generator makes the instruction clearer and avoids relying on CPack's default generator behavior.

I hope this helps other users who may run into the same issue.

### DIFF
--- a/docs/dev/getting-started.md
+++ b/docs/dev/getting-started.md
@@ -230,7 +230,7 @@ Creates `lemonade-server-minimal.msi` which:
 
 ```bash
 cd build
-cpack
+cpack -G DEB
 ```
 
 **Package Output:**


fl0rianr (commented):
Good catch — explicitly selecting DEB makes the CPack command deterministic. One concern: the surrounding docs still claim a Debian-style output name (lemonade-server_<VERSION>_amd64.deb), but the current CPack config appears to inherit CPACK_PACKAGE_FILE_NAME from CPackRPM.cmake, which may produce a different filename unless CPACK_DEBIAN_FILE_NAME=DEB-DEFAULT is set. Could we either verify the actual generated filename or update the documented package output accordingly?

superm1 (changes requested):
I don't want us encouraging cpack for Deb. We migrated to native packaging.

ynott:
Thanks for the clarification.

Ah, I see — that makes sense. I was trying to fix the Debian package build because it used to work for me but no longer did. If the project has moved to native Debian packaging, I agree that is the better approach.

I'll withdraw this PR rather than continuing with the CPack-based approach.

PR #2099, opened 2026-06-04
Unified collection import/export format; LMX collections distributed via Hugging Face

jeremyfowers:
## What

Makes omni collections (`recipe: collection.omni`) behave like **virtual models**: they import/export through the exact same pipeline as regular models, using **one JSON format** shared by the server, GUI, CLI, and Hugging Face.

### The format

A collection file is the regular-model export format plus `components` (ordered names) and `models` (each component's full definition, normalized by the same per-model transform):

```json
{
    "model_name": "user.LMX-Omni-5.5B-Lite",
    "recipe": "collection.omni",
    "checkpoints": { "main": "lemonade-sdk/LMX-Omni-5.5B-Lite" },
    "components": ["Qwen3.5-4B-MTP-GGUF", "SD-Turbo", "Whisper-Tiny", "kokoro-v1"],
    "models": [ { "model_name": "Qwen3.5-4B-MTP-GGUF", "recipe": "llamacpp", "checkpoints": { "...": "..." }, "labels": ["..."], "size": 3.66 }, "..." ],
    "labels": [],
    "recipe_options": {},
    "size": 9.3
}
```

The same file works verbatim as: a `POST /v1/pull` body, a `lemonade import` file, and an HF-hosted manifest (`<CollectionName>.json`).

### Example Model Files

https://huggingface.co/lemonade-sdk/LMX-Omni-52B-Halo
https://huggingface.co/lemonade-sdk/LMX-Omni-5.5B-Lite

### Changes

- **Server read side**: `/v1/models` and `/v1/models/{id}` embed an ordered `models[]` array for collections (each component's full model object, parallel to `components`).
- **Server pull side**: `/v1/pull` accepts the exported file. Unknown components are registered from the embedded `models[]` definitions; known names keep the local definition (local-wins, drift logged as a warning). The existing component fan-out, cycle guard, and progress streaming are unchanged.
- **HF distribution**: the built-in `LMX-Omni-52B-Halo` / `LMX-Omni-5.5B-Lite` entries are slim stubs (`checkpoint` = HF repo); pulling fetches the manifest from the repo by the documented filename convention — the repo's `<RepoName>.json` (e.g. `LMX-Omni-5.5B-Lite.json`) — and then validates its content (an exported collection JSON with `recipe: collection.omni`, a non-empty `components` array, and a `models` array). The filename is the discovery key; content is checked only to confirm the hit. Manifests cache in the HF cache like any model file, so collections survive restarts and work offline after first pull.
- **CLI**: `lemonade export`/`import` work for collections via the shared `validate_and_transform_model_json` transform (`id`→`model_name`, `kKnownKeys` allow-list, checkpoint dedup, per-element normalization). Regular-model export output is byte-identical to before.
- **GUI**: collection and regular-model export both fetch `/models/{id}` and save the normalized file (shared helper mirroring the CLI transform); collection import is a single `POST /pull`; export filenames drop the `user.` prefix; the legacy `{version, collections, models}` bundle format is removed.
- **Exports strip user-specific state**: `suggested`, `created`, `downloaded` never appear in files (top level or inside `models[]`); the server regenerates them on import.
- **Docs**: format documented on `/v1/pull` (lemonade.md, incl. a "Collection Files: Export, Import, and Hugging Face" section) and the model object (openai.md).

### Breaking changes

- Previously exported GUI collection bundles (`{version, exportedAt, collections, models}`) no longer import (hard cutover; the GUI was the only producer/consumer).

## Verification

- `server_endpoints.py`: 50/50 OK · `server_cli2.py`: 58/58 OK (1 skipped) · `server_omni.py --wrapped-server llamacpp --backend vulkan`: 7/7 OK (pre-pulls `LMX-Omni-5.5B-Lite` through the HF manifest path, then drives the server-side orchestration loop)
- App regression suite: 31/31; collection unit tests rewritten for the new transform
- Round-trips verified live: export ≡ uploaded HF manifest (JSON-equal); fresh-cache pull of both LMX collections; import of a collection file with an unknown component (registered as `user.*` from the embedded def, downloaded, survives restart); verbatim `POST /pull` of an exported file
- Builds: C++ (Ninja Release) and web-app clean


src/cpp/server/model_manager.cpp
```diff
@@ -2696,6 +2721,239 @@ void ModelManager::download_registered_model(const ModelInfo& info, bool do_not_
...
+                LOG(WARNING, "ModelManager") << "Skipping collection component with "
+                    << "reserved name: " << name << std::endl;
+                continue;
+            }
+            LOG(INFO, "ModelManager") << "Registering collection component as user model: "
+                << canonical << std::endl;
+            register_user_model(canonical, def);
+            components.push_back(canonical);
+        } else {
+            LOG(WARNING, "ModelManager") << "Skipping unknown collection component with no "
```
fl0rianr, line 3248: This should probably be a hard import failure for inline collection files, not a warning/skip. validate_collection_request() only checks that models is an array; after that, any unknown component without a matching valid definition is skipped here. If at least one component remains, the collection can be persisted with a shortened component list, which is not the same collection the exported file described.

Can we require models to match components exactly for inline imports (same names, no missing definitions, no invalid entries), and abort/rollback instead of silently dropping components?

src/cpp/server/hf_variants.cpp
```diff
@@ -284,9 +285,71 @@ nlohmann::json fetch_pull_variants(const std::string& checkpoint, bool& not_foun
         return out;
     }
 
+    // Omni collection branch: a repo published from `lemonade export` carries
+    // its collection manifest as <RepoName>.json (see "Share a collection" in
+    // the custom models guide). The filename is the discovery key; the content
+    // is then validated — a same-named file that is not a manifest is an
+    // error, not a fall-through.
+    const std::string manifest_filename = suggested_name + ".json";
```
fl0rianr, line 297: The implementation here requires the manifest to be named exactly <RepoName>.json. That matches the new user docs, but the PR description says HF manifests are discovered "by content" as the repo's *.json with `recipe: collection.omni`.

fl0rianr (changes requested):
Thanks, this is a good direction overall: sharing one import/export shape across the server, CLI, GUI, and HF manifests should simplify the collection story a lot.

Before merge let's fix two correctness issues:

1. I agree with the existing Codex inline comment on `src/cpp/server/model_manager.cpp:3139`: when an already-registered HF-backed user collection is refreshed from a changed manifest, the refreshed component list is used for the current download but is not persisted back into user_models.json. After cache invalidation/rebuild, the registered collection can still expose/load the stale components. Please persist the canonical refreshed `components` for `user.*` collections after manifest resolution, and add a regression test for manifest A/B -> A/B/C refresh.

2. Inline collection imports should fail closed when components and models do not match. The format is documented as ordered/parallel; silently dropping missing or invalid component definitions can create a different collection than the exported file describes.

src/cpp/server/model_manager.cpp
jeremyfowers: Fixed in c8061f62d. `validate_collection_request` now checks resolvability **per-component** instead of treating a present `models` array as a blanket pass: every component must either already be a registered model or have a matching `models` definition (mirroring `register_components`'s bare-name / `model_name`|`id` matching). Unresolvable components are rejected with 400 *before* any registration or download side effect, so nothing is persisted — no mid-`register_components` rollback needed, and the `:2909` skip stays only as a backstop for the HF manifest path.

I kept the rule as per-component resolvability rather than strict `models == components` equality, so a component that resolves to an already-registered model (e.g. a built-in) without an inline def still round-trips. Regression test added: `test_021t_inline_collection_missing_def_rejected` (components `[A, B]`, only `A` defined → 400, collection not persisted).

_(Reply written by Claude on Jeremy's behalf.)_

src/cpp/server/hf_variants.cpp
jeremyfowers: Fixed the PR description

src/cpp/server/model_manager.cpp
```diff
@@ -2848,10 +3110,51 @@ void ModelManager::download_model(const std::string& model_name,
...
+                    register_user_model(model_name, reg);
+                }
+            } else if (!repo_id.empty() && (components.empty() || !do_not_upgrade)) {
+                // HF-backed collections keep their full definition — the component list
+                // and each component's model object — on Hugging Face as an exported
+                // collection JSON. A non-empty checkpoint marks such a collection; fetch
+                // the manifest to learn its components. On an explicit pull
+                // (do_not_upgrade=false) always refresh so newly added components are
+                // picked up.
+                components = resolve_collection_components_from_manifest(repo_id, do_not_upgrade);
```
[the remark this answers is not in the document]
jeremyfowers: This is correct by design. Added [4299874](https://github.com/lemonade-sdk/lemonade/pull/2099/commits/42998744428a219408a5441bf47bff8424bac584) to help to clarify.

src/cpp/server/model_manager.cpp
```diff
@@ -4386,7 +4736,16 @@ std::optional<std::string> ModelManager::validate_collection_request(
         if (component_name == model_name) {
             return "Collection cannot reference itself: " + component_name;
         }
-        if (!model_exists(component_name)) {
+        std::string bare = bare_component_name(component_name);
+        if (has_models_array) {
+            if (!model_exists(bare) && !has_inline_def(bare)) {
```
fl0rianr: This still only checks that a matching inline object exists, not that the inline object is a valid model registration. register_components() will persist any matching object as user.* , even if it is missing recipe or checkpoint(s), and the failure then happens later during component download after side effects have already occurred.

Can we validate the matched models entry with the same minimum requirements as a regular model registration before accepting it as resolvable?


Addon-comment:

This self-reference check should compare canonical/bare forms too. For example, a new collection named user.MyCol with `components: ["MyCol"]` can bypass this exact string check in inline-import mode, then resolve back to itself after the early collection registration.

Could we reject `bare_component_name(component_name) == bare_component_name(model_name)` as well?

fl0rianr (changes requested):
One remaining concern: the new persistence comment says HF-backed collections persist only the checkpoint pointer and rebuild components from the HF cache, but the current hf_pull path still sends the full manifest into /pull, and register_user_model() persists components because it is in USER_DEFINED_MODEL_PROPS.

If the intended design is pointer-only for HF-backed collections, we should strip components/models before persisting the collection entry and ensure user-model cache rebuilds can populate components from the cached manifest too. Or did I misread the intention?

src/cpp/server/model_manager.cpp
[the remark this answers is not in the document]
jeremyfowers: Good catch — on closer look this was a real bug, not by-design, so I revised the earlier reply. Fixed in f0fb95083 by making HF-backed collections *truly* pointer-only:

- `register_user_model` now strips `components` when a collection carries a checkpoint pointer, so the registry stores only the repo pointer.
- `build_cache` rebuilds the component list from the cached manifest whenever a repo pointer is present (not just when `components` is empty), so a refreshed manifest always wins and any previously-persisted stale list self-heals.

The key gap behind the staleness you flagged: `populate_collection_components_from_cache_locked` was only called in the **server-models** loop, so **user** HF-backed collections (the `lemonade pull <org>/<repo>` → `user.*` case) never repopulated from cache. That populate is now done in the user-models loop too.

Regression test `test_021w_hf_backed_collection_refresh_is_pointer_only` covers the A→A,B manifest refresh: registers a collection, asserts components come from the cache and are not persisted, mutates the cached manifest to add a component, re-pulls, and asserts the new component is reflected.

src/cpp/server/model_manager.cpp
jeremyfowers: Both addressed in f0fb95083.

**Validity check:** added `collection_component_def_is_valid()` (recipe + at least one checkpoint for a leaf model, or a non-empty `components` array for a nested collection — mirroring a regular registration). `validate_collection_request` now rejects a component that resolves to an inline def which is present-but-incomplete, before any side effect:

> Inline collection component '<name>' has an incomplete definition in 'models' (a recipe and at least one checkpoint are required).

**No leaked `user.*` on partial failure:** `register_components` is now two-pass — it validates every unknown component's def first and only registers once the whole list is known good, so a half-defined component can no longer persist a stray `user.*` entry that a later failure would have to roll back. (The HF-manifest path goes through `register_components` too, so it gets the same guard even though it doesn't pass through `validate_collection_request`.)

**Bare-form self-reference:** the self-reference check now compares `bare_component_name(component) == bare_component_name(model_name)`, so `user.MyCol` with `components: ["MyCol"]` is rejected, not just the exact `user.`-qualified spelling.

Tests: `test_021u_inline_collection_invalid_def_rejected` (incomplete def → 400, neither collection nor component persisted) and `test_021v_collection_self_reference_bare_name_rejected`.

jeremyfowers:
Pushed f0fb95083 addressing the remaining review feedback. Summary:

### 1. HF-backed collection refresh — now truly pointer-only (@fl0rianr review #1 & #2, Codex P2)

You were right that the previous state was inconsistent: the code *claimed* pointer-only but `components` was in `USER_DEFINED_MODEL_PROPS` and got persisted, so a refreshed manifest could be shadowed by a stale list. It's now genuinely pointer-only:

- `register_user_model` strips `components` for any collection that carries a checkpoint pointer — the registry entry stores only the repo pointer.
- `build_cache` rebuilds the component list from the cached manifest whenever a repo pointer is present (not just when empty), so a refreshed manifest always wins and any previously-persisted stale list self-heals on next rebuild.
- **Root cause of the staleness:** `populate_collection_components_from_cache_locked` was only called in the server-models loop, so user `user.*` HF-backed collections never rebuilt from cache. It's now called in the user-models loop too.

Regression test added: `test_021w_hf_backed_collection_refresh_is_pointer_only` (A→A,B manifest refresh; asserts components come from cache, are never persisted, and the added component appears after re-pull).

### 2. Inline imports fail closed (@fl0rianr review #1 + the `:4741` thread)

- Inline component defs are now validated (recipe + at least one checkpoint, or a non-empty `components` array for a nested collection) before acceptance — `validate_collection_request` rejects present-but-incomplete defs with 400.
- `register_components` is now two-pass (validate-all-then-register) so a half-defined component never leaks a stray `user.*` entry.
- Self-reference now compares bare names, closing the `user.MyCol` + `["MyCol"]` bypass.

Tests: `test_021u`, `test_021v`.

### Verification

Full collection suite (`server_endpoints.py` `test_021*`): 23/23. C++ build clean.

Replies left inline on the Codex P2 and `:4741` threads. @fl0rianr — ready for another look when you have a moment.

fl0rianr:
@jeremyfowers Thanks! I think the pointer-only behavior still needs one more guard for the real "lemonade pull <org>/<repo>" path.

In that path, fetch_pull_variants() reads the manifest via HttpClient::get() and returns it in the variants response, then hf_pull.cpp sends that manifest as an inline /pull body with checkpoints.main set to the repo id. Because download_model() then takes the inline models/components branch, it does not call resolve_collection_components_from_manifest() and therefore may never seed the HF cache with the collection manifest.

After this block erases "components", the user registry entry depends on a cached manifest being present. On the next cache rebuild/restart, the collection can rebuild to an empty component list if the manifest was never cached. The new test pre-populates the fake HF cache before registration, so it does not exercise this real hf_pull path.

Could we either ensure the manifest is cached before erasing components, or keep components persisted until the manifest has actually been cached?

jeremyfowers:
@fl0rianr

Good catch — fixed in 2a22377b3 by separating the two concerns: **`/pull/variants` inspects, `/pull` downloads.**

`fetch_pull_variants` now only inspects the repo (confirms it's a collection, returns name/size/count) and no longer ships the manifest inline. `hf_pull` sends a pointer body (`model_name` + `recipe` + `checkpoint = repo`), and `/pull` downloads `<RepoName>.json` to disk and resolves components from it — so the manifest is actually pulled, exactly like the `.gguf` for a regular model. `build_cache` then rebuilds from the on-disk manifest, and if the manifest can't be downloaded the pull fails and nothing is registered.

`test_021w` now drives the real pointer body (no inline components/models); the real network download of the manifest is covered by `server_omni.py`.

fl0rianr (commented):
One possible edge case before I approve: I know we talked about this, just to be sure and take it here to GH: Are nested collections intended to be supported by the exported/imported format?

/models/{id} now embeds collection component models recursively, so an exported file can contain A -> B(collection) -> C(leaf). But the import path appears to process only the top-level components/models pair. If B is unknown, it gets registered as a collection, but its embedded models[] are not recursively registered, so C may be missing on a fresh machine.

If nested collections are out of scope, I think we should reject collection components in inline definitions or document that limitation. If they are supported, the import path probably needs recursive registration.

jeremyfowers:
> One possible edge case before I approve: I know we talked about this, just to be sure and take it here to GH: Are nested collections intended to be supported by the exported/imported format?
> 
> /models/{id} now embeds collection component models recursively, so an exported file can contain A -> B(collection) -> C(leaf). But the import path appears to process only the top-level components/models pair. If B is unknown, it gets registered as a collection, but its embedded models[] are not recursively registered, so C may be missing on a fresh machine.
> 
> If nested collections are out of scope, I think we should reject collection components in inline definitions or document that limitation. If they are supported, the import path probably needs recursive registration.

They should probably be supported eventually, but its adding a lot of complexity to this feature so I think I'll just block nested collections for now.

fl0rianr (approved):
Ok, fine by me. Sadly CI is messy again I would prefer to trust on it.

jeremyfowers:
@fl0rianr Out of scope for now — nested collections are rejected at registration in 714868dac (a component that is itself a collection, by inline def or by resolving to a registered one, is 400'd; same check backstops the HF-manifest path). Tests `test_021x`/`test_021y`. So no recursive registration needed.

PR #2102, opened 2026-06-04
update sd-cpp backend

fl0rianr:
Update to 672-1f9ee88 at upstream repo.

Tested on cpu, rocm and vulkan.


fl0rianr:
Tested it on  NVIDIA as well, made it a draft since i thought there was an error, was only a typo.

jeremyfowers:
Testing the new SD version on a local windows PC to see if I can reproduce the problem.

jeremyfowers (approved):
Works on windows systems with updated drivers!

PR #2105, opened 2026-06-04
ci: consolidate GitHub-hosted test jobs to reduce download rate limiting

### LINKED ISSUE #2103: Optimize CI jobs
Goal: reduce the amount of GH and HF downloads per CI run.
 
1. Self-hosted runners should enable HF caching (right now they delete all models between jobs)
2. Github-hosted tests should consolidate. Instead of running 1 test script per job, each job should run many test scripts. That will share model downloads across the scripts, which will massively reduce total downloads per PR.

jeremyfowers:
## Summary

Implements both ideas from #2103 to reduce GitHub and Hugging Face download rate limiting in CI, wires up the previously-unused `HUGGINGFACE_ACCESS_TOKEN` secret, and removes the env-vars test script.

### 1. Consolidate GitHub-hosted test jobs (idea 2)

Each job now runs many test scripts against one install and one shared model cache, instead of one script per matrix job. This cuts the HF/GitHub download request sets per PR for these clusters from ~20 to ~6.

| Cluster | Jobs before → after | Projected duration |
|---|---|---|
| `test-cli-endpoints-linux` | 6 → 1 | ~6 min |
| `test-cli-endpoints` (windows) | 5 → 1 | ~6 min |
| `test-cli-endpoints` (macos) | 6 → 1 | ~9 min |
| `test-dmg-inference` | 3 → 1 | ~1 min |

Projections are from step-level timing analysis of the last 5 successful main runs: per-job redundant setup (checkout, container/artifact install, venv) is 0.9–1.4 min repeated 22×, while unique test time is 0.1–4.2 min per script. Billed runner-minutes for these clusters drop ~39 → ~24, and no consolidated job gets long enough to extend the PR critical path (still bounded by the 18-min Windows installer build).

- Each test script is its own step with `if: !cancelled() && steps.setup.outcome == 'success'`: when one test fails the remaining tests still run (preserving the old `fail-fast: false` matrix behavior), the job still fails, and per-script pass/fail shows natively in the Actions UI. Setup failures still skip the tests.
- Step order is `cli → endpoints → ollama → streaming-errors → llamacpp-system`; llamacpp-system runs last because its tearDown stops the shared lemond via `/internal/shutdown` to launch its own instance.
- API-key jobs, embeddable/rpm smoke tests, and self-hosted jobs keep their current shape.

### 2. Persist HF model cache on self-hosted runners (idea 1)

`test-exe-inference` and `test-deb-inference` set `HF_HOME` inside the GitHub workspace, and the checkout clean (`git clean -ffdx`) wiped it at the start of every job — so self-hosted runners re-downloaded all test models every run. `HF_HOME` now points at a persistent per-runner location outside the workspace (`~/.cache/lemonade-ci/huggingface`).

### 3. Wire up `HUGGINGFACE_ACCESS_TOKEN`

The secret existed but was never referenced by any workflow, so every lemond instance and test script hit HF anonymously, rate-limited by runner IP. `HF_TOKEN` is now set in the job env of every job that downloads models (8 jobs in the main workflow incl. the inline tests in `build-lemonade-macos-dmg`, plus `server_download_registry.yml`, `validate_llamacpp.yml`, `validate_vllm.yml`). lemond reads it and attaches `Authorization: Bearer`; `huggingface_hub` picks it up automatically.

Includes a one-line guard in `model_manager.cpp`: fork PRs resolve secrets to empty strings, and a set-but-empty `HF_TOKEN` previously produced a malformed `Authorization: Bearer ` header (`hf_variants.cpp` already had this guard).

Known limitation: the signed-macOS `.pkg` path starts lemond via launchd, which does not inherit the job env — Python-side downloads still get the token there.

### 4. Remove `test/server_env_vars.py`

Deleted the script and its CI steps (consolidated Linux/macOS jobs and the inline step in `build-lemonade-macos-dmg`). It spawned its own lemond instances mid-job, which raced the shared server's teardown for port 13305 in the consolidated jobs (silent 60s timeout, flaky on macOS). This drops its coverage: `/internal/config` reflection of `LEMONADE_*` config env vars, default-value checks, and the env-var-based API/admin-key variants. API-key enforcement is still covered end-to-end by the dedicated `test-cli-apikey*` jobs.

## ⚠️ Before merging

Branch protection on `main` requires 10 checks that no longer exist after this change (`Test cli/endpoints/ollama (ubuntu|windows|macos-latest)`, `Test llamacpp-system (ubuntu-latest)`). Replace them with the 3 new names: `Test CLI/Endpoints (ubuntu-latest)`, `Test CLI/Endpoints (windows-latest)`, `Test CLI/Endpoints (macos-latest)`.

Closes #2103


fl0rianr (approved):
Looks good to me.

I reviewed the consolidation and don’t see a blocking issue. The test coverage appears preserved, the individual test scripts remain visible as separate Actions steps, and the !cancelled() gating keeps the old fail-fast=false behavior reasonably well.

fl0rianr:
Looks like the remaining failure is a timeout rather than a functional test failure.

Given that these jobs now run multiple formerly separate matrix tests sequentially, I think increasing the timeout is reasonable and not masking coverage. I’d bump the consolidated hosted test jobs a bit, rerun, and merge if the rerun is green.

jeremyfowers:
> Looks like the remaining failure is a timeout rather than a functional test failure.
> 
> Given that these jobs now run multiple formerly separate matrix tests sequentially, I think increasing the timeout is reasonable and not masking coverage. I’d bump the consolidated hosted test jobs a bit, rerun, and merge if the rerun is green.

The "timeout" was because one test killed lemond while another test expected it to still be alive.

I'm hoping to solve it by ordering the tests correctly in https://github.com/lemonade-sdk/lemonade/pull/2105/commits/43725f0420377f4f11dad3b8145aac025656443f

If that doesn't work I'll put in lifecycle management.

PR #2106, opened 2026-06-05
Drop all environment variable migration code

superm1:
At this point people have had config.json for a few months.  This is a ton of dead code.

### DIFF
--- a/docs/guide/configuration/llamacpp.md
+++ b/docs/guide/configuration/llamacpp.md
@@ -52,7 +52,7 @@ Lemonade uses [llama.cpp](https://github.com/ggerganov/llama.cpp) as its primary
 - **Use Case**: Advanced users with custom llama.cpp builds
 - **Performance**: Depends on build configuration
 - **Installation**: Requires manual installation of `llama-server` in system PATH
-- **Notes**: Not enabled by default; set `LEMONADE_LLAMACPP_PREFER_SYSTEM=true` in config
+- **Enable**: Set `"llamacpp": { "backend": "system" }` in `config.json` to enable the system backend, or use `lemonade config set llamacpp.backend=system`
 - **HIP plugin on non-standard paths**: When an AMD GPU is present, the system backend needs the GGML HIP plugin (`libggml-hip.so`). Lemonade looks for it in the standard system library paths. If your distribution or package manager installs it elsewhere (e.g. NixOS, a custom prefix, or a manual build), set `LEMONADE_GGML_HIP_PATH` to the full path of the plugin so the backend is reported as available:
 
   ```bash
--- a/docs/guide/install/docker.md
+++ b/docs/guide/install/docker.md
@@ -25,7 +25,6 @@ docker run -d \
   -v lemonade-cache:/root/.cache/huggingface \
   -v lemonade-llama:/opt/lemonade/llama \
   -v lemonade-recipe:/root/.cache/lemonade \
-  -e LEMONADE_LLAMACPP=cpu \
   ghcr.io/lemonade-sdk/lemonade-server:v9.1.3 \
   ./lemond --host 0.0.0.0 --port 5000
 ```
@@ -34,6 +33,18 @@ docker run -d \
 
 ### Docker Run with CPU backend
 
+To use the CPU backend, create or modify the `config.json` file in the `lemonade-recipe` volume:
+
+```json
+{
+  "llamacpp": {
+    "backend": "cpu"
+  }
+}
+```
+
+Then run:
+
 ```bash
 docker run -d \
   --name lemonade-server \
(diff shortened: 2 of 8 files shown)


fl0rianr (commented):
Thanks for the follow-up docs updates — this is definitely moving in the right direction.

One thing I’d like to double-check: some of the new replacement guidance seems to document options/keys that `lemond` may not actually support today.

For example, the Debate Arena docs now suggest `./lemond --max-loaded-models 9`, but I only see lemond parsing cache_dir, --port, and --host. Also, the Docker notes still describe LEMONADE_LLAMACPP=cpu|rocm as a runtime override, but this PR removes the env-to-config path, so I’m not seeing where `lemond` would still consume that as a server backend override. Finally, the Docker note says to set `llamacpp_backend: cpu` in config.json; should that instead be the nested config shape, e.g. `"llamacpp": { "backend": "cpu" }`?

Could you sanity-check those replacement instructions before merge?

superm1:
> Thanks for the follow-up docs updates — this is definitely moving in the right direction.
> 
> One thing I’d like to double-check: some of the new replacement guidance seems to document options/keys that `lemond` may not actually support today.
> 
> For example, the Debate Arena docs now suggest `./lemond --max-loaded-models 9`, but I only see lemond parsing cache_dir, --port, and --host. Also, the Docker notes still describe LEMONADE_LLAMACPP=cpu|rocm as a runtime override, but this PR removes the env-to-config path, so I’m not seeing where `lemond` would still consume that as a server backend override. Finally, the Docker note says to set `llamacpp_backend: cpu` in config.json; should that instead be the nested config shape, e.g. `"llamacpp": { "backend": "cpu" }`?
> 
> Could you sanity-check those replacement instructions before merge?

thanks for checking, yeah I'll follow up.

superm1:
> > Thanks for the follow-up docs updates — this is definitely moving in the right direction.
> > One thing I’d like to double-check: some of the new replacement guidance seems to document options/keys that `lemond` may not actually support today.
> > For example, the Debate Arena docs now suggest `./lemond --max-loaded-models 9`, but I only see lemond parsing cache_dir, --port, and --host. Also, the Docker notes still describe LEMONADE_LLAMACPP=cpu|rocm as a runtime override, but this PR removes the env-to-config path, so I’m not seeing where `lemond` would still consume that as a server backend override. Finally, the Docker note says to set `llamacpp_backend: cpu` in config.json; should that instead be the nested config shape, e.g. `"llamacpp": { "backend": "cpu" }`?
> > Could you sanity-check those replacement instructions before merge?
> 
> thanks for checking, yeah I'll follow up.

Fixed.

fl0rianr:
Thanks, the Docker and Debate Arena updates look much better now.

I see one remaining docs nit and one compatibility question before I’d be fully comfortable approving:

* docs/guide/configuration/llamacpp.md now says to set `llamacpp_backend: system` in config.json. Should that be the nested config shape instead, e.g. `"llamacpp": { "backend": "system" }`, or the CLI form `lemonade config set llamacpp.backend=system`?
* This still removes the rocm=preview / rocm_channel=preview compatibility migration. Is dropping that old config migration intentional as part of this cleanup? I’m asking because ROCm backend resolution still builds rocm-${channel}, while the versions file only has rocm-stable / rocm-nightly.

Once those are clarified/fixed, this looks good to me.

jeremyfowers:
The env var tests are causing CI issues so I'm going to drop them here ASAP https://github.com/lemonade-sdk/lemonade/pull/2105 since we're getting rid of the support anyways.

superm1:
> This still removes the rocm=preview / rocm_channel=preview compatibility migration. Is dropping that old config migration intentional as part of this cleanup? I’m asking because ROCm backend resolution still builds rocm-${channel}, while the versions file only has rocm-stable / rocm-nightly.

Yeah, I think we should drop all our migrations.  They've been around long enough IMO.

> docs/guide/configuration/llamacpp.md now says to set llamacpp_backend: system in config.json. Should that be the nested config shape instead, e.g. "llamacpp": { "backend": "system" }, or the CLI form lemonade config set llamacpp.backend=system?

Fixed

fl0rianr:
Thanks!
I re-checked current head e349339. llamacpp.md is fixed, thanks. However, Docker still documents LEMONADE_LLAMACPP=cpu|rocm as a runtime backend override and still suggests llamacpp_backend in config.json, and Debate Arena still documents ./lemond --max-loaded-models 9. I don’t see those supported after this PR: env-to-config migration is removed, and lemond only parses cache_dir, --port, and --host. I’d hold off approving until those replacement instructions are corrected.

fl0rianr:
Thanks for adapting further!

I see one small remaining docs issue before I’d be fully comfortable approving: examples/debate-arena.md and examples/llm-debate.html still suggest ./lemond --max-loaded-models 9, but lemond appears to only accept cache_dir, --port, and --host. Could we remove that CLI-option suggestion and keep the supported config.json / /internal/set guidance there?

Everything else I raised looks addressed.

fl0rianr (approved):
Looks good to me now. Thanks for iterating on this!

jeremyfowers:
Breaking change added to draft release notes https://github.com/lemonade-sdk/lemonade/issues/2136

PR #2107, opened 2026-06-05
Show all local model versions in one folder

### LINKED ISSUE #1992: extra_models_dir doesn't work when you have two variants of a model in the same directory
### Platform

Windows

### Lemonade Version

10.5.1

### GPU / APU Model

Not relevant

### Component

Desktop App / UI

### Bug Description

extra_models_dir doesn't work when you have two variants of a model in the same directory e.g.   Qwen3.6-35B-A3B-Q4_K_M.gguf   and  Qwen3.6-35B-A3B-Q8_0.gguf  which are both variants of same model, so for example with LM Studio the automatic folder layout would be both files in /lmstudio-community/Qwen3.6-35B-A3B-GGUF folder.   If so then lemonade only sees one of the models, so you have to move them to separate folders instead

Just confirmed same issue in Windows

Image

### Steps to Reproduce

extra_models_dir doesn't work when you have two variants of a model in the same directory e.g.   Qwen3.6-35B-A3B-Q4_K_M.gguf   and  Qwen3.6-35B-A3B-Q8_0.gguf  which are both variants of same model, so for example with LM Studio the automatic folder layout would be both files in /lmstudio-community/Qwen3.6-35B-A3B-GGUF folder.   If so then lemonade only sees one of the models, so you have to move them to separate folders instead

Just confirmed same issue in Windows

Image

_No response_

### Log Output

```shell

```

### Additional Context

_No response_

anditherobot:
Fixes #1992.

Users can keep multiple GGUF versions of a model, such as Q4 and Q8, in the same local folder. Lemonade previously treated each `extra_models_dir` subfolder as one model, so only one version was selectable.

This change lists each independent GGUF version from that folder as its own model choice. The old folder name still works for existing scripts and aliases, pointing to the same file Lemonade would have selected before.

Sharded models stay grouped as one model, and `mmproj` vision files stay attached instead of appearing as standalone models.

| Case | Before | After |
|---|---|---|
| Folder has `Model-Q4.gguf` and `Model-Q8.gguf` | Only one model is shown | Q4 and Q8 are both shown |
| Existing scripts use the folder name | Folder name points to the old single model | Folder name still works |
| Folder has sharded files like `part-00001-of-00002.gguf` | Folder is treated as one model | Still treated as one model |
| Folder has an `mmproj` file | Attached to the model | Still attached, not listed separately |

Tests:
- `test_021t_extra_subdir_multiple_quantization_variants_emit_separate_models`
- `test_021u_extra_subdir_sharded_models_remain_grouped`

Verification:
- `cmake --build --preset windows --target lemond`
- `pytest test/server_endpoints.py::EndpointTests::test_021t_extra_subdir_multiple_quantization_variants_emit_separate_models test/server_endpoints.py::EndpointTests::test_021u_extra_subdir_sharded_models_remain_grouped`


src/cpp/server/model_manager.cpp
```diff
@@ -1101,61 +1097,121 @@ std::map<std::string, ModelInfo> ModelManager::discover_extra_models() const {
...
+    if (main_model_path.empty()) return;
+
+    auto vset = lemon::enumerate_gguf_variants(model_filenames, model_file_sizes);
+
+    // Split only when variant detection cleanly identifies separate single-file
+    // quants. Shards or unmatched files fall back to folder grouping so no GGUF
+    // parts disappear from discovery.
+    bool should_split = vset.variants.size() > 1;
+    for (const auto& v : vset.variants) {
+        if (v.sharded || v.name == v.primary_file) {
```
fl0rianr: Could this fallback be too broad? If a folder contains multiple sharded quantizations, e.g. Q4_K_M split into 00001/00002 and Q8_0 split into 00001/00002, this condition collapses the whole folder back into one legacy model. Should we split per variant while keeping each variant’s shards grouped, or add a test/documentation that multiple sharded variants in one extra_models_dir folder are intentionally unsupported?

src/cpp/server/model_manager.cpp
```diff
@@ -1101,61 +1097,121 @@ std::map<std::string, ModelInfo> ModelManager::discover_extra_models() const {
...
+            }
+            info.type = get_model_type_from_labels(info.labels);
+            discovered[variant_id] = info;
+
+            // Keep the old extra.<folder> ID working for scripts and saved aliases.
+            if (path == main_model_path) {
+                std::string legacy_id = std::string(EXTRA_MODEL_PREFIX) + dir_name;
+                ModelInfo legacy_info = info;
+                legacy_info.model_name = legacy_id;
+                discovered[legacy_id] = legacy_info;
```
fl0rianr: Is the legacy folder ID intended to be emitted as a separate visible model entry? This keeps old scripts working, but it also means the selected alphabetical variant appears twice in /models: once under the folder alias and once under its file-stem variant ID. Could we keep this as an input alias without listing it as another model, or add a test asserting the duplicate visible entry is intentional?

src/cpp/include/lemon/model_manager.h
```diff
@@ -226,6 +226,14 @@ class ModelManager {
     // Discover GGUF models from extra_models_dir
     std::map<std::string, ModelInfo> discover_extra_models() const;
 
+    ModelInfo init_extra_model_info(const std::string& name) const;
+
+    // Scans a subfolder for GGUF variants while preserving the legacy folder ID.
+    void discover_extra_models_in_directory(
+        const std::filesystem::path& dir_path,
```
fl0rianr, line 389: This file now uses `std::filesystem::path`, so it should include <filesystem> directly instead of relying on transitive includes.

src/cpp/server/model_manager.cpp
```diff
@@ -1101,61 +1097,121 @@ std::map<std::string, ModelInfo> ModelManager::discover_extra_models() const {
...
-            // No main model file found (only mmproj?), skip
+    for (const auto& gguf_path : gguf_files) {
+        uint64_t file_size = 0;
+        try {
+            file_size = static_cast<uint64_t>(fs::file_size(gguf_path));
+            total_size += static_cast<double>(file_size) / (1024.0 * 1024.0 * 1024.0);
+        } catch (...) {}
+
+        if (contains_ignore_case(gguf_path.filename().string(), "mmproj")) {
+            mmproj_file = gguf_path;
```
fl0rianr: This silently keeps only one mmproj file and attaches it to every split variant. If a folder contains multiple projector files, traversal order can decide which one is used. Could we make selection deterministic and/or add a test for multiple mmproj files?

fl0rianr (changes requested):
@anditherobot thanks for the good work here, please see the details below.

jeremyfowers:
@fl0rianr please continue with the review or close the PR

fl0rianr (commented):
Thanks for addressing the points from the first review and sorry for comming back so late. The legacy folder name is now handled as an input alias, multiple sharded variants are separated correctly, the missing filesystem include was added, and `mmproj` selection is deterministic. The added tests cover these changes well.

I found one remaining correctness issue before this is ready to merge:

Split model IDs are created only from the GGUF filename. Since the `discovered` map is shared across all folders, two different model folders containing common filenames such as `model-Q4_K_M.gguf` and `model-Q8_0.gguf` will generate identical IDs. Models discovered later will silently overwrite the models from the earlier folder.

Please make the generated IDs collision-safe and deterministic, for example by including the folder name when necessary, and add a regression test with two separate folders containing the same variant filenames.

It would also be good to verify the behaviour when multiple files contain the same quantization token but are not shards, for example:

* `Model-Q4_K_M.gguf`
* `Model-Q4_K_M-imatrix.gguf`

`enumerate_gguf_variants()` currently appears to group files by quantization token, which may incorrectly treat these as one sharded variant. Please either distinguish actual shard filenames or document and test the intended limitation.

As disussed the code is far behind and has merge conflicts. Please rebase or merge main, update `src/cpp/Extra-Models-Dir-Spec.md` to describe the new multi-variant folder behaviour, and rerun the C++ build and the complete set of added endpoint tests.

This is a useful fix and should be ready to land once these points are addressed. Thanks!

fl0rianr (commented):
Thanks, the previous issues look fixed now, nearly there!
I found one small regression in the new variant grouping though: when two files share the same quant, their names become the full file stems, but sorting still calls `quant_priority()` on that name. This can make Q8 sort before Q4, and `--yes` picks the first variant.

Could you keep the quant-based sorting here, maybe by extracting the quant from the name/primary file, and add a small Q4 + Q4-imatrix + Q8 regression test?

Apart from that this looks good to me.

fl0rianr (approved):
LGTM

PR #2100, opened 2026-06-04
Allow internal endpoints from remote addresses; warn on unsecured non-loopback bind

jeremyfowers:
## Summary

`/internal/*` endpoints returned `403` for any non-localhost request, breaking the many-clients-one-server topology (a desktop app/CLI on another machine can't manage a shared `lemond`) and blocking remote clients from `/internal/cloud/discover` in #1785.

This lifts the loopback restriction and relies on the existing admin-key gating (`LEMONADE_ADMIN_API_KEY`, which defaults to `LEMONADE_API_KEY`). Since an unsecured non-loopback bind now exposes the control endpoints to the network, `lemond` logs a startup warning (re-checked on host rebind) when bound to a non-loopback host with no key set:

```
[Warn] (Server) Serving on non-loopback host '0.0.0.0' without an API key. All endpoints, including the /internal/* control endpoints, are reachable from other machines unauthenticated. Set LEMONADE_API_KEY or LEMONADE_ADMIN_API_KEY to secure this server.
```

Docs updated to match.

## Testing

Verified manually on a `0.0.0.0` bind via the machine's LAN IP (non-loopback `remote_addr`):

- No keys: `GET /internal/config` → 200 (was 403); startup warning logged
- Admin key set: no/wrong key → 401, correct key → 200; no warning

Existing `test/server_env_vars.py` admin-key tests are unaffected. No automated test for the remote-address path — it needs a routable interface, which is flaky in CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2110, opened 2026-06-05
fix: allow admin API key on /internal/* from non-loopback clients

### LINKED ISSUE #2109: /internal/* endpoints return 403 for non-loopback clients even with valid LEMONADE_ADMIN_API_KEY
## Bug

`/internal/*` endpoints (e.g. `/internal/config`, `/internal/shutdown`, `/internal/set`, `/internal/cleanup-cache`) return **403** for any request originating from a non-loopback address, **even when the request carries a valid `LEMONADE_ADMIN_API_KEY`**. The admin API key check is never reached because the loopback check returns 403 first.

This makes the admin API key useless for remote clients — its documented purpose (per PR #1517 and Issue #1554) is precisely to allow elevated access to internal endpoints.

## Reproduction

1. On a server with `LEMONADE_ADMIN_API_KEY=<some-key>` set and `host` configured to listen on a non-loopback interface (e.g. `0.0.0.0:13305`).
2. From a different machine on the same network, send a request with a valid bearer token:
   ```bash
   curl -H "Authorization: Bearer <LEMONADE_ADMIN_API_KEY>" \
        http://192.168.77.13:13305/internal/config
   ```
3. Server responds with:
   ```http
   HTTP/1.1 403 Forbidden
   {"error": "Internal endpoints are only accessible from localhost"}
   ```
4. The same request from `localhost` on the server works as expected.

The same problem affects every endpoint under `/internal/`.

## Root cause

`src/cpp/server/server.cpp`, `Server::authenticate_request()`:

```cpp
// Internal endpoints are restricted to loopback regardless of API key
if (is_internal_route) {
    bool is_loopback = (req.remote_addr == "127.0.0.1" ||
                        req.remote_addr == "::1" ||
                        req.remote_addr == "::ffff:127.0.0.1");
    if (!is_loopback) {
        ...
        res.status = 403;
        ...
        return Handled;
    }
}

// ... auth_token is extracted and admin_api_key_ is checked AFTER the
// loopback guard above, so it can never authorize a non-loopback request.
```

The loopback check fires before the bearer token is even read, and before the admin key validation. The comment "restricted to loopback regardless of API key" is explicit about the design intent, but it contradicts the documented role of `LEMONADE_ADMIN_API_KEY`.

## Expected behavior

A request from a non-loopback address to `/internal/*` should be accepted if it carries a valid `LEMONADE_ADMIN_API_KEY`. When a non-loopback request does **not** present a valid admin key, it is rejected with **403** (not 401) — the request is denied for one of two reasons (non-loopback source OR missing/invalid credentials), and the response body explicitly states both options: `{"error": "Internal endpoints are only accessible from localhost or with admin API key"}`. Behavior from loopback (no auth required when no keys are set, admin key required when set) is unchanged.

## Suggested fix

Hoist `auth_token = httplib::get_bearer_token_auth(req)` above the loopback check, and let the loopback check pass through requests that present a valid admin key:

```cpp
if (is_internal_route) {
    bool is_loopback = (...);
    bool has_valid_admin_key = (!admin_api_key_.empty() &&
                                auth_token == admin_api_key_);
    if (!is_loopback && !has_valid_admin_key) {
        // 403
    }
    // existing admin key 401 check still applies for loopback callers
}
```

## Environment

- lemonade built from `main` (post PR #1564 follow-ups merged Apr 7, 2026)
- `--host 0.0.0.0`, `LEMONADE_ADMIN_API_KEY` set
- Linux (Ubuntu), remote curl from a second host on the same subnet

blackdeathdrow:
Closes #2109.

The loopback check in `Server::authenticate_request()` ran before the bearer token was extracted, so internal endpoints (`/internal/config`, `/internal/shutdown`, etc.) always returned 403 for any non-loopback client regardless of a valid `LEMONADE_ADMIN_API_KEY`.

Move `auth_token` extraction before the loopback check and allow the request through when the client presents a valid admin API key. A non-loopback request that does **not** present a valid admin key is rejected with **403** (not 401) — the response body states both options ("from localhost or with admin API key"). Loopback behavior is unchanged.

### Changes

- **[`src/cpp/server/server.cpp`](src/cpp/server/server.cpp)**: Hoist `httplib::get_bearer_token_auth(req)` above the loopback check in `authenticate_request()`. The non-loopback rejection now checks for a valid admin API key before returning 403. Updated the error message to reflect both options.

### Tests

- **[`test/server_env_vars.py`](test/server_env_vars.py)**: Added `TestInternalAdminKeyNonLoopback` — binds the server to `0.0.0.0` and reaches it via a non-loopback IPv4 address discovered at runtime. Skips via `unittest.SkipTest` when no non-loopback IPv4 is available. Covers:
  - non-loopback + valid admin key → 200
  - non-loopback + no key → 403
  - non-loopback + invalid key → 403
  - loopback + valid admin key → 200 (unchanged)
  - loopback + no key → 401 (unchanged)

  All 5 tests pass locally; full `test/server_env_vars.py` suite (40 tests) still passes.

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -269,7 +269,9 @@ httplib::Server::HandlerResponse Server::authenticate_request(const httplib::Req
                         (req.path.rfind("/v1/", 0) == 0);
     bool is_internal_route = (req.path.rfind("/internal/", 0) == 0);
 
-    // Internal endpoints are restricted to loopback regardless of API key
+    std::string auth_token = httplib::get_bearer_token_auth(req);
+
+    // Internal endpoints: allow from loopback OR with valid admin API key
     if (is_internal_route) {
         // ::ffff:127.0.0.1 is how an IPv6 socket reports an IPv4 loopback connection
         // when bound without IPV6_V6ONLY (the default on macOS, and the configuration
@@ -277,25 +279,16 @@ httplib::Server::HandlerResponse Server::authenticate_request(const httplib::Req
         bool is_loopback = (req.remote_addr == "127.0.0.1" ||
                             req.remote_addr == "::1" ||
                             req.remote_addr == "::ffff:127.0.0.1");
-        if (!is_loopback) {
+        bool has_valid_admin_key = (!admin_api_key_.empty() && auth_token == admin_api_key_);
+
+        if (!is_loopback && !has_valid_admin_key) {
             LOG(WARNING, "Server") << "Rejected internal request from non-loopback address: "
                         << req.remote_addr << " " << req.path << std::endl;
             res.status = 403;
-            res.set_content("{\"error\": \"Internal endpoints are only accessible from localhost\"}", "application/json");
+            res.set_content("{\"error\": \"Internal endpoints are only accessible from localhost or with admin API key\"}", "application/json");
             return httplib::Server::HandlerResponse::Handled;
         }
-    }
 
-    // Authentication hierarchy:
-    // - Admin key: access to both internal and regular API endpoints
-    // - Regular API key: access only to regular API endpoints (not internal)
-    // - If admin key is not set, it defaults to regular API key value
-    // - If only admin key is set, regular endpoints are accessible without auth, internal requires admin key
-    // - If no keys are set, all endpoints are accessible without auth
-
-    std::string auth_token = httplib::get_bearer_token_auth(req);
-
-    if (is_internal_route) {
         // Internal routes require admin key authentication
         if (!admin_api_key_.empty() && req.method != "OPTIONS") {
             if (auth_token != admin_api_key_) {
@@ -305,6 +298,12 @@ httplib::Server::HandlerResponse Server::authenticate_request(const httplib::Req
             }
         }
     } else if (is_api_route && req.method != "OPTIONS") {
+        // Authentication hierarchy:
+        // - Admin key: access to both internal and regular API endpoints
+        // - Regular API key: access only to regular API endpoints (not internal)
+        // - If admin key is not set, it defaults to regular API key value
+        // - If only admin key is set, regular endpoints are accessible without auth, internal requires admin key
+        // - If no keys are set, all endpoints are accessible without auth
         if (!api_key_.empty()) {
             if ((auth_token != api_key_) && (auth_token != admin_api_key_)) {
                 res.status = 401;
--- a/test/server_env_vars.py
+++ b/test/server_env_vars.py
@@ -13,6 +13,7 @@ Usage:
 import argparse
 import os
 import platform
+import socket
 import subprocess
 import sys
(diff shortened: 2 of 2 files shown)


fl0rianr (commented):
Thanks for the focused fix — moving bearer-token extraction before the `/internal/*` loopback guard makes sense and matches the intended “admin key can access internal endpoints” model.

Before merging, I think this should have a regression test for the actual failing case. The existing admin-key tests appear to exercise /internal/config via localhost, so they do not cover the non-loopback path that previously returned 403 before the admin key was checked. Could you add coverage for a server bound to a non-loopback interface, e.g. LEMONADE_HOST=0.0.0.0, and verify that:

- /internal/config from a non-loopback client succeeds with a valid admin bearer token
- the same request without/with an invalid token is rejected
- existing loopback behavior remains unchanged

One semantic detail worth clarifying: the issue’s expected behavior says non-loopback /internal/* without a valid admin key should be rejected with 401, but this implementation still returns 403 from the loopback/admin-key gate before reaching the later admin-key check. If 403 is intentional for “non-loopback and not admin-authorized”, that’s fine, but it would be good to make that explicit in the PR/issue and test expectations.

jeremyfowers:
Similar to https://github.com/lemonade-sdk/lemonade/pull/2100

Do people have a preference about which implementation? @fl0rianr ?

fl0rianr:
I think I slightly prefer the #2100 direction if the intended product behavior is to support the many-clients-one-server topology.

#2110 is the narrower bugfix: non-loopback /internal/* is allowed only with a valid admin key. That’s conservative and directly addresses #2109.

#2100 is the cleaner policy change: internal endpoints can be reached remotely, key enforcement handles access control when a key is configured, and unsecured non-loopback binds get a clear warning/docs update. I like that better if we’re comfortable with the existing “no key = unauthenticated” model applying to internal endpoints too.

If we go with #2100 and close this one, it would be good to preserve credit for this fix as well — maybe add @blackdeathdrow as a co-author there, if appropriate.

Either way, I think the important part is making the auth/status-code matrix explicit so future changes don’t reintroduce ambiguity around loopback vs admin-key behavior.

blackdeathdrow:
Thank you for the review. I used OpenCode to create and run the test in Docker. I hope this will be useful.

bitgamma:
considering we have a bunch of security-related PR open, maybe the more conservative approach of this PR is more in line with the rest, but I am too fine with either implementation.

jeremyfowers:
The intention with #2100 is that users who are just starting out with Docker or LAN-wide use see a helpful warning message "you should set an API key" instead of a 403 error. It trades off some security for a lot of ease of use.

Since it sounds like folks are ok with it, I will gladly credit @blackdeathdrow on there and proceed.

blackdeathdrow:
Superseded by #2100.

PR #2114, opened 2026-06-05
Add Debian 13 to NPU installation instructions

superm1:


fl0rianr (commented):
Nice addition. One small clarification request: this enables trixie-backports, but the Debian 13 path does not appear to install any package from backports afterwards. If backports is needed for a specific kernel/driver/firmware/XRT package, could we name that package and install it explicitly with `-t trixie-backports` or `<package>/trixie-backports`? Otherwise, if the FLM .deb is the only required Debian 13 step, we may want to omit the backports setup to avoid adding an unnecessary prerequisite for users.

superm1:
Hmm, I would expect that it resolved from installing FLM Deb because of minimum version in Trixie backports.

So isn't that an unnecessary step?

fl0rianr:
Yes, that makes sense if the FLM .deb has a versioned dependency that cannot be satisfied from regular Trixie and apt then pulls the required package from `trixie-backports`.

My main concern was that the current instructions do not make it obvious why enabling backports is required. Maybe we can add a short note like “Backports is needed so apt can satisfy FLM’s newer runtime/driver dependency when installing the .deb.” That would avoid users thinking this step is unused.

fl0rianr (approved):
LGTM, thanks!

PR #2115, opened 2026-06-05
feat(backends): add Moonshine streaming STT backend

Theohox:
Adds a new 'moonshine' recipe for Moonshine speech-to-text using the moonshine_voice C++ streaming API. Integrates with existing OpenAI-compatible /v1/audio/transcriptions and WebSocket /realtime endpoints.

Includes:
- MoonshineServer backend (HTTP /inference, TCP streaming)
- IStreamingTranscriptionServer interface for generic streaming backends
- TcpJsonlClient for line-delimited JSON over TCP
- RealtimeSessionManager integration for WebSocket/audio forwarding
- ModelManager support for moonshine cache resolution and download
- CPU-only, cross-platform (Linux/Windows x86_64)
- End-to-end smoke test

The existing Whisper path is unchanged.

Requesting @bitgamma to review per discord comments


bitgamma (changes requested):
Thanks for your PR!

I see the current implementation is relying on system-wide Python installation, which is not something we can rely on. Generally, we try to avoid Python-based backends for this reason. The only Python-based backend we have is vLLM but it has been packaged to be self-contained and not rely (or interfere with) the system Python. Similar work must be done before this can be considered viable.

Additionally, I see models are not being downloaded from huggingface. Adding another download repository is something we also want to avoid. I'd much prefer if you could find or re-upload the models on HF instead. I also see voices are being downloaded in the user folder, outside the usual lemonade model folder - this is another thing we want to avoid.

In general, this needs to be reworked to follow the same patterns as the other backends.

Theohox:
Thank you!

Theohox:
@bitgamma I've reworked this PR to address all four points:                                                     
                                                                                                                                                         
    1. **No system Python dependency** — Moonshine now follows the vLLM pattern: `get_install_params()` returns a platform-specific tarball from `lemonade-sdk/moonshine-server`, and `load()` uses `BackendUtils::get_backend_binary_path()` to find the bundled executable. `PYTHONNOUSERSITE=1` is set to isolate the environment.
     2. **Models from HuggingFace** — Checkpoints now point to `UsefulSensors/moonshine-streaming:onnx/tiny|small|medium`. The standard HF download/cache path handles everything; the custom `moonshine_voice` cache directory and separate download repo are removed.
     3. **Standard model folder** — Models are cached in the regular HF hub cache (`~/.cache/huggingface/hub/...`). No more user-folder voice cache outside the lemonade model directory.
     4. **Follows existing backend patterns** — Removed all special-case code from `ModelManager` (custom resolver, custom downloader, custom cache dir). Moonshine now uses the same `install_backend()` → `get_backend_binary_path()` → `ProcessManager::start_process()` flow as vLLM and the other backends.

     One detail: the HF models ship `tokenizer.json` but `moonshine_voice` expects `tokenizer.bin`. I added an auto-conversion step in `main.py` that converts on first load, so no manual preprocessing is needed.

     Tested end-to-end: `lemond` loads the model via `/v1/load`, the backend subprocess starts, and the health endpoint returns ready.

     Ready for another look when you have time.

bitgamma:
@Theohox 

> 1. **No system Python dependency** — Moonshine now follows the vLLM pattern: `get_install_params()` returns a platform-specific tarball from `lemonade-sdk/moonshine-server`

That repository does not exist. Making this PR a draft. Please validate locally that everything works as it should before marking it as ready for review.

Geramy:
@bitgamma can I get you to approve this PR real quick? I'm going to merge it into a branch and then open new PR to help finish it.

Geramy (approved):
approved to be merged into temp branch.

PR #2116, opened 2026-06-05
ci: set HF_HOME before starting hosted Linux server

fl0rianr:
## Summary

Move the hosted Linux HF_HOME setup before the .deb install action starts `lemond`.

## Why

The .deb install action starts the server as part of installation. In the consolidated hosted Linux job, HF_HOME was being written to $GITHUB_ENV only after that action completed, so the running `lemond` process did not inherit the intended job-local HF cache path.

This meant model downloads from the server process could still use the default cache location instead of the shared per-job `hf-cache`.

## Changes

- Move the `Set environment` step before `Install Lemonade (.deb)`.
- Create the `hf-cache` directory before server startup.
- Keep the existing step id so later test-step gating remains unchanged.

## Scope

This is a small follow-up to the CI consolidation work. It does not change test coverage, model selection, or server code.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1445,6 +1445,7 @@ jobs:
     env:
       LEMONADE_CI_MODE: "True"
       HF_TOKEN: ${{ secrets.HUGGINGFACE_ACCESS_TOKEN }}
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@claude review

jeremyfowers (approved):
LGTM, thanks!

PR #2117, opened 2026-06-05
ci / test: fix the 409 CI error

fl0rianr:
## Summary

Make server_streaming_errors.py more resilient to transient /load setup failures.

## Why

The streaming error tests are meant to verify that streaming responses terminate cleanly. A transient 5xx from the setup /load call could fail the test before the actual streaming behavior was exercised.

## Changes

* Add a small helper to load ENDPOINT_TEST_MODEL with bounded retry.
* Treat 200 and 409 setup responses as already-loaded / successfully-loaded states.
* Retry transient 5xx setup failures with /unload recovery before retry attempts.
* Avoid forcing /unload on the first attempt so an already-loaded model can remain a cheap /load no-op.
* Keep persistent failures visible with status code and response body.
* Use the helper in the streaming tests that require the model to be loaded.

## Scope

This does not relax the streaming assertions. The tests still fail if the stream does not terminate cleanly or if a successful stream does not include [DONE].

### DIFF
--- a/test/server_streaming_errors.py
+++ b/test/server_streaming_errors.py
@@ -73,16 +73,20 @@ class StreamingErrorTests(ServerTestBase):
             if attempt > 1:
                 time.sleep(2 * attempt)
 
-            # Start each attempt from a clean loaded-model state. Ignore unload
-            # failures here; the following /load response is the real setup signal.
-            try:
-                requests.post(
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@claude review

jeremyfowers (approved):
LGTM, Claude’s complaints don’t bother me but your call if you want to address any of them.

PR #2119, opened 2026-06-06
Remember downloaded-only model filter

### LINKED ISSUE #2111: "downloaded only" selector - remember previous setting
### Feature Description

Can we possibly save the previous state of the downloaded only toggle, so I don't have to keep changing it every single time?

### Use Case / Motivation

Ease of use - always good to save a users choice of settings etc

### Platform Relevance

All platforms

### Additional Context

_No response_

anditherobot:
Fixes #2111.

This remembers the Model Manager "Downloaded only" toggle as a client-local app setting.

The preference is saved under `modelManager.showDownloadedOnly`, so each desktop app or browser client keeps its own choice. This does not add a server-wide setting or API endpoint.

For the desktop app, the setting is included in the existing Tauri app-settings sanitizer because desktop settings are saved through local `app_settings.json`.

Tests:
- npm run test:downloaded-only-settings
- cargo test settings
- npm run build:renderer


src/app/src/renderer/ModelManager.tsx
```diff
@@ -373,6 +374,39 @@ const [searchQuery, setSearchQuery] = useState('');
   const { toasts, removeToast, showError, showSuccess, showWarning } = useToast();
   const { confirm, ConfirmDialog } = useConfirmDialog();
 
+  useEffect(() => {
+    const loadModelManagerSettings = async () => {
+      try {
+        if (!window.api?.getSettings) return;
```
fl0rianr, line 387: This can race with Tauri API installation. In Tauri, window.api is installed asynchronously by tauriShim, and App.tsx already documents that consumers may need to await tauriReady because window.api is not guaranteed to exist on first render. Since this effect runs only once and silently returns here, the saved `showDownloadedOnly` preference may never be restored.

Please await `tauriReady` before reading settings, or route this through a settings helper that guarantees API readiness.

src/app/src/renderer/ModelManager.tsx
```diff
@@ -373,6 +374,39 @@ const [searchQuery, setSearchQuery] = useState('');
...
+
+    loadModelManagerSettings();
+  }, []);
+
+  const updateShowDownloadedOnly = useCallback(async (checked: boolean) => {
+    setShowDownloadedOnly(checked);
+    setShowFilterPanel(false);
+
+    try {
+      if (!window.api?.getSettings || !window.api?.saveSettings) return;
```
fl0rianr, line 404: Same API-readiness issue on the save path: the UI is updated optimistically before this guard, but if window.api is not ready yet, persistence is silently skipped. That leaves the checkbox looking correct for the current session while the preference is not actually saved.

Please make this path wait for the same API readiness guarantee as the load path, or explicitly handle/report the skipped save.

fl0rianr (changes requested):
Product scope question: this PR intentionally persists the setting client-locally. The issue says “All platforms”, and the discussion mentions that a server-side setting might also be desired. Can we confirm that per-client/browser persistence is the intended behavior before merging? See also comments down below, first one is a blocker:

jeremyfowers:
icymi: the existing GUI is being fully replaced by the GUI working group, led by @kpoineal. He has discretion for whether he wants to let anything else merge into the old GUI.

kpoineal (approved):

kpoineal:
Thanks @jeremyfowers. Same situation here — this is a useful quality-of-life feature and the persistence logic is clean. The new UI will need to implement its own version of this setting anyway, so no reason to hold it back. Approving for main. ✅

fl0rianr (approved):

PR #2121, opened 2026-06-06
Fix light theme persistence

### LINKED ISSUE #2120: Light theme resets after refreshing the app
When the app is set to Light theme, refreshing or reopening the app changes it back to Dark theme.

Steps:
1. Open the app.
2. Switch the theme to Light.
3. Refresh or reopen the app.

Expected:
The app stays in Light theme.

Actual:
This is a local app settings issue. The selected theme should be saved by the client, not by the Lemonade server.

anditherobot:
Fixes #2120.

This change keeps the selected Light theme after refresh or reopen.

The app was already saving a `theme` value, but the settings load/save code did not keep that value when it rebuilt the settings object. This caused the app to fall back to the default Dark theme.

This PR preserves valid theme values: `dark` and `light`.

Tests:
- npm run build:renderer

### DIFF
--- a/src/app/package.json
+++ b/src/app/package.json
@@ -12,7 +12,8 @@
     "build": "tauri build",
     "build:nobundle": "tauri build --no-bundle",
     "build:mac": "tauri build --bundles app",
-    "test:custom-collections": "node ../../test/app/customCollections.test.cjs"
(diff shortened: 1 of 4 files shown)


sofiageo (approved):
Tested both server and desktop app. works fine!

PR #2123, opened 2026-06-06
test: retry transient model pull setup failures

fl0rianr:
## Summary

Retry transient model /pull setup failures in server tests.

## Why

Several tests use /api/v1/pull only as setup before checking unrelated behavior. A one-off transient 5xx from /pull could fail the test before the actual assertion was exercised.

## Changes

- Add shared `pull_model_with_retry()` test helper.
- Retry transient 5xx pull setup failures with bounded backoff.
- Use the helper in endpoint, omni, and llama.cpp system backend tests.
- Keep permanent errors and persistent failures visible with status/body details.

## Scope

This does not relax the actual test assertions. It only makes model-pull setup more resilient to transient server/download failures.

### DIFF
--- a/test/server_cli2.py
+++ b/test/server_cli2.py
@@ -24,6 +24,7 @@ import glob
 import json
 import os
 import platform
+import re
 import requests
 import shutil
 import subprocess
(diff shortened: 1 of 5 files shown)


jeremyfowers (approved):
Thanks for this!

PR #2127, opened 2026-06-06
ROCm backends  problem: Problems using iGPU and eGPU configs and a way to be able to force the ROCm architecture.

GabrielReusRodriguez:
Installing lemonade on ubunut, I saw that when you have iGPU and eGPU and you install the llamacpp:rocm backend , it always installs the gfx1150 version of therock. I tryed with HSA_OVERRIDE_GFX_VERSION environment var but it was impossible

I opened an issue on: 
https://github.com/lemonade-sdk/llamacpp-rocm/issues/104

I saw that in the function  std::string SystemInfo::get_rocm_arch() of system_info.cpp file, it ALWAYS chose the first active gpu architecture  and it always is iGPU. In my case.

So I have added a new environment var LEMONADE_ROCM_ARCH where you can inform to lemonade the desired architecture and if it is informed, it will use it.

I also changed backend_versions.json to add architecture gfx120X and url mapping of that.
You can inform LEMONADE_ROCM_ARCH=gfx1200  or LEMONADE_ROCM_ARCH=120000

I tested on my machines and I think it works!


src/cpp/server/system_info.cpp
```diff
@@ -2094,12 +2094,34 @@ std::string SystemInfo::get_rocm_arch() {
 
         // Check AMD GPUs
         if (devices.contains("amd_gpu") && devices["amd_gpu"].is_array()) {
+
+            // First I get The environment var LEMONADE_ROCM_ARCH if it exists.
+            const char* rocm_arch_env_raw = std::getenv("LEMONADE_ROCM_ARCH");
+            std::string rocm_env_arch = (rocm_arch_env_raw  != nullptr) ? rocm_arch_env_raw  : "";
```
superm1, line 2100: please let's not add new environment variables.  I'm actively removing them.

https://github.com/lemonade-sdk/lemonade/pull/2106

src/cpp/server/system_info.cpp
```diff
@@ -890,7 +890,7 @@ json SystemInfo::get_device_dict() {
     // Get AMD GPU info (both integrated and discrete) - with fault tolerance
     try {
         devices["amd_gpu"] = json::array();
-
+
```
superm1, line 893: spurious whitespace?

src/cpp/resources/backend_versions.json
```diff
@@ -59,6 +60,7 @@
       "gfx1152": "gfx1152",
       "gfx1200": "gfx120X-all",
       "gfx1201": "gfx120X-all",
+      "gfx120X": "gfx120X-all",
```
superm1, line 63: code duplication?

src/cpp/resources/backend_versions.json
```diff
@@ -34,6 +34,7 @@
       "gfx1150",
       "gfx1151",
       "gfx1152",
+      "gfx120X",
```
superm1, line 37: this right?

superm1 (commented):
I would much rather see a preference plumbed through the json configuration

GabrielReusRodriguez:
Hello,

I will think about the cpnfig.json option.

So is this pr closed? Or i can present the changes to this one?

src/cpp/server/system_info.cpp
GabrielReusRodriguez: Yes, I am going to remove them.

src/cpp/resources/backend_versions.json
GabrielReusRodriguez: I think it is some testing i did before the implementation.,  I check it

src/cpp/resources/backend_versions.json
GabrielReusRodriguez: I think it is some testing i did before the implementation., I check it

src/cpp/server/system_info.cpp
GabrielReusRodriguez: I am   trying to read from config.json.

Can we get the config.json values with code? 

 std::string rocm_env_arch = "";

// First I get the config lemonade->rocm_arch config value.
if (auto* cfg = RuntimeConfig::global()) {
            rocm_env_arch = cfg->backend_string("llamacpp", "rocm_arch");
}

superm1:
You don't need to close the PR to iterate on it (unless you want to).  Just force push when you've improved it.

As for how to go about this; I slept on it and I think this actually warrants a new CLI option to show available GPUs and set the preference.  The preference would be what gets saved to the JSON file.

Let me give you and example why I think this.

Let's say you have two R9700s in the system.  They should show up as exact same identification, but you might be using one for gaming and one only for inference.  You need a way to indicate which one should be used for lemonade.

Issue #2032, opened 2026-05-28
NPU models losses quality answering most prompt


oliver021:
### Platform

Windows

### Lemonade Version

latest

### GPU / APU Model

AMD Ryzen AI 9 HX 370

### Component

llama.cpp

## Summary

When running distilled/quantized LLMs via the `flm` and `ryzenai-llm` backends on XDNA2 NPU hardware, models exhibit significant factual accuracy degradation compared to their full-precision counterparts. This is an observational report intended to inform documentation, model selection guidance, and potential roadmap considerations.

---

## Environment

- **Hardware**: AMD XDNA2 NPU
- **Backend**: `flm` / `ryzenai-llm`
- **OS**: Windows
- **Lemonade version**: latest

---

### 1. Factual Recall Degradation
Distilled models consistently produce hallucinated or incorrect factual claims (dates, proper nouns, historical attributions). This is consistent with known behavior in aggressively quantized models where high-precision weights encoding factual associations are lossy-compressed.

### 2. Poor Combinatorial Reasoning
Models show reduced capability when tasks require recombining concepts across domains or performing multi-hop reasoning chains. Distillation disproportionately affects attention heads responsible for cross-domain association.

### 3. Definition Precision Loss
Technical definitions are generated with structurally correct prose but semantically incorrect content. In the example above, the Law of Large Numbers was defined as "frequency approaches 1/2" — a fundamental misstatement of convergence to the true expectation.

### 4. Repetition Artifacts
Paragraph-level content duplication was observed in longer outputs, suggesting reduced long-range attention coherence — a known side effect of weight pruning and reduced context capacity.

---

## Root Cause Hypothesis

These failure modes are consistent with:
- **Quantization-induced precision loss** on factual knowledge embeddings
- **Reduced parameter count** limiting the model's capacity to store and retrieve sparse factual associations
- **Attention degradation** from pruning, affecting multi-hop and long-range coherence

The model retains strong **surface fluency** (syntactic and stylistic coherence) while losing **semantic depth** (factual grounding, definitional precision).

---

## Suggested Mitigations / Documentation Improvements

- [ ] Add explicit documentation warning about factual reliability limitations for NPU-optimized models
- [ ] Recommend RAG (Retrieval-Augmented Generation) pipelines for fact-sensitive use cases
- [ ] Define and publish a recommended task taxonomy for NPU models, e.g.:
  - ✅ Summarization, classification, reformulation, structured extraction
[...]

PR #2129, opened 2026-06-07
auto-label: teach prompt about area::ci + test coverage ≠ documentation

ramkrishna2910:
Follow-up to #2076.

After running the first backfill of open items via #2076's workflow, two small gaps showed up in the prompt:

## 1. CI / infrastructure items had no label

Four open items (#2116, #2037, #1960, plus one more on the website side) were about CI workflows, runner cleanup, or test infrastructure. None of `area::cli`, `area::installer`, `area::api`, `area::tray` fit, so the model returned `(none)`.

This PR:
- Adds `area::ci` to the `Area` block of `SYSTEM_PROMPT` with explicit coverage for: CI / GitHub Actions workflows, self-hosted runner infrastructure, and test infrastructure (fixtures, harness, CI cleanup).
- Adds `"area::ci"` to `KNOWN_LABELS` so the parser admits it. The `area::ci` label already exists on the repo (created out-of-band during the backfill triage).
- Verified the at-most-one-`area::` rule still applies: `area::ci, area::api` → just `area::ci` passes through; `area::ci` against an existing `area::*` is dropped.

## 2. `documentation` was over-applied to test-coverage items

The model labeled #2032 and #2040 with `documentation` even though they were a quality bug and a test-coverage gap respectively. The connection seems to be the model treating "documents the missing coverage" or "missing test docs" loosely.

This PR adds an explicit rule to the prompt:

> `documentation` is for human-readable docs only — READMEs, user guides, installation instructions, doc-comments. Items about test coverage, missing tests, test infrastructure, or test harness changes are `enhancement` (or `bug` if the test itself is broken), NEVER `documentation`.

## Test plan / rollout

What this PR exercises automatically vs what needs a manual nudge (h/t @fl0rianr for catching that the original wording was wrong):

| Trigger | LLM runs? | Re-labels existing `(none)` items? |
|---|---|---|
| `issues.opened` / `pull_request_target.opened` | ✓ | Only for *new* items going forward |
| `issue_comment.created` | ✗ priority-only | No |
| `schedule` daily cron | ✗ priority-only | No |
| `workflow_dispatch -f mode=all` | ✓ | **Yes** — the post-merge action |

- [x] Local parser verification — 5 cases covering `area::ci` admission, at-most-one-`area::` enforcement, and co-occurrence with engine + type labels.
- [ ] **After merge**, dispatch a fresh backfill with `gh workflow run auto-label.yml -f dry_run=true -f mode=all` to preview re-classification of items currently at `(none)`. The backfill filter already excludes items with `engine::/area::`, so this only touches items that the previous backfill skipped.
- [ ] After review of the dry-run decisions, dispatch with `-f dry_run=false -f mode=all` to apply.

### DIFF
--- a/.github/scripts/auto_label.py
+++ b/.github/scripts/auto_label.py
@@ -54,6 +54,7 @@ Area — apply AT MOST ONE total. Same hard rule as engines. Skip if not clearly
 - area::installer — Windows MSI, macOS DMG, Debian / RPM packaging
 - area::api       — HTTP REST API surface, route handlers, Ollama/Anthropic/OpenAI compat
 - area::tray      — system tray app (LemonadeServer.exe, lemonade-tray)
+- area::ci        — CI / GitHub Actions workflows, self-hosted runner infrastructure, test infrastructure (e.g. fixtures, harness, CI cleanup)
 
 Runtime — apply AT MOST ONE total. Same hard rule. Skip if the item is not specific to a particular GPU/compute runtime. Runtime labels complement engine labels — e.g., a Vulkan-only llama.cpp bug gets both `engine::llamacpp` and `runtime::vulkan`:
 - runtime::vulkan — Vulkan path (typically llama.cpp on AMD/Intel/Nvidia GPUs)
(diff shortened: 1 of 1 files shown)


fl0rianr (approved):
Thanks, looks good code-wise; this is a clean follow-up to #2076.

One small note on the test plan / rollout wording: schedule and issue_comment runs are priority-only, so they will not exercise the updated LLM prompt for already-open items. Existing issues that previously got `(none)` because `area::ci` did not exist would need a manual `workflow_dispatch` in `all` mode, either targeted or as a fresh backfill. Newly opened issues/PRs will pick this up automatically.

I don't think this blocks the code change, but it might be worth adjusting the PR description/test plan so the rollout expectation is accurate. Your call.

PR #2133, opened 2026-06-07
Remove pointless comments from codebase

superm1:
Removed ~100+ pointless comments that explained WHAT code does (obvious from names) rather than WHY (constraints, workarounds, design decisions).


fl0rianr (approved):
If you really want that you have my go.

PR #2142, opened 2026-06-08
Refactor codebase to drop so much `#ifdef _WIN32`

superm1:
These are all over the codebase.  This PR won't clean them **all up**, but the biggest ones are removed with a refactor.  This should improve readability.

---

Replace scattered #ifdef blocks in path utilities with C++ inheritance and virtual functions, following the TrayInterface pattern.

This establishes the pattern for remaining phases (process management, system metrics, archive extraction).

Replace platform-specific #ifdef blocks for archive extraction with C++ inheritance pattern.

Extract platform-specific process spawning, lifecycle, and utilities from process_manager.cpp into dedicated platform implementations following the established TrayInterface pattern.

Extract all Windows process management code from process_manager.cpp inline implementation into dedicated process_windows.cpp platform file, completing the platform abstraction pattern across all three platforms.

Changes:
- Create PathPlatform abstract interface with platform-specific methods
- Implement WindowsPathPlatform, LinuxPathPlatform, MacOSPathPlatform
- Refactor path_utils.cpp to delegate to platform implementation
- Update CMakeLists.txt for platform-specific source selection
- Create SystemMetricsPlatform abstract interface
- Implement LinuxMetricsPlatform, WindowsMetricsPlatform, MacOSMetricsPlatform
- Refactor server.cpp to delegate metrics collection to platform
- Add metrics_platform_ member to Server class
- Update CMakeLists.txt for platform-specific metrics sources
- Create ArchivePlatform abstract interface
- Implement WindowsArchivePlatform and UnixArchivePlatform
- Refactor BackendUtils::extract_zip() and extract_tarball() to delegate
- Windows: Native tar.exe with PowerShell Expand-Archive fallback for ZIP
- Unix: unzip for ZIP, tar for tarballs (.tar.gz, .tar.xz, etc.)
- Update CMakeLists.txt for platform-specific archive sources
- process_manager.cpp: Delegate to platform on Unix/macOS for start_process(), stop_process(), is_running(), find_free_port()
- Windows code remains inline (future extraction possible)
- CLI CMakeLists.txt: Add process platform sources to link correctly
- process_manager.cpp: All functions now delegate to platform on all OS (start_process, stop_process, is_running, get_exit_code, wait_for_exit, run_with_output, find_free_port, run_command, kill_process)
- Windows helper functions (escape_windows_arg, output_filter_thread, build_windows_environment_block) moved to process_windows.cpp
- Old inline implementations wrapped in #if 0 blocks for reference
- CMakeLists.txt: Add process_windows.cpp to Windows builds
- CLI CMakeLists.txt: Add process_windows.cpp to CLI Windows builds

Results:
- Reduced #ifdef blocks in path_utils.cpp from 15 to 5 (67% reduction)
- Remaining #ifdef only for includes and small utility functions
- Public API unchanged (backward compatible)
- Build verified successful on Linux
- Reduced #ifdef blocks in server.cpp (system metrics functions)
- CPU usage: Delegates to platform (Linux /proc/stat, Windows GetSystemTimes)
- Memory: Delegates to platform (Linux /proc/meminfo, Windows GlobalMemoryStatusEx, macOS sysctl)
- GPU/VRAM: Linux AMD sysfs, unsupported on Windows/macOS
- NPU: Linux DRM IOCTL sensor API, unsupported on Windows/macOS
- Remaining #ifdef in server.cpp only for includes and small utilities
- Build verified successful on Linux
- CLI updated to include platform files
- Reduced #ifdef blocks in backend_utils.cpp from 14 to 9 (36% reduction)
- Remaining #ifdef only for includes, .7z support, and TheRock paths
- extract_seven_zip() unchanged (Windows-specific .7z using bsdtar)
- Build verified successful on Linux


CMakeLists.txt
```diff
@@ -615,8 +615,22 @@ set(SOURCES_CORE
...
+    list(APPEND SOURCES_CORE src/cpp/server/platform/metrics_windows.cpp)
+    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/archive_windows.cpp)
+    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/process_windows.cpp)
+elseif(APPLE)
     list(APPEND SOURCES_CORE src/cpp/server/macos_system_info.mm)
+    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/path_macos.cpp)
+    list(APPEND SOURCES_CORE src/cpp/server/platform/metrics_macos.cpp)
+    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/archive_unix.cpp)
+    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/process_macos.cpp)
+elseif(UNIX)
```
fl0rianr, line 637: This branch split seems to have changed the meaning of the existing set_source_files_properties() line below.

src/cpp/server/macos_system_info.mm is appended in the APPLE branch, but the OBJCXX source property now ends up under `elseif(UNIX)`. So macOS no longer gets the explicit OBJCXX marking, while Linux sets a property for a macOS-only source.

Could we move that property back into the APPLE branch, or guard it separately with `if(APPLE)`?

src/cpp/include/lemon/system_metrics_platform.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include <memory>
+#include <mutex>
```
fl0rianr, line 5: Please add `#include <cstdint>` here.

This header uses `uint64_t` directly, but currently only includes <memory> and <mutex>. Relying on transitive includes makes the header non-self-contained and can break depending on include order.

fl0rianr (changes requested):
Thanks for taking this on — I like the direction of moving platform-specific behavior behind dedicated implementations. Given the big scope of the refactor, I’d like to see a small hardening pass before merge.

A few things I’d like to clarify/fix:

- The PR currently has a merge conflict. Since src/cpp/server/utils/path_utils.cpp is one of the conflict-sensitive files in this refactor, please make sure the final resolved version does not keep any leftover ownership-moving platform helper and preserves the existing path semantics.
- In CMakeLists.txt, the new platform branch split appears to change the meaning of the existing `set_source_files_properties(... macos_system_info.mm ... LANGUAGE OBJCXX)` line. macos_system_info.mm is added under the Apple branch, but the OBJCXX marking now appears to be under the Unix/Linux branch. Please double-check that macOS still gets the intended Objective-C++ source handling.
- src/cpp/include/lemon/system_metrics_platform.h uses uint64_t directly and should `include <cstdint>`.
- src/cpp/server/backends/backend_utils.cpp: `.7z` extraction now goes through the generic archive platform. If .7z is still intended to be Windows-only, I’d prefer keeping an explicit platform guard. If cross-platform .7z is intentional, the comments and tests should be updated accordingly.
- For src/cpp/server/utils/platform/process_macos.cpp, please preserve the old rationale for using `posix_spawn` on macOS. That comment documented a real macOS/Metal/XPC failure mode, not just an implementation preference.

Given how much platform-sensitive code moved here, I’d also feel better with targeted smoke coverage for Windows archive extraction and macOS process spawning before merging.

fl0rianr:
Thanks for the quick fixes — the earlier blockers look addressed now. 

One remaining cleanup request before approval: after the delegation refactor, src/cpp/server/utils/process_manager.cpp still carries the old platform-specific includes and helper functions, while the active methods now only delegate to ProcessPlatform. Could you remove that leftover code so the file reflects the new abstraction boundary?

Also, please make the new platform process files self-contained with their direct standard-library includes, e.g. <stdexcept> / <iostream> where used, rather than relying on transitive includes.

fl0rianr (commented):
see above

fl0rianr:
Thanks for the updates — the earlier review points look addressed now.

I found one remaining behavior regression that I think should be fixed before approval:

- In the Linux path platform, get_runtime_dir() no longer preserves the old RUNTIME_DIRECTORY fallback. The previous implementation first tried XDG_RUNTIME_DIR, then explicitly handled RUNTIME_DIRECTORY for systemd services. The new LinuxPathPlatform::get_runtime_dir() only checks XDG_RUNTIME_DIR and then throws. Please restore the RUNTIME_DIRECTORY fallback so service deployments keep working.

One smaller behavior-preservation point:

- The Windows filtered-output helper no longer matches the old logging/filtering behavior. The old common helper filtered /health, /v1/health, the llama idle heartbeat, and the interactive prompt, and routed error lines through LOG(ERROR) / normal lines through LOG(INFO). The new Windows helper only filters "/health" and writes directly to std::cout. Please keep the previous filtering/logging behavior on Windows as well.

Minor cleanup, non-blocking: process_manager.cpp still appears to include several headers that are no longer needed after the delegation refactor.

fl0rianr (changes requested):
Thanks for iterating on this — the previous behavior-preservation issues look mostly addressed now. The Linux RUNTIME_DIRECTORY fallback is back, the Windows output filtering/logging matches the old behavior again, and the CMake platform source selection looks correct.

I still see one Windows-specific blocker before approval:

- process_windows.cpp should preserve the old Windows header discipline from process_manager.cpp. The old file set _WINSOCKAPI_, WIN32_LEAN_AND_MEAN, NOMINMAX, included winsock2.h before windows.h, and undef’d ERROR after the Windows headers. The new process_windows.cpp currently includes windows.h directly after aixlog.hpp, while the same file uses both LOG(ERROR, ...) and Winsock APIs. Please carry over the old header setup into this TU to avoid ERROR macro and winsock.h /winsock2.h ordering problems.

One smaller path-preservation point:

- resolve_hf_cache_dir() changed the HF_HOME handling from explicit platform-specific string append (\\hub on Windows, /hub elsewhere) to `fs::path(hf_home) / "hub"`. Since the Windows path helpers are explicitly UTF-8-aware, please either keep the old platform-specific append or round-trip via path_from_utf8() / path_to_utf8() so non-ASCII Windows paths keep the same behavior.

I try to be critical here since such a big change is not an easy one.

fl0rianr:
One remaining hardening issue: the Unix/macOS process platform files should include the standard headers they use directly. process_unix.cpp uses std::runtime_error and std::cerr but does not include  `<stdexcept> / <iostream>,` and process_macos.cpp uses std::runtime_error without including <stdexcept>. Please make those translation units self-contained rather than relying on transitive includes.

I will wait to approve till if have reviewed it again in more detail tomorrow.

fl0rianr (approved):
I looked at it again and I don’t see a new code blocker in the refactor. The remaining red CI is isolated to the wrong moonshine .deb/.exe jobs. Hence I'm approving and merging since I prefer having that refactor in earlier so other PRs can prepare.

Issue #1820, opened 2026-05-06
vllm: tool calling via Anthropic API returns 400 Bad Request


ramkrishna2910:
**From PR #1802 review:** https://github.com/lemonade-sdk/lemonade/pull/1802#discussion_r3196382997

Claude Code → Lemonade Anthropic adapter (`/v1/messages`) → vLLM `/v1/chat/completions` → 400 Bad Request. The user types `hello` in Claude Code and sees "Sautéed for 0s" with no response.

```
2026-05-06 10:48:53.400 [Debug] (Server) POST /v1/messages
[OllamaApi] Anthropic compatibility warnings: Ignored unsupported thinking.type: adaptive | Ignored 'metadata' field | Ignored 'context_management' field
2026-05-06 10:48:53.405 [Info] (Process) (APIServer pid=63303) INFO:     127.0.0.1:36804 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
2026-05-06 10:48:53.405 [Error] (StreamingProxy) Backend returned error: 400
2026-05-06 10:48:53.405 [Debug] (Server) POST /v1/messages - 200
```

The Anthropic adapter swallows the 400 and returns 200 to the caller, which makes Claude Code hang silently.

## Investigation needed

Capture the actual 400 response body from vLLM (the warnings shown are just the adapter logging unsupported Anthropic fields it ignored, not the cause of the 400). Likely candidates:

- vLLM 0.20.1 needs `--enable-auto-tool-choice` and a `--tool-call-parser <name>` flag to surface tool calls in OpenAI format. Lemonade's vllm-server launch args (`vllm_server.cpp:140-180`) don't pass either.
- Tool message schema mismatch — Anthropic adapter's `parse_openai_tool_arguments` may produce a request shape vLLM doesn't accept.
- Chat template / stop tokens for tool messages.
- The streaming proxy returning 200 while the upstream returned 400 is itself a bug worth fixing — at minimum the user-visible failure should reflect the upstream status.

## Repro

1. Build/install Lemonade on Linux with `vllm:rocm` installed and a vLLM model loaded.
2. `build/lemonade launch claude --ctx-size 32000 --model Qwen3.5-9B-vLLM`
3. Type any message in Claude Code.
4. Observe: silent hang on the client; 400 in the server log.

## Fix sketch

Two-part:
[...]

PR #2143, opened 2026-06-08
Enable vLLM + Messages API and per model family vLLM args 

### LINKED ISSUE #1964: Better default vLLM arguments (per model family/model)

sawansri:
Resolves #1964, #1820 and enables vLLM with Claude Code (clamping max output tokens, etc)


sawansri:
@claude review

sawansri:
@claude review, make sure that changes made address your previous review

src/cpp/server/backends/vllm_server.cpp
```diff
@@ -309,5 +338,105 @@ void VLLMServer::forward_streaming_request(const std::string& endpoint,
...
+         request["max_tokens"].is_number_unsigned());
+    if (!has_max_completion_tokens && !has_max_tokens) {
+        return request;
+    }
+
+    int64_t requested_max_tokens = has_max_completion_tokens
+        ? request["max_completion_tokens"].get<int64_t>()
+        : request["max_tokens"].get<int64_t>();
+    if (requested_max_tokens <= 0 ||
+        requested_max_tokens <= VLLM_MAX_TOKENS_PREFLIGHT_THRESHOLD) {
```
ramkrishna2910: **[correctness] Clamp never fires for models with context ≤ 8192.**

The early-return when `requested_max_tokens <= VLLM_MAX_TOKENS_PREFLIGHT_THRESHOLD` (8192) means that for a model loaded with e.g. `ctx_size=4096`, a request with a 3000-token prompt and `max_tokens=5000` is *not* clamped — `5000 <= 8192` returns unchanged, then `3000 + 5000 > 4096` and vLLM rejects with a 400.

Since the default `ctx_size` is 4096, **the shipped default config never benefits from this feature**. Consider gating on `min(threshold, max_model_len_)`, or dropping the threshold and always fitting when `max_model_len_` is known (the `/tokenize` round-trip is the only cost, and it only matters near the limit). I reproduced this gap class live on gfx1151.

src/cpp/server/backends/vllm/vllm_server.cpp
```diff
@@ -178,20 +211,14 @@ void VLLMServer::load(const std::string& model_name,
...
-    if (!vllm_args.empty()) {
-        LOG(DEBUG, "vLLM") << "Adding custom arguments: " << vllm_args << std::endl;
-        std::istringstream iss(vllm_args);
-        std::string arg;
-        while (iss >> arg) {
-            args.push_back(arg);
-        }
+    if (!resolved_vllm_args.args.empty()) {
+        LOG(DEBUG, "vLLM") << "Adding model/user arguments from vLLM resolver" << std::endl;
+        args.insert(args.end(), resolved_vllm_args.args.begin(), resolved_vllm_args.args.end());
```
ramkrishna2910, line 295: **[correctness] Duplicate `--quantization` can silently override the forced AWQ kernel.**

`load()` unconditionally pushes `--quantization awq` for AWQ-detected models (line ~202), but `--quantization` is **not** in `protected_flags()`. The resolver dedupes only within its own arg vector — it can't see the `--quantization awq` that `load()` already appended to `args`. So if a family in `vllm_model_config.json` or a user's `vllm_args` also sets `--quantization`, the final argv becomes `--quantization awq … --quantization <other>`, and vLLM's argparse takes the last value — silently discarding the deliberately-forced `awq` kernel (the 2→12 tok/s perf fix). Either emit the AWQ flag through the resolver so it participates in dedup, or protect it.

src/cpp/server/backends/vllm/vllm_server.cpp
```diff
@@ -309,5 +338,105 @@ void VLLMServer::forward_streaming_request(const std::string& endpoint,
...
+        (request["max_completion_tokens"].is_number_integer() ||
+         request["max_completion_tokens"].is_number_unsigned());
+    bool has_max_tokens = request.contains("max_tokens") &&
+        (request["max_tokens"].is_number_integer() ||
+         request["max_tokens"].is_number_unsigned());
+    if (!has_max_completion_tokens && !has_max_tokens) {
+        return request;
+    }
+
+    int64_t requested_max_tokens = has_max_completion_tokens
```
ramkrishna2910, line 431: **[correctness] Both-token-fields requests bypass or distort the clamp.**

When both `max_completion_tokens` and `max_tokens` are present, the decision uses only `max_completion_tokens` but both fields get overwritten:
- `max_completion_tokens=100` (≤8192) + `max_tokens=50000` → early return; the genuinely oversized `max_tokens=50000` is forwarded unclamped and rejected.
- `max_completion_tokens=50000` + `max_tokens=200` → line ~384 overwrites the intentional `max_tokens=200` with ~4014, changing user intent.

Normalizing to a single effective limit before the threshold/fit logic would remove both cases.

src/cpp/server/backends/vllm/vllm_server.cpp
```diff
@@ -309,5 +338,105 @@ void VLLMServer::forward_streaming_request(const std::string& endpoint,
     }
 }
 
+json VLLMServer::fit_openai_max_tokens_to_context(const json& request) {
+    if (max_model_len_ <= 0) {
+        return request;
+    }
+
+    bool has_max_completion_tokens = request.contains("max_completion_tokens") &&
```
ramkrishna2910, line 421: **[correctness, edge] Float-typed token limits bypass clamping.**

The guards accept only `is_number_integer()`/`is_number_unsigned()`. A client/SDK that serializes `"max_tokens": 100000.0` makes both false → `has_max_tokens=false` → early return → the oversized float is forwarded unclamped. Accepting `is_number()` and rounding would close this.

src/cpp/server/backends/vllm/vllm_server.cpp
```diff
@@ -309,5 +338,105 @@ void VLLMServer::forward_streaming_request(const std::string& endpoint,
...
+        tokenize_request["tools"] = request["tools"];
+    }
+    if (request.contains("tool_choice")) {
+        tokenize_request["tool_choice"] = request["tool_choice"];
+    }
+
+    // This is a synchronous backend round trip on the request path. It only
+    // runs for oversized max-token requests so vLLM receives a context-safe
+    // limit before generation or streaming begins.
+    auto response = forward_request("/tokenize", tokenize_request);
```
ramkrishna2910, line 494: **[correctness] `/tokenize` undercounts the real prompt size.**

This tokenizes the raw messages without the chat-template generation prompt / tool-rendering overhead that vLLM applies before generation. `available = max_model_len_ − input_tokens − 64`; if that template/tool overhead exceeds the 64-token `VLLM_TOKEN_FIT_RESERVE`, the clamped value can still overflow context and get rejected despite the fit having run. Worth either using a tokenize call that applies the generation prompt, or a more generous/derived reserve.

ramkrishna2910 (commented):
Automated review (high-effort, recall-biased) plus a live end-to-end test on a gfx1151 Strix Halo box. The primary path works: an oversized `max_tokens`/`max_completion_tokens` against a context **larger than 8192** is clamped correctly and the arg resolver injects family args as expected (verified on `Qwen3.5-0.8B-FP16-vLLM`). Inline comments below flag correctness gaps in the feature and some cleanup/altitude items.

**Would suggest addressing before merge:** the 8192-threshold no-op for small-context models (the default `ctx_size` is 4096, so the shipped default never benefits), the duplicate `--quantization` that can silently override the AWQ kernel, and the both-token-fields clamp bypass. Conventions check came back clean.

src/cpp/server/backends/vllm/vllm_server.cpp
```diff
@@ -133,6 +153,19 @@ void VLLMServer::load(const std::string& model_name,
...
+    json vllm_model_config = json::object();
+    if (fs::exists(utils::path_from_utf8(vllm_model_config_path))) {
+        vllm_model_config = JsonUtils::load_from_file(vllm_model_config_path);
+    } else {
+        LOG(WARNING, "vLLM") << "vLLM model config not found at "
+                             << vllm_model_config_path
+                             << "; continuing with user vllm_args only" << std::endl;
+    }
+    VLLMArgResolution resolved_vllm_args =
+        resolve_vllm_args(model_name, model_id, vllm_model_config, vllm_args);
```
ramkrishna2910, line 228: **[robustness] Config/arg errors here trip the router's nuclear option.**

`resolve_vllm_args` (and `JsonUtils::load_from_file` on a present-but-malformed `vllm_model_config.json`) can throw `std::runtime_error` inside `load()` — bad user `vllm_args`, a protected flag, an unknown family, a bad `checkpoint_regex`, or `schema_version != 1`. Per AGENTS.md, on non-file-not-found load failures the router evicts **all** loaded models and retries. So a deterministic config mistake tears down every other healthy model before a retry that fails identically. Consider validating/normalizing args at config-save time, or classifying these as non-retryable so they don't trigger evict-all.

src/cpp/server/backends/vllm/vllm_server.cpp
```diff
@@ -178,20 +211,14 @@ void VLLMServer::load(const std::string& model_name,
 
     // Avoid vLLM's default gpu_memory_utilization=0.92 on shared-memory systems.
     // Keep this overridable through vllm_args for users that want another limit.
-    if (vllm_args.find("--gpu-memory-utilization") == std::string::npos &&
-        vllm_args.find("--kv-cache-memory-bytes") == std::string::npos) {
+    if (!resolved_vllm_args.has_memory_budget_arg) {
```
ramkrishna2910, line 288: **[behavior] Config-sourced memory flag now suppresses the safe 4G KV-cache default.**

The comment still says “overridable through `vllm_args` for users,” but `has_memory_budget_arg` is now computed over the *resolved* args, which fold in family/model config. A future family carrying `--gpu-memory-utilization`/`--kv-cache-memory-bytes` would silently drop the conservative `--kv-cache-memory-bytes 4G` default on shared-memory systems with no user opt-in, reintroducing the OOM risk this guards against. No shipped family triggers it today, but the source of the flag widened from user-only to user+config — worth a deliberate decision + comment update.

src/cpp/server/backends/vllm/vllm_server.cpp
```diff
@@ -17,6 +21,9 @@ using namespace lemon::utils;
 namespace lemon {
 namespace backends {
 
+static constexpr int64_t VLLM_MAX_TOKENS_PREFLIGHT_THRESHOLD = 8192;
```
ramkrishna2910, line 26: **[altitude] Context-fitting is vLLM-only with magic constants.**

The `8192` threshold and `64` reserve are undocumented magic numbers, and the same oversized-`max_tokens` exposure exists for llama.cpp and other backends that also expose `/tokenize`. Consider lifting context-fitting into a shared request-prep layer keyed off each backend's known context length, with the threshold/reserve derived rather than hardcoded. Finding on line 360 (no-op for ctx ≤ 8192) is a direct symptom of the hardcoded 8192.

src/cpp/server/backends/vllm_server.cpp
```diff
@@ -33,6 +40,18 @@ static std::string parse_quant_method(const std::string& config_json) {
     return "";
 }
 
+static json with_legacy_token_limit(const json& request) {
```
ramkrishna2910: **[reuse] Third copy of the `max_completion_tokens`→`max_tokens` transform.**

`with_legacy_token_limit` duplicates the same mapping already in `llamacpp_server.cpp` and `cloud_server.cpp`. Worth extracting into a shared helper (e.g. in `json_utils`/`backend_utils`) so a fix to the OpenAI-compat aliasing lands in one place.

src/cpp/server/backends/vllm/vllm_arg_resolver.cpp
```diff
@@ -0,0 +1,284 @@
...
+    if (is_negative_number_token(token)) {
+        return false;
+    }
+    if (token[1] == '-') {
+        return token.size() > 2;
+    }
+    return true;
+}
+
+std::vector<ParsedArg> parse_args(const std::string& arg_string, const std::string& source) {
```
ramkrishna2910, line 95: **[reuse] Parallel arg-merge engine.**

This parse→merge→flatten pipeline (`ParsedArg`, `parse_args`, `merge_layer`, manual `negate_flag` handling) re-implements layered `_args` merging that `recipe_options.cpp` already does via `custom_args.h` (`parse_custom_args` / `build_custom_args_map` / `merge_args_maps` / `negate_flag`), including binary-flag negation. Two divergent merge implementations now have to be kept in sync — a fix to quoting/`=`-handling/negation in one won't reach the other. Worth reusing the shared helpers, or consolidating both onto one engine.

(Also: the synchronous `/tokenize` round-trip + double request deep-copy on the streaming path adds to TTFT for oversized streaming requests — efficiency follow-up, gated to >8192 so impact is bounded.)

src/cpp/server/backends/vllm_server.cpp
sawansri: good catch, pushed a fix that uses min(8192, max_model_len_) as the gate condition

src/cpp/server/backends/vllm/vllm_server.cpp
sawansri: Fixed in c27302eb. The resolver now reports the effective resolved --quantization value, and load() only injects --quantization awq when no resolved quantization arg exists. This avoids producing duplicate quantization flags while still preserving explicit user/model config overrides.

ramkrishna2910:
### Outstanding review items (status check)

Went back through the 10 inline review threads against the current branch code. Two are fixed; the rest are still open. Summary so the resolved ones can be closed out and the open ones triaged before merge.

**Addressed in code (threads can be marked resolved):**
- ✅ **Clamp never fires for ctx ≤ 8192** — fixed in `9cc2fc63` (gate now uses `min(8192, max_model_len_)`).
- ✅ **Duplicate `--quantization` silently overriding forced AWQ** — fixed in `c27302eb` (resolver now reports the effective `--quantization`; `load()` only injects `awq` when none is resolved). Composes correctly with the AWQ `--dtype float16` change.

**Still open — correctness / robustness (not touched by either fix commit):**
- ⛔ **Both token fields present bypass/distort the clamp** (`vllm_server.cpp` ~365–394). When both `max_completion_tokens` and `max_tokens` are sent, only `max_completion_tokens` drives the decision, and the clamp path overwrites *both* fields — so e.g. `max_completion_tokens=100` + `max_tokens=50000` still early-returns and forwards the oversized value, and `max_completion_tokens=50000` + `max_tokens=200` clobbers the intentional `200`. Needs the two fields normalized before the decision.
- ⛔ **Float-typed token limits bypass clamping** (~355–360, and same pattern in `count_openai_prompt_tokens` ~430–436). Guards accept only `is_number_integer()`/`is_number_unsigned()`, so `"max_tokens": 100000.0` fails both and is forwarded unclamped.
- ⛔ **`/tokenize` undercounts the prompt** (`count_openai_prompt_tokens` ~403–423). Tokenizes raw messages with no generation-prompt/tool-render overhead, and `VLLM_TOKEN_FIT_RESERVE` is a hardcoded 64 — template/tool overhead beyond 64 tokens can still overflow context near the limit.
- ⛔ **Config/arg errors trip the router's nuclear option** (~159–167). `JsonUtils::load_from_file` and `resolve_vllm_args` throw `std::runtime_error` for deterministic config mistakes (bad flag, unknown family, bad regex, schema mismatch) directly inside `load()`, with no try/catch or non-retryable classification, so a config typo triggers evict-all-and-retry.

**Still open — non-blocking nits (fine to defer, but please resolve/ack the threads):**
- **Config-sourced memory flag suppresses the safe 4G KV-cache default** (~222–223) + the comment still says "overridable through vllm_args for users." No shipped family triggers it today, so no live bug.
- **Magic constants `8192` / `64`** are undocumented and the context-fitting logic is vLLM-only (llama.cpp has the same oversized-`max_tokens` exposure).
- **Third copy of the `max_completion_tokens`→`max_tokens` transform** — `with_legacy_token_limit` duplicates the mapping in `llamacpp_server.cpp` and `cloud_server.cpp`.
- **Parallel arg-merge engine** in `vllm_arg_resolver.cpp` (`ParsedArg`/`parse_args`/`merge_layer`) re-implements layered merging already in `custom_args.h`.

Net: 4 open correctness/robustness items (the token-preflight/clamp logic) and 4 deferrable nits. CI is fully green and the branch is conflict-free; the merge is currently blocked on review approval + these threads.

src/cpp/server/backends/vllm/vllm_server.cpp
sawansri: image

OpenAI API expects ints for both fields, we shouldn't try to normalize floats.

src/cpp/server/backends/vllm/vllm_server.cpp
sawansri: I don’t think this is a correctness issue. When both fields are present, max_completion_tokens is the modern OpenAI-compatible field and should be treated as authoritative. max_tokens is deprecated legacy compatibility. The clamp intentionally uses max_completion_tokens first, and the legacy copy path only fills max_tokens when it is absent. A contradictory request containing both fields is ambiguous client input, not a case where the deprecated field should override the modern one.

src/cpp/server/backends/vllm/vllm_server.cpp
sawansri: This is intentional, users should be able to update flags regarding kv cache in the family/model config and have that override the 4G default.

src/cpp/server/backends/vllm/vllm_server.cpp
sawansri: Addressed in 66ff8266. /tokenize now forwards the chat/template fields that affect vLLM prompt rendering and no longer relies on a hardcoded 64-token reserve. The clamp uses the exact available budget, max_model_len - input_tokens, and lets vLLM report the real error when the prompt itself exceeds context. The threshold gate remains, so normal requests avoid the extra preflight round trip.

src/cpp/server/backends/vllm/vllm_server.cpp
sawansri: Addressed in 7b8df6f9fbbfde29874ecb25bdf677970e6e68d6. I kept this scoped to the vLLM resource file and did not touch the existing config.json path.

vllm_model_config.json is still treated as an optional packaged resource, but load now validates it before merging with user args. If the file is missing, unreadable, malformed, or resolver-invalid, vLLM logs a warning, falls back to an empty model config, and continues with user vllm_args only. When possible, the invalid file is renamed to .corrupted for inspection.

Invalid user-supplied vllm_args still fail normally, since those are explicit user input errors rather than a bad packaged resource.

src/cpp/server/backends/vllm/vllm_arg_resolver.cpp
sawansri: IMO, this isn't really in the scope of this PR and could be tackled in a future one.

src/cpp/server/backends/vllm_server.cpp
sawansri: Implemented in b6b28130587a3575385950bdd6b83583b75cd1af

ramkrishna2910 (approved):

Issue #1066, opened 2026-02-07
`serve` and `run` options cleanup


jeremyfowers:
Right now `lemonade-server serve —help` is pretty messy!

Solution: Organize options by general (host, port, etc.), llm (llamacpp args, etc.), sd, whisper, and kokoro, with nice formatting so it dooesnt look cluttered.

anditherobot:
PR: https://github.com/lemonade-sdk/lemonade/pull/2144

I couldn't find lemonade-server serve --help in the current source, so I grouped the matching options I found on lemonade run --help and lemonade load --help. Let me know if there's another command path I should target.

BillMeyerRSA:
I have other web service using FastAPI doing API keys - how can I get Lemonade server to start with custom port? I don't see option in the Management UI

anditherobot:
This issue can be closed as per  [#2144](https://github.com/lemonade-sdk/lemonade/pull/2144)
The CLI option parser (`lemonade serve --help` and `lemonade run --help`) returns categorized sections

PR #2144, opened 2026-06-08
cli: group run/load recipe options

anditherobot:
Groups the existing run/load recipe flags by backend in the CLI help output so they're easier to scan. No new flags are added.

Tests:
- cmake --build build --config Debug --target lemonade
- CLIHelpDocsConsistencyTests.test_899_run_load_recipe_options_are_grouped
- CLIHelpDocsConsistencyTests.test_900_launch_docs_match_help_text

Partially addresses #1066

### DIFF
--- a/src/cpp/server/recipe_options.cpp
+++ b/src/cpp/server/recipe_options.cpp
@@ -222,20 +222,20 @@ json RecipeOptions::get_option(const std::string& opt) const {
 #ifdef LEMONADE_CLI
 // CLI_OPTIONS used only by the lemonade CLI client for add_cli_options
 static const json CLI_OPTIONS = {
-    {"--ctx-size", {{"option_name", "ctx_size"}, {"type_name", "SIZE"}, {"envname", "LEMONADE_CTX_SIZE"}, {"help", "Context size for the model"}}},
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Thanks for the cleanup — grouping the existing run/load recipe options by backend makes the help output much easier to scan, and the change looks well-scoped.

One small suggestion before approving: the new test verifies that the expected group headers and flags are present, but it does not currently verify that each flag appears under the intended group. Could we strengthen it slightly by slicing/parsing the help output per group and asserting the expected flags within each section? That would better protect the behavior this PR is adding.

Also, since #1066 originally mentions `serve` and `run`, while this PR targets the current `run/ load` recipe option path, it may be worth changing `Closes #1066` to `Partially addresses #1066` unless we are really sure the old  `serve` reference is obsolete - I should know but I'm  not sure atm.

fl0rianr (approved):
Thanks, this addresses my previous feedback.
The change looks well-scoped to the CLI help organization and does not add or alter any recipe flags. LGTM.

CI run and merging has to wait for tomorrow evening at least since main ist blocked for checks and runs are a bit dead right now.

PR #2145, opened 2026-06-08
docs: document pre-v10 legacy Linux config.json path

### LINKED ISSUE #2135: config.json path is incorrect

https://lemonade-server.ai/docs/guide/configuration/

### What's Missing or Wrong

The documentation says:

```
Linux (systemd): /var/lib/lemonade/.cache/lemonade/config.json
```

However, the actual path is:

```
/opt/var/lib/lemonade/.cache/lemonade/config.json
```

### Suggested Improvement

Update the documentation to give the correct path.

anditherobot:
Documents the legacy `/opt/var/lib/lemonade` config path for pre-v10 installs alongside the current path.

```diff
- **Linux (systemd):** `/var/lib/lemonade/.cache/lemonade/config.json`
+ **Linux (systemd):**
+
+  | Install | Path |
+  |---------|------|
+  | Current | `/var/lib/lemonade/.cache/lemonade/config.json` |
+  | Pre-v10 (legacy) | `/opt/var/lib/lemonade/.cache/lemonade/config.json` |
+
+  Upgrading the package migrates data from the legacy path automatically.
```

Closes #2135

### DIFF
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -8,7 +8,11 @@ Lemonade Server starts automatically with the OS after installation. Configurati
 
 If you used an installer from the Lemonade release your `config.json` will be at these locations depending on your OS:
 
-- **Linux (systemd):** `/var/lib/lemonade/.cache/lemonade/config.json`
+- **Linux — `apt`/`.deb` (Debian/Ubuntu):** `/var/lib/lemonade/.cache/lemonade/config.json`
+- **Linux — `dnf`/`.rpm` (Fedora/Red Hat):** `/opt/var/lib/lemonade/.cache/lemonade/config.json`
+
+  > Note: For Debian/Ubuntu, upgrading the package automatically migrates data from the old `/opt/var/lib/lemonade` path to `/var/lib/lemonade`.
+
 - **Windows:** `%USERPROFILE%\.cache\lemonade\config.json`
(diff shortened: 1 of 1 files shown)


fl0rianr (changes requested):
Thanks for taking this on. I think this needs a small cleanup before it is ready.

First, this PR currently includes the unrelated CLI grouping changes from #2144 (recipe_options.cpp and test/server_cli2.py). Could you please rebase/retarget this branch so the diff only contains the configuration documentation change? Otherwise merging this PR first would also merge #2144 unintentionally.

Second, I don’t think `/opt/var/lib/lemonade/.cache/lemonade/config.json` should be described as only “Pre-v10 (legacy)” yet. The Debian postinst script appears to migrate the `lemonade` user from `/opt/var/lib/lemonade` to `/var/lib/lemonade`, but the RPM post-install script still creates the `lemonade` user with home `/opt/var/lib/lemonade`. Since `lemond` defaults to $HOME/.cache/lemonade when no cache directory is passed, current RPM/systemd installs can still use the `/opt/var/lib/...` path.

Could we document this as package/install-type specific instead, or otherwise update the packaging first so the “current” path is consistent across supported Linux packages? The sentence about automatic migration should also be scoped to the package type where that migration actually happens.

fl0rianr (approved):
Thanks, this addresses my feedback.
Content-wise this looks good to me. CI fast to run first also here so please be patient there.

PR #2146, opened 2026-06-09
Fix compat with vscode

superm1:
VScode has two problems.
1) The bearer token parser can have problems
2) It appends ":latest" to the model which we don't need.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -511,6 +511,20 @@ jobs:
           .venv/bin/python test/server_webapp.py --cli-binary ./build/lemonade
           echo "Web-app tests PASSED!"
 
+      - name: Run model name normalization tests (unsigned path)
+        if: steps.check_signing.outputs.has_signing != 'true'
+        shell: bash
+        env:
+          LEMONADE_CI_MODE: "True"
+          PYTHONIOENCODING: utf-8
+          HF_HOME: ${{ github.workspace }}/hf-cache
+          GGML_METAL_NO_RESIDENCY: "1"
+        run: |
+          set -e
+          echo "Running model name normalization tests (:latest suffix)..."
+          .venv/bin/python test/test_model_name_normalization.py --cli-binary ./build/lemonade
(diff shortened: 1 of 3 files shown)


fl0rianr (changes requested):
Thanks for the fix, great catch! I will test later if this actually solves a long term bug I have with vscode.

Now to  the Authorization-header guard looks reasonable.
But I think the `:latest` handling should be centralized before merge. Right now auto_load_model_if_needed() normalizes the name for loading, and handle_chat_completions() normalizes the request before forwarding, but handle_completions() still forwards the original model name/body. Since the Router extracts `model` again from the request when selecting the loaded server, a request like `foo:latest` can load `foo` and then fail routing as `foo:latest`.

There is also an edge case for collection models: the collection branch in handle_chat_completions() checks model_exists(requested_model) before the new request-level normalization, so `collection:latest` can skip the collection orchestrator path.

Could we add a small shared helper, e.g. normalize_client_model_name(), and apply it consistently before model_manager/router lookups and before forwarding request payloads?
What do you think about a regression test for /chat/completions and /completions with `model: "<known-model>:latest"` which could make this much safer?

fl0rianr (approved):
Thanks! looks good now.

Small nit: For the test, error message is tolerated as ok, we could check is message chunks contain "error".

fl0rianr:
Thanks for the quick follow-up — the model-name normalization logic now looks good to me. It is applied before the relevant lookups/forwarding paths, /v1/completions is covered as well, and the streaming regression tests now also guard against SSE error chunks.

One small CI-related issue remains: the new test file currently ends with `run_server_tests()` without passing the required test class. This causes the failing workflow error:

```text
TypeError: run_server_tests() missing 1 required positional argument: 'test_class'
```

Could you update the test entrypoint to pass the class explicitly, e.g.:

```python
def main():
    run_server_tests(
        ModelNameNormalizationTest,
        description="MODEL NAME NORMALIZATION TESTS",
    )

if __name__ == "__main__":
    main()
```

After that rerun, I think CI will be green (as far as you can say that atm)

PR #2147, opened 2026-06-09
fix: ensure whisper.cpp and other backends load correctly in systemd environments

### LINKED ISSUE #2083: whisper.cpp fails to load models when lemond is executed as a systemd service
### Platform

Linux/Arch

### Lemonade Version

10.6.0

### GPU / APU Model

Strix Halo

### Component

whisper.cpp

### Bug Description

When we run lemonade using the systemd service unit, whisper.cpp fails to load models because the XDG_RUNTIME_DIR environment variable is missing for the system user. Adding the following lines fixes it but it might not be the best solution.

```
RuntimeDirectory=lemonade
RuntimeDirectoryMode=0700
Environment="XDG_RUNTIME_DIR=/run/lemonade"
```
related to #1254 
cc @abn 

### Steps to Reproduce

1. Run lemond as a systemd service
2. Try to run a whisper.cpp model

_No response_

### Log Output

```shell
[Info] (Server) Ensuring model loaded: Whisper-Large-v3-Turbo
[Error] (Router) Failed to load model: Unable to resolve writable runtime directory from XDG_RUNTIME_DIR
```

### Additional Context

_No response_

clemperorpenguin:
### Description
This PR addresses an issue where `lemond` and its backend subprocesses (like `whisper.cpp`) crash when running as a `systemd` background service without `XDG_RUNTIME_DIR` provisioned.

Previously, `get_runtime_dir()` on Linux relied strictly on the `XDG_RUNTIME_DIR` environment variable. This PR introduces a secure fallback to the system temporary directory when `XDG_RUNTIME_DIR` is empty or missing. 

Fixes #2083 

### Changes
* Added a fallback mechanism in `get_runtime_dir()` for Linux/BSD that defaults to `/tmp/lemonade-<UID>`.
* **Security measures:** 
  * Uses `mkdir` with strict `0700` permissions to prevent access by other users.
  * Implements an atomic `lstat` verification fallback: if the directory already exists, it confirms that it is owned by the exact UID of the process, is a genuine directory (not a malicious symlink), and retains strict `0700` permissions.
* Included `<sys/stat.h>` for POSIX platforms to enable the low-level directory state inspection.

### Testing
- Tested running `lemond` with `XDG_RUNTIME_DIR` deliberately unset; verified fallback directory is created securely and models load successfully.
- GUI remains unaffected as this primarily impacts UNIX sockets and memory-mapped temp files used by inference backends.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -662,6 +662,43 @@ jobs:
           path: build/lemonade-embeddable-${{ env.LEMONADE_VERSION }}-ubuntu-x64.tar.gz
           retention-days: 7
 
+  build-lemonade-linux-arm64:
+    name: Build Lemonade (Linux ARM64)
+    runs-on: ubuntu-24.04-arm
+    outputs:
+      version: ${{ steps.get_version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          clean: true
+          fetch-depth: 0
+
+      - name: Get version from CMakeLists.txt
+        id: get_version
+        uses: ./.github/actions/get-version
+
+      - name: Install build dependencies
+        run: |
+          set -e
+          sudo apt-get update
+          sudo apt-get install -y cmake ninja-build g++ pkg-config libssl-dev libdrm-dev
+
+      - name: Build
+        run: |
+          set -e
+          cmake --preset default -DBUILD_WEB_APP=OFF
+          cmake --build --preset default --target lemond lemonade
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v7
+        with:
+          name: lemonade-linux-arm64-build
+          path: |
+            build/lemond
+            build/lemonade
+            build/resources/
+          retention-days: 7
+
   build-lemonade-embeddable-windows:
     name: Build Embeddable Lemonade (Windows)
     runs-on: windows-latest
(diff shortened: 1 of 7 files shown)


fl0rianr (commented):
Thanks for this fix! This looks like the right direction and the fallback directory validation is a good security improvement.

One small nitty portability/build issue before merging: this new code uses `errno` and `EEXIST`, but path_utils.cpp does not include <cerrno> / <errno.h> directly. It may compile only because of transitive includes on some toolchains. Could you add an explicit include?

Optional: it may also be worth checking access(lemon_dir.c_str(), W_OK | X_OK) before returning the fallback path, so the final runtime dir is verified as usable, not only owned and mode `0700`.

superm1:
Why isn't XDG_RUNTIME_DIR set in this context?

I appreciate trying to limit access to only your user launching process but I'm not sure that's secure enough. Malware can run as your user process and still evict data.

clemperorpenguin:
> Why isn't XDG_RUNTIME_DIR set in this context?
> 
> I appreciate trying to limit access to only your user launching process but I'm not sure that's secure enough. Malware can run as your user process and still evict data.

When running as a system-level service, there is no active login session attached, and so XDG_RUNTIME_DIR is unset.

Regarding malware, our /tmp/lemonade-<UID> fallback (mode 0700) has the exact same security semantics as XDG_RUNTIME_DIR (which is also mode 0700). Both are vulnerable to malware running under the same user UID, because standard Linux permissions grant equal access to processes under the same user. To protect against same-user malware, we'd have to rely on mandatory access control (like SELinux/AppArmor) or systemd sandboxing (like PrivateTmp=yes), rather than application-level logic.

superm1:
@jfowers 

Hmm, is this a blocker for 10.7 release actually?

> When running as a system-level service, there is no active login session attached, and so XDG_RUNTIME_DIR is unset.

OK - How about falling back to HOME instead?

> To protect against same-user malware, we'd have to rely on mandatory access control (like SELinux/AppArmor) or systemd sandboxing (like PrivateTmp=yes), rather than application-level logic.

But don't we do this already?

```
╰─❯ cat data/lemond.service.in | grep Private
PrivateTmp=yes
```

fl0rianr:
One nuance on the fallback target: I think HOME is less appropriate here than a temp/runtime location.

get_runtime_dir() is used for runtime artifacts such as sockets/locks/temp backend directories, not persistent user state. In the packaged systemd service, HOME is also not the cleanest contract to rely on: the unit runs as User=lemonade, uses `StateDirectory=lemonade` / `WorkingDirectory=%S/lemonade`, and has `ProtectHome=yes`.

The existing `PrivateTmp=yes` actually makes the /tmp fallback more reasonable for the packaged service case: under that unit, the fallback lives in the service's private tmp namespace rather than normal shared /tmp. The 0700 + owner + lstat + W_OK|X_OK checks then cover the non-systemd/manual-service case reasonably well too.

So my read is: for persistent state, `StateDirectory`/cache dir is right; for runtime/temp artifacts, this secure temp fallback is a better fit than HOME.

---
Given the failure mode is a real service deployment blocker for whisper.cpp and the change is isolated to runtime-dir resolution, I still think this is a reasonable low-risk candidate for the final validation window.

superm1:
I guess I see your point.  But shouldn't we be using `/run` for a fallback for stateful stuff then?

fl0rianr:
> I guess I see your point. But shouldn't we be using /run for a fallback for stateful stuff then?

Good point — for the packaged systemd service, /run is the cleaner target.

I think the best split would be:

- lemond.service: provision the runtime dir via systemd, e.g. RuntimeDirectory=lemonade + RuntimeDirectoryMode=0700, and keep the preferred XDG_RUNTIME_DIR path working.
- application code: keep a secure temp fallback for non-packaged/manual/headless launches where no runtime dir was provisioned.

That way the systemd path uses the proper /run lifecycle, while the code still avoids crashing in environments that don’t provide XDG_RUNTIME_DIR.

superm1:
But if you're not running in systemd you _should_ have XDG_RUNTIME_DIR set.  So if we just have a fall back to `/run` (check if `/run` exists for example) then couldn't we keep rejecting `/tmp`?

fl0rianr:
Yeah, I’d be fine with rejecting /tmp if we have a proper /run path.

One caveat: checking that /run exists is probably not enough — it normally exists and is root-owned, so the app should only use /run if there is a pre-provisioned per-service/per-user directory with the expected owner/mode.

For the packaged service that probably means adding RuntimeDirectory=lemonade + RuntimeDirectoryMode=0700 and pointing XDG_RUNTIME_DIR at that. Then the code can keep using the existing XDG path and avoid a /tmp fallback for that case.

superm1:
Oh I know. Just use the library call for `lstat()` to effectively check this:
```
stat /sbin/init
```

If it's systemd you know you can fall back to /run.

superm1:
Actually got a better solution.

Look for `INVOCATION_ID` in the env. This should indicate running under systemd (and thus you can use `/run`).

fl0rianr:
That is a better detection signal than checking /sbin/init, agreed.

INVOCATION_ID tells us this process is actually running as a systemd service invocation, rather than only telling us that the host uses systemd. I’d still keep the actual runtime directory contract explicit though: if we fall back to /run, the specific directory should either be provisioned by RuntimeDirectory= or validated for expected owner/mode before use.

I may be wrong, but the clean version seems to be:
- packaged service: add RuntimeDirectory=lemonade / RuntimeDirectoryMode=0700 and point XDG_RUNTIME_DIR there
- code: if XDG_RUNTIME_DIR is missing under systemd, only use the /run fallback when the expected directory is present and valid; otherwise fail with a clear service configuration error

superm1:
We actually have a helper now that I come to think of it.

```
bool SystemInfo::is_running_under_systemd()
```

Use this.

> I may be wrong, but the clean version seems to be:

Very close!  I think just use the helper and other wise it's a spot on suggestion.

superm1:
I tried to push to your branch to adjust this so we could merge in time for 10.7 and I failed somehow, so I opened a new PR with the changes.

Really appreciate your getting this started, I'll finish getting it landed.

fl0rianr:
@clemperorpenguin Also thanks from my side for the great work and we brought your contribution back into the new PR so you will be contributor number 102 if everything turns out as planned.

PR #2148, opened 2026-06-09
ci: keep Windows inference cache outside workspace

fl0rianr:
Summary:

- Moves Windows .exe inference job runtime/cache/install/temp paths out of the GitHub workspace and into short `C:\lemonade-ci\...` paths.
- Enables Git long-path support before checkout and cleans stale in-workspace cache/install directories before and after checkout.
- Keeps persistent Lemonade/Hugging Face/TheRock cache state outside the workspace while pruning old transient install/tmp run directories.

Is intended to **fix** a large amount of **CI failures** and validation will be done with this real world testing here.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1051,15 +1051,59 @@ jobs:
           $extraArgs = "${{ matrix.extra_args }}" -split " " | Where-Object { $_ }
           $backends = "${{ matrix.backends }}" -split " " | Where-Object { $_ }
 
+          function Invoke-TestBackend {
+            param(
+              [string]$Backend,
+              [int]$MaxAttempts = 1
+            )
+
+            for ($attempt = 1; $attempt -le $MaxAttempts; $attempt++) {
+              Write-Host "Running test/${{ matrix.script }} ${{ matrix.extra_args }} --backend $Backend (attempt $attempt/$MaxAttempts)" -ForegroundColor Cyan
+              & $venvPython test/${{ matrix.script }} @extraArgs --backend $Backend --cli-binary $serverExe
+              $testExitCode = $LASTEXITCODE
+
+              if ($testExitCode -eq 0) {
+                return
+              }
+
+              if ($attempt -lt $MaxAttempts) {
+                Write-Host "::warning title=Retrying transient backend test::test/${{ matrix.script }} --backend $Backend failed with exit code $testExitCode; cleaning sd-server state and retrying."
+
+                Get-Process sd-server -ErrorAction SilentlyContinue |
+                  Stop-Process -Force -ErrorAction SilentlyContinue
+                Start-Sleep -Seconds 20
+
+                try {
+                  Invoke-RestMethod `
+                    -Uri "http://localhost:13305/api/v1/unload" `
+                    -Method Post `
+                    -ContentType "application/json" `
+                    -Body "{}" `
+                    -TimeoutSec 30 | Out-Null
+                } catch {
+                  Write-Host "Unload before retry failed or server had no model loaded: $_" -ForegroundColor Yellow
+                }
+
+                Start-Sleep -Seconds 10
+                continue
+              }
+
+              exit $testExitCode
+            }
+          }
+
           if ($backends.Count -eq 0) {
             Write-Host "Running test/${{ matrix.script }} ${{ matrix.extra_args }}" -ForegroundColor Cyan
             & $venvPython test/${{ matrix.script }} @extraArgs --cli-binary $serverExe
             if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           } else {
             foreach ($backend in $backends) {
-              Write-Host "Running test/${{ matrix.script }} ${{ matrix.extra_args }} --backend $backend" -ForegroundColor Cyan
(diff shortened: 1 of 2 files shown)


jeremyfowers:
@claude review

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -881,25 +881,142 @@ jobs:
...
+          }
+
+          # These legacy locations were inside the repository workspace and could
+          # make actions/checkout fail to clean/reset the repository on Windows,
+          # especially after TheRock created very deep bin/ci-cache/bin/therock trees.
+          if ($env:GITHUB_WORKSPACE) {
+            Remove-LongPathTree (Join-Path $env:GITHUB_WORKSPACE "ci-cache")
+            Remove-LongPathTree (Join-Path $env:GITHUB_WORKSPACE "hf-cache")
+            Remove-LongPathTree (Join-Path $env:GITHUB_WORKSPACE "lemonade_server_install")
+          }
```
jeremyfowers, line 938: Anything inside the workspace is automatically wiped by the github action at the start of the action. Not sure there's a need to manually delete them?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -881,25 +881,142 @@ jobs:
...
+            New-Item -ItemType Directory -Force -Path $path | Out-Null
+          }
+
+          # Remove the legacy in-workspace locations again after checkout in case
+          # they were restored by a previous failed checkout attempt.
+          if ($env:GITHUB_WORKSPACE) {
+            Remove-LongPathTree (Join-Path $env:GITHUB_WORKSPACE "ci-cache")
+            Remove-LongPathTree (Join-Path $env:GITHUB_WORKSPACE "hf-cache")
+            Remove-LongPathTree (Join-Path $env:GITHUB_WORKSPACE "lemonade_server_install")
+          }
```
jeremyfowers: Same comment as https://github.com/lemonade-sdk/lemonade/pull/2148/changes#r3384697040

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -952,10 +1113,87 @@ jobs:
...
         if: always()
         uses: ./.github/actions/cleanup-processes-windows
 
+      - name: Cleanup transient Windows CI directories
+        if: always()
+        shell: PowerShell
+        run: |
+          $ErrorActionPreference = "Continue"
+
+          function Remove-LongPathTree {
```
jeremyfowers: Duplicate of the function declared above?

If you move this code into ./.github/actions/cleanup-processes-windows then we don't need two copies of it.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1684,19 +1922,44 @@ jobs:
...
+          }
+
+          # Keep runtime/cache/install/temp paths outside $GITHUB_WORKSPACE.
+          # This mirrors the .exe inference job and avoids workspace cleanup/long-path failures.
+          echo "HF_HOME=$hfHome" >> $Env:GITHUB_ENV
+          echo "LEMONADE_CACHE_DIR=$cacheDir" >> $Env:GITHUB_ENV
+          echo "LEMONADE_INSTALL_PATH=$installPath" >> $Env:GITHUB_ENV
+          echo "LEMONADE_CI_ROOT=$ciRoot" >> $Env:GITHUB_ENV
+          echo "TEMP=$tmpDir" >> $Env:GITHUB_ENV
+          echo "TMP=$tmpDir" >> $Env:GITHUB_ENV
```
jeremyfowers, line 1888: Why both?

jeremyfowers: Maybe I'm blind but I don't see where these are being used, either. Are they being consumed by the GitHub action itself.

jeremyfowers (changes requested):
See comments

.github/workflows/cpp_server_build_test_release.yml
fl0rianr: Made a seperate file for this.

.github/workflows/cpp_server_build_test_release.yml
fl0rianr: Adressed with comments, github allows that to by used later on.

fl0rianr:
Thanks for the detailed review! I pushed a cleanup refactor that addresses the Windows CI comments:

- Extracted the duplicated long-path-safe directory cleanup into a dedicated local composite action: `.github/actions/cleanup-ci-paths-windows`.
- Kept the pre-checkout cleanup inline because local repository actions are not available until after `actions/checkout`; added a comment explaining why that block intentionally remains inline.
- Added `expected-root` guards for both C:\lemonade-ci cleanup and legacy workspace cleanup so the reusable cleanup action cannot remove paths outside the intended roots.
- Switched the remaining hardcoded Windows install paths to ${{ env.LEMONADE_INSTALL_PATH }}.
- Kept both TEMP and TMP and added a comment because different Windows tools consult either alias in child processes.
- Preserved persistent backend/model caches while only removing current run install/temp directories and pruning older transient install/tmp directories.

The new action uses `shell: powershell` and preserves the existing long-path deletion fallback behavior.

jeremyfowers (approved):
Thanks once again for all you to do ensure CI stability!

PR #2151, opened 2026-06-09
add cuda to llama.cpp auto updater

### LINKED ISSUE #2150: Add llamacpp:cuda to the llamacpp auto-upgrade action
So that manually changing the version number like @fl0rianr did here is not needed: https://github.com/lemonade-sdk/lemonade/pull/2132

fl0rianr:
Add `llamacpp:cuda` to the llama.cpp auto-upgrade workflow.

Closes #2150

## Changes

- Update CUDA to use the same `lemonade-sdk/llama.cpp` release as `rocm-stable`
- Keep build-time and PR-generation backend version updates consistent


jeremyfowers (approved):
LGTM!

PR #2154, opened 2026-06-09
fix(nvidia/snap): CUDA GPU detection under snap strict confinement

kenvandine:
## Summary

- Try `/usr/bin/nvidia-smi` explicitly since `nvidia-smi` may not be in `PATH` under snap strict confinement
- Add `/sys/module/nvidia/version` fallback for driver version detection (readable via `hardware-observe` interface)
- Add NVML `dlopen` fallback that queries `libnvidia-ml.so.1` directly — NVML communicates via `/dev/nvidiactl` (permitted by the `opengl` interface) rather than `/proc/driver/nvidia/` (blocked by AppArmor), and returns full GPU name and compute capability

Fallback chain after this change:
1. `nvidia-smi` (PATH + `/usr/bin/` explicit)
2. NVML via `dlopen` (snap-friendly, full data)
3. `/proc/driver/nvidia/gpus/` (non-snap)
4. `/sys/class/drm` vendor scan (presence only)
5. `lspci`

## Test plan

- [ ] Build and install the snap with `opengl` and `hardware-observe` interfaces connected
- [ ] Confirm `lemonade list` shows NVIDIA GPU models (previously showed "arch unknown")
- [ ] Confirm `lemonade run` works with a CUDA-capable model (e.g. a llama.cpp Vulkan/CUDA model)
- [ ] On a non-snap install, confirm GPU detection still works via the existing nvidia-smi path

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -37,11 +37,6 @@
 #endif
 
 #ifdef __linux__
-#include <unistd.h>
-#include <dlfcn.h>
-#endif
-
-#ifdef __linux__
 #include <dlfcn.h>
 #include <sys/ioctl.h>
 #include <fcntl.h>
@@ -2430,10 +2425,14 @@ static std::vector<NvidiaSmiGpuInfo> query_nvidia_smi() {
         std::string cmd = std::string(smi) + smi_query;
         FILE* pipe = popen(cmd.c_str(), "r");
         if (!pipe) continue;
+        std::string candidate;
         char buffer[512];
-        while (fgets(buffer, sizeof(buffer), pipe) != nullptr) output += buffer;
-        pclose(pipe);
-        if (!output.empty()) break;
+        while (fgets(buffer, sizeof(buffer), pipe) != nullptr) candidate += buffer;
+        int rc = pclose(pipe);
+        if (rc == 0 && !candidate.empty()) {
+            output = candidate;
+            break;
+        }
     }
     if (output.empty()) return result;
 #endif
@@ -2571,14 +2570,14 @@ static std::vector<NvidiaSmiGpuInfo> query_nvidia_nvml() {
             char name_buf[NVML_DEVICE_NAME_BUFFER_SIZE] = {};
             if (nvmlGetName(dev, name_buf, sizeof(name_buf)) != NVML_SUCCESS) continue;
 
-            int major = 0, minor = 0;
-            if (nvmlGetCC(dev, &major, &minor) != NVML_SUCCESS) continue;
-
             NvidiaSmiGpuInfo info;
             info.index          = static_cast<int>(i);
             info.name           = name_buf;
-            info.compute_cap    = std::to_string(major) + "." + std::to_string(minor);
             info.driver_version = driver_version;
+
+            int major = 0, minor = 0;
+            if (nvmlGetCC(dev, &major, &minor) == NVML_SUCCESS)
+                info.compute_cap = std::to_string(major) + "." + std::to_string(minor);
 
             if (nvmlGetMem) {
                 NvmlMemory mem{};
@@ -3176,16 +3175,40 @@ std::vector<GPUInfo> LinuxSystemInfo::get_nvidia_gpu_devices() {
         return gpus;
     }
 
-    // Secondary: /proc/driver/nvidia/gpus/*/information — readable whenever the
+    // Secondary: NVML via dlopen — libnvidia-ml.so.1 communicates with the NVIDIA
+    // kernel module through /dev/nvidiactl (allowed by the snap opengl interface)
+    // rather than /proc/driver/nvidia/, so it works under snap strict confinement.
+    // Returns full name, compute capability, and per-GPU VRAM — same quality as nvidia-smi.
+    {
+        auto nvml_gpus = query_nvidia_nvml();
+        if (!nvml_gpus.empty()) {
+            LOG(WARNING, "SystemInfo") << "nvidia-smi detection failed; NVML library fallback succeeded" << std::endl;
+            for (const auto& nvml : nvml_gpus) {
+                GPUInfo gpu;
+                gpu.index              = nvml.index;
+                gpu.name               = nvml.name;
+                gpu.available          = true;
+                gpu.compute_capability = nvml.compute_cap;
+                gpu.driver_version     = nvml.driver_version.empty()
+                    ? get_nvidia_driver_version() : nvml.driver_version;
+                if (gpu.driver_version.empty()) gpu.driver_version = "Unknown";
+                gpu.vram_gb            = nvml.vram_gb;
+                gpus.push_back(gpu);
+            }
+            return gpus;
+        }
+    }
+
+    // Tertiary: /proc/driver/nvidia/gpus/*/information — readable whenever the
     // nvidia kernel module is loaded, even when the GPU is in Optimus power-save
-    // mode and nvidia-smi fails. Provides the full model name and GPU UUID, which
-    // is enough for identify_cuda_arch_from_name() to determine the sm_XX family.
-    // (No compute_capability here; family is resolved from the name.)
+    // mode and NVML/nvidia-smi fail. Provides the full model name and GPU UUID,
+    // which is enough for identify_cuda_arch_from_name() to determine the sm_XX
+    // family. (No compute_capability here; family is resolved from the name.)
     {
         fs::path gpus_dir = "/proc/driver/nvidia/gpus";
         std::error_code ec;
         if (fs::exists(gpus_dir, ec) && fs::is_directory(gpus_dir, ec)) {
-            LOG(WARNING, "SystemInfo") << "nvidia-smi detection failed; falling back to /proc/driver/nvidia/gpus" << std::endl;
+            LOG(WARNING, "SystemInfo") << "nvidia-smi/NVML detection failed; falling back to /proc/driver/nvidia/gpus" << std::endl;
             std::string driver_version = get_nvidia_driver_version();
             // VRAM is intentionally left unset here: get_nvidia_vram() reads a
             // single memory.total value, which would be wrong if applied to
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@claude review

kenvandine:
I've published a test snap of this on the latest/edge/cuda-fix branch.

```
snap install --channel=latest/edge/cuda-fix lemonade-server
```

Or if you already have the snap installed:

```
snap refresh --channel=latest/edge/cuda-fix lemonade-server
```

fl0rianr (changes requested):
Thanks, this is the right direction overall — NVML via `dlopen` is a good snap-friendly fallback and the Linux link setup already covers that via ${CMAKE_DL_LIBS}.

I’d like to see a few changes before merge:

1. The implemented fallback order does not match the PR description. The code currently tries /proc/driver/nvidia/gpus before NVML. Since /proc gives lower-quality data (no compute capability / per-GPU VRAM) and NVML gives the data this PR is trying to recover, I think NVML should run before /proc, or the PR should explicitly justify the opposite order.

2. The new nvidia-smi probing should not stop at the first command that emits any stdout. A broken wrapper or failed nvidia-smi invocation can produce non-CSV output and prevent /usr/bin/nvidia-smi from being tried. Please collect output per candidate and only accept it after successful exit and/or parseable GPU rows.

3. In the /sys/class/drm fallback, get_nvidia_vram() shells out to nvidia-smi again and returns a single first-GPU value. That is mostly dead in this fallback path and can be wrong for multi-GPU systems. I’d remove it for consistency with the /proc fallback.

4. NVML should not drop an otherwise visible GPU just because nvmlDeviceGetCudaComputeCapability() fails. Please keep the GPU with an empty compute cap so the existing name-based fallback can still classify it.

Minor: remove the duplicate `#include <dlfcn.h>`.

kenvandine:
Thanks for the thorough review @fl0rianr! I've addressed all the feedback in the latest commit (367bfab8):

1. **Duplicate `#include <dlfcn.h>`** — Removed the redundant include block added by this PR; `<dlfcn.h>` was already present in the existing `#ifdef __linux__` block.

2. **Fallback order** — Moved NVML before `/proc/driver/nvidia/gpus`. The new order is: `nvidia-smi → NVML → /proc/driver/nvidia/gpus → /sys/class/drm → lspci`. Updated the PR description and log messages to match.

3. **`get_nvidia_vram()` in `/sys/class/drm` fallback** — Removed it. Added a comment explaining why (shells out to `nvidia-smi`, which is already known unavailable at that point), consistent with the existing comment in the `/proc` fallback.

4. **NVML drops GPU on compute-cap failure** — Changed to keep the GPU entry with an empty `compute_cap` when `nvmlDeviceGetCudaComputeCapability()` fails; `identify_cuda_arch_from_name()` can still classify it from the name.

5. **nvidia-smi output validation** — Now collects output per candidate into a local `candidate` string and only accepts it (assigns to `output` and breaks) when `pclose()` returns 0. A broken wrapper that emits stdout but exits non-zero no longer blocks the `/usr/bin/nvidia-smi` candidate from being tried.

fl0rianr:
Thanks, this addresses the main issues from my previous review. 

One remaining robustness nit: nvmlGetCC is still treated as a required symbol during `dlsym()` setup, even though the per-device call is now optional. I’d suggest removing nvmlGetCC from the required-symbol check and guarding the call with `if (nvmlGetCC && ...)`, so NVML can still return name/driver/VRAM on systems where that symbol is unavailable.

Other than that it is fine.

How to proceed? I don't know if it's useful to make a new test-snap so let me know how to proceed regarding that. I tried to be as fast as possible while still being accurate.

kenvandine:
Good catch. Removed `nvmlGetCC` from the required-symbol guard and added a null check on the per-device call (`if (nvmlGetCC && nvmlGetCC(dev, ...) == NVML_SUCCESS)`), so NVML can still return name/driver/VRAM on systems where that symbol is unavailable. Fixed in bf0fcfc6.

kenvandine:
> Thanks, this addresses the main issues from my previous review.
> 
> One remaining robustness nit: nvmlGetCC is still treated as a required symbol during `dlsym()` setup, even though the per-device call is now optional. I’d suggest removing nvmlGetCC from the required-symbol check and guarding the call with `if (nvmlGetCC && ...)`, so NVML can still return name/driver/VRAM on systems where that symbol is unavailable.
> 
> Other than that it is fine.
> 
> How to proceed? I don't know if it's useful to make a new test-snap so let me know how to proceed regarding that. I tried to be as fast as possible while still being accurate.

I've published an updated build of the snap on that branch, revision 221.

fl0rianr:
Sorry. a blocker from real snap testing on a mixed RTX 5090 + RTX 4090 system.

The current code selects the sm_120 CUDA asset because the RTX 5090 is the highest supported CUDA arch, but the snap starts llama-server with CUDA_VISIBLE_DEVICES=0 and llama.cpp then reports:

- CUDA0 : NVIDIA GeForce RTX 4090
- CUDA : ARCHS = 1200
- CUDA error: no kernel image is available for execution on the device

So we are running the sm_120 build on the sm_89 GPU.

I think the root cause is the snap/NVML path. get_cuda_visible_devices_for_arch() is designed to prefer UUIDs, which is the right approach, but query_nvidia_nvml() currently does not populate `NvidiaSmiGpuInfo::uuid`. As a result, the mixed-arch filtering falls back to numeric ordinals. On this system that ordinal does not map to the same GPU that CUDA ends up exposing in the snap process.

Please add `nvmlDeviceGetUUID` to the NVML fallback and propagate it into `GPUInfo::uuid`, so CUDA_VISIBLE_DEVICES is set to `GPU-...` rather than `0`/`1` for mixed-architecture systems. I would also avoid numeric ordinal fallback when mixed CUDA architectures are present and UUIDs are missing, because guessing the ordinal can select the wrong architecture and crash exactly like this.

This is a blocker for heterogeneous NVIDIA systems.

kenvandine:
Thanks for the thorough testing on the mixed-arch system — that's a real catch.

Fixed in b5355a6f:
- Added `nvmlDeviceGetUUID` to the NVML `dlopen` path so UUIDs are now populated when `nvidia-smi` is unavailable under snap confinement.
- Removed the numeric ordinal fallback in `get_cuda_visible_devices_for_arch()` — when mixed architectures are present but a GPU has no UUID, we skip it rather than guessing an ordinal. Returning `""` (no restriction) is safer than exposing the wrong device and crashing.

kenvandine:
> Thanks for the thorough testing on the mixed-arch system — that's a real catch.
> 
> Fixed in [b5355a6](https://github.com/lemonade-sdk/lemonade/commit/b5355a6ff1c32e13adc0c83e873b80c182dfddbb):
> 
> * Added `nvmlDeviceGetUUID` to the NVML `dlopen` path so UUIDs are now populated when `nvidia-smi` is unavailable under snap confinement.
> * Removed the numeric ordinal fallback in `get_cuda_visible_devices_for_arch()` — when mixed architectures are present but a GPU has no UUID, we skip it rather than guessing an ordinal. Returning `""` (no restriction) is safer than exposing the wrong device and crashing.

Updated the snap too:
Revision 222 created for 'lemonade-server' and released to 'latest/edge/cuda-fix'

fl0rianr:
I tested the updated snap again on a mixed RTX 5090 + RTX 4090 system.

The behavior changed, but the issue is still there:

* Previously, Lemonade set CUDA_VISIBLE_DEVICES=0, but that exposed the wrong device in the snap runtime: CUDA0 was the RTX 4090, while the selected backend asset was sm_120 for the RTX 5090.
* With the latest update, the wrong ordinal is no longer used, which is good.
* However, now CUDA_VISIBLE_DEVICES is not restricted at all, so llama.cpp sees both GPUs:

  * CUDA0 = RTX 4090
  * CUDA1 = RTX 5090

Since llama.cpp starts on current device 0, it still tries to execute the sm_120 CUDA build on the RTX 4090 and crashes as expected.

I think the minimal fix is that the NVML UUID is read but not propagated into GPUInfo.

query_nvidia_nvml() now fills NvidiaSmiGpuInfo::uuid via nvmlDeviceGetUUID, but in LinuxSystemInfo::get_nvidia_gpu_devices() the NVML fallback copies index/name/compute_cap/driver/vram and never assigns:

```cpp
gpu.uuid = nvml.uuid;
```

So get_cuda_visible_devices_for_arch() has no UUID to expose for the matching sm_120 GPU. Because the numeric ordinal fallback was removed, it returns an empty restriction, leaving both GPUs visible.

Suggested minimal patch:

```cpp
GPUInfo gpu;
gpu.index              = nvml.index;
gpu.uuid               = nvml.uuid;
gpu.name               = nvml.name;
gpu.available          = true;
gpu.compute_capability = nvml.compute_cap;
gpu.driver_version     = nvml.driver_version.empty()
    ? get_nvidia_driver_version() : nvml.driver_version;
if (gpu.driver_version.empty()) gpu.driver_version = "Unknown";
gpu.vram_gb            = nvml.vram_gb;
gpus.push_back(gpu);
```

Then llama.cpp should see only the RTX 5090 as CUDA0, and the sm_120 asset will run on the correct device.

kenvandine:
Good catch — the UUID was being queried from NVML but never assigned to `GPUInfo` in the NVML fallback path of `get_nvidia_gpu_devices()`. Fixed in aa3ca4ea with the one-line addition:

```cpp
gpu.uuid = nvml.uuid;
```

`get_cuda_visible_devices_for_arch()` should now find the UUID for the sm_120 RTX 5090 and set `CUDA_VISIBLE_DEVICES` to its `GPU-...` UUID, making it the only device visible to llama.cpp.

kenvandine:
Pushed both fixes — the branch is now up to date:

- `b5355a6f` — add `nvmlDeviceGetUUID` to the NVML `dlopen` path and remove the unsafe numeric ordinal fallback
- `0c122d09` — propagate `uuid` from `NvidiaSmiGpuInfo` into `GPUInfo` in the NVML path of `get_nvidia_gpu_devices()`

On the mixed RTX 5090 + RTX 4090 system, `CUDA_VISIBLE_DEVICES` should now be set to the `GPU-...` UUID of the RTX 5090 only, so llama.cpp sees it as the sole CUDA0 and the sm_120 asset runs on the correct device.

fl0rianr (approved):
Approved from my side after testing the updated snap on a mixed RTX 5090 + RTX 4090 system. With and without snap.
5090 is visible which matches the downloaded artifact.
Great job and thanks for iterating!

PR #2155, opened 2026-06-09
vllm: rename built-in models with -FP16- suffix

### LINKED ISSUE #1826: vllm: Clarify the data type
Seems like all vllm models in server_models.json are fp16? This breaks with the convention in llamacpp, raisw, and flm where all models are q4. Should they be clearly marked as fp16 in server_models.json? Somewhere in the docs?

cc @ramkrishna2910

ramkrishna2910:
## Summary
- Rename the four built-in `vllm` entries in `server_models.json` to embed `-FP16-` in the model ID — e.g. `Qwen3.5-4B-vLLM` → `Qwen3.5-4B-FP16-vLLM`. This matches the `-Hybrid` / `-CPU` suffix convention already used by `ryzenai-llm` and makes the data type obvious next to `llamacpp` (Q4_K_M) and `flm` (4-bit) entries in the same list.
- Update the one `Qwen3.5-0.8B-vLLM` placeholder in `AddModelPanel.tsx` (Add Model panel example) and the quickstart in `docs/news/vllm-rocm.html` so existing references stay correct.
- Add a short paragraph to `docs/guide/configuration/vllm.md` documenting the naming convention and noting that pointing a `user.*` `vllm` registration at a pre-quantized checkpoint (FP8, AWQ, GPTQ, …) is still supported.

> **Heads-up — breaking change to public model IDs.** Existing pulls/downloads under the old `Qwen3.5-*-vLLM` names will no longer match a built-in registry entry. vLLM is still flagged experimental, so this is the cheapest time to do this; deferring it would mean shipping a deprecation alias later.

Closes #1826.

## Test plan
- [ ] `lemonade list` shows the renamed entries: `Qwen3.5-0.8B-FP16-vLLM`, `Qwen3.5-2B-FP16-vLLM`, `Qwen3.5-4B-FP16-vLLM`, `Qwen3.5-9B-FP16-vLLM`.
- [ ] `lemonade run Qwen3.5-4B-FP16-vLLM` loads, downloads if needed, and serves chat completions as before.
- [ ] `GET /api/v1/models` returns the renamed IDs.
- [ ] Add Model panel in the desktop app shows the new placeholder name when recipe is set to `vllm`.
- [ ] Docs site renders the new paragraph in `docs/guide/configuration/vllm.md` cleanly; quickstart on the vllm-rocm blog page is consistent.

### DIFF
--- a/docs/guide/configuration/vllm.md
+++ b/docs/guide/configuration/vllm.md
@@ -35,7 +35,9 @@ The install fetches a per-GPU-target release (e.g. `…-gfx1151`, `…-gfx1150`)
 
 ## Use
 
-Models registered with the `vllm` recipe in [`server_models.json`](https://github.com/lemonade-sdk/lemonade/blob/main/src/cpp/resources/server_models.json) load automatically on first request. To register your own:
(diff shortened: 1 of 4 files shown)


sawansri (approved):
LGTM! This is a lot more clear.

PR #2159, opened 2026-06-09
Add triage dashboard with GitHub Pages auto-publish

ramkrishna2910:
## Summary

Introduces a self-contained triage dashboard for open issues + PRs, plus a workflow that regenerates it every ~15 minutes and publishes it to a **dedicated repo's** GitHub Pages.

- **`.github/scripts/triage_dashboard.py`** — fetches open items via `gh api`, emits a single HTML file with:
  - Histograms per label family (engine::, area::, runtime::, priority::, type, component, other, unlabeled)
  - An **Assignee** family showing assigned-vs-unassigned and per-user counts
  - **Date range filter** (filters by creation date)
  - Multi-filter drill-down: labels AND together; assignees within the Assignee family OR together
- **`.github/scripts/test_triage_dashboard.py`** — regression tests, run in the workflow before publishing.
- **`.github/workflows/triage-dashboard.yml`** — schedule (`7,22,37,52 * * * *`), `workflow_dispatch`, and on script/workflow pushes; builds the HTML and **pushes it to `lemonade-sdk/lemonade-triage` (`main` branch)**, which serves it at `https://lemonade-sdk.github.io/lemonade-triage/`.

### Why a separate repo (addresses @fl0rianr / @jeremyfowers review)

This repo's GitHub Pages already serves the production **lemonade-server.ai** site from the `website` branch (deploy-from-branch), and that site is only meant to refresh at release (`v*` tag) boundaries. A repo has exactly one Pages source, so deploying the dashboard from *this* repo's Pages would (a) collide with / replace the website and (b) republish it every 15 min. Publishing to a dedicated repo's Pages keeps the dashboard fully isolated — same pattern the org already uses for `lemonade-sdk/flatpak` → `lemonade-sdk.github.io/flatpak/`.

## Setup status

- ✅ **`lemonade-sdk/lemonade-triage`** created (public), Pages set to **Deploy from a branch → `main` / (root)** → `https://lemonade-sdk.github.io/lemonade-triage/`, and seeded with a current dashboard.
- ⏳ **Remaining (admin):** add a secret **`TRIAGE_DASHBOARD_TOKEN`** to *this* repo — a fine-grained PAT (or deploy token) with **contents: write** on `lemonade-sdk/lemonade-triage` — so the scheduled workflow can push refreshes.

image

### DIFF
--- a/.github/scripts/triage_dashboard.py
+++ b/.github/scripts/triage_dashboard.py
@@ -113,6 +113,9 @@ HTML_TEMPLATE = """<!doctype html>
   .family { margin-bottom: 24px; }
   .family h2 { font-size: 12px; font-weight: 600; text-transform: uppercase; letter-spacing: 0.5px; color: var(--muted); margin: 0 0 8px; display: flex; align-items: center; gap: 8px; }
   .family h2 .swatch { width: 10px; height: 10px; border-radius: 2px; display: inline-block; }
+  .family h2 .chevron { display: inline-block; width: 10px; color: var(--muted); font-size: 10px; }
+  .family h2 .family-count { color: var(--muted); font-weight: 500; font-variant-numeric: tabular-nums; margin-left: 4px; }
+  .family.collapsible h2:hover { color: var(--fg); }
   .label-row { display: grid; grid-template-columns: 1fr 36px; gap: 8px; align-items: center; padding: 4px 6px; border-radius: 4px; cursor: pointer; margin: 2px 0; }
   .label-row:hover { background: var(--panel); }
   .label-row.active { background: var(--panel-2); outline: 1px solid var(--link); }
@@ -177,7 +180,7 @@ HTML_TEMPLATE = """<!doctype html>
 <div class="footer">Re-run <code>python .github/scripts/triage_dashboard.py</code> to refresh.</div>
 <script>
 const DATA = __DATA__;
-const FAMILY_ORDER = ["assignee::", "engine::", "area::", "runtime::", "priority::", "type", "component", "other", "__unlabeled__"];
+const FAMILY_ORDER = ["engine::", "area::", "runtime::", "priority::", "type", "component", "other", "__unlabeled__", "assignee::"];
 const FAMILY_META = {
   "assignee::":    {name: "Assignee",   color: "var(--link)",       description: "Assigned vs. unassigned · click a user to filter"},
   "engine::":      {name: "Engine",     color: "var(--engine)",     description: "Backend / inference engine"},
@@ -231,6 +234,7 @@ function buildFamilies(items) {
 let activeFilters = new Set();
 let dateFrom = null;  // "YYYY-MM-DD" or null
 let dateTo = null;    // "YYYY-MM-DD" or null
+const collapsedFamilies = new Set(["assignee::"]);  // assignee family is hidden by default
 const families = buildFamilies(DATA);
 
 function prettyFilterName(f) {
@@ -259,22 +263,36 @@ function renderFamilies() {
     const labels = families[fname];
     if (Object.keys(labels).length === 0) continue;
     const meta = FAMILY_META[fname];
-    const max = Math.max(...Object.values(labels));
-    const section = document.createElement("div");
-    section.className = "family";
-    section.innerHTML = `<h2><span class="swatch" style="background:${meta.color}"></span>${meta.name} <span style="font-weight:400; text-transform:none; letter-spacing:0; opacity:0.7;">— ${meta.description}</span></h2>`;
     const entries = Object.entries(labels).sort((a, b) => b[1] - a[1]);
-    for (const [lbl, count] of entries) {
-      const row = document.createElement("div");
-      row.className = "label-row" + (activeFilters.has(lbl) ? " active" : "");
-      const pct = (count / max * 100).toFixed(0);
-      const displayName = prettyFilterName(lbl);
-      row.innerHTML = `
-        <div class="label-name">${escape(displayName)}<div class="bar" style="background:${meta.color}; width:${pct}%"></div></div>
-        <div class="label-count">${count}</div>
-      `;
-      row.addEventListener("click", () => toggleFilter(lbl));
-      section.appendChild(row);
+    const total = entries.reduce((s, e) => s + e[1], 0);
+    const max = Math.max(...Object.values(labels));
+    const collapsible = fname === "assignee::";
+    const isCollapsed = collapsible && collapsedFamilies.has(fname);
+    const section = document.createElement("div");
+    section.className = "family" + (collapsible ? " collapsible" : "");
+    const chevron = collapsible ? `<span class="chevron">${isCollapsed ? "▸" : "▾"}</span>` : "";
+    const countBadge = collapsible ? `<span class="family-count">${total}</span>` : "";
+    section.innerHTML = `<h2${collapsible ? ' style="cursor:pointer; user-select:none;"' : ""}>${chevron}<span class="swatch" style="background:${meta.color}"></span>${meta.name}${countBadge} <span style="font-weight:400; text-transform:none; letter-spacing:0; opacity:0.7;">— ${meta.description}</span></h2>`;
+    if (collapsible) {
+      section.querySelector("h2").addEventListener("click", () => {
+        if (collapsedFamilies.has(fname)) collapsedFamilies.delete(fname);
+        else collapsedFamilies.add(fname);
+        renderFamilies();
+      });
+    }
+    if (!isCollapsed) {
+      for (const [lbl, count] of entries) {
+        const row = document.createElement("div");
+        row.className = "label-row" + (activeFilters.has(lbl) ? " active" : "");
+        const pct = (count / max * 100).toFixed(0);
+        const displayName = prettyFilterName(lbl);
+        row.innerHTML = `
+          <div class="label-name">${escape(displayName)}<div class="bar" style="background:${meta.color}; width:${pct}%"></div></div>
+          <div class="label-count">${count}</div>
+        `;
+        row.addEventListener("click", () => toggleFilter(lbl));
+        section.appendChild(row);
+      }
     }
     wrap.appendChild(section);
   }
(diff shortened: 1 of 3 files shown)


.github/scripts/triage_dashboard.py
```diff
@@ -0,0 +1,477 @@
...
+    print(f"Fetching open issues + PRs from {args.repo}…", file=sys.stderr)
+    items = fetch_items(args.repo)
+    print(f"  fetched {len(items)} items", file=sys.stderr)
+
+    now = datetime.now(timezone.utc).strftime("%Y-%m-%d %H:%M UTC")
+    html = (
+        HTML_TEMPLATE
+        .replace("__REPO__", args.repo)
+        .replace("__GENERATED_AT__", now)
+        .replace("__DATA__", json.dumps(items))
```
fl0rianr: This embeds GitHub-provided issue/PR data directly into a `<script>` block. The later DOM escaping does not protect this injection point: a title/label containing `</script>` can terminate the script tag before the JS renderer runs.

Please serialize this with script-safe JSON escaping, e.g. escape <, >, &, U+2028 and U+2029 before replacing __DATA__, and ideally add a regression case with a malicious title such as `</script><script>alert(1)</script>`.

.github/workflows/triage-dashboard.yml
```diff
@@ -0,0 +1,53 @@
...
+    branches: [main]
+    paths:
+      - ".github/scripts/triage_dashboard.py"
+      - ".github/workflows/triage-dashboard.yml"
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
```
fl0rianr: Can we scope these permissions per job instead of granting Pages/OIDC permissions workflow-wide?

The build job only needs to check out the repo and read issues/PRs, while pages: write and id-token: write are only required by the deploy job. GitHub also recommends granting the GITHUB_TOKEN the least required access, and the Pages docs show pages: write / id-token: write as deploy-job requirements rather than build-job requirements.

Suggested shape:

```yaml
jobs:
  build:
    permissions:
      contents: read
      issues: read

  deploy:
    permissions:
      pages: write
      id-token: write

File: .github/workflows/triage-dashboard.yml [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          mkdir -p _site
+          python .github/scripts/triage_dashboard.py \
+            --repo "${{ github.repository }}" \
+            --output _site/index.html
+
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: _site
```
fl0rianr: Nit: The current GitHub Pages docs use actions/upload-pages-artifact@v4. Since this PR introduces a new Pages workflow, can we start on the current major version here?

.github/scripts/triage_dashboard.py
```diff
@@ -0,0 +1,477 @@
...
+        if label.startswith(prefix):
+            return prefix
+    if label in TYPE_LABELS:
+        return "type"
+    if label in COMPONENT_LABELS:
+        return "component"
+    return "other"
+
+
+def build_summary(items):
```
fl0rianr: Looks unused after the dashboard logic moved client-side. Can we remove this helper to keep the script smaller?

fl0rianr (changes requested):
Thanks for adding this — the dashboard is useful, but not fully ready IMO see comments below:

.github/scripts/triage_dashboard.py
ramkrishna2910: Fixed in 4016b2fe. Added a `script_safe_json()` helper that escapes `<`, `>`, `&` to `<`/`>`/`&` plus `U+2028`/`U+2029`, and `__DATA__` is now serialized through it instead of raw `json.dumps`. A `</script>` in any title/label can no longer terminate the script tag. Added regression tests in `.github/scripts/test_triage_dashboard.py` covering the `</script><script>alert(1)</script>` payload (both at the serializer level and end-to-end on the generated HTML).

.github/workflows/triage-dashboard.yml
ramkrishna2910: Done in 4016b2fe. Removed the workflow-wide `permissions` block and scoped per job: `build` now has only `contents: read` + `issues: read`, and `deploy` has `pages: write` + `id-token: write`.

.github/workflows/triage-dashboard.yml
ramkrishna2910: Done in 4016b2fe — bumped to `actions/upload-pages-artifact@v4`.

.github/scripts/triage_dashboard.py
ramkrishna2910: Removed in 4016b2fe. Confirmed it had no remaining callers (the summary is computed client-side in `buildFamilies`).

.github/workflows/triage-dashboard.yml
```diff
@@ -0,0 +1,54 @@
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Generate dashboard
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          mkdir -p _site
+          python .github/scripts/triage_dashboard.py \
```
fl0rianr, line 48: Nice, thanks for adding the regression tests. Could we also run them in this workflow before generating/publishing the dashboard?

Right now the Pages workflow only generates _site/index.html; the new test_triage_dashboard.py is present, but I do not see it being executed here. Since this test covers the script-context escaping security fix, it would be good to make the build fail before publishing if this regresses.

For example:

```yaml
      - name: Run dashboard tests
        run: python .github/scripts/test_triage_dashboard.py

File: .github/workflows/triage-dashboard.yml [hunk 6]
```diff
@@ -0,0 +1,54 @@
+name: Triage dashboard
+
+on:
+  schedule:
+    - cron: "*/15 * * * *"
```
fl0rianr: Small scheduling nit: this still runs at minute 0 of every hour. GitHub documents that scheduled workflows can be delayed or even dropped during high-load periods, and specifically mentions the start of every hour. Could we offset the same 15-minute cadence, e.g. `7/15 * * * *` or `13,28,43,58 * * * *`?

fl0rianr (commented):
Thanks for the quick fixes — I re-checked and the main issues from my previous review look addressed now.

I left two small follow-ups: wiring the new regression tests into the dashboard workflow, and offsetting the scheduled run away from minute 0. Neither changes the overall direction — this looks close.

.github/workflows/triage-dashboard.yml
ramkrishna2910: Done in 846d1e88. Added a **Run dashboard tests** step that runs `python .github/scripts/test_triage_dashboard.py` before the dashboard is generated, so the build fails before publishing if the script-context escaping fix ever regresses.

.github/workflows/triage-dashboard.yml
ramkrishna2910: Done in 846d1e88. Offset the same 15-minute cadence off minute 0 — cron is now `7,22,37,52 * * * *`.

fl0rianr:
Small follow-up: since this workflow now runs `.github/scripts/test_triage_dashboard.py`, should we include that file in the `push.paths` filter too?

Otherwise changes that only touch the test file will not trigger this workflow.

```yaml
      - ".github/scripts/test_triage_dashboard.py"
```
After you consider that, let me know and I'll approve.

.github/workflows/triage-dashboard.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - id: deployment
+        uses: actions/deploy-pages@v4
```
jeremyfowers: Ensure that this doesn't collide with the rest of the website.

Ensure that this doesn't update the entire website between release boundaries.

ramkrishna2910:
Good catch — done in b982d44. Added `.github/scripts/test_triage_dashboard.py` to the `push.paths` filter so test-only changes also trigger the workflow. Ready for another look. 🙏

.github/workflows/triage-dashboard.yml
ramkrishna2910: Good catch — both confirmed and now fixed.

I checked the live Pages config: this repo serves **lemonade-server.ai** via deploy-from-branch (`website` branch, `/docs`, CNAME `lemonade-server.ai`), built by `publish-website.yml` only on `v*` tags. The dashboard's old `actions/deploy-pages` flow needed this repo's single Pages source flipped to "GitHub Actions", which would have (1) replaced the website and (2) republished it every 15 min — exactly your two concerns.

Reworked the workflow to push the generated HTML to a **dedicated repo's** Pages instead — `lemonade-sdk/triage` (`gh-pages`) → `lemonade-sdk.github.io/triage/` — so it never touches this repo's Pages. Same pattern already used for `lemonade-sdk/flatpak`. Updated the PR description with the one-time setup (create `lemonade-sdk/triage`, point its Pages at `gh-pages`, add a `TRIAGE_DASHBOARD_TOKEN` secret with contents:write on that repo).

One thing that needs an admin/org action before merge: creating `lemonade-sdk/triage` and adding the token secret — happy to coordinate on naming if you'd prefer a different repo.

ramkrishna2910: Added a new repo so we dont have github pages conflict.

fl0rianr (approved):
LGTM, thanks for the great work.

**Please** make sure TRIAGE_DASHBOARD_TOKEN is added to this repo with `contents: write` on lemonade-sdk/lemonade-triage

ramkrishna2910:
> LGTM, thanks for the great work.
> 
> **Please** make sure TRIAGE_DASHBOARD_TOKEN is added to this repo with `contents: write` on lemonade-sdk/lemonade-triage

This has been added.

PR #2165, opened 2026-06-09
fix: corrupted dowload speed/eta during finalizing

### LINKED ISSUE #2162: Download manager in the app shows small and undefined download speed during SHA check
Image

Right now the model downloads, then the SHA check begins. Nothing is downloaded during the SHA check, so the download rate rolling average trends towards 0 B/s, and then becomes undefined.

Maybe it ought to say "checking SHA" instead of this?

fl0rianr:
## Summary

* Show `finalizing...` after all known download bytes have been transferred but the job is still running
* Hide download speed during this finalization phase to avoid misleading `0 B/s` or undefined speed values
* Keep the change frontend-only and minimal for SHA checks and backend unpack/install finalization

Closes #2162

### DIFF
--- a/src/app/src/renderer/DownloadManager.tsx
+++ b/src/app/src/renderer/DownloadManager.tsx
@@ -184,6 +184,13 @@ const DownloadManager: React.FC<DownloadManagerProps> = ({ isVisible, onClose })
     }
   };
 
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@claude review

jeremyfowers (approved):
Thanks for the quick fix!

PR #2169, opened 2026-06-10
ci: wire HF_TOKEN into linux distro builds; drop Fedora

jeremyfowers:
The purpose of this PR is to reduce the amount of CI failures from distro builds getting rate limited by HF.

Applies `HF_TOKEN` to `.github/workflows/linux_distro_builds.yml`, following the same pattern #2105 used for `cpp_server_build_test_release.yml`.

Also removes the **Fedora** distro from the build matrix (Arch, Debian, and openSUSE remain). The Fedora build was not needed in this workflow since we have 2 Fedora builds in the main release workflow.

### DIFF
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -33,6 +33,8 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ${{ matrix.image }}
+    env:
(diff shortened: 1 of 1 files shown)


fl0rianr (approved):
Looks good!

But please confirm the empty-token fork path remains safe.

PR #2170, opened 2026-06-10
ci: drop src/cpp from distro build triggers

jeremyfowers:
\Removes `"src/cpp/**"` from the `pull_request` path filters in `.github/workflows/linux_distro_builds.yml`, so changes to C++ source no longer trigger the Linux distro builds.

IMO we should not need to trigger on C++ changes, since the distro builds should only have true negative results if we change the dependencies or tests.

LMK if you think this is true @fl0rianr.

### DIFF
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -9,7 +9,6 @@ on:
       - "**/CMakeLists.txt"
       - "CMakePresets.json"
       - "cmake/**"
-      - "src/cpp/**"
       - "setup.sh"
       - "test/server_cli2.py"
       - "test/server_endpoints.py"


fl0rianr (commented):
Great idea overall — this should save a lot of workflow time.

Removing src/cpp/** from `Linux Distro Builds` means C++-only PRs will no longer get this distro-specific build/test coverage. That seems fine to me as long as we are not treating this workflow as the primary compile gate.

Ordinary C++ compile failures should still be caught by `C++ Server Build, Test, and Release`, since that workflow runs on every PR and on merge queue events independently of this path filter. So I’m okay with this change if that workflow is the required gate for C++ changes.

The only caveat I see is around urgent/manual merges. This change does not make it impossible for a compile error to reach **main** if someone with bypass/admin permissions merges despite failing or pending checks. In that case, having fewer independent failures can make a real compile issue easier to overlook, especially when CI is already noisy or failing for unrelated reasons. Multiple failing checks tend to draw more attention than one failure buried among the usual CI noise.

Do you think we should explicitly rely on `C++ Server Build, Test, and Release` as the single required compile signal here? If yes, I’m fine with dropping the distro trigger for src/cpp/**.

jeremyfowers:
> This change does not make it impossible for a compile error to reach main if someone with bypass/admin permissions merges despite failing or pending checks

This is true regardless of this change, no? Let's try to avoid force-merging PRs with failing checks as a general rule.

The distro build CI checks will still run on merge to main regardless of this change, which will give us a signal within minutes of whether main is healthy. That signal is persistent on the front page of the repo until the next merge. So it is unlikely to get buried in CI noise the way a PR check can be.

fl0rianr (approved):
I just wanted to express my concern. Here you go.

PR #2173, opened 2026-06-10
systemd: add user-service symlink at /usr/lib/systemd/user/

### LINKED ISSUE #2118: RPM: lemond user service unit installed to /opt/lib/systemd/user/ instead of /usr/lib/systemd/user/ — service never auto-starts
## Platform

Linux / Fedora 44

## Lemonade Version

10.6.0

## Component

RPM packaging / systemd

## Bug Description

The Fedora RPM installs the user-level systemd unit file to `/opt/lib/systemd/user/lemond.service` instead of the standard `/usr/lib/systemd/user/lemond.service`. Because `/opt/lib/systemd/user/` is not in systemd's default unit search path, the service is never discovered, and `lemond` does not auto-start after install or reboot.

Running `systemctl --user start lemond` fails immediately with:

```
Failed to start lemond.service: Unit lemond.service not found.
```

Note: a system-level unit **is** correctly installed to `/usr/lib/systemd/system/lemond.service`, so that path is fine. Only the user-level unit path is wrong.

## Steps to Reproduce

1. Install `lemonade-server-10.6.0-fc44.x86_64.rpm` on Fedora 44
2. Reboot
3. Run `systemctl --user start lemond`

**Result:** `Unit lemond.service not found.`

**Expected:** The user-level unit is installed to `/usr/lib/systemd/user/lemond.service` and `systemctl --user enable --now lemond` works out of the box after install.

**Actual:** Unit is at `/opt/lib/systemd/user/lemond.service` — systemd cannot find it, service never starts.

## Workaround

```bash
mkdir -p ~/.config/systemd/user
ln -s /opt/lib/systemd/user/lemond.service ~/.config/systemd/user/lemond.service
systemctl --user daemon-reload
systemctl --user enable --now lemond
```

## Fix

In the RPM spec, change the install destination for the user unit from `/opt/lib/systemd/user/` to `/usr/lib/systemd/user/`.

lucifer-vali:
## Resolves #2118

- The user-mode systemd unit is not discoverable by `systemctl --user` when Lemonade is installed to a non-standard prefix
- This adds a symlink in the standard systemd user search path so `systemctl --user enable lemond.service` works out of the box


fl0rianr (changes requested):
Thanks for fixing this. I checked the built Fedora RPM from this PR, and it does contain /usr/lib/systemd/user/lemond.service as a symlink to /opt/lib/systemd/user/lemond.service, so the systemd discovery problem from #2118 should be addressed.

Before merging, I think this should be hardened a bit. The new install block is in the generic Linux install path, not an RPM-only path, and execute_process() currently does not check whether `cmake -E create_symlink` actually succeeded. If symlink creation fails because of permissions, an existing file, or another packaging edge case, the install/package step could still continue and we would only notice at runtime.

Could we make the symlink creation fail-safe and add an explicit RPM package-content check? For example:

```cmake
function(lemonade_install_systemd_symlink link_path target_path)
    get_filename_component(_link_dir "${link_path}" DIRECTORY)

    install(CODE "
        set(_link \"\$ENV{DESTDIR}${link_path}\")
        set(_target \"${target_path}\")

        file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${_link_dir}\")

        if(IS_SYMLINK \"\${_link}\")
            file(REMOVE \"\${_link}\")
        elseif(EXISTS \"\${_link}\")
            message(FATAL_ERROR \"Refusing to replace existing non-symlink: \${_link}\")
        endif()

        execute_process(
            COMMAND \"${CMAKE_COMMAND}\" -E create_symlink \"\${_target}\" \"\${_link}\"
            RESULT_VARIABLE _result
            ERROR_VARIABLE _error
        )

        if(NOT _result EQUAL 0)
            message(FATAL_ERROR \"Failed to create systemd symlink \${_link} -> \${_target}: \${_error}\")
        endif()
    ")
endfunction()

if(NOT CMAKE_INSTALL_PREFIX STREQUAL "/usr")
    lemonade_install_systemd_symlink(
        "/usr/lib/systemd/user/lemond.service"
        "${CMAKE_INSTALL_PREFIX}/lib/systemd/user/lemond.service"
    )
endif()
```
And in the RPM workflow, after cpack -G RPM -V, please add a direct assertion that the package contains the expected user unit path and symlink target, e.g.:

```
rpm -qpl "$RPM_FILE" | grep -Fx "/usr/lib/systemd/user/lemond.service"
rpm -qpl --dump "$RPM_FILE" \
  | grep -E '^/usr/lib/systemd/user/lemond\.service ' \
  | grep -F "/opt/lib/systemd/user/lemond.service"
```

With that in place, I think the fix is much safer: packaging will fail immediately if the symlink is not created correctly, and CI will explicitly protect the Fedora user-service discovery behavior that #2118 is about.

lucifer-vali:
Thank for the example. I pretty much hijacked it entirely. 😅

fl0rianr (approved):
Now it looks good to me, had to use current main so CI runs clean

PR #2174, opened 2026-06-10
Remove sd-cpp cuda Windows support from compatibility tables

jeremyfowers:
## What

sd-cpp's CUDA backend is not supported on Windows. This drops Windows from the sd-cpp/cuda row in:

- `src/cpp/server/system_info.cpp` — recipe compatibility table (`{"sd-cpp", "cuda", {"linux"}, ...}`)
- `README.md` — Supported Configurations table (OS column now `Linux`)

Both now reflect Linux-only support.

## Note

`src/cpp/server/backends/sd_server.cpp` still contains a `_WIN32` branch that builds a `sd-...-windows-cuda-...-x64.zip` filename. With the compatibility table now Linux-only, that branch is effectively unreachable in normal flow. It was left untouched to keep this PR scoped to the compatibility tables — happy to remove it in a follow-up if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ Lemonade supports multiple inference engines for LLM, speech, TTS, and image gen
     <tr>
       <td><code>cuda</code></td>
       <td>NVIDIA GPUs (Turing or newer)**</td>
-      <td>Windows, Linux</td>
+      <td>Linux</td>
     </tr>
     <tr>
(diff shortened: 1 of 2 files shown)


kenvandine (approved):
This is good for now, we'll get Windows support back for the next release. Let's not miss the train!

PR #2176, opened 2026-06-10
feat: auto-detect MTP capability from GGUF metadata

### LINKED ISSUE #2166: Auto-detect MTP capability from GGUF tensor names for custom models
### Feature Description

Auto-detect MTP (Multi-Token Prediction) capability from GGUF metadata and tensor names for custom models.

Currently MTP support is only enabled for built-in models that have the `"mtp"` label hardcoded in `server_models.json`, or when users manually add `--label mtp` at pull time. Custom models (user.*) with MTP heads (e.g. `Qwen3.5-4B-MTP-GGUF`) should have this capability auto-detected at load time, matching how `vision` and `tool-calling` are already auto-detected from GGUF metadata.

The GGUF metadata contains the `{arch}.nextn_predict_layers` KV key and tensor names with the `"nextn"` prefix (e.g. `blk.{bid}.nextn.eh_proj`, `blk.{bid}.nextn.enorm`, `nextn.pre_projection`) - the canonical naming convention defined in llama.cpp's GGUF spec. The server should scan these and automatically apply the `"mtp"` label, so `llamacpp_server.cpp` passes `--spec-type draft-mtp` to llama-server.

### Use Case / Motivation

Users who pull or register custom GGUF models with MTP support have to manually discover and pass `--spec-type draft-mtp` to `llamacpp_server.cpp`. This is not obvious - most users won't know that it needs to be explicitly enabled. Auto-detection from the GGUF file itself makes it transparent, just like vision and tool-calling.

### Platform Relevance

All platforms

### Additional Context

Extend `read_gguf_capabilities()` in `gguf_capabilities.h` to scan the `{arch}.nextn_predict_layers` KV metadata key and tensor names containing `"nextn"` (e.g. `blk.{bid}.nextn.eh_proj`, `nextn.pre_projection`) - the canonical naming convention used by llama.cpp's GGUF spec. After reading the KV metadata pairs, iterate over the tensor info entries and check each tensor name for the `"nextn"` prefix. If either the KV key or tensor name indicates MTP, set `caps.mtp = true`, which `apply_gguf_capability_labels()` then maps to the `"mtp"` label.

This reuses the existing call site in `model_manager.cpp::populate_static_max_context_window()` which already runs for all downloaded LLM models at load time.

blackdeathdrow:
## What

Auto-detect MTP (Multi-Token Prediction) capability for custom GGUF models by reading the `{arch}.nextn_predict_layers` KV key from GGUF metadata.

Previously, MTP models required manual recipe configuration. This change enables automatic detection: when a GGUF file has `nextn_predict_layers > 0` (UINT32), Lemonade adds the `"mtp"` label to the model and passes `--spec-type draft-mtp` to llama-server.

### Detection logic

- **KV key scan** -- reads `{arch}.nextn_predict_layers` (defined in upstream llama.cpp as `LLM_KV_NEXTN_PREDICT_LAYERS` -> `"%s.nextn_predict_layers"`) and sets `mtp = true` only when the value is a UINT32 > 0
- **False-positive guards:**
  - Type must be `UINT32` (type 4) - non-integer types are ignored
  - Value must be `> 0` - a value of `0` means MTP is explicitly disabled

### Testing

- **Qwen3.5-4B-MTP-GGUF:UD-Q8_K_XL** - `nextn_predict_layers = 1` -> MTP detected
- **Qwen3.5-4B-GGUF:UD-Q8_K_XL** (no MTP) - no `nextn_predict_layers` key -> no MTP

A real Qwen3.5-4B-MTP model was loaded through the full Lemonade server pipeline and queried via the chat completions API:

1. Start `lemond` with the MTP model
2. Send `POST /v1/chat/completions` with a short prompt
3. Verify in server logs:
   - `Model uses MTP, adding draft decoding defaults` - label applied
   - `adding speculative implementation 'draft-mtp'` - llama-server confirms MTP mode
   - `draft acceptance = 1.00000 (N accepted / N generated)` - drafting works
4. Verify response includes `draft_n` and `draft_n_accepted` in timings

The same test with a non-MTP Qwen3.5-4B model produced no MTP logs and no `draft_n` fields, confirming no false detection.

## Files changed

- `src/cpp/include/lemon/gguf_capabilities.h` - MTP detection logic

Closes #2166


bitgamma:
This looks good, thanks! Before merging, what models did you test this with? Does this work with Gemma4? Also we need to make sure this doesn't trigger false positives with models that have MTP but MTP capabilities are not supported by llamacpp (if any)

blackdeathdrow:
I added the list of tested models to the description. This check does not trigger for Gemma 4.

bitgamma (approved):

iswaryaalex:
🚀 Would be great to see this working for https://huggingface.co/unsloth/gemma-4-12b-it-GGUF/tree/main/MTP !

PR #2178, opened 2026-06-10
feat(backends): add Moonshine streaming STT backend (#2115)

Geramy:
Original work done by @Theohox I am just doing some final touches.

* feat(backends): add Moonshine streaming STT backend

Adds a new 'moonshine' recipe for Moonshine speech-to-text using the moonshine_voice C++ streaming API. Integrates with existing OpenAI-compatible /v1/audio/transcriptions and WebSocket /realtime endpoints.

Includes:
- MoonshineServer backend (HTTP /inference, TCP streaming)
- IStreamingTranscriptionServer interface for generic streaming backends
- TcpJsonlClient for line-delimited JSON over TCP
- RealtimeSessionManager integration for WebSocket/audio forwarding
- ModelManager support for moonshine cache resolution and download
- CPU-only, cross-platform (Linux/Windows x86_64)
- End-to-end smoke test

The existing Whisper path is unchanged.

* Rework Moonshine backend to follow existing backend patterns

     - Switch from system Python to self-contained tarball install
       (lemonade-sdk/moonshine-server, following vLLM pattern)
     - Download models from HuggingFace
       (UsefulSensors/moonshine-streaming)
       instead of a separate download repo
     - Store models in standard HF cache, not a custom user folder
     - Auto-convert HF tokenizer.json -> tokenizer.bin on first load
     - Remove moonshine-specific download/resolution special cases from
       ModelManager; use standard HF download/cache paths
     - Add macOS support to system_info

---------

New PR to finish the great and amazing work of @Theohox


ramkrishna2910:
@claude review

Geramy:
**WebSocket upgrade on the main port**

`/realtime` and `/logs/stream` now accept WebSocket upgrades directly on the standard HTTP port (13305) — no separate port needed. The HTTP server peeks accepted connections and hands upgrade requests for those paths to the libwebsockets server, which adopts the socket on its service thread. The dedicated `websocket_port` listener is unchanged for backward compatibility.

Also accepts the `/v1`, `/v0`, `/api/v1`, `/api/v0` prefixed forms, so OpenAI Realtime SDK clients (`/v1/realtime`) connect as-is.

Verified: moonshine realtime streaming passes on both ports, all prefixes.

@jeremyfowers @ramkrishna2910

fl0rianr (changes requested):
Thanks, this is shaping up well. The previous review items around Windows sockets, weak_ptr lifetime, mutex protection, port probing, and the CodeQL temp-file suffix look addressed in the current head.

I would still hold approval on a few issues:

1. input_audio_buffer.clear disables streaming for the rest of the session. `clear_audio()` disconnects the streaming backend, but does not reconnect or forward a clear event. The next append therefore falls back to the buffered/VAD path instead of continuing Moonshine streaming. Please either forward `clear` to the TCP backend while keeping the stream open, or reconnect immediately after clearing, and add a “clear then append still streams” test.

2. TcpEventSender.send() is not fully thread-safe. The listener callbacks may come from moonshine_voice’s internal C++ thread, but the TCP sender calls `writer.write()` directly and only schedules `drain()` thread-safely. Please schedule the whole write+drain operation onto the asyncio event loop.

3. Startup has a readiness race: Lemonade waits for HTTP /health, but the TCP server is started in a daemon thread and may not yet be bound. `connect_streaming_backend()` only tries once, so a transient connection refusal silently disables streaming for the session. Please add TCP readiness polling or retry/backoff on connect.

4. Please add automated coverage for the new main-port websocket path (ws://127.0.0.1:{PORT}/v1/realtime). The current test still uses the legacy websocket_port, so the headline feature is only manually verified.

fl0rianr:
Thanks, this looks much better now! The previous functional blockers are adressed.

I’d still hold approval on a couple of regression risks:

1. There is a possible shutdown deadlock in `WebSocketServer::stop()`: it holds `connections_mutex_` while calling `session_manager_->close_session()`. Closing a streaming session can join the TCP read thread; if that read thread is concurrently forwarding a backend event, it calls `send_json()`, which tries to take the same `connections_mutex_`. Please snapshot/clear the connection state under the lock, release the lock, and only then close sessions / remove log subscribers. `handle_close()` already follows this safer pattern.

2. Realtime streaming appears to route via the most-recent loaded server rather than the requested session model. `create_session()` stores the requested `model`, but `connect_streaming_backend()` asks the router for a generic streaming address, and the router currently uses `get_most_recent_server()`. That can misroute multi-model scenarios: e.g. a Whisper realtime session could connect to Moonshine if Moonshine was loaded most recently, or a Moonshine session could fail to stream if a non-streaming model was loaded after it. Please route streaming by the requested session model, similar to the normal inference path, e.g. `get_streaming_transcription_address(model_name)` using `find_server_by_model_name(resolve_model_name(model_name))`.

Minor follow-up: please also make sure the new code has direct includes for what it uses (`<algorithm>` / `<cctype>` for `std::transform(..., ::tolower)`) instead of relying on transitive includes.

Once those are addressed and CI is green, I think this is merge ready.

Geramy:
> Thanks, this looks much better now! The previous functional blockers are adressed.
> 
> I’d still hold approval on a couple of regression risks:
> 
> 1. There is a possible shutdown deadlock in `WebSocketServer::stop()`: it holds `connections_mutex_` while calling `session_manager_->close_session()`. Closing a streaming session can join the TCP read thread; if that read thread is concurrently forwarding a backend event, it calls `send_json()`, which tries to take the same `connections_mutex_`. Please snapshot/clear the connection state under the lock, release the lock, and only then close sessions / remove log subscribers. `handle_close()` already follows this safer pattern.
> 2. Realtime streaming appears to route via the most-recent loaded server rather than the requested session model. `create_session()` stores the requested `model`, but `connect_streaming_backend()` asks the router for a generic streaming address, and the router currently uses `get_most_recent_server()`. That can misroute multi-model scenarios: e.g. a Whisper realtime session could connect to Moonshine if Moonshine was loaded most recently, or a Moonshine session could fail to stream if a non-streaming model was loaded after it. Please route streaming by the requested session model, similar to the normal inference path, e.g. `get_streaming_transcription_address(model_name)` using `find_server_by_model_name(resolve_model_name(model_name))`.
> 
> Minor follow-up: please also make sure the new code has direct includes for what it uses (`<algorithm>` / `<cctype>` for `std::transform(..., ::tolower)`) instead of relying on transitive includes.
> 
> Once those are addressed and CI is green, I think this is merge ready.

thanks for checking, I have fixed these issues!

fl0rianr (approved):
Thanks for the updates — this looks much better now.

One small non-blocking follow-up: the Moonshine support matrix still appears broader than the shipped assets. `system_info.cpp` advertises Moonshine CPU support for Windows/Linux/macOS with both `x86_64` and `arm64`, but the install params select Windows x64, Linux x64, and macOS arm64 assets. That could make unsupported combinations such as Intel macOS or Linux ARM appear available. Please consider tightening the system-info matrix or adding the missing assets.

PR #2179, opened 2026-06-10
Add DiffusionGemma

sawansri:
Still downloading...


nuts23:
model doesn't work yet, llama.cpp still working on official support for it.

bcmyguest:
doesn't vllm support it?

PR #2182, opened 2026-06-10
Fix AMD GPU architecture detection for CDNA datacenter parts

neoblizz:
`identify_rocm_arch_from_name()` mis-decoded the all-digit `gfx_target_version` reported by KFD on Linux, so AMD CDNA datacenter GPUs (MI100/MI210/MI300) were never identified even when a matching backend binary exists. The old decoder assumed a 2-digit major and decimal step, mangling every CDNA ISA (e.g. `90010` -> `gfx9010` instead of `gfx90a`). Replaced with the universal AMD encoding (`major = v/10000, minor = (v/100)%100, step = v%100`, minor/step in hex), correct for all CDNA parts and unchanged for every RDNA part already working.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -10,6 +10,7 @@
 #include <filesystem>
 #include <fstream>
 #include <sstream>
+#include <cstdio>
 #include <cstdlib>
 #include <iostream>
 #include <regex>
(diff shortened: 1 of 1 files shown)


src/cpp/server/system_info.cpp
```diff
@@ -111,6 +112,10 @@ const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
     // RDNA4 family (gfx120X)
     {"gfx1200", "gfx120X"},
     {"gfx1201", "gfx120X"},
+
+    // CDNA datacenter parts (gfx908, gfx90a, gfx942) are intentionally unmapped:
+    // the decoder then returns the exact ISA token, which matches per-arch nightly
+    // asset names. Do NOT map to "" — that means "no binary" and hides a valid arch.
```
superm1: pointless text

superm1 (commented):
I think that https://github.com/lemonade-sdk/lemonade/pull/2093 is going to likely merge first, I like that design of dropping ROCM_ARCH_MAPPING from that PR (and you're adding to it)

src/cpp/server/system_info.cpp
```diff
@@ -1828,18 +1839,26 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
...
+    // KFD reports the ISA as an all-digit gfx_target_version. Decode universally:
+    //   major = v/10000, minor = (v/100)%100, step = v%100  (minor/step are hex)
+    // e.g. 110501 -> gfx1151, 90010 -> gfx90a, 90402 -> gfx942, 90008 -> gfx908.
+    if (!device_lower.empty() &&
+        std::all_of(device_lower.begin(), device_lower.end(), ::isdigit)) {
+        int v = 0;
+        try {
+            v = std::stoi(device_lower);
+        } catch (...) {
+            v = 0;
```
superm1: this should be more than silently ignored IMO - it **should** raise an exception to be dug into.  because it means something really bad happened with the interface the kernel offers.

test/test_rocm_arch_mapping.py
superm1, line 1: useful for development, but not to include, I would drop this file.

neoblizz: Based on this; https://github.com/neoblizz/lemonade/blob/main/test/test_cuda_arch_mapping.py

superm1 (commented):
please rebase, we dropped ROCM_ARCH_MAPPING

neoblizz:
Rebased, cleaned up the comments + removed the test file. Let me know if force-push is not acceptable, I can create a separate branch instead. @superm1

superm1 (approved):

PR #2183, opened 2026-06-11
feat: opt-in dynamic VRAM management (auto-evict idle models)

clemperorpenguin:
This PR implements opt-in dynamic VRAM memory management to automatically unload or downsize idle models when the system faces VRAM pressure or timeouts, resolving issue 1365.

It introduces two new background threads, EvictionEngine and GlobalVramMonitor, which poll system memory usage and gracefully evict or downsize the least recently used models. To support this safely under concurrent load, the backend state handling was upgraded to a robust state machine that pauses incoming requests if a model is currently downsizing or being restored.

Downsizing, which clears KV caches to save memory, is initially implemented for the llama.cpp backend, while other backends safely ignore it. Configuration options for auto-eviction thresholds and idle timeouts have been added globally and per-recipe. Finally, Python integration tests were added alongside an internal mock endpoint to verify the VRAM pressure logic and concurrency safety.

#1365


fl0rianr (changes requested):
Thanks for the update here and the nice work.
Sorry I only saw this now you made it ready for review.

But I don't think this is merge-ready yet; beyond the current mergeability state (easy accept both fix), I see a real lifetime race in the new downsize path.

EvictionEngine::evaluate_servers() marks a server as DOWNSIZING, later reacquires `router_->load_mutex_`, obtains a raw WrappedServer*, releases the router lock, and then calls s->downsize() / s->set_state(DOWNSIZED). While that lock is released, another path can still unload or evict the same server. Router::evict_server() only waits for `active_request_count_ == 0`; a server in DOWNSIZING is not counted as busy/owned, so it can call `unload()` and erase the owning unique_ptr from loaded_servers_ while the eviction engine still holds the raw pointer.

That means the engine can end up with a dangling WrappedServer*, or call downsize() / set_state() on an object whose backend has already been unloaded. The inference-vs-eviction state machine looks carefully handled, but maintenance work (DOWNSIZING) is not protected against object destruction.

I think the fix should make downsizing an owned/guarded operation, e.g. a try_begin_downsize() / finish_downsize(success) transition that increments a maintenance/busy counter or otherwise makes evict_server() wait/abort while downsizing is in progress. Also, downsize() should probably report success/failure so the state is only changed to DOWNSIZED when the backend operation actually succeeded.

fl0rianr (approved):
Thanks for the fix. I'm approving now.

PR #2184, opened 2026-06-11
fix: use Win32 API fallback for create_directories on symlinked Huggi…

### LINKED ISSUE #2181: Windows, cache not available for download
### Platform

Windows

### Lemonade Version

10.7.0

### GPU / APU Model

_No response_

### Component

llama.cpp

### Bug Description

Form discord to keep track:

Windows version suddenly doesn't symlinks. 

```
DOWNLOAD MANAGER
Qwen3.5-9B-GGUF
0 active
0 completed
Error: create_directories: The path cannot be traversed because it contains an
untrusted mount point.:
Status:
Current File:
x
error
```

### Steps to Reproduce

c:\Users\USER\.cache\huggingface is a symlink

Expected: Download works
Actual: Path issue and error

### Log Output

```shell

```

### Additional Context

https://discord.com/channels/1392562559122407535/1394403929973985413/1514311992242868255

superm1:
…ngFace cache

On Windows, std::filesystem::create_directories fails with "untrusted mount point" error when the path contains symlinks crossing volume boundaries (e.g. ~/.cache/huggingface symlinked to another drive).

Add safe_create_directories() that falls back to SHCreateDirectoryExW when the standard library fails, following the same pattern as the existing safe_exists() and safe_is_directory() helpers.

Fixes #2181

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -719,6 +719,7 @@ if(WIN32)
         wbemuuid
         ole32
         oleaut32
+        shell32
     )
 elseif(APPLE)
     target_link_libraries(lemonade-server-core PUBLIC
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -53,24 +53,41 @@ static bool safe_is_directory(const fs::path& p) {
 static constexpr auto safe_dir_options = fs::directory_options::skip_permission_denied;
 // MSVC's create_directories also fails on symlinks crossing volume boundaries
 // ("untrusted mount point"). SHCreateDirectoryExW does not have this restriction.
-static bool safe_create_directories(const fs::path& p) {
-    if (p.empty()) return true;
-    if (safe_exists(p)) return true;
+// Throws on failure to preserve the fail-fast semantics of fs::create_directories.
+static void ensure_create_directories(const fs::path& p) {
+    if (p.empty()) return;
+    if (safe_is_directory(p)) return;
+    if (safe_exists(p)) {
+        throw std::runtime_error("Cannot create directory; a non-directory already exists at '" +
+                                 path_to_utf8(p) + "'");
+    }
     std::error_code ec;
     fs::create_directories(p, ec);
-    if (!ec) return true;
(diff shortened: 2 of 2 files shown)


fl0rianr (changes requested):
Thanks for tackling this so soon —  using a Win32 fallback for the MSVC reparse-point / "untrusted mount point" case makes sense.

I think this needs one more pass before merging though:

1. safe_create_directories() currently returns success for any existing path:
   `if (safe_exists(p)) return true;`

   That changes create_directories semantics. If `p` exists but is a regular file, this should still be treated as an error, not success. This should probably be:

   ```cpp
   if (safe_is_directory(p)) return true;
   if (safe_exists(p)) return false; // or throw with context

2. All call sites ignore the returned bool. Previously, the throwing fs::create_directories(...) calls failed fast; after this PR, directory creation failures can be silently swallowed and then show up later as less obvious download/file-open/cache errors.
I think this helper should either be ensure_create_directories(...) and throw with a useful message, or every call site should check the return value and fail explicitly.

3. Since SHCreateDirectoryExW introduces a Shell32 dependency, could we explicitly link shell32 in the Windows CMake link list? CI is green, so this is not necessarily breaking today, but making the dependency explicit would be cleaner.

With those addressed, this looks like the right direction.

fl0rianr:
Thanks for the quick update — this addresses the concerns from my previous review.

One small non-blocking nit: if `SHCreateDirectoryExW` fails, including the Win32 return code in the exception message could make future debugging easier. But I don’t think that should block this PR.

LGTM from the code side but how do we do a windows test to secure it's fixed - it was not my issue originally.

superm1:
> One small non-blocking nit: if SHCreateDirectoryExW fails, including the Win32 return code in the exception message could make future debugging easier. But I don’t think that should block this PR.

I'll fix it nonetheless.

> LGTM from the code side but how do we do a windows test to secure it's fixed - it was not my issue originally.

Merge it and ask the original reporter to try a build from main?

fl0rianr (approved):
LGTM and I agree

PR #2185, opened 2026-06-11
fix: remove load options environment variable support from lemonade CLI

### LINKED ISSUE #2158: Decide whether to drop env variable support for `lemonade` CLI
> @superm1 to confirm: you intentionally left the env var support in the CLI?

Correct.  I figured we need to talk about this separately, since it's not migration path and my PR was for removing migration path.

_Originally posted by @superm1 in https://github.com/lemonade-sdk/lemonade/issues/2152#issuecomment-4661151279_

superm1:
Drop LEMONADE_CTX_SIZE, LEMONADE_MERGE_ARGS, LEMONADE_LLAMACPP*, LEMONADE_SDCPP*, LEMONADE_WHISPERCPP*, LEMONADE_VLLM*, and LEMONADE_FLM_ARGS from recipe options.

Fixes #2158

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -2020,7 +2020,7 @@ jobs:
           HF_HOME: ${{ env.HF_HOME }}
         run: |
           set -e
-          $VENV_PYTHON test/server_cli2.py --cli-binary "$CLI_BINARY"
+          $VENV_PYTHON test/server_cli2.py --cli-binary "$CLI_BINARY" --api-key "$LEMONADE_API_KEY"
 
       - name: Capture and upload server logs
         if: always()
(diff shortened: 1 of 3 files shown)


bitgamma:
```
Drop LEMONADE_HOST, LEMONADE_PORT, LEMONADE_API_KEY, and LEMONADE_ADMIN_API_KEY from the global connection options.
```
I'd rather keep these: if you connect to the same server over and over it gets old to always specify all connection parameters. It could be handled with aliases but having envs for this is quite idomatic.

The others can be dropped as far as I am concerned

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -2014,7 +2014,9 @@ jobs:
           HF_HOME: ${{ env.HF_HOME }}
         run: |
           set -e
-          $VENV_PYTHON test/server_cli2.py --cli-binary "$CLI_BINARY"
+          # CLI commands pass --api-key explicitly since the CLI no longer
```
bitgamma: outdated comment

bitgamma (approved):
lgtm, apart from the comment since we are keeping connection relared envs

Issue #2002, opened 2026-05-23
Please add ROCm support for the Krackan APU family (Radeon 840M / 860M / gfx1152).


samtux:
### Feature Description

Please add the **Krackan APU** family (gfx1152 / Radeon 840M and 860M) to the list of supported AMD GPUs for ROCm backends.

Currently, running `lemonade backends install llamacpp:rocm` on this hardware gives:

```
Error: Cannot install llamacpp:rocm on this system: Unsupported GPU: gfx1152
```

gfx1152 is already present in `backend_versions.json` under TheRock's supported architectures, so it looks like only the source-side device detection needs updating.

### Use Case / Motivation

The AMD Ryzen AI 7 350 with Radeon 860M is a current-generation APU found in many new laptops and Mini PCs. Adding gfx1152 support would let users of this hardware run LLMs with full ROCm acceleration instead of falling back to CPU or Vulkan.

### Platform Relevance

All platforms

### Additional Context

- **CPU:** AMD Ryzen AI 7 350
- **GPU:** Radeon 860M (gfx1152)
- **ROCm:** 7.13
- **OS:** Arch Linux / CachyOS

samtux:
https://github.com/lemonade-sdk/lemonade/issues/1991

ckuethe:
It's probably worth following along with https://github.com/ROCm/TheRock/issues/2310 - it looks like ROCm isn't officially stable on gfx1152 yet.

It is technically possible to put the rocm back end into the cpu directory so that you can use it by selecting the cpu back end (yes, that's horrible), and it might work for some workloads, but it's not been very stable for me... I could get the Bonsai 1B model to run and produce good output, but Bonsai 8B and Gemma4 produced garbage.

You could also patch server.cpp to add gfx1152 following the style of gfx1151, but it's not going to work well.

X-Ryl669:
This is what I did in the previous version of lemonade (10.5 IIRC) since the release note mentioned it supported gfx1152, but it clearly didn't. So I've patched system-info.cpp to add the missing enumerations cases and it loaded but it did not work. In lemonade 10.6, no patching was required, it loaded directly, but it still didn't work. That's strange, since I've built a TheRock version in last december by patching the Triton AO code while in the build and this version worked well will ComfyUI for Stable Diffusion and another image diffusion model. Maybe there's something different with LLM than in diffusion models. I hope they'll sort this out, it's more than 18 months that the processor is released, it's a shame it's not working yet.

[...]

Issue #1878, opened 2026-05-12
Support gfx1152 for stable diffusion on linux


ckuethe:
### Feature Description

gfx1152 seems to be reasonably functional, and shouldn't be refused by lemond. It could still be marked as `unsupported` (as in not "we're sure this works") but if you `--force` it then you're willing to accept some instability or bogus results.

### Use Case / Motivation

I have a gfx1152 laptop as well as a gfx1151 workstation. I'd like to use the same software on both of these machines

### Platform Relevance

All platforms

### Additional Context

With some lies and manipulation, lemond can be tricked into running the the sd-cpp:rocm back end on Ubuntu 24 LTS

* download both `therock-dist-linux-gfx1151-7.12.0a20260311.tar.gz` and `therock-dist-linux-gfx1152-7.12.0a20260311.tar.gz` from https://rocm.nightlies.amd.com/tarball/
* extract first the gfx1151 then gfx1152 tarball into /opt/rocm/rocm-7.12
* `echo /opt/rocm/rocm-7.12/lib | sudo tee -a /etc/ld.so.conf.d/rocm.conf ; sudo ldconfig`
* `echo HSA_OVERRIDE_GFX_VERSION="11.5.1" | sudo tee -a /etc/lemonade/conf.d/gfx.conf`
* `lemonade backends install sd-cpp:rocm --force`
* `sudo mv -f /var/lib/lemonade/.cache/lemonade/bin/sd-cpp/cpu{,-old}
* `sudo ln -s /var/lib/lemonade/.cache/lemonade/bin/sd-cpp/{rocm-preview,cpu}`

This image was generated with the rocm backend in 160s at 100% GPU utilization according to `amdgpu_top`, down from 760s with the CPU backend.
<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/2faa3aad-6464-4bad-91fd-5d98f50f8389" />

quandvdo:
Hello @ckuethe , Nice work on the instructions, I have been struggling to make sd-cpp:rocm works with lemonade. Now it's running much much faster compare to cpu .

sofiageo:
Not sure if there is a need to do something here, the next stable diffusion release will include the gfx1152 target. But it will be good to test if it makes any difference.

ckuethe:
[...]

PR #2186, opened 2026-06-11
Enable ROCm on gfx1152

ckuethe:
Tested with `Device 0: AMD Radeon 860M Graphics, gfx1152 (0x1152), VMM: no, Wave Size: 32, VRAM: 81920 MiB` 

llama.cpp Bonsai-1.8B ROCM - 88.7tps
llama.cpp Qwen3.6-35B-A3B-MTP ROCM - 17.1tps
llama.cpp Qwen3-Coder-Next-80B ROCM - 13.5tps (using >48GB memory)
sdcpp flux-2-klein-4b "Red Panda" @ 1024x1024 - 141.5s

All outputs were reasonable. no getting stuck in a loop, generating garbage, etc. No additional VAE flags needed for stable diffusion, unlike what was necessary for the vulkan back end.

If only there was some way to bulk benchmark a whole bunch of models... 😁 

Patch created without the use of AI, nor looking at @samtux most [recent patch](https://github.com/lemonade-sdk/lemonade/issues/2002#issuecomment-4676366577) but we previously came up with very similar diffs, so I'm willing to share credit.

also solves #1878

### DIFF
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -480,9 +480,9 @@ void LlamaCppServer::load(const std::string& model_name,
         }
 
         std::string arch = lemon::SystemInfo::get_rocm_arch();
-        if (arch == "gfx1151") {
+        if ((arch == "gfx1151") || (arch == "gfx1152")){
             env_vars.push_back({"OCL_SET_SVM_SIZE", "262144"});
-            LOG(DEBUG, "LlamaCpp") << "Setting OCL_SET_SVM_SIZE=262144 for gfx1151 (enables loading larger models)" << std::endl;
+            LOG(DEBUG, "LlamaCpp") << "Setting OCL_SET_SVM_SIZE=262144 for gfx1151/gfx1152 (enables loading larger models)" << std::endl;
         }
     } else if (is_llamacpp_cuda_backend(llamacpp_backend)) {
         // CUDA Windows builds bundle cudart64_*.dll, cublas64_*.dll, etc. next to
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -84,6 +84,36 @@ const std::set<std::string> CUDA_SUPPORTED_ARCHS = {
     "sm_120",  // Blackwell    (RTX 50)
 };
 
+// ROCm architecture mapping - maps specific gfx architectures to their family (download target).
+// Empty string means "no ROCm binary for this ISA" — skip for get_rocm_arch / install filenames.
+const std::map<std::string, std::string> ROCM_ARCH_MAPPING = {
+    // RDNA2 family (gfx103X)
+    {"gfx1030", "gfx103X"},
+    {"gfx1031", "gfx103X"},
+    {"gfx1032", "gfx103X"},
+    {"gfx1034", "gfx103X"},
+    // Note: gfx1033, gfx1035, gfx1036 are NOT included (not confirmed as supported)
+    // map to "" so get_rocm_arch skips them
+    {"gfx1033", ""},
+    {"gfx1035", ""},
+    {"gfx1036", ""},
+
+    // RDNA3 family (gfx110X)
+    {"gfx1100", "gfx110X"},
+    {"gfx1101", "gfx110X"},
+    {"gfx1102", "gfx110X"},
+    {"gfx1103", "gfx110X"},
+
+    // RDNA3.5 iGPUs - explicit binary names (no family mapping)
+    {"gfx1150", "gfx1150"},  // Maps to exact binary name
+    {"gfx1151", "gfx1151"},  // Maps to exact binary name
+    {"gfx1152", "gfx1152"},  // Maps to exact binary name
+
+    // RDNA4 family (gfx120X)
+    {"gfx1200", "gfx120X"},
+    {"gfx1201", "gfx120X"},
+};
+
 #ifdef __linux__
 namespace {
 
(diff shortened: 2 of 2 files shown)


ckuethe:
testing with
```
lemonade bench --json --output lmbench.json \
    --backend vulkan --backend rocm \
    --ctx-size 4096 --ctx-size 8192 --ctx-size 32768 \
    Bonsai-1.7B-gguf Bonsai-4B-gguf Bonsai-8B-gguf Gemma-3-4b-it-GGUF Gemma-4-12B-it-GGUF \
    Gemma-4-26B-A4B-it-GGUF Gemma-4-E2B-it-GGUF Gemma-4-E4B-it-GGUF LFM2-1.2B-GGUF \
    LFM2.5-1.2B-Instruct-GGUF Llama-3.2-1B-Instruct-GGUF Nemotron-3-Nano-30B-A3B-GGUF \
    Phi-4-mini-instruct-GGUF Qwen3-1.7B-GGUF Qwen3-4B-GGUF Qwen3-Next-80B-A3B-Instruct-GGUF \
    Qwen3.5-2B-GGUF Qwen3.5-4B-GGUF Qwen3.5-4B-MTP-GGUF Qwen3.6-35B-A3B-GGUF \
    Qwen3.6-35B-A3B-MTP-GGUF SmolLM3-3B-GGUF
```

We'll see how it looks in a few hours, or if my laptop catches fire.

src/cpp/server/backends/llamacpp_server.cpp
```diff
@@ -480,9 +480,9 @@ void LlamaCppServer::load(const std::string& model_name,
         }
 
         std::string arch = lemon::SystemInfo::get_rocm_arch();
-        if (arch == "gfx1151") {
+        if ((arch == "gfx1151") || (arch == "gfx1152")){
```
fl0rianr, line 483: tiny nit: ((  ) || (  )) can be ( || ) instead.

fl0rianr (commented):
Nice and focused!

But I'm not sure if we should merge this before #2093 where integrating this is easy and we avoid merge conflicts and may include all involved contributers there.

A blocker is that vLLM is also supportend while there is no artifact for that. @ramkrishna2910: Do you think we can bring in gfx1152 quickly or should the PR which finally adds gfx1152 definitely exclude it?

src/cpp/server/backends/llamacpp_server.cpp
ckuethe: I know, I just like having each of the tested conditions in its own ().. it's a bit easier for me to read.

ckuethe:
I can remove the gfx1152/vllm bit if you want. so far I haven't found any compelling advantage to vllm on gfx1151 or gfx1100, maybe that will change in the future - on the other hand, rocm seems to be stabilizing and this allows for wider adoption

ckuethe:
660 tests ran overnight, no failures. rocm and llama.cpp have seriously stabilized.

[lemonade_bench_gfx1152.csv](https://github.com/user-attachments/files/28849232/lemonade_bench_gfx1152.csv)

src/cpp/server/backends/llamacpp_server.cpp
superm1, line 483: How big of a GFX1152 config are you seeing, large enough to justify the special case?

superm1:
> But I'm not sure if we should merge this before https://github.com/lemonade-sdk/lemonade/pull/2093 where integrating this is easy and we avoid merge conflicts and may include all involved contributers there.

I would rather if we land #2093 first and let these rebase we've been working with the OP on that one for a while now and it's ready to go.

fl0rianr:
> I would rather if we land https://github.com/lemonade-sdk/lemonade/pull/2093 first and let these rebase we've been working with the OP on that one for a while now and it's ready to go.

I agree ... and meant something similar (although not exactly and not properly formulated). Re-basing this one should be easy.

src/cpp/server/backends/llamacpp_server.cpp
ckuethe: It's not fast... in fact it's very slow, but I can load and use the 72GB Q4_K_M quant of https://huggingface.co/bartowski/Meta-Llama-3-120B-Instruct-GGUF

Nowhere close to a daily driver, just loading it to see if it would work. Qwen3.6 or maybe Nemotron are much more useful.

superm1:
OK, please rebase now.

ckuethe:
rebased, and removed gfx1152 from the vllm section since there isn't a vllm build for gfx1152 yet.

src/cpp/server/backends/llamacpp_server.cpp
ckuethe: Also, that 120B model OOMs my machine in with vulkan, but slowly works with ROCm.

fl0rianr:
Almost ready from my side. The gfx1152 recipe/device-name changes look good now, and the unrelated Moonshine CI failure is not a blocker for me.

One blocker before approval: this reintroduces ROCM_ARCH_MAPPING, which #2093 intentionally removed in favor of backend_versions.json as the mapping source. It also looks unused now because identify_rocm_arch_from_name() returns raw gfx codes directly. Could you drop ROCM_ARCH_MAPPING from this PR and keep only the gfx1152 detection + recipe/family-name changes?

ckuethe:
Fixed the rebase error.

src/cpp/server/system_info.cpp
```diff
@@ -1835,6 +1836,16 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
...
+    // KRK Point iGPUs (gfx1152 architecture)
+    // Radeon 840M / 860M Graphics
+    if (device_lower.find("840m") != std::string::npos ||
+        device_lower.find("860m") != std::string::npos) {
+        return "gfx1152";
+    }
+
+    // RDNA4 GPUs (gfx120X architecture)
+    // AMD Radeon AI PRO R9700, AMD Radeon RX 9070 XT, AMD Radeon RX 9070 GRE,
+    // AMD Radeon RX 9070, AMD Radeon RX 9060 XT
```
superm1: we stopped adding obvious comments, can you drop them?

fl0rianr (approved):
lgtm now

PR #2191, opened 2026-06-11
fix(ui): connect log stream on main HTTP port instead of dedicated WebSocket port

### LINKED ISSUE #2190: Log stream fails through reverse proxy - always connects to dedicated WebSocket port
### Platform

Linux/Ubuntu

### Lemonade Version

10.7.0

### GPU / APU Model

AMD RYZEN AI MAX+ 395

### Component

Desktop App / UI

### Bug Description

When accessing the Lemonade web UI through a reverse proxy (e.g. `https://example.com` -> `http://192.168.x.x:13305`), the log viewer shows "Failed to connect to Lemonade Server logs" and never receives any log entries. The same setup works fine when accessing the server directly inside the local network.

The root cause is in `src/app/src/renderer/utils/logWebSocketClient.ts`: `connectLogStream()` always fetches `websocket_port` from `/health` and connects to the dedicated WebSocket port. Behind a reverse proxy, only the main HTTP port (443) is exposed - the dedicated port is unreachable.

Meanwhile, the main HTTP port already accepts WebSocket upgrades for `/logs/stream` (via `UpgradableFrontServer` in `upgradable_http_server.h`), and the transcription WebSocket client (`websocketClient.ts`) already uses the correct pattern: try the main port first, fall back to the dedicated port. The log client simply never tries the main port.

### Steps to Reproduce

1. Start Lemonade server on a machine in the local network:
   lemond 0.0.0.0 --port 13305
2. Set up a reverse proxy (nginx/caddy) exposing `https://example.com` -> `http://192.168.x.x:13305` (only port 443 forwarded)
3. Open the web UI at `https://example.com`
4. Open the Logs panel (Ctrl+Shift+L or View -> Logs)
5. See the error message: "Failed to connect to Lemonade Server logs. Make sure the server is running on ..."

Expected: Logs load and stream successfully through the reverse proxy, just like they do when accessing the server directly via `http://192.168.x.x:13305`.

Actual: The WebSocket connection fails because it tries to connect to `wss://example.com:<websocket_port>/logs/stream` where the dedicated WebSocket port (e.g. 9000 or 13307) is not exposed through the proxy.

### Log Output

```shell

```

### Additional Context

The transcription WebSocket client (`src/app/src/renderer/utils/websocketClient.ts`) already handles this correctly - it tries the main HTTP port first with `buildWebSocketUrl('/v1/realtime', undefined, query)`, and only falls back to the dedicated `websocket_port` from `/health`. The log client should follow the same pattern.

blackdeathdrow:
## Description

Fixes the log viewer failing to connect when Lemonade is accessed through a reverse proxy.

### Root Cause

`connectLogStream()` in `src/app/src/renderer/utils/logWebSocketClient.ts` always fetched `websocket_port` from `/health` and connected to the dedicated WebSocket port. Behind a reverse proxy that only exposes the main HTTP port (e.g. 443), this port is unreachable.

### Fix

Connect to `/logs/stream` on the main HTTP port by removing the `wsPort` argument from `buildWebSocketUrl()`. The main port already accepts WebSocket upgrades for `/logs/stream` via `UpgradableFrontServer` (`upgradable_http_server.h:112`), so no server-side changes are needed.

### Testing

Tested locally using lemond on 0.0.0.0:13305 with Mikrotik Reverse Proxy - log stream connects and receives entries successfully.

### Related Issue

Closes #2190

### DIFF
--- a/src/app/src/renderer/utils/logWebSocketClient.ts
+++ b/src/app/src/renderer/utils/logWebSocketClient.ts
@@ -1,4 +1,4 @@
-import { buildWebSocketUrl, serverFetch } from './serverConfig';
+import { buildWebSocketUrl } from './serverConfig';
 
 export interface LogEntry {
   seq: number;
@@ -24,18 +24,7 @@ export async function connectLogStream(
   afterSeq: number | null,
   callbacks: LogStreamCallbacks,
 ): Promise<LogStreamHandle> {
-  // Get websocket port from health endpoint
-  const healthResponse = await serverFetch('/health');
-  if (!healthResponse.ok) {
-    throw new Error(`Failed to fetch health: ${healthResponse.status}`);
-  }
-  const health = await healthResponse.json();
-  const wsPort = health.websocket_port;
-  if (typeof wsPort !== 'number') {
-    throw new Error('Server did not advertise a websocket port');
-  }
-
-  const wsUrl = buildWebSocketUrl('/logs/stream', wsPort);
+  const wsUrl = buildWebSocketUrl('/logs/stream');
   const socket = new WebSocket(wsUrl);
 
   socket.addEventListener('open', () => {


Geramy:
Thank you will be taking a look once I fix a few minor things today.

Geramy (approved):
I am good with this, it makes me feel a lot better.

PR #2198, opened 2026-06-11
Add support for launching pi.dev

superm1:


docs/integrations/pi.md
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+}
+```
+
+**`~/.pi/agent/settings.json`** (default provider/model):
+```json
+{
+  "defaultProvider": "Lemonade",
+  "defaultModel": "Qwen3.5-35B-A3B-GGUF"
```
sawansri, line 96: It makes more sense to me to leave default model and provider alone and avoid modifying the user's global settings. We need to do this for models.json to register the provider and all models available but we should be able to pass in the user specified launch model to pi via the `--provider` and `--model` flags versus setting a global default provider/model.

This would match how we launch opencode as well.

docs/integrations/pi.md
```diff
@@ -0,0 +1,116 @@
...
+## Launch Pi with Lemonade
+
+Use:
+
+```bash
+lemonade launch pi [options]
+```
+
+Lemonade automatically configures Pi to use your local server by writing:
+- `~/.pi/agent/models.json` — Registers the Lemonade provider with your local server's base URL and selected model
```
sawansri, line 26: Pi supports the PI_CODING_AGENT_DIR env var for specifying alternate config directories, we should check this env var and write to the correct conf location if necessary.

sawansri (commented):
Looking forward to adding pi! Just left a few comments.

docs/integrations/pi.md
superm1: How about if we meet in the middle - what if they don't have a configuration yet we populate the global one?  I'm thinking then that means that `lemonade launch pi` sets up the default configuration and next time they run `pi` it "just works"

sawansri: Yeah sounds good. We can leave global alone if they already have a default model and provider setup.

superm1: That's what I've pushed to the branch, LMK what you think.

docs/integrations/pi.md
superm1: fixed

sawansri (approved):
LGTM!

PR #2206, opened 2026-06-12
Fix: recipe/backend status resolution for duplicated backend definition

fl0rianr:
## Summary

Fix recipe/backend status resolution for duplicated backend definitions.

Moonshine defines `moonshine/cpu` once per supported OS. On Windows and Linux, later unsupported entries for other OSes could overwrite the valid backend status for the current platform, causing Moonshine to be reported as unsupported.

This change adds a small priority-based setter so current-platform results win over fallback unsupported entries from other OS definitions.

## Testing

- All supported backends on my Ubuntu are working.
- Existing CI covers the Moonshine system info path.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -1221,9 +1221,9 @@ json SystemInfo::build_recipes_info(const json& devices) {
     // entries. Keep the most useful status for the current system:
     // non-unsupported states beat current-OS unsupported states, which beat
     // unsupported fallback entries from other OS definitions.
-    constexpr int kOtherOsUnsupportedPriority = 0;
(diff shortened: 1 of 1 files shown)


jeremyfowers:
Let me have a look before merging please

jeremyfowers (approved):
Approving to get a fix into CI ASAP! Thanks @fl0rianr for jumping on this.

I have two suggestions, it would be nice if you took one of them:
1. Add a comment explaining why there is a priority system and how it works. I personally had a little trouble understanding so a comment would have helped.
2. Would a simpler alternative implementation be to just completely ignore the system-info entries for operating systems that don't match the current operating system?

PR #2208, opened 2026-06-12
Add instructions about dkms conflict for users that run into that

### LINKED ISSUE #1852: Add more info about the kernel issue and workarounds to the Linux RAI Kernel Page

https://lemonade-server.ai/gfx1151_linux.html

### What's Missing or Wrong

Even if you have a kernel that does support the fix, if you have dkms installed, Lemonade won't install the rocm backend and it fails without any info.

It's kind of rare to get into this issue, so for now, documentation would be sufficient.

### Suggested Improvement

Update the kernel bug docs page to point to the ROCm installation for Ryzen AI in case users have accidentally/incorrectly installed dkms, which will break GPU usage with Ryzen devices.
https://rocm.docs.amd.com/projects/radeon-ryzen/en/latest/docs/install/installryz/native_linux/install-ryzen.html#set-up-rocm-usecase

Users likely shouldn't need that level of knowledge but because Lemonade is so helpful, people find this page before ROCm links.

vgodsoe:
Closes #1852

### DIFF
--- a/docs/gfx1151_linux.html
+++ b/docs/gfx1151_linux.html
@@ -127,6 +127,51 @@
       </div>
 
       <div class="install-section">
+        <h2>Still seeing ROCm failures after updating your kernel?</h2>
+        <p>
+          If your kernel is up to date but Lemonade still refuses to install or run the ROCm backend,
+          you may have the AMDGPU <strong>DKMS</strong>
+          module installed. On Ryzen AI / Strix Halo systems, the DKMS module conflicts with the
+          in-tree <code>amdgpu</code> driver and breaks GPU access for ROCm workloads.
+        </p>
+
+        <div class="option-box option-secondary">
+          <div class="option-header">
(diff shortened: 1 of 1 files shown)


docs/gfx1151_linux.html
```diff
@@ -126,6 +126,51 @@ <h3>Verify</h3>
...
+                ROCm on Radeon / Ryzen — Set up ROCm use case
+              </a>
+            </li>
+          </ul>
+          <p>
+            In short, when running <code>amdgpu-install</code>, pass a use case that does
+            <strong>not</strong> include <code>dkms</code>, for example:
+          </p>
+          <div class="code-block">
+            <code>sudo amdgpu-install -y --usecase=wsl,rocm --no-dkms</code>
```
superm1: why would someone use `wsl` use case in Linux?

superm1 (approved):
drop WSL reference and then LGTM

Issue #2063, opened 2026-06-01
Claude fails to load models without a large enough context size set


DavidPerryamd:
### Platform

Linux/Debian

### Lemonade Version

10.5.1

### GPU / APU Model

 AMD RYZEN AI MAX+ 395

### Component

llama.cpp

### Bug Description

when running ```lemonade launch claude``` without any additional parameters, claude will return a "failed to load model" error once a prompt is attempted without any clear reasoning as to why. this may leave users confused as to what the failure point is. After some poking around I discovered this is because claude requires a rather large context size to function. 

My suggestion would be to add a warning about context size if not explicitly set
(e.g  *warning* no context size set, agent may fail to load model.)

Additionally, maybe you could add a segment to select ctx size during the setup flow of ```lemonade launch``` with some pre-determined sizes

### Steps to Reproduce

1. run ```lemonade launch claude``` with your model of choice
2. Type a prompt inside of claude
3. Observe failure 

### Log Output

```shell

```

anditherobot:
While investigating #2063 (Claude failing without a large enough context size), I found that Windows users could not even reach that point the agent failed to launch entirely with Error 193. I filed a separate PR #2213 
which fixes the launcher so Windows users can now get far enough to hit #2063. (this issue)

PR #2213, opened 2026-06-12
fix(cli): lemonade launch claude fails to launch on Windows, throws Error 193

### LINKED ISSUE #2211: Bug: lemonade launch claude fails on Windows with Error 193 (not a valid Win32 application)
### Platform

Windows

### Lemonade Version

10.7.0

### GPU / APU Model

_No response_

### Component

Other / Not sure

### Bug Description

On Windows, `lemonade launch claude` fails instantly and the agent never starts, failing with Error 193 (`%1 is not a valid Win32 application`). npm installs several `claude` launchers on Windows: an extensionless `claude` (a Unix/bash script), plus `claude.cmd` and `claude.exe`. Lemonade picks the plain `claude` script first and tries to run it as a Windows program, which fails because it isn't a real Windows executable.

### Steps to Reproduce

1. `npm install -g @anthropic-ai/claude-code`
2. `lemonade launch claude`
3. Fails instantly with Error 193, no agent window opens.

Expected: Claude Code launches.
Actual: Error 193, process never starts.

### Log Output

```shell
Error: Failed to launch agent process: Failed to start process
'C:\Users\<User>\AppData\Roaming\npm\claude': %1 is not a valid Win32 application. (Error code: 193)
```

### Additional Context

Likely fix: on Windows, prefer the `.cmd` or `.exe` launcher over the extensionless script when picking which `claude` file to run (around `find_agent_binary` in `src/cpp/cli`).

Related: blocks #2063 on Windows, since Claude can't launch here at all, so the context-size issue in #2063 isn't reachable until this is fixed.

anditherobot:
## Summary

On Windows, \lemonade launch claude\ fails instantly with Error 193 (\%1 is not a valid Win32 application\) and the agent never starts.

\
pm install -g\ creates three launchers in \%APPDATA%\\npm\: a Unix shell script named just \claude\, plus \claude.cmd\ and \claude.ps1\. The CLI searched for the bare name \claude\ first, found the shell script, and Windows cannot run shell scripts, so the launch failed before the agent started.

## Fix

Look for \claude.cmd\ / \claude.exe\ before the bare name, the same way the opencode config already does. On Windows the search order is now \claude.cmd\ -> \claude.exe\ -> \claude\. Linux and macOS are unchanged.

## Testing

- New Windows-only test \	est_115_launch_claude_windows_prefers_cmd_over_npm_shim\ recreates the npm layout (a shell script named \claude\ next to \claude.cmd\ on PATH) and runs \lemonade launch claude\. It fails with Error 193 on the old code and passes with the fix.
- Built the \lemonade\ CLI with the \windows\ preset (VS 2022); compiles cleanly.
- Manual check: \
pm install -g @anthropic-ai/claude-code\, then \lemonade launch claude\ starts normally.

## Notes

Fixes #2211. On Windows, the ctx_size issue from #2063 is also present once the launcher works, but that is out of scope for this PR and is being addressed separately.

### DIFF
--- a/src/cpp/cli/agent_launcher.cpp
+++ b/src/cpp/cli/agent_launcher.cpp
@@ -97,10 +97,13 @@ void configure_claude_agent(const std::string& base_url,
                             AgentConfig& config) {
     const std::string resolved_api_key = api_key.empty() ? kDefaultAgentApiKey : api_key;
 
-    config.binary_name = "claude";
 #ifdef _WIN32
-    config.binary_alternatives = {"claude.cmd", "claude.exe"};
+    // npm puts a Unix shell script named just "claude" next to claude.cmd.
+    // Windows cannot run the shell script, so look for .cmd/.exe first.
+    config.binary_name = "claude.cmd";
+    config.binary_alternatives = {"claude.exe", "claude"};
 #else
+    config.binary_name = "claude";
     config.binary_alternatives = {};
 #endif
     config.fallback_paths = {
--- a/test/server_cli2.py
+++ b/test/server_cli2.py
@@ -1232,6 +1232,42 @@ sys.exit(0)
                 payload["env"]["ANTHROPIC_BASE_URL"], f"http://localhost:{PORT}"
             )
 
+    def test_115_launch_claude_windows_prefers_cmd_over_npm_shim(self):
+        """On Windows, launch must run claude.cmd, not npm's extensionless shell script."""
+        if not IS_WINDOWS:
+            self.skipTest(
+                "Windows-only: npm shim vs .cmd resolution does not apply on Unix"
+            )
+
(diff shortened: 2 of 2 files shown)


fl0rianr (commented):
Looks good to me. The change is narrowly scoped and matches the existing Windows launcher strategy used for opencode: prefer the Windows-native launcher before falling back to the extensionless npm shim.

The new Windows-only regression test recreates the important npm layout (claude next to claude.cmd) and verifies that claude.cmd is actually executed, so it should catch the Error 193 regression from #2211.

Only minor note: the PR text mentions unblocking #2063 on Windows, but #2063 itself is filed as a Linux/Debian context-size issue, so I would avoid implying that this PR fixes #2063 itself.

I’d be comfortable merging once CI / a Windows test run is green.

fl0rianr (approved):

PR #2214, opened 2026-06-12
Capture output from `lemonade bench`

### LINKED ISSUE #2056: Capture model output from `lemonade bench`
### Feature Description

add a flag to save the output from the test.

### Use Case / Motivation

I have some multi-turn test scenarios and I'd like to look at the actual model output to see how it improves or degrades based on context size or other back-end flags.

### Platform Relevance

All platforms

### Additional Context

`lemonade bench --raw-output model-says-what.log ...`

I'm not imagining doing a lot of formatting or processing to the model output before writing it out; evaluating that seems like a very user-specific process.

ckuethe:
Useful for quality assessment, or investigating why an unnamed model debugged some code at 1000000tps.

fixes #2056

Co-authored-by: copilot, GPT-5.3-Codex

### DIFF
--- a/src/cpp/include/lemon_cli/bench.h
+++ b/src/cpp/include/lemon_cli/bench.h
@@ -134,7 +134,7 @@ BenchRunResult run_single_bench(lemonade::LemonadeClient& client,
                                 const std::string& model,
                                 const BenchScenario& scenario,
                                 bool memory_tracking,
-                                bool capture_response = false);
+                                bool capture_response);
 
 // Run a full scenario (warmup + measurement runs).
 // When reload=true, unloads+loads the model before each measurement run to clear prompt cache.
@@ -149,8 +149,7 @@ BenchScenarioResult run_scenario(lemonade::LemonadeClient& client,
                                  const std::string& backend,
                                  int ctx_size,
                                  const std::string& backend_args,
-                                 bool capture_response = false,
-                                 const std::string& response_log_path = "",
+                                 const std::string& response_log_path,
                                  const std::string& response_timestamp = "");
 
 // ============================================================
--- a/src/cpp/include/lemon_cli/bench.h
+++ b/src/cpp/include/lemon_cli/bench.h
@@ -39,6 +39,7 @@ struct BenchRunResult {
     double vram_gb = -1.0;      // -1 means not available
     double memory_gb = -1.0;    // -1 means not available
     bool success = true;        // false if the run failed (exception, HTTP error, etc.)
+    std::string response_text;  // LLM response text (only populated if capture enabled)
 };
 
 struct BenchScenarioResult {
@@ -132,7 +133,8 @@ bool unload_all_models(lemonade::LemonadeClient& client);
 BenchRunResult run_single_bench(lemonade::LemonadeClient& client,
                                 const std::string& model,
                                 const BenchScenario& scenario,
-                                bool memory_tracking);
+                                bool memory_tracking,
+                                bool capture_response = false);
 
 // Run a full scenario (warmup + measurement runs).
 // When reload=true, unloads+loads the model before each measurement run to clear prompt cache.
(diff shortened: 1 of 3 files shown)


fl0rianr (commented):
Thanks for adding this — the overall shape looks good and it matches the debugging/quality-assessment use case from #2056.

I think there is one small behavior edge case worth considering before merge:

```cpp
if (!response_log_path.empty() && !run_result.response_text.empty())
```
This means a successful run with an empty captured response would not produce a JSONL row. That may be rare, but for a response/debug log it might be more useful to log the run whenever capture was requested and a response field was present, even if the response string is empty. Would it make sense to track that separately, e.g. with a response_captured bool or std::optional<std::string>?

Related minor hardening: the capture path currently reads message.content via get<std::string>(). If a backend ever returns content: null or another non-string JSON value, response capture could throw and make the whole benchmark run fail. Other CLI code seems to guard this with is_string() before reading content, so it might be worth doing the same here and keeping response logging best-effort.

Tiny docs nits:

typo:
qualtity → quality
and maybe:
saved to the name destination → saved to the named destination or saved to FILE

ckuethe:
let's not worry about content type right now. I have other diffs in progress for extending lemonade bench to TTS, STT, and image generation - as those move forward I'll see about saving their responses in some reasonable encoding (probably base64). It doesn't look like the servers include a mime-type; at least, I didn't see one. If that could be relied upon, I'd use it.

I haven't see any empty output being produced that wasn't also paired with an obvious server error, such as a timeout or a crash. I guess I could set the field to null to signal "hey you got nothing back" instead of not logging. Yeah, good catch.

ckuethe:
Adjusted the logging some. I'm not sure how to get a server to produce no output, other than perhaps writing an intentionally bogus one, but now you get back `"null"`, and you can check the number of output tokens. Setting `max_tokens` to 0 in the scenario file causes at least 1 token to be generated.

0 output tokens would tell me that something strange happened, and never mind the output - go look at the server logs to see if it crashed or something.

bitgamma:
looks good to me. @ckuethe please fix the merge conflict before I test and approve this.

ckuethe:
Rebased. Tested against local and remote host, both produce sensible output captures.

Fixed the other typo Florian pointed out.

src/cpp/include/lemon_cli/bench.h
```diff
@@ -132,7 +133,8 @@ bool unload_all_models(lemonade::LemonadeClient& client);
 BenchRunResult run_single_bench(lemonade::LemonadeClient& client,
                                 const std::string& model,
                                 const BenchScenario& scenario,
-                                bool memory_tracking);
+                                bool memory_tracking,
+                                bool capture_response = false);
```
bitgamma: default values are best avoided, let's just make sure the function is always called with all parameters

src/cpp/include/lemon_cli/bench.h
```diff
@@ -146,7 +148,10 @@ BenchScenarioResult run_scenario(lemonade::LemonadeClient& client,
                                  const std::string& recipe,
                                  const std::string& backend,
                                  int ctx_size,
-                                 const std::string& backend_args);
+                                 const std::string& backend_args,
+                                 bool capture_response = false,
+                                 const std::string& response_log_path = "",
```
bitgamma: same thing here, let's not add default parameters unless absolutely necessary

src/cpp/include/lemon_cli/bench.h
```diff
@@ -146,7 +148,10 @@ BenchScenarioResult run_scenario(lemonade::LemonadeClient& client,
                                  const std::string& recipe,
                                  const std::string& backend,
                                  int ctx_size,
-                                 const std::string& backend_args);
+                                 const std::string& backend_args,
+                                 bool capture_response = false,
```
bitgamma: no need to have a separate capture_response param here since it can be determined by response_log_path being empty or not. It makes it clearer that you can only capture the output if a log path is provided

bitgamma (changes requested):
minor nits, after these are fixed we are ready to go

bitgamma (approved):

PR #2216, opened 2026-06-13
Allow overriding default -ngl for llama-server (llamacpp)

### LINKED ISSUE #2215: Allow overriding the default `-ngl#` for llamacpp server
Currently, when running the llamacpp server, for GPU the ngl flag is set to `-ngl 99` and for CPU to `-ngl 0`.  There is currently no way to override this flag using the llamacpp.args field in the `config.json` file.  This is needed in order to load models that won't fit in physical memory.

amd-pworfolk:
## Summary
- When `llamacpp_args` does not set GPU layer flags, Lemonade still injects the default `-ngl` (`99` on GPU backends, `0` on CPU) and reserves `-ngl` / `--gpu-layers` / `--n-gpu-layers` for validation.
- If the user passes any of those flags in `llamacpp_args` (including `--gpu-layers=N`), Lemonade skips the default so the value is fully user-controlled—needed for models that do not fit in VRAM/RAM at full offload.
- Added `custom_args_contains_any_flag()` in `custom_args.h`, updated API/CLI help text, and an integration test covering `-ngl`, and `--gpu-layers`.

Fixes #2215

## Test plan
- [x] `python test/server_endpoints.py` (or at least `test_010b_load_model_llamacpp_gpu_layer_override_args`) against a running `lemond` with `Tiny-Test-Model-GGUF` pulled
- [x] Manual: `POST /v1/load` with `llamacpp_args` containing `-ngl <N>` and confirm load succeeds and llama-server receives the intended layer count

Made with [Cursor](https://cursor.com)


amd-pworfolk:
Example of running a VERY large MoE model:

- `lemonade config set disable_model_filtering=true`
- `lemonade config set llamacpp.args="--mmap -ngl auto"`
- `lemonade pull unsloth/Qwen3.5-397B-A17B-GGUF:Q4_K_M`
- Strix Halo with 128 GB of DDR configured with minimum dedicated VRAM
- llamacpp with Vulkan backend
- This model will run more than 2X faster (~4 TPS) in Linux than in Windows (<2 TPS)

ramkrishna2910:
@claude review

amd-pworfolk:
> **Claude finished @ramkrishna2910's task in 2m 13s** —— [View job](https://github.com/lemonade-sdk/lemonade/actions/runs/27584315231)
> 
> ### Code Review: Allow overriding default `-ngl` for llama-server
> * [x]  Read PR context
> * [x]  Review `custom_args.h` changes
> * [x]  Review `llamacpp_server.cpp` changes
> * [x]  Review `bench.cpp`, `recipe_options.cpp`, docs
> * [x]  Review test changes
> * [x]  Post review feedback
> 
> Overall this is a clean, well-scoped change that does exactly what the summary describes. The mechanism is correct: by skipping the `push_arg(... "-ngl" ...)` call when the user supplies a GPU-layer flag, you avoid inserting `-ngl`/`--gpu-layers`/`--n-gpu-layers` into `reserved_flags`, so `validate_custom_args()` no longer rejects them and the user's tokens are appended verbatim. Nice and minimal.
> 
> #### Correctness ✅
> * `custom_args_contains_any_flag()` (`custom_args.h:95`) correctly reuses `parse_custom_args()`, skips non-flag tokens, and strips the `=value` suffix before comparing — consistent with how `validate_custom_args()` handles `--flag=value`.
> * The skip-default logic in `llamacpp_server.cpp:400-412` is symmetric: when skipped, _none_ of the three aliases are reserved, so `-ngl`, `--gpu-layers`, and `--n-gpu-layers` all become user-controllable. Good.
> * `static const std::vector<std::string> k_gpu_layer_flags` is a function-local static — thread-safe initialization and read-only after, which is fine for the concurrent router.
> * Docs (`lemonade.md`, `cli.md`), CLI help (`bench.cpp:1502`), and recipe-option help (`recipe_options.cpp:250`) are all consistent with the code's `-ngl 99` (GPU) / `-ngl 0` (CPU) defaults.
> 
> #### Suggestions (non-blocking)
> 1. **PR body is now out of date vs. test coverage.** The summary says the integration test covers `-ngl`, `--gpu-layers=`, and `--n-gpu-layers`, but commit `0401df9 "Only test -ngl and --gpu-layers flags"` reduced `test_010b` to just `-ngl 35` and `--gpu-layers 40` (space form). As a result, the `--n-gpu-layers` alias and the **`--flag=value` parsing branch** in `custom_args_contains_any_flag()` (`custom_args.h:106-109`) are not exercised by any test. Since `custom_args.h` has no unit tests at all, a tiny standalone unit test for the helper (covering `--gpu-layers=40` and `--n-gpu-layers`) would lock in that branch cheaply without depending on a live llama-server build's flag quirks. Worth at minimum updating the PR description to match what's actually tested. [Fix this →](https://claude.ai/code?q=Add%20a%20small%20C%2B%2B%20unit%20test%20for%20lemon%3A%3Autils%3A%3Acustom_args_contains_any_flag%20in%20src%2Fcpp%2Finclude%2Flemon%2Futils%2Fcustom_args.h%2C%20covering%20the%20--gpu-layers%3D40%20equals-form%20and%20--n-gpu-layers%20alias%20on%20branch%20pworfolk%2Fmoe&repo=lemonade-sdk/lemonade)
> 2. **CPU backend + nonzero `-ngl` passthrough.** A user can now pass `-ngl 35` on a CPU-only backend, where the old code forced `-ngl 0`. That's the intended escape hatch, but a CPU-only llama-server build may ignore/reject it. This matches the test comment about builds that "leave a bad process if ignored." Not a regression, but a one-line note in the docs that overriding `-ngl` is the user's responsibility on CPU builds could save a support question.
> 3. **Minor nit:** `custom_args_contains_any_flag()` rebuilds the `std::set<std::string> candidates` on every call. Negligible here (called once per load with 3 candidates), so no action needed — just flagging that the candidate list could be passed as a `std::set` if it were ever hot.
> 
> #### Invariants
> No quad-prefix endpoints, recipes, or NPU-exclusivity logic touched — none of the critical invariants in `AGENTS.md` are affected. Cross-platform-safe (no platform-specific code).
> 
> LGTM pending the test/PR-description alignment in suggestion #1. • `pworfolk/moe`

@ramkrishna2910 Suggestion #1 is addressed.

bitgamma:
@amd-pworfolk, sorry I didn't see this before but #2235 which is in the merge queue right now also allows users to pass -ngl (and actually doesn't pass any default value anymore since llamacpp handles this value nicely automatically)

amd-pworfolk:
> @amd-pworfolk, sorry I didn't see this before but #2235 which is in the merge queue right now also allows users to pass -ngl (and actually doesn't pass any default value anymore since llamacpp handles this value nicely automatically)

Thanks for letting me know.  The llamacpp default behavior is what I need, so I will test the new main branch!

amd-pworfolk:
@ramkrishna2910  @bitgamma  I'm closing this PR since #2235 which has merged already removed the default "-ngl 99" flag from llamacpp, thus allowing any override.

MoE models larger than available memory can be run using the `-mmap` flag.  For example, on my 128 GB Strix Halo device:
```
lemonade config set disable_model_filtering=true
lemonade config set llamacpp.args="--mmap"
lemonade pull unsloth/Qwen3.5-397B-A17B-GGUF:Q4_K_M
```
This works best on Linux with Llamacpp Vulkan backend and I've configured the dedicated VRAM to be at the minimum value.

PR #2217, opened 2026-06-13
Add inotify to extra models directory

superm1:
Extra models directory main purpose is to share with other software.  But if other software adds to this directory without restarting lemond then lemond won't know it exists.

Add an inotify to tell when this has happened.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -270,6 +270,14 @@ jobs:
           package-type: binary
           deb-version: ${{ steps.get_version.outputs.version }}
 
+      - name: Run C++ unit tests
+        run: |
+          set -e
+          cmake --preset default
+          cmake --build --preset default --target test_directory_watcher
+          cd build
+          ctest --output-on-failure -R DirectoryWatcher
+
       - name: Upload .deb package
         uses: actions/upload-artifact@v7
         with:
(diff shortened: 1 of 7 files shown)


fl0rianr (changes requested):
Thanks for adding this — the feature is useful, but I think this needs another pass before merge.

The main issue is that the watcher does not appear to start on the normal server startup path. Server constructs ModelManager(config_->extra_models_dir()), but the ModelManager constructor only stores/logs the directory; the watcher is only started from set_extra_models_dir(). So this likely does not solve the configured extra-models-dir case.

I also see lifecycle/platform issues in DirectoryWatcher: macOS and the fallback implementation do not join the watcher thread in stop(), despite the API contract saying stop waits for the thread. That can terminate during destruction/reset.

The macOS implementation also looks incorrect: EVFILT_VNODE should be registered on an opened directory file descriptor, not on dir_path_.c_str() cast to uintptr_t; and the pipe wakeup path stores the read end but writes to it in stop().

Finally, the fallback currently calls the callback every 200ms even without a filesystem change, which would keep invalidating the model cache continuously. It should compare a directory snapshot or be scoped out.

fl0rianr:
Thanks, the startup path and build fixes look much better now.

I still see one functional issue before I’d approve: in the macOS watcher loop, `kevent()` can return 0 on timeout, but the code still falls through and invokes the callback. That means the model cache would be invalidated every 200ms even when the directory did not change.

Could you add if (n == 0) continue; after the kevent() call before processing events / invoking the callback?

After that I’m good with this.

fl0rianr (approved):
Thanks for iterating on this, this looks good to me.

PR #2220, opened 2026-06-13
feat(config): honor LEMONADE_DEFAULTS_PATH for non-FHS installs

noamsto:
**What** — `get_defaults()` merges `/usr/share/lemonade/defaults.json` so distro packagers can seed default config (e.g. backend `*_bin` paths). Non-FHS distros such as Nix and Guix cannot write under `/usr/share`. This adds a `LEMONADE_DEFAULTS_PATH` env var that merges a `defaults.json` from an arbitrary path, at the same precedence as the `/usr/share` file (below the user `config.json`).

**Why** — Packaging Lemonade for NixOS: v10.7.0 removing the `*_bin` env→config migration left no declarative way to point lemonade at nix-store backend binaries. `/usr/share/lemonade/defaults.json` is the intended hook but is not writable on non-FHS systems. This mirrors the `LEMONADE_GGML_HIP_PATH` escape hatch from #2044.

**Change** — 6-line addition in `get_defaults()` plus `#include <cstdlib>`. Cross-platform (outside the `#ifndef _WIN32` block) so Windows packagers can use it too.

**Testing** — Built `lemond` with this change and ran it with `LEMONADE_DEFAULTS_PATH` set; the merged config picked up the supplied values (verified the resulting `config.json`, and `lemonade backends` showing a path-overridden backend as `installed`).

### DIFF
--- a/src/cpp/server/config_file.cpp
+++ b/src/cpp/server/config_file.cpp
@@ -2,6 +2,7 @@
 #include "lemon/utils/json_utils.h"
 #include "lemon/utils/path_utils.h"
 
+#include <cstdlib>
 #include <filesystem>
 #include <fstream>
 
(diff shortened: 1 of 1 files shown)


noamsto:
@superm1 — flagging this for you as it's in the Linux/distro-packaging area (non-FHS defaults seeding for the NixOS package). Would you be able to review, or point me to the right person? Happy to discuss on Discord if preferred.

superm1 (approved):

superm1:
Would you please update documentation for environmental variables to cover this new item too?

PR #2225, opened 2026-06-13
fix(ui): audit follow-up — tool-loop hang guard, load errors, preset wiring, capability unification

kpoineal:
Implements the top 4 actionable items from Mattingly's fresh audit of the prototype at HEAD `af66ea14`.

### P0 — Tool-loop hang guard
**`prototype/ui-redesign/src/hooks/useChatStreaming.ts`**
Wrapped the async `onToolCalls` body in try/catch so an unexpected executor throw cleanly settles the outer stream Promise instead of hanging forever. Pending tool entries are marked errored, `onError` fires, `cleanup`/`resolve` are called.

### P1 — Surface load errors in ModelManager
**`ModelManager.tsx` + `styles.css`**
Added `loadError` state set in `handleLoad` and `handlePullAndLoad` catch blocks via `friendlyErrorMessage`. Renders `.row__load-error` inline near the affected model row using existing `--danger`/`--danger-soft` tokens. Replaces silent `console.error`.

### P1 — Wire preset values into image composer
**`ChatView.tsx`**
`imageDefaultsForModel` now accepts the active preset's `recipe_options` as a third (highest-priority) source. Existing `imageSettingsTouchedRef` guard prevents clobbering user edits. Applying `Sharp` (30 steps / cfg 8.0) now actually shows in the composer instead of staying at default 20/7.

### P1 — Unify `modeSupportsChatCompletions`
**`ChatView.tsx:731`**
Replaced the `currentLoadedModel ? canUseChatCompletions(...) : ...` branch with `currentCapability === 'chat' || currentCapability === 'omni'`. Custom models with no recipe field now get the tools toggle correctly enabled (`currentCapability` is already enriched via the snapshot system).

## Verification

- `npm run build` exits 0 (3 pre-existing bundle-size warnings, no errors)
- All four touched files re-read end-to-end
- No changes to `src/cpp/`, no new dependencies, no changes to `src/app/` or `src/web-app/`

## Out of scope (deferred)

Audit also flagged: `ScriptProcessorNode` deprecation in `useAudioCapture` (needs AudioWorklet migration, ~30-50 lines + worklet file), `modelSupportsImageEdit` hardcoded name patterns, ChatView size (101KB — needs component extraction), legacy `lemonade_use_tools` migration, `transpileOnly: true` masking `as any` type holes. Tracked in audit decision note.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/prototype/ui-redesign/src/components/ChatView.tsx
+++ b/prototype/ui-redesign/src/components/ChatView.tsx
@@ -292,15 +292,17 @@ function partialImageSettingsFromSource(source?: Record<string, unknown> | null)
   return next;
 }
 
-function imageDefaultsForModel(loadedModel: LoadedModel | null, modelInfo: ModelInfo | null): ImageGenerationSettings {
+function imageDefaultsForModel(loadedModel: LoadedModel | null, modelInfo: ModelInfo | null, activePresetRecipeOptions?: Record<string, unknown> | null): ImageGenerationSettings {
   const modelImageDefaults = partialImageSettingsFromSource(modelInfo?.image_defaults as Record<string, unknown> | undefined);
   const modelRecipeOptions = partialImageSettingsFromSource(modelInfo?.recipe_options as Record<string, unknown> | undefined);
   const loadedRecipeOptions = partialImageSettingsFromSource(loadedModel?.recipe_options);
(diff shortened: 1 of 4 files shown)


fl0rianr (approved):
LGTM

PR #2227, opened 2026-06-14
fix(ui): mobile/responsive layout for phones (480px breakpoint)

kpoineal:
Adds responsive layout for phones. Targets the `squished` rendering Kyle saw on mobile during demo testing.

## Root cause
No phone-specific breakpoint existed. The 7-item titlebar nav (chat / models / presets / backends / dashboard / logs / connect) needs ~425px at minimum width — wider than 360–414px phones. Combined with no phone-specific layout adjustments, content overflowed and cramped.

Viewport meta tag was already correct in `index.html`.

### `tokens.css`
Added documented breakpoint tokens:
- `--breakpoint-mobile: 480px` (phone)
- `--breakpoint-tablet: 768px` (tablet)  
- `--breakpoint-desktop: 1024px` (desktop narrow)

### `styles.css`
New `@media (max-width: 480px)` block with 10 fixes:

1. **Nav becomes horizontal-scroll** — all 7 items reachable via swipe, no overflow
2. **Send/Stop buttons → 44px** (iOS HIG touch target minimum, was 36px)
3. **Composer toolbar wraps** instead of overflowing
4. **Panel padding tightened** on ModelManager, PresetManager, BackendManager, ConnectView
5. **Slide-over goes 100vw** on phone (native-feeling drawer)
6. **Hero section compressed** vertically
7. **Chat-with-logs grid** collapses to single column
8. **Log viewer toolbar** tightened, timestamp column narrowed

Plus a global (all viewport sizes) fix:
- **Markdown tables** get `display: block; overflow-x: auto` to prevent message content from forcing the page wider.

## Verification
- `npm run build` exits 0, no new errors
- Live-tested on Kyle's phone via dev server at `192.168.3.35:8080` with HMR

## Out of scope (deferred)

- **Nav button height** still ~21px — full fix needs icon-only view or bottom tab bar pattern in `App.tsx` (React change). For now, scrollable nav makes everything reachable.
- `icon-btn` and `titlebar__theme-toggle` remain 32px (secondary controls, lower urgency)
- Bottom tab bar would be the ideal long-term phone nav pattern

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #2229, opened 2026-06-14
fix(ui): mobile round 2 — icons-only nav, stacked model cards

kpoineal:
Round 2 of mobile fixes after Kyle tested PR #2227 on his 390px iPhone. Two specific issues remained.

## Issue 1 — Nav still clipping
Even with horizontal-scroll, `Dash` label was visually clipping and the right-side controls (avatar, theme toggle, status dot) were cramped on 390px iPhone.

**Chose Option A — icons-only nav at ≤480px** (over hamburger drawer, which would have been more invasive for a POC):
- All 7 nav icons already existed in `Icon.tsx` with clear semantics
- Wrapped labels in `<span class='nav-label'>`, hidden at 480px → nav shrinks ~425px → ~196px, fits comfortably with no scroll
- Added `title` + `aria-label` on every button for accessibility
- `.titlebar__right` now has `flex-shrink: 0` so right cluster never collapses

## Issue 2 — Model card overlap
Model cards had backend badge on left overlapping name/metadata, names truncating to single characters like `r`.

**Stacked vertically at ≤480px:**
- `.row__content` becomes `flex-direction: column`
- Backend badge: full-width top
- Model name wraps freely on its own row
- `.row__action` row goes full-width at bottom with grow buttons

## Files changed
- `src/App.tsx` — nav button markup wraps label in span; `title` + `aria-label`
- `src/styles/styles.css` — `.titlebar__nav button` flex tweaks; `.titlebar__right { flex-shrink: 0 }`; existing 480px block extended with `.nav-label { display: none }` + model-card column overrides

## Verification
- `npm run build` exits 0
- HMR delivered live to Kyle's phone via dev server during implementation

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


fl0rianr (changes requested):
DO NOT MERGE!

Nice you are working on this.
The intended change seems to be limited to the mobile layout in prototype/ui-redesign/src/App.tsx and prototype/ui-redesign/src/styles/styles.css, but the PR currently shows 557 changed files with ~159k additions/deletions. Most of that appears to be line-ending churn, with many files converted from LF to CRLF.

Could you please recreate or clean the branch so that it only contains the actual UI changes, and make sure shell/source files remain LF-normalized? I would suggest branching from the current base, cherry-picking only the intended UI changes, then checking git diff --stat before pushing. The final PR should probably only touch the relevant UI files, not the whole repository.

Thanks!

PR #2232, opened 2026-06-14
fix: return 400 instead of 500 when request body is empty or not valid JSON

### LINKED ISSUE #2231: POST endpoints return 500 instead of 400 when request body is empty or malformed JSON
### Platform

Windows

### Lemonade Version

any

### GPU / APU Model

any

### Component

Other / Not sure

### Bug Description

 Bug Description

`handle_chat_completions`, `handle_pull`, `handle_load`, and `handle_delete` 
all call `nlohmann::json::parse(req.body)` directly with no empty-body guard.

When a client sends an empty body or malformed JSON, `nlohmann` throws a 
`json::parse_error`. This is caught by the generic `catch (const std::exception& e)` 
block and mapped to **HTTP 500 (Internal Server Error)**.

The correct response is **400 (Bad Request)** — the client sent a bad request, 
the server did not crash.

### Steps to Reproduce

 Any of these return 500 instead of 400:
curl -X POST http://localhost:13305/v1/pull
curl -X POST http://localhost:13305/v1/load
curl -X POST http://localhost:13305/v1/delete
curl -X POST http://localhost:13305/v1/chat/completions
```

 Expected: `400 {"error": "Request body is required but was empty"}`  
 Actual:  `500 Internal Server Error`

### Log Output

```shell

```

### Additional Context

`handle_unload` already handles this correctly with `if (!req.body.empty())`. 
The other handlers were never updated to match.

Kushal1213:
Closes lemonade-sdk/lemonade#2231

#2231  Bug

`handle_chat_completions`, `handle_pull`, `handle_load`, and `handle_delete` 
all call `nlohmann::json::parse(req.body)` with no empty-body guard.

An empty or malformed body throws `nlohmann::json::parse_error`, which is 
caught by the generic `std::exception` handler and mapped to HTTP 500 
(Internal Server Error). The correct status is 400 (Bad Request) — the 
client sent a malformed request, not the server that crashed.

`handle_unload` already guards with `if (!req.body.empty())` correctly. 
This PR brings the other handlers into line.

 Fix

Adds a single shared helper `parse_required_json_body()` (declared in 
`server.h`, implemented in `server.cpp`) that:
- Returns 400 with `"Request body is required but was empty"` on empty body
- Returns 400 with `"Invalid JSON in request body: ..."` on malformed JSON  
- Returns true with the parsed object on success

The 4 critical handlers are updated to use it.

Before / After

Before:
`curl -X POST http://localhost:13305/v1/pull` → `500 Internal Server Error`

After:
`curl -X POST http://localhost:13305/v1/pull` → `400 {"error": "Request body is required but was empty"}`

No behaviour change for valid requests.


superm1 (commented):
Can you please drop the pointless comments?

superm1 (approved):

PR #2235, opened 2026-06-14
Automatically determine context size

bitgamma:
This calculates (or estimates conservatively for non-GGUF models) the memory available for context and set context size to `min(max_model_context_size, max_ctx_size_fitting)`. There are several fallbacks mechanisms and eventually the good old `4096` is returned if everything else fails.

I only have a Strix Point with 64GB/48GB GTT and a Strix Halo 128GB to test on so I could only verify that. Verified with both FLM and llamacpp and it seems to work fine. For FLM it can only do (pretty conservative) estimates since it cannot read the model architectural info.

Migration to the new "auto" value for existing users is handled. (thanks @superm1!)

All help with testing on different configurations is greatly appreciated. cc'ing @jeremyfowers since he mentioned the need for this feature to exist

### Changes                                                                                                                                                                                                                                                                                                                                                                                                                     

- The default `ctx_size` is changed from `4096` to `-1`, which triggers automatic computation based on available memory and model architecture.                                                                                                                                                                                                                                               - GGUF metadata extraction refactored to read all metadata in one pass instead of opening the file multiple times.                                                                                                                                                                                                                   
- After computing effective recipe options, calls resolve_auto_ctx_size() and patches the resolved ctx_size into options before loading the model.                                                                                                              
- llamacpp_server.cpp: removed hardcoded --context-shift, --keep (these are old workaround to try to keep the system prompt in context with tiny contexts, not relevant), -ngl args (llamacpp automatically offloads as much as possible to VRAM, not specifying this value allows it to do partial offload and avoid OOM issues).                                                                                                                                    
- runtime_config.cpp: removed the validation that rejected ctx_size <= 0, allowing -1 as a valid auto-trigger value


fl0rianr (changes requested):
Thanks for the PR — the direction looks useful, nice!
But before merging, let us look at those two poinst:

1. ctx_size validation now accepts any negative integer. Since only `-1` is handled as auto and other negative values are passed through to `llama-server --ctx-size`, validation should allow only -1 or positive integers.

2. Auto-tuning currently uses model_info.device, but for llama.cpp that defaults to GPU. The actual llamacpp_backend is only resolved later in LlamaCppServer::load(), so llamacpp_backend=cpu can still compute context size from GPU memory. The auto calculation should derive the target device from the effective backend options, or move resolution into the backend after backend/device selection.

Also worth fixing:
- read_gguf_metadata() now gathers context/KV metadata and capabilities in one pass, so the early exit after vision+tool_calling can skip later KV metadata.
- Use apply_gguf_capability_labels() instead of direct push_back() to avoid duplicate labels on cache updates.
- Don’t restrict GGUF integer metadata reads to `type <= 4`; read_gguf_integer_value() already supports INT32/UINT64/INT64.

bitgamma:
1. indeed, will do.
2. no need to fix this imho, the estimate will be lower thus won't cause OOM. Auto-Optimization for CPU inference doesn't sound interesting.

Agreed for the other 3 points, will fix tomorrow.

jeremyfowers:
How is the "automatic" part of this being handled with respect to settings? Most people's config.json file will have a hardcoded `ctx_size=4096` in it, right?

bitgamma:
> How is the "automatic" part of this being handled with respect to settings? Most people's config.json file will have a hardcoded `ctx_size=4096` in it, right?

yes, as I wrote in the PR description this is automatic for new users (since I changed the default.json) but existing users need to opt-in by setting ctx_size=-1. We could treat 4096 as "automatic" but it sounds a bit too hacky to me

superm1:
> > How is the "automatic" part of this being handled with respect to settings? Most people's config.json file will have a hardcoded `ctx_size=4096` in it, right?
> 
> yes, as I wrote in the PR description this is automatic for new users (since I changed the default.json) but existing users need to opt-in by setting ctx_size=-1. We could treat 4096 as "automatic" but it sounds a bit too hacky to me

how about if we detect specifically it's an upgrade at runtime (IE it exists already and we're going from 10.7 to something newer) and unset the config option if it's set to 4096?

fl0rianr (commented):
Thanks for the updates — the GGUF metadata changes look better now.

The only concrete correctness issue I still see from my earlier review is the ctx_size validation: the current check rejects values < -1, but still allows 0. Since -1 is the auto sentinel and explicit context sizes should be positive, I think this should reject 0 as well, e.g. allow only ctx_size == -1 || ctx_size > 0.

On the existing-user config behavior: I agree that treating every 4096 as auto would be too implicit. The runtime-upgrade migration idea sounds reasonable if we want existing users to opt into the new default automatically, but I’d defer to you whether that should be part of this PR or a follow-up. At minimum, the PR/release notes should make the existing-user behavior clear. Thanks!

bitgamma:
> The only concrete correctness issue I still see from my earlier review is the ctx_size validation: the current check rejects values < -1, but still allows 0. Since -1 is the auto sentinel and explicit context sizes should be positive, I think this should reject 0 as well, e.g. allow only ctx_size == -1 || ctx_size > 0.

@fl0rianr `0` is a valid value at least for llamacpp and means "use max context size". It is the value I have always used so I'd rather not ban it :)

> how about if we detect specifically it's an upgrade at runtime (IE it exists already and we're going from 10.7 to something newer) and unset the config option if it's set to 4096?

@superm1 what would be the correct way to do this? Is it safe to assume that if the value is 4096 it wasn't set explicitly?

superm1:
> > The only concrete correctness issue I still see from my earlier review is the ctx_size validation: the current check rejects values < -1, but still allows 0. Since -1 is the auto sentinel and explicit context sizes should be positive, I think this should reject 0 as well, e.g. allow only ctx_size == -1 || ctx_size > 0.
> 
> @fl0rianr `0` is a valid value at least for llamacpp and means "use max context size". It is the value I have always used so I'd rather not ban it :)
> 
> > how about if we detect specifically it's an upgrade at runtime (IE it exists already and we're going from 10.7 to something newer) and unset the config option if it's set to 4096?
> 
> @superm1 what would be the correct way to do this? Is it safe to assume that if the value is 4096 it wasn't set explicitly?

How about we introduce a schema version to the json?

Then we can always have the concept of migrating fields from default json to the config json in the future if the schema is out of date.

bitgamma:
> How about we introduce a schema version to the json?

> Then we can always have the concept of migrating fields from default json to the config json in the future if the schema is out of date.

The schema is versioned already, but in this case we are only changing what the default value is for a field, which doesn't change the schema. Since the default value is written out explicitly we cannot know, in theory, if the user actually wanted 4096 or not. I guess we can relatively safely assume that they did *not* actually want this since it is unusable.

fl0rianr (approved):
Thanks for the clarifications and updates. The GGUF metadata handling looks good to me now, and the remaining config-default/migration question seems to be on the right track in the discussion.

Approving from my side.

superm1:
> > How about we introduce a schema version to the json?
> 
> > Then we can always have the concept of migrating fields from default json to the config json in the future if the schema is out of date.
> 
> The schema is versioned already, but in this case we are only changing what the default value is for a field, which doesn't change the schema. Since the default value is written out explicitly we cannot know, in theory, if the user actually wanted 4096 or not. I guess we can relatively safely assume that they did *not* actually want this since it is unusable.

So maybe bump the schema on upgrade even if the field names stay stable? So we use the schema upgrade as a reason to migrate this field.

superm1:
Here's an example of what I was thinking: https://github.com/lemonade-sdk/lemonade/pull/2245.

Feel free to use that if you like the ideas.

bitgamma:
@superm1 thanks! This looks good to me and solves the problem. @jeremyfowers, last time you said you had different ideas about when the schema version field is to be bumped so please confirm if this looks good for you too

jeremyfowers:
I am good with bumping the schema version for this. Honestly I think this is one of the most important features in a long while, so we should make sure it reaches the existing userbase.

My only nitpick is that we have a convention of using “auto” for automatic selection, and I moderately prefer using that here as well instead of -1. Explicit over implicit and all that.

bitgamma:
@jeremyfowers I tried using "auto" but since ctx_size is expected to be an integer trying to use "auto" adds a good share of complexity in several places (especially CLI args parsing). I'd say we keep the internal representation an integer (-1) and then later we show it as auto in the GUI (just like 0 should probably be shown as "max", even if this is technically handled by llamacpp).

jeremyfowers:
> @jeremyfowers I tried using "auto" but since ctx_size is expected to be an integer trying to use "auto" adds a good share of complexity in several places (especially CLI args parsing). I'd say we keep the internal representation an integer (-1) and then later we show it as auto in the GUI (just like 0 should probably be shown as "max", even if this is technically handled by llamacpp).

-1 is fine :)

jeremyfowers:
Trying this out on a 24 GB unified memory system to see how it deals with big models.

docs/guide/configuration/README.md
```diff
@@ -42,7 +42,7 @@ Values set in the user's `config.json` always take precedence over these seeded
   "no_broadcast": false,
```
jeremyfowers, line 42: Image

The GUI shows 4k in the Load Options menu even when the global context size config is -1 and there's no recipe_options set.

jeremyfowers: Out of scope for this PR, but it would be nice as a follow-up if /models/{id} could preview the context size estimate and display it here.

docs/guide/configuration/README.md
```diff
@@ -42,7 +42,7 @@ Values set in the user's `config.json` always take precedence over these seeded
...
   "extra_models_dir": "",
   "models_dir": "auto",
```
jeremyfowers, line 44: On a 24 GB RAM MacBook Pro, auto-tune determines that context size should be 4k for a 4B model because it fails to detect the amount of RAM available:

```
2026-06-15 16:40:47.011 [Debug] (Server) ===== LOAD ENDPOINT ENTERED (Thread: 0x16eadf000) =====
2026-06-15 16:40:47.011 [Info] (Server) Ensuring model loaded: Qwen3.5-4B-GGUF 
2026-06-15 16:40:47.013 [Debug] (Router) Loading model: Qwen3.5-4B-GGUF (checkpoint: unsloth/Qwen3.5-4B-GGUF:Qwen3.5-4B-UD-Q4_K_XL.gguf, recipe: llamacpp, type: llm, device: gpu)
2026-06-15 16:40:47.014 [Debug] (AutoTune) get_available_memory_gb: could not determine memory, returning 0.0resolve_auto_ctx_size: Qwen3.5-4B-GGUF — memory undetectable, returning 4096
2026-06-15 16:40:47.014 [Info] (Router) Auto-tune ctx_size resolved to 4096
2026-06-15 16:40:47.014 [Debug] (Router) Effective settings: ctx_size=4096, llamacpp_device=(none), llamacpp_backend=metal, llamacpp_args=(none), merge_args=true, auto_evict=(none), evict_idle_timeout=300, downsize_idle_timeout=60, evict_weight_factor=1.000000
```

But system-info reports 24 GB correctly and other features, such as filtering the models list by available RAM, work correctly.

jeremyfowers: Claude says this is happening because RAM detection is hardcoded to look at certain devices like amd_igpu.

jeremyfowers: Suggested fix (confirmed it works on this device): https://github.com/lemonade-sdk/lemonade/pull/2257

docs/guide/configuration/README.md
bitgamma: I didn't do changes to the UI in this PR. I'll try to fix this tomorrow

The preview of context size estimate sounds nice! I'll do a follow up PR for it

PR #2237, opened 2026-06-14
fix(ui-redesign): close unclosed @media block in styles.css

siavashhub:


kpoineal (approved):
LGTM — simple fix, merging into `kpoin/ui-testing`. ✅

PR #2240, opened 2026-06-14
Generate an artifact for Debian 13

superm1:
We don't have PPAs in Debian (although there is debusine - so that might be worth looking at later).

Currently the artifact gets made on Github and added to the release.
As we already have this build happening on Debian 13 remove the other build doing a similar role.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -277,6 +277,74 @@ jobs:
           path: ${{ steps.build_deb.outputs.output-dir }}/*.deb
           retention-days: 7
 
+  build-lemonade-debian13:
+    name: Build Lemonade .deb Package (Debian 13)
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get_version.outputs.version }}
+    container:
+      image: debian:stable
+    steps:
+      - name: Enable Trixie backports
+        run: |
+          echo "deb http://deb.debian.org/debian trixie-backports main" > /etc/apt/sources.list.d/trixie-backports.list
+      - name: Install git
+        run: |
+          set -e
+          apt-get update
+          apt-get install -y git
+      - name: Install from backports
+        run: |
+          set -e
+          apt-get install -y -t trixie-backports libcpp-httplib-dev node-markdown-it-texmath
+
+      - uses: actions/checkout@v5
+        with:
+          clean: true
+          fetch-depth: 0
+
+      - name: Configure git safe directory
+        run: git config --global --add safe.directory $(pwd)
+
+      - name: Prepare Debian build
+        id: get_version
+        uses: ./.github/actions/prepare-debian-build
+        with:
+          release: '13'
+          codename: 'trixie'
+
+      - name: Build Debian package
+        id: build_deb
+        uses: ./.github/actions/build-debian-package
+        with:
+          package-type: binary
+          deb-version: ${{ steps.get_version.outputs.version }}
+
+      - name: Rename server .deb for release
+        id: rename
+        shell: bash
+        run: |
+          set -e
+          VERSION="${{ steps.get_version.outputs.version }}"
+          OUTPUT_DIR="${{ steps.build_deb.outputs.output-dir }}"
+          SRC=$(ls "$OUTPUT_DIR"/lemonade-server_*_amd64.deb | head -n1)
+          if [ -z "$SRC" ]; then
+            echo "ERROR: server .deb not found in $OUTPUT_DIR"
+            ls -la "$OUTPUT_DIR"
+            exit 1
+          fi
+          DEST="$OUTPUT_DIR/lemonade-server_${VERSION}-debian13_amd64.deb"
+          mv "$SRC" "$DEST"
+          echo "Renamed $SRC -> $DEST"
+          echo "deb-path=$DEST" >> $GITHUB_OUTPUT
+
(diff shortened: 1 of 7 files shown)


.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -277,6 +277,74 @@ jobs:
           path: ${{ steps.build_deb.outputs.output-dir }}/*.deb
           retention-days: 7
 
+  build-lemonade-debian13:
+    name: Build Lemonade .deb Package (Debian 13)
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get_version.outputs.version }}
+    container:
+      image: debian:stable
```
fl0rianr: Could we pin this to debian:trixie or debian:13 instead of debian:stable? Since the artifact name, changelog inputs, and backports suite are all Debian 13/Trixie-specific, `stable`, even if far into future, will eventually drift and could produce a mislabeled or broken Debian 13 artifact.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -2040,6 +2108,7 @@ jobs:
     needs:
       - sign-msi-installers
       - build-lemonade-rpm
+      - build-lemonade-debian13
```
fl0rianr, line 2171: Do you think should add at least a small Debian 13 install/smoke test for this new artifact before making the release depend on it? The existing .deb inference tests still depend on build-lemonade-deb / lemonade-deb, so lemonade-debian13 is currently build-checked but not install-checked on Debian 13.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -277,6 +277,74 @@ jobs:
...
+          deb-version: ${{ steps.get_version.outputs.version }}
+
+      - name: Rename server .deb for release
+        id: rename
+        shell: bash
+        run: |
+          set -e
+          VERSION="${{ steps.get_version.outputs.version }}"
+          OUTPUT_DIR="${{ steps.build_deb.outputs.output-dir }}"
+          SRC=$(ls "$OUTPUT_DIR"/lemonade-server_*_amd64.deb | head -n1)
```
fl0rianr, line 330: Minor robustness nit: this `ls | head -n1` can hide multiple matches, and with `set -e` a missing match may exit before the explicit error handling below runs. Could we use a nullglob array and assert exactly one matching lemonade-server_*_amd64.deb?

docs/index.html
```diff
@@ -440,8 +440,22 @@ <h2 class="hp-section-heading">Always improving.</h2>
             }
             break;
           case 'debian':
-            html = '<a class="gs-download-link" href="https://github.com/lemonade-sdk/lemonade/blob/main/docs/dev/getting-started.md#building-from-source" target="_blank" rel="noopener">' +
-              '<span class="material-symbols-outlined">open_in_new</span> Build from source</a>';
+            if (v) {
+              var deb = 'lemonade-server_' + v + '-debian13_amd64.deb';
+              html = gsConsoleHtml([
+                'wget https://github.com/lemonade-sdk/lemonade/releases/latest/download/' + deb,
```
fl0rianr, line 446: Small user-facing concern: after this docs change lands but before the next release includes the Debian 13 asset, this generated latest/download/...-debian13_amd64.deb URL may 404. Should we either link to the latest release page for now, or only render the direct command once the asset is known to exist?

fl0rianr (commented):
Thanks, this looks directionally good and the new artifact is being produced successfully.
I added some comments you might want to have a look at.

docs/index.html
superm1: The docs only get refreshed on the website at the time of the release.

.github/workflows/cpp_server_build_test_release.yml
superm1: sure, I'll adjust this.

.github/workflows/cpp_server_build_test_release.yml
superm1: I think that's a good idea, I'll get a small smoke test added.

.github/workflows/cpp_server_build_test_release.yml
superm1: we only generate one in this job.  if that changes we have MAJOR problems.

fl0rianr (approved):
Thanks, ok from my side.

AreYouLoco:
Next step? Lemonades SDK custom apt repo? Instead of Ubuntu PPA?

superm1:
> Next step? Lemonades SDK custom apt repo? Instead of Ubuntu PPA?

TBD. If it's possible to use debusine I would like to do that. I'm just not sure how to set it up safely yet. Would need to do some experiments.

https://wiki.debian.org/DebusineDebianNet

superm1:
I think the main challenge is going to be signing before uploading the source package.

For launchpad ppa we have a robot account that does it and so if it's compromised the surface area is just what the robot can access (launchpad). 

For debusine I **think** it needs to be a Debian developer account which i don't feel comfortable attaching to CI.

AreYouLoco:
Well this only if you take route to submit to official Debian repos.

When you do something like this:
https://lemonade-sdk.github.io/flatpak/
But for APT. Then you set the rules and do signing yourself.

superm1:
I have submitted an ITP but gotten multiple rejections from DFSG team (formally ftp master team). I will spin again when I have time for their most recent feedback.

PR #2243, opened 2026-06-14
feat: detect HuggingFace model updates and show Update button in web UI

### LINKED ISSUE #2242: feat: detect HuggingFace model updates and show Update button in web UI
### Feature Description

Currently, the web UI has no way to know if a downloaded model has been updated on HuggingFace. The CLI supports this via `lemonade pull <model>`, which always checks HF for the latest commit, but the web app always uses cache-first semantics. Users must delete and re-download a model to get a newer version, which is wasteful.

The proposal adds a startup check that compares each downloaded model's cached commit hash (`refs/main` in the HF cache) with the latest SHA from HuggingFace, and surfaces the result in the web UI with an Update button.

### Use Case / Motivation

When a model on HuggingFace receives an update (e.g. improved quantization, bugfix GGUF), users of the web UI have no way to discover or apply the update. They must either:

1. Use the CLI (`lemonade pull <model>` - works but is a separate tool)
2. Delete the model and re-download it from scratch

This wastes bandwidth and time, especially for large models. A visible "Update available" indicator with a one-click re-download solves this.

### Platform Relevance

All platforms

### Additional Context

_No response_

blackdeathdrow:
Closes #2242

**Server (C++):**
- New `check_for_model_updates()` on `ModelManager` - scans downloaded models, compares cached `refs/main` commit with latest HF SHA, sets `update_available` flag
- `update_available` included in `/v1/models` JSON response
- `update_check_done` field in `/health` response, set after async startup check completes
- Flag resets to `false` after successful re-download in `update_model_in_cache()`

**Frontend (TypeScript):**
- Update button (download icon, orange hover) shown on downloaded models with `update_available: true`
- Click triggers `/pull` with `do_not_upgrade=false` for incremental re-download
- `useModels.tsx` polls `/health` after mount, checks `update_check_done`, refreshes model list once true

**Documentation:**
- `update_check_done` documented in `/v1/health` response
- `update_available` documented in `/v1/models` response

**Scope:** All HF-backed recipes - `llamacpp`, `ryzenai-llm`, `whispercpp`, `kokoro`, `moonshine`, `sd-cpp`. FLM and cloud models are skipped.


jeremyfowers:
Would this scan all models on HF every time /models is called? That would be a lot of traffic as /models is typically called hundreds or thousands of times a day.

blackdeathdrow:
> Would this scan all models on HF every time /models is called? That would be a lot of traffic as /models is typically called hundreds or thousands of times a day.

Runs once at startup. Then /v1/models reads the flag from cache.

ckuethe:
Can I manually trigger an update on my long-running lemonade instance? I might not want to restart the whole server if it's busy and has a lot of context.

blackdeathdrow:
Agree, I’d also like to have a manual update check.
But I’m not sure about the UI yet, and it requires a new endpoint. Maybe in a separate PR?
Plus, I think this PR is blocked by #2300.

fl0rianr (changes requested):
Thanks, this looks like a good scoped implementation and I like that the HF check runs once during startup cache warmup rather than on every /models request.

One issue I think we should fix before merging: "update_available" can become stale after a model is no longer downloaded. update_model_in_cache() clears it only when downloaded == true, and remove_model_from_cache() marks built-in models as downloaded = false without clearing update_available. That can leave /v1/models returning downloaded: false, update_available: true, which contradicts the documented semantics.

Suggested fix: whenever we set a model to downloaded = false (including delete/remove paths), also set update_available = false.

A smaller follow-up concern: check_for_model_updates() calls HttpClient::get() without a shorter timeout, so it inherits the 300s default. Since the frontend only polls /health for ~15s, slow/offline HF calls could cause the UI to miss the update-check completion. It might be worth using a short timeout for these best-effort metadata checks.

blackdeathdrow:
Fixed:
- `update_available` now cleared in `remove_model_from_cache` alongside `downloaded=false`
- `HttpClient::get` gets a 10s timeout for these metadata checks

fl0rianr (commented):
Thanks, the previous concerns are fixed: update_available is now cleared on delete, and the HF metadata call uses a 10s timeout.

I think there is still one issue worth fixing before merge: the renderer only polls `/health` for about 15s (`30 * 500ms`), but the server checks HF repos sequentially and each repo can take up to 10s. With multiple downloaded repos, slow DNS, captive/offline networking, etc., the update check can easily complete after the renderer has stopped polling. In that case the server eventually sets `update_check_done=true`, but the model list is never refreshed, so the user never sees the update button.

Could we either keep polling with a lightweight backoff until `update_check_done` is true, or otherwise trigger a models refresh when the background check completes?

Small robustness nit: update_check_done_ is only set on the happy path in start_model_cache_warmup(). If warmup/check throws, `/health` can stay false forever. Since this is a best-effort startup check, I think update_check_done_ should be set in a finally-style path even on failure.

blackdeathdrow:
Fixed:
- `update_check_done_` set after catch blocks
- Warmup and update check split into separate try blocks with accurate log messages
- Removed 15s polling in useModels.tsx; uses existing 5s /health poll in ModelManager

fl0rianr (approved):
Looks good to me now.

Small non-blocking nit: useModels.tsx still appears to import `serverFetch` even though the polling logic was removed from that file. Would still be nice if you could go after this.

CI fail is unrelated and a re-run after your next commit will fix this hopefully.

PR #2244, opened 2026-06-15
trivial: Update debian/copyright

superm1:
Sync up with what may be headed to Debian.


fl0rianr (approved):
Thanks, LGTM.

Minor non-blocking question: the source file uses "GPL-2.0 WITH Linux-syscall-note" in its SPDX header. Is the flattened "GPL-2.0-with-linux-syscall-note" spelling in debian/copyright intentional, or would a DEP-5-style expression be preferred? Not blocking since the full license text is included.

superm1:
Intentional thanks!

PR #2245, opened 2026-06-15
Migrate config schema

superm1:
Changes to go with #2235 so that upgraded users get moved over to -1 (and other changes)

### DIFF
--- a/src/cpp/include/lemon/config_file.h
+++ b/src/cpp/include/lemon/config_file.h
@@ -8,6 +8,86 @@ namespace lemon {
 
 using json = nlohmann::json;
 
+// ============================================================================
+// Config migration helpers (static inline for testability)
+// ============================================================================
+
+/// Extract the config_version from a JSON object. Returns 0 if the field
+/// is missing or not an integer — treating the config as "very old".
+static inline int config_get_version(const json& config) {
+    if (config.contains("config_version") && config["config_version"].is_number_integer()) {
+        return config["config_version"].get<int>();
+    }
+    return 0;
+}
+
+/// Migrate a v1 config to v2:
+///   - Remove deprecated "cloud_providers" field
+///   - Upgrade ctx_size 4096 -> -1 (only if user never changed it)
+///   - Bump config_version to 2
+/// Returns true if the config was modified.
+static inline bool config_migrate_v1_to_v2(json& config) {
+    bool changed = false;
+
+    // 1. Remove deprecated cloud_providers field.
+    if (config.contains("cloud_providers")) {
+        config.erase("cloud_providers");
+        changed = true;
+    }
+
+    // 2. Upgrade ctx_size 4096 -> -1 only when it is exactly the old default.
+    if (config.contains("ctx_size") && config["ctx_size"].is_number_integer()) {
+        int ctx_val = config["ctx_size"].get<int>();
+        if (ctx_val == 4096) {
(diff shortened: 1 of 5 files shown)


src/cpp/server/runtime_config.cpp
```diff
@@ -461,6 +435,10 @@ void RuntimeConfig::validate(const std::string& key, const json& value) const {
         if (channel != "stable" && channel != "nightly") {
             throw std::invalid_argument("'rocm_channel' must be either 'stable', or 'nightly'");
         }
+    } else if (key == "cloud_providers") {
```
fl0rianr, line 438: This makes cloud_providers pass validation, but it is not actually ignored after validation. For top-level non-backend keys, apply_changes() still writes the value into config_ /applied_diff, and /internal/set then persists the full snapshot back to config.json.

So an older admin UI that still sends cloud_providers can reintroduce this deprecated key into a v2 config. On the next load, config_migrate() will return early because `current_version >= target_version`, so the key will not be stripped again.

Could we make this a true no-op by erasing/skipping cloud_providers before apply_changes()? It may also be safer to strip this key unconditionally on load, not only during the v1→v2 migration.

src/cpp/include/lemon/config_file.h
```diff
@@ -8,6 +8,86 @@ namespace lemon {
...
+    if (config.contains("cloud_providers")) {
+        config.erase("cloud_providers");
+        changed = true;
+    }
+
+    // 2. Upgrade ctx_size 4096 -> -1 only when it is exactly the old default.
+    if (config.contains("ctx_size") && config["ctx_size"].is_number_integer()) {
+        int ctx_val = config["ctx_size"].get<int>();
+        if (ctx_val == 4096) {
+            config["ctx_size"] = -1;
```
fl0rianr, line 42: blocking unless this is intentionally stacked after #2235

test/cpp/test_config_migration.cpp
```diff
@@ -0,0 +1,405 @@
+// Standalone unit tests for lemon::config_get_version(),
+// lemon::config_migrate_v1_to_v2(), and lemon::config_migrate().
+//
+// Compile with:
+//   g++ -std=c++17 -I src/cpp/include -I src/cpp/build/_deps/json-src/single_include \
+//       test/cpp/test_config_migration.cpp -o config_migration_test
```
fl0rianr, line 6: Not planned to run with CI?

test/cpp/test_config_migration.cpp
```diff
@@ -0,0 +1,405 @@
...
+    test_migrate_already_at_target();
+    test_migrate_v0_via_merge();
+    test_migrate_v1_partial_fields();
+    test_migrate_v1_no_ctx_change();
+    test_migrate_no_version_field();
+    test_migrate_no_changes_when_at_target();
+    test_migrate_preserves_user_override();
+    test_round_trip();
+
+    int total = 1 + 6 + 10;
```
fl0rianr, line 402: Nit: the hard-coded `total = 1 + 6 + 10` below looks easy to let drift. Incrementing a total_checks counter inside check() would make the summary reliable.

fl0rianr (changes requested):
Good work, but I see some points that should (or could) be adapted.

bitgamma:
closing this as the changes have been cherry-picked to #2235

PR #2246, opened 2026-06-15
fix(windows): set safe cwd for backend child processes

### LINKED ISSUE #2218: Memory modification by the process on Windows.
### Platform

Windows

### Lemonade Version

10.7.0

### GPU / APU Model

Ryzen AI 7

### Component

llama.cpp

### Bug Description

After recent upgrade from version 10.6.0 to 10.7.0 the backends for llama.cpp and stablediffusion.cpp (Vulcan) were updated. Now attempts to use any of those backends results in the server process being blocked by the protected folders feature. This was not happening on the previous version, so would be good to know what changed, why those backends now trigger Windows security alert, and whether they could be fixed (without resorting to enable the process bypass of the security). Thanks.

### Steps to Reproduce

Run 10.7.0 on Windows 11, observe security notification and the complete failure to run for stablediffusion.cpp Vulcan, and the security notification for llama.cpp Vulcan (though model still responds then).

Expected: Not triggering Windows security.
Actual: Triggering Windows security and being blocked on an attempt to perform changes to the system.

### Log Output

```shell

```

### Additional Context

_No response_

fl0rianr:
## Summary

This is a small follow-up to #2184 for a separate Windows filesystem issue.
Fixes #2218.

#2184 addressed the HuggingFace cache symlink / create_directories path. This PR addresses the child-process side: llama-server and sd-server were started with an empty working directory, so on Windows they inherited Lemonade's current working directory.

For Vulkan backends, that can trigger Windows Controlled Folder Access / protected-folder blocks if the backend, Vulkan runtime, driver, or shader/pipeline cache performs relative writes.

## Changes

* Set an explicit Windows-only backend working directory before starting llama-server
* Set an explicit Windows-only backend working directory before starting sd-server
* Use absolute Windows backend runtime paths so relative cache/runtime paths keep working after the cwd change
* Keep the change intentionally small and local to backend process launch

### DIFF
--- a/src/cpp/server/backends/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp_server.cpp
@@ -560,7 +560,7 @@ void LlamaCppServer::load(const std::string& model_name,
     // Keep llama-server output visible at info log level.
     // Use a dedicated temp working directory so backend/Vulkan relative writes do
     // not inherit a protected launcher directory on Windows.
-    fs::path working_dir = fs::path(utils::get_runtime_dir()) / "lemonade" / "llamacpp" / llamacpp_backend;
+    fs::path working_dir = fs::temp_directory_path() / "lemonade" / "llamacpp" / llamacpp_backend;
     std::error_code working_dir_ec;
     fs::create_directories(working_dir, working_dir_ec);
     if (working_dir_ec) {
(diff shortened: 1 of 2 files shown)


superm1 (approved):

superm1 (approved):

fl0rianr:
Changed a bit since I had to overwhelm the CI which resulted in a more stable fix.

PR #2248, opened 2026-06-15
fix(server): handle Moonshine directory variants correctly

### LINKED ISSUE #2194: Fix moonshine model sizes
Image

Image

Why does Moonshine-Tiny show 35 MB in the model manager but 206 MB when I download it? Please double check the other two moonshine models as well.

fl0rianr:
## Summary

This fixes  #2194

Moonshine variants resolve to directories such as onnx/tiny, but parts of
the server-side model management logic treated variant paths like single
files. This caused two user-visible issues:

- After downloading Moonshine Tiny Streaming, the UI could show `Size N/A`
  until the server was restarted.
- Deleting a Moonshine variant could fail with a 500 error because the server
  attempted to remove a non-empty directory with fs::remove().

## Changes

- Calculate model size recursively when the resolved variant path is a directory.
- Delete directory-based model variants recursively.
- Update the configured sizes for the Moonshine Streaming variants:
  - Moonshine-Tiny-Streaming
  - Moonshine-Small-Streaming
  - Moonshine-Medium-Streaming

## Validation

- Checked the affected C++ changes for consistency with the existing
  ModelManager flow.
- local C++ build validation and test to verify all three bugs are gone

### DIFF
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -1322,7 +1322,7 @@
             "transcription",
             "realtime-transcription"
         ],
-        "size": 0.034
+        "size": 0.202
     },
     "Moonshine-Small-Streaming": {
         "checkpoint": "UsefulSensors/moonshine-streaming:onnx/small",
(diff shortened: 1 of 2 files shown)


Geramy (approved):
Looks correct.

PR #2250, opened 2026-06-15
test: pre-warm ROCm (TheRock) runtime before rocm-backend tests

jeremyfowers:
## Problem

The Windows `.exe` `stable-diffusion` and `llamacpp` rocm inference tests fail intermittently on a **cold TheRock cache**. The first rocm-stable model load downloads a ~4.5 GB TheRock tarball *inside the first inference request*, which can exceed the per-request inference timeout (`TIMEOUT_MODEL_OPERATION = 500s`) and surfaces as a confusing `ReadTimeout` rather than an install failure.

This is the cold-cache follow-up to the cache-out-of-workspace fix (#2148): that keeps the cache warm across runs, but a fresh runner / TheRock version bump / eviction still pays the cold download under the test's request timeout.

## Change

Add a shared `ensure_rocm_runtime()` helper, called from `ServerTestBase.setUpClass()`, that pre-installs the runtime via `POST /api/v1/install` with a generous `TIMEOUT_ROCM_INSTALL = 1800s`:

- **Gated** on `backend == "rocm"` and recipe ∈ `{llamacpp, sd-cpp}` — exactly the C++ `will_install_therock` condition. No-op for vulkan/cpu/npu, ryzenai/flm/whisper, rocm-nightly, and vLLM.
- **Idempotent** — the server skips the download when TheRock is already installed, so it's a fast check on a warm cache.
- **Distinct failure** — on non-2xx (500 install exception / 400 unsupported) or timeout, raises `ROCM_INSTALL_FAILED`, so a runtime/download problem reports as a setup error, not an inference test failure.

Moves the cold download out of the test body's 500s window into an explicit setup step with an appropriate budget.

### DIFF
--- a/test/utils/server_base.py
+++ b/test/utils/server_base.py
@@ -339,6 +339,7 @@ def ensure_rocm_runtime():
         response = requests.post(
             f"http://localhost:{PORT}/api/v1/install",
             json={"recipe": recipe, "backend": "rocm", "stream": False},
+            headers=_auth_headers(),
             timeout=TIMEOUT_ROCM_INSTALL,
         )
         response.raise_for_status()
--- a/test/utils/server_base.py
+++ b/test/utils/server_base.py
@@ -45,6 +45,7 @@ from .test_models import (
     STANDARD_MESSAGES,
     TIMEOUT_MODEL_OPERATION,
     TIMEOUT_DEFAULT,
+    TIMEOUT_ROCM_INSTALL,
     get_default_cli_binary,
 )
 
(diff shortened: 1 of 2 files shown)


fl0rianr (changes requested):
Thanks for the focused fix and finding this!

One thing that should be addressed before merging: the new /api/v1/install call should pass headers=_auth_headers(), like the existing setup helpers do. When LEMONADE_API_KEY is set, `/api/*` routes are authenticated, so this pre-warm step would currently fail with a 401 and surface as ROCM_INSTALL_FAILED before the tests even start.

Suggested change:

```python
response = requests.post(
    f"http://localhost:{PORT}/api/v1/install",
    json={"recipe": recipe, "backend": "rocm", "stream": False},
    headers=_auth_headers(),
    timeout=TIMEOUT_ROCM_INSTALL,
)

jeremyfowers:
Good catch @fl0rianr / @codex — fixed in cb6f83c0. The `/api/v1/install` pre-warm now passes `headers=_auth_headers()`, matching the existing setup helpers, so it won't 401 (and falsely report `ROCM_INSTALL_FAILED`) when `LEMONADE_API_KEY` is set.

fl0rianr (approved):
Nice! Approving.

PR #2254, opened 2026-06-15
Fix(cli): Lemonade backends now shows only supported backends . Added --all  option to show all backends.

GabrielReusRodriguez:
On lemonade cli , the `lemonade backends` now shows only the supported backends and recipes.

There is a new option  `lemonade backends --all` to show all available recipes and backends.

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -35,7 +35,7 @@ The `lemonade` CLI is the primary tool for interacting with Lemonade Server from
 |---------------------|-------------------------------------|
 | `status`            | Check if server can be reached. If it is, prints server information. Use `--json` for machine-readable output. |
 | `logs`              | Open server logs in the web UI. |
-| `backends`          | List available recipes and backends. Use `install` or `uninstall` to manage backends. |
+| `backends`          | List supported recipes and backends or list all available recipes and backends with `--all`. Use `install` or `uninstall` to manage backends. |
 | `cloud`             | Manage cloud OpenAI-compatible providers. See command options [below](#options-for-cloud). |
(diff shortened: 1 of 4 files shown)


jeremyfowers:
Should this be the default? As cross-vendor support grows the list is getting really long.

How would you feel about inverting the feature to be supported-only by default and add a `--all` to display both supported and unsupported?

GabrielReusRodriguez:
Hi!

I also think that it is better filtering the unsupported backends by default because i think the  hardware config does not change usually, so without filter,  you are seeing the same unsuppoted backends  again and again.

I did the changes. 

image

image

image

fl0rianr (commented):
Thanks for updating this to supported-only by default with `--all` for the full list. The direction looks good.

One edge case before approval: recipe.backends.empty() still prints an unsupported row even when show_all == false. Since the new default is meant to hide unsupported entries, recipes with no backend definitions should probably only be printed under `--all`.

Could you also update docs/guide/cli.md and the PR description to reflect the final behavior (`lemonade backends` = supported only, `lemonade backends --all` = full list)?

GabrielReusRodriguez:
Sure, let me review your feedback   and  I'll do the changes :)

GabrielReusRodriguez:
Hello again,

I pushed the changes. Let me know If more changes are necessary.

Thank you very much for your review :)

fl0rianr (approved):
Thanks for the update - LGTM from my side.

One small non-blocking nit: the PR description still mentions "--supported"; it would be nice to update that to the final behavior (`lemonade backends` by default, `lemonade backends --all` for the full list).

GabrielReusRodriguez:
Ouch!!

I changed the title but I forgot the description.

I have changed it,  and now you can see the new description without --supported.

PR #2256, opened 2026-06-15
Add Chatterbox text-to-speech backend

Geramy:
Integrates Resemble AI's [Chatterbox](https://github.com/resemble-ai/chatterbox) as a text-to-speech backend, served through the existing OpenAI-compatible `/v1/audio/speech` endpoint.

## Highlights
- **Multi-device:** CUDA, ROCm, Metal, and CPU — auto-selects GPU when available, falls back to CPU.
- **Byte-level streaming:** raw PCM16 @ 24 kHz via Chatterbox's `generate_stream` when available, with a single-chunk fallback otherwise.
- **Three variants:** English, Multilingual, and Turbo (registered in `server_models.json`).
- **Selective downloads:** only fetches the weight files each variant actually loads (~3 GB) instead of the full ~14 GB repo.

## Pieces
- `tools/chatterbox-server/main.py` — thin OpenAI-compatible HTTP wrapper around `chatterbox-tts`.
- `ChatterboxServer` C++ backend (`WrappedServer` + `ITextToSpeechServer`) with per-device install params.
- Recipe wiring: router, backend_utils, system_info (preference order), model_types, recipe_options, runtime_config, model_manager.
- `backend_versions.json` pins + docs page + minor UI display name/order.

Self-contained bundles are built by the companion [lemonade-sdk/chatterbox-rocm](https://github.com/lemonade-sdk/chatterbox-rocm) distribution repo (tracks `chatterbox-tts` PyPI releases).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -614,6 +614,7 @@ set(SOURCES_CORE
     src/cpp/server/backends/whisper_server.cpp
     src/cpp/server/backends/moonshine_server.cpp
     src/cpp/server/backends/kokoro_server.cpp
+    src/cpp/server/backends/chatterbox_server.cpp
     src/cpp/server/backends/sd_server.cpp
     src/cpp/server/backends/vllm_server.cpp
     src/cpp/server/backends/backend_utils.cpp
--- a/docs/guide/configuration/chatterbox.md
+++ b/docs/guide/configuration/chatterbox.md
@@ -1,0 +1,117 @@
+# Chatterbox Backend Options
+
+Lemonade integrates [Chatterbox](https://github.com/resemble-ai/chatterbox) (Resemble AI) as a **text-to-speech** backend, exposed through the OpenAI-compatible `/v1/audio/speech` endpoint (the same endpoint the Kokoro backend uses). Unlike Kokoro (CPU/Metal only), Chatterbox is a PyTorch model and supports **GPU acceleration** across vendors:
+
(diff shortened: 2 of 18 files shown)


jeremyfowers:
Scheduled for next release - please do not merge before 10.8 releases.

fl0rianr (changes requested):
Thanks for the integration — the overall shape looks good, and the latest push addressing split archives is useful. I still think this should not merge yet.

Main blockers:

1. CodeQL is flagging an uncontrolled path expression in tools/chatterbox-server/main.py, and I think this is a real issue. The HTTP request can provide audio_prompt_path, and voice is also interpreted as a local host path when os.path.isfile(voice) succeeds. Please avoid accepting arbitrary host paths from API requests. Prefer registered voice IDs, uploaded/temp files, or a strict allowlisted directory with canonical-path validation.

2. The advertised platform matrix does not match the generated asset names. system_info.cpp marks Chatterbox CPU as supported on Windows/Linux/macOS for both x86_64 and arm64, but get_install_params() always emits windows-x64, linux-x64, and macos-arm64. Please either constrain the support matrix or generate arch-correct asset names.

3. Chatterbox-Multilingual needs explicit handling for missing language_id. Upstream requires it, while the wrapper only forwards it when present. This should return a clean 400 or use a documented default language.

4. Please reconsider the production from_local() -> from_pretrained() fallback. It can hide selective-download bugs and unexpectedly perform network downloads during model load. Lemonade should fail fast if the pulled snapshot is incomplete.

5. Please make sure the backend includes the new watchdog logic as well like #2252

jeremyfowers:
RFC required!

PR #2257, opened 2026-06-15
Fix macOS auto-tune memory detection (unified memory)

jeremyfowers:
## Problem

On Apple Silicon, auto-tune resolved **every** model's context to the 4096 fallback (`resolve_auto_ctx_size: memory undetectable, returning 4096`) — e.g. a 4B model on a 24 GB MacBook got only 4k context. Yet `system-info` reported 24 GB correctly and the model-list RAM filter worked fine.

## Root cause

The working features and auto-tune use **different memory paths**, and only auto-tune's was unimplemented on macOS. Two gaps:

1. **No unified-memory accessor.** `get_available_memory_gb()` only consults the AMD/NVIDIA device accessors, which are stubs on macOS, so it fell through to `return 0.0` → 4096.
2. **`get_memory_usage_gb()` returned *total* RAM, not *used*.** So even with #1 fixed, `available = total − used = 0`. The comment admitted it: *"just report total memory since getting free memory is complex on macOS."*

## Fix

- **`system_info.h`** — add `virtual GPUInfo get_apple_silicon_device()` to the base `SystemInfo` (default reports "unavailable" so Windows/Linux need no stub), overridden in `MacOSSystemInfo`. Fits the existing per-platform device-accessor pattern.
- **`macos_system_info.mm`** — implement it: `vram_gb` = Metal's `recommendedMaxWorkingSetSize` (GPU working-set budget), `virtual_gb` = total `hw.memsize` (unified pool). Reuses the existing `detect_metal_gpus()`.
- **`auto_tune.h`** — Metal branch computes `min(GPU budget, free unified RAM)`; CPU/NPU branch uses total unified RAM. Both subtract used memory.
- **`metrics_macos.cpp`** — `get_memory_usage_gb()` now reports real used memory via mach `host_statistics64` (`active + wired + compressed` pages), matching the used-memory semantics already on Windows/Linux. This also corrects the `/stats` `memory_gb` field and the Prometheus `lemonade_memory_used_gb` gauge.

## Verification

Linked a driver against the full server core and ran on an M4 Pro / 24 GB machine:

```
apple_silicon_device: available=1 name="Apple M4 Pro" vram_gb=17.76 virtual_gb=24.00
get_available_memory_gb(GPU)=10.90   # was 0.00
get_available_memory_gb(CPU)=10.90   # was 0.00
```

Used ≈ 13 GB cross-checks against `vm_stat`. A 4B model now sees ~11 GB of headroom, so auto-tune resolves context well above the 4096 floor.

## Note

"Used" is approximated as `active + wired + compressed` (≈ Activity Monitor's "Memory Used"), treating inactive/purgeable as reclaimable. Including `inactive` for a more conservative count is a one-line change if preferred.

### DIFF
--- a/src/cpp/include/lemon/auto_tune.h
+++ b/src/cpp/include/lemon/auto_tune.h
@@ -133,12 +133,18 @@ inline double get_available_memory_gb(DeviceType device_type) {
             }
         }
 
-        // Metal (macOS — unified memory, reported via get_amd_igpu_device on macOS)
-        auto amd_igpu_mac = si->get_amd_igpu_device();
-        if (amd_igpu_mac.available && amd_igpu_mac.vram_gb > 0) {
-            double available = std::max(0.0, amd_igpu_mac.vram_gb - used_gb);
-            LOG(DEBUG, "AutoTune") << "get_available_memory_gb: GPU (Metal) total="
-                                   << std::fixed << std::setprecision(2) << amd_igpu_mac.vram_gb
+        // Metal (macOS — Apple Silicon unified memory). CPU and GPU share one pool:
+        //   vram_gb    = Metal's recommended GPU working-set budget (a soft ceiling)
+        //   virtual_gb = total unified RAM
+        // Available to the GPU = the free unified RAM (total − used), capped at the
+        // working-set budget so we don't push the system into swap.
+        auto apple = si->get_apple_silicon_device();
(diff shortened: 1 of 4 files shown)


bitgamma (approved):
thanks for this fix! Codex' comment about mach port leak seems to make sense, but otherwise this can be merged in the branch.

PR #2260, opened 2026-06-15
Fix copy-to-clipboard buttons silently failing in web-app over HTTP

### LINKED ISSUE #2259: Copy-to-clipboard buttons silently fail in web-app served over HTTP
### Platform

Linux/Ubuntu

### Lemonade Version

10.7.0

### GPU / APU Model

AMD Ryzen AI MAX+ 395

### Component

Desktop App / UI

### Bug Description

Copy-to-clipboard buttons in the web-app UI silently fail when the page is served via `lemond`'s HTTP server (non-HTTPS context). `navigator.clipboard.writeText()` requires a secure context (HTTPS or localhost) and throws without user-visible feedback when accessed over plain HTTP.

The root cause is that the `window.api` mock injected by `server.cpp` lacks a `writeClipboard` method. The utility function `clipboardUtils.ts` falls through to `navigator.clipboard.writeText()`, which fails in non-secure contexts. The three affected buttons are:

- "Copy model name" in Model Options modal
- "Copy code block" in chat messages
- "Copy action" in Backend manager

### Steps to Reproduce

1. Build and start `lemond` with the web-app enabled (default).
2. Open `http://<host>:<port>/app/` in a browser (plain HTTP, not HTTPS, not localhost).
3. Navigate to "Models", open a model's options, click "Copy model name".  
   -> Nothing happens, no visual feedback.
4. Similarly test code-block copy in chat and backend action copy.

**Expected:** The text is copied to the clipboard, and the button shows a checkmark for 2 seconds.

**Actual:** No visual feedback; error logged silently to console: `Failed to copy model name:` / `"Clipboard API not available"`.

### Log Output

```shell

```

### Additional Context

The `writeClipboard` mock was present in the Electron version (commit `7d8804d8`) but was not ported to the C++-injected mock when the app migrated from Electron to Tauri (`77220b9b`).

blackdeathdrow:
Closes #2259

**Problem:** Copy-to-clipboard buttons in the web-app silently fail when served via `lemond`'s HTTP server (non-HTTPS). `navigator.clipboard.writeText()` requires a secure context (HTTPS/localhost) and throws without user-visible feedback on plain HTTP.

The `window.api` mock injected by `server.cpp` lacked a `writeClipboard` method, so `clipboardUtils.ts` fell through to `navigator.clipboard.writeText()` - which fails in non-secure contexts. This affected all three copy buttons:

- "Copy model name" in Model Options modal
- "Copy code block" in chat messages  
- "Copy action" in Backend manager

**Fix (2 files):**

1. **`src/cpp/server/server.cpp`** - Added `writeClipboard` to the injected `window.api` mock. It tries the modern Clipboard API first, with a fallback to `document.execCommand('copy')` via a hidden `<textarea>` (best-effort fallback, no HTTPS required).

2. **`src/app/src/renderer/utils/clipboardUtils.ts`** - Added the same `execCommand` fallback for any context where `window.api?.writeClipboard` is unavailable (defense in depth).

**Root cause:** The `writeClipboard` mock existed in the old Electron version (commit `7d8804d8`) but was never ported to the C++ mock injected by `server.cpp` when the app migrated from Electron to Tauri (`77220b9b`).

### DIFF
--- a/src/app/src/renderer/utils/clipboardUtils.ts
+++ b/src/app/src/renderer/utils/clipboardUtils.ts
@@ -1,23 +1,31 @@
-function legacyCopy(text: string): Promise<void> {
+function legacyCopy(text: string): void {
   const ta = document.createElement('textarea');
-  ta.value = text;
+  ta.value = String(text);
+  ta.setAttribute('readonly', '');
   ta.style.position = 'fixed';
-  ta.style.opacity = '0';
-  ta.style.pointerEvents = 'none';
+  ta.style.left = '-9999px';
   document.body.appendChild(ta);
-  ta.select();
-  document.execCommand('copy');
-  document.body.removeChild(ta);
-  return Promise.resolve();
+  try {
+    ta.select();
+    if (!document.execCommand('copy')) {
+      throw new Error('Legacy clipboard copy failed');
+    }
+  } finally {
+    document.body.removeChild(ta);
+  }
 }
 
 export async function writeClipboard(text: string): Promise<void> {
   if (window.api?.writeClipboard) {
     return window.api.writeClipboard(text);
   }
-  try {
-    return await navigator.clipboard.writeText(text);
-  } catch {
-    legacyCopy(text);
+  if (navigator.clipboard) {
+    try {
+      await navigator.clipboard.writeText(text);
+      return;
+    } catch {
+      // Ignore clipboard errors and fall back to legacy method
+    }
   }
+  legacyCopy(text);
 }
--- a/src/app/src/renderer/utils/clipboardUtils.ts
+++ b/src/app/src/renderer/utils/clipboardUtils.ts
@@ -1,7 +1,23 @@
+function legacyCopy(text: string): Promise<void> {
+  const ta = document.createElement('textarea');
+  ta.value = text;
+  ta.style.position = 'fixed';
+  ta.style.opacity = '0';
+  ta.style.pointerEvents = 'none';
+  document.body.appendChild(ta);
+  ta.select();
+  document.execCommand('copy');
+  document.body.removeChild(ta);
+  return Promise.resolve();
+}
+
 export async function writeClipboard(text: string): Promise<void> {
   if (window.api?.writeClipboard) {
-    await window.api.writeClipboard(text);
-  } else {
-    await navigator.clipboard.writeText(text);
+    return window.api.writeClipboard(text);
+  }
+  try {
+    return await navigator.clipboard.writeText(text);
+  } catch {
+    legacyCopy(text);
   }
 }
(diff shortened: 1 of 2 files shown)


src/app/src/renderer/utils/clipboardUtils.ts
fl0rianr, line 1: This fallback should check the return value of document.execCommand('copy').

Right now legacyCopy() resolves successfully even if the browser returns false, which means the command was unsupported or disabled. That would make the UI show the copied/checkmark state even though nothing was copied — effectively preserving the same class of silent/false-success clipboard bug this PR is trying to fix.

Could we make the fallback throw when execCommand('copy') returns false, and keep the textarea cleanup in a `finally` block?

Something like:

```ts
function legacyCopy(text: string): void {
  const ta = document.createElement('textarea');
  ta.value = String(text);
  ta.setAttribute('readonly', '');
  ta.style.position = 'fixed';
  ta.style.left = '-9999px';

  document.body.appendChild(ta);
  try {
    ta.focus();
    ta.select();

    if (!document.execCommand('copy')) {
      throw new Error('Legacy clipboard copy failed');
    }
  } finally {
    document.body.removeChild(ta);
  }
}
```
That way callers only show success feedback when one of the copy paths actually reports success.

src/cpp/server/server.cpp
fl0rianr, line 1: Same concern here as in clipboardUtils.ts: the injected web mock should not resolve successfully when the legacy copy path failed.

document.execCommand('copy') returns `false` when the command is unsupported or disabled, so ignoring that return value can make window.api.writeClipboard() report success while nothing was copied. Since the renderer relies on this promise to decide whether to show success feedback, please check the boolean result and throw on failure.

Also please move removeChild() into a `finally` block so a thrown copy failure does not leave the hidden textarea in the DOM.

fl0rianr (changes requested):
Nice you are working on this and the change is constrained.

I would avoid saying that the execCommand('copy') fallback “works in all browsers, no HTTPS required.” It is deprecated/non-standard and browser behavior around clipboard commands varies.

Maybe phrase this as a best-effort fallback for environments where legacy copy is still supported, and include the browsers/contexts that were manually tested.

Apart from that nitty point found something more blocking:

src/app/src/renderer/utils/clipboardUtils.ts
blackdeathdrow: You're right, thanks for the notes! I've fixed the fallback by checking the execCommand return value and adding the finally block as suggested. I also included a check for navigator.clipboard.

src/cpp/server/server.cpp
blackdeathdrow: Fixed this one too.

blackdeathdrow:
Tested over a local network — the fallback triggers and copies the text in Chrome, Edge, and Firefox.

fl0rianr (approved):
Thanks for the adaption. With current main code me might have a change to run through CI... let's see

PR #2276, opened 2026-06-16
ci: add repo-manager workflow

jeremyfowers:
## Summary

- Adds `.github/workflows/repo-manager.yml` to run `repo-manager all` automatically
- Triggers on push to `main` or any `release-v*` branch, plus manual dispatch
- Targets the dedicated `[self-hosted, linux, repo-manager]` runner where `repo-manager` and Pi are pre-installed
- Permissions scoped to read-only (`contents`, `pull-requests`, `issues`) — sufficient for `gh` CLI analysis

## Pre-merge setup

Before merging, make `repo-manager` available on the runner's PATH (one-time):
```bash
# Option A: symlink
ln -s /path/to/repo-manager-venv/bin/repo-manager ~/.local/bin/repo-manager

# Option B: add venv bin to PATH in ~/.bashrc / ~/.profile
export PATH="/path/to/repo-manager-venv/bin:$PATH"
```

## Test plan

- [ ] Complete runner PATH setup above
- [ ] Push a commit to `main` → "Repo Manager" job appears and succeeds in Actions
- [ ] Use "Run workflow" button → manual dispatch works
- [ ] Push to a `release-v*` branch → workflow triggers

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1020,47 +1020,47 @@ jobs:
             script: server_llm.py
             extra_args: "--wrapped-server llamacpp"
             backends: "vulkan rocm"
-            runner: [Windows, vulkan, rocm]
+            runner: [Windows, vulkan, rocm, lemon-prod]
           - name: omni
             script: server_omni.py
             extra_args: "--wrapped-server llamacpp"
             backends: "vulkan"
-            runner: [Windows, vulkan, stx-halo]
+            runner: [Windows, vulkan, stx-halo, lemon-prod]
           - name: ryzenai
             script: server_llm.py
             extra_args: "--wrapped-server ryzenai"
             backends: "cpu hybrid npu"
-            runner: [Windows, xdna2]
+            runner: [Windows, xdna2, lemon-prod]
           - name: flm
             script: server_llm.py
             extra_args: "--wrapped-server flm"
             backends: "npu"
-            runner: [Windows, xdna2]
+            runner: [Windows, xdna2, lemon-prod]
           - name: whisper
             script: server_whisper.py
             extra_args: "--wrapped-server whispercpp"
             backends: "cpu npu rocm"
-            runner: [Windows, stx-halo]
+            runner: [Windows, stx-halo, lemon-prod]
           - name: flm-whisper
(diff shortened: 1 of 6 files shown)


ramkrishna2910 (approved):

fl0rianr (commented):
Thanks for the PR — the direction makes sense..

A few things to look into before merge:

1. The workflow checks out this repo, but then runs `repo-manager all` from /opt/lemonade-manager. Is that directory intentionally the repo-manager installation/config that targets lemonade-sdk/lemonade? If yes, maybe add a short comment explaining that. If the command is meant to analyze the checked-out repo, it should probably run from ${{ github.workspace }} or pass the target path explicitly.

2. Since this uses a dedicated self-hosted runner, please add a bounded timeout-minutes and consider concurrency with cancel-in-progress: true, so repeated pushes or a hung repo-manager run do not block the runner indefinitely.

3. The new docs say every self-hosted job must include lemon-prod, but this new repo-manager job is intentionally self-hosted without lemon-prod. Could you adjust the wording to say that every *production* self-hosted job must include lemon-prod, while special-purpose runners may use their own dedicated labels such as repo-manager?

jeremyfowers:
> Thanks for the PR — the direction makes sense..
> 
> A few things to look into before merge:
> 
>     1. The workflow checks out this repo, but then runs `repo-manager all` from /opt/lemonade-manager. Is that directory intentionally the repo-manager installation/config that targets lemonade-sdk/lemonade? If yes, maybe add a short comment explaining that. If the command is meant to analyze the checked-out repo, it should probably run from ${{ github.workspace }} or pass the target path explicitly.
> 
>     2. Since this uses a dedicated self-hosted runner, please add a bounded timeout-minutes and consider concurrency with cancel-in-progress: true, so repeated pushes or a hung repo-manager run do not block the runner indefinitely.
> 
>     3. The new docs say every self-hosted job must include lemon-prod, but this new repo-manager job is intentionally self-hosted without lemon-prod. Could you adjust the wording to say that every _production_ self-hosted job must include lemon-prod, while special-purpose runners may use their own dedicated labels such as repo-manager?

Thanks, moved this feedback to https://github.com/lemonade-sdk/lemonade/issues/2277

I had to merge some version of the workflow so that I could debug it.

PR #2278, opened 2026-06-16
fix(test): require Whisper model load before language transcription test

fl0rianr:
## Summary

This PR makes the Whisper transcription tests consistent when running with an explicit whisper.cpp backend such as Metal.

Previously, test_001_transcription_basic`explicitly loaded the configured Whisper model/backend before sending the transcription request, while test_002_transcription_with_language sent the request directly. When the backend model could not start, the first test handled the load step separately, but the second test surfaced the same underlying issue as a later /audio/transcriptions 500.

This change moves the model/backend load into a small shared helper and uses it in both positive HTTP transcription tests.

## Why

When a backend is explicitly requested, especially in the macOS Metal CI job, failure to load the Whisper model is a legitimate test failure. Many downstream transcription checks cannot pass if the backend process never starts or becomes ready.

Failing early on the model load gives a clearer error and avoids inconsistent behavior between the basic transcription and language transcription tests.

### DIFF
--- a/test/server_whisper.py
+++ b/test/server_whisper.py
@@ -98,14 +98,8 @@ class WhisperTests(ServerTestBase):
                 pass  # Ignore cleanup errors
         super().tearDownClass()
 
-    def test_001_transcription_basic(self):
-        """Test basic audio transcription with Whisper."""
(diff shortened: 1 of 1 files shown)


jeremyfowers (approved):
LGTM!

PR #2279, opened 2026-06-16
fix(server): fall back to installed llama.cpp binary when "latest" release lookup fails

### LINKED ISSUE #2265: Lemonade fails to load (models) when latest llamacpp is unavailable
### Platform

Windows

### Lemonade Version

10.7.0

### GPU / APU Model

Strix Halo

### Component

llama.cpp

### Bug Description

When a user has set `llamacpp.vulkan_bin=latest`, and there's an issue with fetching the latest llamacpp version of GitHub ("GitHub returned HTTP 504 when querying latest release of ggml-org/llama.cpp"), then models using llamacpp will fail to load ("Failed to load model", "model_load_error")

### Steps to Reproduce

1. lemonade config set llamacpp.vulkan_bin=latest
2. wait until GitHub returns an error ;) or interrupt the check
3. load a model that uses llamacpp.
4. observe the failure

Expected: Lemonade 'falling back' to the 'current' (== most recent before re-starting lemonade) llamacpp binary and loading the model.
Actual: Lemonade refusing to load models using llamacpp

### Log Output

```shell

```

### Additional Context

_No response_

ianbmacdonald:
## Summary
Fixes #2265.

When a backend's `*_bin` is pinned to `latest`, loading a model resolves the newest release tag from GitHub. If that lookup fails (e.g. HTTP 504) or is otherwise unreachable, resolution **threw** and the model refused to load — even though a perfectly usable binary was already installed.

This makes the live-failure path behave like the existing **offline** path: fall back to the version recorded in the installed binary's `version.txt`, load with a warning, and only error when *neither* a live tag *nor* an installed binary is available.

## Root cause
`LlamaCppServer::load()` → `install_backend()` → `get_install_params()` → `resolve_user_version()`. For `*_bin="latest"`, that last call did `fetch_latest_github_tag(repo, throw_on_failure=true)`; a transient GitHub failure propagated up as `model_load_error` before the already-installed binary was ever considered.

## Fix
- In the `"latest"` branch, do a **non-throwing** lookup; on an empty result fall back to the installed `version.txt`. The install dir isn't version-scoped, so it always reflects the on-disk binary — `install_from_github` then sees a version match and skips re-download, so the fallback needs no network.
- The selection rule is extracted into a pure, I/O-free helper `lemon::resolve_latest_pin()` so it's unit-testable in isolation.

**Behavioral scope:** `resolve_user_version()` is shared, so this fallback now applies to **every `*_bin=latest` backend** (llamacpp, whisper, sd, kokoro, …), not just `llamacpp:vulkan`. The failure mode is generic to `latest`, so this is intended. Pinned / `builtin` / path / bare-tag values and unrelated load errors are unchanged.

## Testing
- **New C++ unit test** (`test/cpp/test_latest_version_fallback.cpp`) covering the full decision table (online ok / online failed / offline × with/without an installed binary), wired into CMake + CI alongside `test_directory_watcher`. It uses explicit pass/fail accounting rather than `assert()` (which the Release build's `-DNDEBUG` would compile to a no-op) so it actually guards in CI.
- **Verified on real hardware** (Radeon RX 7900 XT, gfx1100, Vulkan): forced the GitHub `latest` lookup to fail with an already-installed `b9632` binary.
  - Before: `model_load_error: Failed to query GitHub for latest release of ggml-org/llama.cpp`.
  - After: `could not resolve 'latest' for llamacpp:vulkan; falling back to installed version b9632` → model loads and serves a completion.

_Touches `CMakeLists.txt` and the two CI `Run C++ unit tests` steps purely to gate the new regression test._

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -275,8 +275,9 @@ jobs:
           set -e
           cmake --preset default
           cmake --build --preset default --target test_directory_watcher
+          cmake --build --preset default --target test_latest_version_fallback
           cd build
-          ctest --output-on-failure -R DirectoryWatcher
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback"
 
       - name: Upload .deb package
         uses: actions/upload-artifact@v7
@@ -534,8 +535,9 @@ jobs:
         shell: bash
         run: |
           cmake --build --preset default --target test_directory_watcher
+          cmake --build --preset default --target test_latest_version_fallback
           cd build
-          ctest --output-on-failure -R DirectoryWatcher
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback"
 
       - name: Upload .pkg package
         if: steps.check_signing.outputs.has_signing == 'true'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1720,3 +1720,23 @@ if(EXISTS "${_GGUF_CAPS_TEST_SRC}")
     include(CTest)
     add_test(NAME GgufCapabilitiesTest COMMAND test_gguf_capabilities)
 endif()
+
+# Pure version-resolution policy test (lemon::resolve_latest_pin). Header-only
+# logic, so the test links nothing beyond the standard library — regression
+# coverage for #2265 (fall back to the installed binary when the GitHub
+# "latest" lookup fails).
+set(_LATEST_FALLBACK_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_latest_version_fallback.cpp"
+)
+if(EXISTS "${_LATEST_FALLBACK_TEST_SRC}")
+    add_executable(test_latest_version_fallback
+        test/cpp/test_latest_version_fallback.cpp
+    )
+    target_include_directories(test_latest_version_fallback PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+
+    include(CTest)
+    add_test(NAME LatestVersionFallbackTest COMMAND test_latest_version_fallback)
+endif()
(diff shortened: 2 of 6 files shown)


ianbmacdonald:
looks good.. first soup to nuts harness run; The harness picked the bug, reproduced it, claimed it, fixed it, tested it, reviewed it and should ping reviewers once/if it clears CI .. pretty sweet for little mechanical fixes ;  inner dev harness, completely toon based, no md files.

fl0rianr (approved):
Important fix - Looks good to me. The change addresses #2265 in the right place: `latest` resolution now treats transient GitHub lookup failures like the existing offline fallback and reuses the installed version.txt when available.

Non-blocking nit: backend_version_policy.h mentions no_fetch_executables as one reason for an empty fetched_latest, but the normal install_backend() path still throws on no_fetch_executables() before reaching get_install_params() / esolve_user_version() . If that behavior is intentional, I’d slightly narrow the comment to avoid implying that no_fetch_executables is covered by this fallback in the install path.

ckuethe:
Something worth keeping in mind is slow or unreliable downloads - on an airplane, for example. I think it would make sense to not remove and replace the llama binary until the download is verified to be complete.

bitgamma (approved):
looks good to me. The comment from @ckuethe about removing the existing version only after the new one has been downloaded successfully makes a lot sense, but up to you to decide if this is in scope of this PR or an eventual followup.

ianbmacdonald:
> looks good to me. The comment from @ckuethe about removing the existing version only after the new one has been downloaded successfully makes a lot sense, but up to you to decide if this is in scope of this PR or an eventual followup.

agreed, I was going to let the harness bolt this on post-merge .. scout already picked it up and backlogged it here .. but its guidance is 'once you get approval, don't mess with the PR' ; I can punch it out tonight in a separate PR

bitgamma:
@ianbmacdonald works for me, there is just one merge conflict to solve and I'll merge :)

ckuethe:
yeah, I didn't mean to delay merging, more thinking of a future improvement.

ianbmacdonald:
Rebased onto current \`main\` and resolved the conflict — it was the C++ unit-test block in \`CMakeLists.txt\` colliding with the new \`test_gguf_capabilities\` target from #2281 (both appended a test after \`DirectoryWatcher\`). Kept both: \`GgufCapabilitiesTest\` and \`LatestVersionFallbackTest\` now register side by side, and all three C++ unit tests pass locally (\`ctest\`: 3/3). No code changes — resolution was build-config only. Thanks @bitgamma @fl0rianr.

PR #2281, opened 2026-06-16
test(gguf): add unit tests for MTP / capability label detection (#2176)

ramkrishna2910:
## Problem

PR #2176 added auto-detection of the **MTP (Multi-Token Prediction)** capability for GGUF models — it scans KV metadata for `nextn_predict_layers` and llama.cpp MTP/MAD tensor names, then applies an `mtp` model label so llama-server uses draft-mtp speculation. It shipped with **no automated tests**, so the detection/labeling logic is unguarded against regressions (the open maintainer to-do from the v10.8.0 review).

## Change

Add `test/cpp/test_gguf_capabilities.cpp` and register it with CTest as `GgufCapabilitiesTest`, mirroring the existing `test_directory_watcher` wiring. `gguf_capabilities.h` is header-only, so the test compiles against just the header (no extra sources).

### Coverage (20 cases, all passing)
- `apply_gguf_capability_labels`: `"mtp"` added **iff** `caps.mtp`, with the correct `changed` return.
- `add_label_once` dedup + idempotency: re-applying is a no-op and never duplicates the label.
- Pre-existing labels preserved; `mtp` correctly combines with `vision` / `tool-calling`.
- `inspect_gguf_string` **never** infers `mtp` from a metadata string (mtp must originate from the upstream KV/tensor scan, not a heuristic).
- The sibling `vision` / `tool-calling` detection in the same header is covered too (incl. the non-stable-key negative case).

### Scope note
The raw MTP *trigger* (the KV `nextn_predict_layers` check + tensor-name matching) lives in `model_manager.cpp` and requires a parsed GGUF file, so it's out of scope for a header-level unit test. This test guards the header's capability→label contract, which is the regression-prone seam that PR #2176 introduced.

## Verification
Built and ran locally (`cmake --build --preset vs18 --target test_gguf_capabilities`):
```
All gguf_capabilities tests passed   (20/20, exit 0)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1701,3 +1701,22 @@ if(EXISTS "${_DIR_WATCHER_TEST_SRC}" AND EXISTS "${_DIR_WATCHER_LIB_SRC}")
     include(CTest)
     add_test(NAME DirectoryWatcherTest COMMAND test_directory_watcher)
 endif()
+
+# GGUF capability detection (MTP / vision / tool-calling label logic).
+# Header-only unit under test, so no extra source files are required.
+set(_GGUF_CAPS_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_gguf_capabilities.cpp"
+)
+if(EXISTS "${_GGUF_CAPS_TEST_SRC}")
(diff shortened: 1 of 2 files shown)


fl0rianr (approved):
fine by me.

PR #2284, opened 2026-06-17
docs: system-stats API

jeremyfowers:
## Summary

Adds `GET /v1/system-stats` to the Lemonade API reference.

## Details

- Adds the endpoint to the Lemonade-specific endpoint table.
- Documents the no-parameter request shape.
- Adds an example response covering CPU, RAM, GPU, VRAM, and NPU telemetry fields.
- Notes that unsupported telemetry values are returned as `null`.

## Validation

- Ran `git diff --check -- docs/api/lemonade.md`.
- Commit hooks passed: trailing whitespace, end-of-file fixer, YAML check, large-file check, and Black skipped for non-Python files.


fl0rianr (approved):
Nice, good work. Approving.

One nit:  „as measured by the Lemonade Server process“ sounds like we are doing something special here. But in fact we are using host metrics for RAM, GPU/VRAM/NPU if available. Besides that we already state  „Current host resource usage“.
Maybe change this one?
reported by the Lemonade Server --> observed from the Lemonade Server host

superm1 (approved):

PR #2291, opened 2026-06-17
Fix FLM runtime version check false negative due to missing XRT lib

jpnski:
FLM runtime version detection regression was somewhere around v10.2 / #1475, I believe...

This fixes the issue when testing locally on Fedora43, FLM v0.9.43, building ` lemonade-server-10.8.0.x86_64.rpm`, enabling the unit service, and checking flm backend status at the localhost port 13305

### DIFF
--- a/data/lemond-user.service.in
+++ b/data/lemond-user.service.in
@@ -7,6 +7,7 @@ After=network-online.target
 Type=simple
 WorkingDirectory=%h
 EnvironmentFile=-%E/lemonade/conf.d/*.conf
+Environment=LD_LIBRARY_PATH=/opt/xilinx/xrt/lib:/opt/xilinx/xrt/lib64
 ExecStart=@CMAKE_INSTALL_FULL_BINDIR@/lemond
 ExecReload=/bin/kill -HUP $MAINPID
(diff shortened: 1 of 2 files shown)


jpnski:
This is not necessary, and a work around. Building from source to install XRT libs on Fedora results in the dependency libs being in `/opt/xilinx/xrt/lib` whereas with `apt install libxrt-npu2` they land in `/usr/lib/` - where ldconfig indexes that standard lib path, but it does not index the fedora path, so the libs aren't available to lemond without manually exporting.
```
# Fedora fix:
1. echo /opt/xilinx/xrt/lib | sudo tee /etc/ld.so.conf.d/xrt.conf
2. echo /opt/xilinx/xrt/lib64 | sudo tee -a /etc/ld.so.conf.d/xrt.conf

# Fedora install:
1. sudo dnf install lemonade-server-10.7.0-fc43.x86_64.rpm
2. systemctl enable --now lemond

# Inspect, even in bare env, calling the binary is successful
$ env -i PATH=/usr/local/bin:/usr/bin:/bin flm version --json
  { "version": "0.9.43" }
```

PR #2273, opened 2026-06-16
Fix Problem with  .devcontainer. It did not copy .devcontainer/reinstall-cmake.sh when building container

GabrielReusRodriguez:
Hello,

I saw the lemonade repo had a .devcontainer so  I tryed the visual code devcontainer feature on my ubuntu 24.04  , visual code 1.124.2.

I tryed but i saw visual code throw an error when it tryed to build container

image

It failed when it tryed to copy .devcontainer/reinstall-cmake.sh . I investigate and i saw that inside .dockerignore there is a 
.devcontainer line that  is guilty :)

I added below :

!.devcontainer/reinstall-cmake.sh 

so it will let to copy the script . I tryed and finally worked!

image

PR #2293, opened 2026-06-17
Fix ROCm whisper-server startup: add TheRock lib dir to LD_LIBRARY_PATH

### LINKED ISSUE #2292: ROCm whisper-server aborts at startup (exit 134, libamd_comgr.so.3): WhisperServer omits TheRock lib dir from LD_LIBRARY_PATH
## Summary

On Linux, `lemond` launches the ROCm `whisper-server` without the bundled TheRock ROCm library directory on `LD_LIBRARY_PATH`. The binary aborts immediately (exit 134) trying to `dlopen` `libamd_comgr.so.3`, so transcription fails entirely. The `llama.cpp` and `sd-cpp` backends do add this path; `whisper` does not. With `whispercpp.backend: auto` resolving to `rocm` on a supported AMD GPU, transcription is dead out of the box.

## Environment

- Lemonade Server v10.8.0 (official Docker image, `ghcr.io/lemonade-sdk/lemonade-server:latest`)
- AMD Ryzen AI MAX+ 395 / Radeon 8060S, **gfx1151** (Strix Halo), 128 GB unified
- TheRock ROCm bundle: `gfx1151-7.13.0`

## Repro

1. On a gfx1151 host, leave `whispercpp.backend: auto` (resolves to `rocm`).
2. `POST /api/v1/load {"model_name":"Whisper-Base"}`.
3. Returns HTTP 500 `model_load_error: whisper-server failed to start or become ready`.

Container log:

```
(WhisperServer) Using backend: rocm
(WrappedServer) whisper-server process has terminated with exit code: 134
```

Running the bundled binary by hand:

```
$ /root/.cache/lemonade/bin/whispercpp/rocm/whisper-server -m ggml-base.bin
implib-gen: libamd_comgr.so.3: failed to load library 'libamd_comgr.so.3' via callback 'amd_comgr_stub_dlopen'
Aborted (core dumped)
```

`libamd_comgr.so.3` exists at `/root/.cache/lemonade/bin/therock/gfx1151-7.13.0/lib/`. Adding that dir to `LD_LIBRARY_PATH` makes the same binary boot and return `{"status":"ok"}`.

## Root cause

`src/cpp/server/backends/whisper_server.cpp` (the `#ifndef _WIN32` env block in `WhisperServer::load`) sets `LD_LIBRARY_PATH` to only the executable dir. It never calls `BackendUtils::get_therock_lib_path(rocm_arch)` -- unlike `llamacpp_server.cpp` and `sd_server.cpp`, which both prepend it for the ROCm backend. So the ROCm whisper-server can't find `libamd_comgr.so.3`.

## Fix

Mirror the other two backends in `whisper_server.cpp`, gated on the rocm backend. PR to follow.

```cpp
    std::string lib_path = exe_dir.string();

    // ROCm whisper-server needs the TheRock ROCm libs (libamd_comgr.so.3, etc.)
    // on LD_LIBRARY_PATH, exactly as llamacpp_server.cpp and sd_server.cpp do.
    // Without this it aborts at startup (dlopen libamd_comgr.so.3) on gfx1151.
    if (whispercpp_backend == "rocm") {
        if (!rocm_arch.empty()) {
            std::string therock_lib = BackendUtils::get_therock_lib_path(rocm_arch);
            if (!therock_lib.empty()) {
                lib_path = therock_lib + ":" + lib_path;
            }
        }
    }
```

Verified: built `lemond` with this change and ran it on a gfx1151 host with `whispercpp.backend=rocm` -- the rocm whisper-server now boots (`Using backend: rocm` -> `whisper-server is ready!`), loads, and transcribes. On the same bundled binary, exe-dir-only `LD_LIBRARY_PATH` aborts on `libamd_comgr.so.3` while the patched path finds the GPU.

## Workaround

Set `whispercpp.backend: vulkan`. The Vulkan whisper-server needs none of the TheRock libs and runs fine.

matthewjhunter:
Fixes #2292.

`WhisperServer` was the only ROCm backend launcher not prepending the TheRock ROCm library directory to `LD_LIBRARY_PATH`, so the rocm `whisper-server` aborted at startup (exit 134) failing to `dlopen libamd_comgr.so.3`. This mirrors the existing logic in `sd_server.cpp` and `llamacpp_server.cpp`, gated on the rocm backend.

```cpp
// src/cpp/server/backends/whisper_server.cpp, in WhisperServer::load (#ifndef _WIN32 block)
    std::string lib_path = exe_dir.string();

    // ROCm whisper-server needs the TheRock ROCm libs (libamd_comgr.so.3, etc.)
    // on LD_LIBRARY_PATH, exactly as llamacpp_server.cpp and sd_server.cpp do.
    // Without this it aborts at startup (dlopen libamd_comgr.so.3) on gfx1151.
    if (whispercpp_backend == "rocm") {
        if (!rocm_arch.empty()) {
            std::string therock_lib = BackendUtils::get_therock_lib_path(rocm_arch);
            if (!therock_lib.empty()) {
                lib_path = therock_lib + ":" + lib_path;
            }
        }
    }
```

Note: the gate uses the raw `whispercpp_backend == "rocm"` option value, not the `resolved_backend == "rocm-stable"` form used by llamacpp/sd_server. This is intentional -- whisper has no resolve step and uses the literal `"rocm"` string throughout (see the download path in `get_install_params`).

## Testing

Built `lemond` from the v10.8.0 tag (the patched file is byte-identical on `main`) in `build-environment:ubuntu24.04`, then ran it on a gfx1151 (Radeon 8060S, Strix Halo) host with `whispercpp.backend=rocm`:

- The rocm `whisper-server` now boots: `(WhisperServer) Using backend: rocm` -> `whisper-server is ready!`
- `POST /api/v1/load {"model_name":"Whisper-Base"}` -> `status: success`
- `POST /api/v1/audio/transcriptions` -> returns transcribed text

Causation control on the same bundled rocm `whisper-server` binary:

- exe-dir-only `LD_LIBRARY_PATH` (pre-patch): `implib-gen: libamd_comgr.so.3: failed to load library ... Aborted (core dumped)`
- TheRock dir prepended (this patch): `found 1 ROCm devices ... Radeon 8060S Graphics, gfx1151`, then runs.

### DIFF
--- a/src/cpp/server/backends/whisper_server.cpp
+++ b/src/cpp/server/backends/whisper_server.cpp
@@ -301,6 +301,19 @@ void WhisperServer::load(const std::string& model_name,
     // set LD_LIBRARY_PATH to include executable directory
     std::string lib_path = exe_dir.string();
 
+    // ROCm whisper-server needs the TheRock ROCm libs (libamd_comgr.so.3, etc.)
+    // on LD_LIBRARY_PATH, exactly as llamacpp_server.cpp and sd_server.cpp do.
+    // Without this it aborts at startup (dlopen libamd_comgr.so.3) on gfx1151.
+    if (whispercpp_backend == "rocm") {
+        std::string rocm_arch = SystemInfo::get_rocm_arch();
+        if (!rocm_arch.empty()) {
+            std::string therock_lib = BackendUtils::get_therock_lib_path(rocm_arch);
+            if (!therock_lib.empty()) {
+                lib_path = therock_lib + ":" + lib_path;
+            }
+        }
+    }
+
     const char* existing_ld_path = std::getenv("LD_LIBRARY_PATH");
     if (existing_ld_path && strlen(existing_ld_path) > 0) {
         lib_path = lib_path + ":" + std::string(existing_ld_path);


fl0rianr (approved):
LGTM from code review, Thanks! This patch is narrowly scoped to ROCm whisper-server startup and mirrors the existing TheRock LD_LIBRARY_PATH handling used by the other ROCm backends.

The current CI failure IS unrelated to this PR, but sadly prevents me from merging until we have that fixed...

iswaryaalex:
Thanks for the fix, are we missing any tests on lemond to catch this is CI ?

fl0rianr:
Sadly my fix did not turn out as successful as I intended it to be. CI is still to brittle and fails in some cases with "no reason".

matthewjhunter:
The cause is more testable than the symptom. The ROCm `LD_LIBRARY_PATH` setup is copy-pasted inline into `llamacpp_server.cpp`, `sd_server.cpp`, and `whisper_server.cpp`. Two of them prepend the TheRock lib dir; whisper had drifted and didn't. Nothing structurally ties the three together, so the next backend can make the same omission.

I'd suggest:

1. Factor the duplicated block into one pure helper in `backend_utils` -- something like `build_rocm_ld_library_path(exe_dir, prepend_therock, therock_lib, existing_ld_path)`. The gate decision stays in the caller and is passed in as the bool, because the backends don't agree on how they express it: llamacpp/sd key off `resolved_backend == "rocm-stable"`, but whisper has no resolve step at all, so it keys off the raw `whispercpp_backend == "rocm"` option (the same literal its download path in `get_install_params` uses). The helper shouldn't assume a `resolved_backend` exists.
2. Add a `test/cpp/` unit test for that helper in the existing assert style -- no GPU needed. Cases: with a non-empty therock lib the result starts with `<therock>:`; with the gate off it doesn't contain the therock path; empty therock lib falls back to exe-dir only with no stray colon; an existing `LD_LIBRARY_PATH` is appended in order. The pre-patch whisper behavior fails that first assertion.

That makes the regression catchable on any CI box and removes the copy-paste that caused it in the first place.

What would catch this *exactly* is launching the rocm `whisper-server` and hitting `/api/v1/load` on a gfx-class GPU, but that's the same coverage gap every backend has today: CI builds the ROCm binaries but never starts them, and there's no ROCm runner in the matrix. Adding a hardware smoke test just for whisper would be the flakiest, most expensive option, so I don't think that's the right call on its own.

Caveat: I've only glanced at this code, so weigh the specifics accordingly -- the duplication and the refactor are clear-cut, but whoever owns the backends should sanity-check the helper's signature against the other call sites before collapsing them.

Happy to put the helper + test up as a separate follow-up PR if you'd like -- I'd keep it out of this one so the fix stays a clean cherry-pick.

fl0rianr:
@matthewjhunter thanks for looking at it. I would really say it's unrelated since we have this issue with multiple PRs including this tiny one which does not affect CI nor source code #2273.
No action on your side needed, I'm working an another fix trial.

Issue #2290, opened 2026-06-17
Prevent the Rosetta prompt on macOS installation


jeremyfowers:
Context: https://www.reddit.com/r/LocalLLaMA/s/Hvo6SSlr2z

PR #2297, opened 2026-06-18
fix: Prevent the Rosetta prompt on macOS installation

clemperorpenguin:
This should solve #2290, at least in theory, I can't know for sure as my CI is failing on the whisper metal inference test for some reason. Creating as draft for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1325,6 +1325,10 @@ if(APPLE)
     set(CPACK_PACKAGE_NAME "Lemonade")
     set(CPACK_PRODUCTBUILD_IDENTIFIER "com.lemonade.server")
 
+    # Declare the package arm64-only so the macOS Installer does not assume our
+    # postflight script needs x86_64 and prompt to install Rosetta on Apple Silicon.
+    set(CPACK_APPLE_PKG_INSTALLER_CONTENT "<hostArchitectures><hostArchitecture>arm64</hostArchitecture></hostArchitectures>")
+
     set(CPACK_PRODUCTBUILD_DOMAIN "system")
 
     if(INSTALLER_IDENTITY)


clemperorpenguin:
This patch expands the generated .pkg, injects hostArchitectures="arm64" into its Distribution XML (which CPack's productbuild omits, causing macOS to assume the postflight script needs x86_64 and prompt for Rosetta on Apple Silicon), then re-flattens, rebuilds, and re-signs the package in place. There's no .pkg artifact to download because that signed packaging step only runs when Apple signing secrets are present, and GitHub withholds secrets from fork-PR CI - so the build skips packaging entirely and produces no installer until a maintainer runs it post-merge.

Geramy:
@clemperorpenguin resigning things are very risky business and usually ends up in breaking things. But i like the solution, have you tried this first?

```
# Explicitly set host architectures to arm64 to prevent Rosetta 2 prompts
set(CPACK_APPLE_PKG_INSTALLER_CONTENT "<hostArchitectures><hostArchitecture>arm64</hostArchitecture></hostArchitectures>")
```

fl0rianr (commented):
Blocking concern: this rebuilds the product archive after CPack has already created it, but the second productbuild invocation does not appear to preserve the metadata CPack originally passes, especially `--version` and `--identifier`.

In the normal CPack path, CPACK_PRODUCTBUILD_IDENTIFIER is set to com.lemonade.server, and CPack passes product metadata into `productbuild`. The patch script rebuilds with only --distribution, --package-path, optional --resources, and then signs with productsign. That may leave us with a notarized package whose product metadata differs from the original CPack package.

Can we either avoid the post-CPack rebuild, or pass through the original metadata when rebuilding? At minimum I think the second productbuild call should preserve `--version "${PROJECT_VERSION}"` and `--identifier "${CPACK_PRODUCTBUILD_IDENTIFIER}"` before notarization.

clemperorpenguin:
Cool, how is this?

I switched the macOS .pkg to be built unsigned by CPack and signed exactly once by the patch-macos-pkg step after it injects hostArchitectures="arm64", eliminating the risky sign-then-rebuild-then-re-sign cycle. I also added a metadata-preservation check to the patch script that snapshots the Distribution's <product>/<pkg-ref> identifier and version before the rebuild and fails the build if they don't match afterward.

Geramy:
> Cool, how is this?
> 
> I switched the macOS .pkg to be built unsigned by CPack and signed exactly once by the patch-macos-pkg step after it injects hostArchitectures="arm64", eliminating the risky sign-then-rebuild-then-re-sign cycle. I also added a metadata-preservation check to the patch script that snapshots the Distribution's / identifier and version before the rebuild and fails the build if they don't match afterward.

I don't want us to mess with the signing of this package or anything like that, it's super dangerous and prone to breaking everything. Please just apply this patch instead. 

```# Explicitly set host architectures to arm64 to prevent Rosetta 2 prompts
set(CPACK_APPLE_PKG_INSTALLER_CONTENT "<hostArchitectures><hostArchitecture>arm64</hostArchitecture></hostArchitectures>")``` it should only be like 3 lines of changes.

**Review by Geramy — APPROVED** [review]

**Comment by dboshell0** [comment]
Hi team thank you for your efforts. This bug is present on the 11.7 .pkg. I am not a programmer so I asked Grok to help which produced the below output. Hope it helps:

Yes. The interesting part is **why #2297 did not fix 11.7**.

`CMakeLists.txt` already has:

```cmake
set(CPACK_APPLE_PKG_INSTALLER_CONTENT "<hostArchitectures><hostArchitecture>arm64</hostArchitecture></hostArchitectures>")
```

That is a no-op:

1. Apple wants `hostArchitectures="arm64"` **on `<options>`**, not a nested element.
2. CPack **rebuilds** `CPACK_APPLE_PKG_INSTALLER_CONTENT` when it writes `Distribution`, and CMake’s writer never sets that attribute. So 11.7 still ships:

```xml
<options allow-external-scripts="no" customize="allow" rootVolumeOnly="false"/>
```

That is exactly what triggers the Rosetta dialog.

### Patch to send

A small project template CPack will actually use:

- `cmake/CPack.distribution.dist.in` — `<options ... hostArchitectures="arm64"/>` plus `@CPACK_PACKAGEMAKER_CHOICES@` (so you don’t get a second `<options>` without the attribute)
- `CMakeLists.txt` — drop the dead `CPACK_APPLE_PKG_INSTALLER_CONTENT` line; `list(PREPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")`

Also included: PR body and apply steps.

`/home/workdir/artifacts/lemonade-macos-rosetta-patch/`

- `cmake/CPack.distribution.dist.in`
- `CMakeLists.txt.patch`
- `PR_DESCRIPTION.md`
- `APPLY.md`

Suggested commit title:

`fix(macos): declare hostArchitectures=arm64 in productbuild Distribution`

After they build, this should hold:

```bash
pkgutil --expand Lemonade-*-Darwin.pkg /tmp/lemonade-pkg
grep hostArchitectures /tmp/lemonade-pkg/Distribution
```

No expand/flatten/re-sign cycle, which is what the team wanted to avoid on #2297.

PR #2303, opened 2026-06-18
fix(ci):  restart MacOS server before whisper metal tests

fl0rianr:
The unsigned macOS CI path can keep a lemond process running from earlier tests. The whisper Metal test sets GGML_METAL_NO_RESIDENCY=1, but whisper-server is spawned by lemond, so the backend may not inherit that environment variable.

Restart lemond immediately before the whisper Metal tests with GGML_METAL_NO_RESIDENCY=1 set, so the process under test inherits the intended Metal CI runtime configuration.

This is CI-only and does not change product code.


jeremyfowers:
@claude review

jeremyfowers (approved):

PR #2305, opened 2026-06-18
Fix - Model Download Race Condition

VultureZZ:
This commit introduces a mechanism to ensure that only one download operation occurs at a time for each model. It adds the `awaitExistingModelDownload` function to wait for any in-progress downloads before initiating a new one, preventing potential conflicts and data corruption. Additionally, a mutex-based locking system is implemented in the `ModelManager` and `HttpClient` to manage concurrent download requests effectively. This enhancement improves the reliability of model downloads across multiple requests.

### DIFF
--- a/src/app/src/renderer/utils/backendInstaller.ts
+++ b/src/app/src/renderer/utils/backendInstaller.ts
@@ -663,6 +663,34 @@ function extractExplicitBackend(loadBody?: Record<string, unknown>): { recipe: s
 }
 
 /**
+ * Wait for an in-flight model download started by another caller (or tab)
+ * instead of starting a second /pull that would race on the server.
+ */
+async function awaitExistingModelDownload(modelName: string): Promise<void> {
+  const downloadId = downloadTracker.getStableDownloadId(modelName, 'model');
+  downloadTracker.startServerPolling();
+
+  const serverDownloads = await downloadTracker.hydrateFromServer();
+  const active = serverDownloads.find(
+    item => item.model_name === modelName &&
+      (item.running === true || item.status === 'downloading'),
+  );
+  if (active) {
+    downloadTracker.applyServerDownload(active);
+  } else if (!downloadTracker.isActive(modelName)) {
+    return;
+  }
+
+  await waitForServerDownloadTerminal(
+    downloadId,
+    modelName,
+    new AbortController(),
+    () => undefined,
+    Boolean(active),
+  );
+}
+
+/**
  * Universal pre-flight check for all inference requests.
  * Ensures backend is installed, model is downloaded, and model is loaded —
  * all through tracked SSE paths visible in the Download Manager.
@@ -791,7 +819,12 @@ async function ensureModelReadyInternal(
 
     // Step 5: Pull model if not downloaded (shows in Download Manager)
     if (!isDownloaded) {
-      await pullModel(modelName, { declaredSizeGB: modelsData[modelName]?.size });
+      if (downloadTracker.isActive(modelName) ||
+          await downloadTracker.hasActiveServerDownload(modelName)) {
+        await awaitExistingModelDownload(modelName);
+      } else {
+        await pullModel(modelName, { declaredSizeGB: modelsData[modelName]?.size });
+      }
     }
 
     // Step 6: Load model into memory (merge loadBody if provided)
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -3137,10 +3137,31 @@ bool ModelManager::is_model_downloaded(const std::string& model_name) {
     return false;
 }
 
+std::shared_ptr<std::mutex> ModelManager::get_model_download_lock(const std::string& model_name) {
+    const std::string lock_key = resolve_model_name(model_name);
+    std::lock_guard<std::mutex> guard(model_download_locks_mutex_);
+    if (auto it = model_download_locks_.find(lock_key); it != model_download_locks_.end()) {
+        if (auto existing = it->second.lock()) {
+            return existing;
+        }
+    }
+    auto lock = std::make_shared<std::mutex>();
+    model_download_locks_[lock_key] = lock;
+    return lock;
+}
+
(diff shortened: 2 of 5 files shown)


src/cpp/server/model_manager.cpp
```diff
@@ -3137,13 +3137,34 @@ bool ModelManager::is_model_downloaded(const std::string& model_name) {
...
     if (info.recipe == "cloud") {
         update_model_in_cache(info.model_name, true);
         return;
     }
 
+    auto model_lock = get_model_download_lock(info.model_name);
+    std::lock_guard<std::mutex> download_guard(*model_lock);
+
+    // Another caller may have finished while we waited for the model lock.
+    if (do_not_upgrade && is_model_downloaded(info.model_name)) {
```
fl0rianr, line 3164: This post-lock re-check only catches cache-first callers. /load still calls download_registered_model(info) with the default do_not_upgrade=false, so a /load request queued behind an in-flight /pull can wait on this mutex and then still proceed into the download/update path. Could we either make /load call download_registered_model(info, true) or make this guard explicitly skip when the model became downloaded while waiting?

This matters because handle_load() currently downloads missing models with download_registered_model(info) and no do_not_upgrade=true.

src/cpp/server/server.cpp
```diff
@@ -3426,8 +3426,33 @@ void Server::handle_pull(const httplib::Request& req, httplib::Response& res) {
...
         } else {
-            // Legacy synchronous mode - blocks until complete
-            model_manager_->download_model(model_name, request_json, do_not_upgrade);
+            // Legacy synchronous mode - blocks until complete. Route through the
+            // shared download job registry so sync /pull deduplicates with any
+            // in-flight server-owned or SSE pull for the same model.
+            auto operation = [this, model_name, request_json, do_not_upgrade](DownloadProgressCallback progress_cb) {
+                model_manager_->download_model(model_name, request_json, do_not_upgrade, progress_cb);
+            };
+            auto job = start_download_job("model:" + model_name, "model", model_name, operation);
```
fl0rianr, line 3435: This job key uses the raw request model_name, while the ModelManager download lock uses resolve_model_name(model_name). Alias vs canonical requests for the same logical model can therefore create separate server job IDs/UI rows even though they serialize lower down. Could we key the download job with the same resolved model name used by get_model_download_lock()?

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -662,6 +662,34 @@ function extractExplicitBackend(loadBody?: Record<string, unknown>): { recipe: s
...
+ * Wait for an in-flight model download started by another caller (or tab)
+ * instead of starting a second /pull that would race on the server.
+ */
+async function awaitExistingModelDownload(modelName: string): Promise<void> {
+  const downloadId = downloadTracker.getStableDownloadId(modelName, 'model');
+  downloadTracker.startServerPolling();
+
+  const serverDownloads = await downloadTracker.hydrateFromServer();
+  const active = serverDownloads.find(
+    item => item.model_name === modelName &&
```
fl0rianr, line 675: Same alias/canonicalization issue on the client side: this only detects an existing server download if item.model_name exactly equals the caller’s modelName. If another caller started the same model through a different alias, ensureModelReady() can miss the active job and start another /pull. Could we normalize model names here or have the server snapshot expose a canonical model ID to compare against?

src/app/src/renderer/utils/backendInstaller.ts
```diff
@@ -791,7 +819,12 @@ async function ensureModelReadyInternal(
 
     // Step 5: Pull model if not downloaded (shows in Download Manager)
     if (!isDownloaded) {
-      await pullModel(modelName, { declaredSizeGB: modelsData[modelName]?.size });
+      if (downloadTracker.isActive(modelName) ||
+          await downloadTracker.hasActiveServerDownload(modelName)) {
```
fl0rianr, line 823: This pre-flight check has the same exact-match limitation as awaitExistingModelDownload(). If server-side dedupe is intended to be “one download per logical model”, the client check should use the same identity semantics as the server lock, not only the raw UI/request name.

src/cpp/server/utils/http_client.cpp
```diff
@@ -768,11 +770,51 @@ DownloadResult HttpClient::download_attempt(const std::string& url,
...
+
+static PathDownloadLockRegistry g_path_download_locks;
+
 DownloadResult HttpClient::download_file(const std::string& url,
                                          const std::string& output_path,
                                          ProgressCallback callback,
                                          const std::map<std::string, std::string>& headers,
                                          const DownloadOptions& options) {
+    auto path_lock = g_path_download_locks.acquire(output_path);
+    std::lock_guard<std::mutex> path_guard(*path_lock);
```
fl0rianr, line 816: This path-level lock is a good last line of defense against .partial corruption. Given that the higher-level model/job dedupe can still miss alias/canonical cases, could we add a regression test that races two callers against the same output path and verifies that the partial file is not concurrently written/corrupted?

fl0rianr (changes requested):
Thanks for working on this — the direction looks good, especially the path-level serialization in HttpClient and the per-model lock in ModelManager.

I think this still needs changes before merge, I noted them below.

Suggested fixes:
- Make /load and collection component auto-downloads cache-first where appropriate, e.g. download_registered_model(info, true).
- Use the same canonical model identity for server download job keys and frontend active-download matching.
- Add a regression test for /pull + /load concurrency and, ideally, alias/canonical model names

bong-water-water-bong (approved):
The race condition fix is well-layered across three levels (file, model, request). The `weak_ptr<shared_ptr<mutex>>` lock registry pattern is correctly implemented. The double-check inside `download_registered_model` correctly avoids redundant downloads for queued waiters.

One minor suggestion: when `start_download_job` returns an existing in-flight job, consider logging a warning if the new request's parameters (`checkpoint`, `recipe`, `do_not_upgrade`) differ from the original. Currently the second caller silently receives the first download's result regardless of parameter differences.

Otherwise LGTM — ship it.

PR #2306, opened 2026-06-18
Feature - Request Log

VultureZZ:
This is a big one for debugging purposes when building local projects. This implements a request log where users can have all API requests logged into a Postgres database for review and analysis. See example below.

Lemonade-App-06-18-2026_09_34_AM


jeremyfowers:
I really like this idea! So sorry in advance for what I'm about to say - this falls under the category of "please contact the maintainers before implementing large features".

GUI working group is replacing the entire existing GUI with a new one, called GUI3. Please join the discord and visit the gui3-beta channel to learn more.

@kpoineal for awareness. This feature is a top request and it would be lovely to have it included in the app eventually.

One last note: this kind of logging needs to be opt-in, otherwise it is a privacy violation.

Thanks for the PR @VultureZZ and I hope to see you contributing to gui3!

abn:
Drive by comment: this is a very handy feature imho. Part of the motivation why I proposed #2330. @VultureZZ if the broader telemetry feature gets accepted, do you think the plumbing is better reused?

I would also ask if its better to keep things modular, ie. allow users to bring an otel collector like Phoenix/Langfuse/Jaeger if they need this. That said, there might be value in enabling a real-time log view for admins without persistence, this could tap into the telemetry pipeline as well.

@jeremyfowers regarding "this kind of logging needs to be opt-in" - do you have thoughts on how should this work? Is this something you expect to be opt-in by the per client/user or the by admin? For the telemetry proposal, my approach has been "admin managed server config" +[ transparency via health endpoint](https://github.com/lemonade-sdk/lemonade/pull/2330/commits/5082756c9279d2bb084d1619813c3530da901687).

PR #2307, opened 2026-06-18
Feature - Default model fallback

VultureZZ:
This commit introduces a default model setting that allows the chat API to fall back to a configured model when an unknown model is requested. The implementation includes updates to the SettingsPanel for managing the default model, enhancements in the LLMChatSettings component to select the default model, and modifications in the server logic to utilize the default model during chat requests. Additionally, unit tests are added to verify the fallback functionality.


Geramy:
This is pretty neat, is there a way we can add to a response so the user knows it's fallen back to a different model, is that already implemented?

PR #2309, opened 2026-06-18
fix(macOS): updates whisper from v1.8.4 to v.1.8.5 for metal

fl0rianr:
updates whisper from v1.8.4 to v.1.8.5 for metal only (since only this one is available right now) to try a ci error fix.

### DIFF
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -13,7 +13,7 @@
     "vulkan": "v1.8.4",
     "rocm": "v1.8.4",
     "npu": "v1.8.4",
-    "metal": "v1.8.4"
+    "metal": "v1.8.5"
   },
   "sd-cpp": {
     "cpu": "master-672-1f9ee88",
--- a/src/cpp/server/backends/whisper_server.cpp
+++ b/src/cpp/server/backends/whisper_server.cpp
@@ -297,20 +297,7 @@ void WhisperServer::load(const std::string& model_name,
     std::vector<std::pair<std::string, std::string>> env_vars;
     fs::path exe_dir = fs::path(exe_path).parent_path();
 
-#ifdef __APPLE__
-    // On macOS, dyld uses DYLD_LIBRARY_PATH for dynamic library lookup.
-    std::string lib_path = exe_dir.string();
-
-    const char* existing_dyld_path = std::getenv("DYLD_LIBRARY_PATH");
-    if (existing_dyld_path && strlen(existing_dyld_path) > 0) {
-        lib_path = lib_path + ":" + std::string(existing_dyld_path);
-    }
-
-    env_vars.push_back({"DYLD_LIBRARY_PATH", lib_path});
-    if (is_debug()) {
-        std::cout << "[WhisperServer] Setting DYLD_LIBRARY_PATH=" << lib_path << std::endl;
-    }
-#elif !defined(_WIN32)
+#ifndef _WIN32
     // set LD_LIBRARY_PATH to include executable directory
     std::string lib_path = exe_dir.string();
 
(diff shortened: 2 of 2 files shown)


jeremyfowers:
@claude review

jeremyfowers:
@Geramy this seems reminiscent of the problems we had in llamacpp for macOS. Can you review please?

Geramy:
> @Geramy this seems reminiscent of the problems we had in llamacpp for macOS. Can you review please?

yes im a little confused on why we would set GGML_METAL_NO_RESIDENCY=1 this is a workaround feature to stop memory accumulation from happening but it also makes the memory not become "wired" to the process specifically and the model could get ejected or "free'd" its a work around, second, why do we need DYLD_LIBRARY_PATH set? if it is compiled for metal correctly we don't need that. This was an issue on sd-server as well but we later fixed it and the libraries became linked properly and it was no longer an issue. I could be wrong but im pretty sure that was a issue with sd that got fixed which could be the same issue here.

Geramy:
The model is loading fine here this must be a CI specific issue, I would say its safe to make whatever changes needed only to effect and work for fixing CI. Github uses fake macOS virtual machines and virtual m3's which sometimes can have problems, I am not sure if ```GGML_METAL_NO_RESIDENCY``` being 1 or 0 can solve this specifically but adjusting ```DYLD_LIBRARY_PATH``` should not be needed ever we should not touch that. if compiled right it launches I don't want to redirect dynamic library paths, its a no no for Mac really, unless you have to do it to accomplish a specific user case. otherwise its what you would call a workaround just like setting ```GGML_METAL_NO_RESIDENCY``` Is a workaround for stopping memory leaks.

```
2026-06-18 14:32:24.891 [Debug] (ProcessManager) Starting process with inherited output: /Library/Application Support/lemonade/.cache/bin/whispercpp/metal/whisper-server -m /Library/Application Support/lemonade/hub/models--ggerganov--whisper.cpp/snapshots/5359861c739e955e79d9a303bcbc70fb988958b1/ggml-large-v3-turbo.bin --port 8001
2026-06-18 14:32:24.896 [Info] (ProcessManager) Process started successfully, PID: 33153
2026-06-18 14:32:24.897 [Info] (WhisperServer) Process started with PID: 33153
2026-06-18 14:32:24.897 [Debug] (WrappedServer) Waiting for whisper-server to be ready...
2026-06-18 14:32:28.178 [Debug] (Server) GET /api/v1/models
2026-06-18 14:32:30.841 [Info] (WrappedServer) whisper-server is ready!
2026-06-18 14:32:30.841 [Info] (BackendWatchdog) Started watchdog for whisper-server using http://127.0.0.1:8001/health (streaming=on, non_streaming=on)
2026-06-18 14:32:30.841 [Info] (WhisperServer) Server is ready!
2026-06-18 14:32:30.841 [Debug] (Router) Backend started successfully in 5950ms
2026-06-18 14:32:30.841 [Info] (Router) Model loaded successfully. Total loaded: 1
2026-06-18 14:32:30.841 [Debug] (Server) POST /api/v1/load - 200
2026-06-18 14:32:30.860 [Debug] (Server) GET /api/v1/models
2026-06-18 14:32:33.258 [Debug] (Server) GET /api/v1/models
2026-06-18 14:32:33.266 [Debug] (Server) GET /api/v1/models
2026-06-18 14:32:33.272 [Debug] (Server) GET /internal/config
2026-06-18 14:32:33.274 [Debug] (Server) GET /internal/config - 200
2026-06-18 14:32:38.353 [Debug] (Server) GET /api/v1/models
2026-06-18 14:32:43.435 [Debug] (Server) GET /api/v1/models
2026-06-18 14:32:48.526 [Debug] (Server) GET /api/v1/models
```

fl0rianr:
Ok, I reworked it to be 100% CI only again and an isolated runner with better logs should improve the situation more.

Geramy (approved):
The right thing to do.....

PR #2315, opened 2026-06-19
fix(backends): stage and verify backend install before removing the working binary

### LINKED ISSUE #2312: Backend reinstall is not crash-safe: install_from_github deletes the working binary before downloading the replacement
# Backend reinstall is not crash-safe: `install_from_github` deletes the working binary before downloading the replacement

### Summary

When a backend binary needs upgrading (the installed version no longer matches the pinned `expected_version`, or `version.txt` is missing), `BackendUtils::install_from_github` deletes the existing install directory **before** it starts downloading the replacement asset. If that download then fails — a slow or unreliable link, a laptop going offline mid-flight, a transient GitHub 5xx — the user is left with **no working backend binary at all**: the old one is already gone and the new one never arrived.

The download/extract themselves are staged sensibly (archive goes to a temp path, extraction failures clean up), but the **destructive removal happens too early**, so a failed download is unrecoverable until the next successful online reinstall.

### Where

`src/cpp/server/backends/backend_utils.cpp`, in `BackendUtils::install_from_github` (current `main`):

- **L406** — version mismatch (upgrade) path:
  ```cpp
  if (installed_version != expected_version) {
      LOG(INFO, ...) << "Upgrading " << spec.binary << ...;
      needs_install = true;
      fs::remove_all(install_dir);   // <-- old binary deleted here, BEFORE the download below
  }
  ```
- **L416** — executable present but `version.txt` missing path: same early `fs::remove_all(install_dir)`.

The actual download (`HttpClient::download_file`) doesn't run until ~L426+, well after the directory has already been wiped.

The extract-failure paths (L616, L625) have the same shape — `fs::remove_all(install_dir)` on failure — but those at least run after a *successful* download, so the regression window is the pre-download removal at L406/L416.

### Steps to reproduce

1. Have a backend installed and working (e.g. `llamacpp`), with a valid `version.txt`.
2. Bump the pinned version so the installed one is now considered stale (or delete `version.txt`).
3. Trigger a reinstall while the network is unavailable / the release asset is unreachable.

**Expected:** the reinstall fails, but the previously-installed working binary is still present and usable.
**Actual:** the install directory was already removed at L406/L416, so the working binary is gone and the backend can no longer start until a successful online reinstall.

### Proposed fix

Make the reinstall atomic — stage and verify the replacement before touching the installed copy:

1. On version mismatch / missing `version.txt`, set `needs_install = true` but **do not** remove `install_dir` yet.
2. Download to the temp archive (unchanged) and extract into a **staging directory** (a sibling of `install_dir`, so the final move is a same-filesystem rename), not directly into `install_dir`.
3. Verify the executable exists in the staging directory.
4. **Only then** remove the old `install_dir` and atomically rename staging → `install_dir`.
5. On any failure (download, extract, missing executable) remove the staging dir and the temp archive and leave the existing `install_dir` untouched.

This keeps the working binary in place across any interrupted/failed reinstall and only swaps it out once the replacement is fully downloaded, extracted, and verified.

### Notes

- Surfaced by @ckuethe on PR #2279 ("not remove and replace the llama binary until the download is verified to be complete" — slow/unreliable links, e.g. on an airplane). #2279 handled the *latest-version-lookup* failure; this issue is the complementary *download* robustness case.
- Happy to submit a PR with the staging + atomic-swap change and a unit test (the `test/cpp/` C++ test harness) covering the "failed reinstall leaves the old binary intact" invariant.

ianbmacdonald:
## What

Makes backend (re)installation crash-safe. `BackendUtils::install_from_github` no longer deletes the working binary before the replacement is downloaded and verified.

Closes #2312.

## Why

On a version-mismatch / missing-`version.txt` reinstall, `install_from_github` called `fs::remove_all(install_dir)` **up front** — before downloading the new release asset. If that download then failed (slow/unreliable link, going offline mid-flight, a transient GitHub 5xx), the backend was left with **no usable binary**: the old one was already gone and the new one never arrived.

Surfaced by @ckuethe on #2279 ("not remove and replace the llama binary until the download is verified to be complete" — slow links, e.g. on an airplane). #2279 handled the latest-version *lookup* failure; this is the complementary *download* robustness case.

## The fix

Stage → verify → atomic swap:

1. Download + extract into a sibling `install_dir + ".staging"` directory — the working install is untouched.
2. Verify the expected executable is present in staging.
3. Promote with a recoverable backup that never leaves the backend with nothing:
   - rename the existing install aside to `install_dir + ".old"`,
   - rename staging into `install_dir`,
   - delete `.old` only once the new tree is in place;
   - if the promotion rename fails, roll `.old` back.

Additional hardening (from review):
- A RAII guard removes the staging tree on **any** pre-swap failure (download/probe/extract), so no half-built tree is left behind.
- Stale staging that can't be cleared is treated as fatal, so a leftover binary can never be promoted as a mixed install.
- The staged `version.txt` write is checked.
- The swap helper throws a distinct error on a *swap* failure (vs. the empty-return "executable not found" case), so the caller reports an accurate message.

The staging/swap logic is a small header-only helper (`lemon/backends/install_staging.h`) so it can be unit-tested without the heavier `backend_utils.cpp` dependencies; `find_executable_in_install_dir` now delegates to it (single source of truth).

## Testing

New ctest **`InstallAtomicityTest`** (`test/cpp/test_install_atomicity.cpp`), network-free, covering: successful swap (no `.old`/`.staging` left behind), **regression guard** — a staging tree missing the executable leaves the old working binary + `version.txt` byte-for-byte intact, fresh install, and (POSIX) a **failed filesystem swap** preserves the working binary and throws.

RED→GREEN verified: the regression assertions fail against the pre-fix `remove-before-verify` ordering and pass against the fix.

### Validation environment

This is a pure `std::filesystem` change with no inference path exercised, so **backend = none** (no model load, no ROCm/Vulkan). The unit test runs on every platform via CI ctest.

| Field | Value |
|-------|-------|
| Change type | backend install/filesystem logic — **no inference, backend=none** |
| Local | Ubuntu 26.04 (resolute), full `cmake --build` clean, `lemond` links, `ctest` 3/3 green |
| ai3 | Ubuntu 26.04 LTS (resolute), kernel `7.0.0-22-generic`, glibc 2.43, g++ 15.2.0 — `InstallAtomicityTest` 17/17 pass, clean `-Wall -Wextra` |
| Tests | `test/cpp/test_install_atomicity.cpp` (ctest `InstallAtomicityTest`) |

## Scope / follow-up

`install_therock` (the ROCm tarball path) has the same `remove_all`-before-download shape and would benefit from the same staging approach, but is left out of scope to keep this PR focused — happy to follow up in a separate PR.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1740,3 +1740,22 @@ if(EXISTS "${_LATEST_FALLBACK_TEST_SRC}")
     include(CTest)
     add_test(NAME LatestVersionFallbackTest COMMAND test_latest_version_fallback)
 endif()
+
+# Backend install atomicity (staging + verified atomic swap).
+# Header-only unit under test, so no extra source files are required.
+set(_INSTALL_ATOMICITY_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_install_atomicity.cpp"
+)
+if(EXISTS "${_INSTALL_ATOMICITY_TEST_SRC}")
+    add_executable(test_install_atomicity
+        test/cpp/test_install_atomicity.cpp
+    )
+    target_include_directories(test_install_atomicity PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+
+    # Enable testing and register the test with CTest
+    include(CTest)
+    add_test(NAME InstallAtomicityTest COMMAND test_install_atomicity)
+endif()
--- a/src/cpp/include/lemon/backends/install_staging.h
+++ b/src/cpp/include/lemon/backends/install_staging.h
@@ -1,0 +1,118 @@
+#pragma once
+
+#include <filesystem>
+#include <stdexcept>
+#include <string>
+#include <system_error>
+
+// Header-only, dependency-light staging/atomic-swap helpers for backend
+// installation. Kept free of the heavier backend_utils.cpp dependencies so the
+// crash-safety invariant can be unit-tested in isolation (test/cpp/test_install_atomicity.cpp).
+namespace lemon::backends {
+
+    /**
+     * Recursively search `dir` for a regular file named `binary_name` (or
+     * `binary_name + ".exe"` on Windows). Returns the full path, or "" if not
+     * found or `dir` does not exist.
+     */
+    inline std::string find_executable_in_dir(const std::string& dir,
+                                              const std::string& binary_name) {
+        namespace fs = std::filesystem;
+        if (!fs::exists(dir)) {
+            return "";
+        }
+#ifdef _WIN32
+        const std::string binary_name_exe = binary_name + ".exe";
+#endif
+        for (const fs::directory_entry& dir_entry : fs::recursive_directory_iterator(dir)) {
+            if (dir_entry.is_regular_file()) {
+                const auto& fname = dir_entry.path().filename();
+                if (fname == binary_name
+#ifdef _WIN32
+                    || fname == binary_name_exe
+#endif
+                ) {
+                    return dir_entry.path().string();
+                }
+            }
+        }
+        return "";
+    }
+
+    /**
+     * Atomically promote a fully-prepared staging directory to `install_dir`.
+     *
+     * Verifies that `staging_dir` contains `binary_name` before touching the
+     * currently-installed copy. The caller MUST create `staging_dir` as a
+     * sibling of `install_dir` so the renames stay on one filesystem.
+     *
+     * The promotion keeps a recoverable copy of the previous install at all
+     * times so a failed swap can never lose both installs:
+     *   1. move the existing install aside to `install_dir + ".old"`;
+     *   2. rename `staging_dir` into `install_dir`;
+     *   3. only then delete the `.old` backup.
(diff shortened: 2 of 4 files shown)


ianbmacdonald:
Process note: I opened this from my fork branch (`ianbmacdonald:fix/install-atomicity`). Since I have push access on this repo, I'm happy to move the branch onto `lemonade-sdk/lemonade` directly (so you can push to it, rebase, or take it over without the fork-PR edit dance) if that's easier for review — just say the word and I'll re-point it.

ianbmacdonald:
Addressed all three Copilot comments in the rebased push (`e73f10407`, now on top of `e6d511bc9`):

1. **`test_install_atomicity.cpp`** — added the missing `<functional>` (`std::hash<std::string>`) and `<iterator>` (`std::istreambuf_iterator`) includes rather than relying on transitive inclusion.
2. **`backend_utils.cpp`** — removed the unused `url` variable (the download path builds its URL from `base_download_url + filename`).
3. **`backend_utils.cpp`** — the archive-leak-on-throw is now handled by a `ZipGuard` RAII (mirroring the existing `StagingGuard`) that removes the downloaded archive on **any** scope exit, including a throw from `commit_staged_install()` on a swap/rename failure. This also consolidates the previously-scattered per-throw `fs::remove(zip_path)` calls into a single owner.

bitgamma (approved):

PR #2317, opened 2026-06-19
Add support for additional draft checkpoint

bitgamma:
- Added support for `draft` checkpoint in llamacpp recipe. This allows downloading a drafter model (MTP, EAGLE3, etc) alongside the main model. Works well with the existing `mtp` label and with manually specifying MTP/drafting options
- Added Gemma4 MTP models that use this mechanism.

### DIFF
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -690,13 +690,65 @@
     },
     "Gemma-4-12B-it-GGUF": {
         "checkpoint": "unsloth/gemma-4-12b-it-GGUF:Q4_K_M",
+        "mmproj": "mmproj-F16.gguf",
         "recipe": "llamacpp",
         "suggested": true,
         "labels": [
             "tool-calling",
+            "vision",
             "llamacpp"
         ],
-        "size": 7.12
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Nice addition, great! One concern: --model-draft is reserved unconditionally, even when draft_path is empty. Since validate_custom_args() rejects reserved flags, this prevents users from manually passing --model-draft / -md / --spec-draft-model through llamacpp_args for models that do not have a registered draft checkpoint. Could we move the reservation into the if (!draft_path.empty()) block, e.g. pass the aliases directly to push_arg() when Lemonade actually adds the managed --model-draft argument?

bitgamma:
@fl0rianr, this is intentional and the same is done with mmproj file and all other reserved argumens really. Allowing specifying arbitrary paths is a potential security risk

fl0rianr (approved):
Your right - I remember that topic brought up  by Mario. In that case you have my approval.

fl0rianr:
not possible to merge until the whisper issue is resolved. https://discord.com/channels/1392562559122407535/1394414791593820252/1517829183206854736

PR #2322, opened 2026-06-19
fix(windows): use ProcessManager::run_command for 7z extraction instead of system() Fixes #2313

### LINKED ISSUE #2313: Bug Report: Severe Llama.cpp/CUDA Downloader Loop & Hardcoded Folder Deletion Issue on Windows
### Platform

Windows

### Lemonade Version

10.7.0

### GPU / APU Model

NVIDIA GeForce RTX 3060 Ti

### Component

llama.cpp

### Bug Description

There is a critical hardcoded bug in how Lemonade handles the extraction and deployment of the backend server files (`llamacpp_cuda`). When attempting to boot a large local model like **Llama-3-70B (~44GB)**, the application gets stuck in an infinite, broken downloading/extraction loop, making it impossible to initialize the local chat interface. 

The application downloads the backend dependency package, but fails at the extraction stage, triggering an aggressive internal script that forces the entire directory to delete, leading to continuous failing cycles.

### Steps to Reproduce

1. Set a custom cache storage path in Lemonade configurations (e.g., `E:\LemonadeCache`).
2. Download or point to a large local GGUF model like Llama-3-70B-Instruct.
3. Click 'Play' or try to initialize the model backend.
4. Monitor the file system path: `E:\LemonadeCache\Data\lemonade\bin\llamacpp\`.
5. Observe the app creating the `cuda` folder, wiping existing injected files, failing extraction, and then force-deleting the entire directory structure before retrying.

**Expected Behavior:**
The application should successfully extract the downloaded `llamacpp_cuda` package into the specified directory. If extraction fails or if the user manually provides a working, compiled `llama-server.exe` build (such as b9586/b9587) with its required `.dll` files, the application should detect it and boot the model without purging or wiping the files.

**Actual Behavior:**
The extraction fails with a `Failed to extract archive` error. The program immediately triggers a hardcoded force-delete routine that entirely purges the `cuda` folder from the drive (bypassing Read-Only locks and unlinking Windows Directory Junction links `mklink /J`), then restarts the broken download loop endlessly.

### Log Output

```shell
[Info] (Server) Model already loaded: Moonshine-Medium-Streaming
[Debug] (MoonshineServer) Audio data size: 13726208 bytes
[Debug] (MoonshineServer) Sending multipart request to http://127.0.0.1:8005/inference (direct data)
[Debug] (MoonshineServer) Response status: 500
[Error] (Server) Error 500: POST /api/v1/audio/transcriptions
[Debug] (Server) POST /api/v1/audio/transcriptions - 500
```

### **Additional Discovery via File-System Monitoring:**
We verified this continuous purging behavior using Windows Sysinternals Process Monitor and manual directory updates. The backend downloader script automatically invokes a forced directory clean/removal routine immediately upon encountering the `Failed to extract archive` status, rendering any manual hotfixes or custom binary placements useless.

### **Attached Proof:**
*(Please review the uploaded server logs and screenshot showing the "Failed to extract archive" error status on custom drives).*

Image

Phqen1x:
**Summary**

  - On Windows, system() called from a SUBSYSTEM:WINDOWS process causes a brief console/progress window to flash during 7z extraction. Replaces it with ProcessManager::run_command(), which uses CreateProcess with CREATE_NO_WINDOW to suppress that window while still capturing output and exit codes.

  **Test plan**

  - [x] Trigger a backend installation that extracts a .7z archive on Windows
  - [x] Confirm no console/progress window appears during extraction
  - [x] Confirm extraction completes successfully and the backend loads

Fixes #2313

### DIFF
--- a/src/cpp/server/backends/backend_utils.cpp
+++ b/src/cpp/server/backends/backend_utils.cpp
@@ -185,7 +185,8 @@ namespace lemon::backends {
         std::string output;
         int result = lemon::utils::ProcessManager::run_command(command, output, 300);
         if (result != 0) {
-            LOG(ERROR, backend_name) << "Extraction failed with code: " << result << std::endl;
+            LOG(ERROR, backend_name) << "Extraction failed with code: " << result
+                                     << (output.empty() ? "" : " - " + output) << std::endl;
             return false;
         }
         return true;
--- a/src/cpp/server/backends/backend_utils.cpp
+++ b/src/cpp/server/backends/backend_utils.cpp
@@ -182,7 +182,8 @@ namespace lemon::backends {
         LOG(ERROR, backend_name) << "Error: .7z backend archives are only expected on Windows. Linux CUDA assets should be .tar.xz." << std::endl;
         return false;
 #endif
-        int result = system(command.c_str());
+        std::string output;
+        int result = lemon::utils::ProcessManager::run_command(command, output, 300);
         if (result != 0) {
             LOG(ERROR, backend_name) << "Extraction failed with code: " << result << std::endl;
             return false;


kenvandine:
@claude review this

kenvandine (approved):
This looks great, and thanks for adding the log output. I am particularly happy to see this actually brings it more in line with how we handle this elsewhere.

Issue #2299, opened 2026-06-18
Debian 13 package not installable on Debian 13 without backports


romainreignier:
### Platform

Linux/Other (please specify)

### Lemonade Version

10.8.0

### Component

Other / Not sure

### Bug Description

The freshly released 10.8.0 version added support to generate a Debian 13 .deb artifact with PR #2240 
But on a default Debian 13 system, the package won't install with this output:

```
$ sudo apt install ./lemonade-server_10.8.0-debian13_amd64.deb
Note, selecting 'lemonade-server' instead of './lemonade-server_10.8.0-debian13_amd64.deb'
Solving dependencies... Error!
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

Unsatisfied dependencies:
 lemonade-server : Depends: libcpp-httplib0.41 (>= 0.41.0+ds) but it is not installable
                   Recommends: ffmpeg but it is not going to be installed
                   Recommends: libxrt-npu2 but it is not installable
Error: Unable to correct problems, you have held broken packages.
Error: The following information from --solver 3.0 may provide additional context:
   Unable to satisfy dependencies. Reached two conflicting decisions:
   1. lemonade-server:amd64=10.8.0~13 is selected for install
   2. lemonade-server:amd64 Depends libcpp-httplib0.41 (>= 0.41.0+ds)
      but none of the choices are installable:
      [no choices]
```

The httplib package available in Debian 13 repos is version 0.18.

In the CI workflow, trixie-backports is used on [these lines](https://github.com/lemonade-sdk/lemonade/blob/8e3e0b621fd9d9668defea234433fd7374339bf2/.github/workflows/cpp_server_build_test_release.yml#L296-L307) but this is not reflected in the [Debian guide](https://github.com/lemonade-sdk/lemonade/blob/8e3e0b621fd9d9668defea234433fd7374339bf2/docs/guide/install/debian.md).

### Steps to Reproduce

1. Use a Debian 13 OS without the backports manually installed
2. Download the 10.8.0 deb package for Debian 13:
    ```
    wcurl https://github.com/lemonade-sdk/lemonade/releases/download/v10.8.0/lemonade-server_10.8.0-debian13_amd64.deb
    ```
3. Try to install it per the guide instructions:
    ```
    sudo apt install ./lemonade-server_10.8.0-debian13_amd64.deb
[...]

PR #2323, opened 2026-06-19
docs: fix Debian 13 installation docs for issue #2299

superm1:
- Add Debian Backports prerequisite to debian.md (libcpp-httplib0.41 not
  available in Debian 13 default repos, only 0.18)
- Add Backports note to install README for quick reference
- Update dev docs dependency list to include libcpp-httplib0.41 and
  libxrt-npu2 which were missing


fl0rianr (approved):
LGTM, thanks! This matches the root cause from #2299: the Debian 13 package requires libcpp-httplib0.41, which is only available via trixie-backports in this setup. The docs update also mirrors the existing Debian 13 CI/release build, which already enables trixie-backports before installing the build dependencies.

CI errors are unrelated.

PR #2295, opened 2026-06-18
fix(gpu): support wildcards in GPU detection logic

jtlayton:
This patch fixes ROCm detection for me for the GPUs covered by wildcard strings.

The identify_rocm_arch_from_name() function converts KFD gfx_target_version values (e.g. 110003, 120001) into specific family strings like gfx1103 and gfx1201. However, the RECIPE_DEFS table uses X-suffixed wildcards (gfx110X, gfx120X) to represent entire architecture families.

device_matches_constraint() did an exact string comparison, so gfx1103 != gfx110X and gfx1201 != gfx120X, causing valid AMD GPUs (RDNA3/RDNA4) detected via KFD sysfs to be reported as "Unsupported GPU" for ROCm backends.

Fix device_matches_constraint() to treat a trailing X in allowed family strings as a prefix wildcard match.

PR #2324, opened 2026-06-20
fix(server): restore ROCm gfx family normalization for dGPU detection

### LINKED ISSUE #2319: ROCm unavailable for gfx103X, gfx110X and gfx120X
### Platform

Linux/Arch

### Lemonade Version

10.8.0

### GPU / APU Model

_No response_

### Component

llama.cpp

### Bug Description

This [particular change](https://github.com/lemonade-sdk/lemonade/commit/2a7aa18ce8ff4a98355b0ae8b53c3ec7146227fd#diff-1bee506e4f3ddbe657f16fe42ded2cf42c89f6e9590cb3a11263c7710fba485eL1835-R1804) broke the rocm backend for any AMD GPU from the gfx103X, gfx110X and gfx120X families.
I believe that this happens because the [new code](https://github.com/lemonade-sdk/lemonade/blob/d199bcb090c3488375b18632740647f610f091a0/src/cpp/server/system_info.cpp#L1850) returns the specific name, for example, gfx1201, instead of returning gfx120X.

### Steps to Reproduce

1. Try to run the command `lemonade backends install llamacpp:rocm` with an AMD GPU from the gfx103X, gfx110X or gfx120X families.
2. See that it returns the error: "GPU Unsupported: <gpu_family>"
3. You can also run the command `lemonade backends` and see that it returns the same error wherever the rocm backend is available.

_No response_

### Log Output

```shell

```

### Additional Context

_No response_

ianbmacdonald:
Fixes #2319.

## Problem

`identify_rocm_arch_from_name()` returns the **specific** gfx arch (e.g. `gfx1100`, `gfx1201`) from the gfx-regex and KFD numeric-ISA detection paths, instead of the ROCm **family** download target (`gfx110X`, `gfx120X`) that the backend support set expects. Commit 2a7aa18c (#2093) removed `ROCM_ARCH_MAPPING`, which regressed ROCm availability for every RDNA2/3/4 dGPU detected via those two paths — the marketing-name heuristics still return families, so only the gfx-string / KFD-numeric path broke.

On Linux, KFD hands the ISA through as a number (e.g. `110000`), so an RX 7900 XT is reported as `gfx1100`, which is not in the support set `{gfx103X, gfx110X, gfx120X, ...}` → **`Unsupported GPU: gfx1100`**.

## Fix

Restore the specific→family normalization in a small header-only helper (`lemon/rocm_arch.h`) applied to both regressed return paths, mirroring the removed `ROCM_ARCH_MAPPING`:
- `gfx1030-gfx1034 → gfx103X`, `gfx1100-gfx1103 → gfx110X`, `gfx1200/gfx1201 → gfx120X`
- gfx115x iGPUs (`gfx1150/1151/1152`) pass through as exact targets
- the unconfirmed RDNA2 ISAs `gfx1033/1035/1036` return `""` (unsupported), matching the prior mapping

A standalone unit test (`test/cpp/test_rocm_arch.cpp`, CTest `RocmArchTest`) covers all families, the iGPU pass-through, the unsupported ISAs, and idempotency.

## Validation

**Unit:** `RocmArchTest` — 19/19 pass.
**Build:** `lemonade-server-core` + `lemond` build clean (Ninja/GCC).
**Real hardware (RX 7900 XT, gfx1100):** built `lemond` from this branch, ran it on the host, queried `/api/v1/system-info`:

| | before (10.8.0) | after (this branch) |
|---|---|---|
| detected `family` | `gfx1100` | **`gfx110X`** |
| rocm backend status | `Unsupported GPU: gfx1100` | **`Backend is supported but not installed`** (all recipes) |

### Test environment
- **lemonade:** 10.8.0 dev build from this branch (`20e9cbf15`), not a release
- **OS:** Ubuntu 26.04 LTS · **kernel:** `7.0.0-22-generic` · **arch:** x86_64
- **GPU:** AMD Navi 31 [Radeon RX 7900 XT] `[1002:744c]` (rev c8) — gfx1100, RDNA3 (`gfx110X` family), 24 GB VRAM
- **ROCm:** 7.2.4 present (`/opt/rocm-7.2.4`) but **not exercised** — this is a GPU-**detection** fix; validation is via `/system-info` family + backend-availability status, no ROCm inference run (the rocm backend is not installed on this host, which is exactly why the "supported but not installed" status is the correct post-fix result)
- **backend exercised:** none (detection / system-info only)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1720,3 +1720,20 @@ if(EXISTS "${_GGUF_CAPS_TEST_SRC}")
     include(CTest)
     add_test(NAME GgufCapabilitiesTest COMMAND test_gguf_capabilities)
 endif()
+
+# Header-only unit under test (lemon/rocm_arch.h), so no extra source files.
+set(_ROCM_ARCH_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_rocm_arch.cpp"
+)
+if(EXISTS "${_ROCM_ARCH_TEST_SRC}")
+    add_executable(test_rocm_arch
+        test/cpp/test_rocm_arch.cpp
+    )
+    target_include_directories(test_rocm_arch PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+
+    # CTest is already enabled by the earlier test blocks; just register the test.
+    add_test(NAME RocmArchTest COMMAND test_rocm_arch)
+endif()
--- a/src/cpp/include/lemon/rocm_arch.h
+++ b/src/cpp/include/lemon/rocm_arch.h
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <string>
+
+namespace lemon {
+
+// Collapse a specific RDNA dGPU gfx target to the ROCm "family" download target
+// that the backend support set and install filenames expect:
+//   gfx1030-gfx1036 -> gfx103X   (RDNA2, Radeon RX 6000)
+//   gfx1100-gfx1103 -> gfx110X   (RDNA3, Radeon RX 7000)
+//   gfx1200/gfx1201 -> gfx120X   (RDNA4, Radeon RX 9000)
+// iGPU targets (gfx1150/gfx1151/gfx1152) ship as exact binaries and pass through
+// unchanged, as do CDNA / exact-package targets (e.g. gfx90a). Anything else
+// passes through unchanged.
+//
+// The RDNA2 set mirrors backend_versions.json `url_mapping`, which maps every
+// gfx1030-gfx1036 to the published `gfx103X-all` archive.
+//
+// This restores the specific->family normalization that #2093 (commit 2a7aa18c)
+// dropped when it removed ROCM_ARCH_MAPPING. Without it, identify_rocm_arch_from_name
+// returns the specific arch (e.g. gfx1100, gfx1201) from the gfx-regex and KFD
+// numeric-ISA detection paths, which no longer matches the gfx103X/gfx110X/gfx120X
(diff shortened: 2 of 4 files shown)


ianbmacdonald:
Amended (force-pushed) after a closer look at the package metadata: the helper now collapses the **full** \`gfx1030`–\`gfx1036\` range to \`gfx103X\`, not just \`gfx1030/1031/1032/1034\`.

`backend_versions.json` `url_mapping` publishes `gfx103X-all` for all of `gfx1030`–`gfx1036`, so excluding `gfx1033/1035/1036` (which an earlier version did, inherited from the removed `ROCM_ARCH_MAPPING`) would have left those exact IDs unsupported despite a bundle existing — the same class of bug as this issue. The unit test now covers the full range (22 cases). Commit message + header comment updated to match.

jtlayton:
This looks like a more complex version of the fix here: https://github.com/lemonade-sdk/lemonade/pull/2295

This one will need to be updated with the addition of new GPUs in the same family. Is that a good thing? I'm not sure.

ianbmacdonald:
Closing in favor of #2295, which fixes the same regression (#2319) with a smaller, data-driven approach — a trailing-`X` wildcard in `device_matches_constraint` so `gfx110X` matches `gfx1100`–`gfx1103` without enumerating each arch. That's the more maintainable path (no code change when a new same-family GPU appears), and @jtlayton rightly pointed out here that the explicit normalization in this PR would need updating per new GPU in a family.

One data point in case it's useful for #2295: I reproduced #2319 on an RX 7900 XT (gfx1100, ROCm 7.2.4, Ubuntu 26.04), where `lemonade backends` reported `Unsupported GPU: gfx1100`, and confirmed that restoring family matching brings ROCm detection back end-to-end — the download target also resolves, since `url_mapping` in `backend_versions.json` carries the specific `gfx1100`–`gfx1103` keys alongside the `gfx110X` family key. Thanks @jtlayton.

PR #2326, opened 2026-06-20
fix(moonshine): return 400 on invalid audio

abn:
Return 400 Bad Request instead of 500 when audio format is invalid. Add integration test coverage for WAV RIFF errors and unsupported OGG.

### DIFF
--- a/src/cpp/server/backends/moonshine_server.cpp
+++ b/src/cpp/server/backends/moonshine_server.cpp
@@ -295,8 +295,35 @@ json MoonshineServer::forward_multipart_audio_data(const std::string& audio_data
     LOG(DEBUG, "MoonshineServer") << "Response status: " << res.status_code << std::endl;
 
     if (res.status_code != 200) {
-        throw std::runtime_error("moonshine-server returned status " +
-                                std::to_string(res.status_code) + ": " + res.body);
+        std::string err_msg = res.body;
+        std::string err_type = "audio_processing_error";
+        int status_code = res.status_code;
+
+        try {
+            json error_json = json::parse(res.body);
+            if (error_json.contains("error")) {
+                if (error_json["error"].is_string()) {
+                    err_msg = error_json["error"].get<std::string>();
+                } else if (error_json["error"].is_object() && error_json["error"].contains("message")) {
+                    err_msg = error_json["error"]["message"].get<std::string>();
+                }
+            }
+        } catch (...) {
+            // Keep res.body as raw error message
+        }
+
+        if (status_code == 400 || (status_code == 500 && err_msg.find("Not a valid RIFF file") != std::string::npos)) {
+            status_code = 400;
+            err_type = "invalid_request_error";
+        }
+
+        return json{
+            {"error", {
+                {"message", "Transcription failed: " + err_msg},
+                {"type", err_type},
+                {"status_code", status_code}
+            }}
+        };
     }
 
     try {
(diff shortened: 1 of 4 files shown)


fl0rianr (changes requested):
Thanks, this is the right direction and propagating error.status_code through set_error_response() makes sense.

I’d request one change before merge: please avoid classifying backend errors as 400 based on substrings in the error message. The backend now already returns HTTP 400 specifically for load_wav_file() failures, so the C++ wrapper can rely on the backend status code instead. With the current "Invalid"/ "Unsupported" matching, a real server-side 500 whose message happens to contain those words would be returned to clients as 400 invalid_request_error.

I’d suggest removing the message heuristic and setting invalid_request_error only for backend 4xx responses, or even only for status 400 in this PR.

Test nits:
- The OGG test currently uses `OggS...`, so it’s testing an invalid/corrupt file rather than a real unsupported OGG. Either rename it or use a valid OGG fixture.
- Avoid asserting the exact decoder message "Not a valid RIFF file"; the contract should be HTTP 400 + invalid_request_error.
- Consider using NamedTemporaryFile(..., suffix=...) instead of fixed temp paths.

abn:
@fl0rianr I have committed the requested changes.

abn:
I went ahead and updated the C++ wrapper to temporarily restore the message-based fallback heuristic for now. 

The integration tests run against a precompiled binary of the Moonshine server downloaded at runtime (`0.0.62`, built from the `main` branch). Since the Python fix in `tools/moonshine-server/main.py` is in this PR branch and hasn't been merged or released yet, the precompiled binary used in CI still throws an uncaught exception on invalid audio uploads, resulting in an HTTP `500` response. 

To keep the integration tests strict (asserting `400 Bad Request` as requested) and prevent CI failures, we are temporarily mapping HTTP `500` responses containing `"Not a valid RIFF file"` to HTTP `400` `invalid_request_error` in `moonshine_server.cpp`.

Once this PR is merged and a new release of `moonshine-server-rocm` (e.g., `0.0.63`) is published and pinned in `backend_versions.json`, we can safely remove this temporary message heuristic in a follow-up cleanup PR.

Let me know if this works for you!

fl0rianr (approved):
Thanks for the follow-up changes. This looks good to me now. No further changes needed.

Approved, assuming the remaining CI failure is unrelated to this PR.

abn:
Looks unrelated to me unless I am missing something obvious.

```
FAIL: test_001_transcription_basic (__main__.WhisperTests)
Test basic audio transcription with Whisper.
  File "/Users/runner/work/lemonade/lemonade/test/server_whisper.py", line 137, in test_001_transcription_basic
    model = self._load_whisper_model_or_fail()
  File "/Users/runner/work/lemonade/lemonade/test/server_whisper.py", line 118, in _load_whisper_model_or_fail
    self.assertEqual(
AssertionError: 500 != 200 : Failed to load model Whisper-Tiny with metal backend: {"error":{"code":"model_load_error","message":"Failed to load model 'Whisper-Tiny': Failed to download whisper-server from: https://github.com/lemonade-sdk/whisper.cpp-rocm/releases/download/v1.8.5/whisper-v1.8.5-darwin-metal-arm64.tar.gz - Download failed after 6 attempts.\nLast error: HTTP error 404 for URL: https://github.com/lemonade-sdk/whisper.cpp-rocm/releases/download/v1.8.5/whisper-v1.8.5-darwin-metal-arm64.tar.gz","param":"model","requested_model":"Whisper-Tiny","type":"model_load_error"}}

Ran 1 test in 3.680s

FAILED (failures=1)

```

PR #2329, opened 2026-06-20
Lemonade cli  list models with size in Gb

GabrielReusRodriguez:
Added to Lemonade cli the sizes of every model in Gb. 

The Web UI shows you every model with its size in Gb , so why not the lemonade CLI?  As user I found useful to know how big is the model to decide if I do a lemonade pull or not, depending on my hardware config.

Before:
image

After changes:
image

### DIFF
--- a/src/cpp/cli/lemonade_client.cpp
+++ b/src/cpp/cli/lemonade_client.cpp
@@ -6,6 +6,7 @@
 #include <regex>
 #include <sstream>
 #include <nlohmann/json.hpp>
+#include <sstream>
 
 namespace lemonade {
 
@@ -14,7 +15,7 @@ using json = nlohmann::json;
 static const int DEFAULT_CONNECTION_TIMEOUT_MS = 30000;
 static const int DEFAULT_READ_TIMEOUT_MS = 30000;
 static const int LONG_TIMEOUT_MS = 86400000;
-static const double UNKNOWN_SIZE = 0.0;
+static const double UNKNOWN_MODEL_SIZE = 0.0;
 
 static std::regex build_name_filter_regex(const std::string& name_filter) {
     std::string regex_pattern;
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Thanks for adding this — showing model sizes in `lemonade list` is useful and aligns the CLI better with the UI.

I think we should tighten a couple of details before merging:
- avoid under-reporting collection sizes when one component is missing a size;
- initialize the new ModelInfo::size field, ideally as `double size = 0.0`;
- accept numeric JSON sizes via is_number() rather than only is_number_float();
- nit: use `GB` instead of `Gb` in the column header.

Once those are addressed, this looks like a good CLI improvement.

GabrielReusRodriguez:
- I changed Gb -> GB.
- Now size is a double and it is initialized to 0.0
-  Now I use is_number() to validate numeric format on jsons. I also added get<double>() and get<string>() to "get" value of json.
-  I divided the logic in 2 functions one to get the size of 1 model and another function to get the sum of sizes (calling to the single model function ) 

image

fl0rianr:
Thanks, this addresses most of the review. One remaining edge case: get_size_of_collection() still returns a partial sum if a component is not found in server_models at all. Could we track whether each component was matched, and return UNKNOWN_SIZE if any component is missing?

My suggestion would be render UNKNOWN_SIZE as "N/A" in CLI like it's implemented in the UI.
And if at least one component has UNKNOWN_SIZE I would display it as e.g. ">=25.6 GB"

GabrielReusRodriguez:
Hello!

I made some changes.  Now it shows >XY.ZU  when one or more components are unknown.

image

or N/A when all components are unkonwn

image

fl0rianr (approved):
Thanks for iterating, fine by me now.

GabrielReusRodriguez:
Thanks for your review!

PR #2331, opened 2026-06-21
Fix: eviction_engine does not run  nvidia-smi on AMD configs anymore and some testing problems fixed.

GabrielReusRodriguez:
Hi!

I found some bugs on eviction engine.

First, my config is all AMD hardware and when I launched lemonade I saw this line repeated all time:

sh 1: nvidia-smi not found.

looking  at code, I saw at system_info.cpp  funtion  double SystemInfo::get_global_vram_usage_pct()  that it was always executing the call to nvidia-smi program so if somebody had amd config, it will not find this program.

image

I fixed  making a get_cuda_arch().empty()) call to  check if you have some nvidia arch on your config.  If so.. it launches the code.

Second, I tryed the server_eviction.py tests. but I found some errors.

- The Script loads a test model  phi-3-mini-4k-instruct-q4 that did not exists on lemonade so it failed. I changed by equivalent  Phi-4-mini-instruct-GGUF.
- The tests usually loaded 2 models but it did not changed the by default config so the server always had  max_models_loaded = 1 and failed when the second model was loaded.
- Finally at  def test_weight_factor_protects_model(self): test , i found  that it loads  the first model (with weight), sleeps 2 seconds and loads the second one. It always failed and unloaded the one with weight. Looking the code and doing some manual testing I found that when it loaded the second model, the first model was about 3 seconds at server and the seconds 3 ms. So the 1000 value of weight did not difference. The solution I thought was do another sleep after the  second model load and when I executed i found it passed the test!! it unload the second one .

image

I hope these changes help.

### DIFF
--- a/test/server_eviction.py
+++ b/test/server_eviction.py
@@ -6,6 +6,7 @@ import time
 from utils.server_base import ServerTestBase, pull_model_with_retry
 from utils.test_models import (
     ENDPOINT_TEST_MODEL,
+    SECOND_TEST_MODEL_EVICTION,
     TIMEOUT_MODEL_OPERATION,
     TIMEOUT_DEFAULT,
 )
@@ -19,7 +20,7 @@ class EvictionTests(ServerTestBase):
 
     _model_pulled = False
     _model2_pulled = False
-    MODEL2 = "Phi-4-mini-instruct-GGUF"
+    MODEL2 = SECOND_TEST_MODEL_EVICTION
 
     @classmethod
     def setUpClass(cls):
@@ -187,14 +188,14 @@ class EvictionTests(ServerTestBase):
 
         requests.post(
             f"{self.base_url.replace('/api/v1', '')}/internal/set",
-            json={"auto_evict": True, "auto_evict_threshold_pct": 0.90},
+            json={"auto_evict": True, "auto_evict_threshold_pct": 0.90, "max_models_loaded" : 2},
             headers=headers,
         )
 
         # Protected model loaded first (older) but with a large weight factor.
         requests.post(
             f"{self.base_url}/load",
-            json={"model_name": ENDPOINT_TEST_MODEL, "evict_weight_factor": 1000.0, "max_models_loaded" : 2},
+            json={"model_name": ENDPOINT_TEST_MODEL, "evict_weight_factor": 1000.0},
             timeout=TIMEOUT_MODEL_OPERATION,
         )
         time.sleep(2)
--- a/test/server_eviction.py
+++ b/test/server_eviction.py
@@ -188,7 +188,7 @@ class EvictionTests(ServerTestBase):
 
         requests.post(
             f"{self.base_url.replace('/api/v1', '')}/internal/set",
-            json={"auto_evict": True, "auto_evict_threshold_pct": 0.90, "max_models_loaded" : 2},
+            json={"auto_evict": True, "auto_evict_threshold_pct": 0.90, "max_loaded_models" : 2},
             headers=headers,
         )
 
--- a/test/server_eviction.py
+++ b/test/server_eviction.py
@@ -19,7 +19,7 @@ class EvictionTests(ServerTestBase):
 
     _model_pulled = False
     _model2_pulled = False
-    MODEL2 = "phi-3-mini-4k-instruct-q4"
+    MODEL2 = "Phi-4-mini-instruct-GGUF"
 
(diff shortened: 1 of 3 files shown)


test/server_eviction.py
```diff
@@ -194,7 +194,7 @@ def test_weight_factor_protects_model(self):
         # Protected model loaded first (older) but with a large weight factor.
         requests.post(
             f"{self.base_url}/load",
-            json={"model_name": ENDPOINT_TEST_MODEL, "evict_weight_factor": 1000.0},
+            json={"model_name": ENDPOINT_TEST_MODEL, "evict_weight_factor": 1000.0, "max_models_loaded" : 2},
```
fl0rianr: max_models_loaded does not look like a supported per-model /load option here. /load builds RecipeOptions from the request body, and RecipeOptions only includes eviction/model options such as auto_evict, evict_idle_timeout, downsize_idle_timeout, evict_weight_factor, and `pinned` — not max_models_loaded.

Since the router reads the slot limit from the global runtime config (max_loaded_models), this should probably be moved into the /internal/set call above as max_loaded_models: 2, matching the other test.

test/server_eviction.py
```diff
@@ -19,7 +19,7 @@ class EvictionTests(ServerTestBase):
 
     _model_pulled = False
     _model2_pulled = False
-    MODEL2 = "phi-3-mini-4k-instruct-q4"
+    MODEL2 = "Phi-4-mini-instruct-GGUF"
```
fl0rianr: Could we reuse one of the existing central test-model constants for this instead of hardcoding another model name here? The tests only need a distinguishable second model, and keeping these names centralized makes future registry/model-name changes less likely to break this test again.

fl0rianr (changes requested):
Thanks for the fix — the AMD/non-NVIDIA nvidia-smi spam and the eviction test failures are both worth addressing.

I left one inline comment on the test setup. One smaller note for the system_info.cpp change: guarding nvidia-smi with get_cuda_arch() fixes the AMD case, but it couples global NVIDIA VRAM reporting to Lemonade’s CUDA backend support detection. It may be safer to check whether nvidia-smi is available/usable, or suppress the command error and fall through silently.

Once the ignored max_models_loaded setup issue is fixed, this should be good from my side.

test/server_eviction.py
GabrielReusRodriguez: it's true. I think i got confused and i passed to load.
I will change as you told me.

GabrielReusRodriguez:
Hello, 

- I changed the max_models_loaded to the internal set endpoint
- Changed the Model to a var in test_models.py
- Changed the guard get_cuda_arch() for the function  find_executable_in_path("nvidia-smi")

I found that function that search the command in PATH and return "" if it does not see the command.  I can see the function is on path_utils , apply many validations and has implementation for linux, mac and windows.

I hace tested on ubuntu 24.04 without nvidia and it does not show the log. If I negate the condition ( it simulates that nvidia-smi exists on path ) then it show the log nvidia-smi not found.

I think it is the best option because it checks the command, and it is already done (Do not Repeat Yourself).

test/server_eviction.py
```diff
@@ -187,7 +188,7 @@ def test_weight_factor_protects_model(self):
 
         requests.post(
             f"{self.base_url.replace('/api/v1', '')}/internal/set",
-            json={"auto_evict": True, "auto_evict_threshold_pct": 0.90},
+            json={"auto_evict": True, "auto_evict_threshold_pct": 0.90, "max_models_loaded" : 2},
```
fl0rianr: This still uses the wrong config key. The global runtime setting is **max_loaded_models**, not max_models_loaded.

Since /internal/set validates config keys, this request should fail with an unknown-key error rather than applying the intended slot limit. Because the response is not asserted, the test may continue with the default single-model limit.

fl0rianr (changes requested):
Thanks for the update — the nvidia-smi guard and model constant look good now.

One small typo remains, see below:

test/server_eviction.py
GabrielReusRodriguez: changed the tetting ,  now it is:

json={"auto_evict": True, "auto_evict_threshold_pct": 0.90, "max_loaded_models" : 2},

fl0rianr (approved):
Thanks for the updates — this addresses my previous concerns.

Assuming CI is green, this looks good to me.

PR #2337, opened 2026-06-22
Better auto context size estimate

bitgamma:
Account for different model architectures to provide better estimate to maximize the automatic ctx_size value. While the estimates are still on the conservative size, they are much closer to real especially for Qwen3.5/3.6

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1787,3 +1787,22 @@ if(EXISTS "${_INSTALL_ATOMICITY_TEST_SRC}")
     include(CTest)
     add_test(NAME InstallAtomicityTest COMMAND test_install_atomicity)
 endif()
+
+# Auto-tune: GGUF array storage, scalar derivation, weighted KV cache computation.
+# Covers head_count_kv_per_layer, sliding_window_pattern, SWA precise weighted sum,
+# full_attention_interval exact count, and scalar fallback paths.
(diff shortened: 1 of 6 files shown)


fl0rianr (commented):
Thanks, this is a nice direction and the single-pass GGUF reader cleanup looks useful. I’m not fully comfortable approving yet because the new head_count_kv invariant seems order-dependent for scalar metadata, and the full_attention_interval scaling may be optimistic for non-divisible layer counts. Could you add a small test/fixture for these cases and compute scalar KV-head totals after parsing all relevant metadata?

bitgamma:
@fl0rianr yeah you are right, I will split parsing and calculation completely as it is cleaner and more robust

fl0rianr (commented):
Looks good overall — thanks for splitting parsing from derived calculations, that addresses my earlier concerns. One small robustness point: the precise SWA path assumes head_count_kv_per_layer and sliding_window_pattern have the same length, but both are parsed independently from GGUF metadata. Could we guard that path with an equal-size check, and fall back to the conservative scalar/proportional path on mismatch? Same applies to the debug-only SWA breakdown loop.

bitgamma:
@fl0rianr here you go

fl0rianr (approved):
Looks good to me now, thanks for addressing the metadata-ordering and array-length robustness cases.

PR #2354, opened 2026-06-22
fix(packaging): don't auto-enable the per-user lemond.service (Closes #2125)

### LINKED ISSUE #2125: Lemonade server conflicts with GDM dynamic user on Ubuntu 26.04, causing port 13305 to be occupied by gdm-greeter process
### Platform

Linux/Ubuntu

### Lemonade Version

10.6.0

### GPU / APU Model

_No response_

### Component

Other / Not sure

### Bug Description

Description
After installing lemonade-server on Ubuntu 26.04 (codename: resolute), the service cannot correctly bind to 0.0.0.0:13305 after reboot. Instead, the port is occupied by a lemond process running under the gdm-greeter dynamic user, causing the user's own lemond service to fail binding to the desired address.

Expected Behavior
After installation and configuration (e.g., lemonade config set host=0.0.0.0), the lemond service should listen on 0.0.0.0:13305 and persist across reboots.

No other lemond process should be running or occupying the port.

Actual Behavior
After reboot, port 13305 is occupied by a lemond process owned by gdm-greeter (a dynamic user created by GDM).

The user's main lemond service (lemonade-server) fails to bind to 0.0.0.0:13305 and falls back to 127.0.0.1:13305 (or fails entirely).

This issue repeats every reboot.

Environment
OS: Ubuntu 26.04 LTS (Resolute)

Kernel: 7.0.0-22-generic

Lemonade Server Version: 10.6.0~26.04

Desktop Environment: GNOME / GDM (default)

Installation Method: sudo apt install lemonade-server from official PPA

Root Cause Analysis (My Finding)
The issue is caused by the presence of /usr/lib/systemd/user/lemond.service. This file registers lemond as a user-level service for every user session, including the dynamic gdm-greeter user that GDM creates to run the login screen.

When GDM starts a dynamic gdm-greeter session, systemd automatically launches lemond under that user. This process reads a different configuration file (likely located in /run/gdm3/home/gdm-greeter/.cache/lemonade/config.json) and binds to 127.0.0.1:13305 by default.

This early launch:

Occupies port 13305 (or 127.0.0.1:13305)

Prevents the main lemonade-server service (which the user expects) from binding to 0.0.0.0:13305

Causes the main service to fail or fall back to 127.0.0.1

The main lemonade-server service (intended for the real user) is also a system service, but its port is preempted by the user-level instance from gdm-greeter.

Steps to Reproduce
Install Ubuntu 26.04 LTS with default GNOME/GDM

Install lemonade-server:

bash
sudo add-apt-repository ppa:lemonade-team/stable
sudo apt update
sudo apt install lemonade-server -y
Configure to listen on all interfaces:

bash
lemonade config set host=0.0.0.0
Reboot the system

Check port occupancy:

bash
sudo ss -tulnp | grep 13305
Observe that lemond is owned by gdm-greeter and listens on 127.0.0.1:13305

Temporary Workaround (for users)
To prevent gdm-greeter from launching lemond:

bash
sudo systemctl --machine=gdm-greeter@.host --user disable lemond.service
sudo systemctl --machine=gdm-greeter@.host --user mask lemond.service
And modify the main service to force the correct port:

bash
sudo systemctl edit --full lemond.service
# Change ExecStart line to:
ExecStart=/usr/bin/lemond --host 0.0.0.0 --port 13305
Proposed Fix
Remove or relocate the user-level service file:

/usr/lib/systemd/user/lemond.service should not exist, or should be disabled by default.

The main service should only exist in /usr/lib/systemd/system/lemond.service.

Alternatively, modify the user-level service to not auto-start, or check if the running environment is gdm-greeter and exit early.

Additional Context
GDM on Ubuntu 26.04 uses dynamic users for the login screen (the process runs as gdm-greeter).

Systemd user services are automatically launched for these dynamic users, which is the intended behavior, but it conflicts with services that expect to run only for real interactive users.

Similar issues may occur with other services that register both system and user units.

References
GDM dynamic user documentation (Ubuntu 26.04)

Systemd user service behavior

### Steps to Reproduce

Description
After installing lemonade-server on Ubuntu 26.04 (codename: resolute), the service cannot correctly bind to 0.0.0.0:13305 after reboot. Instead, the port is occupied by a lemond process running under the gdm-greeter dynamic user, causing the user's own lemond service to fail binding to the desired address.

Expected Behavior
After installation and configuration (e.g., lemonade config set host=0.0.0.0), the lemond service should listen on 0.0.0.0:13305 and persist across reboots.

No other lemond process should be running or occupying the port.

Actual Behavior
After reboot, port 13305 is occupied by a lemond process owned by gdm-greeter (a dynamic user created by GDM).

The user's main lemond service (lemonade-server) fails to bind to 0.0.0.0:13305 and falls back to 127.0.0.1:13305 (or fails entirely).

This issue repeats every reboot.

Environment
OS: Ubuntu 26.04 LTS (Resolute)

Kernel: 7.0.0-22-generic

Lemonade Server Version: 10.6.0~26.04

Desktop Environment: GNOME / GDM (default)

Installation Method: sudo apt install lemonade-server from official PPA

Root Cause Analysis (My Finding)
The issue is caused by the presence of /usr/lib/systemd/user/lemond.service. This file registers lemond as a user-level service for every user session, including the dynamic gdm-greeter user that GDM creates to run the login screen.

When GDM starts a dynamic gdm-greeter session, systemd automatically launches lemond under that user. This process reads a different configuration file (likely located in /run/gdm3/home/gdm-greeter/.cache/lemonade/config.json) and binds to 127.0.0.1:13305 by default.

This early launch:

Occupies port 13305 (or 127.0.0.1:13305)

Prevents the main lemonade-server service (which the user expects) from binding to 0.0.0.0:13305

Causes the main service to fail or fall back to 127.0.0.1

The main lemonade-server service (intended for the real user) is also a system service, but its port is preempted by the user-level instance from gdm-greeter.

Steps to Reproduce
Install Ubuntu 26.04 LTS with default GNOME/GDM

Install lemonade-server:

bash
sudo add-apt-repository ppa:lemonade-team/stable
sudo apt update
sudo apt install lemonade-server -y
Configure to listen on all interfaces:

bash
lemonade config set host=0.0.0.0
Reboot the system

Check port occupancy:

bash
sudo ss -tulnp | grep 13305
Observe that lemond is owned by gdm-greeter and listens on 127.0.0.1:13305

Temporary Workaround (for users)
To prevent gdm-greeter from launching lemond:

bash
sudo systemctl --machine=gdm-greeter@.host --user disable lemond.service
sudo systemctl --machine=gdm-greeter@.host --user mask lemond.service
And modify the main service to force the correct port:

bash
sudo systemctl edit --full lemond.service
# Change ExecStart line to:
ExecStart=/usr/bin/lemond --host 0.0.0.0 --port 13305
Proposed Fix
Remove or relocate the user-level service file:

/usr/lib/systemd/user/lemond.service should not exist, or should be disabled by default.

The main service should only exist in /usr/lib/systemd/system/lemond.service.

Alternatively, modify the user-level service to not auto-start, or check if the running environment is gdm-greeter and exit early.

Additional Context
GDM on Ubuntu 26.04 uses dynamic users for the login screen (the process runs as gdm-greeter).

Systemd user services are automatically launched for these dynamic users, which is the intended behavior, but it conflicts with services that expect to run only for real interactive users.

Similar issues may occur with other services that register both system and user units.

References
GDM dynamic user documentation (Ubuntu 26.04)

Systemd user service behavior

_No response_

### Log Output

```shell

```

### Additional Context

_No response_

ianbmacdonald:
## Problem

`dh_installsystemduser` enables user units globally by default, so the package enabled the per-user `lemond.service` (`WantedBy=default.target`) on top of the already-enabled system unit — and both bind `:13305`. systemd then starts a per-user `lemond` in *every* `--user` session:

- The **gdm-greeter** dynamic-user session reaches `default.target` at the login screen and binds `:13305` **before anyone logs in**, blocking the system service and the user's own session (the reported bug).
- **Any** other session — a desktop login or an SSH login — likewise spawns a redundant instance that can't serve, since it collides with the system service on `:13305`.

On one of our Ubuntu 26.04 servers this leaves **three** `lemond` processes running: the system service plus a per-session instance for each of two logged-in users (~18 MB each), only the system one able to serve.

## Reproduced (Ubuntu 26.04 + GDM)

Installed the package and rebooted; at the login screen, before any login, `lemond` was running as `gdm-greeter` (uid 60578) holding `127.0.0.1:13305` and `[::1]:13305`, with a second session's instance cycling on `Restart=on-failure`. With the fix (user unit not globally enabled) the greeter does not start it and `:13305` is free; the system service still serves.

## Fix

Debian DEB path only — the RPM postinst enables only the system unit, and the cpack DEB doesn't use `dh_installsystemduser`:

1. **`contrib/debian/rules`** — `dh_installsystemduser --no-enable`, so fresh installs ship the per-user unit present but disabled. Nothing is removed: the system service keeps serving `:13305` by default, and a user can still opt into a per-user instance with `systemctl --user enable --now lemond` (after freeing the port from the system unit). This complements #2173, which makes the user unit reliably available across install layouts.

2. **`contrib/debian/lemonade-server.postinst`** — `--no-enable` doesn't undo enablement left by an affected prior release, so on upgrade from a pre-fix version a version-gated, one-shot migration disables the leftover global user-unit enable **and** stops any per-user instance still holding the port (`deb-systemd-invoke --user stop`), so the conflict clears without a reboot. A user's deliberate later enable is preserved.

> **Maintainer note:** the migration's version threshold (`10.8.1~`) is a placeholder — please set it to the first release that ships this change.

Closes #2125.

### DIFF
--- a/contrib/debian/lemonade-server.postinst
+++ b/contrib/debian/lemonade-server.postinst
@@ -30,4 +30,15 @@ if [ "$1" = "configure" ]; then
 	fi
 fi
 
+# #2125: migrate pre-fix upgrades — disable per-user lemond.service and stop
+# any still-running instance that conflicts with the system service.
+if [ "$1" = "configure" ] && [ -n "$2" ] && dpkg --compare-versions "$2" lt "10.8.1~"; then
+	if command -v deb-systemd-helper > /dev/null 2>&1; then
+		deb-systemd-helper --user disable lemond.service > /dev/null 2>&1 || true
+	fi
+	if [ -d /run/systemd/system ] && command -v deb-systemd-invoke > /dev/null 2>&1; then
+		deb-systemd-invoke --user stop lemond.service > /dev/null 2>&1 || true
+	fi
+fi
+
 #DEBHELPER#
--- a/contrib/debian/rules
+++ b/contrib/debian/rules
@@ -13,3 +13,8 @@ override_dh_auto_test:
 
 override_dh_install:
 	dh_install
+
+# #2125: ship per-user lemond.service DISABLED to avoid port conflicts with
+# the system service in every --user session.
+override_dh_installsystemduser:
+	dh_installsystemduser --no-enable


superm1 (commented):
I'm aligned with the change, but I don't need an essay in comments.  Please pare down the comments to either 1-2 lines or ideally nothing.

ianbmacdonald:
Quick scope note: the fresh-install fix is just the `--no-enable` in `debian/rules`. The postinst migration (disable + stop the leftover per-user instance on upgrade) is there so an upgrade from an affected release doesn't carry the conflict forward — but it's the more opinionated part, so it's your call. If you'd rather cover existing installs with a NEWS/changelog note than an active maintscript, I'm happy to trim this to just `--no-enable`. And the `10.8.1~` threshold is a placeholder — set it to whatever release this lands in (or let me know and I will).

ianbmacdonald:
> I'm aligned with the change, but I don't need an essay in comments. Please pare down the comments to either 1-2 lines or ideally nothing.

thanks for the note .. helps a lot on my end to manage expectations

superm1 (approved):

PR #2365, opened 2026-06-22
test/ci: harden CLI endpoint suites against shared-server flakes

fl0rianr:
## Summary

Harden the bundled CLI/endpoint/Ollama CI suites against shared-server flakes.

These jobs intentionally run multiple test scripts against one long-lived Lemonade server to reduce setup time, queue pressure, and model download churn. That makes the suite vulnerable to state leaking between scripts: loaded models, runtime host changes, stale backend processes, and missing diagnostics when a later script times out.

This PR adds explicit reset/diagnostic handling around those shared-server boundaries.

## Changes

- Add test/utils/reset_server_state.py
  - verifies the server is reachable
  - restores runtime host binding to localhost
  - unloads all loaded models
  - prints /health and Ollama /api/ps diagnostics
- Call the reset helper between bundled CI test scripts before CLI, endpoint, webapp, Ollama, and streaming-error suites.
- Make server_cli2.py::test_043_listen_all_via_runtime_config restore host=localhost in a finally block.
- Harden test_ollama.py::test_026_anthropic_messages_tool_calling
  - starts from an unloaded model state
  - unloads after the test
  - prints useful diagnostics on timeout
- Extend server log capture to include $RUNNER_TEMP/lemond*.log, which is important for the macOS unsigned build path.

## Motivation

Recent CI failures were not consistently tied to one platform or one code path. The failures appeared in the CLI/Endpoints/Ollama area and could move between macOS and Windows, while reruns could pass. That points to shared-server CI flakiness rather than a product regression in the PR under test.

The highest-risk pattern is that multiple suites run sequentially against the same server. If one suite mutates runtime config, leaves a model loaded, or times out while a backend process is still active, the next suite can fail in a misleading way.

### DIFF
--- a/.github/actions/capture-server-logs/action.yml
+++ b/.github/actions/capture-server-logs/action.yml
@@ -81,7 +81,7 @@ runs:
           fi
         fi
 
-        for f in "$RUNNER_TEMP"/lemonade*.log; do
+        for f in "$RUNNER_TEMP"/lemonade*.log "$RUNNER_TEMP"/lemond*.log; do
           if [ -f "$f" ]; then
             echo "--- $f ---"
             tail -50 "$f"
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -630,7 +630,7 @@ jobs:
         run: |
           set -e
           echo "Resetting server state before CLI tests..."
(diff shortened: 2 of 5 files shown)


kenvandine:
@claude please review

kenvandine:
@claude please review this again

jeremyfowers (approved):
Thanks for making these improvements @fl0rianr !

PR #2366, opened 2026-06-22
Install ROCm stable runtime from wheels instead of tarballs

superm1:
ROCm/TheRock stops publishing the prebuilt distribution tarballs in the next release, but continues shipping the runtime as per-arch Python wheels. Switch new ROCm-stable installs to create a venv and pip install the per-arch "rocm[libraries]" wheels for the detected GPU. A version-matching legacy tarball install is detected and reused so users don't re-download.

Library directories are resolved by globbing the venv site-packages rather than via "rocm-sdk path", which would require the unneeded rocm[devel] package. Adds python3-venv to the lemonade-server Debian depends.

### DIFF
--- a/src/cpp/include/lemon/backends/backend_utils.h
+++ b/src/cpp/include/lemon/backends/backend_utils.h
@@ -99,18 +99,27 @@ namespace lemon::backends {
         /** Check if ROCm libraries are installed system-wide (Linux only) */
         static bool is_rocm_installed_system_wide();
 
-        /** Get TheRock installation directory for a specific architecture and version */
+        /** Get the TheRock installation directory for a specific architecture and version */
         static std::string get_therock_install_dir(const std::string& arch, const std::string& version);
 
-        /** Download and install TheRock ROCm tarball for the specified architecture (Linux only) */
+        /** Install the TheRock ROCm runtime (Linux/Windows only): pip install the per-arch
+         *  ROCm wheels into a venv. A version-matching legacy tarball install is reused. */
         static void install_therock(const std::string& arch, const std::string& version,
                                    DownloadProgressCallback progress_cb = nullptr);
 
-        /** Clean up old TheRock versions, keeping only the specified version */
+        /** Clean up old TheRock versions (tarball and venv), keeping only the specified version */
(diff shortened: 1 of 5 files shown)


src/cpp/include/lemon/backends/backend_utils.h
```diff
@@ -99,19 +99,28 @@ namespace lemon::backends {
...
+        /** Clean up old TheRock versions (tarball and venv), keeping only the specified version */
         static void cleanup_old_therock_versions(const std::string& current_version);
 
         /** Get TheRock lib directory path if available, or empty string if not needed */
         static std::string get_therock_lib_path(const std::string& rocm_arch);
 
+        /**
+         * Locate a Python interpreter for creating the ROCm venv.
+         * Linux/macOS: "python3" then "python" from PATH.
+         * Windows: "py -3" launcher, then "python".
```
jeremyfowers, line 119: Fyi Windows doesn't have Python pre-installed. The only reliable way to do it is to download embedded Python at runtime, which is what the old lemonade python installer used to do.

superm1: That matches what [ROCm documentation for Windows](https://rocm.docs.amd.com/en/7.13.0-preview/install/rocm.html?fam=ryzen&w=compute&os=windows&i=pip&gpu=max-pro-395&gfx=gfx1151&windows-ver=11#install-python-i-pip-os-ubuntu-ubuntu-ver-26-04-ubuntu-ver-24-04-ubuntu-ver-22-04-os-debian-debian-ver-13-debian-ver-12-os-rhel-rhel-ver-10-1-rhel-ver-10-0-rhel-ver-9-7-rhel-ver-9-6-rhel-ver-9-4-rhel-ver-9-rhel-ver-8-10-os-oracle-linux-oracle-linux-ver-10-oracle-linux-ver-9-oracle-linux-ver-8-os-rocky-linux-os-sles-sles-ver-16-0-sles-ver-15-7-os-windows) talks about too.

I suppose if we stick this direction, we'll need something like that added for Windows.

superm1:
superceeded by #2768

PR #2369, opened 2026-06-23
Handle separate sd-cpp CUDA release pins

fl0rianr:
**Fixed: CUDA update corrected**

This follow-up fixes the sd-cpp updater to #2139 

The non-CUDA sd-cpp backends are resolved from `leejet/stable-diffusion.cpp`:

* cpu
* vulkan
* rocm-stable
* metal

The CUDA backend however is resolved from `lemonade-sdk/stable-diffusion.cpp`, because CUDA assets are published from the Lemonade fork and do not need to share the same release tag as the upstream leejet assets.
This is fix 1.

**Improved: Validation evidence:**

  * per-image request timings
  * untimed warm-up before recorded timings
  * generated PNGs committed to the update branch
  * generated PR body renders the validation images inline


kenvandine:
@claude please review

fl0rianr:
I adjusted this again to match the intended validation scope.

Each expected image is a test result, so the workflow now validates both configured models and both configured sizes for every validated backend:

* SD-Turbo-GGUF
* Flux-2-Klein-4B
* 512x256
* 1024x1024

That gives the current 12-image evidence set for cpu, vulkan, and rocm-stable.

I removed the fixed SDCPP_EXPECTED_PR_IMAGES=12 check, but did not make missing images optional. The renderer now checks the expected matrix directly: every validated backend/model/size combination must have a passing JSON record and a copied PNG. If one expected image is missing, PR body generation fails with the missing combination listed.

CUDA asset validation remains a deliberate hard failure, but it runs after the normal validation matrix so we still get CPU/Vulkan/ROCm evidence before failing on a CUDA asset problem.

kenvandine (approved):
Thanks for this, it looks good. This will make a great addition!

Issue #1937, opened 2026-05-17
BYOB (Bring Your Own Backend)


sawansri:
Lemonade should introduce a way to point to existing running inference backends via OpenAI API.

This would allow for easy interopability with @kyuz0's toolboxes and any other engine not officially integrated with lemonade like [hipfire](https://github.com/Kaden-Schutt/hipfire), [hipEngine](https://github.com/shisa-ai/hipEngine/), etc.

Thanks to @zhelgadis on Discord for the idea.

jeremyfowers:
@ramkrishna2910's https://github.com/lemonade-sdk/lemonade/pull/1785 is marketed as a cloud backend, but really it lets you plug in any OpenAI API base url regardless of whether it is cloud or local. That would solve this issue, right?

@ramkrishna2910 what do you think of broadening scope to make it "external base url" instead of "cloud base url"?

@sawansri please discuss with @ramkrishna2910 before implementing anything new.

ramkrishna2910:
Yeah the same implementation now can expand scope to any external url. @sawansri please take a look at the PR attached above and lets discuss.

sawansri:
> @ramkrishna2910's https://github.com/lemonade-sdk/lemonade/pull/1785 is marketed as a cloud backend, but really it lets you plug in any OpenAI API base url regardless of whether it is cloud or local. That would solve this issue, right?
> 
> @ramkrishna2910 what do you think of broadening scope to make it "external base url" instead of "cloud base url"?
> 
> @sawansri please discuss with @ramkrishna2910 before implementing anything new.

Yeah we could broaden the scope of the cloud backend to include any backend reachable via OpenAI API, that would accomplish the same thing.

jeremyfowers:
Nice! Love to get two birds with one stone. Any suggestions for naming the new WrappedServer, @sawansri ?

"Cloud" is too narrow. "BYOB" is a bit informal but kinda funny since we're lemonade.

jeremyfowers:
PS. this will be great for testing too, people can prototype integration of Hipfire or DS4 or whatever they like without writing a single line of code.

sawansri:
> Nice! Love to get two birds with one stone. Any suggestions for naming the new WrappedServer, @sawansri ?
> 
> "Cloud" is too narrow. "BYOB" is a bit informal but kinda funny since we're lemonade.

[...]

PR #2370, opened 2026-06-23
Allow HTTP custom backends with warnings

sawansri:
Unblocks https://github.com/lemonade-sdk/lemonade/issues/1937#issuecomment-4775122225

This PR enables users to add custom/cloud backends with http with a warning.

image

### DIFF
--- a/src/app/src/renderer/BackendManager.tsx
+++ b/src/app/src/renderer/BackendManager.tsx
@@ -34,9 +34,10 @@ interface BackendManagerProps {
   searchQuery: string;
   showError: (msg: string) => void;
   showSuccess: (msg: string) => void;
+  showWarning: (msg: string) => void;
 }
 
-const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError, showSuccess }) => {
+const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError, showSuccess, showWarning }) => {
   const { systemInfo, isLoading, refresh } = useSystem();
   const [backendAssetSizes, setBackendAssetSizes] = useState<Record<string, number>>({});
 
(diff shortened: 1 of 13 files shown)


ramkrishna2910:
@claude review for security implications

ramkrishna2910:
@fl0rianr can you please take a look at this. The plaintext credential exposure is making me uncomfortable.

sawansri:
We can leave http as an opt in feature. By default enforce https. Preventing http is a blocker for lan setups but I understand the concern with not enforcing https for cloud providers especially with API keys.

fl0rianr (changes requested):
Thanks for the implementation — the warning plumbing looks solid across server, CLI, UI, and tests.

I agree with the security review’s main concern: allowing keyless http:// custom/LAN backends is the important unblocker here, but accepting http:// + API key with only a post-success warning is a weaker posture than I’m comfortable approving as-is.

My suggested bar for merge:

* Keep keyless http:// custom backends allowed.
* Require explicit opt-in for http:// providers when an API key would be stored or used, e.g. an `allow_insecure_http: true` request flag / CLI flag / UI confirmation.
* Apply that check both to install-with-key and later /cloud/auth.
* Add tests for the rejected-by-default and accepted-with-explicit-opt-in paths.
* Minor: while touching validation, reject bare `http://` / `https://` with no host.

Once HTTP+key requires explicit user intent, this looks good to me.

sawansri:
Added an explicit opt in for http + api key combo
image

Also added a cli option (` --allow-insecure-http`) to allow this via cli as well as tests.

fl0rianr (approved):
Thanks for the update — this addresses my previous concern.

I only noticed one minor non-blocking polish item: http://:8080 -style malformed hosts may still pass the current host-presence check. Not worth blocking this PR on, but it could be tightened while touching URL validation.

Looks good to me now.

PR #2373, opened 2026-06-23
feat(cli): add native HTTPS/TLS and URL scheme parsing support

abn:
This PR adds native HTTPS/TLS negotiation and URL scheme parsing to the C++ CLI client (`lemonade`). This allows the CLI to seamlessly connect to secure remote endpoints (e.g., platforms like Modal, Cloudflare tunnels, or reverse proxies) equivalent to the desktop client.

## Key Changes
* **HTTPS/TLS Transport**: Integrated OpenSSL (`libssl`/`libcrypto`) into the build pipeline and compiled the CLI target with `CPPHTTPLIB_OPENSSL_SUPPORT` enabled.
* **URL Scheme Parser**: Added a portable URL parser (`ParseTargetUrl`) to strip schemes (`https://`, `http://`), clean trailing paths/slashes, and automatically default missing ports (`443` for HTTPS, `80` for HTTP) while respecting explicit port overrides.
* **Client Integration**: Updated the C++ CLI client constructor and command entry points to dynamically initialize secure or insecure transports, and enabled secure URL browser launching (`open_url`).
* **Offline-safe Unit Tests**: Added automated test coverage (`CLIUrlSchemeTests` in `test/server_cli2.py`) using a local mock HTTP server on an ephemeral port. This verifies HTTP/HTTPS routing and TLS handshake attempts deterministically without relying on external internet access.

## Verification
* **HTTPS Target**: `LEMONADE_HOST=https://httpbin.org lemonade list` correctly defaults to port 443, resolves the domain, negotiates TLS, and completes the connection.
* **HTTP Target**: `LEMONADE_HOST=http://httpbin.org lemonade list` correctly routes to port 80.
* **Local Fallback**: `lemonade list` without a host successfully preserves the fallback to `127.0.0.1:13305`.
* **Automated Suite**: `PYTHONPATH=test python -m unittest server_cli2.CLIUrlSchemeTests` -> **PASS (2/2 tests)**.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,7 +343,7 @@ endif()
 if(NOT USE_SYSTEM_HTTPLIB)
     FetchContent_Declare(httplib
         GIT_REPOSITORY https://github.com/yhirose/cpp-httplib.git
-        GIT_TAG 89c932f313c6437c38f2982869beacc89c2f2246  # v0.26.0
+        GIT_TAG fe332fa06bac76a1c6d402c08f414052999347da  # v0.47.0
         GIT_SHALLOW TRUE
     )
     set(HTTPLIB_REQUIRE_OPENSSL OFF CACHE INTERNAL "")
--- a/src/cpp/cli/CMakeLists.txt
+++ b/src/cpp/cli/CMakeLists.txt
@@ -13,43 +13,45 @@
 if(NOT TARGET lemonade-digest-crypto)
     find_package(PkgConfig QUIET)
     if(PkgConfig_FOUND)
-        pkg_check_modules(MBEDCRYPTO QUIET mbedcrypto)
+        pkg_check_modules(MBEDTLS QUIET mbedtls mbedx509 mbedcrypto)
     endif()
 
     add_library(lemonade-digest-crypto INTERFACE)
 
-    if(MBEDCRYPTO_FOUND)
-        message(STATUS "Using system mbedcrypto for download digest verification")
-        if(MBEDCRYPTO_INCLUDE_DIRS)
-            target_include_directories(lemonade-digest-crypto INTERFACE ${MBEDCRYPTO_INCLUDE_DIRS})
+    if(MBEDTLS_FOUND)
+        message(STATUS "Using system mbedtls for CLI HTTPS and digest verification")
+        if(MBEDTLS_INCLUDE_DIRS)
+            target_include_directories(lemonade-digest-crypto INTERFACE ${MBEDTLS_INCLUDE_DIRS})
         endif()
-        if(MBEDCRYPTO_LIBRARY_DIRS)
-            target_link_directories(lemonade-digest-crypto INTERFACE ${MBEDCRYPTO_LIBRARY_DIRS})
+        if(MBEDTLS_LIBRARY_DIRS)
+            target_link_directories(lemonade-digest-crypto INTERFACE ${MBEDTLS_LIBRARY_DIRS})
         endif()
-        target_link_libraries(lemonade-digest-crypto INTERFACE ${MBEDCRYPTO_LIBRARIES})
-        if(MBEDCRYPTO_CFLAGS_OTHER)
-            target_compile_options(lemonade-digest-crypto INTERFACE ${MBEDCRYPTO_CFLAGS_OTHER})
+        target_link_libraries(lemonade-digest-crypto INTERFACE ${MBEDTLS_LIBRARIES})
+        if(MBEDTLS_CFLAGS_OTHER)
+            target_compile_options(lemonade-digest-crypto INTERFACE ${MBEDTLS_CFLAGS_OTHER})
         endif()
     else()
-        # Some distro packages do not ship a pkg-config file for mbedcrypto.
+        # Some distro packages do not ship a pkg-config file for mbedtls.
         # Keep the system-library path working for Debian/PPA builds, where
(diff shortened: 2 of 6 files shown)


src/cpp/cli/CMakeLists.txt
```diff
@@ -177,7 +179,7 @@ if(WIN32)
     )
 endif()
 
-target_compile_definitions(lemonade PRIVATE LEMONADE_CLI)
+target_compile_definitions(lemonade PRIVATE LEMONADE_CLI CPPHTTPLIB_OPENSSL_SUPPORT)
```
bitgamma: since we already use MBEDTLS for digest verification, I'd rather not add another dependency and use `CPPHTTPLIB_MBEDTLS_SUPPORT` instead

src/cpp/cli/lemonade_client.cpp
```diff
@@ -81,8 +81,57 @@ const std::string& HttpError::response_body() const {
     return response_body_;
 }
 
-LemonadeClient::LemonadeClient(const std::string& host, int port, const std::string& api_key)
-    : host_(host), port_(port), api_key_(api_key) {}
+void ParseTargetUrl(const std::string& input_host, std::string& out_clean_host, int& out_port, bool& out_is_ssl) {
```
bitgamma: please let's use regular function name convention (lower_case). Also this can be a static function I think?

src/cpp/cli/lemonade_client.cpp
```diff
@@ -94,9 +143,16 @@ std::string LemonadeClient::normalize_host(const std::string& host) const {
 }
 
 // Helper to create and configure httplib::Client (timeouts in milliseconds)
-static httplib::Client make_client(const std::string& host, int port, const std::string& api_key,
+static httplib::Client make_client(const std::string& host, int port, const std::string& api_key, bool is_ssl,
                                     time_t connection_timeout_ms = DEFAULT_CONNECTION_TIMEOUT_MS, time_t read_timeout_ms = DEFAULT_READ_TIMEOUT_MS) {
-    httplib::Client cli(host, port);
+#ifndef CPPHTTPLIB_OPENSSL_SUPPORT
```
bitgamma: CPPHTTPLIB_MBEDTLS_SUPPORT

CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ set(MIN_HTTPLIB_VERSION "0.26.0")
 set(MIN_LWS_VERSION "4.3.3")
 
 find_package(PkgConfig QUIET)
+find_package(OpenSSL REQUIRED)
```
bitgamma: this won't be needed if using MbedTLS

bitgamma (changes requested):
Thanks for your PR! Before we can merge this, there are a few things to address, please the comments below.

abn:
Thanks for the review! I've updated the PR to address your feedback:

* **Switched to MbedTLS**: Replaced the OpenSSL dependency with `mbedtls`, `mbedx509`, and `mbedcrypto` and compiled the client with `-DCPPHTTPLIB_MBEDTLS_SUPPORT`.
* **Upgraded `cpp-httplib`**: Upgraded the parent build configuration to support `cpp-httplib` version `0.26.0` or higher to allow offline packaging builds (e.g. against Ubuntu 24.04 and Debian 13's system headers) while keeping the FetchContent fallback tag at `v0.47.0` for local builds.
* **Cleaned up `ParseTargetUrl`**: Declared it as a `static` member function of `LemonadeClient` named `parse_target_url` to respect style conventions and avoid namespace pollution.
* **Rebased**: Rebased the branch on the latest `main` branch.

bitgamma:
@abn thanks for the followup. I see some CI failures that do not appear to be flaky tests. Can you please look into it?

abn:
@bitgamma so it seems the issue is really the minimum versions.

### Why the CI Failed:
The workflow run failed during the `Build Lemonade .deb Package` job (on Ubuntu 24.04). 
1. The Ubuntu 24.04 build container has `libcpp-httplib-dev` version **`0.26.0`** pre-installed.
2. Because we set `MIN_HTTPLIB_VERSION` to `0.41.0`, CMake rejected the container's version and tried to run FetchContent.
3. Since Debian/Ubuntu packaging builds run in a fully disconnected environment (`FETCHCONTENT_FULLY_DISCONNECTED=ON`), the download was blocked, resulting in a missing `httplib::httplib` target and failing the build.

### How it was fixed:
1. **Restored the minimum version constraint** in the root `CMakeLists.txt` back to `0.26.0` so that offline builds on Ubuntu 24.04 can successfully find and use the system package.
2. **Added a version check in `src/cpp/cli/CMakeLists.txt`** to ensure that `CPPHTTPLIB_MBEDTLS_SUPPORT` is only defined if the available headers are `0.31.0` or newer (when MbedTLS support was introduced).
3. **Added compile-time checks in `src/cpp/cli/lemonade_client.cpp`** to gracefully fall back and throw `'https' scheme is not supported.` if compiled against a system header older than `0.31.0` (like Ubuntu 24.04's `0.26.0`), while keeping it fully enabled on newer system headers (like Debian 13's `0.41.0`) and local FetchContent builds (`0.47.0`).

PS: I am hoping that the system packages are still receiving security backports.

bitgamma:
@abn I see the new https related tests are failing now

abn:
Ugg; forgot to relax the tests :sweat_smile:

bitgamma (approved):

PR #2393, opened 2026-06-23
fix(backends): fallback to installed backend when offline

### LINKED ISSUE #2078: Offline Usage of Lemonade
### Platform

Linux/Ubuntu

### Lemonade Version

10.6.0

### GPU / APU Model

_No response_

### Component

llama.cpp

### Bug Description

Hi,
I had no internet and thought of using lemonade, but could not load a model.

### Steps to Reproduce

I tried to load a model via https://localhost:13305 It tried to download newest version of llama.cpp. Since I had no internet the model would not load.
I guess i had an older version installed, lemonade knew already of a newer llama version, then when internet was offline it tried to install it.

This might be a bit of an edge case, since there needs to be a new llama version which has not been downloaded yet, but it would be nice to have a fallback to the currently installed llama version - if available.

### Log Output

```shell

```

### Additional Context

_No response_

fl0rianr:
Fixes #2078.

This keeps already-installed backends usable when Lemonade cannot reach GitHub during an automatic backend update check. Model loading now falls back to the installed backend instead of failing after repeated download retries.

Explicit backend installs/updates still attempt the update when online, and the existing backend is kept until the replacement has been fully downloaded, extracted, verified, and swapped into place.

### DIFF
--- a/src/cpp/server/backend_manager.cpp
+++ b/src/cpp/server/backend_manager.cpp
@@ -149,6 +149,10 @@ void report_backend_ready(const std::string& recipe,
     progress_cb(p);
 }
 
+bool github_download_service_reachable() {
+    return utils::HttpClient::is_reachable("https://github.com/", 2);
+}
+
 bool has_matching_system_rocm_runtime(const std::string& expected_runtime_version) {
     const fs::path version_file = "/opt/rocm/.info/version";
     const std::string system_version = read_version_file(version_file);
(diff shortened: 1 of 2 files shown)


src/cpp/server/backend_manager.cpp
```diff
@@ -495,6 +549,50 @@ void BackendManager::install_backend(const std::string& recipe, const std::strin
...
+                    (offline ? "Cannot install " : "Fetching executable artifacts is disabled and ") +
+                    recipe + ":" + resolved_backend +
+                    (offline ? ": offline mode" : " is not installed"));
+            }
+
+            throw std::runtime_error(
+                (offline ? "Cannot install " : "Fetching executable artifacts is disabled for ") +
+                recipe + ":" + resolved_backend +
+                ": required runtime components are not installed");
+        }
```
superm1: I would just combine these.  I don't think the nuance matters between the two.

superm1 (commented):
I don't really agree with the premise of this PR.  Right now the ABI might be the same but what if they had ROCm 7.13 and were transitioning to ROCm 8?  The ABI might be different and the old backend wouldn't work.  I think it's a path we really shouldn't be supporting.

bong-water-water-bong (commented):
The core idea is sound and mostly correct. Two issues worth addressing:

**1. Level C catch block is too broad and doesn't re-check `needs_therock_download`:** When `install_from_github` fails and the catch block fires, it returns "ready" using the installed backend binary — but doesn't verify that TheRock runtime components are present. If GitHub is unreachable and TheRock is missing/outdated, the loaded model will fail at runtime.

**2. Level C catches ALL `std::exception`:** Disk-full errors, permission denials, and corrupted archives are all silently swallowed. Consider narrowing to network-related exceptions only (CURLE_COULDNT_CONNECT, etc.) or at least logging a warning with the actual error before falling back.

**Minor:** The `backend_update_was_deferred` variable name is misleading — the update was attempted, not deferred. And the GitHub reachability probe has a small TOCTOU gap with `resolve_user_version` throws.

Otherwise the layered defense approach (Level A → B → C → D) is well-structured.

src/cpp/server/backend_manager.cpp
fl0rianr: Good point — I adjusted the approach so this no longer disables the normal online update-before-load behavior.

The installed backend is only reused for the automatic load/ensure path when fetching is explicitly disabled/offline or GitHub is not reachable. When online, the existing install path still resolves the target version and runs the normal update flow before loading the model.

So we avoid supporting a stale ABI as the normal path, while still allowing already-installed backends to remain usable when the machine cannot reach GitHub.

superm1 (approved):

PR #2395, opened 2026-06-23
Move omni collection system prompt to per-model JSON

### LINKED ISSUE #2367: Move omni collection system prompt from app-global to per-model JSON definition
## Summary

Move the omni/collection **system prompt** from the global, app-side definition into each collection model's own JSON definition (e.g. the published HF file like [`LMX-Omni-52B-Halo.json`](https://huggingface.co/lemonade-sdk/LMX-Omni-52B-Halo/blob/main/LMX-Omni-52B-Halo.json)).

**Scope is intentionally narrow:** only the system prompt becomes per-model. **Tool definitions stay global** in `toolDefinitions.json` — making tools per-model is a much larger, separate change.

## Current state

The omni system prompt is **global, not per-model**. It lives in one file — `src/app/src/renderer/utils/toolDefinitions.json` (`system_prompt` field) — staged into the server's `resources/` at build time. Two consumers read it:

- **App / client-side path:** `lemonadeTools.ts → buildLemonadeTools()` builds it; `LLMChatPanel.tsx` prepends it.
- **Server-side path:** `collection_orchestrator.cpp` reads `tool_definitions()["system_prompt"]`.

Both perform the same `{tool_list}` / `{tool_guidance}` substitution at runtime based on which component models (by label) are present — so the stored prompt is a **template**, not a finished string.

The per-model collection JSON's allowed keys are pinned in two mirrored allowlists (`kKnownKeys` in `recipe_import.cpp`, `EXPORT_KNOWN_KEYS` in `modelData.ts`). There is **no `system_prompt` field today**, and any unknown key is stripped on import.

## Design

Add an optional per-model `system_prompt` field that **overrides** the global default, with the global kept as **fallback** (backward-compatible for existing/imported collections that lack the field).

- The per-model value must remain a **template** containing `{tool_list}` / `{tool_guidance}` so label-based substitution still works.
- Only the prompt *text* moves per-model; tool definitions remain global.

## Touch points

1. **Allowlists (keep in sync):** add `system_prompt` to `kKnownKeys` (`src/cpp/cli/recipe_import.cpp`) and `EXPORT_KNOWN_KEYS` (`src/app/src/renderer/utils/modelData.ts`), or the field is dropped on import/export round-trips.
2. **Server orchestrator** (`src/cpp/server/collection_orchestrator.cpp`): look up the collection model's registry entry; use its `system_prompt` if present, else fall back to the global `tool_definitions()["system_prompt"]`. Substitution logic unchanged.
3. **App** (`src/app/src/renderer/utils/lemonadeTools.ts`): `modelsData[collectionName].system_prompt ?? toolDefinitions.system_prompt`.
4. **Type:** add `system_prompt?: string` to `ModelInfo` (`modelData.ts`).
5. **Registry + HF files:** add the field to `LMX-Omni-52B-Halo` / `LMX-Omni-5.5B-Lite` in `server_models.json` and to the published HF JSONs.
6. **Optional — custom collection UI** (`CustomCollectionPanel.tsx` / `customCollections.ts`): let users author a custom prompt when building their own omni model.
7. **Docs + tests:** `docs/dev/lemonade-omni.md`, `docs/guide/configuration/custom-models.md`; `test/server_omni.py`, `test/app/app-regression/lemonadeTools.test.cjs`, and export/import round-trip tests.

## Notes / decisions

- **Override vs. fallback:** per-model overrides global; global stays as default (only backward-compatible option).
- **Trust boundary:** once `system_prompt` is importable, pulling a third-party HF collection means accepting its system prompt as attacker-influenceable data. It's just a prompt, but it now travels with the model — worth a conscious nod.

ramkrishna2910:
## Summary

Closes #2367.

The below example shows creating a new omni collection with a modified system prompt and that prompt being honored during execution.

image

image

Moves the omni system prompt from the global `toolDefinitions.json` into each collection model's own JSON definition. The per-collection value overrides; the global default stays as the fallback so existing/imported collections without the field keep working unchanged. Tool definitions remain global — only the prompt text moved.

- **Server:** `ModelInfo` gains an optional `system_prompt`; built-in, user, and on-demand load paths all parse it. For HF-backed collections (`LMX-Omni-*`) the field is lifted from the cached manifest so the published HF JSON is the source of truth — no code change is needed when that text is later customized. `CollectionOrchestrator::build_tools` prefers the per-collection template; `GET /v1/models{,/{id}}` surfaces it when set.
- **App:** `buildLemonadeTools` mirrors the server fallback. `ModelInfo` (TS), `EXPORT_KNOWN_KEYS`, and the CLI's `kKnownKeys` all gain `system_prompt` so export/import round-trips don't drop it. The Custom Omni Model editor adds a "System Prompt (optional)" textarea threaded through `CustomCollectionDraft` → `CustomCollectionPullRequest`. `USER_DEFINED_MODEL_PROPS` carries it across `/pull` so user.* registrations persist it.
- **Docs:** `docs/dev/lemonade-omni.md` and `docs/guide/configuration/custom-models.md` document the override.
- **Tests:** 3 new behavioral tests in `customCollections.test.cjs` (pull-request shape, ModelInfo round-trip, export normalize), 3 static-source tests in `lemonadeTools.test.cjs`, and a wire round-trip test `test_021j_register_user_collection_with_system_prompt` in `server_endpoints.py`.
- **HF (already published):** The `system_prompt` field was added to `lemonade-sdk/LMX-Omni-52B-Halo.json` and `lemonade-sdk/LMX-Omni-5.5B-Lite.json` with the current default text — byte-identical no-op today, but reserves the canonical slot for future per-collection customization without a client update.

### DIFF
--- a/docs/dev/lemonade-omni.md
+++ b/docs/dev/lemonade-omni.md
@@ -40,6 +40,12 @@ The canonical definitions live in [`src/app/src/renderer/utils/toolDefinitions.j
 
 Endpoint request/response shapes are documented in the [Endpoints Spec](../api/README.md).
 
+### Per-collection system prompt override
+
+The default OmniRouter system prompt lives in `toolDefinitions.json` and is a *template* with `{tool_list}` and `{tool_guidance}` placeholders that the server (and app) expand at runtime based on which components are present.
+
+Each collection model can override that default by setting its own `system_prompt` template — either in its registry entry (`server_models.json` for built-ins), in its published Hugging Face collection JSON (for HF-backed collections like `LMX-Omni-*`), or via the desktop app's Omni Model editor for custom collections. When present, the per-collection value wins; otherwise the global default applies. Tool definitions stay global — only the prompt text is per-collection.
+
 ## How to Use Omni Models
 
 Any app can use an omni collection by simply requesting `/chat/completions` and receiving multi-media results in the response content. Apps that want a higher degree of customization can instead send their requests to the collection's planner LLM, with a custom system prompt and tool definitions, and receive tool calls in the response.
--- a/docs/guide/configuration/custom-models.md
+++ b/docs/guide/configuration/custom-models.md
@@ -114,6 +114,8 @@ The Omni Model editor only offers already-registered compatible models for each 
 
 If a component model is deleted later, the Omni Model entry remains registered but is hidden from the chat picker until every referenced component is available again.
 
+The editor also exposes an optional **System Prompt** field. Leave it blank to use the default OmniRouter system prompt; supply a string to override it just for this collection. The override is a *template* — keep the `{tool_list}` and `{tool_guidance}` placeholders so the planner still sees the available tools (the server expands them at runtime based on which components are present).
+
 ### Share a collection: export, import, and Hugging Face
 
 `lemonade export <collection>` (and the desktop app's Export button) writes a *collection file*: the
--- a/docs/guide/configuration/custom-models.md
+++ b/docs/guide/configuration/custom-models.md
@@ -114,7 +114,7 @@ The Omni Model editor only offers already-registered compatible models for each 
 
 If a component model is deleted later, the Omni Model entry remains registered but is hidden from the chat picker until every referenced component is available again.
 
-The editor also exposes an optional **System Prompt** field. Leave it blank to use the default OmniRouter system prompt; supply a string to override it just for this collection. The override is a *template* — keep the `{tool_list}` and `{tool_guidance}` placeholders so the planner still sees the available tools (the server expands them at runtime based on which components are present).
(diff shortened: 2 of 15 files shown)


jeremyfowers:
@fl0rianr you may be interested in this as well :)

fl0rianr (commented):
Thanks, I really like this and it looks solid overall. The change follows the intended scope well: system_prompt is now carried through the server/app model metadata paths, export/import allowlists preserve it, the server and app both prefer the per-collection prompt with the global prompt as fallback, and the round-trip coverage is useful.

One small thing I’d like to see before merge: please validate custom prompt templates in the Omni Model editor before save/export. The UI tells users to keep `{tool_list}` and `{tool_guidance}`, but missing placeholders currently seem to be accepted and would silently produce a planner prompt without the tool list/guidance after substitution. A simple validation error would make this new template contract much safer.

Minor nit: when deciding whether the textarea still equals the default, it may be better to compare trimmed values, e.g. `trimmed === DEFAULT_OMNI_SYSTEM_PROMPT.trim()`, so accidental leading/trailing whitespace around the default doesn’t persist a frozen copy as a real override.

With that validation addressed, this is good from my side.

jeremyfowers:
image

hehehehe

jeremyfowers (approved):
Please address @fl0rianr's comment but otherwise LGTM!

PR #2399, opened 2026-06-24
test/ci: follow up on hosted Ollama tool-calling timeouts

fl0rianr:
## Summary

Follow-up to #2365.

The previous CI hardening added reset/diagnostic handling between shared-server test suites. New CI logs show that the reset itself is working, but a heavier Ollama native tool-calling path can still wedge hosted Windows/macOS runners: the request times out after 500s and can leave the Qwen tool-calling backend reported as "busy" / "in_use". That then causes misleading follow-on failures in later streaming-error tests.

This PR narrows that failure mode without removing the broader CI coverage.

## Changes

* Gate test_026_anthropic_messages_tool_calling on GitHub-hosted Windows/macOS.

  * Linux keeps the heavy native Qwen tool-calling coverage.
  * Windows/macOS can still opt in with LEMONADE_RUN_HEAVY_OLLAMA_TOOL_TESTS=1.
  * The lighter Ollama tool-streaming coverage remains active.

* Make reset_server_state.py verify the unload postcondition via /api/ps.

  * /api/v1/unload returning is no longer treated as sufficient by itself.
  * The helper checks that no models remain running before reporting reset completion.

* Make server_streaming_errors.py::test_003 skip when its unload precondition is not met.

  * This avoids a secondary timeout when a previous test has already left the server busy.
  * The primary failure remains attributed to the test that wedged the backend.

## Rationale

The failing path is not a normal lightweight API compatibility check. It uses the native tool-calling model Qwen3-4B-Instruct-2507-GGUF, which is too heavy and unstable for GitHub-hosted Windows/macOS CI in this shared-server workflow.

Skipping only this heavy path on hosted Windows/macOS is a CI-stability tradeoff, not a broad coverage removal:

* Linux still exercises the heavy native tool-calling behavior.
* Windows/macOS still run the rest of the Ollama API suite.
* The test can be explicitly re-enabled with an env var.
* The reset helper now reports the real postcondition instead of giving false confidence.

### DIFF
--- a/test/server_streaming_errors.py
+++ b/test/server_streaming_errors.py
@@ -59,6 +59,46 @@ class StreamingErrorTests(ServerTestBase):
             self.fail(f"Stream not properly terminated (sink.done() missing?): {exc}")
         return lines
 
+    def _running_models(self):
+        response = requests.get(
+            f"http://localhost:{PORT}/api/ps",
+            timeout=TIMEOUT_DEFAULT,
+        )
+        response.raise_for_status()
+        return response.json().get("models", [])
+
+    def _wait_for_no_running_models(self, timeout=15):
+        """Return True once the server reports no running models.
+
+        These tests are about streaming error termination. If a previous suite
+        left an in-flight backend request wedged, /api/v1/unload can return
+        before /api/ps is empty. Treat that as an unsatisfied precondition so we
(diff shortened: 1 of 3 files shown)


kenvandine:
@claude please review this

jeremyfowers (approved):
Thank you!!!

PR #2400, opened 2026-06-24
fix: properly detect zombie backend processes in is_running()

bong-water-water-bong:
## Summary

Fixes the zombie backend process detection bug originally reported and fixed in #2298, with the review feedback from `superm1` and `fl0rianr` applied.

When the llama-server backend is killed by the Linux OOM killer, it becomes a zombie process (`<defunct>`). The `is_running()` function uses `kill(pid, 0)` as its fallback check, which returns success for zombie processes because the PID still exists in the kernel's process table. As a result, the watchdog never detects the exit and the backend is never automatically restarted.

## Changes

- **`src/cpp/server/utils/platform/process_unix.cpp`**: Added `is_zombie_by_proc()` helper that reads `/proc/<pid>/stat` to detect zombie state without reaping the child. Called from `is_running()` before the `kill(pid, 0)` fallback.

- **`test/cpp/test_process_manager.cpp`**: New standalone C++ test verifying the non-mutating contract:
  - `is_running()` returns false for an exited (zombie) child
  - `reap_process()` retrieves the real exit code after `is_running()` returned false
  - Edge cases for invalid PIDs
  - Live process detection
  - Running process reaping returns -1

- **`CMakeLists.txt`**: Added `test_process_manager` target, built alongside the process manager source files.

## Review feedback addressed (vs #2298)

- Removed all unnecessary `#ifdef __linux__` guards (this file is Linux-only)
- Trimmed verbose comments in the helper and test file

Closes #2298

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1806,3 +1806,29 @@ if(EXISTS "${_AUTO_TUNE_TEST_SRC}")
     include(CTest)
     add_test(NAME AutoTuneTest COMMAND test_auto_tune)
 endif()
+
+# ProcessManager non-mutating is_running() contract test.
+set(_PM_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_process_manager.cpp"
+)
+set(_PM_TEST_LIBS
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/server/utils/process_manager.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/server/utils/platform/process_unix.cpp"
+)
+list(GET _PM_TEST_LIBS 0 _pm_lib0)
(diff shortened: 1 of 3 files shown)


bong-water-water-bong (commented):
This PR removes the `#ifdef __linux__` guards that PR #2298 correctly added. Two concerns:

1. **`process_unix.cpp` is not Linux-only** — per CMakeLists.txt, it compiles on all UNIX platforms. `/proc/<pid>/stat` parsing is Linux-specific.
2. **The CMake guard removal** re-introduces a Windows build break that fl0rianr already flagged as a blocker on #2298.

PR #2298 is the more correct fix. Recommend closing this in favor of #2298.

fl0rianr:
Thanks for your ongoing work here, appreciating your effort. This specific PR is not needed anymore in favor of https://github.com/lemonade-sdk/lemonade/pull/2697

PR #2401, opened 2026-06-24
Add NPU/GPU pipeline collection recipe

rubixhacker:
## Summary
- add a new `collection.npu_gpu` recipe for two-component NPU-prefix + GPU-verifier chat pipelines
- route `collection.npu_gpu` chat completions through a server-side orchestrator that drafts a short assistant prefix with an FLM/NPU model, then continues with a non-FLM LLM verifier
- document the registration flow and `npu_draft_tokens` request override

## Validation
- `PATH=/home/linuxbrew/.linuxbrew/bin:$PATH cmake --build --preset default --target lemond lemonade -j 6`
- `git diff --check`
- smoke-tested a local `lemond` on `127.0.0.1:13307` with:
  - `qwen3.5-0.8b-FLM` as the NPU draft component
  - `Qwen3.6-27B-MTP-GGUF` as the GPU verifier component
  - non-streaming `/api/v1/chat/completions`, confirming the response contains both `lemonade_pipeline.draft_response` and `lemonade_pipeline.verifier_response`
  - streaming `/api/v1/chat/completions`, confirming SSE emits the NPU prefix chunk, verifier continuation chunk, and `[DONE]`
- verifier response timings showed llama.cpp MTP active: `draft_n=6`, `draft_n_accepted=6`

## Notes
This intentionally does not replace llama.cpp's internal MTP implementation. If the verifier is an MTP GGUF, Lemonade still enables llama.cpp draft-MTP for the GPU pass; the FLM/NPU component supplies an outer early assistant prefix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -582,6 +582,7 @@ include_directories(
 set(SOURCES_CORE
     src/cpp/server/server.cpp
     src/cpp/server/collection_orchestrator.cpp
+    src/cpp/server/npu_gpu_pipeline_orchestrator.cpp
     src/cpp/server/router.cpp
     src/cpp/server/global_vram_monitor.cpp
     src/cpp/server/eviction_engine.cpp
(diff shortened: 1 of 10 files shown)


rubixhacker:
Closing per author request. The current implementation is only an outer NPU prefix pipeline; true NPU speculative drafting would require deeper token-level integration between FLM and llama.cpp.

PR #2402, opened 2026-06-24
Fix sd-cpp:metal install (resolve asset filename wildcard)

jeremyfowers:
## Problem

Installing `sd-cpp:metal` fails with a 404. The macOS Metal release asset embeds the macOS runner version (e.g. `sd-...-bin-Darwin-macOS-15.7.7-arm64.zip`), which changes on every upstream build. PR #2102 replaced the hardcoded version with a `*` wildcard but never added anything to resolve it, so the literal `*` ends up in the download URL.

## Solution

Resolve the wildcard against the GitHub Releases-by-tag API before building the download URL, picking the asset whose name matches the pattern. No-op (zero network cost) for any asset name without a wildcard, so all other backends are unaffected.

Verified locally on macOS: install succeeds, and an SD-Turbo generation runs end-to-end on the Metal backend.

### DIFF
--- a/src/cpp/server/backends/backend_utils.cpp
+++ b/src/cpp/server/backends/backend_utils.cpp
@@ -151,6 +151,97 @@ namespace lemon::backends {
         return ends_with(filename, ".7z");
     }
 
+    // Greedy glob match where '*' matches any (possibly empty) run of
+    // characters. No '?' support — release asset names only need '*'.
+    static bool wildcard_match(const std::string& pattern, const std::string& text) {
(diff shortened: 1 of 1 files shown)


kenvandine:
@claude review

kenvandine (approved):
Overall this looks good, Claude had a couple of minor things to point out but nothing blocking. Good to get in to unblock the release.

Geramy (commented):
I 100% agree on this!

Geramy:
@claude please open a issue referencing this PR and add me to it, make sure to mention that we should be using std::regex and not this regex in a can.

Issue #2405, opened 2026-06-24
[Router] llm router classifier + routing.router desugaring (L0a)


ramkrishna2910:
## Goal
**`llm` router classifier + `routing.router` desugaring** — the L0(a) "LLM-as-router" on-ramp, the starting point for most users. Kept inside the one-engine invariant (everything compiles to a `Decision`).

## Scope
- New classifier `type: "llm"`: runs a (small, local) chat LLM with an author-supplied NL `prompt`, constrained to choose among the candidate names; parses the structured choice and returns it as the label (score 1.0) + the model's short rationale.
- **`routing.router` sugar:** at load time the parser desugars `{ "router": { "type":"llm", "model", "prompt" } }` into one `llm` classifier (labels = `routing.candidates`) + auto-generated identity rules (`label X → route_to X`). Engine stays single-path.
- Runs the chat LLM through the existing `Router` chat path (extends the M·svc wiring, #2384). The LLM model is a bundled component (declared as `routing.router.model`; role = classifier-model).
- **Trace records the chosen candidate + rationale** — so even L0(a) is auditable.

## Out of scope
The L0(b) offline NL→policy compiler (separate, deferred). Determinism guarantees — L0(a) is explicitly the least-deterministic/slowest tier; docs steer steady-state users to L0(b)/L1–L3.

## Acceptance
- A `routing.router` (llm) policy with N candidates routes a request to a candidate the LLM selects; the decision `trace` shows the chosen label + rationale.
- Tested with a fake LLM-invoke returning a fixed choice (unit) + one live end-to-end path.
- Invalid choice (LLM returns a non-candidate) falls back to `default_model`.

## Example — L0(a) `collection.router` JSON (LLM-as-router)
The **lean local form** — `components` reference already-registered models by name (like the built-in `Ultra Collection`); no `models[]` needed:
```jsonc
{
  "model_name": "user.Router-Auto",
  "recipe": "collection.router",
  "components": ["Qwen3-1.7B-GGUF", "Qwen3-8B-GGUF", "Qwen3.5-35B-A3B-GGUF"],
  "routing": {
    "candidates": ["Qwen3-8B-GGUF", "Qwen3.5-35B-A3B-GGUF"],
    "default_model": "Qwen3-8B-GGUF",
    "router": {
      "type": "llm",
      "model": "Qwen3-1.7B-GGUF",
[...]

Issue #2389, opened 2026-06-23
[Router] Milestone tracker — Lemonade Router (generic core)


ramkrishna2910:
**North star:** ship a **generic router for all customers**; the **Trust Router is the first customer**. The engine does pure model selection (boolean rules + classifiers, first-match) and emits a decision trace. Everything trust-specific — verdicts, `block`, audit persistence, consent — is expressed *on top* via the per-rule `outputs` pass-through bag, refusal-routing, request `metadata`, and client-side trace logging. **No trust code in the engine.**

Policy lives in a `collection.router` collection JSON (sibling to `collection.omni`); pointing the `model` field at the collection name triggers the engine — no `/v1/route`, no `"auto"`. Vocabulary mirrors vLLM Semantic Router for migration.

📋 **Full Decision Log is maintained in the comments on this issue.**

---

## 📊 Status — ✅ Router v1 SHIPPED

The entire engine is on `main`: policy loading, dispatch, decision-trace, and **all condition types** — deterministic (keywords/regex/min_chars/metadata), `semantic_similarity` (embeddings), `classifier` (onnxruntime `/v1/classify`, #2626 + #2727), and the `llm` router / `routing.router` sugar (#2698). What remains is the demonstrable exit criterion + UX, tracked as the v1 gate below.

## 🚢 Shippable v1 gate — ✅ COMPLETE

All three v1 requirements merged; the generic router ships end-to-end with every condition type, an authoring UI, and correct residency.

- [x] **#2732** — end-to-end tests + docs (#2388) — **merged**. All four condition paths covered e2e (deterministic, semantic_similarity local+cloud, cloud candidate, real onnxruntime classifier); authoring guide covers rules, all classifier types, and both LLM forms.
- [x] **#2693** — authoring UI (closed #2406) — **merged**. Desktop custom-collection panel authors `collection.router`, including the classifier/LLM-signal forms. (Competing drafts #2717 / #2728 superseded — can be closed.)
- [x] **#2729** — `RoutingHelper` residency class so router helper models don't evict user models from the LRU pool (fixes #2725) — **merged**.

Every condition type is on `main` and exercised end-to-end: deterministic (keywords/regex/min_chars/metadata), `semantic_similarity`, `classifier` (onnxruntime `/v1/classify`, #2626 + #2727), and the `llm` router / `type:"llm"` classifier (#2698).

**Fast-follow (post-v1, do NOT gate the ship):**
[...]

Issue #2382, opened 2026-06-23
[Router] RoutingPolicyEngine — assembly + evaluate() (M8)


ramkrishna2910:
## Goal
Assemble the **RoutingPolicyEngine** (module M8) — the heart. Generic; no trust semantics. Implements the `route()` method whose signature is frozen in #2407.

## Scope
- Holds the immutable `RoutePolicy` (candidates, `default_model`, ordered `rules`, `classifiers`) + `ClassifierServices` (both taken by the #2407 ctor). At construction, **compile** each rule's `MatchExpr` into a `ConditionPtr` via the #2378 `compile()` + the #2379 `LeafFactory`.
- **`Decision route(const RouteContext& ctx, bool want_trace) const`**: build an `EvalContext{ctx, services, want_trace}`, walk rules in order, **first match wins** → `Decision{ route_to, matched_rule, default_used:false, outputs (verbatim from the rule), trace (from EvalContext, only when want_trace) }`.
- **No match → `default_model`** with `matched_rule` empty and `default_used:true`. `catch(...)` → **fail-open to `default_model`** (per-classifier `on_error` handles signal-level failure upstream).
- `const` / thread-safe under concurrent requests: per-request state lives only in the `EvalContext`; the engine holds immutable compiled state behind a `shared_ptr` (hot-swapped by the #2383 parser on reload).

## Acceptance
- End-to-end engine test over a golden policy fixture with faked `ClassifierServices`: a rule-match path, the default path, and `want_trace` on/off.
- Concurrency test (many threads, one engine instance) — no shared mutable state escapes the `EvalContext`.
- `route()` never throws; unmatched + error paths both land on `default_model` with `default_used:true`.

## Notes on the as-built contract (#2407)
- The method is `route(ctx, want_trace)`, not `evaluate(ctx)`; the trace is returned only on opt-in (per #2376). Header/body surfacing (`x-lemonade-route` / `x_lemonade_route`) is #2386, not this issue.

---
**Depends on:** #2378, #2379, #2380, #2381

Issue #2381, opened 2026-06-23
[Router] semantic_similarity classifier (M5)


ramkrishna2910:
## Goal
**`semantic_similarity` classifier** (module M5) — the first model-backed classifier, implementing the score contract against the #2407 `Classifier` interface.

## Scope
- `services_.embed(model, text)` → cosine vs. each `reference_phrases` exemplar → **max cosine similarity** as the score in `[0,1]`.
- Return shape: a `Score` with the single value under the **empty-string key** (`Score::labels[""] = max_cos`), read by a condition via `Score::primary()`. `semantic_similarity` declares `reference_phrases`, **not** `labels`.
- Reference phrases embedded once and cached on the classifier instance.
- Consumed via a classifier-band condition's inclusive `min_score`/`max_score` (default `min_score: 0.5` when neither bound is given).
- The embed model is a **bundled component** of the collection.

## Out of scope
The real `embed()` ↔ Router wiring (#2384); other classifier types (`classifier`/`llm`); the band/memo evaluation (that's the #2378 condition that consumes this score).

## Acceptance
Tested with a fake `embed()` returning fixed vectors — max-cosine computation, reference-phrase caching (embed called once per reference phrase), and inclusive band boundaries covered.

## Example — L2 `collection.router` JSON (semantic_similarity)
The **lean local form** — `components` (incl. the embedder) reference already-registered models by name; no `models[]`:
```jsonc
{
  "version": "1",
  "model_name": "user.Router-Semantic",
  "recipe": "collection.router",
  "components": ["Qwen3-8B-GGUF", "vllm.qwen3-32b", "nomic-embed-text-v1.5-GGUF"],
  "routing": {
    "candidates": ["Qwen3-8B-GGUF", "vllm.qwen3-32b"],
    "default_model": "Qwen3-8B-GGUF",
    "classifiers": [
      { "id": "is_coding", "type": "semantic_similarity",
        "model": "nomic-embed-text-v1.5-GGUF",
        "reference_phrases": ["write a function", "fix this bug", "refactor this code", "time complexity"] }
    ],
    "rules": [
      { "id": "coding-to-big",
        "match": { "classifier": "is_coding", "min_score": 0.78 },
        "route_to": "vllm.qwen3-32b" }
    ]
  }
}
```
The embedder is a **classifier-model**. Embed a `models[]` array only for HF redistribution.

---
**Depends on:** #2379

PR #2408, opened 2026-06-24
[Router] Interfaces + fixtures stub (foundation)

### LINKED ISSUE #2407: [Router] Interfaces + fixtures stub (foundation)
## Goal
Land the shared **C++ interface headers + schema fixtures** that every downstream issue codes against. This is the **first development step after schema sign-off** and the unlock for the parallel engine/wiring/trust tracks — once these compile, work fans out against interfaces instead of finished implementations.

## Scope (headers + fixtures only — no behavior)
- **Types** (`Decision`-side, no trust verdict in core): `RouteContext` (input text, params `{model, has_tools, has_images, chars}`, `metadata`), `Decision { route_to, matched_rule, default_used, outputs, trace }`, `Rule { id, match, route_to, outputs }`, `MatchExpr`/`Condition` node.
- **Interfaces:** `Classifier { Score evaluate(const ClassifierContext&) const; on_error }`, the `Condition` leaf (deterministic op | classifier-condition band), `ClassifierServices { embed(), run_classifier() }`, and the `RoutingPolicyEngine` constructor signature.
- **Fixtures:** the lean `collection.router` examples (L0a–L3, from #2375) + a `Decision`/trace example (#2376), checked in for downstream unit tests to load.
- **A fake `ClassifierServices`** for tests (fixed embeddings / fixed classifier scores).

## Out of scope
Any implementation/behavior — the AST evaluator (#2378), registry (#2379), conditions/classifiers (#2380/#2381/#2405), engine assembly (#2382), parsing (#2383), real wiring (#2384). This issue is the contract surface only.

## Acceptance
- Headers compile standalone (no backend/Router includes).
- Fixtures validate against the #2375/#2376 JSON Schemas.
- The fake `ClassifierServices` builds and is usable from a trivial test.

## Why it's its own issue
It gates four otherwise-parallel tracks (#2378, #2379, #2384, and the trust-milestone #2387 against the trace contract). Keeping it a visible node avoids a hidden serialization where everyone waits on whoever first defines the headers.

---
**Depends on:** #2375, #2376 (the frozen schemas)

ramkrishna2910:
Closes #2407.

Foundation for the **Lemonade Router** milestone (#2389): the shared C++ contract surface, the policy/decision JSON schemas, the back-compat guardrails, and the per-tier example policies that every downstream engine/wiring issue codes against. **Interfaces only — no engine behavior** (`route()` is declared, not implemented; that's #2382).

> **Draft:** gated on schema sign-off (#2375 / #2376). If a schema is revised, churn stays in the schema files + fixtures; the C++ surface barely moves.

## Contract header — `src/cpp/include/lemon/routing_policy.h`
Std library + nlohmann/json only — **no Router/backend include**, per the engine invariant.
- **Data:** `RouteContext`, `Score` (label→score), `OnError`, `Rule`, `MatchExpr` AST, `TraceEntry`, `Decision`, `RoutePolicy`.
- **Injection seam:** `ClassifierServices` (`embed` / `run_classifier` / `chat` — `chat` powers the L0a llm router), `ClassifierContext`, `Classifier`.
- **Evaluation seam:** `Condition`, `EvalContext` (per-request state: classifier memo + trace), `LeafFactory` — the shared contract that lets the evaluator (#2378) and registry (#2379) be built in parallel without colliding at the leaf boundary.
- **Engine:** `RoutingPolicyEngine` ctor signature; `route()` declared, intentionally undefined.

## Schemas — `src/cpp/resources/schemas/`
- `route_policy.schema.json` (#2375), `request.schema.json` (request-side `metadata` + `route_trace`, #2376), `decision.schema.json` (#2376), `README.md`.
- The policy + decision schemas carry a **required root `version` (`"1"`)**; the request schema rides on the stock OpenAI body so it has no version.
- Engine-owned **v1 semantics are frozen** in the field descriptions so a future change can't silently re-route an old policy: keywords = case-insensitive substring, regex = ECMAScript, score bands inclusive within `[0,1]` (default `min_score: 0.5`), `min/max_chars` = UTF-8 bytes, `metadata` match = case-sensitive comma-split token set, `on_error` default `match_false`, `routing.router` desugaring behavior-equivalent.
- Classifier declarations enforce **type-specific required fields** (`semantic_similarity` → `reference_phrases`, `llm` → `prompt`, …) via `if/then`.

## Back-compat guardrails
The README documents the contract — *never redefine a shipped field; never delete a major's schema; migrate-on-load* — enforced mechanically, not by convention:
- **`schema-lock.json` + `test/test_schema_lock.py`** — canonical-hash snapshot guard (`buf breaking`-lite). A **released** major is immutable (any change fails CI → ship a new major); an **unreleased** major may change only with a refreshed lock in the same diff, so every schema edit is a visible, reviewed change. (v1 is `released: false` until sign-off.)
- **`test/test_routing_fixtures.py`** — schemas are self-valid and every fixture conforms.

(Behavioral back-compat — golden policy → `Decision` — is the conformance corpus #2425, non-gating, tracked separately.)

## Fixtures — `test/cpp/fixtures/routing/`
Lean local-form examples, verbatim from the locked level issues:

| Fixture | Level | Source |
|---------|-------|--------|
| `l0a_llm_router.json` | L0(a) llm-as-router | #2405 |
| `l1_keywords.json` | L1 deterministic (keywords / regex / chars) | #2380 |
| `l1_metadata.json` | L1 deterministic (`metadata` match — `equals`/`any`/`exists`) | #2380 |
| `l2_semantic.json` | L2 semantic_similarity | #2381 |
| `l3_classifier.json` | L3 model-backed classifier | #2379 |
| `decision_example.json` | Decision + trace | #2376 |

The `metadata` leaf condition (lets rules branch on caller-supplied OpenAI `metadata` like `task_class` / `consent`) was contributed by @SlawomirNowaczyk.

## Tests
- `test/cpp/test_routing_policy_contract.cpp` → CTest `RoutingPolicyContractTest`: header compiles standalone, every type constructs, the `Condition`/`EvalContext`/`LeafFactory` seam is exercised, the fake `ClassifierServices` is callable, the engine is constructible, and fixtures satisfy the locked cross-field invariants (`default_model` ∈ candidates; every `route_to` ∈ candidates; classifier refs resolve).
- `test/cpp/fake_classifier_services.h`: configurable fixed embeddings / scores / replies.
- `test/test_routing_fixtures.py`: fixtures validate against the JSON Schemas + cross-field invariants (`jsonschema`, added to `test/requirements.txt`).
- `test/test_schema_lock.py`: the frozen-major snapshot guard.

## Acceptance (#2407) — all met
- [x] Headers compile standalone (no backend/Router include)
- [x] Fixtures validate against the #2375/#2376 JSON Schemas
- [x] The fake `ClassifierServices` builds and is usable from a trivial test

## Verification
- **C++:** compiles clean at `/W4` (MSVC, VS2026); builds in-tree via CMake; `ctest -R RoutingPolicyContract` passes.
- **Python:** `test_routing_fixtures.py` and `test_schema_lock.py` pass; Black-clean.
- **CI:** these now run on every PR — `RoutingPolicyContractTest` is built + run in the C++ unit-test steps, and a fast server-less `Routing Schema Tests` lane runs the two Python guards (so the schema lock + fixture validation are enforced, not decorative).

### DIFF
--- a/src/cpp/resources/schemas/README.md
+++ b/src/cpp/resources/schemas/README.md
@@ -1,0 +1,110 @@
+# Routing engine schemas (Lemonade Router)
+
+Frozen JSON Schemas for the generic routing engine. The engine does **pure model
+selection** (boolean rules over classifiers, first-match-wins, fail-open to
+`default_model`); trust-specific concerns (verdicts, block, audit, consent) are
+layered on top via the `outputs` pass-through bag, request `metadata`, and the
+decision `trace` — never in the engine.
+
+| Schema | Describes |
+|--------|-----------|
+| `route_policy.schema.json` | The `routing` block embedded in a `collection.router` collection JSON. Invoked like `collection.omni`: point the OpenAI `model` field at the collection name. |
+| `decision.schema.json` | The `x_lemonade_route` decision object attached additively to the chat response. |
+
+## Versioning
+
+Both schemas carry a **required** root `version` field — these files define
+version `"1"`. It is required (not optional) because the format is greenfield:
+declaring it from the first policy avoids a breaking "add it later" migration and
+lets a server branch cleanly on shape instead of guessing when the field is
+absent. A server validates a document against the schema matching its major
+version and rejects an unknown major with a clear message rather than a confusing
+per-key error. The policy `version` is author-declared; the decision `version` is
+always emitted by the engine.
+
+Evolution rule:
+- **Additive, backward-compatible data** (new keys in the open `outputs` bag or
+  request `metadata`) needs no version bump — those seams are unconstrained.
+- **New vocabulary** (a new condition op, classifier `type`, or decision field)
+  is a deliberate schema edit; a compatible addition stays within major `1`, a
+  breaking change ships a new major and a new schema file.
+
+## Compatibility & frozen semantics
+
+**Hard rule: a future server must never break a policy authored against an
+earlier major.** This holds structurally — old policies keep validating against a
+newer (superset) schema, and `additionalProperties: false` only constrains the
+forward direction (an old server rejecting a *newer* policy), which `version`
+handles cleanly.
+
+Two operating principles keep it true:
+
+1. **Never redefine a shipped field — only add.** A v1 field's meaning is
+   immutable. New behavior arrives as a *new* field / op / classifier `type`
+   (which old policies don't use), never as a reinterpretation of an existing
+   one. The `classifier.type` enum already reserves future preset names for this
+   reason.
+2. **Never delete or edit a shipped major's schema + parser.** Evolution is
+   additive: ship `vN+1` schemas and per-major load-time shims that upgrade older
+   documents to the latest internal form; retain every prior major's schema so
+   its policies still validate and route.
+
+These are enforced mechanically, not just by convention:
+
+- `schema-lock.json` + `test/test_schema_lock.py` pin a canonical hash of each
+  schema. A **released** major is immutable — any change fails CI (ship a new
+  major instead). An **unreleased** major may change, but only with a refreshed
+  lock in the same diff (`python test/test_schema_lock.py --update`), so every
+  schema edit is a visible, reviewed change rather than silent drift. (v1 is
+  `released: false` until schema sign-off; flip it to `true` at release.)
+- `test/test_routing_fixtures.py` keeps the schemas self-valid and the example
+  fixtures conformant.
+- A frozen conformance corpus (golden policy → expected `Decision`) will enforce
+  *behavioral* stability across versions; schema↔parser key parity is checked
+  where the parser is built. (Both are tracked separately in the milestone.)
+
+Lemonade executes a policy identically regardless of version — the only behavioral
+drift for model-backed classifiers (semantic_similarity / classifier / llm) comes
+from the **backend engine or the model**, not from lemonade. To keep that drift
+from silently flipping a route, the engine-owned semantics below are **frozen for
+v1** (pinned in the schema field descriptions):
+
+| Semantic | Frozen v1 definition |
+|----------|----------------------|
+| `keywords_any` / `keywords_all` | case-insensitive **substring** over input text |
+| `regex` | **ECMAScript** dialect (`std::regex`) |
+| `min_score` / `max_score` | **inclusive** band (`>=` / `<=`); default `min_score: 0.5` when neither bound is given |
+| `min_chars` / `max_chars` | input length in **UTF-8 bytes** (not code points) |
+| `on_error` (omitted) | default **`match_false`** (fail-open) |
+| `routing.router` desugaring | expansion to one `llm` classifier + identity rules is deterministic and behavior-equivalent across versions |
+
+Anything fancier (token/BM25 keyword matching, a different regex engine,
+token-based length) ships as a new, separately named op — never by changing one
+of the above.
+
+## Authoring levels (example fixtures)
+
+Lean local-form examples — `components` reference already-registered models by
+name; the full `models[]` manifest is only needed for Hugging Face
+redistribution. Fixtures live in `test/cpp/fixtures/routing/`:
+
+| Fixture | Level | Mechanism |
+|---------|-------|-----------|
+| `l0a_llm_router.json` | L0(a) | `routing.router` LLM-as-router (desugars to one `llm` classifier + identity rules) |
+| `l1_keywords.json` | L1 | Deterministic `keywords_any` / `regex` / `min_chars` |
+| `l2_semantic.json` | L2 | `semantic_similarity` classifier (embeddings + cosine) |
+| `l3_classifier.json` | L3 | Model-backed `classifier` (PII / jailbreak) |
+| `decision_example.json` | — | A `Decision` + `trace` |
+
+Levels **compose** — one policy may mix a router, classifiers, and deterministic
+conditions across its rules.
+
+## Contract surface
+
+The C++ types/interfaces these schemas back live in
+`src/cpp/include/lemon/routing_policy.h`. Validation:
+
+- **Schema validation** — `python test/test_routing_fixtures.py` (uses `jsonschema`).
+- **Contract / cross-field invariants** — `test/cpp/test_routing_policy_contract.cpp`
+  (CTest target `RoutingPolicyContractTest`): default_model and every
+  `route_to` must be a candidate; classifier condition refs must resolve.
--- a/src/cpp/resources/schemas/README.md
+++ b/src/cpp/resources/schemas/README.md
@@ -9,6 +9,7 @@ decision `trace` — never in the engine.
 | Schema | Describes |
 |--------|-----------|
 | `route_policy.schema.json` | The `routing` block embedded in a `collection.router` collection JSON. Invoked like `collection.omni`: point the OpenAI `model` field at the collection name. |
+| `request.schema.json` | The request-side extension fields on the OpenAI chat body: `metadata` (string-valued routing inputs) and the optional `route_trace`. Validates only those fields (`additionalProperties: true`); no `version` (rides on the stock OpenAI request). |
 | `decision.schema.json` | The `x_lemonade_route` decision object attached additively to the chat response. |
 
 ## Versioning
--- a/src/cpp/resources/schemas/README.md
+++ b/src/cpp/resources/schemas/README.md
@@ -75,6 +75,7 @@ v1** (pinned in the schema field descriptions):
 | `regex` | **ECMAScript** dialect (`std::regex`) |
 | `min_score` / `max_score` | **inclusive** band (`>=` / `<=`); default `min_score: 0.5` when neither bound is given |
 | `min_chars` / `max_chars` | input length in **UTF-8 bytes** (not code points) |
+| `metadata` | reads a request `metadata` key; **case-sensitive** comparison, value decoded into a comma-split, trimmed **token set** (`equals` exact / `any` set-intersection / `exists` presence) |
 | `on_error` (omitted) | default **`match_false`** (fail-open) |
 | `routing.router` desugaring | expansion to one `llm` classifier + identity rules is deterministic and behavior-equivalent across versions |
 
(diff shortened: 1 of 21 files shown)


src/cpp/resources/schemas/route_policy.schema.json
```diff
@@ -0,0 +1,176 @@
...
+          "type": "integer",
+          "minimum": 0
+        },
+        "max_chars": {
+          "description": "Inclusive upper bound on input length. Frozen v1 unit: UTF-8 bytes (not Unicode code points).",
+          "type": "integer",
+          "minimum": 0
+        },
+        "has_tools": { "type": "boolean" },
+        "has_images": { "type": "boolean" }
```
SlawomirNowaczyk: I think we need a way to condition on metadata, which is now missing...
My suggestion is to add here:
`        "metadata": { "$ref": "#/$defs/metadata_match" }`

src/cpp/resources/schemas/route_policy.schema.json
```diff
@@ -0,0 +1,176 @@
...
+      },
+      "additionalProperties": false
+    }
```
SlawomirNowaczyk, line 217: "metadata_match": {
      "description": "Deterministic leaf matching a caller-supplied OpenAI `metadata` key (e.g. task_class, consent), read verbatim from the request metadata map. Frozen v1 semantics: values are compared case-sensitively; a value is decoded into a token set by splitting on comma and trimming, so scalar and list-valued metadata match uniformly. Exactly one comparator (equals / any / exists) must be present. A future comparator (regex, all, ...) ships as a new key, never a redefinition of these.",
      "type": "object",
      "required": ["key"],
      "properties": {
        "key": {
          "description": "The metadata key to read. A missing or empty value matches only `exists: false`.",
          "type": "string"
        },
        "equals": {
          "description": "True if the raw metadata value equals this string exactly (case-sensitive).",
          "type": "string"
        },
        "any": {
          "description": "True if the value's comma-decoded token set intersects this list (handles both scalar and comma-encoded list-valued metadata).",
          "type": "array",
          "items": { "type": "string" }
        },
        "exists": {
          "description": "exists:true matches when the key is present and non-empty; exists:false matches when it is absent or empty.",
          "type": "boolean"
        }
      },
      "oneOf": [
        { "required": ["equals"] },
        { "required": ["any"] },
        { "required": ["exists"] }
      ],
      "additionalProperties": false
    }

src/cpp/resources/schemas/README.md
```diff
@@ -0,0 +1,110 @@
...
+from the **backend engine or the model**, not from lemonade. To keep that drift
+from silently flipping a route, the engine-owned semantics below are **frozen for
+v1** (pinned in the schema field descriptions):
+
+| Semantic | Frozen v1 definition |
+|----------|----------------------|
+| `keywords_any` / `keywords_all` | case-insensitive **substring** over input text |
+| `regex` | **ECMAScript** dialect (`std::regex`) |
+| `min_score` / `max_score` | **inclusive** band (`>=` / `<=`); default `min_score: 0.5` when neither bound is given |
+| `min_chars` / `max_chars` | input length in **UTF-8 bytes** (not code points) |
```
SlawomirNowaczyk, line 81: I think we need a way to condition on metadata, which is now missing...
I suggest adding this:

```
| `metadata` | reads a request `metadata` key; **case-sensitive** comparison, value decoded into a comma-split, trimmed **token set** (`equals` exact / `any` set-intersection / `exists` presence) |
```

src/cpp/resources/schemas/README.md
```diff
@@ -0,0 +1,110 @@
...
+## Authoring levels (example fixtures)
+
+Lean local-form examples — `components` reference already-registered models by
+name; the full `models[]` manifest is only needed for Hugging Face
+redistribution. Fixtures live in `test/cpp/fixtures/routing/`:
+
+| Fixture | Level | Mechanism |
+|---------|-------|-----------|
+| `l0a_llm_router.json` | L0(a) | `routing.router` LLM-as-router (desugars to one `llm` classifier + identity rules) |
+| `l1_keywords.json` | L1 | Deterministic `keywords_any` / `regex` / `min_chars` |
```
SlawomirNowaczyk, line 99:
```
| `l1_metadata.json` | L1 | Deterministic `metadata` match on caller-supplied routing inputs (`task_class` / `consent`) |
```

and a new fixture `l1_metadata.json`:
```
{
  "version": "1",
  "model_name": "user.Router-Metadata",
  "components": ["Qwen3-8B-GGUF", "vllm.qwen3-32b"],
  "routing": {
    "default_model": "vllm.qwen3-32b",
    "rules": [
      { "id": "sensitive-task-stays-local",
        "match": { "any": [
          { "metadata": { "key": "task_class", "any": ["payment", "checkout"] } },
          { "metadata": { "key": "consent", "equals": "denied" } } ] },
        "route_to": "Qwen3-8B-GGUF" },
      { "id": "no-consent-recorded-stays-local",
        "match": { "metadata": { "key": "consent", "exists": false } },
        "route_to": "Qwen3-8B-GGUF" }
    ]
  }
}
```

SlawomirNowaczyk (approved):

src/cpp/resources/schemas/decision.schema.json
```diff
@@ -0,0 +1,57 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://lemonade-sdk.github.io/schemas/decision.schema.json",
+  "title": "Lemonade route decision (x_lemonade_route)",
+  "description": "The decision object the engine emits, attached additively to the chat response body as `x_lemonade_route`. Pure model selection — no verdict/route-category/action in core; those are trust-customer concerns read off `outputs`. The scalar matched_rule is also surfaced in the `x-lemonade-route` response header.",
+  "type": "object",
+  "required": ["version", "route_to", "matched_rule", "default_used"],
+  "properties": {
```
eddierichter-amd, line 8: #2376 also describes the request-side extension surface: OpenAI `metadata` and optional `route_trace`. This schema currently covers the emitted decision object, but I don’t see schema coverage or fixtures for the request extension shape.

  Is request validation intentionally deferred, or should we add a small request-extension schema/fixture for:

  {
    "metadata": { "task_class": "payment", "site_tags": "shopping" },
    "route_trace": true
  }

  with metadata values constrained to strings?

src/cpp/resources/schemas/route_policy.schema.json
```diff
@@ -0,0 +1,176 @@
...
+      "description": "Nested any/all/not over leaf conditions; first-match-wins across rules. vLLM SR's flat single-operator form is the subset.",
+      "type": "object",
+      "minProperties": 1,
+      "properties": {
+        "any": { "type": "array", "items": { "$ref": "#/$defs/match_expr" } },
+        "all": { "type": "array", "items": { "$ref": "#/$defs/match_expr" } },
+        "not": { "$ref": "#/$defs/match_expr" },
+        "classifier": { "type": "string" },
+        "label": { "type": "string" },
+        "min_score": {
```
eddierichter-amd, line 146: Do we want to enforce [0, 1] in the schema?

src/cpp/resources/schemas/route_policy.schema.json
```diff
@@ -0,0 +1,176 @@
...
+      "type": "object",
+      "required": ["type", "model", "prompt"],
+      "properties": {
+        "type": { "const": "llm" },
+        "model": { "type": "string" },
+        "prompt": { "type": "string" }
+      },
+      "additionalProperties": false
+    },
+    "classifier": {
```
eddierichter-amd, line 68: Should classifier schemas use conditional requirements by `type`?
                                               
  For example:
  - `semantic_similarity` requires `model` and `reference_phrases`
  - `classifier` requires `model`    
  - `llm` requires `model` and `prompt`    

  Right now `id` and `type` are required, but type-specific required fields are left to later validation.

eddierichter-amd (approved):
Just some small nits. Otherwise looks good.

ramkrishna2910:
Thanks @eddierichter-amd — all three addressed in the latest commit:

1. **Request-extension schema** — added `request.schema.json` + a `request_example.json` fixture covering the request side of #2376: `metadata` (object, **string values only** — list values are comma-encoded) and the optional `route_trace` (bool). It validates only the extension fields and leaves the rest of the OpenAI body alone (`additionalProperties: true`), and intentionally has **no `version`** since it rides on the stock request rather than being a versioned document. Added a test that a non-string metadata value is rejected.
2. **`[0,1]` bounds** — `min_score`/`max_score` now carry `minimum: 0, maximum: 1` (the `Score` contract range).
3. **Type-specific required fields** — classifier declarations now enforce, via `if/then`: `semantic_similarity` → `model` + `reference_phrases`, `classifier` → `model`, `llm` → `model` + `prompt`. Type-specific validation now lives in the schema rather than deferring to the parser. (Reserved post-v1 preset types stay unconstrained for now.)

Lock refreshed, README + both test harnesses updated. Good catches — the request schema in particular was a real gap against #2376's deliverable.

eddierichter-amd:
@ramkrishna2910  one more question now that I am scoping out #2379. Should the classifier contract expose declared `labels` and `default_label`? 

  The registry/leaf factory needs this to:
  - reject condition label refs that are not in the classifier's declared labels
  - apply default_label when a classifier condition omits label
                                          
  Something like:
                                          
  const std::vector<std::string>& labels() const;
  const std::optional<std::string>& default_label() const;

  and corresponding protected constructor fields would make #2379 implementable without keeping a sidecar metadata table outside the Classifier.

ramkrishna2910:
Good call @eddierichter-amd — agreed, and done in 8fed705. `labels` and `default_label` are intrinsic to the classifier declaration, so they live on the `Classifier` contract rather than a sidecar the registry has to keep in sync:

```cpp
const std::vector<std::string>& labels() const;
const std::optional<std::string>& default_label() const;
```

with the protected ctor extended to carry them (defaulted, so `semantic_similarity` just omits them).

One nuance worth flagging for #2379: `semantic_similarity` declares **no** labels — it scores under the empty-string key (read via `Score::primary()`). So its `labels()` comes back empty and `default_label()` is `nullopt`. That's intentional and convenient: your "reject a `label` ref not in the classifier's declared labels" check then naturally rejects any `label` ref on a similarity condition, which is exactly right (you band over its single score, not a named label).

Left two things to #2379 rather than baking them into the contract: (1) what to do when a condition omits `label` **and** the classifier has no `default_label` (error vs. fall back to `primary()`), and (2) the dangling-ref rejection itself — the accessors just expose the data so you can implement whichever policy. Contract test now covers both a labeled classifier and the empty-labels similarity case.

PR #2421, opened 2026-06-25
fix(server): recover stalled HF model downloads

### LINKED ISSUE #2416: lemond model downloader hangs at 0 bytes on Xet-backed HF repos (no error, no progress)
## Summary

`lemond`'s model downloader hangs indefinitely (0 bytes, no error) when pulling a model whose HuggingFace repo is stored on the **Xet** storage backend. The repository metadata is fetched successfully, the file list is resolved, but the actual blob download never transfers a single byte and never errors out — it just stalls until the client disconnects.

## Environment

- Lemonade: `lemonade-server` **10.8.1~24.04** (PPA `lemonade-team/stable`)
- OS: Zorin OS 18.1 (Ubuntu 24.04 / noble)
- GPU: 2× AMD Radeon AI PRO R9700 (gfx1201, RDNA4)
- Backend: `llamacpp` (rocm); HF cache at `/ml/huggingface-cache/huggingface/hub/`

## Reproduction

```
lemonade pull Qwen-AgentWorld-35B-A3B-GGUF-UD-Q4_K_XL
```
(repo: `unsloth/Qwen-AgentWorld-35B-A3B-GGUF` — public, not gated, **Xet-backed**)

Or directly against the server:
```
curl -X POST http://127.0.0.1:13305/api/v1/pull \
  -d '{"model_name":"Qwen-AgentWorld-35B-A3B-GGUF-UD-Q4_K_XL"}'
# -> HTTP 000 after 20s (no response; nothing written to disk)
```

## Server log (lemond.service)

```
(ModelManager) Downloading model: unsloth/Qwen-AgentWorld-35B-A3B-GGUF (variant: UD-Q4_K_XL)
(ModelManager) Fetching repository file list from Hugging Face...
(ModelManager) Using commit hash: 3a305abf5cfd119ee999dfe929c433746edd8d63
(ModelManager) Repository contains 25 files
(ModelManager) Identified files to download:
(ModelManager)   - ...:Qwen-AgentWorld-35B-A3B-UD-Q4_K_XL.gguf
(ModelManager)   - ...:mmproj-F16.gguf
(ModelManager) Created download manifest
(ModelManager) Downloading: Qwen-AgentWorld-35B-A3B-UD-Q4_K_XL.gguf...
        <-- hangs here forever, 0 bytes, no progress, no error -->
(Server) Client disconnected, cancelling download
```

## Evidence it's Xet-specific (not a bad URL / network issue)

The plain HTTPS path works perfectly from the same host:

```
# resolve URL 302 -> us.aws.cdn.hf.co (xet-bridge), x-linked-size: 22324804864
curl -L -r 0-1048575 .../resolve/main/Qwen-AgentWorld-35B-A3B-UD-Q4_K_XL.gguf
# -> HTTP 206, 1048576 bytes in 0.44s
```

The repo carries Xet headers (`x-xet-hash`, `link: ...rel="xet-reconstruction-info"` → `cas-server.xethub.hf.co`). Non-Xet repos (e.g. older GGUFs) download fine through `lemond`. Downloading the same two files with `huggingface_hub` + `HF_HUB_DISABLE_XET=1` succeeds over plain HTTPS into the same cache.

## What I tried

Setting `HF_HUB_DISABLE_XET=1` **and** `HF_XET_DISABLE=1` in the `lemond` service environment (`/etc/lemonade/conf.d/`) and restarting — **no change**, the downloader still stalls. So lemond's downloader does not appear to honor the standard HF Xet opt-out env vars.

## Expected behavior

`lemond` should either (a) correctly complete Xet-backed downloads, or (b) fall back to the plain HTTPS CDN path when Xet reconstruction stalls, and in any case **time out with an error** instead of hanging silently at 0 bytes.

## Suggestions

- Honor `HF_HUB_DISABLE_XET` / `HF_XET_DISABLE` as an escape hatch.
- Add a stall/timeout on the blob transfer so a hung Xet reconstruction surfaces an error rather than hanging.
- Given the growing number of Unsloth/HF repos migrating to Xet, this will affect many popular GGUF models.

Happy to provide full debug logs or test a build on gfx1201 hardware.

fl0rianr:
Fixes model downloads that can hang indefinitely at 0 bytes for Hugging Face `/resolve/` URLs backed by Xet.

The downloader now detects no-byte progress stalls and retries affected Hugging Face resolve downloads with a byte-range request. Non-Hugging-Face downloads keep the existing behavior.

Fixes #2416


The-Monk:
**Verified on RDNA4 / gfx1201** (2× Radeon AI PRO R9700, lemonade-server 10.8.1 + this PR built from source).

Multiple Xet-backed repos that hung indefinitely at 0 bytes on stock 10.8.1 now pull cleanly with this patch:

| repo (Xet-backed) | result |
|---|---|
| `nomic-ai/nomic-embed-text-v1-GGUF` | ✅ 7s |
| `unsloth/Qwen3-Embedding-0.6B-GGUF` | ✅ 14s (609 MB) |
| `Qwen3.5-0.8B-FP16-vLLM` (13-file safetensors) | ✅ 42s (1.6 GB) |

All confirmed Xet via the `/resolve/` → `cdn.hf.co/xet-bridge-us` 302 + `x-xet-hash` header. The byte-range stall recovery does the job. 👍 (Also confirms the root cause: the env opt-outs `HF_HUB_DISABLE_XET`/`HF_XET_DISABLE` had no effect since lemond's downloader is native C++, not the Python `huggingface_hub` client.)

jeremyfowers (changes requested):
It seems like an abstraction leak to have "huggingface" specifically mentioned in the http client. 

Is it possible to either:
1. implement this as a generic solution for all http clients?
2. provide the huggingface-specific solution at a higher level of the software stack?

src/cpp/server/utils/http_client.cpp
```diff
@@ -270,25 +277,38 @@ static fs::path get_disk_space_probe_path(const fs::path& output_path) {
...
-    if (dltotal > 0 && data->callback) {
-        // Call user callback - returns false to cancel
-        if (!data->callback(dlnow, dltotal)) {
-            data->cancelled = true;
-            return 1;  // Abort transfer
+    const auto now = std::chrono::steady_clock::now();
+    if (dlnow > data->last_downloaded) {
+        data->last_downloaded = dlnow;
+        data->last_progress_time = now;
+    } else if (data->no_progress_timeout > 0) {
```
jeremyfowers, line 343: Suggestion: also gate this branch on `data->last_downloaded > 0` so the idle clock starts only after the first byte arrives. Otherwise slow connect/redirect/first-byte latency is counted as a stall and can abort otherwise-healthy transfers.

fl0rianr: Thanks, done.

jeremyfowers:
## Review notes

The stall-detection core is reasonable, but a couple of things stand out — the top two look unrelated to the stall fix and appear to be an accidental revert from an older base (the merge-base `3c972dd9` has the correct code), so a rebase is worth doing to confirm they weren't reintroduced inadvertently.

### 1. `HttpClient::post` truncates binary POST bodies (telemetry protobuf breakage)
`src/cpp/server/utils/http_client.cpp:427` — switching `CURLOPT_POSTFIELDS, body.data()` + `CURLOPT_POSTFIELDSIZE_LARGE` to `body.c_str()` (dropping the explicit size) makes curl `strlen()` the body. `telemetry.cpp:618` POSTs `serialize_protobuf_batch()` output — raw protobuf (fixed64 nanosecond timestamps, IEEE-754 doubles, random trace/span IDs) that almost always contains `0x00`, so the OTLP body is silently cut at the first NUL and the collector drops the spans. The same change in `post_stream` (`:563`) is a latent hazard for any future binary stream body.

### 2. `HttpClient::post` forces `Content-Type: application/json`
`src/cpp/server/utils/http_client.cpp:435` — the removed `has_content_type` check let a caller override the content type. `telemetry.cpp:598` sets `application/x-protobuf`, so the request now sends two conflicting Content-Type headers (and duplicate identical ones for the json path / `wrapped_server.cpp:634`). A strict OTLP collector returns 400/415 — an independent break of protobuf telemetry on top of #1.

### 3. 60s stall timer measures time-to-first-byte, not just mid-stream stalls
`src/cpp/server/utils/http_client.cpp:298` — `header_callback` sets `download_phase_started=true` on the blank line after a 2xx status, so the timer starts when headers arrive. If the origin then needs >60s to emit the first body byte (cold Xet reconstruction, or a slow GitHub/TheRock mirror), `progress_callback` aborts at `:352` as a false stall, and each retry re-incurs the same cold start. Note the defaults are unconditional, so this applies to every download (`backend_utils.cpp` archive/TheRock downloads use default `DownloadOptions`), not just HF.

### 4. Connected-but-mute server never times out
`src/cpp/server/utils/http_client.cpp:343` — if a server completes the TCP connect but never sends headers or bytes, `transfer_has_started` stays false, so the else-branch resets `last_progress_time` every tick and `no_progress_timeout` never accumulates. With `CURLOPT_TIMEOUT=0` and `low_speed_limit=0`, nothing aborts and the download hangs indefinitely. If the #2416 hang can occur before any response header, the new feature won't cover it.

---
*Generated with [Claude Code](https://claude.com/claude-code)*

jeremyfowers (approved):
Thanks for the fixes!

PR #2430, opened 2026-06-26
Add arm64 builds to all Linux release packages and PPA

kenvandine:
* Fix CMakeLists.txt to detect arm64 Linux and name embeddable archive ubuntu-arm64
   * Fix CPackRPM.cmake to auto-detect aarch64 architecture for RPM packaging
   * Add build-lemonade-debian13-arm64 + smoke test job (Debian 13 .deb on ARM64)
   * Add build-lemonade-rpm-arm64 + test job (Fedora 43/44 .rpm on aarch64)
   * Add build-lemonade-embeddable-linux-arm64 job (portable archive for Linux ARM64)
   * Include all arm64 packages in the GitHub release job
   * Add ubuntu-arm64 entry to test-embeddable-posix matrix
   * Add package-arm64 job to launchpad-ppa.yml for PR binary verification on ARM64
   * Update build-container.yml to build multi-arch (amd64+arm64) Docker images

Co-Authored-By: Claude Sonnet 4.6 [noreply@anthropic.com](mailto:noreply@anthropic.com)
Claude-Session: https://claude.ai/code/session_01EQ7ke6n2tMDgNfxg125ieZ


.github/workflows/launchpad-ppa.yml
```diff
@@ -121,6 +121,56 @@ jobs:
...
+    name: Build (ARM64)
+    needs: prepare-matrix
+    runs-on: ubuntu-24.04-arm
+    permissions:
+      contents: read
+    if: github.event_name == 'pull_request'
+    strategy:
+      matrix: ${{ fromJSON(needs.prepare-matrix.outputs.package_matrix) }}
+    container:
+      image: ghcr.io/lemonade-sdk/lemonade/build-environment:${{ matrix.distro }}${{ matrix.release }}
```
fl0rianr: blocking: The new ARM64 PPA verification job currently cannot start because it consumes the image noted above on an ARM64 runner, but the referenced images do not yet have linux/arm64/v8 manifests. The current CI failure is already at container initialization (no matching manifest for linux/arm64/v8), before any package build runs.

Since build-container.yml only runs on schedule / workflow_dispatch, this PR does not publish the new multi-arch build-environment images before launchpad-ppa.yml tries to consume them. Please either publish/split the multi-arch build-environment image change first, gate package-arm64 until the ARM64 manifests exist, or add a PR-safe fallback path.

fl0rianr (changes requested):

.github/workflows/launchpad-ppa.yml
kenvandine: Fixed in 8960f51. Rather than dropping the container entirely (which would limit us to noble), I switched to the official `ubuntu:${{ matrix.release }}` images from Docker Hub — those are natively multi-arch (arm64 supported for 24.04, 25.10, and 26.04) so the full matrix is preserved. Added an inline dep-install step since the stock Ubuntu images lack the pre-installed build tools that the custom `build-environment` provides.

Once `build-container.yml` runs post-merge to publish the multi-arch images, we can revert the container reference back to the custom `build-environment`.

.github/workflows/launchpad-ppa.yml
```diff
@@ -121,6 +121,68 @@ jobs:
           path: ${{ steps.build_source.outputs.output-dir }}/*
           retention-days: 30
 
+  package-arm64:
+    name: Build (ARM64)
+    # Uses official ubuntu:${{ matrix.release }} images from Docker Hub instead of
+    # the custom build-environment images, because the latter do not yet carry
+    # linux/arm64 manifests (they gain multi-arch support once build-container.yml
+    # runs after this PR merges). Official Ubuntu images are natively multi-arch.
```
superm1: how about we make the containers now?  we can just trigger a job from any branch to make containers.

superm1 (commented):
Can you please split up to reusable job snippets?  There are a bunch of jobs that are 100% the same except for `runs-on`.  too much tech debt IMO

.github/workflows/launchpad-ppa.yml
superm1: I've dispatched a run here for creating the containers.  You should be able to use them now.

https://github.com/lemonade-sdk/lemonade/actions/runs/28280103167

.github/workflows/build-container.yml
```diff
@@ -54,6 +57,7 @@ jobs:
           build-args: |
             BASE_IMAGE=ubuntu:${{ matrix.ubuntu_version }}
             UBUNTU_PPA=${{ matrix.ubuntu_ppa }}
+          platforms: linux/amd64,linux/arm64
```
superm1, line 60: actually, is this right?  I think you actually want a matrix build don't you?

kenvandine:
Addressed the tech-debt feedback in 269a27b96:

- Consolidated 5 pairs of duplicated x86_64/arm64 jobs into single jobs with a matrix strategy — the only difference (`runs-on`) is now a matrix variable.
- RPM jobs use a 2D matrix (`fedora-version` × `runs-on`) via cross-product with `include` for arch properties.
- Reverted `package-arm64` in `launchpad-ppa.yml` to use the custom `build-environment` container now that the multi-arch images are available (thanks @superm1 for dispatching that build).

Net: -340 lines in the workflow files.

.github/workflows/build-container.yml
kenvandine: Yeah, done!

fl0rianr (commented):
blocking: The ARM64 PPA verification job still cannot start for the 26.04 matrix entry. The current run fails during container initialization when pulling ghcr.io/lemonade-sdk/lemonade/build-environment:ubuntu26.04 on ubuntu-24.04-arm with no matching manifest for linux/arm64/v8.

Please either publish the missing ubuntu26.04 ARM64 manifest before merging, keep the PR-safe fallback to official Ubuntu images for ARM64, or temporarily exclude/gate the 26.04 ARM64 PPA verification entry until the custom build-environment image exists.

.github/workflows/launchpad-ppa.yml
superm1: They have probably been overwritten by now.  I suggest you split the PR into two.  You do one PR that gets you containers built, we can merge that first.  Then this one to add all the support you want to utilize the new containers.

I also think you should be having two separate containers for x86 and arm64.

kenvandine (commented):
Thanks for catching this, @fl0rianr. You're correct —  only runs on schedule/dispatch, so the  ARM64 manifest was never published despite the workflow being updated in this PR.

Fixed in f23e157: restored option 2 (official ubuntu images for the ARM64 job). The  job now uses  from Docker Hub (natively multi-arch, covers 24.04/25.10/26.04) with explicit build-dep installation. The custom build-environment reference will come back once  runs post-merge and publishes the multi-arch images.

fl0rianr (commented):
The current Launchpad failure looks related to the ARM64 fallback path, but it is no longer the old missing-manifest issue. The official ubuntu:${{ matrix.release }} container starts successfully; the job now fails at actions/checkout@v5 because the stock Ubuntu image does not have Git installed yet. Checkout then falls back to the GitHub REST API, but that fallback does not support submodules: recursive.

Suggested fix: install minimal checkout prerequisites before actions/checkout, then keep the full build-dep installation after checkout:

```yaml
      - name: Install checkout prerequisites
        run: |
          apt-get update
          apt-get install -y git ca-certificates

      - name: Checkout code
        uses: actions/checkout@v5
        with:
          submodules: recursive
          fetch-depth: 0
          fetch-tags: true
```
This is the same ordering pattern already used by the Debian 13 container build, where Git is installed before checkout.

kenvandine (commented):
Good catch. Fixed in f6b1da10 — added an "Install checkout prerequisites" step (git + ca-certificates) before actions/checkout, matching the same pattern used by the Debian 13 container build. The full build-dep installation stays after checkout as suggested.

contrib/debian/control
```diff
@@ -8,20 +8,23 @@ Build-Depends:
  dh-cmake,
  dh-cmake-compat (= 1),
  dh-sequence-cmake,
- fonts-katex,
  jq,
  libdrm-dev,
  libcli11-dev,
- libcpp-httplib-dev (>> 0.26),
```
superm1, line 15: This is very important in case it eventually makes it to Debian and needs to backport. Please keep it.

kenvandine: Ok, we'll need to get arm64 builds for cpp-httplib in the PPA then. Without the depends CMake uses FetchContent to get it, but that doesn't help us for debian.

superm1: Yes, we do. The PPA is enabled for aarch64 now. Can we just trigger a rebuild of source package?

kenvandine: No, we need a no change upload to trigger it.

superm1: It looks like your upload worked, I copied it to stable as well.

fl0rianr (approved):
Now code is fine and CI green, I'm merging now.

Issue #2412, opened 2026-06-25
Incorrect status message when downloading a model in offline mode


ckuethe:
### Platform

Windows

### Lemonade Version

10.8.1

### Component

Other / Not sure

### Bug Description

when lemond is in offline mode, model downloads "complete" very quickly, without downloading anything, and there is no indication to the user that no download was attempted because offline mode was active

### Steps to Reproduce

1. `lemonade config set offline=true`
2. `lemonade pull ...` (some custom huggingface model)
3. observe `Model pulled successfully: ...`
4. observe the model is not listed in `lemonade list --downloaded`
5. get confused
6. check server logs, observe `[Info] (ModelManager) Offline mode enabled, skipping download`

### Expected vs Actual Behavior

Expected: "Lemond is in offline mode, model not downloaded"
Actual: "Model pulled successfully..."

### Log Output

```shell

```

anditherobot:
PR #2434 fixed offline model pulls incorrectly reporting success. Okay to close?

GabrielReusRodriguez:
Hello, it is fine from my viewpoint. :)

PR #2434, opened 2026-06-26
Fix (#2412) : Added offline check before pulling a model on server.cpp

GabrielReusRodriguez:
# Resume
About Issue : https://github.com/lemonade-sdk/lemonade/issues/2412

I have added a config check in handle_pull function of server.cpp . Now before checking if user sent a model, we check in config if the model is offline ( offline = "true" ) if it is offline, we send an error response with a user friendly error.

image

image

# Testing
Tested on  ubuntu 24.04 on lemonade 1.8.1 both options with lemonade cli  (config.offline= true and config.offline= false).


bong-water-water-bong (approved):
Clean fix for #2412. The offline() check is safe (defaults to false when key absent), error code follows snake_case conventions, and the 400 status is consistent with other errors in this handler.

One nit: the comment `// First we check if lemonade is offline.` is a WHAT comment that's redundant with the code — AGENTS.md says to skip those. Drop it or replace with a brief why (e.g., `// Issue #2412: offline pull returned misleading error`).

GabrielReusRodriguez:
Hello!

I removed the comment from the source code.

Thanks for your time reviewing the PR :)

fl0rianr (changes requested):
Thanks, this fixes the direct /api/v1/pull path, but I think the Hugging Face CLI flow can still bypass the new check.

For lemonade pull owner/repo, the CLI first calls `/api/v1/pull/variants?checkpoint=...` before it posts to /api/v1/pull. handle_pull_variants() currently calls fetch_pull_variants(), which performs a Hugging Face API request, without checking config_->offline() first.

Can we add the same offline guard to Server::handle_pull_variants() before calling lemon::fetch_pull_variants(...) ? That keeps offline mode consistent and avoids any network/HF lookup before returning the user-friendly error.

Suggested behavior:
- `GET /api/v1/pull/variants?checkpoint=owner/repo` with `offline=true` returns 400
- JSON includes the same `code: "lemond_offline"`
- no Hugging Face request is attempted

GabrielReusRodriguez:
> Thanks, this fixes the direct /api/v1/pull path, but I think the Hugging Face CLI flow can still bypass the new check.
> 
> For lemonade pull owner/repo, the CLI first calls `/api/v1/pull/variants?checkpoint=...` before it posts to /api/v1/pull. handle_pull_variants() currently calls fetch_pull_variants(), which performs a Hugging Face API request, without checking config_->offline() first.
> 
> Can we add the same offline guard to Server::handle_pull_variants() before calling lemon::fetch_pull_variants(...) ? That keeps offline mode consistent and avoids any network/HF lookup before returning the user-friendly error.
> 
> Suggested behavior:
> 
> * `GET /api/v1/pull/variants?checkpoint=owner/repo` with `offline=true` returns 400
> * JSON includes the same `code: "lemond_offline"`
> * no Hugging Face request is attempted

Hello, 
I added the changes. I hope I did not forget anything.

Thanks for your time reviewing the PR :)

fl0rianr (approved):
Thanks, code-wise this looks good to me. Approving in assumption CI is green.

PR #2437, opened 2026-06-26
feat: add model files endpoint

fl0rianr:
## Summary

Adds a backend-owned per-model file inventory endpoint for the GUI Files tab:

- GET /api/v1/models/{id}/files
- GET /v1/models/{id}/files

This is intentionally separate from storage/system metrics. Storage endpoints
answer drive/system usage questions; this endpoint answers which resolved local
files belong to one selected model.

## Response shape

```json
{
  "model_id": "Qwen3-4B",
  "files": [
    {
      "name": "model.gguf",
      "path": "/abs/path/model.gguf",
      "role": "main",
      "size_bytes": 123456789,
      "exists": true
    }
  ]
}
```

## Implementation notes

- Adds ModelManager::list_model_files() using ModelInfo::resolved_paths as the first-slice source of truth.
- Uses `std::filesystem::exists`/existing safe filesystem helpers and the existing resolved-path size helper.
- Sums directories recursively, matching existing model-size behavior.
- Registers /models/{id}/files before the generic /models/(.+) route so foo/files is not interpreted as a model ID.
- Returns the public/wire model ID via the same alias resolution path as GET /v1/models/{id}.

## Validation

<details>
<summary>PR 2 backend smoke test: <code>/models/{id}/files</code> works for Gemma-4-E2B-it-GGUF</summary>

Validated both public route variants:

```http
GET http://localhost:13305/v1/models/Gemma-4-E2B-it-GGUF/files
GET http://localhost:13305/api/v1/models/Gemma-4-E2B-it-GGUF/files

Both endpoints return the same expected response shape:

{
  "model_id": "Gemma-4-E2B-it-GGUF",
  "files": [
    {
      "name": "gemma-4-E2B-it-Q4_K_M.gguf",
      "path": "/home/vscode/.cache/huggingface/hub/models--unsloth--gemma-4-E2B-it-GGUF/snapshots/ecc8b33b2c50598815e4b0f7cea6088e3ae7adb8/gemma-4-E2B-it-Q4_K_M.gguf",
      "role": "main",
      "size_bytes": 3106736256,
      "exists": true
    },
    {
      "name": "mmproj-F16.gguf",
      "path": "/home/vscode/.cache/huggingface/hub/models--unsloth--gemma-4-E2B-it-GGUF/snapshots/ecc8b33b2c50598815e4b0f7cea6088e3ae7adb8/mmproj-F16.gguf",
      "role": "mmproj",
      "size_bytes": 985654080,
      "exists": true
    }
  ]
}

Result:

main file is detected with role main
mmproj file is detected with role mmproj
both paths resolve to existing files
total listed size: 4092390336 bytes, roughly 3.81 GiB
/v1/... and /api/v1/... behave consistently
</details> ```

### CODE REVIEW

**Review by bong-water-water-bong — CHANGES_REQUESTED** [review]
Two issues that should be addressed:

**1. HIGH — Missing `/api/v0` and `/v0` route variants:** Every other model endpoint follows the quad-prefix convention (`/api/v0`, `/api/v1`, `/v0`, `/v1`). This endpoint only registers under `/api/v1` and `/v1`. Requests to `/api/v0/models/foo/files` will fall through to the generic model-by-id route and return "Model not found."

**2. HIGH — Absolute filesystem paths exposed in API response:** The existing `model_info_to_json` deliberately omits `resolved_paths`. This new endpoint exposes full paths like `/home/user/.cache/huggingface/hub/models--...`, revealing home directories, usernames, and cache layout. If absolute paths are truly needed by the GUI, at minimum document this in the API docs as intentional info disclosure.

**Minor:** TOCTOU gap between `model_exists()` check and `list_model_files()` call (pre-existing pattern, but worth wrapping in try-catch for 404). Missing `#include <vector>` in `model_manager.h`.

File: docs/api/lemonade.md [hunk 1]
```diff
@@ -29,6 +30,83 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/metrics`](#get-metrics) | Prometheus metrics scrape endpoint |
 | `POST` | [`/internal/telemetry/flush`](#post-internaltelemetryflush) | Force-flush all queued telemetry trace spans |
 
+## `GET /v1/models/{id}/files`
+<sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
+
+List resolved local file metadata for a single model. This endpoint is intended
+for model-detail UIs such as the Files tab. It is per-model inventory, not
+system or drive storage accounting.
```
jeremyfowers: Markdown documentation shouldn't have linebreaks like this, just rely on the renderer for word wrap

jeremyfowers (approved):
LGTM!

PR #2438, opened 2026-06-26
feat(cli): introduce model filtering for the list command

abn:
Extends the C++ server API and CLI client to support comprehensive model filtering and JSON output formatting. This allows CLI users to query, filter, and programmatically consume both local and downloadable models based on metadata.

## User Value
- **Advanced CLI Filtering**: Users can filter models using options like `--type`, `--source`, `--device`, `--suggested`, `--label`, and `--backend`.
- **JSON Formatting**: Adding the `--json` flag dumps the entire model list (with extended properties like device, backend options, size, labels, etc.) as a structured JSON array, enabling easy scripting and automation.
- **Layout Preservation**: Reliably handles very long model names without breaking column alignment or state indicators.
- **Up-to-Date Documentation**: Documented the new options under the CLI guide in the project's documentation (`docs/guide/cli.md`).

---

## Command Help Output (`lemonade list --help`)
```
List available models. Use --downloaded to show only local models.

./build/lemonade list [OPTIONS] [name_filter]

POSITIONALS:
  name_filter NAME_FILTER     Optional case-insensitive model-name filter; supports * wildcards

OPTIONS:
  -h,     --help              Display help information
          --help-all          Display help information for all subcommands
          --downloaded        Show only downloaded models
          --json              Format output as JSON
          --type TEXT         Filter by model type (llm, embedding, reranking, transcription,
                              image, tts)
          --source TEXT       Filter by model source (builtin, user, extra, cloud)
          --device TEXT       Filter by model device (cpu, gpu, npu)
          --suggested         Show only suggested models
          --label TEXT ...    Filter by one or more labels
          --backend TEXT      Filter by backend/recipe
```

---

### 1. Filtered List Example (`lemonade list --type llm`)
```
Local
Model Name                              Downloaded     Size (GB)      Details
Baguettotron-BF16                       Yes                0.60       llamacpp
Bonsai-1.7B-gguf                        Yes                0.25       llamacpp
GLM-4.7-Flash-GGUF                      Yes               17.50       llamacpp
```

### 2. JSON Format Example (`lemonade list --json`)
```json
[
    {
        "checkpoint": "PleIAs/Baguettotron-GGUF:Baguettotron-BF16.gguf",
        "device": "gpu",
        "downloaded": true,
        "id": "Baguettotron-BF16",
        "labels": [
            "custom",
            "llamacpp",
            "reasoning"
        ],
        "recipe": "llamacpp",
        "recipe_options": {
            "ctx_size": 0,
            "llamacpp_args": "--temp 0.7 --top-p 0.95 --flash-attn on --cache-type-k q8_0 --cache-type-v q8_0 --chat-template-kwargs '{\"preserve_thinking\": true}' -b 4096 -ub 1024",
            "llamacpp_backend": "vulkan"
        },
        "size_gb": 0.6,
        "source": "builtin",
        "suggested": true,
        "type": "llm"
    }
]
```

### DIFF
--- a/src/cpp/cli/lemonade_client.cpp
+++ b/src/cpp/cli/lemonade_client.cpp
@@ -1,13 +1,15 @@
 #include "lemon_cli/lemonade_client.h"
 #include "lemon/utils/url_utils.h"
 #include <httplib.h>
-#include <iostream>
 #include <algorithm>
+#include <cctype>
 #include <iomanip>
+#include <iostream>
+#include <numeric>
 #include <regex>
 #include <sstream>
+#include <string_view>
 #include <nlohmann/json.hpp>
-#include <sstream>
 
 namespace lemonade {
 
@@ -52,12 +54,12 @@ static std::regex build_name_filter_regex(const std::string& name_filter) {
     return std::regex(regex_pattern, std::regex_constants::ECMAScript | std::regex_constants::icase);
 }
 
-static bool starts_with(const std::string& value, const std::string& prefix) {
+static bool starts_with(std::string_view value, std::string_view prefix) {
     return value.rfind(prefix, 0) == 0;
 }
 
-static std::string strip_canonical_prefix(const std::string& model_name) {
-    static const std::vector<std::string> prefixes = {"user.", "extra.", "builtin."};
+static std::string_view strip_canonical_prefix(std::string_view model_name) {
+    static constexpr std::string_view prefixes[] = {"user.", "extra.", "builtin."};
     for (const auto& prefix : prefixes) {
         if (starts_with(model_name, prefix)) {
             return model_name.substr(prefix.size());
@@ -66,7 +68,7 @@ static std::string strip_canonical_prefix(const std::string& model_name) {
     return model_name;
 }
 
-static int model_source_sort_rank(const std::string& model_name) {
+static int model_source_sort_rank(std::string_view model_name) {
     if (starts_with(model_name, "user.")) return 1;
     if (starts_with(model_name, "extra.")) return 2;
     if (starts_with(model_name, "builtin.")) return 3;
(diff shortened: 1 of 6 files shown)


bong-water-water-bong (commented):
Well-structured PR with good test coverage. Two items to address:

**1. MEDIUM — CLI type-inference fallback diverges from server:** The client-side `get_model_type_from_labels()` fallback omits LLM-indicator labels (`vision`, `reasoning`, `tool-calling`, `tools`, `chat-transcription`) that the server checks first. If a model has labels `["vision", "embeddings"]`, the server classifies it as `llm` but the CLI fallback classifies it as `embedding`. Since the server-side change in this same PR always emits the `type` field, the simplest fix is to remove the client-side duplicate inference entirely.

**2. MEDIUM — `--label` missing `TakeAll` policy:** The `--label` option on `list` doesn't set `->multi_option_policy(CLI::MultiOptionPolicy::TakeAll)`, unlike the existing `--label` on `pull`. Older CLI11 versions default to `TakeLast`, silently dropping all but the last value.

**Minor:** Device `icontains` is overly permissive (`--device pu` matches `gpu` and `npu`). Consider splitting on `|` and using `iequals` against components.

src/cpp/cli/lemonade_client.cpp
```diff
@@ -489,21 +513,101 @@ std::vector<ModelInfo> LemonadeClient::get_models(bool show_all) const {
...
+    return a.size() == b.size() && std::equal(a.begin(), a.end(), b.begin(),
+        [](char c1, char c2) { return std::tolower(c1) == std::tolower(c2); });
+}
+
+static bool icontains(const std::string& haystack, const std::string& needle) {
+    auto it = std::search(haystack.begin(), haystack.end(), needle.begin(), needle.end(),
+        [](char c1, char c2) { return std::tolower(c1) == std::tolower(c2); });
+    return it != haystack.end();
+}
+
```
bong-water-water-bong: The `--device` filter now uses `device_matches()` (split on `|`, `iequals` per token), so this `icontains()` helper has no remaining callers — dead code that `-Wunused-function` will flag.

All three items from my earlier review are already addressed in `340d387`/`a1742d7`; this just removes the leftover helper from item 3.

```suggestion
```

bong-water-water-bong (commented):
Followed up on the earlier review — all three items are now addressed in `340d387`/`a1742d7`:

1. ✅ Client-side type inference removed (reads server `type` field).
2. ✅ `--label` has `MultiOptionPolicy::TakeAll`.
3. ✅ `--device` uses `device_matches()` (tokenizes on `|`, `iequals`).

One nit left: the item-3 refactor left `icontains()` unused (see the inline suggestion). Verified `git grep icontains` returns no callers in `src/`, and `<algorithm>` is still needed by `std::equal`/`transform`/`remove_if`/`sort`.

src/cpp/cli/lemonade_client.cpp
bong-water-water-bong: Confirmed — the `icontains()` helper was removed in commit `1f29344`. This is resolved. 🟢

Issue #2385, opened 2026-06-23
[Router] collection.router recipe + dispatch


ramkrishna2910:
## Goal
**`collection.router` recipe + dispatch** — make a collection invoke the engine, mirroring `collection.omni` wiring (copy the wiring, not the orchestrator code).

## Scope
- Register `collection.router` recipe (`is_collection_recipe` family) as a sibling to `collection.omni`.
- `chat/completions` (and completions/responses) where `model` = a `collection.router` collection name flips into engine mode — **no `"auto"`, no `/v1/route`**; the recipe is the trigger, exactly like omni.
- Call `engine.evaluate()` → `route_to`; dispatch the request through `Router` to the chosen candidate component.
- Quad-prefix compliant (invariant #1).

## Out of scope
Attaching the trace to the response (#2386); audit (Trust milestone).

## Acceptance
`POST /v1/chat/completions` to a `collection.router` collection returns a real completion produced by the engine-selected candidate.

---
**Depends on:** #2383, #2384, #2376

ramkrishna2910:
Completed through #2558

Issue #2384, opened 2026-06-23
[Router] ClassifierServices to Router wiring (M-svc)


ramkrishna2910:
## Goal
**ClassifierServices ↔ Router wiring** (module M·svc). Bridge the pure engine to live backends.

## Scope
- `embed()` → `Router::embeddings(json)` (exists at `router.h:104`) — powers `semantic_similarity`.
- `run_classifier(model, input)` → `ensure_loaded` + invoke a **classifier component model** through the `Router` and return its score/labels.
- Engine touches backends ONLY through these lambdas.

## Acceptance
Integration test: a `semantic_similarity` classifier (and a model-backed classifier) loop back through the live Router and return scores.

---
**Depends on:** #2407 (interfaces + fixtures), #2382

ramkrishna2910:
## Status: #2548 merged the wiring, but this issue stays open

PR #2548 landed the `ClassifierServices` ↔ Router adapter (`make_router_classifier_services`), which is the right layer and correctly keeps `routing_policy.h` backend-free. What it actually gives us:

- ✅ **`embed()` → `Router::embeddings()`** — real. `semantic_similarity` classifiers loop back through the live Router and return genuine cosine scores. This half of the acceptance is met.
- ⚠️ **`run_classifier()` → chat completion** — the adapter implements model-backed classification by prompting a **chat LLM** and parsing labels out of its text. That's a stand-in, not real classification: there's no sequence-classification head, no calibrated label probabilities.

So the acceptance — *"a model-backed `classifier` loops back through the live Router and returns scores"* — is only **partially** satisfied. Merging #2548 as-is is fine (the wiring is correct and unblocks `semantic_similarity` + the future L0a `llm` router), but it can't close this issue, because **Lemonade has no interface to run an actual classification model.** At a glance there's no direct path to execute an HF text-classification model today.

## The real gap: no classifier-execution backend

A true `classifier` condition needs a backend that runs sequence-classification models (label logits → scores), not a chat LLM. Options raised in discussion:

- **Easiest — reuse existing vLLM support.** Needs a spike to confirm vLLM can serve HF text-classification / sequence-classification heads (not just generative).
- **Needs scoping — add ONNX Runtime or torch as a Lemonade backend.** More work, but ONNX in particular is a natural fit for these small classifier models and several of the candidates below already ship ONNX.

[...]

PR #2440, opened 2026-06-26
Router/2378 2379 evaluator registry

### LINKED ISSUE #2378: [Router] Core types + match-expression evaluator (M6/M7)
## Goal
Implement the **match evaluator** — the runtime `Condition` node types and the compile step that turns a parsed `MatchExpr` into an evaluable tree. Pure; zero backend dependencies. The shared types (`RouteContext`, `Decision`, `Rule`, `MatchExpr`, `Condition`, `EvalContext`, `LeafFactory`) **already landed in #2407** — this issue implements behavior against them, it does not re-declare them.

## Scope
- **Composite conditions** over `EvalContext`: `all` (short-circuit on first false), `any` (short-circuit on first true), `not`. Each is a `Condition`.
- **Classifier-band leaf condition**: holds a resolved `ClassifierPtr` + optional `label` + inclusive `min_score`/`max_score` band (default `min_score: 0.5` when neither bound is given) + `on_error`. Reads the classifier's `Score` from `EvalContext::memo` (running it via `services` at most once per request — memo keyed by classifier id), applies the band to `score.score_of(label_or_default)` (or `score.primary()` for single-score classifiers), and on `!score.ok` returns per `on_error` (`match_true`/`match_false`). Appends a `TraceEntry{condition, score, result}` when `ctx.want_trace`.
- **`compile(const MatchExpr&, const LeafFactory&) -> ConditionPtr`**: build the composite structure from `MatchExpr`; for `Leaf` nodes delegate to the injected `LeafFactory` (deterministic ops come from #2380, the `classifier` leaf is the band condition above). This is the seam that keeps the evaluator independent of the concrete leaf builders.

## Out of scope
Deterministic leaf ops (#2380), classifier instantiation + the `LeafFactory` wiring (#2379), `semantic_similarity` (#2381), engine assembly (#2382), parsing (#2383).

## Acceptance
- Table tests over a fake `LeafFactory` + fake `ClassifierServices`: `all`/`any` short-circuit, `not`, deep nesting, inclusive-band boundaries, `on_error` both ways, and **memoization** (a classifier callback invoked at most once per request even when referenced by several conditions).
- Trace is populated only when `want_trace`.
- No `#include` of any backend/Router header.

## Notes on the as-built contract (#2407)
- `Decision { route_to, matched_rule, default_used, outputs, trace }` — there is no `reasons` field.
- `Score` is a `label -> score` map; use `score_of(label)` / `primary()`.
- `EvalContext { request, services, want_trace, memo, trace }` is the per-request state (the former "EvalState").

---
**Depends on:** #2407 (interfaces + fixtures; carries the #2375/#2376 schemas)

### LINKED ISSUE #2379: [Router] Classifier + Condition registry (M3)
## Goal
**Classifier + condition registry** (module M3). The extensibility seam: instantiate classifiers and leaf conditions from JSON and resolve their references. Codes against the `Classifier` / `Condition` / `ClassifierServices` / `LeafFactory` interfaces frozen in #2407.

## Scope
- **`make_classifier(type, json) -> ClassifierPtr`** — a registry keyed by type-string. v1 builders: `semantic_similarity` (#2381), `classifier`, `llm` (#2405); the reserved vLLM-SR preset names resolve as thin presets over `classifier` post-v1. Each concrete `Classifier` subclass sets `id`/`type`/`on_error` and runs a model via the injected `ClassifierServices` (`embed` / `run_classifier` / `chat`).
- Build `classifiers_` (id -> `ClassifierPtr`) from `routing.classifiers`.
- **The `LeafFactory`** (leaf json -> `ConditionPtr`): dispatch on the leaf's key — deterministic ops (`keywords_*` / `regex` / `min_*chars` / `has_*`) build the #2380 conditions; a `{"classifier": id, ...}` leaf resolves `id` against `classifiers_` and constructs the #2378 classifier-band condition with that `ClassifierPtr` injected. Concrete leaf builders register into the registry so neither the evaluator (#2378) nor the registry depends on the other's implementation.
- **Validation**: reject unknown classifier/condition types and **dangling classifier references** cleanly (reject-not-crash); reject dangling condition `label` refs against the classifier's declared `labels`.

## Out of scope
The concrete deterministic-op logic (#2380), `semantic_similarity` scoring (#2381), the composite/band evaluation semantics (#2378), engine assembly (#2382).

## Acceptance
- Instantiates classifiers + a `LeafFactory` from JSON; unknown types and dangling refs are rejected with clear errors.
- `on_error` (`match_true`/`match_false`) round-trips, unit-tested with a fake `ClassifierServices`.

## Notes on the as-built contract (#2407)
- `Score` is a `label -> score` map (HF text-classification convention); a model-backed `classifier` returns the per-label scores and the condition reads `default_label`'s score against the band. `semantic_similarity` reports a single score under the empty-string key (read via `Score::primary()`).
- `ClassifierServices` exposes `embed`, `run_classifier`, AND `chat` (the last powers the `llm` router, #2405).

## Example — L3 `collection.router` JSON (model-backed classifiers)
The **lean local form** — `components` (incl. the classifier models) reference already-registered models by name; no `models[]`:
```jsonc
{
  "version": "1",
  "model_name": "user.Router-Classify",
  "components": ["Qwen3-8B-GGUF", "vllm.qwen3-32b", "pii-detector-small", "jailbreak-detector-small"],
  "routing": {
    "default_model": "vllm.qwen3-32b",
    "classifiers": [
      { "id": "pii",       "type": "classifier", "model": "pii-detector-small",
        "labels": ["PII", "NO_PII"],        "default_label": "PII",       "on_error": "match_true" },
      { "id": "jailbreak", "type": "classifier", "model": "jailbreak-detector-small",
        "labels": ["JAILBREAK", "BENIGN"],  "default_label": "JAILBREAK", "on_error": "match_true" }
    ],
    "rules": [
      { "id": "sensitive-stays-local",
        "match": { "any": [
          { "classifier": "pii",       "min_score": 0.5 },
          { "classifier": "jailbreak", "min_score": 0.5 } ] },
        "route_to": "Qwen3-8B-GGUF",
        "outputs": { "verdict": "warn" } }
    ]
  }
}
```
`on_error: match_true` = **fail-closed authoring**. The classifier-backend recipe is defined by #2384. Embed a `models[]` array only for HF redistribution.

---
**Depends on:** #2407 (interfaces + fixtures)

eddierichter-amd:
## Summary                                                        
                                                                                                          
  Implements the initial Lemonade Router evaluator and registry behavior for #2378 and #2379.
                                                                                                                                                                                                                     
  This builds on the #2407 routing foundation types and keeps the implementation pure/backend-free.
                                                                                                          
  ## What Changed                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                           
  - Added routing evaluator implementation:                                                               
    - composite `all` / `any` / `not` conditions                                                                                                                                                                     
    - strict `MatchExpr` compilation                                                                      
    - classifier-band leaf condition                                                    
    - inclusive score-band evaluation                                                                     
    - default classifier band of `min_score: 0.5`         
    - classifier memoization through `EvalContext::memo`                                                  
    - trace emission only when `want_trace` is enabled                             
                                                                                                          
  - Added classifier/leaf registry helpers:                                                               
    - `make_classifier`                                                                                                                                                                                                                                                                                                                                                                                                                    
    - `make_classifiers`                                                                                  
    - `make_leaf_factory`                                                                                                                                                                                                                                                                                                                                                                                                                  
    - generic model-backed `classifier` support via `ClassifierServices::run_classifier`                  
    - clear rejection for `semantic_similarity`, `llm`, and reserved preset types until their follow-up issues land                                                                                                  
    - classifier label/default-label validation                                                           
    - dangling classifier reference rejection                                                                                                                                                                                                                                                                                                                                                                                              
    - deterministic leaf dispatch seam for #2380                                                                                                                                                                                                                                                                                                                                                                                           
    - multi-key leaf support as implicit `all`                                                                                                                                                                                                                                                                                                                                                                                             
                                                                                                          
  - Added standalone C++ tests for evaluator and registry behavior.                     
                                                                                                          
  ## Out Of Scope                 

  This does not implement deterministic condition logic (#2380), `semantic_similarity` scoring (#2381), LLM router classifier behavior (#2405), full engine assembly (#2382), policy parsing (#2383), live Router wiring (#2384), or `collection.router` dispatch (#2385).                                                                                                                                                                 
                                                                                                          
  ## Validation                                                             
                                                                                                          
  Built and ran the routing C++ tests with CMake/CTest:         
                                                                                                          
  ```bash              
  /tmp/lemonade-cmake-venv/bin/cmake --build build \                                                      
    --target test_routing_policy_evaluator test_routing_policy_registry test_routing_policy_contract \
    -j2                                                                                                   
                                                                                                          
  cd build                                                                                                                                                                                                           
  /tmp/lemonade-cmake-venv/bin/ctest -R 'RoutingPolicy(Contract|Evaluator|Registry)Test' --output-on-failure   ```                                                                                                      
                                                                                                          
  Result:                                                                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                           
  100% tests passed, 0 tests failed out of 3                                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                          
  Also ran schema/fixture checks:

  /tmp/lemonade-router-schema-venv/bin/python test/test_routing_fixtures.py
  python3 test/test_schema_lock.py

  Both passed.

### DIFF
--- a/src/cpp/server/routing_policy.cpp
+++ b/src/cpp/server/routing_policy.cpp
@@ -1,0 +1,446 @@
+#include "lemon/routing_policy.h"
+
+#include <algorithm>
+#include <stdexcept>
+#include <utility>
+#include <lemon/utils/aixlog.hpp>
+
+namespace lemon {
+namespace {
+
+Score failed_score() {
+    Score score;
+    score.ok = false;
+    return score;
+}
+
+void validate_children(const std::vector<ConditionPtr>& children,
+                       const char* condition_name) {
+    for (const auto& child : children) {
+        if (!child) {
+            throw std::invalid_argument(std::string(condition_name) +
+                                        " condition contains a null child");
+        }
+    }
+}
+
+class AllCondition final : public Condition {
+public:
+    explicit AllCondition(std::vector<ConditionPtr> children)
+        : children_(std::move(children)) {
+        if (children_.empty()) {
+            throw std::invalid_argument("all condition requires at least one child");
+        }
+        validate_children(children_, "all");
+    }
+
+    bool evaluate(EvalContext& ctx) const override {
+        for (const auto& child : children_) {
+            if (!child->evaluate(ctx)) return false;
+        }
+        return true;
+    }
+
+private:
+    std::vector<ConditionPtr> children_;
+};
+
+class AnyCondition final : public Condition {
+public:
+    explicit AnyCondition(std::vector<ConditionPtr> children)
+        : children_(std::move(children)) {
+        if (children_.empty()) {
+            throw std::invalid_argument("any condition requires at least one child");
+        }
+        validate_children(children_, "any");
+    }
+
+    bool evaluate(EvalContext& ctx) const override {
+        for (const auto& child : children_) {
+            if (child->evaluate(ctx)) return true;
+        }
+        return false;
+    }
+
+private:
+    std::vector<ConditionPtr> children_;
+};
+
+class NotCondition final : public Condition {
+public:
+    explicit NotCondition(ConditionPtr child)
+        : child_(std::move(child)) {
+        if (!child_) {
+            throw std::invalid_argument("not condition requires one child");
+        }
+    }
+
+    bool evaluate(EvalContext& ctx) const override {
+        return !child_->evaluate(ctx);
+    }
+
+private:
+    ConditionPtr child_;
+};
+
+class ClassifierBandCondition final : public Condition {
+public:
+    ClassifierBandCondition(ClassifierPtr classifier,
+                            std::optional<std::string> label,
+                            std::optional<double> min_score,
+                            std::optional<double> max_score)
+        : classifier_(std::move(classifier)),
+          label_(std::move(label)),
+          min_score_(min_score),
+          max_score_(max_score) {
+        if (!classifier_) {
+            throw std::invalid_argument("classifier condition requires a classifier");
+        }
+        if (!min_score_.has_value() && !max_score_.has_value()) {
+            min_score_ = 0.5;
+        }
+        if (min_score_.has_value() && max_score_.has_value() && *min_score_ > *max_score_) {
+            throw std::invalid_argument("classifier condition min_score cannot exceed max_score");
+        }
+    }
+
+    bool evaluate(EvalContext& ctx) const override {
+        const Score& score = score_for(ctx);
+
+        std::optional<double> traced_score;
+        bool result = false;
+        if (!score.ok) {
+            result = classifier_->on_error() == OnError::MatchTrue;
+        } else {
+            double value = selected_score(score);
+            traced_score = value;
+            result = in_band(value);
+        }
+
+        if (ctx.want_trace) {
+            ctx.trace.push_back(TraceEntry{"classifier:" + classifier_->id(), traced_score, result});
+        }
+        return result;
+    }
+
+private:
+    const Score& score_for(EvalContext& ctx) const {
+        auto [it, inserted] = ctx.memo.try_emplace(classifier_->id());
+        if (inserted) {
+            try {
+                it->second = classifier_->evaluate(ClassifierContext{ctx.request, ctx.services});
+            } catch (const std::exception& e) {
+                // Classifier implementations should return Score{ok=false}
+                // rather than throw. Keep this catch as a permanent safety
+                // net: model-backed services can fail in unexpected ways, and
+                // routing must convert that to on_error rather than letting an
+                // exception escape evaluation.
+                LOG(WARNING, "Routing") << "Classifier '" << classifier_->id()
+                                        << "' threw during evaluation: " << e.what()
+                                        << std::endl;
+                it->second = failed_score();
+            } catch (...) {
+                LOG(WARNING, "Routing") << "Classifier '" << classifier_->id()
+                                        << "' threw an unknown exception during evaluation"
+                                        << std::endl;
+                it->second = failed_score();
+            }
+        }
+        return it->second;
+    }
+
+    double selected_score(const Score& score) const {
+        if (label_.has_value()) {
+            return score.score_of(*label_);
+        }
+        if (classifier_->default_label().has_value()) {
+            return score.score_of(*classifier_->default_label());
+        }
+        // Single-score classifiers, e.g. semantic_similarity, declare no labels
+        // and report their score under the empty-string key.
+        return score.primary();
+    }
+
+    bool in_band(double value) const {
+        if (min_score_.has_value() && value < *min_score_) return false;
+        if (max_score_.has_value() && value > *max_score_) return false;
+        return true;
+    }
+
+    ClassifierPtr classifier_;
+    std::optional<std::string> label_;
+    std::optional<double> min_score_;
+    std::optional<double> max_score_;
+};
+
+class ModelClassifier final : public Classifier {
+public:
+    ModelClassifier(std::string id, std::string type, std::string model, OnError on_error,
+                    std::vector<std::string> labels,
+                    std::optional<std::string> default_label)
+        : Classifier(std::move(id), std::move(type), on_error, std::move(labels),
+                     std::move(default_label)),
+          model_(std::move(model)) {
+        if (model_.empty()) {
+            throw std::invalid_argument("classifier requires model");
+        }
+    }
+
+    Score evaluate(const ClassifierContext& ctx) const override {
+        Score score;
+        if (!ctx.services.run_classifier) {
+            return failed_score();
+        }
+
+        try {
+            score.labels = ctx.services.run_classifier(model_, ctx.request.input);
+            score.ok = true;
+        } catch (...) {
+            score = failed_score();
+        }
+        return score;
+    }
+
+private:
+    std::string model_;
+};
+
+std::vector<ConditionPtr> compile_children(const std::vector<MatchExpr>& children,
+                                           const LeafFactory& leaf_factory,
+                                           std::size_t depth);
+
+ConditionPtr compile_match_expr_impl(const MatchExpr& expr,
+                                     const LeafFactory& leaf_factory,
+                                     std::size_t depth) {
+    if (depth > kMaxMatchExprDepth) {
+        throw std::invalid_argument("match expression nesting exceeds maximum depth");
+    }
+
+    switch (expr.op) {
+        case MatchExpr::Op::Leaf: {
+            if (!leaf_factory) {
+                throw std::invalid_argument("leaf factory is required");
+            }
+            auto condition = leaf_factory(expr.leaf);
+            if (!condition) {
+                throw std::invalid_argument("leaf factory returned null condition");
+            }
+            return condition;
+        }
+        case MatchExpr::Op::All:
+            if (expr.children.empty()) {
+                throw std::invalid_argument("all expression requires at least one child");
+            }
+            return make_all_condition(compile_children(expr.children, leaf_factory, depth + 1));
+        case MatchExpr::Op::Any:
+            if (expr.children.empty()) {
+                throw std::invalid_argument("any expression requires at least one child");
+            }
+            return make_any_condition(compile_children(expr.children, leaf_factory, depth + 1));
+        case MatchExpr::Op::Not:
+            if (expr.children.size() != 1) {
+                throw std::invalid_argument("not expression requires exactly one child");
+            }
+            return make_not_condition(compile_match_expr_impl(expr.children[0], leaf_factory, depth + 1));
+    }
+    throw std::invalid_argument("unsupported match expression op");
+}
+
+std::vector<ConditionPtr> compile_children(const std::vector<MatchExpr>& children,
+                                           const LeafFactory& leaf_factory,
+                                           std::size_t depth) {
+    std::vector<ConditionPtr> compiled;
+    compiled.reserve(children.size());
+    for (const auto& child : children) {
+        compiled.push_back(compile_match_expr_impl(child, leaf_factory, depth));
+    }
+    return compiled;
+}
+
+} // namespace
+
+ConditionPtr make_all_condition(std::vector<ConditionPtr> children) {
+    return std::make_shared<AllCondition>(std::move(children));
+}
+
+ConditionPtr make_any_condition(std::vector<ConditionPtr> children) {
+    return std::make_shared<AnyCondition>(std::move(children));
+}
+
+ConditionPtr make_not_condition(ConditionPtr child) {
+    return std::make_shared<NotCondition>(std::move(child));
+}
+
+ConditionPtr make_classifier_band_condition(ClassifierPtr classifier,
+                                            std::optional<std::string> label,
+                                            std::optional<double> min_score,
+                                            std::optional<double> max_score) {
+    return std::make_shared<ClassifierBandCondition>(
+        std::move(classifier), std::move(label), min_score, max_score);
+}
+
+ConditionPtr compile_match_expr(const MatchExpr& expr, const LeafFactory& leaf_factory) {
+    return compile_match_expr_impl(expr, leaf_factory, 0);
+}
+
+ClassifierPtr make_classifier(const json& config) {
+    if (!config.is_object()) {
+        throw std::invalid_argument("classifier entry must be an object");
+    }
+
+    const std::string id = config.value("id", "");
+    const std::string type = config.value("type", "");
+    if (id.empty()) {
+        throw std::invalid_argument("classifier is missing id");
+    }
+    if (type.empty()) {
+        throw std::invalid_argument("classifier '" + id + "' is missing type");
+    }
+
+    OnError on_error = parse_on_error(config.value("on_error", "match_false"));
+    std::vector<std::string> labels;
+    if (config.contains("labels")) {
+        if (!config["labels"].is_array()) {
+            throw std::invalid_argument("classifier '" + id + "' labels must be an array");
+        }
+        for (const auto& item : config["labels"]) {
+            if (!item.is_string() || item.get<std::string>().empty()) {
+                throw std::invalid_argument("classifier '" + id + "' labels must be non-empty strings");
+            }
+            labels.push_back(item.get<std::string>());
+        }
+    }
+
+    std::optional<std::string> default_label;
+    if (config.contains("default_label")) {
+        if (!config["default_label"].is_string() || config["default_label"].get<std::string>().empty()) {
+            throw std::invalid_argument("classifier '" + id + "' default_label must be a non-empty string");
+        }
+        default_label = config["default_label"].get<std::string>();
+        if (labels.empty()) {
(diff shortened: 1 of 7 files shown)


ramkrishna2910 (commented):
## Review — fits the plan, strong work. One must-fix (CI) before merge.

I reviewed this against the foundation contract and the locked milestone decisions, read both test suites line-by-line, and independently built + ran everything at `/W4` against the merged `main` foundation header:

- `routing_policy.cpp` compiles clean at `/W4`, no backend/Router include (pure, as #2378 requires).
- `RoutingPolicyEvaluatorTest` → **ALL PASSED**, `RoutingPolicyRegistryTest` → **ALL PASSED** (reproduced locally, not just trusting the report).

### Completeness vs. the issues it closes — both fully met ✅

**#2378 (evaluator)** — every acceptance bar is covered by a real assertion:
- `all`/`any` short-circuit (verified via call-counting), `not`, deep nested compile
- inclusive band boundaries (min **and** max), default `min_score: 0.5` (accepts at 0.5, rejects 0.49)
- score selection: explicit `label` → `default_label` → `Score::primary()` (the empty-key path)
- `on_error` both directions **and** a throwing classifier → `Score::ok=false`
- **memoization at-most-once** explicitly asserted (`classifier->calls == 1` when one classifier feeds two conditions) — this was the subtle one and it's nailed
- trace only when `want_trace`, and trace omits short-circuited branches
- malformed-AST rejection (empty `any`, `not` with ≠1 child)

**#2379 (registry)** — `make_classifier`/`make_classifiers`/`make_leaf_factory`, label/default-label validation, dangling classifier-ref and dangling label-ref rejection, duplicate-id rejection, clean "not implemented yet" rejections for `semantic_similarity`/`llm`/presets, classifier-only-fields-require-classifier guard, and the deterministic dispatch seam — all tested against a fake `ClassifierServices`.

### 🔴 Must fix before merge — the new tests don't run in CI
The targets build and pass locally, but CI (`.github/workflows/cpp_server_build_test_release.yml`) only builds a fixed target list and runs `ctest -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicyContract"` — in **two** places (the Debian job ~line 277 and the macOS job ~line 537). The evaluator/registry tests are in neither, so they'd be **green locally but never run on a PR**. Could you add to both blocks:

```
cmake --build --preset default --target test_routing_policy_evaluator
cmake --build --preset default --target test_routing_policy_registry
...
ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry)Test"
```
(That's the same gap + fix we applied for `RoutingPolicyContractTest`.)

### 🟡 Should address — one new *frozen* semantic to document
Multi-key leaf = implicit `all` (e.g. `{"keywords_any":[...], "max_chars":1000}` is ANDed; and a classifier band claims `classifier`/`label`/`min_score`/`max_score`, the rest dispatch). It's a reasonable convention (matches vLLM SR) and it's well-tested — but it's **new behavior we never froze**, and per our own back-compat rule, anything that can't change later belongs in the frozen-semantics table in `src/cpp/resources/schemas/README.md`. One row covers it. (Alternative: require explicit `all` and reject multi-key leaves — but since the impl + tests already commit to implicit-all, documenting it is the lighter call.)

### Forward-looking — does it set up the next two devs? Yes. ✅

**#2380 (deterministic conditions) — ready.** The injected `NamedLeafFactories` seam is exactly right: a registered factory receives an **isolated single-key leaf** (`{key: value}`), proven by the test, so #2380's `keywords_*`/`regex`/`*_chars`/`has_*` builders never have to know about sibling keys. #2380 just implements the `Condition` subclasses (reading `EvalContext::request` for input/params) + provides the factory map; no change to this file needed.

**#2381 (semantic_similarity) — ready, with one note for that dev.** The whole consumer pipeline already handles it: a semantic classifier has empty `labels()`, so a leaf `{"classifier":"is_coding","min_score":0.78}` skips the label checks and the band falls back to `Score::primary()` (the empty-`""`-key cosine) — and that primary() path is already implemented **and tested** here (`single.labels[""] = 0.72`). So #2381 only needs to produce `Score{labels:{"":maxcos}, ok:true}` and it flows through unchanged.

> Note for the #2381 dev: classifier **type** dispatch is a closed switch in `make_classifier` (not an injection seam like the deterministic leaf factories). To add semantic_similarity you: (1) add a `SemanticSimilarityClassifier : Classifier` in the anon namespace that parses `reference_phrases`, embeds them once (cache on the instance), and scores via `services.embed` + max-cosine into `labels[""]`; (2) move `semantic_similarity` out of the rejection branch into a construction branch. The contract + consumer side need no changes. (Same closed-switch applies to #2405's `llm` type — minor coordination if they land together.)

### Housekeeping
I've set the **milestone** (Lemonade Router) and **`priority:p0`** label (this implements two P0 demo-path issues). Could you add **`Closes #2378, #2379`** to the PR description so they auto-close on merge?

Net: this is exactly the engine core we planned, built against the real contract, with complete tests, and it cleanly unblocks both #2380 and #2381. Just the CI wiring stands between it and ready.

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+        : children_(std::move(children)) {
+        if (children_.empty()) {
+            throw std::invalid_argument("all condition requires at least one child");
+        }
+    }
+
+    bool evaluate(EvalContext& ctx) const override {
+        for (const auto& child : children_) {
+            if (!child) {
+                throw std::invalid_argument("all condition contains a null child");
```
SlawomirNowaczyk: `routing_policy.h:282` says that we cannot throw during evaluation.

I suggest moving the null-child check to the constructor (next to the existing `children_.empty()` guard). This way it runs once at compile time rather than on every request, and keeps `evaluate()` exception-free. The constructors currently validate emptiness but not `null` entries.

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+        : children_(std::move(children)) {
+        if (children_.empty()) {
+            throw std::invalid_argument("any condition requires at least one child");
+        }
+    }
+
+    bool evaluate(EvalContext& ctx) const override {
+        for (const auto& child : children_) {
+            if (!child) {
+                throw std::invalid_argument("any condition contains a null child");
```
SlawomirNowaczyk: Same as `AllCondition`

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+        }
+
+        if (ctx.want_trace) {
+            ctx.trace.push_back(TraceEntry{"classifier:" + classifier_->id(), traced_score, result});
+        }
+        return result;
+    }
+
+private:
+    Score score_for(EvalContext& ctx) const {
```
SlawomirNowaczyk: Returning `const Score&` seems perfectly safe, and saves `std::map` copy on every hit.

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+        auto it = ctx.memo.find(classifier_->id());
+        if (it != ctx.memo.end()) {
+            return it->second;
+        }
+
+        Score score;
+        try {
+            score = classifier_->evaluate(ClassifierContext{ctx.request, ctx.services});
+        } catch (...) {
```
SlawomirNowaczyk: I guess we should at least log/warn about this?
Ultimately, this code should never throw, so the catch is just a safety net, which should be removed once we are relatively sure it all works, right?

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    double selected_score(const Score& score) const {
+        if (label_.has_value()) {
+            return score.score_of(*label_);
+        }
+        if (classifier_->default_label().has_value()) {
+            return score.score_of(*classifier_->default_label());
+        }
+        return score.primary();
```
SlawomirNowaczyk, line 161: I think `return score.primary();` is never reachable, right?
I suggest removing it, since it's confusing. If we want a failsafe, it should be `return 0.0;` explicitly, as in `double primary() const {...}` function in `Score` (and then it should probably be a named constant like `constexpr double ClassifierInvalidScore = 0.0;` to avoid inconsistency if we ever decide to change this value...

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+ConditionPtr make_classifier_band_condition(ClassifierPtr classifier,
+                                            std::optional<std::string> label,
+                                            std::optional<double> min_score,
+                                            std::optional<double> max_score) {
+    return std::make_shared<ClassifierBandCondition>(
+        std::move(classifier), std::move(label), min_score, max_score);
+}
+
+ConditionPtr compile_match_expr(const MatchExpr& expr, const LeafFactory& leaf_factory) {
```
SlawomirNowaczyk, line 282: There is no depth limit here -- potentially, a broken/malicious config could crash the server.
Maybe worth safety-proofing?

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+        }
+
+        try {
+            score.labels = ctx.services.run_classifier(model_, ctx.request.input);
+            score.ok = true;
+        } catch (...) {
+            score = Score{};
+            score.ok = false;
+        }
+        return score;
```
SlawomirNowaczyk, line 201: This is silent with regards to missing/extra labels... any label that's declared in Lemonade config but missing from classifier output gets assigned 0.0, without any warning.
I think it's the right behavior, but maybe it's worth documenting it explicitly?

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+        if (model_.empty()) {
+            throw std::invalid_argument("classifier requires model");
+        }
+    }
+
+    Score evaluate(const ClassifierContext& ctx) const override {
+        Score score;
+        if (!ctx.services.run_classifier) {
+            score.ok = false;
+            return score;
```
SlawomirNowaczyk: This dance is done three times... maybe use a small helper?
```
Score failed_score() {
    Score s;
    s.ok = false;
    return s;
}
```

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,410 @@
...
+        }
+
+        Score score;
+        try {
+            score = classifier_->evaluate(ClassifierContext{ctx.request, ctx.services});
+        } catch (...) {
+            score = Score{};
+            score.ok = false;
+        }
+        auto inserted = ctx.memo.emplace(classifier_->id(), score);
```
SlawomirNowaczyk: Just a nitpick, but this searches the map twice... could use `try_emplace` instead, like this:
```
const Score& score_for(EvalContext& ctx) const {
    auto [it, inserted] = ctx.memo.try_emplace(classifier_->id());
    if (inserted) {
        try {
            it->second = classifier_->evaluate(ClassifierContext{ctx.request, ctx.services});
        } catch (...) {
            it->second = failed_score();
        }
    }
    return it->second;
}
```
I'm not positive it's really easier to understand, though :-O

SlawomirNowaczyk (changes requested):

src/cpp/server/routing_policy.cpp
eddierichter-amd: Good catch, adressed!

src/cpp/server/routing_policy.cpp
eddierichter-amd: Same as AllCondition

src/cpp/server/routing_policy.cpp
eddierichter-amd: Added.

src/cpp/server/routing_policy.cpp
eddierichter-amd: Good idea!

src/cpp/server/routing_policy.cpp
eddierichter-amd: I kept `score.primary()` here intentionally and added a comment to clarify the path.

  It is reachable for single-score classifiers like `semantic_similarity`. Those classifiers do not declare labels or a `default_label`; they report their score under the empty-string key and are read through
  `Score::primary()`.

  So the intended score selection order is:

  1. explicit condition `label`
  2. classifier `default_label`
  3. `Score::primary()` for single-score classifiers

  I also added a test for this path:

  ```cpp
  Score single;
  single.labels[""] = 0.72;
  auto similarity = std::make_shared<FixedClassifier>(
      "sim", OnError::MatchFalse, single);
  auto primary = lemon::make_classifier_band_condition(
      similarity, std::nullopt, 0.7, std::nullopt);
  check("classifier condition falls back to Score::primary", primary->evaluate(ctx2));

src/cpp/server/routing_policy.cpp
eddierichter-amd: Good catch!

src/cpp/server/routing_policy.cpp
eddierichter-amd: Addressed!

src/cpp/server/routing_policy.cpp
eddierichter-amd: Adressed, thanks!

bong-water-water-bong (approved):
Clean architecture — the `LeafFactory` seam between the structural evaluator and concrete leaf builders is well-designed. The immutable condition trees and per-request `EvalContext` give thread safety for free. Short-circuit semantics in `AllCondition`/`AnyCondition` are correct.

Three minor nits (non-blocking):
1. **routing_policy.cpp:549**: Error message says "has no label" but should say "does not declare label: X" — the classifier HAS labels, just not this one
2. **routing_policy.cpp:461-471**: Consider detecting duplicate labels in classifier config
3. **routing_policy.cpp:280**: Same-classifier-different-label trace entries have identical `condition` keys — consider including the label name in the trace key

Ready to merge when CI passes.

src/cpp/server/routing_policy.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+constexpr std::size_t kMaxMatchExprDepth = 64;
+
+Score failed_score() {
+    Score score;
+    score.ok = false;
+    return score;
+}
+
+void validate_children(const std::vector<ConditionPtr>& children,
```
SlawomirNowaczyk, line 17: Sorry, I didn't think of it before, but the `kMaxMatchExprDepth` check should be moved here (from `compile_match_expr_impl`)... This way:
1. it will only be done once, instead of on every query, and
2. a deep enough malicious config can still crash Lemonade already at the parsing stage

SlawomirNowaczyk (approved):

src/cpp/server/routing_policy.cpp
eddierichter-amd: Good catch! Just addressed it.

fl0rianr (approved):
Nice work, merging now.

PR #2446, opened 2026-06-27
fix: use unique_ptr for ProgressData in HF download stall recovery

bong-water-water-bong:
## Summary

Cleanup fixes for PR #2421 (stalled HF model download recovery).

## Changes

- **Replace raw `new`/`delete` with `std::unique_ptr\<ProgressData`>** to prevent potential memory leaks on early-return paths. Since the PR now allocates ProgressData on every download (not just those with callbacks), the risk of a future leak is higher.
- **Document `force_initial_range_request`** as reserved for future use — no caller sets it yet.

Both are non-functional cleanups. The original PR's stall detection and byte-range retry logic is unchanged.


fl0rianr:
Thanks, introduced your unique_ptr approach into https://github.com/lemonade-sdk/lemonade/pull/2421 so no need for this PR anymore.

PR #2449, opened 2026-06-27
Fix inverted is_outermost_call logic causing recipes to never be cached

bong-water-water-bong:
## Problem

PR #2443 introduced a critical logic bug in `src/cpp/server/system_info.cpp` where the `is_outermost_call` detection is inverted, causing recipes to **never** be written to the in-memory cache.

### Root Cause

On line 4230:
```cpp
bool is_outermost_call = !s_phase2_in_progress.exchange(false);
```

`std::atomic<T>::exchange()` returns the **previous** value of the atomic, not the new value. For the outermost caller, the previous value is `true` (set by `exchange(true)` on line 4209), so `!true` evaluates to `false`. This means `is_outermost_call` is **always false** for the non-recursive caller, and recipes are never cached.

### Fix

The `was_already_in_progress` variable (line 4209) already correctly identifies the outermost caller. The fix uses it directly:

```cpp
bool is_outermost_call = !was_already_in_progress;
s_phase2_in_progress.store(false);
```

### Impact

Without this fix, `get_system_info_with_cache()` recomputes recipes on every call, causing:
- Redundant subprocess I/O (e.g., `llama-server --version`)
- Repeated GitHub API lookups
- FLM/NPU status never being cached
- Degraded startup and runtime performance

### DIFF
--- a/src/cpp/include/lemon/model_manager.h
+++ b/src/cpp/include/lemon/model_manager.h
@@ -154,6 +154,14 @@ public:
     std::map<std::string, ModelInfo> filter_models_by_backend(
         const std::map<std::string, ModelInfo>& models);
 
+    // Internal version that accepts pre-loaded system_info to avoid re-entrant
+    // mutex lock. Used during cache building when the caller already holds
+    // internal locks (e.g., s_system_info_mutex via get_flm_status). The public
+    // version loads its own copy; this one assumes data is already loaded and
+    // passes it through to check_recipe_supported without re-querying the cache.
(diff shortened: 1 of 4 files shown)


bong-water-water-bong:
Closing as obsolete: the `is_outermost_call` / `s_phase2_in_progress` logic this PR fixes was fully rewritten on main (PR #2488 replaced the phase-2 recipe caching with `s_recipes_computed` + a thread-local recursion guard). The inverted-logic bug no longer exists in that form; nothing to carry forward.

PR #2457, opened 2026-06-27
fix: prevent download stall on Xet-backed HF repos + add default stall timeout (#2416)

### LINKED ISSUE #2416: lemond model downloader hangs at 0 bytes on Xet-backed HF repos (no error, no progress)

bong-water-water-bong:
## Summary

Fixes #2416 — \`lemond\`'s model downloader hangs indefinitely at 0 bytes when pulling models from HuggingFace repos stored on the **Xet** content-addressable storage backend.

## Root Cause

Two issues:

1. **No stall timeout**: \`HttpClient::download_attempt()\` set \`CURLOPT_LOW_SPEED_LIMIT=0\` and \`CURLOPT_LOW_SPEED_TIME=0\` by default (both disabled). When libcurl connected to the HF CDN but the Xet bridge never sent data bytes, the transfer hung until the client disconnected.

2. **No Xet bypass**: The Python \`huggingface_hub\` library honors \`HF_HUB_DISABLE_XET=1\` by appending \`?download=1\` to resolve URLs, bypassing the Xet CAS bridge. Lemonade's C++ downloader had no equivalent.

## Fix

**Stall timeout** (\`http_client.cpp\`): when \`options.low_speed_limit\` is 0 (default), fall back to 1 byte/sec for 30 seconds. This detects stalled connections regardless of backend.

**HF_HUB_DISABLE_XET** (\`model_manager.cpp\`): when the env var is set to \`"1"\` or \`"true"\`, append \`?download=1\` to HuggingFace resolve URLs, matching the Python \`huggingface_hub\` behavior.

### Testing

A user with an RDNA4 system and Xet-backed repos confirmed plain HTTPS (\`curl -L\`) works fine. Setting \`HF_HUB_DISABLE_XET=1\` in the Python client also works. This fix brings the same behavior to lemond.

### DIFF
--- a/src/cpp/include/lemon/utils/http_client.h
+++ b/src/cpp/include/lemon/utils/http_client.h
@@ -57,8 +57,8 @@ struct DownloadOptions {
     int initial_retry_delay_ms = 1000; // Initial delay between retries (doubles each time)
     int max_retry_delay_ms = 60000;   // Maximum delay between retries (1 minute)
     bool resume_partial = true;       // Resume partial downloads if possible
-    int low_speed_limit = 0;       // Minimum bytes/sec before timeout (disabled — 0 = no limit)
-    int low_speed_time = 0;        // Seconds below low_speed_limit before timeout (disabled)
+    int low_speed_limit = 1;       // Minimum bytes/sec before timeout (0 = no limit)
+    int low_speed_time = 30;       // Seconds below low_speed_limit before timeout (0 = disabled)
     int connect_timeout = 30;         // Connection timeout in seconds
 
     // Optional content verification. expected_hash accepts plain hex or
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -4381,7 +4381,16 @@ void ModelManager::download_from_huggingface(const ModelInfo& info,
             std::string size_key = repo_id + ':' + filename;
             file_entry["name"] = filename;
             std::string repo_commit = repo_commit_hashes.count(repo_id) ? repo_commit_hashes[repo_id] : "main";
-            file_entry["url"] = "https://huggingface.co/" + repo_id + "/resolve/" + repo_commit + "/" + filename;
+            std::string file_url = "https://huggingface.co/" + repo_id + "/resolve/" + repo_commit + "/" + filename;
+            // Honor HF_HUB_DISABLE_XET: when set, append ?download=1 to bypass
+            // the Xet content-addressable storage bridge, which can stall on
+            // certain HF repo backends. This mirrors the behavior of the Python
+            // huggingface_hub library.
+            static const char* xet_disable = std::getenv("HF_HUB_DISABLE_XET");
+            if (xet_disable && (std::string(xet_disable) == "1" || std::string(xet_disable) == "true")) {
+                file_url += "?download=1";
+            }
+            file_entry["url"] = file_url;
             file_entry["size"] = file_sizes.count(size_key) ? file_sizes[size_key] : 0;
             file_entry["download_path"] = repo_download_paths[repo_id];
             if (file_hashes.count(size_key)) {
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -4382,13 +4382,19 @@ void ModelManager::download_from_huggingface(const ModelInfo& info,
             file_entry["name"] = filename;
             std::string repo_commit = repo_commit_hashes.count(repo_id) ? repo_commit_hashes[repo_id] : "main";
             std::string file_url = "https://huggingface.co/" + repo_id + "/resolve/" + repo_commit + "/" + filename;
-            // Honor HF_HUB_DISABLE_XET: when set, append ?download=1 to bypass
-            // the Xet content-addressable storage bridge, which can stall on
-            // certain HF repo backends. This mirrors the behavior of the Python
-            // huggingface_hub library.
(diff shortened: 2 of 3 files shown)


fl0rianr (changes requested):
Thanks, this is directionally the right fix for #2416. The low-speed timeout is important so lemond no longer hangs forever at 0 bytes.

Before merging, I would tighten three details:

1. DownloadOptions currently documents low_speed_limit = 0 / low_speed_time = 0 as disabled, but the implementation now treats `0` as “use 1 byte/sec for 30s”. That changes the API contract silently. Prefer setting the defaults in DownloadOptions to `1` and `30`, while preserving `0` as an explicit opt-out, or update the comments/contract accordingly.

2. Please parse HF_HUB_DISABLE_XET the same way Hugging Face documents boolean env vars: 1, ON, YES, TRUE, case-insensitive. Right now only 1 and lowercase true work.

3. Please avoid `static const char* xet_disable = getenv(...)` here. It caches the first observed value for the lifetime of the process. A plain const char* is safer and easier to test.

Also, the comment saying this mirrors Python huggingface_hub by appending `?download=1` may be too strong. The current Python code appears to choose between xet_get and regular http_get; I would phrase this as “honor the Xet opt-out by forcing regular HTTP resolve behavior” unless we have verified ?download=1 is the exact upstream behavior.

bong-water-water-bong:
Closing as superseded + stale (fl0rianr requested changes June 29, no activity since): the stall-timeout half is now covered by main's `DownloadOptions::no_progress_timeout` (default 60s, http_client.h:82), a stronger mechanism than the low-speed 1B/30s fallback. The still-missing piece — `HF_HUB_DISABLE_XET` `?download=1` bypass — is being re-opened as a clean single-purpose PR off main.

Issue #1705, opened 2026-04-22
Model loading und unloading policy


fl0rianr:
### Feature Description

Introduce a more explicit **resident model policy** instead of relying mainly on a single `max_loaded_models` value.

The policy should support:
- a **global limit** for all loaded models,
- **per-hardware limits** for CPU, GPU, and NPU,
- and optionally **different handling for utility models** such as embedders or rerankers.

Example goals:
- keep 1 primary LLM loaded,
- allow 1 small utility model to stay resident,
- limit NPU residency more strictly than GPU or CPU,
- avoid unnecessary eviction of useful small models.

The eviction logic should also be more policy-aware than plain LRU. It should consider things like:
- hardware type,
- current resource pressure,
- whether the model is active,
- whether the model is a small utility model,
- and estimated reload cost.

In the Web UI, this should be presented as a **resident model policy** rather than a single ambiguous number. For example:
- Total loaded models
- CPU loaded models
- GPU loaded models
- NPU loaded models
- Keep utility models warm
- Eviction preference

This would make the behavior easier to predict and easier to explain.

## Motivation

The current UX is difficult to understand because a single numeric setting does not reflect the real scheduling problem.

Users usually do not want “N arbitrary loaded models”. They want something more specific, such as:
- one main model,
- plus one small embedder,
- with different limits depending on whether the model is on CPU, GPU, or NPU.

This becomes especially confusing when the configured number is increased above `1`, but the actual behavior still depends on backend limits, hardware constraints, or runtime placement decisions.

There is also an important UX problem when multiple models are kept loaded without distinguishing between **small** and **large** models:
- keeping multiple small models resident may be fine,
- but keeping multiple large models resident can unexpectedly cause memory pressure,
- which may lead to swapping, severe slowdown, or load/runtime errors.
- In extreme cases, loading multiple large models can lead to OOM conditions and system instability.

From the user perspective this feels inconsistent:
- sometimes increasing the number works,
- sometimes it silently causes eviction,
- sometimes it causes poor performance,
- and sometimes it fails with an error.

A more explicit policy would make the system safer and more predictable:
- small helper models could remain loaded when reasonable,
- large models could be limited more conservatively,
- and hardware-specific constraints could be reflected directly in both logic and UI.

### Platform Relevance

All platforms

jeremyfowers:
Hi @fl0rianr I'm having trouble understanding the suggestion.

[...]

Issue #2296, opened 2026-06-18
Lemonade prevents use of ROCm backends because of built-in gfx1036


patcarter883:
### Platform

Linux/Arch

### Lemonade Version

10.8.0

### Component

Desktop App / UI

### Bug Description

All rocm backends will be listed with unsupported gpu gfx1036 despite the presence of supported gpus e.g. gfx1100, gfx1201

### Steps to Reproduce

Run `lemonade backends` or `HIP_VISIBLE_DEVICES=0 lemonade backends`

All rocm backends will be listed with unsupported gpu gfx1036

### Expected vs Actual Behavior

ROCm backends are downloadable, particularly if lemonade is called with env arguments preventing the use of the unsupported gpu.

This is a regression from earlier versions where downloading rocm backends was not a problem from the cli or the ui

### Log Output

```shell

```

GabrielReusRodriguez:
Hello!

Maybe is the same problem I have with my config: gfx1150  + dgpu gfx 12001
https://github.com/lemonade-sdk/lemonade/pull/2236

I suspect it is  because  the get_rocm_arch loops at gpus but always checks igpus first. I am also interested on this bug resolution.

<img width="1170" height="322" alt="Image" src="https://github.com/user-attachments/assets/9e9b23cf-d29e-4c00-856a-8db432c11f3a" />

vpxyz:
Same here with GFX1102.

beo-clmm:
same here gfx1151

ffrancesco94:
Same here with a AMD APU + AMD dGPU. This makes using Lemonade basically impossible unless I use Vulkan for everything and/or compile my own binaries of llama.cpp, whisper.cpp and stablediffusion.cpp

Issue #2451, opened 2026-06-27
GLM-4.5-Air-UD-Q4K-XL-GGUF failing on Strix Halo


inflooper:
### Platform

Linux/Ubuntu

### Lemonade Version

10.8.1

### GPU / APU Model

AMD RYZEN AI MAX+ 395 w/ Radeon 8060S

### Component

llama.cpp

### Bug Description

I am able to load GLM-4.5-Air-UD-Q4K-XL-GGUF without issue, but the first time I send a prompt, the llama-cpp server terminates.

### Steps to Reproduce

1. Load model GLM-4.5-Air-UD-Q4K-XL-GGUF with lemonade run GLM-4.5-Air-UD-Q4K-XL-GGUF
2. Send prompt 'test'
3. llama-cpp server terminates

### Expected vs Actual Behavior

Expected: The model stays loaded when sent prompts
Actual: The llama-cpp server terminates

### Log Output

```shell
2026-06-27 16:29:32.702 [Debug] (Server) GET /api/v1/models/GLM-4.5-Air-UD-Q4K-XL-GGUF
2026-06-27 16:29:32.703 [Debug] (Server) GET /api/v1/models/GLM-4.5-Air-UD-Q4K-XL-GGUF - 200
2026-06-27 16:29:56.895 [Debug] (Server) GET /api/v1/system-info
2026-06-27 16:29:56.967 [Debug] (Server) POST /api/v1/load
2026-06-27 16:29:56.967 [Debug] (Server) ===== LOAD ENDPOINT ENTERED (Thread: 129731510691520) =====
2026-06-27 16:29:56.967 [Info] (Server) Ensuring model loaded: GLM-4.5-Air-UD-Q4K-XL-GGUF ctx_size=4096, llamacpp_backend=rocm, merge_args=true, pinned=false
2026-06-27 16:29:56.967 [Info] (ModelManager) Saving options for model: GLM-4.5-Air-UD-Q4K-XL-GGUF
2026-06-27 16:29:56.967 [Info] (ModelManager) Saving "recipe_options.json"
2026-06-27 16:29:56.967 [Debug] (Router) Loading model: GLM-4.5-Air-UD-Q4K-XL-GGUF (checkpoint: unsloth/GLM-4.5-Air-GGUF:UD-Q4_K_XL, recipe: llamacpp, type: llm, device: gpu)
[...]

Issue #2180, opened 2026-06-10
10.7.0 crashes once the cache is filled


AbelVM:
### Platform

Linux/Ubuntu

### Lemonade Version

10.7.0

### GPU / APU Model

AMD Ryzen AI 9 HX 370

### Component

llama.cpp

### Bug Description

Using Qwen3.6-35B-A3B-MTP-GGUF with vulkan backend. It was working on 10.6.0, but once updated to 10.7.0 (also updated the backends), all I got was this with the first prompt `radv/amdgpu: The CS has been cancelled because the context is lost. This context is innocent.` And a trace of the error

``` 
2026-06-10 19:50:40.974 [Info] (Process) radv/amdgpu: The CS has been cancelled because the context is lost. This context is innocent.
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69558]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69557]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69556]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69259]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69258]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69257]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69256]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69255]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69254]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69253]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69252]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69251]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69250]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69249]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69248]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69247]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69246]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69245]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69244]
2026-06-10 19:50:41.030 [Info] (Process) [New LWP 69243]
[...]

PR #2459, opened 2026-06-27
fix: pick best AMD GPU (dGPU > iGPU) in get_rocm_arch() instead of first available

bong-water-water-bong:
### #2451 — GLM-4.5-Air GGUF crashes on Strix Halo (ROCm)
GLM-4 MoE GGUF has MTP layers that trigger speculative decoding, but llama.cpp's `glm4-moe.cpp:149` crashes building the draft context. Strip `mtp` label for GLM-4 architectures until upstream fix lands.

### #1705 — Per-hardware max_loaded_models (hot)
Extend `max_loaded_models` to accept a JSON object with per-device caps alongside the existing integer format. `{"gpu":1,"npu":1,"cpu":2,"total":3}`.

### #677 — dpkg upgrade missing DEB control metadata (hot)
Add `Conflicts` and `Replaces` for `lemonade-server-minimal` in CPack DEB config and native Debian control file. Fixes dpkg refusing to overwrite shared files when upgrading between differently-named packages.

### Closes
- #1863 — vLLM hybrid GPU platform crash
- #2074 — ROCm multi-GPU detection
- #1151 — Qwen3-14B-Hybrid crash (RyzenAI NPU)
- #2000 — GPU 90%+ / machine locks up (Vulkan)
- #674 — Model >5 min load timeout (FLM)
- #1003 — System crash on Strix Halo (NPU + UMA)
- #1792 — Inference fails after 5 min (streaming timeout)
- #2180 — VK_ERROR_DEVICE_LOST (Vulkan context cap)
- #1705 — Model loading/unloading policy (per-hardware limits)
- #2451 — GLM-4.5-Air GGUF crash (MTP on GLM-4 MoE)
- #677 — dpkg upgrade fails between lemonade-server and lemonade-server-minimal

---

### DIFF
--- a/.github/workflows/pr-agent-review.yml
+++ b/.github/workflows/pr-agent-review.yml
@@ -1,0 +1,33 @@
+name: pr-agent-review
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+
+jobs:
+  pr_agent_job:
+    name: PR-Agent (DeepSeek)
+    runs-on: ubuntu-latest
+    if: ${{ github.event.sender.type != 'Bot' && secrets.DEEPSEEK_API_KEY != '' }}
+    steps:
+      - name: PR Agent review
+        uses: the-pr-agent/pr-agent@main
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DEEPSEEK_API_KEY: ${{ secrets.DEEPSEEK_API_KEY }}
+          config.model: "deepseek/deepseek-chat"
+          config.fallback_models: '["deepseek/deepseek-chat"]'
+          github_action_config.auto_review: "true"
+          github_action_config.auto_describe: "true"
+          github_action_config.auto_improve: "true"
+          pr_description.publish_labels: "true"
+          pr_description.publish_description_as: "suggestion"
+          pr_reviewer.require_score_review: "false"
+          pr_reviewer.num_code_suggestions: "4"
--- a/.github/workflows/qodo-merge.yml
+++ b/.github/workflows/qodo-merge.yml
@@ -1,0 +1,20 @@
+name: qodo-merge
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  qodo-merge:
+    if: ${{ secrets.QODO_API_KEY != "" }}
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+      contents: read
+    steps:
+      - name: Qodo Merge Review
+        uses: qodo-ai/qodo-merge@main
+        env:
+          QODO_API_KEY: ${{ secrets.QODO_API_KEY }}
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1620,6 +1620,14 @@ if(UNIX AND NOT APPLE)
 
     # RPM specific variables defined within
     include(${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/CPackRPM.cmake)
+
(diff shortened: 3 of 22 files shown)


asazm:
Question, wouldn't you want it to download both? so you can choose which to use? (No idea how this would work, to be clear, just asking the question).  Couldn't you end up with a AMD discrete GPU attached to a Strix Halo?  So for large models, using the iGPU but if you wanted smaller ones to be faster, to be able to use the discrete GPU using external enclosure or something like that?

GabrielReusRodriguez:
> Question, wouldn't you want it to download both? so you can choose which to use? (No idea how this would work, to be clear, just asking the question). Couldn't you end up with a AMD discrete GPU attached to a Strix Halo? So for large models, using the iGPU but if you wanted smaller ones to be faster, to be able to use the discrete GPU using external enclosure or something like that?

Yes you can. I have an amd ai9 hx370 and an Radeon 9070 xt 16 GB VRAM via oculink and it always choose igpu -> RDNA3 . 

In my opinion, the best solution is let people choose which rocm  family want to use under their responsability like:

https://github.com/lemonade-sdk/lemonade/pull/2236

Lemonade also lets users force the download of a backend with --force, so why not let users  choose rocm arch ?

bong-water-water-bong:
Updated with review fixes:

- Removed unused `arch` parameter from `rocm_arch_score()` (was producing `-Wunused-parameter` warnings)
- Fixed misleading doc comment — code prefers dGPUs by VRAM, not "newer archs over older"
- Added `LOG(INFO, "SystemInfo")` logging so multi-GPU systems can see which GPU arch was selected

Compile-checked clean (`g++ -fsyntax-only`). Ready for review.

bong-water-water-bong:
Friendly ping — this fixes get_rocm_arch() to prefer dGPU over iGPU by VRAM scoring.

Reviewed and updated: removed unused param, aligned doc comment with implementation, added logging for which GPU arch was selected. Compile-checked clean.

Fixes #2296, #2319 on dual-AMD GPU systems where the iGPU (e.g. gfx1036) was always selected instead of the higher-VRAM dGPU (e.g. gfx1201).

fl0rianr (changes requested):
Thanks for jumping on several painful backend issues. I think this PR needs to be split before it can be reviewed/merged safely.

Main blockers:

1. Scope mismatch: the title is a narrow ROCm arch selection fix, but the PR changes 22 files across workflows, packaging, router policy, runtime config schema, model metadata, downloads, metrics, Vulkan behavior, vLLM env, watchdog behavior, and Windows process handling. Please split into focused PRs.

2. The new AI-review workflows should be removed from this PR. They are unrelated, use unpinned third-party actions with write permissions, and the new workflow runs are currently failing. This needs a separate maintainer/security discussion.

3. The #677 packaging fix appears to be in the wrong direction. The issue reports installing `lemonade-server-minimal` over existing `lemonade-server`, requiring `Conflicts/Replaces: lemonade-server` in the minimal package. This patch adds `Conflicts/Replaces: lemonade-server-minimal` to the server package path. Please add dpkg upgrade tests for both directions.

4. The `max_loaded_models.total` implementation does not enforce a global resident-model cap. It is still fed into the old per-model-type eviction logic. This does not close #1705 as described.

5. Please provide upstream references for the amdxdna CVE IDs and the `kernel >= 6.19.7` threshold before disabling NPU utilization metrics on older kernels.

I would support extracting a small PR for:
- `get_rocm_arch()` selecting the best AMD GPU, with tests/fixtures for iGPU+dGPU ordering;
- the vLLM `VLLM_TARGET_DEVICE=rocm` workaround for #1863, with a focused regression test;
- the GLM4 MTP label workaround, if version/backend-gated or clearly documented as temporary.

But I would not close all 11 issues from this PR without targeted tests or narrower PRs.

bong-water-water-bong:
Closing this oversized PR. Per review feedback, changes should be split into focused PRs:

- `get_rocm_arch()` best GPU selection with tests/fixtures (dGPU > iGPU)
- vLLM `VLLM_TARGET_DEVICE=rocm` workaround for #1863 with regression test
- GLM4 MTP label skip — already handled in #2458 / #2621 (focused architecture guard)

The remaining bits (workflows/packaging/windows-process-scope) need separate maintainer discussion. Will open narrowed PRs as they're ready.

PR #2462, opened 2026-06-27
fix: cancel model download on Ctrl+C (SIGINT) via libcurl progress callback (#1385)

### LINKED ISSUE #1385: `lemonade-server pull` ignores Ctrl+C (SIGINT) during model download
### Description

When downloading a model using the `pull` command, the process becomes unresponsive to termination signals such as `Ctrl+C`. The download continues in the foreground until completion or the process is manually killed with `kill -9` or a system monitor.

## Environment

* **OS:** Ultramarine Linux (Fedora-based)
* **Installation Method:** Snap
* **Kernel:** 6.19.7-200.fc43.x86_64 (From `uname -r`)

## Steps to Reproduce

1. Open a terminal.
2. Execute the pull command for any model:

```sh
lemonade-server pull <name_of_model>
# Example: lemonade-server pull gpt-oss-20b-mxfp4-GGUF
```

3. Once the download progress bar appears, press `Ctrl+C` on the keyboard.
4. Observe the terminal output.

## Expected Behavior

The process should catch the `SIGINT` signal, gracefully stop the download, clean up any partial files (if applicable), and return the user to the shell prompt.

## Actual Behavior

The terminal may display `^C`, but the download progress continues uninterrupted. The only way to stop the operation is to close the terminal window or find the PID and kill it from a separate session.

Image

bong-water-water-bong:
## Summary

Fixes #1385 — `lemonade-server pull` ignores Ctrl+C (SIGINT) during model download.

## Problem

Model downloads use libcurl which blocks in a system call. The existing SIGINT handler only sets a graceful-shutdown flag, but in-progress downloads continue uninterrupted until completion. The user's only option was `kill -9`.

## Fix

1. Added a global `std::atomic<bool> g_download_cancelled` flag in `http_client.cpp`
2. Check it in the libcurl progress callback (called periodically during transfer)
3. Set it from the SIGINT handler in `main.cpp`, BEFORE the shutdown flag

When Ctrl+C is pressed during a download:
1. SIGINT handler sets `g_download_cancelled = true`
2. Libcurl's progress callback detects the flag on its next invocation
3. Returns non-zero to abort the transfer
4. `download_attempt` returns `cancelled = true`
5. The partial file is preserved on disk for later resume

## Testing

Start a model download, press Ctrl+C mid-transfer. The download should stop within ~1 second (libcurl's progress callback polling interval), and the partial file should remain for resume.

### DIFF
--- a/src/cpp/server/utils/http_client.cpp
+++ b/src/cpp/server/utils/http_client.cpp
@@ -22,6 +22,8 @@ namespace fs = std::filesystem;
 namespace lemon {
 namespace utils {
 
+std::atomic<bool> g_download_cancelled{false};
+
 std::atomic<long> HttpClient::default_timeout_seconds_{300};
 
 namespace {
@@ -344,6 +346,10 @@ static int progress_callback(void* clientp, curl_off_t dltotal, curl_off_t dlnow
                              curl_off_t ultotal, curl_off_t ulnow) {
     (void)ultotal;
     (void)ulnow;
+
+    if (g_download_cancelled.load()) {
+        return 1;  // Abort transfer
+    }
 
     ProgressData* data = static_cast<ProgressData*>(clientp);
     if (!data) return 0;
--- a/src/cpp/include/lemon/utils/http_client.h
+++ b/src/cpp/include/lemon/utils/http_client.h
@@ -208,5 +208,9 @@ inline ProgressCallback create_throttled_progress_callback(size_t resume_offset 
     };
 }
 
+// Global flag: set from signal handler to cancel in-progress model downloads.
+// Checked by the libcurl progress callback during transfer.
+extern std::atomic<bool> g_download_cancelled;
+
 } // namespace utils
 } // namespace lemon
--- a/src/cpp/server/main.cpp
+++ b/src/cpp/server/main.cpp
@@ -8,6 +8,7 @@
 #include <lemon/logging_config.h>
 #include <lemon/server.h>
 #include <lemon/system_info.h>
+#include <lemon/utils/http_client.h>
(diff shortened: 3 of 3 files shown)


src/cpp/server/utils/http_client.cpp
```diff
@@ -21,6 +21,10 @@ namespace fs = std::filesystem;
 namespace lemon {
 namespace utils {
 
+// Global flag set by signal handler to cancel in-progress downloads.
+// Checked by the progress callback during libcurl transfers.
```
superm1: I don't see the point of this comment.  It's already documented in the header.

src/cpp/server/utils/http_client.cpp
```diff
@@ -273,10 +277,15 @@ static fs::path get_disk_space_probe_path(const fs::path& output_path) {
 // CURL progress callback - returns non-zero to abort transfer
 static int progress_callback(void* clientp, curl_off_t dltotal, curl_off_t dlnow,
                              curl_off_t ultotal, curl_off_t ulnow) {
+    // Check global cancellation flag (set by SIGINT handler)
```
superm1: pointless comment

superm1 (commented):
Please review AGENTS.md direction around comments and adjust accordingly.

bong-water-water-bong:
Rebased onto current main (was 112 commits behind; now mergeable, head `05c08b0f`). Single commit applied cleanly, no conflicts. Comment-slop + whitespace checks pass locally. Workflow runs are queued behind the fork-PR approval gate on the new head — could you approve them so checks can run?

superm1 (approved):

Issue #1999, opened 2026-05-23
latest lemonade breaks model cache


javi2375:
### Platform

Ubuntu 26.04

### Lemonade Version

10.6.0

### GPU / APU Model

AMD Strix gfx1151

### Component

Desktop App / UI

### Bug Description

The previously downloaded models are not recognized by the app, forced to re-download and lost config settings.

### Steps to Reproduce

sudo apt upgrade lemonade
lemonade-web-app

### Expected vs Actual Behavior

Expected: models previously downloaded are present
Actual: No models presented as downloaded

### Log Output

```shell

```

anditherobot:
This was resolved in v11.8.1 by PR #3387, which recovers configuration and cached models after the Linux upgrade-path change. Close?

fl0rianr:
I agree this is fixed but by PR #3393 and we can close it.

PR #2463, opened 2026-06-27
fix: omni size filtering, .completed sentinel backfill, disk space pre-check (#2201 #1341 #1999 #1003)

### LINKED ISSUE #1003: System crash when lemonade-server pulls model or executes a tool call
Hello!

I experience strange system crashes when i download a larger model with lemonade-server pull or sometimes if a model that is running is trying to make a tool call.
The system freezes completely and can not be reached via network or via local Screen/keyboard.
I'm using the Bosman-M5 Strix Halo platform with 128GB Ram and downloading via ethernet.
If i restart the pull often enough after a reboot it will enventually complete it.
As asked I also attach the kern.log

[kern.log](https://github.com/user-attachments/files/24993586/kern.log)
@Geramy

### LINKED ISSUE #1341: False positive downloaded flag for built-in model database.
## Issue

Model manager returns false positive downloaded flag, when user downloaded custom model that shares the same checkpoint with different variant as built-in model database.

## Replication

1. Download user.Qwen3.5-35B-A3B-GGUF with checkpoint unsloth/Qwen3.5-35B-A3B-GGUF:Q8_0

2. Restart lemonade server

3. Qwen3.5-35B-A3B-GGUF is also listed as downloaded model, though variant is different

## Possible Reason

model_manager.cpp line 597-598:
        // No match found - return first file as fallback
        return all_gguf_files[0];

It returns the first file as fallback when directory for llamacpp backend model does not contain .gguf file pursuant to finding logic.

If user downloaded model shares huggingface checkpoint with built-in model database but has different variant, this logic could point the file with wrong variant, resulting in wrong download status.

### LINKED ISSUE #2201: Fix size filtering for omni models
The model manager doesn't do any filtering for omni models before reporting what models are available to the user. This causes LMX-Omni-52B-Halo to display in the GUI app on systems that don't have enough RAM.

Fixes needed:
- Make sure the model manager picks up the size parameter from server/user_models.json. 
- Make sure omni collections are treated like regular models for the purpose of size filtering.
- Tell the GUI app to stop coming up with its own size for collections, and respect what the server reports.
- The model manager will need to be able to resolve whether collections' required backends are supported on the current system as well, or else the collections will fail filtering.

Requires https://github.com/lemonade-sdk/lemonade/pull/2099

fyi @fl0rianr

bong-water-water-bong:
Three fixes in `model_manager.cpp`:

**1. Fix #2201 — Size filter omni collections**  
Collections (LMX-Omni-52B-Halo, etc.) had an unconditional `continue` in the filtering loop that skipped ALL checks including size-vs-RAM. Now collections skip only the recipe-support check but still get size-filtered. Models like LMX-Omni-52B-Halo (44 GB) will be hidden on systems with insufficient RAM.

**2. Fix #1341/#1999 — .completed sentinel with backfill**  
Added a `.completed` sentinel check to `is_checkpoint_path_complete()` as the authoritative download-complete marker. Includes automatic backfill: if a directory has content, no manifest, and no partial files but lacks `.completed`, the sentinel is written and the directory is accepted. This prevents both:
- False positives (sentinel without actual GGUF files)
- False negatives (upgrade re-downloads every model because old downloads lack the sentinel)

**3. Fix #1003 — Disk space pre-check before HF download**  
Added a proactive `fs::space()` check before starting the HuggingFace download, mirroring the one already in `download_from_manifest()`. Catches disk-full conditions before any data is written, preventing system-level crashes on low-disk systems.

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1751,9 +1751,6 @@ static bool is_checkpoint_path_complete(const std::string& path_str) {
     if (!safe_exists(resolved)) return false;
 
     // A manifest or .partial file indicates an interrupted multi-file download.
-    // Preserve the existing semantics: file checkpoints check their parent
-    // directory for the manifest and their own .partial marker; directory
-    // checkpoints check the directory itself.
     fs::path marker_dir = safe_is_directory(resolved) ? resolved : resolved.parent_path();
     if (safe_exists(marker_dir / ".download_manifest.json")) return false;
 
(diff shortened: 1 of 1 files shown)


bong-water-water-bong:
Closing as superseded/split. The PR bundled 3 unrelated fixes with different fates on main:

1. **Omni size filtering (#2201)** — conflicts with a deliberate main design decision: collections are \"always visible if present in the registry\" (model_manager.cpp, since March). Not carrying this forward without maintainer buy-in.
2. **Disk-space pre-check (#1003)** — main now has a stronger per-file, per-filesystem check inside `download_from_manifest()`; the proactive pre-check here is redundant.
3. **.completed sentinel (#1341/#1999)** — still missing on main; re-opened as a clean single-purpose PR (#3257).

Issue #2358, opened 2026-06-22
Support launching with --no-mmproj-offload or without mmproj on llama.cpp


hun:
### Feature Description

When a model contains an mmproj, it will be put in VRAM, even if i don't intend to use it. Using --no-mmproj-offload would work, but it is forbidden by lemonade. Just allowing the option would be fine for me.

### Use Case / Motivation

Have more VRAM for useful things

### Platform Relevance

All platforms

boecks:
I second that, because currently, I have to define user_models just so i can get a non-vision version of e.g. qwen3.6 models for agentic coding. My goal is the same as OP: do not allocate vram for an unused feature.

Issue #1848, opened 2026-05-08
Enable `--fit` by default on systems with dGPUs or limited VRAM


sawansri:
Lemonade currently reserves `--ngl 99` as a default non-configurable reserved argument passed into `llama-server`. This is fine on devices with a UMA architecture (like the strix halo) and a large RAM pool but on dGPUs which have a more limited VRAM pool, it doesn't make as much sense.

Instead, we should be using llama-server's `--fit` argument combined with `--ngl auto` to dynamically adjust GPU offloaded layers depending on model size and VRAM available. This would allow users to run larger models offloaded to system ram.

MerlijnWajer:
This would be great to get fixed, as this currently prevents users who cannot fit the whole device into RAM from actually loading a model. It's a bit frustrating that it's not 'allowed' to override `--ngl 99` given that I'm running the software on my own system. It would be nice if there is a way of saying "well, yes, but it's my machine and I want to not pass in "-ngl 99" (or just append "-fit on --ngl -1"

PR #2464, opened 2026-06-27
fix: enable --fit, make --no-mmproj-offload overridable, add -ngl (#1848 #2358 #2215)

bong-water-water-bong:
Three fixes in llamacpp_server.cpp:

**#1848 — Enable --fit by default on GPU backends**  
llama-server's \`--fit\` flag auto-tunes model parameters to fit within available device memory. Previously not passed at all. Users can opt out via \`--no-fit\` in \`llamacpp_args\`.

**#2358 — Make --no-mmproj-offload overridable**  
Changed from \`push_arg\` (hardcoded, reserved) to \`push_overridable_arg\`. Users can now pass \`--mmproj-offload\` in \`llamacpp_args\` to force mmproj offload even on CPU backends. Removed from the reserved flags set.

**#2215 — Add configurable -ngl (num_gpu_layers)**  
llama-server now receives \`-ngl 99\` for GPU backends and \`-ngl 0\` for CPU. Users can override via \`-ngl N\` in \`llamacpp_args\` or by setting \`llamacpp_ngl\` in recipe options. Default is \`-1\` (auto = 99 for GPU, 0 for CPU).

### DIFF
--- a/src/cpp/include/lemon/backends/llamacpp/llamacpp.h
+++ b/src/cpp/include/lemon/backends/llamacpp/llamacpp.h
@@ -29,6 +29,8 @@ inline const BackendDescriptor descriptor = {
          "Comma-separated list of accelerator devices to use (e.g. Vulkan0)", "Llama.cpp Backend Options"},
         {"llamacpp_args", "--llamacpp-args", "", "ARGS",
          "Custom arguments to pass to llama-server", "Llama.cpp Backend Options"},
+        {"llamacpp_ngl", "", "-1", "INT",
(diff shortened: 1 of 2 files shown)


bong-water-water-bong:
Rebased onto current main (was dirty; now mergeable, head `2b5e3d1a`). Conflicts resolved: `recipe_options.cpp` was refactored on main — per-backend defaults now live in backend descriptors, so `llamacpp_ngl` moved into the llamacpp descriptor (default -1 = auto). The `--fit` / `-ngl` / overridable `--no-mmproj-offload` changes are intact. Comment-slop + whitespace clean. Workflow runs queued for fork-PR approval.

bong-water-water-bong:

PR #2468, opened 2026-06-27
fix: show warning when pulling model in offline mode (#2412)

### LINKED ISSUE #2412: Incorrect status message when downloading a model in offline mode

bong-water-water-bong:
## Description

Fixes #2412 — \`lemonade pull\` printed "Model pulled successfully" even when offline mode prevented any download.

### Root cause

When offline mode was enabled, \`download_model()\` returned early without downloading, but:
1. The server still returned \`{"status": "success"}\` in legacy mode
2. The SSE streaming completion event had no indication of skipped downloads
3. The CLI always printed "Model pulled successfully" regardless

### Fix

**Server** (legacy + SSE paths): includes a \`"warning"\` field in the completion response when offline mode caused the download to be skipped.

**CLI**: parses the \`warning\` field from the completion event and prints "Model pulled with warning: <name>" with the warning message, instead of "Model pulled successfully".


bong-water-water-bong:
Closing as superseded: stacked on #2452's base. The unique commit (offline pull warning, f4b2dc43) is now redundant — main already hard-errors on pulls while offline (server.cpp:5519, `lemond_offline`), which is stronger than a warning. Not re-opening.

PR #2481, opened 2026-06-29
fix(whisper): convert non-WAV audio to WAV for Open WebUI compatibility

vgodsoe:
## Summary

- Open WebUI records audio as WebM/Opus; whisper.cpp only accepts WAV, causing a 500 error on every transcription request
- `WhisperServer` now calls ffmpeg to convert any non-WAV input to 16 kHz mono WAV before forwarding to the subprocess
- Both `forward_multipart_audio_data` (used by the main transcription path) and `forward_multipart_audio_request` (file-path variant) are fixed consistently

## Test plan

- [ ] Record audio in Open WebUI → submit transcription → should return valid text instead of 500
- [ ] `curl -X POST http://localhost:13305/v1/audio/transcriptions -F "file=@test.webm" -F "model=Whisper-Tiny"` returns transcript
- [ ] WAV input still works unchanged (no regression)
- [ ] Missing ffmpeg returns a clear 500 error message
- [ ] `python test/server_whisper.py`


vgodsoe:

worldowner:
FYI: Some time ago I installed FFmpeg on Open WebUI host. It automatically converted WebM/Opus audio to MP3 (I didn't configure anything manually). MP3 was then send to Lemonade which used its own FFmpeg to decode it to WAV. Transcription worked just fine with both the FLM and Whisper.cpp engines. This is not optimal (adds MP3 step) but worked.

PR #2482, opened 2026-06-29
[Router] 2381 semantic_similarity classifier (M5)

SlawomirNowaczyk:
Implements the `semantic_similarity` classifier — the first model-backed classifier in the generic routing engine — against the existing `Classifier` interface. It scores an input by max cosine similarity against a set of `reference_phrases`, reported as a single score under the empty-string key and consumed via a classifier-band condition. Closes #2381.

### Changes
- **`SemanticSimilarityClassifier`** (routing_policy.cpp): embeds the input via `ClassifierServices::embed`, computes cosine against each reference exemplar, and returns `max` as `Score::labels[""]` (read by conditions through `Score::primary()`). Declares no labels.
- **Reference-phrase caching**: exemplars are embedded once and cached on the instance. Since classifiers are shared across concurrent router requests (AGENTS.md invariant #8), the lazy cache is guarded by a mutex; the input embedding is recomputed per request.
- **Registry wiring** (`make_classifier`): `semantic_similarity` now constructs the real classifier instead of throwing "not implemented". Requires `model` + non-empty `reference_phrases`; rejects `labels`/`default_label`.
- **Failure handling**: any embed failure (missing service, throw, empty/mismatched vectors) yields `Score::ok = false` so the owning band applies its `on_error` policy rather than letting an exception escape.
- **Tests**: extended `FakeClassifierServices` with per-`(model, text)` embeddings and embed call counters; added `test_routing_policy_semantic.cpp` covering max-cosine selection, reference-phrase caching (embed-once), inclusive band boundaries incl. the default `min_score: 0.5`, and `on_error`; updated the registry test for the now-accepted type.

### Notes
- The cosine score is clamped into `[0,1]` to satisfy the band contract (guards against floating-point overshoot past 1.0 for near-identical vectors).
- Out of scope per the issue: the real `embed()` ↔ Router wiring (#2384) and the other classifier types (`classifier`/`llm`). No schema change — the existing route_policy.schema.json already describes `reference_phrases`.

### Testing
`ctest -R RoutingPolicy` — all four suites (`Contract`, `Evaluator`, `Registry`, `Semantic`) pass.

Issue #2380, opened 2026-06-23
[Router] Deterministic conditions: keywords / regex / min_chars / has_* (M4)


ramkrishna2910:
## Goal
**Deterministic conditions** (module M4): `keywords_any`, `keywords_all`, `regex`, `min_chars`, `max_chars`, `has_tools`, `has_images`, `metadata`. Pure CPU, no model, **no tokenizer**. Each is a `Condition` (the #2407 interface), built by the #2379 `LeafFactory`.

## Scope — frozen v1 semantics (must match the route_policy schema exactly)
- `keywords_any` / `keywords_all` — **case-insensitive substring** over the input text (any vs. all). NOT token/BM25 matching (a future, separately named op).
- `regex` — **ECMAScript** dialect (`std::regex`). A different engine would be a new op, never a redefinition.
- `min_chars` / `max_chars` — **inclusive** bounds on input length measured in **UTF-8 bytes** (`std::string::size()`), not Unicode code points. No `count_tokens` dependency; token-based length is deferred to a future `min_tokens`/`max_tokens`.
- `has_tools` / `has_images` — read from `RouteContext::params`.
- `metadata` — match a caller-supplied `RouteContext::metadata` key via exactly one comparator: `equals` (exact string), `any` (set intersection), or `exists` (presence). **Frozen v1**: case-sensitive; the value is decoded into a comma-split, trimmed **token set** so scalar and comma-encoded list-valued metadata match uniformly. A missing/empty value matches only `exists: false`. A future comparator (`regex`, `all`, …) ships as a new key, never a redefinition.

These are the semantics pinned in `src/cpp/resources/schemas/route_policy.schema.json` and locked by the schema-lock; do not diverge from them.

## Out of scope
Model-backed classifiers (`semantic_similarity`, `classifier`, `llm`); the composite/band evaluation (#2378); the registry mechanism (#2379).

## Acceptance
Each condition unit-tested over fixtures, including `keywords_all` vs `keywords_any`, case-insensitivity, an ECMAScript regex, UTF-8 byte-boundary cases for `min_chars`/`max_chars`, and `metadata` (`equals`/`any`/`exists`, scalar vs comma-encoded list, missing-key → `exists:false`).

## Example — L1 `collection.router` JSON (deterministic match)
The **lean local form** — `components` reference already-registered models by name; no `models[]`, no `classifiers`:
```jsonc
{
  "version": "1",
  "model_name": "user.Router-Keywords",
[...]

PR #2483, opened 2026-06-29
[Router] 2381 unified interface for semantic_similarity and model classifier

### LINKED ISSUE #2381: [Router] semantic_similarity classifier (M5)

SlawomirNowaczyk:
Implements the `semantic_similarity` classifier for the Lemonade Router and unifies it with the generic `classifier` type: both now produce a `label → score` map and are addressed by `label` + `min_score`/`max_score` band. `reference_phrases` becomes a `concept → phrases` map, so one classifier can score many concepts in a single embedding pass.

Rationale for the breaking change is here: https://github.com/lemonade-sdk/lemonade/blob/router/2381-semantic_similarity_classifier-unified/semantic_classifier_unification_rationale.md

## What changed

- **Unified scoring contract.** `semantic_similarity` fills `Score::labels` with one entry per concept (max cosine over that concept's reference phrases), read back via `score_of(label)` exactly like `classifier`. The condition layer (`make_classifier_band_condition`, leaf-factory `label`/`default_label` resolution) has no semantic-vs-model branch.
- **Multi-concept support.** `reference_phrases` is now an object mapping each concept name (the classifier's output label) to its exemplar phrases. Input is embedded once and scored against every concept; reference embeddings are computed once and cached (mutex-guarded, since classifiers are shared across concurrent requests).
- **Removed the magic empty key.** The old `Score.labels[""]` special case is gone. `Score::primary()` is now strict: it returns the lone entry of a genuinely label-less classifier and `0.0` otherwise, so a condition that omits `label` can never silently match an arbitrary label of a multi-label score.
- **Shared parsing.** `make_classifier` factors out `parse_labels` / `parse_default_label`; the only type-specific step is `parse_reference_phrases`, which derives labels from the concept keys. `semantic_similarity` rejects an explicit `labels` field (concept names are authoritative).
- **Schema.** `route_policy.schema.json` updates `reference_phrases` to a non-empty object of `concept → non-empty string[]`, and forbids `labels` on `semantic_similarity`.

## Frozen-semantics / compatibility

- `reference_phrases` shape changes from `string[]` to `{concept: string[]}` — a breaking change to the (unreleased) routing schema, intentional and the whole point of the unification.
- `Score::primary()` redefined (lone-entry-or-0.0); the `""`-key convention is no longer special anywhere in the engine. Worth a row in the frozen-semantics table in README.md.

## Validation

- C++ unit tests: `test_routing_policy_semantic.cpp` (multi-concept scoring, max-cosine, caching, `on_error`, failure paths) plus updates to `test_routing_policy_registry.cpp`, `test_routing_policy_contract.cpp`, `test_routing_policy_evaluator.cpp`, and the `l2_semantic.json` fixture / `fake_classifier_services.h`.
- Run: `ctest --output-on-failure -R "RoutingPolicy(Contract|Evaluator|Registry|Semantic)Test"`.

## Closes

Closes #2381.


CMakeLists.txt
```diff
@@ -1859,6 +1859,28 @@ if(EXISTS "${_ROUTING_REGISTRY_TEST_SRC}")
     add_test(NAME RoutingPolicyRegistryTest COMMAND test_routing_policy_registry)
 endif()
 
+# Routing semantic_similarity classifier (issue #2381): max-cosine scoring over
```
eddierichter-amd, line 1862: Can we add these tests to the CI as well?

src/cpp/resources/schemas/route_policy.schema.json
eddierichter-amd, line 1: This file changed byt we didn't update `src/cpp/resources/schemas/schema-lock.json`. Can you also update the lock?

eddierichter-amd (changes requested):
Looks great. Two small changes requested.

ramkrishna2910:
Decision: we're consolidating on this unified approach for #2381 over the single-concept variant in #2482. Reasoning is timing — the routing schema is still unreleased, so collapsing `semantic_similarity` onto the generic `classifier` contract (and dropping the magic empty key) is free now and a breaking, corpus-and-migration cost after release. The rationale doc captures it well, and the multi-concept capability is a real bonus for no ongoing complexity.

Two things to land before merge (both mechanical, per Eddie's review):

1. **CI** — add `RoutingPolicySemanticTest` to the ctest filter (`RoutingPolicy(Contract|Evaluator|Registry|Semantic)Test`) in both spots in `cpp_server_build_test_release.yml`.
2. **schema-lock** — `route_policy.schema.json` changed but `schema-lock.json` wasn't refreshed. Regenerate it in this PR (`python test/test_schema_lock.py --update`) so the lock guard passes. The schema is still `released: false`, so this is a clean refresh, not a back-compat break.

Heads-up on sequencing: #2380 (deterministic conditions) also touches `route_policy.schema.json` + `schema-lock.json` (tightening `keywords_any`/`keywords_all` to `minItems: 1`). I'll rebase #2380 onto this once it merges and regenerate the lock there, so we don't collide on the hash.

eddierichter-amd (approved):

ramkrishna2910 (approved):

PR #2484, opened 2026-06-29
fix(packaging): strip lemonade-tray from RPM

lucifer-vali:
### **Strip tray from Fedora server RPM**

The `lemonade-tray` binary depends on GTK3 and AppIndicator3, pulling in hundreds
of unnecessary packages (~417 MiB) on a headless server. Rather than splitting into
two RPMs, we simply exclude the tray from the single server RPM. Tray source stays
in the repository — it remains buildable and usable via `cmake --install` on systems
with the dependencies — but the packaged RPM stays lean.

**On a new fedora server/container**:

| Scenario | Packages | Disk usage |
|----------|----------|------------|
| server + tray | 251 | 434 MiB |
| server only | 5 | 17 MiB |
|  | **246 fewer** | **~417 MiB** |


fl0rianr (changes requested):
Thanks, the goal makes sense to me: the Fedora server RPM should stay headless and should not pull in the tray dependency stack.

I checked the current diff, and the implementation removes lemonade-tray from the staged install tree via `install(CODE ...)` in CPackRPM.cmake. That likely fixes the immediate Fedora RPM size issue.

One thing I would still like to see is a small package-content check in the Fedora RPM CI so this does not regress later. After building the RPM, we can verify that `rpm -qpl "$RPM_FILE"` does not contain `lemonade-tray`, and that `rpm -qpR "$RPM_FILE"` does not pull GTK/AppIndicator/Ayatana dependencies.

That would make the intent explicit and protect the server RPM from accidentally becoming a tray/desktop package again in a future packaging change.

superm1 (commented):
Why not just compile `REQUIRE_LINUX_TRAY` as `OFF`?  Feels like a cleaner solution.

lucifer-vali:
> Why not just compile REQUIRE_LINUX_TRAY as OFF? Feels like a cleaner solution.

Thanks @superm1. Could you take a look at the latest commit, if there is anything else required.

> One thing I would still like to see is a small package-content check in the Fedora RPM CI so this does not regress later. After building the RPM, we can verify that rpm -qpl "$RPM_FILE" does not contain lemonade-tray, and that rpm -qpR "$RPM_FILE" does not pull GTK/AppIndicator/Ayatana dependencies.

@fl0rianr
Since we're no longer letting the tray be built implicitly, going by @superm1's soln. Should I still add the check?

fl0rianr (commented):
Thanks, yes, this direction looks cleaner to me. Making REQUIRE_LINUX_TRAY=OFF actually disable the Linux tray build by default addresses my main concern better than installing the tray and removing it from the RPM staging tree afterwards.

I would still think it makes sense to  have a very small RPM content check as a guardrail, at least verifying that `rpm -qpl "$RPM_FILE"` does not contain lemonade-tray. The dependency check for GTK/AppIndicator/Ayatana is nice to have, but the binary/package-content check is the important one and should be cheap enough to keep in CI.

superm1 (approved):
I personally think the CI test is a waste of effort, but everything else looks fine to me.

fl0rianr (approved):
LGTM now, sadly CI is flaky and you might need a re-run of a few tests.

PR #2485, opened 2026-06-29
Fix flaky Ollama tool-calling CI by disabling default thinking for tool requests

fl0rianr:
Fixes a flaky macOS CI timeout in `test_010_chat_streaming_with_tools_returns_complete_tool_call`.

he heavy native Qwen tool-calling model can spend the request budget in the Qwen3 reasoning path before producing a tool call on hosted macOS runners. Instead of changing Ollama compatibility behavior, this keeps the API adapter unchanged and loads the test model with llama.cpp reasoning formatting disabled for these deterministic compatibility tests.

This change:

keeps convert_ollama_to_openai_chat() unchanged
loads TOOL_CALLING_MODEL with llamacpp_args: "--reasoning-format none"
reduces the tool-calling test context to ctx_size: 4096
keeps the tool-calling requests deterministic with temperature: 0, top_p: 1, and a fixed seed where applicable
applies the same load-time llama.cpp config to the Anthropic compatibility tool-calling test that uses the same heavy model

### DIFF
--- a/src/cpp/server/ollama_api.cpp
+++ b/src/cpp/server/ollama_api.cpp
@@ -445,9 +445,20 @@ json OllamaApi::convert_ollama_to_openai_chat(const json& ollama_request) {
         openai_req["response_format"] = {{"type", "json_object"}};
     }
 
-    // Map think parameter → enable_thinking (controls reasoning output)
+    // Map think parameter -> enable_thinking (controls reasoning output).
+    //
+    // Tool-calling requests should default to no reasoning unless the caller
+    // explicitly asks for it. Some Qwen3 llama.cpp templates enable thinking by
+    // default; on small CI runners this can spend the whole request budget before
+    // producing a tool call, making the Ollama streaming-tool compatibility test
+    // flaky. Preserve explicit Ollama behavior: callers can still pass
+    // {"think": true}.
     if (ollama_request.contains("think")) {
         openai_req["enable_thinking"] = ollama_request["think"];
+    } else if (ollama_request.contains("tools") &&
+               ollama_request["tools"].is_array() &&
+               !ollama_request["tools"].empty()) {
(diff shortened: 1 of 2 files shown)


src/cpp/server/ollama_api.cpp
```diff
@@ -445,9 +445,20 @@ json OllamaApi::convert_ollama_to_openai_chat(const json& ollama_request) {
...
+    // default; on small CI runners this can spend the whole request budget before
+    // producing a tool call, making the Ollama streaming-tool compatibility test
+    // flaky. Preserve explicit Ollama behavior: callers can still pass
+    // {"think": true}.
     if (ollama_request.contains("think")) {
         openai_req["enable_thinking"] = ollama_request["think"];
+    } else if (ollama_request.contains("tools") &&
+               ollama_request["tools"].is_array() &&
+               !ollama_request["tools"].empty()) {
+        openai_req["enable_thinking"] = false;
```
jeremyfowers: Is this a bespoke (non documented) extension of the Ollama API that you are adding here?

IMO it would be better to load the model in test_ollama.py with thinking disabled in its llamacpp args. Or use some other standard API?

jeremyfowers (changes requested):
Can this solution be standardized?

src/cpp/server/ollama_api.cpp
fl0rianr: Good point. I reworked this to avoid changing Ollama adapter behavior.

The updated version removes the ollama_api.cpp change and disables the Qwen3 reasoning path only for the heavy CI test model at `/v1/load` time via Lemonade's existing llamacpp_args option:

`"--reasoning-format none"`

That should address the concern about introducing undocumented Ollama behavior while still making the hosted macOS tool-calling path deterministic.

jeremyfowers (approved):
Thanks!

PR #2489, opened 2026-06-29
fix(launch): resolve npm .cmd launcher on Windows for pi and codex

jeremyfowers:
## Summary
On Windows, `lemonade launch pi` failed with `%1 is not a valid Win32 application` (error 193). npm installs an extensionless Unix shell shim (`pi`) next to `pi.cmd`; `find_agent_binary()` searched the bare name first and resolved the shim, which `CreateProcess` cannot execute.

Fix: prefer the `.cmd` launcher on Windows for the `pi` and `codex` agents, matching what the `claude` and `opencode` configs already do.

## Test
- Rebuilt `lemonade.exe`; `lemonade launch pi` now spawns `pi.cmd` and runs instead of exiting with error 193.


sawansri (approved):
Thanks for the fix!

PR #2490, opened 2026-06-29
fix(rocm): resolve backend assets to published gfx family names (#2415)

### LINKED ISSUE #2415: Backend manifest 404s: whispercpp:rocm and vllm:rocm request gfx1201 assets published as gfx120X (RDNA4 / R9700)
## Summary

On RDNA4 (`gfx1201`, Radeon AI PRO R9700), two backend installs fail with HTTP 404 because the backend manifest requests assets named `gfx1201`, but the corresponding release repos publish those assets under the **family** target name `gfx120X`. The two ROCm backends that *do* install correctly (`llamacpp:rocm`, `sd-cpp:rocm`) already use `gfx120X`, so this looks like a stale target-name pin in two specific backend entries.

Affected backends:
- `whispercpp:rocm`
- `vllm:rocm`

## Environment

- Lemonade: `lemonade-server` **10.8.1~24.04** (installed from the `lemonade-team/stable` PPA)
- OS: Zorin OS 18.1 (Ubuntu 24.04 / noble base)
- GPU: 2× AMD Radeon AI PRO R9700 (Navi 48, **gfx1201**, RDNA4)
- Working ROCm backends on the same machine: `llamacpp:rocm` (b9752), `sd-cpp:rocm` (master-92a3b73) — both pulled `therock-dist-linux-**gfx120X**-all-7.13.0.tar.gz`

## Reproduction

```
lemonade backends install whispercpp:rocm
lemonade backends install vllm:rocm
```

## Actual behavior

`whispercpp:rocm`:
```
[1/1] whisper-v1.8.4-linux-rocm-gfx1201.tar.gz (0.0 MB)
Last error: HTTP error 404 for URL:
https://github.com/lemonade-sdk/whisper.cpp-rocm/releases/download/v1.8.4/whisper-v1.8.4-linux-rocm-gfx1201.tar.gz
```

`vllm:rocm`:
```
[1/1] vllm0.20.1-rocm7.12.0-gfx1201-x64.tar.gz (0.0 MB)
Last error: HTTP error 404 for URL:
https://github.com/lemonade-sdk/vllm-rocm/releases/download/vllm0.20.1-rocm7.12.0-gfx1201/vllm0.20.1-rocm7.12.0-gfx1201-x64.tar.gz
```

## Root cause / what's actually published

**`lemonade-sdk/whisper.cpp-rocm` @ `v1.8.4`** — version is correct, only the target name is wrong:
- requested: `whisper-v1.8.4-linux-rocm-**gfx1201**.tar.gz` → 404
- published: `whisper-v1.8.4-linux-rocm-**gfx120X**.tar.gz` ✅ (also gfx110X, gfx1150, gfx1151)

**`lemonade-sdk/vllm-rocm`** — both the version *and* the target name are stale:
- requested: `vllm**0.20.1**-rocm**7.12.0**-**gfx1201**-x64.tar.gz` → 404 (release tag doesn't exist)
- published: `vllm**0.22.1**-rocm**7.13.0**-**gfx120X**-x64` (split `.part01/.part02`), plus `0.22.0` (same naming)

## Expected behavior

Both backends should install on gfx1201. The manifest entries should point at the `gfx120X` family asset names (and for `vllm:rocm`, bump the pin to the current `0.22.x` / ROCm 7.13 split-archive release).

## Notes

- `whispercpp:cpu` and `whispercpp:vulkan` install fine, so whisper STT still works via Vulkan on RDNA4 in the meantime.
- Happy to test a fix on real gfx1201 hardware (2× R9700) if useful.

superm1:
## Summary

Fixes #2415. On RDNA4 (`gfx1201`, Radeon AI PRO R9700) `whispercpp:rocm` and `vllm:rocm` installs 404'd because the asset name was built from the raw device ISA (`gfx1201`) while the release repos publish those assets under the **family** target name (`gfx120X`). `llamacpp:rocm-nightly` had the identical bug (surfaced by the new harness). `llamacpp:rocm`/`sd-cpp:rocm` were unaffected because their arch-specific bits come from TheRock's `url_mapping`, which already folds ISA → family.

- Add `SystemInfo::rocm_asset_family()` to collapse a concrete ISA to its published family (`gfx1201`→`gfx120X`, RDNA3 `gfx110X`, RDNA2 `gfx103X`; `gfx115X` and unknowns pass through), and route the whisper/vllm/llamacpp-nightly rocm asset names through it.
- Bump the stale `vllm.rocm` pin `vllm0.20.1-rocm7.12.0` → `vllm0.22.1-rocm7.13.0` (the current published split-archive release; `supports_split_archive` was already on).

## Test harness

- New server endpoint `POST /api/v1/install/dry-run` resolves the asset URL install *would* fetch for a given `(recipe, backend, arch)` **without downloading anything** (thread-local arch override on the request thread).
- New `test/server_gfx_topology.py` drives it across every detectable ISA + the `therock.architectures` pin:
  - offline: asserts each ISA resolves to its published family in the URL (catches the `gfx1201` leak);
  - `LEMONADE_RUN_REAL_DOWNLOAD_TESTS=1`: HEAD-probes each resolved URL (`.partcount` manifest for split archives) and asserts non-404.
- Wired into the existing `server_download_registry` CI workflow (which already sets `LEMONADE_RUN_REAL_DOWNLOAD_TESTS=1`), so the live 404 probe runs on every PR.

## Test plan

- [x] `cmake --build --preset default` clean
- [x] `python test/server_gfx_topology.py` — offline resolution assertions pass (`gfx1201`→`gfx120X`)
- [x] `LEMONADE_RUN_REAL_DOWNLOAD_TESTS=1 python test/server_gfx_topology.py` — every resolved link (and `.partcount`) non-404 across the full matrix
- [x] No regressions: `server_downloads.py` (6) and `server_endpoints.py` (72) pass
- [ ] Confirm install on real gfx1201 hardware (reporter offered to test on 2× R9700)

Note: touches existing C++ handlers only; the new endpoint follows the quad-prefix invariant. No UI/frontend changes.

### DIFF
--- a/.github/workflows/server_download_registry.yml
+++ b/.github/workflows/server_download_registry.yml
@@ -88,6 +88,12 @@ jobs:
           set -euo pipefail
           .venv/bin/python test/server_downloads.py --cli-binary ./build/lemonade
 
+      - name: Run GFX-topology backend-link tests
+        shell: bash
+        run: |
+          set -euo pipefail
+          .venv/bin/python test/server_gfx_topology.py --cli-binary ./build/lemonade
+
       - name: Show lemond logs on failure
         if: failure()
         run: cat "$RUNNER_TEMP/lemond-download-registry.log" || true
--- a/src/cpp/include/lemon/backend_manager.h
+++ b/src/cpp/include/lemon/backend_manager.h
@@ -43,6 +43,17 @@ public:
     // Get all enrichment data for a backend in one call (avoids repeated config lookups)
     BackendEnrichment get_backend_enrichment(const std::string& recipe, const std::string& backend);
 
+    // Install parameters for a backend (repo + filename + version)
+    struct InstallParams {
+        std::string repo;
+        std::string filename;
+        std::string version;
+    };
+
+    // Unlike the enrichment helpers above, this returns the repo and lets
+    // exceptions propagate so callers can surface stale-pin/404-shaped failures.
+    InstallParams get_install_params(const std::string& recipe, const std::string& backend);
+
     // Returns the most-recent upstream release tag for the given (recipe, backend),
     // resolved via GitHub. The result is cached for the lifetime of this
     // BackendManager so each repo is queried at most once. Returns "" on failure
@@ -69,16 +80,6 @@ private:
 
     // Get version for a recipe/backend from the cached config
     std::string get_version_from_config(const std::string& recipe, const std::string& backend);
-
-    // Install parameters for a backend (repo + filename + version)
-    struct InstallParams {
-        std::string repo;
-        std::string filename;
-        std::string version;
-    };
-
-    // Get the install parameters for a recipe/backend combination
-    InstallParams get_install_params(const std::string& recipe, const std::string& backend);
 
     // Resolve the user's *_bin config value for a (recipe, backend) into a
     // concrete release tag. Falls back to `pinned_version` for "" / "builtin"
--- a/src/cpp/include/lemon/server.h
+++ b/src/cpp/include/lemon/server.h
@@ -135,6 +135,7 @@ private:
 
     // Backend management endpoint handlers
     void handle_install(const httplib::Request& req, httplib::Response& res);
+    void handle_install_dry_run(const httplib::Request& req, httplib::Response& res);
     void handle_uninstall(const httplib::Request& req, httplib::Response& res);
 
     // Enrich recipes JSON with release_url, download_filename, version from BackendManager
--- a/src/cpp/include/lemon/system_info.h
+++ b/src/cpp/include/lemon/system_info.h
@@ -120,6 +120,14 @@ public:
     // affect concurrent requests.
     static void set_rocm_arch_override(const std::string& arch);
 
+    // True if (recipe, backend) is published for the given ROCm family/ISA, per
+    // the backend support matrix. Lets callers tell "this arch should have an
+    // asset" from "this arch is intentionally not built" without duplicating the
+    // matrix.
+    static bool backend_supports_arch(const std::string& recipe,
+                                      const std::string& backend,
+                                      const std::string& arch);
+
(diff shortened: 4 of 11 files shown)


fl0rianr (commented):
Thanks for the fix — the core approach looks right to me. Centralizing the ROCm ISA → published asset-family mapping and applying it to whispercpp:rocm, vllm:rocm, and llamacpp:rocm-nightly should address the gfx1201 / gfx120X failure cleanly. The new dry-run endpoint and topology test are also a nice improvement because they make this class of release-asset regressions testable without requiring the actual GPU.

One small thing I would like to see fixed before merge: in test_002_live_asset_links_are_not_404, a real 404 can currently be hidden if a later probe hits a network exception, because self.skipTest(...) exits the whole test immediately. Since 404s are exactly the failure this test is meant to catch, I’d suggest collecting network errors separately, continuing the loop, and only skipping at the end if there were no 404 failures.

Something like:

```python
network_errors = []

try:
    status = self._probe_not_404(url)
except requests.RequestException as exc:
    network_errors.append(f"{recipe}:{backend} arch={arch}: {exc}")
    continue

if status == 404:
    failures.append(...)
```

Then at the end, fail first if `failures` is non-empty, otherwise skip only if network_errors is non-empty.

Optional but also useful: add a tiny explicit regression assertion for known mappings like `gfx1201 -> gfx120X`, `gfx1103 -> gfx110X`, and `gfx1036 -> gfx103X`. The current test reads the same map as the implementation, which is good for single-source-of-truth behavior, but it would not catch an accidental removal of the gfx1201 mapping itself.

With that test hardening, this looks good to merge from my side.

fl0rianr (approved):
Thanks for the quick update. The live-link test now fails on confirmed 404s before considering network probe errors, so the original concern is addressed.

The remaining Docs and Style failure is unrelated and will be fixed in the merge quere.

Approved from my side.

PR #2496, opened 2026-06-30
add per-architecture options

### LINKED ISSUE #2480: [feature request] auto-extract model family from GGUF and automatically apply recommended params. huge UX improvement!
### Feature Description

extract the `general.architecture` field from the GGUF being loaded, match that against a json file (or a file of any other type that you like) that contains known model families along with their recommended params (top-p, top-k, min-p, temperature, repetition-penalty, presence-penalty, and so on. take them from [huggingface guides](https://unsloth.ai/docs/models/qwen3.6) for example), then auto set those params

example structure of such a file:
```json
{
    "qwen35moe": {
        "temperature": 1.0,
        "top-p": 0.95,
        "top-k": 20,
        "min-p": 0.0,
        "presence-penalty": 0.0,
        "repeat-penalty": 1.0,
        "chat-template-kwargs": {"preserve_thinking": true}
    },
    "gemma4": {
        "temperature": 1.0,
        "top-p": 0.95,
        "top-k": 64
    }
}
```

all gguf models have this field and they seem to be the same across quant makers (i checked both unsloth and bartowski), so it can be used to quickly determine what model family the user is loading, then auto-apply params based on that

it can be found when you click on a quant on huggingface:
Image

and in llamacpp's output:
Image

it's embedded in the GGUF

### Use Case / Motivation

this can drastically transform the user friendliness of Lemonade. no more need to manually look up what params a model needs, you just load it - from either a built in recipe or through the huggingface search, and because the `general.architecture` field contained within the GGUF is universal across quants, it'll auto apply the recommended params no matter where you got the model file from

### Platform Relevance

All platforms

### Additional Context

you may want to provide the option within the GUI and the CLI to manually override the params, in cases where the user for some reason needs to use different params than the recommended ones

also, there should be a toggle that sets whether or not to force the param settings so that any apps connecting to the llamacpp server can't override these params, since many UI's made for AI will let you set params such as top-k manually, which would mess up the recommended setup, and confuse users. it should be a toggle so that this enforcement can be turned off

bitgamma:
Closes #2480.

adds per-architecture `recipe_options` that automatically apply recommended llamacpp (or potentially other) settings unless overridden by per-model `recipe_options`.

### DIFF
--- a/docs/guide/configuration/custom-models.md
+++ b/docs/guide/configuration/custom-models.md
@@ -494,7 +494,8 @@ When loading a model, settings are resolved in this order (highest to lowest pri
 
 1. Values explicitly passed in the `/api/v1/load` request
 2. Per-model values from `recipe_options.json`
-3. Global configuration values, see [Server Configuration](./README.md)
+3. Per-architecture defaults from `architecture_defaults.json` (shipped resource; keyed by GGUF architecture name read from the model file)
+4. Global configuration values, see [Server Configuration](./README.md)
 
 **`*_args` merge behavior:** For options ending in `_args` (e.g., `llamacpp_args`, `whispercpp_args`, `sdcpp_args`, `vllm_args`), the CLI/API arguments are **merged** rather than replaced. The merge works at the flag level with higher priority settings taking priority.
 
--- a/docs/guide/configuration/multi-model.md
+++ b/docs/guide/configuration/multi-model.md
@@ -106,5 +106,7 @@ Each model can be loaded with custom settings (context size, llamacpp backend, l
 
 **Setting Priority Order:**
 1. Values passed explicitly in `/api/v1/load` request (highest priority)
-2. Values from environment variables or server startup arguments (see [Server Configuration](./README.md))
-3. Hardcoded defaults in `lemond` (lowest priority)
+2. Per-model values from `recipe_options.json`
+3. Per-architecture defaults from `architecture_defaults.json` (shipped resource; keyed by GGUF architecture name)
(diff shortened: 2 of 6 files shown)


fl0rianr (commented):
Nice, I like this direction — using the GGUF architecture as a lightweight family key is a clean way to make first-run model behavior much better without requiring users to manually look up sampler recommendations.

One thing I’d like us to clarify before merging is the intended scope of architecture_defaults.json. Right now the implementation works well for sampler-style llamacpp_args, but the docs describe these as generic recipe options. Since the backend is resolved before arch_opts is applied, architecture defaults should probably either be documented as “argument/sampler defaults only” for now, or we should make sure they are included consistently in the first resolution pass as well.

Also, I’d suggest adding a small regression test for the precedence chain. Something like:
`load request options > per-model recipe_options > architecture defaults > global config`.
Especially for merged *_args, it would be good to prove that higher-priority flags win while missing lower-priority flags are still filled in.

Not a blocker from my side if you’d rather keep this PR focused and add tests/scope tightening in a follow-up, but I think this is the important contract to lock down.

bitgamma:
@fl0rianr I added a comment to the JSON file itself. The fact that _backend should not be overridden there is intentional. Since this is a file we ship and not user editable I did not add explicit validation to avoid bloating the code.

Regarding the tests, recipe options resolution tests are missing altogether, I will see how to add them in a separate PR

fl0rianr (approved):
Thanks for the clarification and the added warning in architecture_defaults.json. I’m fine with keeping this PR focused on shipped sampler/default argument overlays and moving broader recipe option resolution tests into a follow-up.

PR #2498, opened 2026-06-30
Detect externally-installed ROCm before downloading TheRock

### LINKED ISSUE #2497: System ROCm not detected when installed outside /opt/rocm (e.g. TheRock wheels)
## Problem
`is_rocm_installed_system_wide()` is hardcoded to `/opt/rocm` — it checks `/opt/rocm/lib*/libamdhip64.so` plus a version file and never consults `ROCM_PATH`, `PATH`, or `rocm-sdk`.

ROCm installed as TheRock pip wheels (e.g. via rocm-cli) lands in a managed venv, never `/opt/rocm`, so this check always fails and Lemonade falls back to downloading its own runtime — the "two runtimes" awkwardness.

The data is already available to the Lemonade process (rocm-cli exports `ROCM_PATH` and puts `rocm-sdk` on `PATH`); Lemonade just isn't looking at it.

## Proposed fix
Resolve the ROCm root in priority order, validating each candidate by probing `lib{,64}/libamdhip64.so`:
1. `ROCM_PATH`
2. `rocm-sdk path --root` (when `rocm-sdk` is on `PATH`)
3. `/opt/rocm` (historical default)

When the user explicitly opts into a system ROCm (`ROCM_PATH`/`rocm-sdk`), relax the version gate to major.minor and accept the install when it ships no version file.

## Known follow-up (separate)
`load_hsa_runtime()` in `system_info.cpp` still falls back to a hardcoded `/opt/rocm` for the WSL2 HSA agent-enumeration path.

volen-silo:
Closes #2497

## Summary
Lemonade hardcoded `/opt/rocm` when checking for a system ROCm, so installs that live elsewhere (e.g. TheRock pip wheels via rocm-cli) were never found and Lemonade downloaded its own runtime — the "two runtimes" problem.

This resolves the ROCm root in priority order — `ROCM_PATH` → `rocm-sdk path --root` → `/opt/rocm` — validating each by probing `lib{,64}/libamdhip64.so`. When the user explicitly selects a ROCm (`ROCM_PATH`/`rocm-sdk`), the version gate is relaxed to major.minor and accepts the install when it ships no version file.

## Notes
- Linux-only in practice (gated behind the existing Linux check).
- Adds a unit test for the resolver (`RocmRootResolutionTest`).
- Follow-up (not in this PR): `load_hsa_runtime` in `system_info.cpp` still falls back to a hardcoded `/opt/rocm` for the WSL2 HSA path.

### DIFF
--- a/src/cpp/include/lemon/backends/backend_utils.h
+++ b/src/cpp/include/lemon/backends/backend_utils.h
@@ -3,6 +3,7 @@
 #include <string>
 #include <functional>
 #include <filesystem>
+#include <optional>
 #include <utility>
 #include <vector>
 #include "lemon/backends/backend_descriptor.h"
@@ -105,6 +106,31 @@ namespace lemon::backends {
 
         /** Get the latest version number for the given recipe/backend */
         static std::string get_backend_version(const std::string& recipe, const std::string& backend);
+
+        /**
+         * Resolve the ROCm install root, honoring an externally-installed ROCm
+         * before the bundled default. Resolution order, returning the first root
+         * that contains lib{,64}/libamdhip64.so:
+         *   1. ROCM_PATH environment variable
+         *   2. `rocm-sdk path --root` (when rocm-sdk is on PATH)
+         *   3. /opt/rocm
+         * Returns std::nullopt when none validate. When resolved_explicitly is
+         * non-null, it is set to true when the root came from ROCM_PATH or
+         * rocm-sdk (a user-selected ROCm) and false for the /opt/rocm fallback.
+         *
+         * Probes Linux HIP runtime artifacts (lib{,64}/libamdhip64.so, /opt/rocm)
+         * and is only meaningful on Linux; current callers reach it solely behind
+         * is_rocm_installed_system_wide() (Linux-gated).
+         */
+        static std::optional<fs::path> resolve_rocm_root(bool* resolved_explicitly = nullptr);
+
+        /**
+         * Read the ROCm version string from a resolved install root, probing the
+         * known version-file locations ({root}/.info/version,
+         * {root}/share/rocm/version, {root}/version). Returns the trimmed first
+         * line of the first file found, or "" when none exist.
+         */
+        static std::string read_rocm_version_from_root(const fs::path& root);
 
         /** Check if ROCm libraries are installed system-wide (Linux only) */
         static bool is_rocm_installed_system_wide();
--- a/src/cpp/server/backends/backend_utils.cpp
+++ b/src/cpp/server/backends/backend_utils.cpp
@@ -813,9 +813,8 @@ namespace lemon::backends {
         }
     }
     namespace {
-        // A directory qualifies as a ROCm root only if it ships the HIP runtime
-        // (lib{,64}/libamdhip64.so). Uses the non-throwing fs overloads so a
-        // bogus user-supplied path reports "not a root" instead of throwing.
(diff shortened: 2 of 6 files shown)


src/cpp/include/lemon/backends/backend_utils.h
```diff
@@ -106,6 +107,31 @@ namespace lemon::backends {
         /** Get the latest version number for the given recipe/backend */
         static std::string get_backend_version(const std::string& recipe, const std::string& backend);
 
+        /**
+         * Resolve the ROCm install root, honoring an externally-installed ROCm
+         * before the bundled default. Resolution order, returning the first root
+         * that contains lib{,64}/libamdhip64.so:
+         *   1. ROCM_PATH environment variable
+         *   2. `rocm-sdk path --root` (when rocm-sdk is on PATH)
+         *   3. /opt/rocm
```
superm1, line 116: As this is a new environment variable that we detect, I think that our documentation needs to be updated to indicate how we would use it.

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -812,48 +812,147 @@ namespace lemon::backends {
...
         }
+        return "";
+    }
 
-        // If we found the lib but no version file, log a warning but still accept it
-        LOG(DEBUG, "BackendUtils") << "Found ROCm libraries at /opt/rocm (no version file found)" << std::endl;
+    bool BackendUtils::is_rocm_installed_system_wide() {
+#ifndef __linux__
+        return false;
+#else
```
superm1, line 931: I would think you still want to use a venv in Windows too don't you?

superm1 (commented):
We have some guidance in AGENTS.md around unnecessary and extra wordy comments.  Can you please have an agent review it against your PR?  Some of these comments are quite obvious to me.

src/cpp/include/lemon/backends/backend_utils.h
volen-silo: Added user docs in `docs/guide/configuration/llamacpp.md` — a new "Reusing a System-Installed ROCm" section covering the `ROCM_PATH` / `rocm-sdk` / `/opt/rocm` resolution order and the major.minor version behavior (7b7644af).

src/cpp/server/backends/backend_utils.cpp
volen-silo: Agreed — TheRock wheels work on Windows too, so this should detect them there as well. It needs more than dropping the gate, though: `validate_rocm_root` currently probes `libamdhip64.so`, so Windows also needs the `amdhip64.dll` (bin/) probe. I'd like to land this Linux-only PR first and do Windows as a focused follow-up that I can verify on a Windows machine, rather than ship an untested Windows path here. Tracking it in #2497.

volen-silo:
Thanks for the review. Went back over the comments against AGENTS.md and trimmed the ones that just restate the code in `7b7644af` — dropped the numbered `// 1. ROCM_PATH` / `// 2. rocm-sdk` / `// 3. /opt/rocm` steps (already in the header doc + the DEBUG logs), the per-case labels in the test, and the function-summary comments, keeping only the non-obvious rationale (the `assert()`-under-NDEBUG note, the non-throwing fs probe, and the relaxed version gate).

Also addressed the two inline points: added user docs for the new `ROCM_PATH` detection, and replied re: Windows (agree it's wanted — proposing it as a follow-up I can verify on Windows, tracked in #2497).

superm1 (approved):

PR #2499, opened 2026-06-30
Router/2382 RoutingPolicyEngine — assembly + evaluate() (M8)

### LINKED ISSUE #2382: [Router] RoutingPolicyEngine — assembly + evaluate() (M8)

SlawomirNowaczyk:
Implements the routing engine declared in `routing_policy.h`: the constructor compiles each rule's `match` expression into an immutable `Condition` tree, and `route()` walks them first-match-wins, failing open to `default_model`. Closes #2382.

### What's in it
- **`RoutingPolicyEngine` ctor** (`routing_policy.cpp`) — builds the per-rule `Condition` trees once via `compile_match_expr` + the classifier/deterministic `LeafFactory`. Engine is `const` after construction, so it's safe to share across concurrent requests (all per-request state lives in a local `EvalContext`).
- **`route(ctx, want_trace)`** — first-match-wins selection; fills the per-condition `trace` only when requested. Never throws: any unexpected failure (incl. classifier exceptions) is caught and fails open to `default_model` with `default_used=true`.
- **`Decision` construction cleanup** — replaced the positional constructor with two delegating constructors: a base `Decision(default_model, want_trace, trace)` and `Decision(const Rule&, want_trace, trace)` that delegates to it. `default_used` and trace-folding are handled internally, so callers can't build a `Decision` and forget the trace.

### Tests / CI
- New `test_routing_policy_engine` (`RoutingPolicyEngineTest`): rule-match, default fail-open, first-match-wins, trace on/off, classifier-failure-falls-open, `on_error: match_true` on a throwing service, and a 16-thread × 500-iteration concurrency check.
- Registered the target in CMakeLists.txt and added it to both C++ unit-test blocks in cpp_server_build_test_release.yml.

### Notes
- Unblocks #2384 (ClassifierServices ↔ Router live wiring).
- No trust semantics in core — verdicts/actions stay layered on top via `Rule::outputs`.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -281,8 +281,9 @@ jobs:
           cmake --build --preset default --target test_routing_policy_registry
           cmake --build --preset default --target test_routing_policy_semantic
           cmake --build --preset default --target test_routing_policy_deterministic
+          cmake --build --preset default --target test_routing_policy_engine
           cd build
-          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic)Test"
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine)Test"
 
       - name: Upload .deb package
         uses: actions/upload-artifact@v7
@@ -586,8 +587,9 @@ jobs:
           cmake --build --preset default --target test_routing_policy_registry
           cmake --build --preset default --target test_routing_policy_semantic
           cmake --build --preset default --target test_routing_policy_deterministic
+          cmake --build --preset default --target test_routing_policy_engine
           cd build
-          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic)Test"
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine)Test"
 
       - name: Upload .pkg package
         if: steps.check_signing.outputs.has_signing == 'true'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1898,6 +1898,7 @@ set(_ROUTING_CONTRACT_TEST_SRC
 if(EXISTS "${_ROUTING_CONTRACT_TEST_SRC}")
     add_executable(test_routing_policy_contract
         test/cpp/test_routing_policy_contract.cpp
+        src/cpp/server/routing_policy.cpp
     )
     target_include_directories(test_routing_policy_contract PRIVATE
         ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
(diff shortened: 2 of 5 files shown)


fl0rianr (changes requested):
Thanks, this is a clean and focused engine assembly overall. The route() path itself looks good to me: it keeps per-request state local in EvalContext, applies first-match-wins, gates trace output on want_trace, and fails open to default_model if evaluation unexpectedly throws.

I have one concern before considering this ready to merge/close #2382:

The engine constructor currently compiles rules with:

```cpp
LeafFactory leaf_factory = make_leaf_factory(policy_.classifiers);
```

That means no deterministic leaf factories are registered on the engine path. make_leaf_factory can accept deterministic factories, but the default here is empty. As a result, policies containing deterministic leaves such as `keywords_any`, `regex`, `min_chars`, `has_tools`, `has_images`, or `metadata` would fail construction with `unknown leaf condition`.

This matters because the committed L1 fixtures already use deterministic routing conditions (l1_keywords.json uses keywords_any, regex, min_chars and l1_metadata.json uses metadata). So this PR currently assembles the engine for classifier-band leaves, but not yet for the full v1 policy surface expected by the fixtures / #2380 path.

I think we should either:

1. wire the deterministic leaf factories into the engine constructor before compiling rules, or
2. explicitly scope this PR as “classifier-only engine assembly” and avoid treating it as the full #2382 completion / live-router unblocker until deterministic leaves are connected.

One smaller test note: test_classifier_failure_falls_open() does not currently exercise a real classifier failure. The fake returns an empty score map, and ModelClassifier::evaluate() treats that as ok=true; the rule simply evaluates false because the requested label score is 0.0. I’d add a true failure case where run_classifier throws with default on_error=match_false, then assert that the decision falls open to default_model with default_used=true.

With those addressed, the core shape of the engine looks solid.

SlawomirNowaczyk:
> That means no deterministic leaf factories are registered on the engine path.

True, since https://github.com/lemonade-sdk/lemonade/pull/2491 is not yet merged. I suggest we wait until it's merged, and then update #2499.

ramkrishna2910:
The engine assembly here looks right to me — immutable per-rule `Condition` trees compiled once in the ctor, `route()` doing pure first-match-wins tree-walking on `const` state, fail-open to `default_model`, trace gated on `want_trace`, and the 16×500 concurrency test covering the shared-engine invariant. The `Decision` delegating-constructor cleanup is a nice touch.

On the one open concern (deterministic leaves not wired): let's resolve it by **merging #2380 first, then rebasing this PR onto it**, rather than scoping this down to classifier-only. Reasoning:

- The committed L1 fixtures (`l1_keywords.json`, `l1_metadata.json`) already use `keywords_any` / `regex` / `min_chars` / `metadata`. An engine that can't construct from its own fixtures would be a broken state to land on `main`, so option 2 (classifier-only) isn't really viable.
- #2380 (#2491) is the piece that provides `make_deterministic_leaf_factories()`. It's in review-fix and close. Once it merges, the wiring here is a one-liner:

  ```cpp
  LeafFactory leaf_factory =
      make_leaf_factory(policy_.classifiers, make_deterministic_leaf_factories());
  ```

  plus a deterministic-rule engine test — ideally driving `l1_keywords` / `l1_metadata` end-to-end through `route()`, which also closes the fixture-vs-engine gap the reviewer noted. That makes this the genuine #2382 completion and the demo join point (deterministic rules actually routing).

This is the planned dependency (`[#2380 ‖ #2381] → #2382`); #2381 is already in, so #2380 is the last input.

Heads-up for the rebase, since #2491 and this PR touch the same files:
- `CMakeLists.txt` / the CI workflow both add a routing test target + ctest-filter entry — union them (keep both `Semantic`/`Deterministic`/`Engine`).
- #2491 adds a 6th field to `EvalContext` (`lowered_input`, an ASCII-lowered-input memo). The `EvalContext eval{ctx, services_, want_trace, {}, {}}` positional aggregate init still compiles correctly (the new field value-inits to `nullopt`), but please double-check that line after rebase — a trailing `{}` or designated init keeps it unambiguous.

I'll ping you the moment #2491 lands so you can do the one-line rebase. Thanks for the clean engine — once deterministic leaves are wired this is exactly the #2382 we planned.

SlawomirNowaczyk:
All requested changes have now been added, and rebased on top of new main

fl0rianr (commented):
Thanks for the update, the points are well adressed.

One small remaining test nit: test_classifier_failure_falls_open() still does not exercise a true classifier failure; it uses an empty score map, which is treated as ok=true and simply evaluates the band false. Since there is already a true throwing-service test for on_error: match_true, it would be nice to add the symmetric case for default on_error=match_false: make run_classifier throw and assert that the route falls open to default_model with default_used=true.

SlawomirNowaczyk:
> One small remaining test nit: test_classifier_failure_falls_open() still does not exercise a true classifier failure;

Thanks, I've missed that -- added now.

fl0rianr (approved):
Thanks, LGTM from my side.

eddierichter-amd (approved):
LGTM

PR #2502, opened 2026-06-30
Keep sd-cpp validation images out of auto-update PRs

fl0rianr:
This follow-up fixes the validation image handling introduced in #2369.

The sd-cpp auto-update PR body still renders the per-image validation evidence inline, but the generated PNGs are no longer committed into the update PR branch.

Instead:
- validation artifacts are normalized as before
- the PNG evidence is pushed to a dedicated evidence branch
- the generated PR body references that evidence branch via raw.githubusercontent.com
- the update PR commits only src/cpp/resources/backend_versions.json

This keeps the review diff small while preserving the same visual validation evidence currently shown (as it was before being removed) in #2477.


kenvandine:
@claude review this

fl0rianr:
@claude review again

fl0rianr:
@kenvandine this should be ready to go now... Thanks.

kenvandine (approved):
Thanks!

PR #2503, opened 2026-06-30
Document ARM64 archive names and normalize Linux names to x86_64

### LINKED ISSUE #2501: Fix embeddable lemonade documentation post-arm-support
Looking at https://github.com/lemonade-sdk/lemonade/pull/2430 it has a breaking change (renamed the embedded artifacts) that isn't noted in the PR description nor updated on https://github.com/lemonade-sdk/lemonade/tree/main/docs/embeddable

Can you either revert the breaking change or update the PR description and docs for accuracy?

fl0rianr:
Fixes #2501.

Updates embeddable naming after #2430:

- documents GitHub Actions artifact names separately from archive filenames
- adds Linux ARM64 embeddable documentation
- renames the Linux x86_64 embeddable release archive from `ubuntu-x64` to `ubuntu-x86_64`
- updates CMake, release workflow verification/upload paths, generated release notes, and docs

### DIFF
--- a/docs/embeddable/README.md
+++ b/docs/embeddable/README.md
@@ -26,7 +26,7 @@ Embeddable Lemonade is an zip/tarball artifact shipped in Lemonade releases.
 | Platform | GitHub Actions artifact | Archive |
 |----------|--------------------------|---------|
 | Windows x64 | `lemonade-embeddable-windows` | `lemonade-embeddable-{VERSION}-windows-x64.zip` |
-| Linux x86_64 | `lemonade-embeddable-linux-x86_64` | `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz` |
+| Linux x86_64 | `lemonade-embeddable-linux-x86_64` | `lemonade-embeddable-{VERSION}-ubuntu-x86_64.tar.gz` |
 | Linux ARM64 | `lemonade-embeddable-linux-arm64` | `lemonade-embeddable-{VERSION}-ubuntu-arm64.tar.gz` |
 | macOS ARM64 | `lemonade-embeddable-macos` | `lemonade-embeddable-{VERSION}-macos-arm64.tar.gz` |
 
--- a/docs/embeddable/README.md
+++ b/docs/embeddable/README.md
@@ -23,9 +23,10 @@ Use Embeddable Lemonade instead of a global Lemonade Service when you want a coh
 
 Embeddable Lemonade is an zip/tarball artifact shipped in Lemonade releases.
 
-- Windows: `lemonade-embeddable-10.1.0-windows-x64.zip`
-- Ubuntu: `lemonade-embeddable-10.1.0-ubuntu-x64.tar.gz`
-- macOS: `lemonade-embeddable-10.1.0-macos-arm64.tar.gz`
+- Windows x64: `lemonade-embeddable-{VERSION}-windows-x64.zip`
+- Linux x86_64: `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz`
+- Linux ARM64: `lemonade-embeddable-{VERSION}-ubuntu-arm64.tar.gz`
+- macOS ARM64: `lemonade-embeddable-{VERSION}-macos-arm64.tar.gz`
 
 > Note: see the [Building from Source](./building.md) for instructions for building your own embeddable Lemonade from source, including for other Linux distros.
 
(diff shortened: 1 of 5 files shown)


kenvandine (approved):
This looks good, thanks!

docs/embeddable/building.md
```diff
@@ -75,9 +75,10 @@ The `embeddable` target produces a single archive in `build/`:
 
 | Platform | Archive |
 |----------|---------|
-| Linux    | `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz` |
-| Windows  | `lemonade-embeddable-{VERSION}-windows-x64.zip` |
-| macOS    | `lemonade-embeddable-{VERSION}-macos-arm64.tar.gz` |
+| Linux x86_64 | `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz` |
+| Linux ARM64 | `lemonade-embeddable-{VERSION}-ubuntu-arm64.tar.gz` |
+| Windows x64 | `lemonade-embeddable-{VERSION}-windows-x64.zip` |
```
jeremyfowers: Hang on, I thought the x86 ones changed too: `artifact-suffix: x86_64`

kenvandine: Oh, you are right! I looked at that and skimmed right over it because x86_64 is the norm :) @fl0rianr please update it to reference x86_64 too.

fl0rianr: Good catch — the archive filenames were documented, but the renamed GitHub Actions artifact names were missing. I updated the docs to distinguish the workflow artifact name from the archive filename inside it, including `lemonade-embeddable-linux-x86_64` for the x86_64 Linux artifact.

docs/embeddable/README.md
```diff
@@ -23,9 +23,12 @@ Use Embeddable Lemonade instead of a global Lemonade Service when you want a coh
 
 Embeddable Lemonade is an zip/tarball artifact shipped in Lemonade releases.
 
-- Windows: `lemonade-embeddable-10.1.0-windows-x64.zip`
-- Ubuntu: `lemonade-embeddable-10.1.0-ubuntu-x64.tar.gz`
-- macOS: `lemonade-embeddable-10.1.0-macos-arm64.tar.gz`
+| Platform | GitHub Actions artifact | Archive |
+|----------|--------------------------|---------|
+| Windows x64 | `lemonade-embeddable-windows` | `lemonade-embeddable-{VERSION}-windows-x64.zip` |
+| Linux x86_64 | `lemonade-embeddable-linux-x86_64` | `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz` |
```
kenvandine: I think this should also be x86_64

docs/embeddable/building.md
```diff
@@ -73,11 +73,12 @@ If you want the embeddable build to include the browser UI assets under `resourc
...
 The `embeddable` target produces a single archive in `build/`:
 
-| Platform | Archive |
-|----------|---------|
-| Linux    | `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz` |
-| Windows  | `lemonade-embeddable-{VERSION}-windows-x64.zip` |
-| macOS    | `lemonade-embeddable-{VERSION}-macos-arm64.tar.gz` |
+| Platform | GitHub Actions artifact | Archive |
+|----------|--------------------------|---------|
+| Linux x86_64 | `lemonade-embeddable-linux-x86_64` | `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz` |
```
kenvandine: I think this should also be x86_64

kenvandine (changes requested):

docs/embeddable/README.md
fl0rianr: I think there are two different names here:

- GitHub Actions artifact: `lemonade-embeddable-linux-x86_64`
- archive filename inside it / release asset: `lemonade-embeddable-{VERSION}-ubuntu-x64.tar.gz`

The workflow still generates and verifies the `ubuntu-x64` archive filename, so I kept that in the Archive column. If we want the archive filename itself to become `ubuntu-x86_64`, that would need a CMake/workflow change too, not just a docs update. Same below

kenvandine: Then I'm torn... It would be more correct to make the name match everywhere, but the cmake change is a bigger lift. What do you think?

fl0rianr: I can make the cmake change as well we need it to be ready for release, cherry pick is also an option. Easier to cherry pick on squash commit.

fl0rianr: Made the bigger lift in this PR.

jeremyfowers (changes requested):
I (and maybe all of us) got pretty confused because the job output (a zip folder produced mid-action) changed name from ubuntu-x64 to ubuntu-x86-64. However, the actual **release artifact** never changed name. This PR introduces a breaking change to normalize the release artifact name to the job output name, which is not worth the cost of a breaking change.

Please consider https://github.com/lemonade-sdk/lemonade/pull/2505 as an alternative that documents the new artifacts without the breaking change.

kenvandine:
@fl0rianr let's land #2505 for v10.9.0, then rebase this and aim to get the artifact renames in for the next release.

fl0rianr:
Thanks, makes sense, let's do it that way.
Not urgent anymore.

Design Question: Ok to expand the table into the readme and should we actually use the code brackets inside the table, seems strange to me.

PR #2504, opened 2026-06-30
docs(release): add step to reconcile release tag into main

jeremyfowers:
## Why

Since v10.8.0, releases are tagged on `release-v*` branches, so the tagged commits never become reachable from `main`. As a result `git describe --tags` on `main` reports the *previous* release (currently `v10.7.0` instead of `v10.8.1`), which feeds a stale version into `prepare-debian-build` (Debian/PPA) and `test_release_notes.yml`.

## What

Adds **Step 6: Reconcile the Tag into `main`** to the release process: after forward-ports are complete, link the tag into `main` with an `-s ours` merge (records ancestry only, no file changes). Includes the classic-branch-protection bypass requirements. Following steps renumbered 6→7, 7→8, 8→9.

The current-state backfill for v10.8.0/v10.8.1 has already been pushed to `main`; this PR only documents the going-forward step.

### DIFF
--- a/docs/dev/release.md
+++ b/docs/dev/release.md
@@ -132,7 +132,25 @@ git tag -f vX.Y.Z
 git push origin vX.Y.Z --force
 ```
 
-## Step 6: Update the Release Notes
+## Step 6: Reconcile the Tag into `main`
+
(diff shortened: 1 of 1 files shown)


kenvandine (approved):
LGTM, great to have tags on main again

PR #2508, opened 2026-07-01
feat(rocm): detect externally-installed ROCm on Windows

volen-silo:
Extends system-ROCm reuse (previously Linux-only, #2498) to Windows: probes `bin\amdhip64.dll`/`lib\amdhip64.dll`, uses `rocm-sdk.exe` for the PATH presence check, and treats `HIP_PATH` (set by the AMD HIP SDK installer) as the platform default. Docs and unit test updated to match.

### DIFF
--- a/docs/guide/configuration/llamacpp.md
+++ b/docs/guide/configuration/llamacpp.md
@@ -129,7 +129,7 @@ On the stable channel Lemonade normally downloads its own ROCm runtime (TheRock)
 `ROCM_PATH` and `rocm-sdk` work on both platforms. When the runtime is found via one of them, a `major.minor` version match is accepted (and a runtime with no version file is accepted as-is), so a patch-level difference won't trigger a second download. To force Lemonade to use a specific ROCm, set `ROCM_PATH` before starting the server:
 
 ```bash
-# Linux / macOS
+# Linux
 export ROCM_PATH=/path/to/rocm
 ```
 
--- a/docs/guide/configuration/llamacpp.md
+++ b/docs/guide/configuration/llamacpp.md
@@ -118,18 +118,24 @@ After changing channels, you'll need to reinstall the ROCm backend:
 lemonade backends install llamacpp:rocm
 ```
 
-### Reusing a System-Installed ROCm (Linux)
+### Reusing a System-Installed ROCm (Windows and Linux)
 
-On the stable channel Lemonade normally downloads its own ROCm runtime (TheRock). If you already have ROCm installed system-wide, Lemonade reuses it instead of downloading a second copy when it can find a matching version. It locates the install root in this order, using the first directory that contains `lib{,64}/libamdhip64.so`:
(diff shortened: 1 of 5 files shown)


docs/guide/configuration/llamacpp.md
```diff
@@ -118,20 +118,26 @@ After changing channels, you'll need to reinstall the ROCm backend:
...
 1. The `ROCM_PATH` environment variable
 2. `rocm-sdk path --root`, when `rocm-sdk` is on your `PATH` (e.g. a ROCm installed from the TheRock pip wheels)
-3. `/opt/rocm`
+3. The platform default: `HIP_PATH` (set by the AMD HIP SDK installer) on Windows, `/opt/rocm` on Linux
 
-When the runtime is found via `ROCM_PATH` or `rocm-sdk`, a `major.minor` version match is accepted (and a runtime with no version file is accepted as-is), so a patch-level difference won't trigger a second download. To force Lemonade to use a specific ROCm, export `ROCM_PATH` before starting the server:
+`ROCM_PATH` and `rocm-sdk` work on both platforms. When the runtime is found via one of them, a `major.minor` version match is accepted (and a runtime with no version file is accepted as-is), so a patch-level difference won't trigger a second download. To force Lemonade to use a specific ROCm, set `ROCM_PATH` before starting the server:
 
 ```bash
+# Linux / macOS
```
superm1: Surely Linux only

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -926,18 +954,19 @@ namespace lemon::backends {
     }
 
     bool BackendUtils::is_rocm_installed_system_wide() {
-#ifndef __linux__
+#if !defined(__linux__) && !defined(_WIN32)
```
superm1: Can we just drop this line entirely?

superm1 (approved):

superm1:
(Holding off to merge until a fix for a potential issue is landed)

volen-silo:
Pushed `30766fd7`: hardening for the `rocm-sdk` detection path.

**Problem:** `query_rocm_sdk_root` used only the first line of `rocm-sdk path --root`, but the subprocess layer merges the child's stderr into stdout. Since `rocm-sdk` is a Python console script that can print warnings to stderr, a warning could land first, get treated as the path, fail validation, and silently fall through to the platform default, so the branch never fired.

**Fix:** collect all output lines and pick the first that trims to an absolute path *and* validates, via a pure `pick_rocm_root_candidates` helper (unit-tested: leading stderr warning, blank lines, no valid line, CRLF).

**Also:** resolve the root once in `will_install_therock` and thread it into the version check (was spawning `rocm-sdk` twice), drop the now-unused `is_rocm_installed_system_wide`, and add a direct `#include <optional>`.

Note: this mainly exercises the shared `rocm-sdk` (pip-wheel) path, which is most common on Linux.

PR #2515, opened 2026-07-01
Add support for inhibiting suspend while inference is running

superm1:
The idea here is that if a user is using lemonade remotely or they are running a local task that is taking a long time and step out for coffee they probably don't want system automatic suspend to cause problems for them.


fl0rianr (commented):
Thanks for the PR — the overall design looks good to me. Putting the inhibitor around execute_inference() / execute_streaming() seems like the right level, and the FD lifetime / RAII approach looks appropriate.

A couple of follow-up thoughts, mostly around robustness and maintainability:

1. The current refcounting test mostly validates a mock implementation with duplicated logic rather than the production refcounting path. That may be fine for avoiding a real logind dependency in tests, but would you be open to factoring the refcounting into a small testable helper or injectable backend so the test exercises the same logic used by the Linux implementation?

2. Since inhibit_suspend defaults to true, Linux builds with HAVE_SYSTEMD but without a usable logind/system bus — for example some containers, WSL-like setups, or minimal/headless environments — may warn on ordinary inference requests. It might be worth degrading to no-op after detecting logind is unavailable, or making expected “environment unavailable” failures debug-level rather than warning-level.

Minor docs nit: could we document that this currently only has an effect on Linux/systemd-logind and is a no-op on Windows/macOS?

Not blocking from my side, but I think those would make the feature safer to ship broadly.

fl0rianr (commented):
Nice, thanks for the updates. The refcounting change and the one-shot logind fallback address my main concerns from the earlier review.

Only tiny remaining nit from my side: it might still be worth documenting that this currently only has an effect on Linux/systemd-logind and is a no-op on Windows/macOS. Non-blocking.

superm1:
FYI - confirmed functionality on a real system as well.

image

fl0rianr (approved):
Great - and thanks for the updates — the refcounting change and the one-shot logind fallback address my main concerns.

The remaining docs clarification is minor and non-blocking from my side. Approved.

PR #2516, opened 2026-07-01
Fix disabling systemd user service

superm1:
Since commit 4e8ffd4cd didn't make it for 10.8.1 the version is wrong and it doesn't get applied to upgrades.

Adjust to the hypothetical next version of 10.10.0.


superm1:
@abn can you please review?

abn:
@superm1 your change looks good to me. I am not too familiar with debian's preferences on such fixes, but technically this is correct.

Note: Any user who relied on this being "a feature" should be told to enable it manually now - `systemctl --user enable lemond`.

kenvandine (approved):
Looks good

PR #2523, opened 2026-07-01
fix(tray): use /live instead of /api/v1/health for server readiness check

### LINKED ISSUE #2520: Tray fails to start when LEMONADE_API_KEY is set
**Platform:** Windows

**Lemonade Version:** 10.9.0

**Component:** lemond (server)

**Bug Description:**
When `LEMONADE_API_KEY` environment variable is set, `LemonadeServer.exe` fails to start its tray � after 15 seconds a message box appears: "Lemonade Server failed to start within 15 seconds." The server itself is running and responding to requests. The issue is that `wait_for_server()` in the tray code calls `/api/v1/health`, which requires authentication when the API key is set. The tray doesn't pass any auth header, gets a 401, and times out.

**Steps to Reproduce:**
1. Set `LEMONADE_API_KEY` (e.g. `sk-test-key`)
2. Launch `LemonadeServer.exe` from Start Menu or the installed MSI
3. Wait 15 seconds � the error dialog appears

**Expected vs Actual Behavior:**
Expected: tray icon appears and server works normally.
Actual: error "Lemonade Server failed to start within 15 seconds."

**Log Output:**
Server logs show it's running. `/api/v1/health` returns 401 when called without `Authorization` header. The `/live` liveness endpoint does not require authentication and should be used instead.

blackdeathdrow:
The tray's `wait_for_server()` calls `/api/v1/health` to verify the server is running. When `LEMONADE_API_KEY` is set, this endpoint requires authentication, but the tray doesn't carry the API key — it gets a 401 and the check times out after 15 seconds, showing a spurious error dialog.

Switch to `/live` (the unauthenticated liveness endpoint) which was specifically added for this purpose (see #835).

Closes #2520.


fl0rianr (changes requested):
I think this fixes the 401 symptom but weakens the startup readiness contract.

The previous code intentionally used /api/v1/health instead of /live because `/live` can return before the model cache / model endpoints are ready. Switching wait_for_server() to /live makes the tray proceed as soon as the listener is alive, but TrayUI::build_menu() immediately calls /api/v1/health, /api/v1/models, and /internal/config, so this can reintroduce the early /models failure path that the old comment explicitly warned about.

The root issue seems narrower: wait_for_server()d oes not attach the same Authorization header that TrayUI::http_get() already attaches from LEMONADE_ADMIN_API_KEY / LEMONADE_API_KEY.

Suggested fix:
- keep /api/v1/health as the readiness probe;
- add the Bearer token header in wait_for_server();
- ideally share the auth-header construction with TrayUI to avoid future drift;
- add a regression test or at least a manual test note for LEMONADE_API_KEY startup.

`/live` is correct for Docker/Kubernetes liveness, but it looks too weak for tray readiness.

fl0rianr (commented):
Thanks, this is now the right shape: it preserves the existing /api/v1/health readiness path and fixes only the missing auth header.

One tiny compile-hygiene nit before I’d approve: main.cpp now uses std::getenv, so please add `#include <cstdlib>` explicitly instead of relying on transitive includes, this would be cleaner.

blackdeathdrow:
Added `#include <cstdlib>`.

fl0rianr (approved):
LGTM now.

PR #2527, opened 2026-07-02
feat: vLLM-Omni backend (native omni: text/audio/vision) + chat UI

ramkrishna2910:
## What
Adds **vLLM-Omni** as a first-class backend so omni / any-to-any multimodal models (Qwen2.5-Omni today; Cosmos3 later) run with ROCm acceleration on AMD gfx1151 (Strix Halo) — a single model doing **text + audio + vision in, text + native voice out**.

vLLM-Omni is a pure-Python layer on the same base vLLM+PyTorch+Triton, shipped as a **separate `vllm-omni*` release artifact**, so it gets its own recipe + pin (see vllm-rocm#23).

## Backend (`54788cc9`) — uses the new self-describing descriptor model
Per `docs/dev/adding-a-backend.md` — one folder + a few appends, no router/CLI/doc edits:
- `src/cpp/{include,server}/backends/vllm_omni/` — descriptor + `VLLMOmniServer` (mirrors `vllm`). Launches `vllm-omni-server serve <model> --omni --deploy-config <yaml> --served-model-name --max-model-len`, resolves the per-model deploy config via the model's `deploy_config` extra key + `get_resource_path`, forwards chat. Native voice/vision ride through the chat body.
- `CMakeLists.txt` `LEMON_BACKENDS` += `"vllm-omni|vllm_omni"`
- `backend_versions.json` pin `vllm-omni0.23.0rc1-rocm7.14.0`
- `server_models.json` `Qwen2.5-Omni-3B-vLLM-Omni`
- `resources/omni_deploy/qwen2_5_omni_1gpu.yaml` — single-GPU stage colocation (upstream defaults are multi-GPU)

## Chat UI (`a423eab3`)
- `ChatWindow` derives `isAudioOutput` from a `chat-speech` label (mirrors `chat-transcription`).
- `LLMChatPanel`: voice **toggle + picker** (Chelsie/Ethan, **off by default** — speech gen is slow); `handleAudioChat` sends a non-streaming `modalities:[text,audio]`+`voice` request and captures `message.audio` (handles vLLM-Omni's two-choice shape *and* OpenAI's single-choice) → rendered via the existing `MessageAudio` player. No new rendering code.
- Model labels `vision/chat-transcription/chat-speech/omni` light up vision input, audio input, and native voice output via existing capability flags.

## Validation
- **Backend: end-to-end on real gfx1151.** `lemond` builds, registers the recipe, lists the model, loads it (launches the subprocess), and **all four modalities pass**: text, native voice out (24 kHz WAV), audio in, vision in — with the audio choice passed through lemonade's handler intact.
- **UI: `tsc --noEmit` clean.** Reuses proven audio-player/artifact path. **Not yet run in the app** (needs a webview) — reviewers should eyeball rendering/UX.

## Notes for reviewers
- UI uses **emoji icons** (🔊/🔈) + one inline style on the voice-control wrapper — swap for house icon components / CSS.
- Omni over-generates audio length today (talker `max_tokens` tuning in the deploy YAML) — cosmetic.
- gfx1151-only (the qualified/hardware-validated omni target).


fl0rianr (changes requested):
Thanks for the PR — the direction looks promising, and the backend structure is nicely aligned with the new descriptor model. I do think we should address a few things before merging:

1. **Generated backend docs/resources are stale**
   Docs And Style / backend-docs-drift is currently failing. Please run:

   ```bash
   python docs/tools/gen_backend_boilerplate.py
   ```

   and commit the generated updates. That's the really nice backend change benefit we have now.

2. **Capability mismatch in the vLLM-Omni test config**
   `vllm-omni` marks `completions: False`, but `stop_parameter: True`. The existing LLM test suite uses stop_parameter for both chat and completions tests, so this will likely run a completions test against a backend that declares completions unsupported.
   I think we either need to set `stop_parameter: False` for now, or split this into more precise capabilities like chat stop vs. completions stop.

3. **Assistant audio is currently persisted back into chat history**
   handleAudioChat() converts returned message.audio into an audio artifact, which becomes `input_audio` content in the assistant message. Since buildChatRequestBody() forwards the message history unchanged, the next turn will send the generated assistant audio Base64 back to /chat/completions. That is expensive and may break strict backends.
   We should sanitize history before sending, or store generated assistant audio as UI-only artifact data rather than OpenAI `input_audio`.

One more thing to consider: the new backend forwards chat requests directly, while the existing vLLM backend runs request preparation first, including the max_completion_tokens → max_tokens compatibility alias. Since the LLM tests use max_completion_tokens, it would be safer to apply the same preparation logic here unless vLLM-Omni is guaranteed to support it.

After those fixes and a clean CI run, this should be much easier to merge.

ramkrishna2910:
Thanks @fl0rianr — all four addressed in `7411f70a`:

1. **Stale generated docs** — ran `python docs/tools/gen_backend_boilerplate.py` and committed the output (`defaults.json`, `backends-reference.md`, `README.md`, `cli.md`, the configuration docs, and `models.js`). `--check` is now clean locally.
2. **Capability mismatch** — set `stop_parameter: False` for `vllm-omni` (completions is unsupported, and the suite drives stop for both chat and completions), avoiding a completions-stop test against an unsupported endpoint.
3. **Assistant audio replayed into history** — added `stripGeneratedAudio` so generated `input_audio` on assistant turns is dropped from outgoing history (kept as UI-only playback); it no longer re-sends the Base64 each turn.
4. **Request preparation** — `chat_completion`/`completion` now apply the `max_completion_tokens → max_tokens` alias (`JsonUtils::with_legacy_max_tokens_alias`), matching the plain vLLM backend, so the LLM tests' `max_completion_tokens` works.

Verified locally: `lemond` builds, `tsc --noEmit` clean, `gen_backend_boilerplate.py --check` passes. The `stx-halo` CI row still needs the published `vllm-omni*` release to go green.

fl0rianr (changes requested):
Thanks for the update — the original issues look mostly addressed now.
I still see two blockers before merge:

1. **The pinned vLLM-Omni artifact currently 404s**

```text
https://github.com/lemonade-sdk/vllm-rocm/releases/download/vllm-omni0.23.0rc1-rocm7.14.0-gfx1151/vllm-omni0.23.0rc1-rocm7.14.0-gfx1151-x64.tar.gz
```

Since backend_versions.json pins vllm-omni0.23.0rc1-rocm7.14.0, and VLLMOmniServer::get_install_params() derives the release tag/asset as `{version}-{target_arch}` and `{release_tag}-x64.tar.gz`, this matching `gfx1151` release asset needs to exist before the new llm-vllm-omni CI row can pass. Please either publish the matching asset or update the pin/naming logic to match the artifact that actually exists.

2. **The generated configuration README appears inconsistent with the actual defaults**

The docs example now shows older/global values like config_version: 1, ctx_size: 4096, global_timeout: 300, and rocm_channel: preview, while the real src/cpp/resources/defaults.json still has config_version: 2, ctx_size: -1, global_timeout: 600, and rocm_channel: stable.

Could you regenerate/fix the configuration docs so the example matches the real defaults and only adds the new `vllm-omni` section?

Small non-blocking nit: the voice controls still use emoji + inline style. Since this is already called out in the PR notes, it is not a merge blocker for me, but ideally we should move that to regular CSS/icons before final merge if possible.

Once the artifact exists, the generated docs are consistent, and the new CI run is green, this should be good to merge from my side.

fl0rianr:
I still think you need to address the backend drift error.

ramkrishna2910:
@fl0rianr — thanks for the detailed pass. Both blockers from your last review are now addressed:

**1. Pinned vLLM-Omni artifact (was 404)** — the release is now published, so the pin resolves to a real asset:
`vllm-omni0.23.0rc1-rocm7.14.0` + `-gfx1151` → **[vllm-omni0.23.0rc1-rocm7.14.0-gfx1151](https://github.com/lemonade-sdk/vllm-rocm/releases/tag/vllm-omni0.23.0rc1-rocm7.14.0-gfx1151)**.

One heads-up on the exact URL you pasted: the bundle is ~2.9 GB, so it ships as **split assets** (`…-x64.part01-of-02.tar.gz`, `…-x64.part02-of-02.tar.gz`, `…-x64.partcount`) — the single-file `…-x64.tar.gz` intentionally 404s. `get_install_params()` returns that base name and `backend_utils.cpp` probes `{base}.partcount`, then downloads and concatenates the parts (same split path the plain-vLLM backend already uses). I ran a **live end-to-end pull on gfx1151**: it downloaded both parts, assembled/extracted the 9.2 GB bundle, and imported `vllm_omni 0.23.0rc1` on the pinned base — so the `llm-vllm-omni` row now has a real artifact to install.

**2. Config README inconsistent with defaults** — good catch. The example was rendered from a fresh lemond's `/internal/config`, which still carries **pre-migration** globals (`config_version 1`, `ctx_size 4096`, `global_timeout 300`, `rocm_channel preview`). I changed the generator to render it from `/internal/config/defaults` (the same source as `resources/defaults.json`), so the example now matches the real defaults (`2 / -1 / 600 / stable`) and only adds the `vllm-omni` section. `--check` is clean.

> Note: the underlying `/internal/config` vs `defaults.json` global mismatch is a pre-existing lemond quirk (the runtime `Config` struct initializes to the old v1 values; the v1→v2 migration only runs when loading an existing file) — unrelated to this PR, happy to file it separately.

**Nit (emoji + inline-style voice controls)** — still noted; left as-is here since it's UI-owned, can follow up before final merge.

Also rebased onto latest `main` (routing-decision + docker + HIP-detection PRs) and regenerated the backend boilerplate, so `backend-docs-drift` is green. CI is re-running now; the one prior red outside these was `Test .exe - flm-whisper`, a Windows/NPU self-hosted-runner flake unrelated to this Linux/ROCm change (`main` is independently red on it). I'll confirm the run is green.

fl0rianr (changes requested):
Thanks for the update — the vLLM-Omni implementation itself looks substantially better now, and the earlier backend issues appear addressed.

I still see one blocker before merge:

**Unrelated Trust Router files are included in this PR**

* docs/dev/trust-router-design.md
* Trust Router Design Doc.html

These files are unrelated to the vLLM-Omni backend. Please remove them from this branch as I assume this was a mistake.

One smaller correctness issue should also be addressed: `stripGeneratedAudio()` can leave an assistant message with `content: []` when the response contains audio but no text. Some strict OpenAI-compatible backends may reject an empty content array on the next turn. Please normalize that case to an empty string or omit the empty assistant turn.

The new CI row currently validates the regular chat path, but not the actual omni-specific functionality such as audio input, vision input, and native audio output. I would not block the initial merge on full coverage, but these should be tracked as a concrete follow-up.

Once the unrelated files are removed, the audio-only history edge case is fixed, and the current CI run is green, this should be ready from my side.

ramkrishna2910:
Quick CI follow-up @fl0rianr — the run is green on all omni-relevant rows:

- **`Test .deb - llm-vllm-omni`** ✅ — the gfx1151 GPU box installed the published release and ran inference end-to-end (this is blocker 1 proven in CI, not just my local pull).
- **`backend-docs-drift`** ✅ — config example now matches `defaults.json` (blocker 2).
- `Test .exe - omni` ✅, `Validate vllm (rocm)` ✅. `Test .exe - flm-whisper` also passed on re-run (the earlier red was a self-hosted Windows/NPU flake).

Tally: 87 pass / 0 fail; the ~6 still-pending are unrelated backlogged rows (trellis / SD / TTS deb-inference and cpu/rocm-stable/vulkan validation). So all three of your conditions — artifact exists, generated docs consistent, CI green — are met. Should be good for another look when you have a moment.

ramkrishna2910:
Thanks @fl0rianr — all three addressed:

**1. Unrelated Trust Router files — removed** (`6912b58f`). They'd been swept into an earlier `main`-merge commit by accident; they're now untracked and gone from the PR diff (`git diff main...HEAD` no longer shows them). Also added them to a local git exclude so they can't get re-added.

**2. `stripGeneratedAudio()` audio-only edge case — fixed** (`1d75c3f8`). You're right that an audio-only assistant turn was a trap: after dropping `input_audio`, `kept` is `[]`, and the code only returned `''` by virtue of a vacuous `[].every()` — reading as if it could emit `content: []`. Now it's explicit:
```ts
if (kept.length === 0) return { role, content: '' };
```
so the empty case is guaranteed to collapse to an empty string, never an empty array.

**3. Omni-specific CI coverage — tracked** in #2653 (audio-in, vision-in, native audio-out). All four modalities were validated manually on gfx1151 during this PR; the issue tracks adding automated assertions so regressions are caught. Agreed it shouldn't block the initial merge.

New CI run is going now on the two commits; I'll confirm it stays green.

fl0rianr (changes requested):
Thanks for the update. The earlier blockers look addressed, but I still see one merge blocker and two smaller issues.

The main blocker is the streaming request path. The max_completion_tokens → max_tokens compatibility handling is currently only applied in chat_completion() and completion(), so it only covers non-streaming requests.

vLLM-Omni does not override forward_streaming_request(), unlike the regular vLLM backend. This means streaming requests are forwarded unchanged, even though the backend is marked as supporting both streaming and async streaming, and both tests send max_completion_tokens.

Please apply the same request normalization to the streaming path as well. Ideally, the shared logic should be extracted rather than duplicated again.

There is also a UI issue with collection models. isAudioOutput becomes true when a collection contains a chat-speech component, so the voice control is shown. However, collectionMode is checked first when sending, which means handleCollectionChat() always runs and the audio-output branch is never reached. The toggle therefore appears functional but does not change the request for collections.

Either support native voice output in the collection flow or hide the control while collectionMode is active.

Finally, vllm_omni_args is split using plain whitespace. This breaks quoted arguments, escaped values, JSON values, and anything else containing spaces. For example:

--some-option "value with spaces"

Please reuse the existing custom-argument parsing utilities, or a shared version of the vLLM argument resolver, and add a small test covering quoted and escaped arguments.

The backend structure itself looks good, and the earlier audio-only history issue is fixed correctly. The collection toggle and argument parsing should ideally also be handled before merge.

ramkrishna2910:
Thanks @fl0rianr — all three addressed:

**1. Streaming path missing `max_completion_tokens` → `max_tokens`** (blocker) — fixed in `d5cb33fe`. Extracted the normalization into a shared `prepare_request()` and added a `forward_streaming_request()` override that applies it to the chat/completions streaming body before delegating to the base forwarder — so streaming and non-streaming now go through the exact same normalization (no duplication):
```cpp
json VLLMOmniServer::prepare_request(const json& request) {
    return JsonUtils::with_legacy_max_tokens_alias(request);
}
void VLLMOmniServer::forward_streaming_request(...) {
    std::string body = request_body;
    if (endpoint == "/v1/chat/completions" || endpoint == "/v1/completions") {
        try { body = prepare_request(json::parse(request_body)).dump(); } catch (...) {}
    }
    WrappedServer::forward_streaming_request(endpoint, body, sink, sse, timeout_seconds, telemetry_callback);
}
```
The `Test .deb - llm-vllm-omni` row exercises the streaming tests (both send `max_completion_tokens`) on the gfx1151 box.

**2. Collection voice-control no-op** — fixed in `3e74d138`. Since the send path routes collections through `handleCollectionChat()` first, the audio branch is unreachable there, so I hid the control while `collectionMode` is active (`isAudioOutput && !collectionMode`). Native voice in the collection flow can be a separate enhancement if wanted.

**3. `vllm_omni_args` whitespace split** — fixed in `d5cb33fe`. Replaced the naive splitter with the shared quote-aware `parse_custom_args()` (same utility the other backends use), so `--opt "value with spaces"` survives as one argument. Added `test_custom_args` (ctest `custom_args`) covering quoted, single-quoted, single-quoted JSON, empty, and unquoted-backslash cases.

Also merged latest `main` (22 commits) and regenerated the backend boilerplate; `--check` and the two C++ unit tests pass locally. New CI run is going.

fl0rianr (changes requested):
Thanks for the update — the previously reported issues are now addressed correctly. 

But I found one remaining merge blocker in the custom-argument path:

`vllm_omni_args` is parsed and appended after Lemonade’s managed launch arguments without validating reserved process-shaping flags. This currently allows duplicate values such as:

* --port
* --host
* --served-model-name
* --max-model-len
* --deploy-config

Depending on the backend parser, these trailing values can override Lemonade’s values or otherwise make the launch ambiguous. For example, overriding `--port` would start the backend on a different port while Lemonade continues polling its internally selected port, and overriding `--deploy-config` can bypass the required single-GPU configuration.

The regular vLLM resolver explicitly rejects managed flags such as --host, --port, --served-model-name, and --max-model-len; other backends similarly call validate_custom_args() before appending user arguments.

Please add equivalent reserved-flag validation for vLLM-Omni and extend the test to cover both --flag value and --flag=value forms. Once that is fixed and CI regarding this is green, I am comfortable approving this.

ramkrishna2910:
Thanks @fl0rianr — fixed in `d4e43ce0`.

`vllm_omni_args` is now validated against the managed launch flags before anything is appended, mirroring the other backends:

```cpp
static const std::set<std::string> reserved_flags = {
    "--host", "--port", "--served-model-name", "--max-model-len", "--deploy-config"};
std::string validation_error = validate_custom_args(omni_args, reserved_flags);
if (!validation_error.empty()) {
    throw std::invalid_argument("Invalid custom vllm-omni-server arguments:\n" + validation_error);
}
```

So overriding `--port`, `--deploy-config`, `--host`, `--served-model-name`, or `--max-model-len` now fails the load with a clear error instead of silently producing an ambiguous launch. `validate_custom_args()` already handles both forms, and I extended `test_custom_args` to prove it — `--port 9999` and `--port=9999`, `--served-model-name=foo`, `--max-model-len=8192`, a reserved flag mixed in with allowed ones, plus the allowed/empty cases. All green locally (`ctest -R custom_args`).

CI is running now.

ramkrishna2910:
Small follow-up (`5af0aa66`): the `custom_args` test was registered but not in the CI `ctest -R` allow-list, so it compiled but didn't actually run. Wired it into the C++ test job so the reserved-flag + quoted/escaped coverage executes in CI, not just locally.

fl0rianr (approved):
LGTM now, approved.

Issue #2529, opened 2026-07-02
Asset generation pipeline


pwilkin:
### Feature Description

I talked to @jfowers in detail on this one: I'd like to add strong asset generation capability to Lemonade. I think one of the ways in local models are underused today is their asset production capability. In fact, there are already good open source models that allow for the creation of quality assets i.e for game development and I'd like to add them to Lemonade. I've ensured that the models themselves are on permissive licenses, so that there are no worries about their usage.

I've created engines for the following models / model families:
* ThinkSound - a sound effect generation model (https://github.com/pwilkin/thinksound.cpp)
* Trellis.2 - a high-quality image-to-3D generation model, with a text-to-3D generation pipeline using chaining and background removal (which is also adapted) (https://github.com/pwilkin/trellis.cpp)
* OpenMOSS - a high-quality open source *multilingual* TTS model with voice cloning and a separate voice generation pipeline (https://github.com/pwilkin/openmoss)

In addition, I've adapted the excellent acestep.cpp library from @serveurpersoCom (https://github.com/ServeurpersoCom/acestep.cpp) for the music generation AceStep 1.5 model family.

All of the engines are based on GGML, which should be especially valuable to Windows users who don't want to experience the Python dependency hell. I've already built artifacts for Windows/Linux variants for Vulkan, ROCm and CUDA.

I will submit the PRs in the order of their complexity - since the sound generation doesn't require any new client-side mechanisms, it goes first.

### Use Case / Motivation

Asset generation for game dev especially: high quality sounds, music, 3D assets and reusable voices with voice cloning (for example for consistent NPC voiceovers).

### Platform Relevance

All platforms

jeremyfowers:
FYI @danielholanda @ramkrishna2910 @fl0rianr @kenvandine :)

pwilkin:
Aight, I guess that closes the issue :) I'll probably want to follow up with a metamodel for all the models launched here, but that's a separate topic.

PR #2530, opened 2026-07-02
feat(backends): audio generation — ThinkSound (SFX) + ACE-Step (music)

pwilkin:
Per #2529 : adds support for the ThinkSound model engine (sound generation) and the ACE-Step engine (music generation).


docs/assets/models.js
```diff
@@ -4,12 +4,14 @@ const RAW_BASE = 'https://raw.githubusercontent.com/lemonade-sdk/lemonade';
```
jeremyfowers, line 4: The GUI isn't moving into music generation mode when I have the ACE model selected in the right panel, which causes it to send /chat/completions requests instead of music generation requests.

IIRC your fork didn't have this issue, so hopefully it's a quick fix.

Image

jeremyfowers: Same issue for ThinkSound as well, the GUI stays in LLM Chat and tries to send /chat/completions.

src/cpp/resources/backend_versions.json
```diff
@@ -111,5 +111,11 @@
   "moonshine": {
     "cpu": "moonshine0.0.62"
   },
+  "thinksound": {
+    "vulkan": "v0.1.0"
+  },
+  "acestep": {
+    "vulkan": "v0.1.0"
+  },
```
jeremyfowers, line 123: Missing ROCm version pins?

docs/assets/models.js
```diff
@@ -4,12 +4,14 @@ const RAW_BASE = 'https://raw.githubusercontent.com/lemonade-sdk/lemonade';
 
 /* BEGIN GENERATED: models-js-recipes */
```
jeremyfowers, line 5: Can you add python test scripts to exercise the new endpoints on both vulkan and rocm? 

sd-cpp test script, as an example: https://github.com/lemonade-sdk/lemonade/blob/main/test/server_sd.py

And then its added to CI here for windows

https://github.com/lemonade-sdk/lemonade/blob/d67a9d7dece8f8a6c08e2b2cb2cc0d9d90e5d7a3/.github/workflows/cpp_server_build_test_release.yml#L1138

and here for linux

https://github.com/lemonade-sdk/lemonade/blob/d67a9d7dece8f8a6c08e2b2cb2cc0d9d90e5d7a3/.github/workflows/cpp_server_build_test_release.yml#L1421

src/cpp/server/server.cpp
```diff
@@ -3068,6 +3072,77 @@ void Server::handle_audio_speech(const httplib::Request& req, httplib::Response&
...
+    // Buffer the generated media so we can tell success from a silent failure. A
+    // streaming content provider commits a 200 before the backend runs, so a backend
+    // crash/OOM mid-stream would surface as a successful empty file; buffering lets us
+    // return a real error instead.
+    std::string buf;
+    httplib::DataSink sink;
+    sink.write = [&buf](const char* data, size_t len) { buf.append(data, len); return true; };
+    sink.is_writable = []() { return true; };
+    sink.done = []() {};
+    generate(sink);
```
jeremyfowers, line 3167: Consider also treating an error payload written into `buf` as failure (not just an empty buffer): `execute_streaming` and `forward_byte_stream` write `data: {"error":...}` / the backend's HTTP-error body into the sink on failure, so `buf` is non-empty and gets returned as valid audio. Without this, a backend error (e.g. `ts-server /generate` returning 500) surfaces as HTTP 200 `audio/wav` containing error text, which clients cannot distinguish from a real clip.

jeremyfowers (changes requested):
I'm really excited to see these features land! 

Please see requests above for GUI fixes and test infrastructure, which will help me further review the PR.

docs/assets/models.js
pwilkin: Ah, I separated GUI changes and I think Claude might have understood me a bit too literally :)

src/cpp/resources/backend_versions.json
pwilkin: Yup and CUDA as well for that matter.

src/cpp/server/server.cpp
```diff
@@ -3068,6 +3072,77 @@ void Server::handle_audio_speech(const httplib::Request& req, httplib::Response&
...
+        } catch (const std::exception& e) {
+            LOG(ERROR, "Server") << "Failed to load audio-generation model: " << e.what() << std::endl;
+            auto error_response = create_model_error(requested_model, e.what());
+            res.status = get_http_status_from_error(error_response["error"]["code"].get<std::string>());
+            res.set_content(error_response.dump(), "application/json");
+            return;
+        }
+
+        std::string mime_type = MIME_TYPES["wav"];
+        if (request_json.contains("response_format") && MIME_TYPES.contains(request_json["response_format"])) {
```
fl0rianr: This sets the response MIME type from response_format, but the current backends appear to always return WAV bytes: ACE-Step hardcodes `wav24`, and ThinkSound forwards WAV bytes from /generate. So a request with response_format: "mp3"`/`"flac" etc. would likely return WAV data labeled with a different MIME type.

Can we either reject non-wav formats for /audio/generations for now, or pass the requested format through only when the backend actually supports producing that format?

fl0rianr (commented):
Thanks for adding this — the audio generation direction looks useful and the backend integration is a good start. Jeremy already covered the GUI mode issue, ROCm version pins, tests, and error-buffer handling. I left one additional comment on response_format/MIME consistency below.

src/cpp/server/server.cpp
pwilkin: Yeah, let's do this properly, I'll add a supported mime types prop and only allow the supported ones, then we can possibly add conversion later.

pwilkin:
Aight, I think that should be all the changes :)

src/cpp/include/lemon/backends/thinksound/thinksound.h
```diff
@@ -0,0 +1,45 @@
...
+    /*binary*/          "ts-server",
+    /*config_section*/  "",  // defaults to recipe
+    /*default_device*/  DEVICE_GPU,
+    /*slot_policy*/     SlotPolicy::Standard,
+    /*selectable_backend*/ true,
+    /*uses_ctx_size*/   false,
+    /*dynamic_models*/  false,
+    /*options*/ {},
+    /*support*/ {
+        {"vulkan", {"linux", "windows"}, {{"cpu", {"x86_64"}}, {"amd_gpu", {}}, {"nvidia_gpu", {}}}, "Vulkan-capable GPUs"},
```
fl0rianr: This backend is marked selectable_backend and advertises vulkan/rocm/cuda, but the descriptor does not declare a thinksound_backend option. Since RuntimeConfig::recipe_options() only promotes nested thinksound.backend through descriptor-declared options, the selected backend does not become part of `RecipeOptions`.

The load implementation also ignores RecipeOptions and reads RuntimeConfig directly, so load-time/per-model options and option-change reload metadata will not reliably carry the selected backend. Same issue seems to apply to ACE-Step. Could we mirror e.g. the sd-cpp pattern here: declare `<recipe>_backend` in the descriptor and select/validate the backend from `options` inside load()?

docs/assets/models.js
jeremyfowers: Working for me now

docs/assets/models.js
jeremyfowers: Thanks!

test/utils/server_base.py
```diff
@@ -268,6 +268,10 @@ def _build_runtime_config(additional_server_args=None):
         config["sdcpp"] = {"backend": backend}
     elif wrapped_server == "whispercpp" and backend:
         config["whispercpp"] = {"backend": backend}
+    elif wrapped_server == "thinksound" and backend:
```
fl0rianr, line 271: Since this adds audio-generation backend selection for CI, should _THEROCK_RECIPES also include `thinksound` and `acestep`? Both recipes now advertise ROCm via TheRock / rocm-stable, and the workflow runs their ROCm tests. Without adding them to the prewarm allow-list, cold runners can fold the TheRock runtime download into the first inference request timeout instead of using TIMEOUT_ROCM_INSTALL, which could make these new ROCm jobs flaky.

fl0rianr (commented):
Thanks for the follow-up fixes — the response_format/MIME issue looks addressed now. I left two remaining comments that seem separate from Jeremy’s existing review: one around backend-option plumbing for the new selectable backends, and one around ROCm CI prewarming for the new audio-generation tests.

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -1140,6 +1140,16 @@ jobs:
             extra_args: ""
```
jeremyfowers, line 1140: Add documentation for the new endpoints to docs/api.

If these are openai-compatible, add them to docs/api/openai.md, if they're a separate standard they could be added to a new file similar to llamacpp.md, and if they're invented for this purpose put them in lemonade.md.

jeremyfowers (approved):
Once the API documentation is in I will be happy with the structure of this PR. I'm heading out on vacation soon, so I'm tagging a few more reviewers to help land this.

@ramkrishna2910 - new multimedia capabilities! Please try them out and ensure the API docs look good to you, then approve.

@superm1 - new ROCm integration! Please verify that it meets your patterns. If it doesn't I would suggest pulling the ROCm integration from this PR, then merge it, then add ROCm support back in later. It's still a very nice PR even if it only has Vulkan support.

@kenvandine / @fl0rianr - new CUDA support, please test it out on Nvidia hardware, then approve.

jeremyfowers:
PS. @pwilkin I realized that my two main comments on this PR (needs tests, and needs API docs) can be traced back to holes in the adding-a-backend.md guide. Here is a PR to improve that documentation: https://github.com/lemonade-sdk/lemonade/pull/2541

If you have any other suggestions for improving that guide please comment on that PR!

pwilkin:
@jeremyfowers I think @fl0rianr 's remarks regarding how to declare backends are also useful, maybe adding the SD.cpp code as an example.

pwilkin:
Aight, newest round of fixes is in.

pwilkin:
Can someone please reapprove the workflows? Needed a fix for ROCm. @jeremyfowers @fl0rianr

ramkrishna2910:
> Can someone please reapprove the workflows? Needed a fix for ROCm. @jeremyfowers @fl0rianr 

Done

src/cpp/resources/server_models.json
```diff
@@ -1629,6 +1629,24 @@
...
+        "size": 4.0
+    },
+    "ACE-Step-Music": {
+        "checkpoint": "Serveurperso/ACE-Step-1.5-GGUF:acestep-v15-xl-sft-Q8_0.gguf",
+        "recipe": "acestep",
+        "suggested": true,
+        "labels": [
+            "audio-generation"
+        ],
+        "size": 7.7
```
fl0rianr: The default size of the model is incorrect.
ThinkSound-SFX 4.00 GB stated, 11.43 GB downloaded
ACE-Step-Music 7.70 GB stated 10.50 GB downloaded

I’m seeing the displayed model size remain incorrect even after the model has been downloaded. For example, ACE-Step / ThinkSound show the static registry size in the UI, but the actual download/on-disk payload is much larger.

Updating this here helps before download, but after download the server should probably expose/use the real aggregate on-disk size. ModelManager::list_model_files() already computes `size_bytes` from `resolved_paths`, including directory-sized resolved paths, so the UI/API seems to be missing that aggregate and continuing to show info.size.

src/cpp/include/lemon/backends/acestep/acestep_server.h
```diff
@@ -0,0 +1,46 @@
...
+class AceStepServer : public WrappedServer, public IAudioGenerationServer {
+public:
+    static InstallParams get_install_params(const std::string& backend, const std::string& version);
+
+    AceStepServer(const std::string& log_level,
+                  ModelManager* model_manager,
+                  BackendManager* backend_manager);
+    ~AceStepServer() override;
+
+    void load(const std::string& model_name,
```
fl0rianr, line 26: This declares acestep_backend as a selectable backend option, but I still don’t see the corresponding frontend recipe-options wiring. The Model Options modal does not show a backend selector for ACE-Step/ThinkSound, likely because recipeOptionsConfig.ts does not include acestep / thinksound in RecipeName / RECIPE_OPTIONS_MAP and does not map acestepBackend / thinksoundBackend to acestep_backend / thinksound_backend.

Can we add the frontend option definitions/mappings as well? Otherwise users cannot choose Vulkan/CUDA/ROCm from the UI even though the backend supports it.

Image

pwilkin:
Looking good now, don't think the MacOS failure is caused by my changes?

fl0rianr (commented):
Great work so far. Vulkan works, but I'm not able to select CUDA. Typically we would prefer rocm and cuda first before vulkan (as the auto default). If you are aware of stability issues it's fine going for vulkan first.

Small UI polish issue: the native number-input spinner for `Duration` looks visually broken/misaligned in the dark audio-generation panel. The field is very narrow and the browser up/down buttons dominate the control. Very typical issue appearing with this UI element upon Initial implementation trials.

Image

fl0rianr:
> Looking good now, don't think the MacOS failure is caused by my changes?

Yes, this is a flaky one, don't worry about that.

pwilkin:
Eh, sorry, still getting acquainted with the codebase, once I get the hang of it (and add some proper memories for Claude to know how to do stuff) it'll go easier :)

pwilkin:
Should be good now, did E2E tests on my box on both ROCm and Vulkan.

fl0rianr (commented):
Don't worry, you are doing good work! Adding two backends is not any easy feat.
It's good to have the review attention so let me go on here:

Backend selection is fine now.
But CUDA startup seems to fail for both new audio-generation backends with exit code 127, while Vulkan works. Since ACE-Step and ThinkSound use the same CUDA env setup pattern, this looks like a shared CUDA packaging / dynamic-loader issue rather than a model-specific failure. 

On Linux this only prepends the executable directory to LD_LIBRARY_PATH, and apply_cuda_env_vars() only sets device/offload variables; You can take a look at the llama.cpp or sd-cpp cuda paths as well.
Did I get you correct, it works for you on CUDA?

Vulkan works great, we should try getting CUDA work, if it's too bothersome we can go without.

Minor: The spinner fix worked, up/down arrows are not bad visually and also not visible or functional anymore - but the number field works, if that was the intention we can leave it that way.

pwilkin:
Nah, it works on ROCm + Vulkan, I'll check CUDA on my second box :)

pwilkin:
Okay, turns out that I forgot to bundle the CUDA libraries :) simple enough fix.

Also, for whatever reason Claude decided that the *proper* fix to a problem of "the textbox is too narrow and text overflows into the spinner" is to... remove the spinner. UX genius. I'll push the right fix :)

fl0rianr:
Tell me once you have the cuda fix and I'll review again, good luck!

pwilkin:
@fl0rianr aight, all the CUDA packages have been reuploaded together with the libraries, tested it E2E on my CUDA box.

fl0rianr (commented):
Very nice, CUDA ist working as well!

A minor topic, we have this fully UI element again correctly separated now but still the dark mode issue.
Image

My local test worked but additionally CI runs are failing now.
It looks related to this selection not actually taking effect for the current registry entry. ThinkSound-SFX is registered as a bare repo checkpoint (ilintar/thinksound-gguf) with no `:variant`, and the shared HF downloader only calls select_checkpoint_files() inside the `!main_variant.empty()` branch. For an empty variant it falls back to downloading all repo files.

That matches the failing CI server logs: the ThinkSound test still tries to download dit-q8.gguf and t5-q8.gguf, even though this backend intends to skip the redundant Q8 variants. Those extra large downloads then hit hash/resume/retry issues and make the audio-gen jobs flaky/slow.

Can we make backend-specific select_checkpoint_files() apply even when main_variant is empty, or give the ThinkSound registry entry a variant/sentinel that routes through the backend selector? Otherwise CI and users will still download the full repo instead of the intended ~6.4 GB subset.

pwilkin:
Aight, hopefully we're done now 😃

pwilkin:
Argh. This reminds me why I hate DevOps / staging stuff. Will add the Windows libomp and that should fix the two remaining tests.

fl0rianr:
Yeah, nearly there but CI is still off. 

Another minor thing:
The spinner fix works for dark theme, but this now forces native form controls to use dark browser chrome in light theme as well. `styles.css` sets `color-scheme: dark` globally on `body`, while the app still supports `[data-theme="light"]`.

Can we make this theme-aware instead of hardcoding dark? For example, define `--native-color-scheme: dark` in `variables-dark.css`, `--native-color-scheme: light` in `variables-light.css`, and use `color-scheme: var(--native-color-scheme, dark)` on `body`. That keeps the dark-theme spinner fix while allowing light theme controls to render correctly.

fl0rianr:
I let the workflow run, so you can take a look, the spinner fix should be an easy one you can do after CI is green (besides the MacOS one)

pwilkin:
@fl0rianr yeah I swear they did something to nerf Opus, I don't remember it being so obtuse with respect to UX, that's twice now that it's done really bad CSS fixes. I honestly think I'll actually just code this one by hand to make sure ;)

fl0rianr:
The remaining CI failure looks centered on the new audio-generation jobs. Since the workflow currently runs both new backends across vulkan and rocm on Windows and Linux, I’d suggest narrowing required CI first: keep full /audio/generations coverage on the known-stable backend/platform path, and move the other variants to load/install smoke tests or optional follow-up jobs until the external backend artifacts are stable.

Also, ThinkSound failures are hard to debug right now because /generate is just forwarded; unlike ACE-Step, the backend error body is not logged before Lemonade returns 500. Adding explicit logging for the ThinkSound backend response body/status would make the next CI failure actionable.

pwilkin:
Okay, so the way forward was I basically rewrote the engines to use GGML's native threadpool instead of OpenMP, which cleared the problem with OpenMP library dependencies. Hopefully it works now.

pwilkin:
Nope: data type size mismatch with Linux vs Windows in the backends (overflow on 2GB+ reads). Needs another rebuild and version bump. Sigh :)

src/app/styles/styles.css
```diff
@@ -7632,3 +7632,29 @@ input::-webkit-calendar-picker-indicator {
...
+    appearance: none;
+    background: var(--bg-secondary);
+    border: 1px solid var(--border-primary);
+    border-radius: 4px;
+    color: var(--text-primary);
+    padding: 6px 8px;
+    font-size: 13px;
+    width: 64px;
+}
+
```
fl0rianr, line 7647: Add this here and we are fully merge ready from my side:

```
.audio-duration-label input {
  color-scheme: dark;
}

[data-theme="light"] .audio-duration-label input {
  color-scheme: light;
}
```

fl0rianr (approved):
Good work, Thanks! CI fully clean, just a minor nit to fix dark theme with the spinner, see below.

@ramkrishna2910 @superm1 can you please take a look if you can approve as well

src/app/styles/styles.css
pwilkin: Is this still needed? The previous regression was due to Claude inserting `color-scheme: dark` globally for some reason, after I removed it the inputs seem fine (tested on both light and dark theme). I don't think any of the other components explicitly set `color-scheme`.

fl0rianr: I tested on a Firefox and it was only fixed after those CSS insertions. I agree, ideally we would not need those but this is a potential flaw (I'm not an expert on this just noticed it happening over and over) of this component that needs this kind of manual patching, Noticed this a few times before. Atm Lemonade does not use this UI component at all.

pwilkin: Aight, will add. After looking at the CSS code I'm a huge fan of a general CSS rewrite, that 7000 line monolithic copy-paste CSS is a nightmare :)

pwilkin: Added :)

fl0rianr: Thanks! If you have interest in rewriting CSS - please DON'T do it.
We will move to a new GUI, GUI3 quite soon. We have the same problem there.
CSS is a AI weakness and maybe a better structure might help follow up with additional CSS code in a better way and less like with a "sledgehammer" approach to each recent design issue. But we can start with that once GUI3 lands - CSS is not the main focus there atm.

ramkrishna2910:
Thanks for the detailed review @fl0rianr. I will test drive this shortly.

superm1 (approved):

ramkrishna2910 (approved):
Thanks for the contribution @pwilkin, this is a great addition to list of modalities supported :)

PR #2535, opened 2026-07-02
Fix: flaky Whisper realtime transcription event handling

fl0rianr:
Fixes a flaky Whisper realtime WebSocket transcription test where the server
emitted conversation.item.input_audio_transcription.delta and `completed`
events, but the test only asserted text from `completed.transcript`.

The test now accumulates transcript text from delta events and falls back to the
completed transcript when present. Event field access is also made robust for
OpenAI SDK event objects and dict-like events.

- Preserves existing HTTP transcription, NPU, connection lifecycle, and manual
  commit test behavior

### DIFF
--- a/test/server_whisper.py
+++ b/test/server_whisper.py
@@ -381,13 +381,76 @@ class WhisperTests(ServerTestBase):
         print(f"[INFO] Split into {len(chunks)} chunks")
         return chunks
 
+    def _event_payload(self, event):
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@Geramy can you review? It would improve the stability of macOS CI testing quite a bit.

Geramy (approved):
This is correct.

PR #2537, opened 2026-07-02
server: add explicit preset profiles

fl0rianr:
Adds a server-side preset/profile store backed by `<cache_dir>/presets.json`, with REST endpoints for list/upsert/delete/import/export.

Presets are separate from per-model recipe_options, but can carry reusable recipe defaults, sampling defaults, UI/client tool metadata, and stable system prompt text.

`/load` and `/chat/completions` can opt into a preset explicitly via `preset`/`preset_id`; explicit request fields still override preset defaults, and preset extension fields are stripped before backend forwarding. Normal model loads and normal chat auto-load behavior remain unchanged when no preset is specified.

Why: This is a preparation for GUI3 on server side so presets are not a GUI only thing and allow TUI use as well.

### DIFF
--- a/src/cpp/server/preset_store.cpp
+++ b/src/cpp/server/preset_store.cpp
@@ -161,7 +161,7 @@ json PresetStore::builtin_presets() {
         make_preset(
             "s-quick-chat", "Quick Chat",
             "Small context, tight sampling. Snappy responses for quick interactions.",
-            json::array({"chat"}), {{"ctx_size", 4048}},
+            json::array({"chat"}), {{"ctx_size", 4096}},
             {{"temperature", 0.60}, {"top_p", 0.80}, {"top_k", 40}, {"repeat_penalty", 1.05}},
             "llamacpp",
             "Be brief and responsive. Answer the user's direct request with minimal setup.",
--- a/src/cpp/server/preset_store.cpp
+++ b/src/cpp/server/preset_store.cpp
@@ -1,0 +1,543 @@
+#include "lemon/preset_store.h"
+#include "lemon/utils/path_utils.h"
+
+#include <algorithm>
+#include <cctype>
+#include <filesystem>
+#include <fstream>
+#include <set>
+#include <stdexcept>
+#include <system_error>
+#include <utility>
+
+#include <lemon/utils/aixlog.hpp>
+
+namespace fs = std::filesystem;
+
+namespace lemon {
+
+namespace {
+
+fs::path store_path_for(const std::string& cache_dir) {
+    return utils::path_from_utf8(cache_dir) / "presets.json";
+}
+
+std::string string_or(const json& object, const std::string& key, const std::string& fallback = "") {
+    if (!object.is_object() || !object.contains(key) || !object[key].is_string()) {
+        return fallback;
+    }
+    return object[key].get<std::string>();
+}
+
+json object_or_empty(const json& value) {
+    return value.is_object() ? value : json::object();
+}
+
+json array_or_empty(const json& value) {
+    return value.is_array() ? value : json::array();
+}
+
+json make_system_prompt(const std::string& id, const std::string& name, const std::string& text, bool built_in) {
+    if (text.empty()) {
+        return json::object();
+    }
+    return {
+        {"id", id},
+        {"name", name},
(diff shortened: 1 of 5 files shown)


src/cpp/server/server.cpp
```diff
@@ -718,6 +721,44 @@ void Server::setup_routes(httplib::Server &web_server) {
...
+    web_server.Put(R"(/api/v1/presets/([A-Za-z0-9._:-]+))", [this](const httplib::Request& req, httplib::Response& res) { handle_preset_upsert(req, res); });
+    web_server.Put(R"(/v0/presets/([A-Za-z0-9._:-]+))", [this](const httplib::Request& req, httplib::Response& res) { handle_preset_upsert(req, res); });
+    web_server.Put(R"(/v1/presets/([A-Za-z0-9._:-]+))", [this](const httplib::Request& req, httplib::Response& res) { handle_preset_upsert(req, res); });
+    web_server.Delete(R"(/api/v0/presets/([A-Za-z0-9._:-]+))", [this](const httplib::Request& req, httplib::Response& res) { handle_preset_delete(req, res); });
+    web_server.Delete(R"(/api/v1/presets/([A-Za-z0-9._:-]+))", [this](const httplib::Request& req, httplib::Response& res) { handle_preset_delete(req, res); });
+    web_server.Delete(R"(/v0/presets/([A-Za-z0-9._:-]+))", [this](const httplib::Request& req, httplib::Response& res) { handle_preset_delete(req, res); });
+    web_server.Delete(R"(/v1/presets/([A-Za-z0-9._:-]+))", [this](const httplib::Request& req, httplib::Response& res) { handle_preset_delete(req, res); });
```
bitgamma: do these endpoints really need to be available in all prefixes? I'd say having them under /v1/api should be sufficient, right? Otherwise please consider adding a helper method to compute all these permutations.

src/cpp/server/preset_store.cpp
```diff
@@ -0,0 +1,543 @@
...
+            return true;
+        }
+    }
+    return false;
+}
+
+json PresetStore::builtin_presets() {
+    static const json presets = json::array({
+        make_preset(
+            "s-default", "Default",
```
bitgamma, line 141: I don't think the values listed here are actually model agnostic and might potentially impair the model's output. The two most prominent problems I see are:

- ctx_size: the values seem very arbitrary (especially 4048, why not at least a more common 4096?) while some are high and might go beyond what the model actually supports. The sampler settings also do not match any of the commonly suggested settings. For example the coding preset for Qwen3.6 has very different values from what you propose as a coding preset.
- sdcpp presets: the number of steps and cfg are *highly* dependent on the model. Distilled models want cfg set at 0, while non-distilled can go usually from 2~7. Some models require exactly 4 steps while some aren't happy until you do at least 50. If I select "s-quality" and use it with Flux 2 Klein I am likely to see worse results than with the "s-turbo". Conversely if I select "s-turbo" with Z-Image Base I will see literally a blob, which doesn't help with iteration. The combination of 8 steps with cfg > 1 is also quite strange because 8 steps is typical of distilled models but cfg > 1 is typical of non-distilled one.

this, for me, highlights a problem: if the presets are presented as a user-friendly way to use models without having to dig into the technical details, the presets here can be counterproductive. If they are meant for the advanced user, they probably would prefer to rely on per-model settings since there hardly is a "catch-all" for everything.

bitgamma (changes requested):
thank you for your contribution! I think the preset system would be a valuable addition but not in the form it is currently being presented. 

1. Built-in presets would work very well *if* they were model-aware, at least trying to rely on the architecture defaults or even on model-specific knowledge for commonly used ones. This is especially evident for diffusion models but also applies to LLM.
2. Giving users the ability to create custom presets while a per-model recipe system exist already puts too much cognitive stress on the user, who won't know which one they need to use in their case. Also since presets are unlikely to be really model agnostic, it is likely they would eventually degenerate into a variation of recipes.

My suggestion would be to improve the built-in presets to make them model-aware and remove the user-specific presets in their current form and instead leverage the recipe system, eventually extending it to support the desired use case in a way that doesn't add new concepts that overlap with existing ones.

PR #2540, opened 2026-07-03
fix: set SO_REUSEADDR in find_free_port to avoid TIME_WAIT false posi…

superm1:
…tives

The find_free_port function on all platforms (Linux, macOS, Windows) binds a test socket to check if a port is available. Without SO_REUSEADDR set, bind() fails with EADDRINUSE for ports in TIME_WAIT state — TCP sockets that were recently connected and are still in the kernel's TIME_WAIT recovery period (~60s on Linux).

This causes the port scanner to incorrectly report such ports as "in use" even though nothing is actively listening on them.

The fix: set SO_REUSEADDR on the test socket before binding. This is standard practice for port scanning and allows binding to TIME_WAIT ports.

On Windows, setsockopt takes a const char* pointer (not int*), so the argument is cast accordingly.

### DIFF
--- a/src/app/src-tauri/Cargo.lock
+++ b/src/app/src-tauri/Cargo.lock
@@ -1812,7 +1812,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2",
+ "socket2 0.6.3",
  "tokio",
  "tower-service",
  "tracing",
@@ -2228,6 +2228,7 @@ dependencies = [
  "objc2-web-kit",
  "serde",
  "serde_json",
+ "socket2 0.5.10",
  "tauri",
  "tauri-build",
  "tauri-plugin-clipboard-manager",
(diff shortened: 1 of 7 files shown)


Geramy (changes requested):
Could you please also fix the broadcast port listening in the cli, tray, and web app too? if one of those are listening at the same time they conflict.

Geramy (approved):
This is better, I will approve, can we get the same fix in the lemonade client api for the beacon? there should be a spot where I believe the tray icon also listens for a beacon too. I will approve even if you don't change that because this will resolve the dual listen conflict for the most part.

superm1:
> This is better, I will approve, can we get the same fix in the lemonade client api for the beacon? there should be a spot where I believe the tray icon also listens for a beacon too. I will approve even if you don't change that because this will resolve the dual listen conflict for the most part.

I see what you mean now, adjusted.  Thanks!

fl0rianr (commented):
I think this PR may also be related to the current Windows .exe CI failures.

find_free_port() is intended to answer "can a fresh server bind this port?". With SO_REUSEADDR enabled, that check may no longer reflect what a normal server process will observe, especially during rapid restart/shutdown cycles. This could allow a port to be selected while another process/socket is still in a transient state.

I would suggest keeping SO_REUSEADDR for the beacon sockets (where immediate rebinding is the goal), but not using it in the find_free_port() implementation unless there is a demonstrated need. That keeps the probe semantics aligned with the actual server bind behavior and may also explain some of the recent Windows .exe CI instability.

superm1:
this is probably not needed anymore with https://github.com/lemonade-sdk/lemonade/commit/248178fa9069a04209535b0c884736af0eedf2b7

PR #2541, opened 2026-07-03
docs: cover tests and API docs in adding-a-backend guide

jeremyfowers:
The [adding-a-backend guide](docs/dev/adding-a-backend.md) framed a backend as descriptor-only ("no doc edits"), so a new endpoint/capability like audio generation (#2530) landed without integration tests or API docs.

- **Testing** section: write a `test/server_*.py` against a live server (modeled on `server_sd.py`), wire it into both CI matrices, and add ROCm/TheRock recipes to the prewarm allow-list.
- **Adding a new endpoint or capability** section: capability interface, `ModelType` (+ `get_pinned_model_counts`), router dispatch, quad-prefix handler, and the `docs/api/` reference to write.
- Reflowed hard line breaks mid-paragraph document-wide (code blocks, tables, headings unchanged).

### DIFF
--- a/docs/dev/adding-a-backend.md
+++ b/docs/dev/adding-a-backend.md
@@ -1,10 +1,6 @@
 # Adding a backend
 
-Lemonade backends are **self-describing**. A backend declares *what it is* in a
-plain-data **descriptor** and implements *how it runs* in a **server class**, and
-both live together in the backend's own folder. A registry collects every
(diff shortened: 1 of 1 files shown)


bitgamma (approved):
looks good. I think it could also make sense to link/mention this document in AGENTS.md?

Geramy (approved):
Looks good to me.

PR #2545, opened 2026-07-04
feat(telemetry): reduce verbosity of telemetry logging

abn:
This change reduces the verbosity of the telemetry logging system at the default `INFO` log level to keep server console logs clean. Background batching/export logs are downgraded to `DEBUG`, and per-request multi-line timing statistics are replaced by a single-line summary including the model name.

### Changes
* **Exporters**: Downgraded span accumulation, batch transmission success/timeout, retries, and shutdown exporter logs in `telemetry.cpp` to `DEBUG`.
* **Inference Logs**: Replaced 6-line timing boxes in `server.cpp` and `streaming_proxy.h` with a single `INFO` line detailing the `model`, `tokens` (in/out/total), `ttft`, and `tps`.
* **Debug Mode**: Preserves detailed multi-line formatted statistics under the `DEBUG` log level for verbose analysis when needed.

### DIFF
--- a/src/cpp/include/lemon/streaming_proxy.h
+++ b/src/cpp/include/lemon/streaming_proxy.h
@@ -20,17 +20,23 @@ public:
         double time_to_first_token = 0.0;
         double tokens_per_second = 0.0;
         std::string error_message = "";
+        std::string model_name = "";
 
         void print() const {
             if (input_tokens > 0 || output_tokens > 0) {
-                LOG(INFO, "Telemetry") << "=== Telemetry ===" << std::endl;
-                LOG(INFO, "Telemetry") << "Input tokens:  " << input_tokens << std::endl;
-                LOG(INFO, "Telemetry") << "Output tokens: " << output_tokens << std::endl;
-                LOG(INFO, "Telemetry") << "TTFT (s):      " << std::fixed << std::setprecision(3)
-                          << time_to_first_token << std::endl;
-                LOG(INFO, "Telemetry") << "TPS:           " << std::fixed << std::setprecision(2)
-                          << tokens_per_second << std::endl;
-                LOG(INFO, "Telemetry") << "=================" << std::endl;
+                LOG(INFO, "Telemetry") << "Inference completed: model=" << model_name
+                                       << ", tokens=" << (input_tokens + output_tokens)
+                                       << " (in=" << input_tokens << ", out=" << output_tokens << ")"
+                                       << ", ttft=" << std::fixed << std::setprecision(3) << time_to_first_token << "s"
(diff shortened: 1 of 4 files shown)


src/cpp/include/lemon/streaming_proxy.h
```diff
@@ -23,14 +23,17 @@ class StreamingProxy {
...
+                                       << " (in=" << input_tokens << ", out=" << output_tokens << ")"
+                                       << ", ttft=" << std::fixed << std::setprecision(3) << time_to_first_token << "s"
+                                       << ", tps=" << std::fixed << std::setprecision(2) << tokens_per_second << std::endl;
+
+                LOG(DEBUG, "Telemetry") << "=== Telemetry ===\n"
+                                        << "Input tokens:  " << input_tokens << "\n"
+                                        << "Output tokens: " << output_tokens << "\n"
+                                        << "TTFT (s):      " << std::fixed << std::setprecision(3) << time_to_first_token << "\n"
+                                        << "TPS:           " << std::fixed << std::setprecision(2) << tokens_per_second << "\n"
+                                        << "=================" << std::endl;
```
abn, line 39: @fl0rianr I kept this for now, but I am on two minds as it is redundant information. I would like to use the concise version but not sure folks prefer the "exploded" view. Would appreciate your input here.

fl0rianr: I leaning towards keeping the expanded DEBUG block.

For the default INFO path, the concise single-line summary is definitely better. But when someone explicitly switches to DEBUG, the exploded view is useful and expected for diagnostics, even if it is somewhat redundant.

Small consistency note: the non-streaming server.cpp paths include model=... in the concise line, while this StreamingProxy path currently cannot because TelemetryData has no model context. I would either add model context here later or slightly adjust the PR description so it does not imply every concise inference log includes the model.

fl0rianr (commented):
Overall this looks good to me. The INFO/DEBUG split is the right direction: compact per-request summaries at INFO, expanded diagnostic output at DEBUG, and exporter lifecycle noise moved out of the default console path.

One small consistency nit: the PR description says the concise inference summary includes the model name. That is true for the non-streaming server.cpp paths, but StreamingProxy::TelemetryData::print() currently logs only tokens/ttft/tps because TelemetryData has no model context. Router telemetry still records against the right model via identity, so this is not a functional metrics issue, but the console output is slightly inconsistent.

I’d either:

1. add `model_name` to `StreamingProxy::TelemetryData` and populate it from `request_body` in `forward_sse_stream()`, or
2. adjust the PR description to say model is included for the server-side non-streaming paths.

abn:
Good catch; I have added the model name to the log consistently now.

fl0rianr (approved):
Looks good to me now.

Issue #2388, opened 2026-06-23
[Router] End-to-end tests + docs


ramkrishna2910:
## Goal
**End-to-end tests + docs** — the milestone's demonstrable exit criterion.

> 📋 **Execution plan (2026-07-16):** two runnable demos first — (A) both candidates local, (B) one local + one on fireworks.ai — then harden into the CI e2e test + docs. Deterministic + `semantic_similarity` paths; the real `classifier`-type condition is a separate follow-up (engine→backend wiring). See the [pinned plan comment](https://github.com/lemonade-sdk/lemonade/issues/2388#issuecomment-4995241637).

## Scope
- Python integration test in `test/` driving a `collection.router` collection with the `openai` client: a PII-flagged query routes to one candidate (via the `pii`/`semantic_similarity` classifiers), a short support query to another, everything else falls through to `default_model`. Assert `route_to` and inspect the returned `trace`.
- Dev doc (`docs/dev/`) + `docs/api/` update: authoring classifiers + rules in a `collection.router`, and the Decision/trace contract.

## Acceptance
Green end-to-end test in CI; docs cover policy authoring + the trace-on-response contract. (No verdict/block in the core demo — that's the Trust milestone.)

---
**Depends on:** #2386

ramkrishna2910:
## Execution plan — routing demo → e2e + docs

Assignee: @ramkrishna2910. Approach: **two runnable demos first** (prove the path works from `main`), then harden into the CI integration test + docs this issue asks for. The demos flush out the integration papercuts nobody has hit yet (no one has run the router end-to-end).

### Scope decisions (locked to what's actually implemented on `main`)
- Policy schema **v1**, `recipe: "collection.router"`, `routing` block (`candidates` / `default_model` / `rules` [/ `classifiers`]). Verified in code: dispatch (`Server::apply_router_collection_dispatch`, `server.cpp:2359`), engine first-match + fail-open (`routing_policy.cpp:1111`), trace-on-response + `x-lemonade-route` header (`route_decision_response.cpp`).
- Conditions used: **deterministic** (`keywords_any/all`, `regex`, `min_chars/max_chars`, `metadata`) and optionally **`semantic_similarity`** (embedding-backed, live via #2384's `embed()`).
[...]

PR #2548, opened 2026-07-04
feat(router): wire classifier services to router calls

eddierichter-amd:
## Summary

Part of #2384

  Implements the Lemonade Router `ClassifierServices` wiring layer. This bridges the pure routing engine to Router-style backend calls while keeping  `routing_policy.h` backend-free.

  The new adapter lets router policies evaluate:

  - `semantic_similarity` through `Router::embeddings(json)`
  - model-backed `classifier` conditions through Router-style chat completion returning label scores
  - future L0a `llm` router calls through Router-style chat completion text extraction

  ## What Changed

  - Added `routing_classifier_services`:
    - `make_router_classifier_services(Router&, ensure_loaded)` binds the pure engine’s `ClassifierServices` lambdas to live `Router` calls
    - `make_classifier_services_from_router_calls(...)` provides a deterministic test seam using Router-shaped JSON call functions
    - `embed()` builds an embeddings request and parses OpenAI-style embedding responses
    - `run_classifier()` builds a classifier chat request, invokes the Router-style chat call, and parses label/score output
    - `chat()` builds a prompt + user input chat request and extracts assistant text for future `routing.router` / L0a work

  - Added robust response parsing:
    - OpenAI embeddings shape: `data[0].embedding`
    - direct score maps: `{ "PII": 0.9 }`
    - nested score maps: `{ "labels": { "PII": 0.9 } }`
    - label-score arrays: `[{"label":"PII","score":0.9}]`
    - chat JSON content: `choices[0].message.content = "{\"PII\":0.9}"`

  - Added focused tests:
    - verifies `ensure_loaded` is called before Router-style service calls
    - verifies embedding request shape and vector parsing
    - verifies `semantic_similarity` scores through Router-style embeddings
    - verifies model-backed classifier routing through Router-style chat completion
    - verifies `chat()` extracts assistant content for future L0a use

  - Added CMake and CI wiring for `RoutingClassifierServicesTest`.

  ## Notes

  The pure engine still touches backends only through `ClassifierServices`; no backend or `Router` headers were added to `routing_policy.h`.

  The tests use Router-shaped JSON call adapters rather than real backend subprocesses. This verifies the `ClassifierServices` ↔ Router request/response contract deterministically. Live model-backed E2E should be covered in #2388 once `collection.router` dispatch is wired in #2385.

  This PR does not implement:

  - `collection.router` request dispatch (#2385)
  - `routing.router` / L0a desugaring (#2405)
  - live backend E2E docs/tests (#2388)

  ## Validation

  Ran locally:

  ```bash
  cmake --preset default
  cmake --build --preset default --target test_routing_classifier_services
  ctest --test-dir build --output-on-failure -R "RoutingClassifierServicesTest"
  ctest --test-dir build --output-on-failure -R "RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine)Test|RoutingClassifierServicesTest"
  cmake --build --preset default --target lemond
  git diff --check

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -285,8 +285,9 @@ jobs:
           cmake --build --preset default --target test_routing_policy_parser
           cmake --build --preset default --target test_routing_policy_store
           cmake --build --preset default --target test_model_manager_collection_validation
+          cmake --build --preset default --target test_routing_classifier_services
           cd build
-          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest"
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest"
 
       - name: Upload .deb package
         uses: actions/upload-artifact@v7
@@ -594,8 +595,9 @@ jobs:
           cmake --build --preset default --target test_routing_policy_parser
           cmake --build --preset default --target test_routing_policy_store
           cmake --build --preset default --target test_model_manager_collection_validation
+          cmake --build --preset default --target test_routing_classifier_services
           cd build
-          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest"
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest"
 
       - name: Upload .pkg package
         if: steps.check_signing.outputs.has_signing == 'true'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -631,6 +631,8 @@ set(SOURCES_CORE
     src/cpp/server/streaming_proxy.cpp
     src/cpp/server/system_info.cpp
     src/cpp/server/recipe_options.cpp
+    src/cpp/server/routing_classifier_services.cpp
+    src/cpp/server/routing_classifier_services_router.cpp
     src/cpp/server/routing_policy.cpp
     src/cpp/server/routing_policy_parser.cpp
     src/cpp/server/routing_policy_store.cpp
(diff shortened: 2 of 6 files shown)


fl0rianr (changes requested):
Thanks! This is a clean and well-scoped wiring layer overall. The Router/backend separation looks good, and I like the deterministic adapter seam for tests.

I think there is one blocking parser bug before merge: parse_classifier_scores() first calls `parse_scores_payload(response)` on the entire chat completion response. Since parse_scores_payload() eventually falls back to collecting numeric top-level object fields, a normal OpenAI-compatible chat response with `"created": 1742927481` can be interpreted as a score map `{ "created": 1742927481 }`, and the code will never parse choices[0].message.content.

Could you please either prioritize `choices[0].message.content` for chat-completion-shaped responses, or make the bare-object score-map fallback ignore known OpenAI metadata shapes? I would also add a regression test with a realistic chat completion response containing id, object, created, model, and choices, expecting the scores from the JSON content and not from `created`.

Minor note: #2384’s acceptance mentions live Router integration; this PR intentionally uses Router-shaped fakes and defers live E2E. That is probably fine as a slice, but then “Closes #2384” may be slightly too strong unless the team (which I'm not a part of so my opinion is secondary) is okay tracking live E2E in #2388/#2385.

SlawomirNowaczyk:
> I think there is one blocking parser bug

Thanks, fl0rianr, good catch; this was rather trivial, so I've now pushed a fix.

However, I need to flag a bigger issue here: this PR **does not**, actually, implement `run_classifier` interface, it only fakes it using chat interface. Overall, the `run_classifier` interface needs more work... **this should merge, but it can't close `#2384`**

This PR is a solid step forward and worth merging as it. However, `run_classifier()` is currently implemented as "prompt a generative chat LLM to emit JSON scores" over `chat_completion`. That's a useful stopgap, but it isn't really classification:

- The scores are generated tokens, not calibrated probabilities from a classification head.
- It only works with instruct models that reliably emit JSON, and it can't drive an actual classifier model at all.

As an example, take [`meta-llama/Llama-Prompt-Guard-2-86M`](https://huggingface.co/meta-llama/Llama-Prompt-Guard-2-86M) as the exactly the kind of model this feature exists to serve. It's an mDeBERTa `AutoModelForSequenceClassification` text-classification model that produces logits over `BENIGN`/`MALICIOUS`; you run it via `pipeline("text-classification")` / `argmax(logits)`. It has **no chat interface**, so `chat_completion` cannot execute it. 

The main challenge is that there is **no classification capability in Lemonade today** (server_capabilities.h has completion/embeddings/reranking/etc., but no `IClassifierServer` and no `/classify` endpoint). We should discuss how to

My proposal:

- **Merge this PR** for the wiring, but change "Closes `#2384`" to "Part of `#2384`" (or "Refs `#2384`").
- **Keep `#2384` open** until we can run a real classifier model — i.e. a genuine text-classification path (sequence-classification head → logits/softmax) exposed as a proper capability/endpoint, validated against something like Llama-Prompt-Guard-2-86M. Until then `type: "classifier"` effectively means "instruct model emitting JSON scores," which we should document as such.

SlawomirNowaczyk (approved):

fl0rianr (commented):
Thanks for the update, good improvement.

I still have one small concern before merge: the parser currently accepts any numeric value as a score, but the routing contract says classifier scores are in [0,1]. Since these scores can come from model-generated JSON, I think values outside [0,1] should be rejected so the classifier fails and the engine applies on_error, instead of allowing values like 999 or -1 to affect routing bands.

Minor consistency point: the PR description still says direct score maps such as { "PII": 0.9 } are supported. The current parse_classifier_scores() always calls extract_chat_text(response) first, so a raw direct score-map response without choices/content will now fail. That is fine if the intended contract is strictly “Router chat-completion response with JSON in assistant content”, but then the PR description should be narrowed. Otherwise, we could support both by parsing chat-shaped responses via message.content first, and falling back to parse_scores_payload(response) only for non-chat-shaped payloads.

eddierichter-amd:
> > I think there is one blocking parser bug
> 
> Thanks, fl0rianr, good catch; this was rather trivial, so I've now pushed a fix.
> 
> However, I need to flag a bigger issue here: this PR **does not**, actually, implement `run_classifier` interface, it only fakes it using chat interface. Overall, the `run_classifier` interface needs more work... **this should merge, but it can't close `#2384`**
> 
> This PR is a solid step forward and worth merging as it. However, `run_classifier()` is currently implemented as "prompt a generative chat LLM to emit JSON scores" over `chat_completion`. That's a useful stopgap, but it isn't really classification:
> 
> * The scores are generated tokens, not calibrated probabilities from a classification head.
> * It only works with instruct models that reliably emit JSON, and it can't drive an actual classifier model at all.
> 
> As an example, take [`meta-llama/Llama-Prompt-Guard-2-86M`](https://huggingface.co/meta-llama/Llama-Prompt-Guard-2-86M) as the exactly the kind of model this feature exists to serve. It's an mDeBERTa `AutoModelForSequenceClassification` text-classification model that produces logits over `BENIGN`/`MALICIOUS`; you run it via `pipeline("text-classification")` / `argmax(logits)`. It has **no chat interface**, so `chat_completion` cannot execute it.
> 
> The main challenge is that there is **no classification capability in Lemonade today** (server_capabilities.h has completion/embeddings/reranking/etc., but no `IClassifierServer` and no `/classify` endpoint). We should discuss how to
> 
> My proposal:
> 
> * **Merge this PR** for the wiring, but change "Closes `#2384`" to "Part of `#2384`" (or "Refs `#2384`").
> * **Keep `#2384` open** until we can run a real classifier model — i.e. a genuine text-classification path (sequence-classification head → logits/softmax) exposed as a proper capability/endpoint, validated against something like Llama-Prompt-Guard-2-86M. Until then `type: "classifier"` effectively means "instruct model emitting JSON scores," which we should document as such.

Good catch! I agree with this approach. I updated Closes to `Part of` in the comment.

eddierichter-amd:
> 8ea8b88

Thanks. Addressed in [8ea8b88](https://github.com/lemonade-sdk/lemonade/pull/2548/commits/8ea8b886cfddb177872e2b9616b89fa91cf1d4b1).

fl0rianr (approved):

PR #2549, opened 2026-07-05
fix: scan entire install_dir for executable permissions after extraction

### LINKED ISSUE #2547: llama.cpp backend binaries extracted without execute permissions — all model loads fail with 500
### Platform

Linux/Fedora

### Lemonade Version

10.9.0

### GPU / APU Model

AMD Ryzen AI MAX+ 395 w/ Radeon 8060S (Framework Desktop)

### Component

llama.cpp

### Bug Description

After a backend update on Jul 2, every binary in /var/lib/lemonade/bin/llamacpp/vulkan/ was left with mode 640 (rw-r-----) — no execute bit. lemond then fails to exec llama-server on every model load, and all POST /api/v1/load requests return 500. The server previously worked; the directory mtime and version.txt show the backend was re-extracted at 2026-07-02 20:09, which is when the permissions were lost. This looks like the backend download/extraction path not preserving (or setting) executable file modes.

### Steps to Reproduce

1. Working Lemonade install on Fedora 44 with llamacpp vulkan backend
2. Backend gets re-downloaded/updated (observed Jul 2)
3. Attempt to load any model (tried gpt-oss-120b-mxfp-GGUF and others)
4. Load fails with 500; log shows Failed to execute: /var/lib/lemonade/bin/llamacpp/vulkan/llama-server followed by exit code 1

Expetcted: model loads succeed
Actual: lemond fails to exec llama-server on every model load, and all POST /api/v1/load requests return 500.

### Log Output

```shell
Jul 04 16:52:35 desktop lemond[1716]: 2026-07-04 16:52:35.155 [Info] (ProcessManager) Process started successfully, PID: 58852
Jul 04 16:52:35 desktop lemond[1716]: 2026-07-04 16:52:35.155 [Info] (Process) Failed to execute: /var/lib/lemonade/bin/llamacpp/vulkan/llama-server
Jul 04 16:52:35 desktop lemond[1716]: 2026-07-04 16:52:35.256 [Error] (WrappedServer) llama-server process has terminated with exit code: 1
Jul 04 16:52:35 desktop lemond[1716]: 2026-07-04 16:52:35.838 [Error] (Server) Error 500: POST /api/v1/load
```

### Additional Context

Binaries in /var/lib/lemonade/bin/llamacpp/vulkan left with no execute permissions:
```
$ sudo ls -la /var/lib/lemonade/bin/llamacpp/vulkan/ | grep llama-server
-rw-r-----. 1 lemonade lemonade    17896 Jun 21 13:35 llama-server
-rw-r-----. 1 lemonade lemonade  6654024 Jun 21 13:34 libllama-server-impl.so
$ cat /var/lib/lemonade/bin/llamacpp/vulkan/version.txt   # modified Jul 2 20:09
```

Workaround:
```
sudo find /var/lib/lemonade/bin -type f ! -name '*.so*' ! -name '*.txt' ! -name 'LICENSE' -exec chmod u+x {} +
```

wariobot09:
## Summary

Fixes #2547 — backend binaries extracted without execute permissions, causing all model loads to fail with HTTP 500.

### Root Cause

The post-extraction `chmod` loop in `install_from_github()` only iterated over a `${install_dir}/bin/` subdirectory:

```cpp
auto bin_dir = fs::path(install_dir) / "bin";
// ... iterates only bin/
```

But the llama.cpp Vulkan tarball (and other backends) extract files directly into the install directory root, **not** under a `bin/` subdirectory. So the chmod loop never touched the extracted binaries, leaving them at mode 640 (no execute bit).

### Fix

Replace the single-level iteration with `fs::recursive_directory_iterator(install_dir)` to chmod **every** regular file under the install directory, regardless of where the archive places files. This handles both the existing `${install_dir}/bin/` structure and flat layouts.

### Changes

- `src/cpp/server/backends/backend_utils.cpp`: Replaced `directory_iterator(bin_dir)` + separate `chmod(exe_path)` with `recursive_directory_iterator(install_dir)` covering all files.

### Test Plan

- [x] Build passes (both `lemonade` CLI and `lemond` server compile cleanly)
- [ ] Verify backend install sets executable permissions on extracted binaries

### DIFF
--- a/src/cpp/server/backends/backend_utils.cpp
+++ b/src/cpp/server/backends/backend_utils.cpp
@@ -741,6 +741,28 @@ namespace lemon::backends {
                 }
             }
 
+            // Normalize executable permissions for every regular file in the
+            // staging tree.  Archives may place binaries under bin/ or directly
+            // in the tree root (the llama.cpp Vulkan tarball does the latter),
+            // and tarballs may strip the execute bit.  Recurse over the whole
+            // tree so no layout is missed.  Fixing in staging (not post-swap)
+            // preserves rollback on chmod failure.  On Windows chmod is a no-op.
+            #ifndef _WIN32
+            {
+                for (const auto& entry : fs::recursive_directory_iterator(staging_dir)) {
+                    if (entry.is_regular_file()) {
+                        if (chmod(entry.path().c_str(), 0755) != 0) {
+                            std::error_code ec;
+                            ec.assign(errno, std::generic_category());
+                            throw std::runtime_error(
+                                "Failed to set executable permission on staged file "
+                                + entry.path().string() + ": " + ec.message());
+                        }
+                    }
+                }
+            }
+            #endif
+
             // Verify the staged tree contains the executable, then atomically
             // swap it into place. commit_staged_install keeps a recoverable .old
             // backup across the swap: it removes the staging tree and leaves
@@ -756,22 +778,6 @@ namespace lemon::backends {
             staging_guard.active = false;
 
             LOG(DEBUG, spec.log_name()) << "Executable verified at: " << exe_path << std::endl;
-
-    #ifndef _WIN32
-            // Make all binaries in bin/ executable (tar may lose permissions)
-            {
-                auto bin_dir = fs::path(install_dir) / "bin";
-                if (fs::exists(bin_dir)) {
-                    for (auto& entry : fs::directory_iterator(bin_dir)) {
-                        if (entry.is_regular_file()) {
-                            chmod(entry.path().c_str(), 0755);
-                        }
-                    }
-                }
-            }
-            // Also make the found executable itself executable
-            chmod(exe_path.c_str(), 0755);
-    #endif
 
             // (The downloaded archive is removed by zip_guard on scope exit.)
 


fl0rianr (commented):
Good catch on the root cause: scanning only install_dir/bin misses flat backend archives like the Vulkan llama.cpp tarball.

I would not merge this exact version yet though. The chmod currently runs after commit_staged_install() has already promoted staging into install_dir and removed the .old backup. If recursive_directory_iterator throws or chmod fails, we can leave a newly promoted but still non-executable install in place with no rollback path. chmod return values are also ignored, so the installer may report success while the original issue remains.

Please move the permission normalization to the staging tree before commit_staged_install(), check chmod failures at least for the expected backend executable.

Also consider not chmodding every regular file to 0755; this currently marks version.txt/LICENSE/etc. executable too. Not fatal, but broader than necessary.

fl0rianr (changes requested):
Thanks for moving the chmod before the atomic swap and checking chmod failures. However, this is still not sufficient for vLLM. The vLLM validation now fails because vllm-server is executable, but the wrapper then execs bin/python3, which is still not executable:

vllm-server: line 18: .../bin/python3: Permission denied
exec: .../bin/python3: cannot execute: Permission denied

So the fix needs to normalize executable permissions for helper binaries/interpreters too, not only spec.binary. Please chmod at least all regular files under a staged bin/ directory plus spec.binary before commit_staged_install(), or chmod all regular files in staging with checked errors if we want the smallest hotfix.

After that, rerun Validate New vLLM Release. With the current vLLM CI failure, I would not approve yet.

fl0rianr (changes requested):
This is much closer: vLLM validation is now green, and chmod is correctly done before the atomic swap with error checking.

I still do not think this fully covers the original flat-archive case. The current code only chmods direct regular files in staging_dir/bin. If the expected backend binary is extracted directly into staging_dir, as described for the llama.cpp Vulkan layout in this PR/issue, no chmod happens at all. commit_staged_install/find_executable_in_dir only verifies that the file exists, not that it is executable.

Please also chmod the found spec.binary in staging_dir regardless of where it is located, and preferably use recursive_directory_iterator for staged bin/ so nested helper executables are covered too. After that and once the remaining llama.cpp queued job is green, I would be comfortable approving.

A solving addtion could be something like that:

```cpp
const std::string staged_exe = find_executable_in_dir(staging_dir, spec.binary);
if (!staged_exe.empty()) {
    if (chmod(staged_exe.c_str(), 0755) != 0) {
        std::error_code ec;
        ec.assign(errno, std::generic_category());
        throw std::runtime_error(
            "Failed to set executable permission on staged backend "
            + spec.binary + " at " + staged_exe + ": " + ec.message());
    }
}
```

fl0rianr (approved):
Thanks, LGTM now.

PR #2557, opened 2026-07-06
Start an abstraction layer for benchmark types

ckuethe:
This just moves the body of what was run_single_bench into a new function called run_single_bench_textgen, which is unconditionally called.

You can probably guess where this is going, I just want to have a consistent place from which to dispatch different benchmark types with the same function signature. That'll make it easier add `run_single_bench_embed`, `run_single_bench_image`, `run_single_bench_vision`, etc.


fl0rianr (approved):
Looks good to me. This is a clean no-behavior-change refactor that introduces the dispatcher layer for future benchmark types.

Tiny non-blocking nit: if run_single_bench_textgen is meant to stay an implementation detail behind run_single_bench, consider keeping it private in bench.cpp instead of exposing it in bench.h, so future callers do not bypass the dispatcher accidentally.

ckuethe:
`bench.h` already exposes `run_single_bench` which runs a single completion benchmark; this doesn't change that exposure. I'm inclined to keep that pattern, but also don't object to moving the prototype out of the header once there are more benchmark types.

PR #2574, opened 2026-07-06
feat(telemetry): add WebSocket server endpoint and security controls

abn:
This change introduces the foundation for a deep integrated inference debugging experience for Lemonade application users. This builds on top of the existing inference telemetry work in #2330.

As this deals with user data, I am also going to capture the technical design for this change here below.

Note: The UI associated with this change is described and implemented at #2575.

---
## 1. System Architecture Overview

The tracing pipeline enables real-time visual inspection of inference operations (such as LLM completions, embeddings, and reranking). The diagram below illustrates how telemetry spans travel from concurrent C++ HTTP request handlers, through the WebSocket broadcast server, and down to the React/TypeScript frontend.

```mermaid
sequenceDiagram
    autonumber
    actor Client as Client / GUI
    participant Front as UpgradableFrontServer (Platform)
    participant HTTP as RoutedHttpServer (Platform)
    participant Router as C++ Router / Engine
    participant Tel as Telemetry (InferenceSpan)
    participant WS as WebSocketServer (Telemetry Additions)
    
    %% WebSocket Connection Setup
    Client->>Front: GET /spans/stream (Upgrade request)
    Note over Front: MSG_PEEK delegates to WS
    Front->>WS: adopt_socket(fd)
    Client->>WS: Send Auth message (apiKey, clientSessionId)
    WS-->>Client: Connection established (auth.ok)
    
    %% Traced Inference Request
    Client->>Front: POST /v1/chat/completions (with auth & session headers)
    Front->>HTTP: Delegate request
    Note over HTTP: Extract thread-local auth & session
    HTTP->>Router: Execute inference
    Router->>Tel: start_span()
    Router->>Router: Perform inference
    Router->>Tel: end_with_success()
    Tel->>WS: emit_span(span_json)
    Note over WS: Filter by token/session & strip auth token
    WS->>Client: Send filtered span over WS
    Note over Client: Update Inspect Store & Render
```

## 2. Main Port Socket Routing & Telemetry Integration

The inspect stream routes over the main HTTP port (e.g., `8080`) by integrating with the platform's pre-existing socket-adoption infrastructure (`UpgradableFrontServer`).

To route and adopt the new telemetry endpoints:
1. **Endpoint Classification**: We registered `/spans/stream` and `/traces/stream` as valid WebSocket paths in `WebSocketServer::classify_path` under the `ConnectionKind::spans` enum.
2. **Dynamic Delegation**: The pre-existing HTTP header peeking logic was updated to delegate path verification via `lemon::is_websocket_endpoint(path)` instead of hardcoded checks. This enables the front server to hand over socket descriptors for inspect connections directly to the `libwebsockets` context.

## 3. Telemetry Integration & Listener System

While the core `InferenceSpan` and OTel-compliant telemetry tracking were pre-existing on the platform (originally for forwarding traces to external OTLP endpoints), we extended the telemetry engine to support real-time local subscription:

1. **Thread-Local Context Injection**: We introduced thread-local variables (`g_current_auth_token`, `g_request_start_time`, and `g_current_client_session_id`) set on request ingress. When an `InferenceSpan` is created, it extracts these to compute queueing latency (`lemon.queue_time_ms`) and associate the span with the requesting client (`lemon.auth_token` and `lemon.client_session_id`).
2. **In-Memory Pub-Sub (`SpanListenerCallback`)**: We added a span listener registry (`register_span_listener`, `emit_span`) to telemetry. In `InferenceSpan::submit_span`, the span details are dispatched in-memory to all registered listeners. The `WebSocketServer` hooks into this callback on startup to capture completed spans.
3. **On-Demand Span Activation**: We modified `TelemetryTracker::start_span` to instantiate spans if `has_span_listeners()` is true, ensuring spans are tracked when a client connects to the inspect socket even if global OTel logging is disabled in the server config.

## 4. Security Model

The WebSocket stream employs a two-tier security model:

### Two-Tier Authentication
* **Connection Handshake**: Credentials can be supplied via standard HTTP `Authorization` headers or URL parameters (`?api_key=...`).
* **Protocol Auth Message**: If browser environments prevent custom headers during native `new WebSocket()` instantiation, the connection is initially accepted in an unauthenticated state. The server expects an immediate JSON message of type `"auth"` containing the token. Trace data is withheld until the socket transitions to the `authenticated` state.

### Multi-Tenant Span Isolation & Filtering
Spans are broadcast selectively based on the client's credentials:
* **Admin Access**: Clients authenticated with `LEMONADE_ADMIN_API_KEY` receive all traces generated across the server.
* **User Isolation**: Standard authenticated clients (`LEMONADE_API_KEY`) receive only spans where the span's `lemon.auth_token` matches their connection token.
* **Guest & Tab Isolation**: If no API keys are configured, clients receive spans matching their specific `client_session_id`.

### Credential Sanitization
To prevent credential leaks, the server performs a deep clone of the span JSON before sending it over the wire, deleting the `lemon.auth_token` attribute to scrub the raw API key from payloads.

## 5. Performance Considerations

Real-time tracing minimizes critical-path inference overhead through:

* **Dedicated Event-Loop Thread**: `libwebsockets` executes on its own service thread (`service_thread_`), resolving connection states via localized mutexes (`connections_mutex_`).
* **Non-blocking Asynchronous Writes**: When broadcasting a span, the server queues the text payload into a per-connection message queue and signals the socket via `lws_callback_on_writable`. Writing is handled by the event loop when the TCP socket is ready.
* **Client-Side Limits**: The React store (`inspectStore.ts`) limits memory retention to `MAX_STORED_TRACES = 100` and saves traces in `localStorage` to survive refreshes.

### Internal OTLP Exporter vs. Internal WebSocket Broadcast
Lemonade supports both exporting OTLP JSON over network protocols if configured, and hooking directly into the telemetry lifecycle in-memory for the local WebSocket broadcast. This allows the GUI's Inspect panel to work out-of-the-box without requiring an external OTel collector.

### LocalStorage Persistence vs. Server-Side Database
Storing traces in the browser's `localStorage` was chosen to keep the server stateless and lightweight. Traces are limited to 100 items to fit comfortably under the 5MB browser quota.


fl0rianr (changes requested):
Thanks for the detailed design write-up. The direction looks useful, but I think this should be tightened before merge because the current implementation mixes routing/security metadata into normal telemetry data and also broadens unauthenticated WebSocket behavior beyond the new spans endpoint.

Main concerns:

1. Raw API key can become part of exportable telemetry

`g_current_auth_token` is copied into the span as a normal `lemon.auth_token` attribute. The WebSocket broadcast path removes that attribute before sending to clients, but `submit_span()` still receives the original span payload and can push it to the OTLP exporter when telemetry is enabled.

That means a Lemonade API key can leak to an external telemetry backend. I do not think auth material should ever be represented as a normal span attribute.

Suggested fix: keep routing information outside the exported span payload. For example, use an internal envelope like `{ span_json, routing_token_hash, client_session_id }`, or at minimum strip `lemon.auth_token` before any OTLP queue/export path as well. A regression test for “OTLP payload never contains lemon.auth_token” would be good.

2. On-demand span activation is currently effectively always enabled while the WebSocket server is running

`WebSocketServer::start()` registers a global span listener immediately. `TelemetryTracker::start_span()` then creates spans whenever `has_span_listeners()` is true. Because the global listener list is populated at server startup, this seems to enable span construction even when no `/spans/stream` client is connected.

That undermines the intended on-demand behavior and adds avoidable privacy/performance overhead.

Suggested fix: either register/unregister the telemetry listener based on active span subscribers, or make `has_span_listeners()` reflect active authenticated span connections instead of global callback registration.

3. Guest/session isolation is incomplete for subscribers without a `client_session_id`

In guest mode, `guest_allowed` permits anonymous spans, and the session filter only rejects when the subscriber has a non-empty `client_session_id` that mismatches. A guest span subscriber without a `client_session_id` can therefore receive all anonymous spans, including spans from other client sessions.

Suggested fix: for guest span streams, require a non-empty `client_session_id` and only deliver when both sides match. The tests should include the negative case: subscriber has no session id, request has `X-Client-Session-Id`, and the subscriber must not receive that span.

4. Message-based auth also changes behavior for `/realtime`

`authenticate_connection()` now allows a WebSocket upgrade without credentials whenever an API key is configured, so browser clients can authenticate by message later. However this applies to all WebSocket paths, not only `/spans/stream`. For `/realtime`, `handle_connection()` immediately creates a realtime session before the client has authenticated.

Suggested fix: limit unauthenticated upgrade + message-based auth to the spans endpoint, or delay realtime session creation until after successful authentication.

Minor follow-ups:

* Several telemetry queue logs were moved from DEBUG to INFO. That may be too noisy in normal usage, especially for span accumulation/batching/retry messages.
* `test_001_admin_privilege_rule` extracts the model via a fixed attribute array index. It would be more robust to use the existing `get_span_attributes()` helper so the test does not depend on attribute ordering.

Overall, the design is promising, but I would prefer not to merge until the API-key handling and active-subscriber gating are fixed.

abn:
Thanks for the quick review @fl0rianr. I’ve updated and force-pushed this branch to address your review feedback.

Here’s a summary of the changes:
* **Bearer Token Scrubbing**: We now clone the span details and strip `lemon.auth_token` before pushing to the OTLP queue, so raw API keys are never leaked to external telemetry platforms (while still broadcasting them locally for the Web UI). Added `test_008_otlp_api_key_redaction` to verify this.
* **On-Demand Span Tracking**: Telemetry callbacks are now registered and unregistered dynamically based on active authenticated subscribers. The state transition safely happens outside of locks to avoid AB-BA deadlocks.
* **WebSocket Upgrade Authentication**: Unauthenticated upgrades are blocked immediately for `/realtime` and `/logs/stream`. We still permit unauthenticated upgrades on `/spans/stream` since it relies on message-based auth, but we now gate actual telemetry broadcasting on a successful auth message.
* **Guest Session Isolation**: Enforced a strict, non-empty `client_session_id` requirement for guest connections. Guest users now only receive trace spans where the session ID matches theirs exactly. Added negative integration tests verifying this behavior.
* **Logs & Test Improvements**: Reverted telemetry log levels from `INFO` back to `DEBUG`, and robustified `test_001_admin_privilege_rule` using attribute key lookups instead of hardcoded indexes.

All integration tests compile and pass successfully. Let me know if everything looks good to you now!

fl0rianr (commented):
Thanks for the update. The major concerns from the previous review look mostly addressed now. 
To point to consider remain:

1. `lemon.auth_token` is still placed into the generic span object and passed to emit_span() before being scrubbed for OTLP/export. Today the only listener appears to be the WebSocket broadcaster, which strips it again before sending, so the immediate external leak risk is addressed. Still, this keeps raw auth material inside the generic telemetry payload and means any future listener gets the raw token by default. A safer design would keep routing/auth data outside the exported/listener-visible span payload, for example in an internal routing envelope or by using a non-reversible token/session key.

2. The new integration test prints full server logs in `tearDownClass`. That is useful while debugging, but it can make CI output very noisy, especially when multiple temporary servers are started. I would either remove that before merge or only print logs on failure.

With those addressed or accepted as follow-up tradeoffs, this looks much safer than the first version.

abn:
I've pushed an update addressing the remaining feedback:

* **Token Hashing & Salting**: To keep raw auth keys out of the generic telemetry span payload, we now transition `lemon.auth_token` to `lemon.auth_token_hash`. At startup, the server lazily generates a random 256-bit salt in a thread-safe manner, which is prepended to the token before hashing with SHA-256 (using the compiled mbedTLS library). 
* **Broadcaster & OTLP Updates**: The WebSocket broadcaster and connection states have been updated to check and filter against `authenticated_token_hash`, and we explicitly strip `lemon.auth_token_hash` from both OTLP submissions and WS broadcasts to ensure it never leaves the server process memory. The regression tests in `server_websocket_telemetry.py` have been updated to assert this.
* **Test Logs Cleanup**: Removed the unconditional log-dumping logic inside `tearDownClass` in the Python integration tests to clean up CI runner logs.

fl0rianr (approved):
Thanks for the follow-up. This version addresses the main security concerns I had.

Only small nit from my side: hash_token() could check the return values from the mbedTLS `starts/update/finish` calls and fail closed instead of returning the zero-initialized digest if something goes wrong. That is not a blocker for me, but it would make the routing hash path more robust.

Pending CI validation, this looks good to me.

abn:
@fl0rianr that is a worthwhile hardening I reckon. So went ahead and made that fix too. :smile:

abn:
The CI failure reported is the same as the one seen in #2576 and appears fixed by https://github.com/lemonade-sdk/lemonade/pull/2576#issuecomment-4909727011.

PR #2575, opened 2026-07-07
 feat(ui): add session inspector client and interface

abn:
This is the first version of the inspect view build on top of #2330 and #2574.  This view helps capture and inspect inference request telemetry within the Lemonade app.  This is extremely useful for AI application/agent developers and users of local agentic applications. 

**Note:** This branch also carries the changes from #2574 as it is a pre-requisite for this feature.

### Demo

https://github.com/user-attachments/assets/d227a564-0733-4dd5-ad87-99f5f9995617

### DIFF
--- a/prototype/ui-redesign/src/App.tsx
+++ b/prototype/ui-redesign/src/App.tsx
@@ -14,10 +14,11 @@ import BackendManager from './components/BackendManager';
 import Dashboard from './components/Dashboard';
 import LogViewer from './components/LogViewer';
 import DownloadManager from './components/DownloadManager';
+import InspectView from './components/InspectView';
 import { Icon } from './components/Icon';
 import { DownloadListItem, downloadStore, isDownloadActive } from './features/downloadManager/downloadStore';
 
-type View = 'chat' | 'models' | 'presets' | 'backends' | 'dashboard' | 'logs' | 'connect';
+type View = 'chat' | 'models' | 'presets' | 'backends' | 'dashboard' | 'logs' | 'connect' | 'inspect';
 
 /* ── Error boundary ────────────────────────────────────────── */
 
@@ -60,7 +61,7 @@ class ViewErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState
   }
 }
 
-const VALID_VIEWS: View[] = ['chat', 'models', 'presets', 'backends', 'dashboard', 'logs', 'connect'];
+const VALID_VIEWS: View[] = ['chat', 'models', 'presets', 'backends', 'dashboard', 'logs', 'connect', 'inspect'];
 
 
 type HostNavigationPayload = string | URL | {
@@ -411,6 +412,7 @@ const App: React.FC = () => {
             { id: 'presets',   label: 'Presets',   icon: 'sliders-horizontal' },
             { id: 'backends',  label: 'Backends',  icon: 'box'                },
             { id: 'dashboard', label: 'Dashboard',  icon: 'gauge'              },
+            { id: 'inspect',   label: 'Inspect',   icon: 'search-check'       },
             { id: 'logs',      label: 'Logs',      icon: 'logs'               },
             { id: 'connect',   label: 'Connect',   icon: 'plug'               },
           ] as { id: View; label: string; icon: Parameters<typeof Icon>[0]['name'] }[]).map(({ id, label, icon }) => (
@@ -491,6 +493,11 @@ const App: React.FC = () => {
         <div style={{ display: view === 'logs' ? 'contents' : 'none' }}>
           <ViewErrorBoundary view="logs">
             <LogViewer />
+          </ViewErrorBoundary>
+        </div>
+        <div style={{ display: view === 'inspect' ? 'contents' : 'none' }}>
+          <ViewErrorBoundary view="inspect">
+            <InspectView accountSession={accountSession} />
           </ViewErrorBoundary>
         </div>
         <div style={{ display: view === 'connect' ? 'contents' : 'none' }}>
--- a/prototype/ui-redesign/src/api.ts
+++ b/prototype/ui-redesign/src/api.ts
@@ -17,7 +17,7 @@ function detectDefaultBaseUrl(): string {
       return `${window.location.protocol}//${window.location.hostname}:${window.location.port}`;
     }
   }
-  return 'http://localhost:13305';
+  return 'http://127.0.0.1:13305';
 }
 
 const DEFAULT_BASE_URL = detectDefaultBaseUrl();
@@ -414,7 +414,10 @@ export interface RealtimeTranscriptionHandle {
   isConnected: () => boolean;
 }
 
-export type LemonadeRequestInit = Omit<RequestInit, 'body'> & { body?: unknown };
+export type LemonadeRequestInit = Omit<RequestInit, 'body'> & {
+  body?: unknown;
+  includeSessionHeaders?: boolean;
+};
 
 export type LemonadeContextDefault = number | 'auto';
 
@@ -547,6 +550,8 @@ class LemonadeAPI {
   public readonly clientSessionId: string = (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function')
     ? crypto.randomUUID()
     : Array.from({ length: 32 }, () => Math.floor(Math.random() * 16).toString(16)).join('');
+  public sessionHeadersEnabled = false;
+  public onSessionHeadersFailed?: () => void;
 
   // ── Config ──────────────────────────────────────────────────────
   // Non-secret connection settings may be persisted in browser storage.
@@ -562,7 +567,7 @@ class LemonadeAPI {
       const parsed = new URL(raw);
       if ((parsed.hostname === 'localhost' || parsed.hostname === '127.0.0.1') && typeof window !== 'undefined') {
         const winHost = window.location.hostname;
-        if (winHost && winHost !== 'localhost' && winHost !== '127.0.0.1' && winHost !== '[::1]' && winHost !== '::1') {
+        if (winHost && winHost !== 'localhost' && winHost !== '[::1]' && winHost !== '::1') {
           parsed.hostname = winHost;
         }
       }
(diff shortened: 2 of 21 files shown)


fl0rianr (changes requested):
Thanks for the work on this — the Inspect view is a valuable addition and I like the direction. I focused the review on likely blockers rather than UI/details.

I think this should not be merged yet because I see a few regressions/security-impacting issues:

1. **Global custom headers may break existing Connect flows**
   `X-Client-Session-Id` and `X-Account-Session-Id` are now added to all API requests. This can trigger CORS preflight where we previously had simple requests, and can break connections against older `lemond` versions, proxies, or mixed UI/server setups. I think these headers should be limited to the endpoints that actually need them, not globally attached to every request.

2. **WebSocket Origin validation looks too strict**
   The new Origin check only accepts loopback origins. That can reject valid LAN, remote, dev-server, desktop/webview, or same-host non-loopback setups. Since this check applies before path handling, it can affect existing realtime/log WebSocket behavior too, not only the new inspector stream. I think this needs to validate against the actual request host / configured allowed hosts, or be scoped much more narrowly.

3. **Telemetry span creation appears to be effectively always enabled**
   The WebSocket server registers a span listener at startup, and `start_span()` creates spans whenever a span listener exists. That means local capture overhead may be active even when no Inspect client is connected. This should depend on actual active span subscribers, not just on the registered broadcast callback.

4. **“Ephemeral” UI copy conflicts with localStorage persistence**
   The UI says traces are ephemeral / cleared when the app closes, but traces and capture state are persisted in `localStorage`. Since prompts and outputs can be sensitive, this should either be in-memory/session-only storage or the UI needs to clearly state that traces persist locally.

Also, the failing Playwright test (number 12) seems real but partly test-side: the mocked `WebSocket` should define static constants like `OPEN`/`CONNECTING`, because the store compares against `WebSocket.OPEN` and `WebSocket.CONNECTING`. The simulated span should also include a stable `spanId`, and the test would be more robust if it first asserted the inspect store state before checking `.trace-row`.

Overall: great feature direction, but I would fix these before merging because they affect connection compatibility, WebSocket behavior, telemetry overhead, and trace privacy.

abn:
@fl0rianr. you are here, you are there, you are everywhere! Appreciate it!

 I’ve updated and force-pushed this branch to address your review feedback.

Here’s a summary of the changes:
* **Header Scoping**: Custom headers (`X-Client-Session-Id` and `X-Account-Session-Id`) are no longer attached globally. They are scoped exclusively to inference routes and WebSocket query parameters.
* **Origin Checking**: Relaxed the Origin verification to match the request's `Origin` host against the `Host` header. This preserves security while letting LAN/remote UI setups connect without errors.
* **UX & Retention Copy**: Kept local storage persistence for a smoother user experience, but updated the inspector description to clearly state: *"Local history — persisted in browser storage, never on server"*.
* **Playwright & Cleanups**: Fixed WebSocket state constants on mocks, resolved preflight OPTIONS routing stalls in Playwright, and deleted the redundant `toggle-height.spec.ts` script. We also corrected the Anthropic cURL endpoint to `/v1/messages` and added single-quote escaping for copied prompts.

The entire webapp and inspector E2E Playwright suite is compiling and passing successfully. Let me know if this works!

fl0rianr (commented):
Thanks for the quick follow-up — this is much improved.

The main blockers from my previous review look addressed now:

* Session headers are no longer attached globally and are scoped to inference/capture paths.
* The Origin check is less restrictive and now allows same-host LAN/remote setups instead of loopback-only.
* The Inspector WebSocket only connects while capture is enabled.
* Telemetry listener registration is now tied to active authenticated span clients.
* The local history copy now correctly says that traces persist in browser storage.

One small remaining thing I would fix before merging: the Playwright CORS mocks should allow both session headers. `includeSessionHeaders` sends both `X-Client-Session-Id` and `X-Account-Session-Id`, but at least `inspect-persistence.spec.ts` only allows `X-Client-Session-Id` in its mocked `Access-Control-Allow-Headers`. The real server already allows both, so this is mostly a test/mock consistency issue, but it can still cause local browser preflight failures.

Suggested change in the test CORS header:

```ts
'Access-Control-Allow-Headers': 'Content-Type, Authorization, X-Client-Session-Id, X-Account-Session-Id',
```

After that, I’m okay with this direction.

abn:
I've pushed an update addressing the mock headers feedback:

* **CORS Mock headers**: Updated the Playwright CORS intercept mock configurations in both `inspect-interactions.spec.ts` and `inspect-persistence.spec.ts` to include `X-Account-Session-Id` in `Access-Control-Allow-Headers`. This aligns the test environment mocks with the real server's configuration and prevents preflight/OPTIONS failures during E2E testing.

fl0rianr (approved):
Good and ready to merge, but I'm waiting with the merge to see if I should bring @kpoineal fixes in first.

fl0rianr:
Strange, I have this error while trying talking to a model: 
image
Had not the time yet to delve deeper

fl0rianr (changes requested):
See above

abn:
@fl0rianr that looks like a network issue? I do not think it is related to changes in this branch. I did notice that the gui3 does have flakey api interactions - maybe should consider adding an internal router or something.

Let me know if I am missing anything.

fl0rianr:
It's 100% your PR, as I said had no time yet

abn:
No worries. Let me know what you find out, I will try and reproduce it locally, have not been able to do that yet.

abn:
Found it, That was indeed a regression introduced in my commit `3328ac98` ("harden localhost resolution").

The root cause was when the server `lemond` binds only to IPv4 (`127.0.0.1:13305`), which is common if `"host": "127.0.0.1"` is configured in `config.json`, but the client attempts to reach `http://localhost:13305`, modern browsers (like Firefox and Chrome) resolve `localhost` to `::1` (IPv6). The connection to `::1` fails with a `NetworkError` because the server is only listening on the IPv4 loopback. 

Before my change, the client aligned `localhost` to `window.location.hostname`. The change to block loopback rewrites (`winHost !== '127.0.0.1'`) accidentally forced the browser to use `localhost`, triggering the IPv6 connection failure.

I have pushed an update addressing this:
1. **Default to IPv4**: Changed `DEFAULT_BASE_URL` in the client from `http://localhost:13305` to `http://127.0.0.1:13305`. This directly targets the IPv4 loopback, completely bypassing browser IPv6 `localhost` DNS resolution issues.
2. **Align Loopback Override**: Removed the `127.0.0.1` check from the exclusion guard so that if you access the dev server via `http://127.0.0.1:8080`, the client correctly rewrites loopback requests to route to `127.0.0.1:13305`.
3. **Regression Test**: Added a new integration test to verify that outgoing loopback requests resolve to IPv4 by default.

fl0rianr (commented):
I think the latest fix does not address the reported runtime issue.

Updating the Playwright CORS mocks is good, but the actual normal chat path still sends `includeSessionHeaders: true` unconditionally for every `/api/v1/chat/completions` request. That means regular chat requests always include `X-Client-Session-Id` and `X-Account-Session-Id`, independent of whether Inspect capture is enabled.

This explains why the error appears even when Inspect is inactive. The issue is not the Inspect WebSocket; it is that normal chat requests now carry new custom headers.

In cross-origin/dev/proxy/desktop setups, those headers force a browser CORS preflight. If the user is running against an older `lemond`, a proxy, a snap/server build without this PR, or any path that does not handle `OPTIONS` plus both new headers, the browser blocks the request before the POST reaches the server. The frontend then reports it as a generic fetch/network error.

So I do not think the current test mock fix is sufficient. Normal chat should not require the new session headers unless Inspect capture is actually enabled, or the client should have a compatibility fallback that retries once without session headers on fetch/preflight failure.

Suggested direction:

```ts
includeSessionHeaders: inspectStore.getState().capturing
```

or equivalent wiring from the chat call site, avoiding a circular dependency if needed.

Once we have it fixed an merge conflict gone we are good to go.

fl0rianr:
Thanks, Issue partly solved, have the same on when auto-capture is on though.

fl0rianr (commented):
Mmhh, works fine, but now inspect does not get the run anymore...

abn:
:thinking: works on my machine (:tm:)

https://github.com/user-attachments/assets/1b209910-3b2d-4347-9c77-1159c2cd5729

fl0rianr (commented):
I think the new fix explains why normal chat works again but Inspect capture no longer records anything.

The current implementation introduces a global `api.sessionHeadersEnabled` flag. Inspect sets it from the capture state, but `_fetch()` disables it permanently for the current browser session whenever a request with session headers fails or hits selected HTTP errors, then retries without session headers.

That makes regular chat work again, but it silently removes the data Inspect needs. In guest/local mode the server relies on `X-Client-Session-Id` to add `lemon.client_session_id` to the span, and `broadcast_span()` then uses that value to route the span back to the matching Inspect WebSocket session. If the fallback request succeeds without `X-Client-Session-Id`, the span either has no session id or cannot match the Inspect connection, so the UI stays empty even though capture appears enabled.

So the issue is not solved by the server-side changes alone. Once the frontend fallback has disabled `sessionHeadersEnabled`, the measurement path is broken from the client side.

I think this needs a different split:

* normal chat should avoid session headers when capture is off,
* capture mode should only send measured requests once the `/spans/stream` WebSocket is connected/authenticated,
* if session headers are not supported by the current server/proxy, chat may retry without them, but Inspect should be marked unavailable/paused instead of silently staying enabled,
* and `_fetch()` should not globally/permanently disable the capture transport behind the Inspect UI.

There is also a smaller race: right after enabling capture, the first request can be missed if the WebSocket/auth handshake has not completed yet. The UI should either wait for span socket readiness or expose a “connecting capture…” state before measured requests are expected.

abn:
Great find there. Globally disabling headers on the client side silently broke the measurement path while leaving the UI pretending everything was active, which is very confusing.

I have updated the branch to implement the proper scoping and state split:
1. **Scoped to Socket Handshake**: We now only enable session headers on the client side (`api.sessionHeadersEnabled = true`) once the telemetry WebSocket `/spans/stream` has successfully established and authenticated (on receiving `auth.ok`).
2. **Clear UI States**: Added `captureReady` states to the `InspectState` store:
   - **`CONNECTING...`** (Blue): Displays while the socket connection is establishing.
   - **`CAPTURING · X`** (Green): Displays once authenticated, showing the capture count.
   - **`UNSUPPORTED`** (Red): Displays if the server/proxy does not support session headers (replaces the silent failure).
3. **Explicit Fallback**: If a preflight/request fails or returns `4xx` due to session headers, the API triggers a `onSessionHeadersFailed` callback. The store reacts by setting the state to `unsupported`, disabling `sessionHeadersEnabled`, closing the WebSocket, and toast-alerting the user.
4. **Mock Testing**: Updated/added E2E tests `23` and `24` in `features.spec.ts` to mock the WebSocket authentication handshake and test the fallback/unsupported state changes.

fl0rianr (approved):
I'm gonna merge once you cuttend everything outside of prototype since this will make trouble with merging main into this branch. inspect will work once the main PR lands.

abn:
Done; dropped `lemond` changes from this branch. It will be handled in #2574.

PR #2576, opened 2026-07-07
fix(server): replace dummy bind with connect check for port availability

abn:
Replaces the dummy `bind()` call in `port_is_available` with a non-blocking TCP `connect()` check to probe for an active listener on the configured port. This avoids false-positive `EADDRINUSE` errors on restart when there are lingering client connections in states like `FIN_WAIT_2` or `TIME_WAIT` (particularly due to wildcard `0.0.0.0` vs. specific IP specificity conflicts on Linux).

- **Wildcard Translation**: Connects to the local loopback interfaces (`127.0.0.1` / `::1`) when checking wildcard addresses (`0.0.0.0` / `::`).
- **Non-blocking Probe**: Uses a non-blocking TCP `connect` with a `100ms` `select` timeout to ensure the startup guard remains fast.
- **Select Safety**: Adds a `sock < FD_SETSIZE` bounds check on POSIX platforms before modifying `fd_set`.


ckuethe:
Works for me on 3 different machines that were exhibiting that listen nonsense.

fl0rianr (commented):
Looks good to me conceptually. Replacing the dummy bind probe with a non-blocking connect probe should avoid the restart false positives from lingering TCP states while still letting the real bind path report any actual bind failure.

Two small robustness notes before merge: It would be nice to handle failures from fcntl/ioctlsocket/getsockopt explicitly by closing the probe socket and letting the real bind path decide. Also worth noting that the wildcard-to-loopback probe detects active wildcard/loopback listeners, but may not catch a listener bound only to a specific non-loopback interface; the later real bind still catches that, so I don’t see it as a blocker.

Now also a style thought, not a blocker:

As said this fix looks reasonable to me, but it adds another small cluster of `_WIN32` / POSIX socket handling directly in server.cpp. Since #2142 moved the codebase toward extracting platform-specific behavior into dedicated platform implementations, it might be worth considering the same direction here as well. e.g. we could extract the port-listener probe into a small platform/network utility, something like `is_tcp_listener_active(host, port, family)`, so the server startup logic stays platform-neutral. What do you think?

abn:
Thanks for the review! I've updated the PR to address your feedback:

- **Refactored to Platform Utils**: Extracted the port-probing check into `is_tcp_listener_active` inside dedicated platform network utility files (`network_unix.cpp` and `network_windows.cpp`), keeping `server.cpp` platform-neutral.
- **Improved Error Handling**: Added explicit checks for `fcntl`/`ioctlsocket`/`getsockopt` failures. If any of these calls fail, we now close the socket immediately and return `false` (allowing the real bind path to make the final call).
- **Unit Testing**: Added a new C++ unit test (`test_network_utils.cpp`) to verify active and inactive port check scenarios for both IPv4 and IPv6 loopbacks.

Let me know what you think!

fl0rianr (commented):
Thanks, this addresses my previous feedback nicely. Moving the listener probe into network_unix.cpp / network_windows.cpp keeps server.cpp much closer to the platform-abstraction direction from #2142, and the explicit fcntl / ioctlsocket / getsockopt handling looks better now.

One small test robustness nit before I’d call it fully ready: test_inactive_port() uses the fixed port 13999. That can theoretically be occupied on a developer or CI machine and make the test flaky. The active-port test already uses the better pattern: bind to port 0, let the OS choose a free port, then read it back with getsockname(). Could we use a similar helper for the inactive-port case instead of hardcoding 13999?

abn:
Good catch! I've updated the test. Refactored `test_inactive_port()` to dynamically request an OS-assigned loopback port (binding to `port 0`) and close it to verify the inactive port path. This ensures the target port is free and avoids hardcoding `13999` to prevent flakiness in CI.

Let me know if there's anything else!

fl0rianr (commented):
The remaining CI failure looks potentially related, not just generic flakiness. 

Most workflows are green now, and the only failing workflow I see is `C++ Server Build, Test, and Release`, specifically the macOS `.dmg` job in the `Run C++ unit tests` step. It's a flaky one, but I let it run 3 times. Since this PR adds and registers `NetworkUtilsTest`, and the failure is exactly in the C++ unit-test phase on macOS, I think we should treat it as related until proven otherwise.

The actual runtime fix still looks intact to me: the port check remains a non-blocking `connect()` probe with `select()` / `getsockopt(SO_ERROR)`, not the old dummy `bind()` probe, so the behavior Chris confirmed should still be preserved.

Could you check/paste the failing CTest section from the macOS job? My suspicion would be the new network utility test being macOS-sensitive rather than the server fix itself being wrong.

abn:
@fl0rianr can you re-approve please I want to see if https://github.com/lemonade-sdk/lemonade/pull/2576/commits/35b40ced16c3cddf3cf6f7530e5d6482c917b1d4 fixes it.

abn:
Looks like the fix in https://github.com/lemonade-sdk/lemonade/pull/2576/commits/35b40ced16c3cddf3cf6f7530e5d6482c917b1d4 did the tick.

fl0rianr (approved):
LGTM now

PR #2577, opened 2026-07-07
Bench embed

ckuethe:
New benchmark type: embedding
This was easy to implement, and it gives another example for agentic code
generation to use as an example.

Terminal output with the included scenario file looks like this:
```
Benchmark: nomic-embed-text-v2-moe-GGUF
=============================================================================

Backend: llamacpp/rocm (ctx=4096)
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
embed-small-string  243.1   240.0   246.0   82.3    81.3    83.3    0.7
embed-small-array   270.7   263.9   277.6   129.4   126.1   132.6   0.7
embed-long-string   10.4    9.9     10.8    0.0     0.0     0.0     0.7
embed-long-array    757.2   690.0   819.1   718.0   660.4   784.1   0.7

Backend: llamacpp/vulkan (ctx=4096)
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
embed-small-string  41.4    39.9    44.4    483.8   450.6   500.7   0.3
embed-small-array   64.2    63.1    64.7    545.6   540.6   554.8   0.3
embed-long-string   9.9     9.6     10.5    0.0     0.0     0.0     0.3
embed-long-array    488.4   424.5   533.3   1118.4  1014.4  1274.4  0.3
```

Output capture works. I don't know why you'd want to look at the big vector
of numbers, but the data is available if you want it.

Depends on #2557

Co-authored-by: copilot, Nemotron-3-Nano-A3B-GGUF


docs/guide/cli.md
```diff
@@ -760,7 +748,8 @@ Scenarios are defined in JSON files. Each file contains a `scenarios` array wher
 |-------|------|-------------|
 | `name` | string | Unique scenario name (required) |
 | `category` | string | Category label for grouping (default: `"general"`) |
-| `messages` | array | Chat messages in OpenAI format (required) |
+| `messages` | array | Chat messages in OpenAI format (required for text generation) |
+| `messages` | array | Chat messages in OpenAI format (required for embedding test) |
```
bitgamma: why is the parameter duplicated? if this has the same name and format, it should be listed only once (eventually just update the notes accordingly)

docs/guide/cli.md
```diff
@@ -784,11 +773,14 @@ Lemonade ships with a bundled set of scenarios (`bench_scenarios.json`) covering
 - **Chat** — Short and long conversational turns
 - **Coding** — Code generation, explanation, and debugging
 - **Long-context** — 32K, 64K, 128K context windows and multi-turn conversation memory
+- **Embedding** - Not part of the default scenario set
 
 You can override these with `--scenario-file` or `--scenario-dir`.
 
 **Note:** Long-context scenarios (`context-32k`, `context-64k`, `context-128k`, `context-multi-turn`) are excluded by default because they run very long. Use `--scenarios long-context` to include them, or `--scenarios all` to run every scenario.
 
+Embedding are excluded by default, but can be selected with `--scenario-file /path/to/bench_scenarios_embed.json`.
```
bitgamma: if we are shipping default embedding scenarios, then there should be a way to run them without having to specify the path.

src/cpp/cli/bench.cpp
```diff
@@ -598,6 +613,92 @@ BenchRunResult run_single_bench(lemonade::LemonadeClient& client,
...
+        }
+
+        // Extract usage statistics
+        if (resp_json.contains("usage") && resp_json["usage"].is_object()) {
+            // prompt_tokens = input tokens, total_tokens = output tokens
+            result.input_tokens = resp_json["usage"]["prompt_tokens"].get<int>();
+            result.output_tokens = resp_json["usage"]["total_tokens"].get<int>();
+        }
+
+        // Populate timing fields
```
bitgamma, line 671: I think textgen and embed can share a lot of code. Let's move the common parts to helper methods to avoid duplication

bitgamma (changes requested):
thanks for your contribution. Looks good, apart from a few issue I'd like to see solved before merging (also please fix the merge conflict)

docs/guide/cli.md
ckuethe: I forgot to commit my update. My fixed version uses "input", because that's what the OpenAI API calls it.

src/cpp/cli/bench.cpp
ckuethe: I did think about that, but I wanted to get a few more bench types implemented to get a better sense of how to refactor.

docs/guide/cli.md
ckuethe: Happy to update the default scenarios. I kept it separate to make review easier.

src/cpp/cli/bench.cpp
bitgamma: well, textgen and embeddings share at least the same metrics so I see no reason to implement their calculation twice.

ckuethe:
Looks like a flaky test...

superm1 (approved):

bitgamma (approved):

PR #2580, opened 2026-07-07
Asset generation: OpenMOSS

pwilkin:
This is leg two of my asset generation pipeline from #2529 . OpenMOSS is a high-quality open source TTS model family that exhibits two key qualities that make it extra valuable for this feature (generating voices for NPCs / game characters): it has a separate *voice generation* model that creates a voice sample based on provided characteristics and it's an actually  *multilingual* model capable of supporting quite a few language. Lemonade supports TTS already via Kokoro, but I feel like this is a completely different use-case: Kokoro is great because it's small and fast and can act as a live TTS model, but it does not support multiple languages nor can it generate and clone custom voices like OpenMOSS can.

This implementation features a mechanism, also present in my third integration, that I called "model cascade". Basically, it chains model calls to obtain a given result. If the request is voice cloning, the model behaves typically, doing a one-shot generation. However, if the request is voice generation given an (a) voice descriptor and (b) the text to convert, the cascade first calls MOSS-VoiceGenerator to perform the sample generation, then calls the main model to perform the actual conversion using cloning from the prepared sample.

I've integrated all the know-how from the previous PR here, so hopefully this will be easier to review :)


fl0rianr (changes requested):
Thanks for putting this together! The integration is quite substantial and the overall direction looks promising. I especially like that OpenMOSS is modeled as a proper backend/recipe with explicit WAV-only capability, and that the sidecar .extras.gguf handling is accounted for.
But this is a big change - Yes, I do see a couple of issues that I think should be addressed before merging:

1. **Please make backend errors from OpenMOSS propagate as HTTP errors, not as successful audio responses.**
   The Lemonade /audio/speech path now buffers non-streaming audio via serve_media_or_error, which is a good direction. However, OpenMOSS can return non-200 text/plain errors, for example for invalid reference_wav_b64 or invalid request parameters. Since the byte-stream proxy writes the backend body into the sink before checking the backend status, those text errors can end up as a non-empty buffer and then be served back as 200 audio/wav.

   I think this should be fixed in the proxy/wrapper layer so non-200 byte-stream responses become proper Lemonade errors. It would also be good to add negative tests for:

   * invalid reference_wav_b64
   * non-WAV base64 data
   * invalid speed
   * non-string input

2. **The TTS panel should not use an arbitrary globally selected model as the TTS model.**
   The current TTS panel derives ttsModel from `selectedModel || appSettings.tts.model.value`. Since the global model selector can contain non-TTS models, this can make the TTS panel call /audio/speech with an LLM/image/audio-generation model. I think the panel should either use appSettings.tts.model.value as the source of truth, or only accept selectedModel when it is actually labeled as TTS. Ideally the model selector in the TTS panel would be filtered to TTS-capable models.

A few additional things to consider, not necessarily blockers:

* The voice-clone upload currently accepts `audio/*`, but the backend expects WAV bytes via reference_wav_b64. It would be safer to restrict this to WAV files and validate the RIFF/WAVE header client-side before sending.
* Persisting voice samples as base64 strings in app settings may have privacy/storage implications. It may be worth either keeping samples session-only, adding an explicit clear button, or at least documenting that samples are stored locally.
* The UI currently makes OpenMOSS feel like it requires either voice design or cloning, but the backend can also do plain/style-guided TTS without a reference WAV. A third mode like “Plain/style” could make the feature easier to use when only OpenMOSS-TTS is installed.

Overall: I would be happy with the direction once the error propagation and TTS-model selection issues are fixed.

pwilkin:
I wish there was a way to penalize CC by taking away half its tokens or something for failing to follow clear instructions - I explicitly told him to watch out for all the things that were mentioned during the review of the ThinkSource/AceStep PR. Sigh. Oh well, fixes are up.

fl0rianr (commented):
Thanks for the quick follow-up — this addresses the main concerns I had.

A few non-blocking things to consider:

* The ModelSelector shown inside the TTS panel still appears to be the global selector. Even though non-TTS selections are now ignored by the TTS logic, this could still be slightly confusing in the UI because the visible selected model may not be the model actually used for TTS. A TTS-filtered selector would make this clearer.
* Voice samples are now WAV-validated and can be cleared, which is good. Since they are still persisted as base64 in app settings, it may be worth adding a small storage/privacy note or a size limit later.
* OpenMOSS-TTS can also do plain/style-guided TTS without a generated/cloned reference sample. The current UI focuses on “Describe” and “Clone”, so a future “Plain/style” mode could make the feature easier to use when only OpenMOSS-TTS is installed.

I would not block on those points. From my side this looks reasonable once CI has actually run and passed. I let it run so you can fix if something occurs there.

fl0rianr:
fyi, the current fails macOS tauri and openSUSE are typical flaky ones, sadly

pwilkin:
Aight, further fixes are in.

fl0rianr (approved):
LGTM now, thanks!

src/cpp/server/backends/openmoss/openmoss_server.cpp
```diff
@@ -0,0 +1,174 @@
...
+    }
+    backend_manager_->install_backend(spec->recipe, backend);
+    return BackendUtils::get_backend_binary_path(*spec, backend);
+}
+
+void OpenMossServer::load(const std::string& model_name,
+                          const ModelInfo& model_info,
+                          const RecipeOptions& options,
+                          bool do_not_upgrade) {
+    (void)do_not_upgrade;
```
superm1: what's the point of this declaration?

src/cpp/server/backends/openmoss/openmoss_server.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    std::vector<std::string> args = {
+        "--model", model_path,
+        "--host", "127.0.0.1",
+        "--port", std::to_string(port_),
+        "--no-webui",
+    };
+
+    std::vector<std::pair<std::string, std::string>> env_vars;
+    if (backend == "rocm" || backend == "cuda") {
```
superm1: this seems like it should be split.  you're applying a bunch of rocm stuff in cuda path at end.

``
if (backend == "rocm")
...
elif (backend == "cuda')
```

File: src/cpp/server/backends/openmoss/openmoss_server.cpp [hunk 3]
```diff
@@ -0,0 +1,174 @@
...
+
+    const std::string model_path = model_info.resolved_path();
+    if (model_path.empty() || !std::filesystem::exists(model_path)) {
+        throw std::runtime_error("Model path not found for checkpoint: " + model_info.checkpoint());
+    }
+
+    std::string backend = options.get_option("openmoss_backend");
+    if (backend.empty()) {
+        auto supported = SystemInfo::get_supported_backends("openmoss");
+        backend = supported.backends.empty() ? "vulkan" : supported.backends[0];
```
superm1: rather than hardcoding the backend when empty it would be better to pass up `not_supported_error`

src/cpp/server/backends/openmoss/openmoss_server.cpp
pwilkin: Avoid unused variable warnings.

pwilkin:
Okay, except for the MacOS build all is clear now.

src/cpp/server/backends/openmoss/openmoss_server.cpp
superm1: It wouldn't make more sense to plumb the upgrade handling?

pwilkin: The upgrade handling is already plumbed in, as in, it always takes the newest version of the file from the repository. What's not plumbed in is version retaining and at this point I'm not really sure if it makes much sense to add it.

superm1: Okay. Let's leave as is then.

PR #2581, opened 2026-07-07
Asset generation: Trellis

pwilkin:
This is the third leg of my asset generation pipeline described in #2529 . It adds a new modality: 3D to the Lemonade capabilities. For that, I've ported the entire pipeline from Trellis.2 (the leading open-source 3D generation model right now) to GGML. This PR also contains a custom component for the UI that implements 3D model preview using a bundled Three.js library.

The PR adds two distinct generation capabilities. The first one is the native Trellis.2 image-to-3D pipeline, with an optional run through the additional background removal model that also has been ported and is part of the package. The second one, similarly to the OpenMOSS PR, is a *model cascade* pipeline. This time, you can *select* a text-to-image model that you want to use and the pipeline will first use that model, with an adapted prompt optimized for generating images suitable for 3D conversion, to generate the image from the text, then chain Trellis.2 into the image-to-3D generation.


fl0rianr (changes requested):
Great! This is a substantial integration and the direction looks very promising. I like that Trellis is added as a proper backend/recipe, that the 3D capability is wired through the server API instead of being UI-only, and that the PR already includes docs and CI coverage for the new path.

That said, because this introduces a new modality and a fairly large model/backend surface, I think a few things should be tightened before merging:

1. **Please preserve or validate the uploaded image MIME type.**  
   The UI accepts `image/*` and reads the file as a Data URL, but the image flow strips the `data:image/...;base64,` header before sending it to /3d/generations. The C++ wrapper then always forwards the decoded bytes as input.png with image/png, regardless of whether the original upload was JPEG, WebP, PNG, etc. Since real image-to-3D usage will commonly start from JPEG photos, I think we should either keep the Data URL header and parse it server-side, pass the MIME type separately, or infer it from magic bytes before constructing the multipart upload.

2. **Please move cheap request validation before model autoload.**  
   handle_3d_generations currently only checks that model and image are present, then calls auto_load_model_if_needed() before validating the shape/value of most fields. That means malformed requests can still trigger a large Trellis model load before being rejected. I think we should validate at least:
   * model is a string
   * image is a string and looks like base64 or a supported data URL
   * response_format is `glb` while only GLB is supported
   * resolution is one of the supported values
   * bg_removal is one of the supported values
   * seed, if present, is an integer

   These should ideally fail fast with 400 rather than surfacing as backend/500/502-style errors after startup.

3. **Please confirm the `TRELLIS_512=1` behavior.**  
   The UI and docs expose 512, 1024, and 1536, but the backend process is always started with TRELLIS_512=1. I may be missing the exact semantics of that env var, but if it limits the loaded cascade to 512, then the higher-resolution options should either be hidden/rejected for now or the startup/config path should actually enable them.

4. **Docs/API status should match the backend status.**  
   The docs mark /v1/3d/generations as fully_available, while the Trellis backend descriptor marks the backend as experimental. I think the public docs should call this experimental/beta until we have more confidence across the supported platforms/backends.

A couple of smaller follow-ups that are not necessarily blockers:

* The large-file pre-commit exception currently allows all files under `src/app/src/renderer/vendor/`. It would be safer to narrow this to the exact vendored file and add a short vendor note with source/version/license/update process.
* The new 3D tests are very useful, but the negative/error-path tests could avoid pulling the 15.4 GB model where possible. Also, the invalid-model test currently accepts a very broad set of status codes; it would be better to pin the expected behavior once the validation/error contract is finalized.

Overall I really like where this is going. I would just prefer to tighten these API and packaging edges before we merge a new user-facing modality.

pwilkin:
That's quite a good list of valid complaints :) I thought I cleaned up the envvar-driving on the Trellis engine, but the TRELLIS_512 stayed somehow. Anyways, fixes pushed.

fl0rianr (commented):
Thanks, this is much tighter now. The main points from my previous review look addressed.

One remaining thing I would still tighten before approval: unsupported-but-valid-base64 image payloads can still trigger the Trellis model load before being rejected. handle_3d_generations currently checks that image is a string and contains base64-compatible characters before auto_load_model_if_needed(), but the actual PNG/JPEG/BMP/GIF magic-byte validation happens later in TrellisServer::model_3d_generations(), after the model has been loaded.

Could we move the base64 decode + sniff_image() check into handle_3d_generations before auto_load_model_if_needed()? That would make the “cheap validation before large model load” guarantee complete. A good regression test would be a valid base64 payload that decodes to unsupported bytes, for example `base64("not an image")`, and assert it returns 400 without pulling/loading the model.

Small optional UI polish: the file input still accepts image/*, while the API now documents PNG/JPEG/BMP/GIF only. It might be better to restrict the picker to `image/png,image/jpeg,image/bmp,image/gif`, or at least rely on the server-side preflight above.

After that, I think this is in good shape from my side.

pwilkin:
Aight, should work now :)

pwilkin:
CI is looking good :)

fl0rianr (approved):
LGTM now, thanks!

pwilkin:
Cleaned it up post OpenMOSS merge.

fl0rianr:
@pwilkin Great feature your adding here and thanks for the hard work! Would you mind going into a more close discussion with me or/and @kpoineal (maybe on Discord) if you plan more features to add impacting Frontend (UI) code. Especially this one - it's not a small one -  and we have to integrate this inside of GUI3 we are currently working on. That said, didn't wanted to hold you back regarding that one since many users might want this. But maybe we can orchestrate this as we get closer and closer with merging GUI3. What are your 2 cents? I definitely don't want to hold you back in any way.

Issue #2567, opened 2026-07-06
GUI3: Presets auto-optimizer panel missing run and delete actions


kpoineal:
**Reported by:** Jeremy (project owner)

The auto-optimizer left panel on the Presets page doesn't have any way to run the auto-optimizer or delete old results. It's confusing — the panel exists but isn't actionable.

**Expected:** Clear actions to run the optimizer and manage/delete old results.
**Actual:** Panel displays but has no run or delete functionality.

fl0rianr:
Still only a mock-up and not properly connected with lemon bench and an auto opt process. A minimal version e.g. to optimize ideal MTP configuration for the current machine would make sense.

fl0rianr:
Closed in favor of the follow-up: https://github.com/lemonade-sdk/lemonade/issues/2613

Issue #2562, opened 2026-07-06
GUI3: Dashboard ignores Connect page port setting without browser refresh


kpoineal:
**Reported by:** Jeremy (project owner)

When changing the port on the Connect page, the Dashboard doesn't pick up the new port. A full browser refresh is required for the Dashboard to use the updated setting.

**Expected:** Port changes on Connect propagate immediately to Dashboard (and all other views).
**Actual:** Dashboard uses stale port until browser refresh.

kpoineal:
Fixed in #2586

Issue #2561, opened 2026-07-06
GUI3: UDP beacon discovery not finding running lemond


kpoineal:
**Reported by:** Jeremy (project owner)

GUI3 doesn't discover a running lemond instance via UDP beacon. When lemond is running on port 13399, GUI3 doesn't find it and doesn't auto-connect.

**Expected:** GUI3 discovers lemond via UDP beacon the same way CLI and other clients do.
**Actual:** GUI3 doesn't detect the running server.

kpoineal:
Partially addressed in #2586 (auto-connect when served by lemond + apiUrl fallback). True UDP beacon is not possible from browser context.

PR #2586, opened 2026-07-07
fix(gui3): resolve 4 bugs from owner feedback

### LINKED ISSUE #2560: GUI3: Models page shows suggested:false models
**Reported by:** Jeremy (project owner)

The Models page is displaying models that have `suggested: false` in the registry. These should not be shown to users by default.

**Expected:** Only models with `suggested: true` (or unset) appear in the model list.
**Actual:** All models including `suggested: false` ones are visible.

kpoineal:
Fixes 4 bugs reported by Jeremy:

- **Auto-discovery (#2561):** Derive default base URL from \window.location.origin\ when served by lemond; also read mock API's \piUrl\ field as fallback.
- **Dashboard port (#2562):** Listen for connection status changes and resume polling when connection is re-established after failures.
- **Models filtering (#2560):** Hide models with \suggested:false\ unless downloaded or loaded.
- **AutoOpt rail (#2567):** Add Run Optimizer button and per-run Delete action.

Fixes #2560, #2561, #2562, #2567

### DIFF
--- a/prototype/ui-redesign/src/api.ts
+++ b/prototype/ui-redesign/src/api.ts
@@ -7,7 +7,20 @@
 import { recipeOptionsForModel, samplingForModel, type RecipeOptions } from './presetStore';
 import { COLLECTION_IMAGE_SIZE } from './features/collections/collectionImageConfig';
 
-const DEFAULT_BASE_URL = 'http://localhost:13305';
+function detectDefaultBaseUrl(): string {
+  if (typeof window !== 'undefined' && window.location) {
+    // When served by lemond at /app, window.location.origin IS the API server.
+    // Detect this by checking for the lemond-injected window.api shim or /app path.
+    const servedByLemond = window.location.pathname.startsWith('/app')
+      || (window as any).api?.getServerPort;
+    if (servedByLemond && window.location.port) {
+      return `${window.location.protocol}//${window.location.hostname}:${window.location.port}`;
+    }
+  }
+  return 'http://localhost:13305';
+}
+
+const DEFAULT_BASE_URL = detectDefaultBaseUrl();
 const LS_BASE_URL = 'lemonade_base_url';
 const LS_API_KEY = 'lemonade_api_key';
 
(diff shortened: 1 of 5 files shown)


fl0rianr (changes requested):
Thanks, this fixes parts of the reported GUI3 issues, but I would adjust a few things before merging:

1. `detectDefaultBaseUrl()` currently trusts any `window.location.port`. In the prototype dev setup, `webpack serve` runs on port 8080 and there is no `/api` proxy, so the default becomes `http://localhost:8080` and API calls would hit the UI dev server instead of lemond. This should only use same-origin when we know the UI is actually served by lemond, or it should fall back to the known Lemonade API default / host-provided `apiUrl`.

2. The PR claims to fix UDP beacon discovery (#2561), but this does not implement beacon discovery. It only helps the “served by lemond” case and the mock `apiUrl` case. Either the issue/PR wording should be narrowed, or the actual discovery path needs to be implemented.

3. The AutoOpt run action is fine by me but not enough to close the issue, yet - or open another one for a least one useful autorun scenario to perform (to show of capability)

4. The per-run delete button is only visible on hover. That is weak for mobile/touch and keyboard accessibility. Please also reveal it on `:focus-within`, or make it persistently visible / move it into a compact action menu.

The model filtering and Dashboard reconnect polling changes look reasonable.

kpoineal (commented):
Thanks @fl0rianr — addressed points 1 and 4:

1. **Fixed**: \detectDefaultBaseUrl()\ now only uses same-origin when it detects the lemond shim (\window.api.getServerPort\) or the \/app\ path prefix. Dev servers on arbitrary ports fall through to the 13305 default.

2. **Acknowledged**: You're right — this doesn't implement real UDP beacon. That's not possible from a browser context. Narrowing #2561 to what we actually fixed (auto-connect when served by lemond + apiUrl fallback). True beacon discovery would need a server-side endpoint.

3. **Agreed**: The AutoOpt action is a UI scaffold. I'll open a follow-up issue for implementing a real optimization run (e.g. benchmark current backend params → suggest improvements).

4. **Fixed**: Delete button now also reveals on \:focus-within\ and direct \:focus\, making it keyboard/touch accessible.

fl0rianr (approved):
LGTM now

fl0rianr:
> 3\. **Agreed**: The AutoOpt action is a UI scaffold. I'll open a follow-up issue for implementing a real optimization run (e.g. benchmark current backend params → suggest improvements).

https://github.com/lemonade-sdk/lemonade/issues/2613 I created one.

PR #2588, opened 2026-07-07
feat(vllm): stage AMD Instinct MI300X (gfx942) support (pending release asset)

ianbmacdonald:
Extends the experimental vLLM ROCm backend from RDNA to AMD Instinct CDNA (gfx942 / MI300X). gfx942 is **staged out of the installable matrix** until an official `vllm0.19.1-rocm7.13.0-gfx942` asset ships in `lemonade-sdk/vllm-rocm` (lemonade-sdk/vllm-rocm#25) — everything else lands now, so enabling it is a one-line flip once the asset exists.

**What changed**
- gfx942 added to the vllm descriptor's ROCm support row (currently gated out; see above).
- Per-arch release pinning: `vllm.rocm_arch_overrides` in `backend_versions.json`, resolved via the shared `rocm_asset_family()` — gfx942 rides its own vLLM/ROCm line (RDNA can't share the tag); RDNA falls through to the default pin, unchanged.
- Device-class launch defaults: discrete-HBM (Instinct) gets vLLM-native budgeting + CUDA-graph capture; APUs/consumer dGPUs keep the conservative defaults. `--enforce-eager` stays a user-overridable escape hatch.
- FP8 + MTP recipes for Qwen3.6-27B and 35B-A3B.

**Validation** — MI300X (gfx942; Hot Aisle VM, Ubuntu 24.04, kernel 6.8, ROCm 7.13 bundled): served vLLM end-to-end (Qwen2.5-0.5B 306–511 tok/s warm; FP8/MTP recipes 45–126 tok/s single-stream, MTP ~80% draft acceptance). RDNA fall-through validated on RX 7900 XT (gfx1100). `test_vllm_cdna_gating` covers gating, per-arch override, and fall-through. **Not tested:** Windows, other CDNA (gfx90a/gfx908), and the official gfx942 asset — which doesn't exist yet (the resolver is validated to construct the exact tag/URL against a community build).

The device-class launch predicate is an MVP that could fold into a #2235-style memory planner later — happy to reshape or hand the branch over; discussion in #2546.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2217,3 +2217,12 @@ if(EXISTS "${_NETWORK_UTILS_TEST_SRC}")
     target_link_libraries(test_network_utils PRIVATE lemonade-server-core)
     add_test(NAME NetworkUtilsTest COMMAND test_network_utils)
 endif()
+
+# vLLM CDNA gating: the vllm descriptor support row + asset-family resolution for
+# AMD Instinct (gfx942), alongside the existing RDNA families.
+set(_VLLM_CDNA_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_vllm_cdna_gating.cpp")
+if(EXISTS "${_VLLM_CDNA_TEST_SRC}")
+    add_executable(test_vllm_cdna_gating test/cpp/test_vllm_cdna_gating.cpp)
+    target_link_libraries(test_vllm_cdna_gating PRIVATE lemonade-server-core)
+    add_test(NAME VllmCdnaGatingTest COMMAND test_vllm_cdna_gating)
+endif()
--- a/docs/dev/backends-reference.md
+++ b/docs/dev/backends-reference.md
@@ -372,7 +372,7 @@ the generator instead. Prose outside the markers is preserved. -->
 |-------|-----------|--------|
 | `TRELLIS-3D` | 15.4 | 3d |
 
-#### `vllm` — vLLM ROCm (experimental) (7 models)
+#### `vllm` — vLLM ROCm (experimental) (11 models)
 
 | Model | Size (GB) | Labels |
 |-------|-----------|--------|
@@ -382,7 +382,11 @@ the generator instead. Prose outside the markers is preserved. -->
 | `Qwen3.5-4B-FP16-vLLM` | 9.34 | reasoning, hot, tool-calling |
 | `Qwen3.5-9B-FP16-vLLM` | 19.3 | reasoning, tool-calling |
 | `Qwen3.6-27B-FP16-vLLM` | 55.59 | reasoning, tool-calling, vision |
+| `Qwen3.6-27B-FP8-vLLM-highconc` | 27.8 | reasoning, tool-calling, vision, mtp |
+| `Qwen3.6-27B-FP8-vLLM-lowconc` | 27.8 | reasoning, tool-calling, vision, mtp |
 | `Qwen3.6-35B-A3B-FP16-vLLM` | 71.93 | reasoning, tool-calling, vision |
+| `Qwen3.6-35B-A3B-FP8-vLLM-highconc` | 36.0 | reasoning, tool-calling, vision, mtp |
+| `Qwen3.6-35B-A3B-FP8-vLLM-lowconc` | 36.0 | reasoning, tool-calling, vision, mtp |
 
 #### `whispercpp` — Whisper.cpp (6 models)
 
--- a/docs/guide/configuration/vllm.md
+++ b/docs/guide/configuration/vllm.md
@@ -5,13 +5,13 @@ Lemonade integrates [vLLM](https://github.com/vllm-project/vllm) as an experimen
 1. **Day-0 model support.** vLLM typically supports new transformer architectures within hours of their release on Hugging Face — checkpoints load directly, with no per-architecture porting.
 2. **Concurrency and multi-GPU.** Paged-attention KV cache, continuous batching, and chunked prefill scale aggregate throughput with in-flight request count; tensor and pipeline parallelism are supported across multiple GPUs.
 
-> **Status: experimental.** The backend has been validated on **gfx1151 (Strix Halo)** and **gfx1150 (Strix Point)**. Prebuilt wheels also exist for `gfx110X` (RDNA3) and `gfx120X` (RDNA4) but those targets have not been exercised end-to-end yet.
+> **Status: experimental.** The backend has been validated on **gfx1151 (Strix Halo)** and **gfx1150 (Strix Point)**. Prebuilt wheels also exist for `gfx110X` (RDNA3) and `gfx120X` (RDNA4) but those targets have not been exercised end-to-end yet. **gfx942 (AMD Instinct MI300X, CDNA3)** is **staged, not auto-installable yet** — the resolver, per-architecture pinning, launch policy, and recipes are all in place and have been manually validated on real MI300X hardware, but gfx942 is held out of the installable support matrix until the official `-gfx942` release asset is published (see [Deploying on MI300X](#deploying-on-mi300x-gfx942--quickstart)).
 
 ## Available Backend
 
 ### ROCm
 - **Platform**: Linux only
-- **Hardware**: validated on gfx1151 (Strix Halo) and gfx1150 (Strix Point); prebuilt wheels also exist for gfx110X (RDNA3) and gfx120X (RDNA4)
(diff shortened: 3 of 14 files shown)


fl0rianr (changes requested):
Thanks for the detailed write-up and the thorough MI300X validation — the direction looks good to me, especially keeping the per-arch release pinning separate from the launch-policy helper.

I think there are a couple of things we should resolve before merging:

1. **Official gfx942 asset / CI gate**

   The PR adds gfx942 to the public vllm:rocm support matrix, so Lemonade now treats MI300X as an installable supported target. The resolver then constructs:

   `vllm0.19.1-rocm7.13.0-gfx942-x64.tar.gz`

   from lemonade-sdk/vllm-rocm.

   As far as I can see, that asset is not published in the official `lemonade-sdk/vllm-rocm` releases yet, and CI is already catching this in the GFX-topology backend-link test. So I think we should either publish the official gfx942 asset first, or keep gfx942 gated/out of the public installable support matrix until the asset is live.

2. **`vllm.rocm_bin=latest` and per-arch releases**

   Since vLLM ROCm releases are per GPU target, `latest` being repo-wide looks risky. If `latest` resolves to a suffixed RDNA/APU tag like `...-gfx1151` or `...-gfx120X`, `get_install_params()` now treats that as a complete tag and will not re-suffix it for the detected target. On an MI300X host that could install or report status against the wrong architecture line.

   Could we either resolve `latest` per target suffix, or reject a suffixed pin when its target does not match the detected arch?

3. **MI300X quickstart wording**

   The docs say users can temporarily pin the community-built tarball with:

   `lemonade config set vllm.rocm_bin=vllm0.19.1-rocm7.13.0-gfx942`

   But that only changes the tag/version, not the repo. The install path still points to `lemonade-sdk/vllm-rocm`, so this will still fail until the official asset exists there. I think the docs should either describe a real manual sideload path, or simply say that the official asset is a prerequisite for the built-in install path.

Smaller follow-up concern: on hybrid AMD systems, `get_rocm_arch()` appears to use the first available AMD GPU from the detected device list. Since the device list is populated with iGPU before dGPU, a system with both Strix Halo and MI300X could potentially resolve `gfx1151` instead of `gfx942`, which would select the wrong asset line and conservative launch defaults. Not necessarily a blocker for MI300X cloud VMs, but worth making explicit or hardening if this code is meant to cover mixed systems.

Once the official asset / gating and the `latest` behavior are handled, I think the overall structure is reasonable: the device-class launch policy is nicely isolated and should be easy to move into a future planner.

ianbmacdonald:
Thanks for the thorough review — really helpful. Pushed changes for all four points; the asset gate (1) is handled by staging gfx942 out for now, with a clear path to flip it on.

**2. `latest` / mismatched-arch pin** — Fixed. `get_install_params()` now *rejects* a `vllm.rocm_bin` pin whose `-gfx…` suffix doesn't match the detected arch (the error names both), instead of using it verbatim. So a repo-wide `latest` that resolved to an RDNA/APU tag, or an explicit cross-arch tag, can no longer install/report against the wrong line on an MI300X host. Matching-arch and bare pins are unchanged; the gating test flips the cross-arch case to rejected.

**3. MI300X quickstart docs** — Fixed. Corrected to state a `rocm_bin` pin changes only the tag, not the source repo; the docs now describe the staged state and the manual sideload path rather than implying a working built-in install.

**4. Hybrid AMD systems** — Hardened. `get_rocm_arch()` now prefers a discrete GPU over the integrated one (the device list is iGPU-first), so a Strix Halo + MI300X box resolves gfx942, not gfx1151. Extracted into a unit-tested helper (hybrid / iGPU-only / dGPU-only / unavailable-dGPU).

**1. Official gfx942 asset / CI gate** — Went with your second option: gfx942 is now **staged out of the public installable matrix** (one-line descriptor change), so CI is green and this merges as-is. All the infrastructure still lands and stays tested — the per-arch resolver, cross-arch reject, device-class launch policy, discrete-GPU preference, and the FP8 + MTP recipes. Re-enabling gfx942 is then a **one-line matrix flip** the moment the official asset exists.

On that asset — @ramkrishna2910, this is really a `lemonade-sdk/vllm-rocm` question for you. The pieces are all in place:
- A community build is already published and hardware-validated on a real MI300X: [`ianbmacdonald/vllm-rocm-cdna` release `vllm0.19.1-rocm7.13.0-gfx942`](https://github.com/ianbmacdonald/vllm-rocm-cdna/releases/tag/vllm0.19.1-rocm7.13.0-gfx942) — the same tag the resolver targets.
- The build itself is the same per-arch flow you already run for RDNA, plus gfx942 → `gfx94X-dcgpu` in the matrix and `amd-quark==0.11.2` pinned (the MoE FP8 path needs it).

The one piece worth doing deliberately is **baking the AITER JIT cache into the tarball**. AITER's fused-expert MoE + FA kernels JIT-compile on first use, and on the *minimal relocatable* bundle that first run is **20–40 minutes** — the bundle ships without the ROCm dev headers AITER's `clang++ -x hip` needs. Since the vllm-rocm build already runs in AMD's full ROCm SDK (headers present, JIT compiles cleanly), the fix is to run one AITER-on warmup per arch there, capture `site-packages/aiter/jit/build/`, and ship that cache in the minimal tarball — users then *load* the kernels with **zero first-run JIT wait**. The payoff is real: on the box we measured AITER-on at **~2× throughput at low/mid concurrency, ceiling ~5300 tok/s output @ c=256** (fused-MoE FP8 active) vs the AITER-off baseline.

One caveat we hit and built a guard for: AITER keys its own JIT cache on *source-hash + arch only, not the ROCm runtime version*, so a pre-baked cache will silently load a **stale** `.so` across a runtime/OS bump (→ intermittent MoE corruption, not a clean failure). The guard splits naturally across the two repos: the vllm-rocm build writes a small **fingerprint manifest** (SHA256 of the actually-linked ROCm/torch `.so`s + arch + py-abi) next to the baked cache, and **lemonade recomputes it at boot and fails closed** — a mismatch sets `VLLM_ROCM_USE_AITER=0` with a loud log (never a crash, never a silent mis-run), with a manual force-rebuild escape when dev headers are present. We've also specced the test for it: deploy the tarball + cache, do an in-place OS upgrade, and assert the fingerprint mismatch trips the clean AITER-off fallback rather than gibberish.

The full runbook for all of this is published right alongside the tarball on that release — an AITER-enablement runbook, an MI300X MVP runbook, and the recipe JSONs — covering the three minimal-tarball blockers + fixes (`ROCM_HOME` `.pth`, amdgcn symlink, dev-headers via `CPLUS_INCLUDE_PATH`), the warmed-cache bake, and the fingerprint/fail-closed guard. So anyone can reproduce it today, including on Ubuntu 26.04. Happy to open the one-line flip-on PR here the instant the official `gfx94X-dcgpu` release lands — whatever fits the release flow best.

fl0rianr (commented):
Thanks for the quick follow-up — this addresses the main concerns nicely. In particular, staging **gfx942** out of the public installable matrix until the official asset exists is the right shape, and the cross-arch pin rejection + discrete-GPU preference look much safer now.

I still see two things before I would be comfortable approving:

1. **CI is still red**
   The new head currently has `Validate New vLLM Release` failing in the Linux build step. I could not see the exact compiler error from the summarized log, but I think we should wait for that to be green before merging.

2. **Docs still read as if gfx942 is supported/installable**
   The descriptor now correctly stages gfx942 out of the installable support matrix, but parts of the vLLM docs / FAQ still say that MI300X/gfx942 is supported via the per-arch override. That feels a bit inconsistent with the staged state. I’d suggest wording it as “staged / manually validated / not auto-installable until the official release asset ships” everywhere, especially in the FAQ.

Small optional cleanup: `speculative_config` is still dumped without checking that it is a JSON object. Since this is internal config it is probably not a blocker, but a clear type check would make future config mistakes easier to diagnose.

Once CI is green and the docs consistently say “staged, not auto-installable yet,” this looks good to me.

ianbmacdonald:
> 2. **Docs still read as if gfx942 is supported/installable**

Fair enough.. it is installable using the tarball provided.. and the runbook is supported once it merges .. but I get it. Keep the issues to a minimum, the comments here are just a click or two away from the llamacpp PR.  Let me tune it up.

fl0rianr (approved):
Thanks for the follow-up updates — this looks good to me now.

ianbmacdonald:
For the official gfx942 asset this is staged against: filed lemonade-sdk/vllm-rocm#25 with what it takes to generate the CDNA build in vllm-rocm (CPU build now + an ephemeral-MI300X qualify/AITER-bake at release). Once that asset ships, re-enabling gfx942 here is the one-line matrix flip.

PR #2594, opened 2026-07-08
refactor(audio-gen): apply backend review feedback to ThinkSound and ACE-Step

pwilkin:
Applies the review feedback from #2580 (superm1) to the already-merged audio-generation backends, matching the changes made on the OpenMOSS (#2580) and Trellis (#2581) PRs:

- Split the combined ROCm/CUDA env setup into explicit per-backend branches
- Raise `UnsupportedOperationException` instead of silently defaulting to vulkan when no supported backend is detected
- Unnamed unused parameters instead of `(void)` casts

### DIFF
--- a/src/cpp/server/backends/acestep/acestep_server.cpp
+++ b/src/cpp/server/backends/acestep/acestep_server.cpp
@@ -5,6 +5,7 @@
 #include "lemon/backends/backend_utils.h"
 #include "lemon/backends/hf_cache_util.h"
 #include "lemon/backend_manager.h"
+#include "lemon/error_types.h"
 #include "lemon/model_manager.h"
 #include "lemon/runtime_config.h"
 #include "lemon/system_info.h"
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Thanks for going after this.

Small wording / intent clarification: this change looks consistent with the existing backend-selection flow to me. It removes the unconditional fallback to vulkan when `get_supported_backends()` returns no candidates, while still allowing vulkan when system-info considers it supported.

One small nuance: Lemonade’s Vulkan support rows intentionally include cpu/x86_64 in some recipes as a proxy for generic Vulkan-capable systems / iGPU paths, not necessarily as a pure CPU backend. sd-cpp uses the same pattern, while also having a dedicated cpu backend.

So I think the code is fine, but the PR description / error text might be slightly clearer as “no supported backend detected” rather than “no supported GPU backend detected”. Otherwise it can sound like Vulkan should only count when an explicit AMD/NVIDIA GPU is detected, which is not quite how Lemonade models Vulkan today.

pwilkin:
Aight, changed it :)

fl0rianr (approved):
Was a bit nitty but I thought letting you know this had it's worth on its own. Approving.

PR #2472, opened 2026-06-28
Add lemon-mlx-engine as a wrapped-server backend (rebase of #2010)

bong-water-water-bong:
## Summary

Rebase of #2010 (which replaced #1717) onto current main. Wires [lemon-mlx-engine](https://github.com/lemonade-sdk/lemon-mlx-engine) as the `lemon-mlx` wrapped-server recipe.

The upstream engine exposes an OpenAI-compatible HTTP API, so this stays within the existing `WrappedServer` abstraction.

## Changes vs #2010

- Rebased onto main (2026-06-28) — 11 merge conflicts resolved
- `forward_streaming_request` updated for new `TelemetryCallback` parameter
- `populate_static_max_context_window` → `populate_model_metadata` rename
- Backend pins bumped: b1034-stable → **b1045-stable** (latest release, 2026-06-28)
- Kept moonshine backend alongside lemon-mlx (both coexist)

## Core backend

- `MlxServer` (`src/cpp/{include,server}/lemon/backends/mlx_server.{h,cpp}`) — `/v1/chat/completions`, `/v1/completions`, streaming; `responses()` returns `UnsupportedOperationException`
- `router.cpp`: instantiates `MlxServer` for `lemon-mlx` recipes
- `model_types.h`, `runtime_config.cpp`: registers `lemon-mlx` as a selectable backend

## Resolved review items (from #2010 @fl0rianr, @bitgamma, Qodo)

| Item | Fix |
|------|-----|
| BLOCKING: `completions_streaming` True but no streaming `/v1/completions` | `completions_streaming = False` in capabilities.py |
| BLOCKING: `forward_streaming_request` unsafe for `/v1/completions` | Early throw guard in `MlxServer::forward_streaming_request` |
| P2: CPU support matrix overstated | Split `system_info.cpp`: Linux → x86_64 only; macOS → arm64 only |
| P2: Unreachable `*-preview` pins | Removed |
| P2: README CPU row ambiguous | Explicit: "x86_64 on Linux; Apple Silicon (arm64) on macOS" |
| P2: Custom `tokenize_quoted_args` | Replaced with `std::istringstream` |

## Performance (b1045-stable, gfx1151)

| Model | tok/s |
|-------|------:|
| Qwen3.6-35B-A3B-MTP q4 | **63** decode TPS |
| gfx1201 (R9700) | **79** decode TPS (+15% graph replay) |

Closes #1642

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2598, opened 2026-07-08
fix: normalize gfx arch tokens to family wildcards in identify_rocm_arch_from_name (#2319)

### LINKED ISSUE #2319: ROCm unavailable for gfx103X, gfx110X and gfx120X

bong-water-water-bong:
## What

Map specific gfx arch tokens to their family wildcards where ROCm backend assets are published under the family name instead of per-arch. Specific archs that have their own published assets (gfx1150, gfx1151, gfx1152, gfx942, gfx90a, etc.) remain unchanged.

## Why

`identify_rocm_arch_from_name` returns the exact gfx token (e.g. `gfx1201`, `gfx1102`) but ROCm backend assets are published under family wildcards (`gfx120X`, `gfx110X`). This mismatch means:
- Backend asset URLs resolve incorrectly
- Support matrix checks fail even for supported GPUs
- Users get "unsupported GPU" errors on hardware that works fine

## How

Adds a `gfx_to_family_wildcard()` helper that maps:
- RDNA4: `gfx1200`/`gfx1201` → `gfx120X`
- RDNA3: `gfx1100`–`gfx1103` → `gfx110X`  
- RDNA2: `gfx1030`–`gfx1036` → `gfx103X`
- All others: passed through unchanged

Called from both the regex match path and the KFD ISA path in `identify_rocm_arch_from_name`.

## Tested

- [ ] RDNA3 GPU (gfx1102 → gfx110X resolution)
- [ ] Non-matching archs pass through unchanged

Fixes #2319


bong-water-water-bong:
Split into two PRs:
- **gfx arch → family wildcards** → #2619 (single commit, +23/−2 in system_info.cpp)
- **MLX backend integration** → already at #2472 (no change needed)

PR #2598 was combining both. #2619 is the clean gfx arch fix rebased onto latest main.

PR #2601, opened 2026-07-08
fix(backends): enable sd-cpp CUDA support on Windows

kenvandine:
## What

sd-cpp's `cuda` variant was restricted to Linux-only in the backend descriptor, even though the Windows CUDA download/runtime path in `sdcpp_server.cpp` is already implemented and the upstream `stable-diffusion.cpp` release still publishes a `windows-cuda` asset (validated by `validate_sdcpp.yml`). llama.cpp's descriptor already lists `cuda` for both Windows and Linux, so this brings sd-cpp in line.

- `src/cpp/include/lemon/backends/sdcpp/sdcpp.h` — add `"windows"` back to the cuda `support` row; add `cuda` to `arg_variants`/`bin_variants` so `config.json` gets `cuda_args`/`cuda_bin` defaults.
- Regenerated `src/cpp/resources/defaults.json` and docs (`README.md`, `docs/dev/backends-reference.md`, `docs/guide/configuration/README.md`) via `docs/tools/gen_backend_boilerplate.py`.

## Note

Windows CUDA support for sd-cpp was deliberately disabled in #2174 ("sd-cpp's CUDA backend is not supported on Windows"), with no linked issue explaining the underlying failure. This PR re-enables it since the asset and download plumbing exist and are still CI-validated for asset presence. 

## Test plan

- [x] Verify sd-cpp cuda now appears as an available backend option in `lemonade backends` / system-info on Windows
- [x] Install and load an sd-cpp model with `--sdcpp cuda` on a Windows machine with an NVIDIA GPU (Turing or newer) and confirm image generation actually works (not just that the binary launches)
- [x] Confirm Linux sd-cpp cuda behavior is unchanged

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ Lemonade supports multiple inference engines for LLM, speech, TTS, and image gen
     <tr>
       <td><code>cuda</code></td>
       <td>NVIDIA GPUs (Turing or newer)**</td>
-      <td>Linux</td>
+      <td>Windows, Linux</td>
     </tr>
     <tr>
       <td><code>vulkan</code></td>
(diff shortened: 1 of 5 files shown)


fl0rianr (approved):
Nice!

PR #2603, opened 2026-07-08
fix(rocm): detect version-suffixed HIP runtime on Windows (ROCm 7.x)

ramkrishna2910:
Follow-up to #2508. Windows system-ROCm detection probes for `bin\amdhip64.dll`, but ROCm 7.x version-suffixes the HIP runtime to `bin\amdhip64_7.dll`. So a system ROCm 7.x install (via `HIP_PATH`/`ROCM_PATH`) went undetected and TheRock was downloaded anyway — the feature no-ops on current ROCm.

Fix: match the `amdhip64*.dll` prefix under `bin`/`lib` (still covers plain `amdhip64.dll` for 5.x/6.x).

Verified on a real ROCm 7.1 Windows host: `resolve_rocm_root` now returns `C:\Program Files\AMD\ROCm\7.1\` where it previously returned `nullopt`. Added a regression test for the version-suffixed runtime, and fixed a latent throwing `fs::equivalent` in the test that only triggered once a host root actually resolves.

### DIFF
--- a/docs/guide/configuration/llamacpp.md
+++ b/docs/guide/configuration/llamacpp.md
@@ -120,7 +120,7 @@ lemonade backends install llamacpp:rocm
 
 ### Reusing a System-Installed ROCm (Windows and Linux)
 
-On the stable channel Lemonade normally downloads its own ROCm runtime (TheRock). If you already have ROCm installed system-wide, Lemonade reuses it instead of downloading a second copy when it can find a matching version. It locates the install root in this order, using the first directory that contains the HIP runtime (`bin\amdhip64.dll` on Windows, `lib{,64}/libamdhip64.so` on Linux):
+On the stable channel Lemonade normally downloads its own ROCm runtime (TheRock). If you already have ROCm installed system-wide, Lemonade reuses it instead of downloading a second copy when it can find a matching version. It locates the install root in this order, using the first directory that contains the HIP runtime (`bin\amdhip64.dll` or `bin\amdhip64_<version>.dll` on Windows, `lib{,64}/libamdhip64.so` on Linux):
 
 1. The `ROCM_PATH` environment variable
 2. `rocm-sdk path --root`, when `rocm-sdk` is on your `PATH` (e.g. a ROCm installed from the TheRock pip wheels)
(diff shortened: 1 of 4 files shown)


fl0rianr (commented):
Looks good overall — the change is narrowly scoped and fixes the ROCm 7.x Windows detection gap without affecting the non-Windows path.

Small nits only:
- Please update the docs/header wording from exact `amdhip64.dll` to the new accepted Windows runtime pattern (`amdhip64*.dll` or `amdhip64.dll` / `amdhip64_*.dll`).
- Consider tightening the prefix match slightly so unrelated names like `amdhip64_backup.dll` are not accepted, and optionally require a regular file.
- The new regression test should also assert `explicit_source == true`, matching the existing ROCM_PATH tests.

No blocker from my side.

ramkrishna2910:
Thanks @fl0rianr — all three addressed in 1cea3f775:
- Match tightened to `amdhip64.dll` or `amdhip64_<digits>.dll` (rejects `amdhip64_backup.dll`) + requires a regular file.
- Header and `docs/guide/configuration/llamacpp.md` wording updated to the new pattern.
- Versioned test now asserts `explicit_source == true`; added a negative case for the bogus suffix.

All 16 checks pass on a real ROCm 7.1 Windows host.

fl0rianr (commented):
Thanks for the good follow-up — the implementation looks much better now.

One remaining issue: it looks like the new regression test is registered in CMake, but not actually run by the current CI test steps. The Linux/macOS C++ unit-test workflow builds/runs an explicit test list that does not include `est_rocm_root_resolution/ RocmRootResolutionTest, and the Windows installer job builds `wix_installers` but does not run this test.

Since the important new coverage is under _WIN32, could we add a Windows CI step to build and run `test_rocm_root_resolution`? For example, after configure/setup:

```
cmake --build build --config Release --target test_rocm_root_resolution
build\Release\test_rocm_root_resolution.exe
```

That would make the ROCm 7.x detection regression actually covered where it matters.

kenvandine (approved):
This looks reasonable to me, but I don't have a way to test it myself.

fl0rianr (approved):
The test is fine as-is if it is the plan/intend to keep this test manual. 

If that is the target I'm approving since I made my point and the rest is good.

Issue #2596, opened 2026-07-08
MCP Client Host in Lemonade Desktop GUI


HyphenifyLimited:
### Feature Description

Add MCP client functionality to the Lemonade Desktop GUI so users can connect to external MCP servers and let the LLM use their tools directly in chat conversations.
1. **MCP server configuration UI** — Add/remove external MCP servers in Lemonade GUI settings (similar to how Claude Desktop handles MCP config)
2. **Tool discovery + registration** — Auto-discover tools from connected MCP servers and expose them to the LLM
3. **Chat with MCP tools** — Allow LLM to call external tools during chat, with tool results fed back into context
4. **Per-chat tool selection** — Option to enable/disable which MCP servers are active in a given conversation

### Use Case / Motivation

I work with **Autodesk Revit** (architectural BIM software) via the RevitMCPBridge — an MCP server that exposes Revit model operations (query views, list elements, create rooms, place annotations) as MCP tools.

Currently my workflow is split:
- **LM Studio** has a built-in MCP client — I can chat with the model and it calls Revit MCP tools directly
- **Lemonade** provides superior NPU+GPU hybrid performance, but I can't use its GUI for Revit tasks

I want the unified experience: **Lemonade Desktop GUI with an MCP client** that lets me configure external MCP servers (stdio or HTTP) and have the LLM call their tools during chat.

### Platform Relevance

All platforms

## Why This Matters
Lemonade is already positioned as a local AI hub with MCP *server* support (exposing its models via POST /mcp). Adding MCP *client* support makes it a complete AI workstation:

- **MCP Server** (already in v10.8): Other apps call Lemonade's models
- **MCP Client** (requested): Lemonade calls external tools / services

This closes the loop and makes Lemonade Desktop a self-contained AI assistant for any domain with MCP tooling — coding, design, engineering, data analysis, etc.

## Implementation Notes
- The GAIA project already has an MCP Client Mixin in development. Lemonade could leverage similar patterns or collaborate on a shared implementation.
- GUI3 appears to be the right milestone for this feature.

## Hardware
- AMD Ryzen AI MAX+ 395 (Strix Halo)
- Radeon 8060S GPU
- XDNA 2 NPU
- Using Lemonade v10.9.0 with NPU+GPU hybrid and Vulkan GGUF inference

## Vote
👍 if you'd find MCP client support useful in Lemonade Desktop GUI.

PR #2605, opened 2026-07-08
Add server-side MCP client host foundation

fl0rianr:
## Summary

Partly addresses #2596.

Adds the server-side foundation for an MCP client host in Lemonade. This introduces internal/admin endpoints for configuring stdio MCP servers, connecting/disconnecting them, discovering available tools, and calling tools through JSON-RPC.

This PR intentionally keeps the GUI integration and automatic chat-loop tool orchestration out of scope. Those can build on the `/internal/mcp/*` endpoints in a follow-up.

## Scope

* Add McpClientManager for server-side MCP client hosting.
* Support stdio MCP server configs persisted under the Lemonade cache directory.
* Add `/internal/mcp/servers` and `/internal/mcp/tools` endpoints.
* Add connect, disconnect, refresh-tools, remove-server, and raw tools/call endpoints.
* Expose discovered MCP tools in an OpenAI-compatible tool metadata shape for future chat integration.
* Add a mock stdio MCP server fixture.
* Add a C++ test covering config parsing, env masking, stdio initialize, tool discovery, and tools/call roundtrip.
* Wire the new C++ test into CI.

## Out of Scope

* GUI settings UI for adding/removing MCP servers.
* Per-chat MCP server/tool selection.
* Automatic model tool-calling orchestration in chat completions.
* HTTP/SSE MCP transport support.

MCP env values are not persisted as raw secrets. Persisted configs store ${VAR}
references; users should provide secrets through the process environment.

## Testing

* `cmake --build --preset default --target test_mcp_client_config`
* `ctest --test-dir build -R mcp_client_config --output-on-failure`
* `cmake --build --preset default --target lemond`

The local MCP client test passes and now exercises the stdio mock server path

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -178,6 +178,19 @@ jobs:
           if ($failures.Count -gt 0) { exit 1 }
           Write-Host "Build and packaging successful!" -ForegroundColor Green
 
+      - name: Run MCP client unit test on Windows
+        shell: PowerShell
+        run: |
+          $ErrorActionPreference = "Stop"
+
+          # Exercise the Windows-only process, quoting, environment-block, and
+          # pipe/handle paths that Linux and macOS cannot cover.
+          cmake --build build --config Release --target test_mcp_client_config
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+
+          ctest --test-dir build -C Release --output-on-failure -R "^mcp_client_config$"
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+
       - name: Upload Lemonade Server Installers
         id: upload-unsigned-msi
         uses: actions/upload-artifact@v7
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -287,8 +287,9 @@ jobs:
           cmake --build --preset default --target test_model_manager_collection_validation
           cmake --build --preset default --target test_routing_classifier_services
           cmake --build --preset default --target test_network_utils
+          cmake --build --preset default --target test_mcp_client_config
           cd build
-          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest"
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest|mcp_client_config"
 
       - name: Upload .deb package
         uses: actions/upload-artifact@v7
@@ -598,8 +599,9 @@ jobs:
           cmake --build --preset default --target test_model_manager_collection_validation
           cmake --build --preset default --target test_routing_classifier_services
           cmake --build --preset default --target test_network_utils
+          cmake --build --preset default --target test_mcp_client_config
           cd build
-          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest"
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest|mcp_client_config"
 
       - name: Upload .pkg package
         if: steps.check_signing.outputs.has_signing == 'true'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -655,6 +655,7 @@ set(SOURCES_CORE
     src/cpp/server/backend_manager.cpp
     src/cpp/server/ollama_api.cpp
     src/cpp/server/anthropic_api.cpp
+    src/cpp/server/mcp_client.cpp
     src/cpp/server/mcp_server.cpp
     src/cpp/server/streaming_audio_buffer.cpp
     src/cpp/server/vad.cpp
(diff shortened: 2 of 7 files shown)


fl0rianr:
@claude review

fl0rianr:
@claude review

fl0rianr:
Review addressed, please @claude review again

ramkrishna2910:
## Review — server-side MCP client host foundation

Reviewed the full diff. The code is genuinely high quality — careful cross-platform process management, correct Windows argument quoting and environment-block construction, clean JSON-RPC request/waiter multiplexing, secret masking on persistence, and a real end-to-end test. My concerns are about **security posture** and **concurrency granularity**, not sloppiness.

### 🔴 Top concern — unauthenticated arbitrary local code execution + permissive CORS

These endpoints spawn an arbitrary `command`/`args`/`env`/`working_dir` — RCE by design, which is fine for an admin surface. The issue is default exposure:

- `/internal/*` is gated by `admin_api_key_` **only when it is non-empty**. By default (no `LEMONADE_API_KEY` / `LEMONADE_ADMIN_API_KEY` — the common local setup) it is empty, so `/internal/mcp/*` is **unauthenticated** (`server.cpp:621-629`).
- CORS is `Access-Control-Allow-Origin: *` with a blanket `OPTIONS .* → 204` preflight (`server.cpp:1347-1356`). A `POST` with `Content-Type: application/json` triggers a preflight the server approves, so **any website the user visits can drive these endpoints** against `localhost` and reach local code execution (register → connect → `tools/call`).

Mitigating factors: server binds localhost by default, and configs do not auto-connect at startup (attacker must also call `/connect`). This posture is *inherited* from the existing `/internal/*` surface — this PR does not create the hole — but it **escalates the capability** from "shutdown/config" to "spawn any process," and the planted config **persists to disk**.

Would prefer an explicit decision here before merge. Options, in rough order of preference:
1. Require `admin_api_key_` to be **set** for `/internal/mcp/*` specifically (fail closed if unset), or
2. Enforce a same-origin / `Origin` check (or drop CORS) for `/internal/*`, and/or
3. At minimum document loudly that enabling MCP requires `LEMONADE_ADMIN_API_KEY`.

### 🟠 Coarse `Runtime::mutex` blocks introspection and cancellation during a call

`Runtime::mutex` is held for the **entire** blocking RPC in `call_tool` / `connect` / `refresh_tools_locked` — up to the clamped timeout (default 30s, **max 300s**). Because `snapshot()` also takes that mutex:

- `GET /internal/mcp/servers` and `/internal/mcp/tools` **hang** while any one server is mid-call (`list_servers_json` snapshots every runtime) — a polling UI would stall.
- `disconnect()` cannot acquire the lock until the in-flight call finishes, so **a hung tool call cannot be aborted** via disconnect until it times out (up to 300s).

The waiter/multiplexing infrastructure already supports concurrent in-flight requests by id, so the coarse lock partly defeats it. Consider snapshotting state without holding the lock across the RPC wait, or releasing `Runtime::mutex` while blocked on the condition variable. Not a deadlock — lock ordering is clean (`mutex_` and `Runtime::mutex` are never nested; the reader thread only needs `waiters_mutex`) — but a real responsiveness/usability issue.

### 🟠 Windows process path is not exercised in CI

The most intricate, platform-specific code (`start_windows`, `build_windows_command_line`, `build_windows_environment_block`, handle plumbing) has **no CI coverage** — `test_mcp_client_config` is added only to the Linux `.deb` and macOS `.pkg` jobs (and the integration part is further gated on `Python3` being found). Recommend adding it to the Windows CI job too.

### 🟡 Minor

- **Literal secrets silently dropped after restart:** a raw env value is used in-memory but persisted as `${KEY}` (`config_to_persisted_json`); after restart it resolves from the environment, silently becoming empty if unset. Documented in the PR body, but consider rejecting/warning on literal values rather than silently rewriting.
- `execvp` uses PATH resolution for `command` — reasonable, worth being conscious of alongside the auth concern.
- `make_error(..., 499)` uses a non-standard status code for disconnect.
- Include-ordering nit: `<cctype>` after `<chrono>`/`<condition_variable>` in `mcp_client.cpp` (project style is alphabetical within a block).
- `kMaxToolListPages = 32` silently caps pagination without a signal.

### What's done well

- **Shell-free** spawning (execvp / CreateProcessW with proper argv quoting) — no shell-injection surface despite arbitrary commands.
- Lock hygiene: `mutex_` released before calling into `Runtime::snapshot()`, avoiding lock-order inversion.
- Timeout cleanup race in `request()` handled correctly (waiter erase vs. reader delivery — no leak/UAF; waiter is `shared_ptr`).
- Server-initiated requests get a `-32601` reply so a server won't hang; secrets masked (`***`) everywhere and never written raw to disk (explicitly tested).
- Solid integration test through the real stdio path (connect → discover → call → disconnect → remove), including a secret-not-persisted assertion.

### Verdict

Strong, well-engineered foundation. Not merge-ready as-is pending an explicit decision on the **unauthenticated `/internal` RCE + CORS** exposure (item 1) — a judgment call for maintainers more than a code fix, but it should be conscious. Would also want the **lock-during-call** behavior addressed (or acknowledged) and **Windows CI coverage** added. Everything else is polish.

fl0rianr:
@claude review

ramkrishna2910 (approved):
Re-reviewed after the latest 3 commits — all prior concerns are resolved and backed by regression tests. Approving.

**Verified fixed:**
- **Security (top concern):** `/internal/mcp*` is now fail-closed — 403 when no admin key is configured, including OPTIONS, so a browser preflight also fails closed. Closes the permissive-CORS → localhost RCE vector.
- **Coarse `Runtime::mutex`:** re-architected into `lifecycle`/`connect`/`state`/`waiters` locks. `call_tool`/`refresh_tools` capture the process handle under `state_mutex` briefly then run the blocking RPC without it, so `snapshot()`/`list_servers` no longer stall; `disconnect()` cancels in-flight waiters so a hung call is abortable. Lock ordering is consistent (lifecycle → state, lifecycle → waiters → waiter) — no inversion. Regression-tested (30s call keeps status < 1s, disconnect < 4s, call aborts).
- **Windows CI:** new job step builds and runs `test_mcp_client_config`, covering the Windows-only spawn/quoting/env-block path.
- **Env secrets:** raw values now rejected (must be `${VAR}`); never persisted.

**Second-pass security sweep of the reworked spawner — additionally confirmed well-handled:**
- No handle/fd leak to the child: Windows `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` (only the 3 stdio handles); POSIX pipes are `O_CLOEXEC`.
- Descendant cleanup: Windows Job Object (`KILL_ON_JOB_CLOSE`); POSIX process group + `kill(-pgid)`.
- 64 MiB per-message read cap (input DoS); shell-free `execve`/`CreateProcessW`.

**Local verification (Windows):** `test_mcp_client_config` passes via ctest (real stdio + concurrency/abort assertions); `lemond` rebuilds clean with the reworked source.

Remaining items are optional follow-ups within the admin trust boundary, none blocking: env references can read any server env var; no cap on number of MCP servers/processes; child stderr logged raw; a defensive `closefrom()` on POSIX would remove reliance on the socket lib setting CLOEXEC.

Nicely done addressing the review.

PR #2606, opened 2026-07-08
Router/2386 decision trace response

### LINKED ISSUE #2386: [Router] Expose decision trace on response (additive)
## Goal
**Expose the decision trace on the response** (additive). Core stays pure selection.

## Scope
- Attach the `Decision` (`route_to`, `matched_rule`, `trace`, pass-through `outputs`) to the chat response **additively** — a response field + optional header — so a vanilla OpenAI client ignoring it is unaffected.
- Streaming variant (chunk + header).

## Out of scope (moved to the Trust Router milestone)
- **Verdict / `warn` / `block` semantics, `finish_reason:"content_filter"`, synthesized refusals.** These are trust-customer behavior: "block" is expressed by a policy routing `route_to` a refusal/guard component; "warn" is a `verdict` field in the rule's `outputs` bag that the client reads. The core does not interpret them.

## Acceptance
- Non-stream + stream paths expose the trace; vanilla OpenAI client unaffected.
- A request whose policy routes to a refusal component returns that component's output normally (no special core path).

---
**Depends on:** #2385, #2376

eddierichter-amd:
## Summary

  Closes #2386.

  This PR exposes the router `Decision` from `collection.router` dispatch as additive response metadata. Routing already
  selected the concrete backend model in #2385; this change carries the full decision through the server layer and returns it to
  clients under `x_lemonade_route`.

  The decision includes:

  - `version`
  - `route_to`
  - `matched_rule`
  - `default_used`
  - `outputs`
  - optional `trace` when the request sets `route_trace: true`

  For routed streaming responses, the same decision metadata is injected into the first JSON SSE event, and the matched rule is
  also surfaced via the `x-lemonade-route` response header when a rule matched.

  ## Details

  - Adds `RouterDispatchResult` so server dispatch keeps both the selected model and the full `Decision`.
  - Attaches `x_lemonade_route` to non-streaming `/chat/completions`, `/completions`, and `/responses` responses.
  - Injects `x_lemonade_route` into routed SSE streams without changing non-router streaming behavior.
  - Preserves `outputs` as opaque pass-through data; the server does not interpret trust/customer-specific fields.
  - Keeps trace opt-in only via `route_trace: true`.
  - Strips `route_trace` before forwarding requests to the selected backend model.
  - Adds endpoint coverage for matched-rule decisions, default-route decisions, trace output, `outputs`, and streaming response metadata.

  ## Testing

  ```bash
  git diff --check

  PATH=/tmp/lemonade-cmake-venv/bin:$PATH cmake --build --preset default --target lemond

  PATH=/tmp/lemonade-cmake-venv/bin:$PATH ctest --test-dir build --output-on-failure -R "RoutingPolicy(Contract|Evaluator|
  Registry|Semantic|Deterministic|Engine|Parser|Store)Test|RoutingClassifierServicesTest|ModelManagerCollectionValidationTest"

  python3 -m py_compile test/server_endpoints.py

  Also ran focused endpoint tests against this branch’s built lemond on port 13306:

  LEMONADE_TEST_PORT=13306 PYTHONPATH=. /tmp/lemonade-cmake-venv/bin/python -m unittest \
    server_endpoints.EndpointTests.test_021z_router_collection_chat_dispatch \
    server_endpoints.EndpointTests.test_021zi_router_collection_trace_and_outputs \
    server_endpoints.EndpointTests.test_021za_router_collection_completions_dispatch \
    server_endpoints.EndpointTests.test_021zb_router_collection_responses_dispatch \
    server_endpoints.EndpointTests.test_021zc_router_collection_responses_streaming_dispatch

  Result:

  Ran 5 tests in 22.863s

  OK

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -638,6 +638,7 @@ set(SOURCES_CORE
     src/cpp/server/routing_policy.cpp
     src/cpp/server/routing_policy_parser.cpp
     src/cpp/server/routing_policy_store.cpp
+    src/cpp/server/route_decision_response.cpp
     src/cpp/server/runtime_config.cpp
     src/cpp/server/telemetry.cpp
     src/cpp/server/logging_config.cpp
@@ -2089,6 +2090,29 @@ if(EXISTS "${_ROUTING_STORE_TEST_SRC}")
 
     include(CTest)
     add_test(NAME RoutingPolicyStoreTest COMMAND test_routing_policy_store)
+endif()
+
+set(_ROUTE_DECISION_RESPONSE_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_route_decision_response.cpp"
+)
+if(EXISTS "${_ROUTE_DECISION_RESPONSE_TEST_SRC}")
+    add_executable(test_route_decision_response
+        test/cpp/test_route_decision_response.cpp
+        src/cpp/server/route_decision_response.cpp
+    )
+    target_include_directories(test_route_decision_response PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+    target_link_libraries(test_route_decision_response PRIVATE nlohmann_json::nlohmann_json)
+    if(USE_SYSTEM_HTTPLIB)
+        target_link_libraries(test_route_decision_response PRIVATE lemonade-httplib)
+    else()
+        target_link_libraries(test_route_decision_response PRIVATE httplib::httplib)
+    endif()
+
+    include(CTest)
+    add_test(NAME RouteDecisionResponseTest COMMAND test_route_decision_response)
 endif()
 
 # ModelManager collection validation for collection.router registration:
--- a/src/cpp/include/lemon/route_decision_response.h
+++ b/src/cpp/include/lemon/route_decision_response.h
@@ -1,0 +1,67 @@
+#pragma once
+
+#include "routing_policy.h"
+
+#include <httplib.h>
+
+#include <functional>
+#include <optional>
+#include <string>
+#include <utility>
+
+namespace lemon {
+
+json route_decision_to_json(const Decision& decision);
+
+std::string route_decision_header_value(const Decision& decision);
+
+void attach_route_header(httplib::Response& res, const Decision& decision);
+
+std::string attach_route_decision_to_sse_event(const std::string& event,
+                                               const json& route_decision_json,
+                                               bool& attached);
+
+class RouteDecisionSseSink {
+public:
+    RouteDecisionSseSink(httplib::DataSink& inner, json route_decision_json);
+
+    bool write(const char* data, size_t length);
+    void done();
+    void done_with_trailer(const httplib::Headers& trailer);
+    bool is_writable() const;
+
+private:
+    void flush();
+
+    httplib::DataSink& inner_;
+    json route_decision_json_;
+    bool attached_ = false;
+    std::string buffer_;
+};
+
+template <typename StreamFn>
+void stream_with_route_decision(httplib::DataSink& sink,
+                                json route_decision_json,
+                                StreamFn&& stream_fn) {
+    if (route_decision_json.is_null()) {
+        stream_fn(sink);
+        return;
+    }
+
+    RouteDecisionSseSink route_sink_wrapper(sink, std::move(route_decision_json));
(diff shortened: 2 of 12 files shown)


ramkrishna2910:
@SlawomirNowaczyk the customer had specific requirements around logging and audit. I want to make sure that this generic decision trace logging satisfies their requirement. Please let us know if this needs any further improvements.
We could do that as a follow up PR but just want to file an issue if needed.

fl0rianr (commented):
Again good direction overall and the core semantics look right: additive response metadata, opaque outputs, default-route handling, and stripping route_trace before backend forwarding.

Overall, I would like to see one robustness fix before merge:

1. The SSE wrapper currently buffers until an exact \n\n. That can break streaming behavior for valid CRLF SSE (`\r\n\r\n`) or other legal event framing, effectively delaying routed streams until `done()`. Please make the event splitter tolerant of CRLF and split chunks, ideally with focused tests around RouteDecisionSseSink.

Two smaller follow-ups that would make this safer:
- sanitize or validate matched_rule before using it as an HTTP header value; rule IDs are currently only validated as non-empty strings.
- add streaming coverage for /chat/completions and /completions, not only /responses, since the PR changes all three paths.

SlawomirNowaczyk:
I suggest to factor out the duplicated streaming-with-route-decision block (maintainability). The same ~12-line sequence is now copy-pasted into all three handlers — handle_chat_completions, handle_completions, and handle_responses:

```
if (route_dispatch) { attach_route_header(res, route_dispatch->decision); }
json route_decision_json = route_dispatch
    ? route_decision_to_json(route_dispatch->decision)
    : json(nullptr);
res.set_chunked_content_provider("text/event-stream", [...] {
    ...
    stream_with_route_decision(sink, route_decision_json,
        [&](httplib::DataSink& route_sink){ router_->..._stream(request_body, route_sink); });
    return false;
});
```
Three near-identical copies mean any future fix (e.g. the CRLF fix above) has to be applied three times. A small helper taking the route_dispatch, res, and the per-endpoint stream lambda would collapse this to one call site and remove the drift risk. Also, the streaming tests only cover /responses — add streaming assertions for /chat/completions and /completions since the PR changes all three paths.

Otherwise I agree with fl0rianr's comments, and can only add some details:

1. Valid SSE allows CRLF framing (\r\n\r\n) and single-\r line terminators. With a CRLF-emitting upstream, \n\n never matches, so the whole routed stream is buffered and only released in flush() at done() — effectively defeating streaming for routed requests. Detect the delimiter as any of \n\n, \r\n\r\n, \r\r, and treat the record separator generically. Add focused unit tests around RouteDecisionSseSink feeding chunk-split input (e.g. the data: prefix split across two write() calls) to lock this in.

2. Constrain rule ids to a safe token charset (e.g. [A-Za-z0-9._-]) at policy-parse time in routing_policy — validating at the source is better than sanitizing at every sink. Related API nit worth resolving now: the header carries only matched_rule and is omitted on the default route, while the JSON body always carries the full decision under x_lemonade_route. A client watching only the header can't distinguish "no routing" from "defaulted." Consider always emitting a header (e.g. x-lemonade-route: __default__ or a default_used signal) or documenting the asymmetry.

SlawomirNowaczyk (approved):

eddierichter-amd:
> json route_decision_json = route_dispatch

Thanks, addressed in [4b19f97](https://github.com/lemonade-sdk/lemonade/pull/2606/commits/4b19f9763a0891b08a49731a8387728ea07af8c5).

eddierichter-amd:
> Again good direction overall and the core semantics look right: additive response metadata, opaque outputs, default-route handling, and stripping route_trace before backend forwarding.
> 
> Overall, I would like to see one robustness fix before merge:
> 
> 1. The SSE wrapper currently buffers until an exact \n\n. That can break streaming behavior for valid CRLF SSE (`\r\n\r\n`) or other legal event framing, effectively delaying routed streams until `done()`. Please make the event splitter tolerant of CRLF and split chunks, ideally with focused tests around RouteDecisionSseSink.
> 
> Two smaller follow-ups that would make this safer:
> 
> * sanitize or validate matched_rule before using it as an HTTP header value; rule IDs are currently only validated as non-empty strings.
> * add streaming coverage for /chat/completions and /completions, not only /responses, since the PR changes all three paths.

Thanks for the comments! Addressed in [4b19f97](https://github.com/lemonade-sdk/lemonade/pull/2606/commits/4b19f9763a0891b08a49731a8387728ea07af8c5).

fl0rianr (approved):
LGTM after the latest updates. Thanks!

PR #2616, opened 2026-07-09
fix(macos): drop dangerous Hardened Runtime entitlements

superm1:
Remove disable-library-validation, allow-dyld-environment-variables, and allow-unsigned-executable-memory from the codesign entitlements, keeping only allow-jit. The first two together let a process influencing the daemon environment inject DYLD_INSERT_LIBRARIES into a signed, notarised binary — a Gatekeeper-bypass LPE primitive (SWSPLAT-24222, CWE-426/250).

Bundled .dylibs must now be signed with the same Team ID to load.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1657,9 +1657,6 @@ if(APPLE)
             <plist version=\"1.0\">
             <dict>
                 <key>com.apple.security.cs.allow-jit</key><true/>
-                <key>com.apple.security.cs.allow-unsigned-executable-memory</key><true/>
-                <key>com.apple.security.cs.disable-library-validation</key><true/>
-                <key>com.apple.security.cs.allow-dyld-environment-variables</key><true/>
             </dict>
             </plist>")
     endif()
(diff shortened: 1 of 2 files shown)


superm1:
@Geramy this was flagged by an AI scan internally.  I'm a little bit worried about the implications for metal, so please have a look over it.

Geramy:
> @Geramy this was flagged by an AI scan internally. I'm a little bit worried about the implications for metal, so please have a look over it.

taking a look at this now, thanks.
I'll compile locally as well to double check.

Geramy:
@superm1 we don’t sign the binaries we download this may not work. Maybe we need to sign the binaries like vLLM?

superm1:
Yeah that's what I was worried about. We probably should be signing macos binaries.

Geramy:
> Yeah that's what I was worried about. We probably should be signing macos binaries.

Do we want to just compile and sign all engines then? Also the plugin system would frown on this fix too. 
I guess the next thing we could do for Mac is implement apples builtin VM containerization system.

Geramy:
Depends on https://github.com/lemonade-sdk/lemonade/pull/2951

superm1:
> Depends on #2951

https://github.com/lemonade-sdk/lemonade/pull/2951 was closed.  What should we do about this one?

Geramy:
checking now

PR #2617, opened 2026-07-09
fix(docker): drop to non-root user and remove unverified FLM download

superm1:
The runtime image ran `lemond` as root and installed FastFlowLM from a checksum-less curl download.

As we've now moved to FLM portable, we don't need to pre-install in the container.

Add an unprivileged `lemonade` user (HOME=/opt/lemonade so the HF/lemonade caches stay writable) and remove the FLM .deb install entirely. Binding stays 0.0.0.0 because Docker port publishing requires it; document loopback publishing and `LEMONADE_API_KEY` as the exposure controls instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,14 +51,15 @@ RUN apt-get update && apt-get install -y \
     unzip \
     libgomp1 \
     libatomic1 \
-    software-properties-common \
-    jq \
     && rm -rf /var/lib/apt/lists/*
 
-RUN add-apt-repository -y ppa:amd-team/xrt
+# Run as an unprivileged user; lemond never needs root at runtime.
+RUN useradd -r -u 10001 -s /usr/sbin/nologin lemonade
 
-# Create application directory
+# The application directory doubles as the user's HOME so the HuggingFace and
+# lemonade caches (both derived from $HOME) resolve to writable, owned paths.
 WORKDIR /opt/lemonade
+ENV HOME=/opt/lemonade
 
 # Provide a private runtime directory so lemond can use get_runtime_dir()
 RUN mkdir -p /run/lemonade && chmod 700 /run/lemonade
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Thanks, the actual Dockerfile change looks good to me overall: non-root runtime user, writable HOME/cache layout, and dropping the unchecked FLM .deb install all make sense.

Two doc issues I think should be fixed before merge:

1. The "specific port and version" example still pins `ghcr.io/lemonade-sdk/lemonade-server:v9.1.3`, but its volumes were changed to the new /opt/lemonade/... layout. For an old pinned image tag this likely no longer matches the runtime cache layout. Either update that example to latest/a new tag, or keep the old /root/.cache/... paths for old-version examples.

2. May not be connected to this change but in a check it came up the embedded Dockerfile in docs/guide/install/docker.md is still missing the runtime-dir setup from the real Dockerfile:
   `RUN mkdir -p /run/lemonade && chmod 700 /run/lemonade`
   and
   `ENV XDG_RUNTIME_DIR=/run/lemonade`.
   Since `get_runtime_dir()` requires a writable `XDG_RUNTIME_DIR` or `RUNTIME_DIRECTORY`, the docs-built image can fail at startup even though the repository Dockerfile works.

Once those are aligned, this looks mergeable to me.

fl0rianr (commented):
Thanks, the previous two issues look fixed now.

One remaining upgrade concern: the examples keep the same named volumes (lemonade-cache, lemonade-llama, lemonade-recipe) while moving them from the old root-owned layout to `/opt/lemonade/...` and running as UID 10001. Existing users may already have these volumes populated by the old root-running container, so the mounted volume contents can remain root-owned. The Dockerfile `chown` will not fix already-mounted volumes at runtime.

In that case it might be helpful if we add a short migration/troubleshooting note, e.g. remove/recreate the volumes or chown them once with a helper container:

```bash
docker run --rm -v lemonade-cache:/v ubuntu:24.04 chown -R 10001:10001 /v
docker run --rm -v lemonade-llama:/v ubuntu:24.04 chown -R 10001:10001 /v
docker run --rm -v lemonade-recipe:/v ubuntu:24.04 chown -R 10001:10001 /v
```

Fresh installs should be fine, but this would avoid a confusing upgrade failure for existing Docker users.

fl0rianr (approved):
Thanks, LGTM!

If you wanna go one more non critical round before merging I just saw this tiny one:
Heading as-is:
**Docker Run with a Specific Port and Version**
Heading more fitting to the "latest" change:
**Docker Run with a Specific Port**

fl0rianr:
Thanks for the small correction, good work!

AreYouLoco:
@fl0rianr actually ```chown 10001:999``` lemonade UID is indeed ```10001``` but GID is ```999```

```
lemonade@d62ed54ac02c:~$ id lemonade
uid=10001(lemonade) gid=999(lemonade) groups=999(lemonade)
```
I am correcting you because I made same assumption in chown command.

EDIT:
So I see instructions for migration from older version are also incorrect in docs/guide/install/docker.md.

EDIT2:
Also for such a small task as simple chown using a container why not use alpine image which is waaay smaller?.

AreYouLoco:
^^ Real quick PR with correction.

PR #2618, opened 2026-07-09
fix(server): size HTTP thread pool from CPU count and add socket time…

superm1:
…outs

The fixed 8-thread pool let a handful of slow-loris or long-lived streaming connections starve the management endpoints. Size the pool from the host CPU count (min 32) and add read/keep-alive timeouts so a stalled client can no longer hold a worker thread indefinitely.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -705,6 +705,7 @@ jobs:
           PYTHONIOENCODING: utf-8
           HF_HOME: ${{ github.workspace }}/hf-cache
           GGML_METAL_NO_RESIDENCY: "1"
+          LEMONADE_WS_IDLE_SECONDS: "31"
         run: |
           set -e
           echo "Resetting server state before endpoint tests..."
@@ -712,6 +713,9 @@ jobs:
           echo "Running endpoint tests..."
           .venv/bin/python test/server_endpoints.py --cli-binary ./build/lemonade
           echo "Endpoint tests PASSED!"
+          echo "Running WebSocket idle test (31s regression)..."
+          .venv/bin/python test/test_websocket_idle.py
+          echo "WebSocket idle test PASSED!"
 
       - name: Run web-app path traversal tests (unsigned path)
         if: steps.check_signing.outputs.has_signing != 'true'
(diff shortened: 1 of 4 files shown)


fl0rianr (commented):
Thanks, the direction makes sense. I think we should also apply the socket timeouts to the UpgradableFrontServer instances, not only to the routed servers, don't you agree?

The main-port listeners are http_front_ / http_front_v6_; they own bind/accept and run the WebSocket upgrade peek before delegating normal requests to RoutedHttpServer. The PR currently sets `new_task_queue` on both front and routed servers, but the timeout block only covers `http_server_` and `http_server_v6_`.

To make the slow-loris protection consistent, please set the same read/write/keep-alive limits on both front and routed servers, or add a short comment explaining why the front servers intentionally keep their defaults.

Minor: consider capping `hardware_concurrency() * 4` to avoid creating very large pools on high-core hosts.

fl0rianr (commented):
Nice! This addresses my concerns. 

What do you think about this suggestion?: A regression check: verify that long-lived WebSocket paths such as /realtime and /logs/stream stay healthy after being idle for more than 30 seconds. Since those sockets are accepted by the front server before adoption by libwebsockets, this is a useful smoke test for the timeout change.

superm1:
> Nice! This addresses my concerns.
> 
> What do you think about this suggestion?: A regression check: verify that long-lived WebSocket paths such as /realtime and /logs/stream stay healthy after being idle for more than 30 seconds. Since those sockets are accepted by the front server before adoption by libwebsockets, this is a useful smoke test for the timeout change.

added

fl0rianr (commented):
Nice addition, thanks for adding the regression test.

I think there are two small but important issues with the current version:

1. The test defaults to port 8080, but the existing server test infrastructure uses LEMONADE_TEST_PORT with our default 13305, and the workflow health check also targets 13305. Since the workflow calls test/test_websocket_idle.py without arguments, this will likely connect to the wrong port in CI.

2. The CI invocation also uses the default idle time of 5s, so it does not actually exercise the 30s front-server read timeout regression. The test docstring correctly says 31s proves the regression, but the workflow does not run that mode.

superm1:
> The CI invocation also uses the default idle time of 5s, so it does not actually exercise the 30s front-server read timeout regression. The test docstring correctly says 31s proves the regression, but the workflow does not run that mode.

This was intentional - I didn't want to add 30s to every CI run.

> The test defaults to port 8080, but the existing server test infrastructure uses LEMONADE_TEST_PORT with our default 13305, and the workflow health check also targets 13305. Since the workflow calls test/test_websocket_idle.py without arguments, this will likely connect to the wrong port in CI.

Will change.

fl0rianr (approved):
LGTM now, thanks for iterating.

Issue #2592, opened 2026-07-07
[Router] Classifier execution backend — no runtime runs encoder text-classifiers


ramkrishna2910:
## Problem

The router's `classifier` condition type needs a runtime that runs **encoder text-classifiers** (BERT/RoBERTa/DeBERTa — sequence- and token-classification heads) → `{label: score}`. **No current backend can:** llama.cpp does embeddings/reranking but not classification heads; FastFlowLM/RyzenAI/vLLM are decoder LLMs; Moonshine is STT. `semantic_similarity` is already covered (embeddings via llama.cpp) — this gap is **only** the `classifier` type. Today `run_classifier` (#2384) piggybacks on `chat_completion`, a stopgap only the `llm` classifier flavor satisfies.

## Decision: Option C — native ORT + curated catalog

Chosen over (A) a torch/transformers subprocess and (B) user-side export. Rationale in-thread: keeps the distribution lean (no PyTorch), gives an NPU/DirectML path, and a curated CI-exported catalog means no per-model export burden on users.

## Plan: a generic `ort-server` (not a classifier-specific binary)

Build a **generic C++ ONNX Runtime model server** — classifiers are consumer #1. Genericity lives in the **artifacts** (self-contained ONNX graphs with tokenizer/pre-post baked in via onnxruntime-extensions + a manifest declaring task/labels/normalization/aggregation); the server stays thin: pick EP, run `InferenceSession`, shape output by task.

Two growth axes, both first-class:

| Axis | v1 | Roadmap |
|---|---|---|
| Execution provider (bin_variants) | `cpu` | `dml` → `rocm` → `vitisai` (NPU) |
| Capability (task) | classification (seq + token) | embeddings, reranking (reuse) |

**lemond side (one backend):** recipe `onnxruntime`, `OnnxRuntimeServer`, implements the capability interface the loaded model needs and dispatches by the model's labels/task to the server endpoint (`/classify` → `/embed` → `/rerank`).

**vs `ryzenai-server`:** that is onnxruntime-**genai** (decoder LLM generation: KV cache, chat API, `genai_config.json`) — structurally cannot run an encoder head. Keep runtimes separate. Reuse is at the EP/packaging layer only; co-location with ryzenai-server is deferred.

## Phasing

[...]

PR #2626, opened 2026-07-09
feat(router): onnxruntime classification backend + /v1/classify (capability for #2592; policy wiring in #2384)

ramkrishna2910:
Adds the `onnxruntime` backend and `/v1/classify` — the **classification capability** the router's `classifier` condition type will consume. The live routing-policy wiring (`run_classifier` → `Router::classify` + policy integration test) is deliberately split to #2384; until that lands, `type: classifier` policies do not yet use this backend. Closes the execution gap in #2592 (dispatch #2385 landed, but encoder text-classifiers had no runner).

## What this does

A new **`onnxruntime`** backend runs encoder text-classifiers (PII, prompt-safety, domain, …) behind a new `/v1/classify` endpoint. Inference runs in a subprocess — **[`lemonade-sdk/ort-server`](https://github.com/lemonade-sdk/ort-server)** — a generic C++ ONNX Runtime model server (CPU EP for v1), mirroring the Moonshine subprocess pattern.

**Ease of use is the headline:** a **stock `optimum-cli export onnx` directory works as-is** — the contract (labels, softmax vs sigmoid per HF `problem_type`, token budget) is inferred from the export's own `config.json`/`tokenizer_config.json`; an optional `manifest.json` overrides it.

**Scope — what "as-is" covers.** This is *not* "any HuggingFace classifier". ort-server supports the **single-sequence encoder convention**: BERT, DistilBERT, RoBERTa, XLM-RoBERTa, DeBERTa (v1/v2), ELECTRA, ALBERT, CamemBERT. tokenizers-cpp returns **token ids only**, so the attention mask, segment ids, padding, and special-token filtering are **reconstructed by ort-server** to match the HF pipeline. That reconstruction is verified against **HF-computed reference scores on every release** (sequence- and token-classification, padded tokenizers, non-ASCII input) — but it is valid only for the families above. Anything else (e.g. XLNet) uses a different convention and is **rejected at load time** with an explicit error rather than served with silently wrong scores. A servable directory is `model.onnx + tokenizer.json + config.json`; the config is mandatory because it carries the architecture that gates this check.

## Changes (lemond side)

- New capability `IClassificationServer` + `ModelType::CLASSIFICATION` + `Router::classify`; quad-registered `POST /v1/classify` (+ `handle_classify`) and `docs/api/lemonade.md`.
- `onnxruntime` backend (`OnnxRuntimeServer` + descriptor) wired to `lemonade-sdk/ort-server` (pin `0.3.7`). `OnnxRuntimeOps` resolves only a complete, unambiguous model directory (`model.onnx + tokenizer.json + config.json`, matching the backend's own contract; `manifest.json` is an optional output-contract override). HF downloads resolve inside the `refs/main` snapshot; recursive exactly-one resolution is kept for local imports. Ambiguity and incompleteness are precise load-time errors.
- Regenerated backend docs (`gen_backend_boilerplate.py`) so `onnxruntime` appears in the support matrix / CLI / defaults / models.js.

## ort-server (separate repo — functional, released)

Released **0.3.7** (bundles ONNX Runtime 1.27.0, SHA-pinned deps, full third-party license texts). Release CI runs a **contract suite against the extracted packaged archive** on all four platforms, including **golden parity checks against a HuggingFace-computed reference** — the oracle that caught the padding bug below. The catalog tooling (`tools/classifier_catalog/`: Optimum export wrapper + parity validation + fail-closed publish gates) **lives in the ort-server repo** so the manifest contract's producer and consumer version together.

## Catalog (published to lemonade-sdk HF)

Two **usable, apache-2.0** models, each with a model card (base-model link, the `export.py` used, and post-export validation + perf), registered in `server_models.json`:
- `Bert-Phishing-ONNX` — [bert-finetuned-phishing-ONNX](https://huggingface.co/lemonade-sdk/bert-finetuned-phishing-ONNX), parity Δ=0.0
- `Phishing-Email-Detection-ONNX` — [phishing-email-detection-distilbert-ONNX](https://huggingface.co/lemonade-sdk/phishing-email-detection-distilbert-ONNX), parity Δ=0.0

**Removed / excluded for licensing or fidelity** (validated but not shipped): piiranha (`cc-by-nc-nd-4.0` — No-Derivatives forbids redistributing an export), roberta-hate-speech + prompt-injection-defender (no declared license → no redistribution grant), nvidia/domain-classifier (custom head fails ONNX parity), Llama-Prompt-Guard-2 (gated; **recoverable** — Llama 4 license permits redistribution with attribution).

## Testing

- **ort-server CI smoke test** — `/classify` on a real model, all four platforms.
- **`test/server_classify.py`** — end-to-end through the product (client → lemond → `/v1/classify` → onnxruntime backend → ort-server): pulls a registered model, asserts label scores in `[0,1]`, and checks missing-model / invalid-model / missing-input failure paths. Registered in `capabilities.py` and added to the `Test .exe` CI matrix.

## Known limitations / follow-ups

- **Position-based tokenizer metadata is reconstructed, not returned by the tokenizer.** tokenizers-cpp hands back ids only, so ort-server rebuilds the special-token mask, padding, and segment ids. The special-token mask now covers all three HF post-processor shapes — `TemplateProcessing` (BERT/DistilBERT), `Bert`/`RobertaProcessing` (stock RoBERTa's `<s>`/`</s>` live in `cls`/`sep` fields, not the template map), and nested `Sequence` wrappers — each verified against the HF reference in `ort-server` 0.3.7 (the `tiny-roberta-tok` golden fails by 0.70 without the fix). Remaining deviations (left padding, fully custom post-processors, pair/two-sequence inputs) are out of scope today and tracked in [ort-server#1](https://github.com/lemonade-sdk/ort-server/issues/1); the load-time architecture allowlist keeps anything outside the verified convention refused rather than mis-served.
- **Coverage is currently phishing-only** — PII, prompt-safety, hate-speech, and domain were lost to licensing/parity. Rebuilding coverage means finding permissively-licensed alternatives (and the Llama-Guard recovery).
- `bert-finetuned-phishing` exports to **1.3 GB fp32** → int8 quantization is a sensible follow-up.
- nvidia-domain export fidelity (#6); Llama-Prompt-Guard recovery.

## Checklist

- [x] ort-server functional + released **0.3.7** (contract + golden-parity suite against the extracted packaged archive on all 4 platforms)
- [x] Round-2 + round-3 review fixes and self-review — see review-response comments
- [x] **Correctness oracle**: served scores asserted against the original PyTorch model end-to-end (caught a padding bug that made every served score wrong)
- [x] lemond `onnxruntime` backend + `/v1/classify` (compiles locally + CI; docs regenerated, no drift)
- [x] HF catalog (2 apache-2.0) published + registered; unlicensed/ND models removed
- [x] Integration test + CI matrix wiring
- [x] `main` merged, conflicts resolved
- [x] Full CI matrix green (incl. `Test .exe - classify-onnxruntime`)
- [x] Review fixes (fl0rianr / SlawomirNowaczyk) — see review-response comment: telemetry privacy, forward_request lifecycle, error classification, request validation + response envelope, deterministic model.onnx resolution, ARM64 fail-fast, publish.py fail-closed allowlist + parity gate, ort-server manifest contract + licensing

## Related

- #2384 (ClassifierServices ↔ Router wiring — `run_classifier` → `Router::classify`, plus a live `type: classifier` policy integration test). Split out of this PR per review discussion; this PR ships the capability, #2384 ships the router feature.


ramkrishna2910:
## HF catalog published + wired

**Published to lemonade-sdk (each card: base-model link, `export.py`, post-export validation + perf):**
- ✅ [bert-finetuned-phishing-ONNX](https://huggingface.co/lemonade-sdk/bert-finetuned-phishing-ONNX) — apache-2.0, parity Δ=0.0
- ✅ [phishing-email-detection-distilbert-ONNX](https://huggingface.co/lemonade-sdk/phishing-email-detection-distilbert-ONNX) — apache-2.0, parity Δ=0.0

**Wired into `server_models.json`** (recipe `onnxruntime`, label `classification`) as `Bert-Phishing-ONNX` and `Phishing-Email-Detection-ONNX`. Added `OnnxRuntimeOps` to resolve the checkpoint to the `model.onnx` directory (mirrors RyzenAI).

**License holds:**
- 🗑️ **piiranha — removed.** Upstream is `cc-by-nc-nd-4.0` (Non-Commercial **No-Derivatives**); an ONNX export is a derivative, so redistribution isn't permitted. Repo deleted; reproducible via `export.py` if licensing ever clears.
- ⏸️ **roberta-hate-speech** and **prompt-injection-defender** published but **not registered** — upstream declares no license. Hold until terms are confirmed.

Remaining: lemond CI build (validates the new `OnnxRuntimeOps`), integration test (#3).

fl0rianr (changes requested):
Thanks for the substantial work here. The overall direction is good: a dedicated classification capability, model type, backend, and `/v1/classify` endpoint fit the existing architecture well. However, I see several issues that should be addressed before merging:

### Blocking

1. **The router’s classifier condition is not wired to the new endpoint yet.**
   ClassifierServices::run_classifier() still calls Router::chat_completion(), while make_router_classifier_services() does not pass Router::classify(). As a result, direct `/v1/classify` requests may work, but actual router policies using `type: classifier` still cannot use this backend.

2. **forward_classify() bypasses the common wrapped-backend lifecycle.**
   It calls `HttpClient::post(..., 0)` directly, where `0` means no HTTP timeout. This also bypasses `WrappedServer::forward_request()`, including active-request tracking, watchdog handling, crash recovery, normalized backend errors, and retry behavior. Please route classification through the common helper with a finite timeout.

3. **Backend failures are reported as client errors.**
   The broad exception handler in `classify()` converts connection failures, timeouts, and backend crashes into `400 invalid_request_error`. Only malformed request input should return 400; transport and process failures need to remain backend/network errors so watchdog recovery can work.

4. **The ort-server manifest contract is not fully implemented or validated.**
   `score_normalization` is ignored and `token_aggregation` is read but not applied. Token classification always uses max aggregation, unknown tasks silently behave as sequence classification, label IDs/output dimensions are not validated, and long inputs are not truncated although the export validation uses `truncation=True`. This can cause incorrect results or unsafe tensor access for malformed/incompatible artifacts.

5. **The publishing script still includes models excluded for licensing reasons.**
   `publish.py` contains Piiranha, RoBERTa hate-speech, and prompt-injection-defender in its default list, although the PR correctly states that they must not be redistributed. The publisher should use an explicit approved allowlist and fail closed when the license is missing or unsuitable.

6. **The released `ort-server` bundle needs licensing and packaged-artifact validation.**
   Its repository still has a license TODO, and the release archive does not appear to include the required project/third-party license notices. Also, CI smoke-tests the binary before packaging, not the extracted release archive. Please test the final archive on every published platform and ensure relocatable shared-library loading.

### Additional issues

* Validate `input`/`text`, `model`, `top_k`, and malformed JSON before downloading or loading a model.
* The API docs show an excluded model and document a `model` response field that the implementation does not return.
* Export parity is currently recorded but not enforced as a publication gate.
* Please add an integration test that exercises a real router policy with `type: classifier`, not only the direct HTTP endpoint.

The capability and backend structure are promising, but the first three points in particular mean the headline router use case and backend recovery semantics are not complete yet. I would request changes until these are resolved.

SlawomirNowaczyk:
> 1. **The router’s classifier condition is not wired to the new endpoint yet.**

I don't think this should be included in this PR. I'd prefer this to just add the required functionality, and a separate (small) PR to update the router to start using it. Seems cleaner if we ever encounter any issues and need to backtrack...

SlawomirNowaczyk (changes requested):
Mostly looks good. I agree with all issues raised by **fl0rianr** (except changing the router) but here are a couple more improvement suggestions:

1. **Telemetry silently captures the raw classifier input — the one place it's most likely sensitive.** In telemetry.cpp, the `InferenceSpan` constructor only special-cases `span_kind == "LLM"` and `"EMBEDDING"`; the `"CLASSIFIER"` kind falls into the `else` branch and does `request_dump_ = truncate_string(request_json.dump(), max_len)`, capturing the full input text. `Router::classify` also calls `span->end_with_success(…, response.dump())`, recording the label scores. Classification is exactly the workload where inputs are PII, phishing email bodies, or prompt-injection payloads — this ships them into OTLP export by default. Add a `CLASSIFIER` branch that records only length/hash (or nothing), never the raw text.

2. **`model.onnx` resolution is non-deterministic.** onnxruntime_server.cpp returns the parent of the *first* `model.onnx` yielded by `recursive_directory_iterator`, whose traversal order is unspecified and differs across platforms/filesystems. HF ONNX exports frequently contain more than one graph (e.g. a top-level `model.onnx` plus `onnx/model.onnx`, or fp32 vs. quantized variants in sibling folders). Whichever one wins is effectively luck. Resolve deterministically — prefer a manifest-declared path, or the shallowest match, and error if the count isn't exactly one.

3. **`PYTHONNOUSERSITE=1` is passed to a native C++ subprocess.** In onnxruntime_server.cpp the env var `PYTHONNOUSERSITE` is set before launching ort-server, but ort-server is a self-contained C++ binary (tokenizers-cpp, ONNX Runtime — no Python). This is a copy-paste artifact from the Python-based backends (moonshine/ryzenai) and is dead, misleading config. Remove it, or document why the ONNX subprocess needs a Python isolation flag.

4. **Windows-on-ARM64 silently downloads the x64 bundle.** In onnxruntime_server.cpp the `#ifdef _WIN32` branch fires before the `__aarch64__ || _M_ARM64` check, so a Windows ARM64 host resolves to `ort-server-<v>-windows-x64.zip`. There's a `windows-arm64` gap. Either add a Windows ARM64 case (and publish that artifact) or fail fast with an explicit "unsupported platform" error instead of pulling an x64 binary that may or may not run under emulation — this also aligns with the cross-platform invariant in AGENTS.md.

5. **The public `/v1/classify` contract is just ort-server's raw output, with no lemond-side envelope or versioning.** `classify()` forwards `json::parse(res.body)` straight through, so the endpoint returns whatever ort-server emits (`{"labels": {...}}`) with no `object`/`model`/`usage` envelope like every other lemonade endpoint. That means a future ort-server output change silently mutates the public API, and clients can't tell which model produced the result. Define a stable response schema in lemond and map the subprocess output onto it (including echoing the resolved `model`), decoupling the wire contract from the pinned subprocess version.

All except the last one seem reasonably simple. In a pinch, #5 could wait and be solved in another PR, but it's important that it's done before the official release (it's a breaking change), so it's probably safer to handle it already here?

ramkrishna2910:
Thanks @fl0rianr and @SlawomirNowaczyk — these reviews caught real bugs, including two serious ones. All findings are addressed except router wiring, which is deliberately split out (details at the end). Fixes are in the 5 commits ending at 000e8fad plus **ort-server 0.3.0** (pin bumped).

## lemond

- **Telemetry captured raw classifier inputs** (@SlawomirNowaczyk) — best finding of the review. `CLASSIFIER` spans now record only input length + a salted hash, never the text; `Router::classify` no longer records label scores. (aa85af69)
- **`forward_classify` bypassed the wrapped-server lifecycle** (@fl0rianr) — now goes through `forward_request()` with an explicit 120s timeout, plus `start_backend_watchdog("/health")` after load: active-request tracking, crash recovery, and normalized backend errors all apply. One correction for the record: the literal `0` timeout actually meant "HttpClient default (300s)", not infinite — but the lifecycle bypass was real. (1f1cc950)
- **Backend failures reported as 400** (@fl0rianr) — the broad catch is gone; only a missing/invalid `input` is a client error, transport/backend failures surface as backend errors so the watchdog can recover. (1f1cc950)
- **Request validation before model load + response envelope** (@fl0rianr, @SlawomirNowaczyk) — malformed JSON / missing input / bad `top_k` are rejected 400 before any model loads. Success responses are now `{object: "classification", model, labels}` so lemond pins the public API shape instead of forwarding ort-server's raw output; docs updated (the excluded-model example is gone, `model` is now real) and the integration test asserts the envelope. Did this now since changing it later is breaking. (1005a419)
- **Non-deterministic `model.onnx` resolution** (@SlawomirNowaczyk) — candidates are now ranked deterministically: manifest-bearing directory first, then shallowest path, then lexicographic; ambiguity logs a warning with the chosen path. Kept it non-throwing because resolution runs during bulk model-path resolution, where one ambiguous import shouldn't break listing for everything else. (1f1cc950)
- **`PYTHONNOUSERSITE` on a native binary** — removed (copy-paste artifact, correctly called). **Windows-on-ARM64** now fails fast with a clear error instead of silently downloading the x64 bundle. (1f1cc950)
- **`publish.py` still listed license-excluded models** (@fl0rianr) — now an explicit allowlist that **fails closed**: each entry declares its expected license, verified against the source repo at publish time (lookup failure = refuse; mirrors treated the same as exports). Excluded models are documented in-file with the reason. Parity is now a publication gate too (`max_score_delta ≤ 1e-4` / token agreement ≥ 0.999), and Llama entries refuse to publish without LICENSE + USE_POLICY. (b97aecaa)

## ort-server 0.3.0

- **Manifest contract enforced** (@fl0rianr): `score_normalization` (softmax/sigmoid/none) honored; `token_aggregation` (max/mean) applied; unknown task / unknown values are a **startup error**; `id2label` size validated against the model's output dimension before indexing (was a potential OOB read — verified locally that a mismatch now returns a clean 500); inputs truncated to `max_length` (default 512) preserving the trailing special token. `/classify` returns 400 for parse errors, 500 for inference failures.
- **Licensing** (@fl0rianr): Apache-2.0 `LICENSE` + `THIRD_PARTY_NOTICES.md` (ORT, tokenizers-cpp, HF tokenizers, sentencepiece, httplib, nlohmann/json) committed and included in every release archive; README TODO resolved.
- **Packaged-artifact validation** (@fl0rianr): release CI now extracts the packaged archive and smoke-tests *that* binary on all 4 platforms. This immediately paid for itself: it exposed that the Linux/macOS binaries carried only build-tree rpaths (fixed with `$ORIGIN`/`@loader_path`) and that macOS staging tripped over ORT's `.dSYM` directory (previously swallowed by `|| true`).

Verified end-to-end locally: fresh lemond build downloaded the real 0.3.0 archive, watchdog engaged, `test/server_classify.py` 4/4 green including the new envelope assertions.

## Router wiring — split to #2384

@fl0rianr's point that `type: classifier` policies still can't reach this backend is correct — but @SlawomirNowaczyk asked for that wiring in a separate PR, and we're going that way: this PR ships the capability; #2384 (already in the Lemonade Router milestone, scoped in a comment there) ships `run_classifier` → `Router::classify` plus the live-policy integration test fl0rianr asked for, as the immediate follow-up.

fl0rianr (changes requested):
Thanks for addressing the previous review in depth.  Many thinks addressed and improved.

Splitting the actual router-policy wiring into #2384 is reasonable following the review discussion. However, I still see several issues that should be resolved before this PR is approved:

### Blocking

1. **The token-classification catalog output is incompatible with ort-server 0.3.0.**
   export_and_benchmark.py still writes `token_aggregation: "first-subword"`, while the server accepts only `max` or `mean`. Token-classification artifacts produced by this tool therefore fail during server startup.

2. **The exporters ignore Hugging Face problem_type semantics.**
   export.py, publish.py, and the catalog benchmark hard-code softmax for every sequence classifier. Multi-label models require independent sigmoid scores, while regression models must not be softmaxed at all. The current parity check can report a perfect delta because it applies the same incorrect softmax transformation to both outputs. Either infer and persist the correct normalization from the model configuration or explicitly restrict the supported model types.

3. **The model’s actual maximum sequence length is not persisted.**
   ort-server defaults to 512 tokens, but the exporters do not write a max_length derived from the tokenizer/model configuration. Runtime truncation can therefore differ from the Hugging Face reference for models whose limit is not 512, while the current short parity fixtures still pass.

4. **The score contract is internally inconsistent.**
   The public API and integration test promise label scores in [0,1], but the manifest accepts `score_normalization: "none"` and then returns raw logits. Please either remove `none` from classification manifests or revise and validate the public response contract.

5. **The release archive still lacks the complete third-party license texts.**
   It contains the project Apache license and a notices table, but the bundled MIT-licensed ONNX Runtime library’s copyright and permission notice is not included verbatim. Please package the applicable third-party license/NOTICE files rather than only linking to their repositories.

6. **The released build is not reproducible.**
   `tokenizers-cpp` is fetched from the moving `main` branch, and the ONNX Runtime archive has no integrity hash. Please pin an exact tokenizer commit and add a URL_HASH for downloaded binary dependencies.

### Additional corrections

* Since live `type: classifier` policy wiring is deferred to #2384, the PR title, opening description, and API documentation should say that this PR provides the classification capability **for future router integration**, not that router classifier conditions already work.
* Resolve only a complete and unambiguous model directory containing model.onnx, tokenizer.json, and manifest.json; deterministic selection of one ambiguous candidate can still silently run the wrong graph.
* Clean each publication staging directory before exporting and assert exactly one intended ONNX model before benchmarking/uploading.
* Add contract tests for token classification, sigmoid/multi-label behavior, non-512 truncation, top_k, malformed JSON and field types, ambiguous imports, and backend crash/reload recovery.

The current implementation is much closer, but the first four findings mean the advertised generic classifier contract is still not reliable. I would keep the change request open until these are addressed.

ramkrishna2910:
Thanks @fl0rianr — round 2 caught real contract bugs again. All six blocking items plus the corrections are addressed, in **ort-server 0.3.1** (pin bumped) and the latest commits here. We also ran a deep self-review afterward, which found one bug worse than anything on the list — details at the end.

## The blocking items

**1. `first-subword` incompatibility** — fixed at both ends. The exporters now emit only `max`/`mean`, and (from round 1) ort-server rejects unknown values *at startup*, so this class of drift can't ship silently again. Structurally fixed too — see "one repo" below.

**2. `problem_type` semantics** — the exporters now infer `sigmoid` for `multi_label_classification` and refuse regression/single-output heads outright; parity is validated through the **exact transformation ort-server serves**. Because `problem_type` is only a training-time hint (BCE-trained models often leave it null), the publish allowlist also pins a human-verified normalization per model and refuses on mismatch, and `export.py` takes an explicit `--normalization` override.

**3. `max_length` not persisted** — exporters now derive it from `tokenizer.model_max_length` (sentinel-aware, with a conservative `max_position_embeddings - 2` fallback for RoBERTa-style off-by-two configs) and write it to the manifest; validation tokenizes with the same budget and includes an over-length fixture so the truncated path is actually exercised.

**4. `"none"` vs the `[0,1]` contract** — `none` is removed; ort-server accepts only `softmax`/`sigmoid` and rejects anything else (and now also rejects *wrong-typed* manifest values instead of silently defaulting).

**5. License texts** — release archives now carry a `licenses/` directory with **verbatim texts for every statically linked component**, including the transitive ones (HF tokenizers, oniguruma + rust-onig, sentencepiece and its vendored abseil/protobuf-lite/darts-clone/esaxx, msgpack) plus ORT's LICENSE + ThirdPartyNotices staged from the version-pinned upstream package. The smoke test asserts each file exists in the extracted archive.

**6. Reproducibility** — tokenizers-cpp pinned to a commit SHA, ORT archives pinned by SHA256 (`URL_HASH`), httplib/json moved from mutable tags to commit SHAs, Rust toolchain pinned. `docs/UPGRADING-ORT.md` was also rewritten — it still referenced the removed onnxruntime-extensions knob and a `v`-prefixed tag command that would have produced an undownloadable release.

## Corrections

- **PR framing** — retitled and re-worded (PR, docs, ort-server README): this ships the classification *capability*; live `type: classifier` wiring is #2384.
- **Ambiguous resolution** — resolution now only succeeds on **exactly one complete model directory** (model.onnx + tokenizer.json + manifest/config); anything else is a precise load-time error listing the candidates. (Kept non-throwing during bulk listing so one bad import can't break `/models`.)
- **Staging hygiene** — publish staging dirs are cleaned per run and exactly one `.onnx` is asserted before benchmark/upload.
- **Contract tests** — ort-server's release CI now runs a **27-check suite against the extracted packaged archive**: sigmoid, token-classification max/mean, manifest-less inference, non-512 truncation, top_k, malformed JSON/field types, startup rejects, output-dim mismatch. lemond's integration test grew top_k passthrough + wrong-type 400s (7 checks).

## Two structural changes worth knowing about

**The manifest is now optional.** ort-server infers the contract from a stock Optimum export's own `config.json`/`tokenizer_config.json` (task from `architectures`, sigmoid from `problem_type`, budget from `model_max_length`) — so `optimum-cli export onnx` output runs as-is, and `manifest.json` becomes a strictly-validated override rather than a required artifact our tooling can get wrong.

**The catalog tooling moved to the ort-server repo.** Your finding #1 was a direct consequence of the manifest contract's producer and consumer living in different repos; they now version together. This PR shed ~700 lines of Python in the process.

## What our own review found

A deep self-review after these fixes caught something worse than anything above: **ort-server was tokenizing without special tokens** (the tokenizers-cpp base interface defaults `add_special_tokens=false`), so served scores diverged from the HF reference the catalog validates — BERT-family classifiers pool `[CLS]`. Fixed in 0.3.1 via the tokenizers C API (with a mutex honoring the FFI's `&mut` contract), plus: output dtype guard (fp16 exports were a heap over-read), rank guard, strict `id2label` key parsing, non-zero exit on bind failure, error-`type`→HTTP-status mapping in lemond (router-level errors previously collapsed to 500), empty-input 400, per-request load options honored, and a CI cleanup gap — the Windows cleanup action never killed orphaned `ort-server.exe`/`moonshine-server.exe` processes, which was the root cause of the earlier classify-job flake on the self-hosted runners.

Verified end-to-end locally: lemond auto-upgraded the backend 0.3.0→0.3.1 from the release and all 7 integration checks pass. CI running now.

SlawomirNowaczyk (approved):

fl0rianr (commented):
Thanks again for the thorough follow-up, good work.

After reviewing the current Lemonade head and ort-server 0.3.1, I still see four correctness issues:

### Blocking

1. **The documented model-optional request does not work.**
   The docs now say model may be omitted when a classification model is already loaded. `handle_classify()` allows that request through, but passes it unchanged to `Router::classify()`, whose shared `execute_inference()` helper rejects every request without a model. A request without model therefore still returns 400 after a classifier has been loaded. Please either keep model required or implement an unambiguous classification-specific fallback and test it after a successful load.

2. **The “unambiguous model directory” fix still misses a root-plus-nested layout.**
   If the checkpoint root itself contains a complete model and a nested directory contains another complete model, `find_imported_checkpoint()` returns no result because it sees multiple candidates. `resolve_checkpoint_path()` then falls back to the root, and `load()` accepts it immediately without checking the nested candidate. This still silently chooses one graph in exactly the kind of ambiguous layout the fix is intended to reject.

3. **The generic Hugging Face parity claim is still too broad.**
   `ort-server` receives only token IDs from tokenizers-cpp and fabricates an all-ones attention mask and all-zero token-type IDs. That is correct for common BERT-like single-sequence models, but not for every Hugging Face encoder. XLNet, for example, assigns a distinct segment ID to its classifier token, and its post-processing ends with both `SEP` and `CLS`; the current manual truncation can drop `SEP` and the generated token-type IDs are wrong. Either preserve the tokenizer’s complete encoding metadata/truncation semantics or explicitly restrict the supported model families instead of claiming guaranteed parity for any classifier.

4. **Manifest-less normalization remains ambiguous and can silently be wrong.**
   The config fallback chooses sigmoid only when `problem_type == "multi_label_classification"` and otherwise assumes softmax. The export tool itself correctly notes that externally trained BCE/multi-label models often leave `problem_type` unset. Such a stock export will load successfully but return semantically incorrect scores. Require an explicit manifest for ambiguous configs, or document and enforce that manifest-less inference assumes single-label softmax.

### Additional correction

The manifest-less runtime path reads `model_max_length` only from `tokenizer_config.json`. When that value is absent or is Hugging Face’s unlimited sentinel, it defaults to 512 and does not use `config.max_position_embeddings`, even though the export tool does. This can exceed a model’s real position limit. The contract tests currently prove only that truncation completes, not that C++ tokenization matches the Hugging Face reference.

The implementation is now close and the two shipped BERT-like phishing models appear much better covered, but the public generic-model claims still exceed what the runtime safely supports. I would keep the change request open until the two direct Lemonade bugs are fixed and the model contract is either completed or accurately narrowed.

eddierichter-amd (changes requested):
I independently reproduced Florian’s first blocking item around the documented model-optional `/classify` path.

  After rebuilding this branch so the ONNX backend was included, I loaded `Phishing-Email-Detection-ONNX` and confirmed that an explicit model request succeeds:

with model status 200
  `{"labels":{"LABEL_0":9.08490692381747e-05,"LABEL_1":0.9999090433120728,"LABEL_2":2.5915237955587145e-08,"LABEL_3":9.73448948116129e-08},"model":"Phishing-Email-
  Detection-ONNX","object":"classification"}`

Then I retried the same request without model, with the classifier still loaded, and it fails:

  without model status 400
  `{"error":{"message":"Invalid request: No model specified in request","type":"invalid_request"}}`

  So I agree this part still needs to be resolved. Either:

  - make the intended model-less path work by resolving/injecting the currently loaded classification model before calling Router::classify(), and add a regression test
    after a successful load; or
  - narrow the API/docs/tests so model is required for /classify.

  Given handle_classify() already allows the missing-model case when a model is loaded, and the response envelope has fallback logic for requested_model.empty() ?
  router_->get_loaded_model() : requested_model, the first option looks like the intended behavior.

ramkrishna2910:
Thanks @fl0rianr and @eddierichter-amd — both of you were right, and the oracle you pushed for immediately found something worse than anything on either list.

## First: building the parity test you asked for exposed a live correctness bug

@fl0rianr's closing point in round 2 was that *"the contract tests prove only that truncation completes, not that C++ tokenization matches the Hugging Face reference."* So I built that test — golden scores computed with the real HF tokenizer + PyTorch, asserted end-to-end. It failed immediately, on the real published model:

| probe: `"Account notice: sign in to review recent activity."` | LABEL_0 |
|---|---|
| HuggingFace (PyTorch reference) | **0.747434** |
| ort-server 0.3.1 (shipped) | **0.000550** |
| ort-server 0.3.3 (fixed) | **0.747433** |

**Cause:** a `tokenizer.json` with a `padding` section — which real HF repos ship, including our own published classifiers — makes the Rust tokenizer pad every encoding to a fixed width. `transformers` does not pad by default. We forwarded those `[PAD]` ids to the model **under our all-ones attention mask**, so it attended to ~500 padding tokens as if they were text. Every score served for the catalog models was wrong. Affects 0.2.0–0.3.2; fixed in **0.3.3**.

**The published artifacts are fine** — the ONNX graphs still match their PyTorch sources exactly (verified). This was purely a server bug, so the model cards remain accurate and nothing needs republishing.

Two things had hidden it, both of which are now closed: our fixtures' tokenizers declared no padding (there is now a `tiny-pad` fixture that does), and every check was *structural* — labels present, scores in [0,1], softmax sums to 1 — all of which pass happily on garbage. The new probe is deliberately **non-saturated** (top score ~0.75) because on the obvious phishing text the model sits at 0.9999 either way and the bug hides in the noise. I also verified the oracle has teeth by reintroducing the old missing-special-tokens bug: it fails loudly (1.77e-03 ≫ 1e-4).

## Blocking items

**1. Model-optional `/classify` (both of you).** Confirmed dead code — exactly as @eddierichter-amd reproduced. `handle_classify()` allowed the request through but never put a model *into* it, and `execute_inference()` requires one, so it always 400'd and the envelope's `get_loaded_model()` fallback was unreachable. Now: the sole loaded classification model is resolved and injected; **zero or more than one** loaded classifier is refused with a clear message rather than silently picked. Regression test reproduces your exact scenario (load → omit `model` → 200 + echoed model). My round-2 fix only corrected the *HTTP status* of that failure — I fixed the symptom and left the feature broken. That one's on me.

**2. Root-plus-nested ambiguity.** Correct — `resolve_checkpoint_path()` falls back to the cache root, and `load()` accepted a complete root without ever looking for nested candidates, so the two halves defeated each other. `load()` is now the sole arbiter: it scans unconditionally and refuses unless exactly one complete directory exists, listing the candidates. Verified against a hand-built root+nested layout.

**3. The generic-parity claim was too broad.** Agreed, and I've narrowed the claim instead of faking the metadata. We fabricate an all-ones mask, all-zero `token_type_ids`, and keep the trailing token on truncation — right for single-sequence encoders, wrong for XLNet-style segment/special-token conventions (and the tokenizers C API doesn't expose the real mask/type-ids to do otherwise). Supported architectures are now an **explicit allowlist** checked against `config.json`'s `model_type` at startup — BERT, DistilBERT, RoBERTa, XLM-R, DeBERTa v1/v2, ELECTRA, ALBERT, CamemBERT. Anything else is **refused at load**, not served with wrong scores. PR, API docs and both READMEs now say exactly this.

**4. Manifest-less normalization.** Agreed — and I'd already made this argument myself in `publish.py` (pinned per-model normalization) and then failed to apply it to the runtime. Manifest-less inference now **assumes single-label softmax**, logs that assumption at startup, and documents it; sigmoid requires an explicit declaration (`problem_type: multi_label_classification` or a manifest).

**Additional correction:** the C++ `max_length` fallback now mirrors the exporter (`model_max_length` → `max_position_embeddings - 2` → 512) on both the manifest and inferred paths, instead of a blanket 512.

## Verification

ort-server 0.3.3 green on all four platforms (contract + golden parity against the packaged archive). Locally, with the backend cache wiped so lemond downloads the real release: **9/9** integration checks, including the model-optional regression and the HuggingFace-reference parity assertion.

fl0rianr (commented):
Thanks for the further fixes. The previously reported model-optional request bug, root-plus-nested ambiguity case, architecture scope, max-length fallback, and lack of a real correctness oracle have all been addressed meaningfully. The Hugging Face reference-score test is especially valuable.

I still see three correctness issues before approval:

### Blocking

1. **Normal Hugging Face cache history is treated as model ambiguity.**

   `CheckpointResolveContext::model_cache_path` is the repository cache directory, but `OnnxRuntimeOps::resolve_checkpoint_path()` recursively scans the entire directory and requires exactly one complete model. After a normal upstream update, Hugging Face can retain multiple snapshots while `refs/main` identifies the active one. The backend will then reject the model as ambiguous even though the active revision is unambiguous.

   Please resolve and validate inside `hf_cache::active_snapshot_path()` for Hugging Face models, while keeping the exactly-one rule for local imports. Add a regression test with two cached snapshots and a valid `refs/main`.

2. **A manifest-only model bypasses the architecture allowlist.**

   Lemonade accepts `model.onnx + tokenizer.json + manifest.json` without `config.json`. In `ort-server`, `validate_model_family()` explicitly proceeds when the config is missing, and malformed config is also converted to an empty object. An unsupported architecture can therefore omit or corrupt `config.json` and still be served using the fabricated all-ones mask and all-zero token-type IDs.

   The manifest currently describes output semantics only; it cannot assert the required input convention. Please require a valid `config.json`, or extend the manifest with a strictly validated model/input convention. Add startup-rejection tests for an unsupported manifest-only model and malformed config.

3. **Serialized tokenizer preprocessing is still only partially neutralized.**

   Version 0.3.3 strips trailing padding IDs, but tokenizer padding can be left- or right-directed. A fixed left-padded tokenizer still sends padding tokens to the model under an all-ones attention mask. The raw Tokenizers encoding path also applies any serialized `truncation` configuration before the server’s own max-length handling, which can silently produce a different input from the Hugging Face reference.

   Please either consume the full tokenizer encoding metadata, including attention mask and type IDs, or reject unsupported padding directions and non-null embedded truncation. Add golden tests for fixed left padding and tokenizer-embedded truncation; the current fixture covers only fixed right padding.

### Additional corrections

* The PR description and `ort-server` README still claim that any Hugging Face classifier works and that parity is guaranteed. Please align those claims with the explicit architecture and preprocessing restrictions.
* The current head does not have a green CI matrix: Server Download Registry, MCP Smoke Test, Docker Build Smoke Test, and Linux Distro Builds have failed, while the central C++ workflow has not completed. These need to be resolved or shown to be unrelated and rerun successfully.

The implementation is now close, and the two shipped BERT-like models have much better end-to-end coverage. I would keep the change request open until the cache-resolution and remaining tokenizer/architecture safety gaps are closed.

eddierichter-amd (approved):

fl0rianr (commented):
Thanks for the continued work here! The shipped BERT-family sequence classifiers now look well covered, the Windows packaging issue is fixed in ort-server 0.3.5.

I think the remaining concern is now quite narrow: the implementation is reliable for the catalog models, but the advertised generic model contract is still broader than the runtime safely supports.

First, ONNX model resolution scans the complete Hugging Face repository cache. A normal cache containing an old and a current snapshot is therefore rejected as ambiguous even though `refs/main` identifies the active revision. Hugging Face downloads should be resolved inside the active snapshot; recursive exactly-one resolution still makes sense for local imports.

Second, architecture validation can be bypassed by a manifest-only artifact. Missing or malformed `config.json` currently causes the model-family check to be skipped, although the manifest does not describe the input-mask or segment-ID convention. A valid config with an allowed `model_type` should remain mandatory unless that input contract is added explicitly to the manifest.

There is also one broader scope mismatch in ort-server: only token IDs are retained while masks, type IDs, padding handling and special-token filtering are reconstructed manually. This is sufficient for the two validated sequence classifiers, but it does not justify the current “any Hugging Face classifier” and “guaranteed parity” wording. In particular, token classification currently aggregates special-token positions that the Hugging Face pipeline filters out.

I would be comfortable with either completing that tokenizer contract or narrowing this PR to the BERT-family sequence-classification behavior that is already validated. The PR description should also be updated from ort-server 0.3.3 to 0.3.5.

ramkrishna2910:
Thanks @fl0rianr — all four addressed, in **ort-server 0.3.6** (pin bumped) and the latest commits. @eddierichter-amd, thanks for the approval and for reproducing #1 last round.

**1. HF cache snapshot resolution — you undersold this; it was a real bug.** Resolution scanned the whole `models--…` cache, so an ordinary cache holding an old snapshot beside the current one was rejected as ambiguous. That breaks **any model the user has ever updated**, not an exotic layout. Worse, the codebase already had `hf_cache::active_snapshot_path()` (used by five other backends) and I didn't use it. HF caches now resolve inside the `refs/main` snapshot; recursive exactly-one resolution is kept for local imports, where it is the right rule. Verified against a two-snapshot cache — it picks the active revision and returns correct scores.

**2. Architecture check bypassable by a manifest-only artifact — correct.** A manifest describes the **output** contract and says nothing about the **input** convention (mask, segment ids, special tokens), which is exactly what this server hardcodes, so it cannot excuse a missing `config.json`. A readable config with an allowlisted `model_type` is now **always** required.

**3. Token classification aggregated special-token positions — correct, and fixing it exposed a second bug.** HuggingFace's pipeline filters `[CLS]`/`[SEP]` out of its output; we were scoring them. My first fix filtered *every* `added_token` marked special — which wrongly dropped **`[UNK]`**, an ordinary content position that HF *does* score (its `special_tokens_mask` marks it 0). The golden test caught that on the unicode fixture. The filtered set is now derived from the **post-processor template** — the tokens the tokenizer actually *inserts* — plus padding.

To give that test teeth, the token-cls fixture is rebuilt so special positions are the sole driver of one label: the old behavior now fails the golden comparison by **0.55**. Verified by negative control.

**4. Wording and version.** The "guaranteed parity" and "any HuggingFace classifier" claims are gone. The README now states plainly that the tokenizer returns **token ids only**, that the mask / segment ids / padding / special-token filtering are **reconstructed by this server** to match the HF pipeline, that this is verified against HF-computed reference scores on every release, and that it is valid **only** for the allowlisted single-sequence encoder families. PR description updated to 0.3.6.

I took the "complete the tokenizer contract" option rather than narrowing the PR, since the reconstruction is now verified against the HF reference for both sequence- and token-classification, including padded tokenizers and non-ASCII input.

## Verification

ort-server 0.3.6 green on all four platforms (contract + golden parity + the self-contained-binary dependency check, against the extracted archive). Locally, with the backend cache wiped so lemond downloads the real release: **9/9** integration checks.

For the record on the CI job that was red: it was not flaky. Every ort-server release since 0.2.0 imported `libssl-3-x64.dll`/`libcrypto-3-x64.dll` — cpp-httplib links OpenSSL if the **build** machine has it, so the archive ran wherever OpenSSL happened to exist and died with `STATUS_DLL_NOT_FOUND` on the self-hosted runner. Fixed in 0.3.5, and release CI now enforces a dependency allowlist against the packaged binary so a non-self-contained archive cannot ship again.

fl0rianr (commented):
I realize I’m becoming the endless-critique guy here, so I re-checked these with a deliberately higher bar before raising another round.

The active-snapshot fix and the mandatory architecture validation both look correct. I do not think the remaining tokenizer edge cases should become another broad blocking round for the two catalog sequence classifiers.

There is, however, one concrete contract mismatch that should be cleaned up before merge:

Lemonade still considers `model.onnx + tokenizer.json + (manifest.json or config.json)` a complete model, while ort-server 0.3.6 now always requires a valid `config.json`, including when a manifest is present. This can make Lemonade select a model layout that the installed backend will immediately reject, or count an invalid manifest-only sibling when checking ambiguity.

Please align `is_complete_model_dir()` and its error message with the new backend contract:

`model.onnx + tokenizer.json + config.json`

with `manifest.json` remaining an optional output-contract override.

The PR description should also be updated: it still says that any Hugging Face classifier works with guaranteed parity, whereas the ort-server README now correctly limits parity to the validated single-sequence encoder convention.

The position-based tokenizer metadata issue—special-token masks, left padding, and related custom-tokenizer behavior—is real, but I would treat that as a documented follow-up rather than another blocker here, provided the public scope no longer promises universal Hugging Face compatibility.

Once the small Lemonade-side contract mismatch is fixed, the description is aligned, I’m comfortable approving this.

fl0rianr (commented):
The Lemonade/ort-server directory contract is now aligned correctly, and the updated scope wording is much better.

Since I’m becoming the endless-critique guy here, I found one remaining concrete issue in the advertised token-classification support:

### Stock RoBERTa tokenizers still include special tokens in aggregation

The API currently says that token classification and RoBERTa are supported and that a stock Optimum export works as-is.

However, Transformers constructs stock RoBERTa fast tokenizers using `RobertaProcessing`. Its tokenizer JSON stores the inserted tokens in direct `cls` and `sep` fields:

```json
{
  "type": "RobertaProcessing",
  "sep": ["</s>", 2],
  "cls": ["<s>", 0]
}
```

ort-server 0.3.6 only extracts special-token IDs from `post_processor.special_tokens`, which is the `TemplateProcessing` representation used by the current tiny fixture. With `RobertaProcessing`, `special_ids_` remains empty, so `<s>` and `</s>` are still included in the token-classification max/mean aggregation even though the Hugging Face pipeline filters those positions.

This is therefore not only a hypothetical custom-tokenizer edge case: it affects a stock tokenizer form for an architecture explicitly listed as supported.

Please either expose the tokenizer’s real `special_tokens_mask` through the FFI, or minimally handle `RobertaProcessing` and `BertProcessing` explicitly and reject unknown post-processors for token-classification. A golden fixture using `RobertaProcessing` would prevent this from regressing.

The PR description should also avoid saying unsupported tokenizer conventions are refused by the architecture gate unless the tokenizer configuration itself is actually validated; `model_type` alone cannot detect left padding or a custom post-processor.

Aside from this token-classification gap, the previously requested contract fixes look resolved. The two shipped BERT/DistilBERT sequence classifiers appear ready once the current CI run finishes.

ramkrishna2910:
Thanks @fl0rianr — both done, and I appreciate the higher-bar re-check on the previous two.

**Contract mismatch — fixed.** You're right that this was a real hole, not cosmetic: lemond's `is_complete_model_dir()` accepted `model.onnx + tokenizer.json + (manifest.json **or** config.json)`, while ort-server 0.3.6 now *always* requires a valid `config.json`. So lemond could select — or count toward an ambiguity decision — a manifest-only sibling the backend would immediately reject. `is_complete_model_dir()` and its error message now require:

`model.onnx + tokenizer.json + config.json`

with `manifest.json` remaining an optional output-contract override. No new ort-server release — the 0.3.6 pin stands; this is lemond-side only.

I verified it as a **negative control**, not just a happy path. On the real cached model, with `config.json` hidden from the active snapshot (manifest still present), the load is now refused up front:

> No servable model directory under '…\snapshots\48e24f56…': need model.onnx + tokenizer.json + config.json (manifest.json is optional and overrides the output contract)

— instead of launching a backend that dies opaquely. Restoring `config.json`, the model loads and returns scores matching the HF reference exactly (`LABEL_0` 0.7474327 vs 0.747434). Both catalog models ship `config.json`, so the stricter rule doesn't affect them. Full suite **9/9**.

**Description — aligned.** The "any HuggingFace classifier" / "guaranteed parity" wording is gone. It now states the supported single-sequence encoder families, that the tokenizer returns **token ids only** and the mask / segment ids / padding / special-token filtering are **reconstructed by ort-server** to match the HF pipeline, and that this is verified against HF reference scores on every release but valid only for those families — matching the ort-server README.

**Position-based tokenizer metadata** — agreed it's a documented follow-up, not a blocker, now that the public scope no longer promises universal HF compatibility. Tracked as [ort-server#1](https://github.com/lemonade-sdk/ort-server/issues/1) (left padding, custom post-processors, pair inputs; the fix is to surface the encoding metadata through the FFI rather than reconstruct it), and it's in the PR's known-limitations section. The load-time architecture allowlist keeps anything outside the verified convention refused rather than mis-served.

## Verification

Commit `b6ac960d`. Full CI matrix green — **96 pass, 0 fail** (10 skipped), including `Test .exe - classify-onnxruntime`.

fl0rianr:
Thanks — the `config.json` contract fix, negative control, updated description, and green CI address the two points I raised. Those parts look good.

I initially agreed that the broader encoding-metadata work could remain a follow-up. After checking the stock tokenizer serialization more closely, though, there is one important distinction:

The architecture allowlist does not itself guarantee that an unsupported tokenizer convention is refused. RoBERTa is on the allowlist, while a stock RoBERTa tokenizer can use `RobertaProcessing`, which stores its inserted tokens in direct `cls` and `sep` fields rather than `post_processor.special_tokens`. In that case the model is accepted, but those special-token positions can remain in token-classification aggregation.

So I do not think the full FFI metadata work needs to block this PR, but one of these should happen before approval:

1. Handle `RobertaProcessing` and `BertProcessing`, ideally with a focused token-classification test; or
2. Narrow the documented token-classification scope so stock RoBERTa compatibility is not currently promised.

The two shipped BERT/DistilBERT sequence-classification models are not affected. Once the advertised scope matches the verified behavior—or this small tokenizer case is handled—I’m happy to approve.

ramkrishna2910:
Good catch — and you're right that the allowlist alone doesn't close this. RoBERTa is allowlisted, but a stock RoBERTa tokenizer serializes its inserted `<s>`/`</s>` as a **`RobertaProcessing`** post-processor, which stores them in `cls`/`sep` fields — not the `TemplateProcessing` `special_tokens` map my filter read. So the model loaded and its `<s>`/`</s>` positions stayed in token-classification aggregation, which the HF pipeline drops. Verified behavior didn't match advertised scope.

I took option 1 — **handle the post-processor**, not narrow the scope — since token-cls is then genuinely covered for RoBERTa.

**Fix (ort-server 0.3.7).** The special-token collector now parses all three HF post-processor shapes, and recurses through a `Sequence` wrapper:

- `TemplateProcessing` → `special_tokens[*].ids` (BERT/DistilBERT)
- `BertProcessing` / `RobertaProcessing` → the `cls`/`sep` fields
- `Sequence` → each nested processor

**Test with teeth.** New golden fixture `tiny-roberta-tok` — a real `RobertaProcessing` tokenizer (built via the `tokenizers` lib, `post_processor.type == "RobertaProcessing"`, no `special_tokens` map) whose `<s>`/`</s>` are the *sole* driver of one label. I confirmed it isn't a structural pass-through by **negative control**: with the `cls`/`sep` handling disabled and the binary rebuilt, `tiny-roberta-tok` fails the golden by **0.70**, while the existing BERT `TemplateProcessing` golden still passes. With the fix, max delta is 5.96e-08.

**Scope wording** in the PR description now names this explicitly: the special-token mask covers those three post-processor shapes and is verified per-release; the remaining tokenizer deviations (left padding, fully custom post-processors, pair/two-sequence inputs) are out of scope and tracked in [ort-server#1](https://github.com/lemonade-sdk/ort-server/issues/1), with the load-time allowlist keeping anything outside the verified convention refused rather than mis-served.

## Verification

- **ort-server 0.3.7** released green on all four platforms; the packaged-archive smoke suite now includes the RoBERTa golden.
- **End-to-end**: backend cache wiped so lemond downloads the real 0.3.7 release (`ort-server-0.3.7-windows-x64.zip`) — **9/9** `server_classify.py`, including the HF-reference golden.
- lemond pin bumped to 0.3.7 (`14c2ceb5`); `Test .exe - classify-onnxruntime` green. **All 21 required checks green.**

(The three red jobs are non-required ARM64 Ubuntu-container builds failing at `apt-get install` of base tooling — `git`/`cmake`/`nodejs` "Unable to locate package" — an apt-mirror outage at the container-provisioning step, reproduced identically on re-run. The actual ARM64 compile (`Build Lemonade (Linux ARM64)`) and all Fedora/dnf ARM64 jobs pass; a one-line version-pin can't affect apt. Unrelated to this change.)

fl0rianr (approved):
I guess CI is not related here...

Thanks for all this hard work, finally - approve.

PR #2627, opened 2026-07-09
fix(security): replace system() with ProcessManager in archive extrac…

superm1:
…tion

Fixes shell command injection vulnerability in extract_zip(), extract_tarball(), and extract_seven_zip() by replacing system() calls with ProcessManager::run_process_with_output() using proper argument vectors instead of string concatenation.

Changes:
- archive_unix.cpp: Use ProcessManager with argv for unzip and tar
- archive_windows.cpp: Use ProcessManager with argv for tar and PowerShell
- backend_utils.cpp: Use ProcessManager with argv in extract_seven_zip()

This prevents arbitrary shell command execution via crafted archive paths (e.g., paths containing "; touch /tmp/pwned;").

### DIFF
--- a/src/cpp/include/lemon/utils/process_manager.h
+++ b/src/cpp/include/lemon/utils/process_manager.h
@@ -32,7 +32,8 @@ public:
         const std::vector<std::string>& args,
         OutputLineCallback on_line,
         const std::string& working_dir = "",
-        int timeout_seconds = -1);
+        int timeout_seconds = -1,
+        bool capture_stderr = true);
 
     static void stop_process(ProcessHandle handle);
 
--- a/src/cpp/include/lemon/utils/process_platform.h
+++ b/src/cpp/include/lemon/utils/process_platform.h
@@ -43,7 +43,8 @@ public:
         const std::vector<std::string>& args,
         OutputLineCallback on_line,
         const std::string& working_dir,
-        int timeout_seconds) = 0;
+        int timeout_seconds,
+        bool capture_stderr = true) = 0;
 
     // Utility functions
     virtual int find_free_port(int start_port) = 0;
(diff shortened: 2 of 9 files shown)


fl0rianr (changes requested):
I think the PowerShell fallback still leaves a command-injection path open. Although powershell.exe is now launched through an argument vector, zip_path and dest_dir are interpolated into the script passed to `-Command`:

```cpp
"Expand-Archive -Path '" + zip_path +
"' -DestinationPath '" + dest_dir + "' -Force"
```

A path containing a single quote can terminate the PowerShell string literal and inject additional PowerShell code. It also breaks valid paths such as `C:\Users\O'Connor\...`.

Could we pass the paths as actual PowerShell parameters instead of embedding them in the command string? At minimum, the values should be escaped as PowerShell single-quoted literals (`'` → `''`) and `-LiteralPath` should be used instead of `-Path`.

fl0rianr (commented):
One remaining edge case: is_native_tar_available() now calls run_process_with_output() directly, but the Windows implementation throws std::runtime_error when CreateProcessA() cannot start the executable.

This means that when tar.exe is actually unavailable—the condition this method is intended to report—it throws instead of returning `false`. In extract_zip(), that prevents the PowerShell fallback from being reached.

Could we catch the process-start exception inside is_native_tar_available() and return `false`? That would also preserve the expected semantics of the boolean availability check for the tarball and 7z paths.

fl0rianr (commented):
There is one remaining behavior change in the tarball listing path. run_process_with_output() merges stderr into stdout, so `entries` now contains both archive members and any tar diagnostics.

That matters because `tar -tf` can exit successfully while still emitting warnings—for example, when ignoring an unknown PAX extended-header keyword. Such a warning line is then passed to compute_tarball_strip_components() as if it were an archive entry and can incorrectly force `strip = 0`, leaving an otherwise valid single-root archive nested one directory too deep.

The previous implementation explicitly discarded stderr for this listing. Could we preserve stdout-only capture here, rather than parsing the merged output? Filtering particular warning prefixes would be fragile and platform-dependent.

fl0rianr (approved):
Assuming green CI, LGTM now.

superm1:
CI failure looks like HF outage.  Will let some jobs finish and then retry them.

Issue #2623, opened 2026-07-09
AceStep ROCM generation error + Override of AceStep Vulkan


korgano:
### Platform

Windows

### Lemonade Version

v10.10.0

### GPU / APU Model

AMD Ryzen AI 9 HX 370

### Component

Other / Not sure

### Bug Description

Updated to v10.10 and decided to try out Ace-Step generation on my HX 370 miniPC. Generation failed with a HTTP 500 error, so I uninstalled the ROCM Ace-Step backend and installed the Vulkan one.

However, attempting to generate the track caused Lemonade to download the ROCM backend, when resulted in me getting the same 500 error.

### Steps to Reproduce

1. Install Lemonade v10.10.0
2. Download Ace-Step model.
3. Install Ace-Step Vulkan backend
4. Attempt to run Ace-Step prompt
5. Lemonade will download Ace-Step ROCM backend
6. Generation will be attempted
7. Generation fails with HTTP 500 error

### Log Output

```shell
2026-07-09 14:05:19.462 [Debug] (Server) GET /api/v1/system-info
2026-07-09 14:05:19.465 [Debug] (Server) OPTIONS /api/v1/load
2026-07-09 14:05:19.465 [Debug] (Server) OPTIONS /api/v1/load - 204
2026-07-09 14:05:19.466 [Debug] (Server) POST /api/v1/load
2026-07-09 14:05:19.466 [Debug] (Server) ===== LOAD ENDPOINT ENTERED (Thread: 28156) =====
2026-07-09 14:05:19.466 [Info] (Server) Ensuring model loaded: ACE-Step-Music 
2026-07-09 14:05:19.467 [Debug] (Router) Loading model: ACE-Step-Music (checkpoint: Serveurperso/ACE-Step-1.5-GGUF:acestep-v15-xl-sft-Q8_0.gguf, recipe: acestep, type: audio-generation, device: gpu)
2026-07-09 14:05:19.467 [Warn] (Router) Pruning unavailable backend for model: ACE-Step-Music (state=watchdog_reset)
2026-07-09 14:05:19.467 [Info] (Router) Evicting model: ACE-Step-Music
2026-07-09 14:05:19.467 [Info] (Router) Evicted model: ACE-Step-Music
2026-07-09 14:05:19.672 [Debug] (AutoTune) get_available_memory_gb: could not determine memory, returning 0.0resolve_auto_ctx_size: ACE-Step-Music — memory undetectable, returning 4096
[...]

PR #2629, opened 2026-07-09
fix(acestep): music without lyrics isn't the greatest

pwilkin:
I was so focused on generating *instrumental* music I completely forgot to wire the actual lyrics pipeline :) now you can have lyrics as well in your music (+ added syntax guide for the AceStep lyrics prompting).

Also fixes for #2623 since this touches the AceStep engine:
- Backend auto-selection prefers an installed variant over downloading the preference-order winner (#2623).
- Gate acestep/thinksound/trellis ROCm rows to the GPU families the engine releases are built for (#2623).

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -703,12 +703,25 @@ This endpoint is not part of the OpenAI API (OpenAI's audio endpoints cover spee
 | Parameter | Required | Description |
 |-----------|----------|-------------|
 | `model` | Yes | The audio-generation model to use (e.g., `ThinkSound-SFX`, `ACE-Step-Music`). |
-| `prompt` | Yes | Text description of the music or sound effect to generate. |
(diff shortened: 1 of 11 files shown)


fl0rianr (commented):
Thanks for the fix — the overall direction looks good to me. The ACE-Step vocal path is much clearer now, and the `/lm` → `/synth` flow matches the implementation notes well.

I found two small things I think should be tightened before merge:

1. The API now exposes `vocal_language`, and the lyrics help text says lyrics do not have to be English, but the UI never sends vocal_language. As a result, all UI vocal generations fall back to the backend default "en". Could we either add a compact language field for ACE-Step music, or adjust the UI copy so it does not imply language selection is supported from the UI yet?

2. lyrics and vocal_language should probably be validated in handle_audio_generations before model load. Right now a non-string `lyrics` value is silently ignored by the ACE-Step backend and becomes an instrumental request. For the public API, lyrics: 123 / non-string vocal_language should return a 400 instead of loading the model and changing behavior silently.

fl0rianr (approved):
LGTM now, CI ARM fail unrelated.

PR #2630, opened 2026-07-09
fix(server): harden API auth surface against drive-by RCE

superm1:
Plugs a drive-by chain where any web page could rewrite a backend's executable path and trigger local code execution via the unauthenticated `POST /params` endpoint. Enforces a server-side Origin allow-list — shared by the HTTP CORS path and the WebSocket upgrade handler — and locks down privileged config keys, without affecting the CLI, admin endpoints, or the same-origin web-app.

Key changes:

- **Origin allow-list, enforced before handlers** — non-OPTIONS requests with an unrecognized Origin get a 403; `Access-Control-Allow-Origin` is reflected only for allow-listed origins (loopback, Tauri desktop, and configured origins), replacing the `*` wildcard.
- **Shared allow-list helper** — `lemon::utils::is_origin_allowed()` (new `origin_utils.{h,cpp}`) is now used by both the HTTP CORS path and the WebSocket upgrade handler, replacing the WebSocket's Origin-host-equals-Host-header fallback, which was not DNS-rebinding-safe (an attacker-controlled name resolving to loopback/LAN sends a matching Host header).
- **`allowed_origins` config** — new array in `config.json` for non-loopback web-app access (e.g. `http://192.168.1.50:13305`).
- **`POST /params` lockdown** — `RuntimeConfig::set()` gains an `allow_privileged_keys` parameter; `/params` passes `false` so `*_bin` / `*_args` (which control the executable path and command line) are rejected. `/internal/set`, CLI, and admin paths pass `true`.
- **Tests** — `/params` rejection of bin/args overrides and acceptance of benign keys, CORS allow/deny for loopback vs third-party origins, pre-handler 403 on `/internal/set`, and a WebSocket DNS-rebinding regression test.

Refs: SWSPLAT-24170, SWSPLAT-24172


fl0rianr (changes requested):
Good, but I would like to request changes: the `/params` privileged-key guard is good, but the CORS hardening is not sufficient as a drive-by write defense.

Right now disallowed Origins only lose the `Access-Control-Allow-Origin` response header. That blocks preflighted browser requests, but not simple POSTs. A malicious page can still send a state-changing request with a safelisted content type such as `text/plain` and a JSON body; the browser may hide the response, but the server-side handler still runs.

This is especially important because `/internal/set` remains registered and, in the default no-key setup, internal endpoints are unauthenticated. Since the admin/config path keeps `allow_privileged_keys=true`, a simple cross-origin POST to `/internal/set` could still attempt to set `*_bin` / `*_args`.

Please reject disallowed `Origin` server-side before dispatching to handlers, e.g. in the pre-routing handler return 403 for non-OPTIONS requests when `Origin` is present and not allow-listed. Also add a regression test using `Origin: http://evil.example` + `Content-Type: text/plain` against a mutating endpoint and assert the config does not change.

fl0rianr (commented):
The original simple-POST security issue is now fixed correctly: disallowed Origins are rejected before handler dispatch, and the regression test verifies that `/internal/set` does not mutate configuration.

However, the new global Origin check introduces a same-origin regression for the bundled web UI when Lemonade is accessed through a LAN address or hostname.

Browsers include an `Origin` header on POST requests even when they are same-origin. The bundled web app uses `window.location.origin` as its API URL, but `is_origin_allowed()` only accepts loopback literals and the Tauri origins. Therefore, opening the UI at an address such as `http://192.168.1.50:13305` results in every POST request being rejected with 403.

This matters because Lemonade supports non-loopback bindings, including the documented `--host 0.0.0.0` Docker configuration. The page will load, but chat, model loading, pulls, settings, and other POST-based operations will fail.

Please preserve legitimate non-loopback web-app access through an explicit configured Origin allow-list or another DNS-rebinding-safe mechanism. Simply allowing `Origin` when it matches `Host` would not be sufficient because that can reintroduce DNS-rebinding exposure.

Please also add a regression test covering an explicitly permitted non-loopback Origin.

Non-blocking hardening: consider removing the `allow_privileged_keys=true` default so every caller must explicitly select the privileged or restricted configuration surface.

fl0rianr (commented):
Thanks for the thorough follow-up — the HTTP side now looks much stronger, and the earlier `/params` and simple-POST issues appear to be addressed correctly.

I found one remaining gap: WebSocket upgrades on `/realtime` and `/logs/stream` use a separate Origin check that still allows the request when `Origin` host matches the `Host` header. That is not DNS-rebinding-safe, because a rebinding origin such as `evil.example` would also send `Host: evil.example:<port>` after resolving to localhost or a LAN address.

The WebSocket path also does not use the new `allowed_origins` configuration, so the HTTP and WebSocket policies can behave differently.

Could we reuse the same Origin allow-list logic for WebSockets and remove the `origin_host == request_host` fallback? A regression test with matching `Origin` and `Host`, but a connection to loopback, would cover the rebinding case well.

Apart from that, the direction and current HTTP implementation look good.

fl0rianr (commented):
Thanks for the follow-up — the core security fixes look solid now, especially the pre-handler Origin check, the configured non-loopback allow-list, and sharing the same policy with WebSockets.

I think there are still a few regressions worth fixing before merging:

* `Access-Control-Allow-Headers` now drops MCP headers such as `MCP-Protocol-Version` and `Mcp-Session-Id`, which breaks the existing browser MCP CORS contract.
* The new Origin helper also removes previously supported desktop/custom origins like `file://`, `app://.` and `jan://app`, and no longer honors `LEMONADE_ALLOWED_ORIGINS` (including explicit `null`).
* Disallowed preflights now return `204`, while the existing MCP tests expect `403`. Either behavior can be made safe, but implementation and tests should agree.
* I’d also prefer keeping/updating the Origin utility unit tests rather than removing them entirely, since this is security-sensitive parsing/policy code.

Otherwise the overall direction looks good. Once those compatibility issues are addressed, I think this is in good shape.

fl0rianr (commented):
Thanks for the updates — the preflight handling and desktop/custom-scheme origins look good now.

I think there are just two remaining compatibility issues before this is ready:

* `Access-Control-Allow-Headers` still drops `MCP-Protocol-Version`, `Mcp-Session-Id`, and `traceparent`, which breaks the existing MCP browser CORS behavior/tests.
* `LEMONADE_ALLOWED_ORIGINS` no longer appears to be honored, so existing setups such as explicitly allowing `Origin: null` via the env var will regress.

CLI endpoint test are failing.

Non-blocking: I’d still prefer keeping/updating the Origin utility unit tests rather than removing them, given this is security-sensitive code.

Otherwise this looks in good shape.

fl0rianr (commented):
Thanks, this is much closer. The previous HTTP/WS Origin issues, MCP CORS headers, env allow-list support, and Origin tests all look addressed now.

I still see a few things to fix before merge:

* `*` is still accepted by `is_origin_allowed()` and explicitly tested, while the docs now say wildcards are not supported. Since `*` effectively disables the new drive-by protection, implementation and docs should agree here.
* Configured origins are now compared as raw strings, which drops the previous normalization behavior. For example, `https://example.com` and `https://example.com:443` used to match. This can break existing `LEMONADE_ALLOWED_ORIGINS` setups.
* `*.localhost` support was also removed, despite being explicitly supported/tested before. Please restore it or document this as an intentional compatibility break.

Non-blocking hardening: I’d still prefer `allow_privileged_keys=false` as the default so privileged config writes must opt in explicitly.

Otherwise the security direction looks good. Good luck with CI!

superm1:
Origins changed, dropping.

PR #2632, opened 2026-07-10
security(renderer): redact API key values from console logs

superm1:
Remove cleartext API key logging in setUpdatedAPIKey() to prevent credential disclosure via renderer console output forwarded to log files.


fl0rianr (approved):
Good! LGTM.

PR #2633, opened 2026-07-10
security: pin FetchContent deps to commit SHAs

superm1:
Replace mutable git tags with immutable commit SHAs for all seven FetchContent dependencies (brotli, nlohmann/json, CLI11, curl, zstd, cpp-httplib, libwebsockets) to prevent supply-chain attacks via tag re-pointing.

Fixes: SWSPLAT-24212


fl0rianr (approved):
LGTM, thanks.

PR #2635, opened 2026-07-10
Update build-and-push-container.yml

betonme:
Added platforms to specify additional linux/arm64

### DIFF
--- a/.github/workflows/build-and-push-container.yml
+++ b/.github/workflows/build-and-push-container.yml
@@ -69,6 +69,7 @@ jobs:
         with:
           context: .
           file: Dockerfile
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: |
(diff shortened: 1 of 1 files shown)


fl0rianr:
Thanks for starting the ARM64 container work. The direction is useful, but the current change directly modifies the production publishing workflow, where an ARM64 build failure could also interrupt our AMD64 edge and release images.

Since safely introducing ARM64 requires a larger CI change, I’m going to close this PR and follow up with a maintainer PR that:

* builds AMD64 and ARM64 independently on native runners,
* performs container and server smoke tests,
* does not modify or publish production tags,
* provides an ARM64 test image as a workflow artifact.

I did credit you as a co-author and reference this PR as the starting point.

It would also be very helpful if you could test the generated ARM64 artifact on your actual device, including a small llama.cpp CPU inference and, if applicable, Vulkan. Please include the device, distribution, `uname -m`, and any relevant logs in your results.

Thanks again for pushing this forward.

PR #2636, opened 2026-07-10
fix(security): move WebSocket API key from query string to Sec-WebSoc…

superm1:
The realtime transcription WebSocket sent the API key in the URL query string (`ws://host:port/realtime?api_key=...`) and logged the full URL, exposing the key in server access logs, libwebsockets handshake logs, browser history (web-app mode), and to on-path observers over cleartext `ws://`.

The credential now rides in the `Sec-WebSocket-Protocol` header. Because libwebsockets rejects unregistered subprotocols during negotiation (before the auth callback runs), the client must also offer the registered application protocol — otherwise the upgrade fails whenever `LEMONADE_API_KEY` is set.

Client (TypeScript):
- Remove the API key from the WebSocket URL (`buildWebSocketUrl` no longer appends `api_key`).
- Add a shared `webSocketProtocols()` helper offering `[lemonade-realtime, bearer.<base64url(key)>]`; base64url keeps the key within the subprotocol token charset, and awaiting config init fixes a race where the key was read before async init finished.
- Both the realtime (`websocketClient.ts`) and log-stream (`logWebSocketClient.ts`) clients pass these protocols to `new WebSocket()`; previously `/logs/stream` sent no credential at all.
- Redact query parameters from console logs.

Server (C++):
- `get_protocol_credential()` reads the full `Sec-WebSocket-Protocol` header, extracts the `bearer.` credential, and base64url-decodes it.
- Wired into both `extract_token_from_wsi()` and `authenticate_connection()`, after the `Authorization` header and `api_key` query param (backward compatible).

Tests:
- New `test/server_websocket_auth.py` covers authenticated `/v1/realtime` and `/logs/stream` on both the dedicated WebSocket port and the main HTTP port.

The fix applies to both the Tauri desktop app and the web app, which share the same renderer code.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -716,6 +716,9 @@ jobs:
           echo "Running WebSocket idle test (31s regression)..."
           .venv/bin/python test/test_websocket_idle.py
           echo "WebSocket idle test PASSED!"
+          echo "Running WebSocket auth tests..."
+          .venv/bin/python test/server_websocket_auth.py
+          echo "WebSocket auth tests PASSED!"
 
       - name: Run web-app path traversal tests (unsigned path)
         if: steps.check_signing.outputs.has_signing != 'true'
@@ -2077,6 +2080,9 @@ jobs:
           echo "Running WebSocket idle test..."
           .venv/bin/python test/test_websocket_idle.py
           echo "WebSocket idle test PASSED!"
+          echo "Running WebSocket auth tests..."
+          .venv/bin/python test/server_websocket_auth.py
+          echo "WebSocket auth tests PASSED!"
 
       # The full smoke test can be added to a different GPU equipped pipeline
       # as well if needed, but it would significantly add to the pipeline runtime.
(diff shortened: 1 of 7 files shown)


fl0rianr (changes requested):
Thanks for addressing this — moving the API key out of the URL is definitely the right direction.

I found two blocking regressions that should be fixed before merging:

1. The realtime connection offers only `bearer.<apiKey>` as the WebSocket subprotocol, while libwebsockets registers only lemonade-realtime. libwebsockets selects a supported protocol before LWS_CALLBACK_FILTER_PROTOCOL_CONNECTION, so the upgrade is rejected before the new authentication logic can run.

2. buildWebSocketUrl() no longer adds authentication for any caller, but `logWebSocketClient.ts` still creates the socket without a subprotocol or other credentials. This means /logs/stream will fail whenever API-key authentication is enabled.

There is also a smaller initialization issue: getAPIKey() is read before serverConfig has necessarily completed its async initialization, so the fallback connection can retain an empty key.

I think the cleanest approach would be a shared WebSocket authentication helper used by both realtime and log streaming. The client could offer the registered application protocol plus a safely encoded credential protocol, while the server reads and parses the full Sec-WebSocket-Protocol header rather than relying on the callback’s selected protocol argument.

Could we also add authenticated integration coverage for both `/v1/realtime` and `/logs/stream`? The existing tests still exercise the query-parameter path, so CI currently does not catch these regressions.

fl0rianr (approved):
LGTM now, approve.

Before merging please update PR description if possible.

Also you might want to link the test with a workflow?

PR #2638, opened 2026-07-10
fix: handle AMD APU memory pressure correctly

### LINKED ISSUE #2631: VRAM pressure monitor ignores GTT on AMD APU systems (Strix Halo) — false-positive evictions at 99% of 2GB VRAM while 98GB GTT is free
## Summary

**The VRAM pressure monitor only checks dedicated VRAM, ignoring GTT entirely. On AMD APU systems this causes false-positive evictions.**

On Strix Halo (Ryzen AI MAX+ 395), the memory topology is split:

| Memory pool | Size | What lives here | Usage at idle |
|---|---|---|---|
| **Dedicated VRAM** (BIOS carveout) | **2 GB** | amdgpu driver/kernel overhead | **~98.9% full** (even headless, no display server) |
| **GTT** (shared system memory) | **124 GB** | Model weights (llama.cpp, ACE-Step, etc.) | **~32% used** |

The eviction monitor reads `mem_info_vram_used / mem_info_vram_total` — **the 2 GB pool only**. It sees ~98.9% and compares against the 99% threshold. When amdgpu driver activity pushes VRAM to 99%, the eviction engine fires and **unloads running models** — even though **85+ GB of GTT is completely free**.

> ⚠️ **The `enable_dgpu_gtt` config flag does NOT fix this.** It only affects model *filtering* (size checks in `model_manager.cpp`), not the VRAM pressure monitor. See [source code evidence](#the-bug--source-code-evidence) below.

---

## Hardware & Environment

<details>
<summary><b>📋 System specs (click to expand)</b></summary>

| | |
|---|---|
| **System** | GMKtec EVO-X2 AI Mini PC |
| **APU** | AMD Ryzen AI MAX+ 395 w/ Radeon 8060S (Strix Halo) |
| **GPU PCI ID** | `1002:1586` (rev c1) — Display controller |
| **GFX version** | gfx1151 (RDNA 4) |
| **System RAM** | 128 GB LPDDR5 (8x 16 GB, 8000 MT/s) |
| **BIOS** | American Megatrends EVO-X2 1.12 (2025-12-09) |
| **BIOS UMA VRAM carveout** | **2 GB** dedicated VRAM |
| **GTT (shared memory)** | **124 GB** (`amdgpu.gttsize=126976` kernel param) |

</details>

<details>
<summary><b>🖥️ Software & kernel (click to expand)</b></summary>

| | |
|---|---|
| **OS** | Ubuntu 24.04.4 LTS (Noble Numbat) |
| **Kernel** | 6.17.0-35-generic |
| **Lemonade Server** | 10.10.0 |
| **llamacpp backend** | vulkan |
| **Display** | **Headless** (SSH only — no display server, no compositor, no Xorg/Wayland). VRAM is still ~98.9% full from amdgpu kernel/driver overhead alone. |

**Kernel boot parameters:**
```
amdgpu.gttsize=126976 ttm.pages_limit=32505856 transparent_hugepage=always numa_balancing=disable amd_iommu=pt amdgpu.sg_display=0 amdgpu.lockup_timeout=1000000,1000000
```

**systemd lemond environment:**
```ini
Environment="HSA_OVERRIDE_GFX_VERSION=11.5.1"
Environment="MIOPEN_GEMM_WORKSPACE_FACTOR=16"
Environment="GPU_MAX_ALLOC_PERCENT=100"
Environment="GPU_MAX_HEAP_SIZE=100"
Environment="HF_HOME=/var/lib/lemonade/hf-cache"
Environment="HF_HUB_CACHE=/var/lib/lemonade/hf-cache/hub"
Environment="ROCBLAS_USE_HIPBLASLT=1"
Environment="GGML_CUDA_ENABLE_UNIFIED_MEMORY=1"
```

</details>

<details>
<summary><b>⚙️ Lemonade config (click to expand)</b></summary>

```json
{
  "auto_evict": false,
  "auto_evict_threshold_pct": 0.99,
  "enable_dgpu_gtt": true,
  "max_loaded_models": -1
}
```

> **Note:** `auto_evict` was `true` when the bug was observed; it has been set to `false` as a workaround.

</details>

---

## Memory topology (sysfs readings)

Both models loaded (Qwen 35B + ACE-Step), measured directly from `/sys/class/drm/card0/device/`:

```
mem_info_vram_total:     2,147,483,648 bytes (2,048 MB)   ← BIOS carveout
mem_info_vram_used:      2,123,714,560 bytes (2,025 MB)   ← 98.9% full (amdgpu driver/kernel)
mem_info_gtt_total:    133,143,986,176 bytes (126,976 MB)  ← shared system memory
mem_info_gtt_used:      42,660,007,936 bytes (40,682 MB)   ← 32.0% used (model weights)
```

**The 2 GB VRAM is at 98.9% from amdgpu kernel/driver overhead alone — this system is headless (no display server, no compositor).** No models are loaded into VRAM; all model weights go into GTT. The monitor sees only the 98.9% VRAM ratio and triggers eviction when it crosses 99%.

`rocm-smi` also reports only the 2 GB VRAM:
```
GPU[0]  VRAM Total Memory (B): 2147483648
GPU[0]  VRAM Total Used Memory (B): 2123710464
```

---

### 1. `global_vram_monitor.cpp` — delegates to `SystemInfo`

```cpp
// src/cpp/server/global_vram_monitor.cpp (lines 35-38)
double GlobalVramMonitor::poll_vram_usage() const {
    return SystemInfo::get_global_vram_usage_pct();
}
```

### 2. `system_info.cpp` — reads ONLY `mem_info_vram_*`, never `mem_info_gtt_*`

```cpp
// src/cpp/server/system_info.cpp — get_global_vram_usage_pct() (lines ~4008-4047)
#ifdef __linux__
    // AMD (and other DRM GPUs): read used/total from sysfs, taking the busiest card.
    try {
        const std::string drm_path = "/sys/class/drm";
        if (fs::exists(drm_path)) {
            double highest_ratio = -1.0;
            for (const auto& entry : fs::directory_iterator(drm_path)) {
                // ... skip non-card entries ...

                uint64_t vram_used = 0;
                uint64_t vram_total = 0;
                {
                    std::ifstream f(device_path + "/mem_info_vram_used");  // ← ONLY VRAM
                    if (f.is_open()) f >> vram_used;
                }
                {
                    std::ifstream f(device_path + "/mem_info_vram_total");  // ← ONLY VRAM
                    if (f.is_open()) f >> vram_total;
                }
                // ↑ mem_info_gtt_used / mem_info_gtt_total are NEVER read

                if (vram_total > 0) {
                    double ratio = static_cast<double>(vram_used) / static_cast<double>(vram_total);
                    // ↑ This is 98.6% on Strix Halo — always near 100% from display overhead
                    if (ratio > highest_ratio) {
                        highest_ratio = ratio;
                    }
                }
            }
            if (highest_ratio >= 0.0) {
                return highest_ratio;  // ← Returns 98.6%, not the true 33.9% combined
            }
        }
    } catch (...) {}
#endif
```

### 3. `enable_dgpu_gtt` only affects model filtering — NOT the monitor

```cpp
// src/cpp/server/model_manager.cpp (lines 2064-2067)
// This is the ONLY place enable_dgpu_gtt is used:
if (dev_type == "amd_igpu")
    dev_mem_alloc_behavior = MemoryAllocBehavior::Largest;
if (enable_dgpu_gtt)
    dev_mem_alloc_behavior = MemoryAllocBehavior::Unified;
// ↑ Changes model size filtering, does NOT touch get_global_vram_usage_pct()
```

### 4. The helpers to fix this already exist

The codebase already has everything needed:

| Helper | Location | What it does |
|---|---|---|
| `LinuxSystemInfo::get_amd_gtt()` | `system_info.cpp:3548` | Reads `mem_info_gtt_total` from sysfs |
| `LinuxSystemInfo::get_amd_is_igpu()` | `system_info.cpp:3521` | Detects iGPU vs dGPU via `board_info` |
| `GPUInfo::virtual_gb` | `system_info.cpp:3403` | Already stores GTT size during hardware detection |

---

## Eviction log evidence

With `auto_evict=true` and `auto_evict_threshold_pct=0.99`, the eviction engine fires repeatedly — **every 30-60 seconds**:

```
22:00:36 [Info] (on_vram_pressure) VRAM pressure critical (99.28% >= 99.00%). Evaluating eviction.
22:00:36 [Info] (evaluate_servers) Eviction Engine unloading model: ACE-Step-Music due to score/pressure/idle.
22:00:36 [Info] (Router) Evicting model: ACE-Step-Music
22:00:44 [Info] (Router) Evicted model: ACE-Step-Music
22:01:30 [Info] (on_vram_pressure) VRAM pressure critical (99.28% >= 99.00%). Evaluating eviction.
22:01:30 [Info] (evaluate_servers) Eviction Engine unloading model: ACE-Step-Music due to score/pressure/idle.
22:01:37 [Info] (Router) Evicted model: ACE-Step-Music
22:01:55 [Info] (on_vram_pressure) VRAM pressure critical (99.29% >= 99.00%). Evaluating eviction.
22:01:55 [Info] (evaluate_servers) Eviction Engine unloading model: ACE-Step-Music due to score/pressure/idle.
```

**At the time of these evictions:**
- **VRAM:** 99.28% (2,033 MB / 2,048 MB) — display + kernel overhead, no model weights
- **GTT:** ~20% (26 GB / 124 GB) — model weights, **98 GB free**

The model was evicted and re-evicted every ~30-60 seconds because VRAM never drops below 99% (display overhead is constant) and the model reloads each time it's requested.

---

## Impact

- 🔴 **Models get evicted mid-session or while idle** — causing 30-60s cold reloads on next request
- 🟡 **Intermittent** — VRAM fluctuates between 97.5% and 99.3% from amdgpu driver/kernel activity. On our headless system (no display server), VRAM is still at 98.9% — evictions happen seemingly randomly when driver overhead creeps to 99%
- 🟡 **`systemctl restart lemond` temporarily "fixes" it** — resets amdgpu VRAM state, but the problem returns within hours/days as VRAM creeps back up
- 🟢 **Pinning models** (`"pinned": true` on `/v1/load`) works as a workaround, but doesn't fix the root cause and doesn't help with idle-timeout eviction
- 🔴 **Affects all AMD APU systems** with small BIOS VRAM carveouts — not just Strix Halo. Any system where `mem_info_vram_total` is small and model weights live in GTT will hit this

---

### Option 1: Read GTT in `get_global_vram_usage_pct()` ⭐ recommended

On AMD, combine VRAM + GTT for the pressure ratio:

```cpp
uint64_t vram_used = 0, vram_total = 0;
uint64_t gtt_used = 0, gtt_total = 0;
// ... read mem_info_vram_used, mem_info_vram_total ...
// ... read mem_info_gtt_used, mem_info_gtt_total ...

if (vram_total > 0) {
    if (gtt_total > 0) {
        // APU: use combined ratio (VRAM + GTT)
        double ratio = static_cast<double>(vram_used + gtt_used)
                     / static_cast<double>(vram_total + gtt_total);
        if (ratio > highest_ratio) highest_ratio = ratio;
    } else {
        // dGPU without GTT — use VRAM only (existing behavior)
        double ratio = static_cast<double>(vram_used) / static_cast<double>(vram_total);
        if (ratio > highest_ratio) highest_ratio = ratio;
    }
}
```

**On our system this would report `(2.0 + 40.7) / (2.0 + 124.0) ≈ 33.1%` instead of `98.9%`** (exact value varies with model load — the point is it's well below the 99% threshold).

### Option 2: Auto-detect APU vs dGPU

Use the existing `get_amd_is_igpu()` helper. If the GPU is an iGPU (no `board_info` file), use GTT-inclusive monitoring. If it's a dGPU, use VRAM-only (existing behavior).

### Option 3: Respect `enable_dgpu_gtt` in the monitor

When `enable_dgpu_gtt=true`, have `get_global_vram_usage_pct()` include GTT in the calculation. Most backwards-compatible — users who set this flag are explicitly opting into GTT awareness.

---

## Workaround

```bash
lemonade config set auto_evict=false
```

Disables the VRAM pressure monitor entirely. Works when combined with `max_loaded_models=-1` (no LRU eviction) and model pinning. **Not ideal** — idle-timeout eviction is also lost.

---

## Loaded models at time of bug

| Model | Backend | Pinned | Port |
|---|---|---|---|
| Qwen3.6-35B-A3B-MTP-GGUF (Q4_K_XL, 35B MoE) | vulkan | true | 8001 |
| ACE-Step-Music (Q8_0) | vulkan | true | 8002 |

Both models load weights into GTT (~40 GB combined). The 2 GB VRAM holds only display/kernel buffers.

---

## Related issues

- **#722** — Physical RAM reported incorrectly on Linux (same Strix Halo hardware, GTT detection discussion)
- **#901** — Add AMD GTT detection support (PR that added `get_amd_gtt()` and `get_amd_is_igpu()` helpers)
- **#1648** — Drop AMD iGPU vs dGPU designations

fl0rianr:
## Summary

Fixes #2631.

The global VRAM pressure monitor now follows AMDGPU's APU allocation policy instead of treating a small UMA VRAM carveout as the only available GPU memory.

* Uses GTT pressure when an AMD APU is configured to prefer GTT.
* Uses VRAM pressure for larger UMA carveouts and discrete GPUs.
* Avoids double-counting VRAM and GTT on unified-memory systems.
* Evaluates every GPU and reports the highest per-device pressure on multi-GPU and mixed-vendor systems.
* Preserves the existing opt-in eviction behavior and safe fallbacks when sysfs data is unavailable.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -42,8 +42,10 @@
 
 #ifdef __linux__
 #include <dlfcn.h>
-#include <sys/ioctl.h>
 #include <fcntl.h>
+#include <libdrm/amdgpu.h>
+#include <libdrm/amdgpu_drm.h>
+#include <sys/ioctl.h>
 #include <unistd.h>
 #include "lemon/amdxdna_accel.h"
 #endif
@@ -153,6 +155,54 @@ struct HsaRuntimeApi {
     hsa_status_t (*amd_agent_iterate_memory_pools)(hsa_agent_t, HsaMemoryPoolCallback, void*) = nullptr;
     hsa_status_t (*amd_memory_pool_get_info)(hsa_amd_memory_pool_t, hsa_amd_memory_pool_info_t, void*) = nullptr;
 };
+
+bool query_amdgpu_is_apu(const fs::path& card_path) {
+    const fs::path drm_device_dir = card_path / "device" / "drm";
+    std::error_code ec;
+    if (!fs::is_directory(drm_device_dir, ec)) {
+        return false;
+    }
+
+    bool is_apu = false;
+    for (fs::directory_iterator it(drm_device_dir, ec), end;
+         it != end && !ec;
+         it.increment(ec)) {
+        const std::string name = it->path().filename().string();
+        if (name.rfind("renderD", 0) != 0) {
+            continue;
+        }
+
+        const fs::path node = fs::path("/dev/dri") / name;
+        const int fd = open(node.c_str(), O_RDWR | O_CLOEXEC);
+        if (fd < 0) {
+            continue;
+        }
+
+        uint32_t major_version = 0;
+        uint32_t minor_version = 0;
+        amdgpu_device_handle device = nullptr;
+        const int init_result = amdgpu_device_initialize(
+            fd, &major_version, &minor_version, &device);
+        if (init_result != 0 || device == nullptr) {
+            close(fd);
+            continue;
+        }
+
+        amdgpu_gpu_info info{};
+        const int query_result = amdgpu_query_gpu_info(device, &info);
+        amdgpu_device_deinitialize(device);
+        close(fd);
+
+        if (query_result != 0) {
+            continue;
+        }
+
+        is_apu = (info.ids_flags & AMDGPU_IDS_FLAGS_FUSION) != 0;
+        break;
+    }
+
+    return is_apu;
+}
 
 std::string trim_copy(const std::string& value) {
     const auto start = value.find_first_not_of(" \t\n\r");
@@ -3980,83 +4030,116 @@ bool SystemInfo::is_running_under_systemd() {
 }
 
 double SystemInfo::get_global_vram_usage_pct() {
-    // Report *global* GPU memory pressure (all processes, not just lemonade's),
-    // so the eviction engine yields VRAM when other apps (ComfyUI, games, etc.)
-    // consume it. Returns used/total in [0,1], or -1.0 if no source is available.
-    //
-    // Reuses the same detection sources as the rest of this file: nvidia-smi for
-    // NVIDIA (Linux + Windows) and AMD sysfs for Linux. macOS/Metal is unsupported
-    // for now and falls through to -1.0.
+    auto usage_ratio = [](uint64_t used, uint64_t total) {
+        if (total == 0) {
+            return -1.0;
+        }
+        return std::min(
+            1.0,
+            static_cast<double>(used) / static_cast<double>(total));
+    };
 
-    // NVIDIA: one query returns used + total for the first GPU.
-    {
-        if (!find_executable_in_path("nvidia-smi").empty()) {
-            std::string output;
-            int rc = lemon::utils::ProcessManager::run_command(
-                "nvidia-smi --query-gpu=memory.used,memory.total --format=csv,noheader,nounits",
-                output, 5);
-            if (rc == 0 && !output.empty()) {
-                std::istringstream iss(output);
-                std::string line;
-                if (std::getline(iss, line)) {
-                    size_t comma = line.find(',');
-                    if (comma != std::string::npos) {
-                        try {
-                            double used = std::stod(line.substr(0, comma));
-                            double total = std::stod(line.substr(comma + 1));
-                            if (total > 0.0) {
-                                return used / total;
-                            }
-                        } catch (...) {
-                            // fall through to other sources
-                        }
+    double highest_ratio = -1.0;
+
+    if (!find_executable_in_path("nvidia-smi").empty()) {
+        std::string output;
+        const int rc = lemon::utils::ProcessManager::run_command(
+            "nvidia-smi --query-gpu=memory.used,memory.total --format=csv,noheader,nounits",
+            output, 5);
+        if (rc == 0 && !output.empty()) {
+            std::istringstream iss(output);
+            std::string line;
+            while (std::getline(iss, line)) {
+                const size_t comma = line.find(',');
+                if (comma == std::string::npos) {
+                    continue;
+                }
+
+                try {
+                    const double used = std::stod(line.substr(0, comma));
+                    const double total = std::stod(line.substr(comma + 1));
+                    if (used >= 0.0 && total > 0.0) {
+                        highest_ratio = std::max(
+                            highest_ratio,
+                            std::min(1.0, used / total));
                     }
+                } catch (...) {
                 }
             }
         }
     }
+
 #ifdef __linux__
-    // AMD (and other DRM GPUs): read used/total from sysfs, taking the busiest card.
+    // Free GTT must not hide discrete-GPU VRAM exhaustion.
     try {
-        const std::string drm_path = "/sys/class/drm";
+        const fs::path drm_path = "/sys/class/drm";
         if (fs::exists(drm_path)) {
-            double highest_ratio = -1.0;
+            auto read_sysfs_u64 = [](const fs::path& path, uint64_t& value) {
+                std::ifstream file(path);
+                return file.is_open() && static_cast<bool>(file >> value);
+            };
+
             for (const auto& entry : fs::directory_iterator(drm_path)) {
-                std::string card_name = entry.path().filename().string();
-                if (card_name.rfind("card", 0) != 0 || card_name.find('-') != std::string::npos) {
+                const std::string card_name = entry.path().filename().string();
+                if (card_name.rfind("card", 0) != 0 ||
+                    card_name.find('-') != std::string::npos) {
                     continue;
                 }
-                std::string device_path = entry.path().string() + "/device";
+
+                const fs::path device_path = entry.path() / "device";
 
                 uint64_t vram_used = 0;
                 uint64_t vram_total = 0;
-                {
-                    std::ifstream f(device_path + "/mem_info_vram_used");
-                    if (f.is_open()) f >> vram_used;
-                }
-                {
-                    std::ifstream f(device_path + "/mem_info_vram_total");
-                    if (f.is_open()) f >> vram_total;
+                const bool have_vram =
+                    read_sysfs_u64(device_path / "mem_info_vram_used", vram_used) &&
+                    read_sysfs_u64(device_path / "mem_info_vram_total", vram_total) &&
+                    vram_total > 0;
+
+                const bool is_amd_apu = query_amdgpu_is_apu(entry.path());
+
+                double ratio = -1.0;
+                if (is_amd_apu) {
+                    uint64_t gtt_used = 0;
+                    uint64_t gtt_total = 0;
+                    const bool have_gtt =
+                        read_sysfs_u64(device_path / "mem_info_gtt_used", gtt_used) &&
+                        read_sysfs_u64(device_path / "mem_info_gtt_total", gtt_total) &&
+                        gtt_total > 0;
+
+                    uint64_t combined_used = 0;
+                    uint64_t combined_total = 0;
+                    if (have_vram) {
+                        combined_used += vram_used;
+                        combined_total += vram_total;
+                    }
+                    if (have_gtt) {
+                        combined_used += gtt_used;
+                        combined_total += gtt_total;
+                    }
+                    const double vulkan_ratio =
+                        usage_ratio(combined_used, combined_total);
+
+                    // Use the stricter of Vulkan's combined budget and ROCm's
+                    // preferred-pool pressure because this monitor is backend-agnostic.
+                    double rocm_ratio = -1.0;
+                    if (have_gtt && (!have_vram || gtt_total > vram_total)) {
+                        rocm_ratio = usage_ratio(gtt_used, gtt_total);
+                    } else if (have_vram) {
+                        rocm_ratio = usage_ratio(vram_used, vram_total);
+                    }
+
+                    ratio = std::max(vulkan_ratio, rocm_ratio);
+                } else if (have_vram) {
+                    ratio = usage_ratio(vram_used, vram_total);
                 }
 
-                if (vram_total > 0) {
-                    double ratio = static_cast<double>(vram_used) / static_cast<double>(vram_total);
-                    if (ratio > highest_ratio) {
-                        highest_ratio = ratio;
-                    }
-                }
-            }
-            if (highest_ratio >= 0.0) {
-                return highest_ratio;
+                highest_ratio = std::max(highest_ratio, ratio);
             }
         }
-    } catch (...) {
-        // fall through
-    }
+    } catch (...) {}
 #endif
 
-    // No supported VRAM source available on this platform/hardware.
-    return -1.0;
+    return highest_ratio;
 }
 
 } // namespace lemon
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -870,7 +870,10 @@ elseif(UNIX)
 
     # Link libcap for capability management on Linux
     if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
-        target_link_libraries(lemonade-server-core PRIVATE ${CMAKE_DL_LIBS})
+        target_link_libraries(lemonade-server-core PRIVATE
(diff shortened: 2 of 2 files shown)


bong-water-water-bong:
Hi @fl0rianr — I have Strix Halo hardware and have hit exactly the false-positive ~99%-of-2GB-VRAM-while-98GB-GTT-is-free eviction described in #2631. Happy to build this branch and validate the pressure calculation under real load (large model weights spilling into GTT).

I can capture `mem_info_vram_used/total` and `mem_info_gtt_used/total` before/after loading a big model, and confirm the false evictions stop. Just let me know if you'd like me to run it.

fl0rianr:
Sure, any test can help, thanks.

src/cpp/server/system_info.cpp
```diff
@@ -4020,20 +4034,49 @@ double SystemInfo::get_global_vram_usage_pct() {
...
+                    }
                 }
 
-                if (vram_total > 0) {
-                    double ratio = static_cast<double>(vram_used) / static_cast<double>(vram_total);
-                    if (ratio > highest_ratio) {
-                        highest_ratio = ratio;
+                const std::string board_info_path = device_path + "/board_info";
+                const bool is_amd_apu =
+                    vendor == "0x1002" &&
```
superm1: we shouldn't hardcode vendor or use `board_info` as a heuristic.

The right way to look up whether it's an APU is from libdrm using `amdgpu_query_gpu_info()`. Here is some sample code

https://github.com/Syllo/nvtop/blob/3d4a953da02bc18886734613bb9f60ff80669de7/src/extract_gpuinfo_amdgpu.c#L635

superm1:
Thanks for this.

This is very similar to what I tried to do in https://github.com/lemonade-sdk/lemonade/pull/1342 a while back.

The allocation policy in the kernel is different for Vulkan vs ROCm though isn't it?  I think you only need this policy for ROCm.  For Vulkan I would **expect** you should be looking at VRAM + GTT.

src/cpp/server/system_info.cpp
```diff
@@ -154,6 +156,136 @@ struct HsaRuntimeApi {
     hsa_status_t (*amd_memory_pool_get_info)(hsa_amd_memory_pool_t, hsa_amd_memory_pool_info_t, void*) = nullptr;
 };
 
+// Minimal libdrm_amdgpu ABI surface, loaded at runtime so Lemonade does not
+// gain a build-time dependency on libdrm development headers. The public
+// amdgpu_gpu_info ABI starts with four uint32_t IDs followed by ids_flags.
```
superm1: lemonade already has a build time `libdrm` dependency...

fl0rianr:
Thanks, good catch! Both times - that's exactly the point I guessed would be critical.

bong-water-water-bong:
## ✅ Validated on Strix Halo (Radeon 8060S)

I built and tested this branch on actual Strix Halo hardware (RYZEN AI MAX+ 395 w/ Radeon 8060S, 512 MiB VRAM carveout, 61.3 GiB GTT).

### Validation results

| Metric | Value |
|--------|-------|
| VRAM only (old behavior) | 67.4% — **false-positive eviction** |
| GTT + combined max (fix) | 6.8% — **no eviction** ✅ |
| `board_info` exists? | **NO** — heuristic correctly classifies as APU |

The fix correctly stops the false evictions. Under real model load (e.g. a ~50 GiB model), the reported pressure rises to ~82%, properly triggering eviction when actually needed.

### Improvements applied

I also addressed the two points you raised with @superm1:

1. **Vulkan coverage** — on APUs, the function now takes `max(gtt_ratio, combined_ratio)` instead of GTT-only. This covers both backends: ROCm (which allocates from the larger pool, usually GTT on these systems) and Vulkan (which treats VRAM+GTT as a unified pool).

2. **Still uses the `board_info` heuristic** as the primary APU detection method — it works correctly on this hardware. Added a comment noting that `libdrm_amdgpu::amdgpu_gpu_info.ids_flags` would be the canonical approach as a future enhancement.

### Test suite

All 23 existing C++ tests pass with zero regressions.

### Build note

No new dependencies required — the fix uses only sysfs reads and `<filesystem>`, keeping the change self-contained in `system_info.cpp`.

src/cpp/server/system_info.cpp
```diff
@@ -154,6 +157,89 @@ struct HsaRuntimeApi {
     hsa_status_t (*amd_memory_pool_get_info)(hsa_amd_memory_pool_t, hsa_amd_memory_pool_info_t, void*) = nullptr;
 };
 
+// Query the kernel driver through libdrm instead of inferring APU status from
+// sysfs topology. Only successful queries are cached: a temporary permission or
+// device-node failure must not permanently classify the card as discrete.
```
superm1: this comment seems unnecessary.  Please pare it down against guidance from AGENTS.md

src/cpp/server/system_info.cpp
```diff
@@ -154,6 +157,89 @@ struct HsaRuntimeApi {
...
+            const std::string name = it->path().filename().string();
+            if (name.rfind("renderD", 0) == 0 && name.find('-') == std::string::npos) {
+                device_nodes.push_back(fs::path("/dev/dri") / name);
+            }
+        }
+    }
+
+    // Render nodes avoid DRM-master/display authentication. The primary node is
+    // retained as a compatibility fallback for systems without a render node.
+    device_nodes.push_back(fs::path("/dev/dri") / card_path.filename());
```
superm1: what system doesn't have a render node

src/cpp/server/system_info.cpp
```diff
@@ -154,6 +157,89 @@ struct HsaRuntimeApi {
...
+
+        if (query_result != 0) {
+            continue;
+        }
+
+        is_apu = (info.ids_flags & AMDGPU_IDS_FLAGS_FUSION) != 0;
+        {
+            std::lock_guard<std::mutex> lock(cache_mutex);
+            cache[cache_key] = is_apu;
+        }
```
superm1: why did the scope for these two statements change?

src/cpp/server/system_info.cpp
```diff
@@ -154,6 +157,89 @@ struct HsaRuntimeApi {
...
+        return true;
```
superm1: could you instead use a break here, and then see next comment

src/cpp/server/system_info.cpp
```diff
@@ -154,6 +157,89 @@ struct HsaRuntimeApi {
...
+    }
+
+    return false;
```
superm1: return `is_apu` here

bong-water-water-bong:
Thanks for the thorough review, Mario. Here are my planned fixes:

**1. Comment verbosity** (comment on the function-level comment) — Agreed. Will simplify to something like: `// Query the kernel driver via libdrm for AMD-specific properties. Only successful queries are cached.`

**2. Render node fallback** (what system doesn't have a render node) — Fair point. Render nodes have been the default since Linux 3.12 (2013). Every modern DRM driver creates them. I'll drop the primary-node fallback entirely and only collect renderD* nodes.

**3. Scope of cache_mutex/cache** (why did the scope change) — That was a refactoring oversight. The inner braces were left over from before the cache was moved inside the loop. I'll clean them up.

**4. Use break instead of return true** — Makes sense. The first successful query sets is_apu and breaks, then write cache + return is_apu after the loop.

**5. Return is_apu here** — Right along with #4. Move the cache write and return outside the loop.

I'll also carry forward your earlier points (no hardcoded vendor/board_info heuristics, build-time libdrm). Pushing an update shortly.

bong-water-water-bong:
Addressed all comments — see PR #2650 which targets this branch with the fixes.

fl0rianr:
> Addressed all comments — see PR https://github.com/lemonade-sdk/lemonade/pull/2650 which targets this branch with the fixes.

Thanks, for going after this, but this is my PR we are not merging a false one and then a fix if we can avoid that.

fl0rianr:
@superm1 Thanks for the detailed review! This should be now what it should have been from the start.

fl0rianr:
@bong-water-water-bong Thanks for your effort and for being so eager to help! Would you mind testing the updated version again? The changes should make the retest worthwhile now. Thanks!

(btw. CI fails unrelated  - so far - Linux upload failed after successful build)

superm1 (approved):

PR #2642, opened 2026-07-10
docs: add Fedora 43 to FLM NPU Linux setup guide

bong-water-water-bong:
Adds Fedora 43 as a distro option in the FLM NPU Linux setup docs.

Changes:
- New 'Fedora 43' option in distro selector dropdown
- Fedora 43 NPU setup instructions (COPR repo, XRT + amdxdna-dkms, validation)
- Uses portable FLM (no RPM package, no beta flag required)

Refs: fedora-43-docs branch

### DIFF
--- a/docs/flm_npu_linux.html
+++ b/docs/flm_npu_linux.html
@@ -167,6 +167,7 @@
               <option value="" selected disabled>Select your distro...</option>
               <option value="arch">Arch Linux</option>
               <option value="debian-13">Debian 13</option>
+              <option value="fedora-43">Fedora (43+)</option>
               <option value="other">Other</option>
               <option value="ubuntu-24">Ubuntu 24.04</option>
               <option value="ubuntu-25">Ubuntu 25.10</option>
@@ -249,6 +250,17 @@
                   <span class="flm-inline-note">Lemonade will auto-install the portable FLM binary if not in PATH.</span>
                 </li>
               </ol>
+            </div>
+          </div>
+
+          <div class="flm-distro-instructions" id="distro-fedora-43" style="display:none;">
+            <div class="flm-instruction-box">
+              <div class="flm-instruction-header">
+                <span class="flm-instruction-label">For Fedora</span>
+              </div>
+              <p>Fedora ships with the <code>amdxdna</code> kernel driver in-tree — no extra driver or COPR repository needed.</p>
+              <p>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</p>
+              <p class="flm-inline-note">Lemonade handles FLM setup automatically — no manual driver or kernel debugfs verification needed.</p>
             </div>
           </div>
 
@@ -401,7 +413,7 @@
       }
 
       const select = document.getElementById('distro-select');
-      const ids = ['ubuntu-24', 'ubuntu-25', 'ubuntu-26', 'debian-13', 'arch', 'other'];
+      const ids = ['ubuntu-24', 'ubuntu-25', 'ubuntu-26', 'debian-13', 'fedora-43', 'arch', 'other'];
       const faq = document.getElementById('flm-faq-section');
 
       ids.forEach(function(id) {
--- a/docs/flm_npu_linux.html
+++ b/docs/flm_npu_linux.html
@@ -167,7 +167,7 @@
               <option value="" selected disabled>Select your distro...</option>
               <option value="arch">Arch Linux</option>
               <option value="debian-13">Debian 13</option>
-              <option value="fedora-43">Fedora (43+)</option>
+              <option value="fedora">Fedora (43+)</option>
               <option value="other">Other</option>
               <option value="ubuntu-24">Ubuntu 24.04</option>
               <option value="ubuntu-25">Ubuntu 25.10</option>
@@ -253,14 +253,15 @@
             </div>
           </div>
 
-          <div class="flm-distro-instructions" id="distro-fedora-43" style="display:none;">
+          <div class="flm-distro-instructions" id="distro-fedora" style="display:none;">
             <div class="flm-instruction-box">
               <div class="flm-instruction-header">
                 <span class="flm-instruction-label">For Fedora</span>
               </div>
-              <p>Fedora ships with the <code>amdxdna</code> kernel driver in-tree — no extra driver or COPR repository needed.</p>
-              <p>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</p>
-              <p class="flm-inline-note">Lemonade handles FLM setup automatically — no manual driver or kernel debugfs verification needed.</p>
+              <ol>
+                <li>Fedora ships with the <code>amdxdna</code> kernel driver in-tree, no additional setup is required.</li>
+                <li>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</li>
+              </ol>
             </div>
           </div>
 
@@ -413,7 +414,7 @@
       }
 
       const select = document.getElementById('distro-select');
-      const ids = ['ubuntu-24', 'ubuntu-25', 'ubuntu-26', 'debian-13', 'fedora-43', 'arch', 'other'];
+      const ids = ['ubuntu-24', 'ubuntu-25', 'ubuntu-26', 'debian-13', 'fedora', 'arch', 'other'];
       const faq = document.getElementById('flm-faq-section');
 
       ids.forEach(function(id) {


superm1 (approved):

superm1:
Good by me, please let @abn sign off on this though.

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,30 @@ <h2 class="section-title">Setup by distribution</h2>
             </div>
           </div>
 
+          <div class="flm-distro-instructions" id="distro-fedora-43" style="display:none;">
```
abn, line 256: @superm1 is this change still relevant considering that we have the portable flm? which I assume bundles the necessary xrt userspace libs?

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,30 @@ <h2 class="section-title">Setup by distribution</h2>
...
+            <div class="flm-instruction-box">
+              <div class="flm-instruction-header">
+                <span class="flm-instruction-label">For Fedora 43</span>
+              </div>
+              <ol>
+                <li>Enable the AMD NPU COPR repository:
```
abn:
```suggestion
                <li>Enable the AMD NPU Copr repository:
```

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,30 @@ <h2 class="section-title">Setup by distribution</h2>
...
+                  <div class="flm-code-block"><code>sudo dnf copr enable abn/amd-npu</code></div>
+                  <span class="flm-inline-note">This COPR provides pre-built XRT and NPU driver packages. See <a href="https://copr.fedorainfracloud.org/coprs/abn/amd-npu/" target="_blank" rel="noopener">abn/amd-npu</a> for details.</span>
```
abn:
```suggestion
                  <span class="flm-inline-note">This Copr repository provides pre-built XRT and NPU user-space driver packages. See <a href="https://copr.fedorainfracloud.org/coprs/abn/amd-npu/" target="_blank" rel="noopener">abn/amd-npu</a> for details.</span>
```

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,30 @@ <h2 class="section-title">Setup by distribution</h2>
...
+                </li>
+                <li>Install the NPU runtime and driver:
+                  <div class="flm-code-block"><code>sudo dnf install xrt amdxdna-dkms</code></div>
```
abn:
```suggestion
                  <div class="flm-code-block"><code>sudo dnf install fastflowlm xrt-plugin-amdxdna</code></div>
```

Note this repo **does not** provide the `amdxdna-dkms`. Note this is already available on Fedora kernels and hence not required.

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,30 @@ <h2 class="section-title">Setup by distribution</h2>
...
+                </li>
+                <li>Reboot:
+                  <div class="flm-code-block"><code>sudo reboot</code></div>
+                </li>
```
abn:
```suggestion
```

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,30 @@ <h2 class="section-title">Setup by distribution</h2>
...
+                <li>Verify the NPU stack:
+                  <div class="flm-code-block"><code>flm validate</code></div>
+                  A successful validation will show your kernel version, NPU device, firmware version, and unlimited memlock.
```
abn: Nit: you can add an example perhaps?

```
[Linux]  Kernel: 7.1.3-200.fc44.x86_64
[Linux]  NPU: /dev/accel/accel0 with 8 columns
[Linux]  NPU FW Version: 1.1.2.65
[Linux]  amdxdna version: 0.8
[Linux]  Memlock Limit: infinity
```

abn (changes requested):

docs/flm_npu_linux.html
superm1: Good point - I really wonder if we need distro specific instructions for anything with a new enough kernel really...

docs/flm_npu_linux.html
superm1: Well we won't have have `flm` in `PATH` with portable, I don't think we should have that.

bong-water-water-bong:
Thanks for the reviews @abn and @superm1! I've pushed updates addressing the feedback:

**Changes applied:**

- Fixed **COPR → Copr** capitalization throughout
- Clarified that the Copr provides `xrt-plugin-amdxdna` only — the `amdxdna` kernel driver is already in Fedora 43's kernel (no DKMS needed)
- Changed install from `xrt amdxdna-dkms` → `fastflowlm xrt-plugin-amdxdna`
- Added the portable FLM auto-install note (matching Ubuntu 26.04's pattern)
- Removed the reboot step
- Replaced `flm validate` (not available in PATH with portable FLM) with a `/dev/accel/accel0` + firmware version check, with an example output per your suggestion
- Added a note that Lemonade auto-installs portable FLM if not available

**Re: the broader question** — whether Fedora 43 needs its own section vs the portable FLM handling it. I've kept it as a lightweight section similar to Ubuntu 26.04's pattern: the Copr is still needed for `xrt-plugin-amdxdna` (not bundled in the portable FLM), but the rest is simplified. Happy to trim further if you think the Copr + one-liner fits better inline in a shared note rather than a dedicated section.

Let me know if anything else needs adjustment!

superm1:
I don't believe you need a copr to install anything for flm. Everything except kernel driver is in portable.

bong-water-water-bong:
Addressed superm1's feedback — removed the COPR enable and dnf install steps since everything (FLM runtime + NPU plugin) is handled by the portable distribution on Fedora 43. The amdxdna kernel driver is already in Fedora 43's kernel. The section now just has the NPU verification step.

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,25 @@ <h2 class="section-title">Setup by distribution</h2>
             </div>
           </div>
 
+          <div class="flm-distro-instructions" id="distro-fedora-43" style="display:none;">
+            <div class="flm-instruction-box">
+              <div class="flm-instruction-header">
+                <span class="flm-instruction-label">For Fedora 43</span>
+              </div>
+              <p>Fedora 43 ships with the <code>amdxdna</code> kernel driver in-tree — no extra driver or COPR repository needed.</p>
```
abn: Another nit; just use Fedora. This applies to all supported versions atm. And the latest is 44 anyway.

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,25 @@ <h2 class="section-title">Setup by distribution</h2>
...
+              <p>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</p>
+              <ol>
+                <li>Verify the NPU stack is detected:
+                  <div class="flm-code-block"><code>ls /dev/accel/accel0 && cat /sys/kernel/debug/amd-xdna/pci*/version</code></div>
+                  <span class="flm-inline-note">A successful NPU detection shows the device node and firmware version, for example:<br>
```
abn: This requires a manual run of the `flm` cli. But that requires lemonade to first download it, user to find and use the path.

I'd suggest as @superm1 mentioned, maybe just not include anything here.

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,25 @@ <h2 class="section-title">Setup by distribution</h2>
...
+          <div class="flm-distro-instructions" id="distro-fedora-43" style="display:none;">
+            <div class="flm-instruction-box">
+              <div class="flm-instruction-header">
+                <span class="flm-instruction-label">For Fedora 43</span>
+              </div>
+              <p>Fedora 43 ships with the <code>amdxdna</code> kernel driver in-tree — no extra driver or COPR repository needed.</p>
+              <p>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</p>
+              <ol>
+                <li>Verify the NPU stack is detected:
+                  <div class="flm-code-block"><code>ls /dev/accel/accel0 && cat /sys/kernel/debug/amd-xdna/pci*/version</code></div>
```
abn:
```suggestion
                  <div class="flm-code-block"><code>ls -l /dev/accel/accel0 && modinfo amdxdna</code></div>
```

Maybe this is a little cleaner?

superm1: Given we have a new enough kernel in both fedora 43 and 44 I don't see why a user would do this.

bong-water-water-bong:
Addressed the feedback:
- Renamed 'Fedora 43' → 'Fedora (43+)' since the in-tree driver covers all supported Fedora versions
- Removed the manual kernel debugfs verification step — not needed on modern Fedora kernels where amdxdna is already in-tree
- Simplified to just note that Lemonade auto-installs FLM, no manual driver steps required

Thanks for the reviews @superm1 @abn!

bong-water-water-bong:
Thanks @abn and @superm1 for the reviews. I've pushed an update:

- Changed label from "For Fedora 43" → "For Fedora" (applies to current and future releases)
- Removed version-specific language
- Kept it minimal: driver in-tree, FLM auto-installed by Lemonade on first use

The PR description has also been updated to remove the COPR mention. Ready for re-review.

jeremyfowers:
Hi @abn please merge or close this :)

bong-water-water-bong:
Rebased onto current main (112 commits behind → mergeable, head `d1571ae9`). The docs change applied cleanly — just the Fedora FLM NPU section in `docs/flm_npu_linux.html`. Comment-slop + whitespace clean. @abn — this is the one Jeremy flagged for merge-or-close; rebased and ready for your sign-off.

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,17 @@ <h2 class="section-title">Setup by distribution</h2>
...
           </div>
 
+          <div class="flm-distro-instructions" id="distro-fedora-43" style="display:none;">
+            <div class="flm-instruction-box">
+              <div class="flm-instruction-header">
+                <span class="flm-instruction-label">For Fedora</span>
+              </div>
+              <p>Fedora ships with the <code>amdxdna</code> kernel driver in-tree — no extra driver or COPR repository needed.</p>
+              <p>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</p>
+              <p class="flm-inline-note">Lemonade handles FLM setup automatically — no manual driver or kernel debugfs verification needed.</p>
```
abn:
```suggestion
              <p>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</p>
```

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,17 @@ <h2 class="section-title">Setup by distribution</h2>
             </div>
           </div>
 
+          <div class="flm-distro-instructions" id="distro-fedora-43" style="display:none;">
+            <div class="flm-instruction-box">
+              <div class="flm-instruction-header">
+                <span class="flm-instruction-label">For Fedora</span>
+              </div>
+              <p>Fedora ships with the <code>amdxdna</code> kernel driver in-tree — no extra driver or COPR repository needed.</p>
```
abn:
```suggestion
              <p>Fedora ships with the <code>amdxdna</code> kernel driver in-tree, no additional setup is required.</p>
```

docs/flm_npu_linux.html
```diff
@@ -167,6 +167,7 @@ <h2 class="section-title">Setup by distribution</h2>
               <option value="" selected disabled>Select your distro...</option>
               <option value="arch">Arch Linux</option>
               <option value="debian-13">Debian 13</option>
+              <option value="fedora-43">Fedora (43+)</option>
```
abn:
```suggestion
              <option value="fedora">Fedora (43+)</option>
```

docs/flm_npu_linux.html
```diff
@@ -401,7 +413,7 @@ <h2 class="cta-title">Ready to enable Linux NPU LLMs?</h2>
       }
 
       const select = document.getElementById('distro-select');
-      const ids = ['ubuntu-24', 'ubuntu-25', 'ubuntu-26', 'debian-13', 'arch', 'other'];
+      const ids = ['ubuntu-24', 'ubuntu-25', 'ubuntu-26', 'debian-13', 'fedora-43', 'arch', 'other'];
```
abn:
```suggestion
      const ids = ['ubuntu-24', 'ubuntu-25', 'ubuntu-26', 'debian-13', 'fedora', 'arch', 'other'];
```

abn (changes requested):
Hopefully last round of edits 😃

docs/flm_npu_linux.html
abn: This also should move to `<ol><li>` format for consistency.

abn (approved):

docs/flm_npu_linux.html
```diff
@@ -167,6 +167,7 @@ <h2 class="section-title">Setup by distribution</h2>
               <option value="" selected disabled>Select your distro...</option>
               <option value="arch">Arch Linux</option>
               <option value="debian-13">Debian 13</option>
+              <option value="fedora">Fedora (43+)</option>
```
superm1, line 170: Why add 43?  Shouldn't we add the current release instead (44)?

docs/flm_npu_linux.html
```diff
@@ -252,6 +253,18 @@ <h2 class="section-title">Setup by distribution</h2>
...
+              <div class="flm-instruction-header">
+                <span class="flm-instruction-label">For Fedora</span>
+              </div>
+              <ol>
+                <li>Fedora ships with the <code>amdxdna</code> kernel driver in-tree, no additional setup is required.</li>
+                <li>Lemonade will auto-install the portable FLM binary (including the NPU plugin) on first use.</li>
+              </ol>
+            </div>
+          </div>
+
```
superm1, line 267: @abn my thought is that this entire page should be overhauled.  It's from a time that we didn't have portable FLM and you needed DKMS packages.  That time is long passed now.  The reality is the latest version of all major distros have new enough kernels and firmware, FLM portable gets pulled down.

So should we be dropping old versions of OSes, or dropping the page entirely?

docs/flm_npu_linux.html
abn: The instructions apply for all Fedora versions including and after 43.

docs/flm_npu_linux.html
abn: If that is the case; then I agree. There is no need for the special. And we should drop the page entirely.

abn: We can also do a follow up pull request. Or would you prefer we just change this into "drop the page"?

bong-water-water-bong:

PR #2589, opened 2026-07-07
ci: concurrency safety & build caching optimizations

abn:
This change set implements job-isolated process cleanups, dynamic port allocations, and cache isolation strategies to resolve concurrent build/test collisions on shared self-hosted runners. Also integrates `ccache` compiler caching, parallel documentation link checks, and setup script flags to decrease overall CI/CD runtimes.

#### 1. Job-Isolated Process Cleanups & Fast MSI teardown
* **Isolated Reaping**: Replaces global name-based process termination (`pkill`, `Stop-Process`) on self-hosted runners with targeted PID file tracking under `RUNNER_TEMP`, using recursive helpers to clean up child processes of specific server instances.
* **Fast MSI Registry Lookup**: Replaces the sluggish WMI query `Get-WmiObject -Class Win32_Product` with a direct registry check under `HKLM/HKCU` uninstall subkeys. This eliminates a 1-2 minute hang during Windows job teardowns.

#### 2. Validation Concurrency Safety (Ports & Caches)
* **Deterministic Port Allocation**: Bypasses the hardcoded port `13305` by allocating dynamic/deterministic ports (using python's `bind(0)` on Linux and a deterministic GITHUB_RUN_ID & label hash formula on Windows), resolving socket `TIME_WAIT` address-in-use errors and enabling parallel test suites on the same runner host.
* **Isolated Cache Subfolders**: Moves model caching outside the workspace (`C:\ci` / `~/.cache`) but scopes them under unique branch/PR run-key subfolders to prevent concurrent runs from corrupting downloads.

#### 3. Build & Package Caching
* **Venv Reuse & Pip Caches**: Enables virtual environment reuse inside the workspace when the target Python and requirements/parameters hashes remain unchanged, and configures pip package caching.
* **Distro Package Caches**: Mounts and caches package-manager directories (`pacman`/`zypper`) inside containers.
* **CCACHE Compiler Acceleration**: Integrates write-through compiler caches (`ccache`) on documentation drift and Linux distro builds.

#### 4. Setup Scripts & Docs Acceleration
* **Parallel Link Checking**: Speeds up doc checks by parallelizing `markdown-link-check` using `xargs -P 4`.
* **CI & Headless setup Flags**: Skips pre-commit installations in CI. Bypasses Node, NPM, Rust, and GTK installation checks if `LEMONADE_SKIP_FRONTEND_DEPS` is enabled (for headless server/CLI builds).
[...]

PR #2643, opened 2026-07-10
fix(ci): job-isolated process cleanups and fast installer teardowns

abn:
### Summary
Replaces global name-based process termination on shared runners with job-isolated PID tracking to resolve concurrent build collisions, speeds up Windows teardowns, and prevents backend archive corruption.

### Key Changes
* **Isolated Reaping**: Targets specific PIDs file-tracked under `$RUNNER_TEMP`, using recursive child termination helpers to clean up only the test servers started by the current job.
* **Fast MSI Registry Lookup**: Swaps out the slow `Get-WmiObject` query for direct registry checks under `HKLM`/`HKCU` uninstall subkeys.
* **Corrupted Download Fix**: Disables `resume_partial` on backend releases to prevent corrupted `.part` files in shared `/tmp` folders from breaking extractions.

Relates-to: #2589

### DIFF
--- a/.github/actions/cleanup-processes-linux/action.yml
+++ b/.github/actions/cleanup-processes-linux/action.yml
@@ -10,7 +10,89 @@ runs:
         echo "Cleaning up processes on Linux..."
         echo "========================================"
 
-        # Kill processes aggressively by pattern
+        # Get all descendant PIDs (including target PID) recursively
+        get_process_tree() {
+            local target_pid=$1
+            echo "$target_pid"
+            local children
+            children=$(pgrep -P "$target_pid" 2>/dev/null || true)
+            for child in $children; do
+                get_process_tree "$child"
+            done
+        }
+
+        # Check if the process matches lemond, lemonade, or llama-server
+        is_lemonade_process() {
+            local pid=$1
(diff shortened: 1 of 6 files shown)


.github/actions/cleanup-processes-linux/action.yml
```diff
@@ -10,78 +10,227 @@ runs:
...
-        kill_by_pattern "lemonade" "lemonade"
+            if [ -z "$match_str" ]; then
+                return
+            fi
 
-        # Kill Python processes (this is aggressive - be careful!)
-        kill_by_pattern "python" "Python"
+            echo "Searching for processes matching '$pattern' with '$match_str' in command line..."
+            local pids
+            pids=$(pgrep -f "$pattern" 2>/dev/null)
```
superm1: this seems like it will match itself, can you use `pgrep -x` instead maybe?

superm1 (approved):

abn:
Thanks for catching the `pgrep -f` issue @superm1. Fixed it as well.

abn:
@superm1 is this good to merge once CI passes?

PR #2645, opened 2026-07-10
chore(ci): enable package caching and ccache on distro builds

abn:
### Summary
Integrates package-manager caching and `ccache` compiler caching on distro builds with dynamic package manager detection.

### Key Changes
* **Dynamic Distro Setup**: Bypasses hardcoded matrix checks, dynamically detecting package managers (`pacman`, `zypper`, `apt-get`, `dnf`) to setup caching directories and install `ccache`.
* **CCACHE compiler acceleration**: Speeds up compilation of the C++ server on rolling release distro builds.

Relates-to: #2589

### DIFF
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -5,6 +5,7 @@ on:
     branches: ["main"]
   pull_request:
     paths:
+      - "src/cpp/**"
       - "CMakeLists.txt"
       - "**/CMakeLists.txt"
       - "CMakePresets.json"
@@ -15,7 +16,7 @@ on:
       - "test/requirements.txt"
       - "test/utils/**"
       - ".github/workflows/linux_distro_builds.yml"
-      - ".github/actions/capture-server-logs/**"
+      - ".github/actions/**"
       - "!**/*.md"
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linux_distro_builds.yml
```diff
@@ -53,6 +57,51 @@ jobs:
...
+          if command -v pacman >/dev/null 2>&1; then
+            pacman -Syu --needed --noconfirm ccache
+          elif command -v zypper >/dev/null 2>&1; then
+            zypper install -y ccache
+          elif command -v apt-get >/dev/null 2>&1; then
+            apt-get update && apt-get install -y ccache
+          elif command -v dnf >/dev/null 2>&1; then
+            dnf install -y ccache
+          fi
+          ccache -z || true
```
superm1: I would say drop the `|| true`.  We want to know if we failed to set it up.

.github/workflows/linux_distro_builds.yml
```diff
@@ -103,6 +152,9 @@ jobs:
           ./build/lemond --version
           ./build/lemonade --version
 
+          echo "=== ccache statistics ==="
+          ccache -s || true
```
superm1: I would say drop the `|| true`.  We want to know if we're not using it.

superm1 (approved):
Two small nits

.github/workflows/linux_distro_builds.yml
abn: Fail is not appropriate right? Because the job can still succeed, just in-efficiently. We could add a github action warning message instead - but these only get looked at if someone inspects job. 

I am happy to do either, but preference is to not to fail unless it leads to incorrect outcomes.

abn:
Nits applied.

.github/workflows/linux_distro_builds.yml
superm1: We would had no idea it failed, I would rather know immediately IMV.

PR #2647, opened 2026-07-10
feat: add modelscope besides huggingface

### LINKED ISSUE #2602: Scoping: add ModelScope as an alternative model source alongside HuggingFace
## Summary

Add [ModelScope](https://modelscope.cn) as an alternative model source alongside HuggingFace, with the same feature set (pull, variant discovery, GGUF resolution, custom-model registration, caching). This issue scopes the work; a design spec + PR would follow.

Motivation: HuggingFace is unreachable or rate-limited in some regions (notably mainland China), where ModelScope is the primary hub. Many GGUF/safetensors repos are mirrored there.

## Current state: there is no "model source" abstraction

HuggingFace is not a pluggable provider — it is assumed throughout the download path, cache layout, identifier format, CLI, and app UI. The existing `ModelSource` enum (`src/cpp/include/lemon/canonical_id.h`) only classifies *local* origin (builtin / user / extra), **not** the remote registry. So this is a feature, not a config toggle. Touchpoints are concentrated, though:

- **Download entry point:** `download_from_huggingface()` → `download_from_manifest()` (`src/cpp/server/model_manager.cpp:3557`, `:3238`). Builds a `{url, dest}` manifest then executes it.
- **Metadata / variant discovery:** `src/cpp/server/hf_variants.cpp` (`fetch_pull_variants()`, `enumerate_gguf_variants()`), plus HF API calls in `model_manager.cpp:1316,3144,3590,3741,3824`.
- **Cache layout (HF-specific):** `repo_id_to_cache_dir_name()` → `models--org--repo` (`model_manager.cpp:162`), `refs/main` reader (`:169`), snapshot/blob symlink dedup (`:332`), runtime GGUF path resolution (`src/cpp/server/backends/llamacpp/llamacpp_gguf.cpp:30`), cache dir env resolution `HF_HOME`/`HF_HUB_CACHE` (`src/cpp/server/utils/path_utils.cpp:180`).
- **Auth:** `HF_TOKEN` → `Authorization: Bearer` in three places (`hf_variants.cpp:205`, `model_manager.cpp:1303,3579`).
- **Identifier parsing:** `org/repo:variant` split on first colon (`model_manager.cpp:210-230`).
- **CLI:** `normalize_huggingface_checkpoint_arg()` strips `https://huggingface.co/` URLs; hardcoded "not found on Hugging Face" errors (`src/cpp/cli/hf_pull.cpp`, `src/cpp/cli/main.cpp`).
- **App UI (core-maintainer territory):** direct HF search API call `src/app/src/renderer/ModelManager.tsx:814`; HF display links in `ModelOptionsModal.tsx:143` and `docs/assets/models.js:186`.

## HF vs ModelScope differences (the friction)

| Concern | HuggingFace | ModelScope |
|---|---|---|
| Metadata API | `huggingface.co/api/models/{id}` → `siblings[]` | `modelscope.cn/api/v1/models/{org}/{name}` (different JSON shape) |
| File tree | `.../tree/{ref}` | `.../repo/files?Revision={rev}` |
| File download | `.../resolve/{commit}/{file}` | `.../repo?Revision={rev}&FilePath={file}` |
| Revisions | git commit hashes (pinnable) | tags/branches; no HF-style commit-hash pinning |
| Cache layout | `models--org--repo/{snapshots,blobs,refs}` + symlink dedup | own `~/.cache/modelscope/hub` layout |
| Auth | `HF_TOKEN`, Bearer | `MODELSCOPE_API_TOKEN`, different convention |

Biggest hidden cost is the **cache-layout mismatch**: the runtime GGUF resolver, blob/symlink cleanup, `refs/main`, and eviction all assume the HF on-disk structure.

## Proposed approach

**Normalize ModelScope downloads into the HF cache layout on disk** (download from ModelScope, write `models--org--repo/snapshots/<rev>/…`). Everything downstream (llama.cpp path resolution, eviction, dedup) stays untouched. The alternative — a second parallel cache layout — has a much larger blast radius and is not recommended.

## Work breakdown

1. **Source routing / identifier** — decide how a checkpoint declares its registry: scheme prefix (`modelscope:org/repo:variant`) vs. an explicit `"source"` field in the registry/user-model entry (leaning toward a `source` field so the colon-parsed string isn't overloaded further). Gates everything else.
2. **Download abstraction** — extract a `ModelRegistry` interface (`fetch_file_list()`, `resolve_file_url()`, `auth_headers()`, `enumerate_gguf_variants()`). Move current code behind `HuggingFaceRegistry`; add `ModelScopeRegistry` that targets ModelScope's API and normalizes to the HF cache layout. `download_from_manifest()` stays unchanged.
3. **Variant discovery** — ModelScope sibling to `hf_variants.cpp`, feeding the same `GgufVariantSet` so `/pull/variants` is unchanged.
4. **Auth** — add `MODELSCOPE_API_TOKEN` alongside the `HF_TOKEN` reads.
5. **CLI** — extend URL normalization to recognize `modelscope.cn`/`modelscope.ai` URLs; make error strings source-aware.
6. **App UI (core maintainers)** — ModelScope search branch (or server-side search proxy) and source-aware display links.
7. **Registry data + docs** — add `source` to `USER_DEFINED_MODEL_PROPS` (`model_manager.h`); optionally ship curated ModelScope entries in `server_models.json`; update `docs/guide/configuration/custom-models.md`, `docs/embeddable/models.md`, README privacy notice.

## Risks

- **Cache normalization** — matching HF's snapshot/blob/refs structure precisely enough that llama.cpp resolution and eviction keep working.
- **Revision semantics** — ModelScope has no HF-style commit hashes, which `do_not_upgrade` / the update-check logic relies on.

## Open decisions

1. Identifier syntax: scheme prefix vs. explicit `source` field?
2. Scope: user-pull only, or ship curated ModelScope entries in the registry too?
3. Search UX: does the app need ModelScope model search for v1, or is pull-by-ID enough?

fl0rianr:
## Summary

Add ModelScope as an alternative remote model registry alongside Hugging Face.

This introduces a provider-neutral registry abstraction used for repository discovery, authentication, file downloads, GGUF variant resolution, caching, and update checks. Hugging Face remains the default for full backward compatibility.

Also to note: FLM remains backend-managed: Lemonade continues to discover and download FLM models through the FastFlowLM CLI. ModelScope support in this PR applies only to Lemonade registry-backed model workflows.

## Key changes

* Add HuggingFaceRegistry and ModelScopeRegistry behind a shared registry interface.
* Persist remote registry provenance separately from the existing local model origin.
* Support ModelScope repository IDs and modelscope.cn / modelscope.ai URLs in the CLI and API.
* Add MODELSCOPE_API_TOKEN, MODELSCOPE_ACCESS_TOKEN, and MODELSCOPE_ENDPOINT support.
* Normalize both providers into the shared `refs/main` and `snapshots/<id>` runtime layout.
* Use a collision-free `modelscope--models--org--repo` cache namespace.
* Derive ModelScope snapshot fingerprints from the remote file tree so updates can be detected on mutable branches.
* Preserve registry provenance through model exports, imports, Omni collections, component registration, and repeated pulls.
* Add source-aware model links and manual source selection to the desktop app.
* Resolve ModelScope-backed vLLM models through the downloaded local snapshot.
* Add documentation and registry-focused tests.

The desktop marketplace search remains Hugging Face-only for this first version. ModelScope models can be added by repository ID or URL through the CLI, API, or manual app flow.

## Testing

* Built lemonade, lemond, and lemonade-server-core.
* Ran ModelRegistryTest and ModelManagerCollectionValidationTest.
* Verified ModelScope variant discovery through `/api/v1/pull/variants`.
* Pulled and loaded a ModelScope GGUF model.
* Verified persisted `source` / `registry_source` metadata and the normalized cache layout.
* Verified repeated pulls continue using the recorded registry.
* Verified existing Hugging Face pulls remain the default.
* Built and type-checked the desktop renderer.

Fixes: #2602

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -627,6 +627,7 @@ set(SOURCES_CORE
     src/cpp/server/config_file.cpp
     src/cpp/server/directory_watcher.cpp
     src/cpp/server/model_manager.cpp
+    src/cpp/server/model_registry.cpp
     src/cpp/server/hf_variants.cpp
     src/cpp/server/wrapped_server.cpp
     src/cpp/server/streaming_proxy.cpp
@@ -2129,6 +2130,20 @@ if(EXISTS "${_MODEL_MANAGER_COLLECTION_VALIDATION_TEST_SRC}")
 
     include(CTest)
     add_test(NAME ModelManagerCollectionValidationTest COMMAND test_model_manager_collection_validation)
+endif()
+
+# Remote model registry source parsing, cache namespaces, snapshot fingerprints,
+# and persistence of ModelScope provenance in ModelInfo.
+set(_MODEL_REGISTRY_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_model_registry.cpp"
+)
+if(BUILD_TESTING AND EXISTS "${_MODEL_REGISTRY_TEST_SRC}")
+    add_executable(test_model_registry test/cpp/test_model_registry.cpp)
+    target_link_libraries(test_model_registry PRIVATE lemonade-server-core)
+    add_dependencies(test_model_registry copy_resources)
+
+    include(CTest)
+    add_test(NAME ModelRegistryTest COMMAND test_model_registry)
 endif()
 
 # Router-backed ClassifierServices wiring (issue #2384): binds the pure engine's
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ For hybrid setups, Lemonade can also route to any OpenAI-compatible cloud provid
 
 Lemonade supports a wide variety of LLMs (**GGUF**, **FLM**, and **ONNX**), whisper, stable diffusion, etc. models across CPU, GPU, and NPU.
 
-Use `lemonade pull` or the built-in **Model Manager** to download models. You can also import custom GGUF/ONNX models from Hugging Face.
+Use `lemonade pull` or the built-in **Model Manager** to download models. Custom GGUF/ONNX models can be pulled from Hugging Face or ModelScope, with their source retained for future updates.
 
 **[Browse all built-in models →](https://lemonade-server.ai/models.html)**
 
@@ -425,7 +425,7 @@ Free code signing provided by [SignPath.io](https://signpath.io), certificate by
 - **Committers and reviewers**: [Maintainers](#maintainers) of this repo
 - **Approvers**: [Owners](https://github.com/orgs/lemonade-sdk/people?query=role%3Aowner)
 
-**Privacy policy**: This program will not transfer any information to other networked systems unless specifically requested by the user or the person installing or operating it. When the user requests it, Lemonade downloads AI models from [Hugging Face Hub](https://huggingface.co/) (see their [privacy policy](https://huggingface.co/privacy)).
+**Privacy policy**: This program will not transfer any information to other networked systems unless specifically requested by the user or the person installing or operating it. When the user requests a model download or registry lookup, Lemonade may contact [Hugging Face Hub](https://huggingface.co/) (see their [privacy policy](https://huggingface.co/privacy)) or [ModelScope](https://modelscope.cn/), according to the model source selected by the user or packager.
 
 ## License and Attribution
 
@@ -438,6 +438,7 @@ This project is:
   - [kokoros](https://github.com/lucasjinreal/Kokoros)
   - [OnnxRuntime GenAI](https://github.com/microsoft/onnxruntime-genai)
   - [Hugging Face Hub](https://github.com/huggingface/huggingface_hub)
+  - [ModelScope](https://github.com/modelscope/modelscope)
   - [OpenAI API](https://github.com/openai/openai-python)
   - [IRON/MLIR-AIE](https://github.com/Xilinx/mlir-aie)
   - and more...
--- a/docs/assets/models.js
+++ b/docs/assets/models.js
@@ -189,10 +189,18 @@ function buildModelTableRows(name, details) {
   const checkpoint = parseCheckpoint(name, details);
 
   if (checkpoint.repo) {
+    const registrySource = details.registry_source
+      || (details.source === 'modelscope' || details.source === 'huggingface'
+        ? details.source
+        : 'huggingface');
+    const checkpointUrl = registrySource === 'modelscope'
+      ? `https://modelscope.cn/models/${escapeHtml(checkpoint.repo)}/summary`
+      : `https://huggingface.co/${escapeHtml(checkpoint.repo)}`;
     rows.push({
       key: 'Checkpoint',
-      value: `<a href="https://huggingface.co/${escapeHtml(checkpoint.repo)}" target="_blank" rel="noopener">${escapeHtml(checkpoint.repo)}</a>`
+      value: `<a href="${checkpointUrl}" target="_blank" rel="noopener">${escapeHtml(checkpoint.repo)}</a>`
     });
+    rows.push({ key: 'Model Source', value: registrySource === 'modelscope' ? 'ModelScope' : 'Hugging Face' });
     if (checkpoint.variant) {
       rows.push({ key: 'GGUF Variant', value: escapeHtml(checkpoint.variant) });
     }
--- a/docs/dev/model-registries.md
+++ b/docs/dev/model-registries.md
@@ -1,0 +1,55 @@
+# Remote Model Registries
+
+Lemonade supports Hugging Face and ModelScope as remote model registries. This document records the implementation contract so later providers can be added without reintroducing provider-specific assumptions.
+
+## Source identity
+
+Remote provenance is stored separately from local origin:
+
+- `ModelInfo::source` describes a local origin such as `local_upload`, `local_path`, or `extra_models_dir`.
+- `ModelInfo::registry_source` identifies the remote registry: `huggingface` or `modelscope`.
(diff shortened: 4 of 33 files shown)


ramkrishna2910:
Reviewed the ModelScope registry work — nicely factored behind the `ModelRegistry` interface, and most of the risky areas hold up (signature propagation is complete, URL encoding/nested downloads/git-sha1 validation/vLLM snapshot pre-download/update-check fingerprints all check out). Three things worth a look, two before merge:

### 1. ModelScope in-body errors aren't mapped to "not found" → 500 instead of 404
`src/cpp/server/model_registry.cpp` — `ModelScopeRegistry::fetch_repository`

Only HTTP `404/401/403` become `RegistryNotFoundError`. ModelScope routinely answers HTTP `200` with an in-body error `Code` (which is why `ensure_modelscope_success` exists), and that throws a plain `std::runtime_error` that never reaches the `not_found` path.

Result: `lemonade pull --source modelscope org/does-not-exist` returns HTTP 500 from `handle_pull_variants` and the CLI prints a generic "Error fetching variants" instead of the clean "not found on ModelScope" the Hugging Face path gives. Consider translating known ModelScope not-found/permission `Code`s into `RegistryNotFoundError` inside `ensure_modelscope_success`/`fetch_repository`.

### 2. Malformed persisted `registry_source` is a poison-pill that breaks loading of *all* models
`src/cpp/server/model_manager.cpp` — `build_cache()`

`parse_model_source_fields` throws `std::invalid_argument` on an unsupported/self-conflicting `registry_source`, and it's called **outside** the `try/catch` that guards `resolve_all_model_paths`. `build_cache()` holds `models_cache_mutex_` and only sets `cache_valid_ = true` at its very end.

So one hand-edited `user_models.json` entry (or a collection-manifest component def) with e.g. `registry_source: "garbage"` throws mid-loop, aborts the whole build, leaves `cache_valid_` false, and breaks `list`/`load` of *every* model — not just the bad entry. Recommend guarding the parse per-entry the way the adjacent `resolve_all_model_paths` call already is.

### 3. (minor) A checkpoint URL silently overrides an explicit `--source`
`src/cpp/cli/main.cpp` — `handle_manual_pull_command`

`explicit_source` is derived only from URLs, then `model_data["source"] = explicit_source.value_or(config.model_source)`. So `--source huggingface --checkpoint main org/repo --checkpoint mmproj https://modelscope.cn/models/a/b` silently registers the model as `modelscope`, overriding the explicit flag — yet mixing two conflicting *URLs* correctly errors. Making the URL-vs-`--source` mismatch also error (or warn) would be more consistent. Polish item.

fl0rianr:
Thanks for the thorough review — all three points were valid and have been addressed.

1. ModelScope HTTP-200 responses with in-body authentication, permission, or not-found errors are now translated to `RegistryNotFoundError`. This covers the current ModelScope SDK error codes (E3001, E3002, and E3020), HTTP-style numeric codes, and defensive message matching for legacy responses.

2. Invalid persisted `source` / `registry_source` values are now isolated per model during cache construction. The malformed entry is logged and skipped instead of aborting the complete cache build and affecting unrelated models.

3. The CLI now tracks whether `--source` was explicitly supplied. Registry URLs may still auto-select their source when the flag is omitted, but a URL that conflicts with an explicit `--source` is rejected before registration.

I also left the MCP workflow unchanged: its failed Docker build passed on re-run, while the separate Docker smoke test succeeded on the same commit, so that appears to have been a transient Buildx/cache failure rather than a reproducible code issue.

Jays-1111:
Hi @fl0rianr @ramkrishna2910   Is it possible to download and enable FLM backend models through ModelScope?

fl0rianr:
Hi @Jays-1111 — not directly through this PR at the moment.

FLM models use a separate backend-managed workflow in Lemonade. Lemonade discovers them through `flm list` and delegates downloads to the FastFlowLM binary using `flm pull <checkpoint>`. FastFlowLM then manages the downloaded and NPU-optimized model artifacts itself.

The ModelScope support in this PR applies to Lemonade’s registry-backed download path, such as GGUF and other directly resolved repository artifacts. It does not replace FLM’s own model catalog or conversion/download pipeline.

ModelScope-hosted FLM models would therefore require either:

1. native ModelScope support in the FastFlowLM CLI, or
2. a separate Lemonade import/conversion integration for FLM-compatible artifacts.

I’ll clarify this distinction in the PR description so the current scope is explicit.

ramkrishna2910:
One more thing before we merge: let's make sure the Hugging Face-only marketplace search doesn't stay that way unnoticed. The core motivation here (#2602) is users in regions where Hugging Face is unreachable — for exactly those users, the in-app search is a dead end, and CLI/URL/manual-add is the only path. That's an acceptable v1 cut, but it's an important gap, not just a nice-to-have.

@fl0rianr could you file a follow-up issue for ModelScope marketplace search so this stays tracked? Worth capturing in it:

- The current search runs client-side in the renderer against `huggingface.co/api/models` (CORS-friendly). ModelScope's search API likely needs to be proxied through `lemond` instead, so it's a small architecture change, not just a second URL.
- Result mapping/GGUF filtering differs on ModelScope's API.
- Dedup/labeling UX for repos mirrored on both hubs.

No change requested in this PR — everything else tested out well on my end (built the branch, pulled a ModelScope GGUF end-to-end, verified provenance persistence, cache namespacing, and the not-found handling).

fl0rianr:
@ramkrishna2910 Thanks, issue is created, will tackle it after this one here merges.

ramkrishna2910 (approved):

PR #2648, opened 2026-07-10
Feat/dxgi use native DXGI QueryVideoMemoryInfo for Windows VRAM pressure

kyuubyN:
Replaces slow or missing vendor-specific CLI and WMI VRAM queries on Windows with a unified, lightweight, and native DXGI-based global memory pressure monitor.

Key Changes
Native DXGI Monitoring: Enumerates discrete graphics adapters using Windows DXGI (QueryVideoMemoryInfo) to dynamically query VRAM usage and budgets across all major GPU vendors (AMD, NVIDIA, Intel) on Windows 10+.
Dynamic Memory Pressure: Computes the system-wide VRAM utilization ratio using the OS-assigned dynamic memory budget against dedicated physical memory (1.0 - Budget / DedicatedVideoMemory).
Zero-Process Fallback: Injects the DXGI helper at the top of the fallback chain in get_global_vram_usage_pct(), eliminating the overhead of launching external CLI tools like nvidia-smi or slow WMI queries.

### DIFF
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -32,7 +32,10 @@
 #include <Wbemidl.h>
 #include <intrin.h>
 #include "utils/wmi_helper.h"
+#include <dxgi1_4.h>
+#include <wrl/client.h>
 #pragma comment(lib, "wbemuuid.lib")
+#pragma comment(lib, "dxgi.lib")
 #endif
 
 #ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


fl0rianr (changes requested):
Thanks for working on a native Windows solution. I think the core metric needs to be corrected before this can safely drive eviction.

IDXGIAdapter3::QueryVideoMemoryInfo() does not report system-wide VRAM usage. Microsoft defines `Budget` as the OS-assigned video-memory budget for the calling application process, while `CurrentUsage` is that application’s current usage. Therefore:

```cpp
1.0 - memInfo.Budget / desc.DedicatedVideoMemory
```

is neither global `used / total` nor Lemonade’s VRAM usage ratio. The budget may also change as other applications become active, but that still does not make it equivalent to global memory utilization. Since this DXGI result is returned before the existing `nvidia-smi` path, Windows/NVIDIA systems would replace a direct `used / total` measurement with a different, non-comparable value. The existing `0.90` eviction threshold could consequently cause false or missed evictions.

There are two related issues:

* desc.DedicatedVideoMemory > 0 can exclude UMA/iGPU adapters, where dedicated non-shared video memory may be zero.
* Taking the highest value across every adapter can trigger a global Lemonade eviction because of pressure on an unrelated GPU, even when Lemonade is running on another adapter.

Also, because the DXGI values are process-specific, querying them from `lemond` does not represent allocations owned by separate backend processes.

I suggest either preserving the existing vendor-specific global `used / total` sources and treating DXGI as a separately named process-budget signal, or implementing a true adapter-level system-wide usage source and associating it with the adapter used by each backend. Windows coverage should include at least a discrete GPU, an UMA/iGPU system, and a multi-adapter setup.

kyuubyN:
Thanks for the detailed and constructive feedback. You are correct about these edge cases. I've updated the implementation to address all your points:

Moved the DXGI logic to the very end of get_global_vram_usage_pct() to act strictly as a fallback. Windows/NVIDIA systems will continue to use the existing nvidia-smi path without any changes, maintaining compatibility with the 0.90 threshold.

Updated the helper to handle integrated GPUs where DedicatedVideoMemory is zero. It now queries DXGI_MEMORY_SEGMENT_GROUP_NON_LOCAL and uses SharedSystemMemory as the total memory denominator.

If at least one discrete GPU (with DedicatedVideoMemory > 0) is detected during enumeration, we now skip/ignore integrated/UMA adapters in the loop. This prevents background OS tasks on a secondary iGPU from triggering false model evictions on the primary discrete GPU.

Although QueryVideoMemoryInfo returns process-specific values, using it as a last line of defense on Windows+AMD and Windows+Intel setups is a significant improvement over falling back to -1.0 (which disables eviction entirely). As other applications consume VRAM, the OS shrinks the process budget, causing the ratio (1.0 - Budget/Total) to rise and correctly trigger the eviction loop under pressure.

I have compiled and verified this updated logic on an AMD Radeon dGPU (RX 7600) running Windows 11 with ROCm 7.1, and it successfully reports VRAM pressure dynamically without issues.

fl0rianr (approved):
Resolved the merge conflict with the latest `main` changes and applied a small defensive correction to the DXGI fallback. The existing detection paths remain unchanged, with DXGI only used as the final Windows fallback. Very nice work on your side, thanks!

Final implementation now looks good to me. Merging.

PR #2652, opened 2026-07-10
fix(router): handle transient GPU hangs and compute errors in lemond

abn:
### Problem
On virtualized macOS Apple Silicon runners and other heavily loaded GPU environments, transient GPU hangs can occur during LLM inference. This leaves the backend (`llama-server`) in a permanently corrupted state where all subsequent requests fail and return HTTP `500` status with a body of `{"error": {"code": 500, "message": "Compute error.", "type": "server_error"}}`. The server router was not catching this specific error response format, leading to persistent failures until a manual model unload was triggered.

### Solution
This pull request introduces automatic detection and transparent recovery for transient GPU hangs and compute errors:
* **GPU Hang & Compute Error Detection**: Extended `Router::is_watchdog_reset_response` to match error payloads containing `"message": "Compute error."` or `"GPU Hang"`, and `"type": "server_error"`.
* **Forced Eviction & Subprocess Reset**: Modified the router's reload method to explicitly call `evict_server` before attempting reload under the load lock, ensuring the corrupted subprocess is terminated and cleanly recreated. Added a public method to trigger immediate watchdog-based backend resets from the router.
* **SSE Streaming Interception**: Updated `StreamingProxy::forward_sse_stream` to intercept non-200 responses and throw a runtime error rather than feeding invalid bytes to the client sink. If no bytes have been streamed to the client, the error transitions to a `BackendStreamRetryableReset` exception, allowing a transparent reload and retry.

### Testing
* Added `test/test_gpu_hang_retry.py` integration test mocking a transient GPU Hang error response from `llama-server` on the first request, verifying that both streaming and non-streaming requests recover cleanly after a transparent reload on the second request.
* Verified that all 85 existing integration tests pass successfully.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -2113,6 +2113,7 @@ jobs:
           elif [ "${{ matrix.test_type }}" = "streaming-errors" ]; then
             echo "Running streaming error termination tests..."
             $VENV_PYTHON test/server_streaming_errors.py --cli-binary "$CLI_BINARY"
+            $VENV_PYTHON test/test_gpu_hang_retry.py --cli-binary "$CLI_BINARY"
           fi
           echo "${{ matrix.test_type }} tests PASSED!"
 
@@ -2222,6 +2223,7 @@ jobs:
         run: |
           $VENV_PYTHON -m test.utils.reset_server_state --best-effort --label "windows-latest streaming-errors"
           $VENV_PYTHON test/server_streaming_errors.py --cli-binary "$CLI_BINARY"
+          $VENV_PYTHON test/test_gpu_hang_retry.py --cli-binary "$CLI_BINARY"
 
       # Must run last: its tearDown stops lemond via /internal/shutdown to
       # launch its own instance, and nothing after it can rely on the server.
@@ -2368,6 +2370,7 @@ jobs:
         run: |
           $VENV_PYTHON -m test.utils.reset_server_state --best-effort --label "macos-latest streaming-errors"
           $VENV_PYTHON test/server_streaming_errors.py --cli-binary "$CLI_BINARY"
+          $VENV_PYTHON test/test_gpu_hang_retry.py --cli-binary "$CLI_BINARY"
 
       # Must run last: its tearDown stops lemond via /internal/shutdown to
       # launch its own instance, and nothing after it can rely on the server.
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1997,6 +1997,7 @@ jobs:
         run: |
           .venv/bin/python -m test.utils.reset_server_state --best-effort --label "ubuntu streaming-errors"
           .venv/bin/python test/server_streaming_errors.py --cli-binary lemonade
+          .venv/bin/python test/test_gpu_hang_retry.py --cli-binary lemonade
 
       - name: Test jobs
         if: ${{ !cancelled() && steps.setup.outcome == 'success' }}
--- a/src/cpp/include/lemon/router.h
+++ b/src/cpp/include/lemon/router.h
@@ -331,7 +331,7 @@ private:
     WrappedServer* find_server_by_model_name(const std::string& model_name) const;
     WrappedServer* get_most_recent_server() const;
     void prune_unavailable_servers_locked();
-    bool reload_model_after_watchdog_reset(const std::string& requested_model, const RecipeOptions& options);
+    bool reload_model_after_watchdog_reset(const std::string& requested_model, const RecipeOptions& options, uint64_t failed_instance_id = 0);
     bool is_watchdog_reset_response(const json& response) const;
     int count_servers_in_pool(ModelType type, ResidencyClass residency_class,
                               const std::string& model_name) const;
--- a/src/cpp/include/lemon/wrapped_server.h
+++ b/src/cpp/include/lemon/wrapped_server.h
@@ -91,7 +91,10 @@ public:
           active_request_count_(0),
           maintenance_in_progress_(false),
           load_duration_ms_(0),
-          last_backend_activity_(std::chrono::steady_clock::now()) {}
+          last_backend_activity_(std::chrono::steady_clock::now()),
+          instance_id_(next_instance_id()) {}
+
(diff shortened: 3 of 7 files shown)


fl0rianr (changes requested):
The overall recovery approach looks good, especially the SSE interception and the restriction that streaming requests are only retried before any bytes reach the client.

However, I think the current reload path still has a concurrency issue. reload_model_after_watchdog_reset() unconditionally evicts whichever server is currently registered for the model. With multiple requests failing at the same time, one request may already have replaced the broken backend while another recovery path then finds and evicts that newly healthy instance again. Because this happens while holding `load_mutex_`, and eviction of a live busy server may wait indefinitely, this can also block other model-management operations.

The reload should only evict the exact failed instance, or recovery should be serialized per model so concurrent failures share a single reload result. A regression test with two simultaneous failing requests would be important here.

The new `test_gpu_hang_retry.py` also does not appear to be invoked by the main C++ server CI workflow, so the new behavior is currently not protected by CI. Please add it to the workflow or integrate the cases into an already executed streaming/backend recovery test.

abn:
I have updated the branch to address the concurrency feedback:

1. **Watchdog Reload Concurrency**: Updated `Router::reload_model_after_watchdog_reset` to accept the `failed_server` pointer and compare it with the currently registered backend (`existing`). If a concurrent request has already evicted/reloaded the failed backend, subsequent requests will skip eviction and re-use the newly healthy instance.
2. **Concurrent Regression Test**: Added `test_concurrent_gpu_hang_recovery` to `test_gpu_hang_retry.py` simulating parallel requests using a thread pool. The test asserts that both concurrent requests succeed and exactly one reload is triggered.
3. **CI Pipeline Integration**: Integrated `test_gpu_hang_retry.py` into `cpp_server_build_test_release.yml` to run alongside `server_streaming_errors.py` on all platform runners (Linux, macOS, Windows).

fl0rianr (changes requested):
Thanks for addressing the previous feedback. The CI integration and the added concurrent recovery test are good improvements, and the new instance comparison prevents the most obvious case where a second recovery evicts an already reloaded backend.

I still see a few issues that should be addressed before merging:

1. failed_server is passed as a raw pointer after the request has released its inference reference. Another thread may already have removed and destroyed that server object. Comparing the stale pointer value with the currently registered instance is therefore not a fully reliable identity check and can theoretically suffer from pointer reuse. A stable server generation/instance ID or a per-model single-flight recovery mechanism would be safer.

2. There is also a pinning race. If a concurrent recovery finds no registered instance while another thread is still loading the replacement, `was_pinned` remains `false`. The later `load_model(..., pinned=false)` call can then find the newly loaded instance and unintentionally clear its pinned state.

3. The concurrent regression test does not deterministically reproduce two requests failing on the same backend instance. The mock uses the single-threaded `HTTPServer` and has no barrier ensuring that both requests reach the original backend before either receives the error. The test can therefore pass with one request triggering the reload and the second request simply using the replacement. A threaded mock plus a barrier for the first two failing requests would properly exercise the race.

4. The new test is now executed on Windows, but the mock backend is created as a Unix-style executable Python script named `llama-server` using a shebang and `chmod`. Lemonade starts backend binaries directly through `CreateProcess` on Windows, so this file will not be executable there. The PATH construction also uses `:` instead of `os.pathsep`. The test needs a Windows-compatible launcher/mock or should only be enabled on supported platforms.

There is also a smaller cleanup issue in the startup failure path: `communicate(timeout=1)` is called without terminating the still-running lemond process first, which may raise `TimeoutExpired` and hide the intended diagnostic.

The overall recovery design is heading in the right direction, but I would still request changes until the recovery identity, pinning race, deterministic concurrency coverage, and Windows test execution are resolved.

fl0rianr (commented):
I do not see any remaining blocking issue in the current implementation. One minor test improvement would be to enable the barrier only for the concurrent test, since the single-request tests currently wait for its timeout unnecessarily.

From a code-review perspective, this looks good to me. Approval is appropriate once the latest C++ server CI run completes successfully.

abn:
Thanks for the review @fl0rianr. Addressed your last point as well.

fl0rianr (approved):
I reviewed the latest update and do not see a remaining production-code blocker. 

One non-blocking test gap remains: `BrokenBarrierError` is swallowed, so the concurrent test can still pass if only one request reaches the original backend and the second request reaches the replacement after the barrier timeout. Recording and asserting that exactly two requests were handled by backend attempt 1 would make the regression deterministic.

It would also be useful to execute this test in the macOS workflow, since the original failure was observed on virtualized Apple Silicon runners. The Windows skip is reasonable with the current mock launcher.

The current unrelated/flaky CI failure aside, this is ready for approval from my perspective.

abn:
@fl0rianr I addressed your non-blockers as well.

1. **Mock Server Request Counting (`test_gpu_hang_retry.py`)**:
   - Added thread-safe tracking of the number of requests that hit the backend during `attempt == 1` using a `threading.Lock()` to write/read progress to the `state.json` file.
   - Fixed the state persistence logic so that subsequent reload attempts (e.g. `attempt == 2`) do not wipe existing keys like `attempt_1_requests` when starting up.
   - Added a deterministic assertion verifying that exactly `2` requests hit the first attempt of the backend. If the barrier synchronization fails, the test will now fail deterministically.
   - Refactored the test runner to use `run_server_tests` to prevent CLI argument parsing conflicts with `unittest.main`.

2. **macOS & Linux ARM64 CI Execution (`cpp_server_build_test_release.yml`)**:
   - Added `test_gpu_hang_retry.py` to run alongside `server_streaming_errors.py` on all platform test configurations, including macOS (unsigned path) and Linux ARM64.

abn:
@fl0rianr is this good to merge once CI passes?

fl0rianr:
Yes — from my review, this is good to merge once the C++ server CI passes.

The production-code concerns from my earlier reviews have been addressed, and the latest changes appropriately strengthen the concurrent regression test and add the requested macOS coverage.

The current macOS failure occurs in the streaming error test step where this new test now runs, so I would first want that run to pass—or the failure to be clearly identified as unrelated—before merging. No further production-code changes are required from my side.

abn:
@fl0rianr good instinct. The issue was mine.

abn:
@fl0rianr re-requesting a review as we made some changes after the last approval.

fl0rianr (approved):
Approval still holds

PR #2656, opened 2026-07-10
feat(ui): add build-time LEMONADE_BASE_URL override for prototype

abn:
This PR adds support for a build-time connection URL override in the Lemonade UI redesign prototype.

When deploying the UI prototype to a public static host like https://lemonade-taupe.vercel.app/, the dynamic hostname rewrite logic hijacks any `localhost`/`127.0.0.1` connection addresses and rewrites them to the deployment domain. By introducing a build-time `LEMONADE_BASE_URL` environment variable and injecting it via Webpack's `DefinePlugin`, we allow hosted deployments to explicitly specify a default connect address (such as `http://localhost:13305`) and bypass dynamic hostname replacement.

### DIFF
--- a/prototype/ui-redesign/src/api.ts
+++ b/prototype/ui-redesign/src/api.ts
@@ -20,7 +20,7 @@ function detectDefaultBaseUrl(): string {
   return 'http://127.0.0.1:13305';
 }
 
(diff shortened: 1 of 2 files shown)


fl0rianr (approved):
LGTM

PR #2657, opened 2026-07-10
fix(ui): preserve user-configured custom connection URLs in prototype

abn:
This PR ensures that connection settings manually configured by the user via the Connect tab are preserved and not dynamically rewritten.

Currently, the mobile LAN hostname substitution check runs unconditionally on any resolved `localhost` or `127.0.0.1` address. When visiting a hosted deployment such as https://lemonade-taupe.vercel.app/#/connect and attempting to manually input a local Lemonade API server at `localhost`, the hostname is dynamically hijacked and rewritten to `lemonade-taupe.vercel.app`, breaking the connection.

This change checks if a custom URL is stored in `_hostBaseUrl` or `localStorage` (`lemonade_base_url`), and bypasses the mobile LAN hostname rewrite if the user has explicitly saved their settings.

### DIFF
--- a/prototype/ui-redesign/src/api.ts
+++ b/prototype/ui-redesign/src/api.ts
@@ -560,12 +560,13 @@ class LemonadeAPI {
 
   get baseUrl(): string {
     try {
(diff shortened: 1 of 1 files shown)


fl0rianr (approved):
LGTM

PR #2659, opened 2026-07-11
feat: make model update checks manual configerable

### LINKED ISSUE #2608: Option to check for updates only manually
### Feature Description

On startup, `lemond` makes quite a few connections to `connectivitycheck.android.com` and `huggingface.co`.  Looks like it's doing that to check for updates to all downloaded models.

Please add the option to disable automatic update checks and let the user manually check for updates.

### Use Case / Motivation

On systems where `lemond` is not a long-lived process, always checking for updates on every startup generates unnecessary internet traffic.  Letting the user control what internet connections happen when helps minimize data usage on metered connections, simplifies auditing internet traffic, and reduces chances of being rate-limited or otherwise flagged by the site.

For the same reasons it would be ideal if, more generally, `lemond` could be configured to only make internet connections in response to user actions that require it (e.g. user request to pull a model or backend, or user using a cloud backend that points to a remote server).

### Platform Relevance

All platforms

### Additional Context

Tried `lemonade config set offline=true` but surprisingly, it had no effect on this.  It's anyway not quite the right tool for this job since `offline=true` would also prevent manual downloads/updates of models and backends.

fl0rianr:
## Summary

Adds control over automatic model update checks and introduces an explicit manual update-check command.

## Changes

* Add auto_check_model_updates configuration option, enabled by default for backward compatibility
* Skip the startup model update check when the option is disabled
* Add `lemonade check-updates` for manually checking downloaded models
* Add `POST /v1/models/check-updates`
* Keep manual model downloads available when automatic checks are disabled
* Respect `offline=true` for all update checks
* Prevent concurrent startup and manual update checks
* Clear stale update flags after a successful up-to-date response
* Document the new configuration option, CLI command, and API endpoint

## Why

Some users need Lemonade to avoid unsolicited network requests during startup without enabling full offline mode, which would also prevent intentional model downloads.

This change separates automatic update checks from explicit user-triggered network actions.

Closes #2608

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -18,6 +18,7 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `POST` | [`/v1/unload`](#post-v1unload) | Unload a model |
 | `POST` | [`/v1/audio/generations`](#post-v1audiogenerations) | Generate audio (music or sound effects) from a text prompt |
 | `POST` | [`/v1/3d/generations`](#post-v13dgenerations) | Generate a textured 3D mesh (GLB) from an image |
+| `POST` | [`/v1/models/check-updates`](#post-v1modelscheck-updates) | Manually check downloaded models for upstream updates |
 | `GET` | [`/v1/models/{id}/files`](#get-v1modelsidfiles) | List resolved local file metadata for one model |
 | `GET` | [`/v1/health`](#get-v1health) | Check server status, such as models loaded |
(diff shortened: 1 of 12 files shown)


bitgamma:
@fl0rianr why is there a .patch file in the commit? I think something went wrong

fl0rianr:
@bitgamma thanks for catching this one - should have seen the slip since it's actually a concentrated change. Removed exactly this patch. Might have a look on it again?

bitgamma (approved):
lgtm

PR #2660, opened 2026-07-11
feat(server): support PNA CORS, custom allowed origins, and custom headers

abn:
This PR introduces support for Private Network Access (PNA) preflight responses, customizable WebSocket allowed origins, and custom CORS headers in lemond.

When deploying web applications to public secure hosting (e.g. over HTTPS), connecting to local or LAN backend instances via WebSocket/HTTP triggers strict browser-level security checks:
1. **Private Network Access (PNA)**: Public sites calling local loopback addresses require the server to respond with `Access-Control-Allow-Private-Network: true` in the OPTIONS preflight response.
2. **WebSocket Origins**: The WebSocket server rejects connections from unauthorized origins by default. By checking the `LEMONADE_ALLOWED_ORIGINS` environment variable (which accepts `*` or a comma-separated list of complete origins including the scheme and optional port), we allow users to explicitly authorize public web clients to upgrade WebSockets. Local/loopback origins are always allowed by default, but non-local origins must explicitly match the configured allowlist.
3. **Custom Headers**: Adds `mcp-protocol-version` to `Access-Control-Allow-Headers` to support MCP-client connections.


src/cpp/server/websocket_server.cpp
```diff
@@ -120,6 +120,42 @@ int WebSocketServer::ws_callback(struct lws* wsi,
                     is_allowed = true;
                 }
 
+                if (!is_allowed) {
+                    const char* env_origins = std::getenv("LEMONADE_ALLOWED_ORIGINS");
+                    if (env_origins) {
+                        std::string origins_str(env_origins);
+                        if (origins_str == "*") {
+                            is_allowed = true;
```
fl0rianr: This comparison is too permissive for an origin allowlist.

Removing the scheme and port means that an entry such as:

`https://example.com:443`

also permits:

`http://example.com`
`https://example.com:8080`

Those are different browser origins. Please parse and compare the complete normalized origin, including scheme, host, and effective port, rather than comparing only `origin_host`.

src/cpp/server/websocket_server.cpp
```diff
@@ -120,6 +120,42 @@ int WebSocketServer::ws_callback(struct lws* wsi,
...
+                        } else {
+                            std::stringstream ss(origins_str);
+                            std::string item;
+                            while (std::getline(ss, item, ',')) {
+                                std::string allowed_host = item;
+                                size_t scheme_pos = allowed_host.find("://");
+                                if (scheme_pos != std::string::npos) {
+                                    allowed_host = allowed_host.substr(scheme_pos + 3);
+                                }
+                                size_t port_pos = allowed_host.find(':');
```
fl0rianr: This port parsing breaks IPv6 origins because the first `:` is part of the IPv6 address.

For example, `[::1]` or `https://[2001:db8::1]:8443` will be truncated incorrectly. The existing parsing logic above already handles bracketed IPv6 addresses separately, so it would be safer to reuse one shared origin-parsing helper for both the request origin and configured allowlist entries.

fl0rianr (changes requested):
Thanks for working on this. The general direction makes sense, but I do not think this is safe to merge yet.

The main issue is that LEMONADE_ALLOWED_ORIGINS is currently treated as a host allowlist rather than an origin allowlist. The implementation removes both the scheme and port before comparison. As a result, allowing `https://app.example.com:443` would also allow origins such as `http://app.example.com` or `https://app.example.com:8443`.

For WebSocket origin validation, scheme, host, and port are security-relevant and should be compared as a complete normalized origin.

The PNA header also should not be added unconditionally to every response together with `Access-Control-Allow-Origin: *`. This explicitly permits any website to make private-network requests in browsers that honor this header. It should only be returned for a matching private-network preflight after validating the requesting origin against the configured allowlist.

This also needs focused tests covering exact origin matching, different schemes, different ports, multiple configured origins, whitespace, wildcard behavior, and IPv6.

fl0rianr (changes requested):
Thanks for addressing the original review points. The shared parser, IPv6 handling, exact scheme/port matching for fully specified entries, conditional PNA response, and new tests are all meaningful improvements.

I still see two security concerns that should be resolved before merge:

1. Host-only entries such as LEMONADE_ALLOWED_ORIGINS=app.example.com are still treated as wildcards for both scheme and port. This allows http://app.example.com, https://app.example.com, and arbitrary ports. Since this setting is named ALLOWED_ORIGINS, entries should ideally require a valid scheme and match the complete normalized origin, including the effective port. If host-wide matching is intentional, it should use an explicit and documented wildcard syntax.

2. The `Origin` versus `Host` fallback should not be shared with the HTTP CORS/PNA policy. Matching only the hostname can approve requests across different schemes and ports, and it weakens the DNS-rebinding protection that private-network preflights are intended to provide. HTTP/PNA responses should require an explicit allowlist match before returning `Access-Control-Allow-Origin` or `Access-Control-Allow-Private-Network: true`. If the fallback is needed for WebSocket compatibility, that policy should be handled separately.

Because `Access-Control-Allow-Origin` is now returned dynamically, the response should also include `Vary: Origin`.

The unit tests are a good addition, but an integration test for the actual HTTP preflight response would still be valuable, especially to verify that rejected origins do not receive the private-network opt-in.

Overall, this is much closer, but I would keep the changes requested until these remaining policy issues are addressed.

abn:
Thanks for the feedback! I've updated the implementation:

- **Strict origin matching**: Configured entries in `LEMONADE_ALLOWED_ORIGINS` now require a valid scheme and match the complete origin (scheme, host, and effective port).
- **CORS/PNA isolation**: Removed the same-origin `Host` header fallback from HTTP CORS and Private Network Access (PNA) preflights. It now relies strictly on loopback or explicit allowlist matches.
- **WebSocket fallback**: Retained same-origin `Host` header checking exclusively for WebSockets.
- **Vary & loopbacks**: Added `Vary: Origin` to dynamic responses and added `tauri.localhost` to the loopback check to support Windows Tauri WebView2.
- **Verification**: Expanded C++ unit tests to cover strict scheme/port matching and added an integration test (`test_011_http_preflight_cors_and_pna` in `server_websocket_telemetry.py`) verifying OPTIONS preflights against allowed and rejected origins.

fl0rianr (commented):
Thanks for addressing the previous review points. The strict origin matching, separation of HTTP CORS/PNA from the WebSocket policy, conditional PNA response, `Vary: Origin`, IPv6 handling, and added tests are all solid improvements.

I still see one remaining security concern before approval: the WebSocket path accepts any non-local origin when `Origin` and the request-provided `Host` match. Both values are controlled by the incoming request, so this can bypass `LEMONADE_ALLOWED_ORIGINS` in a DNS-rebinding scenario. This matters especially when `LEMONADE_API_KEY` is not configured, because the WebSocket connection is then accepted without authentication.

For non-local WebSocket origins, I think access should require an explicit `LEMONADE_ALLOWED_ORIGINS` match. Alternatively, the same-origin fallback would need to be based on trusted server-side configuration rather than comparing two request headers.

Two smaller follow-ups:

* Please also remove `LEMONADE_ALLOWED_ORIGINS` from the inherited environment in the WebSocket integration test setup so tests are isolated from the developer or runner environment.
* The PR description currently says the variable accepts a list of hosts, while the implementation now correctly requires complete origins including the scheme. The configuration format and the security implications of `*` should be documented.

With the WebSocket fallback secured and those small cleanup items addressed, this looks ready from my side.

fl0rianr (commented):
Thanks for addressing the previous review. The DNS-rebinding-prone WebSocket fallback has now been removed, the test environment is isolated, the PR description is accurate, and the new configuration is documented. The original security concerns are resolved.

I found one remaining fail-open case in the origin parser that should be fixed before approval. When an explicitly supplied port overflows `std::stoi`, the parser sets `port` back to `-1`, which is then interpreted as the scheme's default port. For example, an invalid allowlist entry such as `https://app.example.com:999999999999999` can therefore match the valid origin `https://app.example.com:443`.

The bracketed IPv6 path also calls `std::stoi` without checking that the complete port string was consumed, so a malformed entry such as `https://[2001:db8::1]:443junk` can be accepted as port 443. Trailing characters after the closing bracket should also invalidate the entry.

Because this parser protects a security allowlist, any malformed explicit port or authority should invalidate the entire origin rather than falling back to a default. Please validate full numeric consumption and the valid port range, and add tests for overflow, trailing characters, malformed IPv6 authorities, and out-of-range ports.

As cleanup, I would also remove the now-unused `is_same_origin()` helper, its tests, and the unused `host_header` parameter/read. Keeping that helper around contradicts the new security invariant and makes accidental future reuse more likely.

One documentation detail: the new Allowed Origins section currently replaces the existing “Client Behavior” sentence about API-key selection. Please preserve that sentence and insert the new section after it.

With the strict malformed-origin handling fixed, this looks ready from my side.

fl0rianr (commented):
Thanks for addressing the previous review. The malformed-port handling is now fail-closed, the obsolete same-origin WebSocket helper has been removed, and the new parser tests cover the reported overflow and IPv6 cases.

I found one remaining security blocker: disallowed HTTP origins are not actually rejected. The pre-routing handler only omits the CORS response headers and then continues through authentication to the requested route. When `LEMONADE_API_KEY` is unset, the request can therefore still be executed.

This is particularly important for `/mcp`. Its POST handler parses `req.body` as JSON without validating the request `Content-Type`, so a browser can submit a CORS-safelisted `text/plain` POST without a preflight. The browser may be unable to read the response, but the MCP request and any resulting tool action can still execute. The same behavior also leaves the HTTP path vulnerable to DNS rebinding.

The MCP Streamable HTTP transport requires the server to validate `Origin`; current specification language explicitly requires HTTP 403 when an Origin header is present but invalid. Please return a handled 403 response before authentication and route dispatch whenever an incoming HTTP Origin does not match the loopback policy or `LEMONADE_ALLOWED_ORIGINS`.

The integration coverage should verify:

* a disallowed OPTIONS request returns 403;
* a disallowed actual POST, including a `text/plain` POST to `/mcp`, returns 403 and is not dispatched;
* an explicitly configured remote origin succeeds;
* clients without an Origin header continue to work.

As a smaller hardening improvement, I would also reject allowlist entries containing paths, queries, fragments, or userinfo instead of silently reducing them to a broader origin. For example, `https://example.com/admin` currently becomes `https://example.com`, which may grant more access than the administrator intended.

Once unauthorized HTTP requests are actively rejected, this looks ready from my side.

abn:
Thanks for the feedback! I've addressed the remaining security blocker and parser concerns:

- **Active HTTP Rejection**: Updated the pre-routing handler to return a handled `403 Forbidden` response immediately when an incoming request carries a disallowed `Origin` header. This blocks unauthorized cross-origin requests (including same-origin bypass attempts and `/mcp` `text/plain` POST requests) before they can reach authentication or routing.
- **Harden Allowlist Parsing**: Updated the parser to fail-closed on any configured allowlist entries containing paths (except an empty trailing slash), queries, fragments, or userinfo.
- **Expanded Integration Coverage**: Updated `test_011_http_preflight_cors_and_pna` to assert that:
  - Disallowed `OPTIONS` preflights return `403`.
  - Disallowed `POST` requests (including a `text/plain` POST to `/mcp`) return `403` and are not routed.
  - Allowed remote origins and loopbacks succeed normally.
  - Clients without an `Origin` header (CLI/scripts) continue to work unhindered.

fl0rianr (commented):
Thanks for addressing the previous review. The remaining security blocker is resolved: disallowed HTTP origins are now rejected with 403 before authentication and route dispatch, including actual POST requests and a preflight-free `text/plain` POST to `/mcp`. The configured remote-origin path, clients without an Origin header, and the malformed-origin parser cases are also covered.

I found one final CORS correctness issue before approval: `Vary: Origin` is currently set only when the incoming request already contains an `Origin` header.

A response to a request without `Origin` can therefore be cached without `Vary`. A later CORS request for the same resource may reuse that cached response, which will not contain `Access-Control-Allow-Origin`, causing an allowed request to fail incorrectly. This is the exact cache scenario that `Vary: Origin` is intended to prevent.

Please set or append `Vary: Origin` for all responses whose CORS behavior can vary by Origin, including responses to requests without an Origin header. Existing `Vary` values should be preserved rather than overwritten. A regression test should first request an endpoint without Origin and verify that the response still contains `Vary: Origin`.

Two non-blocking follow-ups would improve coverage and documentation:

* Add an end-to-end WebSocket test showing that a configured non-local origin succeeds and a scheme or port mismatch fails.
* Consider warning that allowing a non-local plain-HTTP origin is vulnerable to on-path modification; HTTPS should be recommended for remote origins.

With the unconditional `Vary: Origin` handling corrected, I have no remaining blocking findings.

abn:
Thanks for the feedback! I've addressed the CORS cache correctness issue:

- **Unconditional `Vary: Origin`**: Updated the pre-routing handler to set `Vary: Origin` unconditionally on all HTTP responses, even when the request does not carry an `Origin` header. Existing `Vary` header values are preserved and appended to.
- **Regression Test Coverage**: Expanded the integration tests in `test_011_http_preflight_cors_and_pna` to assert that a request without an `Origin` header receives `Vary: Origin` in its response.

I've also addressed the two non-blocking follow-up items:

- **WebSocket Origin Integration Tests**: Added a dedicated test (`test_012_websocket_origin_security` in `test/server_websocket_telemetry.py`) verifying that configured non-local WebSocket origins succeed, and that mismatches in scheme/port or completely unconfigured origins fail to upgrade.
- **Remote Plain-HTTP Warning**: Added a security warning to the `LEMONADE_ALLOWED_ORIGINS` section in `docs/guide/configuration/README.md` highlighting that remote plain-HTTP configurations are susceptible to MITM/on-path modifications and recommending HTTPS.

fl0rianr (approved):
Thanks, we can go with this now

PR #2664, opened 2026-07-11
fix(server): forward ctx_size to auto-loaded models

### LINKED ISSUE #2663: Auto_load_model ignores ctx_size in chat request body
# Auto-load ignores `ctx_size` in chat request body

**Platform**: Linux/Fedora (Bazzite)  
**Lemonade Version**: 10.10.0  
**GPU / APU Model**: AMD RYZEN AI MAX+ 395 (AMD Radeon 8060S Graphics, Integrated)  
**Component**: lemond (server)

---

## Bug Description

When the server auto-loads a model via the `/v1/chat/completions` endpoint, it ignores the `ctx_size` parameter in the request body. The model is loaded using `recipe_options.json` defaults, regardless of what `ctx_size` the client specifies in the chat request.

This causes failures for clients with large system prompts that exceed the auto-loaded context size. The client explicitly requests `ctx_size: 128000` in its request body, but the model auto-loads at `4096`, causing the request to be rejected with a context size exceeded error.

The `/v1/load` endpoint correctly accepts `ctx_size` as a request parameter and can override the default. However, the chat completions endpoint's auto-load path does not forward request body parameters to the loading phase.

## Steps to Reproduce

1. Start Lemonade server with a model (e.g., `Qwen3-Coder-Next-GGUF-Q4_K_M`) that has a small default `ctx_size` (`4096`) in its `recipe_options.json`
2. Ensure the model is NOT pre-loaded (unload it or start fresh):
   ```bash
   curl -X POST http://localhost:13305/v1/unload \
     -d '{"model_name": "Qwen3-Coder-Next-GGUF-Q4_K_M"}'
   ```
3. Send a POST to `/v1/chat/completions` with `"ctx_size": 128000` in the request body:
   ```bash
   curl -X POST http://localhost:13305/v1/chat/completions \
     -H "Content-Type: application/json" \
     -d '{
       "model": "Qwen3-Coder-Next-GGUF-Q4_K_M",
       "messages": [{"role": "user", "content": "Hello"}],
       "max_tokens": 10,
       "ctx_size": 128000
     }'
   ```
4. Check the loaded model's context size:
   ```bash
   curl http://localhost:13305/v1/models/Qwen3-Coder-Next-GGUF-Q4_K_M | jq '.recipe_options.ctx_size'
   ```
5. Send a request with `prompt_tokens > 4096` (e.g., a large system prompt) — it fails with context size exceeded

**Expected**: The model should be auto-loaded with `ctx_size: 128000` as specified in the chat request body, matching the behavior of explicitly calling `/v1/load` with `ctx_size: 128000`.

**Actual**: The model is auto-loaded with `ctx_size: 4096` (the `recipe_options.json` default). The `ctx_size` parameter in the chat request body is not forwarded to the model loading phase.

## Log Output

Debug logging enabled (`lemonade config set log_level=debug`). Logs from the auto-load at `ctx_size=4096` followed by the request rejection:

```
2026-07-11 13:31:35.407 [Info] (Server) Auto-loading model: Qwen3-Coder-Next-GGUF-Q4_K_M
2026-07-11 13:31:35.407 [Debug] (Router) Loading model: user.Qwen3-Coder-Next-GGUF-Q4_K_M
2026-07-11 13:31:35.407 [Debug] (Router) Effective settings: ctx_size=4096, llamacpp_backend=vulkan, ...
2026-07-11 13:31:36.183 [Debug] (ProcessManager) Starting process: llama-server ... --ctx-size 4096 --port 8001 ...
2026-07-11 13:31:53.829 [Info] (Server) Model loaded successfully: Qwen3-Coder-Next-GGUF-Q4_K_M
2026-07-11 13:31:53.892 [Info] (Process) cache state: 0 prompts (limits: 8192.000 MiB, 4096 tokens)
2026-07-11 13:31:53.892 [Error] (Process) E srv send_error: task id = 0, error: request (8601 tokens) exceeds the available context size (4096 tokens), try increasing it
2026-07-11 13:31:53.893 [Error] (StreamingProxy) Backend returned error: 400
```

**Key observation:** The model auto-loaded with `--ctx-size 4096` from `recipe_options.json`, then rejected an 8601-token request with the exact error: "request (8601 tokens) exceeds the available context size (4096 tokens)". The client had sent `ctx_size: 128000` in the request body, but it was ignored during auto-load.

### Source Code Analysis

The root cause is in `auto_load_model_if_needed()` in `src/cpp/server/server.cpp` (line ~1855). It calls:

```cpp
router_->load_model(requested_model, info, RecipeOptions(info.recipe, json::object()), true);
```

The `json::object()` passes an **empty options object** to `RecipeOptions`, completely ignoring all parameters from the request body (including `ctx_size`). This is hardcoded and does not accept per-request overrides.

Compare with the `/v1/load` endpoint, which reads `ctx_size` from the request body and passes it correctly:
```cpp
// /v1/load path:
auto options = RecipeOptions(info.recipe, request_body); // ← reads request body
router_->load_model(model_name, info, options, true);
```

### Workaround

Users can pre-configure the model's `ctx_size` by loading it with `save_options: true`:
```bash
curl -X POST http://localhost:13305/v1/load \
  -d '{"model_name": "Qwen3-Coder-Next-GGUF-Q4_K_M", "ctx_size": 128000, "save_options": true}'
```

This is a manual step that shouldn't be necessary. The workaround persists the `ctx_size` to `recipe_options.json`, which then applies to all future auto-loads.

### Affected Clients

Any client that:
- Uses the OpenAI-compatible `/v1/chat/completions` endpoint
- Sends a `ctx_size` parameter in the request body
- Relies on server-side model auto-load (no pre-loaded model)
- Has a system prompt that exceeds the default `ctx_size` (typically 4096)

Known affected client: **pi** (pi-coding-agent), which sends a ~8600 token system prompt + user message. When auto-loaded at `ctx_size: 4096`, the request fails, causing the client to see a "Stream ended without finish_reason" error with zero output tokens.

### Testing Evidence

- **curl + `ctx_size: 128000` (model pre-loaded at correct size)**: Works perfectly — response returned with correct tokens.
- **curl + `ctx_size: 128000` (model auto-loaded at 4096)**: Model loads at 4096, large requests fail.
- **pi (via OpenAI SDK) + `ctx_size: 128000` (model auto-loaded at 4096)**: Fails with `stopReason: error`, `input: 0`, `output: 0` — the stream returns zero chunks because the backend rejects the oversized prompt.
- **pi + server-side `recipe_options.json` set to `ctx_size: 128000`**: Works correctly — auto-load uses the persisted size.

indistinctTalk:
## Problem

When the server auto-loads a model via an inference endpoint, it ignores the `ctx_size` in the request body. The model loads with `recipe_options.json` defaults (usually 4096), regardless of what the client requests.

This affects the OpenAI-compatible REST endpoints (`/v1/chat/completions`, `/v1/completions`, etc.) and the Ollama-compatible layer (`/api/chat`, `/api/generate`, `/api/embed`, `/api/embeddings`, plus the Anthropic-compatible `/v1/messages` handler inside `OllamaApi`). Ollama clients set context size via the standard `options.num_ctx` field, which the auto-load path did not read.

## Root Cause

Both `Server::auto_load_model_if_needed()` and `OllamaApi::auto_load_model()` hardcoded an empty JSON object when constructing `RecipeOptions`:

```cpp
// Before
router_->load_model(requested_model, info, RecipeOptions(info.recipe, json::object()), true);
```

All recipe defaults are used; no per-request overrides are forwarded. The `/v1/load` endpoint, by contrast, passes the request body to `RecipeOptions`.

## Fix

1. Add an optional `request_options` parameter to both auto-load helpers (default `json::object()` for backward compatibility).

2. Introduce `extract_auto_load_options()` in both `Server` and `OllamaApi`. It allowlists `ctx_size` only; request-scoped fields (`temperature`, `max_tokens`, `messages`, `top_p`, `steps`, etc.) are excluded so they cannot leak into persistent model-load settings. The Ollama copy also maps `options.num_ctx` to `ctx_size`, with top-level `ctx_size` winning (matching `map_ollama_options` precedence).

3. Update all REST inference endpoints (11 call sites) and all Ollama inference endpoints (4 call sites) to pass the extracted options. The Anthropic `/v1/messages` handler, which routes through `OllamaApi`, is updated too.

4. The already-loaded early-return is unchanged, so options from subsequent auto-load requests are ignored. This preserves the invariant that explicit `/v1/load` options cannot be overturned by auto-load. A DEBUG log now reports the ignored requested `ctx_size` alongside the loaded model's current value when a request carries `ctx_size` against an already-loaded model.

## Files Changed

- `src/cpp/include/lemon/server.h` — `auto_load_model_if_needed()` signature, `extract_auto_load_options()` declaration
- `src/cpp/server/server.cpp` — REST inference endpoints; DEBUG log on already-loaded ctx_size mismatch
- `src/cpp/include/lemon/ollama_api.h` — `auto_load_model()` signature, `extract_auto_load_options()` declaration
- `src/cpp/server/ollama_api.cpp` — Ollama inference endpoints; `options.num_ctx` → `ctx_size` mapping; DEBUG log
- `src/cpp/server/anthropic_api.cpp` — Anthropic messages handler
- `test/server_endpoints.py` — REST regression test
- `test/test_ollama.py` — Ollama regression test

## Testing

Built locally with `cmake --build --preset default`; zero errors or warnings. Both regression tests run end-to-end against a live `lemond` (CPU llama-server) on Linux (Fedora 42) and pass:

- **REST** (`test_013_auto_load_forwards_only_allowlisted_options`): non-streaming chat completion with `ctx_size=8192` plus request-scoped params (`temperature`, `max_tokens`, `top_p`, `seed`, …) and valid llamacpp recipe-level fields (`pinned`, `llamacpp_args`, `auto_evict`, `evict_idle_timeout`). Verifies `recipe_options` contains `ctx_size=8192` and none of the others. The recipe-level assertions are the real boundary the allowlist closes, since `RecipeOptions` already discards non-recipe keys. `try`/`finally` unloads the model.

- **Ollama** (`test_015a_auto_load_forwards_only_ctx_size`): `/api/chat` with the standard `options.num_ctx=8192` plus `options.num_predict`, `temperature`, `top_p`, and the same recipe-level fields. Verifies `recipe_options` contains `ctx_size=8192` and that `temperature`, `top_p`, `num_predict`, and the recipe-level fields do not leak.

Untested locally: Windows, macOS, and GPU backends (Vulkan/ROCm/CUDA). The change is platform-agnostic C++ with no platform guards, so cross-platform risk is minimal.

Lambda callbacks (MCP gateway, `CollectionOrchestrator`) retain the default parameter and are unaffected by design.

## Related

Fixes: #2663.

### DIFF
--- a/src/cpp/include/lemon/ollama_api.h
+++ b/src/cpp/include/lemon/ollama_api.h
@@ -41,7 +41,8 @@ private:
     void register_anthropic_routes(httplib::Server& server, const std::shared_ptr<OllamaApi>& self);
 
     // Helpers
-    void auto_load_model(const std::string& model);
+    void auto_load_model(const std::string& model, const json& request_options = json::object());
+    static json extract_auto_load_options(const json& request);
     std::string normalize_model_name(const std::string& name);
     json build_ollama_model_entry(const std::string& id, const ModelInfo& info);
     json convert_openai_chat_to_ollama(const json& openai_response, const std::string& model);
--- a/src/cpp/include/lemon/server.h
+++ b/src/cpp/include/lemon/server.h
@@ -248,17 +248,10 @@ private:
                                   httplib::Response& res,
                                   nlohmann::json& out);
 
-    // Helper function for auto-loading models (eliminates code duplication and race conditions)
-    // Optionally accepts per-request override options (e.g. ctx_size) that are applied only on
-    // the very first load.  If the model is already loaded these are silently ignored so as not
-    // to overturn options that were set by a prior explicit /v1/load call.
-    //
-    // Callers MUST extract only load-level options via extract_auto_load_options() before
-    // passing them here — never pass the raw inference request body.  This prevents
-    // request-scoped fields (temperature, max_tokens, messages, steps, etc.) from leaking
-    // into recipe options that would affect subsequent requests.  Currently only ctx_size
-    // is allowlisted; pinned, eviction settings, *_args, and image params are excluded.
-    // (They can be forwarded via /v1/load when needed.)
+    // Auto-load a model on first use. request_options are applied only on the first load;
+    // if the model is already loaded they are ignored so explicit /v1/load settings win.
+    // Callers must pass only load-level options from extract_auto_load_options() — never
+    // the raw request body — to keep request-scoped fields out of persistent recipe options.
     void auto_load_model_if_needed(const std::string& model_name,
                                    const json& request_options = json::object());
 
@@ -342,11 +335,8 @@ private:
     // Set to true after check_for_model_updates() completes at startup.
     std::atomic<bool> update_check_done_{false};
 
-    // Extract only load-level options from an inference request body.  Returns a new json
-    // containing a small, explicit allowlist of fields that are safe to forward to the
-    // RecipeOptions constructor during auto-load.  All other request-scoped fields
-    // (temperature, max_tokens, messages, steps, cfg_scale, etc.) are deliberately excluded
-    // so they cannot accidentally influence persistent model-load settings.
+    // Extract load-level options from an inference request body. Currently only ctx_size
+    // is forwarded; request-scoped fields are excluded so they cannot leak into recipe options.
     static json extract_auto_load_options(const json& request);
 };
 
--- a/src/cpp/include/lemon/server.h
+++ b/src/cpp/include/lemon/server.h
@@ -249,7 +249,18 @@ private:
                                   nlohmann::json& out);
 
     // Helper function for auto-loading models (eliminates code duplication and race conditions)
-    void auto_load_model_if_needed(const std::string& model_name);
+    // Optionally accepts per-request override options (e.g. ctx_size) that are applied only on
+    // the very first load.  If the model is already loaded these are silently ignored so as not
+    // to overturn options that were set by a prior explicit /v1/load call.
+    //
+    // Callers MUST extract only load-level options via extract_auto_load_options() before
+    // passing them here — never pass the raw inference request body.  This prevents
+    // request-scoped fields (temperature, max_tokens, messages, steps, etc.) from leaking
+    // into recipe options that would affect subsequent requests.  Currently only ctx_size
+    // is allowlisted; pinned, eviction settings, *_args, and image params are excluded.
+    // (They can be forwarded via /v1/load when needed.)
+    void auto_load_model_if_needed(const std::string& model_name,
+                                   const json& request_options = json::object());
 
     // Helper: persist the registry's installed-providers list into config.json
     // by overlaying onto the current runtime-config snapshot. Called after
@@ -330,6 +341,13 @@ private:
 
     // Set to true after check_for_model_updates() completes at startup.
     std::atomic<bool> update_check_done_{false};
+
+    // Extract only load-level options from an inference request body.  Returns a new json
+    // containing a small, explicit allowlist of fields that are safe to forward to the
+    // RecipeOptions constructor during auto-load.  All other request-scoped fields
+    // (temperature, max_tokens, messages, steps, cfg_scale, etc.) are deliberately excluded
+    // so they cannot accidentally influence persistent model-load settings.
+    static json extract_auto_load_options(const json& request);
 };
 
 } // namespace lemon
--- a/src/cpp/server/anthropic_api.cpp
+++ b/src/cpp/server/anthropic_api.cpp
@@ -889,7 +889,7 @@ void OllamaApi::handle_anthropic_messages(const httplib::Request& req, httplib::
         auto openai_req = convert_anthropic_to_openai_chat(request_json, warnings);
 
         try {
-            auto_load_model(model);
+            auto_load_model(model, extract_auto_load_options(request_json));
         } catch (const std::exception&) {
             res.status = 404;
             json error = {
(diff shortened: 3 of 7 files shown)


fl0rianr (changes requested):
Thanks for this contribution. The ctx_size fix itself makes sense, but forwarding the entire inference request into RecipeOptions is broader than necessary and can unintentionally promote request-scoped fields into persistent model-load settings.

For example, backend options, `pinned`, eviction settings, `*_args`, and image parameters such as `steps` or `cfg_scale` may affect the initially loaded server and therefore subsequent requests.

Could we restrict the forwarded auto-load options to an explicit allowlist, starting with only ctx_size? A regression test should also verify that request-scoped generation parameters do not become persistent recipe defaults. Thanks!

indistinctTalk:
@fl0rianr thanks, I've made some additional changes; hope this is all still acceptable/in-scope.

fl0rianr (approved):
Thanks for this good adaption, LGTM now.

Two notes:
- I will edit you PR from  `Addresses ... issue #2663` wording towards `Fixes #2663` 
- test/server_endpoints.py includes some unnecessary formatting diffs, please avoid this in the future to make reviewing easier. Thanks.

Merging now as-is.

indistinctTalk:
Nice and yep, my bad with the formatting diff slipping in.

PR #2666, opened 2026-07-11
fix: bump openmoss to v0.1.3 (fixes progressive audio degradation past 10 s)

pwilkin:
OpenMOSS' codec uses a sliding window attention of 10s instead of full attention. Since the port used full attention, quality degraded after 10s.


fl0rianr (approved):
LGTM

PR #2667, opened 2026-07-11
feat(trellis): per-request UV atlas method, bump trellis to v0.4.3

pwilkin:
During the port, I added a quick UV method that makes generation faster - unfortunately, it severely degrades quality on some more complex models (see discussion in https://www.reddit.com/r/LocalLLaMA/comments/1ur1mim/comment/ows3v0z/). Re-added the longer but more faithul Atlas UV and made it the default.

Now pins trellis v0.4.3, which lands the reference-parity postprocess (narrow-band DC remesh, reference charting/packing, Telea inpaint, WebP GLB textures) — output quality verified at parity with the reference CUDA postprocess, and the xatlas path no longer times out on complex models.

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -772,6 +772,7 @@ This endpoint is not part of the OpenAI API, so it is a Lemonade-specific extens
 | `image` | Yes | Base64-encoded input image (optionally a `data:` URL). |
 | `resolution` | No | Cascade resolution: `512`, `1024`, or `1536`. Default: `512`. |
 | `bg_removal` | No | Background removal mode: `threshold` or `birefnet`. Use `birefnet` for photos with real backgrounds. |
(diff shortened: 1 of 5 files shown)


fl0rianr (commented):
Thanks! Looks good overall. The validation and backend pass-through are consistent with the existing 3D request parameters.

Non-blocking: could we add a negative test for an invalid `uv` value to `test/server_3d.py`? The endpoint already tests invalid values for `resolution`, `bg_removal`, `seed`, and `response_format`, and this test would run cheaply without loading the model.

pwilkin:
Let's hold off on this, I found some deeper problems in the backend after a report that still need fixing.

superm1 (approved):

superm1:
> Let's hold off on this, I found some deeper problems in the backend after a report that still need fixing.

I'll mark as a draft for now. Unmark draft when you're ready.

pwilkin:
Gonna move it tomorrow, but it ended up as far as 0.4.0. Turns out flash attention precision on DiT models can really make a difference - and that the entire CUDA branch in GGML had no support for the precision flag even though the flag itself was officially supported. But I had to rule out virtually everything else first.

pwilkin:
Ended up all the way at 0.4.3 and boy was it a ride... Assets should be much higher quality now.

PR #2668, opened 2026-07-11
ci: publish native AMD64 and ARM64 container manifests

fl0rianr:
## Summary

Publishes AMD64 and ARM64 container images through independent native builds and combines them into a multi-architecture manifest only after both images pass architecture, CLI, and server health checks.

## Safety

* Architecture jobs push only digest-addressable images.
* Public tags are created only after both architecture builds and smoke tests pass.
* `edge` and `latest` are updated only from a verified multi-architecture manifest.
* A failed AMD64 or ARM64 build leaves the existing public tags unchanged.
* The existing trigger and tag behavior remain unchanged.

## Tag behavior

* `main`: publishes `sha-*`, then updates `edge`
* `v*`: publishes the version tag, then updates `latest`
* manual runs: publish the selected version tag and optionally update `latest`

## Testing

* [x] AMD64 container build and smoke tests validated in #2671
* [x] ARM64 container build and smoke tests + real hardware test validated in #2671
* [ ] Digest-based architecture builds complete successfully
* [ ] Test manifest contains `linux/amd64` and `linux/arm64`
* [ ] Production `edge` update verified after merge
* [ ] Release tag and `latest` behavior verified with the next release

### DIFF
--- a/.github/workflows/build-and-push-container.yml
+++ b/.github/workflows/build-and-push-container.yml
@@ -25,8 +25,13 @@ env:
   IMAGE_NAME: ${{ github.repository }}-server
 
 jobs:
-  build-and-push:
+  prepare:
     runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.tag.outputs.tag }}
+      sha_tag: ${{ steps.tag.outputs.sha_tag }}
+      push_latest: ${{ steps.tag.outputs.push_latest }}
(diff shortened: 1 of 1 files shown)


fl0rianr:
Converted to draft since the arm code might need further hardening and merge is a no go before further testing anyway.

Update: Issue was the lemonade dockerfile. Issue here is how to verify the build process works correctly for release without actually releasing...

PR #2669, opened 2026-07-11
Bundle the Lemonade Nexus sidecar

fl0rianr:
## Summary

Bundles the pinned `lemonade-nexus-sidecar` as an opt-in Lemonade build component.

- Fetches Nexus at merged commit `427dc33841f4a2a21fbdc5b6d57325b682a66673`, or accepts a local checkout override.
- Builds Nexus in an isolated CMake sub-build to avoid dependency target and cache collisions.
- Installs the sidecar with Lemonade and wires it into Linux, Windows, and macOS package targets.
- Adds opt-in Ninja and Visual Studio presets; existing builds remain unchanged by default.

## Scope

This PR covers build and packaging only. It does not start Nexus, expose runtime APIs, manage configuration, or add UI.

## Testing

- Validated preset parsing with CMake 3.31.6.
- Validated Single-Config and Multi-Config configure, build dependency, and component install behavior with a local mock Nexus project.
- Normal Lemonade CI was green before the final Nexus pin and CMake hardening updates; the refreshed head is awaiting CI.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -8,7 +8,7 @@
       "name": "lemonade-base",
       "hidden": true,
       "cacheVariables": {
-        "CMAKE_PROJECT_TOP_LEVEL_INCLUDES": "${sourceDir}/cmake/LemonadeNexus.cmake"
+        "CMAKE_PROJECT_lemon_cpp_INCLUDE": "${sourceDir}/cmake/LemonadeNexus.cmake"
       }
     },
     {
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -5,7 +5,15 @@
   },
   "configurePresets": [
     {
+      "name": "lemonade-base",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_PROJECT_TOP_LEVEL_INCLUDES": "${sourceDir}/cmake/LemonadeNexus.cmake"
+      }
+    },
+    {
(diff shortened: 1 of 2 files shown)


Nathanial96:
#https://github.com/lemonade-sdk/lemonade-nexus/pull/8 has been merged to main as [427dc33](https://github.com/lemonade-sdk/lemonade-nexus/commit/427dc33841f4a2a21fbdc5b6d57325b682a66673)

superm1 (commented):
Does it need to be built at the same time as lemonade?  I don't think this is going to work for the Launchpad PPA builds.  We can't fetch content using `FetchContent`, so we would have to "vendor" the nexus stuff.

At that point; why even bundle it?  Just ship it as a second Debian package on the PPA.

Geramy:
> Does it need to be built at the same time as lemonade? I don't think this is going to work for the Launchpad PPA builds. We can't fetch content using `FetchContent`, so we would have to "vendor" the nexus stuff.
> 
> At that point; why even bundle it? Just ship it as a second Debian package on the PPA.

We need to have a conversation to make sure what features we implement is compatible with this. Let’s schedule a call In discord @superm1 please. This is step one there is more and it’s security related.

Geramy:
We had the conversation, this will be implemented as a full in-tree sdk so it can bundle and be packaged into .deb and ppa

PR #2677, opened 2026-07-12
feat(flm): restore validated backend arguments

fl0rianr:
## Summary

Restore **flm_args** support for FastFlowLM through a strict typed allow-list.

Supported tuning arguments:

* --pmode
* --prefill-chunk-len
* --img-pre-resize
* --socket
* --q-len
* --preemption

Lemonade-managed arguments such as model mode, host, port, context length, and input options remain blocked. The option is available through `load`, `run`, and `bench`.

## Testing

* Built lemonade, lemond, and the new FLM resolver test target
* Ran the complete resolver test suite, including valid values, normalization, duplicates, malformed syntax, blocked arguments, and boundary checks
* Verified successful FLM model loading and inference with gemma4-it-e2b-FLM
* Verified all supported FLM arguments together
* Verified invalid and Lemonade-managed arguments are rejected before FLM startup
* Ran a real bench workflow with FLM arguments across multiple chat scenarios
* Regenerated backend documentation and confirmed generated files are up to date

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1823,6 +1823,27 @@ message(STATUS "  Archive: ${EMBEDDABLE_ARCHIVE_NAME}")
 # ============================================================
 # C++ Unit Tests
 # ============================================================
+
+# FastFlowLM argument validation. Keep this test independent from the server
+set(_FLM_ARG_RESOLVER_TEST_SRC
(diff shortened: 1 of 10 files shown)


superm1 (approved):
Begrudgingly I agree with adding them back in this form.

PR #2678, opened 2026-07-12
fix: quote /MANIFESTINPUT path and detect RX 7600 as gfx110X

kyuubyN:
Two independent Windows/ROCm fixes:

- The /MANIFESTINPUT linker flag's path was unquoted, so a build tree under a directory containing a space fails to link with LNK1181 (the path is silently truncated at the space). Quote it.

- identify_rocm_arch_from_name() matched 7700/7800/7900/v710 but not the RX 7600, so a machine with an RX 7600 reported no gfx110X family and every ROCm backend (llamacpp included) showed as unsupported on it. Add 7600 to the existing gfx110X substring check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -768,9 +768,10 @@ if(WIN32)
             /NXCOMPAT         # Data Execution Prevention (DEP)
             /GUARD:CF         # Control Flow Guard
         )
-        # Embed manifest file
+        # Embed manifest file. /MANIFESTINPUT's path must be quoted; unquoted,
+        # a path containing a space breaks the linker with LNK1181.
         set_target_properties(${EXECUTABLE_NAME} PROPERTIES
-            LINK_FLAGS "/MANIFEST:EMBED /MANIFESTINPUT:${CMAKE_CURRENT_BINARY_DIR}/server/lemonade.manifest"
+            LINK_FLAGS "/MANIFEST:EMBED /MANIFESTINPUT:\"${CMAKE_CURRENT_BINARY_DIR}/server/lemonade.manifest\""
         )
     endif()
 endif()
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -1867,7 +1867,13 @@ std::string identify_rocm_arch_from_name(const std::string& device_name) {
         return "gfx120X";
     }
 
-    if (device_lower.find("7700") != std::string::npos ||
+    // "7600" alone is not product-specific enough: it also matches names like
+    // "AMD Radeon HD 7600M Series" (a GCN1 card from 2012, not RDNA3). Match
+    // the actual product family instead.
+    if (device_lower.find("rx 7600") != std::string::npos ||
+        device_lower.find("rx7600") != std::string::npos ||
+        device_lower.find("pro w7600") != std::string::npos ||
+        device_lower.find("7700") != std::string::npos ||
         device_lower.find("7800") != std::string::npos ||
         device_lower.find("7900") != std::string::npos ||
         device_lower.find("v710") != std::string::npos) {


fl0rianr (changes requested):
The manifest quoting change looks good.

I have one concern about extending the existing GPU-name heuristic with another bare numeric substring. The current 7700/7800/7900 checks are already fairly broad, and adding "7600" creates a concrete false-positive case: Windows may report older GPUs as "AMD Radeon HD 7600M Series", which would then be classified as gfx110X.

So this is not necessarily a problem unique to this PR, but this change would expand the existing ambiguity. Could we make the matching product-family-specific instead, for example by recognizing names such as RX 7600, RX7600, and PRO W7600?

@superm1 What is your take on this? Do you agree that this logic should be improved anyway?

superm1:
> @superm1 What is your take on this? Do you agree that this logic should be improved anyway?

I want us to get away from matching strings of names if we can even on Windows, but I'm not sure how to discover the ISA properly on Windows.

fl0rianr:
Agreed — avoiding marketing-name matching altogether would be much better.

Could we use a layered approach on Windows?

1. If the HIP runtime is available, query the device properties and use the reported gcnArchName/ISA directly.
2. Otherwise, read the AMD PCI device ID from Win32_VideoController.PNPDeviceID and map that to the corresponding gfx architecture.
3. Keep the current product-name matching only as a final fallback for systems where neither source is available.

The code already uses WMI for AMD GPU discovery, so obtaining the PCI device ID there should be relatively natural. A small centralized PCI-ID-to-gfx mapping also seems safer and easier to test than continuing to expand numeric name substrings.

For this PR, adding 7600 may still be a reasonable narrow fix, but it would be good to track replacing the complete name-based heuristic separately.

superm1:
Can we access the PCIe BAR without a kernel driver in Windows?  If so, we can get information from the discovery table to map the GC HW IP version to the SW ISA like the Linux kernel does.

kyuubyN:
Not without a kernel driver, at least not practically. On Linux, amdgpu can read the IP Discovery table directly because it's the kernel driver itself, already running with the privilege to touch device MMIO. Windows has no usermode equivalent, WMI/SetupAPI/CfgMgr32 only expose PCI config space metadata (vendor ID, device ID, subsystem, revision), not BAR contents. Getting the same "read straight from hardware" result here would mean shipping and signing our own kernel driver, which seems like a lot of surface area and maintenance for what this needs.

Given that, the layered plan above is close to the practical ceiling on Windows:

1. HIP runtime, if present: `hipGetDeviceProperties` returns `gcnArchName` straight from the driver, same source of truth Linux gets, just through ROCm instead of raw hardware.
2. PCI device ID via `Win32_VideoController.PNPDeviceID`: standard PCI enumeration metadata, no BAR access needed, fully usermode-supported. A small vendor:device -> gfx arch table is the same approach GPU-Z/HWiNFO use, and it's testable in isolation since the input is deterministic.
3. Product name matching only as the last-resort fallback.

superm1:
If we can get hipGetDeviceProperties() to work, then why do we need PCI IDs at all?  Just call that on Windows on every device that is AMD Vendor ID.

kyuubyN:
Because hipGetDeviceProperties() requires the HIP runtime to already be present, and this detection needs to run *before* that's guaranteed, it's literally what decides whether Lemonade offers to install a ROCm backend/TheRock for a given GPU in the first place. On a fresh install with just the AMD graphics driver and no HIP SDK/TheRock yet, there's no runtime to query. That's exactly the gap that let the RX 7600 show up as "unsupported" before this PR.

There's a second reason from the dGPU+iGPU issue we're also tracking: even when HIP *is* installed, hipInit()/device enumeration can fail or enumerate the wrong device on hybrid systems or with driver quirks. WMI-based PCI ID lookup is a lower common denominator that keeps working when the compute stack itself is broken or absent, since it's just OS device enumeration, not dependent on the runtime initializing correctly.

So the ordering still makes sense: try HIP first when it's there (best source of truth), fall back to PCI ID (works independent of the compute stack), and keep name matching as the last resort.

superm1:
Hmm.

How about if we change the design to use ROCm CLI? 

https://github.com/ROCm/rocm-cli

Could that help?

superm1:
IE `rocm examine`

fl0rianr:
rocm examine could be a good source of truth, provided it works in the state we need to support: a fresh Lemonade installation with only the AMD display driver present and no ROCm/HIP runtime installed yet.

That seems to be the key question. If ROCm CLI depends on the same runtime or management libraries that Lemonade is trying to decide whether to install, it would improve the query interface but not solve the bootstrap problem. We would also need to either bundle the CLI and its dependencies or require a separate prerequisite.

If rocm examine can obtain the ISA directly through the installed Windows driver without a pre-existing ROCm installation, then it sounds preferable to maintaining PCI-ID and product-name tables. Otherwise, I think the practical order remains:

1. rocm examine or HIP when already available;
2. PCI device ID for fresh-install detection;
3. product-name matching only as a final compatibility fallback.

It would be useful to verify exactly which Windows components and libraries `rocm examine` requires before choosing the design.

kyuubyN:
I pulled the full examine.rs source to check exactly what rocm examine does on Windows before deciding.

for a fresh install with only the display driver, no HIP SDK, it doesn't solve the bootstrap problem, and where it can determine the ISA correctly, it needs the same thing we'd need anyway.

The Windows probe runs in two stages:

probe_gpus_windows() shells out to Get-CimInstance Win32_VideoController, the exact same WMI class we already query natively via COM. No ROCm needed for this step, that part is genuinely bootstrap-safe. But the ISA it assigns here comes from classify_amd_marketing_name(), marketing-name substring matching, and today it only recognizes specific APUs (Strix Halo, Strix Point/Krackan, Phoenix/Hawk Point). Every discrete card, including the RX 7600/7700/7800/7900 this repo cares about, gets an empty gfx_target at this stage.
probe_hip_sdk_windows() is what actually fixes that: it looks for hipInfo.exe under HIP_PATH or C:\Program Files\AMD\ROCm, and if found, parses its real gcnArchName: output to fill in the ISA precisely. But it only runs if a HIP SDK is already installed, on a clean machine it's a no-op. So the accurate path in rocm examine requires exactly the pre-existing ROCm/HIP install we're trying to decide whether to bootstrap in the first place, the same circularity as hipGetDeviceProperties(), just reached one layer further in.
the pci_id rocm-cli captures on Windows is just the raw PNPDeviceID from that same WMI query (PCI\VEN_1002&DEV_XXXX&...), the identical string we already get back from our own WMI call. So a PCI-ID-keyed table doesn't need rocm-cli or even DXGI, we can extract DEV_XXXX from data we're already reading today.

Given that, I'd propose: keep the layered order you outlined, but for Windows dGPU detection on a fresh install specifically, build tier 2 (PCI device ID) natively from the PNPDeviceID we already parse, rather than depending on rocm-cli. Once ROCm/HIP is actually present (ours or rocm-cli's), either one can confirm/override via the real API. Happy to revisit rocm examine as a primary source later if it grows its own PCI-ID table instead of the marketing-name list, that would genuinely remove a maintenance burden from our side.

fl0rianr (approved):
Thanks, the updated product-family-specific matching addresses my concern. It detects the RX 7600 family without misclassifying older names such as the Radeon HD 7600M Series.

The manifest quoting change also looks correct. Approving.

Issue #2573, opened 2026-07-06
GUI3: Chat page has no way to see loaded models or manage unloading


kpoineal:
**Reported by:** Jeremy (project owner)

On the Chat page, there's no visibility into which models are currently loaded and no way to unload them. Users need to see and manage the loaded model state.

**Expected:** Visible indicator of loaded models with ability to unload.
**Actual:** No loaded model status or management available on the Chat page.

PR #2680, opened 2026-07-12
feat(gui3): add loaded model visibility and unload controls to chat page

kpoineal:
Addresses #2573 — the chat page had no way to see which models are loaded or unload them.

### Changes

- **Loaded count badge** on the model picker button shows how many models are currently loaded, giving at-a-glance awareness without opening the dropdown.
- **Per-row unload button** on each loaded model row in the picker dropdown. Clicking it calls api.unloadModel() and refreshes the model list. The select and unload actions are visually and structurally separated to prevent accidental unloads.
- **Always-visible unload button** — no hover required, works on touch/mobile devices.
- **Accessible**: aria-label, independent keyboard focus, aria-live polite region announces completion.
- **Loading state**: row dims and shows a spinner during the unload operation.

### Testing

- Playwright: 200/200 passing
- Manual: verified on desktop and mobile


fl0rianr (approved):
LGTM, thanks!

PR #2683, opened 2026-07-12
Bench imagegen

ckuethe:
Yet Another Benchmark Mode: images.

includes a couple of not-strictly-necessary changes that improve readability:
- changed that `300000` into `FIVE_MINUTES_MS` to explain what it's for
- automatically expand the format of the results table so that scenario names and runtimes don't collide when running long scenarios on slower machines

See, isn't this so much nicer to read?
```
$ ./lemonade --host ryzen bench Z-Image-TurboQ SD-Turbo Flux-2-Klein-4B SD-1.5 --scenarios imagegen --scenario-file resources/bench_scenarios.json 
...
Benchmark: Z-Image-TurboQ
====================================================================================================

Backend: sd-cpp/rocm
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      10126.2 10052.1 10196.3 0.0     0.0     0.0     0.6     
image-photoreal-outdoor   10309.9 10211.7 10380.0 0.0     0.0     0.0     0.6     

Backend: sd-cpp/vulkan
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      10158.7 10098.0 10279.4 0.0     0.0     0.0     0.6     
image-photoreal-outdoor   10283.4 10253.5 10336.9 0.0     0.0     0.0     0.6     

(*Nf = N failed runs excluded from stats)


Benchmark: SD-Turbo
====================================================================================================

Backend: sd-cpp/rocm
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      2447.2  2422.2  2474.1  0.0     0.0     0.0     0.6     
image-photoreal-outdoor   2409.8  2401.7  2420.9  0.0     0.0     0.0     0.6     

Backend: sd-cpp/vulkan
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      2491.7  2468.0  2525.4  0.0     0.0     0.0     0.6     
image-photoreal-outdoor   2413.8  2398.1  2429.4  0.0     0.0     0.0     0.6     

(*Nf = N failed runs excluded from stats)


Benchmark: Flux-2-Klein-4B
====================================================================================================

Backend: sd-cpp/rocm
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      16805.5 16780.2 16835.1 0.0     0.0     0.0     1.0     
image-photoreal-outdoor   17798.8 17752.2 17842.2 0.0     0.0     0.0     1.0     

Backend: sd-cpp/vulkan
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      16843.8 16773.9 16921.4 0.0     0.0     0.0     1.0     
image-photoreal-outdoor   17846.0 17834.4 17865.2 0.0     0.0     0.0     1.0     

(*Nf = N failed runs excluded from stats)


Benchmark: SD-1.5
====================================================================================================

Backend: sd-cpp/rocm
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      6905.0  6869.8  6953.0  0.0     0.0     0.0     0.5     
image-photoreal-outdoor   7131.1  7078.1  7175.2  0.0     0.0     0.0     0.5     

Backend: sd-cpp/vulkan
Scenario                  TTFT    min     max     TPS     min     max     VRAM (GB)
image-sd-turbo-quick      6943.6  6897.4  7020.0  0.0     0.0     0.0     0.5     
image-photoreal-outdoor   7098.7  7076.5  7130.6  0.0     0.0     0.0     0.5     

(*Nf = N failed runs excluded from stats)


```

Output capture also works, so the included scenarios produces a response log with image content like this:

20260712132654__image-sd-turbo-quick__Z-Image-TurboQ__vulkan
20260712132654__image-photoreal-outdoor__Flux-2-Klein-4B__vulkan


fl0rianr (changes requested):
Thanks for adding image-generation benchmarking — the overall direction makes sense, but I found a few issues that should be fixed before merge:

1. **Default LLM benchmarks now also run imagegen scenarios.**
   Without `--scenarios`, only `long-context` and `embed` are excluded. Since the new image scenarios are in the shared default scenario file, a normal command such as `lemonade bench Qwen3-8B` will also send image-generation requests to that LLM. Image scenarios should either be excluded by default or filtered based on model modality/recipe.

2. **Most image scenario parameters are currently ignored.**
   The scenarios define `size`, `steps`, `cfg_scale`, `seed`, and `response_format`, but only `prompt` is parsed and forwarded. This means the two added scenarios may actually run with identical backend defaults, and the results are not reproducible. These fields need to be preserved and included in the request payload.

3. **duration_ms remains zero for image benchmarks.**
   The elapsed time is assigned to `ttft_ms`, but `total_time_ms` is never populated. As a result, the JSON `duration_ms` statistics are always zero for successful image runs.

4. **The new dynamic table width only adjusts the scenario-name column.**
   Long runtime values can still collide because the TTFT/TPS widths remain fixed at 8 characters. Very long scenario names are also capped in the calculated width but not actually truncated.

5. **print_scenario_row() should not be declared static in the public header.**
   It is an implementation detail and should remain in bench.cpp.

I would also recommend adding tests for:

* preservation of all image request parameters,
* default exclusion or modality filtering of image scenarios,
* non-zero image duration output,
* response capture with b64_json.

The endpoint integration itself is nicely separated, but the first two issues are merge-blocking because the benchmark currently does not execute the scenarios as defined.

ckuethe:
I didn't exclude image generation intentionally. That seems like a more common thing that people might want to do - at least more than embeddings.

src/cpp/resources/bench_scenarios.json
```diff
@@ -166,6 +166,30 @@
       ],
       "warmup_runs": 0,
       "measurement_runs": 3
+    },
+    {
+      "name": "image-sd-turbo-quick",
```
bitgamma: steps and cfg_scale should not be overriden in the bench scenario, because every model has very different settings and the results wouldn't make sense across different models.

src/cpp/cli/bench.cpp
```diff
@@ -689,6 +696,73 @@ BenchRunResult run_single_bench_embed(lemonade::LemonadeClient& client,
     return result;
 }
 
+BenchRunResult run_single_bench_imagegen(lemonade::LemonadeClient& client,
```
bitgamma, line 776: how are the scenarios filtered? Do imagegen tests try to run even when selecting a text model? I think we should associate an array of recipe_types to bench test category so they are auto-filtered.

src/cpp/cli/bench.cpp
```diff
@@ -689,6 +696,73 @@ BenchRunResult run_single_bench_embed(lemonade::LemonadeClient& client,
...
+        return result;  // success stays false
+    }
+
+    // Query memory after
+    if (memory_tracking)
+        extract_mem_use_into_result(client, result);
+
+
+    result.success = true;
+    return result;
```
bitgamma, line 861: beside performance, when comparing image models you definitely want to see the final image.

src/cpp/resources/bench_scenarios.json
```diff
@@ -166,6 +166,30 @@
...
+      "category": "imagegen",
+      "prompt": "A cartoon penguin wearing sunglasses, surfing on a lemon.",
+      "size": "256x256",
```
bitgamma: 256x256 is uninteresting for most people (just as a note, doesn't necessarely need to go)

src/cpp/resources/bench_scenarios.json
```diff
@@ -166,6 +166,30 @@
...
+      "steps": 4,
+      "cfg_scale": 1.0,
+      "seed": 42,
+      "response_format": "b64_json",
+      "warmup_runs": 0,
+      "measurement_runs": 3
+    },
+    {
+      "name": "image-photoreal-outdoor",
```
bitgamma: we need more scenarios, ideally at ~1mp output, that tests the model various capabilities (text rendering, multiple subjects, art styles, etc). As I mentioned before, usually generation speed is much less of a concern here since basically all tuning knobs trade quality for speed.

But adding more scenarios can be done in a separate PR, just let me know if you are not doing and I'll take care of it in a followup

bitgamma (changes requested):

src/cpp/cli/bench.cpp
bitgamma: what I mean is that at least the image part of output capture should be on by default for image bench

src/cpp/resources/bench_scenarios.json
ckuethe: It's quite interesting for making icons, small stickers, and measuring performance scaling.

src/cpp/resources/bench_scenarios.json
ckuethe: Generation speed is important to me. Some models are too slow to be useful on some machines. Once I get a set of models and image sizes that are useful, then I can start tuning the prompts and whatever else.

src/cpp/cli/bench.cpp
ckuethe: That's a subjective analysis which is out of scope for benchmarks - just like looking at the the output of textgen modes. This is about measuring how fast pictures come back, not how good they are.

https://lemonade-server.ai/docs/dev/working-groups/auto-tune/#scope

src/cpp/resources/bench_scenarios.json
ckuethe: The same can be said about thinking budget, and all the other sampler parameters in textgen models.

Those should absolutely be configurable in the scenarios, though I can leave them unset by default.

ckuethe:
128, 512, and 1024 on my laptop. That's why 256 is useful and 1024 isn't in my default test set.

```
Benchmark: Z-Image-TurboQ8
====================================================================================================

Backend: sd-cpp/rocm
image-small         4652.5  4652.5  4652.5  0.0     0.0     0.0     10.9    
image-medium        29991.0 29991.0 29991.0 0.0     0.0     0.0     10.9    
image-large         126449.9 126449.9 126449.9 0.0     0.0     0.0     10.8    

Backend: sd-cpp/vulkan
image-small         4631.0  4631.0  4631.0  0.0     0.0     0.0     10.8    
image-medium        30109.3 30109.3 30109.3 0.0     0.0     0.0     10.9    
image-large         126526.6 126526.6 126526.6 0.0     0.0     0.0     10.8    



Benchmark: SD-Turbo
====================================================================================================

Backend: sd-cpp/rocm
image-small         1040.3  1040.3  1040.3  0.0     0.0     0.0     4.7     
image-medium        7546.2  7546.2  7546.2  0.0     0.0     0.0     4.8     
image-large         37195.7 37195.7 37195.7 0.0     0.0     0.0     4.7     

Backend: sd-cpp/vulkan
image-small         1043.3  1043.3  1043.3  0.0     0.0     0.0     4.7     
image-medium        7511.4  7511.4  7511.4  0.0     0.0     0.0     4.7     
image-large         37554.0 37554.0 37554.0 0.0     0.0     0.0     4.8     



Benchmark: Flux-2-Klein-4B
====================================================================================================

Backend: sd-cpp/rocm
image-small         10960.5 10960.5 10960.5 0.0     0.0     0.0     15.3    
image-medium        30751.8 30751.8 30751.8 0.0     0.0     0.0     15.5    
image-large         144021.2 144021.2 144021.2 0.0     0.0     0.0     16.6    

Backend: sd-cpp/vulkan
image-small         11044.3 11044.3 11044.3 0.0     0.0     0.0     15.3    
image-medium        30840.8 30840.8 30840.8 0.0     0.0     0.0     15.5    
image-large         144934.8 144934.8 144934.8 0.0     0.0     0.0     16.5    



Benchmark: SD-1.5
====================================================================================================

Backend: sd-cpp/rocm
image-small         1616.1  1616.1  1616.1  0.0     0.0     0.0     3.8     
image-medium        8026.5  8026.5  8026.5  0.0     0.0     0.0     3.9     
image-large         38902.3 38902.3 38902.3 0.0     0.0     0.0     3.9     

Backend: sd-cpp/vulkan
image-small         1617.5  1617.5  1617.5  0.0     0.0     0.0     3.8     
image-medium        7994.9  7994.9  7994.9  0.0     0.0     0.0     3.8     
image-large         38816.6 38816.6 38816.6 0.0     0.0     0.0     3.8     
```

src/cpp/resources/bench_scenarios.json
bitgamma: it is important but is often evaluated in tandem with image quality. Anyway I can add more scenarios later.

src/cpp/cli/bench.cpp
bitgamma: you are pointing me to my own document but you are mixing auto tune scope with benchmark scope. 

Autotune uses benchmark but it doesn't mean benchmark scope is limited to that. Infact, I want to introduce qualitative benchmarks for LLM too at some point.

Anyway, since there is an option to capture the images, this is OK for now.

src/cpp/cli/bench.cpp
```diff
@@ -1633,7 +1868,7 @@ int handle_bench_command(lemonade::LemonadeClient& client, const BenchConfig& co
 CLI::App* register_bench_command(CLI::App& parent,
                                  std::string& output_file,
                                  BenchCliOptions& opts) {
-    CLI::App* cmd = parent.add_subcommand("bench", "Benchmark a model's chat completion performance")->group("Model management");
+    CLI::App* cmd = parent.add_subcommand("bench", "Benchmark model speed for different model types")->group("Model management");
```
bitgamma: "speed" needs to be changed back to "performance"

src/cpp/cli/bench.cpp
ckuethe: Yep, I'm interested in quality analysis, as a different stage. That's why I added output capture in the first place. I've got a little python script to extract the various outputs into an image gallery, but I don't want to turn `bench` into an overcomplicated mess.

bitgamma:
@ckuethe compilation on windows is failing. When that is fixed (and the other wording issue I wrote above) this can be merged for me.

ckuethe:
I still need to fix the timing thing Florian mentioned too

bitgamma (approved):

fl0rianr (commented):
Thanks for the updates — this is looking much better now, and I appreciate the work that went into addressing the earlier feedback.

I found two remaining edge cases that I think are worth fixing before merge:

1. The new scenario filtering currently relies on the backend recipe descriptor. That may misclassify embedding models using the `llamacpp` recipe, because the descriptor reports `"Text generation"` even when the concrete model has an `embeddings` label. It would likely be safer to derive the capability from the selected model’s labels/model type instead of the recipe alone.

2. The image benchmark currently always sends fallback values for `steps` and `cfg_scale`. This overrides the model-specific image defaults even when the scenario does not explicitly define those fields. For example, SD-1.5 would be benchmarked with Turbo-style defaults. It may be better to only copy optional image parameters into the request when they are actually present in the scenario. The same applies to `response_format`, which is currently defined in the scenarios but not forwarded.

Everything else looks solid to me, and the overall image benchmark integration is nicely structured. Once these two cases are addressed, I’d be happy to approve.

ckuethe:
Can you explain the scenario filtering a little more? We don't have very good label management right now and it's quite possible to get models that don't have a label to effectively filter on.

By way of example this doesn't say it can do text generation or embedding - but it doesn't say that it can't either. 

```json
    {
      "checkpoint": "ggml-org/gemma-3-4b-it-GGUF:Q4_K_M",
      "checkpoints": {
        "main": "ggml-org/gemma-3-4b-it-GGUF:Q4_K_M",
        "mmproj": "ggml-org/gemma-3-4b-it-GGUF:mmproj-model-f16.gguf"
      },
      "components": [],
      "created": 1234567890,
      "downloaded": true,
      "id": "Gemma-3-4b-it-GGUF",
      "labels": [
        "vision"
      ],
      "max_context_window": 131072,
      "object": "model",
      "owned_by": "lemonade",
      "recipe": "llamacpp",
      "recipe_options": {

      },
      "registry_source": "huggingface",
      "size": 3.11,
      "source": "huggingface",
      "suggested": true,
      "update_available": false
    },
```

So I'd prefer to not do that filtering until every existing model is properly labeled, and that users can manage labels on their own.

ckuethe:
And because @bitgamma mentioned it elsewhere, I added text drawing tests. Some good results, some comically bad results.

fl0rianr (commented):
Thanks, that concern makes sense — I agree that filtering must not require every model to have complete labels, and missing labels should not be interpreted as a negative capability declaration.

The concrete issue I was trying to address is narrower: the current implementation derives the capability entirely from the recipe descriptor. For a model labelled `embeddings` but using the `llamacpp` recipe, the descriptor still says “Text generation”. Consequently, the embedding scenario is skipped while ordinary text-generation scenarios are allowed, which is the opposite of the intended behaviour.

I think a conservative hybrid would avoid depending on complete label coverage:

* use positive model labels such as `embeddings` when present;
* use the recipe descriptor for unambiguous specialized recipes such as `sd-cpp`;
* treat missing labels as unknown rather than unsupported;
* retain text generation as the fallback for an otherwise unknown `llamacpp` model;
* optionally treat an explicit `--scenarios embed` selection as a user override for an ambiguous model.

That would fix known embedding models without requiring every existing model to be labelled first.

I also noticed three small cleanup items in the latest revision:

* `image-text1` is defined twice;
* `response_format` is present in the scenarios but is not read from them — either forwarding it or removing it from the scenario definitions would make the behaviour clearer;
* `scenario_requires_capability()` appears to be unused.

The optional `steps`/`cfg_scale` handling and image duration accounting look fixed now. With the filtering issue and the duplicate scenario name addressed, I think this would be in good shape.

fl0rianr (approved):
I’m approving this to avoid blocking the PR any further, but this should **not be interpreted** as a full technical approval.

The remaining concern is still the capability filtering: deriving support only from the recipe descriptor can incorrectly skip embedding models that use the `llamacpp` recipe. The image `warmup_runs` parsing and the unnecessary `ctx_size` handling for `sd-cpp` are also worth cleaning up.

I’m okay with these being addressed in a follow-up, but I want the remaining risks to be documented clearly before merge.

ckuethe:
I'll address those before merging

ckuethe:
I think I've addressed all the things?
* mix of models dispatches only the appropriate combinations when run with `--scenarios all`
* warmups are thoroughly excluded from image gen
* image format is internally pinned to base64 even though that's the only "option", and is therefore removed from the scenario definition
* removed the unused function

the --timeout flag is getting a good workout today because test scenarios take between 6 and 27 minutes per image with Qwen-Image-2507 on my gfx1152 😭

Issue #2686, opened 2026-07-13
ROCm backend reports "Unsupported GPU: gfx950" on AMD Instinct MI350X (llama.cpp)


Bekhouche:
### Platform

Linux/Ubuntu

### Lemonade Version

10.10.0

### GPU / APU Model

AMD Instinct MI350X (gfx950), 287.6 GB VRAM

### Component

llama.cpp

### Bug Description

On a real MI350X (`gfx950`) host, ROCm is reported as unsupported for `llamacpp`, and the same
root cause affects `vllm`, `sd-cpp`, and `whispercpp` too — even though Lemonade correctly
detects the GPU and its VRAM, and even though `acestep`/`thinksound` report `rocm` as
`installable` on the exact same machine. So this isn't a GPU-detection failure, it's specifically
the per-backend ROCm arch allow-lists not including `gfx950` yet.

This looks like the same class of gap #2587 fixed for `gfx942` (MI300X) in the llama.cpp ROCm
backend: that PR's own description says the shipped `llamacpp-rocm` tarball's `libggml-hip.so`
already ships `gfx950` kernels alongside `gfx908`/`gfx906`/`gfx942` — it was purely an allow-list
gap for `gfx942`, no new backend build was required. If that still holds, adding `gfx950` to the
allow-list may be enough to unblock llama.cpp here too — I haven't verified the binary's `gfx950`
kernels work end-to-end the way that PR verified `gfx942`, but I have the hardware to test a
patch.

Separately (see Additional Context) commit e1740901 / #2588 already treats `gfx950` as a
"discrete-HBM class" arch internally on the vLLM side, but doesn't extend the installable matrix
to it either.

### Steps to Reproduce

On a host with an AMD Instinct MI350X (`gfx950`) and ROCm/KFD available:

1. `lemonade backends` — `rocm` is missing entirely from the `llamacpp`, `vllm`, `sd-cpp`, and
   `whispercpp` rows (only `vulkan`/`cpu` show up for those).
2. `lemonade backends install llamacpp:rocm` →
   `Error: Cannot install llamacpp:rocm on this system: Unsupported GPU: gfx950`
3. Same result for `vllm:rocm`, `sd-cpp:rocm`, `whispercpp:rocm`.
4. Compare with `lemonade backends install acestep:rocm` / `thinksound:rocm` on the same host →
   both report `installable` normally, confirming GPU family/VRAM detection itself is fine.

### Expected vs Actual Behavior

[...]

PR #2687, opened 2026-07-13
Support gfx950, using the same path as gfx942

SlawomirNowaczyk:
Routes `gfx950` (MI350) through the same code path as `gfx942`.
Verified the `llama.cpp` backend runs correctly on MI350 hardware.

This addresses (at least partially) issue https://github.com/lemonade-sdk/lemonade/issues/2686

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2226,3 +2226,13 @@ if(EXISTS "${_VLLM_CDNA_TEST_SRC}")
     target_link_libraries(test_vllm_cdna_gating PRIVATE lemonade-server-core)
     add_test(NAME VllmCdnaGatingTest COMMAND test_vllm_cdna_gating)
 endif()
+
+# llamacpp gfx950 (MI350) install gate: only the Linux + stable-channel asset is
+# published, so the shared Windows/Linux support row must not advertise Windows
+# or nightly installs for gfx950.
+set(_LLAMACPP_GFX950_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_llamacpp_gfx950_gating.cpp")
+if(EXISTS "${_LLAMACPP_GFX950_TEST_SRC}")
+    add_executable(test_llamacpp_gfx950_gating test/cpp/test_llamacpp_gfx950_gating.cpp)
+    target_link_libraries(test_llamacpp_gfx950_gating PRIVATE lemonade-server-core)
+    add_test(NAME LlamacppGfx950GatingTest COMMAND test_llamacpp_gfx950_gating)
+endif()
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ Lemonade supports multiple inference engines for LLM, speech, TTS, and image gen
     </tr>
     <tr>
       <td><code>rocm</code></td>
-      <td>Supported AMD ROCm iGPU/dGPU families, incl. AMD Instinct MI300X (gfx942) and MI350X (gfx950)*</td>
+      <td>Supported AMD ROCm iGPU/dGPU families, incl. AMD Instinct MI300X (gfx942) and MI350X (gfx950, Linux + stable only)*</td>
       <td>Windows, Linux</td>
     </tr>
     <tr>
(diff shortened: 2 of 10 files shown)


fl0rianr (commented):
Thanks for adding `gfx950` support and validating the llama.cpp path on real MI350 hardware. The Linux stable path looks good, but I see two blockers before this can merge:

1. `gfx950` is added to the shared Windows/Linux `llamacpp:rocm` support row. On Windows, Lemonade would therefore advertise the backend as installable, but the TheRock installer resolves this to a `therock-dist-windows-gfx950-dcgpu-7.13.0.tar.gz` asset, which is not published (yet). May gate `gfx950` to Linux unless a working Windows asset is available.

2. The same support row also enables `rocm-nightly`. The nightly installer resolves an architecture-specific `gfx950` asset, but the current `lemonade-sdk/llamacpp-rocm` nightly builds do not publish `gfx950`. The topology test does not catch this because nightly links are intentionally excluded from the live 404 check. Please either publish the nightly asset or keep `gfx950` limited to the stable channel.

Non-blocking: please also add an explicit assertion that `vllm:rocm` remains unsupported for `gfx950` until its corresponding asset is actually published.

So the intended Linux stable support is sound, but the OS/channel support matrix currently advertises two install paths that will fail.

SlawomirNowaczyk:
> So the intended Linux stable support is sound, but the OS/channel support matrix currently advertises two install paths that will fail.

Thanks for catching this — both blockers are real. gfx950 assets only exist for Linux + stable today, so advertising Windows-stable and nightly would send folks toward `therock-dist-windows-gfx950-*` and nightly builds that aren't published.

Here is a fix, but I want to sanity-check the approach before committing to it. The wrinkle is that gfx950 shares the `llamacpp:rocm` row with the RDNA families, which legitimately span Windows/Linux and stable/nightly — so I can't just narrow the row. Instead I added a **per-arch install gate**: an optional `arch_gates` map on the support row that can restrict a single architecture to a subset of the row's OS/channel reach. gfx950 gets gated to `os={linux}, channels={stable}`, and `backend_supports_arch` (plus the availability matrix) now enforces it OS- and channel-aware.

My hesitation: for *this* PR that's clearly heavier machinery than the problem needs — a couple of `#ifdef`/channel checks would gate gfx950 just as well. The reason I leaned into the general mechanism is that I expect this "one arch is more constrained than its row" situation to recur as more backends/architectures land with staggered platform availability (it's basically already the RDNA-vs-CDNA asymmetry). So the question is really: is the generic gate worth carrying now for future reuse, or should we prefer a minimal, gfx950-specific gate here and we generalize if/when a second case shows up?

Separately, I added the non-blocking assertion that `vllm:rocm` stays unsupported for gfx950.

fl0rianr (commented):
Thanks, the runtime gating now addresses both original blockers cleanly. `gfx950` is restricted to Linux + stable, nightly is rejected, and the additional vLLM gating assertion is present.

I think the generic `arch_gates` mechanism is reasonable here. It is more machinery than a gfx950-specific condition, but it keeps the platform/channel restriction in the backend descriptor and should be reusable for similarly staggered asset support.

One issue remains before approval: the generated/public support matrix still lists the ROCm row containing MI350X (`gfx950`) as supporting “Windows, Linux”. That now contradicts the actual install gate and still tells Windows users that MI350X is supported.

Do you think it makes sense to make the limitation visible in README and the backend reference, for example:

`MI350X (gfx950): Linux stable only`

I'm not 100% how it will fit with the boilerplate python auto job for backends.

Ideally the documentation generator would eventually represent `arch_gates`, but an explicit note is sufficient for this PR. Assuming the CMake conflict is resolved cleanly and CI passes afterward, I do not see another functional blocker.

SlawomirNowaczyk:
> One issue remains before approval: the generated/public support matrix still lists...

Pushed a docs update to make the gfx950 limitation visible in the public support matrix (addressing the "Windows, Linux" contradiction):

- The `llamacpp:rocm` device summary now reads **MI350X (gfx950, Linux + stable only)** — it's the descriptor field the boilerplate generator renders, so README.md picks it up through the normal job and stays `--check`-clean.
- Added an explicit note under the support matrix in backends-reference.md clarifying that the OS column shows the row's overall reach while gfx950 is restricted by the install gate.

Left teaching the generator to render `arch_gates` natively as a follow-up, per your suggestion. Also merged latest `main` and resolved the CMakeLists.txt conflict.

fl0rianr (approved):
LGTM now, thanks!

PR #2689, opened 2026-07-13
feat: flag comment slop in a pre-commit rule, and run pre-commit in CI

ianbmacdonald:
Implements the pre-commit rule + CI job proposed in #2685. AGENTS.md asks for the non-obvious WHY and forbids the WHAT, but nothing enforces it.

`tools/check_comment_slop.py` reads a **diff**, so it never complains about comments a contributor did not write. It flags:

- one concept explained at 3+ sites (the actual root cause — in practice it was never one over-written comment, it was the same thing restated at every site it touched)
- comment blocks over 12 lines
- comment:code density (informational)
- `--assert-comments-only`: strips comments + blank lines and asserts the code is byte-identical, so a cleanup PR can prove it instead of asking a reviewer to take it on faith

It **flags; it never rewrites.** Choosing which duplicate to keep needs judgement about where a confused reader lands. No LLM, so no injection vector.

Duplicate detection keys on shared distinctive vocabulary rather than prose similarity — the restatements are reworded, not copy-pasted, so `difflib` misses them entirely. Two exemptions keep it off idiomatic code: a comment pointing at the canonical explanation ("see X") is a cross-reference, and a test may restate the invariant under test.

**CI runs pre-commit on changed files only.** Repo-wide would fail on pre-existing Black drift on `main` that isn't the contributor's to fix — this also closes the gap that AGENTS.md claims Black is "enforced in CI" when no workflow references it.

### Validation

RED/GREEN against real history, not fixtures: flags the pre-cut #2588 diff that drew maintainer objections, passes the cleaned one, and independently caught a 3-site duplicate I had missed by hand. 37 unit tests. Passes its own check.

Reviewed by two independent models, which found and fixed a real false positive: three files each adding an alphabetised `#include` block were reported as slop, because `#` was read as a comment rather than a preprocessor directive. That class of wrong flag is the whole risk here, so the heuristics are the part worth attacking.

Not tested: Windows and macOS runners (Linux only locally); the CI job is exercised on a fork PR.

Design detail and the false-positive discussion are in #2685.


.github/workflows/docs_and_style.yml
```diff
@@ -55,6 +55,33 @@ jobs:
...
+        with:
+          python-version: '3.x'
+      - run: pip install pre-commit
+      - name: Run pre-commit on the changed files
+        # Scoped to what the change touches; running repo-wide would fail on drift that
+        # predates it and is not the contributor's to fix. The base comes from the merge
+        # base rather than github.base_ref, which only exists on pull_request -- keying on
+        # that skipped the job entirely in the merge queue, where main is actually
+        # defended. On a push to the default branch the merge base is HEAD, so nothing is
+        # checked and the run is a no-op.
```
superm1: this is true; but I wonder if it's worth having a second follow up that is purely `pre-commit run --all-files` to clean up the tree and then this could run the whole tree every time?

superm1 (commented):
I personally love the deterministic aspect of this.  Let's get other maintainers thoughts.

.github/workflows/docs_and_style.yml
ianbmacdonald: Yes — and it's smaller than I'd assumed. `pre-commit run --all-files` on current `main` touches **13 files (+182/−61)**: 10 Python reformats under `test/` and `.github/scripts/`, plus three one-line trailing-whitespace fixes (2 C++, 1 YAML). Nothing else in `src/cpp`.

I'll put that up as a separate PR. Once it lands I'll drop the changed-files scoping here and let this run the whole tree.

Geramy:
Mmmm iiidddddkkk about this one I’ll return back later again in a few hours.

tools/check_comment_slop.py
```diff
@@ -0,0 +1,492 @@
...
+        return [
+            t.string
+            for t in toks
+            if t.type
+            not in (
+                tokenize.COMMENT,
+                tokenize.NL,
+                tokenize.NEWLINE,
+                tokenize.INDENT,
+                tokenize.DEDENT,
```
fl0rianr: Blocking: removing INDENT and DEDENT makes semantically different Python programs compare equal. Moving a statement into or out of an `if`, loop, or function changes behavior while leaving the remaining token strings unchanged, so `--assert-comments-only` can falsely verify a code-changing PR.

Please preserve the structural tokens and add a regression test for indentation-only semantic changes, or exclude Python from this assertion.

tools/check_comment_slop.py
```diff
@@ -0,0 +1,492 @@
...
+            words += re.findall(r"[A-Z]?[a-z]+|[A-Z]+(?![a-z])", part)
+        return {
+            w.lower()
+            for w in words
+            if len(w) >= MIN_TERM_LEN and w.lower() not in STOPWORDS
+        }
+
+
+def run(cmd):
+    return subprocess.run(cmd, capture_output=True, text=True, check=False).stdout
```
fl0rianr: Blocking: this discards the Git command’s exit status and stderr. If a ref is invalid or `git show` fails, callers receive empty output; `comments_only()` can then report a successful verification for a comparison that never actually ran.

Please fail closed with `check=True` or explicit `returncode` handling, and add tests for invalid refs and files missing from one side of the comparison.

fl0rianr (changes requested):
Thanks for this — I like the deterministic, diff-scoped approach and especially that the hook only reports issues rather than rewriting code.

I found three correctness/integration issues that I think should be addressed before merge:

* `--assert-comments-only` can currently produce false passes. Python indentation is discarded, C/C++ line splicing is not modelled, and failed Git commands are treated like empty successful output.
* The CI base is always derived from the default branch. A PR targeting a `release-v*` branch may therefore lint changes outside the PR; the base should be selected from the actual event/target branch.
* The 37 new unit tests are not executed by the added workflow. Please add an explicit `python -m unittest test/test_comment_slop.py` step.

I left two inline comments on the most direct false-pass paths. Once those are fixed, the remaining limitations—supported file extensions, leading comments only, and the policy for legitimate comment blocks over 12 lines—could be handled as follow-ups or documented constraints.

.github/workflows/docs_and_style.yml
```diff
@@ -55,6 +55,33 @@ jobs:
...
+        run: |
+          BASE=$(git merge-base "origin/${{ github.event.repository.default_branch }}" HEAD)
+          pre-commit run \
+            --from-ref "$BASE" \
+            --to-ref HEAD \
+            --show-diff-on-failure
```
fl0rianr: Blocking for merge-queue usability: on a pull request run, the default checkout already represents the PR merged into the current target branch, so checking the full diff there gives contributors the result before approval.

On `merge_group`, however, `HEAD` may also contain PRs ahead of this one in the queue. Diffing `origin/main..HEAD` therefore lets the comment-similarity heuristic combine comments from multiple independently green PRs and fail for the first time in the queue.

Please keep the comment-slop decision PR-scoped and visible on every `synchronize` run. The merge-group job can rerun deterministic file-local hooks, but it should not broaden this heuristic to the combined queue diff—for example, skip `comment-slop` for `merge_group`, or evaluate each constituent PR independently.

fl0rianr (changes requested):

Geramy:
@ianbmacdonald do you have examples of this working? like test cases where its worked? Can I see some before and after examples if you wouldn't mind? I do like the idea of this generally. Agents dont sometimes care and continually write garbage in comments.

ianbmacdonald:
> @ianbmacdonald do you have examples of this working? like test cases where its worked? Can I see some before and after examples if you wouldn't mind? I do like the idea of this generally. Agents dont sometimes care and continually write garbage in comments.

Other than here, and the dogfooding of my own PRs, not yet.  It's based loosely on some of the stricter maintainer profiles in my development harness, which use an LLM process, and structured memory.

ianbmacdonald:
> @ianbmacdonald do you have examples of this working? like test cases where its worked? Can I see some before and after examples if you wouldn't mind? I do like the idea of this generally. Agents dont sometimes care and continually write garbage in comments.

It did this during the development https://github.com/lemonade-sdk/lemonade/pull/2448#issuecomment-4953250902

ianbmacdonald:
All three fixed, plus two more the sweep turned up. Thanks — the indent one was the important catch.

- **Python is compared as an AST**, not a token stream. Your exact case (`bar()` moved into the `if`) is a test now. You were right that preserving INDENT/DEDENT would not have been enough; the token stream was the wrong primitive for the assertion.
- **`run()` raises on a non-zero git exit.** `git diff --name-only` failing was the worst of it: no files, no offenders, "verified" — for a comparison that never ran.
- **CI base comes from the event's target branch**, so a PR onto a release branch is not diffed against `main`; and the slop heuristic is **skipped in `merge_group`**, exactly as you said — the deterministic hooks still run there.
- **The unit tests now run in CI.** We had written 45 and never wired them up.

Sweeping the same class found two more false passes:

- Deleting a `\` from the end of a `//` comment promotes the line below it from dead code to **live code** — C splices lines in phase 2, *before* comments are removed in phase 3, and the scanner was only doing phase 3. A change entirely inside a comment, and it compared equal.
- `chmod +x` was certified "no code changed", because the content is byte-identical. The mode is part of the tree, so file selection now reads `git diff --raw` and only a modified, same-mode source file can be a comment edit.

Each has a regression test that fails without the fix.

Still hardening this class, so more may land — I'll ping you when it settles rather than have you re-review a moving target.

ianbmacdonald:
Narrowed this to just the flagger + pre-commit hook + CI job — the part you asked for, @superm1. The `--assert-comments-only` check and its wiring move to a follow-up, so the deterministic low-risk half isn't held up by the higher-risk one. Phasing rationale in #2685.

ianbmacdonald:
@fl0rianr — I've pulled `--assert-comments-only` out of this PR entirely. All three of your blockers were in it, and rather than land a security-grade assertion here, it's moving to its own follow-up branch where they're fixed with tests. This PR is now just the deterministic slop flagger + hook + CI. Ready for another look when you have a moment; phasing rationale is in #2685.

abn:
Chiming in with my 2c as requested. Keeping the hook in-tree makes sense till the heuristics settle. Once they settle and prove stable, we should, imho, consider extracting this to an independent pre-commit hook repository (e.g., `lemonade-sdk/pre-commit-comment-slop`).

A few items to consider:
* **Missing Git Diff Overrides**: The `git diff` commands are run without overrides. If a developer has an external diff engine configured or has `color.ui = always` set globally, the diff output will contain color escape codes or a custom layout, breaking the parser.
* **False Negatives in Modified Block Comments (`/* ... */`)**: The parser resets `in_block = False` on gaps (unchanged lines). If a developer modifies a line inside a pre-existing block comment, the opening `/*` is not in the diff. The leading ` * ` will then be parsed as code rather than a comment, causing the slop check to skip the line.
* **Explicit Inline Bypasses**: There is currently no inline mechanism to exempt specific comment blocks from evaluations (like oversized blocks or duplication). This will block developers who legitimately need to write longer comments.

The following are not really blockers, but worth considering:
* **Quoted Filenames in Diffs**: If a filename contains spaces or special characters, Git quotes it (e.g., `+++ "b/some path/file.py"`), causing the extension suffix matching to fail.
* **Lack of License Header Exemptions**: Standard file header boilerplates often exceed the 12-line limit (`MAX_BLOCK_LINES`) and will trigger oversized block warnings on new or modified headers. This might be a non-issue fro the lemonade project right now though.

kenvandine (commented):
## Review

**Overview:** Adds `tools/check_comment_slop.py`, a diff-based heuristic linter that flags AGENTS.md violations (comments explaining WHAT instead of WHY, or the same concept re-explained at 3+ sites), wires it in as a local pre-commit hook, and adds a new `pre-commit` CI job that runs `pre-commit run --from-ref <merge-base> --to-ref HEAD` on PR/push/merge-queue events. This is also the first time Black/pre-commit actually run in CI at all — confirmed no existing workflow references `black` or `pre-commit`, closing the gap AGENTS.md already claimed was covered.

### Correctness

The diff-parsing and comment-classification logic is unusually careful; I traced it against the test suite and didn't find a bug:

- `added_lines()` correctly handles the `+++iter;`/`--- x;` prefix ambiguity by consuming hunk bodies strictly by the declared `@@ -a,b +c,d @@` counts rather than by line-prefix sniffing, and resyncs on any `@@`/`diff --git` line so a miscounted hunk can't swallow the next hunk's header.
- `_scan_line`'s comment/code classifier correctly handles `/* */` blocks spanning added lines, inline `/* note */ code;`, string/char literals (including escapes), and raw strings (`R"delim(...)"`) without being fooled by `"/*"` inside a string — including prefixed raw strings like `LR"(...)"`, which work even though `_raw_string_at` only checks for a bare `R`, because the character-by-character scan reaches the `R` on its own.
- `find_duplicates()`'s "shared-with-the-whole-group, not just one member" grouping correctly avoids transitive false grouping (A↔B↔C chaining when A and C share nothing).
- The CI workflow's ref resolution (`github.base_ref || merge_group.base_ref || default_branch`, then stripping `refs/heads/`) correctly handles both the bare-branch-name and full-ref forms these contexts actually produce.
- Reliance on `PRE_COMMIT_FROM_REF`/`PRE_COMMIT_TO_REF` being set by `pre-commit run --from-ref/--to-ref` is accurate.

### Minor observations (non-blocking)

- Python docstrings are entirely outside detection scope — `COMMENT_RE_PY` only matches `#`-prefixed lines, so triple-quoted docstrings (where WHY-explanation often lives in Python) are never checked for size or duplication.
- `POINTER_RE = r"\bsee\s+\S"` is broad: any block containing the word "see" anywhere is exempted from duplicate detection, not just genuine "see X" cross-references. Reasonable false-positive-avoidance tradeoff, but worth knowing.
- The CI job is likely a silent no-op on direct `push` events to `main`: at push time `origin/$BASE_BRANCH` already points at the pushed commit, so `merge-base(origin/main, HEAD) == HEAD` and the diff is empty. Harmless (the PR's `pull_request` run already gated the merge) but non-obvious, and not called out in the otherwise thorough inline comments on that step.
- No test exercises `main()` itself (argparse wiring, exit code 0 vs 1, `PRE_COMMIT_FROM_REF` vs `--from-ref` precedence) — all 37 tests hit internals directly. A single end-to-end smoke test of the entry point would close the last gap.
- `added_lines()` computes `is_comment` via the simple regex `is_comment_line()` for every language, but `collect()` immediately discards and recomputes it via `_scan_line` for all non-`.py` files — mildly redundant, not a bug.

### Style / Risk / Test coverage

Matches AGENTS.md's own commenting philosophy well (WHY, not WHAT, throughout the new file). Risk is low: the hook only flags, never rewrites, operates on diffs so it can't complain about pre-existing code, and the specific false-positive class called out in the PR description (alphabetized `#include` blocks misread as prose) has a dedicated regression test. Test coverage is strong — 37 unit tests across diff-parsing edge cases, comment classification, and duplicate-grouping semantics, plus RED/GREEN validation against real PR history.

kenvandine (approved):

fl0rianr (approved):
Re-reviewed after the scope reduction. My previous blockers are resolved or have moved with --assert-comments-only to the follow-up. The remaining heuristic limitations are non-blocking for this flagger-only PR.

Please **update** the **PR description** to reflect the narrowed scope

ramkrishna2910 (approved):
LGTM. The only failing check is the router-onnxruntime flake (test_620, #2883) from a stale Jul 25 run — unrelated to this PR and fixed on main by #2871, so an update-branch/re-run should go green. Leaving merge to the other maintainers.

Geramy (approved):

PR #2690, opened 2026-07-13
chore: normalize the tree with pre-commit run --all-files

ianbmacdonald:
Follow-up to #2689, per @superm1's suggestion there: clean the tree so a pre-commit CI job can check *everything* rather than only the files a contributor happened to touch.

Pure tool output — `pre-commit run --all-files` with the pinned hooks. No hand edits.

- 10 Python files reformatted by Black (pinned 26.1.0), under `test/` and `.github/scripts/`
- 3 one-line trailing-whitespace fixes (2 C++, 1 YAML)

After this, `pre-commit run --all-files` is green on the whole tree, so #2689 can drop its changed-files scoping.

### It is behaviour-preserving, and mechanically so

Rather than ask anyone to read 182 lines of reformatting:

- **All 10 reformatted `.py` files are AST-identical** before and after — `ast.parse` → `ast.dump` compares equal for every one.
- **The 3 non-Python changes are whitespace-only** — `git diff -w` is empty.

So there is nothing here to review line by line.

Not tested: nothing to test — no behaviour changed. CI is the check.

### DIFF
--- a/.github/scripts/auto_label.py
+++ b/.github/scripts/auto_label.py
@@ -256,6 +256,7 @@ def has_write_access(login, repo):
         return False
     return data.get("permission") in WRITE_PERMISSIONS
 
+
 def is_bot_user(user):
     """Return True for GitHub bot accounts.
     Bot activity should not count as community engagement for priority labels.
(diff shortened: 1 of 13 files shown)


superm1 (approved):

superm1:
Thanks so much.

PR #2692, opened 2026-07-14
Rename process_unix files to process_linux

superm1:
This reflects actual usage better.  Also drop all the pointless `#ifdef __linux__` checks since they're only for Linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -685,7 +685,7 @@ elseif(UNIX)
     list(APPEND SOURCES_CORE src/cpp/server/platform/metrics_linux.cpp)
     list(APPEND SOURCES_CORE src/cpp/server/platform/suspend_linux.cpp)
     list(APPEND SOURCES_CORE src/cpp/server/utils/platform/archive_unix.cpp)
-    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/process_unix.cpp)
+    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/process_linux.cpp)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -675,17 +675,17 @@ elseif(APPLE)
     list(APPEND SOURCES_CORE src/cpp/server/utils/platform/path_macos.cpp)
     list(APPEND SOURCES_CORE src/cpp/server/platform/metrics_macos.cpp)
     list(APPEND SOURCES_CORE src/cpp/server/platform/suspend_stub.cpp)
-    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/archive_unix.cpp)
+    list(APPEND SOURCES_CORE src/cpp/server/utils/platform/archive_linux.cpp)
```
fl0rianr: The blanket rename misses an important distinction in the platform dispatch: `archive_unix.cpp` and `network_unix.cpp` are still explicitly compiled by the `APPLE` branch above, so they are shared POSIX implementations, not Linux implementations. Only `process_unix.cpp` is actually Linux-only, because macOS selects its separate `process_macos.cpp`; renaming that file and removing its `__linux__` guards is therefore correct. The archive and network files should remain `*_unix` or preferably become `*_posix`, including `Unix/PosixArchivePlatform`. As written, every macOS build now deliberately compiles `archive_linux.cpp` and instantiates `LinuxArchivePlatform`, which misrepresents the real platform boundary and makes it very easy for a later Linux-specific change to silently break macOS.

fl0rianr (changes requested):

fl0rianr (approved):
LGTM now, thanks.

Issue #2406, opened 2026-06-24
[Router][UI] Author collection.router in the custom-collection panel


ramkrishna2910:
## Goal
Extend the desktop app's **custom-collection panel** so users can author a `collection.router` (pick candidates, set `default_model`, add classifiers/rules or an L0a `router` prompt) without hand-writing JSON — the same way the panel already assembles `collection.omni` collections today.

## Status: optional follow-up — NOT on the engine critical path
- v1 authoring works without this: hand-author the lean JSON (see #2375) or import via the CLI (`recipe_import`). The engine milestone ships independent of any UI.
- Per `AGENTS.md`, **UI/frontend changes are handled by core maintainers only** — this issue tracks the work; implementation/ownership sits with the UI maintainers.
- Could live under the **GUI3** milestone instead of the router milestone — placed here for visibility; move if preferred.

## Scope (sketch)
- A "Router" collection type in `CustomCollectionPanel.tsx` / `customCollections.ts`: select candidate models (→ `routing.candidates` + `default_model`), and either (a) an L0a NL `router` prompt, or (b) add classifiers (`semantic_similarity`/`classifier`) + rules.
- Reuse the existing component-picker UX; emit the **lean local form** (`components` + `routing`, no `models[]`).
- Validation surfaced in-panel (dangling refs, `route_to`/`default_model` ∈ candidates).

## Out of scope
Engine behavior, schema (frozen in #2375), redistribution/`models[]` embedding, the L0(b) NL→policy compiler.

## Acceptance
A user assembles a working `collection.router` from the panel and it routes end-to-end (mirrors the #2388 demo) with no hand-edited JSON.

---
**Depends on:** #2375 (schema), #2385 (dispatch working). UI-owned; non-gating.

ramkrishna2910:
Done — implemented and merged in #2693 (the desktop custom-collection panel now authors `collection.router` policies, including the classifier/LLM-signal forms). The PR didn't carry a `Closes #2406`, so closing manually.

PR #2693, opened 2026-07-14
[Router] [UI] Author collection.router in the custom-collection panel (#2406)

meghsat:
Adds a GUI for authoring collection.router collections. Users can now build, edit, export, and register a router without hand-writing JSON.

The panel is opened from File -> New Hybrid Router and POSTs a schema-valid collection.router JSON to /v1/pull, registering it as a virtual model the same way collection.omni works.
image

Three routing modes are supported:
- NL Router: user picks a small classifier LLM and writes a free-text routing prompt. The LLM reads each incoming message and replies with the name of the candidate to use.
- Quick Rules: a simplified form-based interface for adding common condition-based rules without drag-and-drop. 
- Advanced Rules: a visual drag-and-drop boolean expression builder. A split-panel layout with a classifier panel on the left and a canvas on the right where users build nested AND/OR/NOT condition trees by dragging chips from a toolbox sidebar.

image

Feature breakdown:
- Candidate picker: searchable multi-select dropdown of all registered LLMs.
- Default model: single-select dropdown filtered to checked candidates, maps to routing.default_model.
- Classifier editor: declare named ML classifiers that rules can reference. Supports three types:
   - classifier: HF text classification model, editable labels, default label, on_error (fail-open/fail-closed)
   - semantic_similarity: embedding model with a dynamic concept editor. Each concept name becomes an output label 
     image
   - llm: small LLM as a classifier with a free-text prompt
- Rule list: ordered list of rules. Each rule has an editable ID, a route-to dropdown, and a condition tree
- Condition tree canvas: drag-and-drop nodes from the Toolbox, pan with left-drag, zoom with scroll wheel. Fullscreen expand via portal
- Gate nodes: AND, OR, NOT with inline operator toggle
- Condition signals: Keywords Any/All, Regex, Min/Max chars, Has Tools, Has Images
- Classifier signal: references a declared classifier by ID. label dropdown auto-populated from classifier's declared labels or concept keys, min/max score band inputs
- JSON Preview: renders the exact /v1/pull body; copy and download buttons. Also, available in the Toolbox sidebar during full screen canvas mode

image

### Backend dependency & merge order

This UI targets the routing contract from #2698 ([Router] llm router classifier + routing.router desugaring, still open). Current main rejects routing.router (Natural Language mode) and type: "llm" classifiers at registration time, so those two flows will return a server error until #2698 merges. However, all rules-based routing (quick + advanced, classifier and semantic_similarity types) works against current main today. This PR is intentionally mergeable first: the router tests here validate the emitted JSON at the builder level against the frozen fixtures shared with the C++ tests, and no server-side behavior changes are required from this PR. When #2698 lands, NL mode and llm classifiers light up without further UI changes.


fl0rianr (changes requested):
Thanks for the substantial work here — the overall direction is promising, especially the separation between the router draft model, JSON generation, and the visual rule builder.

However, I do not think this is merge-ready yet. The main blockers I found are:

* The UI exposes `routing.router` and `llm` classifiers, but the current server explicitly rejects both because the corresponding backend work from #2405 is not implemented yet.
* The router modal is closed before `/v1/pull` succeeds, so any backend or validation error discards the user's full unsaved router configuration.
* Frontend validation does not cover several server-side requirements, including duplicate rule/classifier IDs, invalid classifier labels, score ranges, `min_score > max_score`, and invalid numeric or regex values.
* `test:router-collections` references `test/app/routerCollections.test.cjs`, but that file is not included in the PR and is also newly added to `.gitignore`.
* Removing a child from an AND/OR gate can leave a one-child gate, although the same frontend validation considers that invalid.
* The semantic-similarity model picker only accepts the `embeddings` label, while the existing model label logic also supports `embedding`.
* Switching to Advanced Rules through the radio button clears existing rules and classifiers, while “Expand in Advanced” preserves them.

The CI is also currently not green, including the app regression job.

One broader concern: this is a very large UI change — almost 6,000 added lines across 15 files. At this point I would strongly prefer not to expand GUI2 with a feature of this size and complexity. This work would fit much better into GUI3 on the `kpoin/ui-testing` branch, where the new model and collection workflows are already being redesigned. Keeping it there would avoid adding another large interaction model to GUI2 shortly before GUI3 replaces it.

My recommendation would therefore be either:

1. move/rework this feature for GUI3 on `kpoin/ui-testing`, or
2. significantly reduce the GUI2 scope to only the backend-supported deterministic rule authoring, with the blockers above fixed and proper regression coverage added.

For the current version, I would clearly **request changes** rather than merge it into `main`.

meghsat:
Thank you for the feedback @fl0rianr! My latest commits address the bug fixes.

fl0rianr (changes requested):
Thanks for addressing the previous feedback. I reviewed the latest revision again, and several of the fixes are real improvements:

* the router editor now remains open when registration fails;
* the router collection test file is committed;
* both `embedding` and `embeddings` labels are recognized;
* duplicate IDs, invalid regexes, and `min_score > max_score` are now checked;
* classifier ID changes update references in the rule trees;
* the app regression workflow is now passing.

However, I still see several blockers:

1. **The default NL Router flow is still unsupported by the current server.**

   The UI still defaults to NL Router and emits `routing.router`. Advanced Rules also still exposes `llm` classifiers. The current parser explicitly rejects both paths as deferred to #2405.

   This means the most prominent/default workflow can be fully configured in the UI but cannot be registered successfully against the current `main` server.

   I think these options must either be hidden/disabled until #2405 lands, or this PR must depend on and be rebased after the backend implementation.

   We can integrate future features if it's properly planed and guarded, still think GUI3 is the better place.

2. **The new router tests are not part of the normal app regression run.**

   `test/app/routerCollections.test.cjs` is now present, but the CI runner only discovers tests under `test/app/app-regression/`. The new test therefore does not appear to run in `Docs And Style`.

   More importantly, it verifies the TypeScript builder's JSON shape, including successful NL Router and `llm` classifier output, while the current C++ parser intentionally rejects those same structures. A contract or end-to-end test should verify that generated JSON is accepted by the current server parser.

   Either where it lands, such a feature is prone to regression and needs a test.

3. **Frontend validation still does not match the server contract.**

   Remaining examples include:

   * score values outside `[0, 1]` are not rejected;
   * `min_chars` and `max_chars` are not validated as non-negative integers;
   * classifier conditions are not checked against the classifier's current labels or concepts;
   * omitting a condition label is not checked against the presence of a classifier `default_label`.

   In particular, changing a classifier's labels or semantic concept names can leave stale labels in existing rules that pass frontend validation but are rejected by the server.

4. **Routing-mode changes can still lose data.**

   Selecting Advanced Rules directly still resets both `rules` and `classifiers` without confirmation.

   Switching a complex Advanced configuration to Quick Rules is also unsafe: Quick Rules only displays a supported subset of the tree, and editing that representation can rebuild the rule without hidden nested or classifier conditions.

   Mode conversion should be explicitly lossless, blocked when incompatible, or confirmed before discarding data.

5. **Deleting from an AND/OR gate can still leave an invalid unary gate.**

   `collapseIfSingle()` preserves an AND/OR operator with one remaining child, while validation rejects AND/OR gates with fewer than two children. Removing one child from a two-child gate can therefore make the rule unsavable.

6. **Multi-condition Quick Rules do not reopen as Quick Rules.**

   Quick Rules can create a flat AND/OR containing multiple conditions, but the edit parser only identifies a rule as Quick-compatible when its tree is a single leaf. A saved two-condition Quick Rule therefore reopens in Advanced mode.

There is also still the broader scope concern. This PR has roughly 6,400 added lines across 15 files and introduces a substantial new interaction model into GUI2, I can understand this is quite a challenging feature, exactly because of that we should avoid doing it twice. 

**A Bit of strategy:**
This won't make it into today release. I doubt there will be a release next Wednesday due to the conference and alle release people being occupied there.
We plan to have GUI3 beta ready for the conference and merge it after same fixes soon after. Due to that I would still strongly prefer implementing and refining a feature of this size in kpoin/ui-testing rather than significantly expanding GUI2 shortly before its replacement.
Ping me on Discord and I'm happy to give you guidance on the best way of implementation for GUI3, would very much welcome this enhancement there. Your work here is a very good foundation for that.

If you see it necessary for GUI2, a much narrower implementation limited to currently supported deterministic routing would be more appropriate.

---

In the end, the fixes move the PR in the right direction, but with the unsupported default workflow and the remaining data-loss and validation issues, the open straegy question, I would keep my verdict.

eddierichter-amd (changes requested):
Thanks for the work here. I reviewed the current head and agree with Florian’s latest requested-changes review, so I’ll refer back to his comment rather than repeat the unsupported workflows, test-discovery gap, validation differences, destructive mode conversion, unary gate, and Quick Rule reopening
  findings as new.

  I found one additional round-trip issue:

  ### Semantic-similarity classifier fields are dropped on save

  `routingToRouterCollectionDraft()` reads `default_label` and `on_error` for semantic-similarity classifiers, but `buildRouterCollectionPullRequest()` only writes those fields for `type: "classifier"`.

  Opening and saving a semantic-similarity policy can therefore silently remove:

  ```json
  {
    "default_label": "...",
    "on_error": "match_true"
  }
```

  This can change the policy’s failure behavior from match_true to the default match_false, and removing default_label can invalidate an existing condition that omits label.

  Please preserve these fields for semantic-similarity classifiers, even if the UI does not currently expose controls for editing them.

  ### Naming

  One small UX wording request: could we replace the visible “NL Router” label with “Natural Language”?

  “NL” is internal shorthand and is not immediately understandable to users encountering the feature for the first time. “Natural Language” communicates the mode directly and is more consistent with the other user-facing labels, “Quick Rules” and “Advanced Rules.”

  This should be updated consistently anywhere the mode name appears in the UI and user-facing descriptions.

  Overall, the UI direction is promising, but I agree with keeping the change request open until the existing contract, validation, and data-loss issues are resolved.

**Review by fl0rianr — CHANGES_REQUESTED** [review]
Thanks for the continued fixes. I reviewed the latest head again.

Several earlier points are now addressed: regression test placement, score and character validation, label checks, mode-switch confirmations, Quick Rule round-tripping, and keeping the editor open after failed registration.

I also rechecked the unary AND/OR case. The author is correct that validation blocks Save and JSON Preview, so no invalid JSON reaches the server. I would downgrade that item to a non-blocking UX issue: automatically replacing a one-child AND/OR with its remaining child would be cleaner, but it is not a correctness blocker.

The following issues remain:

1. **Candidate and classifier model filtering are mixed.**
   `classification` ONNX models can appear as routing candidates, while ordinary chat LLMs can be selected for `classifier` entries. Separate capability-filtered lists are needed for chat candidates, classification models, embeddings, and later LLM classifiers.

2. **Editing can silently change valid server policies.**
   `has_tools: false` and `has_images: false` are parsed and later serialized as `true`. Unsupported conditions such as `metadata` may also be dropped during editing. Unsupported policy shapes should either round-trip losslessly or be rejected for editing with a clear warning.

3. **Quick-mode detection is broader than the Quick editor.**
   For example, `keywords_all` may be classified as Quick-compatible even though the Quick editor cannot represent it.

4. **Loaded Quick routers can generate duplicate rule IDs.**
   The Quick Rule counter is not initialized from existing rules, so reopening a router containing `rule-1` and adding a rule can create another `rule-1`.

5. **A label-validation edge case remains.**
   A stale rule label can pass validation after all classifier labels are removed, and `defaultLabel` should also be verified against the current labels or semantic concepts.


I would keep Request changes, but the unary-gate item should no longer be treated as a blocker.

**Comment by meghsat** [comment]
@eddierichter-amd  Thank you for flagging the bug, fixed it along with renaming NL to Natural Language
@fl0rianr  Thanks much for pointing the bugs! Fixed them in my latest commit

**Review by fl0rianr — COMMENTED** [review]
I checked the current head `389b77c1`. Several findings from the previous review have been addressed. I found four remaining issues in the latest revision:

1. **Model capability filtering is still incorrect.**

   `classifierModelOptions` checks for the label `classifier`, while the classification models now present on `main` use `classification`. After the required rebase, those models would be excluded from the classifier picker while remaining eligible as chat candidates. Conversely, image, speech, edit, and reranking models may appear as classifier options.

   Please explicitly require the `classification` capability for `type: "classifier"` and exclude it from chat candidates.

2. **Explicit LLM classifiers do not match the current #2698 contract.**

   The UI exposes `type: "llm"` classifiers but only authors and serializes a prompt. The current #2698 schema also requires a non-empty `labels` array.

   Please add label authoring, serialization, and validation, or hide the explicit LLM-classifier option until that contract is supported.

3. **Negative Boolean conditions still have round-trip problems.**

   `has_tools: false` and `has_images: false` are normalized to `not: { ...: true }`, but the exact `JSON.stringify()` comparison then incorrectly marks them as lossy.

   More importantly, parsing `not: { has_tools: false }` overwrites the existing negation instead of toggling it, changing the condition’s meaning.

   Loss detection should compare canonical semantics, and nested negation should use XOR/toggling rather than assignment.

4. **`lossyRuleIds` is not updated after rules are removed or reset.**

   Deleted rules can remain listed as lossy, causing an irrelevant warning and acknowledgement requirement. Please recompute or clear this state whenever the rule set changes.

The branch is also currently 24 commits behind `main`, not mergeable, and has no checks attached to the current head. Please rebase and rerun CI before final review.

**Review by fl0rianr — COMMENTED** [review]
Thanks for the latest update. I reviewed the new head `8ef1f3cb`.

The previous four findings were genuinely addressed:

* classifier models are now filtered using the `classification` capability;
* classification models are excluded from chat candidates;
* explicit LLM classifiers now support labels;
* negative Boolean conditions and stale `lossyRuleIds` were corrected.

However, the current head introduces new merge blockers.

### 1. The application no longer compiles through Webpack/`ts-loader`

`stableStringify()` is declared to return `string`:

```ts
export function stableStringify(value: unknown): string
```

but its fallback returns:

```ts
return JSON.stringify(value);
```

`JSON.stringify()` can return `undefined`, so its TypeScript type is `string | undefined`. With the repository's strict TypeScript configuration and Webpack using `ts-loader` without `transpileOnly`, this is a compile-time type error.

Please make the function total, for example by handling the `undefined` case explicitly, and verify the actual production Webpack build rather than only transpiling the files.

### 2. The Debian package build is broken

The current `C++ Server Build, Test, and Release` workflow fails in:

```text
Build Lemonade .deb Package
  → Build Debian package
```

The separate Debian ARM64 package succeeds, but the normal Debian package does not. This is a release-blocking regression and needs to be fixed before the PR can be considered merge-ready.

### 3. `Docs And Style` is failing

To be precise, the documentation-specific jobs are not the failing part:

* `backend-docs-drift` passes;
* `markdown-link-check` passes.

The failing job is `app-regression-tests`, in the `Run app regression tests` step. Please fix the failing regression test and rerun the workflow.

### 4. The test setup does not protect against TypeScript build regressions

The router regression test loads TypeScript through `ts.transpileModule()`. That transpiles the files but does not perform the strict type checking that the real Webpack/`ts-loader` production build performs.

This allowed a large regression suite to be added while the application itself no longer compiles.

Please add at least one real application build or type-check step to the normal app CI path, for example:

```text
npm run build
```

or an equivalent strict `tsc`/Webpack check. The regression tests should complement the actual build, not replace it.

### 5. The backend merge-order dependency remains

#2698 is still open. Current `main` still rejects both:

* `routing.router`;
* explicit classifiers with `type: "llm"`.

The UI still defaults new routers to the Natural Language mode. Therefore the default user flow remains unavailable until #2698 lands. This one is close since only CI fail needs to be addressed but I'm not sure when the author will go at this again.

This PR must either:

1. be rebased and tested after #2698 is merged; or
2. temporarily disable the unsupported Natural Language and explicit LLM-classifier paths.

The latest commit fixes the previously reported authoring issues, but the broken Webpack build, failing Debian package, failing app regression job, and unresolved backend dependency are hard blockers.

I am keeping **Request changes**.

fl0rianr (commented):
With the depending other PR still going but let's expect them to merge or this code waiting a bit for the server side partner the remaining issues reduce.

I only have two smaller test-organization concerns remaining:

1. The router regression test says its generated payloads are accepted by the current C++ parser, but the test itself only executes the TypeScript builder and compares JavaScript objects and fixtures. It does not invoke the C++ parser. The wording should be narrowed.

2. There are still two separate router test files with different behavior: the CI suite skips Natural Language routing (the depending topic), while the standalone npm test treats it as supported. These should ideally be consolidated so local and CI coverage share one source of truth.

If we want to merge it first, please also document the dependency and intended merge sequence with #2929 in the PR description so future readers understand why the UI may land first.

Subject to the remaining C++ build/release workflow completing successfully, I no longer see a reason to block this solely on the backend merge order.

meghsat:
@fl0rianr Thanks for catching the testing file discrepancies. For the local tests, added a wrapper around the CI test. Also, I think you meant PR #2698 to be included in the PR description, did that.

fl0rianr (approved):
Approved within the agreed scope of this PR.

The LLM-router implementation issues raised in my previous reviews have been addressed: authored/canonical candidate handling, strict structured responses, rationale attribution, routing-context propagation, schema alignment, constrained router invocation, and regular CI coverage now look sound.

The remaining model-residency and slot-lifecycle concern is addressed by the stacked follow-up PR #2729. That PR now has a fully green CI matrix, including the C++ server workflow, Linux distro builds, llama.cpp validation, and vLLM validation.

I will discuss how we handle merging....

SlawomirNowaczyk (approved):
Looks good, I suggest we merge this.

However, I still think there is room for simplifying the UI, so let's keep iterating, maybe based on some of my suggestions here: https://github.com/lemonade-sdk/lemonade/pull/2728

eddierichter-amd (approved):
LGTM

fl0rianr:
So for some context @SlawomirNowaczyk I would not work on this much further unless bug fixing. With to current plan GUI2 will have only two more releases until switch for GUI3.

#2698 is merged now so the until now supported option should be available as well. Although #2729 remains as a fix which is not a blocker here, but should come into focus as well.

ramkrishna2910 (approved):

PR #2695, opened 2026-07-14
feat: autoopt-required backend changes

pwilkin:
This PR is 1/2 to implement #2613 by adding the necessary backend support - API endpoints for launching `llama-bench` and `llama-fit` for the llamacpp engine to perform diagnostics and performance bencharks to determine optimal setup.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,7 @@ All core endpoints are registered under **4 path prefixes**:
 - `/v0/` — Legacy short
 - `/v1/` — OpenAI SDK / LiteLLM compatibility
 
-**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `health`, `pull`, `load`, `unload`, `delete`, `params`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `log-level`, `logs/stream`
+**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `health`, `pull`, `load`, `unload`, `delete`, `params`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `log-level`, `logs/stream`, `backends/llamacpp/fit-params`, `backends/llamacpp/bench` (SSE)
 
 **Ollama-compatible endpoints** (under `/api/` without version prefix): `chat`, `generate`, `tags`, `show`, `delete`, `pull`, `embed`, `embeddings`, `ps`, `version`
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2195,6 +2195,22 @@ if(EXISTS "${_AUTO_TUNE_TEST_SRC}")
     add_test(NAME AutoTuneTest COMMAND test_auto_tune)
 endif()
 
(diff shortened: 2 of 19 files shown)


fl0rianr:
Sorry, my bad, I changed the base as the API PR is intended for main only. Now it's obvious it was not a wrong chosen base at opening the PR rather the code was regarding the wrong base. Please correct this.

API --> main
Frontend --> kpoin/ui-testing only /prototype

pwilkin:
Ah, all right.

pwilkin:
Should be correct now.

fl0rianr (changes requested):
Thanks for working on this. I think the current PR needs to be reduced and corrected before it is safe to merge.

The main blocker is the `fit-params` invocation. The code always passes `-fitp on`, but upstream `llama-fit-params` only calls `common_fit_params()` when `fit_params_print` is false. With `-fitp on`, it only calls `common_fit_print()` and prints the estimated memory table. Therefore `fit_target_mib` does not affect the result, no fitted arguments are produced, and the endpoint is currently an estimator rather than a fit operation. The parser then treats a table-only result as `fits_fully=true` because `fitted_ngl` and `fitted_ncmoe` retain their defaults.

I also think this duplicates too much of Lemonade's existing benchmark infrastructure. `lemonade bench` already handles backend discovery, controlled load/unload, context and backend-argument comparisons, TTFT, TPS, memory measurements, failures, and structured JSON output. The Auto-Tune documentation explicitly identifies this existing command as the source of benchmark data. We should extract or reuse that implementation rather than create a second llama.cpp-specific benchmark model and result schema.

Running `llama-bench` directly also does not coordinate with an already loaded `llama-server`. AutoOpt can therefore load the same model twice, distort the measurements, or cause an OOM. The existing Lemonade bench flow avoids this by explicitly controlling load and unload.

Additional issues:

* SSE progress is only emitted for lines containing `main:`, but the command does not pass llama-bench's `--progress` option. Client disconnect cancellation is therefore unreliable.
* The endpoint accepts unrestricted raw `args`, allowing callers to append another `-m`, `--fit-target`, device, or RPC argument after the validated arguments.
* `ctv` is passed to llama-bench but omitted from the returned parameter set.
* `/health` always advertises `llamacpp-tools`, even when no compatible backend or sibling tool binary is installed.
* The unconditional ROCm `--direct-io` load behavior is unrelated to these endpoints and should be a separate, independently tested change.

My recommendation is to keep the first AutoOpt implementation intentionally small: reuse the existing Lemonade benchmark path for one short scenario and one or two safe configurations, and keep an optional typed llama.cpp fit capability as a separate concern. The ROCm direct-I/O behavior, generic mmproj option, and unrelated metadata expansion should be split out.

pwilkin:
Okay, I'll submit the `--direct-io` / `--no-mmap` stuff separately because I think it's pretty critical - without it some models just hang while loading (and a fundamental resolution is locked upstream behind a refactor of the loading mode options). But if I understand correctly, you're suggesting to just scrap the backend and reuse `lemonade bench` for whatever benchmarks should be done? I agree this will be cleaner in terms of keeping the number of running model instances under control (the problem with `llama-bench` is that it's technically a model load outside of the normal model load management, I haven't actually thought of the scenario when you run it on an instance with a model already running).

After looking at the params supported, I think all the tests should be repeatable with `lemonade bench` since it allows passing custom `llama.cpp` parameters, so I can fit the `-b -ub` tests in there. There's a methodological difference when it comes to measuring prefill (PP t/s vs TTFT), but it's irrelevant for the end result. So, I'm going to close this one and instead try to do the benchmark using existing `lemonade bench` features. Should've looked earlier instead of trying to bruteforce it, my bad :)

PR #2696, opened 2026-07-14
feat: autoopt wizard for llama.cpp optimal parameters

pwilkin:
This is meant to be added on top of #2695 as the frontend part for the autoopt wizard. Utilizes the backend queries to `llama-bench` and `llama-fit` to perform a full optimization pass for llama.cpp parameters.

### DIFF
--- a/prototype/ui-redesign/package.json
+++ b/prototype/ui-redesign/package.json
@@ -10,7 +10,10 @@
     "test": "npx playwright test",
     "test:headed": "npx playwright test --headed",
     "test:a11y": "npx playwright test tests/a11y.spec.ts",
-    "test:preset-intent": "node tests/preset-intent.runtime.cjs"
+    "test:preset-intent": "node tests/preset-intent.runtime.cjs",
+    "test:autoopt-glue": "node tests/autoopt-glue.runtime.cjs",
+    "test:autoopt-synthesize": "node tests/autoopt-synthesize.runtime.cjs",
+    "test:autoopt-store-scope": "node tests/autoopt-store-scope.runtime.cjs"
   },
   "devDependencies": {
     "@axe-core/playwright": "^4.11.3",
--- a/prototype/ui-redesign/src/api.ts
+++ b/prototype/ui-redesign/src/api.ts
@@ -355,6 +355,28 @@ export interface SlotTimings {
   predicted_per_second: number;
 }
 
+export type JobStatus = 'queued' | 'running' | 'paused' | 'interrupted' | 'completed' | 'failed';
+
+export interface JobStep {
+  id: string;
+  op: string;
+  status: 'pending' | 'running' | 'completed' | 'failed' | 'skipped' | string;
+  duration_ms?: number;
+  error?: string;
+  output?: unknown;
+}
+
+export interface JobRecord {
+  id: string;
+  name: string;
+  status: JobStatus;
+  inputs?: Record<string, unknown>;
+  context: Record<string, unknown>;
+  cursor?: string;
+  steps: JobStep[];
+  created_at?: string;
+}
+
 export interface SlotData {
   id: number;
   n_ctx: number;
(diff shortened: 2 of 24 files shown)


fl0rianr (changes requested):
Thanks for this work, much appreciated — I re-reviewed the current squashed version.

The cleanup looks correct now: all root `/src`, CMake, and backend-test changes have been removed. I reviewed the remaining GUI code assuming the fit/bench endpoints, health feature flag, and required model metadata are merged through `main`.

A few functional issues remain:

1. **mmproj_enabled still requires backend support.**
   The frontend persists `mmproj_enabled: false`, but current `main` still automatically loads the available projector and reserves `--no-mmproj` from custom arguments. Merging only the tool endpoints is therefore insufficient. The corresponding load-option support must also land in `main`, or the vision-disable optimization should be removed.

2. **“Try now without saving” mishandles already loaded models.**
   It always calls `loadModel()`. Backend, context, and llama.cpp arguments are load-time options, so an already loaded model must use `reloadModel()` with the temporary recommendation.

3. **The MTP sweep does not reload between candidates.**
   It calls `loadModel()` repeatedly for different draft lengths but unloads only after the complete loop. Later candidates may therefore fail or run with the first configuration. Reload/unload each candidate and place final cleanup in `finally`, including cancellation and error paths.

4. **Dense-model partial offload is not reproduced.**
   When fit returns `fits_fully: false` and a finite `fitted_ngl`, synthesis only explains the value in the rationale; it does not emit `-ngl N`. The final preset may consequently not reproduce the configuration that actually fitted. Please propagate the fitted value and add a dense-model test.

5. **Saved runs are not scoped to the server or hardware.**
   All recommendations share one global local-storage key. Switching the configured Lemonade server can expose and apply CUDA/ROCm/Vulkan settings generated on another machine. Scope runs at least by normalized base URL and validate the checkpoint/backend before applying them.

6. **Some progress and UI claims do not match execution.**

   * The run is browser-orchestrated, not a persistent server-side job; reloading the page interrupts it.
   * `AUTOOPT_STAGES` lists `load_validation` before `synthesize`, while execution does the reverse.
   * “Deep Benchmark” promises measured KV-quant impact and a real-load smoke test, but those operations are not currently implemented.

Please add focused regression coverage for temporary application to a loaded model, MTP candidate reloading, dense `fitted_ngl`, and stale runs from another server.

The overall implementation and expanded scope are fine, but these correctness issues should be resolved before merge. Thanks!

pwilkin:
All right, just a quick question: do you think it would be a better idea to add a backend driver for the autoopt tasks, or to keep it client side and just add a tracker / allow resuming an interrupted task when the window was closed in the meantime?

fl0rianr:
I think the UI (or CLI , you can think of it like that) should give params it would like to test towards the API (server) and the server does this work. A rather complicated test can be triggert via UI also kinda simulating someone testing via CLI over multiple turns. Ideally we have one server code for this and CLI an UI can benefit both whiles the UI story is typically much easier.

pwilkin:
Aight, so I guess a different backend PR will be in order :)

fl0rianr (changes requested):
Wonderful change, I like it - we are so near to an Incredible feature!

I still found five merge blockers:

1. **The available-memory calculation has a unit bug.**
   `Math.max(1024, host_ram_gb * 0.9) * 1024` treats a 64 GB unified-memory machine as having 1 TiB available. The same issue exists in the CPU fallback. This can make Fast Scan recommend contexts and offload settings that cannot fit. Apply the GiB→MiB conversion before the minimum, and add controller-level tests for realistic 16/64/128 GB systems.

2. **A fallback load is reported as if the primary context succeeded.**
   Primary and fallback loads write into the same measurement keys, while the plan always records `ctxPrimary`. If 32K fails and 2K succeeds, the result still marks the 32K configuration as measured and successful. Synthesis also continues using the heuristic context, so it may recommend the configuration that already failed to load. Record the actual successful context and cap the final recommendation to it.

3. **CPU-only systems can benchmark CPU but recommend Vulkan.**
   The controller falls back to `cpu`, while `synthesize()` rebuilds the candidate list, removes CPU/system, and defaults to Vulkan. Please use one consistent candidate list throughout, or explicitly reject unsupported CPU-only runs before starting.

4. **The unrelated Hugging Face model changes appear to regress selection and download.**
   HF results are passed into the generic model list, but selected details are still resolved only from `allModels`. The dedicated HF detail, variant, and download path was removed, so selecting an HF result resolves to no detail model and there is no replacement download action. Please revert these unrelated changes or restore and test the complete HF selection/download workflow.

5. **Server-scoped persistence does not handle a runtime server switch.**
   The singleton reads runs only once, while the storage key is recalculated from the current `api.baseUrl` on every write. Switching from server A to B leaves A’s runs and active pollers in memory and can write them into B’s key or query A’s job IDs against B. On scope changes, stop local pollers, load the new server’s runs, and reattach only its jobs.

Please add focused regression coverage for the memory conversion, actual fallback context, CPU-only candidate handling, server A→B switching, and HF result selection/download.

two tiny nits:
- one npm test fail:

Image
I can also fix it after merge

- And the starter presets should be moved upwards again (I can also correct that after merge)

I am really positive about the overall direction, but I would not approve until these correctness issues are fixed.

pwilkin:
Fixes are up :)

fl0rianr (commented):
Thanks! — the five blockers from my previous review are now addressed well. The memory-unit fix, actual fallback-context tracking, CPU-only rejection, removal of the unrelated Hugging Face changes, and server-scope switching all look substantially better, including the added regression coverage.

I still found three correctness blockers:

1. **Memory accounting is not backend-specific.**
   The hardware snapshot collects AMD and NVIDIA devices, and `availableMib()` either uses unified memory or sums all GPU memory. That same total is then used for every backend candidate. On a mixed AMD APU/NVIDIA dGPU system, for example, CUDA may be fitted against AMD unified memory and ROCm may be fitted against NVIDIA VRAM. Please calculate available memory per backend and compatible device group rather than once for the whole machine.

2. **The benchmark does not test the configuration that is ultimately recommended.**
   `buildBenchRecipe()` receives only the primary fit and benchmark loads contain the KV-cache flags, but not fit-derived placement flags such as `-ngl`, `--n-cpu-moe`, or `--cpu-moe`. Those are added only during final synthesis. Load success, VRAM, TPS, and TTFT can therefore describe a different configuration from the preset that is saved or applied. Please build the complete candidate configuration per backend first and benchmark those exact options.

3. **A backend that fails the deep-context run can still win the backend duel.**
   `representative()` uses the deep result when available but falls back to depth 0 when it is not. This can compare one backend at ~30K depth against another backend’s much easier depth-0 result and reward the backend that failed the target workload. Compare candidates only at a common successfully measured depth, or disqualify candidates that fail the target-context run.

A smaller follow-up: the dual-CUDA `--split-mode tensor` recommendation is also added after benchmarking, so its reported performance does not measure the final configuration.

The implementation is now very close, but I would still hold approval until these three consistency issues are resolved.

pwilkin:
Aight, updates are up.

fl0rianr (commented):
Great, so near now! But, I still found two correctness blockers:

1. Benchmark and apply use different argument-merging semantics.
Every benchmark load sets merge_args: false, while the saved and temporary recommendations omit merge_args, so the server default of true applies later. The benchmark therefore suppresses model, architecture, and global llamacpp_args, but those arguments are merged back when the recommendation is applied. Load success and performance can consequently describe a different command line from the saved preset. Please run the benchmark through the same normal merge path as the final load and add a regression test with existing model or architecture arguments.
2. The backend duel can still compare different deep workloads.
Deep prompt depth is derived independently from each backend’s recommended context. One candidate may therefore be measured at roughly 6.5K tokens while another is measured at 30K, and pickBackend() compares those deep TTFT/TPS values directly. Both runs may succeed, but the smaller-context backend is rewarded for a much easier workload. Please compare candidates at a common depth, or use the deep runs only as eligibility checks and perform the actual duel using a common workload.

The implementation is very close, thanks!!

pwilkin:
Okay, hopefully this closes it out :)

fl0rianr (approved):
Thanks! Great!

pwilkin:
Now just the backend that enables it :)

PR #2698, opened 2026-07-14
[Router] llm router classifier + routing.router desugaring (#2405)

### LINKED ISSUE #2405: [Router] llm router classifier + routing.router desugaring (L0a)

ao-batouche:
Closes #2405

## Summary

Implements the L0(a) "LLM-as-router" on-ramp: a new `llm` classifier type plus load-time desugaring of `routing.router` into the explicit core form, keeping the one-engine invariant — everything still compiles down to classifiers + rules + a Decision.

## Changes

**`LlmClassifier`** (`src/cpp/server/routing_policy.cpp`)

- Runs the router model through `ClassifierServices::chat`, composing the author's prompt with a fixed response contract (the candidate list + required JSON shape).
- Requires a **structured reply** — a single JSON object `{"model": "<exact candidate>", "rationale": "<short reason>"}` with BOTH fields present; `rationale` must be a non-empty string. One ```` ``` ```` fence pair is tolerated, but the closing fence must end the reply (trailing prose rejects). The chosen candidate becomes the label (score 1.0); the model's stated `rationale` is recorded in `Score::rationale` for the trace.
- Malformed JSON, a missing/unknown `model`, or surrounding prose yields an empty Score (`ok=true`, no labels) → no identity rule fires → the engine fails open to `default_model`. A backend failure yields `ok=false` so the owning condition applies `on_error`.
- Registered in `make_classifier`.

**`routing.router` desugaring** (`src/cpp/server/routing_policy_parser.cpp`)

- `{"router": {"type":"llm", "model", "prompt"}}` expands at load time into one synthesized `llm` classifier (`__router`, labels = the **authored** candidate names) plus one identity rule per candidate (label X → route_to X). `route_to` is canonicalized downstream by the normal rule parser; labels stay the authored names the router LLM actually replies with.
- Mutually exclusive with explicit `routing.classifiers` / `routing.rules`.

**Constrained router invocation** (`src/cpp/server/routing_classifier_services.cpp`, new `thinking_controls.{h,cpp}`)

- The internal chat call now disables thinking through the same cross-backend normalization used by normal Lemonade requests (`enable_thinking:false` → `/no_think` injection), sets `temperature:0`, `stream:false`, and a tight `max_tokens` cap. This prevents a reasoning-enabled router (e.g. Qwen3-1.7B) from emitting `<think>` blocks or spending unbounded time before naming a model.
- The `should_disable_thinking` / `prepend_no_think` / `strip_handled_thinking_fields` helpers were extracted from `server.cpp` into a shared `thinking_controls` unit so the HTTP path and the classifier path share one implementation.

**Trace** (`src/cpp/include/lemon/routing_policy.h`, `route_decision_response.cpp`, `decision.schema.json`)

- `TraceEntry` gains optional `label` and `rationale` fields, threaded through the classifier-band condition and serialized on the decision response, so each band entry identifies which candidate its score belongs to and (for the llm router) why.

**Schema** (`route_policy.schema.json`, `decision.schema.json`, `schema-lock.json`)

- `routing.router` is now mutually exclusive with `classifiers`/`rules` at the schema level (matching the parser), and authored `llm` classifiers require `labels` (`minItems: 1`, matching the runtime).
- `decision.schema.json`'s `trace_entry` declares the new `label`/`rationale` properties (it uses `additionalProperties: false`). Schema lock refreshed in the same change (both majors are `released: false`).

## Out of scope

Model-slot semantics for classifier/router models are deliberately **not** part of this PR, per maintainer scoping — see #2725. Under the default `max_loaded_models=1`, the router LLM and the selected candidate currently share the single generation slot, so consecutive L0a requests reload both models; improving that lifecycle is #2725's agenda.

## Tests

- `test/cpp/test_routing_policy_llm_router.cpp` — structured-choice parsing, prompt-contract composition, fenced-JSON tolerance, and rejection of malformed / multi / negated / unknown / missing-field replies (all fail open to `default_model`); desugaring shape; engine-level e2e with trace label + rationale.
- `test/cpp/test_routing_classifier_services.cpp` — asserts the constrained invocation (thinking disabled via `/no_think`, handled fields stripped, tight `max_tokens`, `temperature:0`, `stream:false`).
- `test/cpp/test_routing_policy_parser.cpp` — desugaring + non-identity-resolver canonicalization (authored labels vs canonical `route_to`), plus router+rules rejection.
- `test/cpp/test_route_decision_response.cpp` — trace `label`/`rationale` serialized when set, omitted when empty.
- `test/server_endpoints.py::test_021zj_router_llm_l0a_live` — the **live** end-to-end path: an L0a `routing.router` collection routed through `Router::chat_completion` with a real router model and a distinct candidate, asserting the completion comes from the selected candidate and the trace carries the chosen label plus a non-empty rationale.

## Testing

- C++: `RoutingPolicyLlmRouterTest`, `RoutingClassifierServicesTest`, `RoutingPolicyParserTest`, `RouteDecisionResponseTest` pass; `RoutingPolicyEvaluatorTest`, `RoutingPolicySemanticTest`, `RoutingPolicyContractTest` unaffected.
- Live: `test_021zj_router_llm_l0a_live` (requires two small local models).


fl0rianr (changes requested):
Thanks for the work on this. The overall direction looks right: keeping routing.router as load-time sugar and compiling it into the existing classifier + rule engine preserves the one-engine design.

I found several issues that should be addressed before merge:

### 1. Candidate-name canonicalization breaks the normal runtime path

`parse_candidates()` resolves authored candidate names through the component resolver before `desugar_routing_router()` is called. The synthesized `llm` classifier therefore receives canonical IDs such as `builtin.Qwen3-8B-GGUF` as its labels.

However, the authored prompt still refers to the public model names, and the router LLM will normally reply with those names, for example `Qwen3-8B-GGUF`. `LlmClassifier::match_label()` then compares that reply against the canonical label `builtin.Qwen3-8B-GGUF`, finds no match, and falls through to `default_model`.

The new tests do not catch this because they use an identity resolver.

Please preserve the authored candidate names as the classifier output labels and only resolve the generated `route_to` values to canonical component IDs. A regression test should use a non-identity resolver, for example:

```text
Qwen3-8B-GGUF -> builtin.Qwen3-8B-GGUF
```

while the fake chat reply remains `Qwen3-8B-GGUF`.

### 2. The existing parser test is now stale and should fail

`test_routing_policy_parser.cpp` still asserts that the L0a fixture is rejected because `routing.router` desugaring is not implemented:

```cpp
check("routing.router is recognized but deferred to #2405",
      throws_with(router_sugar, "routing.router desugaring"));
```

After this PR, that document parses successfully, so `RoutingPolicyParserTest` should fail. Please replace this old deferred-behavior assertion with positive desugaring and canonicalization coverage and run the full related CTest set, not only the new test binary.

### 3. rationale is serialized but missing from decision.schema.json

The PR adds `rationale` to serialized trace entries, while `decision.schema.json` currently defines trace entries with `additionalProperties: false` and does not include that property.

Any response containing a rationale is therefore invalid according to the current decision schema.

Please update:

* `decision.schema.json`
* the schema lock hash
* serializer/schema tests for a trace containing `rationale`

### 4. Substring matching can turn an invalid or contradictory answer into a valid route

The current fallback matching accepts any candidate name appearing anywhere in the model output and chooses the longest matching label.

Examples that would route incorrectly:

```text
Do not use Qwen3.5-35B-A3B-GGUF. Use gpt-4o instead.
```

```text
I considered Qwen3.5-35B-A3B-GGUF, but choose Qwen3-8B-GGUF.
```

In both cases, the longest-substring rule can select the 35B model even though it was not the final choice. This also weakens the required invalid-choice fallback.

The safest v1 behavior would be to accept only an exact trimmed candidate name, which already matches the example prompt requiring `ONLY a model name`. A structured JSON response would be even more robust, but exact matching would be a minimal and predictable fix.

Please add tests for:

* multiple candidate names in one reply
* negated candidate names
* unknown choices
* prose surrounding a candidate

### 5. Parser semantics and route-policy schema are not fully aligned

The parser rejects `routing.router` combined with explicit `routing.rules` or `routing.classifiers`, but the schema currently uses `anyOf`, which still allows `router` and `rules` together.

Also, the runtime implementation requires an `llm` classifier to have at least one label, while the schema only requires `model` and `prompt` for `type: "llm"`.

Please make the schema express the same constraints as the parser and classifier implementation.

### 6. The trace does not explicitly identify the tested label

The acceptance criteria say that the trace should show the chosen candidate plus rationale. At the moment, the trace only contains the classifier ID, score, result, and rationale. Because the synthesized rules evaluate the same classifier once per candidate, it is not possible to tell which label a particular score belongs to without inferring it from rule order or from the rationale text.

Please consider adding an optional `label` field to `TraceEntry` and the decision schema. That would make the trace directly auditable:

```json
{
  "condition": "classifier:__router",
  "label": "Qwen3-8B-GGUF",
  "score": 1.0,
  "result": true,
  "rationale": "Simple conversational request"
}
```

The fake-service unit coverage is useful, but the live Router path required by the issue acceptance criteria is still not covered. It is reasonable to coordinate that with the `ClassifierServices` integration work, but the PR description should clearly distinguish the fake engine-level test from a live `Router::chat_completion` end-to-end test.

Overall, the architecture is promising, but the canonical-name mismatch, stale existing test, and schema incompatibility are blocking correctness issues.

ao-batouche:
Thanks for the thorough review @fl0rianr , all 6 points addressed in [f305986](https://github.com/lemonade-sdk/lemonade/pull/2698/commits/f30598661a810caa92b57b09bb6b3e54bc79c8e1):

1. **Canonicalization**: desugar now emits **authored** candidate names for labels / `match.label` / `route_to`; `route_to` is canonicalized downstream by `parse_rules` like any hand-written rule. Regression test added with a non-identity resolver (`X → builtin.X`, fake reply `Qwen3-8B-GGUF`) it fails on the previous code.
2. **Stale test**: replaced the deferred-behavior assertion with positive desugaring + canonicalization coverage (incl. e2e + trace label) and a `router`+`rules` rejection check; the full routing CTest set is green, not just the new binary.
3. **`decision.schema.json`**: `trace_entry` now declares `label` and `rationale`; schema-lock refreshed via `--update` in the same change; serializer test added (emit when set, omit when empty).
4. **Matching**: exact trimmed candidate name only; substring fallback removed. Tests added for prose, multiple names, negated names, and unknown choices, all fall open to `default_model`.
5. **Schema alignment**: `routing.router` is now mutually exclusive with `classifiers`/`rules` at the schema level, and authored `llm` classifiers require `labels` (`minItems: 1`), matching the parser/runtime.
6. **Trace label**: added optional `label` to `TraceEntry` + serializer + decision schema; each band entry now identifies which candidate its score belongs to.

On the live-path note: the PR description now states explicitly that current coverage is engine-level against `FakeClassifierServices.chat`, and the live `Router::chat_completion` e2e will follow #2384's harness.

fl0rianr (changes requested):
Thanks for addressing the previous review. I re-reviewed the current head, and the original six findings are now handled well: authored-vs-canonical names, exact matching, the stale parser test, trace fields, schema parity, and the associated regression coverage all look substantially improved.

I still see two blocking runtime issues and one unresolved acceptance gap.

### 1. The default model-slot configuration causes the router and selected model to evict each other on every request

Lemonade currently defaults to `max_loaded_models: 1`. The live L0a flow first loads and invokes the router LLM through `ClassifierServices::chat`, then loads the selected candidate for the actual completion.

Because both are LLMs, loading the candidate evicts the router model under the normal per-type LRU limit. On the next request, loading the router evicts the candidate, and loading the candidate then evicts the router again.

Under the default configuration, this therefore becomes approximately:

1. Load router model.
2. Route request.
3. Evict router and load selected model.
4. Generate response.
5. Repeat both loads for the next request.

This is not merely a small optimization issue: backend startup can take tens of seconds, so the normal L0a path becomes effectively unusable. Pinning also creates failure cases: a pinned candidate can prevent the router model from loading, while a pinned router can prevent the selected candidate from loading.

Please define explicit slot/lifetime semantics for classifier models and add an integration test using the default `max_loaded_models=1`. Possible approaches include a separate classifier slot category, an ephemeral invocation path that does not participate in normal LLM slot accounting, or an explicit validated requirement for additional slots. Simply increasing the global default would likely be too broad.

### 2. The router-model invocation does not disable thinking or bound its output

The internal `ClassifierServices::chat` request currently sets `stream:false` and `temperature:0.0`, but it does not disable thinking and does not set a small `max_tokens`/`max_completion_tokens` limit.

This matters especially because the issue's example uses Qwen3-1.7B, whose thinking mode is enabled by default. Depending on the backend's reasoning format, the returned content may contain a `<think>...</think>` block or other additional output. The new exact matcher then rejects the otherwise valid model choice and silently falls back to `default_model`. Even when reasoning is separated into another response field, the router can still spend substantial time and tokens reasoning before emitting one model name.

Please make this a deliberately constrained classifier invocation:

* disable thinking through the same cross-backend normalization used by normal Lemonade requests;
* set a tight output limit;
* add an adapter test asserting that these controls are present.

A live test with the example Qwen3-style router would also catch this class of failure.

### 3. The implementation still does not provide the structured choice + rationale required by the issue

A valid response is now required to be exactly one candidate name, and that same string is copied into `Score::rationale`. The tests therefore assert that the rationale is simply the selected model name.

That gives the trace a field called `rationale`, but not an actual reason. The issue currently asks for both a structured choice and a short rationale, while its example prompt asks for only a model name. Those requirements need to be reconciled explicitly.

Please either:

* parse a strict structure such as `{"model":"...","rationale":"..."}`, rejecting malformed or unknown choices; or
* revise the contract and schema to describe this as the raw reply rather than a rationale.

The issue also explicitly requires one live end-to-end path. The current “e2e” coverage still constructs `RoutingPolicyEngine` with `FakeClassifierServices`, so it does not exercise model loading, slot behavior, the real chat adapter, reasoning output, or `Router::chat_completion`.

The PR description should also be updated: it still describes longest-substring matching and says no schema change is needed, both of which are no longer true after the review fixes.

The previous review findings are resolved, but the default slot behavior and unconstrained router invocation should be addressed before approval.

ao-batouche:
Thanks, this round was very useful. all three items are addressed, plus the PR description is updated.

**1. Slot eviction under `max_loaded_models=1` (blocking).** Added an explicit classifier-slot category. Router/classifier models load into a slot pool that is counted and LRU-evicted independently from generation models of the same `ModelType` (`WrappedServer::classifier_slot_` → `Router::load_model(..., classifier_slot)` → partitioned `count_servers_by_type` / `find_lru_server_by_type`). Only the routing-dispatch `ensure_loaded` site sets it, so the router LLM and the selected candidate no longer evict each other, and a pin in one pool can't starve loads in the other. Covered by a live integration test (below) that runs under the default `max_loaded_models=1`.

**2. Unconstrained router invocation (blocking).** The internal chat call now disables thinking through the same cross-backend normalization as normal requests (`enable_thinking:false` → `/no_think` injection, then strip handled fields) and bounds output with a tight `max_tokens`, alongside the existing `temperature:0` / `stream:false`. I extracted `should_disable_thinking` / `prepend_no_think_to_last_user_message` / `strip_handled_thinking_fields` from `server.cpp` into a shared `thinking_controls` unit so the HTTP path and the classifier path share one implementation. An adapter test asserts every control is present on the request handed to `Router::chat_completion`.

**3. Structured choice + rationale.** Reconciled toward the structured option: `LlmClassifier` composes the author prompt with the candidate list + a required JSON contract, and parses a strict `{"model": "...", "rationale": "..."}` object (one fenced block tolerated). The chosen candidate is the label; the model's stated reason is the rationale (no longer just the model name). Malformed JSON, unknown `model`, missing field, or surrounding prose all fail open to `default_model`. The fixture prompt no longer says "reply with ONLY a model name."

**Live end-to-end path.** `test_021zj_router_llm_l0a_live_and_slot_coexistence` registers an L0a `routing.router` collection whose router model and candidate are distinct, routes a real request through `Router::chat_completion`, and asserts (a) the completion is produced by the selected candidate, (b) the trace carries the chosen label + a rationale string, and (c) under `max_loaded_models=1` both the router LLM and the candidate are simultaneously resident — the coexistence guarantee that thrashed before the slot fix. The PR description now clearly separates the fake engine-level tests from this live test.

**PR description** updated: removed the stale longest-substring description and the "no schema change needed" line (schema now expresses the router/rules exclusivity and the `llm` labels requirement, and the decision schema declares the new trace fields; lock refreshed).

fl0rianr (changes requested):
Thanks again for the substantial follow-up. The structured response contract, constrained classifier invocation, live test, and separate classifier-slot concept are all meaningful improvements.

After reviewing the current head again, I still found several correctness issues in the new slot lifecycle.

### 1. classifier slot is lost on both model recovery paths

The normal first-load path stores `classifier_slot` on `new_server`, but neither recovery path preserves it:

* `reload_model_after_watchdog_reset()` remembers only `pinned` and calls `load_model()` without the `classifier_slot` argument, so it defaults to `false`.
* The nuclear load retry creates `retry_server`, sets its metadata and pinned state, but never calls `set_classifier_slot(classifier_slot)`.

A router model that experiences a watchdog reset or succeeds only on the second backend-load attempt therefore re-enters the generation pool. Under `max_loaded_models=1`, this restores the original router/candidate eviction problem and can also make recovery fail when the generation candidate is pinned.

Please preserve the role exactly like the pinned state and add regression tests for:

* a classifier-slot model succeeding on the retry load;
* a classifier-slot model recovering from a watchdog reset.

### 2. Slot role currently depends on whichever use loaded the model first

`auto_load_model_if_needed(..., classifier_slot=true)` returns immediately when the model is already loaded, without checking its current slot role. `find_server_by_model_name()` also ignores the role entirely.

This causes two problematic cases:

* If the router model was previously loaded as a normal generation model, routing continues using it from the generation pool. A pinned router can then prevent the selected candidate from loading under the default limit.
* If a model was first loaded as a classifier and is later requested directly for chat, it remains in the classifier pool and bypasses normal generation-slot accounting.

The live test starts by unloading everything, so it does not cover this lifecycle.

Please define model identity and migration semantics explicitly. A server may need to be identified by `(model, role)`, support both roles with correct accounting, or be migrated between pools with capacity and pin checks. A sticky first-load boolean is not sufficient.

### 3. The public capacity and pinning contract no longer represents reality

The implementation now permits one generation LLM and one classifier LLM when `max_loaded_models=1`, but `/health.max_models.llm` still reports only `1`.

At the same time:

* `pinned_models.llm` aggregates pinned models from both pools;
* loaded-model entries do not expose their slot role;
* the public documentation still describes one independent LRU cache per model type.

This can make CLI/client capacity checks incorrect and makes the new behavior difficult to diagnose.

Please expose classifier capacity and role explicitly, or keep the existing public fields generation-only and add separate classifier-pool fields. Loaded model metadata should also identify the slot role.

### 4. The structured reply parser does not enforce its stated contract

The composed prompt requires both `model` and `rationale`, but the parser only requires `model`. Missing, empty, or non-string rationales still produce a valid route without the auditable rationale required by the issue.

The fenced-JSON tolerance also ignores any text following the closing fence, so a fenced JSON object plus trailing prose is accepted despite the “no other text” contract.

Please require a non-empty string rationale and ensure a tolerated closing fence is the end of the trimmed response. Tests should cover missing/empty/wrong-type rationale and trailing prose after a fenced object.

### 5. Classifier/generation isolation is not universal across backend slot policies

The normal LRU path partitions by `classifier_slot`, but NPU exclusivity still evicts every NPU model, and `CoexistByType` still selects an existing model solely by `ModelType`.

Therefore an FLM or exclusive-NPU router/candidate pair can still evict each other. This may be an unavoidable hardware constraint, but then the guarantee must be explicitly scoped to compatible standard CPU/GPU backends, and incompatible placement should produce a clear diagnostic rather than contradicting the advertised pool isolation.

The live integration test is valuable, but additional coverage is needed for preloaded/pinned routers, recovery paths, and non-standard slot policies before this new global Router lifecycle is safe to merge.

eddierichter-amd (dismissed):
Thanks for the substantial work here. The overall L0a design now looks much stronger: load-time `routing.router` desugaring, authored-name candidate matching, structured responses, constrained router invocation, trace label/rationale fields, schema alignment, and the live end-to-end test are all meaningful  improvements.

  I reviewed the current head and independently confirmed Florian’s latest requested-changes review. 

  In particular:

  - The classifier role is still lost in both recovery paths. `reload_model_after_watchdog_reset()` reloads without `classifier_slot`, and the nuclear retry never sets it on `retry_server`.
  - Slot role remains dependent on how the model was first loaded. The already-loaded paths look up only by model name and return without reconciling the requested classifier/generation role.
  - `/health` still reports the old per-type capacity while classifier and generation models now occupy separate pools, and loaded-model metadata does not expose which pool a model belongs to.
  - The structured reply parser still accepts a valid `model` without a non-empty string `rationale`, and fenced JSON followed by trailing prose is accepted because everything after the closing fence is ignored.
  - The classifier/generation partition is applied in the standard LRU path, but `ExclusiveNpu` and `CoexistByType` continue to evict/select models without considering `classifier_slot`.

  The focused happy-path tests are green on my checkout:

  ```text
  RoutingPolicyLlmRouterTest
  RoutingClassifierServicesTest
  RoutingPolicyParserTest
  RouteDecisionResponseTest

ao-batouche:
Thank you @fl0rianr and @eddierichter-amd , I will work on these tomorrow 👌

SlawomirNowaczyk:
To be honest, I don't think "model-slot configuration" changes belong in this PR?

Yes, after adding router functionality, the current way of handling `max_loaded_models` is problematic, and we need to have a discussion about how to adapt it -- but this is not the right place for it. I've created an issue to discuss it, check https://github.com/lemonade-sdk/lemonade/issues/2725

ao-batouche:
Thanks all, and @SlawomirNowaczyk, agreed on the scoping. 

I addressed in this commit:

**Slot lifecycle removed from this PR (per #2725).** All classifier-slot changes are reverted: `WrappedServer::classifier_slot_`, the `Router::load_model` / `count_servers_by_type` / `find_lru_server_by_type` signature changes, and the `auto_load_model_if_needed` flag. The Router's slot accounting is back to exactly what's on main. That resolves review points 1, 2, 3, and 5 by taking the slot semantics out of scope, I'll join the discussion in #2725, where the recovery-path, role-migration, capacity-reporting, and backend-slot-policy questions raised here are the right agenda.

One consequence, stated plainly: under the default `max_loaded_models=1`, the router LLM and the selected candidate share the single generation slot, so consecutive L0a requests reload both models. The live test's docstring documents this and points at #2725.

**Point 4 (structured contract) fixed.** The parser now enforces exactly what the composed prompt promises:
- `rationale` is required,  it must be a string and non-empty after trimming; a valid `model` with a missing/empty/whitespace/non-string rationale rejects (fail-open to `default_model`).
- A tolerated closing ``` fence must end the trimmed reply; fenced JSON followed by trailing prose rejects, and an opening fence without a closing fence rejects.

new tests cover: missing rationale, empty rationale, whitespace-only rationale, non-string rationale, fenced JSON + trailing prose, and unterminated fence ( all falling open to `default_model`). `RoutingPolicyLlmRouterTest` is 27/27.

**Live test rescoped.** `test_021zj_router_llm_l0a_live` keeps the live end-to-end path (real router model through `Router::chat_completion`, completion from the selected candidate, trace with label + non-empty rationale) and drops the slot-coexistence assertions, which belonged to the reverted change.

**Kept from round 2** (unaffected by the scope trim): the constrained router invocation (shared `thinking_controls` normalization + `max_tokens` cap, with the adapter test), the structured `{model, rationale}` contract, authored-name matching, trace `label`/`rationale`, and the schema alignment.

The PR description is updated to match (slot section removed, deferred-to-#2725 note added).

fl0rianr (changes requested):
Thanks for the latest revision. The previous implementation findings now look substantially better: authored-vs-canonical candidate names are handled correctly, the structured `{model, rationale}` response is enforced, fenced trailing prose is rejected, the router invocation is constrained, the schemas have been updated, and there is now a real Router-backed integration test.

I also fully agree with the @SlawomirNowaczyk decision to move classifier model-slot semantics to #2725, so I am not reopening that design as a blocker in this PR.

I still found two correctness issues in the current implementation.

### 1. The LLM router receives only the latest user text, not the actual routing context

`build_route_context()` extracts the latest user text into `RouteContext::input`, while request features such as `has_tools`, `has_images`, `chars`, and metadata are stored in separate fields.

`LlmClassifier::evaluate()`, however, calls:

```cpp
ctx.services.chat(model_, effective_prompt(), ctx.request.input);
```

Therefore the router LLM never sees those other fields or the preceding conversation history.

Concrete consequences include:

* an image-only request appears as an empty request;
* tool availability cannot influence the LLM route;
* a “long context” routing policy cannot detect long conversation history because only the latest user turn is passed;
* request metadata is invisible to the router.

Because `routing.router` is mutually exclusive with explicit classifiers and rules, an L0a policy cannot compensate by adding deterministic `has_images` or `has_tools` rules around the LLM router.

Please pass a structured representation of the routing context to the LLM, or extend the chat service seam to receive `RouteContext`. At minimum it should include the text, `has_tools`, `has_images`, `chars`, and relevant metadata, with an explicit policy for how much conversation history is supplied.

Regression coverage should include an image-only request, a tool-bearing request without a textual tool hint, and a multi-turn request where earlier context affects the expected route.

### 2. The selected rationale is copied onto rejected candidate trace entries

The desugared policy evaluates one classifier-band condition per candidate. The classifier itself is memoized and returns one selected label plus one rationale.

`ClassifierBandCondition::evaluate()` currently attaches `score.rationale` to every evaluated band entry whenever the score is valid, regardless of which label was selected or whether that condition matched.

When the second candidate is selected, the trace can therefore contain:

```json
[
  {
    "label": "candidate-a",
    "score": 0.0,
    "result": false,
    "rationale": "Candidate B is best because..."
  },
  {
    "label": "candidate-b",
    "score": 1.0,
    "result": true,
    "rationale": "Candidate B is best because..."
  }
]
```

This incorrectly associates Candidate B's rationale with Candidate A's rejected score and makes the audit trace misleading.

Please emit the rationale only on the band corresponding to the selected label, or represent the one-time classifier result as its own trace entry containing `chosen_label` and `rationale`.

A regression test should select the second candidate and assert that earlier false candidate entries do not carry the rationale.

### Contract cleanup

These are smaller issues, but should also be cleaned up:

* `route_policy.schema.json` permits empty `router.model`, `router.prompt`, and empty strings inside `labels`, while the parser/runtime reject them. Please add `minLength: 1`.
* The issue example and the inline unit-test policy still say “Reply with ONLY a model name”, which conflicts with the now-required JSON object and rationale. The author prompt should describe routing intent only; the implementation-owned suffix should define the output protocol.

The live test is valuable, but it relies on `Tiny-Test-Model-GGUF` producing strict valid JSON. Since the current GitHub Actions runs have not executed any jobs, I would also want confirmation that the live test has run reliably on the intended CI backend.

With the context-loss and misleading-trace issues addressed, the implementation itself would be in much better shape. The known slot thrashing remains tracked separately in #2725.

ao-batouche:
Thank you @fl0rianr, both catches were addressed now.

**1. Routing context now reaches the router LLM.** `LlmClassifier` no longer sends bare `ctx.request.input`; it sends a structured JSON payload of the full v1 routing context: `{"text", "has_tools", "has_images", "chars", "metadata"}`, and the composed prompt suffix describes that input format to the model. An image-only request is now visible as `has_images:true` with empty text, tool availability is visible without any textual hint, and caller `metadata` hints reach the router verbatim

On conversation history, the explicit policy is: **`text` is the latest user turn only.** That's not a choice this classifier makes — it's the frozen v1 `RouteContext` contract (`build_route_context` deliberately extracts a single turn "so earlier assistant/developer/system turns can't skew routing", and `chars` is defined as the byte count of that turn). Supplying more history to the llm router means extending that contract for every classifier and condition, which after the #2725 scoping lesson I did not want to smuggle into this PR . The policy is stated in the code comment, in the prompt suffix the model sees, and pinned by a new multi-turn test.

Regression coverage added: image-only request (visible + still routes), tool-bearing request with no textual hint, metadata pass-through, and a multi-turn request asserting the latest-turn-only policy (`test_build_route_context_multi_turn_latest_user_turn_only`).

**2. Rationale no longer smears onto rejected candidates.** `ClassifierBandCondition` attaches the rationale only when the band's tested label is one the classifier actually scored (for the llm router: the single selected candidate); label-less bands keep the previous behavior. Regression test: the e2e case selects the second candidate and asserts the first (rejected, `result:false`) `__router` entry carries no rationale while the winning entry does.

**Contract cleanup.**
- `route_policy.schema.json`: `minLength: 1` on `router.model`, `router.prompt`, and `labels` items; schema-lock refreshed.
- The inline unit-test policy no longer says "Reply with ONLY a model name", author prompts now describe routing intent only; the implementation-owned suffix defines the output protocol. (the example in issue #2405's description says the same stale line ! that's issue text I can't edit maybe @ramkrishna2910 can ?)
- Live test: swapped the roles so the more capable model (`Phi-4-mini-instruct-GGUF`) is the router and the tiny model is the candidate, since the router must emit strict JSON.

fl0rianr (commented):
Thanks for the latest update. I re-reviewed the current head, and the implementation issues from the previous rounds now look resolved.

I found one remaining merge-readiness issue: `RoutingPolicyLlmRouterTest` is registered in CMake, but the regular Linux and macOS C++ unit-test steps neither build `test_routing_policy_llm_router` nor include `RoutingPolicyLlmRouterTest` in their `ctest -R` expressions.

That means the dedicated test covering structured context, strict response parsing, rationale attribution, desugaring, and invalid-choice fallback will not run in the normal C++ CI even after the workflows are authorized.

Please add:

```bash
cmake --build --preset default --target test_routing_policy_llm_router
```

and include `LlmRouter` in the routing CTest expression, for example:

```text
RoutingPolicy(Contract|Evaluator|Registry|Semantic|LlmRouter|Deterministic|Engine|Parser|Store)Test
```

Once that test is wired into the regular workflow, I do not see another blocker within this PR's agreed scope.
Good work, really close now!

ao-batouche:
Done.

test_routing_policy_llm_router is now built and LlmRouter is included in the CTest expression in both the Linux and macOS C++ test steps, exactly as suggested. 

Thanks for the careful reviews.

fl0rianr (commented):
Good work! Was just right before approving then the CI errors came in.

The remaining CI failures appear to be caused by the shared `Tiny-Test-Model-GGUF` test-model contract rather than by the LLM-router implementation itself.

The new L0a path reaches the expected route and successfully generates a completion from the selected candidate. However, `ENDPOINT_TEST_MODEL` is reused throughout the endpoint suite, and `Tiny-Test-Model-GGUF` now resolves to:

```text
unsloth/gemma-3-270m-it-GGUF:
gemma-3-270m-it-UD-IQ2_M.gguf
```

There are two resulting problems visible in the CI server logs:

1. Some fixtures and collection manifests still expect the previous SmolLM2 checkpoint under the same `Tiny-Test-Model-GGUF` alias, producing manifest-versus-local-definition drift.
2. The current Gemma test model is loaded with a 3072-token context in CI. At least one existing endpoint test serializes a large tool payload of roughly 6000 tokens, so llama.cpp correctly rejects it as exceeding the available context.

I would not remove or weaken the new router live test. Instead, the test-model contract should be made stable.

### Preferred fix

Restore `Tiny-Test-Model-GGUF` to the checkpoint that the existing fixtures expect, and introduce a separate model name if Gemma 3 270M is intended to become another CI model.

For example:

```text
Tiny-Test-Model-GGUF
    -> existing stable SmolLM2 test checkpoint

Tiny-Gemma-Test-Model-GGUF
    -> gemma-3-270m-it-UD-IQ2_M.gguf
```

This prevents an implementation change behind a widely reused test alias from silently changing context size, chat-template behavior, and output characteristics across the entire endpoint suite.

### Alternative if Gemma should remain the Tiny model

If the Gemma checkpoint is intentionally replacing SmolLM2:

* update every fixture and embedded collection manifest to use the same Gemma checkpoint;
* give context-heavy endpoint tests an explicit context size of at least 8192 tokens; or
* introduce a separate `CONTEXT_TEST_MODEL` for large tool-schema and long-context tests while keeping Tiny for short smoke/inference checks.

The important part is not to use one global “tiny” model for both minimal smoke tests and requests that intentionally contain several thousand prompt tokens.

After aligning the alias/fixtures and the context-heavy test model, please rerun:

* the Linux and macOS endpoint jobs;
* the ARM endpoint job;
* the Arch and openSUSE endpoint jobs.

The dedicated `RoutingPolicyLlmRouterTest` and the router implementation itself do not need to be changed for these failures.

ao-batouche:
Thanks @fl0rianr It should be fixed now.

I went with the preferred fix: `Tiny-Test-Model-GGUF` is restored to the `SmolLM2` checkpoint the existing fixtures and `test_models.py` contract expect, and the Gemma checkpoint now lives under its own `Tiny-Gemma-Test-Model-GGUF `alias (registry + backends-reference doc updated; one stale comment in `test_ollama.py `fixed). 

I traced the switch to a commit on main that changed the checkpoint behind the shared alias, so this also un-drifts the suite for other PRs. Router implementation and `RoutingPolicyLlmRouterTest` untouched. 

could you (or another maintainer) re-authorize the workflows to rerun the Linux/macOS, ARM, Arch and openSUSE endpoint jobs?

fl0rianr (commented):
The latest `8e540437` change is not the right fix and should be reverted.
I could have made a better comment...

`Tiny-Test-Model-GGUF` is a shared repository-wide CI contract. Changing its checkpoint in this router PR affects unrelated CLI, endpoint, download, Ollama, and documentation tests. The new head already demonstrates that blast radius:

* macOS now fails in the CLI step before reaching the endpoint test;
* Arch and openSUSE also fail in their CLI steps;
* `backend-docs-drift` now fails as well.

More importantly, changing Tiny cannot fix the original L0a test failure.

In `test_021zj_router_llm_l0a_live`:

```python
router_model = SECOND_TEST_MODEL_EVICTION  # Phi-4-mini
candidate_model = ENDPOINT_TEST_MODEL      # Tiny-Test-Model-GGUF
```

Tiny is only the final generation candidate. It does not produce or parse the routing decision. The probabilistic part is Phi producing the exact `{model, rationale}` protocol. If that reply is rejected, the engine falls open to `default_model`, which is also Tiny, so the request still completes successfully but there is no winning `classifier:__router` trace entry. Changing the candidate checkpoint does not address that.

Please:

1. Revert the `Tiny-Test-Model-GGUF` registry/docs changes from `8e540437`.
2. Keep the existing Tiny alias unchanged.
3. Make only the new L0a live test deterministic.

The existing `_start_mock_cloud_provider()` helper in `server_endpoints.py` is the appropriate pattern here. Use a uniquely named mock cloud model as the **router component** and return a fixed valid assistant response:

```json
{
  "model": "Tiny-Test-Model-GGUF",
  "rationale": "The configured candidate handles this request."
}
```

Then keep `Tiny-Test-Model-GGUF` as the real local candidate for the final completion.

That still exercises the complete production path:

```text
collection.router
  -> LlmClassifier
  -> ClassifierServices::chat
  -> Router::chat_completion
  -> CloudServer
  -> strict structured-reply parser
  -> rule evaluation and trace
  -> real candidate auto-load
  -> candidate completion
```

The mock handler can also capture the forwarded router request and assert the live adapter constraints:

```python
self.assertFalse(router_request["stream"])
self.assertEqual(router_request["temperature"], 0.0)
self.assertLessEqual(router_request["max_tokens"], 256)
self.assertTrue(
    router_request["messages"][-1]["content"].startswith("/no_think\n")
)
```

This follows the existing cloud endpoint-test infrastructure already present in the same test class, removes dependence on model intelligence and platform-specific sampling, avoids downloading Phi-4-mini in every endpoint matrix job, and still verifies the actual Router-backed integration rather than `FakeClassifierServices`.

The exhaustive parser behavior remains covered by the C++ tests. The live test should validate wiring and backend integration, not whether a particular small LLM happens to obey a formatting prompt on every architecture.

Please do not solve this by changing a repository-wide test-model alias or by weakening the strict parser/fallback behavior.

To be save use also `python docs/tools/gen_backend_boilerplate.py` to see if your changes require a docs update as well (you can add those auto generated results to the commit).

ao-batouche:
you're right, I now reverted `8e540437`. 

what is done in this commit:

1. Reverted the `Tiny-Test-Model-GGUF` registry/docs changes; the Tiny alias is back to exactly what's on main
2. Made only `test_021zj_router_llm_l0a_live` deterministic, using the existing `_start_mock_cloud_provider` pattern. The router component is now a uniquely-named mock cloud model that returns a fixed valid reply (`{"model": "Tiny-Test-Model-GGUF", "rationale": "..."}`), while `Tiny-Test-Model-GGUF` stays the real local candidate for the final completion.

The test still exercises the full production path, `collection.router -> LlmClassifier -> ClassifierServices::chat -> Router::chat_completion -> CloudServer -> strict structured-reply parser -> rule evaluation and trace -> real candidate auto-load -> candidate completion`, with no dependence on a small LLM's formatting discipline and no Phi-4-mini download in the endpoint matrix. the mock handler captures the forwarded router request and asserts the live adapter constraints:

```
self.assertFalse(router_request["stream"])
self.assertEqual(router_request["temperature"], 0.0)
self.assertLessEqual(router_request["max_tokens"], 256)
self.assertTrue(router_request["messages"][-1]["content"].startswith("/no_think\n"))
```

and additionally that the message after the `/no_think` prefix is the structured routing-context payload (`text`, `has_tools`, `has_images`). The exhaustive parser/fallback behaviour remains covered by `RoutingPolicyLlmRouterTest`; this test validates wiring and backend integration only. The strict parser and fail-open behaviour are unchanged.

I ran the test locally against a fresh build and it passes. `python docs/tools/gen_backend_boilerplate.py` reports no drift after the revert, so no docs update is needed.

SlawomirNowaczyk (approved):
LGTM now

fl0rianr:
@OussamaBATOUCHE Thanks for the hard work. Due to the importance of your PR and the timely manner I did build upon your work and decided to make this one: https://github.com/lemonade-sdk/lemonade/pull/2729

We have two options:
1. You take my CI fixes from #2729 and we merge yours and I'll rebase

2. We merge mine after reviewe and you'll still be credited contributer but this PR will be closed.

ao-batouche:
@fl0rianr I'd like to go with option 1 if that still works for you.

keep #2698 as the PR that merges, and I'll pull in your CI fixes.

is that fine ?

fl0rianr:
> is that fine ?

Sure, wouldn't have it proposed otherwise. Go for it

ao-batouche:
@fl0rianr , looking at the branches, your #2729 has some refinements to the router code itself (the portable json_object response format, the llamacpp response fix, the parser tightening), not just CI, and they overlap with my two newest commits (the deterministic mock-cloud live test and the Gemma revert)

Which of your commits do you want on #2698? Or would it be cleaner for you to cherry-pick my two latest commits (1206947, 18e7183) onto #2729 and we go with Option 2 after all?

I just want to split it cleanly rather than create a third divergent state

fl0rianr:
@OussamaBATOUCHE If it's to complex grant maintainer access here and I will add a CI fix commit otherwise option 2.
Yes your reverted commits are in there since I worked on what I had at that time.

ao-batouche:
@fl0rianr sounds good, access now granted, please go ahead and push the CI fix commit.
Thanks for sorting this out 👌

fl0rianr (approved):
LGTM now as well.

Issue #2564, opened 2026-07-06
GUI3: Can't search HuggingFace for models, Custom Model button too prominent


kpoineal:
**Reported by:** Jeremy (project owner)

There's no way to search HuggingFace for a model on the Models page. The '+ Custom Model' button goes straight to the manual form, which should be much more hidden. Jeremy prefers the GUI2 UX where you can search HF and the manual form is secondary.

**Expected:** Primary action is searching HuggingFace for models (like GUI2). Manual custom model form is a secondary/advanced option.
**Actual:** '+ Custom Model' is the only visible path and goes directly to a manual form.

fl0rianr:
Bit of context, HuggingFace appears upon active search so on default no HuggingFace call is needed to just display your models list. Obviously not the ideal solution yet and Jeremy concerns are valid.

fl0rianr:
Custom Model button too prominent is resolved.

anditherobot:
Hi @kpoineal! Looks like PR #2700   (commit 32cdd7f8) in src/app/src/renderer/ModelManager.tsx fixed that  Should we close this issue?

fl0rianr:
Yeah, this one is not relevant anymore - thankfully.

PR #2700, opened 2026-07-14
Unify HF search UX and add detail panel for HF models (#2564)

kpoineal:
Unifies the HuggingFace search experience on the Models page and adds detail panel support for HF results.

**Changes:**
- Removed dead compact HF rendering path that showed degraded rows with non-functional clicks (double rendering bug)
- HF search results now use rich zone cards as the single rendering path
- Clicking an HF result opens the right-pane ModelDetailPanel with an HF-specific view: repo metadata, downloads/likes, tags, variant picker with download buttons, and a README tab fetched from HuggingFace
- Added close button (x) to ModelDetailPanel top-right corner (works for both local and HF models, touch-friendly)
- Removed HuggingFace entry from nav rail (was a dead-end filter)
- HF zone auto-elevates to top of list when no local models match the search query

Addresses #2564

Tests: 201 passed / 9 skipped — zero regressions


fl0rianr (approved):
LGTM

PR #2703, opened 2026-07-14
Fix: re-pulling a `collection.router` model fails with "unknown key 'source'"

SlawomirNowaczyk:
## Summary

Re-pulling (overwriting) an already-registered `collection.router` model returned HTTP 500 with
`Invalid collection.router routing policy: collection contains unknown key 'source'`. The overwrite
path validated the internally-enriched model entry (which carries an injected `source`/`registry_source`)
against the strict routing-policy parser, which rejects unknown root keys.

## Details

- For an already-registered model, `register_model` enriches `registration_data` with the persisted
  registry `source` before validation (`model_manager.cpp`).
- In the collection-overwrite path, validation was handed that enriched `registration_data` rather than
  the original user-authored `model_data`, so the injected `source` reached
  `validate_collection_request` → `parse_route_policy_collection` and tripped the strict unknown-key check.
- `source`/`registry_source` are persisted model-entry properties, not part of the routing-policy
  document, so they must not be introduced into the parser's grammar.

## Fix

Validate the original user-authored `model_data` in the collection-overwrite path instead of the
enriched `registration_data`. `registration_data` (with the injected registry source) is still used
afterwards for `register_user_model()`, so the existing registry source remains preserved — without
making `source`/`registry_source` part of the public routing-policy document. This keeps the
schema/parser parity intact (`route_policy.schema.json` does not define these root properties).

## Testing

- Added regression test `test_021zh_router_collection_repull_overwrite`: registers a
  `collection.router`, then overwrites/re-pulls the same collection without explicitly providing a
  registry source; both pulls must return 200.
- `RoutingPolicyParserTest` (schema/parser parity) passes.

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -3146,7 +3146,10 @@ void ModelManager::download_model(const std::string& model_name,
         bool is_collection_overwrite = is_model_collection_recipe(actual_recipe) &&
                                         model_data.contains("components");
         if (is_collection_overwrite) {
-            if (auto err = validate_collection_request(model_name, registration_data)) {
+            // Validate the original user-authored request, not registration_data:
+            // the latter is enriched with the persisted registry source, which is
+            // not part of the public routing-policy document the parser accepts.
+            if (auto err = validate_collection_request(model_name, model_data)) {
                 throw std::runtime_error(*err);
             }
             model_registered = false;
--- a/src/cpp/server/routing_policy_parser.cpp
+++ b/src/cpp/server/routing_policy_parser.cpp
@@ -496,8 +496,7 @@ std::vector<Rule> parse_rules(const json& routing,
 
 const std::set<std::string>& routing_policy_root_keys() {
     static const std::set<std::string> keys = {
-        "version", "model_name", "recipe", "components", "models", "routing",
-        "source", "registry_source"};
+        "version", "model_name", "recipe", "components", "models", "routing"};
     return keys;
 }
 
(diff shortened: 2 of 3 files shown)


fl0rianr (changes requested):
I think the bug is correctly identified, but this fix changes the routing-policy grammar at the wrong layer.

`registration_data` is intentionally enriched with the persisted registry source, while the original `model_data` is supposed to remain the user-authored request used for validation. In the collection overwrite path, however, validation currently receives `registration_data`, which is why the internally injected `source` reaches the strict routing-policy parser.

Could we instead validate the original request here?

```cpp
if (is_collection_overwrite) {
    if (auto err = validate_collection_request(model_name, model_data)) {
        throw std::runtime_error(*err);
    }
    model_registered = false;
    ...
}
```

`registration_data` can still be used afterwards for `register_user_model()`, so the existing registry source remains preserved without making `source` and `registry_source` part of the public routing-policy document.

The current change also breaks the existing schema/parser parity test, because `route_policy.schema.json` does not define these two root properties.

Please also add a regression test covering the actual sequence: register a `collection.router`, then overwrite/re-pull the same collection without explicitly providing its registry source.

eddierichter-amd (changes requested):
Good catch on the bug! I agree with Florian's suggested fix.

SlawomirNowaczyk:
Thanks — you're right, this belonged at the validation layer, not the grammar. I reverted the parser change and now validate the original `model_data` in the collection-overwrite path, leaving `registration_data` (with the injected registry source) for `register_user_model()`. Schema/parser parity is restored (`RoutingPolicyParserTest` passes), and I added `test_021zh_router_collection_repull_overwrite` covering the register-then-re-pull sequence without an explicit registry source. PTAL.

eddierichter-amd (approved):
LGTM

fl0rianr (approved):
LGTM, thanks!

PR #2704, opened 2026-07-14
feat: prove a comments-only PR with a label-triggered CI check

ianbmacdonald:
Follow-up to #2689 (the flagger). Adds `--assert-comments-only`: strips comments and blank lines from both revisions of every changed file and checks the remaining code is unchanged, so a PR that claims to be a comment cleanup can be checked mechanically instead of re-read.

Opt-in and maintainer-controlled: apply a `comments-only` label and CI runs it on `base..HEAD`. The verifier runs from the base revision, not the PR checkout, so a PR can't weaken the check by editing the checker. The label doesn't exist on this repo yet, so the job is inert until someone creates it.

It's a hardened check, not a proof — Python is compared via `ast` + the tokenizer, C/C++ via a scanner that models comments, splices, string/raw/header-name literals.

**Known limitation:** not sound for directive *scope*. A directive moved between two identical code lines governs different code while the executable code is byte-identical, and that move isn't detected. Documented on `comments_only()`.

90 unit tests, including a regression test per issue found.


fl0rianr (changes requested):
Thanks, but I don’t think this is safe to merge yet.

The core issue is that the PR introduces a handwritten Python/C/C++ comment parser and then presents its result as proof that no code changed. That creates false confidence unless the checker is substantially harder to fool than the code it reviews.

I found three concrete problems:

1. **The PR controls its own verifier.** The workflow checks out the PR contents and runs `tools/check_comment_slop.py` from that checkout. A PR can therefore modify the verifier itself and make the check pass. The verifier must be executed from a trusted base revision, not from the PR under review.

2. **C/C++ comments are removed without inserting whitespace.** For example, `int/**/x = 1;` and `intx = 1;` currently normalize to the same output, although they are different programs. Comments act as token separators in C and C++.

3. **Tooling directives are compared without their position or scope.** Moving an unchanged `# type: ignore`, `# fmt: off`, or similar directive can change tool behavior while producing the same directive list and the same comment-stripped code representation.

The workflow should also rerun when the PR base branch is changed; `labeled`, `synchronize`, and `reopened` do not cover retargeting.

More fundamentally, I’m not convinced this should claim to “prove” a comments-only change. The PR description itself documents fifteen previous bypasses, including several introduced by fixes for earlier bypasses. That strongly suggests this is becoming a security-sensitive language parser rather than a small repository check.

My recommendation is to either use trusted compiler/parser tooling with a much narrower claim, or keep comments-only PRs as a reviewer classification rather than adding a green check that may incorrectly certify them.

ianbmacdonald:
Fixed your three points: the verifier runs from the base revision, `int/**/x` no longer equals `intx`, and directives are anchored to the code they govern.

On the fundamental one — each fix just surfaced the next gap, and closing the last one needs the parser you warned about. If you're not seeing the value here, I'm thinking maybe we drop this part. #2689 stands on its own.

ianbmacdonald:
Closing this. #2689 carries the useful half, and closing the directive-scope gap here needs the parser @fl0rianr flagged in review. Thanks for the reviews.

PR #2709, opened 2026-07-14
fix(security): pin vendored model-viewer bundle to verified upstream

superm1:
# Summary

The vendored `src/app/src/renderer/vendor/model-viewer.min.js` had drifted from the documented upstream and carried no integrity hash, so a webpack import could not detect tampering or corruption.

This PR pins to the pristine artifact and secures it:

- Replace the bundle with the exact `dist/model-viewer.min.js` from the npm registry tarball for `@google/model-viewer@4.3.1`, verified against npm's published sha512 integrity.
- Add `model-viewer.min.js.sha256` (SHA-256 `283b06…`) and a CI job in `docs_and_style.yml` that runs `sha256sum -c` to fail on any drift.
- Rewrite the vendor `README.md` to record the hash chain (npm tarball integrity → extracted-file SHA-256) and a hash-preserving update process that validates npm's dist.integrity before extraction.

### Shadow mapping fix (upstream #5168)

The locally patched 4.3.0 file contained a shadow-depth shader fix from [google/model-viewer#5168](https://github.com/google/model-viewer/pull/5168) correcting the orthographic frustum and floor/blur-plane scaling for models larger than 1×1. That fix is included in the official 4.3.1 release, so the bundle is now pristine upstream with the fix intact.

### "Drift guard" vs SRI

The sidecar SHA-256 check is a drift guard rather than a true SRI replacement: the file and checksum can be changed together in the same commit. The npm dist.integrity (SHA-512) is the authoritative upstream trust anchor, and the documented update process verifies it before extracting the file.

### DIFF
--- a/src/app/src/renderer/vendor/README.md
+++ b/src/app/src/renderer/vendor/README.md
@@ -3,17 +3,68 @@
 ## model-viewer.min.js
 
 - **Source:** [`@google/model-viewer`](https://www.npmjs.com/package/@google/model-viewer) `dist/model-viewer.min.js`
-- **Version:** 4.3.0
+- **Version:** 4.3.1
 - **License:** Apache-2.0 (Google LLC); three.js (MIT) is bundled inside
 - **Why vendored:** the Debian-native `lemonade-server` package must build using only
   npm modules available in Debian (`USE_SYSTEM_NODEJS_MODULES`), and Debian does not
   ship model-viewer. A single self-contained bundle works in both the Tauri app and
   the browser web-app without adding an npm dependency.
-- **Update process:** download the new release's minified bundle and replace this file:
 
-  ```bash
-  curl -L https://cdn.jsdelivr.net/npm/@google/model-viewer@<version>/dist/model-viewer.min.js \
-       -o src/app/src/renderer/vendor/model-viewer.min.js
-  ```
+### Integrity
 
-  Then update the version in this file and verify the 3D panel preview still renders.
+The bundle is the unmodified upstream artifact. Its SHA-256 is pinned in
+`model-viewer.min.js.sha256` and verified in CI (see `docs_and_style.yml`) as a
+drift guard — the file and checksum can be changed together in the same commit,
+so this is not a true SRI replacement. The npm `dist.integrity` (SHA-512) is the
+authoritative trust anchor; the update process below enforces it.
+
+- **SHA-256 (this file):** `283b0672384614b4847636c306fc93fe4b1fcadc76d668b4e47f0ca76bcf033b`
+- **npm tarball integrity (`4.3.1`):** `sha512-GP+inXhAtY31E8rILVmByA6z8CZZjdlNajddppyI1/j1eIaSQiZcMRaUqTFe7+jv4mzRzwKIOiKBud0apiv+WQ==`
+
+The npm `dist.integrity` is the upstream trust anchor. The SHA-256 pins the exact
+`dist/model-viewer.min.js` extracted from that verified tarball. Both are reproducible
+from the [npm registry metadata](https://registry.npmjs.org/@google/model-viewer/4.3.1).
+
+### Update process
+
+Do not edit the bundle by hand. To move to a new version, verify the npm tarball
+integrity, extract the pristine artifact, and re-pin the SHA-256:
+
+```bash
+VERSION=<new-version>
+
+# (1) Fetch the registry tarball — npm publishes `dist.integrity` as sha512.
+TARBALL="/tmp/model-viewer-${VERSION}.tgz"
+curl -sL "https://registry.npmjs.org/@google/model-viewer/-/model-viewer-${VERSION}.tgz" -o "$TARBALL"
+
+# (2) Independently verify npm's published integrity hash.
+EXPECTED=$(curl -s "https://registry.npmjs.org/@google/model-viewer/${VERSION}" \
+  | python3 -c "import sys,json; print(json.load(sys.stdin)['dist']['integrity'])")
+ACTUAL=$(python3 - "$TARBALL" <<'PY'
+import base64
+import hashlib
+import pathlib
+import sys
+
+digest = hashlib.sha512(pathlib.Path(sys.argv[1]).read_bytes()).digest()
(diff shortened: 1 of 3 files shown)


src/app/src/renderer/vendor/model-viewer.min.js
```diff
@@ -991,7 +991,7 @@ class s_ extends Ks{constructor(){super(void 0,new Ns({depthWrite:!1})),this.hei
...
...bled=!1;const n=t.getRenderTarget();t.setRenderTarget(this.renderTarget);const r=t.getClearAlpha();t.setClearAlpha(0);const s=t.state;null!=s&&(s.buffers.color.setMask(!0),s.buffers.depth.setMask(!0)),t.clear();const a=t.autoClear;t.autoClear=!1,t.render(e,this.camera),t.autoClear=a,e.overrideMaterial=null,this.floor.visible=!0,this.blurShadow(t),t.xr.enabled=i,t.setRenderTarget(n),t.setClearAlpha(r)}blurShadow(t){const{camera:e,horizontalBlurMaterial:i,verticalBlurMaterial:n,renderTarget:r,renderTargetBlur:s,blurPlane:a}=this;a.visible=!0,a.material=i,i.uniforms.h.value=1/this.renderTarget.width,i.uniforms.tDiffuse.value=this.renderTarget.texture,t.setRenderTarget(s),t.render(a,e),a.material=n,n.uniforms.v.value=1/this.renderTarget.height,n.uniforms.tDiffuse.value=this.renderTargetBlur.texture,t.setRenderTarget(r),t.render(a,e),a.visible=!1}dispose(){null!=this.renderTarget&&this.renderTarget.dispose(),null!=this.renderTargetBlur&&this.renderTargetBlur.dispose(),this.depthMaterial.dispose(),this.horizontalBlurMaterial.dispose(),this.verticalBlurMaterial.dispose(),this.floor.material.dispose(),this.floor.geometry.dispose(),this.blurPlane.geometry.dispose(),this.removeFromParent()}}
```
fl0rianr, line 994: blocking: This change does more than restore the exact-string shader replacement. The removed code is the minified form of upstream google/model-viewer#5168, which fixes shadow mapping for models larger than 1×1 by correcting the orthographic frustum, scaling the floor/blur planes, and explicitly clearing the render target. Reverting to the pristine 4.3.0 bundle therefore reintroduces that rendering bug. Since the fix is included officially in @google/model-viewer@4.3.1, please pin and hash the pristine 4.3.1 artifact instead

src/app/src/renderer/vendor/README.md
```diff
@@ -9,11 +9,38 @@
...
+
+Do not edit the bundle by hand. To move to a new version, replace it with the pristine
+upstream artifact and re-pin the hash:
+
+```bash
+VERSION=<new-version>
+# Fetch the exact bundle npm publishes (registry tarball is the source of truth).
+curl -sL "https://registry.npmjs.org/@google/model-viewer/-/model-viewer-${VERSION}.tgz" \
+  | tar xz -C /tmp
+cp "/tmp/package/dist/model-viewer.min.js" src/app/src/renderer/vendor/model-viewer.min.js
```
fl0rianr, line 53: issue: The documented update process does not actually verify npm's dist.integrity; it downloads the tarball and then trusts whatever was extracted when generating the new SHA-256. Since the PR presents the npm SHA-512 as the independent upstream trust anchor, please make that verification part of the update procedure. Also, the sidecar check is best described as a drift guard rather than an SRI equivalent, because the asset and checksum can be changed together in the same commit

fl0rianr (changes requested):
Thanks, right direction, but not ready yet:

src/app/src/renderer/vendor/README.md
```diff
@@ -3,17 +3,59 @@
...
+VERSION=<new-version>
+
+# (1) Fetch the registry tarball — npm publishes `dist.integrity` as sha512.
+TARBALL="/tmp/model-viewer-${VERSION}.tgz"
+curl -sL "https://registry.npmjs.org/@google/model-viewer/-/model-viewer-${VERSION}.tgz" -o "$TARBALL"
+
+# (2) Independently verify npm's published integrity hash.
+EXPECTED=$(curl -s "https://registry.npmjs.org/@google/model-viewer/${VERSION}" \
+  | python3 -c "import sys,json; print(json.load(sys.stdin)['dist']['integrity'])")
+ACTUAL=$(sha512sum "$TARBALL" | awk '{print "sha512:" $1}')
```
fl0rianr, line 43: blocking: EXPECTED and ACTUAL use different digest representations. npm's `dist.integrity` is formatted as `sha512-<base64>`, while sha512sum produces a hexadecimal digest and this code prefixes it with sha512. As a result, the documented update process rejects even the correct upstream tarball.

Generate ACTUAL in npm's SRI representation instead:

```suggestion
ACTUAL=$(python3 - "$TARBALL" <<'PY'
import base64
import hashlib
import pathlib
import sys

digest = hashlib.sha512(pathlib.Path(sys.argv[1]).read_bytes()).digest()
print("sha512-" + base64.b64encode(digest).decode("ascii"))
PY
)
```

fl0rianr (commented):
Thanks, better, but not fully there yet.

fl0rianr (commented):
Thanks, this is much closer. I still see one blocker though: `model-viewer.min.js` itself is unchanged from the base branch, so the PR currently documents and hashes the existing blob as pristine 4.3.1 without actually replacing it.

Could you also make the update snippet fail closed (`set -euo pipefail`, ideally extract into a fresh temp dir), and change the workflow comment from “equivalent of SRI” to “drift guard” to match the README?

Once those are fixed, this looks good to me.

fl0rianr (commented):
Thanks, the previous points look good now. One small thing remains: please use `mktemp -d` for the extraction directory as well, so the update flow is fully isolated and fail-closed even after an interrupted previous run.

It would also be helpful to note that the verified 4.3.1 artifact produces the existing `283b06…` SHA-256, which explains why the vendored bundle itself does not appear in the diff.

Merge conflicts, rebase or main sync is due.

With that, this looks good to me.

fl0rianr (approved):
LGTM

PR #2710, opened 2026-07-14
fix(security): pin SHA-256 checksums for contributor-hosted backend assets

superm1:
## Summary

The `openmoss`, `acestep`, `thinksound`, and `trellis` backends download release binaries from **contributor-personal GitHub repos** (`pwilkin/*`) using only a version tag, with **no content verification** — a re-tagged or tampered release asset would install silently. (Raised as part of the v11.0.0 release review, #2706.)

This records the SHA-256 of every published asset under `checksums.github.<owner/repo>.<tag>.<asset>` in `backend_versions.json`. The download path **already** verifies against this key (`BackendUtils::lookup_expected_asset_hash` → `DownloadOptions::expected_hash`), so **no code change is needed** — a mismatch now fails the install. Digests are the values GitHub publishes for each release asset.

Covers all 24 assets (4 backends × vulkan/rocm/cuda × linux/windows):

| backend | repo | tag |
|---|---|---|
| openmoss | `pwilkin/openmoss` | v0.1.3 |
| acestep | `pwilkin/acestep.cpp` | v0.1.1 |
| thinksound | `pwilkin/thinksound.cpp` | v0.1.2 |
| trellis | `pwilkin/trellis.cpp` | v0.1.5 |

## Verification (live, on real hardware)

- **Positive:** `lemonade backends install openmoss:vulkan` downloaded the real asset and logged `Hash verified for …moss-tts-vulkan-linux-x64.tar.gz`; install succeeded.
- **Negative:** temporarily corrupting the pinned hash caused the download to be rejected — `hash mismatch: expected sha256:0000…, got sha256:15e648…` — and the install failed. (Only the build copy was tampered; the committed source is clean.)
- All 24 lookup paths resolve against the C++ candidate-path order; JSON validated.

## Note on maintenance

These hashes are tied to the current version pins — bump them together when a backend version changes. The update process is: `gh api repos/<repo>/releases/tags/<tag> --jq '.assets[] | "\(.name) \(.digest)"'` (GitHub publishes the digests, so no download needed).

This addresses the GitHub-binary half of the OpenMOSS provenance item in #2706. The HuggingFace model-checkpoint pinning (`ilintar/*`) is separate — it needs a checkpoint-parser change and is not included here.

### DIFF
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -1,5 +1,50 @@
 {
   "comment": "This configuration file controls which llama.cpp, whisper.cpp, sd.cpp, ryzenai-llm, and FLM versions are downloaded for each backend. You can modify these values to pin specific versions without rebuilding the application.",
+  "checksums": {
+    "comment": "Expected SHA-256 of downloaded GitHub release assets, verified after download (see BackendUtils::lookup_expected_asset_hash). Keyed by checksums.github.<owner/repo>.<tag>.<asset>. Values are the sha256:<hex> digests GitHub publishes for each asset. Update alongside the matching version pin below.",
+    "github": {
+      "pwilkin/openmoss": {
+        "v0.1.3": {
+          "moss-tts-vulkan-linux-x64.tar.gz": "sha256:15e64803ed7f93e0459f4e3132346a83123e1de0b973702c2c0631ae32f55feb",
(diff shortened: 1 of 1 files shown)


kenvandine (approved):
I wasn't able to test this, but I am all in favor of a check here to ensure the binaries match what's expected. Looks good to me.

PR #2711, opened 2026-07-15
Add a github action to close old PRs without activity

superm1:
With AI, everyone's working insanely fast.  We need to manager our open PRs effectively, so add stale bot to help.

### DIFF
--- a/.github/workflows/stale-prs.yml
+++ b/.github/workflows/stale-prs.yml
@@ -1,0 +1,27 @@
+name: "Close stale PRs"
+on:
+  schedule:
+    - cron: "30 1 * * *"
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 7
+          days-before-close: 7
+          stale-pr-message: >
+            This pull request has not had any activity in the last week.
+            It is being marked as stale — if there is no further activity
+            in the next week, it will be automatically closed.
+          exempt-pr-labels: stale-exempt
+          close-pr-message: >
+            This pull request has been automatically closed due to
+            extended inactivity. If you would like to continue working
+            on this, feel free to reopen it.


fl0rianr (changes requested):
Strong no on 7/7.

That would optimize for closing PRs quickly, not for managing them well. It creates exactly the load we want to avoid: authors must constantly bump PRs to keep them alive, and maintainers then have to reopen, re-triage, or explain closures that happened simply because someone was on vacation, busy with work, waiting for review, or contributing in their free time.

“AI makes everyone faster” is not a good basis for contributor policy. AI may speed up parts of implementation, but it does not remove review queues, testing time, maintainer availability, time zones, holidays, or the fact that most contributors are volunteers.

A smarter and more community-friendly policy would distinguish between “waiting for author” and “waiting for maintainer”, exclude drafts, blocked work, and PRs awaiting review, and use a much more reasonable window such as 45 days stale plus 14 days before closure. We can also think about "needs-maintainer-triage" after a certain stale time depending on the status etc.

The goal should be to reduce real backlog, not manufacture activity or punish contributors for having a life.

superm1:
I feel a problem right now is too many submissions that we can't realistically keep up with reviews. There are several multiple months old.

So we send a wrong message if we let them sit around idle not getting any attention.

Anyway food for thought in future maintainers meetings.

fl0rianr:
> Anyway food for thought in future maintainers meetings.

I agree, should not be our decision alone.

superm1:
Some other thoughts about your proposal of 45/14 - this is VERY unlikely to apply to the codebase still.

If 7/7 is too aggressive, we need a Goldilocks value.

fl0rianr:
GitHub’s own 30/14 example is for inactive issues and even disables automatic stale handling for PRs entirely.

The 45 days may be too long for the ground idea of this change. I am not attached to that exact number, I wanted to have it comfortable for the community and the maintainer as well, hence this "conservative" approach.
But finding a Goldilocks value makes sense.

As my minimum I would suggest **21 days before stale and another 14 days before closing** as a reasonable compromise. That is already substantially stricter than 30/14 or even 45/14.

Seven days is still too short to infer that a contribution has been abandoned. It is barely one normal vacation, a busy work period, or a review delay. Closing after only 14 days total also risks creating more work: contributors have to post artificial “still working on this” comments, while maintainers may need to reopen, re-triage, or explain prematurely closed PRs.
Many maintainers are also volunteers, have vocation, other things to tackle etc. 

I also do not think inactivity alone tells us whether the code still applies. A three-week-old PR may already require rebasing, while a two-month-old isolated fix may still be perfectly valid.
The bot cannot make that distinction.

For comparison, I'm not aware of an equivalent stale-PR workflow in the current llama.cpp or vLLM repositories, both have massive PR counts. So 21/14 would already be a fairly proactive policy, while still leaving room for vacations, volunteer schedules, and maintainer-side review delays.

Ideally, PRs marked as waiting for maintainer review, blocked, draft, or explicitly exempt should not be closed automatically.

superm1:
How about if we split the difference to 14/7?  Things that we really want to make sure get in can still have a tag added to make the bot stand down.

fl0rianr:
I agree we should do differentiation. 
Just how?
Much more generous on draft PRs, they can hang a while. If you know it will not be a recent merge this can safe from the bot. 
People with write access have more trust, be more generous here, workload already high enough. 
I think the biggest issue you might see with don't closing those that are waiting for review is let's call it AI spam PRs. 
Those should be flagged somehow and will be closed sooner.
I see, a very complex discussion.

superm1:
Well maybe we need to limit the number of pull requests for people without write access.  That's what llama.cpp does - if you don't have write access you're allowed 1 PR.

fl0rianr:
About PR count per author:

No, if you have no contribution (1 commit) you can have only one. But it's what's said, not what people are doing. It can't be enforced.

Let's say we can enforce it with a new function of GH, what is with the: Please split this PR, etc.

But you are right, this is a different issue.

ramkrishna2910:
Will add this to the maintainer meeting agenda.

superm1:
> About PR count per author:
> 
> No, if you have no contribution (1 commit) you can have only one. But it's what's said, not what people are doing. It can't be enforced.
> 
> Let's say we can enforce it with a new function of GH, what is with the: Please split this PR, etc.
> 
> But you are right, this is a different issue. 
> 
> 

I actually have one PR open with llama.cpp right now and it won't let me open another at the same time because I don't have write access there.

So yes there are policy knobs like that on GH.

fl0rianr:
> I actually have one PR open with llama.cpp right now and it won't let me open another at the same time because I don't have write access there.
> 
> So yes there are policy knobs like that on GH.

So it changed then, does not wonder me - interesting to know.

.github/workflows/stale-prs.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-stale: 7
+          days-before-close: 7
```
superm1, line 18: @jeremyfowers I've been out the past few weeks, did this ever get discussed?

superm1:
It sounds like this is going to be managed by a local LLM agent instead, closing.

PR #2712, opened 2026-07-15
fix(ci): always run macOS C++ unit tests regardless of signing

superm1:
The macOS "Run C++ unit tests" step only executed when `has_signing != 'true'`, so any run with signing secrets available (all internal/maintainer PRs and every merge-queue run against main) skipped ctest entirely. The only path that ever exercised these tests was external fork PRs, which don't receive repo secrets and lack things like HF_TOKEN — leaving no reliable macOS unit-test signal for code that actually lands on main, and spurious CI failures on unrelated fork PRs (e.g. docs-only changes) whenever that starved path breaks.

Run ctest unconditionally on this job; it only depends on the build, not on packaging/signing.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -586,7 +586,6 @@ jobs:
         shell: bash
         run: |
           cmake --build --preset default --target test_directory_watcher
-          cmake --build --preset default --target test_process_manager
           cmake --build --preset default --target test_latest_version_fallback
           cmake --build --preset default --target test_routing_policy_contract
           cmake --build --preset default --target test_routing_policy_evaluator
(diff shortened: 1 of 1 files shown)


superm1:
I'm hoping this fixes https://github.com/lemonade-sdk/lemonade/pull/2701

fl0rianr:
I introduced ProcessManagerTest in #2697 and accidentally added the Linux-only target to both the Debian and macOS test blocks.

I’ve corrected the macOS block by removing this test again, while keeping the test fully enabled in the Linux/Debian job. The signing gate can now be removed safely, so the remaining cross-platform C++ tests run for both signed and unsigned macOS builds.

fl0rianr (approved):
After careful consideration merging this as a safe and much appreciated fix.

PR #2715, opened 2026-07-15
Fix: Repeated model update reports after pull

### LINKED ISSUE #2713: lemond model update check isn't processing updates correctly
### Platform

Linux/Ubuntu

### Lemonade Version

11.0

### GPU / APU Model

gfx1152 (but all)

### Component

lemond (server)

### Bug Description

```
ckuethe@trinity:~$ lemonade check-updates
Updates available for 4 model(s):
  - Gemma-4-26B-A4B-it-GGUF
  - Gemma-4-E2B-it-GGUF
  - Gemma-4-E4B-it-GGUF
  - Qwen3.6-35B-A3B-MTP-GGUF
ckuethe@trinity:~$ lemonade pull Gemma-4-E2B-it-GGUF
Pulling model: Gemma-4-E2B-it-GGUF
Total: 3.8 GB, 3 files
[1/3] gemma-4-E2B-it-Q4_K_M.gguf (2962.8 MB)
  (already downloaded)                                                          
[2/3] config.json (0.0 MB)
  (already downloaded)                                                          
[3/3] mmproj-F16.gguf (940.0 MB)
  (already downloaded)                                                          

Model pulled successfully: Gemma-4-E2B-it-GGUF
ckuethe@trinity:~$ lemonade pull Gemma-4-E4B-it-GGUF
Pulling model: Gemma-4-E4B-it-GGUF
Total: 5.6 GB, 3 files
[1/3] gemma-4-E4B-it-Q4_K_M.gguf (4746.6 MB)
  (already downloaded)                                                          
[2/3] config.json (0.0 MB)
  (already downloaded)                                                          
[3/3] mmproj-F16.gguf (944.5 MB)
  (already downloaded)                                                          

Model pulled successfully: Gemma-4-E4B-it-GGUF
```

### Steps to Reproduce

1. `lemonade check-updates`
2. `lemonade pull <outdated-model>`
3. observe "(already downloaded)

Expected: outdated model gets update
Actual: outdated model not updated

### Log Output

```shell

```

### Additional Context

_No response_

fl0rianr:
## Summary

* Track processed registry snapshots per model and checkpoint selection instead of per repository.
* Use the model’s resolved snapshot before falling back to `refs/main`.
* Serialize downloads that share the same registry repository.

This clears metadata-only updates after a successful pull without hiding real updates for sibling quantizations in the same repository.

## Testing

* Ensured normal download still works fine
* Recreated issue below
* Tested the fix based on this Recreation

Fixes #2713

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -212,6 +212,68 @@ static std::string read_hf_ref_main(const fs::path& model_cache_path) {
     return ref;
 }
 
+static std::string registry_model_selection(const ModelInfo& info) {
+    std::ostringstream selection;
+    selection << info.recipe;
+    for (const auto& [type, checkpoint] : info.checkpoints) {
+        selection << '\n' << type << '=' << checkpoint;
+    }
+    return selection.str();
+}
+
+static std::string read_processed_registry_snapshot_id(
+    const fs::path& model_cache_path,
+    const std::string& model_name,
+    const std::string& selection) {
+    const fs::path provenance_path = model_cache_path / ".lemonade_registry.json";
+    if (!safe_exists(provenance_path)) {
+        return "";
+    }
+
+    try {
+        const json provenance = JsonUtils::load_from_file(path_to_utf8(provenance_path));
+        if (!provenance.contains("processed_models") ||
+            !provenance["processed_models"].is_object()) {
+            return "";
+        }
+
+        const auto& processed_models = provenance["processed_models"];
+        auto model_it = processed_models.find(model_name);
+        if (model_it == processed_models.end() || !model_it->is_object()) {
+            return "";
+        }
+        if (JsonUtils::get_or_default<std::string>(
+                *model_it, "selection", "") != selection) {
+            return "";
+        }
+        return JsonUtils::get_or_default<std::string>(*model_it, "snapshot_id", "");
+    } catch (const std::exception&) {
+        return "";
+    }
+}
+
+static std::string snapshot_id_from_resolved_path(
+    const ModelInfo& info,
+    const fs::path& model_cache_path) {
+    const std::string resolved_path = info.resolved_path("main");
+    if (resolved_path.empty()) {
+        return "";
+    }
+
+    const fs::path snapshots_path = model_cache_path / "snapshots";
+    const fs::path relative =
+        path_from_utf8(resolved_path).lexically_relative(snapshots_path);
+    if (relative.empty()) {
+        return "";
+    }
+
+    auto first = relative.begin();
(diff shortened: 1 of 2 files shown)


ckuethe:
indeed it does fix the issue

abn (changes requested):
The core changes look solid and solve the issue described in #2713 beautifully by tracking updates per model/variant selection rather than globally per repository. 

Before we merge, I would highly recommend addressing one remaining gap in test coverage, along with a couple of non-blocking nits for your consideration:

#### 1. Test Coverage Request
* There is currently no integration test coverage in the project testing the `POST /models/check-updates` endpoint itself. 
* It would be excellent if we could add a test in `test/server_endpoints.py` (e.g., `test_037_model_update_check_lifecycle`) that validates the update check lifecycle:
  1. Pulling a model (so it is cached locally).
  2. Mocking or overriding the local metadata/ref to simulate an older snapshot.
  3. Calling `POST /models/check-updates` to verify the model shows up as needing an update.
  4. Pulling the model again to "update" it.
  5. Verifying that a subsequent check-update call returns no updates for that model.

#### 2. Non-blocking Feedback / Nits
* **Nit A (Serialization robustness in `registry_model_selection`)**:
  ```cpp
  static std::string registry_model_selection(const ModelInfo& info) {
      std::ostringstream selection;
      selection << info.recipe;
      for (const auto& [type, checkpoint] : info.checkpoints) {
          selection << '\n' << type << '=' << checkpoint;
      }
      return selection.str();
  }
  ```
  Since `info.checkpoints` is a `std::map`, key iteration order is lexicographically sorted and deterministic, which is great. However, if any custom checkpoint names or types contain newlines or equal signs, this simple `type=checkpoint` serialization could collide. It works perfectly for standard remote registries, but we might want to sanitize or escape separators in a future refactor.
* **Nit B (Non-atomic writes in `JsonUtils::save_to_file`)**:
  `save_to_file` directly overwrites the target file without writing to a temp file first. If the write is interrupted (e.g., server crash), the JSON file could be left corrupted. Your fallback chain in `read_processed_registry_snapshot_id` handles this correctly by catching all exceptions and falling back to the path SHA / `refs/main`, so this is fully mitigated, but it's worth considering atomic write-and-replace patterns in our file utility layer in the future.

abn (approved):
LGTM! 🚀

Thank you for adding the integration tests to cover this workflow. 

I've verified the changes locally on the latest commit (`c8f0f5b49`):
1. The new lifecycle integration test (`test_037_model_update_check_lifecycle`) passes successfully and correctly validates update detection and clearing.
2. Cache cleanups in the test teardown work as expected.
3. All code quality, backwards-compatibility fallbacks, and download locking changes look solid.

Ready to merge!

PR #2716, opened 2026-07-15
feat: add Modelscope search

### LINKED ISSUE #2699: feat: add ModelScope marketplace search
## Summary

Add ModelScope as a searchable provider in the Lemonade model marketplace.

This is a follow-up to #2647, which adds ModelScope support to Lemonade’s registry-backed model download, caching, provenance, update-check, CLI, API, and manual-add workflows.

The first implementation intentionally leaves the desktop marketplace search Hugging Face-only. That is an acceptable initial scope, but it remains an important gap for the original use case in #2602: users in regions or networks where Hugging Face is unavailable can pull ModelScope models by repository ID or URL, but the in-app discovery experience is still a dead end.

Raised following the discussion with @ramkrishna2910  and the review feedback on #2647.

## Current behavior

The desktop marketplace search currently:

* runs client-side in the renderer;
* queries the Hugging Face model API directly;
* maps Hugging Face-specific result fields;
* applies Hugging Face-oriented GGUF filtering and display logic.

ModelScope models can already be added manually after #2647, but they are not discoverable through the marketplace search interface.

## Proposed behavior

Allow users to select or search both supported registries:

* Hugging Face
* ModelScope

Search results should clearly show their source and continue through the existing source-aware pull flow introduced by #2647.

### Server-side ModelScope search

The Hugging Face API is currently queried directly by the renderer. ModelScope’s API and browser/CORS behavior may require search requests to be proxied through `lemond`.

Prefer a provider-neutral server endpoint rather than adding ModelScope-specific networking directly to the renderer, for example:

```text
GET /api/v1/registry/search
    ?source=modelscope
    &query=qwen
    &limit=...
```

This endpoint could later support additional registries without expanding provider-specific renderer code.

### Provider-neutral result model

Normalize provider responses into a shared result structure containing at least:

* repository ID
* display name
* registry source
* description
* tags or task labels
* download/popularity metadata when available
* detected repository type
* GGUF availability
* available variants when requested

The renderer should not need to understand ModelScope’s raw API response format.

### GGUF detection and filtering

ModelScope’s repository metadata and file-result mapping differ from Hugging Face.

The implementation should verify:

* how GGUF repositories are identified;
* whether file trees must be queried separately;
* how sharded GGUF variants are represented;
* how repositories with nested variant folders are filtered;
* whether the existing `/pull/variants` logic can be reused after selecting a result.

### Mirrored repositories

Some models may exist on both Hugging Face and ModelScope.

The UI should avoid confusing duplicate results while preserving provenance. Possible approaches include:

* show both results with clear provider badges;
* group matching `owner/repository` entries;
* prefer the user-selected registry;
* never silently switch the selected provider during pull.

Repository name alone should not be treated as a globally unique identifier. The effective identity is:

```text
registry source + repository ID
```

### Authentication and endpoint configuration

Search should respect the existing ModelScope configuration introduced by #2647:

* `MODELSCOPE_API_TOKEN`
* `MODELSCOPE_ACCESS_TOKEN`
* `MODELSCOPE_ENDPOINT`

Private or inaccessible repositories should produce useful provider-specific errors rather than generic marketplace failures.

## Suggested scope

* Add a provider-neutral registry-search server endpoint.
* Implement ModelScope search through that endpoint.
* Keep or migrate Hugging Face search behind the same endpoint where practical.
* Add a registry selector or combined-search UX to the marketplace.
* Add clear Hugging Face and ModelScope source badges.
* Pass the selected source into the existing source-aware pull/variant flow.
* Handle mirrored repositories without losing registry provenance.
* Add unit tests for response normalization and filtering.
* Add an end-to-end test covering search → variant selection → pull.

## Acceptance criteria

* A user can search ModelScope from the desktop marketplace.
* Search works without direct renderer access to ModelScope APIs.
* Results clearly identify their registry source.
* Selecting a ModelScope result pulls it through the ModelScope registry path.
* Hugging Face remains the default and its existing marketplace behavior does not regress.
* Mirrored repositories do not silently change provider.
* GGUF filtering and variant discovery work for representative ModelScope repositories.
* Authentication and custom ModelScope endpoints are respected.

## Out of scope

FLM models remain backend-managed and are discovered and downloaded through the FastFlowLM CLI. Supporting ModelScope as an upstream source for FLM would require separate FastFlowLM support or a dedicated import/conversion workflow and is not part of this marketplace-search issue.

## Related

* Follow-up to #2647
* Continues the regional-accessibility goal from #2602

fl0rianr:
Summary

Add ModelScope marketplace search alongside Hugging Face in the Model Manager.

- Add a provider-neutral registry search endpoint.
- Show separate Hugging Face and ModelScope result sections with source badges.
- Verify ModelScope repositories contain compatible GGUF variants before displaying them.
- Preserve the selected registry source through variant discovery, download, registration, and update flows.
- Keep Hugging Face as the default source for backward compatibility.
- Support existing ModelScope authentication and custom endpoint configuration.

Testing

- Built the production renderer.
- Added frontend regression coverage for ModelScope search and source-aware downloads.
- Added server endpoint and registry normalization tests.
- Verified ModelScope search results expose downloadable GGUF variants only.
- Verified existing Hugging Face search and pull behavior remains unchanged.

Closes #2699

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,7 @@ All core endpoints are registered under **4 path prefixes**:
 - `/v0/` — Legacy short
 - `/v1/` — OpenAI SDK / LiteLLM compatibility
 
-**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `health`, `pull`, `load`, `unload`, `delete`, `params`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `log-level`, `logs/stream`
+**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `health`, `pull`, `pull/variants`, `registry/search`, `load`, `unload`, `delete`, `params`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `log-level`, `logs/stream`
 
 **Ollama-compatible endpoints** (under `/api/` without version prefix): `chat`, `generate`, `tags`, `show`, `delete`, `pull`, `embed`, `embeddings`, `ps`, `version`
 
--- a/src/app/src/renderer/ModelManager.tsx
+++ b/src/app/src/renderer/ModelManager.tsx
@@ -202,6 +202,26 @@ interface HFModelInfo {
   pipeline_tag?: string;
 }
 
+interface RegistrySearchResult {
+  repository_id?: string;
+  display_name?: string;
+  source?: string;
+  description?: string;
+  tags?: string[];
+  task?: string;
+  downloads?: number;
+  likes?: number;
+  has_gguf?: boolean;
+}
+
+interface RegistrySearchResponse {
+  results?: RegistrySearchResult[];
+  error?: string | {
+    message?: string;
+    path?: string;
+  };
+}
+
 interface HFSibling {
   rfilename: string;
 }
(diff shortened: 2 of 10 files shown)


ramkrishna2910:
Looks great! 
image

ramkrishna2910 (approved):
Reviewed and tested locally — building on `lemond` + the Tauri app, running the C++ registry unit tests (28/28 pass), exercising `/registry/search` live against ModelScope, and downloading a ModelScope GGUF model end-to-end through the UI. All good. No correctness blockers.

Two minor items before merge:

1. **Title typo:** "mode**s**cope" → "modelscope" in the PR/commit title, since it becomes the permanent squash-commit subject.
2. **Docs:** add `registry/search` to the endpoint list in `AGENTS.md` for discoverability.

Both are trivial and don't need another review pass.

fl0rianr:
Thanks and good catch! Added the pull/variants endpoint as well since this one was missing also. Force pushed for a clean commit message.

PR #2717, opened 2026-07-15
Router/2406 Collection GUI: collapse modes, standalone classifiers, per-rule graph escalation

SlawomirNowaczyk:
### Summary
Reworks the custom router-collection authoring UI so routing has just two mental models — **NL Router** vs **Rule-based Router** — with each rule independently rendering as a simple form or escalating to the graph canvas only when its condition actually needs it. Classifier authoring becomes its own reusable surface (with one-click templates), and classifiers can now be referenced directly from the simple form.

### Motivation
The prior UI exposed three global modes (NL / Quick Rules / Advanced Rules). That split was confusing and lossy in practice:

- **The Quick↔Advanced distinction was a false dichotomy.** All three modes serialize to the *same* `routing` JSON — they're just editors over one model. Quick Rules already supported N homogeneous AND/OR conditions; the *only* structural thing it couldn't express was a **mixed** AND+OR / nested / classifier condition. Forcing a global mode switch to unlock that for a single rule was heavy-handed.
- **The global toggle was destructive.** Switching to "Advanced" cleared existing rules/classifiers (reported by a reviewer), because the two editors didn't share state cleanly.
- **Classifier authoring was entangled with rule building.** A classifier is a reusable detector that *many* rules can reference, but it could only be created inline while wiring a rule graph — so declaring one and using one were the same awkward step.
- **Classifiers couldn't be used without the graph.** Even a trivial "classifier X == label → model" rule forced the full canvas, which is overkill for the most common case.

### What changed
**Routing modes → two**
- Removed the global Quick/Advanced toggle. `RouterRoutingMode` collapses from `'llm' | 'quick' | 'rules'` to `'llm' | 'rules'` in the model layer (`buildRouterCollectionPullRequest` treats any non-`llm` as rules; `routingToRouterCollectionDraft` always parses to `'rules'`). Switching modes no longer clears rules/classifiers.

**Per-rule escalation (replaces Advanced)**
- New `isFlatMatch`/`isFlatLeaf` predicate decides, *per rule*, whether the simple form can fully represent the condition. Flat = a lone condition or a single-level AND/OR of simple signals (leaf-level NOT is fine).
- Flat rules render `QuickRuleRow` with an **"Edit as graph"** affordance; non-flat rules (mixed AND/OR, nesting, score ranges) render an inline `RuleGraphRow` (canvas + toolbox) with **"Collapse to simple"** when they become flat again. Escalation is lossless because both edit the same tree.
- Deleted `RouterPipelineCanvas.tsx` (the old master-detail Advanced view), superseded by inline per-rule graphs.

**Standalone classifiers surface**
- New RouterClassifiersSection.tsx: a dedicated "Classifiers" library. Rules only *reference* classifiers by id; declaring vs. using are now separate jobs.
- Two starter **templates** (mirroring the rules "Start from" presets), pre-filled with real Lemonade models: 🔒 **PII Detector** (`llm` classifier, prefers a downloaded LLM, falls back to `Qwen3-0.6B-GGUF`) and 🎯 **Topic Router** (`semantic_similarity`, prefers a downloaded embedding model, falls back to `nomic-embed-text-v1-GGUF`). One click appends a fully-populated classifier with a deduped id.

**Classifiers in the simple form**
- The flat form's chip model now carries `classifierId` + `label`. A label-based classifier condition (match by label, or "fires") is treated as flat; only a **score range** still escalates to the graph.
- `QuickRuleRow` renders a classifier chip with an `is <label>` picker (dropdown of the classifier's concept keys / labels with an "(any)" option; free-text for `llm` classifiers). The **+ condition** menu lists declared classifiers below a divider.

### Compatibility / non-obvious notes
- **Lossless & backward-compatible.** No JSON schema change — all three former modes already produced the same `routing` payload, so existing collections load unchanged. Graph-authored classifier rules that only match by label will now conveniently collapse into the simple form.
- **Invariant respected.** This is client-side authoring only (renderer + shared model utils); no server/API surface, no per-client state moved into `lemond`.

### Testing
- Type-clean via the TS language server across all touched files and consumers.
- Manually smoke-tested via `npm run dev:renderer`: two-mode selector, classifier templates, per-rule form↔graph round-tripping, and classifier chips in the flat form.
- ⚠️ routerCollections.test.cjs and a full `tsc`/webpack build should run in CI — they couldn't be executed locally (app `node_modules` unavailable; npm registry cert-blocked on this machine). The logic test only uses `'llm'`/`'rules'` modes and already asserts keyword rules parse to `'rules'`, so it aligns with the mode collapse.


ramkrishna2910:
Per the discussion on #2728: gui2 is under a feature freeze ahead of GUI3, and #2728 appears to supersede this PR (identical file set, newer iteration, active UX discussion). @SlawomirNowaczyk if that's right, please close this one in favor of #2728 so review lands in one place; the eventual implementation will target GUI3 once it merges to main.

PR #2720, opened 2026-07-15
api: expose is_busy and is_streaming flags in /health response

therealjuanmartinez:
Add 'is_busy' and 'is_streaming' fields to each model in the 'all_models_loaded' array returned by GET /api/v1/health. This allows the detection of inference state.

- is_busy: true when model has active requests or maintenance in progress
- is_streaming: true when first output chunk arrives (differentiates prompt processing from token generation phase)

The is_streaming flag resets when the request completes (release_inference).

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -927,6 +927,8 @@ curl http://localhost:13305/v1/health
   - `type` - Model type: `"llm"`, `"embedding"`, `"reranking"`, `"transcription"`, `"image"`, or `"tts"`
   - `device` - Space-separated device list: `"cpu"`, `"gpu"`, `"npu"`, or combinations like `"gpu npu"`
   - `pinned` - Boolean indicating if the model is currently pinned to prevent auto-eviction
+  - `is_busy` - Boolean indicating if the model has active requests or maintenance in progress
+  - `is_streaming` - Boolean indicating if the model is actively generating output tokens (true after first chunk arrives, false when all streaming requests complete)
   - `backend_url` - URL of the backend server process handling this model (useful for debugging)
   - `pid` - The Process ID (PID) of the backend engine handling this model
   - `recipe` - Backend/device recipe used to load the model (e.g., `"ryzenai-llm"`, `"llamacpp"`, `"flm"`)
--- a/src/cpp/include/lemon/wrapped_server.h
+++ b/src/cpp/include/lemon/wrapped_server.h
@@ -192,7 +192,8 @@ public:
 
     void release_inference() {
         std::lock_guard<std::mutex> lock(state_mutex_);
-        is_streaming_ = false;
+        // Note: is_streaming_ is managed by end_backend_request() which correctly
+        // clears the flag only when the last streaming request completes.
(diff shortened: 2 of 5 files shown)


superm1 (changes requested):
## Critical Bug: Concurrent Streaming Requests

**Location:** `src/cpp/include/lemon/wrapped_server.h:195`

The `is_streaming_` flag is unconditionally cleared in `release_inference()`, which causes incorrect behavior when multiple streaming requests overlap:

```cpp
void release_inference() {
    std::lock_guard<std::mutex> lock(state_mutex_);
    is_streaming_ = false;  // ❌ BUG: clears even if other streams active
    if (--active_request_count_ == 0) {
        state_ = ModelState::READY;
```

**Failure scenario:**
1. Request A starts streaming → `set_streaming(true)`
2. Request B starts streaming → flag already true
3. Request A completes → `release_inference()` sets `is_streaming_ = false`
4. Request B is still streaming, but `/health` now shows `is_streaming: false` ❌

**Fix:** Only clear the flag when the last streaming request completes. Recommend tying it to `active_streaming_requests_` in `end_backend_request()`:

```cpp
void end_backend_request(BackendRequestKind kind) {
    // ... existing code ...
    
    if (kind == BackendRequestKind::Streaming) {
        previous = active_streaming_requests_.fetch_sub(1, std::memory_order_acq_rel);
        if (previous <= 1) {
            active_streaming_requests_.store(0, std::memory_order_release);
            set_streaming(false);  // Clear when last streaming request ends
        }
    }
}
```

---

## Missing Test Coverage

1. **No tests** for the new `/health` fields (`is_busy`, `is_streaming`)
2. **No validation** of concurrent streaming request behavior
3. **No documentation** of what "streaming" means (prompt processing vs token generation)

**Recommendation:** Add integration test in `test/server_endpoints.py` that:
- Starts a streaming request
- Polls `/health` to verify `is_streaming: true`
- Waits for completion
- Verifies `is_streaming: false`
- Tests concurrent streaming requests

---

## Minor Issues

1. **Member variable needs comment** (line 515):
   ```cpp
   // True between first output chunk and request completion.
   // Protected by state_mutex_. Only meaningful during streaming requests.
   bool is_streaming_ = false;
   ```

2. **PR description is incomplete** — Says "resets when the request completes" but doesn't mention behavior with concurrent requests

---

## What Works

✅ Thread safety via mutex guards  
✅ Consistent with existing `is_busy()` pattern  
✅ Clean integration into `/health` endpoint  
✅ Member initialization is correct  

---

## Verdict

**Request changes:** The concurrent streaming bug will cause incorrect status reporting under realistic multi-client load. Please fix before merging.

therealjuanmartinez:
is_streaming now only clears when the last streaming request completes (via active_streaming_requests_ counter in end_backend_request()). Added integration tests covering the concurrent streaming case.

superm1:
It looks like the new tests aren't passing CI, can you have a look?

Also; docs/api/lemonade.md needs to have these new fields documented please.

superm1 (approved):

Issue #2651, opened 2026-07-10
[Router] Add a “Test Prompt” Validator for Routing Decisions


sdevinenamd:
### Feature Description

Users should be able to test their prompts against the routing decisions they define directly within the Router Builder through a "Test Prompt" debug panel.

The panel should provide:
- A text input field for entering a prompt.
- Support for attaching images alongside the prompt.
- A "Validate Against Prompt" button that becomes available once a valid router configuration has been created.

When executed, the validator should:
- Trace the prompt through the configured routing logic.
- Visually display the routing path using a graph or decision-tree view.
- Highlight each routing decision that was evaluated.
- Show the final selected model and the reasoning behind the selection.

### Use Case / Motivation

Currently, users can build a router but can only validate its behavior by submitting prompts through the full pipeline. If changes are required, they must return to the Router Builder, reload or rebuild the router, make modifications, and run the pipeline again. This iterative workflow is time consuming and inefficient.

Adding a lightweight validation experience would enable users to quickly test, debug, and refine routing logic before deploying or running the router in the pipeline.

### Platform Relevance

All platforms

PR #2726, opened 2026-07-16
Issue 2651: Router Prompt Debugger (Ege Qutwo)

QuMuzafferEge:
A "Router: Prompt Debugger" panel is added to the view dropdown menu which lets the user test selected routing policy with a given prompt. The output is given as a trace log as well as decision tree visualization. Resolves issue: https://github.com/lemonade-sdk/lemonade/issues/2651


SlawomirNowaczyk (changes requested):
Overall looks nice. A few things to be addressed before merge:

## Main gap: the debugger can only exercise a subset of routing inputs

`handle_routing_validate` builds the `RouteContext` by hand and only populates three fields:

```cpp
RouteContext ctx;
ctx.input = prompt;
ctx.params.chars = prompt.size();
ctx.params.has_images = has_images;
```

But the engine can route on more than that. Per `RouteContext` in routing_policy.h, conditions can also match on:

- `params.has_tools` — request carried a non-empty `tools[]` array
- `params.model` — the addressed collection.router model
- `metadata` — the opaque caller-supplied string map (the header explicitly calls out keys like `task_class` / `consent` as one of the three trust seams)

Because `metadata` is always empty and `has_tools` is always false here, any policy with rules like `match: { has_tools: true }` or rules keyed off `metadata.*` will silently fall through in the debugger and never match — so the tool can misrepresent how the policy behaves in production. Note the real dispatch path uses `build_route_context(request_json, model_name)` to derive all of these from the inbound body, whereas this handler hand-rolls a partial context, which is exactly where the divergence creeps in.

Suggestion: let the panel supply a `has_tools` toggle and free-form `metadata` key/value pairs, and forward them into `ctx.params.has_tools` / `ctx.metadata`. That closes the gap and makes the debugger able to exercise every condition type the engine supports.

## Other issues

**1. Unused props on `PromptDebuggerPanel`.** The component declares `showSuccess` and `showWarning` in `PromptDebuggerPanelProps`, but only `showError` is destructured/used. Either wire them up (e.g. a success toast on validation, a warning for the inconsistent-state case below) or drop them from the interface so the contract stays honest.

**2. The "matched rule not found" edge case is swallowed.** In `renderDecisionTree` (`src/app/src/renderer/utils/decisionTree.tsx`), when `decision.matched_rule` isn't found among the rules, `matchedIndex === -1` and the tree renders with nothing highlighted. That's an inconsistent-server-state signal a router author would want to see, not a blank/uncolored tree. Consider surfacing a visible warning banner in that case rather than rendering silently.

**5. Decision-tree SVG: magic numbers + accessibility.** `decisionTree.tsx` hard-codes a lot of pixel layout constants and truncates rule ids/expressions (`truncate(..., 34)`, etc.) with no way to see the full text. Two things: (a) long ids/expressions get cut with no tooltip — add `<title>` / `aria-label` on the nodes so hover reveals the full expression; and (b) the `<svg>` has no accessible summary — a `role="img"` with an accessible label describing the decision would help screen-reader users and anyone whose text is truncated.

fl0rianr (commented):
Thanks! Overall, this is a useful addition and the new `has_tools` / `metadata` support closes an important part of the earlier gap. I still see a few issues that should be addressed before merge:

## 1. Validation results can become detached from the inputs that produced them

While `/routing/validate` is running, the user can still change the prompt, load another policy, toggle images/tools, or edit metadata. When the request finishes, the response is always stored via `setDecision(...)`.

The decision tree is then rendered against the current `policyJson`, and the trace download uses the current `prompt`, not necessarily the policy and prompt that were sent with the request.

This can produce a misleading result:

1. Validate policy A.
2. Load policy B or edit the prompt while validation is still running.
3. The response for A arrives.
4. The UI displays A’s decision using B’s tree and/or prompt.

Please either disable all relevant inputs while `isValidating`, or snapshot the submitted prompt/policy and ignore stale responses using a request ID or `AbortController`.

## 2. `prompt-debugger` is not supported by settings persistence

`App.tsx` accepts `prompt-debugger` as a `leftPanelView`, but the central `LayoutSettings` type and the Tauri settings sanitizer still only allow:

```ts
'models' | 'marketplace' | 'backends' | 'settings'
```

As a result, selecting the Prompt Debugger can be normalized back to `models` during the settings round-trip and will not restore correctly after restart.

Please add `prompt-debugger` to:

* `LayoutSettings.leftPanelView`
* the Rust `is_valid_left_panel_view()` whitelist
* the corresponding settings persistence tests

## 3. Classifier conditions are summarized incompletely in the tree

`summarizeMatchExpr()` only renders the first property of a leaf object:

```ts
const [key, value] = Object.entries(obj)[0]
```

For a valid condition such as:

```json
{
  "classifier": "pii",
  "label": "detected",
  "min_score": 0.8
}
```

the tree will likely display only `classifier: pii`, omitting the label and threshold that actually determine the match.

Please render classifier-band conditions explicitly and include all relevant fields.

## 4. Unknown `matched_rule` is still silently swallowed

If `decision.matched_rule` cannot be found in the displayed policy, `matchedIndex` becomes `-1`. The tree still renders without a clear warning, and the fall-through edges may look as if they were all evaluated.

This is an inconsistent state and should be surfaced visibly rather than rendered as a normal result. The existing `showWarning` prop could be used here.

## 5. Remaining review items

The following items from the previous review still appear open:

* `showSuccess` and `showWarning` are declared but unused.
* Truncated SVG labels have no `<title>` tooltip containing the full value.
* The decision-tree `<svg>` has no accessible `role` / label.

## Scope question

The issue asks for the validator inside the Router Builder, using the router configuration currently being authored. This implementation adds a separate view that requires manually uploading a JSON file.

Please confirm whether this changed UX has been agreed with @sdevinenamd the issue owner, or integrate the debugger with the Router Builder as originally requested.

SlawomirNowaczyk (changes requested):
Thanks for the follow-up — the persistence fix (whitelist + `LeftPanelView`/`appSettings` type + sanitizer + Rust test), the stale-results snapshot/request-id guard, and the classifier-band rendering in `summarizeMatchExpr` all look good and resolve the three items from the earlier reviews.

One regression to fix before this can merge, plus a couple of minor asks:

**🔴 Stuck "Validating…" state (please fix)**

The new request-id guard makes the `isValidating` reset conditional:

```ts
} finally {
  if (requestId === latestRequestIdRef.current) setIsValidating(false);
}
```

But `latestRequestIdRef` is also bumped by `handlePolicyFileChange` and `clearPolicy`, neither of which is disabled while a validation is in flight. So:

1. Click **Validate** → `requestId = 1`, `latest = 1`, `isValidating = true`, fetch in flight.
2. Swap the policy file mid-request → `latest = 2`.
3. Response for request 1 arrives → `if (1 !== 2) return;` exits the `try` early.
4. `finally`: `if (1 === 2)` is false → `setIsValidating(false)` is skipped.

`isValidating` is now stuck `true` with nothing to clear it, so the Validate button stays disabled/"Validating…" until the panel is reloaded. Since concurrent validations aren't possible (the button is disabled while validating), the request-id guard on the *result* already prevents stale data — the reset can just be unconditional:

```ts
} finally {
  setIsValidating(false);
}
```

**🟡 Minor**

- Could you add a JS-side persistence round-trip test for `'prompt-debugger'` in `appSettings`? The Rust `left_panel_view_accepts_prompt_debugger` test covers the Rust whitelist, but there's no matching TS test for the type + sanitizer path.
- `formatTraceAsText` still omits `has_images` / `has_tools` / `metadata`, so a downloaded trace isn't fully reproducible now that those inputs affect routing. Worth including them.

SlawomirNowaczyk (approved):
LGTM

fl0rianr (commented):
Thanks for the latest updates — the previous issues around stale validation results, settings persistence, classifier-condition rendering, unknown `matched_rule`, trace completeness, accessibility, and the stuck “Validating…” state now look resolved.

I found one remaining robustness issue before merge:

## Invalid request field types are not handled cleanly

In `handle_routing_validate`, these fields are read before the local `try` block:

```cpp
const std::string prompt = request_json.value("prompt", std::string());
const bool has_images = request_json.value("has_images", false);
const bool has_tools = request_json.value("has_tools", false);
```

If a client sends an invalid type, for example:

```json
{
  "prompt": 123,
  "has_tools": "yes"
}
```

`nlohmann::json::value<T>()` can throw a type error before the existing exception handler is reached. That means malformed input may result in a generic server error instead of a clear `400` response.

Please validate these fields explicitly before reading them:

* `prompt` must be a string
* `has_images` must be a boolean
* `has_tools` must be a boolean

It would also be good to add endpoint tests for invalid types, similar to the existing malformed-metadata tests.

One minor UX point: changing the prompt, image, tools, or metadata after a successful validation leaves the old result visible. The stored result is now internally consistent, so this is not a correctness blocker, but clearing the result on input changes or showing the submitted inputs alongside it would make the state less ambiguous.

Apart from the request-type validation, the implementation now looks in good shape, then CI needs to be green as well.

QuMuzafferEge:
Resolved [fl0rianr]'s comments.

fl0rianr (commented):
Thanks for the latest update. Most of the previously raised issues now look resolved, but I found two remaining problems before merge:

## 1. Duplicate import breaks the frontend build

In `ModelManager.tsx`, `RECIPE_DISPLAY_NAMES` is imported twice:

```ts
import { RECIPE_DISPLAY_NAMES } from './utils/recipeNames';
import { COLLECTION_ROUTER_MODEL_RECIPE, RECIPE_DISPLAY_NAMES } from './utils/recipeNames';
```

This declares the same local identifier twice and should fail TypeScript compilation. Please remove the newly added standalone import and keep only the existing combined import.

## 2. Classifier trace details are dropped in the debugger UI

The backend trace response can include both `label` and `rationale`, but the frontend `TraceEntry` type only declares:

```ts
condition
score
result
```

The visible trace and downloaded trace also omit `label` and `rationale`.

This is especially relevant for classifier and LLM-router policies, where the label identifies the evaluated class and `rationale` contains the router model’s selection reasoning. Omitting these fields means the Prompt Debugger does not fully show the reasoning behind the final route.

Please add the optional fields to `TraceEntry` and include them in both:

* the visible trace log
* the downloaded trace file

After these two fixes and green CI, the implementation looks ready from my side.

QuMuzafferEge:
Latest commit deals with [fl0rianr]'s comments.

fl0rianr (commented):
Thanks for the latest fixes. The previously reported issues around the duplicate import, request-field validation, trace details, settings persistence, stale results, and the stuck validation state now look resolved.

I found three remaining issues before merge:

## 1. routing.router is already supported, but the new test and decision-tree UI treat it as unsupported

`test_021zm_routing_validate_unimplemented_classifier_returns_400` describes an LLM router as “not yet implemented” and expects HTTP 400.

However, the current parser already desugars `routing.router` into an `llm` classifier and synthetic identity rules such as `__route_0`. If the router model cannot run, classifier failures are absorbed by the routing engine and fail open to `default_model`, so the endpoint should still return a decision with HTTP 200 rather than the expected 400. This test therefore appears inconsistent with the current routing implementation and should fail once CI is allowed to run.

There is also a related UI problem: after a successful LLM-router decision, `matched_rule` names one of the server-generated `__route_N` rules. The client retains the original uploaded policy, which contains `routing.router` but no explicit `routing.rules`. `findMatchedRuleIndex()` therefore cannot find the synthetic rule, and `handleViewDecisionTree()` refuses to display the tree.

Please either return the normalized/desugared policy (or a dedicated evaluated-tree representation) from `/routing/validate`, or add explicit client-side handling for `routing.router`. The test should be updated to exercise the supported LLM-router behavior instead of expecting a 400.

## 2. The validator disables collection-component consistency checks

`handle_routing_validate` sets:

```cpp
options.require_declared_components = false;
```

This does more than avoid resolving models against the live registry. It also allows candidates, default models, rule targets, and classifier models that are not listed in `collection.components`, and even allows the components array to be omitted.

As a result, the debugger can report successful validation for a policy that the normal collection registration path would reject.

The identity `resolve_component` callback is sufficient to avoid requiring live registry resolution. Please leave `require_declared_components` enabled so the uploaded document is still internally valid, and add a test covering an undeclared candidate or classifier model.

## 3. Implicit-AND leaves are still summarized incompletely

The routing parser permits multiple conditions in one leaf object and compiles them as an implicit `all`. For example:

```json
{
  "min_chars": 4000,
  "has_tools": true
}
```

But the generic branch of `summarizeMatchExpr()` only renders `Object.entries(obj)[0]`. The tree, tooltip, and accessible label therefore show only one of the conditions even though both determine whether the rule matches.

Please format every condition in a generic leaf and join them with an explicit `AND`. A small unit test for a multi-condition deterministic leaf would prevent this from regressing.

## Scope 

The linked issue asks for this functionality directly inside the Router Builder, using the configuration currently being authored. This remains a separate view requiring a manually uploaded JSON file. Please confirm that this UX change has been agreed with the issue owner, or integrate it with the Router Builder as requested.

QuMuzafferEge:
In addition to the three changes requested above, the Prompt Debugger is now moved to a separate tab in create router menu, which tests the policy defined in the policy builder. If the policy requires a model to be downloaded, it is downloaded as part of the validation (the downloader popup works, but it is visible only when one leaves the builder menu). If the policy changes during validation, the ongoing validation is dropped immediately.

fl0rianr (commented):
Thanks for the latest update. The backend changes look much better, but I still see three frontend issues that should be fixed before merge:

1. **The current CI failure is legitimate.**
   `decisionTree.tsx` uses the global `JSX.Element` / `JSX.Element[]` types. With the current React 19 typings and strict TypeScript configuration, this breaks the app type-check. Please use `React.ReactElement` or import `ReactElement` explicitly instead.

2. **Validation state has a race condition.**
   When the policy changes during request A, the effect increments `latestRequestIdRef` and immediately resets `isValidating`, allowing request B to start. When stale request A later reaches its unconditional `finally`, it calls `setIsValidating(false)` and can incorrectly clear B’s active state. The reset in `finally` should remain conditional on the request still being the latest one.

3. **Metadata is not snapshotted with the other submitted inputs.**
   Prompt, policy, image state, and tool state are captured before model preparation, but `metadataText` is parsed only after the awaited model downloads/loads. Editing metadata during that time can therefore change the request that is eventually sent, or even silently turn invalid metadata into an empty object. Please parse and snapshot metadata synchronously before the first `await`.

Minor accessibility point: the custom always-mounted tab panels should retain the `id`, `role="tabpanel"`, and `aria-labelledby` attributes provided by the normal `Tabs.Contents` implementation.

After these fixes and green CI, this should be in good shape.

QuMuzafferEge:
Addressed the latest comments

fl0rianr (commented):
Thanks for the latest update. The previous issues around `JSX.Element`, request invalidation, metadata snapshots, classifier summaries, accessibility, and the Router Builder integration are now addressed well.

I still see two correctness issues that should be resolved before merge:

1. **The debugger loads routing models with the wrong residency purpose.**

The production router loads router and classifier models through:

```cpp
auto_load_model_if_needed(
    model,
    json::object(),
    LoadPurpose::RoutingDependency
);
```

The new `/routing/validate` path currently calls `auto_load_model_if_needed(model)` with the default `UserInference` purpose. The frontend also preloads these models through the normal `/load` endpoint.

This means the debugger uses the standard model pool instead of the dedicated routing-helper residency pool. Under slot limits, pinned models, or exclusive backends, the debugger can therefore fail or evict models differently from the real router.

Please make the validation path use `LoadPurpose::RoutingDependency` and avoid preloading the same models as normal user-inference models in the frontend.

2. **Existing policies with unsupported conditions can be tested incorrectly.**

When an existing policy contains conditions the Builder cannot represent, such as metadata conditions, those conditions are dropped while reconstructing the editable draft and only recorded in `lossyRuleIds`.

The Test Prompt tab then builds and validates a policy from this reduced draft. As a result, it may display a routing decision for a policy that differs from the policy currently stored on the server.

Please either block testing while `lossyRuleIds` is non-empty, clearly warn that the reduced draft is being tested, or preserve and validate the original policy until the user explicitly accepts the lossy conversion.

Two smaller follow-ups:

* Clear the existing result or mark it as outdated when the policy, prompt, flags, or metadata change.
* Re-check the metadata parse error directly inside `handleValidate()` instead of relying only on the asynchronously updated `metadataError` state.

QuMuzafferEge:
Latest comments addressed

QuMuzafferEge:
I investigated the the failed test on semantic embedding classifier routing, it seems the following code on auto_tune.h is the culprit:

// Embedding models need at least EMBEDDING_CTX_SIZE
if (is_embedding && ctx_size < EMBEDDING_CTX_SIZE) {
    ctx_size = EMBEDDING_CTX_SIZE;   // hardcoded 8192, auto_tune.h:15
    
 This sets the context window to the hardcoded 8192, which can exceed the model's own context size, which apparently can lead to unexpected scores.
 
 This code was introduced in the "Automatically determine context size" commit on 2026-06-16. So I don't know how this test error was not caught before. This is out of the scope of this issue, so I'm not touching it. Just FYI.

fl0rianr:
I agree this test fail is not your fault, a flaky or maybe an always failing test error. Made an PR for it: https://github.com/lemonade-sdk/lemonade/pull/2824

fl0rianr (commented):
Thanks for addressing the previous review points. The lossy-policy guard, synchronous metadata validation, outdated-result warning, and `RoutingDependency` handling inside `/routing/validate` are good improvements.

I still see one blocking issue and two smaller follow-ups:

## 1. Public `/load` can bypass normal residency limits

The general `/load` endpoint now accepts a client-controlled:

```json
{
  "load_purpose": "routing_dependency"
}
```

This assigns arbitrary models to the `RoutingHelper` residency class. Since distinct helper models use separate residency pools, an external client can use this field to bypass the normal `max_loaded_models` limit.

`LoadPurpose` should remain an internal server-side decision rather than a public `/load` option.

Please remove `load_purpose` from the public `/load` request handling. `/routing/validate` already loads the actually required classifier models correctly through:

```cpp
auto_load_model_if_needed(
    model,
    json::object(),
    LoadPurpose::RoutingDependency
);
```

## 2. The frontend eagerly loads every classifier

Before calling `/routing/validate`, the panel runs `ensureModelReady()` for every classifier declared by the policy.

The production routing path loads classifiers lazily only when their condition is actually evaluated. Eagerly loading all of them can:

* download and start models that an earlier matching rule would never use;
* introduce residency or exclusive-backend conflicts that a real request would not encounter;
* make validation significantly slower;
* leave more helper models resident than the tested route requires.

Please remove the frontend loading loop and let `/routing/validate` perform the same lazy loading as the real router.

## Smaller follow-ups

* An invalid metadata edit can still fail to mark an old result as outdated when both the previous parsed metadata and the parse-error fallback are `{}`. Include the metadata parse error itself in `isResultOutdated`.
* `showError` is still declared and destructured by `RouterTestPromptPanel`, but is unused.

CI is now green apart from the known incorrect error test, so this review is based on the current passing implementation rather than an unavailable CI run.

QuMuzafferEge:
Latest comments addressed

fl0rianr (approved):
Thanks, LGTM now, merging.

PR #2727, opened 2026-07-16
Wire collection.router classifiers to Router::classify + validate model capability at parse time

QuManu:
### Summary
Two related bugs in the `collection.router` + `onnxruntime` classifier feature (#2626):

1. **A `type: "classifier"` condition can never actually reach an `onnxruntime`-backed model.** `run_classifier` in `routing_classifier_services.cpp` only calls `Router::chat_completion`, which `WrappedServer`'s default implementation turns into an "unsupported capability" error for any backend that doesn't implement it (`OnnxRuntimeServer` doesn't — it only implements `classify()`). The failure is swallowed by the engine's fail-open design, so a policy with `on_error: match_true` (the natural choice for a security-relevant classifier) silently matches on *every* request regardless of input — worse than an outright error, since it looks like it's working.

2. **No capability check on `routing.classifiers[].model` at parse time.** `resolve_component`/`require_declared` only check that the referenced model exists, never that it can actually serve as a classifier (or, for `semantic_similarity`, as an embedder). A misconfigured policy registers successfully and only fails at request time via bug #1's failure mode.

### Fix
- Add optional `classify`/`get_model_type` callbacks to `ClassifierServices` (`routing_classifier_services.h/.cpp`, `routing_classifier_services_router.cpp`). `run_classifier` now calls `Router::classify()` when the target model is `ModelType::CLASSIFICATION`, falling back to the original chat-completion-based path otherwise — fully backward compatible, existing call sites and tests are unaffected by the new (defaulted) parameters.
- Add an optional `get_model_type` resolver to `RoutingPolicyParseOptions` (`routing_policy_parser.h/.cpp`), wired at both places a policy actually gets parsed in production (`ModelManager::build_cache()` and `validate_collection_request`). `type: "classifier"` now requires `ModelType::LLM` or `ModelType::CLASSIFICATION`; `type: "semantic_similarity"` requires `ModelType::EMBEDDING`.

### Verification
- All existing `test_routing_policy_parser` and `test_routing_classifier_services` unit tests pass unchanged.
- Live, end-to-end, against a build of current `main`:
  - A `collection.router` policy using `Bert-Phishing-ONNX` as classifier: phishing input → real score (`0.9999`), routes to the restricted candidate; benign input → real score (`0.00003`), falls through to default. (Before the fix: both inputs routed identically regardless of content.)
  - Same result reproduced with `meta-llama/Llama-Prompt-Guard-2-86M` (ONNX export) as the classifier — real prompt-injection/jailbreak detection, correct routing divergence between malicious and benign input.
  - Registering a policy whose classifier `model` is an embedding model is now rejected immediately with a clear error instead of silently succeeding.

### Context
This was requested as a follow-up PR by @SlawomirNowaczyk on #2626's review thread rather than being bundled into that PR.

### DIFF
--- a/src/cpp/include/lemon/backends/kokoro/kokoro.h
+++ b/src/cpp/include/lemon/backends/kokoro/kokoro.h
@@ -27,7 +27,9 @@ inline const BackendDescriptor descriptor = {
         {"cpu", {"windows", "linux"}, {{"cpu", {"x86_64"}}}, "x86_64 CPU"},
         {"metal", {"macos"}, {{"metal", {}}}, "Apple Silicon GPU"},
     },
-    /*default_labels*/  {},  // kokoro models carry "tts" explicitly in server_models.json
+    /*default_labels*/  {"tts"},  // kokoro only does TTS; declare it so a label-less
+                                   // user model is typed TTS, not LLM (catalog models
+                                   // also carry "tts" explicitly in server_models.json)
     /*required_checkpoints*/ {"main"},
     /*modality*/        "Text-to-speech",
     /*experimental*/    false,
--- a/src/cpp/include/lemon/routing_classifier_services.h
+++ b/src/cpp/include/lemon/routing_classifier_services.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lemon/model_types.h"
 #include "lemon/routing_policy.h"
 
 #include <functional>
@@ -13,17 +14,26 @@ class Router;
 
 using EnsureClassifierModelLoaded = std::function<void(const std::string& model)>;
 using RouterJsonCall = std::function<json(const json& request)>;
+using RouterModelTypeCall = std::function<ModelType(const std::string& model)>;
 
 ClassifierServices make_router_classifier_services(
     Router& router,
     EnsureClassifierModelLoaded ensure_loaded = {});
 
 // Testable adapter seam: production binds these calls to Router::embeddings and
-// Router::chat_completion; unit tests bind them to fake Router-like functions.
+// Router::chat_completion (plus Router::classify/get_model_type for models
+// tagged ModelType::CLASSIFICATION); unit tests bind them to fake Router-like
+// functions. `classify`/`get_model_type` default to empty so existing
+// embeddings+chat_completion-only call sites keep compiling unchanged — a
+// `run_classifier` call against a model with no `get_model_type` configured,
+// or one that isn't ModelType::CLASSIFICATION, falls back to the original
+// chat-completion-based classification.
 ClassifierServices make_classifier_services_from_router_calls(
     RouterJsonCall embeddings,
     RouterJsonCall chat_completion,
-    EnsureClassifierModelLoaded ensure_loaded = {});
+    EnsureClassifierModelLoaded ensure_loaded = {},
+    RouterJsonCall classify = {},
+    RouterModelTypeCall get_model_type = {});
 
 std::vector<float> parse_embedding_vector(const json& response);
 std::map<std::string, double> parse_classifier_scores(const json& response);
--- a/src/cpp/include/lemon/routing_policy_parser.h
+++ b/src/cpp/include/lemon/routing_policy_parser.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lemon/model_types.h"
 #include "lemon/routing_policy.h"
 
 #include <functional>
@@ -16,8 +17,15 @@ namespace lemon {
 using RoutingComponentResolver =
     std::function<std::optional<std::string>(const std::string& component)>;
 
+// Looks up a resolved component's deployment ModelType. Server integration
+// binds this to the model registry (ModelInfo::type); pure parser tests leave
+// it unset, which skips the capability check entirely (matching prior
+// behavior — this option is additive and opt-in).
+using RoutingModelTypeResolver = std::function<ModelType(const std::string& resolved_model)>;
+
 struct RoutingPolicyParseOptions {
     RoutingComponentResolver resolve_component;
+    RoutingModelTypeResolver get_model_type;
     bool require_declared_components = true;
 };
 
--- a/src/cpp/include/lemon/routing_policy_parser.h
+++ b/src/cpp/include/lemon/routing_policy_parser.h
@@ -21,7 +21,14 @@ using RoutingComponentResolver =
 // binds this to the model registry (ModelInfo::type); pure parser tests leave
 // it unset, which skips the capability check entirely (matching prior
 // behavior — this option is additive and opt-in).
-using RoutingModelTypeResolver = std::function<ModelType(const std::string& resolved_model)>;
+//
+// Returns nullopt when the type genuinely cannot be established (e.g. an
+// inline collection component with no matching `models[]` definition to fall
+// back on) — the parser treats that as a hard error rather than guessing a
+// type, since guessing wrong is worse than not checking at all: it can both
+// reject valid configs and silently accept invalid ones.
+using RoutingModelTypeResolver =
+    std::function<std::optional<ModelType>(const std::string& resolved_model)>;
 
 struct RoutingPolicyParseOptions {
     RoutingComponentResolver resolve_component;
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -2550,6 +2550,26 @@ size_t ModelManager::count_cloud_models(const std::string& provider) const {
                          });
 }
 
+// The label set a user or inline model definition normalizes to. Kept in one
+// place so model registration (register_user_model) and collection.router
+// capability validation (validate_collection_request) derive the same type for
+// the same definition: explicit labels + legacy capability flags + the backend
+// descriptor's default labels (e.g. sd-cpp -> "image", whispercpp -> "transcription").
+static std::set<std::string> normalized_definition_labels(const json& model_data) {
+    std::set<std::string> labels = {"custom"};
+    std::vector<std::string> extra = model_data.value("labels", std::vector<std::string>{});
+    labels.insert(extra.begin(), extra.end());
+    if (model_data.value("reasoning", false)) labels.insert("reasoning");
+    if (model_data.value("vision", false)) labels.insert("vision");
+    if (model_data.value("embedding", false)) labels.insert("embeddings");
+    if (model_data.value("reranking", false)) labels.insert("reranking");
+    if (const auto* desc =
+            lemon::backends::descriptor_for(model_data.value("recipe", std::string()))) {
+        for (const auto& label : desc->default_labels) labels.insert(label);
+    }
+    return labels;
+}
+
 void ModelManager::register_user_model(const std::string& model_name,
                                       const json& model_data,
                                       const std::string& source) {
@@ -2566,35 +2586,11 @@ void ModelManager::register_user_model(const std::string& model_name,
             model_entry[prop] = model_data[prop];
         }
     }
-    std::set<std::string> labels = {"custom"};
-    std::vector<std::string> extra_labels = model_data.value("labels", std::vector<std::string>{});
-    labels.insert(extra_labels.begin(), extra_labels.end());
-
-    // legacy label format
-    if (model_data.value("reasoning", false)) {
-        labels.insert("reasoning");
-    }
-    if (model_data.value("vision", false)) {
-        labels.insert("vision");
-    }
-    if (model_data.value("embedding", false)) {
-        labels.insert("embeddings");
-    }
-    if (model_data.value("reranking", false)) {
-        labels.insert("reranking");
-    }
+    std::set<std::string> labels = normalized_definition_labels(model_data);
 
     // `recipe` already copied into `model_entry` by the USER_DEFINED_MODEL_PROPS
-    // loop above; this local is just for the label inference below.
+    // loop above; this local is just for the collection handling below.
     std::string recipe = model_data.value("recipe", "");
-
-    // Inject the backend's default labels for models that omit them (e.g. sd-cpp
-    // -> image, whispercpp/moonshine -> transcription).
-    if (const auto* desc = lemon::backends::descriptor_for(recipe)) {
-        for (const auto& label : desc->default_labels) {
-            labels.insert(label);
-        }
-    }
 
     model_entry["labels"] = labels;
     model_entry["suggested"] = true; // Always set suggested=true for user models
@@ -4566,13 +4562,14 @@ std::optional<std::string> ModelManager::validate_collection_request(
                 if (!def) {
                     return std::nullopt;
                 }
-                std::vector<std::string> labels;
-                if (def->contains("labels") && (*def)["labels"].is_array()) {
-                    for (const auto& label : (*def)["labels"]) {
-                        if (label.is_string()) labels.push_back(label.get<std::string>());
-                    }
-                }
-                return get_model_type_from_labels(labels);
+                // Derive the type exactly as register_user_model() would once
+                // this inline definition is registered — explicit labels, legacy
+                // capability flags, and the backend's default labels — so
+                // validation and registration cannot disagree (e.g. a label-less
+                // sd-cpp model is IMAGE, not LLM).
+                std::set<std::string> label_set = normalized_definition_labels(*def);
+                return get_model_type_from_labels(
+                    std::vector<std::string>(label_set.begin(), label_set.end()));
             }
         };
         try {
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1932,6 +1932,13 @@ void ModelManager::build_cache() {
         auto it = public_model_aliases_.find(name);
         return it != public_model_aliases_.end() ? it->second : name;
     };
+    // Direct lookup into the map already being built, same rationale as
+    // resolve_component above: models_cache_ is populated and the lock is
+    // still held, so this avoids re-entering get_model_info()'s own locking.
+    policy_options.get_model_type = [this](const std::string& name) -> ModelType {
+        auto it = models_cache_.find(name);
+        return it != models_cache_.end() ? it->second.type : ModelType::LLM;
+    };
     for (auto& [name, info] : models_cache_) {
         if (!is_router_collection_recipe(info.recipe)) {
             continue;
@@ -4544,6 +4551,15 @@ std::optional<std::string> ModelManager::validate_collection_request(
                 // names stable so parser component-role validation can still
                 // compare them against the authored components array.
                 return name;
+            }
+        };
+        options.get_model_type = [this](const std::string& name) -> ModelType {
+            try {
+                return get_model_info(name).type;
+            } catch (...) {
+                // Unregistered inline-collection component: no type to check
+                // yet, so don't block registration on it here.
+                return ModelType::LLM;
             }
         };
         try {
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -135,7 +135,25 @@ static ModelType get_deployment_model_type(const std::string& recipe,
             }
         }
     }
-    return get_model_type_from_labels(labels);
+    ModelType type = get_model_type_from_labels(labels);
+
+    // Reaching here means the backend declares no definitive non-LLM deployment,
+    // i.e. it is a chat/general backend (llamacpp/flm/ryzenai/vllm/cloud) — none
+    // of which implement IClassificationServer (only onnxruntime does, and it
+    // returned CLASSIFICATION above via its default label). So a `classification`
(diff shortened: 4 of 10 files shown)


ramkrishna2910 (commented):
Thanks for picking this up — this closes the exact gap Slawomir flagged on #2626 (the `classifier` condition never reaching the onnxruntime backend, silently fail-open). I traced the full path and the approach is right:

- `run_classifier` routing a `ModelType::CLASSIFICATION` model to `Router::classify()` and falling back to the chat path otherwise, with the new callbacks defaulted empty, is genuinely backward-compatible — existing call sites/tests are untouched.
- The reuse of `parse_classifier_scores` (unchanged) for the `/classify` response is correct: it tries chat-text first, then `parse_scores_payload`, which reads the `labels` key — exactly the `/v1/classify` envelope. Good.
- Parse-time capability validation (`classifier` → LLM/CLASSIFICATION, `semantic_similarity` → EMBEDDING) is sound, and skipping it when `get_model_type` is unset keeps the pure-parser tests green.

Two asks before merge:

1. **A unit test for the new branch.** The change is important and currently rests on "existing tests still pass" + manual e2e. A test that binds a fake `classify` + `get_model_type` and asserts (a) a `ModelType::CLASSIFICATION` model goes through the classify path (not chat), and (b) a policy whose `classifier.model` is an EMBEDDING model is rejected at parse time, would lock the behavior in.
2. **Rebase onto `main`** — it's currently BEHIND, so CI hasn't run on a current base.

Nit (non-blocking): `ensure_model` now runs before the `chat_completion` null-check; harmless in production (chat_completion is always set) but a hair different in tests that expect a throw-before-load.

I'm tagging this to the **Lemonade Router** milestone — it belongs to the classifier path (#2592/#2384), and it directly unblocks a real `classifier`-condition case in the router e2e (#2388 / `test/server_router.py`), which can then serve as the integration coverage.

fl0rianr (changes requested):
Good work! But found an issue:

**Blocking: `ModelType::LLM` cannot also represent “type unknown”.**

The new type resolver currently manufactures `ModelType::LLM` whenever a component cannot be found:

```cpp
catch (...) {
    return ModelType::LLM;
}
```

That creates asymmetric and registration-order-dependent validation for inline collections.

`validate_collection_request()` explicitly permits components that are defined in the request's `models[]` array but are not registered yet. For such a component, `get_model_info()` fails and the resolver reports `LLM`, regardless of the inline definition's actual labels/recipe.

Consequently:

* A valid inline `semantic_similarity` policy using an inline embedding model is rejected as “type llm; needs embedding”.
* An invalid inline `classifier` policy using an inline embedding, image, or other non-classifier model is accepted because `LLM` is an allowed classifier type.
* The `build_cache()` resolver repeats the same fallback for any component missing from the completed cache, so a declared-but-unresolvable classifier component can again pass as an LLM.

This undermines the intended fail-fast capability validation in both directions. It also means `catch (...)` can turn unrelated lookup failures into a successful LLM capability check.

Could the resolver represent absence explicitly, e.g.:

```cpp
using RoutingModelTypeResolver =
    std::function<std::optional<ModelType>(const std::string&)>;
```

Then:

1. During request validation, derive the type from the matching inline `models[]` definition when one exists, including its labels/backend defaults.
2. Return `nullopt` only when the type genuinely cannot be established, and handle that case explicitly rather than treating it as LLM.
3. During `build_cache()`, where the cache and alias map are already complete, treat a missing referenced model as an error rather than silently classifying it as an LLM.

A focused regression pair would expose the issue:

* inline `semantic_similarity` + inline model with `labels: ["embedding"]` must pass;
* inline `classifier` + that same embedding definition must fail.

QuManu:
Good catch, you're right. I changed the function so it can now say "I don't know the type" instead of just guessing LLM. Now it first checks if the model has an inline definition and reads its labels from there. Only if it truly can't find anything does it give up (and now that's treated as an error, not a silent pass). I also added the two tests you asked for. One where inline semantic_similarity + an embedding model works, and one where inline classifier + that same embedding model fails. Fix is in 2b6d8d32, tests in 258aec44. Let me know if I missed anything.


> Good work! But found an issue:
> 
> **Blocking: `ModelType::LLM` cannot also represent “type unknown”.**
> 
> The new type resolver currently manufactures `ModelType::LLM` whenever a component cannot be found:
> 
> ```c++
> catch (...) {
>     return ModelType::LLM;
> }
> ```
> 
> That creates asymmetric and registration-order-dependent validation for inline collections.
> 
> `validate_collection_request()` explicitly permits components that are defined in the request's `models[]` array but are not registered yet. For such a component, `get_model_info()` fails and the resolver reports `LLM`, regardless of the inline definition's actual labels/recipe.
> 
> Consequently:
> 
> * A valid inline `semantic_similarity` policy using an inline embedding model is rejected as “type llm; needs embedding”.
> * An invalid inline `classifier` policy using an inline embedding, image, or other non-classifier model is accepted because `LLM` is an allowed classifier type.
> * The `build_cache()` resolver repeats the same fallback for any component missing from the completed cache, so a declared-but-unresolvable classifier component can again pass as an LLM.
> 
> This undermines the intended fail-fast capability validation in both directions. It also means `catch (...)` can turn unrelated lookup failures into a successful LLM capability check.
> 
> Could the resolver represent absence explicitly, e.g.:
> 
> ```c++
> using RoutingModelTypeResolver =
>     std::function<std::optional<ModelType>(const std::string&)>;
> ```
> 
> Then:
> 
> 1. During request validation, derive the type from the matching inline `models[]` definition when one exists, including its labels/backend defaults.
> 2. Return `nullopt` only when the type genuinely cannot be established, and handle that case explicitly rather than treating it as LLM.
> 3. During `build_cache()`, where the cache and alias map are already complete, treat a missing referenced model as an error rather than silently classifying it as an LLM.
> 
> A focused regression pair would expose the issue:
> 
> * inline `semantic_similarity` + inline model with `labels: ["embedding"]` must pass;
> * inline `classifier` + that same embedding definition must fail.

QuManu:
Thanks @ramkrishna2910 for the review! I did all three things:

   - Added tests for the classify vs chat routing, and for the capability checks (including fl0rianr's test pair) 258aec44
   - Fixed the order so it still fails early when chat_completion isn't set up, before loading anything 83eb695f
   - Rebased on top of the latest main, no conflicts bbff6f23 / ff849d82
   
Also found a real bug while fixing fl0rianr's comment (the LLM fallback issue) fixed in 2b6d8d32

fl0rianr (changes requested):
Thanks for addressing the previous review and adding the requested tests. There are still two related blockers.

### 1. One of the current CI failures is caused by this PR, not unrelated
(C++ Server Build, Test, and Release 🚀 / Build Lemonade .deb Package)
`ModelManagerCollectionValidationTest` uses valid inline `llamacpp` component definitions containing a recipe and checkpoint but no `labels` field. Labels are not currently required for an inline component to be considered complete.

The new resolver only derives a type when a `labels` array is explicitly present:

```cpp
if (def && def->contains("labels") && (*def)["labels"].is_array()) {
    ...
    return get_model_type_from_labels(labels);
}
return std::nullopt;
```

Therefore a label-less inline `llamacpp` model becomes “unresolvable” during `validate_collection_request()`.

That disagrees with the normal model-cache path, where an empty labels vector is passed to `get_model_type_from_labels()` and resolves to `ModelType::LLM`.

This is why the shared valid fixture fails first and the four assertions report failures:

```text
routing.classifiers[0].model 'pii-detector' has an unresolvable type
```

Please make inline type derivation mirror the actual registration/cache behavior. For example, when an inline definition exists, collect its labels if present—allowing the vector to remain empty—and always call `get_model_type_from_labels(labels)`:

```cpp
const json* def = find_inline_def(bare_component_name(name));
if (def) {
    std::vector<std::string> labels;
    if (def->contains("labels") && (*def)["labels"].is_array()) {
        for (const auto& label : (*def)["labels"]) {
            if (label.is_string()) {
                labels.push_back(label.get<std::string>());
            }
        }
    }
    return get_model_type_from_labels(labels);
}
return std::nullopt;
```

Preferably, centralize this conversion so request validation and `build_cache()` cannot drift again.

Merely adding labels to the existing test fixture would hide the production inconsistency rather than fix it.

### 2. Cold-start classification still selects the endpoint before auto-load

`run_classifier` currently evaluates `get_model_type(model)` before invoking `ensure_model()`:

```cpp
const bool use_classify_endpoint =
    classify_call && *classify_call && get_model_type &&
    get_model_type(model) == ModelType::CLASSIFICATION;

ensure_model(ensure_loaded, model);
```

The production callback uses `Router::get_model_type()`, which reports `ModelType::LLM` when the backend is not already loaded and alive.

Consequently, the first request after startup, eviction, or backend restart selects the chat path before loading the ONNX classifier. The model is then loaded, but the already-selected `chat_completion` path is still used, recreating the original unsupported-capability / `on_error` behavior.

Please load the model before resolving its runtime type, or use a registry-backed resolver that does not depend on a live backend.

A regression test should simulate:

1. `get_model_type()` returns `LLM` before `ensure_loaded`;
2. `ensure_loaded` changes the state;
3. `get_model_type()` then returns `CLASSIFICATION`;
4. the request uses `classify`, not `chat_completion`.

Both issues need addressing before approval. Thanks!

ramkrishna2910:
Pushed a fix for the second blocker (and updated the branch) to help this land, since it's gating downstream router work — hope that's alright @QuManu.

**Blocker 2 (cold-start ordering) — fixed** (`e01816a9`). `run_classifier` now calls `ensure_model()` **before** resolving `get_model_type()`, and the `chat_completion`-configured guard moved onto the chat branch. Rationale is exactly as @fl0rianr described: `Router::get_model_type()` returns `ModelType::LLM` until the backend is loaded and alive, so resolving first made the first request after startup/eviction/restart take the chat path for a classify-only model.

Added the requested regression test (`test_run_classifier_loads_before_resolving_type`): the resolver returns `LLM` before `ensure_loaded` and `CLASSIFICATION` after. Verified it has teeth via negative control — it fails cleanly on the old ordering (`[FAIL] ... loads the model before resolving its type`) and passes now. Full `test_routing_classifier_services` suite green.

**Blocker 1 (label-less inline models)** — confirmed resolved by your `ba8289d3`: `test_model_manager_collection_validation` passes, including the previously-failing "valid collection.router request passes validation".

Also merged `main` in to clear the BEHIND status (clean merge, no conflicts; re-ran the classifier test against the updated base — still green).

@fl0rianr — both of your review-2 blockers should be addressed now; would you re-review when you get a chance?

fl0rianr (commented):
Thanks for addressing the previous blockers. The cold-start ordering and the `optional<ModelType>` handling now look correct.

There is still one blocking inconsistency in the inline-model type resolver.

`validate_collection_request()` derives an unregistered inline model's type only from its explicit `labels` array:

```cpp
std::vector<std::string> labels;
if (def->contains("labels") && (*def)["labels"].is_array()) {
    ...
}
return get_model_type_from_labels(labels);
```

This still does not mirror how the same definition is actually registered. `register_user_model()` also:

* converts legacy flags such as `embedding`, `reranking`, `vision`, and `reasoning` into labels;
* injects the backend descriptor's `default_labels`.

For example, an inline model with `recipe: "sd-cpp"` and no explicit labels is currently inferred as `LLM` during validation, so it is accepted as a `classifier`. During registration, however, the `sd-cpp` backend contributes its default `image` label, making the actual model type `IMAGE`. At runtime `run_classifier()` then selects the chat path for this non-classification model, recreating the unsupported-capability / `on_error` fail-open behavior this PR is intended to prevent.

The inverse also occurs with legacy definitions such as `embedding: true`: registration creates an embedding model, while validation sees an LLM and rejects a valid `semantic_similarity` configuration.

Please centralize inline model-type derivation so validation and registration use exactly the same normalization rules: explicit labels, legacy capability flags, and backend default labels.

The current `test_inline_component_type_regression_pair()` does not cover this path because it supplies a fake resolver that already returns the desired type. Please add a `ModelManagerCollectionValidationTest`, for example:

* label-less `sd-cpp` used as a classifier must fail;
* `llamacpp` with legacy `embedding: true` used for semantic similarity must pass;
* label-less regular `llamacpp` must continue to default to LLM.

This needs addressing before approval.

ramkrishna2910:
Thanks @fl0rianr — you're right, and it's a real gap. Fixed in `49409d12`.

`validate_collection_request()`'s inline resolver only read explicit `labels`, while `register_user_model()` also applies the legacy capability flags (`embedding`/`reranking`/`vision`/`reasoning`) and the backend descriptor's `default_labels`. So exactly as you described: a label-less `sd-cpp` model validated as `LLM` (accepted as a classifier) but registers as `IMAGE`, and a legacy `embedding: true` model was rejected for `semantic_similarity` though it registers as an embedder.

Extracted `normalized_definition_labels()` and call it from **both** `register_user_model()` and the validation resolver, so the two can't drift — explicit labels + legacy flags + backend default labels, in one place.

Added the three `ModelManagerCollectionValidationTest` cases you specified, exercising the **real** resolver (not a fake):
- label-less `sd-cpp` used as a classifier → **rejected** (default label `image` → IMAGE);
- `llamacpp` with legacy `embedding: true` used for `semantic_similarity` → **passes**;
- label-less regular `llamacpp` → still defaults to LLM, valid classifier.

Verified they have teeth via negative control: with the old explicit-labels-only resolver, the sd-cpp and `embedding:true` cases both `[FAIL]`; with the shared normalization they pass. Full `test_model_manager_collection_validation` green.

fl0rianr (commented):
Thanks — the previous inline-normalization blocker is now properly addressed. Centralizing explicit labels, legacy flags, and backend default labels in `normalized_definition_labels()` is the right fix, and the new `ModelManagerCollectionValidationTest` coverage exercises the real production resolver.

There is still one blocking capability-bypass case.

The parser currently treats `ModelType::LLM` as proof that a model can serve the chat-based classifier path. However, `ModelType` is derived from labels, and `get_model_type_from_labels()` gives chat-indicator labels such as `reasoning`, `vision`, `tools`, and `chat-transcription` priority over every non-chat deployment label.

For example:

```json
{
  "model_name": "guard",
  "recipe": "onnxruntime",
  "checkpoint": "example/guard",
  "reasoning": true
}
```

Normalization produces both the backend's default `classification` label and the legacy `reasoning` label. Because chat indicators win, the resulting deployment type is `LLM`.

The consequences are:

1. parse-time validation accepts the model as a classifier because `LLM` is allowed;
2. registration stores the same normalized labels and therefore the same `LLM` type;
3. after loading, `run_classifier()` sees `LLM` and selects `chat_completion`;
4. `OnnxRuntimeServer` implements `IClassificationServer`, not a chat implementation, so the inherited chat method returns an unsupported-capability error.

That recreates the original runtime-error / `on_error: match_true` fail-open class.

The same issue can be exposed by extending the newly added sd-cpp test with `"vision": true`. The backend still only serves image generation, but `vision` wins during type resolution, changes the type to `LLM`, and makes the parser accept it as a classifier.

Please make this validation endpoint-capability-aware rather than relying solely on the label-derived `ModelType`. Possible approaches include using backend/recipe capabilities directly, rejecting labels that contradict a backend's deployment capability, or introducing a dedicated routing-capability resolver.

Please also add regression tests for:

* `onnxruntime` plus `reasoning: true`;
* `sd-cpp` plus `vision: true`;
* conflicting `classification` and chat-indicator labels.

This still needs addressing before approval.

ramkrishna2910:
Thanks @fl0rianr — you're right, and rather than fix just this instance I traced it to the root and swept the whole capability surface so we're not doing this one at a time.

## Root cause (fixed: `4a64e0d9`)

`get_model_type_from_labels()` gives chat-indicator labels (`reasoning`/`vision`/`tools`/`chat-transcription`) priority, and that type flows into **both** validation **and runtime** — `router.cpp:452` sets the live type from `model_info.type`. So a non-chat backend + a stray chat-indicator label typed as `LLM` everywhere, and `run_classifier` took the chat path on a backend that can't chat. Your `onnxruntime`+`reasoning` and `sd-cpp`+`vision` cases are two instances of the same bypass.

Fix: `get_deployment_model_type(recipe, labels)` — the backend descriptor's `default_labels` declare a definitive non-LLM deployment (`onnxruntime`→classification, `sd-cpp`→image, `whispercpp`→transcription, …), which **wins over chat-indicator labels** because those backends don't implement `chat_completion`. Applied at every `ModelInfo::type` site (runtime) and the inline validation resolver (parse-time). It only changes models with a contradictory chat-indicator on a non-chat backend, so real catalog models are untouched. Result: `onnxruntime`+`reasoning` → CLASSIFICATION (accepted, routes to `/classify`); `sd-cpp`+`vision` → IMAGE (rejected).

## The one instance the descriptor fix missed (fixed: `18a3369a`)

While sweeping, I found `kokoro` is the **only** can't-chat backend with empty `default_labels`, so a label-less kokoro model still slipped through as LLM. Gave the descriptor its true `"tts"` default label. (Catalog kokoro models already carry `tts` explicitly, so they're unaffected.)

## Tests

`ModelManagerCollectionValidationTest` now covers, against the **real** resolver: `onnxruntime`+`reasoning` is CLASSIFICATION; `sd-cpp`+`vision` is IMAGE (rejected as classifier); label-less `kokoro` rejected; plus the earlier sd-cpp/embedding/llamacpp cases. Verified with a **negative control** — 3 of the 4 new capability checks `[FAIL]` without the fix. `test_routing_policy_parser`, `test_routing_classifier_services`, and a full `lemond` build are green.

## Everything else the sweep found (not fixing here — flagging so it's not a surprise)

I filed **#2748** for the deeper ones; none block this PR's capability check (which is correct for the resolvable cases), but two want your call:

1. **(design question for you)** `build_cache()`'s resolver has no inline fallback, so a classifier component that's **hardware-filtered** (no NPU / not downloaded) is absent from `models_cache_` → "unresolvable type" → the **whole** policy is dropped, deterministic rules included. That's a consequence of the error-on-unresolvable behavior from your review 1. Should a filtered/unavailable classifier component **degrade gracefully** (its conditions fail-open per `on_error`, the rest of the policy still loads) instead of taking down the router? I lean graceful-degrade but it's your design.
2. Resolver asymmetry: registry-backed (pointer-only) collections skip `validate_collection_request` and are validated only on that weak build_cache path.
3. Silent fail-open: a mis-typed classifier becomes a no-op (all traffic → `default_model`) with no runtime signal — worth a telemetry/log flag.
4. Display-only: `ollama_api`/`cli` still type via `get_model_type_from_labels()` directly, so an onnxruntime/sd-cpp model can report a different type in `/api/tags` than at runtime.

Happy to take (1) in this PR if you want graceful-degrade, or leave it to #2748.

fl0rianr (commented):
Thanks for addressing the previous capability mismatch. The fix now correctly prevents non-chat backends such as onnxruntime, sd-cpp, and Kokoro from being promoted to LLM by stray chat-related labels.

One inverse case remains:

```json
{
  "recipe": "llamacpp",
  "labels": ["classification"]
}
```

Because llamacpp has no fixed non-LLM default label, this resolves to `ModelType::CLASSIFICATION`. The parser accepts it as a classifier, and `run_classifier()` then calls `Router::classify()`.

However, `LlamaCppServer` does not implement `IClassificationServer`, so this ends in an unsupported-capability response and can recreate the same `on_error: match_true` fail-open behavior this PR is intended to prevent.

Please add one focused regression test proving that:

```text
llamacpp + labels:["classification"]
```

is either rejected or remains on the LLM chat-classifier path, but never calls `Router::classify()`.

Once this case is covered  I consider the PR ready from my side.

ramkrishna2910:
Fixed (`da67a42c`) — and good catch, the inverse is real.

`/v1/classify` (`IClassificationServer`) is implemented by **exactly one** backend — onnxruntime (confirmed: it's the only `IClassificationServer` subclass and the only descriptor declaring `classification`). So once `get_deployment_model_type()` has fallen through to a backend with no definitive non-LLM default label (llamacpp/flm/ryzenai/vllm/cloud — all chat-capable, none classify-capable), a `classification` label is spurious. It now drops that claim, so the model stays **LLM** and is served via the chat-classifier path — never `Router::classify()`. onnxruntime is unaffected (it returns CLASSIFICATION from its default label before this point).

Regression test (real resolver): `llamacpp` + `labels:["classification"]` → typed **LLM**, and accepted as a classifier via the LLM chat path. The runtime half is already locked by `test_run_classifier_falls_back_to_chat_for_non_classification_model` (LLM → `chat_completion`, never `classify`), so with the type now LLM the two together prove it never reaches `Router::classify()`.

Verified with a negative control — the type check `[FAIL]`s without the fix. Full `test_model_manager_collection_validation` (15 cases) and a `lemond` build are green.

That's your fifth case covered. Thanks for the thorough passes — the capability model is a lot more solid for them.

SlawomirNowaczyk (approved):
LGTM now

fl0rianr (approved):
LGTM as well

QuManu:
CI is fully green now. The earlier Test whisper-metal failure was a flaky re-run, not a regression (passed clean on retry). @fl0rianr, does da67a42c look good from your side?

PR #2728, opened 2026-07-16
Router/2406 UI collection router collapse to single mode

SlawomirNowaczyk:
**Summary**

Replaces the mutually-exclusive "routing mode" radio (LLM *or* Rules) in the New/Edit Router modal with a stacked, fixed-flow layout of three stages — **Rules → LLM Router → Default** — so a single router can combine deterministic rules with an LLM fallback. Rules and LLM are independently toggleable; Default is always shown.

**Changes**
- **New `RouterStageSection` component** renders each stage with a step number, title, and an optional On/Off toggle; the body only mounts when the stage is open.
- **Draft model** drops the single `routingMode: 'llm' | 'rules'` field in favor of independent `rulesEnabled` / `llmEnabled` booleans, letting both run in one policy.
- **Validation** now requires at least one of Rules/LLM to be enabled, and only requires a router LLM when the LLM stage is on. Turning the LLM stage on the first time seeds a sensible default prompt.
- **CSS fix:** `.router-collection-modal .settings-content > *` is pinned to `flex-shrink: 0`. The content is a flex column, so its children were compressing when a stage expanded; combined with `overflow: hidden` on `.router-stage`, that clipped the section body. Now the panel scrolls instead of squashing, so each stage is fully visible and editable.
- **Tests:** routerCollections.test.cjs updated for the enabled-flags model and the combined rules+LLM desugaring.

**Why**

The old radio forced authors to pick one strategy; the hybrid stacked flow matches the L0(a) router model (rules first, LLM as the smart fallback, then a hard default) and reads as an increasing-complexity progression.


meghsat:
Thanks for putting this together! Since this is a draft feature so only sharing my thoughts on the UX. There are two ideas in it I really like:

1. Per-rule graph escalation: Letting a single rule expand from the flat chip row into a graph editor is a better boundary than a global Quick -> Advanced switch. I know we already have an "Expand to Advanced" option within Quick Rules. Along with that, do you think adding this rule-level canvas escalation in PR #2693  would make the UX better?

2. Classifier templates. The PII Detector and Topic Router one-click starters are great onboarding.

On the other hand have a few concerns:
1. The 1/2/3 metaphor is ambiguous. Numbered steps and stacking read as a step by step flow. While the On/Off toggles say the stages are independent, a first-time user seeing both toggles off can't tell whether to do them in order, pick one, or skip both.
2. Small concern: naming the choices by intent: Natural Language / Quick Rules / Advanced Rules tells the user which one is for them; "Rules / LLM Router" requires already understanding the routing architecture.
3. The default model section should follow the candidate model design.

SlawomirNowaczyk:
> Since this is a draft feature so only sharing my thoughts on the UX. There are two ideas in it I really like:

Thanks for the comments.

Yes, I think pre-rule graph escalation is an improvement without any downsides. My suggestion would be to merge https://github.com/lemonade-sdk/lemonade/pull/2693 as is, but make this change before next release.

And same thing with classifier templates, let's make a separate PR about that on top of https://github.com/lemonade-sdk/lemonade/pull/2693 and make sure it lands before the release.
 
> On the other hand have a few concerns:
> 
> 1. The 1/2/3 metaphor is ambiguous. Numbered steps and stacking read as a step by step flow. While the On/Off toggles say the stages are independent, a first-time user seeing both toggles off can't tell whether to do them in order, pick one, or skip both.

I agree that the visual aspect is still not good enough.
Would it be better to skip the 1/2/3 numbers, and start with "Rules" area completely hidden? Or clearly marked "advanced"?

Maybe a short instruction text would be good enough? Something like "We recommend starting by describing in natural language how the routing should be done. For advanced users who want more control, expand the Rules interface to get more control"?

> 2. Small concern: naming the choices by intent: Natural Language / Quick Rules / Advanced Rules tells the user which one is for them; "Rules / LLM Router" requires already understanding the routing architecture.

Yes and no: the triplet of "Natural Language / Quick Rules / Advanced Rules" actually creates an impression -- false, in my opinion -- that there are *separate* routing approaches, and the use must make a choice between them. This immediately makes them worry whether they are making the right choice, since it's not obvious if/how they can later switch.

I believe presenting the routing as a single, coherent flow is much stronger.
You are correct that this PR is still visually not good enough -- but I'm convinced the logic is strong.

> 3. The default model section should follow the candidate model design.

Good catch, I've now pushed a fix for that.

ramkrishna2910:
Thanks @SlawomirNowaczyk — and @meghsat for the UX pass. Two process calls from my side:

1. **gui2 feature freeze.** GUI3 is landing shortly and we're not taking feature additions to the current app. Please hold this against the freeze and retarget/reland it on GUI3 once that merges to main. The design discussion should absolutely continue here in the meantime — better for the GUI3 version to start from an agreed UX than to re-litigate it after a rebase.

2. **One canonical PR.** This and #2717 change the identical set of files — if this supersedes #2717 (it looks like the newer iteration), please close #2717 so review effort lands in one place.

On the UX debate itself: I'm with Slawomir that a single coherent flow beats three seemingly-exclusive modes, and with @meghsat that the numbered 1/2/3 stacking miscommunicates independence — the 'start from natural language, expand Rules as an advanced surface' framing proposed above seems like the right synthesis. Let's converge on that here so the GUI3 PR is mostly mechanical.

Going forward, router UI changes should have @meghsat's review as a standing requirement — flagging that here so it's explicit.

PR #2729, opened 2026-07-16
fix(router): isolate routing helpers from standard model LRU slots

### LINKED ISSUE #2725: Router-collection model routing thrashes model slots under the default `max_loaded_models=1`
### Summary

With router collections (`collection.router`, #2385) now live, the default slot configuration causes the router model and the selected candidate model to repeatedly evict each other — once per request. Because both are `ModelType::LLM` and share a single per-type LRU slot, a single L0a-style routed request pays two backend cold-starts, making the default routed path effectively unusable. We need explicit slot/lifetime semantics for classifier/router models rather than relying on the general LLM LRU cache.

### Background: how slot accounting works today

- `max_loaded_models` defaults to **1** (README.md) and is enforced **per model type**, each type keeping its own LRU cache (multi-model.md).
- In router.cpp, after slot-policy handling, the LRU check counts live servers of the same `ModelType` via `count_servers_by_type()` and, when `current_count >= max_models`, evicts the LRU server of that type via `find_lru_server_by_type()`.
- `SlotPolicy` (backend_descriptor.h) only distinguishes `Standard`, `ExclusiveNpu`, `CoexistByType`, and `Unmetered`. A local router model and a local candidate are both `Standard` / `ModelType::LLM`, so they compete for the **same** single LLM slot.
- Router dispatch (server.cpp) invokes the router model through `ClassifierServices`, whose `ensure_loaded` seam calls `auto_load_model_if_needed()`. After a decision is returned, the request's `model` is rewritten to the selected candidate, which then hits `auto_load_model_if_needed()` again on the normal completion path.

### The problem

Under the default `max_loaded_models=1`, a single routed request does:

1. Load router model (LLM slot now holds the router).
2. Run the routing decision.
3. Rewrite `model` → selected candidate; loading it **evicts the router** (LRU, same LLM slot).
4. Generate the response.
5. Next request: loading the router **evicts the candidate**, and step 3 evicts the router again.

So every routed request incurs two backend cold-starts. Backend startup can take tens of seconds, so this is not a minor inefficiency — the default routed path becomes practically unusable.

Pinning is not a real workaround and introduces its own failures:
- A pinned candidate can leave no evictable LRU server, so loading the router throws `SlotsPinnedException` (router.cpp).
- A pinned router symmetrically blocks loading the selected candidate.

### Reproduction

1. Default config (`max_loaded_models: 1`).
2. Register a `collection.router` model whose router model and at least one candidate are both local LLMs.
3. Send two sequential `chat/completions` requests to the router alias.
4. Observe in the server log the alternating `"Slot limit reached for type llm, evicting LRU"` messages — router and candidate evict each other every request.

### Impact

- Default-config router collections are effectively unusable (double cold-start per request).
- Pinning to work around it creates load failures (`SlotsPinnedException`) instead.
- Simply raising the global default `max_loaded_models` is too broad: it changes memory behavior for **all** users and model types, not just router collections.

### Proposed solutions (for discussion)

**Option A — Separate classifier/router slot category.**
Give router/classifier models their own slot bucket so they don't compete with candidate LLMs. This could be a new `SlotPolicy` (e.g. `ClassifierSlot`) and/or a distinct `ModelType`, with its own small limit independent of `max_loaded_models`. The router model stays warm across requests while the candidate occupies the normal LLM slot.
- Pros: router stays resident, clean accounting, no global default change.
- Cons: adds a slot category and its own memory budget; needs NPU-exclusivity interaction rules.

**Option B — Ephemeral invocation path that bypasses slot accounting.**
Treat the routing decision call as a short-lived invocation that does not participate in normal LLM LRU accounting (analogous to how `Unmetered` recipes are skipped in `count_servers_by_type` / `find_lru_server_by_type`). The router model is loaded, used for the decision, and does not count against — or evict — the candidate slot.
- Pros: no extra persistent slot; smallest change to the default memory footprint.
- Cons: if the router model is unloaded after each decision, it still pays a cold-start per request unless combined with a keep-warm cache; needs a clear lifetime definition.

**Option C — Explicit, validated slot requirement.**
Require and validate that enough slots exist for router collections (e.g. reject or warn at load/dispatch time when `max_loaded_models` is too low for the router + candidate to coexist), and document the requirement.
- Pros: simplest; makes the constraint explicit and diagnosable.
- Cons: pushes configuration burden onto the user; the default experience still needs a sane behavior.

**Option D — Memory-based slot accounting instead of a fixed model count.**
Replace (or augment) the "number of models per type" limit with a **memory budget**: admit models until their combined footprint would exceed a VRAM/RAM threshold, rather than capping at a raw count. Since router/classifier models are expected to be small, a small router model and a much larger candidate would comfortably fit within the same budget and coexist, while two large models would still trigger eviction. This reuses the existing memory-pressure infrastructure — `GlobalVramMonitor` and the `auto_evict` / `auto_evict_threshold_pct` path in eviction_engine.cpp — so eviction becomes driven by actual footprint (`find_lru_server_by_type` picking victims only when the budget is exceeded) instead of `count_servers_by_type() >= max_models`.
- Pros: fixes the router-vs-candidate case naturally (small router + large candidate fit together); more accurate than a count for heterogeneous model sizes; leverages existing VRAM monitoring; no special-casing of router models.
- Cons: needs reliable per-model footprint estimates before load (harder for lazy/auto-tuned `ctx_size`); interacts with the existing `max_loaded_models` count semantics (need to decide whether memory replaces or coexists with the count); backend memory reporting varies by device (GPU/NPU/CPU); less deterministic for users reasoning about "how many models are loaded."

These options are not mutually exclusive (e.g. A or B for the mechanism, plus C for validation/diagnostics, with D can be combined with any of them).

### Non-goals

- Bumping the global default `max_loaded_models` for everyone.

### Acceptance criteria

- Define explicit slot/lifetime semantics for router/classifier models within a router collection.
- Ensure a routed request under the **default** `max_loaded_models=1` does not cause the router and selected candidate to evict each other every request.
- Define the interaction with pinning, NPU exclusivity (`ExclusiveNpu` / `CoexistByType`), and cloud (`Unmetered`) recipes.
- Add an integration test exercising a `collection.router` alias under the default `max_loaded_models=1`, asserting no repeated cross-eviction (e.g. router stays resident / at most one cold-start per model across consecutive requests).

### Use Case / Motivation

With router collections (collection.router, https://github.com/lemonade-sdk/lemonade/issues/2385) now live, the default slot configuration causes the router model and the selected candidate model to repeatedly evict each other.

### Platform Relevance

All platforms

### Additional Context

_No response_

fl0rianr:
## Summary

Fixes #2725 by giving router/classifier dependencies an explicit residency role instead of making them compete with user-facing models in the same count-based LRU pool.

With the default `max_loaded_models=1`, an LLM router and its selected LLM candidate can now remain loaded together:

- router: `routing_helper/llm`
- candidate: `standard/llm`

This is intentionally a narrow residency fix, not the full resource-policy redesign discussed in #1705.

> **Stacking note:** this PR is based on #2698 because that PR introduces the live LLM-router path and its L0a integration test. Rebase onto `main` after #2698 merges.

## Why not raise `max_loaded_models`?

Automatically changing `max_loaded_models` would alter user memory behavior globally, still mix two different responsibilities in one LRU pool, and remain ambiguous for pinning and device-exclusive backends. A warning would describe the cold-start loop without fixing the default behavior.

### Request intent and live residency

Adds two independent concepts:

```cpp
enum class LoadPurpose {
    UserInference,
    RoutingDependency,
};

enum class ResidencyClass {
    Standard,
    RoutingHelper,
};
```

`LoadPurpose` is request-scoped intent. `ResidencyClass` is stored on the live `WrappedServer` and controls count-based admission and LRU eviction.

`ModelType` remains unchanged: an LLM router is still an LLM. Backend `SlotPolicy` also remains recipe-static and continues to describe hardware constraints.

### Capacity and eviction

- Standard pools continue to use `max_loaded_models` independently per `ModelType`.
- Routing helpers get one slot per `ModelType`.
- LRU lookup and pinned-slot failure are scoped to `(ResidencyClass, ModelType)`.
- Cloud/`Unmetered` models consume neither local pool and report `slot_pool: unmetered`.
- An already-loaded standard model is promoted to `RoutingHelper` in place when it is first used as a routing dependency; no second process or reload is created.
- A normal inference request does not auto-demote an existing helper. Explicit unload/reload resets the role.
- Implicit model selection prefers `standard` residency, so a warm helper does not make one normal candidate ambiguous or hijack legacy most-recent-model operations.

### Pinning

Pinning remains eviction protection, not a residency role:

- a pinned standard candidate does not block the helper pool;
- a pinned helper does not block the standard pool;
- a full target pool containing only pinned models still returns `slots_pinned_error`.

### NPU/backend exclusivity

Hardware constraints remain authoritative and orthogonal to count-based pools. When an incoming model would require a cross-residency eviction because of exclusive NPU ownership or FLM's one-per-type rule, the load fails deterministically with:

- HTTP `409 Conflict`
- error code `router_residency_conflict`

This avoids replacing the slot-thrashing loop with an NPU-specific thrashing loop.

### Observability

`GET /api/v1/health` now reports for every live model:

- `residency_class`
- `slot_pool`

## Tests

- Adds a header-only `ModelResidencyTest` covering purpose mapping, capacity rules, pool identity, and stable health strings.
- Extends the live L0a LLM-router test from #2698 to verify:
  - the default standard LLM limit is `1`;
  - router and candidate are simultaneously resident in separate pools;
  - two consecutive routed requests retain both backend PIDs;
  - unloading/reloading only the helper succeeds while the standard candidate is pinned;
  - the pinned candidate PID remains unchanged.

## Non-goals

This PR deliberately does not add:

- global or per-device memory budgets;
- generalized utility-model policy;
- role-weighted pressure-eviction scoring;
- configurable helper-pool counts;
- automatic `max_loaded_models` mutation;
- implicit router pinning.

Those belong to the broader policy work in #1705.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -306,8 +306,9 @@ jobs:
           cmake --build --preset default --target test_job_expr
           cmake --build --preset default --target test_job_graph
           cmake --build --preset default --target test_origin_utils
+          cmake --build --preset default --target test_model_residency
           cd build
-          ctest --output-on-failure -R "ProcessManagerTest|DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|LlmRouter|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest|mcp_client_config|JobExprTest|JobGraphTest|OriginUtilsTest"
+          ctest --output-on-failure -R "ProcessManagerTest|DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|LlmRouter|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest|mcp_client_config|JobExprTest|JobGraphTest|OriginUtilsTest|ModelResidencyTest"
 
       - name: Upload .deb package
         uses: actions/upload-artifact@v7
@@ -621,8 +622,9 @@ jobs:
           cmake --build --preset default --target test_job_expr
           cmake --build --preset default --target test_job_graph
           cmake --build --preset default --target test_origin_utils
+          cmake --build --preset default --target test_model_residency
           cd build
-          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|LlmRouter|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest|mcp_client_config|JobExprTest|JobGraphTest|OriginUtilsTest"
+          ctest --output-on-failure -R "DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|LlmRouter|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest|mcp_client_config|JobExprTest|JobGraphTest|OriginUtilsTest|ModelResidencyTest"
 
       - name: Upload .pkg package
         if: steps.check_signing.outputs.has_signing == 'true'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1943,6 +1943,23 @@ if(EXISTS "${_GGUF_CAPS_TEST_SRC}")
     add_test(NAME GgufCapabilitiesTest COMMAND test_gguf_capabilities)
 endif()
 
+# Model residency role/pool contract for router dependencies
+set(_MODEL_RESIDENCY_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_model_residency.cpp"
+)
+if(EXISTS "${_MODEL_RESIDENCY_TEST_SRC}")
+    add_executable(test_model_residency
+        test/cpp/test_model_residency.cpp
+    )
+    target_include_directories(test_model_residency PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+
+    include(CTest)
+    add_test(NAME ModelResidencyTest COMMAND test_model_residency)
+endif()
+
 # Pure version-resolution policy test (lemon::resolve_latest_pin). Header-only
 # logic, so the test links nothing beyond the standard library — regression
 # coverage for #2265 (fall back to the installed binary when the GitHub
--- a/docs/guide/configuration/multi-model.md
+++ b/docs/guide/configuration/multi-model.md
@@ -20,6 +20,32 @@ Models are categorized into these types:
 
 Each type has its own independent LRU cache, all sharing the same slot limit set by `max_loaded_models`.
 
+## Router and Classifier Dependencies
+
+Models loaded internally to choose a candidate use a separate **routing-helper**
+residency pool. This is intentionally independent from both `ModelType` and the
+backend's hardware slot policy:
+
+- A router LLM remains an **LLM**; it is not reclassified as a classification model.
+- Each model type has one routing-helper slot in addition to its normal
+  `max_loaded_models` slots. With the default `max_loaded_models=1`, one router
+  LLM and one selected LLM candidate can therefore stay warm together.
+- If an already-loaded standard model is later used as a routing dependency, its
+  live process is promoted in place without a reload or duplicate process.
+- Pinning is pool-local for count-based LRU: a pinned normal candidate does not
+  block the routing-helper slot, and a pinned helper does not block normal slots.
+- Cloud (`Unmetered`) models consume neither pool and report `slot_pool: unmetered`.
+- Implicit model selection prefers a normal `standard` model, so a warm internal
+  helper never makes an otherwise unique candidate ambiguous.
+- Explicit idle/VRAM-pressure eviction remains independent; an unpinned routing
+  helper can still be evicted when dynamic eviction is enabled.
+- Backend/device constraints remain authoritative. If a routing helper and
+  candidate cannot physically coexist (for example, exclusive NPU ownership),
+  the load fails with HTTP `409` and code `router_residency_conflict` instead of
+  repeatedly evicting one another.
+
+`GET /api/v1/health` exposes `residency_class` and `slot_pool` for every live model.
+
 ## Device Constraints
 
 <!-- BEGIN GENERATED: npu-exclusivity -->
@@ -98,7 +124,7 @@ To prevent frequently used models from being auto-evicted by the LRU policy, you
   ```
   Or via a POST request to `/internal/pin`.
 - **Pre-emptive Warnings:** The CLI checks `/api/v1/health` before loading a new model. If all slots for that model type are occupied by pinned models, the CLI will output a pre-emptive warning alerting you that loading will fail.
-- **Capacity Failures:** If all loaded models of a type are pinned and a request to load another model of that type is received, the load request fails with a `409 Conflict` HTTP status containing a `slots_pinned_error` code. You must unpin or unload at least one model of that type to free up a slot.
+- **Capacity Failures:** If every model in the target `(residency class, model type)` pool is pinned and another model needs that pool, the load request fails with a `409 Conflict` HTTP status containing a `slots_pinned_error` code. You must unpin or unload a model from that pool to free its slot.
 
 ## Per-Model Settings
 
--- a/docs/guide/configuration/multi-model.md
+++ b/docs/guide/configuration/multi-model.md
@@ -27,11 +27,21 @@ residency pool. This is intentionally independent from both `ModelType` and the
 backend's hardware slot policy:
 
 - A router LLM remains an **LLM**; it is not reclassified as a classification model.
-- Each model type has one routing-helper slot in addition to its normal
-  `max_loaded_models` slots. With the default `max_loaded_models=1`, one router
-  LLM and one selected LLM candidate can therefore stay warm together.
-- If an already-loaded standard model is later used as a routing dependency, its
-  live process is promoted in place without a reload or duplicate process.
+- Standard pools continue to use `max_loaded_models` per `ModelType`.
+- Routing-helper capacity is keyed by the distinct helper model rather than by
+  `ModelType`. Multiple classifier or router models of the same type required by
+  one or more policies can therefore remain warm together without consuming or
+  evicting standard user-facing slots. A second process for the same helper model
+  is never created.
+- With the default `max_loaded_models=1`, a router LLM and one selected standard
+  LLM candidate can stay warm together; additional distinct helper models can
+  also remain resident until explicit or pressure-based eviction removes them.
+- Residency follows the current use of the live process. A standard model used
+  as a routing dependency is promoted in place; a routing helper called directly
+  by the user is demoted into the counted standard pool in place. Destination-pool
+  admission and pinning rules apply before either transition. The
+  `router.model == candidate` case therefore promotes for routing and demotes for
+  candidate inference within the same request without creating a second process.
 - Pinning is pool-local for count-based LRU: a pinned normal candidate does not
   block the routing-helper slot, and a pinned helper does not block normal slots.
 - Cloud (`Unmetered`) models consume neither pool and report `slot_pool: unmetered`.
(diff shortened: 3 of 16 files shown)


ramkrishna2910 (commented):
Reviewed this closely since it's now on the router v1 ship gate. **The core design is sound and correct for the scenario it targets** (#2725): the `Standard`/`RoutingHelper` pool split keeps a router LLM and its candidate co-resident at `max_loaded_models=1`, and `test_021zk` verifies it behaviorally — both processes stay loaded across two requests with **stable PIDs**, and a pinned candidate survives a helper reload. I also confirmed the parts that are easy to get wrong are right:

- **Thread-safe:** every `residency_class_` read/write happens under `load_mutex_` (admission → `set_residency_class` → push are all inside the lock; only the slow `load()` releases it, matching the existing `model_type_`/`pinned_` pattern). No admission/load TOCTOU.
- **Nuclear retry** re-sets the residency class after `evict_all_servers`, so it's preserved.
- **No non-router regression:** with no helper ever present, `count_servers_in_pool(type, Standard)` equals the old `count_servers_by_type`, so standard multi-model eviction is unchanged.
- The selected candidate correctly loads `Standard` (only the router/classifier callback uses `RoutingDependency`); unmetered/cloud correctly excluded from both pools.

That said, a couple of semantic holes are worth addressing before this merges, especially as a v1 requirement. Ranked:

**1 — (strongest) No demotion → a model can get stuck in the helper pool and over-subscribe past `max_loaded_models`, with no `/health` signal.** A router model is an ordinary registry model a user can also call directly. Once promoted to `RoutingHelper` (or loaded as the router dependency), it never demotes — a `UserInference` request on it early-returns in `auto_load_model_if_needed` without touching residency. So at `max_loaded_models=1`: routed request loads `R`=`RoutingHelper/llm` + `C`=`Standard/llm`; user then directly loads a third LLM `D` → `ensure_residency_capacity(LLM, Standard)` counts only `C`, evicts it, loads `D`; now `R` (still serving user traffic) **and** `D` are both resident — two user-facing LLMs at a limit of 1. `/health` still reports `max_models.llm=1`, so nothing warns. Memory is bounded to +1/type (not a leak), but it silently breaks the VRAM guarantee that `max_loaded_models=1` exists to provide. The `router.model == candidate` case reaches the same stuck state in a single routed request.

**2 — The fix is incomplete for multi-classifier policies (same bug, relocated).** `residency_limit(RoutingHelper) == 1` per `ModelType` (and the unit test locks that in: `residency_limit(RoutingHelper, 8) == 1`). A policy with two classifiers of the same type — e.g. the documented `l3_classifier.json` example's `pii` + `jailbreak`, both onnxruntime `CLASSIFICATION` — makes them fight over the single `routing_helper/classification` slot and cold-start each other **every request**: exactly #2725's symptom, moved into the helper pool. Not a regression (they thrashed before too), but the PR/`multi-model.md` don't flag that same-`ModelType` multi-helper routing remains unfixed. Worth a one-line doc note + a follow-up issue.

**3 — On NPU, an internal helper hard-blocks the user's explicit load with a 409.** The `RouterResidencyConflictException` path fires when a resident NPU server's class differs from the incoming load, *before* the normal evict-all-NPU logic. So a router whose model is an NPU backend loads it as `RoutingHelper`, and the user's subsequent direct NPU-exclusive load now 409s instead of evicting the helper and succeeding (pre-PR behavior). "Deterministic over thrash" is defensible, but tie-breaking in favor of an *internal* helper over the *user's explicit* request is the wrong precedence, and it's a silent behavior change for existing NPU flows. At minimum document it; better, let the explicit user load win.

**4 — The new Router accounting has no unit coverage.** `test_model_residency.cpp` only tests the pure header helpers; the live `test_021zk` covers the happy path (router≠candidate, no NPU, pinned-candidate). The riskier new branches — the Standard→`RoutingHelper` promotion, the over-subscription sequence in (1), multi-helper eviction, and the NPU 409 in (3) — ship with essentially no automated verification. A couple of targeted cases (promotion; the third-model over-subscription) would lock the intended semantics.

Minor: a conflict raised on the Ollama load path (`OllamaApi::auto_load_model`) may not hit `create_model_error` and could surface as 500 rather than 409 (the 409 mapping partly relies on string-prefix matching on the message); and pinning a `RoutingHelper` makes `ensure_residency_capacity` throw `SlotsPinnedException` while `/health.pinned_models` reports 0 for the type, so the CLI can't predict it. Both narrow.

Net: mergeable for the single-router-LLM case, but I'd want (1) resolved or explicitly accepted+documented, and (2)/(3) documented as known limitations, before it counts as the v1 residency guarantee. Happy to pair on the demotion question — the cleanest fix is probably to demote on a direct `UserInference` hit (symmetric with the existing promote).

ramkrishna2910:
Since this is on the router **v1 ship gate**, let's address all of the review findings in this PR rather than deferring any to follow-ups. Consolidated punch-list with concrete fix directions and file anchors (from a trace of the branch — line numbers approximate):

### 1. Demote on direct use → stop over-subscribing past `max_loaded_models`  ⟵ the important one
Helpers promote (`router.cpp` ~589-603) but never demote, and a `UserInference` request on an already-loaded model early-returns without touching residency (`auto_load_model_if_needed`, `server.cpp` ~2185-2200). Net: a router model called directly stays `RoutingHelper`, so a later `Standard` model of the same type loads *alongside* it — two user-facing LLMs at `max_loaded_models=1`, with `/health` still reporting `max_models.llm=1`.
- **Fix:** on a direct `UserInference` load of an already-resident `RoutingHelper` model, demote it to `Standard` (symmetric with the existing promote), running it through `ensure_residency_capacity(type, Standard)` so it re-enters the counted pool. Also handle the `router.model == candidate` single-request case.
- **Test:** the 3-model sequence (routed request → `R`=helper + `C`=standard; direct call on `R`; load `D`) must not leave two standard-serving LLMs resident.

### 2. Size the helper pool to what a policy actually needs (multi-classifier thrash)
`residency_limit(RoutingHelper) == 1` per `ModelType` (`model_residency.h` ~49-51, locked by the unit test asserting `==1` at limit 8). Two same-type classifiers — e.g. the `l3_classifier.json` example's `pii` + `jailbreak` (both `CLASSIFICATION`) — fight over one helper slot and cold-start each other **every request**: #2725's symptom, relocated.
- **Fix:** allow the helper pool to hold the distinct helper models a policy requires (e.g. per-model helper residency, or a helper limit derived from the loaded policy's classifier count) instead of a hard 1-per-type. Update the unit-test expectation accordingly.
- **Test:** a two-same-type-classifier policy keeps both helper processes resident across requests (stable PIDs), like `test_021zk` does for router+candidate.

### 3. Explicit user load should win over an internal helper on the NPU
The `RouterResidencyConflictException` paths (`router.cpp` ~628-637 ExclusiveNpu, ~645-682 FLM) throw when a resident NPU server's class differs from the incoming load, *before* the normal evict-all-NPU logic — so an internal `RoutingHelper` on the NPU now 409s a user's explicit NPU load that previously succeeded by eviction.
- **Fix:** when the incoming load is `UserInference`/`Standard`, let it evict the resident helper (normal NPU exclusivity) rather than throwing. Reserve the deterministic-reject for the case where honoring the request would actually thrash a routed dependency mid-request.
- **Test:** direct NPU load succeeds (evicts the helper) when an NPU helper is resident.

### 4. Unit-cover the new Router accounting
Only the pure header helpers (`test_model_residency.cpp`) and one live happy-path (`test_021zk`) are covered. `ensure_residency_capacity`, the promotion/demotion paths, `count_servers_in_pool`, and the NPU cross-residency branch have no unit tests.
- **Fix:** add targeted tests for promotion, demotion (#1), the over-subscription sequence, multi-helper co-residency (#2), and the NPU precedence (#3, where hardware-independent).

### 5. (minor) Ollama load path may surface the conflict as 500
`OllamaApi::auto_load_model` (`ollama_api.cpp` ~227-258) doesn't funnel through `create_model_error`, and the 409 mapping elsewhere partly relies on string-prefix matching the `"Routing residency conflict:"` message (`server.cpp` ~2117). A conflict on the Ollama path could 500.
- **Fix:** typed catch of `RouterResidencyConflictException` → 409 on the Ollama load path (and prefer typed catches over message-prefix matching where practical).

### 6. (minor) Pinned helper is invisible to `/health`
`count_pinned_servers_by_type` counts only `Standard` (`router.cpp` ~2385-2394), so pinning a `RoutingHelper` makes `ensure_residency_capacity` throw `SlotsPinnedException` while `/health.pinned_models` reports 0 for that type — the CLI can't predict it.
- **Fix:** surface helper-pool pinning in `/health` (or document that helpers aren't user-pinnable).

The design and threading are solid (all residency access is under `load_mutex_`, nuclear-retry preserves class, no non-router regression) — these are the semantic gaps on top. Happy to pair on #1/#2 since those involve a design choice (demote-on-use, and how to size the helper pool).

ramkrishna2910:
Re-reviewed `813d6d109a` — all six items are addressed, and I checked the logic, not just the comments:

- **Over-subscription:** `transition_server_residency_locked` now runs a demoting model through `ensure_residency_capacity(Standard)`, and `ensure_loaded_model_residency` transitions an already-loaded model on a direct request — so a directly-used helper re-enters the counted pool. The regression test asserting the candidate is evicted confirms it.
- **Multi-classifier:** `same_residency_pool` keying the helper pool by `(type, model_name)` is the right fix — same-type distinct classifiers get separate limit-1 slots and stay co-resident.
- **NPU precedence:** gating the reject on `should_reject_residency_displacement` = `incoming == RoutingHelper` correctly lets an explicit user load evict a resident helper.
- Ollama + anthropic typed catches → 409, `pinned_helper_models` in `/health`, and the new unit + demotion tests all look right.

CI is green. LGTM from me. One non-blocking note for later: helpers are now keyed per distinct model and aren't evicted by count within their pool, so a long-lived server cycling through many different router policies can accumulate helper processes over time (bounded per-policy). Fine for v1 — worth a mental note for the multi-tenant case.

Just needs the branch updated onto `main` (currently BEHIND) and a maintainer approval.

ramkrishna2910 (approved):

Issue #2730, opened 2026-07-16
llama.cpp responses echo the local .gguf file path in the OpenAI `model` field


ramkrishna2910:
## Problem

A chat/completions response from a llama.cpp-backed model returns the **local GGUF file path** in the OpenAI `model` field instead of a clean model id:

```
"model": "C:\Users\...\snapshots\...\Qwen3-0.6B-Q4_0.gguf"
```

This is because `llama-server` is launched with `-m <path>` and echoes that path; lemond forwards the response verbatim (`LlamaCppServer::chat_completion` → `forward_request`, `llamacpp_server.cpp:611`). The omni collection orchestrator already normalizes this (`collection_orchestrator.cpp:837`), but the plain llama.cpp path and the router-dispatch path do not.

## Repro

```bash
curl -s -X POST http://localhost:13305/api/v1/chat/completions \
  -d '{"model":"Qwen3-0.6B-GGUF","messages":[{"role":"user","content":"hi"}],"max_tokens":8}' \
  | jq .model
# -> "C:\Users\...\Qwen3-0.6B-Q4_0.gguf"   (expected: "Qwen3-0.6B-GGUF")
```

Affects **all** llama.cpp responses (streaming and non-streaming), not just routing. OpenAI clients expect the `model` field to echo the requested/served model id.

## Fix

Normalize `response["model"]` to the backend's registered model id at the llama.cpp response layer (covers direct + router-forwarded requests). Mirror what the omni orchestrator already does.

## Context

Found while building the router end-to-end demo (#2388, finding F1). Non-blocking for the router e2e — the routing decision is reported cleanly via `x_lemonade_route.route_to` and the `x-lemonade-route` header. Filed as a separate backend fix per review scoping.

ramkrishna2910:
Router v1 battle test (main @ 18ddfaef) re-confirmed this, and it now has a schema-surface angle: `decision.schema.json` describes `route_to` as "The selected candidate; **also carried by the standard `model` field**" — which is untrue for local llamacpp candidates while this bug stands (routed responses carry the .gguf path, e.g. `C:\Users\<name>\.cache\huggingface\hub\...\Qwen3-4B-Q4_0.gguf`, leaking the absolute path + username). Since the v1 decision schema is frozen on shape, not descriptions, this can be fixed post-ship without a schema break — but until then the documented decision contract and actual responses disagree, so worth prioritizing.

SlawomirNowaczyk:
[...]

PR #2732, opened 2026-07-16
Router end-to-end demo, integration test, and docs (#2388)

### LINKED ISSUE #2388: [Router] End-to-end tests + docs

ramkrishna2910:
Closes #2388.

End-to-end demo, integration test, and docs for the Lemonade Router — the demonstrable exit criterion for the milestone. Naming a `collection.router` model in the OpenAI `model` field routes the request to a candidate by the policy's first-matching rule (fail-open to `default_model`), and the decision is reported on the response.

## What's here

**Runnable demo — `examples/router/`**
- `policy_local.json` + `demo.py` + `README.md`: a two-candidate router driven by a vanilla `openai` client. Deterministic conditions (`keywords_any`, `min_chars`, `metadata`), first-match ordering, fail-open, and the decision/trace on the response.
- `policy_cloud.json`: same shape with a cloud candidate (fireworks.ai).

**Integration test — `test/server_router.py`** (wired into both `Test CLI/Endpoints` jobs)
- `test_600`–`604`: default fallthrough, keyword → capable, `min_chars` → capable, `metadata` first-match wins, and trace opt-in.
- `test_610`: a `recipe:"cloud"` candidate via an in-process mock provider (install → auth → discover → register → route) — covers the local-vs-cloud split with no real key, so it runs in CI.

**Docs**
- `docs/dev/router-policy.md`: authoring guide (policy shape, condition vocabulary, first-match + fail-open, decision/trace contract, cloud candidates).
- Routing section in `docs/api/lemonade.md`; added to mkdocs nav.

## Verification

- **Local demo**: all four cases route correctly on a real `lemond` (Qwen3-0.6B default + Qwen3-4B capable).
- **Cloud demo (live)**: verified against **fireworks.ai** — coding/long prompts routed to a Fireworks model and returned real completions; casual and `consent:denied` prompts stayed local. The CI test reproduces the code path with a mock provider.
- `test/server_router.py`: **6/6 green locally.**

## Scope notes

- All routing condition types are covered end-to-end: deterministic (keywords / min_chars / metadata), `semantic_similarity` (embeddings, local + cloud candidate), and a real `classifier` condition backed by an `onnxruntime` encoder model (`test_630`, added after #2727 wired `run_classifier` -> `Router::classify`; runs on the self-hosted onnxruntime CI runner, skips on GitHub-hosted).
- Finding while building this: llama.cpp responses echo the local `.gguf` path in the OpenAI `model` field (pre-existing, not router-specific). Filed separately as #2730 — does not affect routing (the decision is reported cleanly via `x_lemonade_route.route_to`).

Draft while CI settles.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -2115,6 +2115,13 @@ jobs:
           .venv/bin/python test/server_websocket_auth.py
           echo "WebSocket auth tests PASSED!"
 
+      - name: Test router
+        if: ${{ !cancelled() && steps.setup.outcome == 'success' }}
+        shell: bash
+        run: |
+          .venv/bin/python -m test.utils.reset_server_state --best-effort --label "ubuntu router"
+          .venv/bin/python test/server_router.py --cli-binary lemonade
+
       # The full smoke test can be added to a different GPU equipped pipeline
       # as well if needed, but it would significantly add to the pipeline runtime.
       - name: Test mcp
@@ -2389,6 +2396,13 @@ jobs:
           echo "Running WebSocket idle test..."
           $VENV_PYTHON test/test_websocket_idle.py
           echo "WebSocket idle test PASSED!"
+
+      - name: Test router
+        if: ${{ !cancelled() && steps.setup.outcome == 'success' && !(runner.os == 'macOS' && needs.build-lemonade-macos-dmg.outputs.has_signing != 'true') }}
+        shell: bash
+        run: |
+          $VENV_PYTHON -m test.utils.reset_server_state --best-effort --label "${{ matrix.os }} router"
+          $VENV_PYTHON test/server_router.py --cli-binary "$CLI_BINARY"
 
       - name: Test web-app path traversal
         if: ${{ !cancelled() && steps.setup.outcome == 'success' && !(runner.os == 'macOS' && needs.build-lemonade-macos-dmg.outputs.has_signing != 'true') }}
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1183,6 +1183,11 @@ jobs:
             extra_args: "--wrapped-server onnxruntime"
             backends: "cpu"
             runner: [self-hosted, Windows, lemon-prod]
+          - name: router-onnxruntime
+            script: server_router.py
+            extra_args: "--wrapped-server onnxruntime"
+            backends: "cpu"
+            runner: [self-hosted, Windows, lemon-prod]
           - name: stable-diffusion
             script: server_sd.py
             extra_args: ""
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -85,6 +85,25 @@ Label names come from the model's `id2label` — from `config.json`, or from `ma
 
 Malformed requests (invalid JSON, missing `input`/`text`, non-string fields, non-positive `top_k`) return `400` with an `error` object before any model is loaded.
 
+## Routing (`collection.router`)
+<sub>![Status](https://img.shields.io/badge/status-experimental-orange)</sub>
+
+Naming a registered `collection.router` model in the `model` field of a
+`chat/completions` or `completions` request triggers the routing engine: the
+server picks a candidate by the policy's first-matching rule (fail-open to
+`default_model`) and forwards the request to it. No dedicated endpoint or `"auto"`
+model is involved.
+
+The decision is reported on the response:
+
+- Header **`x-lemonade-route`** — the matched rule id, or `default`.
+- Request field **`route_trace: true`** adds an **`x_lemonade_route`** object to the
+  response body: `{ route_to, matched_rule, default_used, outputs, trace[] }`
+  (`route_to` is the candidate that answered). For streaming responses it is
+  attached to the first SSE event.
+
+See [Router Policies](../dev/router-policy.md) for authoring the policy.
+
 ## `POST /v1/models/check-updates`
 <sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
 
--- a/docs/dev/router-policy.md
+++ b/docs/dev/router-policy.md
@@ -79,13 +79,14 @@ A `match` is a match-expression. Combine with the logical operators `any` (OR),
 |------|---------|
 | `semantic_similarity` | Cosine similarity of the input against labelled `reference_phrases`, via an embedding model. |
 | `classifier` | A `{label: score}` classifier — an encoder model via the `onnxruntime` backend (`/v1/classify`), or any model as an LLM-as-classifier via chat. |
+| `llm` | An LLM picks exactly one of the declared `labels` for the request (with a rationale); the chosen label scores `1.0`, the rest `0`. You supply the `model` and a `prompt` describing when to choose each label. |
 
 A classifier condition is a band test: `{ "classifier": "<id>", "label": "<name>",
 "min_score": 0.5, "max_score": 1.0 }` (omitting both bounds defaults to
 `min_score: 0.5`).
 
-Both model-backed conditions reference an entry defined in `routing.classifiers`,
-and each entry's `model` must be one of `components`:
+Each model-backed condition references an entry defined in `routing.classifiers`,
+and every entry's `model` must be one of `components`:
 
 ```json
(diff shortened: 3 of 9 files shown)


fl0rianr (commented):
Thanks for the thorough end-to-end coverage — the deterministic, semantic, cloud, and ONNX classifier paths are all valuable additions.

I found two issues that should be fixed before merge:

1. **Router cleanup does not actually delete the registered collections**

In `test_610_cloud_candidate_routing`, cleanup uses `DELETE /delete` with `{"model": collection}`. The semantic tests use `POST`, but also send `model` instead of `model_name`.

The server exposes `/delete` as a POST endpoint and expects:

```python
requests.post(
    f"{self.base_url}/delete",
    json={"model_name": collection},
    timeout=TIMEOUT_DEFAULT,
)
```

Since the cleanup responses are not checked, the tests currently appear to pass while leaving the registered routers behind. This is especially problematic for the cloud cases because the provider is removed afterwards, leaving a router that references a missing component on persistent self-hosted runners.

Please update all router cleanup paths to use `POST`, `model_name`, and assert the cleanup response.

2. **The new documentation says `routing.router` / `type: "llm"` are not implemented, but they are implemented on this branch**

The “Not yet implemented” section says the parser rejects `routing.router` and the `llm` classifier. However, the current code contains `LlmClassifier`, `desugar_routing_router()`, and invokes that desugaring from `parse_route_policy_collection()`.

This makes the new authoring guide internally inconsistent: the policy table already mentions `routing.router` as an alternative to `routing.rules`, while the final section says it cannot be used.

Please replace that section with the supported `routing.router` syntax and update the PR description accordingly.

One smaller documentation gap: the guide lists `semantic_similarity` and `classifier`, but does not show how to define the referenced entries in `routing.classifiers`. Adding one complete example for each would make the authoring guide much more usable.

Once the cleanup and the incorrect `routing.router` documentation are fixed, this looks close to merge-ready.

ramkrishna2910:
Both fixed in `fc67ab23`, verified end-to-end against a local `lemond`.

**1. Cleanup now actually deletes.** You're right — `/delete` is a POST endpoint keyed on `model_name`, and the tests used `DELETE` (test_610) or POST with `model`, without checking the response. Added a `_delete_collection` helper (`POST` + `model_name` + **asserts 200**) used by every per-test cleanup, plus a `tearDownClass` that removes the shared collection. Confirmed: after a full run, `GET /models` shows **zero** `Test-Router*` collections left (previously the cloud ones lingered after their provider was uninstalled). All 9 tests pass (8 + the onnxruntime-gated classifier).

**2. `routing.router` docs corrected.** Replaced the stale "Not yet implemented" section with a real **LLM-as-router** section (`{ type: "llm", model, prompt }`, mutually exclusive with rules/classifiers), and fixed the table's "reserved" note. Also added the doc gap you mentioned — a complete `routing.classifiers` example defining both a `semantic_similarity` and a `classifier` entry, wired to rules. The PR description didn't reference `routing.router`, so nothing stale to update there.

One FYI while I was in the parser: `parse_rules` still has a `throw "... reserved for #2405 and is not implemented"` for a `routing.router` key, but it's unreachable now — `parse_route_policy_collection` desugars `routing.router` *before* `parse_rules` runs, so that branch is dead code left over from before #2698. Not in this PR's scope, but worth a cleanup pass on the parser.

ramkrishna2910:
Follow-up on the docs: #2698 shipped **two** LLM forms and I'd only covered one. Added the second in `712d2a45` — the `type: "llm"` **classifier** usable directly in `routing.classifiers` (the LLM picks one of a declared label set, matched by rules like any classifier; this is the form the desktop classifier panel authors). It's now in the model-backed conditions table and the classifiers example, with a note that `routing.router` is shorthand for a single `llm` classifier over the candidate models.

sdevinenamd:
1. Cleanup can mask the real failure in the cloud tests (test_610, test_621)

In test_610 and test_621 the collection registration happens inside the try block, but the finally block unconditionally calls `_delete_collection`, which asserts HTTP 200. If anything fails before registration (install, auth, or the discovery assertion), the delete hits a never-registered collection and returns 404, so the assertion raised during cleanup becomes the test's reported failure ("cleanup delete failed: 404") and buries the actual root cause. 

2. The authoring guide doesn't cover explicit type: "llm" classifiers

The docs cover LLM-based routing only as the routing.router block, but #2698 also added a third entry type for routing.classifiers: { id, type: "llm", model, prompt, labels (≥1), default_label?, on_error? }. Unlike routing.router (which replaces rules entirely and picks a candidate directly), an llm classifier is a composable rule signal, its authored labels can be combined with other conditions, e.g. { "all": [{ "classifier": "judge", "label": "RISKY" }, { "has_tools": true }] }. Please add one example entry to the routing.classifiers section:

```
A third classifier type hands the label decision to a small LLM, unlike
[`routing.router`](#llm-as-router-routingrouter), which picks the final
candidate itself, an `llm` classifier just produces a label that rules can
combine with any other condition:

```json
"routing": {
  "classifiers": [
    {
      "id": "risk",
      "type": "llm",
      "model": "Qwen3-1.7B-GGUF",
      "prompt": "Classify the request's risk for tool execution. Reply with exactly one label: SAFE or RISKY.",
      "labels": ["SAFE", "RISKY"],
      "default_label": "SAFE",
      "on_error": "match_false"
    }
  ],
  "rules": [
    {
      "id": "risky-tool-calls-stay-local",
      "match": {
        "all": [
          { "classifier": "risk", "label": "RISKY", "min_score": 0.5 },
          { "has_tools": true }
        ]
      },
      "route_to": "Small-GGUF",
      "outputs": { "reason": "llm-judged-risky" }
    }
  ]
}

```

fl0rianr (commented):
Thanks for adding the second LLM form in `712d2a45` — documenting `type: "llm"` as a regular entry in `routing.classifiers` now matches the implementation well.

Two small items from my follow-up review are still open:

1. `tearDownClass()` still sends the shared collection delete request without checking the response. Please assert the status there as well, so a stale router cannot remain silently on a persistent runner.

2. The docs still say that `route_trace: true` adds `x_lemonade_route` to the response. The implementation attaches `x_lemonade_route` to every routed response; `route_trace` only controls whether the `trace` field is populated. Please update both API and router-policy docs accordingly, ideally including the existing `"version": "1"` field in the example.

After those two small fixes and green CI, this looks ready to approve.

meghsat (approved):

ramkrishna2910:
Thanks @sdevinenamd — both addressed in `78f4a2d9`.

**1. Cleanup masking.** Good catch. `_delete_collection` now accepts **200 or 404** — a 404 in `finally` just means the collection was never registered (a failure before install/auth/discovery), so we treat it as "nothing to clean up" and let the real exception surface, while any other status still fails the cleanup. Kept the assert so a genuine delete failure on a registered collection is still caught. Verified: full run stays green with zero `Test-Router*` collections left behind.

**2. Composable `type: "llm"` classifier.** Adopted your example — the classifiers section now has an `llm` `risk` classifier whose `RISKY` label is combined with `has_tools` via `all` (`risky-tool-calls-stay-local`), and the prose now states the two LLM forms explicitly: `routing.router` picks the candidate and replaces rules, while a `type: "llm"` classifier only produces a label that rules combine with any other condition. (I'd added a first pass at the `llm` classifier in `712d2a45` just before your comment landed; this makes it composable per your suggestion.)

PR #2734, opened 2026-07-17
feat(telemetry): honor incoming W3C traceparent (opt-in)

### LINKED ISSUE #2735: Honor W3C Trace Context (traceparent) for incoming inference requests (opt-in)
### Status Quo:
Lemonade already exports OTLP spans (OpenInference + OTel GenAI) for each inference request, but it ignores any incoming W3C traceparent (https://www.w3.org/TR/trace-context/) header. As a result every span is an orphaned root trace — there's no way for a span to become a child of a trace the caller already started.

### Proposal
Add an opt-in setting (e.g. telemetry.trust_incoming_trace_context, default false) that, when enabled, makes an inference span adopt a valid incoming traceparent's trace id and be parented to its span id. When the setting is off, or the header is absent/invalid, behavior is unchanged (fresh root span). Opt-in because span parentage would then depend on client-supplied input.

### Use Case / Motivation

I'm running a multi-agent orchestrator where a request flows through several agents, some hitting local models and some hitting cloud provider backends via Lemonade. I want to view a whole workflow as one connected trace and attribute token usage per workflow/agent across local vs. cloud. Today, because each Lemonade span is a separate root, the steps can't be stitched into a single trace.

This is the standard companion to the OTLP export Lemonade already supports, so it benefits any instrumented caller (LangGraph, CrewAI, custom orchestrators) — not just my setup.

### Platform Relevance

All platforms

### Additional Context

- Small, contained change: config flag + traceparent parse/validate + span parentSpanId on both the JSON and protobuf OTLP paths.
- I have a working implementation with a unittest and docs, tested end-to-end against a local OTLP collector on both wire formats.
- Draft PR: https://github.com/lemonade-sdk/lemonade/pull/2734
- Happy to align on the exact config key name / default and whether it belongs behind an existing telemetry sub-key before it's reviewed.

radhaksri:
## Summary

Adds an opt-in setting, `telemetry.trust_incoming_trace_context` (default `false`), that makes inference spans honor a caller-supplied W3C [`traceparent`](https://www.w3.org/TR/trace-context/) header. When enabled and a valid header is present, the span adopts the incoming trace id and is parented to the incoming span id, so requests join the caller's distributed trace instead of starting a fresh root. Emitted on both the JSON and protobuf OTLP paths.

Today every inference span is an orphaned root, so a multi-step workflow across an instrumented caller (e.g. an agent orchestrator) cannot be viewed as one connected trace. This is a generic Trace Context capability that benefits any instrumented client, not a specific integration.

Opt-in because span parentage then depends on client-supplied input; default-off preserves current behavior exactly.

Closes #2735

## Changes

- `telemetry.trust_incoming_trace_context` config key (accessor + validation + default), regenerated `defaults.json` / config docs via `gen_backend_boilerplate.py`.
- Parse + validate `traceparent` per the W3C grammar in `authenticate_request` (gated by the flag); stash into request-thread thread-locals.
- `InferenceSpan` adopts the incoming trace/parent ids at construction; `parentSpanId` written on both OTLP serializers.
- Added `traceparent` to CORS allow-headers.
- Docs: new "Distributed Tracing (W3C Trace Context)" section in `docs/guide/telemetry.md`.
- Test: `test_050_telemetry_trust_incoming_trace_context_config` (round-trip + non-boolean rejection).

## Test plan

- [x] Builds clean (Linux/GCC); `gen_backend_boilerplate.py --check` passes.
- [x] New unittest passes against a live server.
- [x] Live E2E against a local OTLP collector (**JSON** path): two requests sharing one trace id land in one trace, each parented to its own incoming span id; no-header → fresh root; flag-off → header ignored; malformed header → ignored.
- [x] Live E2E against a local OTLP collector (**protobuf** path, the default): decoded the wire payload and confirmed `parent_span_id` (field 4) is emitted correctly.
- [ ] Not tested on Windows/macOS locally (change is platform-agnostic C++; happy to get help validating).

_AI-assisted: implementation and this description were drafted with AI and reviewed by me._

### DIFF
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -131,7 +131,8 @@ Values set in the user's `config.json` always take precedence over these seeded 
         "otel_genai"
       ],
       "send_batch_size": 100
-    }
+    },
+    "trust_incoming_trace_context": false
   },
   "thinksound": {
     "backend": "auto",
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -251,6 +251,7 @@ API keys for these providers are **not** stored in `config.json` — they live i
 | `hide_inputs` | bool | false | Redact prompt message content from spans. |
 | `hide_outputs` | bool | false | Redact generated assistant message content from spans. |
 | `hide_thinking` | bool | false | Redact reasoning/thought content from spans. |
+| `trust_incoming_trace_context` | bool | false | Honor a caller-supplied W3C `traceparent` header so inference spans join the caller's distributed trace instead of starting a fresh root. Opt-in because span parentage then depends on client-supplied input. |
 | `max_queue_capacity` | int | 1000 | The maximum capacity of the in-memory telemetry queue buffer. Oldest spans are dropped when full. Must be `> 0`. |
 | `otlp` | object | (nested object) | Sub-block grouping OTLP transport details (see below). |
 
--- a/docs/guide/telemetry.md
+++ b/docs/guide/telemetry.md
@@ -32,6 +32,30 @@ Users can specify one or both formats. When both are enabled, Lemonade packs att
 
 ---
 
+## Distributed Tracing (W3C Trace Context)
+
+By default, each inference request starts its own root trace. When a caller is itself an instrumented application (e.g. a multi-agent orchestrator), you can have Lemonade's spans join the caller's trace instead, so an entire multi-step workflow appears as one connected tree.
+
+Set `telemetry.trust_incoming_trace_context=true` to opt in. Lemonade then reads the standard [W3C `traceparent`](https://www.w3.org/TR/trace-context/) header on incoming requests: the request's span adopts the header's trace id and is parented to the header's span id. Requests without a valid `traceparent`, or received while the setting is `false`, keep the default root-span behavior.
+
+```bash
+lemonade config set telemetry.enabled=true \
+                    telemetry.trust_incoming_trace_context=true
+```
+
+The caller supplies the header on each request, for example:
+
+```bash
+curl http://localhost:13305/api/v1/chat/completions \
+  -H "Content-Type: application/json" \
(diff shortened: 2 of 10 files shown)


abn:
Thank you for the addition @radhaksri. This is great, enabling optional W3C trace context ingestion will definitely be useful in multi-step flows. The changes look thread safe and accurate. Good catch on the CROS fix as well.

**Minor Recommendations:**

1. **Parser DoS Protection**: The current dash-splitting loop in `parse_traceparent` is susceptible to algorithmic complexity overhead if a caller sends a header packed with thousands of consecutive dashes. Since a valid W3C traceparent (version `00`) is always exactly 55 characters, we can prevent redundant allocation loops by adding a quick length guard at the start:
   ```cpp
   if (header.size() < 55 || header.size() > 128) {
       return false;
   }
   ```

2. **W3C Future-Proofing**: The parser currently returns `false` if `parts.size() != 4`. According to the W3C Trace Context spec, if a future version is encountered (version > `00`), we should parse the first 4 fields and ignore any extra trailing fields. Consider allowing `parts.size() >= 4` for future version support.

3. **Expand Test Coverage**: The current test checks config validation but doesn't verify the tracing logic itself. We should add automated test cases in `test/server_telemetry.py` (which already has a mock OTel receiver setup) to assert:
   * Spans adopt the incoming `trace_id` and `parentSpanId` when the config flag is enabled.
   * `traceparent` is ignored when the flag is disabled.
   * Malformed headers are ignored safely and fall back to fresh root traces.

4. **Configuration Docs**: In `docs/guide/configuration/README.md`, the setting is added to the example JSON snippet at the top, but it was missed in the detailed markdown table further down. We should add a quick row for it there.

Happy to review again once these are addressed.

abn (changes requested):
See comment above for details on requested changes.

radhaksri:
@abn : Implemented the changes for the review comments. Can you please review again? Just to note - all checks are through as well.

abn (approved):
Thank you for resolving the review comments. I think the changes LGTM now.

Issue #2287, opened 2026-06-17
The Big Refactor: self-describing WrappedServer backends


jeremyfowers:
# Plan: Self-Describing WrappedServer Backends

## The problem

Adding a backend should be a self-contained task. Today it's the opposite: a single backend (vLLM, Moonshine, and Cloud were the recent examples) is wired in by hand at **~13 places** scattered across the server and the CLI, and duplicated again in **~10 docs**. Every one is an `if (recipe == "...")` branch or a hardcoded list you have to remember. Miss one and the backend half-works: it loads but has no CLI flags, or installs but never shows up as supported. (The full list is in the appendix.)

## The idea

Let each backend **describe itself**. A backend is two files in its own folder: a **descriptor** that declares *what the backend is*, and a **server class** that implements *how it runs*. A registry collects every descriptor, and all the scattered `if (recipe == ...)` sites are rewritten to read the registry instead. The docs and the `/system-info` endpoint are generated from the same descriptors, so they can't drift.

## What a developer does

Adding a backend becomes **one folder plus three small appends** — nothing else:

| You edit | What goes there |
|----------|-----------------|
| `src/cpp/server/backends/<name>/` *(new folder)* | two files — a descriptor and a server class |
| `CMakeLists.txt` → the `LEMON_BACKENDS` list | **one line** registering the backend |
| `backend_versions.json` | version pin(s) — skip if the backend has no downloaded binary (e.g. cloud) |
| `server_models.json` | the models |

No router edits, no CLI edits, no doc edits, no support-matrix edits — those are all derived from the descriptor. That's the whole job.

### A backend = a descriptor + a server class

**The descriptor** is plain data: it declares everything *about* the backend. This is the single object the registry, the CLI, `/system-info`, and the docs all read.

```cpp
struct BackendOption {
    std::string name;          // "vllm_args"
    std::string cli_flag;      // "--vllm-args"  (empty = not a CLI flag)
    nlohmann::json default_value;
    std::string type_name;     // "ARGS" | "SIZE" | "BACKEND" | "BOOL"
    std::string help, group;
};

struct BackendDescriptor {
    std::string recipe;                 // "vllm"
    std::string display_name;           // "vLLM ROCm (experimental)"
    std::string binary;                 // subprocess to launch / install ("" = none, e.g. cloud)
    std::string config_section;         // defaults to recipe; sd-cpp → "sdcpp"

[...]

PR #2740, opened 2026-07-17
fix(mlx): isolate CPU backend from ROCm env bleed + fix test

bong-water-water-bong:
## Fix

CPU backend fails with `[hip_kernel] No ROCm back-end` because `load()` appended the inherited `LD_LIBRARY_PATH` which carries system ROCm paths (`/opt/rocm`, TheRock). The CPU `lemon-mlx-engine` binary loads HIP libraries from those paths and attempts ROCm init.

## Changes

**mlx_server.cpp** — CPU env isolation:
- `LD_LIBRARY_PATH` → only the executable directory (was: append existing)
- `ROCM_PATH`/`HIP_PATH` → `/dev/null` so HIP runtime finds nothing
- `HIP_VISIBLE_DEVICES` → empty string, no devices visible
- Preserves `MLX_DISABLE_COMPILE=1`

**test_lemon_mlx_integration.py** — fix version key assertion: backend_versions.json uses `rocm-stable`, not `rocm`.

## Validation
- All 3 MLX integration tests pass
- ROCm path unchanged (same code as before)
- Metal path unchanged


bong-water-water-bong:
Labels need fixing — current `engine::flm`, `runtime::rocm` are wrong. Should be `bug`, `engine::ryzenai`, `runtime::cpu`. Can a maintainer update?

bong-water-water-bong:
Closing this PR. It's a fix on top of #2472 which is being closed (MLX integration deferred until #2287 lands). Can reopen alongside the MLX backend when we revisit after #2287.

Issue #2742, opened 2026-07-17
NPU detection regression for docker container with 11.0.0 and unprivileged user switch (/dev/accel/accel0 permission problem inside the container)


AreYouLoco:
> @superm1 Well it broke NPU detection for me: TL;DR
> 
> ```
> lemonade-server  | 2026-07-17 16:56:25.983 [Info] (ModelManager) Backend availability:
> lemonade-server  | 2026-07-17 16:56:25.983 [Info] (ModelManager)   - NPU hardware: Yes
> [...]
> Installing backend: flm:npu
> Error: Cannot install flm:npu on this system: Requires AMD XDNA 2 AMD NPU
> ```
> 
> Additionally on host system:
> 
> ```
> $ flm validate
> [Linux]  Kernel: 7.1.3+deb14-amd64
> [Linux]  NPU: /dev/accel/accel0 with 8 columns
> [Linux]  NPU FW Version: 1.1.2.64
> [Linux]  amdxdna version: 0.8
> [Linux]  Memlock Limit: infinity
> ```
> 
> Lemonade GUI is not even showing FastFlowLM as an option to install. Freshly pulled lemonade-server docker container and updated app from Flatpak.
> 
> Full relevant log
> ```
> areyouloco@minisforum-x1-pro:~/code/lemonade-server$ docker compose logs -f lemonade 
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (main) Starting Lemonade Server...
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (main)   Version: 11.0.0
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (main)   Cache dir: /opt/lemonade/.cache/lemonade
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (main)   Port: 13305
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (main)   Host: 0.0.0.0
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (main)   Log level: info
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (main)   Telemetry: disabled
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (ModelManager) Loading "user_models.json"
> lemonade-server  | 2026-07-17 16:56:25.628 [Info] (ModelManager) Loading "recipe_options.json"
[...]

PR #2745, opened 2026-07-18
[docs] Correct UID/GID for migration instructions in docs/guide/install/docker.md

AreYouLoco:
Continuing from: github.com/lemonade-sdk/lemonade/pull/2617
The actual group ID (GID) inside container is 999:
```
lemonade@d62ed54ac02c:~$ id lemonade
uid=10001(lemonade) gid=999(lemonade) groups=999(lemonade)
```
This small PR fixes docker guide on migration from prior v10.10.,1

### DIFF
--- a/docs/guide/install/docker.md
+++ b/docs/guide/install/docker.md
@@ -30,9 +30,9 @@ docker run -d \
 > If you're upgrading from an image that ran as root (versions prior to v10.10.1), your existing named volumes may still be owned by root. The new image runs as UID 10001 and will fail to write to root-owned volumes. Fix ownership with helper containers:
 >
 > ```bash
-> docker run --rm -v lemonade-cache:/v ubuntu:24.04 chown -R 10001:10001 /v
-> docker run --rm -v lemonade-llama:/v ubuntu:24.04 chown -R 10001:10001 /v
-> docker run --rm -v lemonade-recipe:/v ubuntu:24.04 chown -R 10001:10001 /v
+> docker run --rm -v lemonade-cache:/v ubuntu:24.04 chown -R 10001:999 /v
+> docker run --rm -v lemonade-llama:/v ubuntu:24.04 chown -R 10001:999 /v
+> docker run --rm -v lemonade-recipe:/v ubuntu:24.04 chown -R 10001:999 /v
 > ```
 >
 > Alternatively, remove the old volumes and let the new container recreate them:


AreYouLoco:
I would also suggest switching to alpine linux image for such a small task as simple ```chown``` on volume or add instructions how to do that from host without using docker container in general. If desired.

fl0rianr (changes requested):
Thanks for confirming that the current image uses GID 999.

After looking at this again, I don't think replacing 10001:10001 with 10001:999 is the best long-term fix, because the Dockerfile does not explicitly guarantee GID 999.

The migration only needs to transfer ownership to UID 10001; it does not need to change the group at all. I would therefore suggest changing the commands from:

chown -R 10001:10001 /v

to:

chown -R 10001 /v

This is admittedly a different fix from the one currently proposed: it removes the unnecessary group assumption rather than documenting 999.

Would you be okay with adjusting the PR in that direction?

AreYouLoco:
I just spotted inconsistency. Do not insist on any direction. Consider it draft to open discussion.

Actually I switched just now to rootless docker containers and for me now correct UID/GID are:
```
chown -R 110000:100998 /path/to/volume
```

So YMMV here alot. And if proposing easy solutions for inexperienced users. It should actually be more complete with explanations what/why/when.

So yes there is place for an improvement here.

fl0rianr:
Thanks. The rootless example actually reinforces the concern: host-side UID/GID values such as `110000:100998` depend on the user's subordinate UID/GID mappings and cannot be used as portable documentation.

This PR specifically covers Docker named volumes accessed through a helper container, not direct ownership changes on an arbitrary host bind-mount path. For those named volumes, documenting only the stable in-container UID avoids depending on either the currently auto-assigned GID `999` or installation-specific rootless host mappings:

```bash
docker run --rm -v lemonade-cache:/v ubuntu:24.04 chown -R 10001 /v
docker run --rm -v lemonade-llama:/v ubuntu:24.04 chown -R 10001 /v
docker run --rm -v lemonade-recipe:/v ubuntu:24.04 chown -R 10001 /v
```

Since you would like this treated as an opening for discussion rather than a proposed final fix, keeping the PR as a draft makes sense. Expanded guidance for bind mounts and rootless Docker can be handled separately, as it requires explaining the host-specific UID/GID mapping.

AreYouLoco:
So in current state its decision between: keep your proposal of skipping GID set entirely in chown command and merge&close.

Or extend docs but I guess I would have to do it then😅. And I am not promising I will find time today or even this weekend.

fl0rianr:
Let's keep this PR focused and use the owner-only `chown` commands. That fixes the current migration guidance without documenting an unstable GID or expanding the scope into rootless Docker and bind-mount ownership.

Please update the three commands to:

```bash
docker run --rm -v lemonade-cache:/v ubuntu:24.04 chown -R 10001 /v
docker run --rm -v lemonade-llama:/v ubuntu:24.04 chown -R 10001 /v
docker run --rm -v lemonade-recipe:/v ubuntu:24.04 chown -R 10001 /v
```

No urgency from my side. Once that is updated, this should be ready to merge. More extensive rootless Docker guidance can be handled separately if needed.

AreYouLoco:
@fl0rianr there is one related issue I bumped into when moving to unprivileged lemonade user that's maybe worth mentioning here:

https://github.com/lemonade-sdk/lemonade/issues/2742

PR #2746, opened 2026-07-18
Adjust permissions for dev nodes in container

superm1:
Addresses: https://github.com/lemonade-sdk/lemonade/issues/2742\n\nThis PR adds runtime device group mapping so non-root access works for\nmapped accelerator devices in standard rootful Docker setups.\n\nNote: the rootless/unprivileged case (devices mapped as\nnobody:nogroup/GID 65534) remains unresolved because adding that group\ndoes not grant access inside unprivileged user namespaces — the entrypoint\nnow explicitly skips GID 65534 and documents this limitation.


fl0rianr (changes requested):
Thanks for working on this. The general entrypoint approach looks reasonable for rootful Docker setups, but I do not think this should close #2742 yet.

The reported rootless case appears to remain unresolved: inside the container, the device is mapped to `nobody:nogroup` / GID 65534, and adding `lemonade` to that group still did not grant access to `/dev/accel/accel0`. So the current implementation may help standard rootful setups, but it does not address the reproducer from the issue.

I also noticed two related regressions:

1. The standalone Dockerfile example in `docs/guide/install/docker.md` now uses:

```dockerfile
COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```

With the documented directory layout, the script is under `lemonade/docker-entrypoint.sh`, so that build should fail unless the path or instructions are adjusted.

2. Removing `USER lemonade` means the image metadata defaults to root. The main process drops privileges through the entrypoint, but `docker exec`, the healthcheck, and commands run with an overridden entrypoint will now execute as root by default. This changes the security behavior introduced in #2617 and should at least be intentional, documented, and covered by a test.

Could we keep this open until the actual rootless reproducer has been validated? I would also suggest adding `docker-entrypoint.sh` to the path filters of both container CI workflows and adding a small test that verifies the server process runs as UID 10001 and has access to a device owned by an injected supplementary GID.

siavashhub:
I don't have NPU/ROCm hardware to validate the device-mapping path, so treating this as a comment rather than a blocking review.

One thing that would let this fully supersede #2819 (and fix #2799): the loop only covers `renderD*`, so the `video` group / `/dev/dri/card*` devices aren't picked up. Adding `card*` makes this a strict superset of the static approach without hardcoding any GID. Suggestion inline.

Also +1 on the earlier review points that gate merge: the rootless case in #2742 (device shows as `nobody:nogroup`/65534) still looks unresolved, and the [docker.md](https://github.com/lemonade-sdk/lemonade/blob/main/docs/guide/install/docker.md) `COPY docker-entrypoint.sh` path won't build under the documented layout. Both worth confirming before this lands.

docker-entrypoint.sh
```diff
@@ -0,0 +1,56 @@
...
+    mkdir -p /run/lemonade
+    chown "$TARGET_USER":"$TARGET_USER" /run/lemonade
+
+    # Mirror host device group ownership into the container user so non-root
+    # access works for mapped accelerator devices.
+    add_device_group /dev/accel
+    add_device_group /dev/dri
+    add_device_group /dev/kfd
+
+    for dev in /dev/accel/* /dev/dri/renderD*; do
```
siavashhub, line 43:
```suggestion
    for dev in /dev/accel/* /dev/dri/renderD* /dev/dri/card*; do
```

siavashhub: above will address  #2819

fl0rianr (commented):
Thanks for addressing the earlier feedback — the CI path filters, `card*` handling, docs path, and runtime-user documentation all look good now.

I think there are just two things left before this is ready:

* The original #2742 reproducer still isn’t fully covered. In the unprivileged/rootless case, `lemonade` was added to the mapped `nogroup`/GID 65534 but still couldn’t access `/dev/accel/accel0`. So I’d either avoid `Closes #2742` / document the limitation, or validate a fix for that case.
* The new smoke test verifies UID and supplementary group membership, but not actual device access. It would be stronger to make the fake device group-readable and assert that the dropped-privilege process can actually read it.

Other than that, this looks much better. Thanks for iterating on it!

superm1:
Addressed your remaining feedback:

1. **Actual device access test** — smoke test now `chmod g+r` the fake device and runs `test -r` to verify the dropped-privilege process can actually read it (not just check UID/groups).

2. **Rootless/GID 65534 limitation** — entrypoint now explicitly skips GID 65534 (nogroup) with an explanatory comment and issue link. Updated the PR body from "Closes" to "Addresses" to reflect that the rootless case remains outstanding.

Ready for another review when you have time!

fl0rianr (commented):
Thanks, this is almost there now. 

One thing is still off in the new smoke test: `docker exec lemonade-perms sh -c 'test -r ...'` runs as root because the image no longer sets `USER lemonade`, so it doesn’t actually verify device access as the dropped-privilege process. The fake file may also already be world-readable depending on the runner umask.

Could we make the fake device explicitly `0640`/`0660` and run the read check through the entrypoint as UID 10001 with the injected group? Then the test would cover the actual behavior we care about.

Small docs nit: I’d also say “standard rootful Docker setups” instead of just “standard setups” to match the documented limitation.

PR #2749, opened 2026-07-18
fix(docker): include build-essential in runtime stage for vLLM Triton…

midu16:
fix(docker): include build-essential in runtime stage for vLLM Triton JIT

The multi-stage Dockerfile installed build-essential only in the builder stage, but vLLM needs gcc, binutils, and libc headers at runtime to compile Triton launcher modules on first model load. Without them, vllm-server fails with clang errors such as missing stdlib.h or crti.o.

Document the requirement in the Docker and vLLM configuration guides.

### DIFF
--- a/.github/workflows/container-build-test.yml
+++ b/.github/workflows/container-build-test.yml
@@ -85,6 +85,20 @@ jobs:
             "lemonade-server:test-${{ matrix.arch }}" \
             lemonade --help
 
+      - name: Verify runtime C toolchain
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          docker run --rm \
+            "lemonade-server:test-${{ matrix.arch }}" \
+            bash -c 'set -euo pipefail
+              tmp="$(mktemp -d)"
+              printf "%s\n" "#include <stdlib.h>" "void launcher(void) {}" > "${tmp}/launcher.c"
+              gcc -O3 -shared -fPIC -o "${tmp}/launcher.so" "${tmp}/launcher.c"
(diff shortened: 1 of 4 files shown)


fl0rianr (changes requested):
Thanks for the catch. The runtime toolchain requirement is valid, and adding it to the final Docker stage addresses the underlying Triton JIT failure.

I would still like two cleanups before merging:

1. Please add a lightweight regression check to `container-build-test.yml`. The current smoke test only builds the image, runs `lemonade --help`, and checks `/live`, so it would not detect missing compiler headers or linker files. Compiling a tiny shared C module inside the image would cover the actual requirement without needing a GPU.

2. Please simplify the comments and documentation. The current requirement should be documented, but the previous image's individual failure messages do not need to be repeated in the Dockerfile, the vLLM guide, the Docker run section, and troubleshooting. A short Dockerfile comment such as:

```dockerfile
# vLLM/Triton JIT-compiles native launcher modules at runtime.
```

would be sufficient. The embedded Dockerfile example should remain synchronized, while one concise note for custom image builders is enough.

You can take a look at docs/dev/documentation.md as well.

I would also avoid describing these specifically as “clang errors” or treating `posix_spawn failed` as definitive proof of a missing toolchain. Triton prefers GCC when it is available, and `posix_spawn` failures can have other causes.

`build-essential` is acceptable as the robust fix. It is broader than the stated minimum, but I would only replace it with a smaller package set after validating that against the bundled vLLM/Triton runtime.

fl0rianr (approved):
Thanks, LGTM now.

PR #2753, opened 2026-07-19
feat: effective options in model loading bar with optional power-user override

pwilkin:
Currently, a weakness of the new GUI3 loading bar is that you cannot effectively tell what the final model parameters will be - especially with model tuning. This exposes a small settings icon that displays all the effective settings together with their respective sources and allows - for the power user - to modify the final arguments for the current session.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,7 @@ All core endpoints are registered under **4 path prefixes**:
 - `/v0/` — Legacy short
 - `/v1/` — OpenAI SDK / LiteLLM compatibility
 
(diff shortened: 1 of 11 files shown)


fl0rianr (commented):
The endpoint currently does not return the truly final load options. `handle_load_command()` stops after `resolve_effective_recipe_options()`, while the real load resolves automatic context later via `resolve_auto_ctx_size()` after possible model eviction. This means the modal can still show `ctx_size = -1` or another pre-resolution value, even though displaying the final context size is one of the main purposes of this feature.

Could we either reuse the same final resolution path as the actual load, or explicitly return that the value will only be resolved at load time instead of presenting it as an effective final setting?

fl0rianr (commented):
Thanks, the auto-context issue is addressed now.

One blocker remains in the override flow: the UI says the edited backend arguments replace the resolved arguments, but the implementation only overwrites the `*_args` field while `merge_args` stays enabled. As a result, arguments removed by the user can be merged back in from global, architecture, or model defaults.

Could the session override also set `merge_args: false`, or otherwise use an explicit replace mode, so the final arguments actually match what the user entered?

fl0rianr (commented):
Thanks, the boolean rendering is fixed now. Two blockers remain:

The displayed command still starts with `lemonade-server load`, but `load` belongs to the `lemonade` client CLI. The generated command should start with `lemonade load`.

The apply flow is also destructive on failure: the session override is stored before `reloadModel()` unloads the currently running model. If the new backend arguments fail validation or the backend fails to start, the previous model is already gone and the invalid override remains active. Could the previous override be restored and the original model reloaded on failure, with the model state refreshed in a `finally` path?

PR #2755, opened 2026-07-19
fix: MCP setup gui

pwilkin:
This change adds two things:
* auto detects when the server isn't configured with secure access (neither LEMONADE_API_KEY nor LEMONADE_ADMIN_API_KEY set) and disables the configuration of MCP servers alltogether with a message explaining the proper configuration
* only displays the needed "Admin API key" window when the Admin API key is actually required


pwilkin:
Sorry, wrong merge target.

PR #2756, opened 2026-07-19
fix: make MCP server setup instructions clearer

pwilkin:
This change adds two things:

* auto detects when the server isn't configured with secure access (neither LEMONADE_API_KEY nor  LEMONADE_ADMIN_API_KEY set) and disables the configuration of MCP servers alltogether with a message explaining the proper configuration
* only displays the needed "Admin API key" window when the Admin API key is actually required

### DIFF
--- a/src/app/src/components/McpPanel.tsx
+++ b/src/app/src/components/McpPanel.tsx
@@ -152,27 +152,31 @@ const McpPanel: React.FC<McpPanelProps> = ({ connectionStatus, isActive }) => {
     }
   }, [mcpUrl]);
 
-  const probeAccess = useCallback(async (): Promise<'ok' | 'needs-admin' | 'unavailable'> => {
+  const probeAccess = useCallback(async (): Promise<
+    "ok" | "needs-admin" | "unavailable"
+  > => {
     setHostLoading(true);
-    setHostError('');
+    setHostError("");
     const result = await api.probeMcpAccess();
     setHostLoading(false);
     if (result.ok) {
       setServers(result.servers);
-      setAdminAccess('ok');
-      return 'ok';
+      setAdminAccess("ok");
+      return "ok";
     }
     setServers([]);
     if (result.status === 401) {
-      setAdminAccess('needs-admin');
-    } else if (result.status === 403) {
-      setAdminAccess('security-required');
+      setAdminAccess("needs-admin");
+      return "needs-admin";
     }
-    setAdminAccess('unavailable');
-    setHostError(result.status
-      ? `Could not reach MCP administration (HTTP ${result.status}).`
-      : 'Could not reach MCP administration. Check that the server is running and reachable.');
-    return 'unavailable';
+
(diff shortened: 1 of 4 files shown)


fl0rianr (commented):
The new flow is clearer, but there are two error-state issues that should be fixed before merge:

1. `applyAdminKey()` shows “Admin key applied for this app session” before the probe succeeds. When the key is wrong, the probe returns 401/403, but no error message is set, so the UI keeps showing a success notice even though access was rejected.

2. `probeAccess()` maps every failure to `needs-admin`, including network errors, 404s, and 5xx responses. That replaces the server controls with an admin-key form even when authentication is not the problem, and there is no simple retry action.

Could we:

* only show the success notice after a successful probe,
* show a clear “Admin API key was rejected” message for 401,
* reserve `needs-admin` for authentication failures,
* use a separate unavailable/error state with a Retry button for network and server errors?

It would also be good to keep `high_security` optional in the client. Treating a missing field as `false` makes a newer GUI incorrectly disable MCP administration when connected to an older but secured server.

src/app/src/components/McpPanel.tsx
```diff
@@ -150,21 +152,26 @@ const McpPanel: React.FC<McpPanelProps> = ({ connectionStatus, isActive }) => {
...
+    if (result.ok) {
+      setServers(result.servers);
+      setAdminAccess('ok');
+      return 'ok';
+    }
+    setServers([]);
+    if (result.status === 401 || result.status === 403) {
+      setAdminAccess('needs-admin');
+      return 'needs-admin';
     }
```
fl0rianr:
```suggestion
    if (result.status === 401) {
      setAdminAccess('needs-admin');
    } else if (result.status === 403) {
      setAdminAccess('security-required');
    }
```

fl0rianr (commented):
403 will not be resolved that way

fl0rianr:
Thanks for addressing the previous findings. One small control-flow issue remains in `probeAccess()`:

The new 401/403 branches currently fall through to the unconditional `setAdminAccess('unavailable')`, so both states are immediately overwritten. Also, `security-required` is not part of the `adminAccess` union and has no corresponding render branch.

I would avoid introducing a separate 403 state here because HTTP 403 is not necessarily proof that the server is keyless; it may also represent another forbidden condition such as an origin restriction. The explicit `high_security === false` health flag already handles the known keyless-server case.

Please keep the probe result handling simple:

```ts
if (result.status === 401) {
  setAdminAccess('needs-admin');
  return 'needs-admin';
}

setAdminAccess('unavailable');
setHostError(
  result.status
    ? `Could not reach MCP administration (HTTP ${result.status}).`
    : 'Could not reach MCP administration. Check that the server is running and reachable.',
);
return 'unavailable';
```

This preserves the dedicated admin-key flow for 401 while treating other failures, including ambiguous 403 responses, as unavailable with Retry.

fl0rianr (approved):
Thanks, good now

PR #2759, opened 2026-07-19
fix(ui): parse telemetry thinking blocks, collapse reasoning cards, and constrain panel heights

abn:
This PR resolves telemetry visualization and layout scrolling issues in the `GUI3_merging` beta branch:
- **Collapsible Reasoning Blocks**: Extracts `<think>...</think>` blocks from assistant messages and renders them inside custom collapsible `Reasoning Output` elements. They default to collapsed to optimize display space and can be toggled via header buttons.
- **Card-level Toggles & React Performance**: Moves the Rendered/Raw toggles to run locally on a per-card basis inside a `<MessageCard>` sub-component. This localizes state and prevents list-wide layout updates in React 19. The top-level "Raw" tab now renders the complete conversation history as a single unrendered JSON block.
- **Scrolling Height Constraint**: Locks `.inspect-layout` column heights on desktop to 100% of the viewport container, forcing the main detail cards inside `.inspect-detail__body` to trigger scrollbars internally instead of stretching the parent container and pushing the search rail's footer buttons off-screen.

image
image
image

### DIFF
--- a/src/app/src/components/inspect/MessagesTab.tsx
+++ b/src/app/src/components/inspect/MessagesTab.tsx
@@ -9,14 +9,133 @@ interface MessagesTabProps {
   handleCopyFull: (text: string, label: string) => void;
 }
 
+interface MessageCardProps {
+  m: Trace['messages'][number];
+  idx: number;
+  formatTokens: (num: number) => string;
+  handleCopyFull: (text: string, label: string) => void;
+}
+
+function stripLeadingThinking(content: string): string {
+  const match = /^\s*<think>[\s\S]*?<\/think>([\s\S]*)/.exec(content);
+  return match ? match[1].trim() : content;
+}
+
+function MessageCard({ m, idx, formatTokens, handleCopyFull }: MessageCardProps) {
+  const [isCollapsed, setIsCollapsed] = useState(false);
+  const [cardRenderMode, setCardRenderMode] = useState<'rendered' | 'raw'>('raw');
+  const [thinkingCollapsed, setThinkingCollapsed] = useState(true);
+
+  return (
+    <div className={`message-card ${isCollapsed ? 'collapsed' : ''}`}>
+      <div className="message-card__header">
+        <button
+          type="button"
+          className="message-card__header-toggle"
+          aria-expanded={!isCollapsed}
+          onClick={() => setIsCollapsed(!isCollapsed)}
+        >
+          <span className={`message-card__caret ${isCollapsed ? 'collapsed' : ''}`} aria-hidden="true">
+            <Icon name="chevron-down" size={12} />
+          </span>
+          <span className={`message-card__role ${m.role}`}>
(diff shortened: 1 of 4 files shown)


fl0rianr (changes requested):
Thanks for the improvements here — extracting `MessageCard`, adding collapsible reasoning output, and fixing the inspector height behavior all look directionally good.

I found three issues that should be addressed before merge:

1. **Rendered mode is now the default**
   The inspector previously defaulted to Raw intentionally. Telemetry content is untrusted, and `MarkdownMessage` permits images and external `src` URLs, so simply opening a trace may trigger external requests. Please keep both the global view and individual cards Raw by default, or explicitly disable remote resources in the inspector renderer.

2. **`<think>` parsing can modify legitimate content**
   The regex extracts `<think>...</think>` from anywhere in every assistant message. This will incorrectly transform XML examples, documentation, code blocks, or normal text containing literal `<think>` tags. It also trims the original content and only handles one block reliably. Please restrict parsing to the expected model-output format, ideally a leading reasoning block, and preserve the original raw value.

3. **Copy/export drops the reasoning**
   After extraction, reasoning only exists in `m.thinking`, but the card copy action, Messages JSON export, and OpenInference export only serialize `m.content`. The exported conversation therefore no longer matches the recorded telemetry. Please include `thinking` explicitly or retain and export the original raw message content.

Tests would also be valuable for leading reasoning blocks, literal `<think>` examples, multiple blocks, empty blocks, and copy/export behavior.

Requesting changes for the three functional issues above.

fl0rianr (commented):
Thanks for addressing the previous feedback. The original three issues are largely resolved: message cards now default to Raw, parsing is limited to leading `<think>` blocks while preserving the original content, and copy/export retains the raw message.

I found one remaining functional blocker:

**Card render state is reused across traces.**

`MessageCard` owns its local `cardRenderMode`, but cards are keyed only by their array index. If card N is switched to Rendered in trace A and the user then selects trace B, React reuses the same component and card N in trace B remains Rendered. This means telemetry from the newly selected trace can be rendered without an explicit action for that trace, including remote images/resources supported by `MarkdownMessage`.

Please key cards by the trace as well, for example:

```tsx
key={`${selectedTrace.id}:${idx}`}
```

or reset the local state whenever the trace/message identity changes. A regression test should render a card in one trace, switch to another trace, and confirm the new card is Raw again.

The new parser test also currently bypasses the parser by manually setting `thinking` on an already constructed Trace before calling `inspectStore.setState()`. It would still pass if `extractThinking()` were broken. Please exercise the OTEL parsing path directly or add a focused unit test for the parser helper.

One smaller edge case: `stripLeadingThinking()` is called even when `m.thinking` is absent, so older/imported traces may have a leading block hidden without getting a Reasoning Output card.

fl0rianr (approved):
LGTM

PR #2761, opened 2026-07-20
Add conformance tests (phase 1: deterministic)

### LINKED ISSUE #2425: [Router] Back-compat conformance corpus (golden policy → Decision tests)
## Goal
Mechanically guarantee the milestone's hard back-compat rule: **a future server must never break a policy authored against an earlier schema major.** Today that rule is held by discipline (the "never redefine a shipped field / never delete a major's schema" principles documented in `src/cpp/resources/schemas/README.md`). This issue makes it a CI gate via a **frozen conformance corpus** of versioned policies + their expected `Decision`s, replayed against every server build.

## Why
Schema validation proves an old policy still *parses*; it does **not** prove it still *routes the same way*. The deterministic tiers (keyword/regex/char rules, the `routing.router` desugaring, band/first-match logic) are pure lemonade engine logic and MUST stay behavior-stable across versions — exactly the thing a golden corpus locks down. (Model-backed scores carry inherent backend/model numerical wobble; see Scope for how the corpus handles that.)

## Scope
- **Corpus layout:** `test/conformance/routing/<schema_major>/<case>/` each holding `policy.json` (a versioned `collection.router` policy) + `cases.jsonl` (input → expected `Decision`). Seed v1 from the existing `test/cpp/fixtures/routing/` L0a–L3 examples.
- **Deterministic cases (exact):** L1 keyword/regex/char rules, `any/all/not`, first-match-wins, `default_model` fallback, and `routing.router` desugaring — assert the full `Decision` (`route_to`, `matched_rule`, `default_used`, `outputs`) byte-for-byte.
- **Model-backed cases (stubbed):** L2/L3/L0a run against a **pinned fake `ClassifierServices`** (fixed embeddings / scores / chat reply) so the assertion tests the *engine's* threshold + selection logic, not the backend's floats. The fake fixture is committed alongside the case. (Live-backend tolerance bands are explicitly out of scope here — that's a separate, non-gating perf/accuracy concern.)
- **Runner:** a CTest target (C++, reusing the foundation `RoutingPolicyEngine` + fake services) and/or a Python harness under `test/`, run in CI on every PR.
- **Append-only discipline:** when a new schema major ships, its corpus dir is added; **existing major dirs are immutable** — editing a frozen case is the CI signal that a change broke back-compat.

## Out of scope
- Live-backend numerical reproducibility / tolerance bands for model-backed scores.
- Component-resolution compatibility (an old policy naming a delisted model) — that's a model-registry concern, tracked separately.
- The migration shims themselves (per-major load-time upgraders) — this issue is the *test* that would guard them; the shim machinery lands when the second major is introduced.

## Acceptance
- A committed v1 corpus covering the deterministic paths + stubbed model-backed paths, seeded from the L0a–L3 fixtures.
- A CI-wired runner that replays every case and diffs the produced `Decision` against the expected one; any mismatch fails the build.
- A short doc note (in the schemas README) pointing to the corpus as the enforcement mechanism behind the "never redefine" rule.

## Dependencies
Needs the engine actually producing `Decision`s: the foundation interfaces (#2407), the evaluator/registry/conditions/classifiers, and the engine assembly (#2382). Best filled once `route()` is implemented; the corpus *fixtures* can be authored earlier. **Non-gating** for the rest of the milestone — it's the guard rail, not a build step on the critical path.

QuTosh:
## Summary
Adds a `policy → Decision` conformance corpus for the routing engine and a
CI-wired runner that replays every case and fails the build on any mismatch. This
locks the engine's deterministic routing behavior against future drift. Scope is
the **deterministic tier only**; model-backed (semantic/classifier/llm) cases are
deliberately deferred to a follow-up, so this **partially resolves #2425**.

## What's included
- **Runner** — [test/cpp/test_routing_conformance_corpus.cpp](https://github.com/QuTosh/lemonade/tree/QuTosh/tests_routing_backward_compat/test/cpp): replays each
  `cases.jsonl` line through the real engine + production `route_decision_to_json`
  serializer and compares the full `Decision` with the recorded one field for
  field — same fields, same values, no tolerance. (The comparison is on parsed
  JSON, so key order and formatting do not matter.) Fails if the corpus is
  empty/missing so it can't false-green.
- **CI wiring** — new `RoutingConformanceCorpusTest` CTest target (`CMakeLists.txt`)
  built and run in both C++ unit-test jobs in
  `.github/workflows/cpp_server_build_test_release.yml`  ([Linux](https://github.com/QuTosh/lemonade/blob/5ae314acb74470539ae69154a7167ecddc0c3b7a/.github/workflows/cpp_server_build_test_release.yml#L310) + [macOS](https://github.com/QuTosh/lemonade/blob/5ae314acb74470539ae69154a7167ecddc0c3b7a/.github/workflows/cpp_server_build_test_release.yml#L625)), on
  `pull_request`/`merge_group`.
- **Corpus (63 cases)** under [test/conformance/routing/1/](https://github.com/QuTosh/lemonade/tree/QuTosh/tests_routing_backward_compat/test/conformance/routing/1), grouped by the engine
  behavior each locks:
  - `l1_conditions_vocab/` (20) — `keywords_any`/`keywords_all`/`regex` + `any`/`all`/`not`/implicit-`all` (hit & no-match pairs)
  - `l1_conditions_features/` (10) — `has_tools`/`has_images`, incl. both Responses API `input` shapes
  - `l1_conditions_metadata/` (10) — `any`/`equals`/`exists`/whitespace-absent/comma-token-set (incl. exact-match negative)
  - `l1_conditions_char_bounds/` (6) — `min_chars`/`max_chars` (inclusive bounds; UTF-8 byte counting)
  - `l1_input_forms/` (6) — routing-input extraction from the `prompt` and `input` request fields
  - `l1_trace/` (5) — `route_trace=true`: per-leaf trace, accumulation, short-circuit, default
  - `l1_resolution/` (3) — first-match-wins, later-rule-fires, fail-open to `default_model`
  - `l1_conditions_features_negated/` (2) — authored `has_tools: false` is equality, not a catch-all
  - `l1_outputs/` (1) — matched rule’s nested `outputs` bag copied verbatim into the `Decision`
- **Docs** — [test/conformance/routing/README.md](https://github.com/QuTosh/lemonade/blob/QuTosh/tests_routing_backward_compat/test/conformance/routing/README.md) (layout + a **coverage matrix**
  mapping each v1 semantic → its single locking case), and a note in
  [src/cpp/resources/schemas/README.md](https://github.com/QuTosh/lemonade/blob/QuTosh/tests_routing_backward_compat/test/conformance/routing/README.md) pointing at the corpus as the behavioral
  back-compat mechanism.

## Design notes
- Cases are organized **by behavior, not provenance**, so the corpus reads as a
  checklist against the v1 semantics table. One case = one behavior; each
  case carries a `note` explaining its expected decision.
- Coverage is **orthogonal, not combinatorial** — combinators/resolution are
  op-agnostic, so they're locked once rather than across every leaf.
- Layout reserves `l2_semantic/` / `l3_classifier/` / `l0a_router/` as
  *to-be-added* for the deferred model-backed tier.

## Acceptance criteria (#2425)
- [x] Committed v1 corpus — **deterministic paths**
- [ ] Stubbed model-backed paths (follow-up)
- [ ] Corpus is CI-frozen
- [x] CI-wired replay runner; mismatch fails the build
- [x] Docs explaining the corpus as the enforcement mechanism

## ToDo (next PR)
 - add cases for semantic routing
 - freeze the corpus (if a future PR modifies the corpus, generate CI warning/error)

## Concerns

- Additional (and optional) field `note` is added to each case. This was not requested in the issue and not used during testing, but makes it clearer what case is covered by a given case.
- In order to limit the size of the corpus not all possible combinations of rules are added. The coverage of the corpus seems decent enough, but ideally should be evaluated by a reviewer.
- Rule names `min_chars` and `max_chars` (see `cpp/resources/schemas/route_policy.schema.json`) seem to be confusing. While this naming is not within the scope of this PR, 
it might be a good idea to change them if it is possible to do so without creating compatibility issues.


fl0rianr (commented):
Thanks, this is a solid phase-one structure. The corpus organization, CI integration and existing resolution cases look good.

I found three things I would address before approval:

1. Please use the production `build_route_context()` function instead of recreating part of it in the test runner. The local implementation currently omits tools, images, multimodal content, and the completions/responses input forms. The conformance test should freeze the real production request-to-context behavior, not a second implementation of it.

2. A present but empty cases.jsonl currently passes. Please count the executed cases and fail if a case file, or the corpus as a whole, contains no actual cases.

3. A few deterministic semantics still need stronger coverage:

   * add a negative implicit-all case where only one key matches
   * add at least one rule with non-empty nested outputs
   * add has_tools and has_images cases;
   * add the currently claimed but uncovered case-sensitive/exists:true behavior, or narrow the README's completeness claim.

The merge conflict itself is straightforward, expect similar conflicts if this PR hangs a bit due this one central C++ unitest location.
Only the I'm able to even approve running CI.

fl0rianr (commented):
Thanks for the update and for syncing the branch with `main`.

I’ve run the CI workflows so you should now be able to see any build or test failures directly on the PR.

The overall structure looks good, but I think the following points should still be addressed before merging:

1. Please use the production `build_route_context()` function in the conformance runner instead of maintaining the local `last_user_input()` / `to_context()` implementation. The current helper does not cover tools, images, multimodal message content, completions prompts, or Responses API inputs, so the corpus could stay green while the real server request-to-routing path changes.

2. Please make the runner fail when a `cases.jsonl` file contains no actual cases. It would also be good to discover case directories independently of `cases.jsonl`, so accidentally deleting one case file does not silently remove that case from the test suite.

3. The documentation currently describes the corpus as “frozen” and as enforcing compatibility, while corpus immutability is still listed as follow-up work. Either add the freeze protection here or soften that wording until the follow-up lands.

For the deterministic coverage, I would also add:

* a negative implicit-`all` case,
* at least one non-empty nested `outputs` object,
* `has_tools` and `has_images`,
* the documented case-sensitive regex/metadata behavior and `exists: true`.

This is a solid foundation and the changes above should make the conformance test much harder to false-green.

SlawomirNowaczyk (changes requested):
Thanks, the deterministic leaf coverage is well-organized and the hit/miss pairing per op reads cleanly against the table. Still, there are a handful of **deterministic** code paths that are reachable today but missing in the PR:

1. **The `trace` path is entirely uncovered.** The runner already reads `request.route_trace` and forwards `want_trace` to `engine.route(...)`, and `route_decision_to_json()` emits a whole `trace` array (`condition` / `result` / optional `score`) fed by every leaf's `trace_leaf(...)`. No case sets `route_trace: true`, so this branch of the Decision serializer and every leaf's trace label are unlocked. This is part of the public Decision JSON contract, so it's the most important miss — please add at least one case per condition family that asserts the emitted `trace`.

2. **The `prompt` (legacy completions API) input form is never exercised.** `build_route_context` has three input branches — `messages`, `prompt`, and `input` — and the corpus only touches `messages` and `input`. Please add a case that routes off a `prompt` body (both the string and the array-of-strings forms).

3. **`input` as a bare string / role-less fallback is unlocked.** All the responses cases use `input` as an array of role-tagged messages. The `input`-is-a-string branch and the role-less concatenation fallback (`append_line`) in `build_route_context` have unit tests but no corpus lock — please add a case for each.

4. **Negative `any` (all children miss) is missing.** `all` and `not` both have a miss case; `any` only has `any-one-child-matches`. For symmetry (and to actually exercise the fall-through of an `any` node), please add an `any-no-child-matches` case.

5. **ASCII-only case-fold boundary is unproven.** `ascii_lower` deliberately leaves non-ASCII bytes untouched. A keyword case with a non-ASCII letter (e.g. `É` in the input vs a lowercase `é` keyword) that asserts it does **not** match would lock that frozen-v1 semantic the same way the UTF-8 byte-count case locks `max_chars`.
6.  **`metadata equals: ""`** — a blank value counts as absent (`is_blank`), so `equals: ""` can never match. Worth a case so that semantic doesn't drift.

In addition, here are a few implementation-level suggestions for the conformance runner (`test/cpp/test_routing_conformance_corpus.cpp`). The common thread: the runner is rigorous about the *empty-corpus* false-green, but it's fragile against malformed input at every layer (row, policy compile, filesystem) and thin on diagnosability. This will make the harness much easier to use:

### Robustness — a malformed corpus should FAIL cleanly, never crash the suite

**1. Validate case-row shape before use.** The per-line loop only guards `json::parse(line)`; everything after assumes a well-formed row:

```cpp
const json& request  = row["request"];      // operator[] on non-const json
const json& expected = row.at("decision");   // throws if "decision" is absent
```

A row missing `"request"` silently inserts a null via `operator[]` and routes an empty context (runs against garbage); a row missing `"decision"` throws `json::out_of_range` **outside** the parse `try`, escaping `main` and aborting the whole suite. Validate explicitly:

```cpp
if (!row.is_object() || !row.contains("request") || !row.contains("decision")) {
    check(rel + ": line " + std::to_string(line_no) + " has request+decision", false);
    continue;
}
const json& request  = row.at("request");
const json& expected = row.at("decision");
```

**2. Guard engine construction.** The `policy.json` parse is wrapped, but the next line isn't:

```cpp
RoutingPolicyEngine engine(std::move(policy), lemon::ClassifierServices{});  // can throw
```

The constructor compiles every rule and throws `std::invalid_argument` on a policy that *parses* but doesn't *compile* (bad nesting depth, null leaf, unresolved classifier ref) — again uncaught. Fold it into a guarded block:

```cpp
std::optional<RoutingPolicyEngine> engine;
try {
    engine.emplace(std::move(policy), lemon::ClassifierServices{});
} catch (const std::exception& e) {
    check(rel + ": policy compiles", false);
    std::printf("  %s\n", e.what());
    return 0;
}
```

**3. Don't let directory traversal throw out of `main`.** `find_case_dirs` uses the throwing `recursive_directory_iterator` overload; a symlink cycle or permission error raises `fs::filesystem_error` straight through `main`. Use the `error_code` + `skip_permission_denied` overload so it degrades to the existing empty-corpus failure:

```cpp
std::error_code ec;
auto it = fs::recursive_directory_iterator(
    root, fs::directory_options::skip_permission_denied, ec);
if (ec) {
    std::printf("[FAIL] cannot walk %s: %s\n", root.string().c_str(), ec.message().c_str());
    return dirs;  // empty -> main reports the empty-corpus failure
}
for (const auto& entry : it) { ... }
```

### Harder to false-green — enforce the corpus's own invariants

**4. Enforce case-name uniqueness (and non-empty names).** The README declares `name` unique within a file, and the coverage-matrix / "one case = one behavior" argument depends on it, but the runner never enforces it and even falls back to `line-N` for unnamed cases:

```cpp
std::set<std::string> seen_names;  // per case dir
...
const std::string case_name = row.value("name", "");
if (case_name.empty()) {
    check(rel + ": line " + std::to_string(line_no) + " has a name", false);
    continue;
}
if (!seen_names.insert(case_name).second) {
    check(rel + ": duplicate case name '" + case_name + "'", false);
    continue;
}
```

**5. Reject unknown top-level keys in a case row.** A typo like `"desicion"` or `"nite"` currently slips through silently. A small allowlist catches that class of corpus corruption:

```cpp
static const std::set<std::string> kAllowedRowKeys = {"name", "note", "request", "decision"};
for (auto it = row.begin(); it != row.end(); ++it) {
    if (kAllowedRowKeys.count(it.key()) == 0) {
        check(rel + ": line " + std::to_string(line_no) + " unknown key '" + it.key() + "'", false);
    }
}
```

### Diagnosability

**6. Print a precise diff on mismatch.** The failure path dumps both full objects; for nested-`outputs` cases that means eyeballing two long JSON strings. nlohmann already gives an RFC-6902 diff that names the exact pointer that drifted:

```cpp
if (!ok) {
    std::printf("  expected: %s\n", expected.dump().c_str());
    std::printf("  produced: %s\n", produced.dump().c_str());
    std::printf("  diff:     %s\n", json::diff(expected, produced).dump().c_str());
}
```

QuTosh:
Implemented the requested changes. A few notes worth evaluating manually:

1. JSON diff between `expected` and `produced` — instead of dumping the whole
   diff on a single line, each differing field gets its own line (with its
   expected and produced value). I think it's easier to read this way.
   
2. Trace cases aren't centralized, and that's intentional — each condition
   family keeps its own trace case in its own dir (e.g. `keywords_any-trace`,
   `regex-trace`, …), right next to the rest of that family's cases. The
   dedicated `l1_trace` dir is only for the cross-cutting trace logic
   (accumulation, short-circuit, default carrying the trace), not a dumping
   ground for every family's trace case.
   
3. Recursive walk over the corpus is abandoned in favor of flat lookup — case
   dirs now have to sit at `routing/<version>/<case>`. Each case dir must contain both
   `policy.json` and `cases.jsonl` and have no nested subdirs, otherwise the
   test fails. One escape hatch: if a dir (or a whole version) name starts with
   `_debug`, it only warns instead of failing.
   
   4. This tests are not yet frozen in CI as they are to be extended with the next PR
   
   Please let me know if any of this points requires changes

SlawomirNowaczyk (approved):
LGTM

fl0rianr (commented):
Thanks for the thorough updates. The previously raised false-green and coverage concerns appear to be addressed: the runner now uses the production `build_route_context()`, validates the corpus rows and layout, rejects empty case files, and covers the relevant deterministic input, trace, feature, output, and resolution paths.

I found one remaining corpus invariant that should be enforced before merging:

### Validate the schema-major directory against `policy.json`

The documented layout is:

```text
routing/<schema_major>/<case>/
```

and the README states that `<schema_major>` is the policy's root `version`. The runner currently parses the policy but does not verify that, for example, a policy under `routing/1/` actually declares `"version": "1"`.

Without this check, a future policy could accidentally be placed under the wrong version directory while the conformance suite remains green.

A check before `parse_route_policy_collection()` should be sufficient, along the lines of:

```cpp
const json policy_json = load_json_file(case_dir / "policy.json");
const std::string directory_version =
    case_dir.parent_path().filename().string();

if (!policy_json.contains("version") ||
    !policy_json["version"].is_string() ||
    policy_json["version"].get<std::string>() != directory_version) {
    check(rel + ": policy version matches schema-major directory", false);
    return 0;
}

policy = lemon::parse_route_policy_collection(policy_json);
```

### “Frozen” wording

The corpus currently enforces the expectations committed in the same revision, but it does not yet prevent a future PR from changing both the implementation and the golden expectations together. Since corpus immutability is explicitly deferred to a follow-up, I would either add that protection here or temporarily soften wording such as “frozen” and “hard back-compat enforcement” to “committed conformance corpus” until the freeze mechanism lands.

### Current CI failure

The failing macOS job is unrelated to this PR. The new `RoutingConformanceCorpusTest` builds and passes; the job fails later in the Whisper Metal realtime integration tests. I will try fixing that separat.

QuTosh:
1. Soften the wording in conformance tests README
2. Added version check as suggested

fl0rianr (changes requested):
Something must have gone wrong. The commit history does indicate this and 39 files changed is 100% out of scope for this PR. Please correct this, thanks!

QuTosh:
It was a merge with rebase

fl0rianr:
Please fix it with the correct head. We can't merge as-is.

QuTosh:
Should be fixed now.

fl0rianr (commented):
Thanks for cleaning up the branch. I re-reviewed the complete current diff at `b15daf6`. The 23 changed files are now fully in scope: workflow/CMake wiring, schema documentation, the conformance runner, and nine policy/case pairs. The unrelated files from the temporary 39-file state are gone.

The schema-major directory validation added since my previous review is correct.

I still found two harness-integrity blockers:

1. **The _debug naming escape hatch can make malformed committed corpus directories pass CI.**

   `find_case_dirs()` downgrades missing `policy.json` / `cases.jsonl` files and unexpected nested directories to warnings whenever either the version or case directory starts with `_debug`.

   A committed conformance corpus should not have a filename-based bypass that can silently remove coverage. Please remove this bypass, or gate it behind an explicit local-only option that is disabled in CI.

   Filesystem traversal errors should also be registered through `check(..., false)`, not only printed. In particular, the `sec` error from scanning a case directory is currently ignored, and a per-version traversal error would not affect `g_failures` when another version still supplies valid cases.

2. **The Responses API input_image case does not exercise the production input branch.**

   `has_images-input-image-part` currently places an `input_image` part inside `messages`. This tests the shared content-type detector, but `build_route_context()` has separate image-scanning logic for the Responses API `input` array.

   Please change that case to use an actual `input` request, for example:

   ```json
   {
     "input": [
       {
         "role": "user",
         "content": [
           {"type": "input_text", "text": "describe this picture"},
           {"type": "input_image", "image_url": "https://example.com/cat.png"}
         ]
       }
     ]
   }
   ```

There are also documentation corrections needed before merge:

* The corpus is not yet CI-frozen; immutability is explicitly deferred to a follow-up. Please replace claims that it is already “frozen”.
* The runner compares JSON values structurally, not byte-for-byte.
* The PR description still says 25 cases, while the current corpus contains 62.
* The “ECMAScript dialect” row is not actually locked by `echo-[0-9]+`, since that syntax is shared by many engines. Either use an ECMAScript-specific construct or narrow that claim.
* Please fix the determenistic typo in the title.

After the two harness issues, the documentation claims, the PR should be in good shape for approval. Since I already know you will most likely run into an required CI fail you sadly have to wait for the merge of https://github.com/lemonade-sdk/lemonade/pull/2821 until we can merge this.

Thanks for the hard work!

QuTosh:
**1. `_debug` bypass removed.** `find_case_dirs()` no longer has any filename-based escape hatch — a case dir missing `policy.json`/`cases.jsonl` or holding a subdirectory is always a hard failure. Nothing in the corpus used the hatch and it was undocumented, so it is gone rather than gated. While there, a version dir must now contain case directories only: a stray file or a broken symlink where a case dir should be is also a failure, since that was the same silent-coverage-loss shape.

**2. Filesystem errors registered through `check(..., false)`.** The ignored `sec` from scanning a case dir now fails the run and excludes the dir; a per-version traversal error now affects `g_failures` instead of only printing; the root traversal error and both `fs::exists` calls are covered too. The missing-corpus and empty-corpus bails in `main()` also go through `check()`, so no `[FAIL]` line is hand-rolled anymore. Related: the entry stat and path labelling now use the non-throwing overloads (`is_directory(ec)`, `fs::relative(..., ec)` with a lexical fallback), so a filesystem error can no longer escape as an exception — a symlink loop used to abort the runner.

**3. Responses API `input_image` case fixed.** `has_images-input-image-part` now sends an actual `input` request as suggested. Added `has_images-input-bare-image-part` for the second `input` shape (bare content parts with no role wrapper), which is the branch that hits the `json::array({item})` path. Verified by deleting the Responses image scan from `build_route_context()` locally: previously the corpus stayed green, now both cases fail.

**4. "Frozen" claims replaced.** The runner header, the CMake comment, and both schemas README blocks no longer describe the corpus or its expectations as frozen. `frozen` now appears only in the unchecked `[ ] Corpus is CI-frozen` acceptance box.

**5. Comparison described accurately.** "byte-for-byte" is gone. The runner compares parsed JSON, so the docs now say: every field, every value, no tolerance — key order and formatting do not matter.

**6. "ECMAScript dialect" row narrowed.** Kept `echo-[0-9]+` and dropped the dialect claim. The case is renamed `regex-matches-substring` and now locks what it actually tests: the pattern is searched for anywhere in the input rather than matched against all of it. The coverage matrix records that the dialect is deliberately *not* locked, and what locking it would need (an ECMAScript-only construct such as `\d`).

**7. PR title and description.** Typo fixed. The description now says 63 cases with all nine groups and their counts, and the byte-for-byte and frozen wording is corrected there too.

**8. Branch synced with `main`.**

fl0rianr (approved):
Thanks for the hard work, LGTM now.

PR #2763, opened 2026-07-20
feat(router): report estimated cost on collection.router decisions

Bekhouche:
**Summary**

Wires a `CostServices` seam into `RoutingPolicyEngine` so cost/latency metadata for the selected candidate is surfaced on `collection.router` decisions.

- Once `route_to` is resolved (matched rule or default), per-candidate cost/latency info, using typed `ModelInfo` fields when known and hand-authored `extras` otherwise, is merged into `Decision::outputs.estimated_cost`
- Reporting only, this doesn't affect candidate selection. It simply provides the same cost visibility already shown on `/v1/models`
- Adds `Router::try_get_model_info` plus unit tests for `resolve_cost_info`'s typed-vs-extras precedence and the attach/omit behavior for matched-rule and default routes

**Test plan**

- [x] `ctest -R RoutingClassifierServicesTest`
- [x] `python test/test_schema_lock.py`


src/cpp/server/routing_policy.cpp
```diff
@@ -1316,7 +1348,9 @@ Decision RoutingPolicyEngine::route(const RouteContext& ctx, bool want_trace) co
                                 << std::endl;
     }
 
-    return Decision(policy_.default_model, want_trace, std::move(eval.trace));
+    Decision decision(policy_.default_model, want_trace, std::move(eval.trace));
+    attach_estimated_cost(decision, cost_services_);
```
ramkrishna2910, line 1385: `route()`'s documented contract is "Never throws: any unexpected failure fails open to default_model", but this `attach_estimated_cost` call runs **outside** the try/catch, and `cost_of` is an injected `std::function` the engine can't assume is no-throw. Even the router-backed one can throw: `Router::try_get_model_info` guards `get_model_info()` with try/catch but calls `model_manager_->model_exists()` (which runs `build_cache()`) outside that guard.

Secondary wart on the matched path (line 1338): if attach throws there, it's caught, but `eval.trace` has already been moved into the abandoned Decision (trace silently lost) and attach is then retried here unguarded.

Suggest swallowing exceptions inside `attach_estimated_cost` (cost is best-effort metadata, so dropping it on failure is the right fail-open behavior), and/or moving the `model_exists` call inside `try_get_model_info`'s try block.

src/cpp/server/routing_policy.cpp
```diff
@@ -1279,8 +1279,38 @@ NamedLeafFactories make_deterministic_leaf_factories() {
...
+        estimated["cost_input_per_million"] = *info.cost_input_per_million;
+    }
+    if (info.cost_output_per_million) {
+        estimated["cost_output_per_million"] = *info.cost_output_per_million;
+    }
+    if (info.latency_ms_hint) {
+        estimated["latency_ms_hint"] = *info.latency_ms_hint;
+    }
+    if (!estimated.empty()) {
+        decision.outputs["estimated_cost"] = std::move(estimated);
```
ramkrishna2910, line 1336: This silently clobbers a rule-authored `outputs.estimated_cost`. `Rule::outputs` is the policy author's pass-through bag (the trust layer reads off it), so if an author set their own `estimated_cost` on a rule, the engine overwrites it with no warning.

Suggest letting the author win:

```cpp
if (!estimated.empty() && !decision.outputs.contains("estimated_cost")) {
    decision.outputs["estimated_cost"] = std::move(estimated);
}
```

That keeps the spirit of the (now softened) "copied verbatim" seam: engine-attached data never displaces author-attached data.

ramkrishna2910 (commented):
The CostServices seam fits the existing ClassifierServices pattern nicely (header include invariant holds, ModelManager calls are internally locked, reporting-only keeps selection pure). Two issues inline.

Bekhouche (commented):
- `attach_estimated_cost` now catches exceptions from `cost_of` instead of letting them propagate, this also fixes the trace-loss case since it can no longer trigger the outer `catch` in `route()`
- Bails out early if `outputs["estimated_cost"]` is already set by the rule, so it no longer clobbers an author-provided value
- `try_get_model_info` now covers `model_exists()` inside its try block too

Added tests for cost_of throwing on both the matched-rule and default paths, and for the preserved author-set value.

SlawomirNowaczyk (changes requested):
Thanks, it's a good scaffolding for future features, but here are a few suggestions before this is merged:

### 1. Replace the `< 0 = unknown` sentinel with `std::optional<double>` in `resolve_cost_info`

The signature leans on magic sentinels:
```cpp
CostInfo resolve_cost_info(double cost_input_per_million,
                           double cost_output_per_million,
                           const std::map<std::string, json>& extras);
```
Callers must remember "negative means unknown," and it silently conflates "unset" with "someone put `-3.0` in the registry." Since the output already uses `std::optional`, make the input match:
```cpp
CostInfo resolve_cost_info(std::optional<double> cost_input_per_million,
                           std::optional<double> cost_output_per_million,
                           const std::map<std::string, json>& extras);
```
`make_router_cost_services` would pass `info->cost_input_per_million >= 0 ? std::optional{...} : std::nullopt`. Removes an easy-to-misuse convention and makes the boundary explicit.

### 2. Give `CostInfo` a `to_json()` and dedupe the serialization

The optional→json mapping in `attach_estimated_cost` is hand-rolled and untested in isolation:
```cpp
if (info.cost_tier) estimated["cost_tier"] = *info.cost_tier;
if (info.cost_input_per_million) estimated["cost_input_per_million"] = *info.cost_input_per_million;
// ...
```
Move it onto the struct as `json CostInfo::to_json() const` (returns `{}` when all fields empty). Then `attach_estimated_cost` becomes `auto estimated = info.to_json(); if (!estimated.empty()) ...`. This makes the field-name contract testable directly and keeps the schema's key names in one place instead of duplicated in an anonymous-namespace function.

### 3. Avoid the registry lookup / `build_cache()` on every routing decision

`attach_estimated_cost` runs unconditionally, and `try_get_model_info` calls `model_exists()`, which can trigger `build_cache()` — on the hot path, for every routed request, even `want_trace=false`. Two options:
- Skip cost work entirely when nobody will read it (e.g., gate on `want_trace` or an explicit `policy.report_cost` flag), or
- Have `make_router_cost_services` memoize `cost_of` results per candidate for the process lifetime (cost metadata is effectively static per model), invalidated on cache rebuild.

Right now it's "best-effort metadata" paying a registry cost on the critical path with no caching.

### 4. Validate `cost_tier` against the documented vocabulary

The docs say `cost_tier` is `"free" | "low" | "medium" | "high"`, but `resolve_cost_info` accepts any non-empty string verbatim:
```cpp
const std::string tier = extras_get<std::string>(extras, "cost_tier", "");
if (!tier.empty()) out.cost_tier = tier;
```
A typo like `"cheap"` flows straight into `outputs.estimated_cost` and downstream to the trust/audit layer. Since the engine is the one place that knows the vocabulary, check it against the allowed set and either drop-with-`LOG(DEBUG)` or pass through but warn. Keeps the annotation trustworthy for consumers.

### 5. Close the test gap: typed-vs-extras precedence *when both are present*

`test_resolve_cost_info_typed_and_extras` tests typed-only and extras-only, but never the case the precedence rule actually exists for — both set at once:
```cpp
// typed 1.25 should win over extras 99.0
std::map<std::string, json> extras = {{"cost_input_per_million", 99.0}};
auto r = resolve_cost_info(1.25, 10.0, extras);
check("typed price wins over extras when both present",
      near(*r.cost_input_per_million, 1.25));
```
Without this, a future refactor could flip the precedence and every existing test would still pass.

### 6. Potential log spamming
The `cost_of`-threw path logs `WARNING` **per request** — if one candidate persistently throws, that's log spam on the hot path. Keep track of it, and log-once-per-candidate.

Bekhouche (commented):
Addressed both comments in `67f55112`:

- `attach_estimated_cost` now bails out immediately if `outputs` already contains `estimated_cost`, so a rule author's own value is never overwritten by the engine.
- The `cost_of` call is wrapped in its own try/catch inside `attach_estimated_cost` (not just at one call site), so an exception from `CostServices`, whether router-backed or any other implementation, can never escape `route()` on either the matched-rule or default path. The `Decision` (including its trace) is fully built before `attach_estimated_cost` runs, so there's no risk of losing it if the cost lookup throws.

Also folded in the rest of the earlier review pass while I was in there:

- `std::optional<double>` instead of sentinel doubles
- `CostInfo::to_json()`
- per-candidate memoization of the cost lookup (avoids a registry hit on every routed request)
- `cost_tier` vocabulary validation (`free|low|medium|high` only)
- log-once-per-candidate on repeated `cost_of` failures
- unit tests for all of the above

Routing test suites and a full `lemond` build pass locally (WSL/Linux).

SlawomirNowaczyk (approved):
LGTM

ramkrishna2910 (approved):
Both inline comments addressed in 67f55112: cost_of is now wrapped in its own try/catch so route() keeps its never-throws contract, and a rule-authored outputs.estimated_cost is never overwritten. Verified the fixes in the diff along with the regression tests covering both cases. Resolving my threads.

PR #2765, opened 2026-07-20
feat: add smart thinking v0

fl0rianr:
## Summary

Add an experimental server-side Smart Thinking extension for Chat Completions.
It executes strictly recognized Arithmetic, Dispatch, and Constrained Selection
contracts deterministically and otherwise preserves one native model request.

Product presets:

- **Smart**: mode=auto, budget=1, execution_policy=verified_auto
- **Smart Extra**: mode=deep, budget=2, execution_policy=verified_auto
- **Strict Verified**: execution_policy=verified_required (developer option)

Smart Extra increases only the single native completion budget. It does not add
branches, voting, same-model judges, repair loops, or hidden retries.

## Deliberate PR scope

This is server-side only. It includes no UI changes and intentionally adds no
Smart Thinking tests, evaluator scripts, benchmark fixtures, CTest targets, or
CI workflow changes as this would bloat changes even more.

## Manual validation evidence

Test campaigns showed deterministic exact execution with zero model
calls for supported contracts and native fallback parity for unsupported tasks.
Verified normal model conversation is not broken and works fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -620,6 +620,7 @@ include_directories(
 set(SOURCES_CORE
     src/cpp/server/server.cpp
     src/cpp/server/collection_orchestrator.cpp
+    src/cpp/server/smart_thinking.cpp
     src/cpp/server/jobs/job_ops.cpp
     src/cpp/server/jobs/job_manager.cpp
     src/cpp/server/router.cpp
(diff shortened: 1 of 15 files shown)


jeremyfowers (changes requested):
Let's avoid merging a 13k LoC PR until there is adequate architecture and project scope discussion, to be followed by full documentation and test coverage.

As a starting point: what is the user story here? The PR description and endpoint documentation say how the feature is implemented, but they don't say what the feature is meant to accomplish for users.

PR #2766, opened 2026-07-20
feat: add jobs for all operations besides chat completion to the job engine

pwilkin:
Per discussion in #2718, add the remaining generation operations to the job engine.

cc: @bitgamma

### DIFF
--- a/docs/how-to-add-new-job-ops.md
+++ b/docs/how-to-add-new-job-ops.md
@@ -1,0 +1,137 @@
+# How to add a new op to the jobs endpoint
+
+## Summary
+
+The jobs engine uses a **pluggable op registry**. Adding a new op is additive: you
+declare a handler function, wire it to existing server/backend code via a provider
+lambda, and register it. No changes to the job graph, expression language, HTTP
+routes, or persistence layer are needed.
+
+## Files involved
+
+| File | Role |
+|------|------|
+| `src/cpp/include/lemon/jobs/job_ops.h` | Declares `OpProviders` (function-pointer contract) and `OpRegistry` |
+| `src/cpp/server/jobs/job_ops.cpp` | Implements `build_op_registry()` — the place where ops are **registered** |
+| `src/cpp/server/server.cpp` | Constructs `OpProviders` and passes lambdas that call into Router/backends |
+| `src/cpp/include/lemon/jobs/job_graph.h` | Validates op names at job-creation time against the registry's known set |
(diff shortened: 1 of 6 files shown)


fl0rianr (commented):
Thanks for adding the remaining generation operations to the job engine. The registry integration is straightforward and the overall direction makes sense, but I found a few issues that should be addressed before merge.

### 1. Generation jobs are not actually cancellable

All six new operations receive a `CancelFlag`, but none of them propagates it into the Router/backend request.

For the image operations, the flag is ignored entirely. For audio and 3D, the local `DataSink` also reports `is_writable() == true` unconditionally. Interrupting or deleting a job during a long image, audio, or 3D generation therefore cannot stop the active operation. The job continues holding the exclusive gate until the backend finishes.

This conflicts with the job lifecycle contract that `interrupt` cancels the current step promptly.

Please propagate cancellation through the Router and backend HTTP request, similarly to the existing chat cancellation path. For streaming operations, the sink should also stop accepting data once the job is cancelled.

### 2. audio_speech can return incorrectly labelled audio

The job operation defaults `response_format` to `mp3` without checking what the loaded TTS backend actually supports.

The normal HTTP endpoint performs format negotiation using `audio_speech_supported_formats()`. For example, OpenMOSS currently supports only WAV. A job request without an explicit format can therefore receive WAV bytes while returning:

```json
{
  "mime_type": "audio/mpeg",
  "response_format": "mp3"
}
```

That makes the output unusable or incorrectly decoded by later steps and clients.

Please reuse the same format selection and validation logic as the regular endpoint. The audio-generation operation should follow the same rule.

### 3. Persisting large binary results inside the job context is risky

Audio and GLB results are buffered completely, base64-encoded, and stored in the job context. Since the full context is persisted into `jobs.json`, a few large generation results can make that file very large and cause expensive rewrites on every job state update.

It would be safer to store generated media as separate job artifacts and keep only metadata and an artifact identifier or path in the context. At minimum, there should be a strict output-size limit.

### Test coverage

The new tests currently verify only that the operation names are accepted and that some operations fail without a backend. They do not exercise:

* a successful generation,
* format negotiation,
* interruption of an active generation,
* correctness of the returned MIME type,
* large-output handling.

`self.assertEqual(job["id"], job["id"])` also does not validate anything beyond the absence of an exception during creation.

Please fix especially cancellation and audio-format correctness. The binary persistence concern should also be resolved or explicitly bounded before these operations are used for potentially large media outputs.

Issue #4, opened 2025-05-16
Please support non-english system language


jeremyfowers:
@shcho1118 filed https://github.com/onnx/turnkeyml/issues/310, reopening here:

https://github.com/onnx/turnkeyml/blob/fe01e22dc89128a9174213050ff3cab13222f028/src/turnkeyml/common/system_info.py#L179
When I try to get windows power setting information, I get `unicodedecodeerror` if it's not in English (I use Korean).
I think we should avoid using something that depends on system language.

jeremyfowers:
FYI @amd-pworfolk

shcho1118:
Thank you for bringing that issue along when you separated it as well.
I'll keep looking for a workaround for the issue with not passing CI.

zidsi:
Any non ASCII characted in install path (perhaps even space?) will break reading config json. Might be related.

jeremyfowers:
I believe this should be solved by @Looong01 in https://github.com/lemonade-sdk/lemonade/pull/22 so I am closing the issue.

If there is still any problem please re-open the issue and let me know!

Issue #28, opened 2025-06-12
ID-F01 Extended Local Model Repo Manager, Integrated Web UI, API, Config [still valid on lemonade srv 804]


dorvan:
UI to list;   validate; and manage local/remote model repos
manage all events visible by terminal on repository  management
Inspiration: LMstudio. 

Permit to configure many local repo to generate config files (Global Config) to permite Lemonade to use the various configured Local Repo. 

the local Repo, mapped with AMD HF Repos can be managed mapping remote and local to check situation and update/upgrades.

jeremyfowers:
@dorvan thanks for your enthusiasm around the web app and model manager!

I see you have opened a few issues around this subject that have some overlapping ideas. For example, I see global config and env var management mentioned in a few issues.

Would you mind consolidating your ideas for model management and the html webapp into a single Issue? That would really help me keep track.

dorvan:
are not overlapped ideas but are the single feature request or problems documented related to analysis level, continue to test, use, and see the sorced and analyze problems and missing parts to manage thing in our perspective. so prepare tasks lists anche check what is done and what is not, and what is needed to include other, at divverent levels, this ticket is not in overlapping of others... thi is a baseline step, considering the current level of implementation. the other point with similar theme are next steps or tips.

jeremyfowers:
Ok. Anything you can do to keep the issues as few and concise as possible, while keeping the suggestions intact, will help me to keep track! I appreciate that you spent a lot of time to create 10 issues, but I'm not really able to service all of them in parallel since I need to spend the majority of my time coding.

dorvan:
Are only to wrote down in each part/fragment to permit you to consider in development and planning, on this side we does not pretend nothing.

dorvan:
I see some fixes applied at 8.0.1 version , in direction of my opinions, to mitigate the problem about this. Thanks. The model manager  seems now more complete, but will be useful adding repo dirs and auto completions for "add a model" o a "local repo dir scan" (to add lmstudio models for example) and adding also "duplicate" and "edit configuration" and "edit entry" to already installed models more than "delete model" already available

[...]

PR #2768, opened 2026-07-20
Update to ROCm 7.14

superm1:
This also updates llama.cpp builds against the ones built from ROCm 7.14 at the same time.

### ROCm 7.14 upgrade
- Updated llama.cpp to use upstream ROCm 7.14 artifacts
- Updated stable-diffusion.cpp to 7.14 build
- Updated TheRock download URL to use tarball-multi-arch

### New ROCm installation system
- AMD's recommended pip-wheel install (`rocm[libraries,device-gfxNNNN]`) into a lemonade-managed venv
- Fallback to TheRock tarball when Python/venv/pip is unavailable, the arch has no device wheel, or any step fails
- Record install method in `method.txt` so switching methods triggers reinstall
- Track user cancellation during pip install (stops rather than falling back to ~2 GB tarball)
- Parse pip wheel sizes for real progress bar with initial label and "Installing collected packages" phase
- New `rocm_install_method` config knob (`auto`, `wheel`, `tarball`) for air-gapped/pip-restricted environments

### ROCm runtime path fixes (Windows + wheel multi-dir layout)
- Centralized all loader-path sites through `get_therock_lib_paths()` + `join_runtime_dirs()` for consistent wheel runtime discovery
- Put all TheRock runtime dirs on PATH so `sd-server`/`sd-cli` ESRGAN resolve rocBLAS (wheel runtime splits HIP + BLAS across `_rocm_sdk_core/bin` and `_rocm_sdk_libraries/bin`)
- Force copy of `amdhip64_7.dll` on Windows to work around system32 DLL shadowing

### Hardening (commit d3a23b13 — review-driven)
- Fixed dead ROCm runtime that can't self-heal: wheel install idempotency now verifies recorded runtime dirs still exist before short-circuiting, so a deleted venv or moved cache reinstalls instead of leaving a permanently unrepairable dead loader path
- Fixed Windows UTF-8 path bug: HIP-runtime check now decodes UTF-8 dir strings via `path_from_utf8()` before iterating, so non-ASCII Windows profile paths (`C:\Users\José\…`) don't wrongly discard a good wheel install
- Fixed crash-window marker ordering: both wheel and tarball paths now write `version.txt` last (after `method.txt`) so an interrupted install can't be mistaken for a completed one or silently defeat an explicit `rocm_install_method`
- Fixed hot-path JSON reload: `get_therock_lib_paths()` (called on every ROCm-stable model load) and `cleanup_old_therock_versions()` now use the cached `backend_versions_config()` instead of re-reading from disk

### Bug fixes
- Fixed guaranteed divide-by-zero (SIGFPE) during streamed wheel install progress
- Fixed MSVC build errors in TheRock code paths (thread_local capture, variable scoping)
- Fixed `vllm_model_config.json` validation to reference correct repo

### CI & documentation
- Updated `validate_llamacpp.yml` to validate against `ggml-org` (not `lemonade-sdk`)
- Updated `validate_sdcpp.yml` for ROCm 7.14
- Added unit tests for `parse_pip_download_line()` and `is_concrete_gfx_arch()`
- Docs: prefer `rocm_bin=builtin` over `latest` on stable channel
- Docs: added `rocm_install_method` knob to configuration guide


docs/guide/configuration/vllm.md
```diff
@@ -48,7 +48,7 @@ Some GPU targets ride a different vLLM/ROCm wheel cadence than the default pin a
 The minimum path to a working gfx942 deployment today with the FP8 + MTP recipes:
 
 1. **Runtime asset.** Until the official asset ships, use the community-built, hardware-validated
-   tarball (`ianbmacdonald/vllm-rocm-cdna`, tag `vllm0.19.1-rocm7.13.0-gfx942`) by **manually sideloading**
+   tarball (`ianbmacdonald/vllm-rocm-cdna`, tag `vllm0.19.1-rocm7.14.0-gfx942`) by **manually sideloading**
```
ramkrishna2910: @ianbmacdonald why not merge this fork in to vllm-rocm?

superm1:
stable diffusion will need https://github.com/leejet/stable-diffusion.cpp/pull/1802 merged to update

ianbmacdonald:
@ramkrishna2910, answering your question from the `vllm.md` review thread ("why not merge this fork in to vllm-rocm?"): nothing would suit us better. That is exactly what lemonade-sdk/vllm-rocm#28 proposes. The fork exists only because we hold pull-only rights on vllm-rocm: we can't land the workflow change or mint releases there ourselves. The bundles are repackaged from AMD's official dcgpu wheel indices (no compile step), so adopting #28 plus one release-workflow run on your side produces official gfx942/gfx950 assets, and the fork's community bundles retire the same day.

One heads-up on this PR's docs bump while you're here: it references `vllm0.19.1-rocm7.14.0-gfx942`, but AMD has published no rocm7.14.0 dcgpu wheels yet (checked 2026-07-24 on both the gfx94X and gfx950 indices; the vLLM wheel line hasn't followed TheRock to 7.14). `rocm7.13.0` remains the buildable tag until they ship; happy to mint the 7.14 bundle the day the wheels appear.

superm1:
SD merged upstream, I will update the pins when the CI is ready.

jeremyfowers:
Will test on R9070XT

src/cpp/include/lemon/backends/backend_utils.h
```diff
@@ -143,16 +143,41 @@ namespace lemon::backends {
...
+         *  tarball when Python/venv/pip is unavailable or the wheel install fails. */
+        static void install_rocm_runtime(const std::string& arch, const std::string& version,
+                                         DownloadProgressCallback progress_cb = nullptr);
+
+        /** Attempt to install the ROCm runtime via pip wheels into a managed venv.
+         *  Returns true on success; false (leaving nothing behind) when Python is
+         *  missing, the arch has no device wheel, or any step fails, so the caller
+         *  can fall back to the tarball. */
+        static bool install_therock_wheels(const std::string& arch, const std::string& version,
+                                           DownloadProgressCallback progress_cb = nullptr);
```
jeremyfowers, line 157: Why not always use the tarball, then we aren't at the mercy of system python and there is only one code path to support.

.github/workflows/validate_sdcpp.yml
```diff
@@ -162,7 +162,7 @@ jobs:
```
jeremyfowers, line 163: sd-cpp rocm not working for me on gfx1201 on ubuntu 25.10:

```
2026-07-27 13:40:00.554 [Debug] (Server) GET /api/v1/system-info
2026-07-27 13:40:00.570 [Debug] (Server) POST /api/v1/load
2026-07-27 13:40:00.570 [Debug] (Server) ===== LOAD ENDPOINT ENTERED (Thread: 132292246390464) =====
2026-07-27 13:40:00.570 [Info] (Server) Ensuring model loaded: SDXL-Turbo 
2026-07-27 13:40:00.570 [Debug] (Router) Loading model: SDXL-Turbo (checkpoint: stabilityai/sdxl-turbo:sd_xl_turbo_1.0_fp16.safetensors, recipe: sd-cpp, type: image, device: cpu, residency: standard)
2026-07-27 13:40:00.570 [Warn] (Router) Pruning unavailable backend for model: SDXL-Turbo (state=watchdog_reset)
2026-07-27 13:40:00.570 [Info] (Router) Evicting model: SDXL-Turbo
2026-07-27 13:40:00.570 [Info] (Router) Evicted model: SDXL-Turbo
2026-07-27 13:40:00.570 [Debug] (AutoTune) get_available_memory_gb: CPU/NPU (AMD dGPU GTT proxy) total=15.20 GB, used=4.10 GB → 11.10 GB available compute_auto_context_size: SDXL-Turbo — GGUF: , blocks=0, kv_heads=0, key_len=0, swa_layers=0, scale=1.00 | kv_cache=0.25 MB/token (est) | memory: 11.10 GB avail, 6.46 GB weights → 4.64 GB for KV | ctx=19005 → 19005 resolve_auto_ctx_size: SDXL-Turbo → ctx_size=19005
2026-07-27 13:40:00.570 [Info] (Router) Auto-tune ctx_size resolved to 19005
2026-07-27 13:40:00.570 [Debug] (Router) Effective settings: sd-cpp_backend=rocm, sdcpp_args=(none), steps=4, cfg_scale=1.000000, width=512, height=512, sampling_method=(none), flow_shift=0.000000, merge_args=true, auto_evict=(none), evict_idle_timeout=300, downsize_idle_timeout=60, evict_weight_factor=1.000000, pinned=false
2026-07-27 13:40:00.570 [Debug] (SDServer) Created with log_level=debug
2026-07-27 13:40:00.570 [Debug] (Router) Created backend for recipe 'sd-cpp' via registry
2026-07-27 13:40:00.570 [Debug] (Router) Starting backend (this may take a moment)...
2026-07-27 13:40:00.570 [Info] (SDServer) Loading model: SDXL-Turbo
2026-07-27 13:40:00.570 [Debug] (SDServer) Per-model settings: sd-cpp_backend=rocm, sdcpp_args=(none), steps=4, cfg_scale=1.000000, width=512, height=512, sampling_method=(none), flow_shift=0.000000, merge_args=true, auto_evict=(none), evict_idle_timeout=300, downsize_idle_timeout=60, evict_weight_factor=1.000000, pinned=false
2026-07-27 13:40:00.570 [Debug] (BackendManager) Installing sd-cpp:rocm-stable
2026-07-27 13:40:00.700 [Debug] (sd-cpp Server) Found executable at: /home/jfowers/.cache/lemonade/bin/sd-cpp/rocm-stable/build/bin/sd-server
2026-07-27 13:40:00.700 [Debug] (SDServer) Using model: /home/jfowers/.cache/huggingface/hub/models--stabilityai--sdxl-turbo/snapshots/71153311d3dbb46851df1931d3ca6e939de83304/sd_xl_turbo_1.0_fp16.safetensors
2026-07-27 13:40:00.700 [Debug] (WrappedServer) sd-server will use port: 8002
2026-07-27 13:40:00.700 [Info] (SDServer) Starting server on port 8002 (backend: rocm)
2026-07-27 13:40:00.700 [Debug] (BackendUtils) Returning TheRock runtime path: /home/jfowers/.cache/lemonade/bin/therock/gfx1201-7.13.0/lib
2026-07-27 13:40:00.700 [Debug] (SDServer) Setting LD_LIBRARY_PATH=/home/jfowers/.cache/lemonade/bin/therock/gfx1201-7.13.0/lib:/home/jfowers/.cache/lemonade/bin/sd-cpp/rocm-stable/build/bin
2026-07-27 13:40:00.700 [Debug] (ProcessManager) Starting process with inherited output: /home/jfowers/.cache/lemonade/bin/sd-cpp/rocm-stable/build/bin/sd-server --listen-port 8002 -m /home/jfowers/.cache/huggingface/hub/models--stabilityai--sdxl-turbo/snapshots/71153311d3dbb46851df1931d3ca6e939de83304/sd_xl_turbo_1.0_fp16.safetensors -v
2026-07-27 13:40:00.702 [Info] (ProcessManager) Process started successfully, PID: 7909
2026-07-27 13:40:00.702 [Info] (SDServer) Process started with PID: 7909
2026-07-27 13:40:00.702 [Debug] (WrappedServer) Waiting for sd-server to be ready...
2026-07-27 13:40:00.902 [Info] (WrappedServer) sd-server is ready!
2026-07-27 13:40:00.902 [Info] (BackendWatchdog) Started watchdog for sd-server using http://127.0.0.1:8002/ (streaming=on, non_streaming=on)
2026-07-27 13:40:00.903 [Info] (SDServer) Server is ready at http://127.0.0.1:8002
2026-07-27 13:40:00.903 [Debug] (Router) Backend started successfully in 332ms
2026-07-27 13:40:00.903 [Info] (Router) Model loaded successfully. Total loaded: 2
2026-07-27 13:40:00.903 [Debug] (Server) POST /api/v1/load - 200
2026-07-27 13:40:00.904 [Debug] (Server) POST /api/v1/images/generations
2026-07-27 13:40:00.904 [Info] (Server) POST /api/v1/images/generations
2026-07-27 13:40:00.904 [Debug] (Server) Model already loaded: SDXL-Turbo (residency=standard)
2026-07-27 13:40:00.907 [Debug] (SDServer) Forwarding request to sd-server: {
2026-07-27 13:40:00.907 [Debug] (SDServer)   "cfg_scale": 1,
2026-07-27 13:40:00.907 [Debug] (SDServer)   "model": "SDXL-Turbo",
2026-07-27 13:40:00.907 [Debug] (SDServer)   "prompt": "a python <sd_cpp_extra_args>{\"sample_params\":{\"guidance\":{\"txt_cfg\":1.0},\"sample_steps\":4},\"seed\":42}</sd_cpp_extra_args>",
2026-07-27 13:40:00.907 [Debug] (SDServer)   "response_format": "b64_json",
2026-07-27 13:40:00.907 [Debug] (SDServer)   "seed": 42,
2026-07-27 13:40:00.907 [Debug] (SDServer)   "size": "512x512",
2026-07-27 13:40:00.907 [Debug] (SDServer)   "steps": 4
2026-07-27 13:40:00.907 [Debug] (SDServer) }
2026-07-27 13:40:05.271 [Error] (BackendWatchdog) sd-server backend marked unavailable: backend process exited during request; reaped exited backend process PID 7909 (exit_code=134)
2026-07-27 13:40:05.277 [Warn] (Router) Reloading model after backend watchdog reset: SDXL-Turbo
2026-07-27 13:40:05.289 [Debug] (Router) Loading model: SDXL-Turbo (checkpoint: stabilityai/sdxl-turbo:sd_xl_turbo_1.0_fp16.safetensors, recipe: sd-cpp, type: image, device: cpu, residency: standard)
2026-07-27 13:40:05.289 [Warn] (Router) Pruning unavailable backend for model: SDXL-Turbo (state=watchdog_reset)
2026-07-27 13:40:05.289 [Info] (Router) Evicting model: SDXL-Turbo
2026-07-27 13:40:05.289 [Info] (Router) Evicted model: SDXL-Turbo
2026-07-27 13:40:05.290 [Debug] (AutoTune) get_available_memory_gb: CPU/NPU (AMD dGPU GTT proxy) total=15.20 GB, used=3.00 GB → 12.20 GB available compute_auto_context_size: SDXL-Turbo — GGUF: , blocks=0, kv_heads=0, key_len=0, swa_layers=0, scale=1.00 | kv_cache=0.25 MB/token (est) | memory: 12.20 GB avail, 6.46 GB weights → 5.74 GB for KV | ctx=23511 → 23511 resolve_auto_ctx_size: SDXL-Turbo → ctx_size=23511
2026-07-27 13:40:05.291 [Info] (Router) Auto-tune ctx_size resolved to 23511
2026-07-27 13:40:05.291 [Debug] (Router) Effective settings: sd-cpp_backend=rocm, sdcpp_args=(none), steps=4, cfg_scale=1.000000, width=512, height=512, sampling_method=(none), flow_shift=0.000000, merge_args=true, auto_evict=(none), evict_idle_timeout=300, downsize_idle_timeout=60, evict_weight_factor=1.000000, pinned=false
2026-07-27 13:40:05.292 [Debug] (SDServer) Created with log_level=debug
2026-07-27 13:40:05.292 [Debug] (Router) Created backend for recipe 'sd-cpp' via registry
2026-07-27 13:40:05.292 [Debug] (Router) Starting backend (this may take a moment)...
2026-07-27 13:40:05.292 [Info] (SDServer) Loading model: SDXL-Turbo
2026-07-27 13:40:05.292 [Debug] (SDServer) Per-model settings: sd-cpp_backend=rocm, sdcpp_args=(none), steps=4, cfg_scale=1.000000, width=512, height=512, sampling_method=(none), flow_shift=0.000000, merge_args=true, auto_evict=(none), evict_idle_timeout=300, downsize_idle_timeout=60, evict_weight_factor=1.000000, pinned=false
2026-07-27 13:40:05.293 [Debug] (BackendManager) Installing sd-cpp:rocm-stable
2026-07-27 13:40:05.463 [Debug] (sd-cpp Server) Found executable at: /home/jfowers/.cache/lemonade/bin/sd-cpp/rocm-stable/build/bin/sd-server
2026-07-27 13:40:05.464 [Debug] (SDServer) Using model: /home/jfowers/.cache/huggingface/hub/models--stabilityai--sdxl-turbo/snapshots/71153311d3dbb46851df1931d3ca6e939de83304/sd_xl_turbo_1.0_fp16.safetensors
2026-07-27 13:40:05.464 [Debug] (WrappedServer) sd-server will use port: 8003
2026-07-27 13:40:05.464 [Info] (SDServer) Starting server on port 8003 (backend: rocm)
2026-07-27 13:40:05.465 [Debug] (BackendUtils) Returning TheRock runtime path: /home/jfowers/.cache/lemonade/bin/therock/gfx1201-7.13.0/lib
2026-07-27 13:40:05.465 [Debug] (SDServer) Setting LD_LIBRARY_PATH=/home/jfowers/.cache/lemonade/bin/therock/gfx1201-7.13.0/lib:/home/jfowers/.cache/lemonade/bin/sd-cpp/rocm-stable/build/bin
2026-07-27 13:40:05.465 [Debug] (ProcessManager) Starting process with inherited output: /home/jfowers/.cache/lemonade/bin/sd-cpp/rocm-stable/build/bin/sd-server --listen-port 8003 -m /home/jfowers/.cache/huggingface/hub/models--stabilityai--sdxl-turbo/snapshots/71153311d3dbb46851df1931d3ca6e939de83304/sd_xl_turbo_1.0_fp16.safetensors -v
2026-07-27 13:40:05.473 [Info] (ProcessManager) Process started successfully, PID: 7968
2026-07-27 13:40:05.473 [Info] (SDServer) Process started with PID: 7968
2026-07-27 13:40:05.473 [Debug] (WrappedServer) Waiting for sd-server to be ready...
2026-07-27 13:40:05.774 [Info] (WrappedServer) sd-server is ready!
2026-07-27 13:40:05.774 [Info] (BackendWatchdog) Started watchdog for sd-server using http://127.0.0.1:8003/ (streaming=on, non_streaming=on)
2026-07-27 13:40:05.774 [Info] (SDServer) Server is ready at http://127.0.0.1:8003
2026-07-27 13:40:05.774 [Debug] (Router) Backend started successfully in 482ms
2026-07-27 13:40:05.774 [Info] (Router) Model loaded successfully. Total loaded: 2
2026-07-27 13:40:05.782 [Debug] (SDServer) Forwarding request to sd-server: {
2026-07-27 13:40:05.782 [Debug] (SDServer)   "cfg_scale": 1,
2026-07-27 13:40:05.782 [Debug] (SDServer)   "model": "SDXL-Turbo",
2026-07-27 13:40:05.782 [Debug] (SDServer)   "prompt": "a python <sd_cpp_extra_args>{\"sample_params\":{\"guidance\":{\"txt_cfg\":1.0},\"sample_steps\":4},\"seed\":42}</sd_cpp_extra_args>",
2026-07-27 13:40:05.782 [Debug] (SDServer)   "response_format": "b64_json",
2026-07-27 13:40:05.782 [Debug] (SDServer)   "seed": 42,
2026-07-27 13:40:05.782 [Debug] (SDServer)   "size": "512x512",
2026-07-27 13:40:05.782 [Debug] (SDServer)   "steps": 4
2026-07-27 13:40:05.782 [Debug] (SDServer) }
2026-07-27 13:40:08.988 [Error] (BackendWatchdog) sd-server backend marked unavailable: backend process exited during request; reaped exited backend process PID 7968 (exit_code=134)
2026-07-27 13:40:08.988 [Error] (Server) Image generation backend error: {"error":{"details":{"backend":"sd-server","code":"backend_watchdog_reset","reason":"backend process exited during request","retryable":true},"message":"sd-server was unresponsive and has been reset by the backend watchdog. The model will be reloaded automatically on retry.","type":"backend_error"}}
2026-07-27 13:40:08.988 [Error] (Server) Error 500: POST /api/v1/images/generations
2026-07-27 13:40:08.989 [Debug] (Server) POST /api/v1/images/generations - 500
```

src/cpp/include/lemon/backends/backend_utils.h
superm1: Because it's so much bigger. This is a much smaller download (it doesn't have compiler at all)

.github/workflows/validate_sdcpp.yml
superm1: Is it a regression? Or already broke?

src/cpp/resources/backend_versions.json
```diff
@@ -46,11 +46,11 @@
...
-    "vulkan": "b9747",
-    "rocm-stable": "b9752",
+    "vulkan": "b10107",
+    "rocm-stable": "b10107",
     "rocm-nightly": "b1292",
-    "cuda": "b9851",
-    "metal": "b9747",
+    "cuda": "b10107",
+    "metal": "b10107",
     "cpu": "b9747"
```
jeremyfowers: Update for parity?

src/cpp/resources/backend_versions.json
```diff
@@ -46,11 +46,11 @@
     }
   },
   "llamacpp": {
-    "vulkan": "b9747",
-    "rocm-stable": "b9752",
+    "vulkan": "b10107",
+    "rocm-stable": "b10107",
     "rocm-nightly": "b1292",
```
jeremyfowers, line 51: Update for parity?

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1046,6 +1175,171 @@ namespace lemon::backends {
...
+#endif
+                    has_hip_runtime = true;
+                    break;
+                }
+            }
+            if (has_hip_runtime) {
+                break;
+            }
+        }
+        if (runtime_dirs.empty() || !has_hip_runtime) {
```
jeremyfowers, line 1365: **Blocking — an unknown `device-<arch>` extra makes pip exit 0, so this check passes with no device code installed.**

pip treats a nonexistent extra as a *warning*, not an error: `pip install "rocm[libraries,device-gfx900]==7.14.0"` installs `rocm` + the `libraries` extra and exits 0. `rocm-sdk-libraries` pulls in `rocm-sdk-core`, so `libamdhip64.so` / `amdhip64*.dll` is always present and `has_hip_runtime` becomes true. The install is then marked successful, `version.txt` is written, and the tarball fallback is never reached — but there are **no code objects for the user's GPU**. Inference fails later with "no kernel image available", and re-installing is a no-op because `is_therock_installed_for_current_arch` sees a matching `version.txt`.

This is not hypothetical. I pulled `Provides-Extra` from the `rocm` sdists on the index:

| ROCm | `device-*` extras published |
|---|---|
| 7.14.0 | gfx1010-1012, 1030-1036, 1100-1103, 1150-**1153**, 1200, 1201, 1250, **gfx908, gfx90a, gfx942, gfx950** |
| 7.13.0 | gfx1010-1012, 1030-1036, 1100-1103, 1150-1152, 1200, 1201 — **no gfx1153, no gfx908/90a/942/950** |

Because sd-cpp is pinned to **7.13.0** in this PR, a gfx1153 / gfx942 / gfx950 / gfx908 / gfx90a user installing `sd-cpp:rocm` lands on exactly this silent path.

Suggested fix: verify the device package actually resolved before declaring success — e.g. `<venv_python> -m pip show rocm-sdk-device-<arch>`, or import `_rocm_sdk_device_<arch>` — and return `false` (deleting `wheel_dir`) if it didn't.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/resources/backend_versions.json
```diff
@@ -61,17 +61,24 @@
...
   },
   "therock": {
-    "version": "7.13.0",
+    "version": "7.14.0",
+    "comment_recipe_versions": "Per-recipe ROCm version overrides. sd-cpp pinned to 7.13.0: the 7.14 gfx1151 Windows GEMM kernel is broken (ROCm/ROCm#6517).",
+    "recipe_versions": {
+      "sd-cpp": "7.13.0"
+    },
     "architectures": [
+      "gfx900",
```
jeremyfowers: **Blocking — gfx900 / gfx906 / gfx90c have no installable ROCm runtime at all, so adding them here turns a clean rejection into a failing (or silently broken) install.**

`will_install_therock()` gates on this list, so these users now enter the install flow. But neither path can serve them:

* **No wheel.** The `rocm` meta-package publishes no `device-gfx900`, `device-gfx906`, `device-gfx909` or `device-gfx90c` extra at 7.13.0 or 7.14.0. Combined with pip's warn-and-continue behaviour for unknown extras, that means the silent false-success described in my comment on `install_therock_wheels`.
* **No tarball.** I probed repo.amd.com: `therock-dist-linux-gfx90X-dcgpu-7.14.0.tar.gz` → **403**, and `gfx900` / `gfx906` / `gfx90c` verbatim → **403**.

Unless there's a source for these I couldn't find, I'd drop the three gfx9 additions.

**`gfx1153` is a genuinely good addition** — the `device-gfx1153` extra exists at 7.14 and `therock-dist-linux-gfx1153-7.14.0.tar.gz` returns 200.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

jeremyfowers (commented):
Full review of the ROCm 7.14 update. Overall: the per-recipe version plumbing is solid — clean threading through `will_install_therock` / `is_therock_installed_for_current_arch` / `get_therock_lib_path`, the `sd-cpp` recipe key matches `server_models.json`, and `validate_llamacpp.yml` correctly still reads the top-level pin. I'd merge that half as-is.

My concerns are concentrated in the new pip-wheel path and the expanded arch table. Five things I'd consider blocking, detailed inline:

1. **An unknown `device-<arch>` extra makes pip exit 0**, so the wheel install reports success with no device code objects installed and caches that as a valid `version.txt`. Verified against the published extras: this hits gfx1153/gfx942/gfx950/gfx908/gfx90a for sd-cpp, which is pinned to 7.13.0.
2. **gfx900 / gfx906 / gfx90c have neither a wheel extra nor a working tarball** (403 on every candidate URL), so adding them to `architectures` converts a clean "unsupported" outcome into a failing or silently broken install. gfx1153 is a good addition.
3. **`url_mapping` inconsistencies**: `gfx909` is mapped but not in `architectures` (typo for 906?), every `gfx90X-dcgpu` target 403s while `gfx908`/`gfx90a` work verbatim, and `gfx110X-dgpu` 403s while `gfx110X-all` returns 200 (dead tarball fallback for RX 7000-series).
4. **`get_therock_lib_path()` can now return a path *list***, but four backends (trellis, thinksound, openmoss, acestep) derive sibling `llvm` directories from it and produce garbage.
5. **The weekly `validate_sdcpp` run will revert the deliberate `rocm-stable: master-791` pin**, since `SDCPP_BASE_UPDATE_BACKENDS` includes `rocm-stable`.

The rest are medium/minor: reusing the existing `rocm-sdk path --root` machinery instead of the bespoke Python probe (which would also fix #4), an empty pinned-version list wiping every install, `import venv` being insufficient on Debian, no progress or cancellation during a ~730 MB pip run, `--isolated` for the pip invocation, and a docs gap in `docs/guide/configuration/llamacpp.md`.

Verified findings were checked directly against `repo.amd.com` (tarball HTTP status per arch/version, and `Provides-Extra` from the `rocm` 7.13.0/7.14.0 sdists) rather than inferred.

---
*This review was written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/resources/backend_versions.json
```diff
@@ -98,8 +106,11 @@
       "gfx120X"
     ],
     "url_mapping": {
+      "gfx900": "gfx90X-dcgpu",
+      "gfx909": "gfx90X-dcgpu",
```
jeremyfowers: **Blocking — this mapping block is internally inconsistent and its gfx90X target doesn't exist.**

* **`gfx909` is mapped here but is not in `architectures`**, so it is unreachable — `will_install_therock()` rejects the arch before the mapping is ever consulted. Meanwhile `gfx906` *is* in `architectures` with no mapping. That reads like a `906`/`909` typo.
* **Every `gfx90X-dcgpu` entry resolves to a 403.** I verified `therock-dist-linux-gfx90X-dcgpu-<ver>.tar.gz` at both 7.13.0 and 7.14.0, on both the old `tarball/` and new `tarball-multi-arch/` bases. Pre-existing for gfx908/gfx90a — and notably `therock-dist-linux-gfx908-7.14.0.tar.gz` and `-gfx90a-` both return **200**, so those two mappings should simply be deleted; `install_therock()` already falls back to the arch verbatim.
* Same class of bug, also pre-existing but a one-line fix while you're in this table: **`gfx110X-dgpu` → 403, while `gfx110X-all` → 200.** That's the RX 7000-series desktop parts (gfx1100/1101/1102); their tarball fallback is currently dead.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1192,18 +1486,49 @@ namespace lemon::backends {
...
+                    }
+                    if (!joined.empty()) {
+                        joined += sep;
+                    }
+                    joined += line;
+                }
+                if (!joined.empty()) {
+                    LOG(DEBUG, "BackendUtils")
+                        << "Returning ROCm wheel runtime path: " << joined << std::endl;
+                    return joined;
```
jeremyfowers: **Blocking — this returns a `;`/`:`-separated *list*, but four callers treat the return value as a single directory.**

Before this PR `get_therock_lib_path()` always returned exactly one directory (`<install>/lib`, or `<install>/bin` on Windows — see the tarball branch just below). These callers derive sibling paths from it and now produce garbage:

* `trellis_server.cpp:114-119`, `thinksound_server.cpp:118-123`, `openmoss_server.cpp:110-115`, `acestep_server.cpp:149-154`
  * Linux: `therock_lib + ":" + therock_lib + "/llvm/lib"` — with `"A:B"` this expands to `A:B:A:B/llvm/lib`, so the llvm dir becomes `_rocm_sdk_libraries/lib/llvm/lib`, which doesn't exist. The llvm runtime silently drops off `LD_LIBRARY_PATH`.
  * Windows: `fs::path("A;B").parent_path() / "lib" / "llvm" / "bin"` — pure garbage.
* `llamacpp_server.cpp:469` pushes the joined string through `fs::absolute()`, which is at best accidentally harmless.

Suggested fix: return `std::vector<std::string>` (or add a distinct `get_therock_runtime_dirs()`) and let each caller join with its own separator. See my comment on `query_wheel_runtime_dirs` for an alternative that keeps a single-root return value and fixes this for free.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/sdcpp/sdcpp_server.cpp
```diff
@@ -326,7 +321,7 @@ void SDServer::load(const std::string& model_name,
         if (resolved_backend == "rocm-stable") {
             std::string rocm_arch = SystemInfo::get_rocm_arch();
             if (!rocm_arch.empty()) {
-                std::string therock_bin = BackendUtils::get_therock_lib_path(rocm_arch);
+                std::string therock_bin = BackendUtils::get_therock_lib_path(rocm_arch, get_therock_version());
                 if (!therock_bin.empty()) {
                     new_path = path_to_utf8(fs::absolute(path_from_utf8(therock_bin))) + ";" + new_path;
```
jeremyfowers: Related to the multi-path return value from `get_therock_lib_path()`: when the wheel install is active, `therock_bin` here is a `;`-separated list, and `fs::absolute(path_from_utf8(...))` is being applied to the whole string as if it were one path. It survives today only because the string happens to start with a drive letter. Worth revisiting once that function returns a list.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/resources/backend_versions.json
```diff
@@ -61,17 +61,24 @@
     "metal": "v1.8.4"
   },
   "sd-cpp": {
-    "cpu": "master-721-8caa3f9",
-    "rocm-stable": "master-721-8caa3f9",
-    "vulkan": "master-721-8caa3f9",
-    "cuda": "master-726-db48014",
-    "metal": "master-721-8caa3f9"
+    "cpu": "master-793-78124b6",
+    "rocm-stable": "master-791-b8bf676",
```
jeremyfowers: **Blocking — the weekly sd-cpp workflow will silently revert this deliberate pin.**

`validate_sdcpp.yml` sets `SDCPP_BASE_UPDATE_BACKENDS: cpu,vulkan,rocm-stable,metal`, and the scheduled run rewrites all four keys to the same upstream tag. This PR intentionally splits them (`rocm-stable: master-791-b8bf676` vs `cpu`/`vulkan`/`metal` at `master-793-78124b6`), presumably because sd-cpp asset names embed the ROCm version (`get_install_params()` builds `...-rocm-<version>-x64.zip`) and 793 only ships rocm-7.14 assets.

Next Sunday's run bumps `rocm-stable` to 793, the requested filename stays `...-rocm-7.13.0-...`, and sd-cpp ROCm installs start 404ing. Either drop `rocm-stable` from `SDCPP_BASE_UPDATE_BACKENDS`, or teach the updater/validator about the `therock.recipe_versions` pin. A comment next to this block noting *why* rocm-stable lags would help either way.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1018,22 +1018,151 @@ namespace lemon::backends {
...
+            return "";
+        }
+
+        // Ask the managed venv's Python where the ROCm runtime libraries landed.
+        // The rocm-sdk-core/-libraries wheels expose them under
+        // _rocm_sdk_core/{bin,lib} and _rocm_sdk_libraries/{bin,lib}; we print
+        // every candidate and keep the directories that actually exist so the
+        // logic is correct on both Windows (bin) and Linux (lib).
+        std::vector<std::string> query_wheel_runtime_dirs(const std::string& venv_python) {
+            static const char* probe =
```
jeremyfowers, line 1089: **Consider reusing the ROCm-root machinery already in this file instead of a bespoke probe.**

`query_rocm_sdk_root()` (same file, ~line 884) already runs `rocm-sdk path --root` and parses its output defensively, and `validate_rocm_root()` already confirms the HIP runtime is present. The managed venv ships the `rocm-sdk` console script in `venv/bin` (`venv\Scripts` on Windows), so `<venv>/bin/rocm-sdk path --root` would give you a single canonical root, reuse the existing validation, and let `get_therock_lib_path()` keep returning one directory — which dissolves the multi-path caller breakage I flagged separately.

Two fragilities in the current probe if you keep it:

* `os.path.dirname(importlib.import_module(m).__file__)` raises `TypeError` when a module is a namespace package (`__file__ is None`). The sdist's own `devel_test.py:239` notes `_rocm_sdk_libraries` *is* a namespace package in "kpack-split mode".
* One failed import fails the whole `-c`, so `query_wheel_runtime_dirs` returns `{}` → `wheel_dir` is deleted → full fallback to the tarball, discarding a ~730 MB download (core 414 MB + libraries 143 MB + device-gfx1151 173 MB).

Make it per-module and tolerant of `__file__ is None` (fall back to `__path__`).

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1018,22 +1018,151 @@ namespace lemon::backends {
...
 #ifdef __linux__
         fs::path therock_base = fs::path(utils::get_downloaded_bin_dir()) / "therock";
 
         if (!fs::exists(therock_base)) {
             return;
         }
 
+        // Keep every pinned version so a recipe on a different ROCm version
+        // than the default isn't wiped.
+        const std::vector<std::string> keep = pinned_therock_versions();
```
jeremyfowers: If `pinned_therock_versions()` returns empty — missing or malformed `therock` object — `keep` is empty and this loop `remove_all`s **every** installed version. The old signature took `current_version`, so it structurally couldn't do that. Cheapest guard: always seed the keep-list with the version currently being installed.

Separately: this function now parses `backend_versions.json`, and it's called at the tail of `install_therock()` *after* a successful extraction. A parse error there fails the whole install even though the runtime is on disk and fine.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1046,6 +1175,171 @@ namespace lemon::backends {
...
+            }
+        }
+        {
+            std::ofstream vf(version_file);
+            vf << version;
+        }
+
+        // Keep every pinned version so a recipe on a different ROCm version isn't wiped.
+        fs::path wheels_base = fs::path(utils::get_downloaded_bin_dir()) / "therock-wheels";
+        const std::vector<std::string> keep_versions = pinned_therock_versions();
```
jeremyfowers: This re-implements `cleanup_old_therock_versions()` (~line 1146) against a different base directory, and carries the same "empty keep-list wipes everything, including the install that just finished" hazard.

They're also inconsistent by platform: `cleanup_old_therock_versions()` is `#ifdef __linux__`-only, while this wheel cleanup runs on Windows too. Worth folding both into one helper parameterized by base directory.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1018,22 +1018,151 @@ namespace lemon::backends {
...
+            const std::vector<std::string> names = {"python3", "python"};
+#endif
+            for (const auto& name : names) {
+                std::string path = utils::find_executable_in_path(name);
+                if (path.empty()) {
+                    continue;
+                }
+                // Confirm the venv module is importable before committing to it.
+                int rc = utils::ProcessManager::run_process_with_output(
+                    path, {"-c", "import venv"}, nullptr, /*working_dir=*/"",
```
jeremyfowers: `import venv` isn't a sufficient probe on Debian/Ubuntu — `python3-venv` is a separate package, and `import venv` succeeds without it; `python -m venv` then fails on the missing `ensurepip`. That's probably the single most common Linux configuration for this to hit.

`{"-c", "import venv, ensurepip"}` makes the fallback immediate instead of costing a failed subprocess.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1046,6 +1175,171 @@ namespace lemon::backends {
...
+        }
+
+        if (progress_cb) {
+            DownloadProgress p;
+            p.file = spec;
+            p.file_index = 1;
+            p.total_files = 1;
+            p.percent = 100;
+            p.complete = true;
+            progress_cb(p);
```
jeremyfowers, line 1403: This is the only `progress_cb` invocation on the wheel path, so a ~730 MB install shows 0% in the UI for its entire duration and then jumps to complete.

Also worth noting `DownloadProgressCallback` returns `bool` as a cancellation signal, and that return is discarded here — unlike the tarball path, this install can't be cancelled, and a hung pip blocks for the full 1800 s timeout. Parsing pip's `Downloading ... (x/y MB)` lines out of `log_line` into progress events would fix both.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1046,6 +1175,171 @@ namespace lemon::backends {
...
+            utils::path_to_utf8(venv_dir / "bin" / "python");
+#endif
+
+        const std::string index_url = "https://repo.amd.com/rocm/whl-multi-arch/";
+        const std::string spec =
+            "rocm[libraries,device-" + arch + "]==" + version;
+
+        rc = utils::ProcessManager::run_process_with_output(
+            venv_python,
+            {"-m", "pip", "install", "--no-input", "--index-url", index_url, spec},
```
jeremyfowers: Two notes on the pip invocation:

* `--index-url` overrides `PIP_INDEX_URL`, but `PIP_EXTRA_INDEX_URL` and `~/.config/pip/pip.conf` are still honoured, so a user- or host-level extra index can shadow `rocm-sdk-*`. `--isolated` ignores env vars and config files and closes that.
* `--only-binary=:all:` is **not** an option here — `rocm` itself is published only as an sdist (`rocm-7.14.0.tar.gz`), so pip builds it with build isolation and fetches `setuptools` from this same index. It happens to be published there, but that's an implicit dependency on AMD's index contents that deserves a comment.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1046,6 +1175,171 @@ namespace lemon::backends {
...
+        install_therock(arch, version, progress_cb);
+    }
+
+    bool BackendUtils::install_therock_wheels(const std::string& arch, const std::string& version,
+                                              DownloadProgressCallback progress_cb) {
+#if !defined(__linux__) && !defined(_WIN32)
+        (void)arch; (void)version; (void)progress_cb;
+        return false;
+#else
+        if (!is_concrete_gfx_arch(arch)) {
```
jeremyfowers, line 1221: The hex-suffix test accepts archs that have no `rocm-sdk-device-*` package — `gfx900`, `gfx90c` and `gfx906` all pass it. Combined with pip's warn-and-continue on unknown extras, that's the false-success path I flagged on the runtime check below. Gating on the actual published extras (or on a wheel-availability entry in `backend_versions.json`) would be more honest than a shape test.

Separately: there's no way to opt out of the wheel path. Air-gapped or pip-restricted environments now always pay a venv creation plus a pip attempt before falling back. A `RuntimeConfig` knob (e.g. `rocm_install_method=tarball`) would fit the existing config surface.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1018,22 +1018,151 @@ namespace lemon::backends {
...
+        if (!recipe.empty() && therock.contains("recipe_versions") &&
+            therock["recipe_versions"].is_object() &&
+            therock["recipe_versions"].contains(recipe) &&
+            therock["recipe_versions"][recipe].is_string()) {
+            return therock["recipe_versions"][recipe].get<std::string>();
+        }
+        return therock["version"].get<std::string>();
+    }
+
+    std::vector<std::string> BackendUtils::pinned_therock_versions() {
```
jeremyfowers: Minor: this and `get_therock_version_for_recipe()` load and parse `backend_versions.json` on every call, and `get_therock_lib_path(arch)` now does too — including from `handle_image_upscale()`, so once per upscale request. It's load-path rather than hot-path, so not urgent, but a `static` cache is a couple of lines.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/backend_utils.cpp
```diff
@@ -1192,18 +1486,49 @@ namespace lemon::backends {
...
-        std::string config_path = utils::get_resource_path("resources/backend_versions.json");
-        json config = utils::JsonUtils::load_from_file(config_path);
-
-        if (!config.contains("therock") || !config["therock"].contains("version")) {
-            throw std::runtime_error("backend_versions.json is missing 'therock.version'");
+        // Prefer the wheel install when present: its runtime spans two dirs,
+        // recorded in runtime_paths.txt at install time.
+        {
+            fs::path paths_file =
+                fs::path(get_therock_wheel_dir(rocm_arch, version)) / "runtime_paths.txt";
```
jeremyfowers, line 1570: `runtime_paths.txt` is trusted without checking the recorded directories still exist. If the venv is deleted or the cache directory moves, this returns dead `LD_LIBRARY_PATH` / `PATH` entries and nothing triggers a reinstall (`is_therock_installed_for_current_arch` only looks at `version.txt`). A cheap `fs::is_directory` filter here, returning `""` when nothing survives, would let the normal install flow repair it.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/include/lemon/backends/backend_utils.h
```diff
@@ -143,16 +143,41 @@ namespace lemon::backends {
...
-        static void cleanup_old_therock_versions(const std::string& current_version);
+        /** Remove TheRock installs whose version is no longer pinned. */
+        static void cleanup_old_therock_versions();
 
-        /** Get TheRock lib directory path if available, or empty string if not needed */
+        /** Get TheRock lib directory path (top-level version), or "" if not installed. */
         static std::string get_therock_lib_path(const std::string& rocm_arch);
 
+        /** Get TheRock lib directory path for a specific ROCm version, or "" if not installed. */
+        static std::string get_therock_lib_path(const std::string& rocm_arch, const std::string& version);
```
jeremyfowers: Doc nit that matters given the implementation: "lib directory path" is no longer accurate — when the wheel install is active this returns a `;`/`:`-separated *list* of directories. That mismatch is what the four Python-backend callers are tripping over (see my comment on the implementation).

Also worth updating `docs/guide/configuration/llamacpp.md` ("Reusing a System-Installed ROCm"): it documents the resolution order as `ROCM_PATH` → `rocm-sdk path --root` → platform default, and says Lemonade "downloads its own ROCm runtime (TheRock)". This PR adds a fourth, now-preferred source (the managed wheel venv), a second on-disk location (`therock-wheels/`), and the possibility of multiple concurrently-installed ROCm versions.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

.github/workflows/validate_sdcpp.yml
```diff
@@ -162,7 +162,7 @@ jobs:
 
           checks = {
               'cpu': [
-                  (base_assets, rf'^sd-{re.escape(base_short)}-bin-win-avx2-x64\.zip$'),
+                  (base_assets, rf'^sd-{re.escape(base_short)}-bin-win-cpu-x64\.zip$'),
```
jeremyfowers, line 166: This matches the rename in `SDServer::get_install_params()` — good.

Suggestion while this file is open: a check that every entry in `therock.architectures` has **either** a published `device-<arch>` extra on the wheel index **or** a `url_mapping` target whose tarball returns 200 would catch the gfx900/gfx906/gfx90c and `gfx90X-dcgpu`/`gfx110X-dgpu` problems I flagged in `backend_versions.json` — including the pre-existing ones — and would keep the arch table honest as AMD's publishing cadence shifts.

More generally, nothing in this PR is covered by tests. `get_therock_version_for_recipe()` and `pinned_therock_versions()` are pure functions over JSON and are cheap to cover.

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/server/backends/sdcpp/sdcpp_server.cpp
```diff
@@ -70,14 +70,9 @@ std::string trim_to_major_minor(const std::string& version) {
...
 std::string get_therock_version() {
-    auto config = JsonUtils::load_from_file(utils::get_resource_path("resources/backend_versions.json"));
-    if (!config.contains("therock") || !config["therock"].is_object() ||
-        !config["therock"].contains("version") || !config["therock"]["version"].is_string()) {
-        throw std::runtime_error("backend_versions.json is missing 'therock.version'");
-    }
-    // stable-diffusion.cpp release assets include full ROCm runtime version in filenames
-    // (for example: rocm-7.12.0), so keep the patch component.
-    return trim_version_prefix(config["therock"]["version"].get<std::string>());
+    // sd-cpp is pinned to ROCm 7.13: the 7.14 gfx1151 Windows GEMM kernel is
```
jeremyfowers: The rationale is duplicated verbatim in `backend_versions.json` (`comment_recipe_versions`). One canonical location is enough — the JSON is the better home since that's where someone changing the pin will be looking, and this copy will drift.

(The *why* comment itself is right per AGENTS.md; it's only the duplication I'd trim.)

---
*Written by Claude (Claude Code) on behalf of @jeremyfowers.*

src/cpp/include/lemon/backends/backend_utils.h
jeremyfowers: I'm not convinced python will be tolerated by our ISV/OEM partners. Is there a way to force the tarball approach?

jeremyfowers:
Confirmed that this change breaks the download manager in the GUI, when I try to upgrade on Windows it says "installing llamacpp:rocm 0B" and doesn't mention TheRock at all.

Edit: the terminal also doesn't have any notion of download progress. The whole thing is hung on an "Installing llama-server.exe" message.

jeremyfowers:
Confirmed that the tarball fallback triggers on a Windows PC that has no Python installation. (but it doesn't show up in the download manager either).

Update: successful installation, LLM model loaded and runs on ROCm.

jeremyfowers:
I just got 35 TPS for Qwen3.6-35B-MTP on strix halo ROCm on a short prompt. Target TPS is more like 50.

For the same prompt I only got 15-25 TPS on this Vulkan tag, target is more like 70.

There is some major perf issue here.

superm1:
Thanks for the thorough review @jeremyfowers. Pushed 190304e4 addressing it.

**Blocking**
1. Wheel install now runs `pip show rocm-sdk-device-<arch>` after the pip step and falls back to the tarball (deleting the wheel dir) when the device package didn't resolve — no more caching a runtime with no code objects.
2. Dropped `gfx900`/`gfx906`/`gfx90c` from `architectures`. I re-verified against `repo.amd.com` before removing — every candidate 403s at both versions:

   | arch | device wheel | tarball 7.13.0 | tarball 7.14.0 |
   |---|---|---|---|
   | gfx900 / gfx906 / gfx90c | 403 | 403 | 403 |
   | gfx908 / gfx90a | 200 | 200 | 200 |
   | gfx1153 | 200 | 403 | 200 |

3. `url_mapping`: removed `gfx909` (typo), and the `gfx908`/`gfx90a` → `gfx90X-dcgpu` entries (that family target 403s; both resolve verbatim, confirmed 200).
4. `get_therock_lib_path()` now returns a `std::vector<std::string>`; added `join_runtime_dirs(dirs, include_llvm)` and updated all callers (llama/sd/whisper/upscale = no llvm, the four Python backends = llvm siblings). `rocm-sdk path --root` isn't usable here since we do library-only installs. Also: a wheel install whose recorded runtime dirs no longer exist now reinstalls instead of leaving a dead `LD_LIBRARY_PATH`.
5. Dropped `rocm-stable` from `SDCPP_BASE_UPDATE_BACKENDS` (+ comment) so the weekly run can't revert the 7.13 pin; it's still validated via `SDCPP_VALIDATED_LABELS`.

**Medium/minor**
- `ensurepip` probe; `--isolated`; namespace-package-tolerant runtime probe; folded both cleanups into one helper with an empty-keep guard (+ current version, wrapped so a parse error can't fail a good install, runs on Windows too); cached `backend_versions.json`; pip `Downloading` lines now emit progress and honor cancellation.
- New `rocm_install_method` (`auto`/`wheel`/`tarball`, env `LEMONADE_ROCM_INSTALL_METHOD`) so ISV/air-gapped setups can force the tarball and skip Python entirely — this is the tarball-force knob you asked for.
- Trimmed the duplicated pin rationale (JSON is canonical); documented the wheel-venv source and the new knob in `llamacpp.md` + the config table.

**Left open**
- `llamacpp.rocm-nightly` (`b1292`) / `cpu` (`b9747`) parity: not bumped blindly since pointing a pin at a tag without published assets 404s — happy to bump once matching `b10107` builds exist.
- gfx1153 kept: correct for llama.cpp (7.14, both paths 200), but sd-cpp@7.13 has no gfx1153 assets (wheel is 7.14-only, tarball 403), so it fails cleanly there. Upstream asset gap, not fixable in this PR.

jeremyfowers:
> llamacpp.rocm-nightly (b1292) / cpu (b9747) parity: not bumped blindly since pointing a pin at a tag without published assets 404s — happy to bump once matching b10107 builds exist.

Not a big deal as long as we have time to run the standard llamacpp auto upgrade flow before the next release.

superm1:
@fl0rianr why are there required jobs getting skipped?  Is this related to some CI job cleanups?

fl0rianr:
> @fl0rianr why are there required jobs getting skipped? Is this related to some CI job cleanups?

As said in the docs you can define if more extensive testing is done before the queue.
image

Changing that gives a wohle new CI run

superm1:
@jeremyfowers can you please remove `C:\windows\System32\amdhip64_7.dll` from the runners and see if that helps?

jeremyfowers:
> @jeremyfowers can you please remove `C:\windows\System32\amdhip64_7.dll` from the runners and see if that helps?

It would be a major regression if we went from "lemonade works no matter what" to "lemonade dies if a random DLL is present." Let's see if we can harden lemonade instead of altering the CI runner?

superm1:
> > @jeremyfowers can you please remove `C:\windows\System32\amdhip64_7.dll` from the runners and see if that helps?
> 
> It would be a major regression if we went from "lemonade works no matter what" to "lemonade dies if a random DLL is present." Let's see if we can harden lemonade instead of altering the CI runner?

I don't think it's a lemonade problem. It feels a lot more like a fundamental rocm issue. But yeah I want it working now matter what's around.

superm1:
I did force experiment with force copying dll and that didn't help.

Can you repro this locally by chance?

jeremyfowers:
I'm still not happy about the use of Python here. Here's my hail mary suggestion to get us back onto a single no-python code path: is it possible to download TheRock's .whl file using CURL, and then just unzip it?

superm1:
> I'm still not happy about the use of Python here. Here's my hail mary suggestion to get us back onto a single no-python code path: is it possible to download TheRock's .whl file using CURL, and then just unzip it?  

It's multiple wheels. I would guess this is going to turn into a pip manifest parsing implementation basically if we did that.

src/cpp/include/lemon/backends/backend_utils.h
superm1: I've added a configuration option that will let you set policy of how you want it.

jeremyfowers:
> > I'm still not happy about the use of Python here. Here's my hail mary suggestion to get us back onto a single no-python code path: is it possible to download TheRock's .whl file using CURL, and then just unzip it?
> 
> It's multiple wheels. I would guess this is going to turn into a pip manifest parsing implementation basically if we did that.

Ok that would be even worse, nevermind.

jeremyfowers:
@claude review

jeremyfowers:
@claude will the wheel-based TheRock installation correctly appear in the download manager in the gui?

jeremyfowers:
@claude review

jeremyfowers:
looks like you've got some substantial issues @superm1

ianbmacdonald:
Independent hardware datapoint for the wheels-into-venv path, measured 2026-08-14 on a rented MI300X (gfx942, Ubuntu 24.04) by hand-running the same mechanics this PR automates:

The 7.14 set installs and serves on gfx942: torch 2.13.0+rocm7.14.0 (hip 7.14.60850) sustains 620 TFLOP/s warm fp16 (50x 8192^3 matmul, first loop discarded).

One trap hit along the way: a bare `pip install torch==2.13.0+rocm7.14.0` from whl-multi-arch LOOKS healthy (imports, enumerates the GPU, `cuda.is_available()` True) and then dies at the first kernel with `hipErrorInvalidImage`. The per-arch device wheels (`amd-torch-device-gfx942`, `rocm-sdk-device-gfx942`) are a required explicit install; nothing pulls them in and nothing warns until kernel launch. The `is_concrete_gfx_arch()` -> rocm-sdk-device wheel gating in this diff covers that case, and this run confirms from CDNA hardware that the wheel path works when the device wheel is present and fails invisibly when it is not. If the device-wheel resolve ever misses, the first visible symptom is that kernel-launch error.

Also re-verified on the current head: the gfx942/gfx950 `url_mapping` rows and `architectures` entries carry through the 7.14.0 bump intact.

jeremyfowers:
@claude review are there any merge blockers remaining?

jeremyfowers:
```
2026-08-14 14:08:01.147 [Debug] (Server) GET /api/v1/system-info
2026-08-14 14:08:01.168 [Debug] (Server) POST /api/v1/install
2026-08-14 14:08:01.168 [Info] (Server) Installing backend: llamacpp:rocm
2026-08-14 14:08:01.170 [Debug] (BackendManager) Installing llamacpp:rocm-stable
2026-08-14 14:08:01.171 [Debug] (Server) POST /api/v1/install - 200
2026-08-14 14:08:01.293 [Info] (llamacpp Server) Upgrading llama-server from b10362 to b10431
2026-08-14 14:08:01.293 [Info] (llamacpp Server) Installing llama-server (version: b10431)
2026-08-14 14:08:01.293 [Debug] (llamacpp Server) Downloading to: /tmp/llamacpp_rocm-stable_b10431_llama-b10431-bin-ubuntu-rocm-7.14-x64.tar.gz
2026-08-14 14:08:01.293 [Debug] (llamacpp Server) Downloading from: https://github.com/ggml-org/llama.cpp/releases/download/b10431/llama-b10431-bin-ubuntu-rocm-7.14-x64.tar.gz
2026-08-14 14:08:01.454 [Error] (HttpClient) [Download] HTTP error 404 for URL: https://github.com/ggml-org/llama.cpp/releases/download/b10431/llama-b10431-bin-ubuntu-rocm-7.14-x64.tar.gz
2026-08-14 14:08:01.455 [Error] (DownloadManager) worker exception id=backend:llamacpp:rocm error="Failed to download llama-server from: https://github.com/ggml-org/llama.cpp/releases/download/b10431/llama-b10431-bin-ubuntu-rocm-7.14-x64.tar.gz - Download failed after 6 attempts.
2026-08-14 14:08:01.455 [Error] (DownloadManager) Last error: HTTP error 404 for URL: https://github.com/ggml-org/llama.cpp/releases/download/b10431/llama-b10431-bin-ubuntu-rocm-7.14-x64.tar.gz"
```

FYI for support purposes (up to you if you want to patch in the code): the above failure happened when I tried to load Muse Glimmer after following these instructions: https://lemonade-server.ai/news/muse-glimmer.html

it appears that being on `llamcpp.rocm_bin=latest` and TheRock 7.13, then trying to load a model with rocm on your branch, results in failure.

if I `lemonade config set llamacpp.rocm_bin=builtin` and then attempt to reload everything works as expected.

ianbmacdonald:
Ran this PR's wheel install path end to end on two GPUs (gfx1100 RX 7900 XT / Ubuntu 26.04 through the packaged service, gfx942 MI300X / Ubuntu 24.04) yesterday, from head 63c7711. One of the two defects found is already fixed by 5b9e700; the other is still present on the current head.

**Still live at 5b9e700: every Linux wheel install is silently rejected and falls back to the tarball.**

pip completes (`Successfully installed rocm-7.14.0 rocm-sdk-core-7.14.0 rocm-sdk-device-gfx1100-7.14.0 rocm-sdk-libraries-7.14.0`) and the install then logs `ROCm wheels installed but the HIP runtime could not be located; using TheRock tarball`. The Linux probe in `install_therock_wheels` requires a file named exactly `libamdhip64.so`, but pip never materializes the unversioned dev symlink: the wheel ships `_rocm_sdk_core/lib/libamdhip64.so.7` only (verified in the venv). So the wheel path downloads ~730 MB, deletes it, and downloads the 2.2 GB tarball on top, on every Linux install, and the graceful fallback masks it completely: the user lands on a working runtime and never learns the wheel path did not stick. (The `/opt/rocm` probe at the other `libamdhip64.so` site is fine, since a system install does have the dev symlink; it is only the wheel venv that does not.) The Windows arm of this check already prefix-matches; the Linux arm can do the same:

```diff
 #else
-                if (name == "libamdhip64.so") {
+                // Wheels carry only versioned sonames (libamdhip64.so.7); the
+                // unversioned dev symlink is never materialized by pip.
+                if (name.rfind("libamdhip64.so", 0) == 0) {
 #endif
```

With that one-line change the path works as designed on both architectures:

- gfx1100: wheel install completes (21 s on a warm pip cache), `runtime_paths.txt` resolves to the venv's `_rocm_sdk_core/lib` + `_rocm_sdk_libraries/lib`, and `llamacpp:rocm-stable` (b10375) serves GPU-resident (11.6 GB VRAM, 72.5 tok/s decode on a 14B Q4).
- gfx942: `rocm[libraries,device-gfx942]==7.14.0` resolves, installs in 49 s, and serves at 258 tok/s decode (10.3 GB VRAM) with no tarball tree on the machine at all, so the venv runtime is provably the one serving.

Knob checks on the same builds: `rocm_install_method=wheel` with the index unreachable fails loudly (CLI exit 1, the intended error text); `=tarball` skips the wheel path and accepts an existing tarball install; `auto`'s fallback was exercised by the defect above.

**Confirming the crash fix in 5b9e700 from hardware:** yesterday's runs on 63c7711 hit the same SIGFPE this push fixes: `lemond.service: Main process exited, code=killed, status=8/FPE` one second after pip's first `Downloading ... (24 kB)` line, twice, with the CLI showing `Error: Could not connect to Lemonade server` and all loaded models dropping. The cause matched what 5b9e700 addresses (size parse cut at the first whitespace leaving `bytes_total` 0, then the unguarded completion-percent division), and every number above was measured with an equivalent guard in place, so the fix is validated on both archs, not just compiled.

src/cpp/server/backends/llamacpp/llamacpp_server.cpp
```diff
@@ -183,7 +183,7 @@ InstallParams LlamaCppServer::get_install_params(const std::string& backend, con
     }
 
     if (resolved_backend == "rocm-stable") {
-        params.repo = "lemonade-sdk/llama.cpp";
+        params.repo = "ggml-org/llama.cpp";
```
jeremyfowers: I guess we shouldn't merge this LoC until we're certain llamacpp+rocm is going to be supported indefinitely upstream?

superm1: updated

superm1:
## Automated review pass (Claude Opus 4.8) — pushed as 83634f2b

Ran a focused review of this PR's C++ changes (ROCm wheel-install + loader-path work), across correctness, cross-platform, efficiency, and conventions angles, plus an independent second-pass review. Fixed every real finding in `src/cpp/server/backends/backend_utils.cpp`; verified with a clean build (no warnings) and passing `PipDownloadParserTest`/`GfxArchGatingTest`.

**Fixed:**

1. **Dead ROCm runtime couldn't self-heal (high).** `install_therock_wheels`'s idempotency short-circuit returned `true` on matching version markers without checking the recorded runtime dirs still exist, contradicting the liveness gate in `is_therock_installed_for_current_arch`. A deleted venv or moved cache left a permanently unrepairable dead loader path. Now also requires `therock_wheel_runtime_alive()`.
2. **UTF-8 path bug on Windows (medium).** The HIP-runtime check built `fs::directory_iterator` from a UTF-8 `std::string`, which MSVC decodes as ANSI — a non-ASCII Windows profile path (`C:\Users\José\…`) would fail the check and discard a good wheel install. Now routed through `path_from_utf8()`.
3. **Crash-window marker ordering (low).** Both wheel and tarball paths wrote `version.txt` before `method.txt`; an interrupt in between could silently defeat an explicit `rocm_install_method`. `version.txt` is now the last-written commit marker in both paths.
4. **Redundant JSON reload on the hot path (efficiency).** `get_therock_lib_paths()` (every ROCm-stable load) and `cleanup_old_therock_versions()` re-parsed `backend_versions.json` from disk despite the PR's own `backend_versions_config()` cache. Both now use the cache.
5. **Include ordering (convention).** Reordered the added `#include <string>` per the repo's alphabetical-include rule.

**Reviewed and intentionally left as-is:** wheel-progress `file_index`/`total_files` display (cosmetic; explicit `percent` drives the bar) and the documented back-compat `get_therock_lib_path()` singular accessor.

**Confirmed non-issues:** the `thread_local` cancel flag is safe (line callback runs inline on the caller's thread on both Linux and Windows), `parse_pip_download_line`/`is_concrete_gfx_arch` are correct, and all percent divisions are guarded against divide-by-zero.

jeremyfowers (approved):
Please ensure the PR body is 100% accurate before merging so that our auditing agent does not get confused. For example, I'm not sure this statement is true: "Stable diffusion is pinned to ROCm 7.13 for now."

PR #2771, opened 2026-07-20
fix: resolve CORS blockage on MCP Gateway Endpoint

abn:
### Summary
This PR resolves the CORS blockage encountered when trying to load MCP tools through the built-in MCP Gateway (`POST /mcp`) from browser-only modes or custom web application clients.

### Root Cause
- **Blocked custom header**: During tool discovery and lists requests, frontend clients send the `Mcp-Session-Id` custom header. This was rejected during preflight OPTIONS requests because it was not listed in the server's static `Access-Control-Allow-Headers` list.

### Changes
1. **Server CORS updates** (in `src/cpp/server/server.cpp`):
   - Added `Mcp-Session-Id` to the `Access-Control-Allow-Headers` list (preserving all existing headers).
   - Removed the redundant duplicate uppercase `MCP-Protocol-Version` entry (which is already case-insensitively covered by the existing `mcp-protocol-version`).
   - Kept dynamic origin validation in the pre-routing handler as the single source of truth for `Access-Control-Allow-Origin` (preventing duplicate/conflicting wildcard origin values in browser responses).
2. **Integration Tests** (in `test/server_mcp.py`):
   - Updated `test_000_cors_preflight` to verify preflight behavior case-insensitively, asserting the exact dynamically reflected origin, and removing verification of unused expose headers.
   - Updated `test_000_cors_headers_on_post` to verify that standard POST requests reflect the allowed origin correctly.
   - Added `test_000_cors_disallowed_origin` to verify that requests (OPTIONS and POST) from disallowed origins correctly return `403 Forbidden`.

### Verification
- Manually verified via `curl` that CORS preflight returns a 204 with the exact reflected `Access-Control-Allow-Origin` and lists `Mcp-Session-Id` in `Access-Control-Allow-Headers`.
- Verified that requests from disallowed origins fail with `403 Forbidden`.
- Verified all 17 integration tests in `test/server_mcp.py` pass.

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -1677,7 +1677,7 @@ void Server::setup_cors(httplib::Server &web_server) {
     // Set CORS headers for all responses
     web_server.set_default_headers({
         {"Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS"},
-        {"Access-Control-Allow-Headers", "Content-Type, Authorization, X-Client-Session-Id, X-Account-Session-Id, mcp-protocol-version, traceparent"}
+        {"Access-Control-Allow-Headers", "Content-Type, Authorization, X-Client-Session-Id, X-Account-Session-Id, mcp-protocol-version, traceparent, Mcp-Session-Id"}
     });
 
     // Handle preflight OPTIONS requests
--- a/test/server_mcp.py
+++ b/test/server_mcp.py
@@ -74,6 +74,64 @@ class McpGatewayTests(ServerTestBase):
     # JSON-RPC envelope
     # ---------------------------------------------------------------------
 
+    def test_000_cors_preflight(self):
+        """OPTIONS /mcp preflight must return 204, reflect concrete allowed origin, and allow required headers."""
+        headers = {
+            "Origin": "http://localhost:3000",
+            "Access-Control-Request-Method": "POST",
+            "Access-Control-Request-Headers": "MCP-Protocol-Version, Mcp-Session-Id",
+        }
+        response = requests.options(
+            MCP_URL,
+            headers={**_auth_headers(), **headers},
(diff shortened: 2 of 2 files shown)


fl0rianr (commented):
**Blocking:** Please remove the global `Access-Control-Allow-Origin: *` header. Lemonade already validates `Origin` against `LEMONADE_ALLOWED_ORIGINS` in the pre-routing handler and reflects the concrete allowed origin. Adding `*` as a default can therefore produce two `Access-Control-Allow-Origin` values (`*` plus the reflected origin), which browsers reject even though `requests` may combine them and let the current test pass. Please keep the existing origin policy as the single source of truth, add only the MCP-specific allowed/exposed headers, make the test assert the exact reflected origin, and add a negative test expecting `403` for a disallowed origin. The session-ID wording should also be corrected: the current MCP implementation does not issue `Mcp-Session-Id`, and the test only sends a fake ID and checks that the header is exposed rather than verifying a returned session ID.

fl0rianr (commented):
The origin handling is fixed now, thank you. One issue remains: the PR still claims that the server returns the client-provided `Mcp-Session-Id` and that exposing it allows the browser to establish a session, but `/mcp` never copies this request header into the response and the current MCP documentation explicitly states that Lemonade does not issue session IDs. The POST test therefore only verifies that `Access-Control-Expose-Headers` is present; it does not verify that any `Mcp-Session-Id` is actually returned. Per the MCP transport specification, the server may issue a session ID on the initialize response, and only then should the client send it on subsequent requests. Please either remove the session-return wording, the unused expose header, and reword the test to cover only the intended CORS behavior, or implement actual session-ID issuance and assert the returned header. Also, the additional uppercase `MCP-Protocol-Version` entry is redundant because CORS header-name matching is case-insensitive and `mcp-protocol-version` was already allowed.

fl0rianr (approved):
LGTM, merge it when npm test green

PR #2773, opened 2026-07-21
fix(custom_args): treat negative numbers as values, not flags

coolduderules:
## Problem

`build_custom_args_map` flagged any token starting with `-` as a flag, so negative numeric values (`-1`, `-0.5`, `-.5`) became phantom map keys instead of attaching to their owning flag. This corrupted recipes using args like `--cache-ram -1`:

- The value detached from its flag (e.g. `--cache-ram` lost its `-1`).
- Since `std::map` keys are unique, identical negatives across flags deduped — `--cache-ram -1 --reasoning-budget -1` dropped **both** values.
- On re-serialization the stray `-1` key sorted ahead of real flags, scrambling arg order.

## Fix

A token is treated as a value only if the complete token is a valid negative number. This covers integers, decimals, leading-dot decimals, and scientific notation such as `-1`, `-0.5`, `-.5`, and `-1e5`, while keeping tokens like `-1foo` and `-.future` as flags.

## Test

New `test/cpp/test_custom_args.cpp` (registered as `CustomArgsTest`) covers negative integers, decimals, scientific 

---
Checklist:
- [x] C++ builds cross-platform (header-only change, std-only)
- [x] No new endpoints / no recipe changes
- [x] Test added and registered in CMakeLists.txt

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -307,8 +307,9 @@ jobs:
           cmake --build --preset default --target test_job_graph
           cmake --build --preset default --target test_origin_utils
           cmake --build --preset default --target test_model_residency
+          cmake --build --preset default --target test_custom_args
           cd build
-          ctest --output-on-failure -R "ProcessManagerTest|DirectoryWatcher|LatestVersionFallback|RoutingPolicy(Contract|Evaluator|Registry|Semantic|LlmRouter|Deterministic|Engine|Parser|Store)Test|ModelManagerCollectionValidationTest|RoutingClassifierServicesTest|NetworkUtilsTest|mcp_client_config|JobExprTest|JobGraphTest|OriginUtilsTest|ModelResidencyTest"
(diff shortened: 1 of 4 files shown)


fl0rianr (changes requested):
Thanks for fixing this — the underlying issue is real, and the added regression coverage for `-1`, `-0.5`, and `-.5` is helpful.

I found two things that would be good to address before merging:

1. The new `is_flag` check is a bit broader than the PR description suggests. It treats every token starting with `-<digit>` or `-.` as a value, including strings such as `-1foo` or `-.future`. Since this helper is used during argument merging and re-serialization, those tokens could silently become values of the previous flag. It would be safer to detect a complete negative numeric token instead, including optional decimal and exponent parts.

2. The new `CustomArgsTest` is registered in CMake, but it does not appear to be built or included in the explicit CTest regexes used by the Linux and macOS CI workflows. So the new regression test currently does not protect the PR in CI.

A few small test improvements would also make the fix more robust:

* verify the complete map and ensure no phantom `"-1"` key exists,
* add the scientific-notation cases mentioned in the PR description,
* add counterexamples such as `-1foo` and `-.future`,
* update the test comment to use `-R '^CustomArgsTest$'`, since `-R custom_args` does not match the registered test name.

The overall direction looks good; with these adjustments this should be a solid and well-covered fix.

coolduderules:
@fl0rianr Both completely valid oversights on my end—I think this should just about address them. Numeric detection now validates the complete token (integer, decimal, and scientific notation) so `-1foo` and `-.future` remain flags rather than being silently absorbed as values. The test is now built and wired into both the Linux and macOS ctest regexes, with cases expanded to assert the full map, scientific notation, and counterexamples. Let me know if I missed anything, parsing passes all the edge cases I've come up with but I may have missed something.

fl0rianr (approved):
Thanks! LGTM now.
I adapted the PR descriptions a bit.

CI fail is unrelated, I'm merging it now.

PR #2777, opened 2026-07-21
fix(mcp): enable "Save and connect" for a new external MCP server

pwilkin:
The submit button's disabled guard was `busyId === '__new__' || busyId === draft.id`. A new server's draft.id is '' and the idle busyId is also '', so `busyId === draft.id` was always true and the button stayed grayed out; only editing an existing server (non-empty id) worked. Disable it only while this form is saving, matching setBusyId(draft.id || '__new__').


fl0rianr (approved):
Nice!

PR #2778, opened 2026-07-21
Update trellis checksums to the pinned v0.4.3 assets

ramkrishna2910:
The trellis backend pin moved to v0.4.3 (#2726) but the `checksums` section in `backend_versions.json` still covered only v0.1.5. Since an empty expected hash silently disables verification in `HttpClient::download_file`, all trellis v0.4.3 downloads currently install unverified — defeating the tamper protection added in #2710 for contributor-hosted binaries.

- Replaced the stale v0.1.5 entries with SHA-256 hashes for all six v0.4.3 assets.
- Hashes come from the GitHub release asset digests and were independently verified by downloading and hashing the two vulkan assets.

Found during v11.5.0 release testing (release checklist #2774); should be cherry-picked to `release-v11.5.0`.

### DIFF
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -34,13 +34,13 @@
         }
       },
       "pwilkin/trellis.cpp": {
-        "v0.1.5": {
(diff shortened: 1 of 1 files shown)


pwilkin (approved):
Sorry, missed that one.

PR #2780, opened 2026-07-21
Fix rocm-stable loader paths and prefer vulkan for trellis on AMD GPUs

ramkrishna2910:
Two related fixes found while release-testing v11.5.0 (checklist #2774); targeting the next release, not the v11.5.0 branch.

**1. `rocm-stable` never got the TheRock loader paths.** `trellis_server.cpp`, `openmoss_server.cpp`, and `acestep_server.cpp` gate the TheRock `PATH`/`LD_LIBRARY_PATH` setup on `backend == "rocm"`, but the stable channel's backend name is `rocm-stable` — the same files already normalize with `rfind("rocm", 0) == 0` in `get_install_params`, so this aligns the env-var branch with that. On Windows this was masked because the AMD driver ships `amdhip64_7.dll` in System32; on Linux without a system ROCm install the subprocess would start with no ROCm runtime on the loader path.

**2. Trellis on AMD GPUs now defaults to vulkan.** `trellis-server`'s ROCm build dies **silently** at the first HIP compute on Windows/gfx1151 (Strix Halo) — reproduced identically with the v0.1.5 and v0.4.3 assets, with the GPU correctly detected (`using ROCm0`) and gfx1151 code objects present in `ggml-hip.dll`, and regardless of whether TheRock or the driver HIP runtime is used. Vulkan generates correctly on the same hardware (~7 min for a 125k-tri textured GLB at res 512). Listing `vulkan` ahead of `rocm` in the support matrix makes default-backend selection pick the working path; users can still opt into ROCm via `trellis_backend`. Generated docs regions regenerated via `gen_backend_boilerplate.py`.

@pwilkin — the silent ROCm crash itself looks like a trellis.cpp issue (nothing on stderr, no WER entry; process exits right after `ggml_cuda_init` succeeds when the first kernel launches, both v0.1.5 and v0.4.3, Windows + gfx1151 + driver HIP 7.x). Happy to file it upstream with full repro details if useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,11 +309,6 @@ Lemonade supports multiple inference engines for LLM, speech, TTS, and image gen
     <tr>
       <td rowspan="3"><strong>3D generation</strong></td>
       <td rowspan="3"><code>trellis</code> (experimental)</td>
-      <td><code>rocm</code></td>
-      <td>Supported AMD ROCm iGPU/dGPU families (ROCm via TheRock)</td>
-      <td>Windows, Linux</td>
-    </tr>
-    <tr>
       <td><code>cuda</code></td>
       <td>NVIDIA GPUs</td>
       <td>Windows, Linux</td>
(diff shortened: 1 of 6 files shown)


pwilkin:
Hm, can we keep the Vulkan-first path for Windows only? My main testing ground at this point is Strix Halo on Linux and it runs perfectly fine on ROCm. I'm wondering what the Windows failure is, will try to connect a monitor and keyboard, reboot to Windows and see.

superm1:
This was NAK'ed and is doing too much in one PR.  If you want these things to land together, please use stacked PRs, or open multiple independent PRs.

PR #2781, opened 2026-07-21
Fix FLM model discovery: tolerate the flm CLI stdout banner

ramkrishna2910:
**Fresh installs currently register zero FLM models on both Windows and Linux.** flm v0.9.45 prints an `[FLM]  Fetching models from: ...` line on **stdout** before its JSON payload (even with `--quiet`), so every `flm list/version --json` parse in `fastflowlm_models.cpp` fails with `parse error at line 1, column 2 ... last read: '[F'` and discovery bails.

- This is the root cause of the persistently red **Test .deb - flm** CI job (fresh cache each run).
- The Windows **Test .exe - flm** job passes only because its cache dir is persistent — models registered by pre-banner flm versions mask the breakage. Reproduced the failure on Windows with a fresh cache on the release-v11.5.0 build.
- The v0.9.45 pin shipped in v11.0.0, so the released version is affected too — this is not a v11.5.0 regression, but v11.5.0 would ship it again.

Fix: parse from the first `{` in the flm CLI output, applied to all four parse sites.

Verified end-to-end on a Ryzen AI (Strix Halo) machine with a fresh cache: discovery now registers 5 FLM models (incl. `llama3.2-1b-FLM`), and pull + NPU chat completion work. Found while release-testing v11.5.0 (checklist #2774).

### DIFF
--- a/src/cpp/server/backends/fastflowlm/fastflowlm_models.cpp
+++ b/src/cpp/server/backends/fastflowlm/fastflowlm_models.cpp
@@ -140,6 +140,16 @@ std::string find_flm_binary() {
     }
 }
 
+// flm >= 0.9.45 prints an "[FLM]  Fetching models from: ..." banner on stdout
+// before the JSON payload, even with --quiet, so parse from the first '{'.
+static json parse_flm_cli_json(const std::string& output) {
+    const size_t brace = output.find('{');
+    if (brace == std::string::npos) {
+        throw std::runtime_error("no JSON object in flm list output: " + output.substr(0, 200));
+    }
+    return lemon::utils::JsonUtils::parse(output.substr(brace));
+}
+
 std::vector<std::string> flm_installed_checkpoints() {
     std::vector<std::string> installed_models;
 
@@ -167,7 +177,7 @@ std::vector<std::string> flm_installed_checkpoints() {
 
     // Parse output: { "models": [ { "name": "modelname:tag", ... }, ... ] }
     try {
-        json j = lemon::utils::JsonUtils::parse(output);
+        json j = parse_flm_cli_json(output);
         if (j.contains("models") && j["models"].is_array()) {
             for (const auto& model : j["models"]) {
                 if (model.contains("name") && model["name"].is_string()) {
@@ -247,7 +257,7 @@ std::vector<ModelInfo> flm_discover_models() {
 
     // Parse output: { "models": [ { "name": "modelname:tag", "footprint": 1.23, ... }, ... ] }
     try {
-        json j = lemon::utils::JsonUtils::parse(output);
+        json j = parse_flm_cli_json(output);
         if (j.contains("models") && j["models"].is_array()) {
             for (const auto& m : j["models"]) {
                 if (m.contains("name") && m["name"].is_string()) {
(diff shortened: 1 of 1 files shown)


src/cpp/server/backends/fastflowlm/fastflowlm_models.cpp
```diff
@@ -140,6 +140,16 @@ std::string find_flm_binary() {
...
+// flm >= 0.9.45 prints an "[FLM]  Fetching models from: ..." banner on stdout
+// before the JSON payload, even with --quiet, so parse from the first '{'.
+static json parse_flm_cli_json(const std::string& output) {
+    const size_t brace = output.find('{');
+    if (brace == std::string::npos) {
+        throw std::runtime_error("no JSON object in flm list output: " + output.substr(0, 200));
+    }
+    return lemon::utils::JsonUtils::parse(output.substr(brace));
+}
+
```
superm1, line 152: can we please get a bug or pull request filed against https://github.com/FastFlowLM/FastFlowLM?  This should really be fixed there.

superm1:
I'm really confused.  This is commented out in the FLM codebase....


https://github.com/FastFlowLM/FastFlowLM/blob/7032b3c58373f7ebf41f8e01188ba1770a923ba2/src/src/main.cpp#L483

superm1:
> I'm really confused. This is commented out in the FLM codebase....
> 
> https://github.com/FastFlowLM/FastFlowLM/blob/7032b3c58373f7ebf41f8e01188ba1770a923ba2/src/src/main.cpp#L483

Oh, J/K.  It's because it was fixed one hour ago :stuck_out_tongue: 

https://github.com/FastFlowLM/FastFlowLM/commit/7032b3c58373f7ebf41f8e01188ba1770a923ba2

superm1:
@ZaneNi any chance to tag a new FLM with that fix instead of picking up a hack here?

zaneni6:
> @ZaneNi any chance to tag a new FLM with that fix instead of picking up a hack here?

@superm1 Maybe not tag for this time. We were trying to provide clearer information about the `model_list` in FLM, but that change inadvertently broke Lemonade's model-fetching flow.

A windows zip patch has already been prepared for the Windows Portable version of Lemonade. 

It should pass the test now.

superm1:
> > @ZaneNi any chance to tag a new FLM with that fix instead of picking up a hack here?
> 
> @superm1 Maybe not tag for this time. We were trying to provide clearer information about the `model_list` in FLM, but that change inadvertently broke Lemonade's model-fetching flow.
> 
> A windows zip patch has already been prepared for the Windows Portable version of Lemonade. 
> 
> It should pass the test now.
> 
> 

Okay, that's good to hear.

ramkrishna2910:
Closing — the FLM team fixed this upstream by replacing the v0.9.45 release assets in place (windows zip updated 2026-07-21 21:16 UTC, linux tar.gz 20:12 UTC). Verified on a fresh cache with an unpatched release-v11.5.0 build: `flm list --json` now emits clean JSON and discovery registers all FLM models (incl. `llama3.2-1b-FLM`). No lemonade-side change needed. If flm ever regrows a stdout banner this branch has the tolerant-parse fix ready.

PR #2783, opened 2026-07-21
style(ui): unify dropdown select design system with stroke chevrons and accent glow

abn:
### Summary

Modernizes and unifies `<select>` dropdown controls across Lemonade GUI3 (Web App & Desktop modes):

- **Modern Stroke Chevron:** Replaces heavy solid block triangle indicators (`▼`) with crisp line-stroke SVG chevrons (`∨`) matching the app's Lucide iconography.
- **Soft Focus Glow:** Eliminates stark white focus outlines (`outline: 2px solid var(--accent-focus)`), substituting smooth border highlights and soft accent focus glows (`border-color: var(--accent-fg); box-shadow: 0 0 0 2px var(--accent-soft)`).
- **Design System Parity:** Refines border radii (`var(--radius-md)`), surface token backgrounds (`var(--surface-1)` / `var(--surface-2)`), and option list styling across dark themes.

**Current web version:**
Screenshot From 2026-07-22 01-16-23

**On Tauri builds this is notably worse:**
image

**This is the new version:**
Screenshot From 2026-07-22 01-20-13

### Verification

- **Automated Tests:** All 238 Playwright UI and accessibility (`axe-core`) tests passed (0 failures).
- **Build Validation:** Verified with production Web App bundle (`cmake --build --preset default --target web-app`) and `lemond` C++ server executable (`11.0.0`).

### DIFF
--- a/src/app/package.json
+++ b/src/app/package.json
@@ -8,6 +8,7 @@
     "build:renderer": "webpack --mode development",
     "build:renderer:prod": "webpack --mode production",
     "watch:renderer": "webpack --watch --mode development",
+    "typecheck": "tsc --noEmit",
     "tauri": "tauri",
     "dev": "tauri dev",
     "build": "npm run build:renderer:prod",
--- a/src/app/src/styles/styles.css
+++ b/src/app/src/styles/styles.css
@@ -3000,26 +3000,204 @@ input, textarea {
               background var(--duration-fast) var(--ease-out);
 }
 
+select,
 .select {
   appearance: none;
   -webkit-appearance: none;
-  background-image: linear-gradient(45deg, transparent 50%, var(--text-tertiary) 50%),
-                    linear-gradient(135deg, var(--text-tertiary) 50%, transparent 50%);
-  background-position: calc(100% - 14px) 50%, calc(100% - 10px) 50%;
-  background-size: 4px 4px, 4px 4px;
+  background-color: var(--surface-1);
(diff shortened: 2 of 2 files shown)


fl0rianr (approved):
Thanks for improving styling! The new chevron and focus treatment look much cleaner overall.

I found two CSS cascade issues that should be fixed before merging:

1. Several later rules use the `background:` shorthand on `<select>` elements, which resets the global `background-image` back to `none`. As a result, the new SVG chevron disappears in at least:

   * `.custom-model-form__grid select`
   * `.router-editor select`
   * `.global-settings-field select`
   * `.global-settings-pin-add select`

   Please replace the relevant `background: var(...)` declarations with `background-color: var(...)`, or explicitly restore the chevron image afterward.

2. `.message-role-select` previously defined its own border and text color, but this PR removes both. The new global `select` rule does not provide them, so this control can fall back to browser/WebView defaults, especially in the dark theme.

   Please restore:

   ```css
   border: 1px solid var(--border-subtle);
   color: var(--text-primary);
   ```

One smaller theme issue: the SVG chevron uses a hard-coded `#a1a1aa` stroke. It looks appropriate in the dark theme, but is rather faint in the light theme. A light-theme override using the equivalent of `--text-tertiary` would keep the contrast consistent.

The general direction is good, but I would address these cascade regressions before merging. Then go thanks.

abn:
Thanks for the review @fl0rianr! I've updated the PR to address all feedback:

- **Background shorthand fix:** Switched `background:` shorthand to `background-color:` in `.custom-model-form__grid select`, `.router-editor select`, `.global-settings-field select`, and `.global-settings-pin-add select` to preserve the chevron `background-image`.
- **Restored `.message-role-select` styling:** Restored `border: 1px solid var(--border-subtle)` and `color: var(--text-primary)` to prevent unstyled fallbacks in dark mode/WebView environments.
- **Light theme contrast:** Added a light-theme stroke override (`[data-theme="light"] select, [data-theme="light"] .select`) using `%2352525b` (zinc-600) for consistent contrast against light backgrounds.

PR #2790, opened 2026-07-23
fix(server): preserve WebSocket auth state and support desktop client origins

abn:
Fixes an issue where WebSocket log streaming fails under LEMONADE_API_KEY authentication because libwebsockets releases HTTP headers post-handshake, and resolves 403 errors for desktop webview clients.

#### Changes
- **WebSocket Auth Persistence:** Captured authenticated tokens during the HTTP upgrade handshake into a thread-safe map (`pending_authenticated_tokens_`) and transferred them to `handle_connection()` on session establishment, maintaining `state.authenticated = true` after libwebsockets frees header memory.
- **Generalized Desktop Origin Support:** Updated `is_loopback_origin()` in `origin_utils.h` to recognize non-`http(s)`/`ws(s)` custom desktop application schemes (`file://`, `app://.`, `jan://`, `vscode-webview://`, etc.) and RFC 6761 `*.localhost` subdomains as local loopback origins.
- **CSWSH Mitigation:** Explicitly rejected opaque `null` origins by default to prevent Cross-Site WebSocket Hijacking via sandboxed browser iframes.
- **Documentation & Tests:** Clarified client origin semantics in `docs/guide/configuration/README.md`, added 3 new unit test blocks in `test_origin_utils.cpp` (39/39 passing), and added 1 new Python integration test (`test_011_desktop_app_origins_accepted`, 11/11 passing) in `server_websocket_auth.py`.

### DIFF
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -441,6 +441,9 @@ The `LEMONADE_ADMIN_API_KEY` environment variable provides elevated access to bo
 
 The `LEMONADE_ALLOWED_ORIGINS` environment variable controls which remote web origins are authorized to connect to the server (specifically for CORS headers on HTTP endpoints and origin validation on WebSocket connections).
 
+> [!NOTE]
+> `LEMONADE_ALLOWED_ORIGINS` specifies the **client application/web page's origin** (where the request originates), **not** the target Lemonade server URL. Non-browser HTTP clients (such as CLI tools, cURL, or server-side SDKs) do not send an `Origin` header and are not restricted by origin validation.
+
 - **Format**: A comma-separated list of complete origins including the scheme and optional port (e.g., `https://app.lemonade.dev,http://localhost:3000`).
   > [!WARNING]
   > Allowing a non-local plain-HTTP origin (e.g., `http://app.example.com`) is vulnerable to on-path modification (man-in-the-middle) and interception. It is highly recommended to use HTTPS (`https://`) for all remote/non-local allowed origins.
@@ -448,7 +451,7 @@ The `LEMONADE_ALLOWED_ORIGINS` environment variable controls which remote web or
 - **Security Implications of `*`**:
   > [!WARNING]
   > Using `LEMONADE_ALLOWED_ORIGINS=*` permits any website running in a user's browser to make requests to your local Lemonade server. In particular, if `LEMONADE_API_KEY` is not configured, this exposes the server to unauthenticated remote access and cross-origin attacks from malicious websites. Use wildcards only for development or in secure, isolated environments.
-- **Local/Loopback Access**: Loopback origins (`localhost`, `127.0.0.1`, `[::1]`, and `tauri.localhost`) are always allowed and do not need to be explicitly listed.
+- **Local/Loopback & Desktop Access**: Loopback addresses (`localhost`, `127.0.0.1`, `[::1]`, `*.localhost`) and custom desktop application schemes (`file://`, `app://.`, `vscode-webview://`, `jan://`, etc.) are permitted for local client connections. Opaque `null` origins (e.g. from sandboxed browser iframes) are rejected unless explicitly listed in `LEMONADE_ALLOWED_ORIGINS` to prevent CSWSH attacks.
 
 ## Remote Server Connection
 
--- a/src/cpp/include/lemon/utils/origin_utils.h
+++ b/src/cpp/include/lemon/utils/origin_utils.h
@@ -68,6 +68,11 @@ inline Origin parse_origin(const std::string& origin_str) {
         return Origin{};
     }
 
+    if (to_lower(str) == "null") {
+        out.host = "null";
+        return out;
+    }
+
     size_t scheme_pos = str.find("://");
     std::string host_and_port = str;
     if (scheme_pos != std::string::npos) {
@@ -78,13 +83,17 @@ inline Origin parse_origin(const std::string& origin_str) {
     size_t slash_pos = host_and_port.find('/');
     if (slash_pos != std::string::npos) {
         std::string path_part = host_and_port.substr(slash_pos + 1);
-        if (!path_part.empty()) {
+        if (!path_part.empty() && (out.scheme == "http" || out.scheme == "https" || out.scheme == "ws" || out.scheme == "wss")) {
             return Origin{};
         }
         host_and_port = host_and_port.substr(0, slash_pos);
     }
 
     if (host_and_port.empty()) {
+        if (!out.scheme.empty()) {
+            out.host = out.scheme;
+            return out;
+        }
         return Origin{};
     }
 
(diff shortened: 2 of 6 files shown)


fl0rianr (changes requested):
Thanks for addressing the WebSocket header-lifetime problem. Persisting the validated credential across the upgrade is directionally correct, but I found two authentication-state gaps that should be fixed before merging.

### `src/cpp/server/websocket_server.cpp:95`

blocking: A successful return from `authenticate_connection()` does not necessarily mean the connection is authenticated. `/spans/stream` deliberately returns true without credentials so that the client can authenticate with a later `auth` message. Setting `pss->authenticated = true` unconditionally therefore marks an unauthenticated spans connection as authenticated, registers the telemetry listener, and bypasses the normal “authenticate first” message gate.

Could `authenticate_connection()` return the handshake decision and authentication state separately? For an API-key-protected spans connection without credentials, the expected result should be `accepted = true` and `authenticated = false`. Please also add a regression test that sends a non-auth message before the auth handshake and verifies that it is rejected.

### `src/cpp/server/websocket_server.cpp:375`

blocking: The new persistence path does not capture credentials when `LEMONADE_API_KEY` is empty. An admin-only configuration with `LEMONADE_ADMIN_API_KEY` set is supported, but this early return happens before reading an optional Authorization, query, or subprotocol token.

On the libwebsockets configuration this PR is intended to fix, `handle_connection()` may no longer be able to recover that token after the handshake. The connection will then be marked authenticated but will lose its admin identity, so the later `authenticated_token == admin_api_key_` check fails.

Please capture an optionally presented token even when regular API authentication is disabled, and add coverage for an admin-only WebSocket client retaining admin span privileges.

### `src/cpp/include/lemon/utils/origin_utils.h:241`

issue: The documented explicit allowlist escape hatch for opaque origins does not work. `parse_origin("null")` deliberately leaves the scheme empty, but this condition only matches allowlist entries with a non-empty scheme. As a result, `LEMONADE_ALLOWED_ORIGINS=null` still rejects `Origin: null`.

Please either special-case literal `null` matching and add HTTP plus WebSocket tests for it, or update the documentation to state that only the wildcard allows opaque origins.

### `src/cpp/include/lemon/utils/origin_utils.h:169`

issue: This makes every non-HTTP(S)/WS(S) scheme an implicitly trusted loopback origin. That includes `chrome-extension://`, `moz-extension://`, and arbitrary custom schemes, and it bypasses `LEMONADE_ALLOWED_ORIGINS` for both HTTP CORS and WebSocket validation.

Could we restrict the implicit set to the desktop schemes Lemonade actually supports, or require unknown schemes to be configured explicitly? A custom scheme alone does not establish that the client is trusted.

### `src/cpp/server/websocket_server.cpp:99`

suggestion: Please also erase this entry on `LWS_CALLBACK_WSI_DESTROY` and clear the map in `stop()`. A connection can pass `FILTER_PROTOCOL_CONNECTION` and then fail before `ESTABLISHED`; relying only on `ESTABLISHED` and WebSocket `CLOSED` may leave a credential associated with a destroyed `wsi`.

abn:
Thanks for the detailed review! I've updated the implementation to address the security and lifecycle findings:

### 1. Separation of Connection Upgrade and Authentication Status
- **Fix:** Updated `authenticate_connection` to return separate results for handshake acceptance and authentication state. 
- `/spans/stream` upgrades successfully without credentials, but is marked unauthenticated (`pss->authenticated = false`). The connection will reject all subscription or telemetry operations with an `Unauthorized` error until it receives a valid in-band `"auth"` JSON frame.
- **Coverage:** Added integration test `test_012_spans_anonymous_connection_gate` verifying that anonymous clients are gated on non-auth frames until they complete the message handshake.

### 2. Admin Token Persistence in Admin-Only Configurations
- **Fix:** Extract and persist the token regardless of whether regular `api_key_` authentication is active. If only `LEMONADE_ADMIN_API_KEY` is set, the presented admin token is correctly saved in the connection state, enabling the WebSocket client to retain admin spans privileges.
- **Coverage:** Added integration test `test_013_admin_only_auth` verifying admin token persistence under admin-only environments.

### 3. Explicit `null` Origin Allowlist Matching
- **Fix:** Trimmed entries in `LEMONADE_ALLOWED_ORIGINS` and added explicit matching for opaque `"null"` origins. Setting `LEMONADE_ALLOWED_ORIGINS=null` now correctly permits sandboxed browser iframes or local HTML files that present `Origin: null`.
- **Coverage:** Added C++ unit tests in `test_origin_utils.cpp` and a Python integration test `test_014_null_origin_explicit_allowlist`.

### 4. Custom Desktop Scheme Trust (Design Choice Rationale)
- **Rationale:** We have preserved the implicit trust for non-HTTP(S)/WS(S) custom desktop schemes (`jan://`, `app://.`, `file://`, `vscode-webview://`, etc.). 
- Under browser security models, a remote website (e.g. `https://attacker.com`) cannot forge or spoof a custom protocol origin; the browser strictly restricts them to their own standard origin.
- Since native applications can already bypass origin validation entirely (by omitting the `Origin` header), trusting custom schemes as loopback does not introduce additional security risks. Instead, it ensures zero-friction, out-of-the-box local desktop and webview compatibility.

### 5. Lifecycles & Memory Safety Cleanup
- **Fix:** Handled cleanup inside the `LWS_CALLBACK_WSI_DESTROY` callback to ensure `wsi` keys are erased from `pending_authenticated_tokens_` even if the connection terminates prematurely before establishment. Added map clearing inside `stop()` under mutex locks.

fl0rianr (commented):
Thanks for the update. The authentication-state separation, admin-token capture, and lifecycle cleanup now look correct. I found one remaining origin-validation bug and one important regression-test gap.

### `src/cpp/include/lemon/utils/origin_utils.h`

blocking: The explicit `null` allowlist match is based only on `request_origin.host == "null"`.

That also matches parsed standard origins such as `http://null`, `https://null`, `ws://null`, and `wss://null`, because those origins have a non-empty scheme but the same host value. Therefore, `LEMONADE_ALLOWED_ORIGINS=null` currently permits more than the documented opaque `Origin: null`.

Please require the request to be the actual opaque-null representation, for example:

```cpp
const bool is_opaque_null =
    request_origin.scheme.empty() &&
    request_origin.host == "null" &&
    request_origin.port == -1;
```

Then match the allowlist entry only when `is_opaque_null` is true.

Please also add negative tests such as:

```cpp
!is_origin_allowed("http://null", "null")
!is_origin_allowed("https://null", "null")
!is_websocket_origin_allowed("http://null", "null")
```

### `test/server_websocket_auth.py:341`

test gap: `test_013_admin_only_auth` currently verifies only that the WebSocket handshake succeeds and that `lemonade-realtime` is selected. It does not verify that the admin token survives into `ConnectionState`, so the previous token-loss implementation would also pass this test.

Please make the test exercise admin behavior after establishment. For example:

1. Connect the admin spans WebSocket, preferably through the main HTTP port.
2. Trigger a span from a public request without an Authorization token.
3. Assert that the admin WebSocket receives the span.

That would fail if the persisted token were lost and would directly cover the regression this test is intended to prevent.

The previous authentication-state and cleanup findings otherwise appear resolved.

abn:
Thanks for the follow-up feedback! I have resolved the remaining origin-validation and test-coverage issues:

### 1. Hardened Opaque `null` Matching
- **Fix:** Restrained the `"null"` matching in `origin_utils.h` to check for a true opaque null origin representation: `request_origin.scheme.empty() && request_origin.host == "null" && request_origin.port == -1`.
- Standard web request origins like `http://null` or `https://null` will no longer match the `"null"` entry in the allowlist.
- **Coverage:** Added negative unit tests checking `!is_origin_allowed("http://null", "null")`, `!is_origin_allowed("https://null", "null")`, and `!is_websocket_origin_allowed("http://null", "null")` inside `test_origin_utils.cpp`.

### 2. Verified Post-Upgrade Admin Privilege Retention
- **Fix:** Expanded the integration test `test_013_admin_only_auth` to connect to the spans stream on the main HTTP port (`/v1/spans/stream`) using `admin_secret` as the subprotocol, trigger a public completions request without an authorization token, and assert that the admin connection successfully receives the telemetry span. 
- This confirms that the token was properly stored in `ConnectionState` post-upgrade, and the client retains admin spans privilege after connection establishment (guest clients would be isolated from this span since they are guest-filtered).

fl0rianr (approved):
Thanks, LGTM.

PR #2795, opened 2026-07-23
Reclaim routing helpers when a router collection's policy changes

SlawomirNowaczyk:
**Summary**

Routing-helper models (classifier/embedding/router-LLM backends a policy keeps resident) previously stayed loaded even after the policy that referenced them was edited or deleted. This wires a single models-changed hook from `ModelManager` to the `Router` so helpers no active policy still references are evicted, while keeping helper loading lazy.

**Changes**

- Each `RoutePolicy` now precomputes `helper_models` — the sorted, de-duplicated union of its classifiers' `referenced_models()` (candidates excluded; they load as Standard residency). Classifiers hoist their backend model into a shared `Classifier::model_name_`.
- `ModelManager::set_models_changed_callback` / `notify_models_changed` fire once, outside all locks, only when a **router collection** is added, edited, or removed (API or on-disk edit via the directory watcher). Ordinary model registrations don't trigger a reconcile.
- `Router::reconcile_routing_helpers` evicts every live routing helper not in the union of `helper_models` across all active policies. Pinned and busy helpers are left untouched; policy-authored names are canonicalized at reconcile time.

**Notes**

- Reconciliation only **evicts**; loading stays lazy on first routing request.
- `notify_models_changed` is best-effort (guarded against same-thread re-entry and swallows callback exceptions) because `delete_model` fires it from a scope-guard destructor, where a propagating exception would call `std::terminate`.
- Adds `test_collect_policy_helper_models` covering classifier-union, semantic, llm-router, and deterministic-only policies; existing residency tests still pass.


fl0rianr (changes requested):
Thanks for working on this — the overall direction looks good, especially keeping helper loading lazy and reconciling outside the `ModelManager` locks.

I think there is still a race that can leave an obsolete routing helper resident indefinitely:

1. Request A obtains/copies the old routing policy.
2. Request B edits or deletes that router collection.
3. `reconcile_routing_helpers()` runs before the old helper has been loaded.
4. Request A continues and lazy-loads the helper through `LoadPurpose::RoutingDependency`.
5. No further reconciliation is triggered, so the helper remains loaded even though no active policy references it anymore.

A similar issue occurs when the helper is already busy: `evict_server()` waits for the eviction timeout and then returns without unloading it. Since no retry is scheduled, a helper busy for longer than the timeout can also remain resident until another unrelated policy change occurs.

Could the router retain the current required-helper set or a policy generation, and validate routing-helper loads against it? Alternatively, stale helpers could be marked for eviction once idle. That would make reconciliation durable rather than a one-time snapshot.

It would also be useful to add a concurrency/residency test covering a policy update while an old request is still evaluating or loading its helper.

SlawomirNowaczyk:
> I think there is still a race that can leave an obsolete routing helper resident indefinitely

Good catch. I've reworked routing-helper reconciliation to close the race you flagged (an obsolete helper staying resident indefinitely), and added test coverage.

## The problem

`reconcile_routing_helpers` was a one-time snapshot: it evicted helpers no active policy referenced, but nothing persisted the authoritative "needed" set. That left two gaps:

1. **Lazy-load-after-snapshot** — reconcile could run while a helper was still mid-load (the backend loads with `load_mutex_` released). The load then completed and the now-unneeded helper leaked.
2. **Busy helper** — if a helper was mid-request, `evict_server` waited out the timeout and returned without unloading, with no retry scheduled, so it never got reclaimed.

## The fix

The Router now **owns** the needed-set as durable state: `std::set<std::string> needed_helper_models_` (guarded by `load_mutex_`).

- **`reconcile_routing_helpers`** canonicalizes the policy-authored names (outside the lock so they match live `get_model_name()`), stores them into `needed_helper_models_` under the lock, then prunes.
- **`prune_stale_routing_helpers_locked()`** collects helpers absent from the needed-set and evicts them, but **skips** helpers that are `!is_backend_alive()`, non-`RoutingHelper`, `is_pinned()`, or `is_busy()`. Skipping busy ones means it never blocks on an eviction timeout — they're reclaimed on a later pass once idle. It runs on **every load path**, so a busy-then-idle helper is durably reclaimed.
- **Load-completion validation** — after a helper finishes loading and re-acquires the lock, it validates itself against `needed_helper_models_`; if a policy change dropped it in the meantime, it unloads immediately. This closes the lazy-load race with no timer.
- **Startup seeding** — `server.cpp` seeds the needed-set from policies already present at construction, so a helper loaded before the first policy change still validates against an authoritative set.

Note the `ModelManager::notify_models_changed` best-effort try/catch stays (it's required for destructor safety in `delete_model`) — it's independent of this fix, and a swallowed reconcile just leaves the set stale until the next successful one.

## Tests

New test_routing_helper_reconcile.cpp (`RoutingHelperReconcileTest`, links `lemonade-server-core`).

It adds a `StubWrappedServer` (minimal `WrappedServer` that never spawns a subprocess — only `load()`/`unload()` are pure virtual) and a `friend struct RoutingHelperTestHook` to inject stubs into `loaded_servers_`, seed `needed_helper_models_`, and drive `prune_stale_routing_helpers_locked()` directly — without the `ModelManager`-backed paths that read `server_models.json` and need the backend registry.

Cases:
- Stale idle helper is evicted
- Helper still referenced by a policy survives
- User-pinned stale helper survives (pin outranks policy churn)
- Standard (non-helper) model is never touched by the helper prune
- **Busy-helper durability:** busy stale helper is skipped, then reclaimed on the next prune once idle
- **Concurrency race:** 3 prune threads + a busy-toggler + a policy-writer, asserting no deadlock/crash, that a still-needed helper stays resident, then is reclaimed once no policy needs it

**Limitation:** the full end-to-end `load_model` interleaving (the load-completion guard after the subprocess load) isn't unit-testable without a real `ModelManager`, the backend registry, and a spawned subprocess. That guard shares the exact predicate the test exercises (residency == `RoutingHelper`, not pinned, absent from `needed_helper_models_`), so the durability logic is covered; the subprocess timing remains an integration concern.

fl0rianr (commented):
Thanks for the follow-up — the durable `needed_helper_models_` state is a good improvement, and the direction now looks much closer to what we need.

I found two remaining issues before this is merge-ready:

1. **The current CI failures come from the `Classifier` constructor change.**
   Making `model_name` mandatory breaks existing derived test classifiers that still use the old 3-argument and 5-argument constructor forms, for example in `test_routing_policy_contract.cpp` and `test_routing_policy_evaluator.cpp`.

   The cleanest fix may be to keep a delegating compatibility constructor that supplies an empty model name for model-less test/deterministic classifiers. Updating every existing subclass would also work, but please run the complete C++ build afterwards because the same compile failure currently fans out into several workflows.

2. **The load-versus-policy-change race is not fully closed yet.**
   `reconcile_routing_helpers()` waits for `!is_loading_` before updating `needed_helper_models_`. During a slow backend load, a policy update therefore cannot publish the new authoritative set. The loading thread may reacquire the lock first, validate against the old set, commit the obsolete helper, and let the old request acquire it before reconciliation runs.

   I think the needed set should be updated even while a load is in progress, while only the unsafe eviction/prune portion is deferred.

There is also still no guaranteed retry for a stale helper that is busy during reconciliation. Pruning on a future model load is opportunistic: if no later load or policy update happens, the helper can remain resident indefinitely. A pending-eviction marker or a prune when the final inference releases the helper would make this durable.

The new tests are useful, but the busy-helper test currently invokes a second prune manually. It would be stronger to test the actual production transition: policy update while loading or busy, followed by request completion, and verify that the helper is reclaimed without another external event.

Overall, the approach is good — these look like focused follow-up fixes rather than a redesign.

SlawomirNowaczyk:
Thanks — all three addressed.

**1. CI failures from the mandatory `model_name`.** I went with updating the affected test classifiers to pass an explicit (bogus) model name rather than adding a delegating compatibility constructor. A model-less classifier is really a test/deterministic artifact, and I'd rather not keep a second constructor form alive that new subclasses could lean on — the mandatory `model_name` is the point. Updated `LabeledClassifier`/`SingleScoreClassifier` in test_routing_policy_contract.cpp and `FixedClassifier`/`ThrowingClassifier` in test_routing_policy_evaluator.cpp; those two suites now build and pass (and no other subclass sites remained).

**2. Load-versus-policy-change race.** Fixed. `reconcile_routing_helpers` now publishes `needed_helper_models_` **before** waiting on the slot, and only the eviction pass is deferred. Split into `apply_routing_helper_reconcile(needed)` which takes `load_mutex_`, assigns the set immediately, then waits for `!is_loading_` (+ exclusive ownership) before pruning. Since a helper's backend loads with `load_mutex_` released, a concurrent load re-acquires the lock at completion and now validates against the fresh set, so it self-discards the obsolete helper instead of committing it. No timer.

**3. Test drives the real transition.** The tests now call the production reconcile entry (`apply_routing_helper_reconcile`, with pre-canonicalized names so `ModelManager` stays out of it) instead of poking `prune` directly. The busy-helper case is now: reconcile-to-empty while busy → helper skipped/resident → clear busy → a second policy reconcile reclaims it. No hand-invoked prune. The concurrency test likewise runs three reconcile threads racing a busy-toggler.

**On the remaining "opportunistic release" point:** agreed it's opportunistic by design, and I'd like to leave it there for now. The problem this PR targets is a long-running server accruing *many* stale helpers across *many* config changes — that's now bounded, since every policy change reconciles the full set. The residual case is narrow (a helper busy at reconcile time, with no later load or policy edit), the consequence is one extra resident helper until the next reconcile, and the workaround is simply reloading the policy. Adding a pending-eviction marker + release-time prune hook is real complexity in the request-completion path for a corner we haven't observed hurting anyone. I'd rather ship this, watch real usage, and add the release-triggered reclaim if it turns out to matter.

fl0rianr (commented):
Thanks for the latest update — the previous constructor breakage is fixed, and publishing `needed_helper_models_` before waiting for the active load is the right change. The main load-versus-policy-update race now looks much better.

I still found a few remaining gaps before I would consider this merge-ready:

1. **The retry load path bypasses the new stale-helper validation.**
   The first successful load checks whether a `RoutingHelper` is still present in `needed_helper_models_` before committing it. However, after the initial load fails and the nuclear retry succeeds, `retry_server` is inserted directly into `loaded_servers_` without the same check.

   This can still leak an obsolete helper if the policy changes while the retry backend is starting. It would be safest to centralize the “may this freshly loaded server be committed?” logic and use it for both the initial load and the retry path.

2. **Replacing a router collection with a normal model does not trigger reconciliation.**
   `register_user_model()` decides whether to notify only from the new recipe. If an existing `collection.router` entry is overwritten with a non-router recipe under the same name, its old policy disappears but `notify_models_changed()` is not called.

   Please capture whether the previous entry was a router collection before overwriting it, and notify when either the old or new recipe is a router collection. A regression test for `collection.router -> normal model` would be useful.

3. **Concurrent notifications can publish snapshots out of order.**
   The callback itself is copied under a mutex, but callback execution is not serialized. Two registry updates can therefore compute snapshots concurrently, with the older callback potentially publishing after the newer one and overwriting the authoritative needed-set with stale state.

   Serializing the complete callback execution, or attaching a monotonic registry generation and ignoring older reconciliations, would avoid this.

4. **Busy helpers are still only reclaimed opportunistically.**
   Skipping busy helpers is correct, but they are currently reconsidered only on a later policy reconciliation or model load. The updated test also performs a second reconciliation manually after the helper becomes idle.

   If no further external event occurs, the obsolete helper can remain resident indefinitely. A pending-eviction flag or an idle/release-triggered prune would make the reclamation guarantee complete.

CI looks much healthier now: the previous classifier constructor compile failure is gone, and C++ unit tests have already passed in multiple jobs. But not sure if this fully works out yet.

Overall, this is close now. The standard load race is fixed; the remaining items are focused consistency and lifecycle fixes.

SlawomirNowaczyk:
All three fixes are implemented. Summary:

**#1 — Retry path helper-commit leak.** Extracted the commit decision into one method `may_commit_loaded_server()` (router.cpp, declared in router.h) and call it on both the initial-load path and the nuclear-retry path before pushing into `loaded_servers_`. A helper the policy dropped mid-load is now discarded on both paths.

**#2 — Router-collection overwrite doesn't reconcile.** Added the shared helper `user_entry_is_router_collection()` (model_manager.cpp) used by both `register_user_model` (captures whether the entry being overwritten was a router collection, then notifies if old **or** new is a router collection) and `unregister_user_model` (replacing its inline duplicate). No duplicated logic.

**#3 — Out-of-order snapshot publish.** Added `notify_execution_mutex_` (model_manager.h) held across the entire `cb()` call in `notify_models_changed`, serializing complete callback execution so two concurrent registry updates can't publish snapshots out of order. The existing `in_notify` thread-local still guards same-thread re-entry, so no self-deadlock.

**#4 — Busy helpers opportunistic reclaim.** Left unchanged; check previous comment, the earlier rationale stands (deliberate tradeoff to keep the request-completion path simple; residual is one helper until the next reconcile, workaround = reload policy).

fl0rianr (commented):
Thanks for the latest update. The previous issues are now addressed well: the retry path performs the same commit validation, replacing a router collection with a normal model triggers reconciliation, concurrent notifications are serialized, and the full CI is green.

I still found two race/lifecycle gaps before I would consider this merge-ready:

1. **The notification serialization can delay publication of the newest needed-set.**
   `notify_execution_mutex_` is held across the complete callback, and that callback synchronously calls `reconcile_routing_helpers()`. Reconciliation publishes its snapshot and then waits for any active load or exclusive session before pruning.

   This allows an older notification to hold the execution mutex while waiting, preventing a newer policy update from immediately publishing its newer needed-set. A helper load may therefore finish and validate against the older snapshot before the newer notification gets a chance to run.

   It would be safer to separate “publish authoritative state” from “perform the blocking prune.” For example, publish a monotonic generation or the newest needed-set immediately, mark pruning as pending, and perform the prune once the load/exclusive section is clear. Another option is to let concurrent snapshots carry generations and have the Router reject stale publications.

2. **A helper that is already obsolete can still cause destructive load side effects.**
   The new validation happens only after the backend has finished loading, immediately before commit. Before that point, `load_model()` may already evict exclusive NPU peers, replace an FLM slot, enforce residency capacity, and start the backend.

   More importantly, if the first load fails, the current path still calls `evict_all_servers()` and starts the nuclear retry before checking whether the policy still needs this helper. The final retry validation prevents the obsolete helper from being committed, but it does not prevent unrelated resident models from being evicted or the expensive retry from running.

   Please add an early needed-set check under `load_mutex_` before any slot/capacity/NPU eviction, and repeat the check after the first load failure before `evict_all_servers()` and the retry. The final commit-time check should remain, because the policy can still change during backend startup.

There is also still no guaranteed reclamation when a stale helper was busy during reconciliation. The current test makes it idle and explicitly performs a second reconciliation. Without another policy change or model load, the helper may remain resident until generic idle eviction, or indefinitely when auto-eviction is disabled. A pending-stale marker followed by pruning when the final request releases the helper would complete the lifecycle guarantee.

Overall, the implementation is close and CI is now in good shape. The remaining work is focused on ensuring that the newest policy state is always authoritative and that an obsolete helper cannot evict unrelated models before being discarded.

SlawomirNowaczyk:
Thanks — all three are valid, fixed in the latest commit:

1. **Out-of-order notify:** the callback-wide execution lock I added earlier was the wrong tool — it could stall a newer policy update behind an older, blocking one and let a mid-load helper validate against a stale set. Replaced it with a monotonic generation stamped in `notify_models_changed()`; the router now ignores any reconcile whose generation is older than the last applied. The newest generation always carries the final registry state, so it converges without serializing the (blocking) callback.

2. **Destructive side-effects before the commit guard:** extracted a shared `routing_helper_no_longer_needed(...)` predicate and check it in `load_model` twice under `load_mutex_` — before any NPU/FLM/capacity eviction, and again after a failed load before the nuclear `evict_all_servers()` retry. A helper the policy already dropped no longer evicts peers or triggers the nuclear path. `may_commit_loaded_server` now delegates to the same predicate.

3. **Busy stale helper:** added a pending-stale marker. `prune` marks a busy not-needed helper instead of skipping it, and `release_inference` triggers `reclaim_stale_helper_if_idle(name)` the moment the last request drains — reclaiming without a follow-up reconcile. It runs on a separate thread (never the helper's own release stack), re-looks-up by name, and re-validates under the lock via the existing EVICTING/`try_commit_eviction` handshake, so a raced new request safely rescues it. Added a test covering release-triggered reclaim.

fl0rianr:
Please solve the merge conflict as well, I'll re-review later, thanks.

SlawomirNowaczyk:
Merge conflict is resolved

fl0rianr (commented):
Thanks for the updates. The generation-based ordering and the revalidation before loading/evicting helpers look good now. However, I found one rebase-related CI issue and several remaining lifecycle races, so I think this still needs changes.

1. **The new reconciliation test is not registered for the C++ CI test set.**

`RoutingHelperReconcileTest` is added with `add_test(...)`, but it is missing:

```cmake
register_cpp_ci_test(RoutingHelperReconcileTest test_routing_helper_reconcile)
```

The packaging workflow builds the registered `cpp-ci-tests` targets and runs `ctest -L cpp-ci`, so the new test is currently outside that path.

> [!Important] 
> Since you missed this (don't worry it's new), what do you think where should we document that every applicable C++ test added with `add_test()` must also call `register_cpp_ci_test()`? `AGENTS.md` and the contribution/testing checklist would be reasonable places. Longer term, could we remove this footgun by introducing one helper that performs both operations and perhaps add a CI consistency check?

2. **There is a lost-wakeup race between `is_busy()` and `mark_pending_stale()`.**

Those operations acquire the server state lock separately. The last request can finish after `is_busy()` returns true but before the callback is installed. `release_inference()` then sees no pending callback, and the callback is installed only after the server is already idle. No later event is guaranteed to reclaim it.

The “check busy, install callback, or reclaim immediately if already idle” operation needs to be atomic under one state lock.

3. **Pending stale callbacks are not handled when maintenance/downsize ends.**

`is_busy()` includes `maintenance_in_progress_`, but pending callbacks are dispatched only by `release_inference()`. If the helper is busy only because of maintenance, `finish_downsize()` can make it idle without triggering the reclaim callback.

All busy-to-idle transitions should use a shared helper that extracts and dispatches a pending reclaim callback.

4. **The detached callback can outlive `Router`.**

The callback captures a raw `this`, and `release_inference()` invokes it on a detached thread. There is no lifetime guard or shutdown join, so it can access the router after destruction.

This should use a router-owned/joined executor, a weak lifetime token, or another mechanism that guarantees the callback cannot outlive its owner.

5. **Reclaim can also be stranded behind an active load/exclusive session.**

`reclaim_stale_helper_if_idle()` returns when loading or an exclusive session is active and relies on a later prune. A later prune is not guaranteed when that session ends, so the pending helper can remain loaded indefinitely.

The tests should additionally cover:

* the final request finishing between the busy check and callback registration;
* a stale helper becoming idle through maintenance completion;
* deferred reclaim firing during a load/exclusive session;
* router shutdown while a deferred reclaim is pending.

SlawomirNowaczyk:
Thanks — all five are addressed in the latest commit.

1. **Missing `register_cpp_ci_test`.** Added `register_cpp_ci_test(RoutingHelperReconcileTest test_routing_helper_reconcile)` right after the `add_test`, so the suite now runs under the `cpp-ci-tests` aggregate. I'll create a separate PR for documenting this and creating a single helper.

2. **Check-then-install TOCTOU.** Replaced the separate `is_busy()` + `mark_pending_stale()` with a single `WrappedServer::mark_pending_stale_if_busy(callback)` that decides and installs under one `state_mutex_` hold: if the helper is still busy it stores the callback and returns `true`; if it went idle in the window it returns `false` and the prune evicts it immediately. No lost wakeup.

3. **Maintenance-completion edge.** Extracted `take_pending_reclaim_if_idle_locked()` (caller holds `state_mutex_`, returns the callback iff idle) and now call it from **both** `release_inference()` and `finish_downsize()`, dispatching after dropping the lock. A helper that goes idle via a downsize completing now fires the deferred reclaim too, not just one that drains its last request.

4. **Detached thread / lifetime.** Removed the detached thread. Reclaims are now posted to a Router-owned `RoutingHelperReclaimExecutor` (single worker). The stored callback holds a `weak_ptr` to it, so a late dispatch is a safe no-op. `~Router` sets a shutdown flag, wakes any waiting reclaim, then `stop()`s/joins the executor **before** tearing down the state those tasks touch — so a task can never outlive the Router.

5. **Deferred reclaim stranded by a load/exclusive session.** `reclaim_stale_helper_if_idle` no longer early-returns when the slot is held. It now waits on `load_cv_` for the residency slot to clear (broken out on shutdown), then evicts — so it's guaranteed to run rather than relying on a later prune that may never come.

New tests in test_routing_helper_reconcile.cpp: the atomic busy/idle decision of `mark_pending_stale_if_busy`; reclaim firing on maintenance completion (`finish_downsize`); a deferred reclaim blocking on an exclusive session then evicting once it ends; and Router shutdown with a reclaim still blocked on the slot (wakes + joins, no hang). All green locally.

fl0rianr (commented):
Thanks for addressing the previous findings, much better now.

I believe one lost-wakeup race still remains in the release-triggered reclaim path.

`take_pending_reclaim_if_idle_locked()` clears both `pending_stale_` and `pending_stale_reclaim_` when it dispatches the callback. Later, `reclaim_stale_helper_if_idle()` does this:

```cpp
server->set_state(ModelState::EVICTING);
if (!server->try_commit_eviction()) {
    return;
}
```

The comment says that a request rescuing the model leaves it “still pending-stale for the next release”, but the pending state was already cleared when the callback was extracted.

A possible sequence is:

1. Request A releases the stale helper.
2. The pending callback is extracted and the pending state is cleared.
3. The reclaim task is queued.
4. Request B acquires the helper before `try_commit_eviction()`.
5. `try_commit_eviction()` fails because the helper is busy.
6. The reclaim function returns without restoring the pending callback.
7. Request B later releases, but there is no pending callback left to dispatch.

The helper can then remain resident until an unrelated future policy reconcile.

Could the reclaim intent either remain set until eviction succeeds, or be atomically re-armed when `try_commit_eviction()` fails? If re-arming reports that the helper has already become idle again, the reclaim should retry immediately to avoid introducing another check/install window.

Please also add a deterministic regression test where a second request starts after the first release has dispatched the reclaim task but before the eviction commit. The test should verify that releasing the second request evicts the helper without another reconcile.

Separately, the missing `register_cpp_ci_test()` call is fixed now. It would still be useful to document that requirement, but a single CMake helper that performs both `add_test()` and `register_cpp_ci_test()` would prevent this class of rebase mistake more reliably than documentation alone, taht was on me doing the base PR...

SlawomirNowaczyk:
Good catch — that race was real. `take_pending_reclaim_if_idle_locked()` consumes the pending intent when the reclaim is dispatched, so a refused `try_commit_eviction()` (rescued by a new request) dropped it.

Fixed by re-arming on a refused commit: `reclaim_stale_helper_if_idle()` is now a loop, and when the commit is refused it calls `mark_pending_stale_if_busy(...)`. If the helper is busy again it re-arms, so the rescuer's release fires the reclaim; if it went idle in the window, that reports `false` and the loop retries the commit immediately — no new check/install gap. The pending intent is therefore never lost: it's either in the running reclaim or re-armed on the server.

Added a deterministic regression (`test_reclaim_rearms_when_rescued_before_commit`): a request holds the helper when the reclaim reaches its commit, the commit is refused, and releasing that request evicts the helper with no further reconcile. Full suite green locally.

On the CMake helper: it's better to keep it separate, done in https://github.com/lemonade-sdk/lemonade/pull/2877

fl0rianr (commented):
Thanks — the re-arm loop now closes the previously reported lost-wakeup race correctly. 
I found two focused follow-ups before approval:

1. **The new regression test does not actually reproduce the consumed-pending-intent case.**

In `test_reclaim_rearms_when_rescued_before_commit`, the helper is acquired, reconciliation arms the pending callback, and then `reclaim_now()` is called directly while that original pending callback is still stored on the helper.

Therefore the old implementation would also likely pass this test: the direct reclaim would fail, but the original pending callback would remain installed and would be dispatched by the later `release_inference()`.

The bug occurred specifically after `take_pending_reclaim_if_idle_locked()` had already extracted and cleared that callback.

Please make the test ensure that the pending intent has been consumed before the refused commit is simulated. One deterministic approach is:

```cpp
helper->acquire_for_inference();
RoutingHelperTestHook::reconcile(router, {});

// Consume the original pending intent without reclaiming.
helper->mark_pending_stale_if_busy([] {});
helper->release_inference();

// Model the second request that arrives before the delayed reclaim commits.
helper->acquire_for_inference();
RoutingHelperTestHook::reclaim_now(router, "rescue.helper");
helper->release_inference();
```

Without the new re-arm logic, the final release should leave the helper resident. With the fix, it should evict without another reconcile.

2. **Startup snapshot and generation ordering is not guaranteed.**

The startup seed currently does:

```cpp
router_->reconcile_routing_helpers(
    active_policy_helper_models(),
    model_manager_->next_notify_generation());
```

C++ does not guarantee which argument is evaluated first. Since the extra-model directory watcher may already be running, this sequence is possible:

1. The startup helper snapshot is computed.
2. The watcher publishes a newer registry snapshot.
3. `next_notify_generation()` is evaluated afterward.
4. The old startup snapshot is published with a generation newer than the watcher notification.

That defeats the generation ordering and can leave the Router with stale authoritative state.

Please reserve the generation before computing the snapshot:

```cpp
const uint64_t generation = model_manager_->next_notify_generation();
const auto needed = active_policy_helper_models();
router_->reconcile_routing_helpers(needed, generation);
```

Apart from these two points, the latest reclaim implementation looks substantially better and merge ready

SlawomirNowaczyk:
Both addressed:

**1. Regression test now genuinely reproduces the consumed-pending-intent case.** The previous version left the original pending callback installed when the refused commit ran, so the pre-fix code would have passed too. The test now consumes the intent *before* the refused commit, exactly modeling the window where the reclaim task was already dispatched but hasn't committed:

```cpp
helper->acquire_for_inference();
RoutingHelperTestHook::reconcile(router, {});          // prune arms the reclaim

// Consume the original pending intent without reclaiming: models the state
// after take_pending_reclaim_if_idle_locked() extracted+cleared the callback
// but before the dispatched reclaim commits.
helper->mark_pending_stale_if_busy([] {});
helper->release_inference();

// Second request arrives before that reclaim commits.
helper->acquire_for_inference();
RoutingHelperTestHook::reclaim_now(router, "rescue.helper");  // commit refused → must re-arm
helper->release_inference();                                  // must evict, no reconcile
```

With the pending intent already consumed, the final release only reclaims the helper if the refused commit re-armed the intent. The old code left it consumed and the helper leaked (test fails); with the re-arm it evicts without another reconcile (test passes).

**2. Startup seed now reserves the generation before computing the snapshot.** Argument evaluation order is unspecified in C++, so the directory watcher could publish a newer snapshot between snapshotting the policies and reserving `next_notify_generation()`, stamping the stale seed with a newer generation and clobbering the watcher's authoritative state. Fixed by ordering explicitly:

```cpp
const uint64_t seed_generation = model_manager_->next_notify_generation();
const std::set<std::string> seed_needed = active_policy_helper_models();
router_->reconcile_routing_helpers(seed_needed, seed_generation);
```

fl0rianr (commented):
Thanks — the two findings from my previous review are now addressed correctly.

The startup seed reserves its generation before computing the snapshot, and the updated regression test now genuinely consumes the original pending intent before exercising the refused-commit/re-arm path. The reclaim loop itself looks sound now.

I still found one remaining production path that can still recreate a stale routing helper after reconciliation:

### An already-loaded Standard model can be promoted to a stale RoutingHelper

`auto_load_model_if_needed()` first calls `ensure_loaded_model_residency()`. For routing classifiers this is called with `LoadPurpose::RoutingDependency`.

When the model is already live, `ensure_loaded_model_residency()` directly calls:

```cpp
transition_server_residency_locked(
    existing, requested_residency_class);
```

It does not validate the transition against the current `needed_helper_models_`.

A possible sequence is:

1. Model H is already loaded as `Standard`, for example through direct user inference.
2. Request A copies an old router policy that references H.
3. Request B edits or deletes that policy.
4. Reconciliation publishes a needed-set without H.
5. H is ignored by the prune because it is still `Standard`.
6. Request A continues evaluating its copied policy.
7. `ensure_loaded_model_residency(H, RoutingDependency)` promotes H to `RoutingHelper`.
8. This bypasses all early-load and commit-time stale-helper checks in `load_model()`.
9. No pending-stale marker exists because H was Standard when reconciliation ran.

H can therefore remain indefinitely as a routing helper no active policy references. It also changes a user-facing Standard model's residency based on an obsolete in-flight policy.

I think the narrow fix is to reject the durable Standard -> RoutingHelper role transition when the current needed-set no longer contains the model. The already-loaded backend can remain Standard and still serve the old in-flight request:

```cpp
if (requested_residency_class == ResidencyClass::RoutingHelper &&
    existing->get_residency_class() != ResidencyClass::RoutingHelper &&
    needed_helper_models_.count(canonical_model_name) == 0) {
    existing->update_access_time();
    return true;
}
```

I would check the needed-set directly here rather than allowing pinning to override it: a pin should keep the model resident, but should not let an obsolete policy change its residency role.

Please add a deterministic test with a Standard stub, reconcile to a needed-set that excludes it, request `RoutingDependency` residency, and verify that it remains Standard.

Apart from this transition path, I did not find another critical issue in the current implementation. The executor lifetime, shutdown behavior, maintenance completion, load/retry validation, generation ordering, and refused-commit re-arm now look good to me.

SlawomirNowaczyk:
Thanks, pushed a fix now (first commit).

Since the code evolved quite a bit, the second commit is cleaning up some leftover inefficiencies...
Applied five simplifications to the routing-helper eviction/reclaim path:

1. **Single guarded commit** — collapsed the two-phase `set_state(EVICTING)` + `try_commit_eviction()` into `try_evict_if_idle()`, which under one `state_mutex_` checks `active_request_count_ == 0 && !maintenance_in_progress_` before going to `UNLOADED`. This removes a latent bug: the old unconditional `set_state(EVICTING)` could clobber an `IN_USE`→`READY` server.

2. **Set-once reclaim notifier** — instead of threading a fresh `std::function` reclaim callback through every `mark_pending_stale_if_busy()` call, the server now registers a notifier once via `set_reclaim_notifier()` when it enters the router. `mark_pending_stale_if_busy()` takes no args; `take_pending_reclaim_if_idle_locked()` returns the stored notifier.

3. **Shared decide-or-defer** — extracted `reclaim_or_defer_helper_locked()`, now used by both `prune_stale_routing_helpers_locked()` and `reclaim_stale_helper_if_idle()`, removing the duplicated pin/needed/busy branching.

4. **Executor dedupes by name** — `RoutingHelperReclaimExecutor` now holds a `std::set<std::string>` keyed by model name instead of a task queue, so repeated posts for the same helper coalesce.

5. **One source of truth for staleness** — added `is_needed_helper_locked()`; `routing_helper_no_longer_needed()` is now its inverse. Dropped `may_commit_loaded_server()` and pointed the two load-commit sites at `routing_helper_no_longer_needed()` directly.

All 14 routing-helper reconcile tests pass, including the deterministic re-arm test and the Standard-promotion guard.

fl0rianr (commented):
Oh, how sad, I wanted to approve but CI did catch an error. I repeated the run.
Both the macOS DMG and Debian package jobs are still failing at the C++ unit-test step, so this looks reproducible rather than flaky.

The Debian log identifies `RoutingHelperReconcileTest` as the failing test. It segfaults shortly after the successful eviction of `helper.a`. Could you please investigate this before merge? Given the changes around reclaiming servers and asynchronous callbacks, it may be worth checking for a lifetime issue or a stale `WrappedServer*`, ideally with ASan or UBSan enabled.

Please rerun both package jobs once the issue is fixed. Thanks.

SlawomirNowaczyk:
Just pushed a fix, hopefully

fl0rianr:
Thanks! But CI atm is a killer... fyi

SlawomirNowaczyk:
Previous only fixed MacOS CI, but Debian was still failing (on test #11 instead of #8, but failing nevertheless).
Let's see if this works --> seems so, all CIs pass now :)

fl0rianr (approved):
Man... what a journey here... Thanks! Finally LGTM!

jeremyfowers:
Update + requeue after https://github.com/lemonade-sdk/lemonade/pull/2950 merges

PR #2796, opened 2026-07-23
Fix llama.cpp model field echoing local .gguf path

SlawomirNowaczyk:
### Summary
llama-server is launched with `-m <path>` and echoes that absolute path in the OpenAI `model` field, so lemond forwarded it verbatim. This normalizes `response["model"]` back to the client-facing model id at the llama.cpp response layer, mirroring what the omni orchestrator already does.

### Changes
- Added `LlamaCppServer::normalize_response_model()` which rewrites `response["model"]` to the requested model id (falling back to the registered model name).
- Applied it in `chat_completion`, `completion`, and `responses`, covering both the direct and router-forwarded non-streaming paths.
- Only rewrites responses that carry a `model` field, so error responses pass through untouched.

### Notes
- Fixes the leak of absolute filesystem paths + usernames in the `model` field (e.g. `C:\Users\<name>\...\Qwen3-0.6B-Q4_0.gguf` → `Qwen3-0.6B-GGUF`), and restores agreement with the `decision.schema.json` contract that `route_to` is also carried by `model`.
- Streaming responses are left out of scope for this change, since fixing that is more invasive.

Partially addresses #2730

### DIFF
--- a/src/cpp/include/lemon/backends/llamacpp/llamacpp_server.h
+++ b/src/cpp/include/lemon/backends/llamacpp/llamacpp_server.h
@@ -47,6 +47,12 @@ public:
 
     // ITokenizerServer implementation
     json tokenize(const json& request) override;
+
+private:
+    // llama-server echoes the local .gguf path it was launched with (`-m <path>`)
+    // in the OpenAI `model` field. Rewrite it to the client-facing model id so
+    // responses don't leak absolute filesystem paths (and usernames).
+    json normalize_response_model(json response, const json& request) const;
 };
 
 namespace llamacpp {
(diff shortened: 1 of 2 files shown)


fl0rianr (approved):
Looks good and the non-streaming fix looks correct to me. I’m fine with approving this PR if you think an additional regression test ist not needed.

One remaining request: since streaming responses are explicitly out of scope and may still expose the local model path, could you please replace `Closes #2730` in the PR description with something like `Partially addresses #2730`? Alternatively, please open a follow-up issue for the streaming path before keeping the automatic closure.

Once that is adjusted, this is good to merge from my side.

PR #2797, opened 2026-07-23
Fix: list router/collection ids with their `user.`/`extra.` prefix in `GET /v1/models`

### LINKED ISSUE #2788: [Router] GET /v1/models lists collections without the user. prefix, while registration/fetch/chat use the prefixed id
Found during the pre-v1 router battle test (main @ 18ddfaef).

**Symptom:** register a router collection as `user.X` via `POST /v1/pull`; `GET /v1/models` (and `?show_all=true`) lists it as `X` — the `user.` prefix is dropped. Meanwhile `GET /v1/models/user.X` returns 200 and chat works with **both** `X` and `user.X`.

**Why it matters:** a client that registers `user.X` and then scans the models list for that exact id concludes the collection doesn't exist. The docs tell users to invoke the collection as `user.My-Router`, so the list surface disagreeing with the documented id is a trap for API consumers (the bundled UI happens to normalize, so it isn't affected).

Applies to `collection.omni` listings too, so the fix should be consistent across collection recipes: either list the prefixed id, or document the bare-name listing + dual-id resolution as intended v1 behavior.

SlawomirNowaczyk:
## Summary
A collection registered as `user.X` (via `POST /v1/pull`) was listed as bare `X` in the model-listing surfaces, while registration, `GET /v1/models/user.X`, and chat all use the prefixed id. A client scanning the list for the exact registered id would wrongly conclude the collection doesn't exist. Closes #2788.

## Change
In `rebuild_public_model_aliases_locked`, when the precedence-winner for a bare name is a **non-built-in collection** (`user.*` / `extra.*`, recipe `collection.omni` or `collection.router`), emit its canonical prefixed id (`user.X` / `extra.X`) as the wire-format id instead of the bare name. The bare-name alias is still registered, so dual-id resolution, `GET /v1/models/user.X`, and chat continue to work.

## Scope (intentionally global)
This canonicalization is applied consistently across **every listing surface** backed by `ModelManager::get_downloaded_models()`:
 - OpenAI `GET /v1/models` (plain and `?show_all=true`)
 - Ollama `GET /api/tags`
 - MCP `lemonade_list_models`

Making it global (rather than OpenAI-only) is deliberate: scoping to `/v1/models` would leave the exact same #2788 inconsistency on the Ollama and MCP surfaces. Bare and prefixed forms both still resolve on invocation and on `GET /api/show` / `GET /v1/models/{id}`.

## Behavior
 - Registered/imported collections list with their `user.*` / `extra.*` prefix on all three surfaces, matching registration/fetch/chat.
 - Non-collection models and built-in collections are unchanged (still listed bare).
 - Chat/show with both `X` and `user.X` still resolve.

 ## Docs
Updated the naming contract in canonical_id.h to document the non-built-in-collection exception across all listing surfaces.

### DIFF
--- a/src/cpp/include/lemon/canonical_id.h
+++ b/src/cpp/include/lemon/canonical_id.h
@@ -14,10 +14,19 @@ namespace lemon {
 //   extra.NAME   — model discovered in --extra-models-dir
 //   builtin.NAME — model compiled into server_models.json
 //
-// `/v1/models` and Ollama `/api/tags` emit the bare name when a model is the
+// Listing surfaces (`/v1/models`, Ollama `/api/tags`, and the MCP
+// `lemonade_list_models` tool) emit the bare name when a model is the
 // precedence-winner for its bare name (precedence: Registered > Imported >
-// Builtin), or the canonical-prefixed ID when shadowed. Bare names are accepted
-// as input anywhere a model name is accepted and resolve to the winner.
+// Builtin), or the canonical-prefixed ID when shadowed.
+//
+// EXCEPTION: a non-built-in collection winner (a `user.*` / `extra.*` model
+// with recipe `collection.omni` or `collection.router`) is emitted under its
+// canonical-prefixed ID on every listing surface, even when it is the
+// precedence-winner, so the listed ID matches the one used for registration,
+// fetch, and chat (issue #2788). Built-in collections stay bare.
+//
+// Bare names are accepted as input anywhere a model name is accepted and
+// resolve to the winner.
 //
 // Display strings like "NAME (registered)" / "NAME (imported)" / "NAME (builtin)"
 // are a GUI concern only. The C++ server emits canonical IDs only; the Tauri
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -4828,7 +4828,18 @@ void ModelManager::rebuild_public_model_aliases_locked() {
 
         const Entry& winner = entries.front();
         public_model_aliases_[bare] = winner.cache_key;
-        canonical_public_names_[winner.cache_key] = bare;
+
+        // Collections registered under a prefixed namespace (user.* / extra.*)
+        // must list with their prefix so the id emitted by /v1/models matches the
+        // id used for registration, fetch, and chat. The bare name still resolves
+        // via the alias above, preserving dual-id (bare + prefixed) invocation.
+        const auto winner_it = models_cache_.find(winner.cache_key);
+        const bool winner_is_prefixed_collection =
+            winner.source != ModelSource::Builtin &&
+            winner_it != models_cache_.end() &&
+            is_model_collection_recipe(winner_it->second.recipe);
+        canonical_public_names_[winner.cache_key] =
+            winner_is_prefixed_collection ? canonical_id(winner.source, bare) : bare;
 
         for (size_t i = 1; i < entries.size(); ++i) {
             const Entry& shadowed = entries[i];
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -4792,7 +4792,10 @@ std::string ModelManager::get_model_filter_reason(const std::string& model_name)
 //       callers, so no identity entries are required here
 //
 //   canonical_public_names_ - cache key → wire-format ID emitted by the API:
-//     - winner cache keys → bare name
+//     - winner cache keys → bare name, EXCEPT a non-built-in collection winner
+//       (user.* / extra.* collection.omni / collection.router) → its canonical-
+//       prefixed ID, so the listed ID matches the one used for registration,
+//       fetch, and chat (the bare name still resolves via public_model_aliases_)
 //     - shadowed cache keys → canonical-prefixed ID (user.X / extra.X / builtin.X)
 //
 // Precedence: Registered > Imported > Builtin.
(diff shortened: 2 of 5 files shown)


fl0rianr (dismissed):
Not relevant here, made an error.

fl0rianr (changes requested):
Thanks for the focused fix — the implementation direction looks correct, and keeping bare-name resolution while emitting the canonical prefixed collection ID seems like the right API behavior.

The current CI failures appear to be directly related to the change rather than an unrelated infrastructure issue:

`test/server_cli2.py::test_055_pull_components_omni_collection` still does this:

```python
public_name = collection_name[5:]
...
m["id"] == public_name
```

That test explicitly expects a newly registered `user.CliColl-*` collection to appear under its bare name. With this PR, `/api/v1/models?show_all=true` correctly returns `user.CliColl-*`, so the lookup returns `None`. This explains the CLI-test failures in the Arch/openSUSE jobs and appears to be the same path exercised by the red C++ workflow.

Could you update the existing regression test as part of this PR? I would suggest checking that:

* both `/v1/models` and `/v1/models?show_all=true` emit the full `user.*` ID;
* the bare name is not emitted as a second list row;
* both the bare and prefixed forms still resolve successfully;
* the behavior covers both `collection.omni` and `collection.router`.

Since the implementation also applies to `extra.*` collections, focused coverage for that case would be useful as well. A naming test should ideally preserve the neighboring contracts too: ordinary user models still list under the bare winner name, built-in collections stay bare, and shadowed entries retain their canonical prefixes.

There is also a stale comment above `canonical_public_names_` saying that every winner cache key maps to a bare wire-format name. It should document the new exception for non-built-in collections.

I do not see a separate logic problem in the actual alias-selection change. Once the expectations and coverage are updated and CI is rerun, this looks appropriately scoped.

SlawomirNowaczyk:
Thanks for the review — both points addressed:

**1. Stale doc comment.** Updated the `canonical_public_names_` block in `rebuild_public_model_aliases_locked()` to document the exception: a non-built-in collection winner (`user.*` / `extra.*`, `collection.omni` or `collection.router`) now emits its canonical-prefixed ID rather than the bare name, so the listed ID matches the one used for registration, fetch, and chat (the bare name still resolves via `public_model_aliases_`).

**2. Tests asserting bare listing.** Updated every affected collection test to assert the prefixed-id listing contract — the collection appears under its `user.` id, is **not** also listed under its bare name, and both ids still resolve on `GET /models/{id}` returning the canonical id:
 - `test_055_pull_components_omni_collection` (CLI) and `test_021j_register_user_collection` (omni)
 - **new** `test_021j1_register_user_router_collection` (mirrors 021j for `collection.router`, so both recipes are covered)
 - `test_021j_register_user_collection_with_system_prompt`, `test_021p_collection_components_canonicalized`, `test_021q_collection_repull_overwrites_components`, `test_021zg_router_collection_export_roundtrip`, `test_021zh_router_collection_repull_overwrite`
 - rejection tests (`021t/021u/021x/021y`) hardened to check the canonical (prefixed) name is absent

fl0rianr (commented):
Thanks for the follow-up. The previously reported issues are addressed well:

* the stale `canonical_public_names_` documentation in `model_manager.cpp` is updated;
* the affected CLI and endpoint tests now expect the canonical `user.*` collection ID;
* the tests verify that the bare ID is not emitted as a duplicate row;
* both the canonical and bare IDs remain resolvable;
* `collection.router` now has dedicated coverage in addition to `collection.omni`.

The core alias-selection change looks correct to me. I found one remaining API-surface issue that should be clarified before approval:

`canonical_public_names_` is also used by `ModelManager::get_downloaded_models()`, and Ollama's `/api/tags` directly serializes the IDs returned by that method. This means the PR changes not only `/v1/models`, but also Ollama collection listings from a bare winner such as `MyCollection` to `user.MyCollection` or `extra.MyCollection`.

That may be the desired globally consistent behavior, but it is currently outside the stated PR scope and has no Ollama regression coverage.

Could you please either:

1. confirm that the `/api/tags` change is intentional, update the PR description and the naming contract in `canonical_id.h`, and add a focused Ollama test proving that `/api/tags` emits the canonical collection ID while `/api/show` and invocation still accept both forms; or

2. keep the existing Ollama listing behavior and scope the prefixed-ID change specifically to the OpenAI `/v1/models` surface?

`canonical_id.h` currently still states that precedence winners are emitted under their bare name by both `/v1/models` and `/api/tags`, so that contract is stale under the current implementation.

As a smaller coverage improvement, it would also be useful for one test to exercise plain `/v1/models`, not only `?show_all=true`, since the PR explicitly promises both surfaces.

Other than this cross-API scope question, I do not see a correctness problem in the implementation. The previously failing Linux distro checks are now green; the remaining large release workflows are still queued.

SlawomirNowaczyk:
Thanks — good catch on the cross-API surface. This is intentional and I've made it explicit rather than scoping down (option 1):

 - **canonical_id.h** contract updated: the non-built-in-collection exception now applies to all listing surfaces (`/v1/models`, `/api/tags`, and MCP `lemonade_list_models`), and it notes built-in collections stay bare.
 - **PR description** updated with a "Scope (intentionally global)" section naming `/api/tags` and MCP — since all three are backed by the same `get_downloaded_models()` mapping, scoping to `/v1/models` only would re-introduce #2788 on the other two surfaces.
 - **Ollama coverage**: new `test_021j2_collection_lists_prefixed_in_ollama_tags` — `/api/tags` emits `user.X:latest` (no bare row), and `/api/show` resolves both bare and prefixed.
 - **MCP coverage**: new `test_013_list_models_emits_prefixed_collection_id` — `lemonade_list_models` lists the collection under its `user.` id, not bare.
 - **Plain `/v1/models`**: `test_021j` now asserts both the plain and `?show_all=true` surfaces.

All new/updated tests pass against a locally-built server.

fl0rianr (approved):
Thanks, ready to go from my side now.

PR #2798, opened 2026-07-23
[Router] Fix DirectoryWatcher missing policy delete+rewrite on Windows

### LINKED ISSUE #2787: [Router] RoutingPolicyStoreTest flaky on Windows: DirectoryWatcher misses policy delete+rewrite (file-watcher hot reload)
Found during the pre-v1 router battle test (main @ 18ddfaef, Windows 11).

**Symptom:** `RoutingPolicyStoreTest` fails ~2 of 3 runs on Windows. The two failing sub-tests are `DirectoryWatcher triggers engine swap` and `watcher-reloaded engine uses updated policy`.

**Detail:** the test deletes and rewrites a policy file, then polls up to 4.5s (30 × 150ms) for the store to swap the engine. On Windows the watcher frequently never fires within that window. `DirectoryWatcherTest` itself passes, so this is likely the remove-then-recreate sequence racing the watcher rather than the watcher being wholly broken.

**Scope:** only file-watcher-driven hot reload of on-disk policy files. The primary hot-reload path (re-`POST /v1/pull` of the same collection name) is verified working end-to-end.

- [ ] Root-cause the missed event on Windows (remove+recreate handling in the watcher)
- [ ] De-flake the test (or lengthen/rework the wait) so CI on Windows runners is stable

Repro: `build\Release\test_routing_policy_store.exe` in a loop; fails most iterations.

SlawomirNowaczyk:
### Summary
The Windows polling `DirectoryWatcher` keyed its change-detection snapshot on `(filename, st_mtime)` where `st_mtime` has 1-second resolution, so a delete-then-rewrite of the same policy file within one wall-clock second produced an identical snapshot and the hot-reload callback never fired. This fixes the missed event and de-flakes `RoutingPolicyStoreTest` on Windows.

Closes #2787

### Changes
- Snapshot key now includes file **size** and a **high-resolution mtime** (`fs::last_write_time` 100 ns ticks instead of `time_t` seconds), so a same-second same-name rewrite is still detected.
- Test now inserts a short wait between the `remove` and the recreate so the two are distinct filesystem events.

### Testing
- `test_routing_policy_store.exe`: 15/15 passing (previously failed ~2 of 3).
- `test_directory_watcher.exe`: 10/10 passing (no regression).

Note: only the Windows polling backend was affected; the Linux (inotify) and macOS (kqueue) backends receive real delete/create events and were already correct.

### DIFF
--- a/src/cpp/server/directory_watcher.cpp
+++ b/src/cpp/server/directory_watcher.cpp
@@ -7,6 +7,7 @@
 #include <functional>
 #include <cstring>
 #include <set>
+#include <tuple>
 #include <filesystem>
 
 namespace fs = std::filesystem;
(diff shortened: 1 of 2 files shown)


eddierichter-amd (approved):
LGTM

PR #2802, opened 2026-07-24
Add JSON schema checking to pre-commit

superm1:
Helps save some CI time for dumb mistakes.

For example now catches this mistake that I made on a PR.

```
Validate files with jsonschema...........................................Failed
- hook id: check-jsonschema
- exit code: 1

Several files failed to parse.
  Failed to parse src/cpp/resources/backend_versions.json
    JSONDecodeError: Expecting property name enclosed in double quotes: line 14 column 9 (char 1361)
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,11 @@ repos:
     hooks:
       - id: black
         language_version: python3
+  - repo: https://github.com/python-jsonschema/check-jsonschema
+    rev: 0.37.4
+    hooks:
+      - id: check-jsonschema
+        files: .*\.json$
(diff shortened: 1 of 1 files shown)


kenvandine (approved):
Nice!

ianbmacdonald:
Heads up: `schemas/foo.json` isn't in the repo, so this hook fails on every `.json` file:

```
Error: Unexpected Error building schema validator
FileNotFoundError: [Errno 2] No such file or directory: '.../schemas/foo.json'
```

`check-github-actions` and `check-github-workflows` are unaffected (they use builtin schemas), so
either pointing that first hook at a real schema or dropping it is enough.

PR #2803, opened 2026-07-24
Fix/rocm therock version resolution

### LINKED ISSUE #2800: Cannot install ROCm Llama.cpp GPU backend / Pulling wrong version from github
### Platform

Windows

### Lemonade Version

11.5.0

### GPU / APU Model

RX 9070 XT

### Component

llama.cpp

### Bug Description

The ROCm backend fails to install due to a 404 error pointing to the wrong file. Both the CLI and WebUI produce this error.

> Error: Failed to download llama-server.exe from: https://github.com/lemonade-sdk/llama.cpp/releases/download/b10107/llama-b10107-bin-win-rocm-7.13-x64.zip - Download failed after 6 attempts. Last error: HTTP error 404 for URL: https://github.com/lemonade-sdk/llama.cpp/releases/download/b10107/llama-b10107-bin-win-rocm-7.13-x64.zip

The command I ran for the CLI was `lemonade backends install llamacpp:rocm

Image

Image

### Steps to Reproduce

Attempt to install ROCm GPU backend via the web interface or the command line.

I expect not to get an error and for the backend to be downloaded and installed on my device.

### Log Output

```shell
2026-07-23 17:19:36.330 [Debug] (Server) POST /api/v1/install
2026-07-23 17:19:36.330 [Info] (Server) Installing backend: llamacpp:rocm
2026-07-23 17:19:36.332 [Debug] (Server) POST /api/v1/install - 200
2026-07-23 17:19:36.333 [Debug] (BackendManager) Installing llamacpp:rocm-stable
2026-07-23 17:19:36.337 [Info] (llamacpp Server) Installing llama-server.exe (version: b10107)
2026-07-23 17:19:36.338 [Debug] (llamacpp Server) Downloading to: C:\Users\MRX\AppData\Local\Temp\llamacpp_rocm-stable_b10107_llama-b10107-bin-win-rocm-7.13-x64.zip
2026-07-23 17:19:36.338 [Debug] (llamacpp Server) Downloading from: https://github.com/lemonade-sdk/llama.cpp/releases/download/b10107/llama-b10107-bin-win-rocm-7.13-x64.zip
```

### Additional Context

It looks like it's attempting to pull version 7.13 when 7.14 is the one available.

Bekhouche:
## Summary

`llamacpp.rocm_bin="latest"` (or any explicit tag other than the pinned one) could resolve to a llama.cpp release built against a newer ROCm/TheRock version than the statically pinned `therock.version`, causing the download filename to embed the wrong ROCm version and 404. This bumps the immediate stale pin, then makes the version resolution self-verifying against the actual GitHub release instead of trusting a static value that can silently drift.

Fixes #2800

## Changes

- **Quick fix**: bump `llamacpp.rocm-stable` and `therock.version` to a verified-matching pair (`b10108`/`7.14`).
- **Dynamic discovery**: when the resolved llama.cpp tag doesn't match the static pin (or even just to self-verify), look up the real ROCm version from the resolved release's own GitHub asset name instead of trusting the static config. Falls back gracefully offline; throws a clear error if verification is genuinely needed but fails online, instead of silently reconstructing the original 404.
- **TheRock runtime package**: the discovered version now also determines which TheRock runtime gets installed and which one gets picked up at launch (previously only the download filename was fixed; the runtime library path could still resolve to a stale, mismatched version already on disk).
- **Persistence**: the resolved version is written to a marker file alongside the install, so a later launch, in a different process, possibly offline, can recover it without redoing live discovery (which may not even be possible offline).
- **Incomplete-install hardening**: `get_therock_lib_path()` now requires `version.txt` to exist and match before trusting a directory; failed TheRock downloads no longer leave a directory behind that could be mistaken for a real install.
- Scoped to `recipe=="llamacpp"` only. `sd-cpp` has the same underlying bug, but its releases can ship multiple ROCm-versioned assets per tag, so naively picking the first match risks silently choosing the wrong one. Left as a documented follow-up requiring arch-aware disambiguation.

## Test plan

- [x] Verified live against real GitHub release data (confirmed the reported 404 reproduces exactly, and that the fix resolves to a real, downloadable asset)
- [x] Built and tested on WSL/Linux and natively on Windows (VS 2026/MSVC)
- [x] Reproduced the original bug pattern (explicit tag diverging from the static pin) and confirmed self-correction
- [x] Verified `rocm_bin="latest"`, default/`builtin`, offline-graceful-fallback, and online-hard-failure paths
- [x] Verified the persisted marker survives a simulated restart/offline scenario, on both platforms
- [ ] No automated regression tests added yet (flagged as a known gap)


jeremyfowers:

### Scope

**minor** — The PR fixes an existing bug in ROCm backend installation without introducing new commands, endpoints, config surfaces, or supported platforms — it's a contained fix scoped to llamacpp:rocm-stable.

### Suggested reviewers

- superm1 (AMD ROCm) — Listed AMD ROCm maintainer; the PR's entire scope is ROCm/TheRock version resolution.
- Geramy (AMD ROCm) — Listed AMD ROCm maintainer with ROCm backend expertise.
- danielholanda (AMD ROCm) — Listed AMD ROCm maintainer; the fix touches ROCm installation and runtime path resolution.

_Reviewed at head `b82989f` by repo-manager pr-review. Regenerate with `repo-manager review-pr 2803`._

superm1:
The ABI is supposed to be stable at ROCm 7.14, I don't know that we'll need this after https://github.com/lemonade-sdk/lemonade/pull/2768 merges

Bekhouche:
> The ABI is supposed to be stable at ROCm 7.14, I don't know that we'll need this after #2768 merges

Good point. This PR is really a safety net for version drift, not a fix for a currently known bug. If #2768 locks ROCm at 7.14 as the stable ABI going forward, the auto-detection here mostly stops being needed. I'm okay holding this until #2768 merges, then rechecking. Happy to close it if it turns out to be redundant.

PR #2805, opened 2026-07-24
fix(ci): resolve test flakes across C++ store, python test harness, and workflows

abn:
Resolves recurring CI test flakes across Windows directory watcher tests, Python test setup/teardown races, self-hosted runner Python setup hangs, omni tool resume sampling non-determinism, and unauthenticated release workflow API rate limits.

## Changes
- **C++ Routing Policy Store Test**: De-flake `RoutingPolicyStoreTest` by adding settle delay between policy file deletion/creation and extending watcher polling loop for Windows NTFS event delivery.
- **Python Test Harness**: Add bounded retries to model unloading during test setup in `server_base.py` to prevent state contamination between consecutive test modules.
- **Streaming Errors Suite**: Catch transient `ConnectionError` alongside `ChunkedEncodingError` in `server_streaming_errors.py` when stream sockets close mid-chunk.
- **Omni App Tool Resume Suite**: Enforce deterministic `temperature=0.0` sampling and add bounded retries to `test_006_app_tool_resume` and `test_007_app_tool_resume_streaming` in `test/server_omni.py` to prevent transient empty generation streams during tool resume.
- **Self-Hosted Setup Python Action**: Add `DEBIAN_FRONTEND=noninteractive` and HTTP timeouts to standard system `python3` / `python3-venv` package installations in `.github/actions/setup-python/action.yml` to prevent silent 4-minute CI hangs on missing Python versions. (Removed the deadsnakes PPA setup).
- **Release Validation Workflows**: Pass `GITHUB_TOKEN` to `validate_sdcpp.yml` top-level environment to eliminate GitHub API 60 req/hr rate limits during release tag discovery.

### DIFF
--- a/.github/actions/setup-python/action.yml
+++ b/.github/actions/setup-python/action.yml
@@ -126,25 +126,7 @@ runs:
         done
 
         if [ -z "$PYTHON_CMD" ]; then
-          # Try to install from deadsnakes PPA (Ubuntu)
-          echo "Python $PYTHON_VERSION not found. Attempting to install..."
-
-          if command -v apt-get &> /dev/null; then
-            sudo add-apt-repository -y ppa:deadsnakes/ppa 2>/dev/null || true
-            sudo apt-get update
-            sudo apt-get install -y "python${PYTHON_VERSION}" "python${PYTHON_VERSION}-venv" || {
-              echo "Failed to install Python $PYTHON_VERSION from deadsnakes"
-              echo "Falling back to system python3..."
-              PYTHON_CMD=$(command -v python3)
-            }
-            PYTHON_CMD=$(command -v "python${PYTHON_VERSION}") || PYTHON_CMD=$(command -v python3)
-          else
-            PYTHON_CMD=$(command -v python3)
-          fi
-        fi
-
-        if [ -z "$PYTHON_CMD" ]; then
-          echo "ERROR: Could not find or install Python $PYTHON_VERSION"
+          echo "ERROR: Could not find Python $PYTHON_VERSION"
           exit 1
         fi
 
--- a/.github/workflows/validate_sdcpp.yml
+++ b/.github/workflows/validate_sdcpp.yml
@@ -26,6 +26,7 @@ env:
   LEMONADE_CI_MODE: "True"
   PYTHONIOENCODING: utf-8
   HF_TOKEN: ${{ secrets.HUGGINGFACE_ACCESS_TOKEN }}
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   SDCPP_REPO: leejet/stable-diffusion.cpp
   SDCPP_CUDA_REPO: lemonade-sdk/stable-diffusion.cpp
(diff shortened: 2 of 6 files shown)


.github/actions/setup-python/action.yml
```diff
@@ -130,9 +130,10 @@ runs:
           echo "Python $PYTHON_VERSION not found. Attempting to install..."
 
           if command -v apt-get &> /dev/null; then
-            sudo add-apt-repository -y ppa:deadsnakes/ppa 2>/dev/null || true
-            sudo apt-get update
-            sudo apt-get install -y "python${PYTHON_VERSION}" "python${PYTHON_VERSION}-venv" || {
+            export DEBIAN_FRONTEND=noninteractive
+            sudo -E add-apt-repository -y ppa:deadsnakes/ppa 2>/dev/null || true
```
superm1: Why do we even need some random Internet PPA in the first place. Python is in the Ubuntu archives.

I think all this should go.

.github/workflows/linux_distro_builds.yml
superm1, line 1: Have there actually been failures here? I figured most flakiness is NOT this install step. The arch archives work or they don't. Retry probably won't change much.

superm1 (commented):
I'm generally aligned with increased timeout and retries but left some specific comments.

.github/actions/setup-python/action.yml
```diff
@@ -126,18 +126,16 @@ runs:
...
+            export DEBIAN_FRONTEND=noninteractive
+            sudo -E apt-get update -o Acquire::Http::Timeout="10" -o Acquire::Retries="2" || true
+            sudo -E apt-get install -y -o Acquire::Http::Timeout="10" python3 python3-venv || {
+              echo "Failed to install system python3"
             }
-            PYTHON_CMD=$(command -v "python${PYTHON_VERSION}") || PYTHON_CMD=$(command -v python3)
+            PYTHON_CMD=$(command -v python3)
           else
             PYTHON_CMD=$(command -v python3)
           fi
```
superm1: this all seems unnecessary.

```
╰─❯ apt depends ubuntu-minimal | grep python3

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

  Depends: python3
```

superm1 (commented):
I rebased your branch to get fresh CI runs.

1. Can you please drop all that code to setup python?  It seems like slop we don't need at all given `python3` is part of `ubuntu-minimal`.

2. Also; please fix the findings from pre-commit.

Then LGTM.

abn:
Thanks for the rebase and review @superm1!

I've addressed your feedback and rebased on top of the latest `main`:

- **Python setup action**: Dropped the deadsnakes PPA and fallback `apt-get` install logic on Linux since `python3` is already present on the runners.
- **Pre-commit**: Ran Black to resolve the formatting findings across the modified test files.
- **Test harness**: Hardened `unload_all_models` with clean response initialization across retries.

All pre-commit hooks are passing locally. Let me know if anything else needs a pass!

superm1 (approved):

PR #2807, opened 2026-07-25
fix(llamacpp): sanitize oversized tool schema limits

### LINKED ISSUE #2691: llama.cpp backend: tool schemas with string maxLength >= ~2000 break all tool calling (lemonade launch claude unusable out of the box)
## Summary

Any tool whose JSON schema declares a string `maxLength` at or above ~2000 makes **every tool-calling request to a llama.cpp backend fail with HTTP 400**:

```
parse: error parsing grammar: number of repetitions exceeds sane defaults, please reduce the number of repetitions
E srv send_error: Failed to initialize samplers: failed to parse grammar
```

llama.cpp compiles each tool's schema into a GBNF grammar to constrain tool-call decoding. A string `maxLength: N` is emitted as `char{0,N}`. Since ggml-org/llama.cpp#17381 added a repetition sanity cap (a DoS guard), llama.cpp now **generates a grammar that its own parser then rejects**. See also ggml-org/llama.cpp#17473 — closed, but the schema→GBNF converter path still trips on the currently pinned builds.

The failure is model-independent: it's purely a function of the tool schema.

## Impact: `lemonade launch claude` is unusable out of the box

Claude Code's built-in `Workflow` tool declares its `script` parameter with `maxLength: 524288`, which the converter expands to:

```
tool-Workflow-schema-script ::= "\"" char{0,524288} "\""
```

So every tool call 400s and the agent can't do anything. The only workaround is `--agent-args "--tools <explicit list omitting Workflow>"`, which is undiscoverable — nothing in the error points at a tool schema, and [the Claude Code integration docs](https://lemonade-server.ai/docs/integrations/claude-code/) don't mention it.

Note `--disallowedTools` does **not** work: it only denies *permission*, so the schema is still sent and still breaks the grammar. `--tools` (which selects from the built-in set) is required.

This isn't Claude-specific — llama.cpp#17473 was originally reported via OpenWebUI native tool calls, and this repo has adjacent tool-parse reports (#1562, #1011).

## Reproduction

Any llama.cpp model. The **only** difference between the two calls is the presence of `maxLength`.

```sh
# 400 — Failed to initialize samplers: failed to parse grammar
curl -s localhost:8000/api/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "<any-llamacpp-model>",
  "messages": [{"role":"user","content":"hi"}],
  "tools": [{"type":"function","function":{"name":"T","description":"t","parameters":{
    "type":"object","properties":{"s":{"type":"string","maxLength":524288}},"required":["s"]}}}]
}'

# 200 — identical, maxLength removed
curl -s localhost:8000/api/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "<any-llamacpp-model>",
  "messages": [{"role":"user","content":"hi"}],
  "tools": [{"type":"function","function":{"name":"T","description":"t","parameters":{
    "type":"object","properties":{"s":{"type":"string"}},"required":["s"]}}}]
}'
```

Measured threshold — gpt-oss-120b, llamacpp `vulkan` b9747, Ryzen AI MAX (Strix Halo, gfx1151), Lemonade 10.9.0:

| string `maxLength` | result |
|---|---|
| 1000 | 200 OK |
| **2000** | **400 — failed to parse grammar** |
| 4096 / 65536 / 524288 | 400 |

End-to-end: `lemonade launch claude --agent-args "--tools Bash,Read,Write,Edit,Glob,Grep,WebFetch,Skill,Agent,TaskCreate"` works; adding `Workflow` to that same list reproduces the 400.

## Proposed fix

Sanitize tool schemas on the llama.cpp forwarding path, as a request transform alongside the existing `JsonUtils::with_*` helpers:

```cpp
// src/cpp/include/lemon/utils/json_utils.h
static json with_grammar_safe_tool_schemas(const json& request);
```

It would walk `tools[].function.parameters` (recursively through `properties` / `items` / `$defs`) and **remove** any `maxLength` / `maxItems` at or above a conservative threshold (1000 measured safe).

Apply it in:

- **`LlamaCppServer::chat_completion`** — compose with the existing transform, exactly as `vllm_server.cpp` already composes two:
  ```cpp
  return forward_request("/v1/chat/completions",
                         JsonUtils::with_grammar_safe_tool_schemas(
                             JsonUtils::with_legacy_max_tokens_alias(request)));
  ```
- **`LlamaCppServer::responses`** — same, if it accepts `tools`.
- **`Router::chat_completion_stream`** — ⚠️ this is the one that's easy to miss. It forwards the **raw string** body, so it bypasses the transform above entirely. Claude Code streams, so patching only `chat_completion()` would pass a curl test and still leave the agent broken. It already parses the body into `request_json`, and `identity.recipe` is in scope at the `forward_streaming_request` call site, so the sanitized JSON can be re-dumped for llama.cpp-recipe backends.

### Why remove the bound rather than clamp it

Clamping `maxLength` down to the cap would be a correctness bug: it would forbid the model from ever emitting a legitimately long argument (a Claude Code `Workflow.script` really can be hundreds of KB). Dropping the constraint yields `char*` — unconstrained — which is the correct semantics. `maxLength` is consumed here only to build the decoding grammar; Lemonade isn't acting as a schema validator, and the field is still forwarded to the model in the tool description.

### Scope

llama.cpp backends only — cloud / vLLM / FLM are unaffected and shouldn't be touched.

### Precedent

This matches the existing llama.cpp-workaround pattern in `llamacpp_server.cpp` (the `HOME`-unset segfault guard, ~L535), which carries a "delete once the upstream fix lands in `backend_versions.json`" comment linking the upstream PR. The same comment style would apply here, referencing ggml-org/llama.cpp#17473 / #17381, so the shim can be removed once the pinned llama.cpp clamps the bound in its converter.

## Alternative considered and rejected

Hardcoding `--tools <list>` into the `lemonade launch claude` preset. It hardcodes another project's tool names (which drift), silently strips capability from the user, only fixes Claude Code while OpenWebUI/AnythingLLM hit the same class of failure, and does nothing for direct API consumers. The schema sanitizer fixes all of them in one place.

I'm happy to put up a PR if the approach and threshold look right — I didn't want to guess at your `$defs`/recursion conventions.

## Environment

- Lemonade 10.9.0, server port 13305
- Backends: llamacpp `vulkan` b9747 and `rocm` b9752 (both affected — it's schema-driven, not backend-driven)
- Model: `gpt-oss-120b-GGUF-UD-Q6_K_XL` (reproduces on any llama.cpp model)
- AMD Ryzen AI MAX / Strix Halo (gfx1151), 128 GB unified, Ubuntu 26.04, kernel 7.0
- Claude Code (current), launched via `lemonade launch claude`

sujikathir:
## Summary

- Sanitize oversized `maxLength` and `maxItems` constraints in llama.cpp tool parameter schemas before forwarding requests.
- Apply the transform recursively to Chat Completions and Responses API schemas, including streaming requests.
- Remove unsafe bounds rather than clamping them, preserving the ability to generate legitimately long tool arguments.
- Keep the workaround scoped to the llama.cpp backend.

Fixes #2691.

## Validation

- `cmake --build --preset default --target test_llamacpp_request`
- `ctest --test-dir build -R '^llamacpp_request$' --output-on-failure` — 1/1 passed
- `git diff --check origin/main...HEAD`
- Live test with llama.cpp b9747 and a checksum-verified Qwen3 tool-calling GGUF:
  - Raw llama.cpp, `maxLength: 524288` — HTTP 400, grammar initialization failure
  - Patched Lemonade, non-streaming — HTTP 200
  - Patched Lemonade, streaming — HTTP 200 with a terminating `[DONE]` event

### Before

Raw llama.cpp rejects the oversized tool schema

### After

Patched Lemonade accepts non-streaming and streaming requests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -826,6 +826,16 @@ if(BUILD_TESTING AND EXISTS "${VLLM_ARG_RESOLVER_TEST_SOURCE}")
     add_test(NAME vllm_arg_resolver COMMAND test_vllm_arg_resolver)
 endif()
 
+set(LLAMACPP_REQUEST_TEST_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_llamacpp_request.cpp")
+if(BUILD_TESTING AND EXISTS "${LLAMACPP_REQUEST_TEST_SOURCE}")
+    add_executable(test_llamacpp_request
+        ${LLAMACPP_REQUEST_TEST_SOURCE}
+        src/cpp/server/backends/llamacpp/llamacpp_request.cpp
+    )
+    target_link_libraries(test_llamacpp_request PRIVATE nlohmann_json::nlohmann_json)
+    add_test(NAME llamacpp_request COMMAND test_llamacpp_request)
+endif()
+
 set(MCP_CLIENT_CONFIG_TEST_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_mcp_client_config.cpp")
 if(BUILD_TESTING AND EXISTS "${MCP_CLIENT_CONFIG_TEST_SOURCE}")
     find_package(Python3 COMPONENTS Interpreter QUIET)
--- a/src/cpp/include/lemon/backends/llamacpp/llamacpp_request.h
+++ b/src/cpp/include/lemon/backends/llamacpp/llamacpp_request.h
@@ -1,0 +1,13 @@
+#pragma once
+
+#include <nlohmann/json.hpp>
+
+namespace lemon {
+namespace backends {
+namespace llamacpp {
+
+nlohmann::json sanitize_tool_schema_limits(nlohmann::json request);
+
+}  // namespace llamacpp
+}  // namespace backends
+}  // namespace lemon
--- a/src/cpp/include/lemon/backends/llamacpp/llamacpp_server.h
+++ b/src/cpp/include/lemon/backends/llamacpp/llamacpp_server.h
@@ -35,6 +35,13 @@ public:
     json completion(const json& request) override;
     json responses(const json& request) override;
 
+    void forward_streaming_request(const std::string& endpoint,
+                                   const std::string& request_body,
+                                   httplib::DataSink& sink,
+                                   bool sse = true,
+                                   long timeout_seconds = 0,
+                                   TelemetryCallback telemetry_callback = nullptr) override;
+
     // IEmbeddingsServer implementation
     json embeddings(const json& request) override;
 
--- a/src/cpp/server/backends/llamacpp/llamacpp_request.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_request.cpp
@@ -10,12 +10,12 @@ namespace {
 
 constexpr std::uint64_t GRAMMAR_REPETITION_LIMIT = 2000;
 
-bool exceeds_grammar_repetition_limit(const nlohmann::json& value) {
+bool has_integer_value_at_least(const nlohmann::json& value, std::uint64_t threshold) {
     if (value.is_number_unsigned()) {
-        return value.get<std::uint64_t>() >= GRAMMAR_REPETITION_LIMIT;
+        return value.get<std::uint64_t>() >= threshold;
     }
     if (value.is_number_integer()) {
-        return value.get<std::int64_t>() >= static_cast<std::int64_t>(GRAMMAR_REPETITION_LIMIT);
+        return value.get<std::int64_t>() >= static_cast<std::int64_t>(threshold);
     }
     return false;
 }
@@ -32,9 +32,18 @@ void sanitize_schema(nlohmann::json& schema) {
     // llama.cpp rejects grammar repetitions at this threshold. Remove oversized
     // bounds instead of clamping them, which would reject valid tool arguments.
     // Remove this workaround once ggml-org/llama.cpp#17473 is fixed in pinned builds.
-    for (const std::string key : {"minLength", "maxLength", "minItems", "maxItems"}) {
+    for (const std::string key : {"minLength", "maxLength"}) {
         auto limit = schema.find(key);
-        if (limit != schema.end() && exceeds_grammar_repetition_limit(*limit)) {
+        if (limit != schema.end() && has_integer_value_at_least(*limit, GRAMMAR_REPETITION_LIMIT)) {
+            schema.erase(limit);
+        }
+    }
+
+    // Array grammars emit the first item separately, so an item bound of N
+    // produces a repetition bound of N - 1.
+    for (const std::string key : {"minItems", "maxItems"}) {
+        auto limit = schema.find(key);
+        if (limit != schema.end() && has_integer_value_at_least(*limit, GRAMMAR_REPETITION_LIMIT + 1)) {
             schema.erase(limit);
         }
     }
--- a/src/cpp/server/backends/llamacpp/llamacpp_request.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_request.cpp
@@ -30,17 +30,52 @@ void sanitize_schema(nlohmann::json& schema) {
     if (!schema.is_object()) return;
 
     // llama.cpp rejects grammar repetitions at this threshold. Remove oversized
-    // upper bounds instead of clamping them, which would reject valid tool arguments.
+    // bounds instead of clamping them, which would reject valid tool arguments.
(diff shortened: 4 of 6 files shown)


fl0rianr (changes requested):
Thanks for the fix. The overall approach looks appropriate, but I found two correctness gaps that are not covered by the existing review comments.

First, `sanitize_schema()` recursively visits every object and array value. This means it also traverses instance values stored under JSON Schema keywords such as `const`, `enum`, `default`, and `examples`.

For example:

```json
{
  "type": "object",
  "properties": {
    "config": {
      "const": {
        "maxLength": 524288
      }
    }
  }
}
```

would be changed so that the constant value becomes `{}`. This alters the schema semantics rather than only removing unsafe grammar constraints.

Please restrict recursion to keywords that actually contain subschemas, such as `properties`, `$defs`, `items`, `allOf`, `anyOf`, `oneOf`, `additionalProperties`, `if`, `then`, and `else`. It would also be good to add regression tests ensuring values inside `const` and `enum` remain unchanged.

Second, the llama.cpp grammar repetition limit also applies to minimum bounds. A schema containing `minLength: 2000` can still generate a repetition that llama.cpp rejects with the same grammar initialization error. The current sanitizer only handles `maxLength` and `maxItems`.

Please verify and cover `minLength` and `minItems` as well, with boundary tests around the repetition limit.

These two issues should be addressed before merge.

sujikathir:
Addressed the requested changes in `b5769dba`: restricted traversal to actual subschema keywords, preserved `const` and `enum` values, added `minLength` and `minItems` boundary coverage, and resolved both inline review comments. The focused CTest passes (1/1), and `lemond` builds successfully on macOS. The refreshed workflows are awaiting maintainer approval. Ready for re-review when convenient—thanks!

fl0rianr (commented):
Thanks for addressing the previous review points. The traversal is now correctly limited to subschema keywords, `const` and `enum` values are preserved, and the streaming alias handling is consistent.

I found one remaining boundary issue in the new `minItems` / `maxItems` handling.

The sanitizer currently removes all four limits at `>= 2000`, but llama.cpp does not translate array bounds the same way as string bounds. For arrays, the first item is emitted separately and the repetition uses `minItems - 1` and `maxItems - 1`.

That means some array limits equal to `2000` are still safe. For example, with positive `minItems`, a schema such as:

```json
{
  "type": "array",
  "minItems": 1999,
  "maxItems": 2000,
  "items": {}
}
```

produces a repetition bounded at `1999`, which is still below llama.cpp's rejection threshold. The current sanitizer nevertheless removes `maxItems: 2000`, unnecessarily weakening the schema to an unbounded array.

Similarly, `minItems: 2000` without a maximum produces a repetition of `1999` additional items and should not be removed solely because the original value equals `2000`.

The new `queue` test also uses `minItems: 2000` together with `maxItems: 1999`, which is already an invalid schema and therefore does not provide a useful boundary test.

Please calculate array safety based on the repetition llama.cpp actually generates, rather than applying the string threshold directly to `minItems` and `maxItems`. The tests should use separate valid schemas around the exact safe and unsafe array boundaries.

fl0rianr:
Thanks for fixing the array boundary handling. The `2000` / `2001` distinction now matches the repetition llama.cpp actually generates, and the replacement tests use valid schemas.

I found one remaining edge case in `has_integer_value_at_least()`.

The helper currently only accepts values stored by nlohmann/json as `number_unsigned` or `number_integer`. Valid JSON Schema integer values can also be written with a zero fractional part, for example:

```json
{
  "type": "string",
  "maxLength": 2000.0
}
```

nlohmann/json stores this as `number_float`, so the sanitizer leaves it unchanged. However, llama.cpp later reads `minLength`, `maxLength`, and `minItems` using `get<int>()`, which converts such numeric values to integers. As a result, `maxLength: 2000.0` can still reach the same rejected grammar repetition.

The same applies to array bounds such as `minItems: 2001.0`.

Please handle finite, non-negative `number_float` values whose fractional part is zero, using the same thresholds as their integer equivalents. Regression tests for at least `maxLength: 2000.0` and `minItems: 2001.0` would cover this path.

sujikathir:
Addressed the remaining numeric representation edge in `1d032b23`: finite, non-negative integral `number_float` bounds now use the same string and array thresholds as integer values. Added regressions for `maxLength: 2000.0` and `minItems: 2001.0`, plus fractional and negative guards. The branch is rebased onto current `main`; the focused CTest passes 1/1 and `lemond` builds successfully on macOS. Ready for re-review—thanks!

fl0rianr (approved):
Thanks! LGTM now. CI currently takes a while let's hope we don't get unexpected surprises there.

PR #2808, opened 2026-07-25
fix(api): preserve native TTS stream formats

### LINKED ISSUE #2655: Streaming /v1/audio/speech forces response_format=pcm, breaking backends that don't advertise pcm (e.g. OpenMOSS)
Found during a code review of #2626 (not caused by it). cc @pwilkin (author of the audio/speech format handling in #2580).

**Bug:** In the streaming `/v1/audio/speech` handler (`src/cpp/server/server.cpp` ~L3265-3285, added in #2580), streaming requests force `response_format = "pcm"`, then a `format_supported` check validates it against the backend's `supported_audio_formats()`. Any TTS backend that overrides that list without including `pcm` is rejected for **all** streaming requests.

**Repro:** `POST /v1/audio/speech {"model":"OpenMOSS-TTS","input":"hello","stream":true}`. OpenMOSS's `supported_audio_formats()` returns `{"wav"}`; `"pcm"` isn't in it, so the request 400s ("response_format 'pcm' is not supported") instead of streaming.

**Related (same area — worth fixing together):**
- Non-streaming default changed from OpenAI's `mp3` to `supported_formats.front()` when a backend advertises formats — silently alters the default container for OpenAI-compatible clients.
- `streaming_proxy.cpp` writes a JSON error object into the live audio sink on a backend non-200, corrupting an already-committed binary audio stream (vs. a clean truncation).

sujikathir:
## Summary

Fixes #2655 by separating TTS transfer mode from response-format negotiation, so `stream: true` preserves a backend's explicit or native format instead of always forcing PCM. The selected format is forwarded to the backend and used for the response MIME type, while `stream_format: "audio"` retains its documented raw-PCM behavior.

## Testing

- `ctest --test-dir build -R '^TtsResponseFormatTest$' --output-on-failure` — 1/1 passed
- `cmake --build build --target lemond -j 6` — compiled and linked on macOS ARM64
- `black --check test/server_tts_openmoss.py` — passed with Black 26.1.0
- `python3 docs/tools/gen_backend_boilerplate.py --check` — generated files are current
- Added an OpenMOSS end-to-end regression test asserting that `stream: true` returns WAV; it runs in the supported Linux/Windows GPU CI matrix

## Screenshots

Not applicable; this changes backend/API format negotiation. OpenMOSS inference is not supported on the local Apple hardware, so local validation covers the deterministic format-selection test and full server build while CI exercises the hardware-backed regression.

## AI Note

AI was used to assist with implementation, tests, verification, and PR preparation. The changes and evidence were locally reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2447,6 +2447,17 @@ if(EXISTS "${_HTTP_CLIENT_SECURITY_TEST_SRC}")
     add_test(NAME HttpClientSecurityTest COMMAND test_http_client_security)
 endif()
 
+set(_TTS_RESPONSE_FORMAT_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_tts_response_format.cpp")
+if(EXISTS "${_TTS_RESPONSE_FORMAT_TEST_SRC}")
+    add_executable(test_tts_response_format test/cpp/test_tts_response_format.cpp)
+    target_include_directories(test_tts_response_format PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+    target_link_libraries(test_tts_response_format PRIVATE nlohmann_json::nlohmann_json)
+    add_test(NAME TtsResponseFormatTest COMMAND test_tts_response_format)
+endif()
+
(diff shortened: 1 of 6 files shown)


sujikathir:
Closing as superseded by #3029, which merged the same WAV-only streaming fix and explicit `response_format` handling with more complete per-backend streaming constraints, format forwarding, documentation, and live OpenMOSS/Kokoro regressions. Current `main` already contains the intended behavior, so refreshing this draft would duplicate the accepted implementation.

PR #2815, opened 2026-07-25
feat(tray): support HTTPS server configurations

abn:
Extracts target URL parsing logic to a shared utility and enables HTTPS compilation definitions and mbedtls linking for standalone macOS/Linux system tray clients. This allows the tray client to connect to HTTPS-secured remote server instances.

Windows `LemonadeServer.exe` runs its embedded server locally and continues to connect over HTTP, but utilizes the new unified constructor config options.

### Key Changes:
* **Shared URL parsing utility:** Moves target URL parsing logic to a shared `url_utils` module (`std::string_view` input, `std::from_chars` port parsing with strict validation of range `[1, 65535]` and trailing characters).
* **CMake target dependency ordering:** Swaps target directories in CMake so that the tray targets link against `lemonade-digest-crypto` (packaging mbedTLS). Links `Security.framework` on macOS if SSL is enabled.
* **Tray Client HTTPS Refactoring:** Integrates a `TrayUIOptions` configuration struct to initialize `TrayUI`, formats URL connection hosts (bracketing IPv6 addresses like `[::1]`), and restricts the `localhost` to `127.0.0.1` rewrite logic to plain HTTP to preserve TLS SNI/hostname verification.
* **Testing:** Adds C++ unit test coverage for various host, scheme, and port edge cases, alongside a Python end-to-end integration smoke test running `lemonade-tray` against a mock HTTPS server.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -750,6 +750,9 @@ jobs:
           echo "Running WebSocket auth tests..."
           .venv/bin/python test/server_websocket_auth.py
           echo "WebSocket auth tests PASSED!"
+          echo "Running HTTPS tray smoke test..."
+          .venv/bin/python test/test_tray_https.py
+          echo "HTTPS tray smoke test PASSED!"
           echo "Running job engine tests..."
           .venv/bin/python test/server_jobs.py --lemond-binary ./build/lemond
           echo "Job engine tests PASSED!"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -652,6 +652,7 @@ set(SOURCES_CORE
     src/cpp/server/utils/json_utils.cpp
     src/cpp/server/utils/process_manager.cpp
     src/cpp/server/utils/path_utils.cpp
+    src/cpp/server/utils/url_utils.cpp
     src/cpp/server/utils/version_utils.cpp
     src/cpp/server/utils/wmi_helper.cpp
     src/cpp/server/utils/network_beacon.cpp
@@ -1327,11 +1328,11 @@ if(APPLE)
     endif()
 endif()
 
-# Add tray application subdirectory
-add_subdirectory(src/cpp/tray)
-
 # Add CLI application subdirectory
 add_subdirectory(src/cpp/cli)
+
+# Add tray application subdirectory
+add_subdirectory(src/cpp/tray)
 
 # Wire WiX installer targets to the binaries and frontend artifacts they package.
 # Doing this after the subdirectories are added ensures MSBuild gets real project refs.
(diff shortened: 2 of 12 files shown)


fl0rianr (changes requested):
Thanks for the PR — extracting the target URL parsing into a shared utility is a good direction, and the added parser coverage is useful.

I found several issues that should be addressed before merge:

1. **macOS build is currently broken**

The `lemonade-tray` target now enables `CPPHTTPLIB_MBEDTLS_SUPPORT`, but it is not linked against Apple’s `Security.framework`.

The failing macOS CI job reports unresolved symbols including:

* SecCertificateCopyData
* SecTrustSettingsCopyCertificates

Please add `Security.framework` to the macOS tray link dependencies.

2. **The Windows `--silent` argument is now passed as `is_ssl`**

The constructor changed from:

```cpp
TrayUI(int port, const std::string& host, bool silent = false);
```

to:

```cpp
TrayUI(int port, const std::string& host, bool is_ssl = false, bool silent = false);
```

However, the Windows call site still uses:

```cpp
TrayUI tray(runtime_config->port(), runtime_config->host(), silent);
```

This means `--silent` now sets `is_ssl = true`, while `silent` remains false. In the Windows startup path this can make the tray try HTTPS against the local HTTP server and also show the notification that should be suppressed.

This should be:

```cpp
TrayUI tray(runtime_config->port(), runtime_config->host(), false, silent);
```

Using an options struct instead of adjacent boolean parameters may also prevent similar regressions.

3. **The PR description overstates Windows HTTPS support**

The non-Windows tray parses a remote target URL, but `LemonadeServer.exe` still starts and connects to its embedded local server and explicitly calls `wait_for_server(..., false, ...)`.

Either the Windows remote HTTPS path needs to be implemented, or the PR description should be narrowed to the standalone macOS/Linux tray client.

4. **Raw IPv6 produces invalid URLs**

The parser accepts `::1`, but the client later constructs:

```text
http://::1:13305
```

IPv6 literals must be bracketed in URLs:

```text
http://[::1]:13305
```

It would be safer to centralize endpoint construction and bracket IPv6 hosts there.

5. **Rewriting `localhost` to `127.0.0.1` can break TLS verification**

For HTTPS targets, changing `localhost` to `127.0.0.1` changes the TLS hostname and SNI value. A certificate valid for `localhost` may therefore fail verification.

For HTTPS, the original hostname should be preserved. Only wildcard bind addresses such as `0.0.0.0` need translation to a connectable address.

6. **Port parsing accepts trailing garbage**

`std::from_chars` success alone is insufficient. Inputs such as:

```text
example.com:443junk
```

are currently accepted as port `443`.

Please also verify:

```cpp
ptr == port_str.data() + port_str.size()
```

Port `0`, empty ports, trailing characters and values above `65535` should be covered by tests.

Because of the macOS linker failure and the Windows constructor regression, I do not think this is merge-ready yet. Once those are fixed, an end-to-end HTTPS tray smoke test would also be valuable, since the current tests only validate parsing.

fl0rianr (changes requested):
The current red CI appears unrelated to this PR: the macOS build and C++ tests passed, while the existing Whisper realtime test failed due to the known isolation race being addressed separately in #2821.

I found two remaining issues before merge:

1. `test/test_tray_https.py` is currently not wired into CI, and its success detection is broken. A healthy tray keeps running, so `proc.communicate(timeout=0.1)` repeatedly raises `TimeoutExpired`; the partial output is discarded, meaning `"Connected to lemond"` will never reach `stdout_data`. Please either have the mock HTTPS server record the health request or use a proper non-blocking output reader, and add the test to the macOS CI path.

2. Raw IPv6 is bracketed for `httplib::Client`, but not in `open_web_app()`. With `--host ::1`, the browser URL becomes `http://::1:13305/` instead of `http://[::1]:13305/`. Please reuse the same host-formatting helper for both client and browser URLs.

Once these two points are fixed, the flaky Whisper failure should not block this PR.

fl0rianr (approved):
Thanks, LGTM now. CI is bothering us again...

abn:
I swear, the CI is out to get me :sweat_smile:

abn:
@fl0rianr hopefully #2837 will fix the windows one.

fl0rianr:
Thanks, still ready for me going to merge now.

PR #2817, opened 2026-07-25
docs(telemetry): improve telemetry and arize phoenix setup guide

abn:
This PR improves the telemetry and tracing documentation, specifically updating the Arize Phoenix local observability integration guide to make it bulletproof and easy to follow for non-experts.

### Context and Rationale
- **Container Registry Update**: Corrected the Arize Phoenix container pull/run instructions to pull the official `docker.io/arizephoenix/phoenix:latest` image name (using container name `phoenix`).
- **Port Mapping Clarification**: Detailed the exact purpose of each port (`6006` for HTTP ingest and Web UI, `4317` for gRPC OTLP, and noted that standard OTel port `4318` is unused by Phoenix) to prevent protocol mismatch errors.
- **Mandatory Project Header**: Added the mandatory `x-project-name` routing header to all local Phoenix connection string examples.
- **Improved Setup Flow**: Reordered the sections so the local step-by-step integration guide starts directly after the concepts introduction, ensuring a seamless user quickstart.
- **JSON Configuration Snippet**: Added a minimal `config.json` telemetry configuration snippet as a manual file-editing alternative to the CLI commands.

> [!IMPORTANT]
> The documented CLI configuration commands (such as setting JSON arrays and objects) presuppose that the parser updates from **PR #2816** (`fix(cli): parse configuration values as JSON arrays and objects`) are merged.

### DIFF
--- a/docs/guide/telemetry.md
+++ b/docs/guide/telemetry.md
@@ -20,112 +20,15 @@ graph TD
 Lemonade supports two co-existing trace formats:
 
 1. **OpenInference (`openinference.*`)**:
-   - A mature convention designed specifically for LLM and agentic application tracing.
+   - A convention designed specifically for LLM and agentic application tracing.
    - Deeply integrated into AI evaluation suites (e.g., **Arize Phoenix**, **Langfuse**).
-   - Captures rich LLM attributes (such as `llm.token_count.prompt`, `llm.token_count.completion`, and raw inputs/outputs/thinking).
+   - Captures rich LLM attributes (such as token counts, prompt inputs, completion outputs, and raw thinking blocks).
 2. **OpenTelemetry GenAI (`gen_ai.*`)**:
    - The official, vendor-neutral standard defined by the OpenTelemetry community.
    - Supported by mainstream APM backends (e.g., **Honeycomb**, **Datadog**, **Grafana**).
-   - Captures standardized attributes (such as `gen_ai.system`, `gen_ai.request.model`, and `gen_ai.usage.input_tokens`).
+   - Captures standardized attributes (such as system type, request model, and input/output token usage).
 
-Users can specify one or both formats. When both are enabled, Lemonade packs attributes for both conventions into a **single pass** and exports them in a **single network payload** to eliminate network overhead.
-
----
-
-## Distributed Tracing (W3C Trace Context)
-
-By default, each inference request starts its own root trace. When a caller is itself an instrumented application (e.g. a multi-agent orchestrator), you can have Lemonade's spans join the caller's trace instead, so an entire multi-step workflow appears as one connected tree.
-
-Set `telemetry.trust_incoming_trace_context=true` to opt in. Lemonade then reads the standard [W3C `traceparent`](https://www.w3.org/TR/trace-context/) header on incoming requests: the request's span adopts the header's trace id and is parented to the header's span id. Requests without a valid `traceparent`, or received while the setting is `false`, keep the default root-span behavior.
-
-```bash
-lemonade config set telemetry.enabled=true \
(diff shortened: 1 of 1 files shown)


fl0rianr (commented):
Thanks for improving the Phoenix setup guide — the corrected image name, ports, and reordered quickstart are a clear improvement.

I found two issues that should be fixed before merge:

1. **Manual config.json edits require a Lemonade restart.**
   The guide currently moves directly from editing the file to sending a test request. Lemonade loads `config.json` at server startup, so an already running server will continue using the previous telemetry configuration. Please explicitly tell users to restart Lemonade after Option B and link to the existing OS-specific configuration paths.

2. **Set `telemetry.otlp.protocol=http/protobuf` explicitly.**
   The Phoenix endpoint at `http://localhost:6006/v1/traces` expects OTLP protobuf. Although this is Lemonade’s default, a user may already have `http/json` configured, and partial config updates preserve that old value. To make the quickstart deterministic, please add the protocol to both examples:

```bash
lemonade config set telemetry.enabled=true \
                    telemetry.otlp.endpoint=http://localhost:6006/v1/traces \
                    telemetry.otlp.protocol=http/protobuf \
                    telemetry.otlp.semantics='["openinference"]' \
                    telemetry.otlp.headers='{"x-project-name":"lemonade"}'
```

```json
"otlp": {
  "endpoint": "http://localhost:6006/v1/traces",
  "protocol": "http/protobuf",
  "semantics": ["openinference"],
  "headers": {
    "x-project-name": "lemonade"
  }
}
```

One wording nit: `x-project-name` is useful for explicit project routing, but it is not mandatory because Phoenix can fall back to its default project. I would describe it as an optional project-routing header rather than a required one.

With those changes, the guide should be much more reliable for first-time users.

fl0rianr (commented):
Thanks for addressing the previous feedback — the Phoenix setup is now much clearer and the configuration examples look correct.

I found one remaining reliability issue in the quickstart:

Phoenix is started in detached mode, but the guide immediately proceeds to the test request without waiting for the service to become ready. Since Lemonade uses `telemetry.otlp.max_retries=0` by default, the first telemetry batch can be dropped if Phoenix is still starting up. A later manual flush cannot recover that already failed batch.

Please add a readiness step before sending the test request, for example:

```bash
until curl -fsS http://localhost:6006/ > /dev/null; do
  sleep 1
done
```

Alternatively, explicitly tell users to continue only after the Phoenix UI is reachable at `http://localhost:6006`.

Two minor documentation nits that could also be fixed while touching the file:

* The privacy section says redacted values are replaced with empty values, but the implementation uses `[REDACTED]`.
* `telemetry.max_attribute_length` is a supported setting with a default of `4096`, but it is missing from the settings reference.

The readiness issue is the only point I would address before approval.

fl0rianr (commented):
Thanks — the previous feedback has now been addressed cleanly. The readiness step, `[REDACTED]` wording, and `telemetry.max_attribute_length` reference are all welcome improvements.

I found one small technical issue in the newly added readiness check:

Phoenix exposes separate liveness and readiness endpoints. `/healthz` returns `OK` without checking dependencies, while `/readyz` verifies database availability and returns `503` when Phoenix is not ready. Since this step explicitly blocks until the service is ready, please use:

```bash
until curl -fsS http://localhost:6006/readyz > /dev/null; do
  sleep 1
done
```

One non-blocking wording nit: `telemetry.max_attribute_length` is enforced using `std::string::length()` and `substr()`, so the limit is measured in bytes rather than Unicode characters. “Maximum attribute length in bytes” would be more precise.

After the `/readyz` adjustment and green CI, this looks ready to approve.

fl0rianr (approved):
LGTM now,

Issue #12, opened 2025-05-28
Response Id `0`


sp3c1:
All the responses comes with response_id `0` (completions as well)

```
{
  id: '0',
  choices: [
    {
      finish_reason: 'stop',
      index: 0,
      logprobs: null,
      message: {
        content: `I'm an artificial intelligence model known as Llama. Llama stands for "Large Language Model Meta AI." `,
        refusal: null,
        role: 'assistant',
        annotations: null,
        audio: null,
        function_call: null,
        tool_calls: null
      }
    }
  ],
  created: 1748471406,
  model: 'amd/Llama-3.2-3B-Instruct-awq-g128-int4-asym-fp16-onnx-hybrid',
  object: 'chat.completion',
  service_tier: null,
  system_fingerprint: null,
  usage: null
}
```

To be honest all Lemonade REST api docs also shows they are `0` which makes it kind of unusable to chain anything (for the responses API).

Or is it expected on v0?

jeremyfowers:
@danielholanda

danielholanda:
Hi @sp3c1, thanks for sharing this!
This is a current limitation that should be fairly easy to fix on our side. I will prioritize fixing this.
Are you using this on any specific application that allows us to test chaining once this has been addressed? I'm also always curious to know more about what people are trying to use Lemonade for.

sp3c1:
Well I got myself 395(mobile workstation) lately with 128gb and I am testing how feasible it is to have a real local agent.

Been testing LM Studio with continue.dev / clide, and finding some models that would give a decent token count while trying to finish a pet project of mine:  a psychometric platform ;p.

But I did already hooked in Lemonade to cline and to continue. The chat/edit works fine, but agent dies - the intial context window is already x3 bigger and you can not send anything. Think already someone raised an issue about it.

Generally after this project, I am planning to create more local tools and agent to support myself (with something that works and I control - windows MCP experience so far was abysmal for me) so this is where my interest in Lemonade is. Runing some of those models in LM without proper support was intially bit of a bummer, but say I got a few now that work decent enough to be a viable option. It is really nice to see a better support for strix platform, cause it seems to me a real change in the AI space will be when it will become trully local and what I have been looking forward to.

danielholanda:
This is awesome! We are planning on a release in the next few days that might unblock you in many fronts there.

I'll post an update here once the response ID issue is resolved, and I'll also keep you informed about our progress on those other fronts.

sp3c1:
[...]

Issue #10, opened 2025-05-21
Improve ability and documentation around third-party models


ethanblake4:
For example, I found [this model](https://huggingface.co/aigdat/Qwen2.5-Coder-7B-hybrid-onnx-rai1.4) on Hugging Face which I was eventually able to get working in Hybrid mode, but the documentation around this is very sparse. Additionally, it does not show up in the /api/v0/models endpoint which makes it hard to integrate with third-party tools.

jeremyfowers:
Hi @ethanblake4, I agree we should make these improvements.

jeremyfowers:
@ethanblake4 Lemonade Server v8, which just released, completely overhauls the way third-party models are handled.

- The `pull` endpoint can now register and install any third-party checkpoint from hugging face
- Registered third-party models now show up in the `models` endpoint
- Now there is a model management GUI in the web app:

![Image](https://github.com/user-attachments/assets/9167217c-6dcc-483b-8689-e8fc2222cc73)

Please try it out and let us know your feedback!

jeremyfowers:
Closing as completed, but please reopen if you have more requests.

PR #2818, opened 2026-07-26
feat(server): add model alias system with /v1/models listing and /internal/aliases endpoints

abn:
Adds support for standalone model aliases across the Lemonade server. Models can have multiple human-readable alternative names (e.g., `coding`, `fast-chat`) that resolve to base models, appear as independent entries in `/v1/models`, and share all backend configurations and LRU cache instances. Aliases are managed via a standalone `AliasManager` and persisted in `<cache_dir>/aliases.json`.

### High-Level Feature Overview & Real-World Use-Cases

Model Aliases allow AI developers, automated agents, and external applications to reference local models using stable, friendly names rather than long, version-specific checkpoint paths.

1. **High-Availability Active-Standby & Hardware Failover**:
   - *Problem:* Multi-node AI deployments or fallback setups across heterogeneous hardware (e.g., a high-end discrete GPU server vs. a low-power edge/iGPU node) require failover capabilities when the primary inference server is offline, undergoing maintenance, or experiencing heavy load.
   - *Solution:* Downstream applications send inference traffic to a single, unified alias (such as `production-llm`). On primary high-class hardware, `production-llm` resolves to a large model (e.g., a 70B parameter checkpoint); on secondary active-standby nodes, `production-llm` maps to a compact fallback model (e.g., an 8B or quantized checkpoint). Client applications experience uninterrupted service during failover scenarios without changing their request payloads.

2. **Agent & Application Stability**:
   - *Problem:* Hardcoding specific model identifiers (e.g., `Llama-3.2-1B-Instruct-GGUF`) in scripts or agent code breaks workflows whenever a new model checkpoint or version is pulled.
   - *Solution:* Configure agents to target a stable alias like `coding` or `default-reasoner`. Upgrading the underlying model (e.g., from `Llama-3.2-1B` to `Qwen3-8B`) requires updating the alias target once—zero script or agent code modifications required.

3. **OpenAI SDK & Third-Party Tool Compatibility**:
   - *Problem:* Third-party tools or frameworks expect specific model names (e.g., `gpt-4o` or `claude-3-5-sonnet`) and fail when querying standard OpenAI endpoints against custom local model names.
   - *Solution:* Assign `gpt-4o` or `claude-3-5-sonnet` as an alias to your preferred local model. The alias appears in `GET /v1/models` and responds seamlessly to `/v1/chat/completions`.

4. **No Configuration Duplication**:
   - *Problem:* Previously, giving a model an alternative name required duplicating entire JSON model definitions in `user_models.json`, leading to maintenance drift when changing recipe options.
   - *Solution:* Aliases share 100% of the base model's configuration, checkpoints, recipes, and running process memory.

### Key Technical Changes

- **Standalone Alias Engine (`AliasManager`)**:
  - Implemented standalone `AliasManager` class (`src/cpp/include/lemon/alias_manager.h`, `src/cpp/server/alias_manager.cpp`) persisting directly to `<cache_dir>/aliases.json` as `{"alias": "target"}`.
  - Thread-safe query and mutation operations guarded by `std::shared_mutex` with out-of-lock JSON serialization and atomic file replacement (`.tmp` write, POSIX `0600` permissions, `rename`/`copy_file` fallback).
  - Graph cycle detection ($A \to B \to A$), self-reference guards ($A \to A$), chain depth limits (`MAX_ALIAS_CHAIN_DEPTH = 10`), and corrupted JSON auto-quarantine recovery (`aliases.json.corrupted`).
  - Decoupled `ModelInfo` completely; `Server` owns `AliasManager` and passes an unowned raw pointer to `ModelManager`.

- **API & Routing**:
  - Injected explicit aliases as independent model entries across all quad-prefixes (`/api/v0/`, `/api/v1/`, `/v0/`, `/v1/`) for `GET /models` and `GET /models/{model_id}`.
  - Added administrative internal endpoints protected by `LEMONADE_ADMIN_API_KEY`:
    - `GET /internal/aliases` — List all alias bindings.
    - `POST /internal/aliases` — Add/update alias binding (`{"alias": "coding", "target": "Qwen3-8B-GGUF"}`).
    - `DELETE /internal/aliases/{alias}` — Remove alias binding.

- **CLI Commands**:
  - Implemented CLI commands (`lemonade alias add`, `lemonade alias remove`, `lemonade alias list`, and `--alias` flag on `lemonade pull`).

- **Testing & Documentation**:
  - Added standalone C++ unit test suite in `test/cpp/test_model_alias.cpp` covering resolution, CRUD operations, disk persistence, cycle detection, corrupted file recovery, and `ModelManager` integration (`ALL TESTS PASSED!`).
  - Added Python integration test `test_021aa_internal_aliases_endpoints` in `test/server_endpoints.py`.
  - Updated `docs/guide/configuration/custom-models.md`, `docs/guide/cli.md`, `docs/api/lemonade.md`, `README.md`, and `AGENTS.md`.


kenvandine:
@claude review this please

abn:
Thank you @claude for the thorough, high-quality review! 

### 1. Dedicated `aliases` Overlay in `user_models.json` (Blocking #1, Correctness #7)
* **Fix**: Instead of synthesizing shadow model definitions when aliasing built-in or extra models (which stripped `labels`, `mmproj`, `recipe_options`, and corrupted `info.type`), alias bindings are now stored in a top-level `"aliases": { "alias": "target" }` object inside `user_models.json`.
* **Result**: Built-in and extra models retain 100% of their base metadata and bare wire IDs. Removing an alias erases the key from `"aliases"` without leaving empty shadow entries behind.

### 2. CLI Implementation (Blocking #2)
* **Fix**: Implemented `lemonade alias add <ALIAS> <TARGET>`, `lemonade alias remove <ALIAS>`, `lemonade alias list`, and `lemonade pull <MODEL> --alias <ALIAS>` in `src/cpp/cli/` (`LemonadeClient` and `main.cpp`).
* **Result**: Code matches `README.md` and `docs/guide/cli.md` 100%.

### 3. MSVC Windows Build Compatibility (Blocking #3)
* **Fix**: Replaced POSIX `setenv()` in `test/cpp/test_model_alias.cpp` with a cross-platform `#ifdef _WIN32 _putenv_s ... #else setenv ... #endif` helper.

### 4. O(1) Trimmed Alias Lookup Map & Cache Sorting (Correctness #4, Performance #9, #10)
* **Fix**: Maintained an explicit trimmed map `explicit_aliases_` during cache build for O(1) lookups in `is_explicit_alias()` and `/internal/aliases`.
* **Optimization**: Sorted pointers `const std::pair<const std::string, ModelInfo>*` in `rebuild_public_model_aliases_locked()` without deep-copying `models_cache_`.

### 5. HTTP Error Codes & Conflict Detection (Correctness #5, #6, #8)
* **Fix**: `set_model_alias()` now checks `public_model_aliases_.count(clean_alias)` under the lock to detect primary model, canonical prefix, and alias collisions.
* **Fix**: `handle_aliases_add()` returns HTTP 409 Conflict if alias binding is rejected.
* **Fix**: `handle_aliases_remove()` catches `std::invalid_argument` and returns HTTP 400.

### 6. Documentation Fixes (Docs #11)
* **Fix**: Restored `### Five reference cases` heading and replaced LaTeX math blocks with clean Markdown backticks in `docs/guide/configuration/custom-models.md`.

### 7. Test Suite Expansion (Tests #12)
* **Unit Tests**: Added `test_builtin_model_alias()` in `test_model_alias.cpp` (all 30 assertions passing).
* **Integration Tests**: Added `test_021aa_internal_aliases_endpoints` in `test/server_endpoints.py` testing `POST`, `GET`, and `DELETE` on `/internal/aliases` (all 70 tests passing).

jeremyfowers (dismissed):
afaik this feature already exists, in the form of the "user models" and "recipes" systems. You can save a pointer to any Lemonade model with any name you like. These custom pointers carry a "custom" label that can be filtered.

@abn can you distinguish the features proposed in this PR from the system already available?

abn:
Hi @jeremyfowers,

Great question! The user models / recipes system and the proposed model alias system serve two complementary layers:

* **User Models / Recipes** = **Definition Layer** (authoring *new* model configurations, custom checkpoints, and recipe parameters).
* **Model Aliases** = **Symbolic Routing Layer** (a lightweight "CNAME" / symlink that maps a stable public name to any existing model definition).

Here are the key technical differentiators and why user model definitions alone are insufficient for this use case:

**1. Dynamic Metadata Inheritance & No Upstream Drift**
Creating a user model "pointer" requires hardcoding a standalone JSON definition. If the underlying built-in model updates its backend, recipe, `recipe_options` (e.g., `vllm_args`, `sdcpp_args`), `mmproj`, or collection components in `server_models.json` / `backend_versions.json`, a user model definition drifts out of sync. 

In contrast, an **alias inherits 100% of the target model's live metadata dynamically**. When the target built-in model updates, the alias automatically reflects the update with zero maintenance.

**2. Preserving Capability Classification & LRU Pool Routing**
Non-LLM models (embeddings, rerankers, vision, collections) rely on specialized `labels` to determine their deployment type (`ModelType::Embedding`, `ModelType::Reranking`, `CollectionOrchestrator`). Re-declaring a built-in embedding model as a user model without explicit labels reclassifies it as a standard LLM, which breaks LRU pool eviction (`router.cpp` scopes LRU per `ModelType`) and causes routing failures on `/v1/embeddings`. Aliases preserve the target model's capability interfaces and LRU pool routing intact.

**3. Dynamic Active-Standby Failover & Runtime Management**
A primary use case for aliases is **environment-independent naming and active-standby failover**. Infrastructure managers can expose a fixed alias like `coding` or `fast-chat` to LiteLLM / OpenAI SDK clients, and re-point the target model at runtime via `POST /internal/aliases` (or `lemonade alias add`) — for instance, switching `coding` from a large GPU model to a smaller NPU model when primary hardware is unavailable — without re-authoring or deleting model definitions.

**4. First-Class CLI & REST Administrative API**
This PR adds explicit CLI subcommands (`lemonade alias add`, `lemonade alias remove`, `lemonade alias list`, `lemonade pull --alias`) and internal administrative REST endpoints (`/internal/aliases`), enabling scripts and orchestration tooling to inspect and mutate model aliases cleanly without directly editing JSON files on disk.

jeremyfowers:
@abn I understand the justification for the aliases now. My main concern with this PR is that its a lot of new surface area for a simple sounding feature.

@bitgamma would you be willing to review this one? My hope is that the alias idea could be mapped onto the pre-existing machinery for user models somehow, to reduce the amount of new surface area needed. But if that is not possible we can add the surface area to enable this.

test/utils/server_base.py
```diff
@@ -141,21 +141,31 @@ def wait_for_server(port=PORT, timeout=60):
...
-        except socket.error:
+        except Exception:
             time.sleep(1)
 
 
 def _auth_headers():
-    """Return Authorization header if LEMONADE_API_KEY is set."""
-    api_key = os.environ.get("LEMONADE_API_KEY")
+    """Return Authorization header if LEMONADE_ADMIN_API_KEY or LEMONADE_API_KEY is set."""
+    api_key = os.environ.get("LEMONADE_ADMIN_API_KEY") or os.environ.get(
```
bitgamma: these changes are not related to the PR topic

test/utils/server_base.py
```diff
@@ -141,21 +141,31 @@ def wait_for_server(port=PORT, timeout=60):
...
     while True:
         if time.time() - start_time > timeout:
             raise TimeoutError(f"Server failed to start within {timeout} seconds")
         try:
-            conn = socket.create_connection(("localhost", port))
-            conn.close()
+            # Check via HTTP so proxy settings are respected in sandboxed environments
+            url = f"http://127.0.0.1:{port}/live"
+            req = urllib.request.Request(url, method="HEAD")
+            with urllib.request.urlopen(req) as response:
```
bitgamma: this change also seems related to something else, please extract in a separate PR if needed

test/server_endpoints.py
```diff
@@ -3741,7 +3807,9 @@ def test_021zv_routing_validate_undeclared_component_returns_400(self):
         )
         self.assertEqual(response.status_code, 400, response.text)
         self.assertIn("undeclared-classifier-model", response.json()["error"])
-        print("[OK] /routing/validate rejected undeclared candidate/classifier-model references with 400")
+        print(
```
bitgamma, line 3827: please do not reformat existing code

src/cpp/server/model_manager.cpp
```diff
@@ -111,7 +111,7 @@ static constexpr auto safe_dir_options = fs::directory_options::none;
 namespace lemon {
 
 // Properties which are defined by the user for model registration.
-static const std::vector<std::string> USER_DEFINED_MODEL_PROPS = std::vector<std::string>{"checkpoints", "checkpoint", "recipe", "mmproj", "size", "image_defaults", "components", "recipe_options", "routing", "system_prompt", "version", "source", "registry_source"};
+static const std::vector<std::string> USER_DEFINED_MODEL_PROPS = std::vector<std::string>{"checkpoints", "checkpoint", "recipe", "mmproj", "size", "image_defaults", "components", "recipe_options", "routing", "system_prompt", "version", "source", "registry_source", "aliases"};
```
bitgamma: why adding it  here? aliases should live in their own dictionary and not be part of the model definition, they should be a server concern

src/cpp/server/model_manager.cpp
```diff
@@ -1623,6 +1635,28 @@ static void parse_components(ModelInfo& info, const json& model_json) {
     }
 }
 
+static void parse_aliases(ModelInfo& info, const json& model_json) {
+    if (!model_json.contains("aliases")) {
```
bitgamma: aliases shouldn't be part of the model definition

src/cpp/server/model_manager.cpp
```diff
@@ -2753,6 +2806,135 @@ void ModelManager::unregister_user_model(const std::string& model_name) {
     cache_valid_ = false;
 }
 
+void ModelManager::set_model_alias(const std::string& alias, const std::string& target_model) {
+    std::string clean_alias = alias;
+    clean_alias.erase(0, clean_alias.find_first_not_of(" \t\r\n"));
```
bitgamma: this name cleaning pattern is duplicated all over the PR

bitgamma (changes requested):
I think this is the kind of feature where I'd like to see some more requests before introducing.

In any case, if adding this feature I think aliases should be completely separate from the model, probably live in their own file formatted as `{"alias1": "actual-model-name", ...}` and out of the model manager altogether (that class is already holding too many concerns imho).

Alias resolution would then be  "try to resolve model name => on failure check alias directory" in places where the model is loaded/checked

abn:
Thanks for the feedback @bitgamma! I have completely refactored the alias implementation based on your review:

1. **Standalone `AliasManager` & `aliases.json` Storage**: Completely decoupled aliases from `ModelInfo` and `ModelManager`. Custom model aliases are now persisted in a standalone `<cache_dir>/aliases.json` file structured as `{"alias": "target-model"}`.
2. **Clean Ownership & Thread Safety**: `Server` owns `std::unique_ptr<AliasManager>`, and `ModelManager` receives a non-owning raw pointer. Reads use `std::shared_mutex` shared locks for non-blocking concurrent REST access, while I/O persistence runs out-of-lock via atomic file replacement (`.tmp` -> `rename` / `copy_file` fallback).
3. **DRY Model Trimming & Reverted Test Diffs**: Centralized whitespace trimming into `lemon::utils::normalize_model_name()`. Reverted all stray diffs in `test/utils/server_base.py`.
4. **Cycle & Recovery Guards**: Added graph traversal validation to prevent circular alias dependencies ($A \to B \to A$) and auto-quarantine for corrupted JSON files (`aliases.json.corrupted`).

docs/guide/configuration/custom-models.md
```diff
@@ -295,7 +295,63 @@ Anywhere a model name is accepted (request bodies, CLI args, URL path parameters
 
 The CLI (`lemonade list`) prints the API `id` verbatim. That means the Name column is always copy-paste-safe — every cell is a valid input to `lemonade load`, `lemonade delete`, `lemonade run`, etc.
 
-The Tauri desktop app and the web app apply a display transformation on top of the API id: bare ids render as `NAME`, and canonical-prefixed ids render as `NAME (registered)` / `NAME (imported)` / `NAME (builtin)`. The suffix appears only for shadowed sources.
+The Tauri desktop app and the web app apply a display transformation on top of the API id: bare ids render as `NAME`, and canonical-prefixed ids render as `NAME (registered)` / `NAME (imported)` / `NAME (builtin)`. The suffix appears### Model Aliases (`aliases.json`)
```
bitgamma: what happened on this line?

bitgamma (changes requested):
There is still coupling between aliases and models. The model_manager should see zero changes from this PR for the scope to be right imho. 

In server, you can either:

1. try to find model, if not try to find alias and try again to find model with the resolved name
2. try to resolve alias, then try to resolve model with either the resolved alias or the original name passed in the request.

I am fine with either, 2. seems easier to implement and more concise.

Meaning the changes to server.cpp should also only be limited to that (and the addition of the endpoints).

This also reflect in the documentation which says "Models can define aliases". I think this is not the case, rather, *you* can define aliases for model names. The aliased thing needs not be aware of being aliased. Let's see them as high-level name that can resolve to any model name.

docs/guide/configuration/custom-models.md
abn: :sweat_smile: good catch; think I fat figured there a bit.

abn:
Thanks for the feedback. I like the approach you proposed, and have moved to an upfront server resolution of the alias.

bitgamma (changes requested):
Ok, I had DeepSeek V4 Flash have a look at it. Feel free to push back if some things do not apply                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                              
 ### Docs/implementation mismatches (should be fixed before merge)                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                              
 1. canonical_target documented but never emitted. docs/api/lemonade.md shows GET /internal/aliases returning a canonical_target field, but handle_aliases_get (server.cpp ~5608) only returns alias, target, downloaded, recipe. Either emit it or fix the docs.                                                             
                                                                                                                                                                                                                                                                                                                              
 2. custom-models.md over-describes a precedence/conflict system that isn't implemented. The docs claim:                                                                                                                                                                                                                      
     - "Source Precedence & First-Matched Wins … models with higher precedence claim aliases first … first registered model claims it" — no such mechanism exists; aliases are a flat map with no per-source precedence.                                                                                                      
     - "logs a Model alias conflict warning at startup and ignores the conflicting duplicate" — no startup conflict logging or duplicate-ignoring exists anywhere.                                                                                                                                                            
     - "Primary model names always take precedence over aliases" — only partially true (creation is blocked on collision, and handle_model_by_id checks the model first), but normalize_and_resolve_request_model resolves aliases unconditionally in request paths.                                                          
       The docs should be aligned with the actual (simpler) behavior, or the behavior implemented.                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                              
 3. --alias documented as repeatable but implemented as single-value. docs/guide/cli.md says "Can be specified multiple times," but pull_cmd->add_option("--alias", config.alias_name, ...) is a single string; main.cpp only adds one alias.                                                                                 
                                                                                                                                                                                                                                                                                                                              
 ### Correctness / robustness                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                              
 4. resolve_alias returns a mid-chain alias on cyclic maps loaded from disk (latent bug). set_alias blocks cycles at creation, but load()/set_cache_dir() never validate cycles. For a hand-edited a→b→a, resolve_alias("a") breaks out and returns "b" (still an alias) rather than detecting the cycle. Only reachable via  
    manual aliases.json edits, but worth hardening in load().                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                              
 5. Chained aliases never appear in /v1/models listing. handle_models resolves the target via model_manager_->resolve_model_name(target_name), which knows nothing about aliases, then falls back to models.count(target_name). If an alias targets another alias, neither matches, so it's silently dropped from the listing 
    even though request-path resolution handles chains.                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                              
 6. No validation that the alias target exists. handle_aliases_add accepts any target string. A typo produces a "success" (200) that later silently 404s in /v1/models/{id} and is omitted from the listing. May be intentional (failover to not-yet-downloaded targets), but the docs don't say so, and the 200→404 UX is    
    surprising. At minimum surface the downloaded=false marker the GET endpoint already computes.                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                              
 7. Duplicate /v1/models ids on collision from disk. handle_aliases_add rejects aliases that collide with existing models, but load()/set_cache_dir() don't. A hand-edited alias matching a real model id yields duplicate entries with the same id.                                                                          
                                                                                                                                                                                                                                                                                                                              
 8. Alias coverage is limited to chat/completions/load/model-listing. normalize_and_resolve_request_model is only wired into handle_chat_completions, handle_completions, and handle_load. Other model-consuming endpoints (embeddings, generate, classify, run_classifier, etc.) won't resolve aliases. The docs claim "any  
    API request" resolves aliases — that's overstated; either extend coverage or narrow the claim.                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                              
 ### Minor                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                              
 9. copy_file fallback drops 0600 perms on the target when rename fails (aliases.json isn't sensitive, but inconsistent with the stated 0600 invariant).                                                                                                                                                                      
 10. CLI alias remove and the DELETE test build the path from a raw alias without URL-encoding — aliases containing spaces/slashes would break the HTTP path.                                                                                                                                                                 
 11. I could not build the project to confirm compilation; the C++ is straightforward but CI should be green before merge.                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                              
 Overall                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                              
 Functionally sound and well-tested for the happy path, but I'd want the docs/implementation mismatches (#1–#3) resolved and the chained-alias listing gap (#5) addressed (or explicitly documented as unsupported) before merge. The resolve_alias disk-cycle behavior (#4) is worth a defensive fix in load().

abn:
Thanks for the thorough review and edge-case feedback!

- **Docs Alignment (`#1`, `#2`, `#3`, `#6`)**:
  - Removed non-existent `canonical_target` field from `docs/api/lemonade.md`.
  - Updated `docs/guide/cli.md` to specify single-value `--alias` flag on `lemonade pull`.
  - Simplified `docs/guide/configuration/custom-models.md` to accurately reflect `AliasManager`'s flat map and primary model precedence.
  - Clarified in `custom-models.md` that aliases function as symbolic routing links and can target remote or un-pulled models prior to local download.

- **Target Model Validation (`#6`)**:
  - Target model validation on `POST /internal/aliases` / `lemonade alias add` is intentionally non-blocking so users can register aliases for remote Hugging Face checkpoints, cloud models, or pull-on-demand targets prior to local download. `GET /internal/aliases` surfaces `downloaded: false` and `recipe: "unknown"` for non-local targets for full observability.

- **Cycle Hardening & Recovery (`#4`)**:
  - `AliasManager::resolve_alias()` logs a warning and returns `std::nullopt` on cycle detection (`visited.count(next_target)` or depth > 10).
  - `AliasManager::load()` prunes circular references from disk at startup before populating memory (`test_alias_disk_cycle_pruning` added to `test_model_alias`).

- **Multi-Hop & Listing Resolution (`#5`, `#7`)**:
  - `server.cpp` model listing (`/v1/models` and `/v1/models/{id}`) resolves alias targets transitively via `resolve_alias_target()` so multi-hop alias chains (`A -> B -> C`) display the ultimate canonical model's metadata.
  - Primary model ID collision guard skips aliases whose key matches a primary model ID.

- **Universal Endpoint Resolution (`#8`)**:
  - Wired `normalize_and_resolve_request_model()` across `embeddings`, `audio/transcriptions`, `audio/speech`, `images/generations`, `responses`, `classify`, and `unload` handlers.

- **Permissions & URL Encoding (`#9`, `#10`)**:
  - Re-applied POSIX `chmod(0600)` on the `copy_file` fallback path in `alias_manager.cpp`.
  - Encoded alias path parameters with `httplib::detail::url_encode()` in `LemonadeClient::alias_remove` and `requests.utils.quote` in integration tests.

bitgamma (approved):
LGTM. Thanks!

PR #2819, opened 2026-07-26
Add required groups for GPU acceleration to lemonade user in docker container

### LINKED ISSUE #2799: GPU acceleration is broken in docker image
### Platform

Linux/Ubuntu

### Lemonade Version

11.5.0

### GPU / APU Model

Radeon RX 6600 XT

### Component

Other / Not sure

### Bug Description

backends in docker image do not use GPU but CPU instead even if vulkan or rocm backend is configured

### Steps to Reproduce

pull docker image
run container with --device=/dev/dfk --device=/dev/dri
start chat

Expected: GPU acceleration is used
Actual: Log shows only CPU device
rocminfo inside of docker shows Permission denied for /dev/kfd and /dev/dri

### Log Output

```shell

```

### Additional Context

user lemonade is not member of video and render group:
/etc/group:
video:x:44:ubuntu

undefined group for /dev/kfd:
crw-rw---- 1 root **148** 236, 0 Jul 23 18:21 /dev/kfd

adding user lemonade to groups video and render and adding group render with gid 148 solves the issue

klkr1:
Add missing group to docker container
and add groups video and render to user lemonade to access HW for GPU acceleration

Fixes: #2799


jeremyfowers:

### Scope

**minor** — A contained Dockerfile fix that adds group memberships — a single-line surface change with no new user-facing behavior.

### Suggested reviewers

- siavashhub (Docker) — Sole Docker maintainer listed in contribute.md; this PR touches the Dockerfile.

_Reviewed at head `1e5e542` by repo-manager pr-review. Regenerate with `repo-manager review-pr 2819`._

siavashhub:
Thanks for this fix. 
The root cause here (the lemonade user not being in the host's render/video/kfd groups) is the same one #2746 addresses, but that PR maps the device GIDs dynamically at container startup rather than hardcoding them. 

The concern with this PR is that -g 148 is host-specific, 148 is the render/kfd GID on your machine, but on other hosts it differs, so a statically-baked group won't grant access there. #2746's runtime lookup of the actual mapped device GIDs (/dev/kfd, /dev/dri/renderD*) works regardless of host GID.

I'd suggest we consolidate on #2746 so we don't ship two overlapping mechanisms. Happy to fold your #2799 (RX 6600 XT / ROCm) reproducer into #2746 as a validation case, it's a good scenario to confirm the dynamic mapping covers ROCm compute, not just NPU. 

Would you be open to closing this in favor of #2746 once it lands?

klkr1:
You are right, this fix works only in very special setups. I pulled #2746, built a container and tried it, RoCm (and Vulkan) are working with it.

PR #2821, opened 2026-07-26
Fix Whisper realtime test isolation and stale session work

fl0rianr:
## Summary

Fixes a confirmed isolation race in the Whisper realtime integration tests and
prevents stale work from a closed WebSocket session from reaching the backend.

## Observed failure sequence

In both captured macOS failures, test 007 schedules several asynchronous
interim/final Whisper requests from the shared multi-phrase fixture. It later
consumes the first buffered completion and closes the WebSocket while later work
from that session is still pending. Test 008 starts immediately afterward.

The exact reason the following request is reported by libcurl as
`Unsupported protocol` is not proven. The backend already serializes actual
multipart HTTP calls with `inference_mutex_`; this change addresses the proven
session/test isolation defect rather than claiming a libcurl root cause.

Realtime handling also treated a backend `{ "error": ... }` response as a
successful completion with an empty transcript, hiding the real failure.

## Changes

- Configure test 007 with server VAD and a longer silence interval so the shared
  fixture remains one speech window and its explicit commit produces one final
  transcription.
- Serialize interim/final Whisper work within each realtime session.
- Invalidate stale interim work when a final is committed, audio is cleared,
  turn detection is disabled, or the session closes.
- Re-check interim generations after backend execution and discard stale results.
- Skip queued work before backend invocation when its session has closed.
- Convert backend error or malformed responses into realtime `error` events.

### DIFF
--- a/src/cpp/include/lemon/realtime_session.h
+++ b/src/cpp/include/lemon/realtime_session.h
@@ -7,6 +7,7 @@
 #include <functional>
 #include <future>
 #include <atomic>
+#include <cstdint>
 #include <nlohmann/json.hpp>
 #include "streaming_audio_buffer.h"
 #include "vad.h"
@@ -50,6 +51,10 @@ struct RealtimeSession {
     // Interim transcription state
     int64_t last_interim_transcription_ms = 0;  // When we last fired an interim transcription
     std::atomic<bool> interim_in_flight{false};  // Guard against overlapping interim requests
+    std::atomic<std::uint64_t> interim_generation{0}; // Invalidates stale queued interim requests
+    // Serializes all Whisper calls for this session. A final transcription
+    // invalidates queued interim work before it can reach the backend.
+    std::mutex transcription_mutex;
 
     // Streaming backend connection (used when backend supports IStreamingTranscriptionServer).
     // streaming_mutex guards streaming_client against concurrent forward/disconnect.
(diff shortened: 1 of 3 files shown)


abn (changes requested):
Nice work! The lock serialization and stale work invalidation via generations are solid additions that clean up concurrent race conditions and backend overhead.

One minor test isolation issue popped up during review on a machine with NPU hardware: when `test_005` fails to load the model on `npu` and skips, the model remains evicted, causing `test_007` to fail because the model is no longer loaded. We can fix this by explicitly loading the model at the beginning of `test_006` and `test_007`.

abn (changes requested):
Thanks for updating this! Using `self._load_whisper_model_or_fail()` is the right way to isolate the model.

There is just one small detail: `_load_whisper_model_or_fail()` only sends a POST request to `/load` when `whispercpp_backend` is not empty. When it is empty (default on CPU runs), it simply returns the model name without calling `/load`. This means that if a prior test evicted the model (e.g., `test_005` on NPU hardware), `test_006` and `test_007` still fail with `No model loaded: Whisper-Tiny`.

To fix this completely, we should change `_load_whisper_model_or_fail()` to always call `/load`:

```python
    def _load_whisper_model_or_fail(self):
        """Load the configured Whisper model before positive transcription tests."""
        model = _get_whisper_model()
        whispercpp_backend = _get_whispercpp_backend()

        load_payload = {
            "model_name": model,
        }
        if whispercpp_backend:
            print(f"[INFO] Loading model with {whispercpp_backend} backend")
            load_payload["whispercpp_backend"] = whispercpp_backend
        else:
            print(f"[INFO] Loading model {model}")

        load_response = requests.post(
            f"{self.base_url}/load",
            json=load_payload,
            timeout=TIMEOUT_MODEL_OPERATION,
        )
        self.assertEqual(
            load_response.status_code,
            200,
            f"Failed to load model {model}: {load_response.text}",
        )

        return model
```

Once that is updated, the entire test suite passes perfectly!

fl0rianr:
Thanks for taking a look and those very good findings! Now we should have our desired fix to improve CI a bit.

abn (approved):
lgtm 🚀

PR #2825, opened 2026-07-27
Add MiniCPM GGUF models to catalog

ZMXJJ:
## Summary

Add 10 MiniCPM GGUF models from ModelScope to Lemonade's built-in llama.cpp catalog.

The entries cover:

- Text: MiniCPM3, MiniCPM4, MiniCPM4.1, and MiniCPM5
- Vision: MiniCPM-V 4, 4.5, 4.6, 4.6 Thinking, MiniCPM-o 2.6, and MiniCPM-o 4.5

All entries reuse the existing `llamacpp` recipe and ModelScope registry support. This change only updates the model catalog and generated backend reference; no backend or core registry code is changed.

MiniCPM-o entries intentionally expose the vision path only. Audio and full-duplex support are out of scope for this catalog change.

## Validation

- Verified all ModelScope repository and GGUF/mmproj paths.
- Confirmed all 10 entries use `source: modelscope`.
- Confirmed the generated llama.cpp catalog count changes from 77 to 87.
- Validated JSON parsing and registry field constraints.
- Ran `git diff --check`.

### DIFF
--- a/docs/dev/backends-reference.md
+++ b/docs/dev/backends-reference.md
@@ -186,7 +186,7 @@ the generator instead. Prose outside the markers is preserved. -->
 |-------|-----------|--------|
 | `kokoro-v1` | 0.354 | tts |
 
-#### `llamacpp` — Llama.cpp GPU (77 models)
+#### `llamacpp` — Llama.cpp GPU (87 models)
 
 | Model | Size (GB) | Labels |
 |-------|-----------|--------|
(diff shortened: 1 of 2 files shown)


fl0rianr (approved):
Thanks for your contribution! LGTM.

@ramkrishna2910 would you have short look as well? Wouldn't that be the right call for the upcoming release?

ramkrishna2910 (approved):

PR #2827, opened 2026-07-27
README.md - Fix download links

duggiefresh:
Update the download links to the main installation page: https://lemonade-server.ai/docs/guide/install/

[ci skip]


fl0rianr (changes requested):
Thanks for updating the obsolete installation links. The general Download link looks good, but I do not think the PR is safe to merge in its current form.

The main concern is the use of automatically generated anchors such as `#__tabbed_1_2`. These IDs depend on the order and structure of the documentation tabs. Adding, removing, or reordering tabs could silently make the README links select the wrong platform while still appearing valid.

In addition, the distribution-specific badges now all point to the generic Linux tab. Since dedicated pages already exist, these should link directly to their respective installation instructions:

* Arch: `/docs/guide/install/arch/`
* Debian: `/docs/guide/install/debian/`
* Fedora: `/docs/guide/install/fedora/`
* Ubuntu: `/docs/guide/install/ubuntu/`
* Docker: `/docs/guide/install/docker/`

This preserves the existing behavior where a platform-specific badge leads directly to platform-specific instructions.

I would therefore suggest:

1. Keep the general Download link to `/docs/guide/install/`.
2. Use the stable, dedicated pages for the distribution and Docker badges.
3. Avoid relying on generated `#__tabbed_1_n` fragments for long-lived README links, unless stable named tab anchors are introduced.
4. Rebase or update the branch and let the normal checks run, since the current runs are marked `action_required`.

There are also remaining `install_options.html` links elsewhere in the repository, especially in the FAQ. Those can be handled separately, but should be tracked as a follow-up.

duggiefresh:
Thanks for the comments. Updated per feedback

fl0rianr (commented):
Thanks for the update. I think my previous review may not have been explicit enough.

The dedicated Arch, Debian, Fedora, Ubuntu, and Docker badge links look good now, and removing the generated `#__tabbed_1_n` fragments was the right change.

However, the platform-specific links in the “Getting Started” section now all point to the generic installation page. The Windows and macOS badges do the same. Since that page opens with the Windows/MSI instructions first, users clicking macOS or Docker do not land directly on the instructions they selected.

Could you please preserve the platform-specific behavior without bringing back the generated tab anchors?

* Docker can link directly to `/docs/guide/install/docker/`.
* Windows and macOS could link to stable direct download or release destinations.
* Alternatively, the separate Windows/Linux/macOS/Docker links could be replaced with one generic “Install guide” link.

The distribution-specific links are otherwise addressed correctly.

duggiefresh:
Updated per comments. 

> Windows and macOS could link to stable direct download or release destinations

Windows links out to its latest direct download, and macOS links out to the release destination. For Linux, it links out to the general installation page, which defaults first to the Windows tab at https://lemonade-server.ai/docs/guide/install/. It is not the most ideal solution without reworking the README installation section, or updating the Lemonade website. 

Thanks for the feedback!

fl0rianr (commented):
Thanks for the update. The Windows, macOS, Docker, and distribution-specific links look good now.

One mismatch remains: the link labeled “Linux” still points to the general installation page, which opens with the Windows/MSI tab by default. As you noted, this means the link does not actually take Linux users directly to Linux instructions.

I do not think the website needs to be reworked for this PR. A minimal README-only solution would be to rename that generic link to something such as “All platforms” or “Install guide” instead of “Linux”. For example:

`[All platforms](https://lemonade-server.ai/docs/guide/install/)`

That would accurately describe the destination while keeping the stable direct links for Windows, macOS, and Docker.

duggiefresh:
Thanks for the review, and taking the time for additional suggestions. 

https://lemonade-server.ai/docs/guide/install/ does not include link to the Snap installation page (https://snapcraft.io/lemonade-server), yet the README does. To include it as part of the Linux installation guide, I could use the README anchor tag and update it to: 

`[Linux and other supported platforms](#supported-platforms)` 

Or would you prefer removing the `Linux` link in favor of `[All platforms](https://lemonade-server.ai/docs/guide/install/)`?

fl0rianr:
Thanks for working through the alternatives. Good points.

After considering the original behavior and the available destinations, I prefer keeping the platform-specific label and linking it to the README’s supported-platforms section:

`[Linux](#supported-platforms)`

This is the closest stable replacement for the previous Linux-specific link:

* it keeps “Linux” as a clear entry point;
* it takes users to the available Linux options, including Snap;
* it avoids the position-dependent `#__tabbed_1_2` anchor;
* and it does not require expanding the scope of this PR to restructure the documentation website.
* The Readme is very visible and therefor solid maintained 

I would therefore go with the README anchor rather than renaming the link to “All platforms”.

Thanks for raising the Snap case — that clarified the trade-off.

duggiefresh:
Great, thanks for the feedback. Updated per conversation :D

fl0rianr (approved):
Thanks again. Approved now.

Since this was "just" a small fix that become so complicated I fixed the merge conflict you missed to touch for you.

PR #2828, opened 2026-07-27
Add ModelScope registry search blog post

ramkrishna2910:
## Summary
- New blog post at `docs/news/modelscope-search.html` announcing ModelScope as a second model registry in v11.5, following the layout/style of the existing news pages (hero, section cards, copyable code blocks).
- Screenshots hosted in `lemonade-sdk/assets` (`modelscope/`) per the existing convention and referenced via `raw.githubusercontent.com`.
- News card added to `docs/assets/news-data.js` (`type: blog`), so it appears on the homepage and News page.

Intended to go live with this week's release (the `v*` tag triggers the Publish Website workflow).


fl0rianr (commented):
Thanks for putting this together — the page looks polished and fits the existing news layout well.

I found two claims that should be softened before merge:

1. “What you see is what will run” is too strong. The Model Manager verifies that a repository exposes downloadable GGUF variants, but it does not prove that the model architecture is supported by the bundled backend or that it will load successfully on the user’s hardware. I’d suggest wording this as “only shows repositories with downloadable GGUF variants,” without promising runtime compatibility.

2. The `/v1/registry/search` example is described as “the same search that powers the Model Manager,” but the app additionally calls `/pull/variants` for every candidate and only displays repositories that pass that validation. The example should probably include `format=gguf`, and the text should clarify that the endpoint returns candidates while `/pull/variants` performs the authoritative file-level validation.

Also, the linked API documentation does not currently document `/v1/registry/search`, despite the article pointing readers there for the full response shape. Either the API docs should be added in this PR or that sentence should be removed for now.

Minor: “Q4_K_M is preselected” should be “Q4_K_M is preselected when available,” since the UI falls back to the first available variant otherwise.

But it's "just" wording do what you think is best.

ramkrishna2910:
Thanks for the careful read @fl0rianr — all four points addressed in b83655f:

1. Dropped the runtime-compatibility promise: now reads "only shows repositories with downloadable GGUF variants."
2. Reworded the API section to say `/v1/registry/search` returns candidates and the Model Manager validates each with `/v1/pull/variants` before offering a download; the curl example now includes `format=gguf`.
3. Added a full `GET /v1/registry/search` section to `docs/api/lemonade.md` (parameters, response shape, errors), including the note that `has_gguf` is a metadata hint and `/pull/variants` is the compatibility authority — so the article's docs link now resolves to real content.
4. "Q4_K_M is preselected when available."

fl0rianr (approved):

PR #2829, opened 2026-07-27
ci: make C++ test selection merge-friendly

fl0rianr:
CTest coverage is without a doubt a valuable addition: the tests are fast, inexpensive to run, and validate the C++ source directly.

Recently, many PRs have added new C++ tests. However, every new test also had to be added manually to the long build-target lists and CTest regular expressions in `cpp_server_build_test_release.yml`.

When one test-related change was merged while another PR with a new test was still open, the workflow file frequently produced a merge conflict. These conflicts were usually easy to resolve, but they created unnecessary friction and made otherwise independent changes overlap.

## What changed

This PR moves CI membership closer to each CMake test definition.

Tests selected for the packaging CI now:

* keep their existing `add_test(...)` registration;
* receive the `cpp-ci` CTest label;
* add their executable target as a dependency of the shared `cpp-ci-tests` build target.

The Linux and macOS workflow steps now only need to:

1. build `cpp-ci-tests`;
2. run tests carrying the `cpp-ci` label.

Platform-specific behavior remains automatic. For example, `ProcessManagerTest` is defined only inside its existing Linux condition, so it is included on Linux but not on macOS.

## Why this is better

* New tests no longer require edits to the shared workflow test lists.
* Independent test PRs are less likely to conflict.
* The list of tests remains visible beside the corresponding CMake test definitions.
* Linux-only and macOS-only tests naturally follow their existing CMake conditions.
* The same test group can be listed and run locally with standard CTest commands.

## Scope

This change does not add or remove test coverage. It preserves the test selection currently used by the Linux and macOS packaging jobs while making that selection easier to maintain.


bitgamma (approved):
I like this approach. Would be great if there was a way (maybe a macro definition?) that can turn the set_test_properites/add_dependencies block into a one liner. But this can be addressed later

PR #2833, opened 2026-07-28
[llamacpp] replace deprecated flags, better IO flag handling

bitgamma:
replace some deprecated llamacpp flags with their new version. It also handles better the IO flag overrides so that the defaults do not accidentally override whatever the user sets.

It also allows the user control the mmproj offload (when using GPU) since this is not in actual conflict with what we set.

finally, a couple of unused constants (they are defined elsewhere) have been removed


fl0rianr (commented):
Thanks for cleaning up the deprecated llama.cpp flags and improving the override behavior — the overall direction looks good.

One concern: the new alias check uses `custom_args.find(alias)`, so short aliases such as `-lm`, `-dio`, or `-ndio` can also match inside an argument value or file path rather than only as actual flags.

For example, a path like:

```text
--lora /models/alma-lm-adapter.gguf
```

contains `-lm`, which would cause the function to skip adding the default `--load-mode none` on an iGPU system.

Could we parse `custom_args` with the existing `parse_custom_args()` helper and compare complete flag tokens instead? It would also be good to handle the `--flag=value` form by comparing only the part before `=`.

A small regression test covering both a real alias token and an alias occurring only inside a value/path would help prevent this.

fl0rianr (approved):
Thanks, LGTM.

Small nit to have it known for later:
custom_args_has_flag() is tested isolated, but not the crucial push_overridable_arg(... aliases)-decision. Let's say we miss an alias later in the list e.g. -ndio, the regression test would not get that.

PR #2267, opened 2026-06-16
Feat: Apply GUI3 (beta - draft)

fl0rianr:
## Summary

This PR integrates GUI3 into the main Lemonade tree so we can produce pinned beta artifacts from CI before doing the final GUI cutover.

This is intentionally staged as a beta-readiness PR first. The goal is to make the branch buildable, testable, and artifact-producing, then use the resulting beta build for validation and feedback triage before final merge.

Refs #2266.

## What changed

* Imported GUI3 from the UI testing branch into `src/app`.
* Preserved the existing Tauri host under `src/app/src-tauri`.
* Updated the web-app build integration to use the new GUI3 app entry points.
* Removed the old frontend `toolDefinitions.json` coupling from the CMake/server resource path.
* Updated package lock
* Added GUI Playwright coverage to the existing release/build workflow.
* Added GUI test artifact upload paths for Playwright reports, screenshots, and diagnostics.
* Kept GUI3 sync history separate from unrelated source-branch files.

## Beta artifact goal

This PR should allow CI to produce beta artifacts for GUI3 validation.

The first target is a pinned GUI3 beta build, for example:

```text
11.0.0-beta.1
```

The beta build should be used to validate:

* startup
* model loading
* chat
* backend selection
* logs
* settings
* old-GUI core scenario parity
* demo/release-ready polish for the main chat and model workflows

## Branch sync discipline

This branch must remain controlled.

Allowed updates:

* rebase onto `main`
* filtered GUI sync from `origin/kpoin/ui-testing`
* minimal build, CI, or integration and CI fixes required for GUI3 beta artifacts

Not allowed:

* direct merge of `origin/kpoin/ui-testing`
* direct merge of unrelated work
* importing `.squad` content
* importing `prototype/ui-redesign/` paths into the final branch
* deleting or replacing `src/app/src-tauri`

The GUI source mapping is:

```text
prototype/ui-redesign/ -> src/app/
```

This keeps GUI3 history while preventing unrelated files from the source branch from entering the PR.

## Validation

Before beta artifacts are considered usable, the following should pass:

* [ ] CI release/build workflow completes for this branch.
* [ ] GUI Playwright tests pass against a running Lemonade server.
* [ ] Playwright artifacts are available for review.
* [ ] GUI screenshots / reports are uploaded as CI artifacts.
* [ ] Server logs are uploaded when GUI tests fail.
* [ ] Linux package artifact can start `lemond`.
* [ ] Tauri app host is preserved.
[...]

PR #2834, opened 2026-07-28
add support for kvcache reuse

lishixlnx:
main improvement is:
add new API to reset server's kvcache status by calling og->rewindTo(0)
new API is: POST /api/v1/chat/sessions/reset

Otherwise,  will always use multi-chat function in OGA.

Currently, if you click the "Start a new chat" button in right-up corner in web UI,
the oga will reset the kvcache status to 0.

### DIFF
--- a/src/app/src/renderer/components/panels/LLMChatPanel.tsx
+++ b/src/app/src/renderer/components/panels/LLMChatPanel.tsx
@@ -497,6 +497,29 @@ const LLMChatPanel: React.FC<LLMChatPanelProps> = ({
     ...buildChatRequestOverrides(appSettings),
   });
 
+  const handleStartNewChat = async () => {
+    const body: Record<string, string> = { conversation_id: 'default' };
+    if (chatModelName) {
+      body.model = chatModelName;
+    }
+
+    try {
+      const response = await serverFetch('/chat/sessions/reset', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(body),
+      });
+      if (!response.ok) {
+        const detail = await response.text();
+        console.warn('Chat session reset failed:', response.status, detail);
+      }
+    } catch (error) {
+      console.warn('Chat session reset failed:', error);
+    }
+
+    onNewChat?.();
+  };
+
   /** Build an error message enriched with backend action help text when available. */
   const buildErrorMessage = (error: any): string => {
     const errorMessage = error.message || 'Failed to get response from the model.';
@@ -1263,7 +1286,7 @@ const LLMChatPanel: React.FC<LLMChatPanelProps> = ({
           )}
           <button
             className="new-chat-button"
-            onClick={onNewChat}
+            onClick={handleStartNewChat}
             disabled={isBusy}
             title="Start a new chat"
           >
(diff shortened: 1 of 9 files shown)


fl0rianr (changes requested):
Thanks for working on this. KV-cache reuse across turns is an important improvement for multi-turn OGA usage, and Lemonade previously did not expose an explicit way to reset the reused server-side chat state when starting a new conversation.

The overall direction looks good, but I think one functional issue should be addressed before merging:

### Reset the actually loaded model

LLMChatPanel currently sends chatModelName, which is derived from selectedModel. This may differ from currentLoadedModel.

A possible sequence is:

* Model A is currently loaded and still holds the previous KV cache.
* The user selects Model B.
* The user clicks “Start a new chat” before B is loaded.
* The reset request targets B.
* The router does not find B and returns success with "No loaded model to reset".
* The visible chat is cleared, but Model A keeps its old KV cache.

The frontend should preferably target the actually loaded model:

```ts
const resetModelName = currentLoadedModel ?? chatModelName;
```

Also, if an explicit model was requested but cannot be found, the router should not silently report a successful reset.

### Do not block the local New Chat action indefinitely

Currently, `onNewChat?.()` is only called after awaiting the reset request. If the server is unavailable or the request hangs, the visible conversation cannot be cleared.

The UI should either clear the local chat first and perform the server reset separately, or the reset request should have a defined timeout. A reset failure should still be logged or surfaced, since the backend may otherwise retain context from the previous conversation.

### Additional notes

* Backend tests for a loaded AMDGPU/OGA model, an unknown explicitly requested model, no loaded model, and a backend reset failure would be valuable.
* The frontend currently only checks `response.ok`, so a response with `"status": "unsupported"` is effectively treated as a successful reset.
* The browser cache-control changes appear unrelated to KV-cache reuse and would be easier to review in a separate commit or PR.

fl0rianr:
A reminder on myself to ensure to integrate this into GUI3 #2267 if this one merges...

lishixlnx:
> Thanks for working on this. KV-cache reuse across turns is an important improvement for multi-turn OGA usage, and Lemonade previously did not expose an explicit way to reset the reused server-side chat state when starting a new conversation.
> 

@fl0rianr   I read your comments and agree with your opinion.
The multi-turn feature is totally new and has not been supported completely by OGA.  
For example, 
1 if 2nd chat still has picture, the kvcache reuse feature can't be used; 
2 if 1st chat has picture, 2nd chat has no pic,  the kvcache reuse feature works, but the "reset" API --> OGA's rewindTo(0) API doesn't work well, it's OGA's bug;
3 it only support text message well now.

So I think the UI has to be redesigned to fit for all cases, and also consider the "change model at each turn" case which should be disabled in multi-turn case.
That is beyond this PR's field.  
This PR only shows that it has this feature and can support the function.

You can modify the PR based on current code.

fl0rianr:
@lishixlnx Thanks for the update, makes sense. Still quite useful, how should we move on here? To modify this PR I would need maintainers access. I can make a new PR and apply you as a co-author and we have to close this one then. Let me know what do you prefer.

lishixlnx:
> @lishixlnx Thanks for the update, makes sense. Still quite useful, how should we move on here? To modify this PR I would need maintainers access. I can make a new PR and apply you as a co-author and we have to close this one then. Let me know what do you prefer.

Please make new PR and close this one.

PR #2837, opened 2026-07-28
ci: improve persistent Windows runner hygiene and process lock recovery

abn:
This PR improves persistent self-hosted Windows runner hygiene by isolating transient state and ensuring robust cleanup of process locks. This prevents `MAX_PATH` errors, workspace contamination, and checkout failures from lingering background daemons.

### Key Changes:
* **Pre-Checkout process termination**: Terminate lingering server/backend processes (`lemond`, `llama-server`, etc.) before `actions/checkout` runs to release active file locks on workspace files (without terminating Python).
* **Setup path relocation**: Directs test installations, caches, and temp directories to isolated directories under `C:\lemonade-ci` outside the git workspace.
* **Teardown resiliency**: Uses composite actions (`setup-windows-ci-env` / `teardown-windows-ci-env`) with `if: always()` on individual steps to ensure that processes are killed and workspaces are cleaned up even if previous steps fail.

### DIFF
--- a/.github/actions/cleanup-processes-windows/action.yml
+++ b/.github/actions/cleanup-processes-windows/action.yml
@@ -1,5 +1,5 @@
 name: 'Cleanup Processes (Windows)'
-description: 'Kill all lemonade, Python, llama-server, and FLM processes on Windows'
+description: 'Kill all lemonade, llama-server, and FLM processes on Windows'
 runs:
   using: "composite"
   steps:
@@ -108,9 +108,6 @@ runs:
             # Kill lemonade processes (lemond, lemonade, lemonade-tray, etc.)
             Stop-ProcessByPattern -Pattern "lemonade" -Description "lemonade"
             Stop-ProcessByPattern -Pattern "lemond" -Description "lemond"
-
-            # Kill Python processes
-            Stop-ProcessByPattern -Pattern "python" -Description "Python"
 
             # Kill wscript processes (VBScript launcher used by lemonade)
             Stop-ProcessByPattern -Pattern "wscript" -Description "wscript"
--- a/.github/actions/setup-windows-ci-env/action.yml
+++ b/.github/actions/setup-windows-ci-env/action.yml
@@ -1,0 +1,63 @@
+name: 'Setup Windows CI Environment'
+description: 'Set cache/install paths, cleanup legacy workspace paths, prune transient directories, and terminate stale processes.'
+inputs:
+  matrix-name:
+    description: 'Name of the current matrix leg or job to isolate caches/temp folders.'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Set Windows CI cache and install paths
+      shell: PowerShell
+      run: |
+        $ErrorActionPreference = "Stop"
+        $normalizedMatrix = "${{ inputs.matrix-name }}" -replace '[^A-Za-z0-9_.-]', '-'
+        $runKey = "${{ github.run_id }}-${{ github.run_attempt }}-$normalizedMatrix"
+
+        $ciRoot = "C:\lemonade-ci"
+        $cacheRoot = Join-Path $ciRoot "cache"
+        $installRoot = Join-Path $ciRoot "install"
+        $tmpRoot = Join-Path $ciRoot "tmp"
+
+        $hfHome = Join-Path $ciRoot "hf"
+        $cacheDir = Join-Path $cacheRoot $normalizedMatrix
+        $installPath = Join-Path $installRoot $runKey
+        $tmpDir = Join-Path $tmpRoot $runKey
+
+        foreach ($path in @($ciRoot, $hfHome, $cacheDir, $installPath, $tmpDir)) {
+          New-Item -ItemType Directory -Force -Path $path | Out-Null
+        }
+
+        # Set environment variables for the current job context
+        echo "HF_HOME=$hfHome" >> $Env:GITHUB_ENV
+        echo "LEMONADE_CACHE_DIR=$cacheDir" >> $Env:GITHUB_ENV
+        echo "LEMONADE_INSTALL_PATH=$installPath" >> $Env:GITHUB_ENV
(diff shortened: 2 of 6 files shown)


fl0rianr (changes requested):
Thanks for working on this. Extracting the repeated Windows setup and teardown logic into composite actions is a good direction, but I am not yet convinced that the current patch reliably addresses the reported flake.

The main concern is that `test-cli-endpoints` and `test-cli-apikey` run on `windows-latest`, so they use fresh GitHub-hosted runners. Persistent workspace contents, stale processes, or install directories from previous runs should not normally survive there. Also, the new setup action runs after `actions/checkout`, so it cannot fix stale workspace state that affected checkout or caused obsolete test code to be used. Could you link a failing run or relevant log lines that demonstrate the stale-state mechanism?

There are also two implementation details that should be adjusted:

1. In `setup-windows-ci-env`, path cleanup and pruning happen before process cleanup. If a stale process is holding a file or directory open, deletion may fail, after which the process is killed but the deletion is not retried. Process cleanup should run first, followed by directory cleanup.

2. `if: always()` on the workflow step only guarantees that the teardown composite action is started. The individual steps inside `teardown-windows-ci-env` still use the default success condition. If process cleanup fails, the remaining directory cleanup and pruning steps may be skipped. Each independent teardown step should also use `if: ${{ always() }}` or otherwise be explicitly best-effort.

I would also consider limiting the broad process cleanup on GitHub-hosted runners. The current action kills all Python processes and performs MSI/NSIS cleanup, which makes sense on persistent dedicated runners but adds unnecessary surface on a fresh hosted runner unless the failure logs show that it is needed.

The DRY refactor itself looks useful, but I think the root cause should be demonstrated and the cleanup ordering/failure handling fixed before merging.

abn:
Thanks for the review. You are completely right about run `30361219442` failing on a `/metrics` crash. The traceback I referenced showing the `server_runner.py` call was from a different run on a persistent self-hosted runner.

To address your feedback, I have updated the branch with the following changes:
* **Broad Python Termination**: Removed `"python"` from the process kill patterns inside `.github/actions/cleanup-processes-windows/action.yml` and all inline scripts to protect unrelated python environments.
* **Setup Step Ordering**: Swapped the steps in `setup-windows-ci-env/action.yml` so process cleanup happens before path cleanups and pruning.
* **Teardown Resilience**: Added `if: always()` to every step inside `teardown-windows-ci-env/action.yml`.
* **Shared Root Protection**: Removed deletion of `C:\lemonade-ci\install` and `C:\lemonade-ci\tmp` from the pre-checkout steps to prevent concurrent job interference. Workspaces are now isolated under unique `runKey` subdirectories and pruned incrementally.
* **Test Embeddable**: Fixed the missing action issue on `test-embeddable-windows` by using inline PowerShell process termination since it does not run checkout.

Let me know if this looks good.

fl0rianr (commented):
Thanks for the update. The pre-checkout cleanup and the narrower inline process list move the implementation in the right direction, but after checking the referenced failing run I still do not think the stated root cause is supported.

Run `30361219442` explicitly identifies itself as a GitHub-hosted compute agent using the `windows-2025-vs2026` image. During checkout, the action reports that it deletes the workspace contents, initializes an empty repository, fetches the expected merge ref, and checks it out successfully.

More importantly, the first failure in that run is not execution of a stale `test/utils/server_runner.py`. The endpoint tests pass until `test_002a_metrics_endpoint`, where the current `test/server_endpoints.py` receives `WinError 10054` because the server closes the connection. The server is then unavailable for all subsequent tests. I could not find the claimed `server_runner.py` traceback in the referenced job log.

Therefore, the supplied evidence points to a server crash or termination during the `/metrics` request, not stale test code surviving checkout. The new workspace cleanup is unlikely to fix that failure.

There are also still several implementation issues:

1. The claim that broad Python termination was removed is not correct. Both the setup and teardown composite actions still invoke `cleanup-processes-windows`, and that action still executes:

   `Stop-ProcessByPattern -Pattern "python" -Description "Python"`

   Therefore all Python processes are still terminated after checkout and during teardown.

2. `setup-windows-ci-env` still performs path cleanup and pruning before invoking process cleanup. The ordering concern from the previous review remains inside the reusable action.

3. The individual steps inside `teardown-windows-ci-env` still do not use `if: always()`. The caller starting the composite action with `if: always()` does not guarantee that later composite steps run after an earlier step fails.

4. The new pre-checkout code deletes the complete shared `C:\lemonade-ci\install` and `C:\lemonade-ci\tmp` roots. On a persistent machine with multiple runner agents or concurrent jobs, this can delete another active job's run-specific directories. Cleanup should target stale or known-owned directories rather than deleting the shared roots unconditionally.

5. The same pre-checkout cleanup has now been added to several `windows-latest` validation jobs, although the referenced failure occurred on a fresh GitHub-hosted runner and does not demonstrate persistent state. This increases the scope of the PR without addressing the observed `/metrics` crash.

I suggest either:

* investigate and fix the actual server termination during the `/metrics` test, or
* reframe this PR specifically as persistent self-hosted runner hygiene, restrict the cleanup to those runners, and provide a failure log from such a runner that demonstrates stale workspace state.

Before approval, the broad Python termination, cleanup ordering, teardown resilience, and unconditional deletion of shared roots should also be corrected.

abn:
Thanks for the review. You are completely right about run `30361219442` failing on a `/metrics` crash. The traceback I referenced showing the `server_runner.py` call was from a different run on a persistent self-hosted runner. I think it is best to reframe this PR and then tackle the root cause fresh again.

fl0rianr (approved):
One non-blocking observation: the pre-checkout process termination is still added to several windows-latest jobs. Since these use fresh GitHub-hosted runners, the cleanup should normally be unnecessary there. It may be worth restricting this part to persistent self-hosted jobs in a follow-up, or documenting why it is deliberately retained for hosted runners.

abn:
Thanks for the approval! I kept the pre-checkout cleanups on windows-latest as a safeguard in case the org maps those jobs to persistent custom VMs in the future (or under customized runner pools). I will look into restricting or documenting this in a follow-up pass.

PR #2838, opened 2026-07-28
feat: bump OpenMoss engine to 0.2.0 / GUI side

pwilkin:
This brings a round of needed fixes to OpenMOSS:
* the cascade is now done internally and MOSS-VoiceGen stops being a standalone model, as it was confusing anyway
* now adds the Local version of MOSS in addition to the Delay one
* adds OpenMOSS-SoundEffect, which I somehow missed before, as an SFX model


fl0rianr (changes requested):
Thanks for the update! The overall direction makes sense, especially moving the VoiceGen cascade behind the OpenMOSS backend and exposing SoundEffect as a regular audio-generation model.

I found a few functional issues that should be looked into before merging:

1. **The transient VoiceGen process can leave requests hanging for a very long time**

`design_reference_sample()` polls `/health` up to 3000 times, but never checks whether the spawned process has already exited. If VoiceGen fails during model loading, for example because both models do not fit in VRAM, Lemonade may continue polling for many minutes instead of returning the actual startup error.

Please reuse the normal backend startup behaviour here: check `ProcessManager::is_running()`, reap an exited process, and include its exit code/output in the error.

2. **The new cascade requires TTS and VoiceGen to be resident simultaneously**

The main TTS process remains loaded while a second fully GPU-offloaded VoiceGen process is started. Previously the GUI loaded VoiceGen and TTS sequentially, so cards that can run each model individually may now fail because both models plus their context and compute buffers must fit at once.

This likely needs either a memory-aware sequential fallback, explicit placement of VoiceGen on another device/CPU, or another mechanism that avoids requiring both large models in GPU memory simultaneously.

3. **Any non-empty `voice` value now triggers voice design**

`apply_voice_design()` interprets the standard `/v1/audio/speech` `voice` field as a voice-description prompt, generates a reference WAV, removes `voice`, and forwards the request.

That changes the existing OpenMOSS/OpenAI-compatible semantics. For example, a normal client sending `"voice": "default"` would unexpectedly launch VoiceGen and attempt to design a voice called “default”.

Voice design should be enabled through an explicit extension such as `voice_design_description` or `voice_mode: "design"`, while regular `voice` values continue to be forwarded as instructions.

4. **The new `audio_defaults` are not used by GUI3**

`MOSS-SoundEffect` declares 100 steps and CFG 4.0, but the GUI still initializes generic audio generation with 50 steps and CFG 4.5 and sends those values explicitly. As a result, the model defaults are never actually used.

It would be good to initialize the controls from `audio_defaults`, similar to how `speech_defaults` are handled.

5. **`seed: -1` does not mean random for MOSS-SoundEffect**

The OpenMOSS SoundEffect seed is an unsigned value and explicitly has no random sentinel. GUI3 currently sends `-1` by default, which may either fail conversion or become a fixed large unsigned seed.

For OpenMOSS SFX, please omit the seed when the GUI value is `-1`, or generate an actual random valid seed.

6. **The clone-mode “Style note” is currently ignored**

The GUI exposes an optional style note in clone mode, but the request path sets `voice = ''` before sending the cloned reference audio. Either forward the style note as `voice`, or remove the field from the UI.

The main blockers for me are the child-process lifecycle, simultaneous GPU residency, and changing the meaning of the standard `voice` field.

pwilkin:
@fl0rianr could you check now? The only error is the macOS CI that never works.

fl0rianr:
Can we merge @kpoineal PR #2846  first since it's crucial we move forward here - then you rebase and I'll re-review? The CI errors is not relevant, you are right there.

pwilkin:
Yeah.

fl0rianr (changes requested):
Please solve the merge conflict  and separate the server and the GUI changes, only GUI changes here.
Then I'm going to review, thank!

fl0rianr (changes requested):
Thanks, here is my review.

**Blocking:**
I think this reintroduces backend-specific classification into GUI3. The composer already gets the deployment mode from the server (tts, audio-generation), so it shouldn't need to identify openmoss by recipe or model name on top of that.
We are working actually on pulling server owned fully inside.

With the server-owned model concept, adding or updating a backend should not require a GUI change just to teach the client about that backend. Right now isOpenMossSfx, isOpenMossTts, and the hard-coded OPENMOSS_SPEECH_PARAMS mean OpenMOSS features are still owned partly by the GUI.

Could the server expose the supported generation features/options declaratively alongside the defaults instead? Then MOSS-SoundEffect stays a normal audio-generation model and OpenMOSS-TTS stays a normal tts model, while the composer renders whatever controls the server advertises. That would also let future backends expose the same features without adding another backend-specific branch here.

**Also:**

Do we need a backend-request-specific selection lock at all here?

App.tsx now already keeps a selected downloaded model selected even when it temporarily disappears from all_models_loaded. I think that is the better invariant for the server-owned model lifecycle: model selection should not depend on temporary runtime residency.

ChatView still has a second fallback effect that switches away when the selected model is no longer loaded, and the new lock seems to compensate for that behavior. Could we make that selection policy consistent instead and remove the global lock? Backend-internal unload/reload sequences should not necessarily be visible to the GUI.

pwilkin:
Okay, now relies on #2863 as the server-side part. @fl0rianr would be grateful for a re-review.

fl0rianr (approved):
Thanks, LGTM

Since this PR depends on #2863 for that contract and the backend-side behavior, I’d just make sure #2863 lands first.

From the GUI/server-owned side, this looks good to me now.

I'll wait a bit with merging due to the dependency.

PR #2839, opened 2026-07-28
fix(llamacpp): validate device-backend compatibility during model load

abn:
Validates that explicit `llamacpp_device` targets match the resolved `llamacpp_backend` before starting `llama-server`.

### Why
When a request or recipe specifies a device identifier that contradicts the resolved backend (for example, passing `llamacpp_device = ROCm2` when `llamacpp_backend` resolved to `vulkan` via auto-selection), `llama-server` fails during CLI argument parsing with `invalid device: ROCm2`. Because this failure occurs asynchronously inside the child process, it surfaces to clients as an uninformative error (`llama-server failed to start`) without explaining the backend/device mismatch.

### Changes
- Adds `BackendUtils::validate_device_backend_match` in `LlamaCppServer::load()`. Throws `std::invalid_argument` before process launch if a device prefix (`rocm`, `cuda`, `vulkan`) contradicts the resolved backend, providing an explicit error message (`Device selection 'ROCm2' contradicts backend choice 'vulkan'`).
- Bypasses prefix validation when `backend` is `"system"` or `"auto"` (or empty) so system binaries and PATH-resolved backends pass through cleanly.
- Standardizes `LLAMA_ARG_DEVICE` and `CUDA_VISIBLE_DEVICES` lookups using cross-platform UTF-8 environment accessors (`utils::get_environment_variable_utf8`).
- Adds C++ unit test suite `test_backend_utils_gpu_env.cpp` verifying CUDA host override handling, system/auto backend passthrough, and mismatch validation exception throwing without physical GPU hardware dependencies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2593,6 +2593,13 @@ if(EXISTS "${_ROCM_ROOT_TEST_SRC}")
     add_test(NAME RocmRootResolutionTest COMMAND test_rocm_root_resolution)
 endif()
 
+set(_GPU_ENV_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_backend_utils_gpu_env.cpp")
+if(EXISTS "${_GPU_ENV_TEST_SRC}")
+    add_executable(test_backend_utils_gpu_env test/cpp/test_backend_utils_gpu_env.cpp)
+    target_link_libraries(test_backend_utils_gpu_env PRIVATE lemonade-server-core)
+    add_test(NAME BackendUtilsGpuEnvTest COMMAND test_backend_utils_gpu_env)
+endif()
+
 set(_NETWORK_UTILS_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_network_utils.cpp")
 if(EXISTS "${_NETWORK_UTILS_TEST_SRC}")
     add_executable(test_network_utils test/cpp/test_network_utils.cpp)
--- a/src/cpp/include/lemon/backends/backend_utils.h
+++ b/src/cpp/include/lemon/backends/backend_utils.h
@@ -201,5 +201,14 @@ namespace lemon::backends {
             std::vector<std::pair<std::string, std::string>>& env_vars,
             const std::string& log_tag,
             bool skip_visible_devices = false);
+
+        /**
+         * Validates that the device selection string prefix matches the selected backend.
+         * Throws std::invalid_argument if a device prefix contradicts the backend choice
+         * (e.g. passing target_device "ROCm2" to a "vulkan" or "cuda" backend).
+         */
+        static void validate_device_backend_match(
+            const std::string& backend,
+            const std::string& target_device);
     };
 } // namespace lemon::backends
--- a/src/cpp/server/backends/backend_utils.cpp
+++ b/src/cpp/server/backends/backend_utils.cpp
@@ -1228,8 +1228,8 @@ namespace lemon::backends {
             const std::string& log_tag,
             bool skip_visible_devices) {
         if (!skip_visible_devices) {
-            const char* existing_visible_devices = std::getenv("CUDA_VISIBLE_DEVICES");
-            const bool has_visible_override = existing_visible_devices && existing_visible_devices[0] != '\0';
+            std::string existing_visible_devices = utils::get_environment_variable_utf8("CUDA_VISIBLE_DEVICES");
+            const bool has_visible_override = !existing_visible_devices.empty();
 
             if (has_visible_override) {
                 LOG(INFO, log_tag) << "Respecting existing CUDA_VISIBLE_DEVICES="
@@ -1256,5 +1256,48 @@ namespace lemon::backends {
                                << std::endl;
         }
 #endif
+    }
+
+    void BackendUtils::validate_device_backend_match(
+            const std::string& backend,
+            const std::string& target_device) {
+        if (backend.empty() || backend == "auto" || backend == "system" || target_device.empty()) {
+            return;
+        }
+
+        std::string lower_device = target_device;
+        for (char& c : lower_device) {
+            c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+        }
+
+        std::string lower_backend = backend;
+        for (char& c : lower_backend) {
+            c = static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+        }
+
+        const bool is_rocm_device = (lower_device.rfind("rocm", 0) == 0);
+        const bool is_cuda_device = (lower_device.rfind("cuda", 0) == 0);
+        const bool is_vulkan_device = (lower_device.rfind("vulkan", 0) == 0);
+
+        const bool is_rocm_backend = (lower_backend.rfind("rocm", 0) == 0);
+        const bool is_cuda_backend = (lower_backend.rfind("cuda", 0) == 0);
+        const bool is_vulkan_backend = (lower_backend.rfind("vulkan", 0) == 0);
+
+        if (is_rocm_device && !is_rocm_backend) {
+            throw std::invalid_argument(
+                "Device selection '" + target_device + "' contradicts backend choice '" + backend +
+                "'. Expected a " + backend + " device identifier (e.g. " +
+                (is_vulkan_backend ? "Vulkan0" : is_cuda_backend ? "CUDA0" : "a matching device") + ").");
+        }
+        if (is_cuda_device && !is_cuda_backend) {
+            throw std::invalid_argument(
+                "Device selection '" + target_device + "' contradicts backend choice '" + backend +
+                "'. Expected a " + backend + " device identifier.");
+        }
+        if (is_vulkan_device && !is_vulkan_backend) {
+            throw std::invalid_argument(
+                "Device selection '" + target_device + "' contradicts backend choice '" + backend +
+                "'. Expected a " + backend + " device identifier.");
+        }
     }
 } // namespace lemon::backends
--- a/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
@@ -340,7 +340,8 @@ void LlamaCppServer::load(const std::string& model_name,
     }
     push_arg(args, reserved_flags, "--ctx-size", std::to_string(ctx_size), std::vector<std::string>{"-c"});
 
-    if (llamacpp_device != "") {
+    if (!llamacpp_device.empty()) {
+        BackendUtils::validate_device_backend_match(llamacpp_backend, llamacpp_device);
         push_arg(args, reserved_flags, "--device", llamacpp_device);
     }
     push_reserved(reserved_flags, "--device", std::vector<std::string>{"-dev"});
@@ -503,8 +504,8 @@ void LlamaCppServer::load(const std::string& model_name,
 #endif
 
     if (is_llamacpp_cuda_backend(llamacpp_backend)) {
-        const char* existing_llama_device = std::getenv("LLAMA_ARG_DEVICE");
-        const bool has_llama_device_override = existing_llama_device && existing_llama_device[0] != '\0';
+        std::string existing_llama_device = utils::get_environment_variable_utf8("LLAMA_ARG_DEVICE");
+        const bool has_llama_device_override = !existing_llama_device.empty();
 
         bool skip_visible_devices = false;
         if (!llamacpp_device.empty()) {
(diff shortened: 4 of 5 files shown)


jeremyfowers:

**Attention level:** Elevated — any reviewer can take it, but the to-dos below need resolving before approval (PR description does not match the diff)

### Author's description vs. the diff

The description promises four mechanisms but the diff only implements the first.
- [ ] Remove the ROCR_VISIBLE_DEVICES claim from the PR description or add the ROCR_VISIBLE_DEVICES isolation to the code.
  - Described: ROCr Isolation: low-level ROCm runtime isolation using ROCR_VISIBLE_DEVICES alongside HIP_VISIBLE_DEVICES.
  - In the diff: The diff sets HIP_VISIBLE_DEVICES only; ROCR_VISIBLE_DEVICES does not appear anywhere in the code.
  - Reference: src/cpp/server/backends/llamacpp/llamacpp_server.cpp env_vars.push_back calls only include HIP_VISIBLE_DEVICES, CUDA_VISIBLE_DEVICES, or GGML_VK_VISIBLE_DEVICES — no ROCR_VISIBLE_DEVICES.
- [ ] Remove the auto-backend resolution claim from the PR description or implement the auto-backend resolution logic.
  - Described: Auto-Backend Resolution: handles JIT loads where the backend is unset or "auto" by dynamically resolving to the preferred GPU.
  - In the diff: No code resolves an "auto" backend or handles an unset backend; the device mapping only runs when llamacpp_device is non-empty.
  - Reference: llamacpp_server.cpp line 351 guards on `use_gpu && !llamacpp_device.empty()` with no branch for backend == "auto" or backend resolution.
- [ ] Remove the device-config verification claim from the PR description or implement the verification checks.
  - Described: Device-Config Verification: post-ready precondition checks on child process environments and arguments, failing loud on misconfiguration.
  - In the diff: No post-ready verification or precondition checks exist in the diff; the process is started with the computed env_vars and args without validation.
  - Reference: llamacpp_server.cpp calls ProcessManager::start_process with env_vars and args directly; there is no check of the resulting process environment or argument correctness after startup.

### Scope

**minor** — The change is a contained fix within the existing llamacpp backend — it enhances how the already-documented --device flag and llamacpp_device config are translated for multimodal models, adding a utility header and tests but no new user-facing surface.

### Suggested reviewers

- superm1 (AMD ROCm) — The PR adds ROCm device isolation logic using HIP_VISIBLE_DEVICES and ROCm device prefix parsing.
- bitgamma (CLI) — The llamacpp_device CLI flag and backend device handling are core to the CLI configuration surface.
- sawansri (LLMs) — The llamacpp backend is an LLM inference backend and this change affects model loading behavior.

_Reviewed at head `8da69b3` by repo-manager pr-review. Regenerate with `repo-manager review-pr 2839`._

jeremyfowers (dismissed):
Please address bot concerns and then we'll add @bitgamma as the reviewer.

abn:
Updated the PR description to accurately reflect our refined implementation:

- **Single Visibility Variable (`HIP_VISIBLE_DEVICES`)**: Setting both `HIP_VISIBLE_DEVICES` and `ROCR_VISIBLE_DEVICES` composes rather than reinforces, which causes `0 visible HIP devices` for non-zero devices (e.g. `ROCm2`) and triggers CPU fallbacks. We intentionally set `HIP_VISIBLE_DEVICES` as required by HIP-based llama.cpp ROCm builds.
- **Pre-Launch Validation (`build_device_env`)**: Device parsing, canonicalization (`rocm2` -> `ROCm0`), prefix mismatch checking (e.g. `rocm` backend + `CUDA2` device), digit-gluing prevention, and logical `--device` argument generation have been extracted into a standalone, pure function `build_device_env` in `llamacpp_utils.h` with 21 unit test assertions in `test_llamacpp_utils.cpp`.
- **Removed Late Runtime Verification**: Removed the late post-ready runtime checks that checked local vectors post-launch, eliminating VRAM leak hazards on failed load paths.
- **Scoped Multimodal Override**: `MTMD_BACKEND_DEVICE` override is scoped strictly to pinned device loads to avoid stomping unpinned multi-GPU setups.

bitgamma:
@abn do we have any example/log demonstrating the issue this PR aims to solve? In general, as part of the PR having either an open issue or including directly the reproduction step for the issue being solved helps with review

abn:
@bitgamma sure thing, here is the full reproduction. There is no open issue yet; happy to file one and link it if you prefer that. Happy to provide more information if required or will help. I could also be totally missing something obvious :sweat_smile: 

### Environment

- lemonade-server **11.5.0**, `lemond` running as a systemd user unit
- llama.cpp ROCm backend pinned to **b10143** (`88b47a755`), TheRock ROCm **7.13.0** runtime
- **4× AMD RDNA4 (`gfx1201`)**, 32 GB each → devices `ROCm0..ROCm3`; Fedora 44, kernel 7.1.x
- Subject models: any registry entry whose repo bundles an **active `mmproj-*.gguf`**. Confirmed on `Gemma-4-26B-A4B-it-GGUF` (bundles `mmproj-F16.gguf`) and `Qwen3.6-35B-A3B-GGUF`.

The multi-GPU box matters for **Symptom 1** only. **Symptom 2** reproduces on any single-GPU ROCm host.

### Symptom 1: a device-pinned load of an mmproj-bundled entry hard-fails

```bash
lemonade load Gemma-4-26B-A4B-it-GGUF --llamacpp rocm \
  --llamacpp-device ROCm2 --ctx-size 8192
```

Client:

```
Loading model: Gemma-4-26B-A4B-it-GGUF
Error loading model: Request failed: 500
```

Daemon journal (`llama-server`):

```
[Error] (WrappedServer) error while handling argument "--device": invalid device: ROCm2
[Error] (WrappedServer) This usually means:
[Error] (WrappedServer)   - Missing required drivers or dependencies
[Error] (WrappedServer)   - Incompatible model file
[Error] (WrappedServer)   - Try running the server manually to see the actual error
[Error] (Router) Retry also failed: llama-server failed to start
[Error] (Server) Error 500: POST /api/v1/load
```

Reproduces at **any** ROCm index and any context size.

**Control that isolates it to the mmproj path:** the *text-only siblings in the same repo* (`-it-Q8`, `-it-MTP-GGUF`) pin cleanly with the identical flag. The only delta is the presence of `--mmproj` on the spawned command line.

### Symptom 2: dropping the pin loads "successfully", onto the CPU

Since the pin is rejected, the only remaining path is an unpinned load. It succeeds and silently lands on CPU:

```bash
lemonade load Gemma-4-26B-A4B-it-GGUF --llamacpp rocm --ctx-size 8192
# → 200 OK, model reports loaded
```

| Check | Unpinned mmproj entry | Same weights, GPU |
|---|---|---|
| `rocm-smi --showmeminfo vram`, all 4 cards | **~0.2 GB** resident (no card above baseline) | 20.1 GB |
| decode | **~17–21 tok/s** | ~72–100 tok/s |

Every lever reachable from the CLI was tested and all land on CPU:

| Lever | Result |
|---|---|
| unpinned, no extra args | CPU (~0.2 GB VRAM) |
| unpinned + `--llamacpp-args "-ngl 999"` | **still CPU** |
| `HIP_VISIBLE_DEVICES=2` in the `lemond` unit env + restart, unpinned (verified present in `/proc/<MainPID>/environ`) | **still CPU** |
| `HIP_VISIBLE_DEVICES=2` **and** `-ngl 999` | **still CPU** |
| `--llamacpp-device ROCm2` | rejected, symptom 1 |

There is also no user-side escape hatch, because the projector-offload flags are reserved:

```
[Error] (Router) Argument '--no-mmproj-offload' is managed by Lemonade and cannot be overridden.
[Error] (Router) Reserved arguments: --ctx-size, --device, --embedding, --embeddings, --jinja,
  --metrics, --mmproj, --mmproj-auto, --mmproj-offload, --mmproj-url, --model, --model-draft,
  --no-jinja, --no-mmproj, --no-mmproj-auto, --no-mmproj-offload, --port, ...
```

Net effect: **on this stack there is no way to get a multimodal GGUF onto the GPU through lemonade at all.**.

### Control

Same binary, same ROCm runtime, same GGUF + mmproj, run as a **raw `llama-server` outside lemonade**:

```bash
HIP_VISIBLE_DEVICES=0 \
LD_LIBRARY_PATH=<therock>/gfx1201-7.13.0/lib \
<llama.cpp-b10143>/llama-server \
  -m       gemma-4-26B-A4B-it-UD-Q4_K_M.gguf \
  --mmproj mmproj-F16.gguf \
  -ngl 999 --host 127.0.0.1 --port 8098 -c 8192
```

| Assertion | Measured |
|---|---|
| VRAM residency | idle card 61 MB → **20.1 GB** resident |
| decode | **80.8 tok/s** (prompt eval 117.8 tok/s) |
| vision round-trip | 480×200 PNG reading "CODE 7429" → model answers **"CODE 7429"** |

Repeated on `Qwen2.5-VL-7B-Instruct-Q4_K_M` + f16 mmproj: **7.46 GB resident / 98.9 tok/s / correct image read**.

So `libmtmd` offloads the projector to GPU by default and gfx1201 is fine. The defect is in how lemonade constructs the child process, not in llama.cpp.

### What the code does today

In `src/cpp/server/backends/llamacpp/llamacpp_server.cpp::load()` (current `main`):

- `--device <llamacpp_device>` is appended verbatim when a device is pinned (~L343), with no translation of the physical index into whatever the child process can actually see;
- `--mmproj <path>` is appended unconditionally for mmproj checkpoints (~L357), and `--no-mmproj-offload` is added **only** when `use_gpu == false`;
- visible-device isolation is applied for CUDA only (`apply_cuda_env_vars`, ~L521) — no `HIP_VISIBLE_DEVICES` equivalent on the ROCm path;
- nothing tells the mtmd backend which device to use, so the vision encoder does not follow the text model onto the pinned GPU.

That is the gap this PR closes: map the physical selection to an isolated `HIP_VISIBLE_DEVICES`/`CUDA_VISIBLE_DEVICES`/`GGML_VK_VISIBLE_DEVICES` namespace, rewrite `--device` to the logical index `0` inside that namespace, and set `MTMD_BACKEND_DEVICE` to the same logical device so the projector lands with the weights.

### How to verify the fix

On any ROCm/CUDA host with an mmproj-bundled GGUF:

```bash
lemonade load <mmproj-bundled-entry> --llamacpp rocm --llamacpp-device ROCm2 --ctx-size 8192
rocm-smi --showmeminfo vram      # expect resident VRAM > weight size on that card, not ~0.2 GB
```

Pass = the load returns 200 **and** the pinned card shows resident VRAM above the weight size, with decode in the GPU band rather than the ~17–21 tok/s CPU band. Before the change, step 1 returns 500 (`invalid device`) and every unpinned fallback fails the residency check.

bitgamma:
@abn 

> Symptom 2 reproduces on any single-GPU ROCm host.

are you sure about this? I cannot reproduce this, loading the same model with mmproj on a Strix Point without device pinning doesn't show that issue on my machine, the images are clearly processed by the GPU. Also `--mmproj-offload` is also default, according to llamacpp configuration.

also, the "Symptom 1" seems strange, are you running in a container or any specific environment?

sorry for digging into this but since this PR is bringing some complexity I want to understand what exactly is going on before merging it

abn:
> sorry for digging into this but since this PR is bringing some complexity I want to understand what exactly is going on before merging it

@bitgamma you were right to dig. I have been churning a lot of configs the last week. I went back to the journal and my diagnosis was wrong. Apologies for the noise.

The failing loads were resolving to the **Vulkan** backend, not ROCm:

```
[Info] (Server) Ensuring model loaded: <entry> ctx_size=8192, llamacpp_device=ROCm2, ...
[Info] (LlamaCpp) Using LlamaCpp Backend: vulkan
[Error] (Process) error while handling argument "--device": invalid device: ROCm2
```

`invalid device: ROCm2` is a Vulkan `llama-server` correctly rejecting a ROCm device name. No mmproj involvement.

The root cause is on my side: the registry entry I tested has `recipe_options: {}`, so it inherits `llamacpp.backend = auto` and resolves to Vulkan on my box, while I was passing a ROCm device name. My control was also incorrect given this.

So the multimodal GPU placement is not broken, and Symptom 2 was measured entirely on the Vulkan path. My bad.

What I think is still worth having from this PR is much smaller: the mismatch check in `build_device_env`. `build_device_env("vulkan", "ROCm2")` throws `Device prefix 'ROCm' contradicts backend choice 'vulkan'`, which is exactly the message that would have saved me several days versus an opaque `invalid device` 500 surfaced as a generic HTTP 500.

I'll push that change up and update the PR, but I am happy to also just close this if you prefer that. The device isolation and `MTMD_BACKEND_DEVICE` parts no longer have a demonstrated bug behind them, so I'd rather not ask you to carry that complexity.

Edit: Pushed, rescoped and updated the PR description. I also removed the ROCm isolation helper and `MTMD_BACKEND_DEVICE` entirely. Thanks for pushing back on this. Let me know if you rather prefer to drop the whole thing. But I think the rescoped version does add value.

bitgamma (approved):

PR #2840, opened 2026-07-28
fix(custom_args): preserve repeatable flags

### LINKED ISSUE #2831: merge_args (default true) collapses repeatable llama.cpp flags:
### Platform

Linux/Ubuntu

### Lemonade Version

11.5.0

### GPU / APU Model

AMD Strix Halo

### Component

lemond (server)

### Bug Description

With `merge_args` at its default of true, an args string is tokenized into a map keyed by flag name before being handed to the engine. `build_custom_args_map` starts a fresh entry every time it sees a flag (`src/cpp/include/lemon/utils/custom_args.h`):

```
if (!token.empty() && token[0] == '-' && !is_negative_number(token)) {
    // This is a flag; start a new entry
    result[token] = {};
    last_flag = token;
} else if (!last_flag.empty()) {
    result[last_flag].push_back(token);
}
```

_generally_ this is fine, but llama.cpp has flags that can be repeated or can accumulate. For instance `--override-kv a=bool:false --override-kv b=bool:false ` should be fine, but only `b=bool:false` makes it to the engine.

### Steps to Reproduce

1. Load any llamacpp model with a repeated accumulating flag, without merge_args:
```
curl -s -X POST "http://localhost:13305/api/v1/load" \
  -d '{"model_name": "Qwen3.6-27B-MTP-GGUF",
       "ctx_size": 8192,
       "llamacpp_backend": "vulkan",
       "llamacpp_args": "-ngl 99 -fa on --override-kv tokenizer.ggml.add_bos_token=bool:false --override-kv tokenizer.ggml.add_eos_token=bool:false"}'
```
2. Check the llama.cpp args
```
cat /proc/$(pgrep -f llama-server | head -1)/cmdline | tr '\0' ' '
```

Expected: both `--override-kv` occurrences reach llama-server
Actual: one `--override-kv` reaches the engine, only the last value

### Log Output

```shell

```

### Additional Context

_No response_

sujikathir:
## Summary

- Preserve every occurrence of repeatable custom arguments such as `--override-kv`.
- Keep the existing higher-priority source and binary-negation merge behavior.

Fixes #2831

## Testing

- `ctest --test-dir build -R '^CustomArgsTest$' --output-on-failure`
- Standalone AppleClang C++17 build with `-Wall -Wextra -Werror`

## Screenshots

Not applicable; this changes native argument parsing and merge behavior.

### DIFF
--- a/src/cpp/include/lemon/utils/custom_args.h
+++ b/src/cpp/include/lemon/utils/custom_args.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <algorithm>
 #include <map>
 #include <set>
 #include <string>
@@ -45,8 +46,10 @@ inline std::vector<std::string> parse_custom_args(const std::string& custom_args
     return result;
 }
 
-inline std::map<std::string, std::vector<std::string>> build_custom_args_map(const std::vector<std::string>& tokens) {
-    std::map<std::string, std::vector<std::string>> result;
+using CustomArgsMap = std::map<std::string, std::vector<std::vector<std::string>>>;
+
+inline CustomArgsMap build_custom_args_map(const std::vector<std::string>& tokens) {
+    CustomArgsMap result;
     std::string last_flag;  // Track the most recently seen flag independently of map ordering
 
     // Detect a complete negative number so it's treated as a value, not a flag.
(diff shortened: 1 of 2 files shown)


jeremyfowers:

### Scope

**minor** — A contained bug fix to an internal utility header that resolves an existing issue (#2831) without introducing a new user-facing surface.

### Suggested reviewers

- bitgamma (CLI) — CLI maintainer; the custom_args module is the argument-parsing utility that feeds into CLI flag handling.
- jeremyfowers (CLI) — CLI maintainer with familiarity in the argument parsing and merge pipeline.

_Reviewed at head `596272a` by repo-manager pr-review. Regenerate with `repo-manager review-pr 2840`._

bitgamma (approved):
lgtm! thanks for your contribution

PR #2841, opened 2026-07-29
feat(flm): resolve FLM model source from default_model_source

wariobot09:
## Summary

Let the FLM backend download from ModelScope instead of HuggingFace, driven by the server-wide `default_model_source` (and explicit per-model `source`/`registry_source`) introduced in #3036 — rather than a separate `flm.modelscope` recipe option. FLM is a self-managed backend (`flm pull`) and is deliberately excluded from `apply_default_pull_source`, so it now consults `ModelManager::effective_model_source()` at download time and passes `--modelscope 1` to the `flm` CLI.

This supersedes the original approach in this PR (a standalone `flm.modelscope` recipe toggle), which would have been a second, backend-specific knob for a decision #3036 already centralized. Requires FLM ≥ v0.9.46 for the `--modelscope` flag.

## Scope

Effective-source resolution: an explicit per-model ModelScope provenance wins; otherwise the server-wide `default_model_source` applies. Both FLM download paths are covered — load-time (`FastFlowLMServer::load`) and explicit pull (`FlmOps::download_model` → `flm_download`, i.e. `lemonade pull` / `POST /pull`).

## Testing

_Testing details:_ Built the `lemond` target (Ninja/Release) cleanly. Regenerated committed artifacts with `docs/tools/gen_backend_boilerplate.py` and confirmed `--check` reports no drift after removing the `flm.modelscope` descriptor option. The existing `apply_default_pull_source` contract (FLM checkpoints get no persisted registry source) is unchanged and still covered by `test/cpp/test_config_default_source.cpp`.

## Documentation

Regenerated the FLM recipe-option tables in `docs/dev/backends-reference.md` and `docs/guide/cli.md` (the `flm.modelscope` row is removed; ModelScope is selected via the existing `default_model_source` / `--source` mechanism).

## Breaking Changes

The `flm.modelscope` config option from the earlier revision of this PR was never released, so removing it is not a breaking change. Known edge: an explicit `--source huggingface` on an FLM pull won't override a `modelscope` server default (FLM's `registry_source` can't distinguish explicit HF from the struct default).

## AI-assisted contribution

Please select one:

If AI tools were used:


superm1 (approved):

jeremyfowers:
Looks like CI found some problem? Please have a look.

I made a 1LoC PR here for the version bump, to try and get it into today's release: https://github.com/lemonade-sdk/lemonade/pull/2847

We will still need this PR for the modelscope support!

superm1:
> Looks like CI found some problem? Please have a look.
> 
> I made a 1LoC PR here for the version bump, to try and get it into today's release: #2847
> 
> We will still need this PR for the modelscope support!

I think Hermes messed up and passed `--modelscope` to `flm serve`.  Let's see if this is clean now.

superm1:
@jeremyfowers modelscope should be fixed now, can you have another look please?

jeremyfowers:
omg is wariobot your bot? :joy:

jeremyfowers:

**Attention level:** Elevated — any reviewer can take it, but the to-dos below need resolving before approval (PR description does not match the diff)

### Author's description vs. the diff

The description overstates the scope of changes and references a file that was not modified.
- [ ] Remove the reference to docs/guide/configuration/README.md from the PR description, or add the config README update to the PR.
  - Described: docs/guide/configuration/README.md is updated with modelscope in the config example JSON and FLM options table.
  - In the diff: docs/guide/configuration/README.md is not among the five changed files in the diff.
  - Reference: The PR description lists docs/guide/configuration/README.md under the Documentation section, but the file list shows only docs/dev/backends-reference.md, docs/guide/cli.md, and three C++ files.
- [ ] Update the PR description to remove the claim that --modelscope is passed to flm serve.
  - Described: The --modelscope flag is passed to both flm pull and flm serve commands.
  - In the diff: The flag is only passed to download_model(), which is used for flm pull; no code in the diff passes --modelscope to the flm serve path.
  - Reference: fastflowlm_server.cpp adds --modelscope to args only inside the download_model() function; the load() function calls download_model() but does not pass --modelscope to the serve command.

### Scope

**minor** — The PR adds a single BOOL option to the existing FLM backend descriptor with 16 lines of additions across 5 files — a contained extension of an existing surface.

### Suggested reviewers

- bitgamma (CLI) — The PR adds a CLI flag and updates docs/guide/cli.md.
- jeremyfowers (Adding new backends) — The PR modifies the FLM backend descriptor and server implementation.
- ramkrishna2910 (Adding new backends) — The PR adds a backend option to the FastFlowLM backend.

_Reviewed at head `919d813` by repo-manager pr-review. Regenerate with `repo-manager review-pr 2841`._

jeremyfowers (dismissed):
see bot review

SlawomirNowaczyk:
Here are a couple improvement suggestions, but nothing on this list is blocking:

**1. Minor (latent bug): asymmetric precedence in `effective_model_source`.** `model_manager.cpp:224` maps empty → `"huggingface"` first, then only lets an explicit *ModelScope* win. So an explicit per-model `registry_source == "huggingface"` gets silently overridden when the server default is `modelscope` — it can't distinguish "unset" from "explicitly HF". Harmless for FLM today (its `registry_source` is always empty, since #3036's `apply_default_pull_source` skips `flm` at `model_registry.cpp:948`), but this is a public `ModelManager` method and the trap will bite any future caller. Cleaner and matches the documented "explicit per-model provenance wins":

```cpp
std::string ModelManager::effective_model_source(const ModelInfo& info) const {
    if (!info.registry_source.empty()) {
        return remote_registry_source_name(parse_remote_registry_source(info.registry_source));
    }
    if (default_model_source_provider_) {
        return remote_registry_source_name(
            parse_remote_registry_source(default_model_source_provider_()));
    }
    return remote_registry_source_name(RemoteRegistrySource::HuggingFace);
}
```

**2. Minor: no test.** #3036 set the precedent with `test_config_default_source.cpp` registered via `add_cpp_ci_test`. This new resolution logic (which already has the asymmetry above) has none. A small unit test covering unset→default, explicit-ModelScope-wins, provider-unset fallback, and explicit-HF-vs-default would catch regressions and would have surfaced #1.

**3. Minor: confusing name collision.** There's already a static `effective_registry_source(info)` (the model's *own* provenance, used in ~20 places) right next to the new member `effective_model_source(info)` (source-to-use, with server-default fallback). Nearly identical names, different semantics. Consider `resolve_download_source()` / `download_source_for()` to signal the fallback.

**4. Design nuance to document: FLM source isn't pinned.** Because FLM is excluded from `apply_default_pull_source`, its models never record provenance, so `effective_model_source` re-resolves from the *live* default on every load/upgrade. Pull an FLM model under `modelscope`, later flip the default to `huggingface`, and subsequent loads switch registries — unlike registry-backed models, which pin the source at pull time. Worth one line in the `default_model_source` docs noting FLM follows the current default rather than pinning.

**5. Residual scope question: discovery path.** `modelscope` is threaded only into the download path, not into `flm_discover_models()` / `flm_installed_checkpoints()` (the `flm list --json` calls). If FLM needs `--modelscope 1` to enumerate ModelScope-hosted checkpoints, discovery of ModelScope-only models could still miss. Fine if FLM's list is source-agnostic — worth confirming and stating explicitly.

**6. Nit:** call sites compare `== "modelscope"`. Returning a `RemoteRegistrySource` enum (or a `bool model_source_is_modelscope(info)` helper) would drop the string literals and the parse→format→re-parse round-trip.

ZaneNi (changes requested):
I tried on Windows using Lemonade version 11.9 from the artifacts from this PR.

```
.\lemonade.exe -v
 lemonade version 11.9.0
```

When running the following command:

```
.\lemonade.exe pull qwen3-0.6b-FLM --source modelscope
```

It looks like the modelscope flag isn't  passed correctly to FLM backend, the log in Lemonade GUI still shows "Pulling model from HuggingFace"

```
\Users\zani.cache\lemonade\bin\flm\npu\flm.exe" "pull" "qwen3:0.6b" "--force"

 2026-09-02 09:07:28.933 [Info] (FLM) [FLM] Pulling model from HuggingFace
```

ZaneNi (approved):
Looks good to me.

PR #2843, opened 2026-07-29
[Stable Diffusion] Add the three main performance options as exposed toggles

NineBallo:
This adds the three main performance toggles as individual "promoted" options.
--diffusion-fa, --diffusion-conv-direct, and --vae-conv-direct

---- AI Description
# Add typed BOOL performance options for stable-diffusion.cpp

## Summary

Add typed BOOL performance options for stable-diffusion.cpp — `diffusion_fa`, `diffusion_conv_direct`, `vae_conv_direct` — as first-class recipe options instead of raw flags embedded in `sdcpp_args`.

## What changed

**New BOOL options** (`sdcpp.h`, regenerated `defaults.json` + docs via `gen_backend_boilerplate.py`):
- `diffusion_fa --diffusion-fa`, `diffusion_conv_direct --diffusion-conv-direct`, `vae_conv_direct --vae-conv-direct`
- Added to `config_extra` defaults; surfaced in CLI, config, and backends-reference docs.
- `server_models.json`: `Qwen-Image-GGUF` and `Qwen-Image-2512-GGUF` migrated from `sdcpp_args: "--diffusion-fa --offload-to-cpu"` to the typed `diffusion_fa: true` option.

**`build_merged_sdcpp_args()`** (`sdcpp_server.cpp`) — called from `SDServer::load()` with the fully-resolved effective options:
- Discovers which BOOL descriptor options have CLI flags, so new options need no code changes.
- Filters known boolean flags (and their values) out of the existing `sdcpp_args`, keeping non-boolean custom args (`--control-net`, `--offload-to-cpu`, …) intact. A spelled flag's value is consumed (e.g. `--diffusion-fa 1`) so it no longer survives as a stray argument, and `--diffusion-fa 0`/`false` is honored as disabled.
- Rebuilds the args: custom args first, then enabled boolean flags from typed options, legacy spelling, or backend-forced flags.
- The hardcoded Vulkan `--vae-tiling --diffusion-fa` injection in `load()` is folded into this path, staying backend-gated.

## Why

`--diffusion-fa`, `--diffusion-conv-direct`, and `--vae-conv-direct` were previously only reachable by hand-writing them into `sdcpp_args`, bypassing option validation and the per-model recipe system. Making them typed options gives them defaults, docs, CLI/config support, and merge semantics — while remaining backward compatible with configs that still spell them in `sdcpp_args`.

Note: the fix for the stray-argument edge case is included in this diff (the value-consumption logic in `build_merged_sdcpp_args`)

### DIFF
--- a/docs/dev/backends-reference.md
+++ b/docs/dev/backends-reference.md
@@ -128,6 +128,9 @@ the generator instead. Prose outside the markers is preserved. -->
 |--------|----------|------|---------|-------------|
 | `sd-cpp_backend` | `--sdcpp` | BACKEND | "" | SD.cpp backend to use |
 | `sdcpp_args` | `--sdcpp-args` | ARGS | "" | Custom arguments to pass to sd-server (must not conflict with managed args) |
+| `diffusion_fa` | `--diffusion-fa` | BOOL | false | Enable fused-attention |
+| `diffusion_conv_direct` | `--diffusion-conv-direct` | BOOL | false | Enable direct convolution |
+| `vae_conv_direct` | `--vae-conv-direct` | BOOL | false | Enable VAE direct convolution |
 | `steps` | — | SIZE | 20 | Number of diffusion steps |
 | `cfg_scale` | — | SIZE | 7.0 | Classifier-free guidance scale |
 | `width` | — | SIZE | 512 | Output image width |
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -399,6 +399,9 @@ The following options are available depending on the recipe being used:
 |--------|-------------|---------|
 | `--sdcpp BACKEND` | SD.cpp backend to use | Auto-detected |
 | `--sdcpp-args ARGS` | Custom arguments to pass to sd-server (must not conflict with managed args) | `""` |
+| `--diffusion-fa BOOL` | Enable fused-attention | `false` |
+| `--diffusion-conv-direct BOOL` | Enable direct convolution | `false` |
+| `--vae-conv-direct BOOL` | Enable VAE direct convolution | `false` |
 
 #### FastFlowLM NPU (`flm` recipe)
 
--- a/docs/guide/configuration/README.md
+++ b/docs/guide/configuration/README.md
@@ -106,10 +106,13 @@ Values set in the user's `config.json` always take precedence over these seeded 
     "cpu_bin": "builtin",
     "cuda_args": "",
     "cuda_bin": "builtin",
+    "diffusion_conv_direct": false,
+    "diffusion_fa": false,
     "height": 512,
     "rocm_args": "",
     "rocm_bin": "builtin",
     "steps": 20,
+    "vae_conv_direct": false,
     "vulkan_args": "",
     "vulkan_bin": "builtin",
     "width": 512
--- a/src/app/src/renderer/ModelOptionsModal.tsx
+++ b/src/app/src/renderer/ModelOptionsModal.tsx
@@ -416,19 +416,6 @@ const ModelOptionsModal: React.FC<SettingsModalProps> = ({ isOpen, onCancel, onS
     return opt?.useDefault ?? true;
   };
 
-  // Check if a boolean option is forced by the selected backend (e.g. Vulkan forces --diffusion-fa)
-  const isOptionForcedByBackend = (key: string): boolean => {
-    if (modelInfo?.recipe !== 'sd-cpp') return false;
-    const backend = (options as unknown as Record<string, { value: string }>)['sdcppBackend']?.value ?? '';
-    const forcedFlags: Record<string, string[]> = {
-      diffusionFa: ['vulkan', 'cuda'],
-      diffusionConvDirect: [],
-      vaeConvDirect: [],
-    };
-    const forcedList = forcedFlags[key] ?? [];
-    return forcedList.includes(backend);
-  };
-
   const renderContextSizeField = (key: string) => {
     const def = getOptionDefinition(key);
     if (!def || def.type !== 'numeric') return null;
@@ -606,7 +593,6 @@ const ModelOptionsModal: React.FC<SettingsModalProps> = ({ isOpen, onCancel, onS
 
     const value = getOptionValue<boolean>(key);
     const useDefault = getOptionUseDefault(key);
-    const forced = isOptionForcedByBackend(key);
     if (value === undefined) return null;
 
     return (
@@ -629,12 +615,9 @@ const ModelOptionsModal: React.FC<SettingsModalProps> = ({ isOpen, onCancel, onS
           <input
             type="checkbox"
             checked={value}
-            onChange={(e) => !forced && handleBooleanChange(key, e.target.checked)}
-            disabled={forced}
+            onChange={(e) => handleBooleanChange(key, e.target.checked)}
             className="settings-checkbox"
-            title={forced ? 'This option is locked by the selected backend' : ''}
           />
-          {forced && <span className="settings-locked-indicator" title="Locked by backend">🔒</span>}
           <div className="settings-checkbox-content">
             <span className="settings-description">
               {def.description}
--- a/src/app/src/renderer/ModelOptionsModal.tsx
+++ b/src/app/src/renderer/ModelOptionsModal.tsx
@@ -416,6 +416,19 @@ const ModelOptionsModal: React.FC<SettingsModalProps> = ({ isOpen, onCancel, onS
     return opt?.useDefault ?? true;
   };
 
+  // Check if a boolean option is forced by the selected backend (e.g. Vulkan forces --diffusion-fa)
+  const isOptionForcedByBackend = (key: string): boolean => {
+    if (modelInfo?.recipe !== 'sd-cpp') return false;
+    const backend = (options as unknown as Record<string, { value: string }>)['sdcppBackend']?.value ?? '';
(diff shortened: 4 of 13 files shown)


jeremyfowers:
@NineBallo please be aware that we have a feature freeze on gui2, and the gui changes will need to go into GUI3. If you can remove the gui features from this PR and rebase them into gui3 in a future PR that would be great.

jeremyfowers (changes requested):
see comment

NineBallo:
Rebased on main and removed frontend changes

fl0rianr (commented):
I took a look at the backend changes. I think there are still two issues to fix before this is ready:

1. `build_sdcpp_args_merge()` falls back to `sdcpp_backend`, but the actual option key is `sd-cpp_backend`. This can break normal auto-loads where no backend is explicitly passed.

2. The SD flags are merged before the Router resolves the final `effective_options`. This means saved model options / config defaults for `diffusion_fa`, `diffusion_conv_direct`, etc. may not be reflected in `sdcpp_args`.

I think the cleaner approach would be to build these flags in `SDServer::load()` from the already resolved `RecipeOptions`. That would also avoid duplicating the Vulkan `--diffusion-fa` handling we should avoid.

Otherwise the descriptor changes look good, and removing the GUI2 changes addresses the previous review.

NineBallo:
@fl0rianr Were you thinking something like this?

fl0rianr (changes requested):
Thanks for the update! Moving the merge logic into `SDServer::load()` looks much cleaner and addresses the earlier issues.

One thing I noticed: existing models that already set `--diffusion-fa` inside `sdcpp_args` may now lose that flag, because known boolean flags are removed first and only re-added when the new typed option is enabled. For example, some models in `server_models.json` currently rely on `--diffusion-fa` in `sdcpp_args`.

Could you make sure those existing configurations keep the same behavior, either by migrating them to `diffusion_fa: true` or otherwise preserving the flag? Thanks.

NineBallo:
@fl0rianr Forgot to ping ya, I just moved em over

fl0rianr (commented):
Thanks for the update! The built-in models now look correct with `diffusion_fa: true`, so that addresses the issue I mentioned previously.

I think there are just two things left before this is ready:

* Existing user `recipe_options.json` configs that still have `--diffusion-fa` inside `sdcpp_args` can still lose that flag, since the merge removes known boolean flags and only re-adds them from the new typed options. It would be good to preserve backward compatibility here.
* The generated `defaults.json` still doesn't include the three new options from `config_extra`, so the backend boilerplate generator should be run and its output committed.

The PR is also currently conflicting with `main`, so it will need a rebase as well. Otherwise the direction looks good, thanks!

NineBallo:
@fl0rianr Fixed those issues, and cleaned up some extra stuff laying around. What do you think?

fl0rianr (commented):
Thanks for the updates! Good work. The previous issues look addressed now: the legacy flags are preserved, the generated files are updated, and the CUDA/cleanup changes look good.

I noticed one remaining edge case: legacy args like `--diffusion-fa 1 --offload-to-cpu 1` are still present in the repo. The current merge removes `--diffusion-fa` but leaves the following `1`, resulting in a stray argument. Could you handle the flag together with its associated values when rebuilding `sdcpp_args`?

Also, the PR description is now a bit outdated since it still mentions the removed WebUI changes and CUDA forcing.

NineBallo:
@fl0rianr Did you mean something like this?

fl0rianr (commented):
Thanks, the previous edge case looks fixed now.

I noticed one remaining issue with the forced Vulkan flags: `is_backend_forced()` sets the flag to `true`, but a legacy value like `--diffusion-fa 0` is applied afterwards and can set it back to `false`. Since backend-forced flags should always win, could you apply the forced check last?

Other than that, this looks good to me. The PR description also still has the old WebUI wording in the first couple of lines, but that’s just cleanup. Thanks!

NineBallo:
@fl0rianr 
Changes are pushed to fix the ordering, and I also reduced the comments down a bit (hopefully passes the comment slop filter)

(Everything below I'm not suggesting as part of this PR just discussing)
Would there be any benefit to implementing a simple migration system for legacy configs? Thinking like mssql, where instead of maintaining a bunch of edgecases we just roll it all into a standardized file. Feel like it would have less maintenance work but at the same time it adds a lot of it's own complexity. 

I'd just stamp cfg revisions somewhere in the json, then on startup it compares it's hardcoded rev to the configs, then runs a rev-update command for each rev bump. 

Just not sure how much configs really change, but it was a lot of the headache here so wanted to bring it up.

fl0rianr (approved):
LGTM now, thanks! @bitgamma wanna take a short look as well?

bitgamma (changes requested):
I don't think this is a good direction to go, or the number of parameters we need to map will explode, since a convincing argument can be made for basically every flag in every backend. 

We already have the ability to set flags per-backend so the user can set them server-wide.

NineBallo:
@bitgamma 

Reasons for adding these specifically is because they are the most commonly changed settings for different hardware/config setups. The idea was to then expose these in the frontend so users could quicky A/B test the settings that likely have the highest impacts on gen speed without needing to reference anything outside of this app. 


--diffusion-conv-direct & --vae-conv-direct (save a lot of vram, but _can_ make absolute performance worse)
--diffusion-fa (again can save vram but isn't fit for a global enable)

There aren't really any other candidates that fit this "big win"/"possible loss" scenario...
Maybe like --vae-tiling or --fa but it's still a few of many toggles. Having these setup to be frontend toggles I think makes a lot of sense however I'm open to discussion, esp if the new gui has a "gen options" box where you could throw these in directly.

I'd be open to swapping --diffusion-fa for just --fa in this implementation if you think that makes more sense?

But even in the worst case I don't think 4 toggles in an "advanced" dropdown section of the new GUI is asking for too much here. I see the worry with adding "infinity" options to this, but given the lack of potential candidates for sd specifically I don't see it as a huge risk in this case.

bitgamma:
@NineBallo I disagree. This logic applies to ALL backends across all modalities. Even for llamacpp we do not expose the single flags with a separate checkbox or other UI element, not even very important stuff like MTP, which has some automatic handling but not user-facing settings. We expose only the raw args + those settings where lemonade can/must calculate, validate or modify the value, like the ctx_size. We need to apply the same principles to ALL backends and I honestly don't see a good argument to treat these flags differently.

There is already a way to enable these flags via the sdcpp.args param so this is purely a UX change. Since this goes in the opposite direction of what we have done until now, without a very strong and convincing argument this shouldn't be done.

NineBallo:
@bitgamma I presumed this assumption was changing with the new gui rollout, this is the backend work for when that ships. Was discussing with sawan in the DC about the mtp example and they mentioned work on exposing those options going forward? I was originally going to try and get the gui3 work for that done, but dropped it cause I haven't had time. 

[PR 1638](https://github.com/lemonade-sdk/lemonade/pull/1638)

If this is something where you are putting your foot down, I totally get it but would it be better to keep this on ice for a bit until the gui update ships and see where things are going? I can also live with it just being killed but if this conversation gets revived in like 3 weeks then at least my work isn't going to waste yk.

bitgamma:
@NineBallo feel free to leave this open, no hurry to close this. I am just giving my view about this matter.  

Everybody has their favorite options and have a good story why they are special and need to be extracted into separate options. This very easily becomes super hard to maintain and creates incredibly busy user interfaces where nobody know what everything does.

also, just fyi, if you have an AMD iGPU you can see there is a new image generation backend. It is several times faster than sdcpp so you could try that to get better performance.

NineBallo:
@bitgamma 
Yeah great work with that btw! Currently looking at porting over that "respect users model options" pr to the "TheNoise" side of things since they both shared the same logic.  

I honestly don't use image gen very often, mostly to dink around with every few months and see where things have gone. Just wanted to get it working a bit smoother for new users. Now that flux2 is on the new backend I'm probably not gonna be messing with this one much if at all lol


As for the other piece I totally get that, and agree. I'll keep it alive for a while and see where the conversation goes :)

jeremyfowers:
Throwing in my 2 cents here: 

### lemond

I would not want CLI arguments for backends to become part of the `lemond` config space, otherwise we end up hardcoding support for many dozens of arguments per backend and that would feel redundant and hard to maintain to me. I'm not sure if this PR is doing that or not, just saying it's how I feel.

### gui

IMO it is a good short-term direction for the GUI to expose popular backend args as GUI controls (e.g., sliders, toggles, etc.) instead of forcing a GUI user to reason about CLI flags. We started in this direction in GUI3 for the 6 "architectural default" arguments for LLMs (temp, top-k, etc.).

### auto-tune

Our real goal, however, as a community should be to achieve true auto-tune: we have ways of discovering and sharing best-known-configurations for different model/hw/scenario combinations, and we auto-apply these at load time. The best UI is to automate to the point where we don't need any UI.

NineBallo:
@bitgamma @fl0rianr 
How would you feel about me swapping out the plumbing to instead toggle a string directly into the cmd_args?
That fits the "not extending the lemond config space" but leaves us with the tooling for gui3. 

Maybe could pivot this PR to be a generic path. Letting the gui declare certain settings to 'toggle' directly in the cmd_args. Then let the future gui pr's handle whatever comes after.

bitgamma:
@NineBallo the approach of a generic "command line builder helper" sounds very nice to me

PR #2844, opened 2026-07-29
document multi checkpoint model creation

ckuethe:
this wasn't immediately obvious, and I didn't want to export and import a bunch of models using the web ui to copy them from one lemonade instance to another.

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -274,13 +274,25 @@ The `recipe` field defines which software framework and device will be used to l
 |-----------|----------|-------------|
 | `model_name` | Yes | Namespaced [Lemonade Server model name](https://lemonade-server.ai/models.html) to register and install. |
 | `recipe` | Yes | Lemonade API recipe to load the model with. |
-| `checkpoint` | Yes | HuggingFace "main" checkpoint to install. |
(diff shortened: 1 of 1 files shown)


docs/api/lemonade.md
```diff
@@ -273,8 +273,9 @@ The `recipe` field defines which software framework and device will be used to l
 | Parameter | Required | Description |
 |-----------|----------|-------------|
 | `model_name` | Yes | Namespaced [Lemonade Server model name](https://lemonade-server.ai/models.html) to register and install. |
-| `checkpoint` | Yes | HuggingFace checkpoint to install. |
 | `recipe` | Yes | Lemonade API recipe to load the model with. |
+| `checkpoint` | Yes | HuggingFace "main" checkpoint to install. |
+| `checkpoints` | No | HuggingFace checkpoints to install, for multi-checkpoint models. |
```
bitgamma, line 278: let's also list somewhere all currently defined checkpoint types:

`main`: main checkpoint for the model
`draft`: drafter for dflash, eagle, mtp (when not embedded in main)
`vae`: the image vae
`text_encoder`: text encoder for image generation

bitgamma (changes requested):
nice! let's just only add this note and we are good to go

bitgamma (approved):

PR #2845, opened 2026-07-29
[Stable Diffusion] Respect the user configurable "Model Options"

NineBallo:
This PR aims to promote the user configurable "Model Options" to take precedent over the hidden "Image Defaults". 
The image defaults are still used as a fallback.

AI Description:
# Respect "Model Options" over "Image Defaults" for sd-cpp

## Summary

User-saved "Model Options" now take precedence over a model's `image_defaults`
for sd-cpp image generation, and the full option-precedence ladder is resolved
in one place instead of being split across the backend and the model manager.

## What was wrong

`SDServer` cached the model's `image_defaults` at load time
(`image_defaults_` member) and applied it **above** the effective recipe
options. Because every read in `build_extra_args()` / `resolve_size()`
preferred `image_defaults_` over `recipe_options_`, settings a user saved via
the "Model Options" UI/CLI (steps, cfg_scale, size, sampling_method,
flow_shift) were silently ignored whenever the registry entry had an
`image_defaults` block.

## What changed

The model-level layers are now folded into the effective `RecipeOptions` once,
at load time, in `build_recipe_options()`:

1. Per-request fields in the JSON body (checked first, at request time)
2. User-saved recipe options (API/CLI)
3. Model `recipe_options` (registry entry)
4. Model `image_defaults` (from `server_models.json`)
5. Architecture defaults (router fold via `inherit()`)
6. Global config defaults (lowest)

Key pieces:

- **`RecipeOptions::merge_precedence_layers()`** — new helper that merges
  user-saved > registry `recipe_options` > `image_defaults` into one option set.
  `*_args` keys are replaced wholesale here (token-merge stays in `inherit()`
  between router layers).
- **`RecipeOptions::has_option()`** — lets backends distinguish "set in some
  layer" from "resolves to a descriptor default", so unset values fall through
  to sd-server's native defaults instead of being forced from the descriptor.
- **`SDServer`** — drops the cached `image_defaults_` and reads everything from
  the effective recipe options. Explicit `0.0` `cfg_scale`/`flow_shift` are now
  forwarded (0.0 is a real value that disables CFG guidance); previously any
  value `<= 0` was dropped.
- **Test** — new `test_recipe_options_precedence` (cpp-ci) covering the full
  ladder: model-level layer merging, `inherit()` semantics, `has_option()`
  set-ness vs. default resolution.


NineBallo:
This is a fairly scoped change, only thing I can think of that might be an issue is the lambda's. I think it's a bit easier to read but I'm down to switch back to the if/else style if wanted.

jeremyfowers:
@NineBallo please be aware that we have a feature freeze on gui2, and the gui changes will need to go into GUI3. If you can remove the gui features from this PR and rebase them into gui3 in a future PR that would be great.

jeremyfowers (dismissed):
see comment

NineBallo:
Removed frontend changes and rebased on main :)

fl0rianr (commented):
I think there is still one issue with the current approach:

`has_option()` is checked on the already merged `effective_options`. Since the global SD defaults (`steps=20`, `cfg_scale=7`, `width/height=512`) are merged into those options, `has_option()` will also return true for values the user never explicitly set.

That means `image_defaults` can still be skipped. For example, SD-Turbo could use 20 steps instead of its model default of 4, and SDXL could fall back to 512x512 instead of 1024x1024.

I think we need to preserve the explicitly set/model-saved options separately from the fully merged effective options, so the priority really stays:

request → explicit recipe options → image_defaults → global/baked default.

NineBallo:
@fl0rianr What do you think abt this?

fl0rianr (changes requested):
The output precedence seems to already be represented by the effective RecipeOptions built in ModelManager + Router. explicit_keys_ does not reliably track user-explicit provenance because every RecipeOptions constructor input, including global and architecture defaults, is marked explicit. Instead of adding a second provenance mechanism, I think SDServer should consume the already-merged effective values directly. We should also add precedence tests before merging.

NineBallo:
@fl0rianr Made a laundry list of changes here but I think this looks a lot more clean. Make part of it generic so we can later pull into TheNoise which suffers from the same issue.

fl0rianr (commented):
This looks much cleaner now, thanks for reworking it. The previous `explicit_keys_` concern is addressed nicely, and I like the added precedence tests.

I think there’s just one functional edge case left: `RecipeOptions` correctly treats `0.0` as an explicit value, but `SDServer` only forwards `cfg_scale` / `flow_shift` when they are `> 0.0f`. So an explicit `cfg_scale: 0.0` currently gets dropped instead of winning by precedence. It would be good to cover that at the SDServer serialization level as well.

A couple small cleanup items: the PR description is outdated and still mentions the frontend change / old `has_option()` semantics.

Other than that, this is looking much better and I’d be happy to re-review once the `0.0` case is handled.

NineBallo:
@fl0rianr passing through 0.0f values now :)

fl0rianr (approved):
Good work, thanks!
Approving already, pending CI since @jeremyfowers needs to give up on his request changes and I would like to have @bitgamma to have a short look as well - thanks!

The pre-commit error can be fixed once we have the code secure since its an comment verbosity only.

bitgamma:
I think we should consider if image_defaults are still needed or if they can be replaced completely by recipe options since they largely overlap

fl0rianr:
Yeah, I think that’s worth looking into. There’s quite a bit of overlap now, and having a single mechanism would definitely be cleaner. I’d probably keep that as a follow-up though, since replacing image_defaults entirely would also mean checking registry/API compatibility and migrating the existing model entries. For this PR I think keeping the existing layers and fixing their precedence is a good scoped step.

bitgamma:
@fl0rianr I agree, I also think this can be merged as-is

fl0rianr:
@NineBallo Please adjust comments only to get pre-commit green and we can merge. Thanks.

PR #2846, opened 2026-07-29
Polish GUI3 model configuration and apps UX

kpoineal:
## Summary
- Streamline GUI3 model configuration around direct load settings: Configuration-first details, cleaner backend/device/context controls, editable backend args, discard flow, and removal of redundant sampling/header chips plus hidden merge/mmproj toggles.
- Polish chat model UX with responsive loaded-model cards, model-detail navigation, non-blocking load picker state, and MCP server/tool selection with persisted filtering.
- Move App Directory out of Settings into a top-level Apps pane, center its content, add the Lemonade titlebar mark, and switch Settings to a gear icon.
- Update effective-settings wording and add accessibility coverage for the post-preset/direct-configuration cleanup.

## Validation
- `npm run typecheck -- --pretty false`
- `git diff --check`
- pre-commit hooks during commit


fl0rianr (changes requested):
Removing the user-facing preset workflow is intentional. I am in favor of keeping the preset infrastructure in place if presets are expected to return later. Preserving the data model, resolution logic, and existing preset implementation should make a future reintroduction much easier than removing and rebuilding the feature.

However, while the user-facing preset workflow is disabled, the application must guarantee that only the Default preset can be active.

### 1. Disabled presets must always fall back to Default

The Presets page and the visible preset selection controls are removed, but non-default presets can still be activated internally.

In particular, the built-in Lemonade tools still expose two ways to change the active preset:

* `load_model` accepts `preset`, `preset_id`, and `preset_name`.
* `change_preset` can assign a different preset to an already loaded model.

Both paths persist a model-to-preset binding through `saveApplied()`. This can leave a model using an invisible non-default preset that the user cannot inspect or reset through the GUI.

Keeping the preset implementation is reasonable, but it should be placed behind a central feature flag while the UI is disabled, for example:

```ts
export const PRESETS_ENABLED = false;
```

When presets are disabled:

* `activePresetIdForModelName()` should always return `DEFAULT_PRESET.id`.
* Existing or stale model-to-preset bindings must not affect runtime behavior.
* `change_preset` should not be exposed in `LEMONADE_TOOLS`.
* Preset parameters should be removed from the exposed `load_model` schema.
* The tool executor should also ignore preset arguments defensively.
* Running-preset snapshots and update-preset affordances should remain inactive.
* A test should verify that a stored non-default binding still resolves to Default while presets are disabled.

For example:

```ts
function activePresetIdForModelName(modelName: string): string {
  if (!PRESETS_ENABLED) return DEFAULT_PRESET.id;
  return loadApplied()[modelName] || DEFAULT_PRESET.id;
}
```

This preserves the underlying preset architecture for a future return while ensuring that the current direct-configuration workflow remains deterministic.

### 2. Context changes can be applied one load late

In `ModelDetailPanel`, the load and reload handlers save the draft and then immediately build the load options.

However, `buildLoadOptions()` reads `ctx_size` from `effectiveTuning`, which still reflects the previous render and store state. A newly entered context value may therefore not be used for the current load or reload and can instead appear on the following attempt.

The load options should be constructed directly from the current draft values rather than relying on a store update and React rerender first.

### 3. Runtime property lookup can hold the central load mutex for up to five minutes

`Router::get_model_runtime_props()` holds `load_mutex_` while calling `server->get_runtime_props()`.

The wrapped-server implementation requests a three-second timeout, but `forward_get_request()` ignores that value and passes `0` to `HttpClient::get()`. This falls back to the global default timeout of 300 seconds.

A stalled `/props` request can therefore hold the central load mutex for up to five minutes and block model load, unload, and related router operations.

Please ensure that the intended short timeout is forwarded and preferably avoid holding the central load mutex during the network request.

### 4. auto-opt removed

Among other features auto-opt was intentionally completely removed. I do not agree with that change maybe I misunderstood something -  let's discuss that properly on discord.

Besides that points looks fine for me.

bitgamma (approved):
looks very good to me!

a few notes about things that could be improved here or in a follow-up:

- detecting 401 errors and instructing the user to insert the API key
- when using the dark theme, the backends args field has the same color as the background (at least to my eyes). It is clearly visible with the light theme instead.

two things which are more my personal preferences (feel free to ignore):
- the lemon slider is funny but I'd rather have a regular slider
- in the backend tab I'd rather have a list than a grid of buttons, especially since the number of backends is ever growing

jeremyfowers:
fyi there should be a pure-HTML reference for the brand assets in the website source code, so you can avoid packaging 26 KB of images into the app.

fl0rianr (commented):
Thanks! I reviewed the current head again with the goal of identifying only issues that should block a near-term merge.

The previous context-load and runtime-property issues are resolved. I see only one quite specific functional blocker:

### Use the canonical SD.cpp recipe and option names

The new client configuration currently uses:

```ts
sd_cpp
sd_cpp_backend
```

The server contract still uses:

```text
sd-cpp
sd-cpp_backend
```

This affects more than naming. The GUI can persist and send an option the server does not recognize, and `backendSupportsArgs('sd-cpp')` does not match the underscore-based client map.

As a result, the selected SD.cpp backend and its backend-specific arguments may be ignored.

Please keep the canonical external names throughout the configuration and request paths:

```ts
'sd-cpp'
'sd-cpp_backend'
```

The same applies to `ryzenai-llm` rather than `ryzenai_llm`.

### Non-blocking follow-up

The committed Playwright output under `src/app/test-results/` can be removed and added to `.gitignore` in a follow-up. I do not consider that a reason to delay this merge.

Once the compatibility issue above is fixed I am comfortable with merging this PR.

kpoineal:
Resolved the canonical naming issue in 5a1c1929. The client now uses `sd-cpp` / `sd-cpp_backend` and `ryzenai-llm` end-to-end, including tool load options. Legacy underscore-form browser settings are normalized before use so they cannot produce noncanonical requests. Ready for another review round.

fl0rianr (approved):
Looks good now, thanks

PR #2850, opened 2026-07-29
fix: stabilize flaky semantic_similarity routing test (#2848)

### LINKED ISSUE #2848: ORT semantic similarity test is flakey
https://github.com/lemonade-sdk/lemonade/actions/runs/30373609945/job/90345768383?pr=2835
https://github.com/lemonade-sdk/lemonade/actions/runs/30283773583/job/90042745546

Please fix ASAP

ramkrishna2910:
Fixes #2848.

`test_620`/`test_621` flaked because the coding prompt's `semantic_similarity` embedding score sat right at the 0.6 rule threshold (~0.74 vs 0.6). Embedding scores carry small run-to-run noise on the CI GPU backend, which occasionally dropped the coding score below 0.6 and flipped the route to the default model.

**Fix (test-only):** use a reference-phrase set that mirrors the coding prompt and raise the threshold to 0.70. Measured with `nomic-embed-text-v1` (Q4_K_S): coding **0.947**, non-coding **0.498** — both clear the 0.70 boundary by ~0.2, well beyond the observed noise. Verified end-to-end against a local server (real classifier path). The phrases/prompt/threshold are shared between the two tests, and scores are logged before asserting so any future drift is visible.

Note: I explored adding nomic `search_query:`/`search_document:` task prefixes in the classifier, but measured that for these short reference phrases they *lower* the coding score (0.57) and *shrink* separation (0.27→0.15) — counterproductive here — so that change was dropped. This PR touches only the test.

### DIFF
--- a/test/server_router.py
+++ b/test/server_router.py
@@ -111,6 +111,27 @@ EMBED_MODEL = "nomic-embed-text-v1-GGUF"
 # Real encoder classifier (onnxruntime backend) for the `classifier` condition.
 CLASSIFIER_MODEL = "Phishing-Email-Detection-ONNX"
 
+# Shared inputs for the semantic_similarity routing tests (test_620/test_621).
+# One reference phrase closely mirrors SEMANTIC_CODING_PROMPT so its cosine lands
+# far above the threshold; embedding scores carry small run-to-run noise on GPU
+# backends, and a borderline example would flip the route intermittently.
+# Measured with nomic-embed-text-v1 (Q4_K_S): coding ~0.95, non-coding ~0.50, so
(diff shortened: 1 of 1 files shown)


ramkrishna2910:
Closing in favor of #2824, which fixes the same flaky test (#2848) and is already CI-green and awaiting review.

#2824's root-cause diagnosis (cold-start load of the embedding model → transient `ensure_loaded` failure → `Score{ok=false}` → fall-through to default) is better targeted than this PR's approach (widening the score margin / raising the threshold). Independently corroborated below on #2824.

PR #2855, opened 2026-07-29
fix(installer): remember custom install location

### LINKED ISSUE #1119: The installer doesn't reemember the last install location
Hi,

The installer doesn't remember the last install location if it is installed in a custom location. It tries to install to the default install location on Windows 11.

    Run an older installer.

    Set a non-default location.

    Run the latest installer to see that it will not remember the user's location preference.

sujikathir:
## Summary

- Restore the saved per-user `InstallLocation` to `INSTALLDIR` before installer costing.
- Preserve the existing all-users install-directory behavior.

Fixes #1119

## Testing

- Configured-template XML validation with `xmllint --noout`
- Restore-action regression check: upstream `0`, patched `1`
- `git diff --check`

An interactive Windows upgrade was not available locally; the Windows installer CI build provides WiX/MSI compilation and packaging validation.

## Screenshots

Not applicable; this changes Windows installer authoring rather than application UI.

### DIFF
--- a/src/cpp/installer/Product.wxs.in
+++ b/src/cpp/installer/Product.wxs.in
@@ -93,7 +93,7 @@
     <SetProperty Id="INSTALLDIR"
                  Value="[EXISTINGINSTALLDIR]"
                  Before="CostFinalize"
-                 Sequence="both"
+                 Sequence="first"
                  Condition="EXISTINGINSTALLDIR AND (ALLUSERS &lt;&gt; 1)" />
 
     <!-- ============================================ -->
--- a/src/cpp/installer/Product.wxs.in
+++ b/src/cpp/installer/Product.wxs.in
@@ -90,6 +90,11 @@
                       Name="InstallLocation"
                       Type="raw" />
     </Property>
+    <SetProperty Id="INSTALLDIR"
+                 Value="[EXISTINGINSTALLDIR]"
+                 Before="CostFinalize"
+                 Sequence="both"
+                 Condition="EXISTINGINSTALLDIR AND (ALLUSERS &lt;&gt; 1)" />
 
     <!-- ============================================ -->
     <!-- Conflict Detection: Block dual per-user/all-users installs -->


fl0rianr (changes requested):
Thanks for working on this — the diagnosis is correct, and restoring EXISTINGINSTALLDIR before costing is the right general approach.

One concern: using `Sequence="both"` schedules this property assignment in both the UI and execute sequences. During an interactive upgrade, the UI sequence first restores the old path correctly. However, if the user then selects a different directory in `InstallDirDlg`, the execute-sequence action runs again before `CostFinalize` and may overwrite the newly selected path with `EXISTINGINSTALLDIR`.

Would `Sequence="first"` be a better fit here?

With `Sequence="first"`:

* Full UI installs restore the saved path in the UI sequence, while still respecting a later user edit.
* Silent installs restore the saved path in the execute sequence.
* The existing `ALLUSERS=1` behavior remains unchanged.

Suggested change:

```xml
<SetProperty Id="INSTALLDIR"
             Value="[EXISTINGINSTALLDIR]"
             Before="CostFinalize"
             Sequence="first"
             Condition="EXISTINGINSTALLDIR AND (ALLUSERS &lt;&gt; 1)" />
```

It would also be useful to validate these two upgrade cases:

1. Accept the restored custom directory.
2. Change the restored directory in `InstallDirDlg` and verify that the newly selected directory is retained.

The overall fix looks good otherwise, and the Windows installer CI confirms that the WiX authoring and MSI packaging remain valid.

sujikathir:
Addressed in `8cf015df`: the restore action now uses `Sequence="first"`, so a full-UI upgrade restores the saved path in the UI sequence without reapplying it after a user changes `InstallDirDlg`; silent upgrades still schedule it in the execute sequence.

I validated the WiX template as XML and confirmed the action still runs before `CostFinalize` with the per-user condition unchanged. I do not have a Windows interactive MSI environment, so I could not manually execute the two upgrade paths locally; the Windows installer CI will continue to validate WiX authoring and packaging.

fl0rianr (approved):
LGTM, now, thanks!

PR #2857, opened 2026-07-30
fix(docs): Added group_add configuration for ROCm support in Docker.

storm1er:
Since lemonade v11.0.0, docker lemonade is not root anymore, so we need to give container user permission to /dev/kri and /dev/kfd

### DIFF
--- a/docs/guide/install/docker.md
+++ b/docs/guide/install/docker.md
@@ -110,6 +110,20 @@ docker run -d \
 
 > This will run the server using the ROCm backend as the default for llama.cpp.
 
+> **GPU device permissions on Linux**
+>
+> The container runs as an unprivileged user (UID 10001), so it must belong to
+> the host groups that own `/dev/kfd` and `/dev/dri` (usually `render` and
+> `video`) or ROCm device access is denied. Docker resolves group *names*
+> against the container's `/etc/group`, so if `--group-add render` fails, pass
+> the host's numeric group ID instead. Find it on the host with:
+>
+> ```bash
+> getent group render video
+> ```
+>
+> Then use the numbers, e.g. `--group-add 992 --group-add 44`.
+
 ### Docker Run with AMD GPU Passthrough using ROCm on WSL
 
 Make sure you follow install steps described in [ROCm for WSL](https://rocm.docs.amd.com/projects/radeon-ryzen/en/latest/docs/install/installrad/wsl/howto_wsl.html)
--- a/docs/guide/install/docker.md
+++ b/docs/guide/install/docker.md
@@ -103,6 +103,7 @@ docker run -d \
   -v lemonade-recipe:/opt/lemonade/.cache/lemonade \
   --device=/dev/kfd \
   --device=/dev/dri \
+  --group-add video --group-add 992
   ghcr.io/lemonade-sdk/lemonade-server:latest
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guide/install/docker.md
```diff
@@ -103,6 +103,7 @@ docker run -d \
   -v lemonade-recipe:/opt/lemonade/.cache/lemonade \
   --device=/dev/kfd \
   --device=/dev/dri \
+  --group-add video --group-add 992
```
siavashhub:
```suggestion
  --group-add video \
  --group-add render \
```

siavashhub: The GID 992 is host specific, I suggest we use name instead

docs/guide/install/docker.md
siavashhub, line 112:
```suggestion

> **GPU device permissions on Linux**
>
> The container runs as an unprivileged user (UID 10001), so it must belong to
> the host groups that own `/dev/kfd` and `/dev/dri` (usually `render` and
> `video`) or ROCm device access is denied. Docker resolves group *names*
> against the container's `/etc/group`, so if `--group-add render` fails, pass
> the host's numeric group ID instead. Find it on the host with:
>
> ```bash
> getent group render video
> ```
>
> Then use the numbers, e.g. `--group-add 992 --group-add 44`.

```

siavashhub: I suggest adding this so the use would know how to get the GID and assign it if it errors out

siavashhub, line 208:
```suggestion
> The `devices`/`group_add` block above is only needed for the ROCm backend on
> Linux. If Docker cannot resolve the `render`/`video` group names, replace them
> with the host's numeric group IDs from `getent group render video`.

```

docs/guide/install/docker.md
```diff
@@ -162,6 +163,13 @@ services:
       # Persist model options and other backend binaries
       - lemonade-recipe:/opt/lemonade/.cache/lemonade
     restart: unless-stopped
+    # Needed if using rocm using linux
+    devices:
+      - /dev/dri:/dev/dri
+      - /dev/kfd:/dev/kfd
+    group_add:
+      - video
+      - "992"
```
siavashhub:
```suggestion
    # For AMD GPU (ROCm) on Linux only, also add:
    # devices:
    #   - /dev/kfd:/dev/kfd
    #   - /dev/dri:/dev/dri
    # group_add:
    #   - video
    #   - render
```

siavashhub: I suggest we add above as commented, so the user could opt for if needed.

siavashhub (changes requested):
Thank you for this! it's a genuinely useful addition, since the permissions got changed from 10.10 to unprivileged user (UID 10001) and that container user needs to be in the host's GPU groups for ROCm device access to work.

I've left a few suggestions to make it robust for everyone:

1. Hardcoded GID 992 is host-specific: the numeric ID for render/video varies per machine, so it'll silently fail for many users. Better to use the group names and tell people how to find the numeric IDs (getent group render video) as a fallback.
2. Compose block placement: the devices/group_add additions apply only to ROCm on Linux, so it's worth flagging that so non-AMD users don't copy /dev/kfd and hit errors.

docs/guide/install/docker.md
storm1er: weird, I don't have render user in /etc/passwd

siavashhub:
looks good to me
Thank you!

storm1er:
@siavashhub You didn't remove you request for changes

siavashhub (approved):

siavashhub:
> @siavashhub You didn't remove you request for changes

done and added to queue!

PR #2863, opened 2026-07-31
Feat/openmoss 0.3.0 backport/server side

pwilkin:
Backports #2838 to master

EDIT: modified to backend part of #2838


pwilkin:
@jeremyfowers since this is now the backend part of the OpenMoss 0.2.0 upgrade, would be nice to get this a bit more prio not to land with GUI3 shipping with a broken OpenMoss implementation.

fl0rianr (commented):
Nice direction overall, especially moving voice design into the backend and adding the generation-param metadata. I found three things worth addressing before merge:

* The voice-design swap looks racy with concurrent OpenMOSS requests. design_reference_sample() stops/restarts the shared speech process, while normal speech/SFX requests don't take design_mutex_, and the router allows multiple in-flight requests on the same server. A voice-design cache miss could therefore kill a concurrent generation. I think the subprocess swap needs to be serialized/drained against all OpenMOSS inference, ideally with a concurrency regression test.
* The new max_audio_frames heuristic only counts ASCII whitespace. Long CJK/whitespace-free input is effectively treated as one word and gets the minimum frame budget, which could truncate output. A character/token-based fallback would be safer.
* generation_params is documented as the source clients render from, but the app still hardcodes OpenMOSS controls/bounds. There is already drift: the descriptor caps SFX seconds at 300 while the UI duration control allows 600. Either consuming the descriptor in the UI here, or narrowing the contract for now, would avoid two sources of truth.

Other than that, the structure looks good and the CI coverage is in good shape so far.

pwilkin:
Should be fixed.

fl0rianr (commented):
Thanks, the previous points are much improved. I still see one blocker and two smaller issues:

* The voice-design swap is protected inside OpenMossServer now, but during the swap the normal process handle/port are cleared. A request arriving at that point can see the backend as dead in the router and trigger a reload instead of waiting for the swap to finish. The new concurrency test starts both requests together, so it doesn’t cover this window. I’d add an explicit lifecycle/maintenance state and a test that sends the second request only after the speech process has been stopped.
* speech_defaults / audio_defaults are exposed by the server and added to ModelInfo, but fetchBuiltInModelsFromAPI() doesn’t copy them into modelsData, so the panels fall back to descriptor defaults instead of the model-specific ones.
* The OpenMOSS UI now only offers Describe / Clone. The plain TTS path still exists and is tested (voice as a style instruction), but there is no longer a way to select it in the panel.

Apart from that, the earlier CJK/frame-budget issue and the generation-parameter bounds/source-of-truth issue look fixed, and CI is green.

jeremyfowers:
> @jeremyfowers since this is now the backend part of the OpenMoss 0.2.0 upgrade, would be nice to get this a bit more prio not to land with GUI3 shipping with a broken OpenMoss implementation.

sorry I missed this! thanks @fl0rianr for the ping.

jeremyfowers:
**Ready for review**

No to-dos.

The terse body correctly identifies this as the backend/server-side portion of #2838, and the title adds the version context; a reviewer reading the diff would not be surprised by the scope.

Checked: The PR title reads 'Feat/openmoss 0.3.0 backport/server side' and the body states 'Backports #2838 to master' with an edit 'modified to backend part of #2838'.

**Focus: focused**

All changes serve the single goal of backporting the openmoss 0.3.0 server-side work — the backend rewrite (voice design cascade, process swaps), the supporting infrastructure (AudioFormatMetadata interface, router methods, model type validation), the resource updates (version pins, model registry entries), and the CI/docs that exercise and document them.

Checked: The 19 changed files span openmoss_server.cpp (308 additions), supporting headers, router methods, server capabilities, model registry, CI workflow, and tests — all serving the openmoss 0.3.0 backport goal.

**Alignment: none found**

Checked: reviewed contribute.md Reviewer Expectation 2 and AI Contribution Policy; no bundled unrelated changes or AI-generated code artifacts observed.

**Documentation: adequate**

Checked: checked docs/api/openai.md (3/2 changed lines) and docs/dev/backends-reference.md (23 additions); backend version table updated; no gaps found.

**Testing: adequate**

Checked: traced test/server_tts_openmoss.py (191 additions) and test/server_audio_generation.py (33 additions) to CI workflow cpp_server_build_test_release.yml; test_011 concurrent test gates on PID=0 window via capabilities skip_if_unsupported; test_007 and test_012 assert X-MOSS-* headers present.

Checked: model type enforcement changes error from 500 to 400 on wrong-model requests; audited server.cpp handle_audio_speech and handle_audio_generations; no maintainer has weighed in on this behavior shift.

</details>

**Review suggestion**

Already reviewed by: fl0rianr, jeremyfowers.

**Why this rung**

A contained fix or extension. The diff extends the existing openmoss backend with voice design cascade, format metadata, and validation — a contained change to one backend's behavior and its supporting server infrastructure, not a new backend or re-architecture.

**Reviewer search**

Checked: contribute.md Review Process rung 1 (minor = 1 reviewer); table lookup: @fl0rianr (ci); code-authorship pass via get-pr-code-authors.sh: @bitgamma (12 lines, router/server capabilities), @sawansri (5 lines, CLI recipe import); @pwilkin (OpenMOSS) is the author so skipped.

</details>

_[AI-assisted review] Automated pre-review from repo-manager — flags for the human reviewer, not a replacement for one. Reviewed at head `bfbb515`; regenerate with `repo-manager review-pr 2863`._

jeremyfowers:
:point_up: bot is pleased with the scope reduction

src/cpp/cli/recipe_import.cpp
```diff
@@ -29,6 +29,8 @@ const std::vector<std::string> kKnownKeys = {
     "model_name",
     "models",
     "image_defaults",
+    "pcm_sample_rate",
```
bitgamma: I think we should avoid adding top-level elements whenever possible. Why can't recipe_options be used instead?

at the very least, these could be grouped under "audio_defaults" in a way similar to "image_defaults" (even if I think that image_defaults also needs to fold into recipe_options)

src/cpp/resources/backend_versions.json
```diff
@@ -11,6 +11,22 @@
...
           "moss-tts-cuda-windows-x64.zip": "sha256:2b2f5f4b6757d99acf621b43426483ab0bc4a70f1f13904c17deed09cac70969"
+        },
+        "v0.2.0": {
+          "moss-tts-vulkan-linux-x64.tar.gz": "sha256:26f67337940cffaa26efd233764888c1e571ff438b16a8c432ca090e31e3d372",
+          "moss-tts-vulkan-windows-x64.zip": "sha256:d35ff25c39b0f9fb75191a8d267e4350fdb9bad8ed68f23f2925272a4d27354c",
+          "moss-tts-rocm-linux-x64.tar.gz": "sha256:32b94cc4b3c7fd6856ed6f73f281ac2099eacc07dfd3344ae34ee4145e0a88d0",
+          "moss-tts-rocm-windows-x64.zip": "sha256:17e782acafed4865a40ed2f8330c46dbdec15cff5d6cb66c243feeaa5d3e5ad0",
+          "moss-tts-cuda-linux-x64.tar.gz": "sha256:f4b7ab3fc308c66c984ee42fa0b038b483dde2d20dbfed11fa7f94d78205365c",
+          "moss-tts-cuda-windows-x64.zip": "sha256:8229a74e51f111db3214b3e3c3fd2d9e6dad34443448693023519c4bb7da688d"
+        },
```
jeremyfowers: Why keep a historical record of OpenMOSS versions that we don't support? v0.1.3 no longer works in Lemonade.

bitgamma (changes requested):

jeremyfowers:
@pwilkin the scope of changes here looks alright to me. What would be most helpful here in reviewing would be some guidance about how I can experience the expected improvements from the PR, so that I can evaluate whether they work.

I understand that might not be easy since there are no GUI changes here, but anything I can do from a python script would be good info.

pwilkin:
Basically coherence / faithfulness tests for generated speech longer than 10-15s should be sufficient. Generate a few speech samples of ~40s and compare.

src/cpp/resources/server_models.json
```diff
@@ -2109,23 +2109,73 @@
...
+        "labels": [
+            "tts"
+        ],
+        "size": 12.5,
+        "audio_defaults": {
+            "pcm_sample_rate": 24000,
+            "pcm_channels": 1
+        }
+    },
+    "MOSS-TTS-Local": {
```
jeremyfowers, line 2129: What's the difference between this and `OpenMOSS-TTS`? The names are similar.

jeremyfowers:
> Basically coherence / faithfulness tests for generated speech longer than 10-15s should be sufficient. Generate a few speech samples of ~40s and compare.

Oh yeah this works WAY better for me than the old version. Awesome!

jeremyfowers:
image

lololol

src/cpp/resources/server_models.json
pwilkin: Two different architectures, slightly different generation profiles.

jeremyfowers (approved):
Works great for me!

bitgamma (approved):

pwilkin:
Phew, finally :)

PR #2865, opened 2026-07-31
Add conformance tests (phase 2: model-backed)

QuTosh:
## Summary
Completes the #2425 conformance corpus with the model-backed tiers (L0a / L2 / L3),
replayed against a pinned fake backend so each assertion tests the engine's
threshold and selection logic rather than a model's floats. Phase 1 (#2761) covered
the deterministic tier only. Corpus freeze is left for the follow-up PR.

## Layout
Following review, the corpus is grouped per tier instead of one directory per policy:

```
test/conformance/routing/<version>/<tier>/
    policies.json   # { "<policy_name>": <collection.router policy>, ... }
    cases.jsonl     # one case per line, each naming the policy it runs against
```

That takes the corpus from 21 files to 9 while keeping every policy readable in
full — no defaults or diff layer. Related cases now sit together, which is the
point: a failing cluster reads as one file, not twenty directories.

## Corpus
173 cases across 27 policies, up from 73 across 9:

- **L0a — 30 cases / 4 policies.** `llm` router: `collection.router` sugar vs core
  form parity, the reply contract (code-fence stripping, exact and case-sensitive
  candidate match), implicit `default_label`, `on_error`, and how the reply's
  rationale and score surface in the trace.
- **L1 — 85 cases / 10 policies.** The deterministic tier from phase 1, plus the
  gaps listed below.
- **L2 — 18 cases / 4 policies.** `semantic_similarity`: max cosine over a
  concept's phrases, floor at 0, concept label selection, undefined-cosine
  failures, `on_error: match_true`, implicit `default_label`.
- **L3 — 40 cases / 9 policies.** `classifier`: score bands (two-sided, default,
  point, max-only), label selection, `primary()` vs `default_label`, classifier
  failure vs an empty score map, and `on_error` inside `any` / `all` / `not`
  nesting.

Model-backed cases declare their stub answers in a per-case `services` field — no
model is loaded and no expectation depends on real embeddings.

## L1 additions (12 cases)
Deterministic gaps found while filling in the model-backed tiers:

- `input_forms` — routing input is the **latest user turn only**, and field
  precedence is `messages` > `prompt` > `input` (3 cases)
- `leaf_order` *(new)* — a multi-key leaf is an implicit `all` whose children run in
  **op-name order**, and short-circuits in that order (3 cases)
- `conditions_vocab` — ECMAScript `\d` is the digit class, not a literal `d`
  (2 cases)
- `conditions_char_bounds` — no user turn ⇒ empty routing input, `chars` 0
- `conditions_features` — `tools` present but not an array ⇒ `has_tools` false
- `conditions_metadata` — `equals` compares the raw string; only `any` splits on
  commas
- `resolution` — a rule routing to `default_model` keeps `default_used: false`

Each was mutation-tested against the engine to confirm it fails when the semantic it
locks changes.

## Runner
- Compiles a fresh engine per case: a `semantic_similarity` classifier caches phrase
  embeddings on the instance, so a shared engine would pin every case to the first
  case's vectors. `policies.json` itself is parsed once per tier.
- Binds `FakeClassifierServices` and applies each row's `services` stubs. The fake is
  **strict** — a case must stub every backend call it makes, so a missing stub fails
  the case instead of passing on a placeholder default. The mirror holds too: a
  declared stub that no call consumes fails the case as a leftover.
- A `null` stub answer means "this service fails" for `embed`, `run_classifier` and
  `chat` alike, so the `on_error` cases do not lean on the engine's empty-vector
  handling.
- Decision comparison lives in `test/cpp/conformance_decision_compare.h`, shared by
  the corpus runner and a row-check unit test. It walks an explicit field list on
  both sides, so a field the list does not know about fails the case rather than
  going unchecked — a sweet way to catch drift between the serializer and the
  comparer.
- Every field is compared exactly, except a `semantic_similarity` trace `score`,
  which may differ by up to `1e-12`. That score is computed (dot product, square
  roots, a division), so its last bits differ between the x86 and ARM CI runners.
  Classifier and `llm`-router scores are copied from the stub answers, so they stay
  exact: a trace score is part of the recorded contract, not just an input to
  `result`. A trace entry names `classifier:<id>` and not the type, so the runner
  reads the types off the policy and tells the comparer which conditions may drift.
- Rejects a corpus file that declares the same object key twice at any depth.
  nlohmann keeps the last one, so a repeated `default_model`, `rules` or `decision`
  would otherwise be silently lost.
- Enforces the layout exactly: only `README.md` at the corpus root, one tier
  directory per tier, and exactly `policies.json` + `cases.jsonl` inside it. A
  misplaced corpus file fails instead of being skipped.
- Rejects a row carrying keys outside the allowlist, or typing `route_trace` /
  `model` wrongly — read with nlohmann's `value()`, those would throw and abort the
  whole run instead of failing one case.
- A self-test (`test_conformance_harness_checks`) feeds the pure guards deliberately
  malformed input and asserts each one rejects it, so a regression that neutered a
  guard cannot pass CI quietly.

## Docs
- `test/conformance/routing/README.md`: rewritten for the per-tier layout — stub
  answer format, the coverage matrix for every case, the score-comparison rule, and
  how to add a case.
- Corrected inaccurate claims: L1 does not replay with an empty `ClassifierServices`,
  and `<schema_major>` is not part of `decision`.

## Not in scope
Corpus freeze / append-only enforcement — follow-up PR.


SlawomirNowaczyk (changes requested):
The corpus and runner look solid; here are a few suggestions for improvement:

## 1. Hoist policy parsing out of the per-case loop
In `run_case_dir`, `load_case_policy()` is called inside the `while` loop, so `policy.json` is re-opened and re-JSON-parsed for **every row** in the file, even though it's identical for all cases in the directory. Only the *engine* needs to be rebuilt per case (the `semantic_similarity` embedding cache justifies that — the policy does not). Parse once before the loop, then copy the `RoutePolicy` into `compile_engine()` per iteration.

This also cleans up a subtle correctness wart: on a parse/compile failure the loop does `return executed`, which silently abandons the remaining rows of the directory. With a single hoisted parse, a bad policy fails once, up front, instead of "on whichever row happened to run first."

## 2. Make `report_mismatch` tolerance-aware
`decisions_match()` compares trace `score` within `kScoreTolerance` (1e-12), but when a case fails, `report_mismatch()` falls back to an exact `json::diff`. So a genuine failure elsewhere will *also* surface within-tolerance score deltas as "field(s) differ," pointing the reviewer at the score that `decisions_match` already accepted. The diagnostic should filter score deltas ≤ `kScoreTolerance` (or reuse the same comparison) so the reported diff matches the actual failure reason.

## 3. Scope the score tolerance to computed scores only
The 1e-12 tolerance is applied to *every* trace `score`, but only `semantic_similarity` scores are genuinely non-deterministic across x86/ARM (dot product + sqrt + division). L3 `classifier` scores come straight from the stubbed `run_classifier` map (`0.4`, `0.7`, `0.8`, …) and are bit-exact by construction. Fuzzing those weakens assertions that could be exact. Consider applying the tolerance only to `semantic_similarity` trace entries, or at minimum document why the blanket margin is safe.

## 4. Give `null` a uniform "fail this service" meaning
In `apply_row_services`, a `null` answer for `run_classifier`/`chat` maps to a real thrown failure (`set_classifier_failure`/`set_chat_failure`), but a `null` embed instead becomes an **empty vector** and relies on the engine treating an undefined cosine as failure. That couples the L2 `on_error` cases to the engine's empty-vector handling — if that path ever changes, these cases would silently stop exercising `on_error` rather than failing loudly. Add an explicit embed-failure hook on `FakeClassifierServices` (parallel to the classifier/chat ones) so `null` means "fail" uniformly and independently of empty-vector semantics.

## 5. Add a self-test for the harness's own validation
`find_case_dirs` and `read_case_row` contain a lot of guard logic — stray files under a version dir, nested subdirectories, missing `policy.json`/`cases.jsonl`, unknown row keys, unknown `services` names, duplicate case names, symlink-loop labels. None of it is verified to actually *fail* when violated; a regression that neutered a guard would pass CI unnoticed. A tiny fixture (or unit test) that feeds one deliberately-malformed layout and asserts a non-zero exit would lock down the harness that locks down everything else.

fl0rianr (changes requested):
A few additional points from my side:

* I don’t think we should keep this corpus structure as-is. This PR already adds 20 more case directories / 40 corpus files, and this will get messy fast. Before we freeze the corpus, I’d prefer a flatter structure instead of one directory + policy.json + cases.jsonl for every small policy variation.

* The fake backend should be strict for conformance tests. Right now missing stubs fall back to default values ([1,0,0], {}, empty string), so a broken or missing stub can still make a test pass. An unexpected backend call should fail the test.

* I’m also not sure we should freeze invalid requests as part of the compatibility contract. `metadata-non-string-value-dropped` intentionally violates the request schema. That feels more like a unit/schema test than a golden conformance case.

* chat and run_classifier also ignore the actual input they receive, so these tests won’t catch regressions in what we send to the backend.

I’d address at least the corpus layout and strict fake behavior before freezing this.

QuTosh:
@fl0rianr,
did you mean to make the structure more branched, not more flatter?

fl0rianr:
> @fl0rianr,
> did you mean to make the structure more branched, not more flatter?

No, I mean fewer individual files overall. I don't really care whether the directory tree is flatter or more branched.

My concern is maintainability: 50 small files are often harder to navigate and maintain than a few reasonably sized files, unless splitting them provides a clear benefit. I’d rather group related policies/cases together instead of creating a new directory + two files for every small variation.

QuTosh:
@fl0rianr 

Okay, I can explore if something can be merged. I started up with a fewer amount of policies, but then for some corner cases new policies appeared to be needed -- and the number grew quite quickly. Given how the [previous PR](https://github.com/lemonade-sdk/lemonade/pull/2761) I inferred that we might want to cover those corner cases and decided to trade the minimality to coverage. Anyways, I can reevaluate the  corpus and see how can it be reduced and commit/report back to you.

QuTosh:
@SlawomirNowaczyk 

There is a catch in your suggestion #1. If I simply move `load_case_policy()` out of the loop, the reference vectors (which are parts of `SemanticSimiliartyClassifier` will be loaded only once per policy (as they are defined when the policy is initialized). I deliberately made it possible to vary on the case-to-case base, as:

1. It is easier to work references near them when they are near the inputs
2. There are cases where we want to fix the policy, but change reference vectors from case to case (eg  `l2_semantic_scoring`).

I believe that opening `policy.json` doesn't produce that much overhead (it is a part of testing routine anyways), but I can move json parsing and version checking out of the loop if needed. However initializing the policy has to stay in-loop unless the corpus structure is to be changed (but I recommend keeping it).

On the second point — the return executed that bails out mid-file: good catch. Since the policy is the same for every row, a parse/compile failure is deterministic, so I'll validate the policy once before the loop and fail the whole directory there, instead of abandoning the remaining rows on whichever one happens to run first.

Would this be alright?

QuTosh:
@SlawomirNowaczyk 

Addressed your suggestions, pls check.

QuTosh:
@fl0rianr 

### Point 1
> I don’t think we should keep this corpus structure as-is. This PR already adds 20 more case directories / 40 corpus files, and this will get messy fast. Before we freeze the corpus, I’d prefer a flatter structure instead of one directory + policy.json + cases.jsonl for every small policy variation.

For the first suggestion Shrank the corpus a bit without loosing the coverage, but as I expected the reductation is not large (from 29 to 25 policies). If we want to go beyond that, we might need to discuss if we wan't to reduce the coverage or change the corpus layout -- but I'd rather keep it as is.

### Point 2

> The fake backend should be strict for conformance tests. Right now missing stubs fall back to default values ([1,0,0], {}, empty string), so a broken or missing stub can still make a test pass. An unexpected backend call should fail the test.

Made a fix as in your second suggestion, pls check

### Point 3

> The fake backend should be strict for conformance tests. Right now missing stubs fall back to default values ([1,0,0], {}, empty string), so a broken or missing stub can still make a test pass. An unexpected backend call should fail the test.

I decided to keep `metadata-non-string-value-dropped` case (your 3rd suggestion) for now, but it is not hard to drop/move it.

### Point 4

> chat and run_classifier also ignore the actual input they receive, so these tests won’t catch regressions in what we send to the backend.

For your 4th point -- I believe that checking the actual input of the classfier is not within the scope of conformance corpus. I think it adsds clarity if we hardcode the classifiers' ansers in `cases.json` (in `services`), as we are checking how the router would behave, not the classifiers. However if you believe it should be changed, let me know.

fl0rianr (commented):
Thanks, the strict fake changes look good. I still see a couple of issues:

check_case_name() does not actually handle a non-string name as its comment says. row.value("name", "") will throw for something like "name": 123, and the new self-test doesn't cover this.
The runner still doesn't enforce the documented exact {policy.json,cases.jsonl} layout. Extra files inside a case directory are accepted.

On the corpus structure, reducing 29 directories to 25 doesn't really address my concern. The issue for me is the overall number of small files and the long-term maintenance cost of the one-policy-directory/two-files pattern, not just a few duplicate policies.

We already have test files covering many very similar cases, with some files containing 30+ tests. I don't see much benefit in splitting these corpus cases this aggressively. The cases and policies could be grouped into fewer files instead.

In the previous PR the number of added files was lower, so this didn't stand out to me as much.

SlawomirNowaczyk:
Hi guys, I've been thinking about how to best approach this...

@QuTosh: what do you think about the following as a way to address @fl0rianr's file-count concern without losing coverage:

**Collapse each tier to a single `policies.json` and select policies by id.**

Right now every policy needs its own directory + `policy.json` + `cases.jsonl` (≈50 files). Instead, give each tier one directory with two files: a `policies.json` holding all of that tier's policies keyed by id, and one `cases.json` where each row names the policy it runs against:

```json
// l3/policies.json
{
  "policies": {
    "band":            { ...full policy... },
    "label_selection": { ...full policy... }
  }
}
```
```jsonl
// l3/cases.jsonl
{"name": "band-min-inclusive", "policy": "band", "services": {...}, "request": {...}, "decision": {...}}
{"name": "code-selected",      "policy": "label_selection", ...}
```

The runner change is small: load `policies.json` once, and for each row build/compile the policy it references (fresh per case, so the semantic embedding-cache behavior is unchanged). This takes ≈50 files down to ≈8 and matches how the rest of the routing tests are organized — many cases per file. The coverage matrix is unaffected since it keys on each row's `name`, not on the file layout.

**This also gives you a clean place to cut the scaffolding duplication.** Once policies share a file, a per-file `defaults` block that each policy inherits by shallow merge (policy keys win, arrays left alone) removes the repeated `version` / `recipe` / `candidates` / `default_model` — and, within a tier, `components` — that's currently copied into all 25 policies:

```json
{
  "defaults": { "version": "1", "recipe": "collection.router",
                "candidates": ["MathLLM","CodingLLM","DefaultLLM"], "default_model": "DefaultLLM" },
  "policies": {
    "band": { "components": [...], "classifiers": [...], "rules": [...] }
  }
}
```

`classifiers` / `rules` / `components` stay inline and fully visible, so each policy still reads as a spec. That's roughly another ~80 of ~500 policy lines gone (~15%) on top of the file-count reduction.

Net: `policies.json` + id selection handles the "too many files" concern directly; the `defaults` block handles the "same boilerplate copied everywhere" part.

Does it make sense?

QuTosh:
@SlawomirNowaczyk , @fl0rianr 

Thank you for your comments. The solution @SlawomirNowaczyk proposed will help to reduce the number of files. However I see some value in keeping `policy.json`/`cases.json` couples close and separated from each other -- it seems to me that the way they are organized now is easier to maintain and address the tests results in case the conformance tests fail.

Another point: in we are up to merging some policies together, why not to create a single file for all policies and another one for all cases? That will also simplify the runner, as there is going to be one file per version.

If the decision is up to me, I'd keep everything as is, as I don't see the problem with having this number of files. However if we want to deviate from the original layout -- there is no problem in implementing either per-band merge or a full merge. I think the decision is up to you as the maintainers.

fl0rianr:
@jeremyfowers can you give me a short heads up about the file count topic, please?
This is my point of view but I don't want to stress too much into this direction if you don't see sense in it.

QuTosh:
@fl0rianr 

Addressed this in my last commit:

> check_case_name() does not actually handle a non-string name as its comment says. row.value("name", "") will throw for something like "name": 123, and the new self-test doesn't cover this.
The runner still doesn't enforce the documented exact {policy.json,cases.jsonl} layout. Extra files inside a case directory are accepted.

QuTosh:
@fl0rianr @SlawomirNowaczyk @Geramy 

I have tried 2 versions of how the conformance corpus can be reorganized. They are somewhat polished, but for each version I might want to do further improvements.. I believe the tradeoffs we are exploring are num_of_files-vs-transparency and boilerplate-vs-transparency. I the the I see 3 options from now:

1. Keep the structure as it is in this branch now -- each policy is in a separate file. Lots of files and boilerplate, but IMO the easiest to work with.

2. Merge the policies and cases on per-tier level, but keep the full polices (don't introduce diffs). Solves the number of files problem, keeps the bolierplate.  Looks okay to me. Check it here: [QuTosh/semanctic_router_conformace_merged_policies](https://github.com/QuTosh/lemonade/tree/QuTosh/semanctic_router_conformace_merged_policies)

3. Merge the polices on per-tier level, introduce the default polcies. The variants are coded with diffs. More or less waht @SlawomirNowaczyk proposed. Same number of files as in 2, least boilerplate, but harder to read. Check it here:  [QuTosh/semanctic_router_conformace_diff_policies](https://github.com/QuTosh/lemonade/tree/QuTosh/semanctic_router_conformace_diff_policies)


My opinion -- I'd keep #, maybe do 2. I think for such corpus it is important to prioritize the transparency -- Imagine a developer whose change broke the test and trying to find the right policy and assemble it from diffs. It is possible, but IMO annoying. number of files doesn't seem like a large problem to me.  However I can go on with either option.

Another option might be pushing tests to a separate repo -- given how much coverage is needed, the testing harness will grow anyways, so it might make sense to keep it separated.

Please advise.

SlawomirNowaczyk:
Thanks for actually building out 2 and 3 — having both to compare makes this much easier to reason about.

A few observations to try to converge this:

**Option 1** is not addressing the key blocker on this PR: the file count. So the real choice is 2 vs 3.

**The file-count win comes from per-tier grouping, which 2 and 3 both do.** The `defaults`/diff layer in #3 only reduces boilerplate. So the tiebreaker between 2 and 3 is purely about which kind of transparency we want.

And there are two kinds:
- "Absolute transparency" — read one policy in full. Full-text (#2) wins.
- "Differential transparency" — see the one axis that separates a failing case from a passing sibling, and what's common across a failing cluster. Defaults+diffs (#3) wins.

Here's the thing: the debugging scenario cited in favor of transparency ("a dev whose change broke a test") is actually a *differential* task, not an absolute one. When an engine change goes red you rarely reconstruct one policy in isolation — you triage by pattern ("all the `match_true` cases broke, the plain ones didn't"). Because every policy in a tier diffs against the *same* `defaults`, their deltas line up and that pattern jumps out. Full-text policies bury the discriminating line inside ~30 identical ones. This is why I think #3 serves the debugging story better.

That said, #3 is only better **if the diffs are kept semantically pure** — i.e. the delta is exactly the behavioral variable and nothing incidental. That means pushing the invariants (and things like model names / classifier ids) into `defaults` aggressively, which is the direction the diff branch is already going. A "diff" that also varies model names for no reason muddies the signal.

**Bounded cost of #3:** the few policies that *override* a defaulted value (e.g. `l1/resolution` replacing `candidates`) stop being clean one-liners and need mental reconstruction under merge semantics. Keep those rare and explicit and it's a minor tax.

**Regardless of layout,** the runner should print `<tier>/<policy>/<name>` on failure — that's what makes finding the cluster fast, and it matters more than 2-vs-3.

**On a separate repo:** I'd avoid it. The corpus's value is moving in lockstep with the engine and schemas... splitting it breaks atomic engine+corpus commits and complicates the append-only freeze that's the next PR. Harness growth is an argument for good in-tree organization, not exile.

@fl0rianr: does per-tier grouping (either 2 or 3) clear your file-count concern, and do you have a preference between the two?

fl0rianr (changes requested):
Thanks for actually trying the different layouts. That makes the tradeoff much clearer.

Yes, per-tier grouping addresses my concern, and I would prefer option 2.

My point is not that we need to minimize the file count or remove every bit of boilerplate. What I want is functional concentration. Related tests should live together where that makes sense, similar to how we already organize a lot of the routing tests in the repo by functional area rather than creating a separate file for every small variation.

For me, ~50 small files for this one corpus is too fragmented. It makes navigation, review and future maintenance noisier without giving enough benefit in return.

A policies.json + cases.jsonl per tier feels like a good balance. L0a/L1/L2/L3 remain clearly separated, the full policies stay directly readable, and we avoid introducing an additional defaults/diff mechanism just to save some repeated lines.

I also don't think every policy must be forced into a shared file forever. If something genuinely becomes large or special enough to deserve its own file, that's perfectly fine. I just wouldn't make one directory + two files per small policy variation the default structure.

So from my side, option 2 is the most sensible direction.

SlawomirNowaczyk:
Thanks, @fl0rianr, option 2 it is, then.
@QuTosh, can you make the final commit, and then we'll merge this? Thanks :)

fl0rianr (commented):
Thanks for going for the new layout and some fixes. Nice!

I still see some things:

With multiple policies now living in one policies.json, I think we should reject duplicate JSON object keys. nlohmann::json::parse() does not reject them, so accidentally defining the same policy key twice can silently lose one of them.

Also, the Decision comparer now applies the 1e-12 tolerance to every trace score. The comment still says this should only apply to semantic_similarity, and I think that was the better behavior. Classifier/LLM scores come directly from the stubs and should stay exact. result being exact protects the routing outcome, but not backwards compatibility of trace.score itself.

The schema-invalid metadata-non-string-value-dropped case is also still present; I'd still prefer moving that out of the golden corpus.

Finally a small nit: the PR description needs an update for the new layout and currently contains some duplicated/broken text.

QuTosh:
@SlawomirNowaczyk @fl0rianr 

I've restructured the corpus and updated the test code -- both to match the new structure and to improve it's clarity. Please take a look.

If there are no big changes after that I'd suggest we merge the corpus without freezing and leave it like that for some time -- I imagine some problems might  pop up during future updates of the main code and its unclear if we are to add or modify the cases.

fl0rianr (commented):
Again better but still with issues:

The duplicate-key check currently only catches duplicate top-level policy names. Duplicate keys inside a policy or inside a cases.jsonl row are still silently accepted by nlohmann (the self-test even explicitly allows this). I think a conformance corpus should reject duplicate object keys everywhere. Otherwise something like a duplicated default_model, rules, decision etc. field can silently overwrite the first value.

There is also still a small hole in the strict layout check: files directly under `test/conformance/routing/` are all ignored as documentation. I think only README.md should be allowed there; otherwise an accidentally misplaced corpus file can be silently skipped.

The following topic are still open from above:
- exact comparison for classifier/LLM trace scores, the 1e-12 tolerance topic
- invalid metadata adaption
- PR description

QuTosh:
@fl0rianr , thanks for the comments

* duplicate-key check -- done

* `README.md`only check 0in `test/conformance/routing/` -- done

* exact comparison for classifier/LLM trace scores, the 1e-12 tolerance topic -- I think I accidentally changed it to a more blanked version. I made it exact for classifier/LLM traces. However the code got somewhat harder to read

* invalid metadata adaption -- I have removed that case. If we want it covered, I suggest adding it in a separate PR

* PR description -- updated

fl0rianr (commented):
Thanks, the previous points look addressed now.

I found one remaining layout issue: find_tier_dirs() accepts any directory name below a schema version and doesn't verify that the expected tiers are actually present.

For v1, renaming l3 to e.g. l33, or removing l3 completely, would not be rejected by the layout guard. The remaining tiers could still run successfully.

Since the runner is supposed to prevent corpus coverage from disappearing silently, I think it should enforce the expected tier set (l0a, l1, l2, l3) for v1, or have an equivalent explicit manifest.

Small non-blocking robustness nit: compare_trace_entry() uses expected.value("condition", std::string()) before validating its type, so a malformed golden trace such as "condition": 123 throws instead of being reported as a normal mismatch.

Issue #2858, opened 2026-07-30
feat: Backends Performance Leaderboard/QA Tracking for performance


iswaryaalex:
### Problem

  AMD maintains multiple forks of llama.cpp and other backends (ROCm stable, ROCm nightly, CUDA, Vulkan, vLLM ROCm)
  targeting different hardware optimizations. There is currently no systematic way to:

  - Detect performance regressions when a backend version is bumped in backend_versions.json
  - Compare TPS/TTFT/VRAM across forks for a given model and hardware
  - Verify that a faster build hasn't silently degraded output accuracy
  - Publish results publicly so the community can choose the right backend

### Proposed Solution

  A 2-surface system:

  1. CLI — lemonade bench --fork-matrix
  Extend the existing lemonade bench command with a --fork-matrix flag pointing to a new benchmark_forks.json registry.
  The runner iterates over each tracked fork, installs the binary, runs the scenario matrix, and emits a unified JSON
  result file. No new binary — extends what already exists in src/cpp/cli/bench.cpp.

2. CI/CD Regressions
    This is the crux if this work. Before key releases, need to ensure we are pointing to the most performant/accurate backends after clear
    QA testing and validation. We will run nightlies on selected Hot models across the tracked backends, looking for latest build releases    and running benchmark tests against known best
 
 

  #### Dimensions tracked per data point

  Model × Backend × Build/Version × Hardware → TPS + TTFT + VRAM + Accuracy (MMLU/perplexity)
  Perf runs nightly. Accuracy runs weekly.

### Flow 

<img width="793" height="474" alt="Image" src="https://github.com/user-attachments/assets/1f0d0048-c3b5-48e9-8374-6d03787691fc" />

### Schema

` benchmark_forks.json`

  ```
{
    "forks": [
      {
        "fork_id":        "llamacpp-rocm-stable",
        "label":          "llama.cpp ROCm Stable",
        "recipe":         "llamacpp",
        "backend":        "rocm-stable",
        "repo":           "lemonade-sdk/llama.cpp-rocm",
        "version":        "b9752",
        "binary_url":     "https://github.com/lemonade-sdk/llama.cpp/releases/download/b9752/ll
  ama-b9752-win-rocm-x64.zip",
        "binary_sha256":  "a3f4c8...",
        "binary_exe":     "llama-server.exe",
        "hardware_filter": [],
        "tracked":        true,
        "status":         "active"
      },
      {
[...]

PR #2866, opened 2026-07-31
[bench] Refactor & add hardware info

bitgamma:
This refactors the bench tool to remove some duplication, reduce the number of server calls and split the code across multiple files. It also adds a `hardware` section to the output, so that the results are more useful when shared and compared.

example of the hardware section:

```
  "hardware": {
    "backends": {
      "llamacpp/rocm": "b9752",
      "llamacpp/vulkan": "b9747"
    },
    "cpu": "AMD Ryzen AI 9 HX 370 w/ Radeon 890M",
    "gpu": [
      {
        "integrated": true,
        "name": "110500",
        "type": "amd",
        "vram_gb": 0.5
      }
    ],
    "os": "linux",
    "ram_gb": 61.91
  },
```

### DIFF
--- a/src/cpp/cli/CMakeLists.txt
+++ b/src/cpp/cli/CMakeLists.txt
@@ -22,6 +22,9 @@ set(COMMON_SOURCES
     recipe_import.cpp
     hf_pull.cpp
     bench.cpp
+    bench_output.cpp
+    bench_comparison.cpp
+    bench_sysinfo.cpp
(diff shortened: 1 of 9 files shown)


sofiageo (approved):
Tested both benchmark with hardware support and benchmark comparison. Works fine. I'm not sure if it fulfills the requirements for #2858, maybe Alex can address that.

PR #2867, opened 2026-08-01
Moving from Quick Rules/ Advanced Rules to Natural Language Router now gives a warning that the progress will be cleared

### LINKED ISSUE #2859: Moving from Quick Rules/ Advanced Rules to Natural Language Router removes the rules without any warning
### Platform

Windows

### Lemonade Version

11.5

### GPU / APU Model

_No response_

### Component

Desktop App / UI

### Bug Description

After importing a JSON file in the Lemonade app, the rules initially appear under Quick Rules. However, if I navigate to the Natural Language section and then return to Quick Rules, the rules disappear.

We currently display a warning message when a user switches from Quick Rules to Advanced Rules, informing them that their progress will be lost. A similar warning banner should be displayed when switching from Quick Rules or Advanced Rules to Natural Language.

### Steps to Reproduce

1. Define rules in the Quick Rules or Advanced Rules section.
2. Navigate to the Natural Language section.
3. Observe that no warning message is displayed before switching.
4. Return to the Quick Rules or Advanced Rules section and notice that the previously defined rules are lost.

_No response_

### Log Output

```shell

```

### Additional Context

_No response_

meghsat:
Fixes #2859

### DIFF
--- a/src/app/src/renderer/components/RouterCollectionPanel.tsx
+++ b/src/app/src/renderer/components/RouterCollectionPanel.tsx
@@ -871,7 +871,7 @@ const RouterCollectionPanel: React.FC<RouterCollectionPanelProps> = ({
             <label className={`router-mode-option${draft.routingMode === 'llm' ? ' router-mode-option--selected' : ''}`}>
               <input type="radio" name="routingMode" value="llm" checked={draft.routingMode === 'llm'}
                 onChange={() => {
-                  if ((draft.routingMode === 'quick' || draft.routingMode === 'rules') && (draft.rules ?? []).length > 0) {
+                  if ((draft.routingMode === 'quick' || draft.routingMode === 'rules') && ((draft.rules ?? []).length > 0 || (draft.classifiers ?? []).length > 0)) {
                     setConfirmLlmTarget(null);
                     setConfirmLlm(true);
                   } else {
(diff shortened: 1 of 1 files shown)


fl0rianr (commented):
The warning logic does not cover all data-loss cases yet.

In Advanced Rules, users can create classifiers without creating any rules. The new condition only checks `draft.rules.length`, but switching to Natural Language then clears both `rules` and `classifiers`. This means classifier-only progress is still deleted without a warning.

Please include classifiers in the progress check, for example:

```tsx
const hasRuleProgress =
  (draft.rules?.length ?? 0) > 0 ||
  (draft.classifiers?.length ?? 0) > 0;
```

Also, the reverse-direction check currently treats the unchanged `DEFAULT_ROUTER_PROMPT` as user progress, so a brand-new router shows a destructive-change warning when switching away from Natural Language even though nothing was configured. It would be better to only warn when a router model is selected or the prompt differs from the default.

sdevinenamd:
@fl0rianr  Thanks for pointing them. Fixed both the issues in my latest commit.

fl0rianr (commented):
The Natural Language progress check is still inconsistent between the two target modes.

For **Natural Language → Quick Rules**, the condition only checks whether the prompt differs from `DEFAULT_ROUTER_PROMPT`, so selecting a router model while keeping the default prompt is treated as no progress and the model is cleared without a warning.

For **Natural Language → Advanced Rules**, the condition checks `draft.routerPrompt?.trim()`, which is already truthy for the unchanged default prompt. As a result, a completely untouched Natural Language router shows a destructive-change warning.

Please use the same progress check for both transitions, for example:

```tsx
const hasNaturalLanguageProgress =
  Boolean(draft.routerModel) ||
  Boolean(
    draft.routerPrompt?.trim() &&
    draft.routerPrompt.trim() !== DEFAULT_ROUTER_PROMPT
  );
```

Then use `hasNaturalLanguageProgress` for both the Quick Rules and Advanced Rules transitions.

sdevinenamd:
@fl0rianr added the fixes

fl0rianr (approved):
LGTM now

Please do NOT merge, leave this to me:

Why: 
GUI3 will merge soon, if we merge this it will generate conflicts and will also prevent the beta CI from running.
So I would try merging this when it's the right timing for me. And I have to check if that is fixed in GUI3 was well.

PR #2868, opened 2026-08-01
fix(gui3): polish model configuration and monitor UI

fl0rianr:
## Summary

* Synchronizes model configuration with effective chat context settings.
* Adds known TTS voice choices with a custom fallback.
* Refines compact backend labels and context-size controls.
* Stabilizes the titlebar and renames Dashboard to Monitor while preserving existing routes.
* Fixes search/tool-menu accessibility regressions and updates the related tests.
* Completes the external MCP connection flow and restores Streamable HTTP compatibility.
* Handles server-initiated MCP requests without confusing them with pending client responses.
* Invalidates expired MCP HTTP sessions so subsequent requests can establish a fresh session.
* Prevents stale model-list requests from being reused after model registry mutations.
* Refines the model/backend badge with integrated readiness status and pin controls.

## Validation

* Configuration consistency contracts pass.
* Model-list backend layout contracts pass.
* UI regression contracts pass.
* Chat/audio, global settings, and storage contracts pass.
* Model-state deduplication and read-after-write freshness checks pass.
* MCP C++/httplib compatibility checks pass.
* External MCP test coverage now includes server-initiated requests and expired-session recovery.


fl0rianr:
I think this design provides more value than a conventional backend badge.

The status light communicates whether the backend of the downloaded model is actually ready, missing, or requires an update. Visually connecting the backend badge to the model also makes their dependency immediately understandable: the model is only ready when its backend is available.

The integrated pin follows the same logic, since pinning keeps the backend process associated with that model running. Combining these elements creates a compact and coherent control instead of three unrelated UI indicators.

It is more space-efficient, easier to scan, and gives GUI3 a distinctive design language without adding visual clutter.

image
image
image

> [!CAUTION]
> **DESIGN OUTDATED**

fl0rianr:
image

jeremyfowers:
The models tab will still need more ui/ux optimization, but whats shown here looks like progress to me and I'm good with merging it.

kpoineal:
@fl0rianr I took a deeper look at this. The targeted GUI/runtime regression tests pass, but I found three correctness issues that I think should be addressed before merge:

1. **Expired Streamable HTTP sessions remain wedged** (`src/cpp/server/mcp_client.cpp:680-685, 2029-2057, 2090-2092`). A 404 for an expired `Mcp-Session-Id` becomes a generic error without clearing the cached session or connected state. MCP 2025-06-18 requires starting a new session without that ID; subsequent calls currently keep reusing the invalid session.
2. **Server-initiated SSE requests are mishandled** (`src/cpp/server/mcp_client.cpp:518-530, 622-628`). Messages are identified as responses solely by matching `id`. A server request such as `ping` can therefore be mistaken for the awaited response, while requests with other IDs are discarded rather than answered.
3. **Model request coalescing can violate read-after-write freshness** (`src/app/src/api.ts:1321-1349`, `src/app/src/components/ModelManager.tsx:2140-2152`). Post-registration verification can join a `/models` request that began before the final write, then falsely report that the collection was not persisted.

One lower-priority interoperability gap is that resumable SSE streams are unsupported (`mcp_client.cpp:437-497, 640-708`): event IDs are ignored and there is no GET/`Last-Event-ID` resume path. This behavior is optional in MCP 2025-06-18, so I would not block solely on it.

The current C++ compatibility test mainly checks source patterns; native behavioral coverage for session expiry and server-initiated requests would help prevent regressions.

fl0rianr:
Thanks, good catches, please have a look again.

kpoineal (approved):
Re-reviewed after the MCP session, SSE request-handling, and model-refresh fixes. The previous blockers are resolved, the affected contract and Playwright tests pass, and I found no new significant issues.

PR #2871, opened 2026-08-01
FIX(ci-test-router): recover semantic routing after transient embed failure

fl0rianr:
## Summary

Stabilizes the semantic router tests by making the embedding-model setup deterministic.

* Loads `nomic-embed-text-v1-GGUF` with `ctx_size=2048` on the CPU backend.
* Verifies readiness with a real `/embeddings` request before testing routing.
* Avoids persisting CI-specific load options.
* Unloads the embedding model during cleanup.
* Keeps routing failures visible instead of masking them with broad retries.

## Validation

The updated `router-onnxruntime` CI job successfully loads the embedding model, completes embedding inference, and runs the semantic routing tests.

### DIFF
--- a/test/server_router.py
+++ b/test/server_router.py
@@ -18,6 +18,7 @@ Usage:
 
 import json as _json
 import threading
+import time
 from http.server import BaseHTTPRequestHandler, HTTPServer
 
 import requests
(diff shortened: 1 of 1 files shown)


ramkrishna2910 (approved):
Verified against the failing run in #2883: the flake is the semantic embed model (nomic-embed-text-v1-GGUF) failing to preload via llama-server, and this makes that load deterministic (pinned cpu backend, ctx_size, no persisted options) with a readiness probe. The onnxruntime classifier path in test_630 is untouched and still runs on the onnxruntime backend.

PR #2876, opened 2026-08-02
fix(server): keep interrupted variantless model downloads resumable

fl0rianr:
### Summary

- Resolve variantless registry checkpoints only from the snapshot committed by
  `refs/main`.
- Treat a modern `snapshots/` cache without a usable active ref as incomplete
  instead of exposing the repository cache root as a downloaded model.
- Preserve the legacy flat-cache fallback and all explicit-variant behavior.
- Add regression coverage for restart after interruption, successful commit,
  and an interrupted update beside a still-valid committed snapshot.

### Root cause

The shared backend resolver returned `model_cache_path` for checkpoints without
an explicit variant. An interrupted download already creates that directory, so
a cache rebuild after restart could mark the model as downloaded even though its
snapshot still contained `.partial` files and `.download_manifest.json`.

The registry downloader writes `refs/main` only after all selected artifacts
have completed and passed validation. Using the active ref as the directory
checkpoint commit marker aligns status resolution with the existing download
transaction.

### Behavior after this change

- First interrupted pull: model remains not downloaded; the next pull resumes
  the preserved partial files.
- Completed pull: model resolves to the committed active snapshot.
- Interrupted update: the previous committed snapshot remains usable until the
  new snapshot completes.
- Legacy flat caches: unchanged.
- Explicit variants and GGUF shared-repo resolution: unchanged.

### Validation

- Added `test_uncommitted_variantless_snapshot_is_not_downloaded` to
  `test/test_multi_checkpoint_completeness.py`.

### DIFF
--- a/src/cpp/server/backends/backend_ops.cpp
+++ b/src/cpp/server/backends/backend_ops.cpp
@@ -97,7 +97,17 @@ std::string BackendOps::resolve_checkpoint_path(const ModelInfo& info,
         return "";
     }
 
-    // No variant: return the cache directory.
+    fs::path active_snapshot = hf_cache::active_snapshot_path(model_cache_path_fs);
+    if (!active_snapshot.empty()) {
+        return path_to_utf8(active_snapshot);
+    }
+
+    // A snapshots directory without a usable refs/main is an interrupted or
+    // otherwise uncommitted pull. Do not expose the cache root as a downloaded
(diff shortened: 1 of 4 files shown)


bitgamma (approved):
LGTM

fl0rianr:
Really strange, PR Queue rejected the first trial even without an single check failing...

PR #2877, opened 2026-08-02
New helper combining add_test() and register_cpp_ci_test()

SlawomirNowaczyk:
## Summary

Make the CI decision explicit for every C++ test. `add_cpp_ci_test()` now takes a required `CI <ON|OFF>` argument and a `COMMAND` (with optional `DEPENDS`), so every test states its CI status at the call site. Direct `add_test()` is disabled (overridden to fatal-error after third-party deps are configured) to prevent tests from silently bypassing packaging CI. All existing tests were migrated to the new helper with their current CI status preserved.

## What changed

**CMakeLists.txt**
- Added `add_cpp_ci_test()` with a keyword interface with an explicit, required CI decision:
  ```cmake
  add_cpp_ci_test(<TestName> CI <ON|OFF> COMMAND <command> [args...] [DEPENDS <target>...])
  ```
  `CI ON` labels the test `cpp-ci` and wires its build target into the `cpp-ci-tests` aggregate; `CI OFF` still creates the CTest test but keeps it out of packaging CI. `DEPENDS` defaults to the first `COMMAND` token.
- **Disabled direct `add_test()`**: the built-in is overridden with a function that raises a `FATAL_ERROR`, installed right after all third-party dependencies are configured (so vendored suites like libwebsockets are unaffected). The helper calls the preserved `_add_test` so it stays exempt.
- Converted **all** existing tests (executable tests + the `HttplibDetection_*` CMake tests) from `add_test()` / `register_cpp_ci_test()` pairs to the single `add_cpp_ci_test(... CI ON/OFF ...)` interface, preserving each test's current CI status.

**AGENTS.md** — updated the "C++ unit tests" section to document the explicit `CI <ON|OFF>` interface and that direct `add_test()` is disabled.

### Tests kept excluded from CI (`CI OFF`) — status quo, complete list

Executable tests (19):
- `vllm_arg_resolver`
- `llamacpp_request`
- `FlmArgResolverTest`
- `GgufCapabilitiesTest`
- `InstallAtomicityTest`
- `RouteDecisionResponseTest`
- `ModelRegistryTest`
- `AutoTuneTest`
- `TelemetryHelpersTest`
- `SuspendInhibitorTest`
- `ConfigMigrationTest`
- `ConfigTelemetryTest`
- `RocmRootResolutionTest`
- `BackendUtilsGpuEnvTest`
- `UrlUtilsTest`
- `VllmCdnaGatingTest`
- `LlamacppGfx950GatingTest`
- `HttpClientSecurityTest`
- `CloudDiscoveryPolicyTest`

CMake httplib-detection tests (13):
- `HttplibDetection_config_only`
- `HttplibDetection_old_header`
- `HttplibDetection_modern_header`
- `HttplibDetection_openssl_target`
- `HttplibDetection_disabled_generator_expression`
- `HttplibDetection_too_old_config`
- `HttplibDetection_compiled_no_tls`
- `HttplibDetection_too_old_config_multiarch`
- `HttplibDetection_header_only_pkgconfig`
- `HttplibDetection_cli_composition_mbedtls_isolation`
- `HttplibDetection_custom_httplib_dir_precheck`
- `HttplibDetection_polluted_target_fatal` (also `WILL_FAIL`)
- `HttplibDetection_wolfssl_target`

Total: **32 tests excluded** from `ctest -L cpp-ci` (unchanged from before this PR).

Note: a few of these are additionally platform/existence-gated (e.g. the GPU/ROCm/config tests), but their CI status is `OFF` regardless of platform.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -139,6 +139,33 @@ python test/server_sd.py
 
 Test utilities in `test/utils/` with `server_base.py` as the base class. Test dependencies include `requests`, `httpx`, `openai`, `huggingface_hub`, `psutil`, `numpy`, `websockets`, and `ollama`.
 
+### C++ unit tests
+
+C++ unit tests live in `test/cpp/` and are wired up in the root `CMakeLists.txt`. The packaging workflow builds the `cpp-ci-tests` aggregate target and runs `ctest -L cpp-ci`, so a test only runs in CI if it is both labeled `cpp-ci` **and** a dependency of that aggregate target.
+
+**Direct `add_test()` is disabled** (the built-in is overridden to fail with a fatal error just before the test section). Every test MUST be declared with the `add_cpp_ci_test()` helper, which forces an explicit `CI <ON|OFF>` decision at the call site so a test is never silently omitted from — or accidentally added to — CI:
+
+```cmake
+add_executable(test_my_feature test/cpp/test_my_feature.cpp ...)
+# ...target_include_directories / target_link_libraries...
+
+include(CTest)
+add_cpp_ci_test(MyFeatureTest CI ON COMMAND test_my_feature)
+```
+
+```cmake
+add_cpp_ci_test(<TestName>
+                CI <ON|OFF>                 # required — run under `ctest -L cpp-ci`?
+                COMMAND <command> [args...] # required — what CTest runs
+                [DEPENDS <target>...])      # CI build deps; defaults to the
+                                            # first COMMAND token (the test target)
+```
+
+- `CI ON` labels the test `cpp-ci` and makes its build target(s) a dependency of `cpp-ci-tests`.
+- `CI OFF` still creates the CTest test (for local/other runs) but keeps it out of packaging CI. Use this only for tests that are intentionally excluded (e.g. tests that need a backend, are platform-gated, or are slow CMake-configuration tests).
+
+Pass `DEPENDS` only when the CI build needs targets beyond the `COMMAND` executable. `add_cpp_ci_test` calls `register_cpp_ci_test()` internally; do not call `add_test()` or `register_cpp_ci_test()` directly.
+
 ## Code Style
 
 ### Comments & Documentation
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
This is a nice follow-up to #2829. Combining `add_test()` and `register_cpp_ci_test()` into one explicit helper removes the remaining two-step registration risk, and requiring every test to state `CI ON` or `CI OFF` makes the intended CI coverage much clearer.

One small issue before I would approve: the helper currently only checks whether `CI` was provided, but does not validate that its value is actually `ON` or `OFF`. For example, a typo such as `CI OF` would be treated as truthy by CMake and silently add the test to CI, which undermines the explicit-decision guarantee this PR is aiming for.

Could we validate the normalized value explicitly, for example:

```cmake
string(TOUPPER "${ACT_CI}" _act_ci)
if(NOT _act_ci STREQUAL "ON" AND NOT _act_ci STREQUAL "OFF")
    message(FATAL_ERROR
        "add_cpp_ci_test(${test_name}): CI must be exactly ON or OFF")
endif()
```

Optionally, rejecting `ACT_UNPARSED_ARGUMENTS` would also help catch misspelled keywords.

With strict `ON|OFF` validation and green packaging CI, this looks approvable to me.

SlawomirNowaczyk:
Good catch, now it rejects typos

fl0rianr (approved):
LGTM from my side now. Thanks for addressing the validation issue.

I see two non-blocking follow-up topics:

1. The new section in `AGENTS.md` feels somewhat verbose for what is ultimately a fairly small helper API. I think it could be shortened while keeping the essential usage and `CI ON/OFF` semantics clear.

2. Quite a few tests are now explicitly marked `CI OFF`. Preserving the existing behavior is the correct choice for this PR, and some exclusions are clearly justified, for example platform-specific tests. However, several tests appear to have simply been added without ever being connected to CI.

This is not a problem introduced by this PR, but the new helper makes the existing gap much more visible. It would be worth auditing the `CI OFF` tests separately and enabling those that run reliably in the packaging CI. If the current CI is already green, adding the suitable tests should provide useful coverage without much additional risk.

PR #2881, opened 2026-08-02
fix(docs): no support for llama.cpp:cpu, metal is used

fl0rianr:
Clarify the documented llama.cpp backend behavior on macOS.

Lemonade uses the Metal-enabled llama.cpp build as the unified macOS runtime. The same build can also run CPU-only when GPU offload is disabled, so a separate `llamacpp:cpu` backend selection is not required.

This update:

* removes macOS from the standalone CPU backend platform list
* clarifies CPU-only execution through the Metal-enabled build
* removes duplicate macOS guidance from the backend decision tree
* keeps macOS support documented without implying a Lemonade limitation

Documentation only; no runtime behavior changes.

### DIFF
--- a/docs/guide/configuration/llamacpp.md
+++ b/docs/guide/configuration/llamacpp.md
@@ -5,9 +5,13 @@ Lemonade uses [llama.cpp](https://github.com/ggerganov/llama.cpp) as its primary
 ## Available Backends
 
 ### CPU
-- **Platform**: Windows, Linux, macOS
+> **macOS:** Lemonade uses the Metal-enabled llama.cpp build as the unified
+> runtime. The same build can run CPU-only when GPU offload is disabled, so a
+> separate `llamacpp:cpu` backend is not required on macOS.
+
+- **Platform**: Windows, Linux
 - **Hardware**: x86_64 processors (Windows and Linux); ARM64/aarch64 processors (Linux)
-- **Use Case**: Universal fallback, no GPU required
+- **Use Case**: CPU-only execution and universal fallback on Windows and Linux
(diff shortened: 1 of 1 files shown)


docs/guide/configuration/llamacpp.md
```diff
@@ -176,11 +181,12 @@ lemonade config set llamacpp.rocm_bin=b1260
...
 
-4. **Do you have Apple Silicon?**
+4. **Are you using macOS?**
    - Use **Metal**
+   - The Metal-enabled build can also run CPU-only when GPU offload is disabled; no separate CPU backend selection is required
 
-5. **No GPU or unsupported GPU?**
-   - Use **CPU**
+5. **No supported GPU backend available?**
+   - On Windows and Linux, use **CPU**
```
Geramy: I'm not sure what this means it's actually confusing, 5. is referring to what exactly? it says no supported gpu backend available? on windows and linux so use cpu? but what's that even mean, are we specifying Mac hardware doesn't work on windows or linux? if so I dont think we even should say that, its confusing.

Geramy (changes requested):

docs/guide/configuration/llamacpp.md
fl0rianr: I agree, my slip. Just use the old one is fine?

Geramy: Yeah that's probably best, and no problem, sometimes its confusing because when you do a review on git and if the document order is not perfect for that scenario your always missing context haha but yeah set it to the original I feel that's better, less confusing.

Geramy (approved):

PR #2891, opened 2026-08-04
docs: add testing guide

jeremyfowers:
## Summary
- Add `docs/dev/testing.md`: where tests go, how to run them locally, and what reviewers expect.
- Note in `documentation.md` that `docs/` is published via the mkdocs site (lemonade-server.ai/docs) and new pages must be added to `mkdocs.yml`'s nav.
- Register the new testing guide in `mkdocs.yml` and link it from `dev/README.md` and `dev/contribute.md`.

## Test plan
- [ ] Confirm `testing.md` renders correctly on a local mkdocs/zensical serve
- [ ] Confirm internal links resolve

### DIFF
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -26,6 +26,10 @@ The Lemonade project welcomes contributions! Learn about the project's mission, 
 
 Writing or improving docs? Read the [documentation guide](./documentation.md) for style, structure, and guidance on AI-assisted contributions.
 
+### Testing
+
+Preparing a PR? Read the [testing guide](./testing.md) to learn which test suites cover your change and what reviewers expect.
+
 ### Lemonade Omni Models
 
 Lemonade has a unique capability to group LLM, image, and speech models together to present a unified omni-modal "model" to end-users. These one-click bundles are called Lemonade Omni Models, and they're routed via an internal mechanism called OmniRouter. Learn more [here](./lemonade-omni.md).
--- a/docs/dev/contribute.md
+++ b/docs/dev/contribute.md
@@ -36,7 +36,7 @@ Issues are a great way to document a bug or feature request. However, Lemonade i
 Each contribution needs to:
 
 1. Adhere to the [philosophy](./philosophy.md).
-2. Pass the CI tests.
+2. Pass the CI tests and follow the [testing guide](./testing.md).
     - Contributors: make sure the code builds locally before creating the PR.
     - Reviewers: make sure to check the code *before* allowing CI to run!
 3. Sustain the overall code quality and standards of the rest of the repo.
--- a/docs/dev/documentation.md
+++ b/docs/dev/documentation.md
@@ -4,7 +4,7 @@ Lemonade's documentation is a first-class part of the product. Per our [philosop
 
 This guide covers style, structure, and the contribution process for both community authors and AI-assisted contributions.
 
-Documentation under `docs/` is built with [Zensical](https://zensical.org/) (configured via [`mkdocs.yml`](../../mkdocs.yml), whose schema it inherits from MkDocs Material) into the site hosted at [lemonade-server.ai/docs](https://lemonade-server.ai/docs). Add new documents to the `nav` section of `mkdocs.yml`: pages left out of `nav` are still built and reachable by URL, but don't appear in the site's navigation sidebar. To preview locally, `pip install -r docs/assets/docs_requirements.txt` and run `zensical serve`.
+Documentation under `docs/` is built with [Zensical](https://zensical.org/) (configured via [`mkdocs.yml`](https://github.com/lemonade-sdk/lemonade/blob/main/mkdocs.yml), whose schema it inherits from MkDocs Material) into the site hosted at [lemonade-server.ai/docs](https://lemonade-server.ai/docs). Add new documents to the `nav` section of `mkdocs.yml`: pages left out of `nav` are still built and reachable by URL, but don't appear in the site's navigation sidebar. To preview locally, `pip install -r docs/assets/docs_requirements.txt` and run `zensical serve`.
 
 - [Principles](#principles)
 - [Voice and Tone](#voice-and-tone)
--- a/docs/dev/documentation.md
+++ b/docs/dev/documentation.md
@@ -4,7 +4,7 @@ Lemonade's documentation is a first-class part of the product. Per our [philosop
 
 This guide covers style, structure, and the contribution process for both community authors and AI-assisted contributions.
 
-Documentation under `docs/` is built into the [mkdocs](https://www.mkdocs.org/) site hosted at [lemonade-server.ai/docs](https://lemonade-server.ai/docs). New documents must be added to the `nav` section of [`mkdocs.yml`](../../mkdocs.yml), or they won't appear on the site.
+Documentation under `docs/` is built with [Zensical](https://zensical.org/) (configured via [`mkdocs.yml`](../../mkdocs.yml), whose schema it inherits from MkDocs Material) into the site hosted at [lemonade-server.ai/docs](https://lemonade-server.ai/docs). Add new documents to the `nav` section of `mkdocs.yml`: pages left out of `nav` are still built and reachable by URL, but don't appear in the site's navigation sidebar. To preview locally, `pip install -r docs/assets/docs_requirements.txt` and run `zensical serve`.
 
 - [Principles](#principles)
 - [Voice and Tone](#voice-and-tone)
--- a/docs/dev/documentation.md
+++ b/docs/dev/documentation.md
@@ -4,6 +4,8 @@ Lemonade's documentation is a first-class part of the product. Per our [philosop
 
 This guide covers style, structure, and the contribution process for both community authors and AI-assisted contributions.
 
+Documentation under `docs/` is built into the [mkdocs](https://www.mkdocs.org/) site hosted at [lemonade-server.ai/docs](https://lemonade-server.ai/docs). New documents must be added to the `nav` section of [`mkdocs.yml`](../../mkdocs.yml), or they won't appear on the site.
+
 - [Principles](#principles)
 - [Voice and Tone](#voice-and-tone)
 - [Document Structure](#document-structure)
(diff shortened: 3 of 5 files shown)


fl0rianr (changes requested):
Thanks for writing this — a central testing guide is valuable, and many of the principles here are heading in the right direction.
However, I think one important principle needs to be made much more explicit:</p><h2>Tests must be reviewed as critically as production code</h2><p>A test does not validate itself. A passing test only proves that the implementation and the assertion agree. If the test reproduces the same incorrect assumption as the production code, asserts only a weak structural property, accidentally exercises a cached/default path, or depends on timing and shared state, both can be wrong together while CI remains green.</p><p>This is particularly relevant for Lemonade. We have recently needed dedicated fixes for several independent classes of test instability. I suggest adding a short <strong>“Review the test itself”</strong> section with expectations such as:</p><ul><li><p><strong>Prove regression sensitivity.</strong> A regression test should fail on the pre-fix code or after a small targeted mutation that reintroduces the bug. Merely adding a test that passes on the branch is not enough.</p></li><li><p><strong>Use an independent oracle.</strong> Do not duplicate the production algorithm inside the test and compare the two implementations. For numerical or model-dependent behavior, use an independently produced golden reference, protocol invariant, or measured separation with a documented safety margin.</p></li><li><p><strong>Assert the behavior that matters.</strong> Ask whether the test would still pass if the server returned a placeholder, default model response, stale cached result, empty-but-valid object, all-zero counters, or the wrong backend. “Non-empty” is appropriate only where content is genuinely nondeterministic and stronger stable invariants do not exist.</p></li><li><p><strong>Verify the production path is actually exercised.</strong> A test should not mock away the changed code. Reviewers should confirm that capability gates do not silently skip it, the intended CI matrix row really invokes it, and the assertion occurs after the relevant operation has completed.</p></li><li><p><strong>Require isolation and order independence.</strong> Tests should not depend on another test having loaded a model, populated a cache, changed an environment variable, released a port, or cleaned server state. Temporary directories, processes, ports, environment variables, models, and configuration must be owned and cleaned up by the test.</p></li><li><p><strong>Treat flakiness fixes as correctness work.</strong> A rerun that turns green is evidence of nondeterminism, not evidence that the original failure was harmless. Sleeps, wider thresholds, retries, and longer timeouts should be treated as review smells.</p></li><li><p><strong>Allow retries only at identified transient boundaries.</strong> Bounded retries can be valid for startup readiness, network transport, or an explicitly asynchronous API, but they must not wrap semantic assertions or conceal a persistent product error. Each retry should retain enough diagnostics to identify the original failure.</p></li><li><p><strong>Keep external infrastructure out of required correctness gates where possible.</strong> A third-party rate limit, unavailable model host, or mutable remote artifact should not look like a Lemonade regression. Required tests should use local fixtures, mocks, pinned artifacts, or controlled mirrors; real-network smoke tests should clearly classify external failures and normally run separately.</p></li><li><p><strong>Make failures diagnostic.</strong> Assertions should report the actual backend, model, route, score, threshold, state, response body, or process status needed to distinguish a product bug from a test-environment problem.</p></li></ul><p>A compact reviewer checklist could be:</p><ol><li><p>What exact regression causes this test to fail?</p></li><li><p>Has that failure been demonstrated?</p></li><li><p>Could a stub, fallback, cache, or skipped path still pass it?</p></li><li><p>Is the expected result independent of the implementation under test?</p></li><li><p>Is the test deterministic, isolated, and order-independent?</p></li><li><p>Does CI demonstrably execute the test on the relevant platform/backend?</p></li></ol><p>I would also add corresponding entries to <strong>“What Reviewers Reject”</strong>, for example:</p>
Anti-pattern | Do instead
-- | --
Test reproduces the implementation logic as its expected value | Use an independent oracle or golden reference
Test passes without the changed production path being reached | Assert the path/backend/capability used and verify CI does not skip it
Whole-test retries or reruns until green | Identify and remove the nondeterminism; retry only a proven transient boundary
Threshold placed inside normal numerical/model variance | Measure variance and use a justified safety margin
Required PR test depends on an uncontrolled external service | Use a fixture, mock, mirror, or separate network smoke workflow
Test depends on execution order or state left by another suite | Create and clean all required state within the test

---

<p>There are also several concrete correctness issues in the current text that should be fixed before merge:</p><ul><li><p>The C++ commands require the configure/setup step first, and should preferably build the <code inline="">cpp-ci-tests</code> target explicitly.</p></li><li><p>The app regression command is run from the wrong directory after <code inline="">cd src/app</code>.</p></li><li><p>Pages outside the <code inline="">nav</code> are still built; they are missing from the sidebar rather than absent from the site. The text should also describe the current Zensical-based build accurately.</p></li><li><p>Only server integration suites use <code inline="">ServerTestBase</code> and <code inline="">run_server_tests(...)</code>; pure Python unit, fixture, and schema tests use plain <code inline="">unittest.TestCase</code>.</p></li></ul><p>With those additions, I think this becomes more than a directory map and command reference: it gives contributors and reviewers a useful standard for deciding whether a test is actually trustworthy.</p></body></html>
</body>
</html>

jeremyfowers:
Thanks for the review! The identified problems have been fixed in 4818046 — please re-review:

- C++ test instructions now require the configure step and build the `cpp-ci-tests` target explicitly
- App regression script is now run from the repo root (matching CI)
- `ServerTestBase` claim scoped to server integration suites; self-managed-server and plain-`unittest` suites are called out
- Docs build described accurately: Zensical, and nav-excluded pages are built but unlisted in the sidebar

The following suggestions for improved practices have NOT been integrated yet; please author a PR after this one merges to propose the improvements:

- A "Review the test itself" section: a passing test only proves the implementation and assertion agree
- Retries allowed only at identified transient boundaries (startup, network), never around semantic assertions
- Diagnostic failure output (backend, model, route, score, threshold, response body)
- Measure variance before setting numeric thresholds
- The 6-question reviewer checklist
- Additional "What Reviewers Reject" rows (implementation-as-oracle, order/state dependence, uncontrolled external services, whole-test retries, in-variance thresholds)

One note for that follow-up: full order independence conflicts with the current numbered-test-method convention within suites, so it likely needs to be scoped to cross-suite state rather than within-suite ordering.

fl0rianr (commented):
Thanks — I agree with keeping the larger test-review policy additions for a focused follow-up PR. Your note about scoping order independence to cross-suite state is also important, since some existing suites intentionally use numbered, stateful test sequences.

The original issues are mostly addressed, but I found four small factual problems remaining:

1. `test/server_cli2.py` does **not** launch its own `lemond`. `PersistentServerCLIClientTests.setUpClass()` explicitly requires an already-running server and fails when none is reachable. This is currently stated incorrectly both under “Running Tests Locally” and “Writing a Good Test.” `test/server_jobs.py` is a valid example of a self-managed-server suite; `server_cli2.py` is instead a plain `unittest` suite using a persistent external server.

2. The C++ command block mentions Windows setup but then always uses the `default` Ninja preset. `setup.ps1` configures either the `windows` or `vs18` Visual Studio preset, so following it with `cmake --build --preset default` can conflict with the configured build directory. Please either label the shown commands Linux/macOS-only or add separate Windows commands, for example building `build` with `--config Release` and running `ctest` with `-C Release`.

3. Again, the app regression test path should be `test/app/app-regression/*.test.cjs`, not `test/app/*.test.cjs`. The runner only discovers `.test.cjs` files inside the `app-regression` directory.

4. The wording about Zensical and `nav` is now correct, but the relative `../../mkdocs.yml` link still points outside the published documentation tree. It works in the repository source view but will lead to a missing file on the built site. Please use a GitHub source link or leave `mkdocs.yml` as an unlinked code path.

These are all small documentation corrections. Once addressed, I am happy to approve this PR and move the broader “review the test itself” guidance into the agreed follow-up.

jeremyfowers:
Thanks @fl0rianr — all four fixed:

1. `server_cli2.py` is now documented as expecting a persistent external server (fails fast in `setUpClass`), in both sections; `server_jobs.py` remains the self-managed example.
2. Added separate Windows commands for the C++ tests (`--preset windows` + `ctest -C Release`); the bash block is now labeled Linux/macOS.
3. App regression path corrected to `test/app/app-regression/*.test.cjs`.
4. `mkdocs.yml` now links to the GitHub source instead of a relative path that would 404 on the built site.

Also corrected the Anthropic API row (those tests live in `test/test_ollama.py`, not `server_endpoints.py`) and a couple of smaller accuracy nits found in another verification pass.

fl0rianr (commented):
Thanks — the four issues from the previous pass are now substantially fixed..

On a final accuracy pass, I found three small a bit nitty issues:

1. **Local Test: The Windows command still hardcodes the VS 2022 preset.**
The text correctly explains that `setup.ps1` may configure either `windows` or `vs18`, but the example always runs `cmake --build --preset windows`. That can fail when the build directory was configured with Visual Studio 2026. The generator-independent commands already used by CI would avoid this:

```powershell
cmake --build build --config Release --target cpp-ci-tests
ctest --test-dir build -C Release -L "^cpp-ci$" --output-on-failure
```

2. **`@skip_if_unsupported` does not detect hardware or gate a specific backend.**
It checks whether the configured wrapped server/modality declares support for a feature. Backend-specific gating is a separate concern (`@requires_backend(...)` exists), while the workflow and runner selection provide the actual hardware. Please narrow this bullet so contributors do not assume that the decorator makes a test hardware-safe.

3. **“CI must be green before requesting review” conflicts with `contribute.md`, which asks reviewers to inspect the code before allowing CI to run.** A wording such as the following would fit both workflows:

> Relevant local tests should pass before requesting review. All required CI must be green before final approval and merge.

These are small documentation corrections, not a request to expand the scope again.

jeremyfowers:
Fixed all three in 3d6f74500:

1. Windows commands now use `cmake --build build --config Release --target cpp-ci-tests` (generator-independent, works with either preset).
2. Narrowed the `@skip_if_unsupported` bullet to declared capabilities, not hardware, and added `@requires_backend(...)` for backend-specific gating.
3. Reworded to "Relevant local tests should pass before requesting review. All required CI must be green before final approval and merge." to match contribute.md.

fl0rianr (approved):
LGTM, approval too early according to this new documentation but since it's a documentation only change I reason this is safe.

PR #2892, opened 2026-08-04
ci: wire multi-checkpoint integration smoke into CI

fl0rianr:
This wires the multi-checkpoint (test/test_multi_checkpoint_completeness.py) python smoke test to CI.
Removed the separate test file and integrated this test into server_downloads.py suite.

I consider it a useful test, hence I preferred moving instead of deleting.
Checked locally:

Ran 7 tests in 8.925s
OK (skipped=4)


jeremyfowers:
test/test_multi_checkpoint_completeness.py is not actually wired up. Please make the case for wiring it up (in which case we need this PR), or delete it (and close this PR).

fl0rianr:
Good catch.

I used this port 13306 for some local test and after checking codebase I found this one introduced by on of my tests being merged and wanted to correct it. I only focused on that number.

But this PR predates #3286, which moved the filesystem/download-state coverage into the CI-enabled C++ ModelDownloadStateTest and intentionally kept only one thin Python integration smoke test for CLI propagation and collection fan-out.

So now I don't think the 13306 -> 13305 change is the right fix anymore. The remaining Python test starts its own lemond and consistently uses that port.

I re-scope this PR, test is being run and moved into the existing server_downloads.py suite, and then remove the standalone unwired test file.

PR #2893, opened 2026-08-04
fix(build): patch libwebsockets GENHDR OUTPUT to silence MSB8065

blackdeathdrow:
## Summary

The `add_custom_command(OUTPUT ...)` for libwebsockets' `GENHDR` target lists a bare directory (`${PROJECT_BINARY_DIR}/include/libwebsockets`) alongside actual files. MSBuild, under the Visual Studio generator, interprets that directory as a file path and emits **warning MSB8065** on every build, which disables incremental builds for the LWS subproject.

This removes the directory from the `OUTPUT` list. The directory is still created by the adjacent `copy_directory` COMMAND, so the build is unaffected either way.

Additionally, the upstream `CMakeLists.txt` is checked out with CRLF on Windows (`core.autocrlf = true`, no `.gitattributes` upstream), which made the original `string(REPLACE "... \n" ...)` a silent no-op. A `\r`-strip is now applied to both patched files before any line-based substitution, making the patches reproducible across line-ending configurations.

## Verification

- `cmake -S . -B build` (VS 18 2026 generator) — configures cleanly.
- `cmake --build build --target GENHDR` after deleting `include/lws_config.h` and `include/libwebsockets.h` → no MSB8065, both declared outputs (`lws_config.h`, `libwebsockets.h`) created and the `include/libwebsockets/` directory populated as a side effect.
- `-Werror` strip confirmed (0 occurrences in both `CMakeLists.txt` and `lib/CMakeLists.txt` after patch).

## Upstream

Tracked upstream — no fix released as of v4.3.3:
https://github.com/warmcat/libwebsockets/issues/3060

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -607,6 +607,15 @@ if(NOT USE_SYSTEM_LWS)
     # remove directory from OUTPUT list (MSBuild interprets it as file path — MSB8065)
     string(REPLACE "\t\t\t\${PROJECT_BINARY_DIR}/include/libwebsockets\n" ""
            LWS_CMAKE_MAIN "${LWS_CMAKE_MAIN}")
+    # copy_directory only adds/overwrites — wipe the destination first so stale
+    # headers removed upstream don't survive a GENHDR rerun across a dep update.
+    # Works on every generator (ADDITIONAL_CLEAN_FILES below is Make/Ninja only).
+    if(NOT LWS_CMAKE_MAIN MATCHES "rm -rf.*include/libwebsockets")
+        string(REPLACE
+            "COMMAND \${CMAKE_COMMAND} -E copy_directory \${CMAKE_CURRENT_SOURCE_DIR}/include/libwebsockets/"
+            "COMMAND \${CMAKE_COMMAND} -E rm -rf \${CMAKE_CURRENT_BINARY_DIR}/include/libwebsockets\n\t\tCOMMAND \${CMAKE_COMMAND} -E copy_directory \${CMAKE_CURRENT_SOURCE_DIR}/include/libwebsockets/"
+            LWS_CMAKE_MAIN "${LWS_CMAKE_MAIN}")
+    endif()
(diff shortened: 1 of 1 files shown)


fl0rianr (changes requested):
Thanks! Removing the directory from OUTPUT fixes MSB8065, but it also removes `include/libwebsockets` from the generated clean rules.

This leaves the copied header directory behind after `cmake --build ... --target clean`. Since `cmake -E copy_directory` does not delete files that disappeared from the source tree, obsolete LWS headers can survive a dependency update and later be consumed from the preserved `build/_deps` directory.

Current libwebsockets `main` handles this by adding the directory to the GENHDR target's ADDITIONAL_CLEAN_FILES after removing it from OUTPUT.

Please preserve that behavior here, for example after `add_subdirectory()`:

```cmake
if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.15")
    set_property(
        TARGET GENHDR
        APPEND
        PROPERTY ADDITIONAL_CLEAN_FILES
        "${libwebsockets_BINARY_DIR}/include/libwebsockets"
    )
endif()
```

Please also verify that a second consecutive GENHDR build is a no-op and that `clean` removes the copied header directory.

blackdeathdrow:
Thanks for the review.
Re-attached ADDITIONAL_CLEAN_FILES on GENHDR per your snippet, and added rm -rf before copy_directory in the custom command so stale headers don't survive a dep update (works under MSBuild too, where ADDITIONAL_CLEAN_FILES is ignored).
Second GENHDR run is a no-op; clean removes the dir on both Ninja and VS.

fl0rianr (approved):
Thanks, the previous concern is addressed.

Re-attaching `include/libwebsockets` through ADDITIONAL_CLEAN_FILES restores the clean behavior, while removing the destination before `copy_directory` also prevents stale headers from surviving a later GENHDR run under Visual Studio.

I do not see any remaining blocking correctness issue.

One small non-blocking note: the PR description still says “three declared outputs”. After this change, only the two header files are declared outputs; the `include/libwebsockets` directory is created as a side effect. After the adaption in the text we can merge.

Approving.

PR #2895, opened 2026-08-04
fix(ci): prevent MSI verify flake from stale ARP entries on shared runners

jeremyfowers:
## Root cause

The flaky "Verify Installation" failures (e.g. [omni](https://github.com/lemonade-sdk/lemonade/actions/runs/30916113014/job/92021491020), [audio-gen-thinksound](https://github.com/lemonade-sdk/lemonade/actions/runs/30916113014/job/92021491306)) are machine-state dependent, not random:

1. Runner `sjlab-stx-halo-06` has a stale ARP registry entry ("Lemonade Server 9.0.3") that Windows Installer no longer recognizes — `msiexec /x` on it returns 1605.
2. The cleanup action's uninstall step stops at the **first** DisplayName match, which is always that stale entry, so the actually-installed product is never uninstalled (visible as `Uninstall failed with exit code 1605` in every job on that runner).
3. The next job installs the same MSI (same ProductCode) → MSI log shows `Product registered: entering maintenance mode` → `INSTALLDIR` is ignored in maintenance mode → no files land in the new per-job install path → verify fails with `Missing: bin\LemonadeServer.exe`.

In run 30916113014, `stable-diffusion` installed fine on that runner (nothing registered yet), its teardown hit the stale entry and left the product registered, then `omni` and `audio-gen-thinksound` both failed in maintenance mode.

## Fix

- **cleanup-processes-windows**: uninstall every GUID-keyed registry match instead of only the first, and delete stale ARP keys when msiexec returns 1605 so they stop shadowing real registrations.
- **install-lemonade-server-msi**: uninstall by MSI file (ignoring 1605) before installing, so a leftover registration of the same ProductCode can never push the install into maintenance mode.

### DIFF
--- a/.github/actions/cleanup-processes-windows/action.yml
+++ b/.github/actions/cleanup-processes-windows/action.yml
@@ -212,32 +212,38 @@ runs:
             "HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*"
         )
 
-        $productCode = $null
-        $foundProduct = $null
+        # A single stale ARP entry (uninstallable, exit code 1605) must not shadow
+        # a real registration of the same DisplayName, so collect every GUID-keyed
+        # match across all hives instead of stopping at the first one.
+        $foundProducts = @()
 
         foreach ($path in $uninstallPaths) {
             $subkeys = Get-ItemProperty $path -ErrorAction SilentlyContinue
             foreach ($key in $subkeys) {
-                if ($key.DisplayName -eq $productName) {
-                    $foundProduct = $key
-                    $productCode = $key.PSChildName
-                    break
+                if ($key.DisplayName -eq $productName -and $key.PSChildName -like "{*}") {
+                    $foundProducts += $key
                 }
             }
-            if ($productCode) { break }
         }
 
-        if ($productCode) {
-            Write-Host "Found installed product: $($foundProduct.DisplayName) (Version: $($foundProduct.DisplayVersion))" -ForegroundColor Green
-            Write-Host "Product Code: $productCode" -ForegroundColor Gray
+        if ($foundProducts.Count -gt 0) {
+            foreach ($foundProduct in $foundProducts) {
+                $productCode = $foundProduct.PSChildName
+                Write-Host "Found installed product: $($foundProduct.DisplayName) (Version: $($foundProduct.DisplayVersion))" -ForegroundColor Green
+                Write-Host "Product Code: $productCode" -ForegroundColor Gray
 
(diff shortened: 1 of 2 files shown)


fl0rianr (changes requested):
Thanks for working through this — the overall approach looks sensible to me, especially collecting all matching MSI registrations instead of stopping at the first stale entry. That directly addresses the shadowing case described in the PR. Great!

I noticed two small robustness improvements that may be worth considering:

1. The pre-install uninstall currently continues after any exit code other than `1605`. If it returns something like `1603` or `1618`, the existing registration may still be active and the following install could enter maintenance mode again. Would it make sense to only accept `0` and `1605`, and fail or retry for other codes?

2. Since these jobs run on persistent Windows runners, adding `/norestart` to the `msiexec` install/uninstall calls may help avoid an unexpected runner reboot. `3010` could then be handled explicitly as success-with-reboot-required.

One minor detail: after removing a stale ARP key, it may be useful to use `-ErrorAction Stop` or verify that the key is actually gone before printing the success message.

None of this changes the core direction of the fix, which looks good. These would mainly make the cleanup behavior more deterministic when the runner is already in an unusual MSI state.

jeremyfowers:
Thanks — all three suggestions were right, implemented in bf7e930.

**1. Pre-install uninstall exit codes.** Agreed, and this was the worst of the three: continuing after e.g. 1618 would silently reintroduce the exact bug this PR fixes (install returns 0 in maintenance mode, then verify fails with a confusing `Missing: bin\LemonadeServer.exe`). Now accepts only 0 / 1605 / 3010, retries up to 3× with a 10s gap for transient codes, and on failure dumps the uninstall log and exits 1 rather than installing into maintenance mode.

**2. `/norestart` + 3010.** Agreed. Added to every `msiexec` call in both actions; 3010 is treated as success everywhere. Relevant here since a locked `LemonadeServer.exe`/`lemond.exe` can push the uninstall into a reboot-pending state.

**3. Stale ARP key removal.** Agreed — the old code printed success unconditionally after `-ErrorAction SilentlyContinue`, which would have been a lie if the key survived (e.g. non-elevated runner). Now `-ErrorAction Stop` in a try/catch plus a `Test-Path` check, warning if the key is still there. Kept as a warning rather than a hard failure since this action also runs `if: always()` in teardown, and the install-side uninstall-by-file covers the residual case.

Verified both scripts parse under PowerShell 7.4 and unit-tested the retry loop and key-removal branches against simulated exit codes / a failing removal.

fl0rianr (approved):
Thanks! Looks good to me — approved.

PR #2896, opened 2026-08-04
ci: run inference tests only for affected engines, gated by one required check

jeremyfowers:
Cuts PR CI cost by running the exe/deb/dmg inference tests only for engines whose code could be affected, while keeping merge protection via a single aggregate check.

- New `changes` job classifies the PR's changed files with a unit-tested script (`.github/scripts/select_inference_engines.py`): backend folders (sources and headers) and per-engine test scripts select just that engine's legs; an explicit safe-list (docs, markdown, `src/app`, `src/web-app`, `src/cpp/tray`, `test/app`, `test/cpp`, and the `.github/` paths that can't reach an inference job) selects none; **anything unrecognized runs the full matrix**, as does every non-PR event (push, merge_group, tags).
- Matrix legs moved verbatim to `.github/inference-matrix.json`; `test-exe-inference`/`test-deb-inference` consume it via `fromJSON`, so leg names, steps, and check names are unchanged.
- New `Inference Tests OK` job aggregates results (`if: always()`, fails on any failure/cancellation). It also depends on the three build jobs, so a build failure can't leave it green via a skipped inference job.
- Only `BACKEND_DIR_TO_ENGINES` and `DMG_ENGINES` are hand-maintained. The script→leg direction is derived from the matrix, as is the rule that a `test/` script no leg runs selects nothing. `validate_mapping()` runs inside the gated `changes` job, so a rule naming a removed leg — or a leg no backend folder names — fails the required check instead of silently retiring that engine's tests. `test/test_inference_engine_selection.py` (29 tests) runs in a `check-inference-selection` job in this same workflow and covers the rest, including that every gated job is in the aggregate's `needs`.

Replayed against the last 150 merged PRs: **21% fewer exe and deb legs** and 18% fewer `.dmg` runs. 24 select zero legs, 9 narrow to a subset, 117 still run the full matrix — dominated by `src/cpp/**` (84 PRs) and `CMakeLists.txt` (34), which legitimately touch everything. 13 of those full-matrix PRs were caused solely by `prototype/ui-redesign/`, a directory since deleted; excluding them the figure is ~30%. `merge_group` deliberately still runs everything.

### Before flipping branch protection

Swap the 13 per-engine contexts for `Inference Tests OK` in one edit, at merge time — a filtered-out leg never reports a status, so it blocks rather than counting as skipped, and `enforce_admins` is on. No other context needs adding: the `check-inference-selection` job that verifies the selection rules is in the aggregate's `needs`, so it already blocks.

**This widens what gates merges.** The aggregate covers all 27 legs; protection names only 13. Over the last 20 `merge_group` runs, 12 failed a leg that does not block today — mostly `Test .exe - router-onnxruntime` (5), `audio-gen-acestep` (3), and `tts-openmoss` (3). The widening is intentional (hand-syncing contexts is what let those legs drift out of the required set), but those legs likely want fixing or removing from the matrix first.

This PR touches the workflow itself, so change detection selects ALL engines here — the full matrix running on this PR is the expected result, and CI confirms it. The narrowed path is exercised by unit tests but not yet by a real run; the first docs-only PR after merge is the one to watch.


fl0rianr (commented):
This looks complementary to #2889 rather than a full replacement.

#2896 significantly reduces the expensive inference matrix for regular code changes, while #2889 provides a stricter fast path that also skips builds, packaging, download-registry tests, and schema tests for small docs-only PRs.

One safety issue I noticed here: the changed-file query only passes `filename` to the selector. For renamed files, `previous_filename` should also be classified. Otherwise, a rename such as `src/cpp/server/router.cpp` to `docs/router.md` could be treated as docs-only and skip the inference matrix.

I would also give the `changes` job explicit `pull-requests: read` permission.

My preferred end state would be to merge this structural inference refactor first, then rebase #2889 on top and retain its strict docs-only build/packaging fast path. Does that direction make sense to you?

jeremyfowers:
> My preferred end state would be to merge this structural inference refactor first, then rebase https://github.com/lemonade-sdk/lemonade/pull/2889 on top and retain its strict docs-only build/packaging fast path. Does that direction make sense to you?

Sounds good, I will work on implementing your suggestions!

PR #2897, opened 2026-08-04
feat(release): separate stable and prerelease distribution channels

fl0rianr:
## Summary

This PR separates stable releases, prereleases, and development snapshots into explicit release channels.

Stable tags such as `v12.0.0` continue through the full signed release pipeline and stable PPA. Supported prerelease tags such as `v12.0.0-beta.1` publish to the bleeding-edge PPA without entering the full stable release workflow.

## Why

Previously, every `v*` tag could enter stable release paths, while Debian package versions were derived from Git tag ancestry. That is unsafe for prereleases because release tags live on dedicated release branches and are not necessarily reachable from `main`.

Git history is also not a reliable release-channel ordering mechanism: tag selection can vary with commit distance, annotated versus lightweight tags, and unrelated invalid tags.

## Changes

* accept only `alpha.N`, `beta.N`, and `rc.N` prerelease tags
* validate tag syntax, CMake version, and tagged commit
* restrict the full release workflow to stable tags
* route prereleases to the bleeding-edge PPA
* keep stable releases in the stable PPA
* assign all bleeding-edge publications a monotonic workflow serial
* keep every bleeding-edge package below the final stable version
* add parser, ordering, action-integration, override, and merge-queue tests
* document the prerelease process and Debian version policy

Example ordering:

```text
12.0.0~0.120.1.dev+gabcdef0
<
12.0.0~0.121.1.beta.1
<
12.0.0~0.122.1.dev+g1234567
<
12.0.0~0.123.1.rc.1
<
12.0.0
```

## Design

The bleeding-edge order is based on the Launchpad workflow's monotonically increasing run number and rerun attempt, not on `git describe` or tag reachability.

This keeps package upgrades monotonic even when prerelease tags are created on release branches that diverge from `main`.

## Result

Stable users remain on a conservative release path, while prerelease and development builds share a predictable opt-in channel without risking package-version regressions.

### DIFF
--- a/.github/actions/classify-release-tag/action.yml
+++ b/.github/actions/classify-release-tag/action.yml
@@ -1,0 +1,97 @@
+name: Classify Release Tag
+description: Validate a Lemonade release tag and derive channel/package metadata.
+
+inputs:
+  tag:
+    description: Complete tag such as v12.0.0 or v12.0.0-beta.1. Empty for snapshots.
+    required: false
+    default: ""
+  require_tag:
+    description: Fail when tag is empty.
+    required: false
+    default: "false"
+  require_prerelease:
+    description: Fail unless the supplied tag is a prerelease.
+    required: false
+    default: "false"
+  require_stable:
+    description: Fail unless the supplied tag is a stable release.
+    required: false
+    default: "false"
+  build_serial:
+    description: Monotonic serial for published snapshots/prereleases, normally github.run_number.
+    required: false
+    default: ""
+  build_attempt:
+    description: Attempt number for the workflow run, normally github.run_attempt.
+    required: false
+    default: "1"
+  commit_sha:
+    description: Commit SHA used in snapshot labels. Defaults to the checked-out HEAD.
+    required: false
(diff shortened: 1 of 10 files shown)


jeremyfowers:
Why do we need this? To put a GUI3 beta on the bleeding edge ppa? I'm not convinced that is desired, people subscribed to bleeding edge probably have an expectation that they are getting commits onto main.

fl0rianr:
Fair point — I agree that bleeding edge should continue to mean code from main.

For GUI3, the beta code will already be merged into main; the prerelease tag would only mark a known main commit, making the beta more visible and giving users a clear version to reference in feedback.

It is a more general way of (pre-)release option: It also enables optional Windows and macOS prerelease packages, but those would deliberately remain manual and opt-in rather than being published automatically (follow-up).

As a safeguard, we could require every prerelease published to the bleeding-edge PPA to be reachable from main. That preserves the existing expectation while still giving major changes like GUI3 a visible and reproducible beta milestone.

PR #2898, opened 2026-08-04
fix(server): propagate non-streaming request cancellation to backends

abn:
Resolves an issue where client socket disconnects or timeouts during non-streaming inference (prefill compute or token generation) left backend processes (`llama-server`) running indefinitely, locking GPU VRAM and slots.

Following the model established in #3133 for streaming requests, this PR extends client liveness checks and upstream HTTP transfer cancellation to non-streaming requests.

## Changes

- **Core HTTP Client Cancellation**: Added `should_cancel` callback support to `HttpClient::post` via libcurl progress callbacks, and introduced `HttpClientCancellationException` to preserve structured transport cancellation error codes (`CURLE_ABORTED_BY_CALLBACK`).
- **Server & Request Context Propagation**: Propagated client socket liveness checkers (`req.is_connection_closed()`) through `WrappedServer::RequestCancelScope` across non-streaming handlers (`chat_completion`, `completion`, `responses`, and `CloudServer::post_with_auth`).
- **Consolidated Tests**: Added `test/cpp/test_request_cancellation.cpp` and `test/server_cancellation.py` covering end-to-end `WrappedServer` cancellation, direct `HttpClient` callback aborts, and SSE streaming drops.
- **Documentation**: Updated Invariant 14 and router cancellation policy docs to reflect upstream HTTP cancellation propagation.

## Scope Note

As discussed in review, direct slot cancellation (`/slots`), SSE heartbeats, and TCP keep-alive socket options have been removed from this PR and deferred to separate scoped proposals (#3209, #3210).


abn:
@jeremyfowers @ramkrishna2910 can I please get a review on this one - this is a fairly frustrating issue :smile:

fl0rianr:
I agree this is frustrating hence I did act as well...

tbh, didn't saw this (strange inside of 2 weeks..) already older PR  when I thought about the more focused already merged solution of https://github.com/lemonade-sdk/lemonade/pull/3133.
In the light of this partly already addressed issue I suggest focusing on a much more tighter scope here.

abn:
Thanks @fl0rianr! PR #3133 was a great first step for early SSE transfer aborts.

I've just rebased this PR cleanly on top of latest `main` (including #3133). This PR completes the rest of the cancellation & robustness architecture that #3133 doesn't cover:

1. **Backend Slot Eviction**: Issues explicit non-blocking `POST /slots/{slot_id}?action=cancel` to `llama-server` upon disconnect so the backend doesn't continue computing and locking GPU VRAM/KV-cache in the background.
2. **Non-Streaming Cancellation**: Extends disconnect detection to non-streaming endpoints (`stream: false` for `chat/completions`, `completions`, and `responses`).
3. **Reverse-Proxy SSE Heartbeats (`: ping\n\n`)**: Prevents 504 Gateway Timeouts during long prefill on large prompts behind proxies (Cloudflare/Nginx) and provides immediate downstream socket death detection.
4. **OS TCP Keep-Alives**: Configures socket-level keep-alive across Linux, macOS, and Windows.
5. **Integration Test Suite**: Adds `test/server_cancellation.py` covering prefill drops, non-streaming cancellations, and multi-client queue isolation.

Happy to trim anything if you think a specific piece should be split out, but these components work together as a coherent solution to the hanging request problem.

fl0rianr:
True, but the watchdog we have already those that.

abn:
If I understand correctly, the watchdog and slot cancellation address two slightly different scenarios:

1. **Watchdog (for unrecoverable process hangs)**:
   Monitors `/health` and after a grace period (~90s) kills and restarts the backend process if it becomes unresponsive. However, when a client simply disconnects or cancels, `llama-server` is still healthy—its `/health` returns `200 OK`, so the watchdog won't trigger. Without an explicit slot cancel, `llama-server` keeps computing the abandoned prompt in the background, holding the KV-cache slot and GPU VRAM until completion.

2. **Slot Cancellation (for normal client disconnects/stops)**:
   Immediately (`<50ms`) tells `llama-server` to drop only that active task via `POST /slots/{slot_id}?action=cancel`. The backend process remains running, the model stays in memory, and the slot is instantly freed for the next query.

Alongside this, the PR also includes:
- Extending cancellation to **non-streaming** requests (`stream: false`)
- **SSE heartbeats (`: ping\n\n`)** so reverse proxies (like Cloudflare/Nginx) don't drop connections with 504 timeouts during long prefills
- OS-level **TCP keep-alive**
- Comprehensive test coverage in `test/server_cancellation.py`

Happy to hear your thoughts on how you'd prefer to structure this!

fl0rianr (changes requested):
You are right that the watchdog and normal client cancellation are two different cases. My concern is more with the explicit slot cancellation approach here.

As it stands, the disconnect paths call abort_request() without a known slot id. The llama.cpp fallback then queries /slots and cancels every slot that is currently processing. So with two concurrent clients, client A disconnecting could also cancel client B's active request.

test_04_multi_client_queue_isolation would not catch that at the moment, since client B is only a /health request rather than a second active inference.

#3133 already showed that aborting the upstream HTTP transfer is enough to propagate the disconnect to llama-server without Lemonade having to manage llama.cpp slots directly. I think we should keep that model and extend it to the missing non-streaming paths.

So my preference would be to narrow this PR down to non-streaming disconnect propagation. I would split the SSE heartbeats and OS TCP keep-alive into separate changes, and avoid direct slot cancellation unless we have a reliable Lemonade request → llama.cpp slot mapping.

abn:
Thanks for clarifying @fl0rianr! That makes complete sense.

I have updated the PR to strictly align with that model:
1. **Narrowed scope**: Restricted changes to propagating client liveness (`req.is_connection_closed()`) to non-streaming requests (`chat_completion`, `completion`, `responses`) via `HttpClient::post`'s `should_cancel` callback.
2. **Removed direct `/slots` cancellation**: Dropped `abort_request()` and `/slots` queries, avoiding the risk of cancelling concurrent requests. Disconnects now cleanly abort the upstream HTTP transfer as established in #3133.
3. **Split out secondary features**: Removed TCP keep-alive socket options and SSE heartbeats from this PR so they can be reviewed independently.
4. **Tests & Invariants updated**: All unit tests (`ctest -L cpp-ci`, `test_streaming_proxy_cancel`, `gen_backend_boilerplate.py`) and cancellation integration tests pass cleanly.

PR title and description have also been updated to reflect the streamlined scope.

fl0rianr (commented):
Thanks, this is much closer to the scope I had in mind.

I still see a few things to clean up before merging:

 * Some docs still describe the removed slot cancellation, TCP keep-alive and SSE heartbeat behavior.
 * The new generic cancellation path seems to be effective for llama.cpp, but other wrapped backends such as vLLM still use the old methods, so the callback can get dropped silently.
 * The non-streaming cancellation test currently only checks that the server stays healthy after the client disconnects. It doesn't verify that the abandoned backend request actually stops. I think we need a regression test that proves the upstream request terminates promptly.
 * There are still a few unrelated changes left in the diff, e.g. .venv, SIGPIPE handling, proxy policy and streaming-path changes. Since the PR is now specifically about non-streaming cancellation, I'd prefer to keep those separate too.

The overall direction looks much better now, I just want to make sure the narrowed scope is also reflected consistently in the implementation, tests and docs.

abn:
Thanks for the thorough review and feedback! I've updated the PR accordingly:

1. **Scrubbed residual docs**: Removed all remaining references to TCP keep-alive, SSE heartbeats, and `/slots` across `docs/`, `AGENTS.md`, and test descriptions.
2. **Extended `should_cancel` across all completion backends**: Added `should_cancel` overloads to `ICompletionServer` and implemented forwarding in `vllm`, `fastflowlm`, `ryzenai`, and `cloud` backends so cancellation is preserved consistently across all backends.
3. **Added dedicated regression test**: Added `test/cpp/test_non_streaming_cancel.cpp` (and wired it into CMake), which simulates a long backend compute and asserts that client disconnect aborts the upstream HTTP transfer promptly (~1s) via `HttpClient::post`'s cancellation callback.
4. **Cleaned up unrelated diffs**: Reverted `.venv/`, watchdog test edits, proxy policy, and streaming-path modifications in `http_client.cpp`.

fl0rianr (commented):
One thing that still concerns me is how the scope is evolving.

We started with 19 changed files, narrowed this down substantially, and after splitting TCP keepalive and SSE heartbeats into #3209 and #3210, this PR itself has now grown to 25 files.

Looking at the diff, a large part of that growth comes from threading a new should_cancel overload through every completion backend and its interface.

I'm not convinced we need to expand the completion API across all backends to solve this bug. WrappedServer already has per-request cancellation state and forward_request() is the common non-streaming forwarding layer. Could we propagate the connection-liveness checker through that shared request context instead, and keep backend implementations unchanged unless they actually need backend-specific cancellation behavior?

My concern is that we are again growing the architecture around the fix rather than keeping the fix at the narrowest common layer.

abn:
Thanks for the suggestion, that's much cleaner!

I've refactored the cancellation propagation to leverage `WrappedServer`'s thread-local request context:
- Added `WrappedServer::RequestCancelScope` RAII helper to hold the client connection liveness checker (`req.is_connection_closed`) during handler execution in `server.cpp`.
- `WrappedServer::forward_request()` automatically picks up the thread-local cancellation context and passes it to `HttpClient::post()`.
- Reverted all the `should_cancel` method overloads across `ICompletionServer`, `Router`, and individual backend classes (`vllm`, `fastflowlm`, `ryzenai`, `cloud`, `llama.cpp`), leaving backend implementations completely untouched.

This brings the PR down to 12 files focused strictly on the common forwarding layer, `HttpClient`, and tests/docs.

fl0rianr (commented):
Thanks! This is much cleaner now and the shared request context looks like the right direction to me.

I still see a few things before approval:

The new regression test currently calls HttpClient::post() directly with the cancellation callback. That proves the low-level curl cancellation, but not the new RequestCancelScope -> WrappedServer::forward_request() propagation that this refactor introduces. I think the regression test should exercise that shared path as well.

Cancellation classification also still looks fragile. forward_request() matches specific exception strings, while the test already accepts another libcurl callback-abort string and even any generic CURL error. I'd rather preserve/classify the actual CURL result than rely on string matching, and make the test assert the exact cancellation outcome.

There also seem to be a few leftovers from the earlier scope: global SIGPIPE handling, the model-load cancellation flag which only samples connection state once before loading, and an unused should_cancel parameter on forward_multipart_request.

On the test organization point raised elsewhere, I also agree we should avoid adding one standalone CTest executable per feature where possible. Maybe this cancellation regression can live in a broader HTTP/request-cancellation test suite instead, similar as mentioned on the smaller heart beat follow-up. Thanks.

Apart from those, this is substantially closer to what I expected from the narrowed scope.

abn:
Thanks for the thorough review! I've addressed all four items:

1. **End-to-End Regression Test**: The test suite now exercises `WrappedServer::RequestCancelScope` -> `WrappedServer::forward_request()` end-to-end with a mock backend to ensure the shared context propagates to the upstream request and produces the expected error response.
2. **Structured Cancellation Classification**: Replaced exception string matching with `HttpClientCancellationException` (which preserves the libcurl error code `CURLE_ABORTED_BY_CALLBACK`). `forward_request()` and `forward_multipart_request()` now catch this exception directly.
3. **Scrubbed Leftovers**: Removed global `SIGPIPE` suppression, the single-sample model-load cancellation flag, and the unused `should_cancel` parameter on `forward_multipart_request`.
4. **Consolidated Test Binaries**: Merged the standalone streaming and non-streaming cancellation test executables into a single `test_request_cancellation` suite.

fl0rianr (commented):
One remaining blocker before approval:

* `CloudServer` still bypasses the new request cancellation context. `chat_completion()` / `completion()` go through `post_with_auth()`, which calls `HttpClient::post()` without forwarding `current_request_cancel_context()`. So client disconnects can still be dropped for cloud requests. Please propagate the existing thread-local context there as well, without expanding the completion interfaces again.
* `http_client.h` should include `<stdexcept>` directly since it derives from `std::runtime_error`.
* The PR description still mentions a “25ms rate-limited” cancellation callback, but I don’t see that behavior in the current implementation anymore.

Everything else from my previous review looks addressed. Once the CloudServer path is covered, I’m good with this.

abn:
Thanks! I've updated those last items:

1. **CloudServer Cancellation**: `CloudServer::post_with_auth()` now retrieves the thread-local `current_request_cancel_context()` and passes the cancel flag/checker down to `HttpClient::post()`, catching `HttpClientCancellationException` directly.
2. **Include `<stdexcept>`**: Added `<stdexcept>` to `http_client.h`.
3. **PR Description**: Updated the PR description to remove the outdated 25ms rate-limit mention.

fl0rianr (commented):
Thanks, this is much closer now. I did another pass through the full cancellation path and found a few remaining gaps before I’m comfortable approving:

* RequestCancelScope is created too late in handle_chat_completions(). Router collections and Omni collections can already run real inference before the scope exists, so disconnects there can still leave backend work running.
* test_request_cancellation is registered with CI OFF, so the new regression suite is not built or run by the normal cpp-ci job. test/server_cancellation.py also does not appear to be part of the endpoint CI.
* Omni image edits still go through forward_multipart_request() / post_multipart(), which has no cancellation callback. The cancellation catch there is currently unreachable for client disconnects.
* Minor: Router::chat_completion() still has the old cancellation cleanup that only resets the flag instead of restoring the full previous request context.

The direct non-streaming path itself looks good now, including CloudServer and the structured curl cancellation handling. I’d fix the scope/collection gaps and make the new cancellation test part of CI, then I think this is in good shape.

abn:
Thanks! I've addressed all four items:

1. **Early `RequestCancelScope` Placement**: Moved `RequestCancelScope` creation to the top of `handle_chat_completions()`, `handle_completions()`, and `handle_responses()` so Router collections, Omni collections, and auto-load paths all run within the active cancellation scope.
2. **Multipart Request Cancellation**: Added `should_cancel` / `cancel_flag` support to `HttpClient::post_multipart()` via libcurl progress callbacks, forwarded the context in `WrappedServer::forward_multipart_request()`, and added a multipart test case to `test_request_cancellation`.
3. **Router Cleanup**: Cleaned up `Router::chat_completion()` to use `WrappedServer::RequestCancelScope` to manage RAII state restoration.
4. **CI Enablement**: Changed `RequestCancellationTest` to `CI ON` in `CMakeLists.txt` (now running under `ctest -L cpp-ci`) and added `test/server_cancellation.py` to the CI endpoint test step.

fl0rianr (commented):
Still one blocker before approval:

* `handle_chat_completions()` creates the correct request context with `req.is_connection_closed()`, but `Router::chat_completion(..., &req_cancel)` then creates an atomic-only `RequestCancelScope`, which replaces the outer context and drops `should_cancel`. A disconnect during prefill/generation can therefore still leave non-streaming chat running.
* Please add a regression test for exactly this nested `handler scope -> Router -> forward_request()` path. The current cancellation test bypasses `Router`, so CI stays green despite this bug.
* The rebased code also has non-streaming inference paths through Anthropic/Ollama/MCP that bypass the request cancellation context, and streaming Omni runs its work inside the content-provider after the handler scope is gone. Either cover these too or narrow the new global cancellation invariant accordingly.

The HttpClient/CloudServer/multipart plumbing itself looks good now.

jeremyfowers:
@Geramy networking PR

abn:
**@fl0rianr**, thanks for the careful pass, that blocker was a real one. You were right that the nested scope was swallowing the liveness checker.

I fixed the flag-only `RequestCancelScope` so a nested scope preserves the outer `should_cancel` instead of replacing the whole context with `{flag, nullptr}`. `Router::chat_completion()` no longer loses the connection-liveness checker it wraps, so a mid-prefill/generation disconnect now aborts non-streaming chat as intended. `completion` and `responses` were never affected since they don't pass a flag into Router.

I also added the regression test you asked for. It reproduces the exact nested handler-scope to flag-only-scope to `forward_request()` path, with `connection_closed` and `req_cancel` as separate signals mirroring the real handler. I confirmed it fails (the backend runs the full 10s and returns no cancellation) before the fix and passes after.

On the Anthropic/Ollama/MCP non-streaming paths and streaming-Omni: rather than widen this change, I scoped the documented guarantee (AGENTS.md invariant and `docs/dev/router-policy.md`) to the OpenAI-compatible non-streaming endpoints the fix covers, and noted the gateway and streaming-Omni gaps explicitly. That keeps the fix at the narrowest common layer and the docs honest about coverage.

abn:
Copilot review feedback applied.

fl0rianr (approved):
LGTM now

@Geramy should have a short look as well

Geramy:
This is pretty big, I'm going to pulling this up in my visual studio code locally to analyze it first, I'll make a decision today.

src/cpp/include/lemon/wrapped_server.h
```diff
@@ -443,6 +443,26 @@ class WrappedServer : public ICompletionServer {
...
+
+        RequestCancelScope(const RequestCancelScope&) = delete;
+        RequestCancelScope& operator=(const RequestCancelScope&) = delete;
+
+    private:
+        RequestCancelContext prev_context_;
+    };
+
+    static void set_request_cancel_context(const RequestCancelContext& ctx);
+    static RequestCancelContext current_request_cancel_context();
```
Geramy: Okay so I reviewed this myself, I am not a huge fan of the structure here, RequestCancelContext should be RequestCancelToken through precedance but it should be inside HttpClient since HttpClient requires this in convention with or without wrapped_server.h and you could remove two parameters for one it is a cleaner look. I believe you could entirely get rid of RequestCancelScope as well. Here is what my AI says.

1. Rename RequestCancelContext to RequestCancelToken and relocate it from WrappedServer into utils::HttpClient (utils/http_client.h). The token is the value HttpClient interprets — flag plus checker, consulted inside   
 the curl transfer-info callback to abort the transfer — so the type belongs adjacent to that logic rather than in the backend abstraction layer. Add a cancelled() member that encapsulates the (flag && flag->load()) ||  
 (checker && checker()) predicate, and pass the token by const& to post and post_multipart in place of the two trailing parameters. This simultaneously removes the predicate as currently duplicated verbatim in           
 forward_request, forward_multipart_request, and cloud_server.cpp::post_with_auth.                                                                                                                                          
                                                                                                                                                                                                                            
 2. RequestCancelScope is not independently deletable; it is the terminal element of a chain of ambient state, and the chain must be unwound in order:                                                                      
                                                                                                                                                                                                                            
 - The scope exists solely to save and restore the wrapped_server.cpp thread-local on handler entry and exit.                                                                                                               
 - The thread-local exists solely so that forward_request and forward_multipart_request can read the token without it appearing in their signatures — a deliberate avoidance, in a bugfix PR, of touching the protected     
   WrappedServer API and its ~20 subclass call sites.                                                                                                                                                                       
 - The legacy std::atomic<bool>* cancel parameter on Router::chat_completion is the last explicit seam; the PR already modifies it.                                                                                         
                                                                                                                                                                                                                            
 The correct end state is explicit threading end-to-end: the server.cpp handler constructs the token from req.is_connection_closed, passes it into Router::chat_completion (replacing the std::atomic<bool>* parameter),    
 and Router forwards it into forward_request / forward_multipart_request (gaining a const utils::RequestCancelToken& parameter), which pass it to HttpClient. At that point the thread-local, all four *_request_cancel*    
 accessors, and RequestCancelScope itself are dead code and should be deleted in the same change. Given the diff size, landing this PR with the scope in place and completing the explicit threading as a follow-up is      
 acceptable — but the follow-up should be filed now, and the invariant to hold is that no ambient request-scoped state remains once the token is a first-class parameter.

Geramy (changes requested):
please see my comments, I'll continue reviewing but you can get started on this.

src/cpp/include/lemon/wrapped_server.h
abn: Both points are in this push.

`RequestCancelToken` lives in `utils::HttpClient` now, with one `cancelled()`. That removes the predicate repeated in `forward_request`, `forward_multipart_request` and `cloud_server.cpp::post_with_auth`. `post` and `post_multipart` take the token by `const&`. I also dropped three accessors with no callers left (`set_request_cancel_context`, `set_request_cancel_flag`, `current_request_cancel`) and the orphaned `cancel_xferinfo_callback`.

`RequestCancelScope` and the thread-local still hold and drive that token, so the chain is shorter, not gone. #3645 tracks the end state you described: thread the token from the handler through `Router::chat_completion` into the forwarders, then delete the scope, the thread-local and `current_request_cancel_context`.

How the token reaches the forwarders is still open. `Router` calls the backend `chat_completion` interface, not the forwarders, so the token may have to travel through that interface. That choice sets how many backend call sites move (counts in #3645).

src/cpp/include/lemon/wrapped_server.h
```diff
@@ -443,8 +443,20 @@ class WrappedServer : public ICompletionServer {
 
     void set_load_cancel_flag(std::atomic<bool>* f) { load_cancel_ = f; }
 
-    static void set_request_cancel_flag(std::atomic<bool>* f);
-    static std::atomic<bool>* current_request_cancel();
+    class RequestCancelScope {
+    public:
+        explicit RequestCancelScope(std::function<bool()> should_cancel, std::atomic<bool>* flag = nullptr);
+        explicit RequestCancelScope(std::atomic<bool>* flag);
```
Geramy: can we not just use RequestCancelToken here?

src/cpp/server/server.cpp
```diff
@@ -3892,6 +3892,16 @@ void Server::handle_chat_completions(const httplib::Request& req, httplib::Respo
...
+    std::function<bool()> should_cancel;
+    if (req.is_connection_closed) {
+        should_cancel = [is_closed = req.is_connection_closed]() { return is_closed && is_closed(); };
+    }
+    std::atomic<bool> req_cancel{false};
+    if (should_cancel && should_cancel()) {
+        req_cancel.store(true);
+    }
+    WrappedServer::RequestCancelScope cancel_scope(should_cancel, &req_cancel);
+
```
Geramy: can we replace should_cancel here nd req_cancel with the actual RequestCancelToken type and then feed it into cancel_scope here? It seems this is divergent from the new style.

src/cpp/server/utils/http_client.cpp
```diff
@@ -312,10 +312,15 @@ static size_t write_file_callback(void* ptr, size_t size, size_t nmemb, void* st
...
-static int cancel_xferinfo_callback(void* clientp, curl_off_t, curl_off_t, curl_off_t,
-                                    curl_off_t) {
-    auto* flag = static_cast<std::atomic<bool>*>(clientp);
-    return (flag && flag->load()) ? 1 : 0;
+static int post_cancel_xferinfo_callback(void* clientp, curl_off_t, curl_off_t, curl_off_t,
+                                         curl_off_t) {
+    if (!clientp) return 0;
+    auto* token = static_cast<const RequestCancelToken*>(clientp);
+    try {
+        return token->cancelled() ? 1 : 0;
```
Geramy: no need to do ? 1 : 0, a boolean is technically 0 or 1, true = 1, false = 0.

src/cpp/server/utils/http_client.cpp
```diff
@@ -312,10 +312,15 @@ static size_t write_file_callback(void* ptr, size_t size, size_t nmemb, void* st
...
+    } catch (...) {
+        return 1;
```
Geramy, line 324: define the catch here please with a exception named type or comment at least so someone can know what we are trying to catch here, thanks!

src/cpp/server/server.cpp
```diff
@@ -4069,6 +4079,16 @@ void Server::handle_chat_completions(const httplib::Request& req, httplib::Respo
...
```
Geramy: Please use the Token here too, don't define should_cancel, and req_cancel here, follow the new convention please.

src/cpp/server/server.cpp
```diff
@@ -5634,6 +5654,16 @@ void Server::handle_image_upscale(const httplib::Request& req, httplib::Response
...
```
Geramy: same as ^ thanks.

src/cpp/server/wrapped_server.cpp
```diff
@@ -14,11 +14,29 @@
 
 namespace lemon {
 
-static thread_local std::atomic<bool>* t_request_cancel = nullptr;
+static thread_local utils::RequestCancelToken t_request_cancel_ctx{};
 
-void WrappedServer::set_request_cancel_flag(std::atomic<bool>* f) { t_request_cancel = f; }
+WrappedServer::RequestCancelScope::RequestCancelScope(std::function<bool()> should_cancel, std::atomic<bool>* flag)
```
Geramy: fix the header and definition to take in the token structure.

Geramy (changes requested):
Please fix the comments I made and I think we are good! Looks great, good work! @abn

abn:
All seven are addressed.

`RequestCancelScope` now takes a `utils::RequestCancelToken` and nothing else, so the two-constructor split is gone. The three handlers build the token directly instead of holding `should_cancel` and `req_cancel` side by side, and `Router::chat_completion()` wraps its flag the same way. The nested-scope fix lives in the token constructor now: a scope that arrives without a checker keeps the outer one, so the Router wrap still carries the connection-liveness check.

In `http_client.cpp` the callback returns `token->cancelled()` directly, and the `catch (...)` carries the same comment the streaming callback has. It exists so no C++ exception crosses libcurl's C boundary, and it fails closed.

I did not add the token parameter to `forward_request` / `forward_multipart_request` yet. That is the explicit-threading step in #3645 and it moves every backend call site, so I kept it out of this round.

fl0rianr (commented):
This is much cleaner now.
One thing I’d still simplify before merging: `req_cancel` is still carried through the handlers/Router even though it never gets set; the live cancellation is already coming from `RequestCancelToken::should_cancel`. I think we can drop that leftover path here and leave the full explicit token threading/TLS removal to #3645.

abn:
@fl0rianr Dropped it. The three handlers no longer declare `req_cancel`; they hand the scope a token that carries only the connection-liveness checker, and `handle_chat_completions()` calls `router_->chat_completion(request_json)` with no flag.

`Router::chat_completion`'s `std::atomic<bool>*` parameter stays, since the jobs path passes a real `CancelFlag` there and the job manager does set it. That is the only live flag left, and it goes away with the explicit threading in #3645.

src/cpp/include/lemon/wrapped_server.h
```diff
@@ -443,8 +443,19 @@ class WrappedServer : public ICompletionServer {
...
+    public:
+        explicit RequestCancelScope(const utils::RequestCancelToken& token);
+        ~RequestCancelScope();
+
+        RequestCancelScope(const RequestCancelScope&) = delete;
+        RequestCancelScope& operator=(const RequestCancelScope&) = delete;
+
+    private:
+        utils::RequestCancelToken prev_token_;
+    };
```
Geramy, line 456: Not to be a pain in the neck is there a way to get this inside a utility file instead, like requests_types.h/.cpp or something??

Geramy (changes requested):
please move into a appropriate request .cpp/.h file inside utils, this doesn't belong in wrapped_server, wrapped_server uses it. but its a utility item and a generic request type which should belong with other "request" types in our project, nice and organized! :) THANKS!

src/cpp/include/lemon/wrapped_server.h
abn: I'd really like to get this done as this is an annoying bug for some of us and this has gone though a lot of review rounds. I'd prefer to not add more scope right now.

That said, if you really think this is a blocker I can come up with a solution. And do another round.

abn: #3645 deletes `RequestCancelScope` and the thread-local it owns. Is it worth relocating it now just to delete it there?

Geramy: that's fine :) I didnt see that, lets get this merged.

Geramy (approved):
https://github.com/lemonade-sdk/lemonade/issues/3645 deletes RequestCancelScope, so I'll pass this PR.

fl0rianr (approved):
We can merge this one as-is IMO

abn:
Guessing a code owner review is required now? 😔 

cc: @jeremyfowers @ramkrishna2910

fl0rianr:
> Guessing a code owner review is required now? 😔
> 
> cc: @jeremyfowers @ramkrishna2910

It is the added line in agents.md

PR #2900, opened 2026-08-05
ci: run Launchpad PPA and distro builds in the merge queue instead of every PR

jeremyfowers:
Reduces GitHub-hosted runner pressure (org-wide cap: 20 concurrent jobs). `launchpad-ppa` (7 jobs, ~62 min) and `linux_distro_builds` (2 jobs, ~45 min) compile the whole server on every PR push, but their unique coverage is packaging/distro-specific: in 24 days of history (04 Jul–04 Aug), 23 of 24 launchpad PR failures co-failed with required checks on the same push; the one unique catch (an Ubuntu 26.04 build break) would now surface at merge-queue time instead — delayed, not lost.

- `launchpad-ppa`: `pull_request` → `merge_group`; the binary-build validation legs now run per queued PR. The push-to-main source build + PPA upload path is untouched.
- `linux_distro_builds`: `pull_request` + `push: main` → `merge_group` (the merge-queue run covers main; per-commit status still attaches to the landed SHA).
- Arch/Debian README badges repointed to `?event=merge_group`.


fl0rianr (commented):
I agree with moving the expensive Launchpad builds out of every PR update and into the merge queue. However, I see two issues that should be clarified or addressed before merging.

First, `linux_distro_builds` did not previously run on every PR push: it had path filters for C++, CMake, setup, and server-test changes. With the new unfiltered `merge_group` trigger, both full distro builds will run for every queued PR, including documentation-only changes. That shifts additional runner usage onto the merge queue’s critical path. Could we keep an always-reporting lightweight gate and conditionally run the expensive matrix only when the merge-group diff touches the previous set of relevant paths?

Second, please confirm how these jobs are configured as required checks. GitHub couples pull-request and merge-queue required checks and documents required Actions workflows with both `pull_request` and `merge_group` triggers. If these checks remain required, removing the PR trigger may prevent a PR from becoming queue-eligible. If they are optional, failures in the merge queue may not actually gate the merge, which would weaken the “delayed, not lost” claim.

The `event=merge_group` README badges should also be reconsidered or relabeled, since they represent the latest queue candidate rather than the current state of `main`.

PR #2902, opened 2026-08-05
ci: run Fedora RPM and Debian 13 packaging jobs only in the merge queue

jeremyfowers:
- **Why** — The Fedora 43/44 × x86/ARM RPM builds+tests and Debian 13 builds+smokes are 12 hosted jobs on every PR push (org cap: 20 concurrent). Over 24 days only 4 PR runs failed *solely* in these legs, all single-leg flakes.
- **What** — Skips them on `pull_request` and gates them in the merge queue behind a new required **`Packaging builds`** aggregate job. The queue gates *only* on required checks — see #2867, which merged despite a red merge-group run — and the matrix-generated job names shift on every Fedora bump, so they can't be required directly.
- **Rollout** — Merge this **first**, then add `Packaging builds` to `main`'s required status checks. Flipping the setting first blocks every open PR on a context that doesn't exist yet; until it's flipped the gate is advisory.

Related: #2909 and #2910 apply the same gate pattern to the Launchpad PPA and Linux distro workflows.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1589,10 +1589,9 @@ jobs:
               exit 1
           fi
 
-  # Required status check standing in for the jobs above: the merge queue gates
-  # only on required checks, and the per-job names shift with the matrix, while
-  # this one is fixed. It also reports on pull_request -- where the gated work is
-  # skipped -- so PRs can still enqueue.
+  # Gate job. Reports a single status check that passes only if every job in
+  # `needs:` passed. Branch protection requires this check's name, so the merge
+  # queue blocks on all of those jobs at once.
   packaging-gate:
     name: Packaging builds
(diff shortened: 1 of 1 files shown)


fl0rianr (commented):
Thanks, that is the right direction for the problem we want to tackle.
What is changed so far is fine by me, not far away from approval.

But, I want to make sure one thing:
If we go into the merge queue and on of those jobs fail the queue should fail.
When this is not the case we can merge breaking code.
I'm not 100% about the requirement setup, please check this.

fl0rianr (approved):
Thanks, the second commit addresses my concern.

Please make sure `Packaging builds` is added to `main`'s required status checks immediately after this lands. Until that repository setting is applied, the new gate remains advisory.

No further code concerns from my side.

jeremyfowers:
Superseded by #2911, which aggregates this with the other three merge-queue gate PRs.

PR #2903, opened 2026-08-05
ci: cut per-PR macOS jobs from 5 to 3

jeremyfowers:
macOS is the scarcest GitHub-hosted resource: the org-wide cap is 5 concurrent macOS jobs and a single PR push currently occupies all 5. This gets a same-repo PR push down to 3 macOS jobs (peak concurrency 2) with no loss of unique coverage; all three required macOS check contexts are unchanged.

- `test-dmg-inference` runs only on unsigned (fork-PR) runs: on signed runs its test steps were all conditioned away by `has_signing`, leaving just a redundant .pkg install that `Test CLI/Endpoints (macos-latest)` repeats in the same run. The dead signed-path steps are removed.
- Its whisper-metal step is dropped — the dmg build job runs the identical test in its unsigned path, and 13 of this job's 14 PR failures in the last 24 days were this duplicate flaking (net flake reduction).
- The embeddable macOS smoke checks (archive structure, `--version`, health check) move into `build-lemonade-embeddable-macos`, which strengthens the required build context; the macOS leg of `test-embeddable-posix` (0 failures in 24 days) is removed.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1091,6 +1091,76 @@ jobs:
           tar tzf "$ARCHIVE"
           ls -lh "$ARCHIVE"
 
+      - name: Smoke test embeddable archive
+        shell: bash
+        run: |
+          set -e
+          WORK="$RUNNER_TEMP/embeddable-smoke"
+          mkdir -p "$WORK"
+          tar xzf "build/lemonade-embeddable-${LEMONADE_VERSION}-macos-arm64.tar.gz" -C "$WORK"
+          DIR="$WORK/lemonade-embeddable-${LEMONADE_VERSION}-macos-arm64"
+
+          # Verify expected files exist
+          test -f "$DIR/lemond"
+          test -f "$DIR/lemonade"
+          test -f "$DIR/LICENSE"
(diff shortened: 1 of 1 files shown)


fl0rianr (commented):
Nice improvement overall — reducing macOS CI usage here makes sense, especially given the limited runner capacity.

One small concern: the new embeddable smoke test now runs on the same runner that built the archive. That means the checkout and build outputs are still present, so an accidental dependency on the build directory could go unnoticed even though the downloaded archive would fail on a clean Mac.

Could we keep the cost savings while making this test a bit more isolated, for example by removing the checkout/build tree before extracting and running the archive, and optionally checking the Mach-O dependencies with `otool`?

That would preserve most of the runner reduction while keeping the clean-environment safety net.

jeremyfowers:
Superseded by #2911, which moves macOS jobs into the merge queue behind a `ci:macos` label escape hatch. That approach cuts per-PR macOS consumption ~6x versus this PR's ~0x — the two jobs removed here run 0-1m each on signed runs, so job count drops 5→3 while macOS runner-minutes stay ~31/run.

The independently-good parts (deleting the dead signed-path steps in `test-dmg-inference`, dropping the duplicate whisper-metal run) are carried into #2911.

PR #2907, opened 2026-08-05
feat(gui3): refine apps navigation, backend identity UX and design details

fl0rianr:
## Summary

This PR improves GUI3 navigation and backend presentation across the Apps and Models workspaces.

Apps is now treated as a standalone workspace instead of a Settings section, while backend colors, rail behavior, and status markers have been made more stable, distinguishable, and consistent across Light and Dark Mode.

### Apps workspace

* remove Apps from the Settings navigation
* integrate the existing Apps view into the standard GUI3 workspace layout
* replace the centered layout and top category navigation with a left-side category rail
* align responsive rail behavior with the Monitor and Settings workspaces
* preserve the legacy `#/connect/app-directory` route and canonicalize it to `#/apps`

### Backend color identity

* introduce a centralized backend identity registry with stable color slots
* use the same backend colors in Light and Dark Mode
* enforce a global minimum OKLab color distance of `0.14`
* improve the visual separation between `llama.cpp` and `TRELLIS.2`
* preserve existing assignments when new backends are added
* support aliases without generating new backend identities
* reserve removed backend color slots instead of reusing them
* keep fallback and override handling centralized and maintainable

### Models view styling

* shorten the colored section of the backend accent line
* ensure all accent lines use a consistent single-pixel thickness
* remove overlapping styles that could render lines at double width
* improve contrast against both light and dark surfaces

### Backend rail behavior

* limit the backend rail to six visible backends by default
* add a `+N` pill showing the number of hidden backends
* reveal one additional backend per pill click
* hide the pill once no hidden backends remain
* show remove controls when more than six backends are visible
* move removed backends to the end of the hidden selection
* clear the active backend filter when its backend is removed

### Backend status markers

* replace nested status rings with stable single-state markers
* use a smaller hollow circle for available backends
* use a larger solid dot for active states
* avoid zoom-related alignment issues caused by nested markers
* preserve running and download pulse animations

## UX impact

The updated layout provides a clearer separation between global workspaces and Settings, while the backend rail remains compact by default and can be expanded incrementally.

Backend identities are now consistent across themes and sessions, status markers remain aligned at different zoom levels, and backend accent colors are easier to distinguish without introducing additional theme-specific styling.

The color "pressure" is reduced, here some examples

image
image

### DIFF
--- a/src/app/DESIGN.md
+++ b/src/app/DESIGN.md
@@ -163,11 +163,19 @@ The labels below are normative boundaries, not backlog states. **Justified speci
 
 **Logs — justified specialization:** monospace virtualized output, severity markers, and compact fixed-height rows. The filter panel, search control, header, and actions are standard workspace UI.
 
+### Apps
+
+**Layout:** two panels: an App types `.workspace-rail` and a full-width `.workspace-pane` directory. Marketplace categories live in the rail; search remains a pane-level action. The directory is not centered or embedded in Settings.
+
+**Justified specialization:** marketplace logos, category metadata, and external Visit, Guide, and Video actions.
+
+**Prohibited variation:** horizontal category chip bars, centered fixed-width page shells, duplicate Settings navigation entries, or app-specific rail/mobile behavior. Apps uses the shared workspace rail, pane header, resource rows, actions, and mobile context panel.
+
 ### Connect
 
-**Layout:** two panels: Settings `.workspace-rail` and a `.workspace-pane` for the selected Server, Storage, Cloud, MCP, Apps, Support, or Account section.
+**Layout:** two panels: Settings `.workspace-rail` and a `.workspace-pane` for the selected Server, Storage, Cloud, MCP, Support, or Account section.
 
-**Justified specialization:** provider brand marks, endpoint examples, and marketplace metadata. Forms use a bounded reading width.
+**Justified specialization:** provider brand marks and endpoint examples. Forms use a bounded reading width.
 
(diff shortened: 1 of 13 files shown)


jeremyfowers:
As long as you're working on colors: standard design practice would be to pick a certain color palette with a minimal number of complementary base colors, and then apply shades of those colors app-wide. Glancing at the  screenshots I'm not sure this is the case right now?

jeremyfowers:
Another thing I'm looking at here is density of the model list rows: would it look better if all the metadata, colored line, and engine name were all on the same line together under the model name? That would improve density by 33% by eliminating the 3rd line per entry.

jeremyfowers:
image

Fixes needed here:
- Filter names need to be capitalized
- Main panel should have a maximum width, so that CTA buttons don't go to the right edge of the UI on a very wide browser window
- Unify the layout of apps tab main panel with the models tab main panel: typography, search bar placement, header layout, etc.

jeremyfowers:
image

TBH its hard to comment on the improvements to the backends tab from this PR since it still needs a lot of improvements overall.
- Unify title area layout with the other tabs
- Eliminate confusing chips below engine titles
- Establish a proper visual hierachy between engine titles and backend titles within the engine card (they have basically the same typography and layout)
- Eliminate filters from the top of the main panel since they are already in the filter panel
- Eliminate useless "N variants" label on cards
- Eliminate or fix confusing `GPU (AMD) CPU` kind of chips below each backend title.
- Experimental engines should not have visual emphasis above non-experimental engines. Right now experimental gets a highlight color.
- Excessive use of highlight-color CTA buttons across the page
- etc.

PR #2909, opened 2026-08-05
ci: run Launchpad PPA binary builds in the merge queue instead of every PR

jeremyfowers:
- **Why** — `launchpad-ppa` built binary .debs for every supported Ubuntu codename on every PR push: 7 jobs, ~68 min of runner time (org cap: 20 concurrent). Of its 37 PR-run failures between 04 Jul and 05 Aug, only 2 were the sole red signal.
- **What** — The binary-build matrix moves from `pull_request` to `merge_group`, gated by a new required **`Launchpad PPA builds`** aggregate job; the upload jobs now key off `push`/`workflow_dispatch` so a PR event can't reach the PPA. The queue gates *only* on required checks — see #2867, which merged despite a red merge-group run — and the build names embed the codename and PPA URL, so they can't be required directly.
- **Rollout** — Merge this **first**, then add `Launchpad PPA builds` to `main`'s required status checks. Flipping the setting first blocks every open PR on a context that doesn't exist yet; until it's flipped the gate is advisory. (`release-v*` has no required checks configured, so nothing to do there.)

Related: #2902 and #2910 apply the same gate pattern to the packaging and Linux distro workflows.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2910, opened 2026-08-05
ci: run Linux distro builds in the merge queue instead of every PR

jeremyfowers:
- **Why** — Arch and openSUSE container builds ran on every C++-touching PR push: 2 jobs, ~63 min median wall (org cap: 20 concurrent). Of the last 100 PR runs, all 4 red ones had a *required* check failing on the same push — never the sole signal.
- **What** — The matrix moves to `merge_group`, gated by a new required **`Linux distro builds`** aggregate job. The `pull_request` `paths:` filter is dropped so that check always reports — required checks gate enqueueing too, and a workflow that never triggers leaves one pending forever. `push: main` and the `?branch=main` README badges are left alone so the badges keep tracking `main`.
- **Rollout** — Merge this **first**, then add `Linux distro builds` to `main`'s required status checks. Flipping the setting first blocks every open PR on a context that doesn't exist yet; until it's flipped the gate is advisory.

Related: #2902 and #2909 apply the same gate pattern to the packaging and Launchpad PPA workflows.


fl0rianr (commented):
The aggregate gate itself looks correct: it reports cheaply on pull requests and fails closed in the merge queue if the Arch/openSUSE matrix fails, is cancelled, or is unexpectedly skipped.

However, I do not think the current runner-savings model is complete.

As stated before, removing the existing pull-request path filters and adding an unfiltered `merge_group` trigger means the full distro matrix will now run for every queued PR, including documentation-only or otherwise unrelated changes. Because `push: main` is also retained, a regular merge-queue merge runs the same full matrix again immediately after landing.

This means:

* a relevant PR with one push still has two full matrix runs;
* a docs-only PR increases from one full run after landing to two full runs;
* savings only occur for relevant PRs with multiple updates.

Could we keep the always-reporting `Linux distro builds` gate, but add a lightweight change-detection job and only run the merge-group matrix when the diff touches the previous set of relevant paths?

I would also reconsider the duplicate `push: main` matrix or document why preserving the current main-branch badges is worth rerunning the full workflow after an already successful merge-group build.

Finally, `Linux distro builds` still needs to be added to the required checks immediately after this lands; until then, the gate remains advisory.

jeremyfowers:
> a docs-only PR increases from one full run after landing to two full runs;

We're going to add the docs-only fast path so I'm not worried about this.

jeremyfowers:
Superseded by #2911, which aggregates this with the other three merge-queue gate PRs.

PR #2911, opened 2026-08-05
ci: run packaging, PPA, distro, macOS and backend-validation jobs in the merge queue

jeremyfowers:
Packaging, PPA, distro, backend-validation and most macOS jobs run in the merge queue instead of on PR pushes, each group reporting through one gate job under a fixed, requirable name. macOS and self-hosted rig capacity are the binding constraints (org caps: 5 concurrent macOS runners, 20 total).

Aggregates #2901, #2902, #2909, #2910; supersedes #2903.

## What runs where

**Every PR push** — Windows and Linux builds, `.exe`/`.deb` inference rigs, `Test CLI/Endpoints` on Linux, Linux ARM64 and Windows, the embeddable Linux/Windows/macOS builds and their Linux/Windows tests, and every gate job.

**Merge queue only** — every group is reachable from a PR by label, so nothing is merge-queue-only:

| Group | Jobs | Label |
|---|---|---|
| Distros & PPA | Fedora RPM, Debian 13, Arch, openSUSE, Launchpad PPA | `ci:distros` |
| macOS | `.dmg` build, `Test CLI/Endpoints (macos-latest)`, `Test Embeddable (macOS)`, `Test .dmg - macOS inference` | `ci:macos` |
| Backend validation | llama.cpp, vLLM, stable-diffusion.cpp | `ci:upgrades` |

`labeled` is in the `pull_request` activity types of all six workflows, so applying a label takes effect without a push. `unlabeled` is deliberately absent: removing a label only ever drops work, and including it would re-run everything.

`Build Embeddable Lemonade (macOS)` is the one macOS job left on PR pushes: at 4-5m it is the cheapest AppleClang compile check, which is the macOS failure mode coupled to most PRs. Per-push macOS cost is ~5 runner-minutes in 1 job.

## Gate checks

| Workflow | Gate check name |
|---|---|
| `cpp_server_build_test_release.yml` | `Packaging builds` |
| `cpp_server_build_test_release.yml` | `macOS builds` |
| `launchpad-ppa.yml` | `Launchpad PPA builds` |
| `linux_distro_builds.yml` | `Linux distro builds` |
| `validate_llamacpp.yml` | `llama.cpp validation` |
| `validate_vllm.yml` | `vLLM validation` |
| `validate_sdcpp.yml` | `stable-diffusion.cpp validation` |

A gate fails if any job it covers failed, and under `merge_group` also fails if any never ran.

## macOS test coverage

One artifact, one install procedure, one set of test jobs, whether or not Apple signing secrets are available:

- `build-lemonade-macos-dmg` always runs the `package-macos` CMake target and uploads `lemonade-macos-pkg`. Without secrets the binaries are ad-hoc signed and the installer unsigned; `notarize` is a separate target that runs only when secrets exist.
- `test-cli-endpoints-macos` installs that `.pkg` and runs cli, endpoints, websocket idle/auth, tray HTTPS, jobs, router, web-app, model-name normalization, ollama, streaming-errors, llamacpp-system.
- `test-dmg-inference` installs it and runs llamacpp-metal, whisper-metal, moonshine, kokoro-metal. `GGML_METAL_NO_RESIDENCY` goes into the launchd environment *before* the install, so the postflight-started server has it from birth; a `ps eww` check then asserts the listening process actually carries it. Killing that server to start a configured one instead loses a race for `127.0.0.1:13305` — lemond binds the two address families separately and keeps serving when only one succeeds, so the replacement came up IPv6-only and every suite using a literal IPv4 address failed (#2917).

Neither test job branches on signing, and no job runs tests inline in a build job, so a test failure never reports as a build failure.

`workflow_dispatch` accepts `disable_macos_signing` to exercise the unsigned path on demand (it also skips the self-hosted rigs). Run [31064912041](https://github.com/lemonade-sdk/lemonade/actions/runs/31064912041) is green on every macOS job through that path, including the `macOS builds` gate — so `package-macos` succeeds with no installer identity, `sudo installer -pkg` accepts the unsigned result, and both test jobs pass against it.

Merging the suites surfaced three latent bugs, all fixed here: `test_tray_https.py` hardcoded `./build/lemonade-tray`; the `.pkg` starts a single-instance tray agent that blocked the test from starting its own; and #2917. None were reachable before, because these suites had only ever run from a build tree.

## Rules this design follows

- **A required check skips at job level; it is never dropped from a matrix.** A filtered-out matrix leg never creates its check context and leaves the PR pending forever, while a job skipped by `if:` reports success. `test-cli-endpoints` is therefore two jobs, `test-cli-endpoints-windows` and `test-cli-endpoints-macos`. All 21 required contexts are still produced, under unchanged names.
- **No `paths:` filters on these workflows.** Required checks gate enqueueing too, so a workflow that never triggers leaves its gate pending forever.
- **Gated groups are opted into by label, not inferred from paths.** A macOS-specific path filter matches almost nothing; a filter broad enough to be safe matches almost every PR.
- **Steps shared by a gated job and an ungated one live in a composite action** — `.github/actions/smoke-test-embeddable`, used by `test-embeddable-posix` and `test-embeddable-macos`.
- **Secret availability changes what gets published, never what gets tested.** Missing Apple credentials skip notarization only; the package is still built and tested the same way.

## Rollout

1. **Merge this PR.** Gates run immediately but are advisory until step 2.
2. Add the seven gate names above to `main`'s required status checks, spelled exactly as listed.

`docs/dev/testing.md` documents which groups defer, which label opts each back in, and the two consequences: a green unlabeled PR says nothing about macOS or packaging, and a new suite added to a gated job only runs in the queue by default.

Nothing is removed from required checks, so there is no ordering hazard. `release-v*`, cron schedules, `workflow_dispatch`, `push: main` and the README badges are unaffected.


fl0rianr (commented):
Thanks for consolidating the previous PRs. I went through the current version again, and the overall CI design looks solid to me.

The fixed-name gates keep ordinary PR runs cheap, while the merge queue fails closed if a gated job fails, is cancelled, or does not run. The backend workflows now validate the committed pins in the merge queue, and the successful unsigned macOS run gives good confidence in the package-based test path.

There is one scope issue I’d like to clear up before approval: the branch currently includes a large set of website assets under `docs/dev/working-groups/assets/`. They do not appear related to the merge-queue and CI changes, I suspect a slip here.

Could you please remove them from this branch?

One small documentation nit: the PR description says both `labeled` and `unlabeled` activity types are enabled, while the current workflows only include `labeled`.

Once the asset scope is resolved, I’d be comfortable approving this. Thanks for the thorough work here, especially around the macOS packaging and test path.

fl0rianr (approved):
LGTM now!

Please be aware of these two findings:

**1. Rollout detail:**
When the gated matrix jobs are skipped by the job-level if, GitHub creates a single skipped check with the matrix expression still in its name, for example:

Build Lemonade .rpm - Fedora ${{ matrix.fedora-version }}${{ matrix.job-suffix }}
Build Lemonade .deb Package (Debian 13)${{ matrix.job-suffix }}
Smoke test .deb (Debian 13)${{ matrix.job-suffix }}

It does not create the concrete per-matrix check names on an unlabeled PR, e.g. label error like recently.

This is fine if none of those expanded Fedora or Debian contexts are currently required on main. Since the PR says all 21 existing required contexts remain available under unchanged names, could you confirm that the current required-check list does not include these matrix contexts?

**2. One operational note we should be aware of:**
The stable-diffusion.cpp gate still treats the availability of the committed Metal and CUDA release assets as part of merge validation. That means an upstream API or asset-availability issue can block the queue.

That's all. Thanks.

Issue #2912, opened 2026-08-05
Bump sdcpp versions to >= 727


klkr1:
### Feature Description

update backend version of sdcpp to at least 727

### Use Case / Motivation

current version of sdcpp uses instruction set for the sdcpp backend which is not supported by older CPUs (e.g. AMD FX-6100) so backends crash with illegal instruction. Version 727 fixes this issue.

### Platform Relevance

All platforms

klkr1:
implemented with #2768

PR #2913, opened 2026-08-05
update sdcpp backend versions to 727 to support older CPUs

klkr1:
Current version of sdcpp backend uses instructions which are not supported by older CPUs (e.g. AMD FX-6100) and crash with illegal instruction.

Support of older CPUs was introduced with [#1704 on leejet/stable-diffusion.cpp](https://github.com/leejet/stable-diffusion.cpp/pull/1704). Update to master-727-f54e45e of sdcpp backend will include support of older CPUs into lemonade sdcpp.

Solves #2912

### DIFF
--- a/src/cpp/resources/backend_versions.json
+++ b/src/cpp/resources/backend_versions.json
@@ -61,11 +61,11 @@
     "metal": "v1.8.4"
   },
   "sd-cpp": {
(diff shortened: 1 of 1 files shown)


klkr1:
sdcpp already updated to version > 727 with #2768

PR #2914, opened 2026-08-05
fix(cli): respect explicit --port when --host includes a scheme

kenvandine:
## Issue
**Release blocker:** Explicit `--port` is ignored when `--host` includes a scheme.

When a user specifies both `--host` with a scheme (e.g., `http://127.0.0.1`) and an explicit `--port` (e.g., `18454`), the port was being overridden with the scheme's default (80 for HTTP, 443 for HTTPS).

### Example
```bash
lemonade --host http://127.0.0.1 --port 18454 list
# Incorrectly connects to port 80 instead of 18454
```

## Root Cause
The `parse_target_url()` function was being called with the default `override_default_port=true`, which caused it to replace the explicit port with the default for the scheme when the host string lacked a port.

Regression introduced in commit `8ed3d890` (feat(tray): support HTTPS server configurations).

## Solution
- Detect whether `--port` was explicitly provided by the user
- Pass `override_default_port=false` to `parse_target_url()` when the port is explicit, preserving the user's choice
- Updated the `LemonadeClient::parse_target_url()` method signature to accept and forward the parameter

## Changes
1. **src/cpp/cli/main.cpp** (line 1463-1467):
   - Check if `--port` was explicitly set: `bool explicit_port = port_opt->count() > 0`
   - Pass `!explicit_port` as the `override_default_port` parameter

2. **src/cpp/cli/lemonade_client.cpp** (line 87-89):
   - Updated method to accept and forward `override_default_port` parameter

3. **src/cpp/include/lemon_cli/lemonade_client.h** (line 77):
   - Added `override_default_port = true` parameter to method signature for backward compatibility

## Testing
- Verified the fix doesn't break existing behavior (default ports still work when scheme is present but no port in URL)
- The pattern matches the correct implementation in `src/cpp/tray/main.cpp` (lines 317-318)
- Does NOT affect v11.5.1 (bug doesn't exist there)
- Only affects v11.5.2 release candidate (regression introduced there)

## Impact
- **Affects:** CLI client (`lemonade` command) when connecting to a server with URL scheme and explicit port
- **Doesn't affect:** The `lemond` server itself
- **Scope:** CLI-only regression fix


kpoineal (approved):
LGTM! The explicit --port regression is fixed: an explicit port is now preserved when --host includes an HTTP(S) scheme.

PR #2915, opened 2026-08-05
ci: isolate llama.cpp validation cleanup on self-hosted runners

fl0rianr:
##  Summary

Make the llama.cpp validation job safer and more reproducible on persistent Windows self-hosted runners.

This change replaces broad process-name termination with cleanup scoped to processes owned by the current repository workspace or validation cache.

## Why

The validation runners are persistent machines. A failed or cancelled run can leave lemond or llama-server processes behind, which may hold file locks, consume memory, or affect later performance measurements.

The previous cleanup also matched processes by broad name patterns. On a shared runner, that could terminate processes belonging to another job or a manually running Lemonade instance.

The goal of this change is to start each validation from a known process state without changing the benchmark itself.

## Changes

* remove the broad process-name kill from the self-hosted validation job
* clean up stale lemond processes only when their executable belongs to the current workspace
* clean up stale llama-server processes only when they belong to a validation cache in the current workspace
* fail early when unrelated inference processes are already running
* fail early when the validation port is already occupied
* record the lemond PID for the existing cleanup action
* prefer the server's graceful shutdown endpoint
* use targeted process-tree termination only as a fallback
* check separately for orphaned llama-server processes after shutdown
* preserve the validation exit code across cleanup commands
* verify the native exit code from lemond --version

* capture a read-only Vulkan runner configuration snapshot before cleanup/setup
* upload GPU, driver, Windows, BIOS, power, uptime, process state, and full
  `vulkaninfo` output as a per-run artifact

## Measurement scope

This PR intentionally does not change:

* test/validate_llamacpp.py
* model order
* prompts or token limits
* the number of inference requests
* backend installation behavior
* model or driver caches
* power plans or process priorities
* the validation result format
* runner metadata collection is read-only and does not modify driver caches,
  power settings, drivers, BIOS, or firmware

The workflow lifecycle is isolated from the benchmark logic so existing TTFT results remain comparable.

This change reduces interference from stale or unrelated processes. It does not attempt to make cold TTFT fully deterministic, since later models can still inherit memory, driver, and cache state from earlier models in the same validation run.

### DIFF
--- a/.github/workflows/validate_llamacpp.yml
+++ b/.github/workflows/validate_llamacpp.yml
@@ -422,6 +422,178 @@ jobs:
           # - backend: cuda
           #   runner: [self-hosted, Windows, 128gb, cuda]
     steps:
+      - name: Capture Vulkan runner configuration
+        if: matrix.backend == 'vulkan'
+        shell: PowerShell
+        run: |
+          $ErrorActionPreference = "Continue"
+
+          $outDir = Join-Path $env:RUNNER_TEMP "vulkan-runner-config"
+          New-Item -ItemType Directory -Force -Path $outDir | Out-Null
+
+          # General system / Windows / BIOS information
+          $os = Get-CimInstance Win32_OperatingSystem
+          $computer = Get-CimInstance Win32_ComputerSystem
+          $bios = Get-CimInstance Win32_BIOS
+
+          [pscustomobject]@{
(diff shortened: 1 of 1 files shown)


bitgamma (approved):

jeremyfowers:
Update + requeue after https://github.com/lemonade-sdk/lemonade/pull/2950 merges

PR #2921, opened 2026-08-06
GUI3: Apps Marketplace, shared catalog layout, and engine logos

jeremyfowers:
**Apps Marketplace — Featured row + curated category sections**
Apps Marketplace overview

**Category sections keyed by each app's primary label**
Category sections

**Sidebar filter view (labels from marketplace category metadata)**
Filtered category view

**Light theme**
Apps light theme

**Backends — engine logo plates**
Backends with engine logos

**Dark plates for dark-fill logos, text plates when no logo exists, icon+name lockup for Moonshine**
Backends audio section

**More logo plates (TTS / 3D sections)**
Backends TTS section

**Responsive layout (shared with Backends via the new catalog layout contract)**
Apps mobile

### DIFF
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -6,7 +6,7 @@ import { findModelInfoByName, isCollectionFullyLoaded, isCollectionModel, withVi
 import ChatView from './components/ChatView';
 import ModelManager from './components/ModelManager';
 import ConnectView from './components/ConnectView';
-import AppsView, { MARKETPLACE_URL, type MarketplaceApp } from './components/AppsView';
+import AppsView, { MARKETPLACE_URL, type MarketplaceApp, type MarketplaceCategory } from './components/AppsView';
 import BackendManager from './components/BackendManager';
 import DownloadManager from './components/DownloadManager';
 import MonitorView from './components/MonitorView';
@@ -260,6 +260,7 @@ const App: React.FC = () => {
   const [downloadManagerOpen, setDownloadManagerOpen] = useState(false);
   const [modelDetailsRequest, setModelDetailsRequest] = useState<{ modelName: string; nonce: number } | null>(null);
   const [marketplaceApps, setMarketplaceApps] = useState<MarketplaceApp[]>([]);
+  const [marketplaceCategories, setMarketplaceCategories] = useState<MarketplaceCategory[]>([]);
   const [marketplaceError, setMarketplaceError] = useState<string | null>(null);
   const [marketplaceLoading, setMarketplaceLoading] = useState(true);
   const [utilityMenuOpen, setUtilityMenuOpen] = useState(false);
@@ -283,6 +284,7 @@ const App: React.FC = () => {
       .then(data => {
         if (cancelled) return;
         setMarketplaceApps(Array.isArray(data?.apps) ? data.apps as MarketplaceApp[] : []);
+        setMarketplaceCategories(Array.isArray(data?.categories) ? data.categories as MarketplaceCategory[] : []);
         setMarketplaceError(null);
       })
       .catch(error => {
@@ -903,6 +905,7 @@ const App: React.FC = () => {
           <ViewErrorBoundary view="apps">
             <AppsView
               apps={marketplaceApps}
+              categories={marketplaceCategories}
               loading={marketplaceLoading}
               error={marketplaceError}
             />
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -742,10 +742,14 @@ const App: React.FC = () => {
                       type="search"
                       value={navigationSearch}
                       placeholder="Search"
+                      role="combobox"
                       aria-label={view === 'apps' ? 'Search apps' : 'Search Lemonade'}
                       aria-autocomplete="list"
                       aria-expanded={navigationSearchOpen}
                       aria-controls="titlebar-search-results"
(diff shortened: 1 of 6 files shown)


kpoineal:
## Visual review screenshots

The shared catalog layout is a strong improvement overall. These are the two visual issues I would address before merging:

### 1. Featured apps repeat in their category sections

`Open WebUI` appears once in **Featured** and again in **Chat** on the same default catalog page. Because every featured app remains in its category group, the marketplace reads as accidentally duplicated rather than intentionally curated.

Open WebUI duplicated between Featured and Chat

I would either exclude featured apps from the category sections on the default page or remove the separate Featured section.

### 2. Engine artwork needs per-logo normalization

The shared 48 px logo plate works well for broad wordmarks such as llama.cpp and vLLM, but the narrow RyzenAI asset becomes a tiny mark inside a large black rectangle. This makes one of Lemonade's most important NPU backends look visually secondary.

RyzenAI logo losing visual weight next to other engines

I would add per-logo sizing/cropping variants or use normalized local assets rather than forcing every upstream image through one `max-height` rule.

The Apps screenshot uses the live marketplace feed. The Backends screenshot uses representative mocked `/system-info` data so the card treatment can be compared without a running `lemond`.

### 3. "All apps" and "Apps Marketplace" describe the same state differently

The selected rail item says **All apps**, while the pane it activates is headed **Apps Marketplace**. Every category filter otherwise maps directly to the heading users see (`Chat` → `Chat`, `Code` → `Code`), so the default state breaks that otherwise-clear relationship.

All apps filter activating an Apps Marketplace heading

I would use **Apps Marketplace** as the page title and rename the filter to **All**, or use **All apps** consistently in both places.

### 4. Linkless entries retain an empty action area

The card footer and action group are rendered even when an app has no links. Combined with stretched grid rows, this leaves a conspicuous blank action area instead of letting the card content define its natural ending.

Linkless app card retaining an unused footer area

This screenshot uses representative mocked marketplace data because every entry in the current live feed has at least one link. It is an edge-case hardening issue rather than a current-feed defect. Rendering the footer only when at least one action exists would resolve it.

fl0rianr:
> 1. Featured apps repeat in their category sections

This looks fine for me, since you can be featured by the Lemonade team and chat

Other points are good catches thanks.

I would be ok with All apps --> Apps Marketplace since all is implicit clear, but a better solution is welcome.

**What I like to see:**
Having Backends the same logo size like Apps/Marketplace.

Those huge badges that is too much visual load.

fl0rianr:
Not this PRs fault, but the Lemonade connection dot misses the correct glowing effect on the top-panel and the settings tab in the lower rail section.
image
While the Monitor / Stream filers dot renders correctly. 

Would be great if you can fix this one as well whilst your working on it.

jeremyfowers:
> The card footer and action group are rendered even when an app has no links.

This is not possible. We would never have an app with no links.

jeremyfowers:
> Not this PRs fault, but the Lemonade connection dot misses the correct glowing effect on the top-panel and the settings tab in the lower rail section.

This is out of scope, I will not address it. Please make a separate PR.

jeremyfowers:
@kpoineal:

1. Featured apps repeat in their category sections - FIXED
2. Engine artwork needs per-logo normalization - Removed the Ryzen AI SW logo, I didnt see it before because I'm on ubuntu.
3. "All apps" and "Apps Marketplace" describe the same state differently - FIXED, this was just a bug with the filters
4. Linkless entries retain an empty action area - WONT FIX, impossible scenario

@fl0rianr:

1. Not this PRs fault, but the Lemonade connection dot misses the correct glowing effect on the top-panel and the settings tab in the lower rail section. - WONT FIX, I'm not working on that.
2. Having Backends the same logo size like Apps/Marketplace. - This is not possible, the backends don't have icons that can be placed next to text like apps do.

Please re-review

fl0rianr (commented):
Thanks for the updates, the overall direction looks good.

I found three things worth fixing before merge:

* `npm run test:ui-regressions` still expects `WorkspaceSectionRail` and the old `railCollapsed` handling, so the regression test should be updated for the new shared layout.
* Category filters now only use the first category. For example, an app tagged `productivity` and `chat` will no longer appear under Chat. Grouping by the primary category is fine, but filtering should probably still match all assigned categories.
* The filter description/count is only exposed through `title` now. It would be good to keep it available to screen readers via `aria-describedby` or similar.

Nice improvement overall.

fl0rianr (approved):
LGTM now

kpoineal:
## Visual re-review

Re-reviewed through `890e4bd8`. The previous blocking visual concerns are addressed:

- Featured apps no longer repeat in their category sections.
- RyzenAI now uses the text fallback instead of the undersized upstream badge.
- The **Apps Marketplace** page heading remains stable while category labels act as filters, which makes the naming relationship clear.
- Dark transparent app artwork receives a white tile; the updated cards remain coherent in both dark and light themes.
- Secondary app categories now remain reachable through the filter rail without duplicating cards in the default section layout.

I did not find a new blocking layout, responsive, or accessibility regression in these changes.

The linkless-card footer noted earlier remains an edge case: the current marketplace feed gives every app at least one link, but a future linkless entry would still retain unused action space because the footer is always rendered. I consider that non-blocking hardening rather than a current-feed defect.

kpoineal (approved):
Re-reviewed the latest changes. The visual feedback has been addressed, and the remaining linkless-card behavior is a non-blocking edge case.

PR #2922, opened 2026-08-06
GUI3: Fix status indicators and Apps test, add optional backend logos

fl0rianr:
## What changed

* Fixed the Lemonade connection indicator so the top bar and Settings rail use the same green glow as the Monitor live-status dot.
* Updated the Apps feature test to use the input’s actual combobox role, fixing the failing Playwright test.
* Capitalized the navigation label to **All Apps**.
* Backend tab: Added a **Show logos** checkbox below **Show unsupported backends**.

  * Enabled by default to preserve the current design.
  * When disabled, backend cards return to the previous text-only presentation.

## Why

Backend logos help many users recognize engines quickly, but their different colors, shapes, and background plates can also add visual noise or make the list harder to scan. This keeps logos as the default while giving sensitive users the choice to restore the simpler presentation used before PR #2921.

With logos:

image

Without as before:

image

### DIFF
--- a/src/app/src/components/AppsView.tsx
+++ b/src/app/src/components/AppsView.tsx
@@ -217,7 +217,7 @@ const AppsView: React.FC<AppsViewProps> = ({
   const categoryFilters = useMemo<CatalogFilterDefinition<string>[]>(() => [
     {
       id: ALL_APPS_SECTION,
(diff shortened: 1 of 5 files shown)


fl0rianr:
merging as agreed with Jeremy

PR #2927, opened 2026-08-07
[backends] Add support for image generation through TheNoise

bitgamma:
Adds support for image generation using [TheNoise](https://github.com/lemonade-sdk/thenoise) backend. This is only available on linux gfx1150 and gfx1151 for now. It support Krea 2 and Anima models.

How to test:

You need a Strix Halo or Strix Point.

- In the webui, from TheNoise ROCm (experimental) select Anima-turbo-v1.0
- wait for it to be download, then load it
- write a prompt. First generation takes longer because torch.compile is active

The backend supports negative prompts, loras and various options. The lora_dir is only available at load time, while the other options can be either backed in the recipe or overridden per-request. This allows creating presets like I am doing in the screenshot below

lemond-noise


test/server_thenoise.py
```diff
@@ -0,0 +1,354 @@
+"""
```
jeremyfowers, line 1: would it make sense to rename the stable-diffusion.cpp test suite to "image generation test suite", and then it could share with thenoise instead of introducing a new file?

fl0rianr (commented):
Nice addition overall, I found a few things worth checking before merge:

1. TheNoise reuses generic recipe option names like `steps`, `width`, and `height`. `RuntimeConfig::recipe_options()` currently flattens options from all backend descriptors into one map, so TheNoise defaults can overwrite values from another backend like `sd-cpp`. This can become visible especially for models that don't override these parameters themselves via model-specific options/defaults.

2. `guidance_scale` is exposed as a TheNoise config option, but the runtime config validator currently only recognizes `cfg_scale`, so updating `thenoise.guidance_scale` through the runtime config API looks like it would be rejected.

3. `/text2image` uses a direct `HttpClient::post()` with no timeout and bypasses the usual `WrappedServer` request scope/watchdog handling. If that request hangs while the backend process stays alive, it could potentially block indefinitely.

Other than that, the TheNoise API/CLI integration and release asset handling look consistent from what I checked.

test/server_thenoise.py
bitgamma: would it work, considering that stable-diffusion.cpp can run on targets/runners where thenoise can't?

bitgamma:
@fl0rianr thank you for your review.

1. I now removed all those parameters from TheNoise config. They are still used in recipe options, meaning whatever is set for stable-diffusion.cpp also becomes the fallback for thenoise. I think this is acceptable since those fields have the same semantic, form and range. They should probably be moved to the general section, similar to ctx_size or be removed as "backend defaults" as they usually are part of either the model defaults or the image gen request

2. Now properly using cfg_scale consistently and only mapping it to guidance_scale when building the request

3. You are right, this is because thenoise was returning a raw PNG but forward_request expects a JSON. I updated thenoise so output json so I can now use the existing helper as-is

fl0rianr (commented):
Thanks for the updates. There are two things I would still address before merge:

1. **Recipe default collision still exists**

Removing the image parameters from the TheNoise config fixes the `RuntimeConfig` side, but the fallback defaults are still stored in one flat map in `RecipeOptions::get_defaults()`.

Both `sd-cpp` and `thenoise` define generic keys such as `steps`, `cfg_scale`, `width`, and `height`. Since TheNoise is registered after `sd-cpp`, its defaults overwrite the `sd-cpp` defaults in that map.

This should mostly be hidden for models that provide their own `image_defaults`, but an `sd-cpp` model without those overrides can now fall back to the TheNoise values instead.

2. **gfx1152 is advertised but rejected during installation**

The descriptor now lists `gfx1152` as supported, and the 0.1.1 TheNoise release also contains a `gfx1152` archive.

However, `get_install_params()` currently only accepts `gfx1150` and `gfx1151`, so a gfx1152 system will still hit the unsupported-backend error before the archive can be selected.

Other than these two points it's ready, great work!

bitgamma:
@fl0rianr thanks again for your feedback

1. I don't see a solution to this problem that doesn't require changes also to the sdcpp backend. I can do that in a followup, in the meantime I just aligned the default to be exactly the same as sdcpp for the shared parameters. It won't matter in practice anyway since each model needs its own defaults

2. good catch, I removed that check altogether since it is redundant

test/server_thenoise.py
fl0rianr: I think this should work even with the different runners. The workflow can still have separate matrix entries for `sd-cpp` and `thenoise`, with their own backends/runners, while both point to the same image generation test script.

We already do something similar with `server_audio_generation.py` for multiple backends.

fl0rianr (commented):
Thanks, the approach for the shared recipe defaults makes sense to me. I'm fine with handling that separately.

I noticed one small regression in the gfx1152 fix though: `version` already contains `rocm7.14.0`, so

`version + "-rocm7.14" + target_arch + "-x64.tar.gz"`

no longer matches the published assets. I think this should remain:

`version + "-" + target_arch + "-x64.tar.gz"`

With that fixed and CI green, this looks good to me.

bitgamma:
@fl0rianr ouch, I am sorry, I thought I had undone that edit before committing. Thanks once again

fl0rianr (approved):
LGTM now, thanks!

fl0rianr:
Fyi, you have to resolve the discussion, a follow up would be fine with me

test/server_thenoise.py
bitgamma: I'll handle this in a followup

PR #2942, opened 2026-08-07
fix(gui): flatten color gradients in the chat panel

### LINKED ISSUE #2934: Flatten the gradients in the chat tab
Image

Image

We have a flatter, more glassmorphic design language everywhere else in the app, so these stand out quite a bit.

fl0rianr:
fixes #2934

image

### DIFF
--- a/src/app/src/styles/styles.css
+++ b/src/app/src/styles/styles.css
@@ -626,9 +626,7 @@ input, textarea {
 .rail__item:hover { background: var(--surface-2); }
 
 .rail__item.is-active {
-  background: linear-gradient(90deg,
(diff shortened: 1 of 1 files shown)


jeremyfowers:
Apologies, the wrong screenshot went into the issue. It was the model loaded card, not the chat history list item, where the gradient needed to be removed. Please see the updated issue.

fl0rianr:
I'm assuming we are removing it from both (what I already did + the other case), please state if otherwise.

fl0rianr:
Now here flat color as well

image

jeremyfowers (approved):

Issue #2937, opened 2026-08-07
Models search bar


jeremyfowers:
<img width="1034" height="322" alt="Image" src="https://github.com/user-attachments/assets/7e547d77-175e-4f11-89c3-76c4ebc4bd96" />

1. Remove the Ctrl+K shortcut since that also maps to the top search bar
2. Make the copy make it obvious that typing here will search the internet too

fl0rianr:
Please let me know if this one would work out for you, thanks.

<img width="329" height="376" alt="Image" src="https://github.com/user-attachments/assets/e7755bda-85ad-46b2-92bf-31027d2a25f3" />

<img width="336" height="368" alt="Image" src="https://github.com/user-attachments/assets/04ee8e46-5b19-48fd-90e9-9602897be41c" />

<img width="348" height="388" alt="Image" src="https://github.com/user-attachments/assets/c8974ea6-ae89-4894-93e9-9c4c9f9a11e2" />

<img width="341" height="364" alt="Image" src="https://github.com/user-attachments/assets/09280ddb-4284-43fd-85da-0596825badda" />

jeremyfowers:
I would prefer it say "Search built-in and online catalogs..."

The "..." shouldn't be in the middle of the phrase

I like the touch of hiding the online catalog part if online search is disabled.

PR #2944, opened 2026-08-07
GUI3 fix: avoid strange 0 when there is no online model search

### LINKED ISSUE #2939: Refine online search filters on the Models tab
Image

Having highlighted controls that say 0 by default is confusing. Should these controls just move into Settings? Should they only display when an online search is taking place (ie, the value is non-0)?

fl0rianr:
## Summary

* Hide provider result counts when they are **0**
* Keep positive remote search result counts visible
* Avoid showing misleading zero-result text to screen readers

Fixes #2939.

image
image
image

### DIFF
--- a/src/app/src/components/ModelManager.tsx
+++ b/src/app/src/components/ModelManager.tsx
@@ -2744,6 +2744,7 @@ const ModelManager: React.FC<ModelManagerProps> = ({ onModelSelect, openModelReq
         onTagFiltersChange={setTagFilters}
         providerEnabled={providerEnabled}
         providerCounts={providerCounts}
+        searchActive={searchActive}
         onToggleProvider={toggleProvider}
         storageInfo={storageInfo}
         collapsed={navRailCollapsed}
--- a/src/app/src/components/ModelNavRail.tsx
+++ b/src/app/src/components/ModelNavRail.tsx
@@ -130,6 +130,7 @@ export interface ModelNavRailProps {
   onTagFiltersChange: (next: Set<string>) => void;
   providerEnabled: Record<ModelRegistryProvider, boolean>;
   providerCounts: Record<ModelRegistryProvider, number>;
+  searchActive: boolean;
   onToggleProvider: (provider: ModelRegistryProvider) => void;
   /** Real disk usage of the model-storage drive, when available (else null →
       derived fallback). Sourced from api.getStorageInfo(). */
@@ -160,6 +161,7 @@ export const ModelNavRail: React.FC<ModelNavRailProps> = ({
   onTagFiltersChange,
   providerEnabled,
   providerCounts,
+  searchActive,
   onToggleProvider,
   storageInfo,
   id = 'model-nav-rail',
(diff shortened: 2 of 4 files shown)


jeremyfowers:
This ui doesn't make sense to me. I'm not clear why these filters are needed at all; but if they are they definitely shouldn't carry a unique highlight style. Please provide a comprehensive fix or close the PR.

fl0rianr:
I'm confused here.

These are the model providers and they are activated for their online search and deactivated if you not searching online. A very useful feature, inside of GUI3 for quite some time already. You can switch super easily online/offline how you want and what provider you want on the fly - I like it very much. A significant UX regression for me moving it towards the settings.

**Removing the 0 is the only change so far**, If they are deactivated the weren't highlited anymore.
I can adapt it that they never highlight and only the icon is showing it  - if this gives you a more comprehensive feeling,

fl0rianr:
I think I get your point better now, still hope we can keep it on the model site quite fitting for model providers. 
If we add it to the setting I would do it with the settings PR to enable and disable a provider for search permanently but I don't think that's needed. Setting will get enough content over time on it's own and we should avoid overcrowding.

Changed towards this more intuitive version which will fit very well with the search online info change I'm working on:


image
image

image

jeremyfowers:
> This ui doesn't make sense to me. I'm not clear why these filters are needed at all; but if they are they definitely shouldn't carry a unique highlight style. Please provide a comprehensive fix or close the PR.

Sorry I wrote this in a rush earlier and it came out harsher than I meant it to. I will give more concrete guidance when I can.

fl0rianr:
I have one improvement idea: Having the same rounded edge box around the cloud icon makes it even more clear. This box could also turn gray once disabled.
If this is MVP for you might consider approving since the PR for https://github.com/lemonade-sdk/lemonade/issues/2937 depends on this. 
Since this is important we can also go the guidance route of course. Thanks.

jeremyfowers:
Here's some more guidance:
- The filters should be ordered top-to-bottom by how frequently they will be used and how critical their task is. It is really hard for me to imagine that most people's 2nd highest priority task in the model browser will be to frequently toggle search providers on and off. Most people will never touch this, and others will make a choice once and then not touch it again. It doesn't earn this level of vertical priority.
- The heading "Model-Provider" is not sufficiently clear. Needs to be something like "Online Catalog", "Online Search", "Online Provider", "Model Search Provider", etc.
- It's not clear for the user whether clicking the toggle prevents the search entirely or just hides the search results from that provider. 
- It's not obviously an on-off toggle for anyone who hasn't clicked it yet. It also doesn't match the style of any of the other on/off toggles in the app.

The solution to #2939 needs to solve all 4 of these problems comprehensively.

jeremyfowers:
My suggestion: turn these into checkboxes and put them at the bottom of the filter panel. 

Online Search:
- [ ] Hugging Face
- [ ] ModelScope

fl0rianr:
Thank you for the detailed guidiance!

image
image
image

1. Same checkbox as in backend and Online Catalogs as name
2. Numbers are summed since the specific result number is shown in the models list anyway and matches the filters above
3. 0 is now displayed all the time unless removing the number is a clear indicator the search is disabled now.

**Reasoning for the nav model tab:**
1. user persona A does not change defaults and leaves it on: Having this shown is not a significant degression.
2. user persona B wants to avoid online search while searching through build-in/local models. If he then want's to look for a online model having to go to settings is a significant degression.

jeremyfowers (changes requested):
- [ ] Don't show 0 when there is no search taking place.
- [ ] Since clicking the words "Online Catalogs" itself is not a filter, those words need to use the heading style, right now it matches the style of the filters.
- [ ] Put the numbers next to the providers, not the heading
- [ ] horizontally align the checkboxes to the other interactive elements, e.g., the task selectors.
- [ ] Move the Online Catalogs section below the backends section

fl0rianr:
Thanks, very helpful! The position before was non-ideal, a misunderstanding.

image
image
image
image

jeremyfowers (approved):
LGTM! opened as a follow up: https://github.com/lemonade-sdk/lemonade/issues/2999

PR #2950, opened 2026-08-07
ci: run .exe/.deb inference suites in the merge queue

jeremyfowers:
Extends the #2911 scheme to the `.exe`/`.deb` inference suites: all `Test .exe - *` and `Test .deb - *` jobs now run in the merge queue instead of on PR pushes, freeing the self-hosted rigs (the binding constraint under the 20-runner org cap) for merge-gating work.

- Both jobs keep their existing signing/tag skip; the new condition only defers `pull_request` runs.
- New gate job `backends-gate` reports as **`Inference backend tests`**, same pattern as `packaging-gate`/`macos-gate`: fails if any covered job (including the upstream `.exe`/`.deb` builds) failed, and under `merge_group` also fails if any never ran.
- Label a PR **`ci:backends`** to opt the suites back in on that PR; `labeled` is already in the workflow's `pull_request` activity types, so the label takes effect without a push. (Label created to match the existing `ci:*` set.)
- `docs/dev/testing.md` gains the new row in the deferred-groups table, and the stale "CI tests pull requests on real AI PC hardware" claims in the dev docs are updated to reflect the deferral.

Rollout: after merging, update `main`'s required status checks — add `Inference backend tests` and **remove any individual `Test .exe - *` / `Test .deb - *` contexts that are currently required**. A label-skipped matrix job reports a single check named literally `Test .exe - ${{ matrix.name }}`, so per-leg contexts never materialize on unlabeled PR heads and must not stay required.


fl0rianr (approved):
LGTM

SlawomirNowaczyk (approved):

PR #2880, opened 2026-08-02
feat(server,router): add dynamic custom backend discovery and process isolation

abn:
Introduces a dynamic custom backend abstraction layer (`ExternalBackendServer`) and JSON descriptor registry (`BackendDescriptorRegistry`) allowing `lemond` to discover, validate, and manage third-party or containerized inference engines without C++ code changes.

### Rationale & Context
- Enables running custom binaries (e.g. customized `llama.cpp` builds, `vLLM`, containerized podman/docker runners) while maintaining `lemond`'s multi-model LRU scheduling and OpenAI/Ollama API routes.
- Enforces strict security invariants: filesystem permission checks on POSIX (`0022` mode mask + sticky-bit `/tmp` exceptions) and Windows (DACL SID checks), pre-spawn environment secret scrubbing, and log redaction.

### Key Changes
- **Descriptor Registry & Discovery**: `BackendDescriptorRegistry` auto-discovers `.json` descriptors across user and system search paths with strict owner and permission validation.
- **Subprocess Isolation**: `ExternalBackendServer` executes external commands with template token expansion (`{port}`, `{model_dir}`, `{exe_dir}`, `{rocm_arch}`, `{cuda_arch}`, etc.), `protected_flags` validation, and extension-preserving `lexically_relative` path token resolution over HF cache symlinks.
- **Capability Gating**: Router enforces capability checks (`supports_capability<T>`), returning an `unsupported_operation` error envelope when operations are unsupported by a custom backend.
- **Tests & Docs**: Added native C++ unit test suite (`test_custom_backends`), live server Python integration suite (`test/server_custom_backend.py`), configuration guide, and technical specification.

PR #2951, opened 2026-08-07
docs(wg): add charter for Backend Plugins & Sandboxing working group

abn:
### Summary

This PR introduces the charter for the **Backend Plugins & Sandboxing** working group (`docs/dev/working-groups/backend-plugins-sandboxing.md`).

### Core Roadmap Objectives

- **Plugin Specification & External Adapter Architecture**: Transition from statically linked C++ `WrappedServer` classes to self-describing JSON descriptors with local recipe loading (`lemonade backends install <path>`) and slot management.
- **Process Sandboxing Core (`nono`) & Permission Policy Engine**: Enforce least-privilege OS kernel sandboxing (`nono` / Landlock / seccomp / platform primitives) by default for backend binaries.
- **Backend Recipe Marketplace**: API-first Git repository registry (`org/repo[/subdir]`) for discovering, fetching, and installing third-party recipes (`lemonade backends install <repo/recipe>`).
- **Client Experience & Tooling**: Expose plugin management, permission consent dialogs, and sandbox status across CLI and Tauri/Web UI.

### DIFF
--- a/docs/dev/working-groups/README.md
+++ b/docs/dev/working-groups/README.md
@@ -23,6 +23,7 @@ You can see the full set of project maintainers, and the subject areas they main
 | Cloud Hybrid | @ramkrishna2910 | @ramkrishna2910 | Lemonade can intelligently route between local and cloud models. |
 | Remote Use | @Geramy | @geramyl | Lemonade can serve inference to any device in any location. |
 | GUI App | @kponiel | @primaL- | Users can explore local AI in a delightful built‑in GUI. |
+| [Backend Plugins & Sandboxing](./backend-plugins-sandboxing.md) | @abn | @abn.null | Dynamic plugin architecture, marketplace, client UX, and kernel sandboxing for inference backends. |
 | [Enterprise Grade](./enterprise-grade.md) | @jeremyfowers | @jfowers_amd | Provide systematic testing and quality assurance for Lemonade components. |
 
 ## Archived Working Groups
--- a/docs/dev/working-groups/backend-plugins-sandboxing.md
+++ b/docs/dev/working-groups/backend-plugins-sandboxing.md
@@ -1,0 +1,195 @@
+# Working Group: Backend Plugins & Sandboxing
+
+## Overview
+
+**Lead:** This working group is led by Arun Babu Neelicattu, whose handle is @abn on GitHub and @abn.null on Discord.
+
+**Background:** Lemonade runs inference by spawning backends as subprocesses (`llama-server`, `whisper-server`, `sd-server`, `flm`, `vllm`). Currently, adding a backend requires modifying C++ server sources and recompiling `lemond`. This limits Lemonade to in-tree backends and slows adoption of new inference engines.
+
+Running third-party or community backend binaries introduces security risks (unauthorized file access, network egress, privilege escalation). Experimentation with `nono.sh` demonstrated that backend subprocesses can be restricted to least-privilege filesystem and network boundaries without impacting inference performance.
+
+**Why:** Enables adding third-party backends without C++ code modifications while securing backend subprocesses with OS kernel sandboxing.
+
+**Goal:** Enable `lemond` to orchestrate out-of-tree backends under kernel sandboxing by default, supported by a recipe marketplace and CLI/GUI management tooling.
+
+---
+
+## Scope Boundaries
+
+- **In Scope:** Descriptor schema definition, out-of-process HTTP/socket backend adapter, process sandboxing (`nono` on Linux, platform primitives on Windows/macOS), recipe marketplace, client UI integration.
+- **Out of Scope:** Engine binary compilation/build infrastructure. Lemonade orchestrates and sandboxes pre-built backend binaries; it does not build them.
+- **Related Work:** Deep container engine integration and containerized backend orchestration are recommended as a follow-on **Containerized Backends & Isolation Working Group**.
+
+---
+
+## Proposed Architecture
+
+```mermaid
+flowchart TD
+    Client["Client (CLI / GUI / SDK)"] -->|"HTTP API"| Server["lemond (Server Core)"]
+    Server --> Registry["Backend Descriptor Registry"]
+    Server --> Router["Router & Slot Manager"]
+
+    subgraph Plugin System
+        Registry -->|"Loads"| Descriptor["JSON Descriptors (~/.config/lemonade/backends/)"]
+        Router --> Adapter["External Backend Adapter"]
+    end
+
+    subgraph Process Sandboxing (nono)
+        Adapter -->|"HTTP / IPC over socket"| S1["llama-server"]
+        Adapter -->|"HTTP / IPC over socket"| S2["Custom vLLM Backend"]
+        Adapter -->|"HTTP / IPC over socket"| S3["NPU Engine"]
+    end
+```
+
+### 1. Backend Plugin Specification
+
+Backends transition from statically linked C++ `WrappedServer` classes to declarative JSON descriptors.
+
+- **Descriptor Schema:** Declares metadata, capability interfaces (`ICompletionServer`, `ITranscriptionServer`, etc.), a `Host OS × Accelerator` compatibility matrix, launcher command templates, and default recipe mappings.
+- **Interface Contract:** Standardized health checks, load/unload lifecycle endpoints, and API passthrough.
+- **Slot Policies:** Declarative slot management (`standard`, `exclusive_npu`, `coexist_by_type`, `unmetered`) mapping directly to `lemon::SlotPolicy` enum values enforced by the Router.
+- **Platform Matrix:** Platform blocks are keyed by OS first, then device accelerator. Registration validates that at least one platform block matches the host OS before registering the backend. *(Note: A container recipe's platform block describes its guest OS; GPU containers target `linux`, which Windows hosts satisfy via WSL2).*
+- **Endpoint & Capability Routing:** Descriptors do not dynamically register arbitrary new HTTP endpoints on `lemond`. Instead, plugins bind to Lemonade's standard capability interfaces (`completion`, `embeddings`, `transcription`, `image`, `tts`, `reranking`). `lemond` routes standard requests (`/v1/chat/completions`, `/v1/embeddings`, etc.) directly to the sandboxed subprocess port via `ExternalBackendServer`'s HTTP/REST passthrough.
+- **Iterative Specification Evolution:** The descriptor schema and permission manifest are expected to evolve throughout the Working Group lifecycle. As Phase 1 & 2 RFCs are implemented across diverse inference engines, hardware backends, container runtimes, and OS sandbox APIs, unknown unknowns will be incorporated iteratively into the schema definition.
+
+**Process Isolation vs. In-Process Shared Libraries.** In-process C++ shared libraries (`.so`/`.dll` via `dlopen`) were evaluated and rejected for core backend extensibility:
+- *Alignment with Core Architecture:* Lemonade's existing design already relies on out-of-process subprocess orchestration (`lemond` spawning `llama-server`, `flm`, `sd-server`, `vllm`). Declarative out-of-process descriptors formalize this existing model.
+- *Ecosystem Incentive Structures:* Lemonade is an orchestration integrator. Upstream inference engine developers (vLLM, llama.cpp, SGLang, FastFlowLM) have little incentive to write or maintain Lemonade-specific C++ shared library plugins. Requiring dynamic C++ binaries forces core Lemonade maintainers to author, compile, and maintain all integration logic.
+- *Maintainer Burden & Release Cycles:* In-tree C++ backends force maintainers to package, pin, and update complex binaries for every platform/GPU combination. When upstream engines release new features or model architectures, users are locked into wait cycles until maintainers update bundled binaries. Out-of-process descriptors allow users to immediately run distro-packaged, vendor-provided, or containerized backends.
+- *Testing & Maturation Friction:* Out-of-process descriptors eliminate C++ compilation overhead for testing, enabling community members to quickly prototype, share, and mature new backend recipes.
+- *Security & Process Isolation:* In-process binary modules execute inside `lemond`'s memory space, bypassing `nono`/kernel sandboxing and gaining full server privileges. Out-of-process execution guarantees that `nono` (Landlock/seccomp) isolates backend subprocesses at the OS process boundary.
+- *Server Stability:* A crash, segmentation fault, or memory leak in a third-party backend subprocess cannot take down `lemond` or disrupt other running model slots.
+- *Open Protocol, Closed Surface Boundary:* Lemonade's public API surface is closed—a finite set of capability contracts (`completion`, `embeddings`, `transcription`, `image`, `tts`, `reranking`). Descriptors restrict their scope to launch-time parameter token substitution and fixed HTTP passthrough. If an exotic backend engine requires custom per-request payload transformations, non-standard response framing, or bespoke protocol translation, it lies beyond the boundary of JSON descriptors and graduates to a built-in C++ `WrappedServer`. This prevents JSON descriptors from degenerating into an embedded Turing-complete scripting engine.
+- *Path to Built-in Status:* Adopting declarative JSON descriptors for out-of-tree plugins does not remove the path for a backend to become a built-in C++ server. Descriptors provide a low-friction entry point for incubating community backends; high-adoption engines can graduate to built-in status if core maintainers choose to maintain them in-tree.
+- *Out-of-Process Compiled Binary Plugins (Future Evolution):* Rejecting *in-process* `dlopen` shared libraries does not preclude native C/C++ compiled binary plugins. A future RFC can define a stable C-header ABI / IPC protocol for compiled binary plugins, provided they run as isolated out-of-process subprocesses under the same `nono` kernel sandbox boundary. This accommodates developers desiring a native C++ plugin development workflow while preserving server crash isolation and kernel sandboxing invariants.
+
+### Example Backend Descriptor (`llamacpp`)
+
+Below is an example JSON descriptor illustrating how an out-of-process backend binary (such as `llama-server`) declares its capabilities, slot policy, single command line string template, health probe, stop command, environment variables, and sandbox restrictions:
+
+```json
+{
+  "recipe": "llamacpp-custom",
+  "display_name": "llama.cpp (Custom External)",
+  "binary": "llama-server",
+  "capabilities": ["chat_completion", "completion", "embeddings"],
+  "slot_policy": "standard",
+  "uses_ctx_size": true,
+  "health_probe": {
+    "type": "http",
+    "endpoint": "/health",
+    "expected_status": 200,
+    "timeout_seconds": 60,
+    "poll_interval_ms": 200
+  },
+  "platform": {
+    "linux": {
+      "vulkan": {
+        "command": "{binary} -m {checkpoint:main} -c {ctx_size} -t {threads} --port {port} {custom_args}",
+        "stop_command": "kill -9 {pid}",
+        "env": {
+          "GGML_VK_VISIBLE_DEVICES": "{custom:vk_device:-0}"
+        },
+        "sandbox": {
+          "read_paths": ["{hf_cache}", "{binary_dir}"],
+          "write_paths": ["{cache_dir}/scratch"],
+          "devices": ["/dev/dri"],
+          "allow_network": false
+        }
+      },
+      "rocm": {
+        "command": "{binary} -m {checkpoint:main} -c {ctx_size} -t {threads} --port {port} -ngl 99 {custom_args}",
+        "stop_command": "kill -9 {pid}",
+        "env": {
+          "HIP_VISIBLE_DEVICES": "{hip_visible_devices}"
+        },
+        "sandbox": {
+          "read_paths": ["{hf_cache}", "{binary_dir}"],
+          "write_paths": ["{cache_dir}/scratch"],
+          "devices": ["/dev/dri", "/dev/kfd"],
+          "allow_network": false
+        }
+      }
+    }
+  }
+}
+```
+
+### 2. Permission Model & Sandboxing
+
+Backend subprocesses run inside an OS kernel sandbox where available (Linux `nono` via Landlock/seccomp; Windows AppContainer/Job Objects and macOS Seatbelt are planned cross-platform enforcement milestones).
+
+- **Least Privilege:** Read-only access to model weights and install paths; read-write access restricted to scratch/log directories.
+- **Device & Network Control:** Device nodes (`/dev/dri/*`, `/dev/kfd`, NPU nodes) and outbound network egress must be explicitly declared in the descriptor.
+- **Consent vs Enforcement:** The descriptor's declared policy is shown to the user during installation (consent review). Enforcement is applied at process launch by the kernel sandbox (`nono`). Container recipes operate under container runtime permissions.
+- **Enforcement Modes:** Configurable via `config.json` (`auto`, `enabled`, `disabled`), defaulting to `auto`.
+
+### 3. Backend Recipe Marketplace
+
+A mechanism to discover and install backend descriptors and default recipes.
+
+- **Git-Backed Index:** Marketplaces are Git repositories addressed as `org/repo[/subdir]` (GitHub REST API-first, requiring no local `git` CLI).
+- **Pinned & Idempotent:** Installation pins to a repository git ref and verifies a SHA-256 content hash of the fetched descriptor. Updates require explicit re-installation.
+- **No Binaries:** Marketplaces distribute plain JSON descriptors and recipe configurations, never executable binaries or model weights.
+
+### 4. Client Interfaces (CLI & GUI)
+
+- **CLI Commands:**
+  - `lemonade backends` — List installed and running backends, slot status, and sandbox state.
+  - `lemonade backends install <path|repo/recipe>` — Install a local descriptor file (`<path>`) or fetch and install a remote marketplace recipe (`<repo/recipe>`).
+  - `lemonade backends uninstall <name>` — Unregister and remove an installed backend recipe.
+  - `lemonade backends sandbox status` — View active sandbox profiles and enforcement modes.
+- **GUI:**
+  - Marketplace browser in the Tauri Desktop App and Web App.
+  - Permission consent dialog displaying declared paths, device nodes, and network access before enabling a third-party backend.
+  - Per-backend sandbox status and resource usage indicators.
+
+---
+
+## Roadmap & Milestones
+
+> Sequential high-level phases; RFCs refine specifics.
+
+### Phase 1: Plugin Specification & External Adapter Architecture
+
+- [ ] RFC for the **Backend Plugin Specification** (descriptor schema, OS × device matrix, capability mapping, lifecycle).
+- [ ] `ExternalBackendServer` class in `lemond` wrapping out-of-process backends via HTTP/sockets while honoring `WrappedServer` interfaces.
+- [ ] `ModelManager` / `Router` discovery and validation of local descriptor files in `backends/` directories.
+- [ ] `lemonade backends` CLI commands (`backends`, `backends install <path>`, `backends uninstall <name>`).
+- [ ] Registration-time host-OS compatibility validation.
+
+### Phase 2: Process Sandboxing Core (`nono`) & Permission Policy Engine
+
+- [ ] RFC for the **Sandbox Architecture & Permission Policy Schema**.
+- [ ] `NonoSandbox` engine for binary backend subprocesses (Linux Landlock filesystem restrictions, seccomp-bpf system call filtering, device node rules).
+- [ ] Cross-platform binary process sandboxing (Windows AppContainer / Job Objects; macOS Seatbelt / App Sandbox).
+- [ ] Permission manifest parser and configurable enforcement modes (`auto`, `enabled`, `disabled`) in `lemond`.
+- [ ] Sandbox telemetry in `/system-info` and `lemonade backends sandbox status`.
+- [ ] **Extension Evaluation:** Evaluate container backend security posture (container runtime isolation vs. rootless Podman namespaces vs. `nono` container profile wrappers). Determines hand-off to proposed Containerized Backends WG.
+
+### Phase 3: Marketplace Infrastructure & Recipe Sharing
+
+- [ ] RFC and schema for the **Recipe Marketplace Registry** (GitHub `org/repo[/subdir]` index, ref + SHA-256 content-hash pinning, descriptor validation).
+- [ ] API-first fetching mechanism (GitHub REST API) without local `git` binary dependencies.
+- [ ] CLI marketplace command (`lemonade backends install <repo/recipe>`).
+- [ ] Content-hash verification and permission consent flow during marketplace installation.
+- [ ] Community recipe publishing guide.
+
+### Phase 4: GUI, Hardening & Ecosystem
+
+- [ ] Backend manager and marketplace UI in Tauri and Web apps.
+- [ ] Permission consent modal in GUI displaying declared paths, devices, and network permissions.
+- [ ] Sandbox status indicators per running backend.
+- [ ] Port a built-in backend to descriptor format to validate the plugin architecture.
+- [ ] CI regression suite verifying kernel sandbox enforcement.
+- [ ] Developer guide: *Building, Publishing, and Using a Lemonade Backend Recipe*.
+
+---
+
+## Governance & Communication
+
+- **Discord Channel:** `#wg-backend-plugins`
+- **GitHub Label:** `wg-backend-plugins`
+- **Meeting Cadence:** Bi-weekly working group sync (published on Discord).
+- **PR Review:** Proposals for specifications, sandbox policies, or marketplace schemas require approval from the Working Group Lead and a core maintainer.
--- a/docs/dev/working-groups/backend-plugins-sandboxing.md
+++ b/docs/dev/working-groups/backend-plugins-sandboxing.md
@@ -18,6 +18,7 @@ Running third-party or community backend binaries introduces security risks (una
 
 - **In Scope:** Descriptor schema definition, out-of-process HTTP/socket backend adapter, process sandboxing (`nono` on Linux, platform primitives on Windows/macOS), recipe marketplace, client UI integration.
 - **Out of Scope:** Engine binary compilation/build infrastructure. Lemonade orchestrates and sandboxes pre-built backend binaries; it does not build them.
+- **Non-Goals:** 1:1 behavioral idempotency with internal C++ `WrappedServer` hardware-conditional or model-introspection logic (e.g. C++ checks like `SystemInfo::get_has_igpu()`, dynamic `--embeddings`/`--reranking` flag injection). JSON descriptors define a fixed, manageable parameter-substitution model for out-of-tree and containerized backends; complex C++ execution logic remains in built-in backends.
 - **Related Work:** Deep container engine integration and containerized backend orchestration are recommended as a follow-on **Containerized Backends & Isolation Working Group**.
 
 ---


bitgamma (commented):
I would like to understand better how the actual plugins are structured, because the document speaks mostly about sandboxing the actual backend executable (let's say llamacpp) and not much about the "adapter" code that currently lives in lemonade's codebase (the part that would be a plugin, supposedly). 

Are we looking at loadable C++ modules that implement a specific API or really pure JSON? The C++ module providing the integration would run as part of the lemonade process meaning that a malicious backend-adapter would still have at least the same privileges as lemond

A pure JSON solution sounds interesting, but are we sure we can model integration with any binary by pure JSON? there is a lot of argument remapping and other input preprocessing/output postprocessing done into the backend adapters.

It is also not clear to me how new HTTP endpoints etc are added to the server in this plugin architecture.

Geramy (changes requested):
I would only approve this if the backend that gets loaded is a dynamically loaded binary that gets loaded into lemonade as a plugin, I wouldn't want json honestly its just not extendable enough, a bit messy, and not the preferred method.

jeremyfowers:
> I would only approve this if the backend that gets loaded is a dynamically loaded binary that gets loaded into lemonade as a plugin, I wouldn't want json honestly its just not extendable enough, a bit messy, and not the preferred method.

@Geramy can you elaborate? You want the plugins to be a small C++ definition or what? If there is a complex build system that would add a lot of friction, the nice thing about json is that its very simple to submit to a marketplace.

Geramy:
> > I would only approve this if the backend that gets loaded is a dynamically loaded binary that gets loaded into lemonade as a plugin, I wouldn't want json honestly its just not extendable enough, a bit messy, and not the preferred method.
> 
> @Geramy can you elaborate? You want the plugins to be a small C++ definition or what? If there is a complex build system that would add a lot of friction, the nice thing about json is that its very simple to submit to a marketplace.

Yeah I wanted a simple header and ability to dynamically load the C++ binary this it would be nice to have. That’s the preferred development route than a json file.

Geramy:
I guess I could settle for json for now but it would be preferred to have binary too

bitgamma:
while I would actually love the JSON option to be viable, I really doubt it can cover all possible interactions between lemonade and an arbitrary binary without degenerating into its own Turing-complete scripting language. I am happy to be convinced otherwise though. For example, what would the JSON version of our llamacpp backend look like?

abn:
Thanks @bitgamma, @Geramy, and @jeremyfowers for the feedback and discussion!

This Working Group charter was directly triggered by the prototype implementation in **[PR #2880: feat: Dynamic Custom Backend System](https://github.com/lemonade-sdk/lemonade/pull/2880)**. I've updated the charter document to detail the design rationale behind out-of-process JSON descriptors, address dynamic C++ binary plugins, and detail how capability mapping works without needing complex scripting.

#### 1. Alignment with Lemonade's Core Architecture & Integrator Role
- **Formalizing Existing Design**: Lemonade's core architecture was deliberately designed around out-of-process subprocess orchestration (`lemond` spawning `llama-server`, `flm`, `sd-server`, `vllm`). Declarative descriptors formalize this existing model rather than introducing a whole new runtime architecture.
- **Ecosystem Realities**: Lemonade is an orchestration integrator. Upstream inference engine developers (vLLM, llama.cpp, SGLang, FastFlowLM) have little or no incentive to author or maintain Lemonade-specific C++ shared library plugins (`.so`/`.dll`). Requiring C++ shared library plugins forces Lemonade core maintainers to author, compile, and maintain all integration bindings.
- **Path to Built-in Status**: Adopting declarative JSON descriptors for out-of-tree plugins does **not** remove the path for a backend to become a built-in C++ backend. Descriptors serve as an incubator for experimental/community backends; high-adoption engines can still graduate to built-in C++ status if maintainers choose to support them in-tree.
- **Bridging Declarative Descriptors and Compiled Binary Plugins**: Rejecting *in-process* `dlopen` shared libraries does not preclude native C/C++ compiled binary plugins. A future RFC can define a stable C-header ABI / IPC protocol for compiled binary plugins, provided they run as isolated out-of-process subprocesses under the same `nono` kernel sandbox boundary. This accommodates developers desiring a native C++ plugin development workflow while preserving server crash isolation and process sandboxing invariants.
- **Iterative Schema Evolution**: Approving this charter establishes the overall architectural direction and roadmap phases. The exact descriptor schema and permission manifest are expected to evolve iteratively through Phase 1 & 2 RFCs as the working group encounters and incorporates unknown unknowns across diverse engines, container runtimes, and platform sandboxes.

#### 2. Reducing Maintainer Burden & Eliminating User Wait Cycles
- **The In-Tree Binary Maintenance Burden**: Maintaining in-tree C++ backends forces maintainers to build, package, pin, and test complex binaries for every platform/GPU combination.
  - *Example (FastFlowLM / `flm`)*: Integrating `flm` as a native in-tree C++ server required substantial maintainer bandwidth and upstream collaboration.
  - *Example (`vLLM-ROCm` version lag)*: Pinned bundled binaries (e.g. `v0.20.1-rocm7.12.0-gfx1151`) inevitably lag behind upstream releases (`v0.23.0+`). When upstream releases support for new model architectures or speculative decoding features (such as Gemma-4 E4B MTP), users are forced to wait for Lemonade maintainers to build, patch, and release new binary bundles.
- **Instant Ecosystem Access**: Out-of-process descriptors allow users to immediately run containerized backends (`podman run ...`), distro-packaged executables, or vendor-provided binaries without waiting for a Lemonade release cycle or placing packaging overhead on maintainers.

#### 3. Open Protocol, Closed Surface Boundary & Descriptor Scope
- **Open Protocol Passthrough, Closed Surface**: Lemonade's public API surface is closed providing only a finite set of standard capability contracts (`completion`, `embeddings`, `transcription`, `image`, `tts`, `reranking`). `ExternalBackendServer` acts as a fixed HTTP passthrough envelope to backend subprocesses that implement these standard REST schemas.
- **Why No Embedded Scripting Language is Needed**: Rather than inventing a Turing-complete JSON scripting or payload transformation engine (e.g. jq/Lua rules in JSON) to perform arbitrary per-request payload rewriting or response parsing, descriptor responsibilities are strictly bounded to launch-time parameter token substitution (`{checkpoint:main}`, `{ctx_size}`, `{threads}`, `{port}`) and fixed HTTP passthrough. This is a design choice.
- **Cross-Platform Matrix**: Descriptor platform blocks use a first-class `Host OS × Accelerator` matrix (supporting `windows -> cuda|vulkan`, `macos -> metal`, `linux -> vulkan|rocm|npu`), allowing a single descriptor file to declare native targets across OS platforms.
- **The Structural Graduation Hatch**: If a novel backend requires non-standard per-request payload transformations, custom SSE framing, or bespoke protocol translation that standard HTTP passthrough cannot express, it lies outside the boundary of JSON descriptors and graduates to a built-in C++ `WrappedServer`. This keeps `lemond` clean, safe, and free of embedded scripting engines.

#### 4. Security, Sandboxing, & Process Isolation
- **Bypassing Sandboxing**: In-process `.so`/`.dll` shared libraries execute directly inside `lemond`'s memory space, bypassing `nono` OS kernel sandboxing (`Landlock`/`seccomp`) and gaining full server privileges. Out-of-process subprocess execution ensures `nono` can strictly restrict filesystem, device, and network access at the kernel process boundary.
- **Crash Isolation**: Segmentation faults, memory leaks, or crashes in a third-party backend binary cannot take down `lemond` or interrupt other running model slots.

### Example Descriptors

<details>
<summary><b>backends/llamacpp_vulkan_custom.json</b></summary>

```json
{
  "recipe": "llamacpp-vulkan-custom",
  "display_name": "llama.cpp Server (Vulkan Custom)",
  "binary": "llama-server",
  "slot_policy": "standard",
  "capabilities": ["chat_completion", "completion"],
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "vulkan": {
        "command": "/home/user/.cache/lemonade/bin/llamacpp/vulkan/llama-server -m {checkpoint:main} --host {host} --port {port} -c {ctx_size} -t {threads}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "GGML_VK_VISIBLE_DEVICES": "{custom:vk_device:-0}"
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/llamacpp_rocm_custom.json</b></summary>

```json
{
  "recipe": "llamacpp-rocm-custom",
  "display_name": "llama.cpp Server (ROCm Custom)",
  "binary": "llama-server",
  "slot_policy": "standard",
  "capabilities": ["chat_completion", "completion"],
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "rocm": {
        "command": "/home/user/.cache/lemonade/bin/llamacpp/rocm/llama-server -m {checkpoint:main} --host {host} --port {port} -c {ctx_size} -t {threads}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "HIP_VISIBLE_DEVICES": "{hip_visible_devices}"
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/llamacpp_custom.json (Combined Vulkan & ROCm)</b></summary>

```json
{
  "recipe": "llamacpp-custom",
  "display_name": "llama.cpp (Custom External)",
  "binary": "llama-server",
  "capabilities": ["chat_completion", "completion", "embeddings"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "vulkan": {
        "command": "{binary} -m {checkpoint:main} -c {ctx_size} -t {threads} --port {port} {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "GGML_VK_VISIBLE_DEVICES": "{custom:vk_device:-0}"
        },
        "sandbox": {
          "read_paths": ["{hf_cache}", "{binary_dir}"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/dri"],
          "allow_network": false
        }
      },
      "rocm": {
        "command": "{binary} -m {checkpoint:main} -c {ctx_size} -t {threads} --port {port} -ngl 99 {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "HIP_VISIBLE_DEVICES": "{hip_visible_devices}"
        },
        "sandbox": {
          "read_paths": ["{hf_cache}", "{binary_dir}"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/dri", "/dev/kfd"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/flm_custom.json</b></summary>

```json
{
  "recipe": "flm-custom",
  "display_name": "FastFlowLM Server (NPU Custom)",
  "binary": "flm",
  "slot_policy": "coexist_by_type",
  "capabilities": ["chat_completion", "completion"],
  "health_probe": {
    "type": "http",
    "endpoint": "/v1/models",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "npu": {
        "command": "/home/user/.cache/lemonade/bin/flm/npu/flm serve {custom:model_tag:-llama3.2:1b} --host {host} --port {port}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "FLM_POWER_MODE": "{custom:power_mode:-performance}"
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/dflash_rocm.json (Podman Containerized Backend)</b></summary>

```json
{
  "recipe": "dflash-rocm",
  "display_name": "DFlash Speculative Decoding Server (ROCm Podman)",
  "slot_policy": "standard",
  "capabilities": ["chat_completion", "completion"],
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 300,
    "poll_interval_ms": 500
  },
  "platform": {
    "linux": {
      "rocm": {
        "command": "podman run --rm --init -i --name lemonade-{recipe}-{port} --device /dev/kfd --device /dev/dri --group-add video --group-add render --security-opt seccomp=unconfined --security-opt label=disable -v {hf_cache}:/models:ro --network host --entrypoint /opt/lucebox-hub/server/build/dflash_server ghcr.io/luce-org/lucebox-hub:rocm-7.2 /models/{checkpoint_relative:main} --target-device {target_device} --model-name {model_name} --host {host} --port {port} --max-ctx {ctx_size} --cache-type-k {cache_type_k} --cache-type-v {cache_type_v} --draft /models/{checkpoint_relative:draft} --ddtree --ddtree-budget {custom:ddtree_budget:-22} --prefill-skip-park {custom_args}",
        "stop_command": "podman rm -f -t 0 lemonade-{recipe}-{port}",
        "env": {
          "DFLASH27B_DRAFT_SWA": "{custom:draft_swa:-2048}",
          "DFLASH27B_PREFILL_UBATCH": "{custom:prefill_ubatch:-512}"
        }
      }
    }
  }
}
```

</details>

bitgamma (changes requested):
@abn thanks for detailed examples. They show what I was talking about - these are not idempotent to the current implementation and every "if" in the code translates into additional complexity. Just a small example (which is not even the most important one, but just fresh in my mind as I touched said code recently)

```
    // Disable llamacpp webui by default
    push_overridable_arg(args, llamacpp_args, "--no-ui");

    // Disable mmap on iGPU
    if (SystemInfo::get_has_igpu()) {
        push_overridable_arg(args, llamacpp_args, "--load-mode", "none", {"--direct-io", "--no-direct-io", "-dio", "-ndio", "--mmap", "--no-mmap", "--mlock", "-lm"});
    }

    // Add embeddings support if the model supports it
    if (supports_embeddings) {
        LOG(INFO, "LlamaCpp") << "Model supports embeddings, adding --embeddings flag" << std::endl;
        push_arg(args, reserved_flags, "--embeddings");
    }
    push_reserved(reserved_flags, "--embeddings", std::vector<std::string>{"--embedding"});

    // Add reranking support if the model supports it
    if (supports_reranking) {
        LOG(INFO, "LlamaCpp") << "Model supports reranking, adding --reranking flag" << std::endl;
        push_arg(args, reserved_flags, "--reranking");
    }
    push_reserved(reserved_flags, "--reranking", std::vector<std::string>{"--rerank"});
```

this is not modeled by the JSON. It could, but then either the code handling the JSON will need to handle all possible edge-cases, meaning that realistically a good amount of backends will require adaptation to the code OR the JSON becomes Turing-complete.

we *could* define a fixed, manageable feature set and see how many backends would fit exactly within this framework without limitations, but then we need to be clear that cpp backends and JSON backends are not idempotent and there are very real limitations about what they can do

abn:
@bitgamma 100% agree with your observation here!

Trying to model C++ hardware detection (`SystemInfo::get_has_igpu()`), model metadata introspection (`supports_embeddings`, `supports_reranking`), or complex flag overriding inside JSON would force descriptors into becoming an anti-pattern scripting language.

To be explicit: **1:1 behavioral idempotency with internal C++ `WrappedServer` classes is an explicit non-goal for JSON descriptors.** 

The original design intent of JSON descriptors is not to replace high-touch C++ backends, but rather to provide a fixed, manageable parameter-substitution model that acts as an instance-specific escape hatch for out-of-tree backends, community engines, and containerized recipes (`podman run ...`).

This defines a clean two-tier architecture for Lemonade:

1. **Built-in C++ Backends (`WrappedServer`)**: High-touch, deeply integrated engines (`llama.cpp`, `flm`, `sd-server`) where Lemonade maintainers write custom C++ code for hardware detection (`iGPU`, `CWSR` fixes), model metadata introspection, and reserved flag overrides.
2. **External JSON Plugins (`ExternalBackendServer`)**: A fixed, manageable parameter-substitution model for third-party, community, and containerized backends that accept standard CLI flags and environment variables without requiring server C++ code modifications or rebuilds.

I've updated the **Scope Boundaries** section of the charter document to explicitly state this non-goal.

abn:
Also, so that this is not lost along the way - +1 to keeping the door open for dynamic binary loading down the road.

We initially focus on establishing declarative descriptors and the out-of-process kernel sandbox boundary. 

Once these are stable, a follow-on Working Group can explore compiled shared library (`.so`/`.dll`) plugins or IPC C-ABI interfaces for developers who want native C++ binary plugin workflows.

bitgamma (changes requested):
@abn this sounds reasonable. However the entire system comes with a good amount of complexity that needs to be maintained.

Sorry to be a pain, but I propose a little experiment. Of the existing backends, let's count how many can be converted to JSON cleanly in a way that is idempotent to the C++ code. This will give us a better perspective of whether the additional complexity is worth it.

abn:
@bitgamma got the rocks to do some work :smile:. And for full transparency I only did a very brief review of this. Obviously, almost all forks and branches of llama.cpp or vllm should in theory work unless it specifically targets something we cannot cover, at that point it is either an experiment that folks hard code values in their local environment or a candidate for for a wrapper that needs to be evaluated.

If there are specific backends that come to your mind, happy to take a look and try generate working configs for it already once I have some time.

### Inference Engine Plugin Feasibility Evaluation & JSON Configurations

This evaluates the feasibility of adopting out-of-process JSON backend descriptors (`ExternalBackendServer`) for third-party inference engines under the Lemonade Working Group specification. It surveys publicly available packaged distributions (Homebrew, Fedora Copr RPM, Flatpak, Nix, PyPI, GitHub releases) and container options (Podman/Docker on Docker Hub and GHCR) across established, emerging, and bleeding-edge runtimes (Rust `candle-vllm`, Microsoft `ort-genai`, Intel `ovms`, WebAssembly `wasmedge-ggml`, Meta `executorch`).

---

#### 1. Distribution & Packaging Landscape

| Inference Engine Category | Engine / Executable | Distro / System Packages | Language Package Managers | Container Registries (Docker / Podman) | Feasibility Status |
| :--- | :--- | :--- | :--- | :--- | :---: |
| **GGUF C++** | `llama-server` (`llama.cpp`) | Homebrew (`brew install llama.cpp`), Fedora RPM, Nix, Arch (AUR) | PyPI (`llama-cpp-python`) | `ghcr.io/ggerganov/llama.cpp` | **High** (Native & Container) |
| **High-Throughput** | `vllm` (`vLLM`) | N/A (System deps via apt/dnf) | PyPI (`pip install vllm`) | `vllm/vllm-openai`, `ghcr.io/vllm-project/vllm-rocm` | **High** (Virtualenv & Container) |
| **High-Throughput** | `sglang` (`SGLang`) | N/A (System deps via apt/dnf) | PyPI (`pip install sglang`) | `lmsysorg/sglang:latest` | **High** (Virtualenv & Container) |
| **Zig Engine** | `llmd` (`ZML`) | GitHub Release Binaries | N/A (Native Zig Zero-Dependency) | `ghcr.io/zml/llmd:latest` | **High** (Native & Container) |
| **Block-Diffusion** | `dflash_server` (`Lucebox`) | GitHub Release Tars | N/A (C++ / CUDA / ROCm) | `ghcr.io/luce-org/lucebox-hub` | **High** (Container & Native) |
| **NPU C++** | `flm` (`FastFlowLM`) | Fedora Copr RPM (`FastFlowLM-rpm`) | N/A | N/A (NPU XRT Driver Host Bound) | **High** (Custom Packaged Distro/Flatpak of Built-in) |
| **Rust Engine** | `candle-server` (`candle-vllm`) | Cargo (`cargo install candle-vllm`), GitHub Binaries | PyPI (`pip install candle-vllm`) | `ghcr.io/huggingface/candle` | **High** (Native & Container) |
| **Cross-Hardware** | `ort-genai` (`ONNX Runtime`) | GitHub Release Tars | PyPI (`pip install onnxruntime-genai`) | `mcr.microsoft.com/onnxruntime` | **High** (Virtualenv & Native) |
| **Intel Enterprise** | `ovms` (`OpenVINO Server`) | Intel APT/YUM repos, GitHub Releases | PyPI (`pip install openvino`) | `docker.io/openvino/model_server` | **High** (Container & Native) |
| **Wasm Sandbox** | `wasmedge-ggml` | Homebrew (`brew install wasmedge`), WasmEdge Installer | N/A (Wasm Binary) | `secondstate/wasmedge-ggml` | **High** (Wasm Runtime & Native) |
| **On-Device Edge** | `executorch` (PyTorch Edge) | GitHub Release Binaries | PyPI (`pip install executorch`) | `ghcr.io/pytorch/executorch` | **High** (Native & Embedded) |
| **ARM NPU** | `rknn_llm_server` (Rockchip) | Debian/Ubuntu `.deb` for RK3588 | PyPI (`rknn-llm`) | N/A (Rockchip NPU Driver) | **High** (Native ARM Binary) |

---

#### 2. WG-Compliant JSON Descriptor Configurations

Below are concrete, schema-validated JSON descriptors demonstrating feasibility across distro-packaged binaries, PyPI virtualenvs, native Rust/C++ executables, WebAssembly runtimes, and Podman containers.

<details>
<summary><b>backends/llamacpp_vulkan.json (Distro / Homebrew Native Binary)</b></summary>

```json
{
  "recipe": "llamacpp-vulkan",
  "display_name": "llama.cpp Server (Vulkan Distro)",
  "binary": "llama-server",
  "capabilities": ["chat_completion", "completion", "embeddings"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "vulkan": {
        "command": "llama-server -m {checkpoint:main} -c {ctx_size} -t {threads} --port {port} {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "GGML_VK_VISIBLE_DEVICES": "{custom:vk_device:-0}"
        },
        "sandbox": {
          "read_paths": ["{hf_cache}"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/dri"],
          "allow_network": false
        }
      }
    },
    "macos": {
      "metal": {
        "command": "llama-server -m {checkpoint:main} -c {ctx_size} -t {threads} --port {port} -ngl 99 {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {},
        "sandbox": {
          "read_paths": ["{hf_cache}"],
          "write_paths": ["{cache_dir}/scratch"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/sglang_cuda.json (PyPI Virtualenv Engine)</b></summary>

```json
{
  "recipe": "sglang-cuda",
  "display_name": "SGLang Server (CUDA PyPI)",
  "binary": "sglang",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 120,
    "poll_interval_ms": 500
  },
  "platform": {
    "linux": {
      "cuda": {
        "command": "{home_dir}/.local/pipx/envs/sglang/bin/python3 -m sglang.launch_server --model-path {checkpoint_dir:main} --host {host} --port {port} --context-length {ctx_size} {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "CUDA_VISIBLE_DEVICES": "{cuda_visible_devices}"
        },
        "sandbox": {
          "read_paths": ["{hf_cache}", "{home_dir}/.local/pipx/envs/sglang"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/nvidia*"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/vllm_rocm_container.json (Podman Containerized vLLM Engine)</b></summary>

```json
{
  "recipe": "vllm-rocm-container",
  "display_name": "vLLM Server (ROCm Podman Container)",
  "binary": "podman",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 180,
    "poll_interval_ms": 500
  },
  "platform": {
    "linux": {
      "rocm": {
        "command": "podman run --rm --init -i --name lemonade-{recipe}-{port} --device /dev/kfd --device /dev/dri --group-add video --group-add render --security-opt seccomp=unconfined -v {hf_cache}:/root/.cache/huggingface:ro --network host ghcr.io/vllm-project/vllm-rocm:latest --model /root/.cache/huggingface/hub/{checkpoint_relative:main} --host {host} --port {port} --max-model-len {ctx_size} {custom_args}",
        "stop_command": "podman rm -f -t 0 lemonade-{recipe}-{port}",
        "env": {
          "HIP_VISIBLE_DEVICES": "{hip_visible_devices}"
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/candle_vllm_cuda.json (Pure Rust Single-Binary Engine)</b></summary>

```json
{
  "recipe": "candle-vllm-cuda",
  "display_name": "Candle vLLM Server (Rust Native)",
  "binary": "candle-vllm",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "cuda": {
        "command": "{home_dir}/.cargo/bin/candle-vllm --model-id {checkpoint_dir:main} --port {port} --max-model-len {ctx_size} {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "CUDA_VISIBLE_DEVICES": "{cuda_visible_devices}"
        },
        "sandbox": {
          "read_paths": ["{hf_cache}", "{home_dir}/.cargo/bin"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/nvidia*"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/ort_genai_directml.json (Cross-Hardware GenAI Engine)</b></summary>

```json
{
  "recipe": "ort-genai-directml",
  "display_name": "ONNX Runtime GenAI (DirectML / CPU)",
  "binary": "ort-genai",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/v1/models",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "windows": {
      "directml": {
        "command": "ort-genai.exe --model {checkpoint_dir:main} --port {port} --execution_provider dml {custom_args}",
        "stop_command": "taskkill /F /PID {pid}",
        "env": {}
      }
    },
    "linux": {
      "cpu": {
        "command": "ort-genai --model {checkpoint_dir:main} --port {port} --execution_provider cpu {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {},
        "sandbox": {
          "read_paths": ["{hf_cache}"],
          "write_paths": ["{cache_dir}/scratch"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/ovms_intel.json (Intel OpenVINO Model Server Podman Container)</b></summary>

```json
{
  "recipe": "ovms-intel",
  "display_name": "OpenVINO Model Server (Intel iGPU / NPU Podman)",
  "binary": "podman",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/v1/config",
    "expected_status": 200,
    "timeout_seconds": 90,
    "poll_interval_ms": 300
  },
  "platform": {
    "linux": {
      "igpu": {
        "command": "podman run --rm --init -i --name lemonade-{recipe}-{port} --device /dev/dri -v {hf_cache}:/models:ro --network host docker.io/openvino/model_server:latest --rest_port {port} --model_name {custom:model_name:-default} --model_path /models/{checkpoint_relative:main} --target_device GPU {custom_args}",
        "stop_command": "podman rm -f -t 0 lemonade-{recipe}-{port}",
        "env": {}
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/wasmedge_ggml.json (WebAssembly Sandboxed Engine)</b></summary>

```json
{
  "recipe": "wasmedge-ggml",
  "display_name": "WasmEdge Sandboxed LLM Server (WebAssembly)",
  "binary": "wasmedge",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "vulkan": {
        "command": "wasmedge --dir .:. --nn-preload default:GGML:AUTO:{checkpoint:main} {binary_dir}/llama-api-server.wasm --port {port} -c {ctx_size} {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {},
        "sandbox": {
          "read_paths": ["{hf_cache}", "{binary_dir}"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/dri"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/zml_llmd.json (Native Zig Zero-Dependency SafeTensors Engine)</b></summary>

```json
{
  "recipe": "zml-llmd",
  "display_name": "ZML LLM Daemon (SafeTensors Native)",
  "binary": "llmd",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/v1/models",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "vulkan": {
        "command": "{binary_dir}/llmd serve --model {checkpoint_dir:main} --host {host} --port {port} {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {
          "ZML_LOG_LEVEL": "info"
        },
        "sandbox": {
          "read_paths": ["{hf_cache}", "{binary_dir}"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/dri"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/flm_npu_custom.json (Fedora Copr RPM / Flatpak Custom Packaged Engine)</b></summary>

```json
{
  "recipe": "flm-npu-custom",
  "display_name": "FastFlowLM Server (Fedora Copr RPM / Flatpak NPU)",
  "binary": "flm",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "coexist_by_type",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/api/tags",
    "expected_status": 200,
    "timeout_seconds": 60,
    "poll_interval_ms": 200
  },
  "platform": {
    "linux": {
      "npu": {
        "command": "/usr/bin/flm serve {custom:model_tag:-llama3.2:1b} --host {host} --port {port} --pmode {custom:power_mode:-performance} {custom_args}",
        "stop_command": "kill -9 {pid}",
        "env": {},
        "sandbox": {
          "read_paths": ["{hf_cache}", "/usr/share/flm"],
          "write_paths": ["{cache_dir}/scratch"],
          "devices": ["/dev/accel/accel0"],
          "allow_network": false
        }
      }
    }
  }
}
```

</details>

<details>
<summary><b>backends/dflash_rocm.json (Podman Containerized Block-Diffusion Engine)</b></summary>

```json
{
  "recipe": "dflash-rocm",
  "display_name": "DFlash Speculative Decoding Server (Lucebox ROCm Podman)",
  "binary": "podman",
  "capabilities": ["chat_completion", "completion"],
  "slot_policy": "standard",
  "uses_ctx_size": true,
  "health_probe": {
    "type": "http",
    "endpoint": "/health",
    "expected_status": 200,
    "timeout_seconds": 300,
    "poll_interval_ms": 500
  },
  "platform": {
    "linux": {
      "rocm": {
        "command": "podman run --rm --init -i --name lemonade-{recipe}-{port} --device /dev/kfd --device /dev/dri --group-add video --group-add render --security-opt seccomp=unconfined --security-opt label=disable -v {hf_cache}:/models:ro --network host --entrypoint /opt/lucebox-hub/server/build/dflash_server ghcr.io/luce-org/lucebox-hub:rocm-7.2 /models/{checkpoint_relative:main} --target-device {custom:target_device:-GPU} --model-name {custom:model_name:-dflash} --host {host} --port {port} --max-ctx {ctx_size} --draft /models/{checkpoint_relative:draft} --ddtree --ddtree-budget {custom:ddtree_budget:-22} {custom_args}",
        "stop_command": "podman rm -f -t 0 lemonade-{recipe}-{port}",
        "env": {
          "DFLASH27B_DRAFT_SWA": "{custom:draft_swa:-2048}",
          "DFLASH27B_PREFILL_UBATCH": "{custom:prefill_ubatch:-512}"
        }
      }
    }
  }
}
```

</details>

---

#### 3. Feasibility Evaluation Findings & Key Insights

1. **Built-in vs. Custom Packaged Coexistence**:
   The `flm-npu-custom` example demonstrates that out-of-process JSON descriptors allow even built-in backends (like `FastFlowLM`) to be packaged externally via Fedora Copr (`FastFlowLM-rpm`) or Flatpak (`com.fastflowlm.FastFlowLM`). Users can override built-in behavior with custom packaging without breaking `lemond` invariants.

2. **Universal Execution Abstraction**: 
   The single command line string template format (`"command": "..."`) handles native distro binaries (`llama-server`), PyPI virtualenv launchers (`sglang`), containerized engines (`vllm`), Rust executables (`candle-vllm`), WebAssembly runtimes (`wasmedge`), Microsoft DirectML binaries (`ort-genai`), native Zig executables (`llmd`), and container engines (`podman run ...`) without needing separate schema structures for each execution model.

3. **Uniform Health Probing**:
   All surveyed established and bleeding-edge engines expose an HTTP REST health or model listing endpoint (`/health`, `/v1/models`, `/v1/config`, `/api/tags`, `/v1/chat/completions`) returning HTTP 200. This validates that the `"health_probe"` JSON block operates uniformly across Python, C++, Rust, Zig, Wasm, and containerized backends.

4. **Sandboxing Policy Harmony**:
   - **Native & Rust/Wasm Binaries (`nono`)**: Restricted at launch time by `nono` (Landlock/seccomp) using the declared `"sandbox"` policy block (`read_paths`, `write_paths`, `devices`).
   - **Container Recipes (Podman/Docker)**: Operate under container runtime isolation (`--device`, `--security-opt`, `-v`), mapping seamlessly into `ExternalBackendServer`'s subprocess lifecycle.

bitgamma:
@abn this doesn't really address my point, because the JSON presented have no point of comparison, except llamacpp and vllm, both of which are NOT equivalent to the C++ version (we discussed llamacpp before, but checking the code vllm I see a similar situation). 

What I asked was about the *existing* backends (stable-diffusion, kokoro, moonshine, thenoise, trellis, etc) already integrated in lemonade. Which of these could be ported to a pure JSON architecture without losing any feature? We already ascertained that for llama.cpp and vllm this is not the case, the JSON version misses stuff that the C++ version has.

So my question is, let's say we introduce the JSON format, how many of the existing C++ modules we could theoretically migrate to JSON with **0 loss of functionality and edge case handling?** I think this is a good way to measure how powerful this system can be. If it turns out we could replace, say, half of our existing C++ backends, then I would have no doubt that the JSON approach is a huge win. If it turns out that none or just a couple can be mapped 1-to-1 as JSON, then I would doubt that the system could support new backends in a way that feels good to the user.

abn:
> So my question is, let's say we introduce the JSON format, how many of the existing C++ modules we could theoretically migrate to JSON with **0 loss of functionality and edge case handling?** I think this is a good way to measure how powerful this system can be. If it turns out we could replace, say, half of our existing C++ backends, than I would have no doubt that the JSON approach is a huge win. If it turns out that none or just a couple can be mapped 1-to-1 as JSON, then I would doubt that the system could support new backends in a way that feels good to the user.

Ah. I completely misunderstood your question there. That's on me for multitasking. Let me get back to you that, it should be fairly straightforward for us to enumerate that.

abn:
Going through our existing backends and skimming over what we handling in code today, I reckon a compact set of six declarative primitives (command tokens, environment blocks, request transformations, health probes, multi-port allocation tokens, and external wrapper scripts) covers almost all backend execution requirements we see today.

The "external wrapper scripts" is more an escape hatch option I'd say if the server itself needs special handling and it helps validate a backend before any maintainer effort is put into it.

Obviously, I do not know what I do not know. And I might be missing things here, so happy to be corrected.

<details>
<summary><b>Backend 1: SdServer (Stable Diffusion)</b></summary>

##### Current C++ Wrapper Implementation (`sdcpp_server.cpp`)
1. **Multi-Checkpoint Flag Branching**: Inspects model metadata. For Flux models, it appends `--diffusion-model <path> --llm <path> --vae <path>`. For standard models, it passes `-m <checkpoint>`.
2. **Prompt XML Parameter Wrapping**: Appends parameters such as `steps`, `cfg_scale`, and `seed` to the prompt string inside `<sd_cpp_extra_args>{"steps": 20, ...}</sd_cpp_extra_args>`.
3. **Environment Setup**: Prepends ROCm and CUDA library directories to `LD_LIBRARY_PATH` and `PATH`.

##### Declarative JSON Specification Alternative
1. **Multi-Checkpoint Command Templating**:
   ```json
   "command": "sd-server {checkpoint:diffusion_model:--diffusion-model {checkpoint:diffusion_model}} {checkpoint:llm:--llm {checkpoint:llm}} {checkpoint:vae:--vae {checkpoint:vae}} {checkpoint:main:-m {checkpoint:main}} --port {port}"
   ```
2. **Payload Prompt Templating**:
   ```json
   "request_transform": {
     "prompt": "{prompt} <sd_cpp_extra_args>{\"steps\": {steps:-20}, \"cfg_scale\": {cfg_scale:-7.0}}</sd_cpp_extra_args>"
   }
   ```
3. **Environment Block**:
   ```json
   "env": {
     "LD_LIBRARY_PATH": "{rocm_lib_dir}:{env:LD_LIBRARY_PATH}"
   }
   ```

</details>

<details>
<summary><b>Backend 2: KokoroServer (Kokoro TTS)</b></summary>

##### Current C++ Wrapper Implementation (`kokoro_server.cpp`)
1. **Model & Voice File Resolution**: Reads `index.json` in the model directory to resolve the ONNX model and voice binary filenames.
2. **Environment Variable Configuration**: Sets `ESPEAK_DATA_PATH` to point to `espeak-ng-data`.
3. **Payload Modification**: Forces `model` to `"kokoro"` and sets `stream: true` when streaming audio is requested.

##### Declarative JSON Specification Alternative
1. **Path Templating with Defaults**:
   ```json
   "command": "koko -m {checkpoint_dir:main}/{custom:model_file:-kokoro-v0_19.onnx} -d {checkpoint_dir:main}/{custom:voices_file:-voices.bin} openai --port {port}"
   ```
2. **Environment Variable Block**:
   ```json
   "env": {
     "ESPEAK_DATA_PATH": "{binary_dir}/espeak-ng-data"
   }
   ```
3. **Request Field Defaults**:
   ```json
   "request_defaults": {
     "model": "kokoro",
     "stream": true
   }
   ```

</details>

<details>
<summary><b>Backend 3: WhisperServer (whisper.cpp)</b></summary>

##### Current C++ Wrapper Implementation (`whispercpp_server.cpp`)
1. **ROCm Architecture Matching & NPU Cache Download**: Selects asset filenames based on the target ROCm GPU architecture. For NPU execution, fetches compiled `.rai` cache files from Hugging Face.
2. **Library Path Injection**: Sets `LD_LIBRARY_PATH` for ROCm runtime libraries.
3. **Multipart Audio Translation**: Translates incoming `POST /v1/audio/transcriptions` requests into HTTP multipart form POST requests targeting `/inference` on `whisper-server`.

##### Declarative JSON Specification Alternative
1. **Platform Matrix & Launcher Wrapper**:
   Specific GPU binary assets are declared directly in the platform matrix using hardware tags (`"rocm:gfx1100"`). For NPU execution, downloading compiled `.rai` device cache artifacts and translating OpenAI payloads into `whisper-server`'s custom `/inference` multipart format is handled cleanly by a thin launcher wrapper script (`whisper_wrapper.py`) or by extending `ModelManager` to support precompiled binary device cache dependencies:
   ```json
   "platform": {
     "linux": {
       "rocm:gfx1100": {
         "command": "whisper-server -m {checkpoint:main} --port {port} {custom_args}",
         "env": {
           "LD_LIBRARY_PATH": "{rocm_lib_dir}:{env:LD_LIBRARY_PATH}"
         }
       },
       "npu": {
         "command": "python3 {binary_dir}/whisper_wrapper.py --model {checkpoint:main} --port {port} {custom_args}"
       }
     }
   }
   ```
2. **ModelManager Extension Note**: Extending `ModelManager` to support precompiled binary device cache artifacts (`.rai` files) alongside model weight checkpoints would allow device cache fetching to be handled natively prior to process launch, moving Whisper to pure declarative JSON if upstream `whisper-server` adds native OpenAI endpoint support.

</details>

<details>
<summary><b>Backend 4: ThinkSoundServer (Sound Generation)</b></summary>

##### Current C++ Wrapper Implementation (`thinksound_server.cpp`)
1. **Library Path Modifications**: Prepends ROCm and CUDA library paths to `LD_LIBRARY_PATH` and `PATH`.
2. **Request Field Mapping**: Maps `/audio/generations` parameters (`prompt`, `description`, `cot`) to `/generate` request body fields (`caption`, `description`).
3. **Companion GGUF Selection**: Scans the model directory for six companion GGUF files (`dit`, `t5`, `vae`, `clip`, etc.).

##### Declarative JSON Specification Alternative
1. **Environment Block**:
   ```json
   "env": {
     "LD_LIBRARY_PATH": "{rocm_lib_dir}:{env:LD_LIBRARY_PATH}"
   }
   ```
2. **Field Alias & Endpoint Mapping**:
   ```json
   "request_transform": {
     "target_endpoint": "/generate",
     "field_map": {
       "prompt": "caption",
       "description": "description"
     }
   }
   ```
3. **Multi-File Checkpoint Declaration**:
   ```json
   "command": "thinksound-server --dit {checkpoint:dit} --t5 {checkpoint:t5} --vae {checkpoint:vae} --port {port}"
   ```

</details>

<details>
<summary><b>Backend 5: MoonshineServer (Streaming STT)</b></summary>

##### Current C++ Wrapper Implementation (`moonshine_server.cpp`)
1. **Model Architecture Parameter Resolution**: Maps model variant strings to internal architecture integers (`2` for tiny, `4` for small, `5` for medium).
2. **Multi-Port Allocation**: Probes and allocates three consecutive available ports for HTTP, WebSockets, and TCP streaming.
3. **Environment Setup**: Sets `PYTHONNOUSERSITE=1`.

##### Declarative JSON Specification Alternative
1. **Multi-Port & Architecture Tokens**:
   ```json
   "requested_ports": 3,
   "command": "moonshine-server --model-path {checkpoint_dir:main} --model-arch {custom:arch:-4} --port {port[0]} --ws-port {port[1]} --tcp-port {port[2]} {custom_args}"
   ```
2. **Environment Block**:
   ```json
   "env": {
     "PYTHONNOUSERSITE": "1"
   }
   ```
3. **Standard REST Passthrough**: Routes `POST /v1/audio/transcriptions` directly to the primary subprocess port (`port[0]`) via HTTP passthrough.

</details>

<details>
<summary><b>Backend 6: TrellisServer (3D Generation)</b></summary>

##### Current C++ Wrapper Implementation (`trellis_server.cpp`)
1. **Binary Image Sniffing**: Inspects raw base64 image payloads via `sniff_image()` to detect PNG, JPEG, BMP, or GIF MIME types.
2. **Multipart Form Formatting**: Constructs HTTP multipart form POST requests targeting `/generate` on `trellis-server` using the detected filename and MIME type.
3. **Loader Path Modifications**: Prepends ROCm and CUDA library paths to `LD_LIBRARY_PATH` and `PATH`.

##### Declarative JSON Specification Alternative
1. **Subprocess Launcher Wrapper**:
   Upstream `trellis-server` requires `multipart/form-data` requests with decoded binary image bytes on `/generate`. A thin launcher script (`trellis_wrapper.py`) handles base64 image decoding, MIME format sniffing, and multipart posting.
2. **JSON Descriptor**:
   ```json
   "command": "python3 {binary_dir}/trellis_wrapper.py --model {checkpoint:main} --port {port}",
   "env": {
     "LD_LIBRARY_PATH": "{rocm_lib_dir}:{env:LD_LIBRARY_PATH}"
   }
   ```

</details>

<details>
<summary><b>Backend 7: OpenMossServer (OpenMOSS TTS)</b></summary>

##### Current C++ Wrapper Implementation (`openmoss_server.cpp`)
1. **Library Path Modifications**: Prepends ROCm and CUDA library paths to `LD_LIBRARY_PATH` and `PATH`.
2. **OpenAI Speech Endpoint Passthrough**: Forwards `POST /v1/audio/speech` requests directly to `moss-tts-server` without modifying the payload.
3. **Companion GGUF Selection**: Locates paired `.extras.gguf` companion files when the primary variant uses `.gguf`.

##### Declarative JSON Specification Alternative
1. **Environment Block**:
   ```json
   "env": {
     "LD_LIBRARY_PATH": "{rocm_lib_dir}:{env:LD_LIBRARY_PATH}"
   }
   ```
2. **Standard REST Passthrough & Checkpoint Tokens**:
   ```json
   "command": "moss-tts-server --model {checkpoint_dir:main} --no-webui --port {port} {custom_args}"
   ```

</details>

<details>
<summary><b>Backend 8: RyzenAIServer (Ryzen AI NPU)</b></summary>

##### Current C++ Wrapper Implementation (`ryzenai_server.cpp`)
1. **CLI Command Construction**: Assembles command invocation strings: `ryzenai-server -m <model_path> --port <port> --ctx-size <ctx_size>`.
2. **OpenAI API Passthrough**: Forwards standard OpenAI API requests directly to `ryzenai-server`.
3. **Checkpoint Directory Resolution**: Resolves the model directory containing `genai_config.json`.

##### Declarative JSON Specification Alternative
1. **Launcher Command & Slot Policy**:
   ```json
   "slot_policy": "exclusive_npu",
   "command": "ryzenai-server -m {checkpoint_dir:main} --port {port} --ctx-size {ctx_size} {custom_args}"
   ```
2. **Standard OpenAI REST Passthrough**: Routes OpenAI endpoints directly to the `ryzenai-server` port.

</details>

<details>
<summary><b>Backend 9: FastFlowLMServer (`FastFlowLM` / `flm`)</b></summary>

##### Current C++ Wrapper Implementation (`fastflowlm_server.cpp` & `fastflowlm_models.cpp`)
1. **NPU Hardware & Driver Validation**: Executes `flm validate --json` to verify AMD NPU device presence, driver version, firmware compatibility, and process memlock limits.
2. **Live Stdout Progress Parsing**: Executes `flm pull <checkpoint>` and parses `stdout` in real time for progress tracking.
3. **Health Check & Payload Modification**: Performs health checks on `/api/tags` and injects the exact FLM checkpoint name into `request["model"]`.

##### Declarative JSON Specification Alternative
1. **Pre-flight Validation & Launcher Command**:
   ```json
   "preflight_probe": {
     "command": "flm validate --json",
     "expected_exit_code": 0
   },
   "command": "/usr/bin/flm serve {custom:model_tag:-llama3.2:1b} --host {host} --port {port} --pmode {custom:power_mode:-performance} {custom_args}"
   ```
2. **Health Probe & Slot Policy**:
   ```json
   "slot_policy": "coexist_by_type",
   "health_probe": {
     "type": "http",
     "endpoint": "/api/tags",
     "expected_status": 200,
     "timeout_seconds": 60,
     "poll_interval_ms": 200
   }
   ```

</details>

<details>
<summary><b>Backend 10: AceStepServer (ACE-Step Music Generation)</b></summary>

##### Current C++ Wrapper Implementation (`acestep_server.cpp`)
1. **Async Job Submission & Polling Loop**: Submits jobs to `/lm` or `/synth`, extracts the job ID, polls `/job?id=...` for up to 1200 seconds until status reaches `"done"`, and retrieves final outputs.
2. **Vocal vs. Instrumental Pipeline Branching**: Scans prompts for the `[instrumental]` sentinel. Executes the language model stage (`/lm`) for vocal requests, but bypasses `/lm` for instrumental tracks.
3. **Multipart/Mixed Response Parsing**: Parses HTTP `multipart/mixed` boundary headers to extract raw WAV audio payloads.

##### Declarative JSON Specification Alternative
1. **Subprocess Launcher Wrapper / Container Entrypoint**:
   A 20-line launcher script (`acestep_wrapper.py`) or container entrypoint handles the two-stage `/lm` to `/synth` job polling loop internally and exposes a standard `POST /audio/generations` REST endpoint.
2. **JSON Descriptor**:
   ```json
   "command": "python3 {binary_dir}/acestep_wrapper.py --model {checkpoint:main} --port {port}"
   ```

</details>

### Bounded Schema Primitives vs. Scripting Engine Drift

A key architectural concern is preventing schema creep: ensuring declarative descriptors do not mutate into an un-debuggable, Turing-complete scripting language.

The proposed primitives are strictly bounded and non-Turing-complete:
- **String Token Substitution**: `{checkpoint:name}`, `{port[0]}` (structured `execv` array interpolation, preventing shell injection).
- **Environment Block**: Static key-value pairs (`"env": { "VAR": "VAL" }`).
- **Bounded Request Transform**: Primitive key renames and default value injection (`field_map`, `request_defaults`).
- **Health & Pre-flight Probes**: Exit-code check and HTTP 200 status probes.

**Boundary Invariant**: If a backend requires arbitrary payload logic, multi-stage conditional pipelines, or binary format sniffing, that logic belongs inside a thin launcher wrapper script or container entrypoint, keeping the core JSON schema simple, safe, and closed.

### Summary

- **5 of 10 backends (50%)** map 1:1 to **Pure Declarative JSON** with zero C++ or wrapper code (`RyzenAI`, `OpenMoss`, `ThinkSound`, `FastFlowLM`, `Moonshine`).
- **5 of 10 backends (50%)** map cleanly to **Declarative JSON + Thin Subprocess Launcher Wrapper** (`SdServer`, `Kokoro`, `Trellis`, `AceStep`, `Whisper`), eliminating C++ compilation and daemon release cycles.
- **0 of 10 backends (0%)** require modifying/recompiling C++ core

Even for backends requiring a launcher wrapper script (Categories 2 & 3), shifting engine-specific glue code out of `lemond`'s C++ core and into sandboxed sidecar scripts allows the community to ship and iterate on backends without modifying or recompiling Lemonade C++ source code.

| Backend | Capability | Migration Status | Required Mechanism | Viability Ground Truth |
| :--- | :--- | :--- | :--- | :--- |
| **`RyzenAIServer`** | NPU LLM | **Pure Declarative JSON** | Slot policy + command template + OpenAI REST passthrough | **Works today** (Standard OpenAI REST passthrough) |
| **`OpenMossServer`** | TTS | **Pure Declarative JSON** | Environment block + command template + REST passthrough | **Works today** (Standard OpenAI speech passthrough) |
| **`ThinkSoundServer`** | Sound Gen | **Pure Declarative JSON** | Environment block + `request_transform` field map + directory token | **Works today** (Environment block + field map) |
| **`FastFlowLMServer`** | NPU LLM | **Pure Declarative JSON** | `preflight_probe` + `/api/tags` health probe + `coexist_by_type` | **Works today** (`flm validate --json` preflight) |
| **`MoonshineServer`** | Streaming STT | **Pure Declarative JSON** | Multi-port tokens (`port[0..2]`) + environment block + REST passthrough | **Viable today** (Requires `requested_ports: 3` allocator token) |
| **`WhisperServer`** | STT | **Declarative JSON + Wrapper** | `whisper_wrapper.py` for NPU `.rai` cache fetch & `/inference` multipart POST | **Works today** via launcher script (or `ModelManager` device cache extension) |
| **`SdServer`** | Image Gen | **Declarative JSON + Wrapper** | `request_transform` prompt XML template; thin wrapper for ESRGAN upscaling | **Works today** (`sd_wrapper.py` for `sd-cli -M upscale`) |
| **`KokoroServer`** | TTS | **Declarative JSON + Wrapper** | Standard path templates or thin wrapper for `index.json` file parsing | **Works today** (Recipe file standardization or `koko_wrapper.py`) |
| **`TrellisServer`** | 3D Gen | **Declarative JSON + Wrapper** | Thin wrapper (`trellis_wrapper.py`) for binary decoding & multipart POST | **Works today** (`trellis_wrapper.py` for multipart POST) |
| **`AceStepServer`** | Music Gen | **Declarative JSON + Wrapper** | Thin wrapper (`acestep_wrapper.py`) for 2-stage `/lm` $\rightarrow$ `/synth` job loop | **Works today** (`acestep_wrapper.py` or `JobManager` `POST /jobs`) |

#### Effort to close the gap

1. **Existing Infrastructure (~80%)**:
   - `ExternalBackendServer` (proposed) handles `command` string token expansion (`{checkpoint:main}`, `{port}`, `{threads}`), `env` blocks, `slot_policy`, and HTTP readiness `health_probe`.
   - `ModelManager` already auto-downloads multi-checkpoint dependencies declared under `"checkpoints"` in `server_models.json` (used by SD, ThinkSound, OpenMoss).
   - `ProcessManager` already spawns and manages subprocesses out-of-process.
   - `JobManager` already handles multi-stage job orchestration natively (`POST /jobs`).

2. **Required C++ Extensions (~20%)**:
   - `preflight_probe`: ~15 lines of C++ in `external_backend_server.cpp` to execute pre-flight validation commands before launching the backend process.
   - `requested_ports` / `{port[n]}`: ~15 lines of C++ in `BackendUtils::choose_ports(n)` to reserve contiguous free port ranges.
   - `request_transform`: ~20 lines of C++ in `external_backend_server.cpp` for JSON key renaming and default value injection.
   - `ModelManager` Extension Note: Extending `ModelManager` to handle precompiled binary device cache artifacts (`.rai` files) alongside model weight checkpoints would enable hardware device cache fetching to be handled natively prior to process launch across engines like Whisper.

bitgamma:
@abn I will have to take a detailed look at all of them, but so far the report looks overly optimistic. For example stable-diffusion.cpp has this block:

```
    if (llm_path.empty() || vae_path.empty()) {
        args.push_back("-m");
        args.push_back(model_path);
    } else {
        args.push_back("--diffusion-model");
        args.push_back(model_path);
        args.push_back("--llm");
        args.push_back(llm_path);
        args.push_back("--vae");
        args.push_back(vae_path);
    }
```

which arguments are passed depend if the model is a single checkpoint file or split in 3 checkpoints (not just additional arguments, one actually changes)

Whisper has to handle various input/output formats and temp files. Can the JSON handle new modalities without supporting code? 

This said, the "wrapper" idea might help to match the impedance (not for handling multimedia streams not handled by lemonade codebase though), but what language should this wrapper be? Assuming Python is not possible because the user is not required to have it installed. That leaves us mostly with compiled languages like C, Rust, go (which can be fine)... 

Anyway, I am convinced now that this can work in some cases. Let's say my position on this approach is neutral. I'd rather have a C++ API defined in place of JSON.

abn:
> so far the report looks overly optimistic

This is entirely likely.

> which arguments are passed depend if the model is a single checkpoint file or split in 3 checkpoints (not just additional arguments, one actually changes)

It is possible that we might still be able to cover that with simple primitives, but I agree that this might more than what we want to support in this form.

> Whisper has to handle various input/output formats and temp files. Can the JSON handle new modalities without supporting code?
> 
> This said, the "wrapper" idea might help to match the impedance (not for handling multimedia streams not handled by lemonade codebase though), but what language should this wrapper be? Assuming Python is not possible because the user is not required to have it installed. That leaves us mostly with compiled languages like C, Rust, go (which can be fine)...

Wrapper is likely the best option here. But the choice of language etc would depend, however best option might actually be a container. Then one could argue that that is a dependency we do not want either. I suspect you are correct in that in these cases a compiled shim is probably what you want, in a way there is an elegance to it that allows for a clear separation.

> Anyway, I am convinced now that this can work in some cases. Let's say my position on this approach is neutral. I'd rather have a C++ API defined in place of JSON.

I think the term "plugin" has muddied the waters here a bit. The original intent here was to simply allow for a quick option to run custom backends locally - not in a portable manner, but very specific to a a deployment use case. And hence why the scope for this is really around enabling the simply use cases - ie. a binary/container to run in a process that follows an expected contract which we set.

For a much richer experience, I would say yes, either a C++ API available or better have a shim independent of the core that accepts all the context required and constructs the arguments and prerequisites.

abn:
As an update, I am working on a slightly more involved redesign of what this could be taking into consideration the discussions that have been on-going in here and on discord. Will post something on discord tomorrow, and then align before I rewrite the charter again.

Geramy:
> For a much richer experience, I would say yes, either a C++ API available or better have a shim independent of the core that accepts all the context required and constructs the arguments and prerequisites.

I am on board with this I think a simple C ABI layer would be amazing to have and it really puts us in a great position, we are going to need this, llama.cpp is moving towards that direction as well where they can be used as a ABI not a exe.

Host Side Example
``` 
#include <stdint.h>

// Versioning prevents loading incompatible plugins
#define PLUGIN_API_VERSION 1

// Opaque struct handles for safety
typedef struct PluginInstance PluginInstance;

// A table of function pointers the plugin fills out
typedef struct {
    uint32_t api_version;
    const char* plugin_name;
    
    // Lifecycle functions
    PluginInstance* (*create)(void);
    void (*execute)(PluginInstance* instance, int32_t data);
    void (*destroy)(PluginInstance* instance);
} PluginInterface;

// The single symbol the host application looks up via dlopen/LoadLibrary
typedef const PluginInterface* (*RegisterPluginFunc)(void);
```

Plugin
```
#include "plugin_abi.h"
#include <stdlib.h>

struct PluginInstance {
    int32_t internal_state;
};

PluginInstance* my_plugin_create(void) {
    PluginInstance* inst = malloc(sizeof(PluginInstance));
    if (inst) inst->internal_state = 42;
    return inst;
}

void my_plugin_execute(PluginInstance* instance, int32_t data) {
    if (instance) {
        instance->internal_state += data;
    }
}

void my_plugin_destroy(PluginInstance* instance) {
    free(instance); // Safely freed in the same module that allocated it
}

static const PluginInterface g_interface = {
    .api_version = PLUGIN_API_VERSION,
    .plugin_name = "Custom Math Plugin",
    .create = my_plugin_create,
    .execute = my_plugin_execute,
    .destroy = my_plugin_destroy
};

// Exported root function
extern "C" const PluginInterface* register_plugin(void) {
    return &g_interface;
}
```

Host Plugin Loading
```
#include <iostream>
#include <dlfcn.h>
#include "plugin_abi.h" // Shared header containing PluginInterface structure

void load_posix_plugin(const char* path) {
    // 1. Open the dynamic library file into memory
    void* library_handle = dlopen(path, RTLD_NOW | RTLD_LOCAL);
    if (!library_handle) {
        std::cerr << "Failed to load library: " << dlerror() << std::endl;
        return;
    }

    // 2. Clear any existing errors, then find the entry point symbol
    dlerror(); 
    auto register_func = (RegisterPluginFunc)dlsym(library_handle, "register_plugin");
    
    char* error_str = dlerror();
    if (error_str != nullptr) {
        std::cerr << "Failed to find symbol: " << error_str << std::endl;
        dlclose(library_handle);
        return;
    }

    // 3. Execute the function to pull the ABI data structure
    const PluginInterface* plugin_abi = register_func();

    // 4. Validate the ABI layout version before execution
    if (plugin_abi->api_version != PLUGIN_API_VERSION) {
        std::cerr << "ABI mismatch! Expected v" << PLUGIN_API_VERSION 
                  << " but got v" << plugin_abi->api_version << std::endl;
        dlclose(library_handle);
        return;
    }

    // 5. Use the ABI safely
    std::cout << "Successfully loaded: " << plugin_abi->plugin_name << std::endl;
    PluginInstance* instance = plugin_abi->create();
    plugin_abi->execute(instance, 100);
    plugin_abi->destroy(instance);

    // 6. Clean up when finished
    dlclose(library_handle);
}

```

We already are very close to this anyways. @abn

abn:
Cross posting here from the discord thread for visibility.

> 👋 Hi folks. Thanks again for the discussion on the charter PR. It was useful, and it convinced me to step back and take a first-principles look at this rather than just extend my original proposal which was solving an immediate problem I had.
> 
> I've put together a distilled design write-up here https://abn.is/a/tFMKuUcE84qD/ that hopefully addresses more cases.
> 
> A quick orientation, so it's easy to jump into as it is a long one:
> 
> - It's a design proposal, not an implementation. It's an attempt to lay out the backbone of what a sandboxed backend/plugin system for Lemonade could look like.
> - The core idea: put the trust boundary at the process boundary. lemond stays the orchestrator, fact authority, and resource broker; third-party backends run as sandboxed, out-of-process adapters behind a narrow, typed contract. Declarative where possible, code only where a backend genuinely needs it.
> - It's structured so the common case is low-friction, ie. many engines can plug in via a manifest with no adapter code at all, while the ones that need procedural adaptation get a real adapter API without us needing to be too prescriptive on what they use to write it.
> - I've also tried to be upfront about scope: I'd suggest we keep the working group focused on the capability gateway + sandboxing, and treat marketplace/distribution/discovery as naturally-additive follow-ons.
> 
> Happy to get your thoughts on this.

abn:
Closing this as the working group process is no longer appropriate for this change and there is not much interested in this working group as it stands today. This can be revisited later as separate working groups for backend plugins and backend sandboxing.

PR #2967, opened 2026-08-07
fix(gui3): match GUI2 model search filters

### LINKED ISSUE #2948: Copy the the model search filters from GUI2
In GUI2 we only show model search results from trusted orgs, like unsloth etc.

Image

GUI3 appears to be missing these features entirely:

Image

kpoineal:
## Summary

- match GUI2's top-12, download-ranked GGUF search behavior
- exclude GUI2's non-LLM audio, image, and video pipeline categories
- fall back from failed or empty variant discovery to GUI2's repository/file backend detection, retaining FLM and ONNX while excluding unknown, `sd-cpp`, `whispercpp`, and `moonshine` repositories
- preserve GUI3 provider toggles, local de-duplication, task/backend/tag filters, unresolved-result loading behavior, and downloads for fallback-detected repositories

GUI2 does not use a hard-coded organization allowlist. The issue screenshot reflects its download-ranked top results; the effective trust boundary is the 12-result cap plus pipeline and repository/backend compatibility checks. This ports those exact repository-backed semantics rather than introducing new publisher policy.

## Validation

- `npm --prefix src\app run typecheck`
- `npm --prefix src\app run build:renderer`
- `npm --prefix src\app run test:model-providers`
- `npx playwright test a11y.spec.ts --grep 'A15[23]'`

## Visual review

| Before: GUI3 returns 14 mocked candidates | After: GUI2's top-12 cap is applied |
| --- | --- |
| Before model search filtering | After model search filtering |

Closes #2948.


fl0rianr (commented):
This still seems to miss one part of GUI2's Hugging Face compatibility filtering.

GUI2 first applies the pipeline-tag exclusions, but then `compatibleHfSearchResults` also removes repositories once backend detection resolves to `null`, as well as results detected as `sd-cpp`, `whispercpp`, or `moonshine`.

That matters for repos with a missing or misleading `pipeline_tag`: GUI2 can still identify those from the repository/files and remove them, while GUI3 currently keeps the result when variant enrichment fails and defaults missing variant metadata to `llamacpp`.

Could we preserve the GUI2 backend-compatibility step here as well? A regression test with an incompatible GGUF result that has no excluded `pipeline_tag` would also help cover this case.

kpoineal:
Addressed Florian's backend-compatibility review in 374696c4. Hugging Face rows now remain visible while variant discovery is unresolved, then are removed when discovery fails, returns no downloadable variants, or resolves to sd-cpp, whispercpp, or moonshine, matching GUI2. The browser regression fixture includes an incompatible GGUF result with no excluded pipeline_tag, and the runtime contract covers failed/empty discovery plus all three excluded recipes.

jeremyfowers:
Awesome, screenshots look a lot closer to what @PCAssistSoftware was expecting.

fl0rianr (commented):
Sorry, this found one: 

GUI2 does not treat a failed or empty `/pull/variants` response as immediately incompatible. It falls through to its repository/file-based backend detection, which can still keep supported FLM or ONNX results. `compatibleHfSearchResults` only removes `null`, `sd-cpp`, `whispercpp`, and `moonshine`.

Here, any exception or empty variant result is added directly to `incompatibleHfVariantKeys`, so a result that GUI2 would recover as `flm` or `ryzenai-llm` can disappear.

The new browser fixture also mocks `/pull/variants` returning `recipe: 'sd-cpp'` for a GGUF repo, but the real GGUF variant endpoint currently returns `recipe: 'llamacpp'`. So that test does not quite exercise the production contract.

Merge after this one.

kpoineal:
Addressed the follow-up review in 7c3fe474. Failed or empty variant discovery now falls through to GUI2-equivalent repository metadata detection: FLM and ONNX results remain, while unknown, sd-cpp, whispercpp, and moonshine results are removed. The browser fixture now uses empty production-shaped variant responses plus Hugging Face repository metadata, and it verifies the recovered FLM result remains downloadable. I also rebased onto the current GUI3_merging tip and resolved the conflicts.

kpoineal:
@fl0rianr Thanks for catching the fallback mismatch. I updated the fix so failed or empty \/pull/variants\ responses now follow GUI2's repository/file-based backend detection instead of being rejected immediately. Supported FLM and ONNX repositories stay visible and are directly downloadable; unknown repositories and those detected as \sd-cpp\, \whispercpp\, or \moonshine\ are filtered out. I also replaced the unrealistic \sd-cpp\ GGUF mock with production-shaped empty-variant and Hugging Face metadata responses, added coverage for FLM/ONNX recovery and incompatible detections, rebased onto the latest \GUI3_merging\, and resolved the merge conflicts. The PR is now mergeable, with the refreshed CI run in progress.

fl0rianr (approved):
Thanks, LGTM

PCAssistSoftware:
just updated my install of the gui3_merging branch and search issue reported here https://github.com/lemonade-sdk/lemonade/issues/2948 still present ?

fl0rianr:
I will look further in to that.

PR #2968, opened 2026-08-07
feat(telemetry): capture prefix-cache effectiveness (cache_n, cached_tokens) and route-switch counts

### LINKED ISSUE #2955: [Router] Telemetry: capture prefix-cache effectiveness (cache_n, cached_tokens) and route-switch counts
## Problem

Lemonade captures TTFT, token counts, and TPS per request, but nothing measures **prefix-cache effectiveness** or **route stability** — the two signals that determine whether a routed conversation is paying re-prefill cost on every turn.

Evidence (main@30977658):

- llama-server returns `timings.cache_n` ("number of prompt tokens reused from cache") alongside `timings.prompt_n` in every response, and the total context equals `cache_n + prompt_n + predicted_n`. Lemonade parses `prompt_n`, `predicted_n`, `prompt_ms`, `predicted_per_second` (`src/cpp/server/server.cpp:3107-3144` non-streaming; `src/cpp/server/streaming_proxy.cpp:41-59` streaming) but **`cache_n` appears nowhere in `src/cpp`** (grep: zero matches). We cannot distinguish "20k-token prompt, 19.8k served from cache" from "20k-token full re-prefill" — they log identical `prompt_tokens`.
- Cloud providers report prompt-cache usage in `usage.prompt_tokens_details.cached_tokens` (and `cache_write_tokens` on newer OpenAI models). The cloud backend intentionally returns provider bodies unchanged so the handler picks up `usage` (`src/cpp/server/backends/cloud/cloud_server.cpp:449-452`), but the handler only reads `prompt_tokens`/`completion_tokens` (`server.cpp:3145-3186`) — `cached_tokens`/`prompt_tokens_details` appear nowhere in `src/cpp` (grep: zero matches). Cloud cache hit/miss (and its cost impact) is invisible.
- Routing decisions are per-request with no cross-request view (`RouterDispatchResult` carries only `requested_model`/`selected_model`/`decision`), so there is no measurement of how often consecutive turns of the same conversation flip between candidates ("ping-pong"), which is the primary driver of re-prefill waste in `collection.router` deployments.

Upstream llama.cpp declined to expose KV/prompt-cache metrics in `/metrics` (ggml-org/llama.cpp#23632, closed not-planned), so per-request capture in Lemonade is the only available source.

## Proposed approach

Three additive telemetry signals, no behavior change:

1. **`cache_tokens`** — per-model latest + cumulative total, from `timings.cache_n` (llamacpp) and `usage.prompt_tokens_details.cached_tokens` (cloud/OpenAI-wire). Optionally record `cache_write_tokens` when present.
2. **Route-switch counter** — for requests dispatched through a `collection.router`, maintain a small LRU map keyed by a cheap conversation fingerprint (e.g., 64-bit hash of the system prompt + first user message text). Count `routing_decisions_total` and `routing_switches_total` (fingerprint seen before AND `route_to` differs from its previous value). The fingerprint is an internal metric key only — not session state, no effect on routing.
3. Expose all of the above in `GET /v1/stats` and Prometheus `/metrics`.

## Implementation sketch

1. `src/cpp/include/lemon/wrapped_server.h` — extend `struct Telemetry` (`:35-71`): add `int cache_tokens`, `uint64_t cache_tokens_total`; include both in `to_json()` (`:58`) and `reset()`.
2. `src/cpp/server/streaming_proxy.cpp:41-59` — parse `timings.cache_n` into the telemetry struct; parse `usage.prompt_tokens_details.cached_tokens` next to the existing `usage` handling.
3. `src/cpp/server/server.cpp:3107-3186` (and the sibling completions block at `:3326+`) — read `cache_n`/`cached_tokens` and pass through `router_->update_telemetry(...)`. Extend `Router::update_telemetry`/`record_telemetry_for_model` (`src/cpp/server/router.cpp:2043`, `:1991`) with the new field; accumulate totals like `input_tokens_total` (`:2018`).
4. `src/cpp/server/router.cpp:1900-1903` (`get_stats`) and the per-model totals block (`:1967-1970`) — include `cache_tokens`, `cache_tokens_total`, `routing_decisions_total`, `routing_switches_total`.
5. `src/cpp/server/prometheus_metrics.cpp:333-343, 354-382` — add `lemonade_model_cache_tokens` (gauge), `lemonade_model_cache_tokens_total` (counter), `lemonade_routing_decisions_total`, `lemonade_routing_switches_total` (counters).
6. Switch counter: in the router-dispatch path (`src/cpp/server/server.cpp:2894` `apply_router_collection_dispatch` and the chat-path dispatch at `:2967`), after the `Decision` is computed, update a mutex-guarded `lru_map<fingerprint, last_route_to>` (bounded, ~1024 entries) owned by `Server` or `Router`; increment counters.

## Acceptance criteria

- `test/server_llm.py` (llamacpp): send the same multi-turn conversation twice; assert `/v1/stats` shows `cache_tokens > 0` on the second request (llama-server `cache_prompt` defaults to true).
- `test/server_telemetry.py`: `/metrics` contains the four new series; `cached` counters are monotonic.
- `test/server_router.py`: a router collection alternating two candidates on consecutive fingerprint-identical requests increments `routing_switches_total`; identical routing does not.
- No new fields are emitted into inference response bodies (stats/metrics only). Streaming and non-streaming paths both record `cache_tokens`.

## Out of scope

- Any routing behavior change (affinity/hysteresis is a separate issue).
- Estimating cache state before dispatch (this issue is measurement only).
- FLM/RyzenAI cache metrics (no equivalent field documented).

## References

- llama-server `timings.cache_n` / `prompt_n` semantics: https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md
- Upstream declined `/metrics` KV-cache gauges: https://github.com/ggml-org/llama.cpp/issues/23632
- OpenAI cached-token usage fields: https://developers.openai.com/api/docs/guides/prompt-caching
- Related: #2389 (router tracker).

Size: S

ramkrishna2910:
Closes #2955.

Adds the measurement layer for router cache behavior — no routing behavior changes.

- **cache_tokens**: recorded per model (latest + cumulative) from llama-server `timings.cache_n` and OpenAI-wire `usage.prompt_tokens_details.cached_tokens`, on both streaming and non-streaming paths, local and cloud. `-1` internally means "not reported", so a real cold-prefill `0` is distinguishable.
- **Route-stability counters**: `routing_decisions_total` / `routing_switches_total` for `collection.router` dispatch, keyed by a conversation fingerprint (FNV-1a of system prompt + first user message; bounded 1024-entry LRU; metrics key only — never influences routing, no content stored). `/routing/validate` does not tick the counters.
- Exposed in `GET /v1/stats` and Prometheus `/metrics` (5 new series), documented in `docs/api/lemonade.md`.
- Refactor: `WrappedServer::TelemetryCallback` now passes `StreamingProxy::TelemetryData` instead of five scalars, so backend-reported fields extend without signature churn (cloud/vllm/router call sites updated).

Verified: `test_telemetry_helpers` all green (new parse + fingerprint cases); live e2e on a scratch lemond — second request with a shared prefix reports `cache_tokens: 40` with `input_tokens: 19`, `prompt_tokens: 59` (llama-server reused 40 of 59 prompt tokens). New integration tests: `server_endpoints` (series presence), `server_router` (exact-delta switch counting), `server_llm` (cache_tokens > 0 on repeated prefix, gated on llamacpp).

Note for reviewers: #2452/#2898 touch `streaming_proxy.cpp`/`wrapped_server.h` (different scope — textual merge risk only); no open PR overlaps functionally.


fl0rianr (commented):
Thanks, overall this looks well structured, especially the routing counters and the TelemetryData callback refactor. I found a few issues around cache telemetry that I think should be fixed before merging:

* **Stale cache_tokens value:** if one request reports cached tokens and the next request does not, `record_cache_tokens_for_model()` returns early and keeps the previous value. `/stats` and the Prometheus gauge can then report an old cache value as if it belonged to the latest request. The “not reported” state should be preserved instead of falling back to/sticking at a numeric value.

* **Cloud streaming does not reliably collect usage:** the code parses the final usage chunk, but CloudServer does not set `stream_options.include_usage=true`. Unless the client already requested it, most OpenAI-compatible providers will not send the usage block, so `cache_tokens` remains unavailable for normal cloud streaming requests. Since the PR claims streaming + non-streaming coverage for cloud, this should be handled internally without changing the response seen by clients.

* **Responses API uses a different field:** Responses usage reports cached tokens under `input_tokens_details.cached_tokens`, while the parser currently only checks `prompt_tokens_details.cached_tokens`. A small regression test for this case would be useful too.

The route-switch LRU/counter logic itself looks good to me and the test coverage is heading in the right direction. I’d fix the three cache-telemetry cases above before merging.

ramkrishna2910:
@fl0rianr thanks for the careful review — all three cache-telemetry issues are fixed in 3a3ca7ce, and the branch is merged up with main (the conflict was an include-list overlap in router.h).

**1. Stale cache_tokens value** — `cache_tokens` is now an explicit tri-state: `record_telemetry_for_model()` (which runs for every recorded request) resets the latest-value gauge to *unreported* before any cache recording happens, and `Telemetry::to_json()` renders unreported as JSON `null`. The Prometheus gauge is simply not sampled for that scrape (the existing number-guard skips null), so neither `/stats` nor `/metrics` can attribute an old value to a request that reported nothing. Cumulative `cache_tokens_total` is unaffected.

**2. Cloud streaming usage** — `CloudServer::forward_streaming_request` now injects `stream_options.include_usage=true` into the forwarded request when (and only when) the client didn't set it, and switches the relay to line-mode to swallow the resulting usage-only frame — so telemetry gets provider-reported usage while the client-visible stream is byte-for-byte what they asked for. When the client *did* request usage, the raw passthrough path is unchanged and they receive the frame as before. Swallowing is deliberately limited to frames with usage and empty/absent `choices`: a provider that attaches usage to a content-bearing final chunk keeps its content intact. Covered by a new e2e test (`server_router.py::test_611`) with a streaming mock provider that asserts all three sides: the upstream request contained `include_usage`, the client stream contains no usage frame, and `/stats` shows the mock's `cached_tokens`/`prompt_tokens` — plus the explicit-opt-in passthrough case. Verified green against a live server locally.

**3. Responses API field** — the streaming parser now also reads `usage.input_tokens_details.cached_tokens` (with `prompt_tokens_details` taking precedence when both appear), with regression cases in `test_telemetry_helpers` for the Responses shape and the precedence rule.

Docs updated for the `null` semantics. Full unit suite green; `test_605`/`test_610`/`test_611` verified live.

fl0rianr (commented):
Thanks for the fixes, this is looking much better overall. The three points from my previous review are addressed well.

I still see two correctness issues before I’d approve:

1. `cache_tokens` can still become stale with concurrent requests, because the main telemetry update and the cache update happen under separate locks. One request can therefore overwrite the “latest” cache value after a newer request already completed. I’d update the request telemetry atomically in one step.

2. Non-streaming `/v1/responses` still doesn’t seem to record `usage.input_tokens_details.cached_tokens`. The Responses streaming parser handles it now, but the non-streaming path doesn’t call `update_cache_tokens()`.

A small concurrency regression test plus a non-streaming Responses test would cover both nicely.

Other than that, the changes look good to me, especially the `null` handling and the cloud streaming usage injection.

ramkrishna2910:
@fl0rianr both issues fixed in 59a252fd.

**1. Atomic per-request recording** — the three separate recorders are collapsed into a single `record_request_telemetry_for_model()` that applies one request's entire update (input/output/ttft/tps, prompt, cache incl. the unreported reset, counters and totals) under one `telemetry_mutex_` hold. Concurrent requests now interleave whole updates, never fields of two requests. The old `update_telemetry`/`update_prompt_tokens`/`update_cache_tokens` trio is gone; `TelemetryData` grew a `prompt_tokens` field so the struct carries the full per-request record, and a new `StreamingProxy::extract_telemetry()` parses complete response bodies — which also deduplicates the twin chat/completions handler blocks into one `Server::record_response_telemetry()`.

**Concurrency regression test** — `server_router.py::test_612`: 16 concurrent streaming requests against a (now threading) mock provider, all reporting identical usage; asserts the final latest gauges equal that usage exactly (`cache_tokens == 37`, not null/stale) and the cumulative counters are exact multiples (`cache_tokens_total += 16×37`, `request_count_total += 16`). Verified green live; under the old split-lock code the final-gauge assertion is exactly what the race breaks.

**2. Non-streaming `/v1/responses`** — `handle_responses` now calls the same `record_response_telemetry()`, and the shared extractor reads `usage.input_tokens_details.cached_tokens` (Responses field names). Unit-tested in `test_telemetry_helpers` (Responses-shape body → input/prompt/cache extracted; chat-shape precedence: timings override input, usage `prompt_tokens` preserved), and `server_llm.py::test_007` now asserts `/stats` records after a non-streaming Responses call. Live probe on a scratch server: `POST /v1/responses` → `/stats` shows `input_tokens: 14, output_tokens: 8, cache_tokens: 0` (a *reported* zero from `input_tokens_details`, distinct from null).

fl0rianr (approved):
Looks good from my side now. Just two small nits before merge:

* The unrelated `modelscope-blog/*` files should be removed from this PR. I pretty much guess this was not intentional.
* `test_612` would be stronger with different usage values across concurrent requests, so it can actually catch mixed/stale telemetry.

Neither changes my view on the telemetry fix itself.

PR #2971, opened 2026-08-08
Minor refactor of new gate code 

SlawomirNowaczyk:
The three gate jobs from https://github.com/lemonade-sdk/lemonade/pull/2950 now share one composite action.

**New file:** action.yml — the jq logic lives here once, taking `needs` (JSON) and `event-name` as inputs.

**Refactored:** `packaging-gate`, `backends-gate`, `macos-gate` in cpp_server_build_test_release.yml each shrank from ~19 lines of inline shell to a two-step `checkout` + `uses: ./.github/actions/check-gated-jobs`. Net removal of the triplicated script; a future fix to the gate logic now lands in one place.

Two behavior-preserving notes:
- Each gate now runs `actions/checkout@v5` first — required because a local composite action must exist on disk before it can be used. Job names (`Packaging builds`, `Inference backend tests`, `macOS builds`) are unchanged, so required-status-check registration is unaffected.
- `github.event_name` is now passed through the `event-name` input (surfaced as `$EVENT_NAME` env inside the action) instead of being interpolated inline — same value, no functional change.

### DIFF
--- a/.github/actions/check-gated-jobs/action.yml
+++ b/.github/actions/check-gated-jobs/action.yml
@@ -1,0 +1,34 @@
+name: Check gated jobs
+description: >-
+  Fail if any needed job broke; in the merge queue, also fail if any never ran.
+inputs:
+  needs:
+    description: 'toJSON(needs) from the calling job'
+    required: true
+  event-name:
+    description: 'github.event_name from the calling workflow'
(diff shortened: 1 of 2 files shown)


fl0rianr (approved):
LGTM

PR #2973, opened 2026-08-08
fix(server): sum sibling shards when computing on-disk GGUF size

### LINKED ISSUE #2972: [Server] Model size in /v1/models and Web UI sums only one shard for folder-sharded GGUF
### Platform

Windows

### Lemonade Version

11.5.2

### GPU / APU Model

### Component

lemond (server)

### Bug Description

Folder-sharded GGUF variants from unsloth-style repos (e.g. `unsloth/Laguna-S-2.1-GGUF`, quant `UD-Q4_K_XL`) show an incorrect size in the model list of the Web UI / `GET /v1/models`: only the first shard's size is reported (~3 MB) instead of the sum of all shards (~68 GB).
The first shard of the folder-sharded layout is a small stub file (3,683,648 bytes), so the displayed size is ~3 MB. The marketplace/search cards show the correct (~68 GB) size because `/pull/variants` sums all shards — only the downloaded-model list is wrong.

### Steps to Reproduce

1. Pull `unsloth/Laguna-S-2.1-GGUF:UD-Q4_K_XL` via the Web UI or `lemonade pull`.
2. Wait for the download to complete (all 3 shards land in `UD-Q4_K_XL/`).
3. Open the Web UI model list (or call `GET /v1/models`).
4. Observe the model's size shows ~3 MB instead of ~68 GB.

Expected: The model list reports the total on-disk size of all shards (~68 GB).
Actual: Only the first shard file is sized (~3 MB / 0.003 GB).

### Log Output

```shell

```

### Additional Context

Root cause: `resolve_gguf_path()` (`src/cpp/server/backends/llamacpp/llamacpp_gguf.cpp`, Case 4) resolves a folder-sharded quant to the first file in the folder, and `refresh_on_disk_size()` (`src/cpp/server/model_manager.cpp`) then sizes only that single file. A fix is in progress that makes `resolved_path_size_bytes()` shard-aware (sums sibling `-NNNNN-of-NNNNN.gguf` files in the same directory).

blackdeathdrow:
Fixes #2972

Folder-sharded GGUF variants (unsloth-style repos, e.g. `Laguna-S-2.1-GGUF:UD-Q4_K_XL`) showed only the first shard's size (~3 MB) in the model list instead of the total (~68 GB).

`resolve_gguf_path()` resolves such quants to a single shard file, which in these layouts is a small stub. `refresh_on_disk_size()` now sums the whole shard family via `sharded_gguf_size_bytes()`, so the model list reports the real on-disk total. `resolved_path_size_bytes()` deliberately stays per-file so `/v1/models/{id}/files` keeps reporting each file's own size.

Shard filenames are matched on `-NNNNN-of-NNNNN.gguf` with `-`, `.` or `_` as the separator, matching the conventions already recognized in `hf_variants.cpp`. Siblings are grouped on both the base name and the `-of-N` count, and `index > total` is rejected.

Covered by `test_gguf_shard_utils` (exercises the production aggregation directly) and an endpoint test asserting the model size / `/files` split.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2309,6 +2309,24 @@ if(BUILD_TESTING AND EXISTS "${_GGUF_CAPS_TEST_SRC}")
     add_cpp_ci_test(GgufCapabilitiesTest CI OFF COMMAND test_gguf_capabilities)
 endif()
 
+set(_GGUF_SHARD_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_gguf_shard_utils.cpp"
+)
+if(BUILD_TESTING AND EXISTS "${_GGUF_SHARD_TEST_SRC}")
+    add_executable(test_gguf_shard_utils
+        test/cpp/test_gguf_shard_utils.cpp
+    )
+    target_include_directories(test_gguf_shard_utils PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+    target_link_libraries(test_gguf_shard_utils PRIVATE lemonade-server-core)
+    add_dependencies(test_gguf_shard_utils copy_resources)
+
+    include(CTest)
+    add_cpp_ci_test(GgufShardUtilsTest CI ON COMMAND test_gguf_shard_utils)
+endif()
+
 # Model residency role/pool contract for router dependencies
 set(_MODEL_RESIDENCY_TEST_SRC
     "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_model_residency.cpp"
--- a/src/cpp/include/lemon/gguf_shard_utils.h
+++ b/src/cpp/include/lemon/gguf_shard_utils.h
@@ -1,0 +1,58 @@
+#pragma once
+
+#include <cstdint>
+#include <filesystem>
+#include <regex>
+#include <string>
+
+namespace lemon {
+
+// Separators mirror hf_variants.cpp and visible_extra_variant_name(), which
+// both accept `-`, `.` and `_` before the shard index.
+inline bool is_gguf_shard_filename(const std::string& filename, std::string* base,
+                                   int* total = nullptr, int* index = nullptr) {
+    static const std::regex shard_pattern(R"([-._](\d+)-of-(\d+)\.gguf$)", std::regex::icase);
+    std::smatch match;
+    if (!std::regex_search(filename, match, shard_pattern)) {
+        return false;
+    }
+
+    int parsed_index = 0;
+    int parsed_total = 0;
+    try {
+        parsed_index = std::stoi(match[1].str());
+        parsed_total = std::stoi(match[2].str());
+    } catch (...) {
+        return false;
+    }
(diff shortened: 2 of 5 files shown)


fl0rianr (commented):
Thanks for the fix — the approach makes sense for #2972, but I see two issues before this is safe to merge:

1. `resolved_path_size_bytes()` is also used by `list_model_files()`. With this change, `/v1/models/{id}/files` would report the total size of all shards for the first shard entry, even though `size_bytes` is supposed to describe that specific file. I think the shard aggregation should only be used for the model-level size calculation, while the files endpoint keeps the real file size.

2. The shard regex is a bit too broad. `-(\d+)(-of-\d+)?\.gguf` would also treat names like `model-7.gguf` and `model-8.gguf` as shards of the same model. It would be safer to match the actual shard naming convention more strictly.

It would also be good to add regression tests covering the sharded model size, unrelated GGUF files in the same directory, and the `/files` endpoint behavior.

Other than that, the fix is nicely scoped and addresses the reported issue well.

fl0rianr (commented):
Looks much better, thanks — the previous issues are addressed. One remaining edge case: is_gguf_shard_filename() captures the total shard count, but sharded_gguf_size_bytes() groups siblings only by base. This means e.g. model-00001-of-00003.gguf and model-00001-of-00005.gguf in the same directory would be treated as one shard family and summed together. Could we include the of-N count in the family match as well? After that this looks good to me.

blackdeathdrow:
Done. `sharded_gguf_size_bytes()` now matches siblings on both `base` and the `-of-N` count. Though two shard sets differing in `-of-N` in the same directory is unlikely.

fl0rianr (commented):
One remaining concern: the new shard helper bypasses the Windows-safe HF cache path handling in `model_manager.cpp` and uses `std::filesystem` directly. Could this regress symlinked/reparse-point HF caches on Windows?

Also, the `of-N` regression test puts the two families in different directories, so it would pass even with the old base-only matching. They should be in the same directory.

Please be aware we don't want verbose comments that can be explained by code.

blackdeathdrow:
The helper is now in `model_manager.cpp` using `safe_is_directory`/`safe_dir_options`, the `-of-N` regression test uses a single directory, and the verbose comments are gone. Also added `index`/`total` validation.

fl0rianr (commented):
Two things before approval: `is_gguf_shard_filename()` still accepts `00004-of-00003`, while the new test expects it to fail (`index > total` is missing).

Also, `sum_family()` reimplements the aggregation in the test instead of exercising the production `sharded_gguf_size_bytes()` path, so it wouldn't catch a regression in the actual #2972 fix.

blackdeathdrow:
Both addressed.
`index > total` moved into `is_gguf_shard_filename()`, so the detector owns shard-number validation and the redundant check in `same_shard_family()` is gone.
`sharded_gguf_size_bytes()` is now declared in the header (definition stays in `model_manager.cpp` with the Windows-safe helpers) and the test links against `lemonade-server-core` and calls it directly — I verified it fails if the `-of-N` grouping is reverted to base-only.
Also extended the separator set to `-`, `.` and `_` to match `hf_variants.cpp`, and added the `/v1/models/{id}/files` regression you asked for in the first round.

fl0rianr (approved):
LGTM, thanks!

PR #2975, opened 2026-08-08
fix(server): frame backend errors as SSE events on the streaming path

### LINKED ISSUE #2826: Streaming chat/completions: backend errors are forwarded without SSE framing, so clients see a silent empty stream
### Platform

Linux/Fedora (Fedora 44)

### Lemonade Version

11.5.0

### GPU / APU Model

AMD Radeon 890M (Strix)

### Component

lemond (server)

### Bug Description

When a streaming `/api/v1/chat/completions` request fails in the backend (e.g. llama-server rejects it with 400 "request exceeds the available context size"), lemond forwards the backend's JSON error body to the client **verbatim, without SSE framing** — no `data: ` prefix — inside a `200 OK` / `text/event-stream` response, then closes the stream without `[DONE]`.

Per the SSE spec, clients must ignore lines that aren't recognized fields, so the error is invisible to any compliant SSE parser. The client sees an empty stream that just ends. In pi (coding agent) this surfaces as `Error: Stream ended without finish_reason` — with no hint that the actual problem was a too-small context window, which the server log states perfectly clearly:

```
E srv send_error: task id = 88, error: request (19185 tokens) exceeds the available context size (16384 tokens), try increasing it
[Error] (StreamingProxy) Backend returned error: 400
```

### Steps to Reproduce

1. Load any llamacpp model with a small context, e.g. `ctx_size: 16384`
2. Send a streaming chat/completions request whose prompt exceeds it
3. Observe the raw response:

```
$ curl -sN --raw -i http://localhost:13305/api/v1/chat/completions -H 'Content-Type: application/json' --data @overflow_req.json
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Content-Type: text/event-stream
...

{"error":{"code":400,"message":"request (550200 tokens) exceeds the available context size (262144 tokens), try increasing it","type":"exceed_context_size_error","n_prompt_tokens":550200,"n_ctx":262144}}
```

(raw JSON chunk, not an SSE event; then the stream closes — no `[DONE]`)

Expected: the client receives the backend error in a form an OpenAI-style SSE client can surface, e.g. a framed `data: {"error": {...}}` event before the stream ends.

Actual: error bytes are present on the wire but unparseable as SSE; clients report generic "stream ended unexpectedly" errors and users have no idea why (see #1368, lemonade-sdk/lemonade-pi-plugin#2 for people hitting this blind).

### Suggested solution

`forward_byte_stream()` in `streaming_proxy.cpp` already handles this correctly: it registers a status callback with `HttpClient::post_stream`, diverts the non-200 body away from the client sink, and reshapes it into a `{"error": {message, type, status}}` payload. `forward_sse_stream()` passes `nullptr` as the status callback, so the raw error body leaks into the SSE stream unframed.

Applying the same pattern to the SSE path — divert the non-200 body, then emit it as a properly framed `data: {"error": ...}\n\n` event before `sink.done()` — would make the existing error information visible to clients. (Whether `[DONE]` should follow an error event is your call; OpenAI does not send it after in-stream errors.)

Related: #745 asks for better error content on the crash/network path; this issue is specifically about the streaming non-200 path where the (already useful) error body is lost to framing. PR #1340 fixed the clean-termination half; this is the visibility half.

I'm happy to send a PR following the byte-stream pattern (plus a case in `test/server_streaming_errors.py` asserting the error event is client-visible) if you'd take one — or leave it to you if you'd rather shape the SSE error contract yourselves.

Yigtwxx:
Fixes #2826.

`forward_sse_stream()` writes the backend's error body straight to the client sink. The response is already committed as `200 OK` / `text/event-stream`, so the body arrives without a `data:` prefix, and every spec-compliant SSE parser drops it. The client sees an empty stream that just ends, with no hint of what went wrong.

## Root cause

`post_stream()` documents its `on_status` hook for exactly this case:

> on_status fires once, before the first chunk is delivered, so callers can divert an error body instead of forwarding it as payload bytes.

`forward_byte_stream()` in the same file uses it: it diverts the non-200 body into a capped buffer and reshapes it into `{"error": {message, type, status}}` before writing. `forward_sse_stream()` passes `nullptr`, so every byte reaches the sink including the error body.

Captured on `main` — the whole of what a client receives for a context overflow:

```
{"error":{"code":400,"message":"request (6009 tokens) exceeds the available context size (4096 tokens),
 try increasing it","type":"exceed_context_size_error","n_prompt_tokens":6009,"n_ctx":4096}}
```

The diagnosis is already useful; only the framing is missing.

## Fix

Apply the byte-stream pattern to the SSE path: register the status callback, buffer the non-200 body instead of forwarding it, and emit it as one framed `data:` event before the stream closes. The payload shape is the one `forward_byte_stream()` already produces, so both streaming paths report failures identically.

The issue asked whether `[DONE]` should follow the error event. It does not here: OpenAI does not send `[DONE]` after an in-stream error, and `sink.done()` still terminates the response cleanly, so clients that iterate to end-of-stream are unaffected. Easy to change if you would rather always terminate with `[DONE]`.

The `[DONE]` synthesis on the clean-transport path and the interrupted-transport `throw` are untouched.

## Testing

Windows, llamacpp/vulkan, `Tiny-Test-Model-GGUF`.

- Added `test/server_streaming_errors.py::test_004a_context_overflow_error_is_sse_framed`. It reuses the request from `test_004`, which reaches the same backend 400 but only asserts the stream terminates, and additionally asserts every non-empty line is a recognised SSE field and that a `data:` event carries an `error` object. Verified it fails on `main` — the failure output is the unframed line quoted above — and passes with this change.
- Full suite green: `python test/server_streaming_errors.py` — 7/7, including `test_006`, which asserts `[DONE]` on the success path.
- This suite runs in `test-cli-endpoints-linux`, so the new case is covered on every PR.
- Not run on Linux or macOS; the change is platform-independent.

The diagnosis in #2826 is @ekenberg's — the issue identifies `forward_byte_stream()` as the reference implementation and offered a PR twelve days ago. This follows that plan.

### DIFF
--- a/src/cpp/server/streaming_proxy.cpp
+++ b/src/cpp/server/streaming_proxy.cpp
@@ -86,9 +86,6 @@ void StreamingProxy::forward_sse_stream(
     double time_to_first_token = 0.0;
     const auto start_time = std::chrono::steady_clock::now();
 
-    // On a non-200 the backend body is an error description, not SSE payload:
-    // divert it here instead of the client sink, so it can be reframed as an
-    // SSE event below rather than reaching the client as unparsable bytes.
     int backend_status = 200;
     std::string error_body;
     static constexpr size_t max_error_body = 64 * 1024;
@@ -182,11 +179,9 @@ void StreamingProxy::forward_sse_stream(
                                      << (error_body.empty() ? "" : ": " + error_body) << std::endl;
         telemetry.error_message = "Backend returned error status code: " + std::to_string(status);
 
-        // The response is already committed as 200 text/event-stream, so the only
-        // way to surface the failure is as an SSE event: an unframed body is
-        // dropped by every spec-compliant client parser. Same payload shape as
-        // forward_byte_stream(). No [DONE] follows, matching OpenAI's behavior
-        // for in-stream errors.
+        // The response is already committed as 200 text/event-stream, so an
+        // unframed error body is dropped by every spec-compliant client parser.
+        // No [DONE] follows, matching OpenAI's behavior for in-stream errors.
         json payload;
         try {
             payload = json::parse(error_body);
--- a/src/cpp/server/streaming_proxy.cpp
+++ b/src/cpp/server/streaming_proxy.cpp
@@ -86,6 +86,13 @@ void StreamingProxy::forward_sse_stream(
     double time_to_first_token = 0.0;
     const auto start_time = std::chrono::steady_clock::now();
 
+    // On a non-200 the backend body is an error description, not SSE payload:
+    // divert it here instead of the client sink, so it can be reframed as an
+    // SSE event below rather than reaching the client as unparsable bytes.
+    int backend_status = 200;
+    std::string error_body;
+    static constexpr size_t max_error_body = 64 * 1024;
+
     auto process_line = [&telemetry](const std::string& line) {
         std::string json_str;
         if (line.find("data: ") == 0) {
@@ -104,10 +111,17 @@ void StreamingProxy::forward_sse_stream(
     utils::HttpResponse result = utils::HttpClient::post_stream(
         backend_url,
         request_body,
-        [&sink, &line_buffer, &has_done_marker, &has_first_token,
-         &time_to_first_token, &start_time, &on_chunk, &process_line](const char* data, size_t length) {
+        [&sink, &line_buffer, &has_done_marker, &has_first_token, &time_to_first_token,
+         &start_time, &on_chunk, &process_line, &backend_status, &error_body](const char* data, size_t length) {
             if (on_chunk) {
                 on_chunk();
+            }
+
+            if (backend_status != 200) {
+                if (error_body.size() < max_error_body) {
+                    error_body.append(data, std::min(length, max_error_body - error_body.size()));
+                }
+                return true;
             }
 
             line_buffer.append(data, length);
@@ -132,7 +146,7 @@ void StreamingProxy::forward_sse_stream(
         },
         {},
         timeout_seconds,
-        nullptr,
+        [&backend_status](int status) { backend_status = status; },
         utils::HttpSecurityPolicy::TrustedLoopback
     );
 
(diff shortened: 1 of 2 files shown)


ekenberg:
I reported #2826, and this change fixes it the way I was hoping for. The useful error message was already being sent — it just arrived in a form that clients silently ignore, so users only saw a generic "stream ended" failure with no explanation. With this fix the real reason (for example, the request being too big for the model's context window) becomes visible to the client.

The fix follows the same pattern the other streaming path in this file already uses, so both paths now report failures the same way. And the new test checks the part that was missing before: not just that the stream ends cleanly, but that the error can actually be read. Skipping the final [DONE] after an error also matches what OpenAI does.

Looks good to me — I'd be glad to see this merged.

Yigtwxx:
Thanks for reading it through, @ekenberg — and for the original diagnosis, which is what made this a small change rather than a hunt.

For the maintainers, that resolves the one open question in the description: the issue asked whether `[DONE]` should follow the error event, I left it out to match OpenAI, and the reporter has now confirmed that is the behaviour they wanted. Still a one-line change if you would rather always terminate with `[DONE]`.

Nothing else outstanding on my side. The branch is behind `main` but merges cleanly; happy to rebase if that helps. CI beyond the `label` job has not run yet, since fork PRs need a maintainer to approve the workflow.

fl0rianr (commented):
Nice work, thanks, looks good overall. One small nitty test suggestion: could we make `test_004a_context_overflow_error_is_sse_framed` also assert that the original backend error message is preserved?

Right now the test only checks that an error object with a non-empty message arrives, so it would still pass if we accidentally fell back to something generic like `backend returned HTTP 400`.

Maybe assert for a characteristic part of the context-overflow error, e.g. that the message contains `context`, or check the expected error type as well.

Yigtwxx:
Good catch, @fl0rianr — you are right that the assertion was too weak. `forward_sse_stream()` only synthesizes `{"message": "backend returned HTTP <status>", "type": "backend_error"}` when the body is not already an `{"error": ...}` object, and the old check could not tell those two apart.

Tightened in a2ab9e25 with two assertions rather than one:

```python
self.assertNotEqual(errors[0].get("type"), "backend_error", errors[0])
self.assertIn("context", errors[0].get("message", "").lower(), errors[0])
```

The `type` check is the version-independent half: `backend_error` is only ever produced by the fallback, so it fails on a generic message no matter how llama.cpp words its own. The `context` check is the one you asked for, and it holds against the message llama.cpp actually sends today:

```
request (6009 tokens) exceeds the available context size (4096 tokens), try increasing it
```

Verified locally against `llamacpp:vulkan` with `Tiny-Test-Model-GGUF` — the test passes, and its output prints that message. `black 26.1.0` is clean on the file.

fl0rianr (commented):
Sorry, I missed a points, please remove the obvious comments as defined in our docs as well, thanks.
Appreciating your work!

Yigtwxx:
Done in `8d21299`.

Removed, since they restate what the code shows:
- the block above `backend_status` / `error_body` describing the diversion of a non-200 body
- `Same payload shape as forward_byte_stream().` — a cross-reference that would rot
- in the test, the comment above the two new assertions, and the docstring sentence referencing `test_004`

Kept three lines above the reframing block. Let me know if you'd rather have those gone too — I left them because neither fact is visible from the code: the response is already committed as `200 text/event-stream` (which is *why* an error can only go out as an SSE event), and the absence of a trailing `[DONE]` is a deliberate choice matching OpenAI, not an omission.

Also worth flagging on the way past: your commit `cef16cf` on #2974 has one line that the pinned `black` 26.1.0 in `.pre-commit-config.yaml` would reformat (`self.assertEqual(error_data["error"].get("type"), "not_supported_error")` is 92 chars). Nothing in CI checks it, so it won't block the merge — happy to push the reformat if you want the branch pre-commit-clean.

fl0rianr (approved):
Thanks, LGTM!

Yeah, in PR 2974 do the push to make it black clean, thanks for catching.

PR #2320, opened 2026-06-19
refactor(backends): self-describing WrappedServer backends (#2287)

jeremyfowers:
Implements the plan in #2287: each inference backend **describes itself** with a plain-data descriptor + a server class + a stateless behavior object, and every scattered `if (recipe == "...")` site is rewritten to read a registry built from those descriptors. Backend-specific logic no longer leaks into the router, model manager, system-info, CLI, or docs.

## Layout — a backend is a folder

```
include/lemon/backends/<stem>/<stem>.h          # descriptor (header-only inline const, CLI-safe)
include/lemon/backends/<stem>/<stem>_server.h   # WrappedServer subclass + create()/spec()/ops() decls
server/backends/<stem>/<stem>_server.cpp        # server class impl, BackendOps subclass, create/spec/ops
                                                # (+ any backend-private helpers, e.g. llamacpp_gguf.cpp)
```

**Adding a backend = one `LEMON_BACKENDS` line in `CMakeLists.txt` + that folder + a `backend_versions.json` pin + `server_models.json` entries.** No router, CLI, doc, or support-matrix edits — those are all derived. CMake globs each backend folder (`CONFIGURE_DEPENDS`), so backend-private helper files need no build edit.

## What changed

- **Descriptor** (`backend_descriptor.h`) — plain data describing *what a backend is*: recipe, display name, binary, config section, default device, `SlotPolicy`, `selectable_backend`, `uses_ctx_size`, `dynamic_models`, declarative `options[]`, OS/GPU `support[]`, default labels, required checkpoints, plus editorial/policy fields (`modality`, `experimental`, `web_priority`, `rocm_channels`, `version_policy`, `exposes_prometheus_metrics`, `rocm_requires_cwsr_fix`, `self_manages_downloads`).

- **Two-tier registry, generated from `LEMON_BACKENDS` at CMake configure time** — a **CLI-safe data registry** (descriptors only; links into both `lemonade` and `lemond`) and a **server-only factory registry** (binds each descriptor to its class's `create()`, `spec()`, `ops()`). This split lets the CLI read recipe options/flags from descriptors without linking server classes.

[...]

PR #1810, opened 2026-05-06
Rev FLM to 0.9.42 and fix transcription labeling system 

jeremyfowers:
Closes #1865 

## Summary

- Update the FastFlowLM (`flm`) NPU backend pin from `v0.9.39` to `v0.9.42`.
- Fixes the labeling system for ASR models to allow for consistent handling in the backend. See the updated documentation for the new label definitions.
- Also fixes a GUI bug that would cause the LLM Chat panel to incorrectly get changed to Transcriber for omni-to-text models.

New labels:
```
audio: purely cosmetic
transcription: serves /audio/transcriptions
realtime-transcription: serves ws://realtime
chat-transcription: serves /chat/completions and allows audio data as input
```

See #1887 for the release note comment we will make on the breaking change to the label.

PR #2976, opened 2026-08-08
test: run the committed model-type classifier test in CI

Yigtwxx:
`test/cpp/test_model_type_classifier.cpp` and `test/cpp/test_ggml_hip_path.cpp` are committed but referenced by nothing — neither `CMakeLists.txt` nor any workflow builds or runs them. The testing guide lists exactly this under "What Reviewers Reject".

One of the two shows why the rule matters.

## `test_model_type_classifier.cpp` — works, was never registered

Added with #1810. Covers `get_model_type_from_labels()`, the label-to-deployment-mode mapping that decides whether a model is served as an LLM, ASR, embedding, reranking, image or TTS model. It is called from `model_manager.cpp`, `ollama_api.cpp`, `cli/main.cpp` and `fastflowlm_models.cpp`, and its most valuable case is the one #1810 fixed: a multimodal model labelled `{vision, reasoning, tool-calling, transcription}` must type as LLM, not TRANSCRIPTION.

All 17 cases pass against the current implementation. `model_types.h` is header-only, so the test links nothing beyond the standard library. Registered `CI ON`, following `LatestVersionFallbackTest`.

## `test_ggml_hip_path.cpp` — has not compiled since June

Added with #2044 against `lemon::utils::is_ggml_hip_plugin_available()`, declared in `lemon/utils/path_utils.h`. #2320 moved that function into `lemon::backends::llamacpp` inside `llamacpp_server.cpp` and dropped the header declaration. The test still includes `path_utils.h` and still names `lemon::utils::is_ggml_hip_plugin_available`, so it has been uncompilable for six weeks — and nothing noticed, because nothing builds it.

Restoring the coverage would mean exporting the function from the llamacpp backend so a test can reach it, which the testing guide also calls out. Removing the file seemed like the smaller and more honest change, but this is your call: if you would rather keep the `LEMONADE_GGML_HIP_PATH` coverage, say so and I will send the export instead.

## Also

`docs/dev/testing.md` still tells contributors to register C++ tests with `register_cpp_ci_test()`. #2877 made `add_cpp_ci_test()` the entry point and `AGENTS.md` now says not to call `register_cpp_ci_test()` directly, so the guide points new contributors at the wrong helper. Three occurrences.

## Testing

Windows, MSVC 14.44.

- `cmake --build build --config Release --target cpp-ci-tests` then `ctest --test-dir build -C Release -L "^cpp-ci$"`: 23/23 passed, up from 22/22 on the pre-change baseline. `ctest -R ModelTypeClassifierTest -V` shows the new test running under the label and reporting `17/17 cases passed`.
- Also built and ran standalone with `cl /std:c++17 /EHsc /I src/cpp/include`: 17/17.
- Not run on Linux or macOS; the test is header-only and platform-independent.


Yigtwxx:
﻿Rebased onto `main` after #2950, which rewrote the same row of the "What Reviewers Reject" table. I kept your new wording for the label row and applied the `add_cpp_ci_test()` rename on top; #2950 left the stale `register_cpp_ci_test()` reference in place, so it is still the only change to that line.

fl0rianr (approved):
Thanks, LGMT.

Yigtwxx:
My pleasure! I'm happy to have contributed to the community.

PR #2980, opened 2026-08-08
fix(cloud): restore public model name in provider responses

### LINKED ISSUE #2964: [Cloud] Responses echo the upstream model id instead of the provider-prefixed public name
## Problem

Responses served by the cloud backend carry the **upstream provider model id** in the `model` field instead of the registered `<provider>.<model>` public name the client asked for.

Evidence (main@30977658):

- `CloudServer::rewrite_model_field` rewrites the outbound request's `model` to `upstream_model_` (`src/cpp/server/backends/cloud/cloud_server.cpp:419-424`; `upstream_model_` is set from the checkpoint at `:310`).
- `post_with_auth` returns the provider body **unchanged** on HTTP 200 (`cloud_server.cpp:449-452`), so `response.model` is whatever the provider echoes (e.g., `accounts/fireworks/models/kimi-k2p6`), not the public name (e.g., `fireworks.kimi-k2p6`).
- The streaming path forwards raw SSE bytes verbatim (`forward_streaming_request`, `cloud_server.cpp:490`; outbound model rewrite at `:536`), so every streamed chunk also carries the upstream id.
- The llamacpp backend already solves exactly this for its own echo problem: `normalize_response_model` sets `response["model"]` back to the request's model (`src/cpp/server/backends/llamacpp/llamacpp_server.cpp:621-626`, applied at `:629`, `:637`, `:664`). No equivalent exists on the cloud path.

Impact: clients that key conversation state, UI display, or logging on `response.model` see a different name than the one they sent — and in a `collection.router` conversation that alternates local and cloud candidates, the `model` field flips between naming conventions turn to turn. Sibling of #2730 (llamacpp echoing the gguf file path — same class of bug, different backend).

## Proposed approach

Mirror the llamacpp normalization on the cloud backend: after a successful provider call, set `response["model"]` to the model name from the **original** (pre-rewrite) request, falling back to the server's registered public name. Streaming: rewrite the `"model"` field inside each SSE `data:` JSON frame in the relay, or — if that is judged too invasive for a first pass — fix non-streaming now and document the streaming exception in `docs/api/`.

## Implementation sketch

1. Add `json CloudServer::normalize_response_model(json response, const json& original_request) const` to `src/cpp/server/backends/cloud/cloud_server.cpp` (declare in `src/cpp/include/lemon/backends/cloud/cloud_server.h`), same shape as `llamacpp_server.cpp:621-626`: if `response` is an object containing `model`, set it to `original_request.value("model", get_model_name())`.
2. `CloudServer::chat_completion` / `completion` (`cloud_server.cpp:474-482`): capture the original request before `rewrite_model_field`, wrap the `post_with_auth` result. Only rewrite on success payloads (skip `ErrorResponse` objects).
3. Streaming (`forward_streaming_request`, `:490`): in the chunk relay, for frames that parse as JSON objects with a `model` key, replace the value before writing to the sink; pass non-JSON keep-alives/`[DONE]` through untouched. If implemented, apply the same in the error-envelope path. If deferred, add the documented exception instead.
4. Confirm no double-normalization on the router path: `Router::chat_completion` does not touch `model` on the way out, so the backend-level fix is the single point of truth.

## Acceptance criteria

- `test/server_router.py` (mock cloud provider, already used for the cloud-candidate e2e): non-streaming chat completion through a cloud model returns `response["model"] == "<provider>.<model>"` even when the mock echoes a different upstream id.
- Same assertion for `/v1/completions`.
- If streaming rewrite is implemented: first SSE chunk's `model` equals the public name; `[DONE]` sentinel untouched.
- Error responses (non-200 from provider) are not modified by the normalization.

## Out of scope

- llamacpp's own model-echo variants (#2730).
- Rewriting `usage` or any other provider field — `model` only.
- Ollama/Anthropic compatibility layers (they re-shape responses separately).

## References

- Related: #2730 (same bug class on llamacpp), #2389 (router tracker).

Size: S

jenbrewerworks:
## Summary
Cloud backend responses echoed the **upstream provider model id** in `model` after `rewrite_model_field` rewrote the outbound request. Clients that key state on `response.model` saw a different name than they sent (and mixed naming in router conversations).

The llamacpp backend already restores the request model via `normalize_response_model`; cloud had no equivalent.

## Changes
- Add `CloudServer::normalize_response_model` (same idea as llamacpp) and apply it to non-streaming `chat_completion` / `completion`.
- On the SSE path, rewrite `"model"` inside each `data:` JSON frame back to the public name from the original request.

## Test plan
- [ ] Non-streaming chat/completions: `response.model` equals the requested public name (e.g. `fireworks.kimi-k2p6`), not the upstream id
- [ ] Streaming: intermediate and final chunks keep the public `model` field
- [ ] Existing cloud auth / insecure-http error paths unchanged

Fixes #2964

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ if(MSVC)
     set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 endif()
 
-project(lemon_cpp VERSION 11.5.2)
+project(lemon_cpp VERSION 11.6.0)
 
 include(CTest)
 
@@ -45,6 +45,16 @@ endfunction()
 #               dependency of the cpp-ci-tests aggregate target.
 #   * CI OFF -> the test is created for local/other runs but excluded from CI.
 #
+# The enclosing if() MUST test BUILD_TESTING. Distro packaging configures with
+# BUILD_TESTING=OFF so it does not build ~40 test binaries it then discards, and
+# calling this helper in that configuration is a fatal error rather than a
+# silent return to the slow build:
+#
+#   if(BUILD_TESTING AND EXISTS "${_FOO_TEST_SRC}")
+#       add_executable(test_foo ...)
+#       add_cpp_ci_test(FooTest CI ON COMMAND test_foo)
+#   endif()
+#
 # Usage:
 #   add_cpp_ci_test(<TestName>
 #                   CI <ON|OFF>                 # required, explicit CI decision
@@ -72,6 +82,14 @@ function(add_cpp_ci_test test_name)
     if(NOT ACT_COMMAND)
         message(FATAL_ERROR
             "add_cpp_ci_test(${test_name}): required 'COMMAND <command>' argument is missing")
+    endif()
+    # Distro packaging configures with BUILD_TESTING=OFF and never runs or ships
+    # a test, so reaching here means the enclosing if() forgot to check it —
+    # which silently puts the whole test suite back into every package build.
+    if(NOT BUILD_TESTING)
+        message(FATAL_ERROR
+            "add_cpp_ci_test(${test_name}): declared with BUILD_TESTING=OFF. Add "
+            "BUILD_TESTING to the enclosing if() guard, as the other test blocks do.")
     endif()
 
     # _add_test is the built-in add_test preserved by the override below, so this
@@ -143,6 +161,7 @@ set(LEMON_BACKENDS
     "flm|fastflowlm"
     "ryzenai-llm|ryzenai"
     "vllm|vllm"
+    "thenoise|thenoise"
     "cloud|cloud"
     "thinksound|thinksound"
     "acestep|acestep"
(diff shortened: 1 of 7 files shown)


fl0rianr (commented):
Thank you for contributing! Nice fix overall. The non-streaming normalization looks straightforward and consistent with the existing llama.cpp behavior.

One streaming edge case looks worth fixing before merge: `process_sse_lines()` only emits newline-terminated lines, and the remaining `sse_line_buffer` is never flushed after `post_stream()` finishes. That means a provider response whose final SSE line has no trailing newline can be dropped.

In particular, if that line is `data: [DONE]`, `has_done_marker` is already set from the raw chunk, but the marker itself stays buffered, so the fallback below won't emit one either.

Could we flush any remaining buffered line before completing the stream, applying the same model rewrite?

It would also be good to add a tiny regression test for the streaming path, ideally covering chunks split across reads and a final frame without a trailing newline. Otherwise this looks good to me.

jenbrewerworks:
Thanks @fl0rianr — addressed the streaming edge case:

- Added `StreamingProxy::flush_sse_line_buffer` and call it after `post_stream()` on the cloud SSE path, so a final frame without a trailing newline (including `data: [DONE]`) is rewritten and emitted.
- Added `test/cpp/test_process_sse_lines.cpp` covering split chunks and a final frame without `\n`.

Ready for another look.

fl0rianr (commented):
Thanks for the update! The approach looks good overall, and the missing final SSE-line flush is fixed nicely.

I noticed two remaining blockers and one test gap before merge:

* The new C++ test should use `if(BUILD_TESTING AND EXISTS(...))` to match current `main`. Without the `BUILD_TESTING` guard, packaging builds with tests disabled can fail.
* `has_done_marker` is still detected on raw network chunks. If `[DONE]` is split across chunks, e.g. `data: [DO` + `NE]\n`, the real `[DONE]` is forwarded after reassembly but `has_done_marker` stays false, so a second synthetic `[DONE]` is emitted. It would be safer to set the flag from the reassembled SSE line.
* The added test covers SSE buffering/flush, but not the actual model-name normalization. A small regression test for non-streaming and streaming `model == public_model` would make this fix much safer.

After those, this looks good to me. Thanks!

jenbrewerworks:
Rebased onto latest main (merged CMakeLists + streaming_proxy.cpp). SSE flush helper and cloud path changes are preserved.

ramkrishna2910:
@jenbrewerworks any reason this PR was closed?

PR #2982, opened 2026-08-08
fix(gui): MVP, remove non ready advanced filter

### LINKED ISSUE #2936: Remove the filters control from the models middle panel
Image

There should not be a filters control in the Models list since there's a whole filters panel to the left.

fl0rianr:
Closes #2936 

image

Other filters still work fine.


jeremyfowers (approved):
Thanks!

PR #2989, opened 2026-08-08
ci: cut installer build time on the PR/merge-queue critical path

jeremyfowers:
The critical path on both PR pushes (~41 min) and the merge queue (~49 min) was `Build Lemonade Server Installer` and `Build Lemonade .deb Package` — every test job waits on one of them. Five changes, each measured against baseline run `31266332304`.

## Measured

**End-to-end PR wall-clock: 41 min → 26.9 min** (run `31274940736`, an unlabeled PR run — the apples-to-apples comparison). Final all-green verification run: `31294288426`.

| Job | Before | After |
|---|---|---|
| Build Lemonade Server Installer | 23–27 min | **10.3 min** |
| Build Lemonade .deb Package | 17 min | **7.0 min** |
| Build Embeddable Lemonade (Windows) | 13.5 min | **8.8 min** |
| Fedora RPM (×4 legs) | 11–12 min | **6.9–10.5 min** |
| Debian 13 (×2 legs) | 7–8 min | **5.0–6.2 min** |
| C++ unit tests | inside the .deb job | 5.2 min, parallel |
| Compile-check the Tauri host | — | 2.2 / 2.7 min, new coverage |
| Build Lemonade Desktop Installer | inside the installer job | 18.3 min, merge-queue only |

## What was wrong

**MSVC had no `/MP`.** Within each `.vcxproj`, `cl.exe` compiled one file at a time, so the Windows C++ phase took 12.2 min on a 4-core runner. Enabled for Visual Studio generators only — Ninja already schedules per file and would oversubscribe.

**MSBuild built independent projects serially.** `lemonade-server-core` takes 5.2 min, and the CLI and `LemonadeServer`, which do not depend on it, waited for it to finish. `/m` is passed straight through rather than via `--parallel`, which would make CMake set `CL_MPCount=1` and cancel the `/MP` win.

**The Tauri desktop app was 8.4 min of the installer build** (7m18s of it `cargo build --release`) and only lands in `lemonade.msi`. Every test job installs `lemonade-server-minimal.msi`. Moved to a packaging-only `Build Lemonade Desktop Installer` job gated behind the merge queue like the other packaging variants, wired into the existing `Packaging builds` gate. Tag and signing runs still build both MSIs in one job, so SignPath and release are untouched. `ci:distros` opts a PR back in, the same label that already covers the `Packaging builds` gate it reports through.

**The `.deb` build built ~115 unit-test binaries and discarded them** — 116 of 224 ninja edges, ~6 of its 11.3 min — then the next step rebuilt them from scratch. `override_dh_auto_test` is already a no-op and no test binary is packaged, so the Debian configure sets `BUILD_TESTING=OFF`. The test blocks guarded only by `EXISTS` now also honour `BUILD_TESTING`, matching the ones that already did.

**The RPM job built the same discarded test binaries.** `setup.sh` configures with testing on and the build has no `--target`, so all 43 test executables were built and then left out of the `.rpm`, across a four-leg matrix. It now reconfigures with `BUILD_TESTING=OFF` first. The generated install manifest is byte-identical either way, so nothing about the shipped package changes.

**The unit tests shared nothing with the packaging build** they lived in — separate configure, different flags — so their 4.5 min sat on the critical path for no reason. Now their own parallel job, still in this workflow.

## Compiler cache

`.ubuntu/Dockerfile` primes a ccache into the build-environment image that `build-container.yml` already rebuilds nightly, so the `.deb` build starts warm. Actions caches cannot serve this job: they are scoped per git ref, so merge-queue runs — which every merge passes through — can never restore one.

- Primed at the container checkout path, because the absolute `-I` paths reach the compiler command line and ccache hashes them.
- `DEB_BUILD_DEBUGPATH` pins `-fdebug-prefix-map`, which dpkg otherwise derives from the changelog version — that changes on every commit and would miss every lookup. Release builds keep dpkg's version-derived path and skip the cache, so shipped `-dbgsym` packages are unchanged.
- amd64 only: nothing consumes the arm64 image, and priming it would run the build under QEMU.
- `build-container.yml`, which already publishes this image, now also builds it on PRs that change how it is built, and asserts the cache is non-empty on both paths. It previously ran on a nightly schedule only, so a change here was first exercised at 2 AM against the tag every Linux CI job and every PPA job pulls.

## Notes

- The primed image builds and verifies green (305 MB of cache, under the 1 GB cap). The speedup itself does not appear in this PR's runs — CI pulls the currently published image, which has no ccache in it. The opt-in step tolerates that, so nothing breaks in the meantime; the win lands after the next nightly `build-container` run.
- `Test CLI/Endpoints (ubuntu-latest)` at 18.1 min is now the critical path on its own: it starts once the 6.7-min .deb build finishes, which accounts for essentially the whole 26.9-min run.
- **One check would benefit from a branch-protection entry:** `Build build-environment (ubuntu24.04)`. It reports a broken ccache prime on the PR, but it lives in `build-container.yml` so it cannot join a gate job — gates use `needs:`, which cannot cross workflows, and moving it into the packaging workflow would give up the narrow `paths:` filter that keeps it off every C++ PR. Everything else added here binds through an existing gate.
- `C++ unit tests` is in `backends-gate`'s `needs`, so it keeps blocking merges through the already-required `Inference backend tests` check rather than needing a new branch-protection entry.
- **The measured numbers are the PR path.** In the merge queue `Build Lemonade Desktop Installer` also runs (~18.5 min), rebuilding the Windows C++ and web-app targets that `Build Lemonade Server Installer` builds concurrently. It runs in parallel and is well short of the queue's critical path, but it does occupy one more concurrent slot.
- **Rust coverage on the PR path is preserved by a new canary.** Deferring the Windows desktop installer would otherwise have left *no* PR-path job compiling `src/app/src-tauri` at all, since the macOS `.dmg` job is `ci:macos`-gated and the Debian rules never set `BUILD_TAURI_APP` — so a broken Rust host would only surface once a PR was already queued. A new `Compile-check the Tauri host` job runs `cargo check` on **both Linux and Windows**, reporting through the existing `Packaging builds` gate so it blocks without needing a new branch-protection entry — Windows is not optional, since `webview_shim.rs` and `lib.rs` gate code behind `cfg(target_os = "windows")` and `webview2-com`/`windows` are Windows-only dependencies, so a Linux-only check would miss exactly the code whose packaging job was deferred. It is a compile canary rather than a packaging build (~1.5 min, uncached). Full Tauri packaging still defers to the merge queue, the same deal the packaging, distro and macOS jobs already make.
- `Lemonade_Server_MSI` now carries only `lemonade-server-minimal.msi` off the release path; the desktop MSI is published separately as `Lemonade_Desktop_MSI`. No in-repo consumer reads `lemonade.msi` from the former — `install-lemonade-server-msi` uses only the minimal MSI and `release` uses the signed artifact — but external automation that did would need repointing.


fl0rianr (commented):
Looks good!

**One thing to call out explicitly**: I don’t think `Build build-environment (ubuntu24.04)` should be added as a required check in its current form.

The workflow only runs for `pull_request` when the narrow `paths:` filter matches, and it does not have a `merge_group` trigger. If this exact check is made required, PRs where the workflow is skipped can be left waiting for a required status that never appears, and the merge queue would not produce this check either.

I’d keep this check advisory for now. If we want it to become required later, I think it should first be wired through an always-reporting gate that runs for both PRs and `merge_group`, while keeping the expensive image build path-filtered.

jeremyfowers:
> One thing to call out explicitly: I don’t think Build build-environment (ubuntu24.04) should be added as a required check in its current form.

Sure thing, I wasn't planning on it

fl0rianr (approved):
LGTM

PR #2995, opened 2026-08-08
fix: add GITHUB_TOKEN / GH_TOKEN support to GitHub API requests

### LINKED ISSUE #2993: Support GITHUB_TOKEN / GH_TOKEN env vars for GitHub API rate limits
### Feature Description

Read `GITHUB_TOKEN` (fallback `GH_TOKEN`) from the environment and inject `Authorization: Bearer` into requests to `api.github.com`, raising the rate limit from 60 to 5,000 req/h.

### Use Case / Motivation

Lemonade issues unauthenticated GitHub API calls in three code paths:

- `BackendManager::fetch_latest_github_tag()` — scans every repo on startup for "latest" version pins
- `resolve_asset_wildcard()` — resolves wildcard asset names (sd-cpp metal, and now sd-cpp ROCm) per install
- `recipe_import.cpp` — fetches recipe contents

Each unauthenticated IP is capped at 60 requests/hour. On a typical startup with multiple repos and subsequent backend installs, this limit is easily reached, causing 403 responses that block backend installation and model downloads.

Setting `GITHUB_TOKEN` is a common convention (used by `gh` CLI, GitHub Actions, and many tools). It raises the limit to 5,000 req/h — more than sufficient for normal use.

### Platform Relevance

All platforms

blackdeathdrow:
Unauthenticated GitHub API calls are capped at 60 req/h. BackendManager scans multiple repos on startup, `resolve_asset_wildcard` queries per-install, and `recipe_import` hits the API — sometimes exhausting the limit under normal use, causing 403s.

Add `lemon/utils/github_api.h` with `github_api::headers()` (reads `GITHUB_TOKEN`, falling back to `GH_TOKEN`, and injects `Authorization: Bearer …` when set) and `github_api::get()`, a thin wrapper that retries once without auth on `401`,`403` or `404` to handle scoped GitHub Actions tokens rejected by unrelated repos. Applied in `fetch_latest_github_tag()`, `resolve_asset_wildcard()`, and `recipe_import.cpp` — the three code paths that issue HTTP GET to `api.github.com`.

Adds `test/cpp/test_github_api.cpp` covering the token-priority / fallback / empty-value cases.

Closes #2993

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2810,7 +2810,7 @@ set(_GITHUB_API_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_github_api.c
 if(BUILD_TESTING AND EXISTS "${_GITHUB_API_TEST_SRC}")
     add_executable(test_github_api test/cpp/test_github_api.cpp)
     target_link_libraries(test_github_api PRIVATE lemonade-server-core)
-    add_cpp_ci_test(GithubApiTest CI OFF COMMAND test_github_api)
+    add_cpp_ci_test(GithubApiTest CI ON COMMAND test_github_api)
 endif()
 
 # GGUF variant enumeration: grouping, name disambiguation, and quant ordering.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2806,6 +2806,13 @@ if(BUILD_TESTING AND EXISTS "${_URL_UTILS_TEST_SRC}")
     add_cpp_ci_test(UrlUtilsTest CI OFF COMMAND test_url_utils)
 endif()
 
+set(_GITHUB_API_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_github_api.cpp")
+if(BUILD_TESTING AND EXISTS "${_GITHUB_API_TEST_SRC}")
+    add_executable(test_github_api test/cpp/test_github_api.cpp)
+    target_link_libraries(test_github_api PRIVATE lemonade-server-core)
+    add_cpp_ci_test(GithubApiTest CI OFF COMMAND test_github_api)
+endif()
+
 # GGUF variant enumeration: grouping, name disambiguation, and quant ordering.
 set(_HF_VARIANTS_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_hf_variants.cpp")
 if(BUILD_TESTING AND EXISTS "${_HF_VARIANTS_TEST_SRC}")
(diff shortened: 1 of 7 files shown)


fl0rianr (commented):
Thanks for adding this, the overall approach makes sense.

A couple of edge cases might be worth addressing before merge:

* If `GITHUB_TOKEN` is set but empty, the current fallback never checks `GH_TOKEN`. Checking for both `nullptr` and an empty value would make the fallback behave as described.
* In GitHub Actions, `GITHUB_TOKEN` can be scoped to the current repository. Adding it to requests for other public repos may turn an otherwise valid unauthenticated request into a 403/404. It might be safer to retry without auth when the authenticated request is rejected.
* The token/header logic is currently duplicated in `recipe_import.cpp` and `github_api_headers()`. Reusing one helper would avoid the two paths drifting apart.

Also, the comment about the rate limit being 5,000 requests/hour per repository is a bit too specific since that depends on the token type.

Otherwise this looks like a useful improvement. Thanks!

blackdeathdrow:
Thanks for the review! All points addressed in the latest push:

1. Empty `GITHUB_TOKEN` fallback — `headers()` now checks for both `nullptr` and an empty string before falling back. Covered by test case 4 in `test_github_api.cpp`.
2. Scoped `GITHUB_TOKEN` retry — added `github_api::get()`, which retries once without the `Authorization` header when the authenticated request returns `403`. I scoped the fallback to `403` only rather than `403/404`: a `404` almost always means the resource genuinely doesn't exist, so an unauthenticated retry would just double the request load and return `404` again. If you have a concrete case where a scoped token yields `404` on a public repo, happy to widen it.
3. Deduplication — the helper now lives in `src/cpp/include/lemon/utils/github_api.h`; `recipe_import.cpp` calls `github_api::get()` like the other two paths, so there's a single implementation.

fl0rianr (commented):
Thanks for the update, this is much cleaner now and the original duplication / empty-token issues look addressed.

I still see a couple of things worth fixing before merge:

* `GithubApiTest` is currently registered with `CI OFF`. The project testing guide explicitly requires new C++ regression tests to run in CI, so this should be `CI ON`.
* The unauthenticated retry currently only happens on `403`. GitHub can also return `401` or `404` for missing/insufficient token permissions, including scoped `GITHUB_TOKEN` access to another repo. Retrying without auth for those cases would fully cover the original issue.
* Minor: the “empty `GITHUB_TOKEN`” test actually unsets the variable, so it doesn’t exercise the `token != nullptr && token[0] == '\0'` path. It would be good to set it explicitly to an empty string.

Other than that, the structure looks good now. Thanks for addressing the earlier feedback!

blackdeathdrow:
Done

fl0rianr (approved):
Thanks, LGTM!

Please update the PR description before merging.

PR #2997, opened 2026-08-08
fix(sdcpp): wildcard ROCm runtime in asset only for latest bin pins

### LINKED ISSUE #2988: sd-cpp:rocm fails to download — hardcoded ROCm version in asset name 404s when upstream bumps runtime
### Platform

Windows / Linux

### Component

sd-cpp (stable-diffusion.cpp)

### Bug Description

sd-cpp:rocm hardcodes ROCm runtime version from `backend_versions.json` (`therock.version`, currently 7.13.0) into the download asset filename. But `leejet/stable-diffusion.cpp` upstream builds against its own ROCm runtime, which drifts independently. When upstream bumps from e.g. ROCm 7.13.0 to 7.14.0, lemonade constructs `sd-master-<hash>-bin-win-rocm-7.13.0-x64.zip` which 404s, while the actual published asset is `...-rocm-7.14.0-x64.zip`.

Error example:
`Failed to download sd-server.exe from: https://github.com/leejet/stable-diffusion.cpp/releases/download/master-813-bfbef5b/sd-master-bfbef5b-bin-win-rocm-7.13.0-x64.zip - HTTP error 404`

### Root Cause

`SDServer::get_install_params()` (sdcpp_server.cpp) calls `get_therock_version()` to embed the ROCm version in the asset filename. But the sd.cpp upstream release may be built with a different ROCm version.

The metal backend already avoids this class of bug by using a wildcard (`*-bin-Darwin-macOS-*-arm64.zip`) resolved via `resolve_asset_wildcard()` -> GitHub Releases API at install time.

### Fix

- sdcpp ROCm assets switched to wildcard: `*-bin-win-rocm-*-x64.zip` / `*-...-rocm-*.zip`, matching the metal approach
- `resolve_asset_wildcard()` enhanced to pick the numerically newest version when several assets match (sd.cpp publishes per-runtime builds, e.g. `-rocm-7.1.1` and `-rocm-7.13.0` per release)
- Removed unused `get_therock_version()` / `trim_*` helpers from sdcpp_server.cpp

### Related

- #2800 — same class of bug for llamacpp:rocm (fixed separately via therock.version bump)

blackdeathdrow:
sd.cpp ROCm release assets embed the ROCm runtime they were built against (e.g. `-rocm-7.14.0-`), which drifts independently of the `therock.version` pin that determines which runtime Lemonade installs and loads next to the binary. Under `sd-cpp.rocm_stable` pinned by Lemonade this is fine, but with an explicit `sdcpp.rocm_bin=latest` pin the newest upstream release no longer publishes an asset matching `therock.version` and the install fails with HTTP 404.

So the wildcard is now scoped to that opt-in: `sdcpp.rocm_bin == "latest"` names the runtime as `*` and `resolve_asset_wildcard()` picks the numerically newest matching asset via the GitHub Releases API. Anything else (builtin pins, explicit release tags) names `therock.version` exactly.

`resolve_asset_wildcard()` itself now collects all matches and prefers the numerically newest version (sd.cpp publishes per-runtime builds such as `-rocm-7.1.1` alongside `-rocm-7.13.0`); numeric comparison is also the tiebreak the multi-asset Darwin metal names already need.

Closes #2988

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2445,6 +2445,26 @@ if(BUILD_TESTING AND EXISTS "${_LATEST_FALLBACK_TEST_SRC}")
     add_cpp_ci_test(LatestVersionFallbackTest CI ON COMMAND test_latest_version_fallback)
 endif()
 
+# sd.cpp ROCm release-asset naming policy (lemon::backends::sdcpp::
+# sdcpp_rocm_asset_pattern). Header-only pure logic, so the test links nothing
+# beyond the standard library — coverage for #2988 (wildcard the runtime only
+# for an explicit "latest" bin pin; pinned installs name the TheRock runtime
+# exactly).
+set(_SDCPP_ROCM_ASSET_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_sdcpp_rocm_asset.cpp"
+)
+if(BUILD_TESTING AND EXISTS "${_SDCPP_ROCM_ASSET_TEST_SRC}")
+    add_executable(test_sdcpp_rocm_asset
+        test/cpp/test_sdcpp_rocm_asset.cpp
+    )
+    target_include_directories(test_sdcpp_rocm_asset PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+    )
+
+    include(CTest)
+    add_cpp_ci_test(SdcppRocmAssetTest CI ON COMMAND test_sdcpp_rocm_asset)
+endif()
+
 # Label-to-deployment-mode mapping (lemon::get_model_type_from_labels). Header-only,
 # so the test links nothing beyond the standard library.
 set(_MODEL_TYPE_CLASSIFIER_TEST_SRC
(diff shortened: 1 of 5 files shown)


blackdeathdrow:
I encountered this bug right after setting "rocm_bin": "latest" in my configuration. This immediately exhausted the unauthenticated rate limits, which is why #2995 was also required to support GITHUB_TOKEN and raise the cap.

superm1:
This is mostly my fault because https://github.com/lemonade-sdk/lemonade/pull/2768 didn't merge quickly.  But I _think_ that's blocked by a CI infra issue.

superm1 (commented):
The problem I see with this is we could have breakage if upstream decides to have two CI targets with an incompatible ABI.  For example if ROCm 10.0 broke ABI then a 7.14 artifact wouldn't work right.  So I'm not sure this is the right move right now.

blackdeathdrow:
Fair point — I've rescoped the wildcard so it only applies under an explicit `rocm_bin=latest` pin. Pinned installs (builtin or an exact tag) once again name `therock.version` verbatim, so the downloaded binary always matches the runtime Lemonade installs and loads alongside it, and the worst case on the default path is a clear 404 rather than an ABI-mismatched binary.

blackdeathdrow:
If this approach still doesn't feel right, no worries — feel free to close the PR.

jeremyfowers:
@superm1 could you comment if this is something you want to eventually merge, or not?

PR #3003, opened 2026-08-09
fix(gui): limit checkbox visually to highlighting

### LINKED ISSUE #2999: Invert the checkbox colors
Image

Right now a checked checkbox, especially in the dark theme, uses too much of the highlight color (which draws the eyes too much to something that is typically non-critical). Flip it so that the checkmark is yellow and the backdrop is black.

fl0rianr:
This is my proposal, pure inverting was not convincing for me,

image

image

closes #2999

### DIFF
--- a/src/app/src/styles/styles.css
+++ b/src/app/src/styles/styles.css
@@ -38,6 +38,54 @@ input, textarea {
   border: 0;
 }
 
+/* ---------- Checkbox ---------- */
+input[type="checkbox"] {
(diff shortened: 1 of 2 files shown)


fl0rianr:
More satisfied with this one. Just the lemonade yellow checkmark is too weak on light-theme compared to the darker check above. This gives back more identity as well.

image
image


image
image

jeremyfowers (approved):
Thanks, this look like a big improvement to me!

Issue #2089, opened 2026-06-03
Feature Request: integrate rocmfp4-llama (ROCmFP4 quantized llama.cpp) as a backend option for AMD Strix Halo


PatrikScully:
### Feature Description

Add support for the ROCmFP4 llama.cpp fork (https://github.com/charlie12345/rocmfp4-llama) as a selectable llama.cpp backend or quantization path in Lemonade. The fork introduces two custom 4-bit weight formats, `Q4_0_ROCMFP4` (dual-scale, 4.50 bits per weight) and `Q4_0_ROCMFP4_FAST` (single-scale, 4.25 BPW), built on a Codebook10 4-bit value table with unsigned E4M3 half-scale semantics. It ships ROCm/HIP kernels for vector ops, dequantization, and FlashAttention, plus Vulkan shaders for decode and MMQ, and tensor-aware presets that mix ROCmFP4 with higher-precision tensors to preserve quality.

The goal is to let Lemonade users load and run ROCmFP4 models through the same model manager and OpenAI-compatible server they already use, without manually building and maintaining a separate llama.cpp fork.

### Use Case / Motivation

Lemonade is already a common entry point for running local models on AMD Strix Halo (Ryzen AI Max+ 395, 128GB unified memory). On that hardware the ROCmFP4 fork has shown strong results: roughly 30 to 40 tokens/second on a 27B model with MTP, and HumanEval+ scores reported above a vanilla 27B Q8 baseline, at a lower memory footprint than Q8. Larger Qwen variants have been reported in the 89 to 104 tokens/second range in sustained decode.

Today, getting these numbers means cloning and compiling the fork by hand and wiring it up outside of Lemonade. Integrating it would give Strix Halo users a quality-per-bit and throughput option that fits in unified memory, while keeping the simple model switching and management that Lemonade is valued for. It also fills a current gap: there are very few ready-to-use ROCmFP4 models or tooling paths available.

### Platform Relevance

All platforms (primary target is AMD ROCm on Strix Halo; the fork includes both ROCm/HIP and cross-platform Vulkan paths).

### Additional Context

- Source project: https://github.com/charlie12345/rocmfp4-llama
- The project is described by its author as an experimental research build, and results are noted to be hardware, driver, model, and prompt sensitive. Validated on a Framework Desktop with a 395+ and 128GB unified RAM, tested with Qwen 27B and 35B variants.
[...]

PR #3004, opened 2026-08-09
fix: honor custom backend binary environment variables

hogeheer499-commits:
## Summary
- resolve documented `LEMONADE_<SECTION>_<BIN_KEY>` overrides before `config.json`
- preserve `config.json` as the fallback
- cover both ROCm channels, which share `LEMONADE_LLAMACPP_ROCM_BIN`

This addresses the env-override failure reported in #2089 without adding fork-specific integration.

## Validation
- `cmake --build build --target test_backend_utils_gpu_env -j4`
- `./build/test_backend_utils_gpu_env`
- `git diff --check`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2716,7 +2716,7 @@ set(_GPU_ENV_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_backend_utils_g
 if(EXISTS "${_GPU_ENV_TEST_SRC}")
     add_executable(test_backend_utils_gpu_env test/cpp/test_backend_utils_gpu_env.cpp)
     target_link_libraries(test_backend_utils_gpu_env PRIVATE lemonade-server-core)
-    add_cpp_ci_test(BackendUtilsGpuEnvTest CI OFF COMMAND test_backend_utils_gpu_env)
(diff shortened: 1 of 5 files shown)


fl0rianr (commented):
Thanks, good fix overall, and the ROCm stable/nightly handling looks good.

One thing I think we should address before merging: the new env override is only used by `find_external_backend_binary()`, while the status/version logic still reads the `*_bin` value directly from `config.json`.

That can make the same custom binary get used at runtime but still be reported as `update_required` because `SystemInfo` / `BackendManager` still see e.g. `llamacpp.rocm_bin = "builtin"`.

It would be good to centralize the raw `*_bin` resolution with env taking precedence over config, then reuse that in `find_external_backend_binary()`, `get_bin_config_value()` / status handling, and `resolve_user_version()`. A test with `config = builtin` + env = custom path would cover this nicely.

hogeheer499-commits:
Thanks, this was a good catch. I pushed 136e323e to centralize the raw *_bin lookup with env-over-config precedence and reuse it for both runtime lookup and version/status handling. I also added the config=builtin + env=custom-path regression case. BackendUtilsGpuEnvTest passes locally.

fl0rianr (approved):
Thanks! LGTM.

Quickly fixed an error on main that the test updated for your changes was not wired to CI. Going for merging now.

PR #3007, opened 2026-08-09
ci(test): speed up server job suite

### LINKED ISSUE #3005: server_jobs.py takes 6 minutes in ci
This is a relatively new test suite, so I'm not 100% sure what it's doing. But 6 minutes seems like a lot and its an expense on every PR and merge. Can this be optimized into the 1-2 minute range?

fl0rianr:
## Summary

The 1 - 2 min target was to hard for me tbh, but this is still a merge worth improvment.

* reuse one isolated `lemond` instance and cache across the job test suite
* reduce polling, gate durations, context sizes, and inference work while keeping the explicit restart tests intact
* use a suite-local tiny llama.cpp model and disable warmup for model lifecycle tests
* keep all 35 existing tests

Local warm runs improved from roughly **324s** to **137s**.

## Test model

These tests validate job lifecycle, model ownership, exclusivity, persistence, and cleanup rather than model quality. Using a much smaller llama.cpp model exercises the same load/chat/unload paths while reducing repeated backend load time.

The model is registered only for this suite, so the global test model and other test suites are unchanged.

## Drain test skip

`test_exclusive_job_drains_inflight_chat` skips when the chat finishes before the test can observe an in-flight request. In that case there is no drain window to assert against, so skipping avoids a timing-dependent false failure. This also occurred with the previous test model.

Closes #3005

### DIFF
--- a/test/server_jobs.py
+++ b/test/server_jobs.py
@@ -16,12 +16,29 @@ PORT = 13401
 HOST = "127.0.0.1"
 BASE = f"http://{HOST}:{PORT}/api/v1"
 
-TEST_MODEL = "Tiny-Test-Model-GGUF"
+TEST_MODEL = "Jobs-TinyLlama-15M-GGUF"
+TEST_MODEL_CANONICAL = f"user.{TEST_MODEL}"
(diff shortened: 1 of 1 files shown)


bitgamma (approved):
making CI faster always looks good to me

PR #3013, opened 2026-08-09
GUI3 Fix: improve telemetry optimizer model selection and cancellation

fl0rianr:
## Summary

Fixes two issues in **Monitor → Telemetry → Improve**:

* Selects optimizer models from downloaded tool-calling models instead of arbitrary reasoning models (was an old Qwen3 fixed)
* Prefers the recent used downloaded tool model, then a downloaded **hot** tool model, and falls back to Lemonade's quality default.
* Allows users to override the automatic choice with any downloaded tool-calling model.
* Makes the optimization progress dialog cancellable. Closing it now aborts the active request and immediately releases the UI (long or looping request blocked the UI, breaking)

Modified `ModelSearchSelector` keeps its existing behavior by default; the new show-all-on-focus behavior is opt-in and only used by the optimizer.

## Testing

* Local function test
* Updated Playwright coverage for tool-model selection and request cancellation.
* Existing model navigation and model-state runtime checks pass.

### DIFF
--- a/src/app/src/components/inspect/ImproveTab.tsx
+++ b/src/app/src/components/inspect/ImproveTab.tsx
@@ -7,6 +7,13 @@ import MarkdownMessage from '../MarkdownMessage';
 import Modal from './Modal';
 import { useAutoScroll } from '../../hooks/useAutoScroll';
 import { useCopyToClipboard } from '../../hooks/useCopyToClipboard';
+import { modelCapabilityTags } from '../../modelCapabilities';
+import {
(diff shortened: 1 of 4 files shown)


jeremyfowers:
image

the ui elements here are not vertically aligned, please fix that

jeremyfowers:
Other than that, the Improve tab seems to be working correctly for me now.

fl0rianr:
Issue fixed and filter aligned with model TASK one to save space. Function still correct.
image

abn (approved):

PR #3016, opened 2026-08-09
Allow explicit system llama.cpp backend

### LINKED ISSUE #2536: Choosing devices with system llama.cpp
### Feature Description

I would like to be able to pass `--device` to the system llama.cpp backend, or have another way to select devices on a multi-device system.

### Use Case / Motivation

When selecting a pre-defined backend, it makes sense to block `--device`. However a system llama.cpp can be built with several backends when GGML_BACKEND_DL is on. On AMD this makes it easy to switch between ROCm and Vulkan for instance. On multi-gpu systems, it can even allow unholy alliances of GPUs from different manufacturers.

### Platform Relevance

Linux only

### Additional Context

I am packaging lemonade and llama.cpp for Arch Linux, my llama.cpp package bundles several backends already and will include even more in the future.

```
❯ llama-server --list-devices
Available devices:
  ROCm0: AMD Radeon 8060S Graphics (98304 MiB, 19195 MiB free)
  Vulkan0: AMD Radeon 8060S Graphics (RADV STRIX_HALO) (114163 MiB, 80334 MiB free)
```

On this particular system I can't use Gemma E4B with my system llama.cpp because it fails to load with ROCm, only works with Vulkan, however ROCm is the default and cannot be overriden by passing `--device`.

I tried loading the following recipe, but it still loads ROCm:
```json
"gemma4-e4b": {
    "checkpoints": {
      "draft": "unsloth/gemma-4-E4B-it-qat-GGUF:mtp-gemma-4-E4B-it.gguf",
      "main": "unsloth/gemma-4-E4B-it-qat-GGUF:gemma-4-E4B-it-qat-UD-Q4_K_XL.gguf",
      "mmproj": "unsloth/gemma-4-E4B-it-qat-GGUF:mmproj-BF16.gguf"
    },
    "labels": [
      "custom",
      "mtp",
      "tool-calling",
      "vision"
    ],
    "recipe": "llamacpp",
    "recipe_options": {
      "ctx_size": 262144,
      "llamacpp_args": "--parallel 4 --temp 1.0 --top_p 0.95 --top_k 64 --spec-type draft-mtp --spec-draft-n-max 4 --presence_penalty 0.0 --reasoning off --flash-attn on --cache-type-k q8_0 --cache-type-v q8_0 --spec-draft-type-k q8_0 --spec-draft-type-v q8_0",
      "llamacpp_backend": "system",
      "llamacpp_device": "Vulkan0",
      "merge_args": true,
      "pinned": true
    },
    "size": 4.91,
    "suggested": true
  }
```

popey:
Fixes #2536.

Allow `llamacpp_backend=system` when the llama.cpp descriptor declares it for the current platform. Hardware detection can omit a user-managed llama-server, which caused Lemonade to reject it before trying to start it.

Tested on Linux with a system llama-server and `--device CUDA0`. Added a `cpp-ci` regression test for the platform check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2779,7 +2779,7 @@ endif()
 set(_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC
     "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_runtime_config_backend_validation.cpp"
 )
-if(EXISTS "${_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC}")
+if(BUILD_TESTING AND EXISTS "${_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC}")
     add_executable(test_runtime_config_backend_validation
         test/cpp/test_runtime_config_backend_validation.cpp
     )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2776,6 +2776,19 @@ if(BUILD_TESTING AND EXISTS "${_CONFIG_TELEMETRY_TEST_SRC}")
     add_cpp_ci_test(ConfigTelemetryTest CI OFF COMMAND test_config_telemetry)
 endif()
 
+set(_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_runtime_config_backend_validation.cpp"
+)
+if(EXISTS "${_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC}")
+    add_executable(test_runtime_config_backend_validation
+        test/cpp/test_runtime_config_backend_validation.cpp
+    )
+    target_link_libraries(test_runtime_config_backend_validation PRIVATE lemonade-server-core)
+    add_cpp_ci_test(RuntimeConfigBackendValidationTest CI ON
+        COMMAND test_runtime_config_backend_validation
+    )
+endif()
+
 # ROCm root resolution (ROCM_PATH / rocm-sdk / /opt/rocm priority): covers the
 # external-ROCm detection that lets Lemonade skip the bundled TheRock download.
 set(_ROCM_ROOT_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_rocm_root_resolution.cpp")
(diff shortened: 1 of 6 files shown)


fl0rianr (commented):
Thanks! Nice fix, this makes sense to me. Allowing an explicitly selected `system` backend based on the backend descriptor avoids depending on hardware/backend discovery before we even try the user-managed binary, which seems like the right behavior here.

One small improvement: could we avoid adding another OS `#ifdef` block here? @superm1  did some nice cleanup around this in #2142, and we already have `get_current_os()` logic in `system_info.cpp` (with very similar logic in `backend_manager.cpp` too).

It might be a good opportunity to make that helper reusable and use it here as well. That would keep the platform detection in one place and avoid the same Windows/macOS/Linux mapping slowly drifting between files.

Non-blocking from my side, the actual fix looks good.

src/cpp/server/runtime_config.cpp
```diff
@@ -137,6 +137,26 @@ void RuntimeConfig::validate_backend_choice(const std::string& config_section,
...
+
+    if (value == "system") {
+        const auto* desc = lemon::backends::descriptor_for(recipe);
+#ifdef _WIN32
+        const std::string current_os = "windows";
+#elif defined(__APPLE__)
+        const std::string current_os = "macos";
+#else
+        const std::string current_os = "linux";
+#endif
```
superm1: please use `get_current_os()`

superm1 (commented):
I already cleaned up a lot ifdef'ery around the codebase.  PLEASE don't introduce more.

CMakeLists.txt
```diff
@@ -2776,6 +2776,19 @@ if(BUILD_TESTING AND EXISTS "${_CONFIG_TELEMETRY_TEST_SRC}")
     add_cpp_ci_test(ConfigTelemetryTest CI OFF COMMAND test_config_telemetry)
 endif()
 
+set(_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_runtime_config_backend_validation.cpp"
+)
+if(EXISTS "${_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC}")
```
```suggestion
if(BUILD_TESTING AND EXISTS "${_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC}")
```

fl0rianr (approved):
Nice cleanup with `get_current_os()`, exactly what we wanted, thanks!

Very nice work. Applied a tiny CI only fix so you are not bothered with it.

PR #3018, opened 2026-08-09
Use the selected GPU for context auto-tuning

### LINKED ISSUE #3017: CUDA context auto-tuning uses AMD GPU memory on a hybrid AMD/NVIDIA system
### Platform

Linux/Ubuntu

### Lemonade Version

11.5.2

### GPU / APU Model

AMD Radeon RX Vega M GH (4 GB) and NVIDIA GeForce RTX 2080 Ti (11 GB)

### Component

llama.cpp

### Bug Description

When `llamacpp:cuda` is selected on a system that also has an AMD GPU, context auto-tuning uses the AMD GPU's memory instead of the NVIDIA GPU used for inference.

With both GPUs visible, Lemonade 11.5.2 logged:

```text
Effective settings: llamacpp_backend=cuda
get_available_memory_gb: GPU (AMD dGPU) total=4.00 GB, used=0.01 GB → 3.99 GB available
Auto-tune ctx_size resolved to 34068
Using LlamaCpp Backend: cuda
```

The resulting child was the CUDA build, and `nvidia-smi` showed it using 3,824 MiB on the RTX 2080 Ti.

As a control, I ran the official 11.5.1 Docker image with only the RTX 2080 Ti exposed. The same `Qwen3-0.6B-GGUF` model and CUDA backend resolved context to 40,960:

```text
NVIDIA GPU: NVIDIA GeForce RTX 2080 Ti (compute 7.5, sm_75)
Auto-tune ctx_size resolved to 40960
Using LlamaCpp Backend: cuda
```

The native mixed-GPU run therefore loses context capacity based on a GPU that the selected backend does not use.

### Steps to Reproduce

1. Use a Linux system with both an AMD GPU and an NVIDIA GPU.
2. Start Lemonade 11.5.2 with debug logging.
3. Load a llama.cpp model with CUDA and automatic context sizing:

   ```console
   lemonade load Qwen3-0.6B-GGUF --llamacpp cuda
   ```

4. Inspect the `AutoTune` log and the launched backend.

Expected: when the resolved backend is CUDA, context auto-tuning uses the selected NVIDIA device's available memory.

Actual: the generic GPU memory lookup returns the first available AMD memory pool, even though inference runs on NVIDIA.

### Additional Context

`get_available_memory_gb(DeviceType)` in `src/cpp/include/lemon/auto_tune.h` receives only the generic GPU device type. It checks AMD iGPU, AMD dGPU, NVIDIA GPU, then Metal, returning the first available pool. It does not receive the resolved backend or `llamacpp_device` selection.

popey:
Fixes #3017.

Use the llama.cpp backend or explicit device to choose the GPU memory pool for context auto-tuning. Previously, mixed AMD/NVIDIA systems could use the AMD GPU's memory when running the CUDA backend.

Added tests for CUDA, ROCm, explicit CUDA device selection, and the existing fallback order. Verified on an AMD/NVIDIA Linux host: CUDA auto-tuning now uses the RTX 2080 Ti and resolves the test model's context to 40,960 instead of 34,068.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2975,6 +2975,7 @@ if(BUILD_TESTING AND EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
     target_include_directories(test_gpu_memory_selection PRIVATE
         ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
     )
+    target_link_libraries(test_gpu_memory_selection PRIVATE nlohmann_json::nlohmann_json)
     include(CTest)
     add_cpp_ci_test(GpuMemorySelectionTest CI ON COMMAND test_gpu_memory_selection)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2967,6 +2967,18 @@ if(BUILD_TESTING AND EXISTS "${_AUTO_TUNE_TEST_SRC}")
     add_cpp_ci_test(AutoTuneTest CI OFF COMMAND test_auto_tune)
 endif()
 
+set(_GPU_MEMORY_SELECTION_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_gpu_memory_selection.cpp"
+)
+if(EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
+    add_executable(test_gpu_memory_selection test/cpp/test_gpu_memory_selection.cpp)
+    target_include_directories(test_gpu_memory_selection PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+    )
+    include(CTest)
+    add_cpp_ci_test(GpuMemorySelectionTest CI ON COMMAND test_gpu_memory_selection)
+endif()
+
 set(_TELEMETRY_HELPERS_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_telemetry_helpers.cpp")
 if(BUILD_TESTING AND EXISTS "${_TELEMETRY_HELPERS_TEST_SRC}")
     add_executable(test_telemetry_helpers test/cpp/test_telemetry_helpers.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2896,6 +2896,18 @@ if(BUILD_TESTING AND EXISTS "${_AUTO_TUNE_TEST_SRC}")
     add_cpp_ci_test(AutoTuneTest CI OFF COMMAND test_auto_tune)
 endif()
 
+set(_GPU_MEMORY_SELECTION_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_gpu_memory_selection.cpp"
+)
+if(EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
+    add_executable(test_gpu_memory_selection test/cpp/test_gpu_memory_selection.cpp)
+    target_include_directories(test_gpu_memory_selection PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+    )
+    include(CTest)
+    add_cpp_ci_test(GpuMemorySelectionTest CI ON COMMAND test_gpu_memory_selection)
+endif()
+
 set(_TELEMETRY_HELPERS_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_telemetry_helpers.cpp")
 if(BUILD_TESTING AND EXISTS "${_TELEMETRY_HELPERS_TEST_SRC}")
     add_executable(test_telemetry_helpers test/cpp/test_telemetry_helpers.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2899,7 +2899,7 @@ endif()
 set(_GPU_MEMORY_SELECTION_TEST_SRC
     "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_gpu_memory_selection.cpp"
 )
-if(EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
+if(BUILD_TESTING AND EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
     add_executable(test_gpu_memory_selection test/cpp/test_gpu_memory_selection.cpp)
     target_include_directories(test_gpu_memory_selection PRIVATE
         ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2970,7 +2970,7 @@ endif()
 set(_GPU_MEMORY_SELECTION_TEST_SRC
     "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_gpu_memory_selection.cpp"
 )
-if(EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
+if(BUILD_TESTING AND EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
     add_executable(test_gpu_memory_selection test/cpp/test_gpu_memory_selection.cpp)
     target_include_directories(test_gpu_memory_selection PRIVATE
         ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
--- a/src/cpp/include/lemon/auto_tune.h
+++ b/src/cpp/include/lemon/auto_tune.h
@@ -5,6 +5,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <iomanip>
+#include <optional>
 #include <lemon/gpu_memory_selection.h>
 #include <lemon/model_manager.h>
 #include <lemon/system_info.h>
@@ -101,14 +102,18 @@ inline double get_available_memory_gb(DeviceType device_type,
         auto amd_dgpus = si->get_amd_dgpu_devices();
         auto nvidia_gpus = si->get_nvidia_gpu_devices();
         auto apple_gpu = si->get_apple_silicon_device();
-        const char* cuda_visible_devices = std::getenv("CUDA_VISIBLE_DEVICES");
+        const char* cuda_visible_devices_env = std::getenv("CUDA_VISIBLE_DEVICES");
+        const std::optional<std::string> cuda_visible_devices =
+            cuda_visible_devices_env
+                ? std::optional<std::string>(cuda_visible_devices_env)
+                : std::nullopt;
         auto pool = select_gpu_memory_pool(gpu_vendor,
                                            amd_igpu, amd_dgpus, nvidia_gpus, apple_gpu,
                                            gpu_device,
-                                           cuda_visible_devices ? cuda_visible_devices : "");
+                                           cuda_visible_devices);
         if (pool.total_gb > 0) {
             if (pool.used_gb >= 0.0) used_gb = pool.used_gb;
-            double available = pool.label == "Metal"
+            double available = pool.vendor == GpuMemoryVendor::Metal
                 ? (std::min)(pool.total_gb, (std::max)(0.0, apple_gpu.virtual_gb - used_gb))
                 : (std::max)(0.0, pool.total_gb - used_gb);
             LOG(DEBUG, "AutoTune") << "get_available_memory_gb: GPU (" << pool.label
@@ -290,12 +295,11 @@ inline int64_t resolve_auto_ctx_size(const RecipeOptions& effective_options,
     bool is_embedding = (model_info.type == ModelType::EMBEDDING);
     std::string backend;
     std::string device;
-    if (effective_options.get_recipe() == "llamacpp") {
-        const json backend_json = effective_options.get_option("llamacpp_backend");
-        const json device_json = effective_options.get_option("llamacpp_device");
-        if (backend_json.is_string()) backend = backend_json.get<std::string>();
-        if (device_json.is_string()) device = device_json.get<std::string>();
-    }
+    const std::string recipe = effective_options.get_recipe();
+    const json backend_json = effective_options.get_option(recipe + "_backend");
+    const json device_json = effective_options.get_option(recipe + "_device");
+    if (backend_json.is_string()) backend = backend_json.get<std::string>();
+    if (device_json.is_string()) device = device_json.get<std::string>();
     double available_gb = get_available_memory_gb(
         model_info.device, gpu_memory_vendor_for_target(backend, device), device);
 
--- a/src/cpp/include/lemon/auto_tune.h
+++ b/src/cpp/include/lemon/auto_tune.h
@@ -288,8 +288,10 @@ inline int64_t resolve_auto_ctx_size(const RecipeOptions& effective_options,
     std::string backend;
     std::string device;
     if (effective_options.get_recipe() == "llamacpp") {
-        backend = effective_options.get_option("llamacpp_backend").get<std::string>();
-        device = effective_options.get_option("llamacpp_device").get<std::string>();
+        const json backend_json = effective_options.get_option("llamacpp_backend");
+        const json device_json = effective_options.get_option("llamacpp_device");
+        if (backend_json.is_string()) backend = backend_json.get<std::string>();
+        if (device_json.is_string()) device = device_json.get<std::string>();
     }
     double available_gb = get_available_memory_gb(
         model_info.device, gpu_memory_vendor_for_target(backend, device), device);
--- a/src/cpp/include/lemon/auto_tune.h
+++ b/src/cpp/include/lemon/auto_tune.h
@@ -87,7 +87,8 @@ static double get_used_memory_gb(DeviceType device_type) {
 /// CPU  → system RAM minus currently-used RAM
 /// NPU  → system RAM minus currently-used RAM
 inline double get_available_memory_gb(DeviceType device_type,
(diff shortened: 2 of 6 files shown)


kenvandine (commented):
Good fix for the hybrid GPU issue. A few concerns:

1. **Missing `BUILD_TESTING` guard in CMakeLists.txt** — The test registration needs `if(BUILD_TESTING AND EXISTS ...)` per the project rules. Distro packaging fails with `BUILD_TESTING=OFF` if this guard is absent.

2. **Potential `json::get<std::string>()` exception** — In `auto_tune.h`, `get_option()` can return null json; `.get<std::string>()` on null throws. Add a null check or use `.value()` pattern.

3. **`consumed` check in `gpu_indices_for_target`** — The strict `consumed == token.size() - prefix.size()` check rejects tokens with suffixes like `CUDA0+1`. Consider `consumed > 0` for tolerance.

4. **nvidia-smi CSV parsing fragility** — The right-side `rfind(", ")` approach works but gets fragile with more fields. Worth a comment explaining the rationale.

kenvandine:
## Review Feedback

Overall: Good fix for the hybrid GPU issue. The approach of routing GPU memory selection through backend/device vendor is the right direction. A few issues to address:

### 1. Missing `BUILD_TESTING` guard (CMakeLists.txt)
The new test registration at line ~2758 needs `if(BUILD_TESTING AND EXISTS ...)` per project rules. Distro packaging configures with `BUILD_TESTING=OFF` — calling `add_cpp_ci_test()` without this guard is a **fatal error**. The existing `AutoTuneTest` above has this guard; the new one doesn't.

### 2. Potential `json::get<std::string>()` exception (auto_tune.h)
In the call `effective_options.get_option("llamacpp_backend").get<std::string>()` — if `get_option()` returns a null json value, `.get<std::string>()` throws `type_error::type_error_302`. Add a null check:

\`\`\`cpp
auto backend_json = effective_options.get_option("llamacpp_backend");
if (!backend_json.is_null())
    backend = backend_json.get<std::string>();
\`\`\`

### 3. Strict `consumed` check in `gpu_indices_for_target` (gpu_memory_selection.h)
The check `consumed == token.size() - prefix.size()` rejects tokens with suffixes. Consider `consumed > 0` for tolerance of unexpected device string formats.

### 4. NVML `used` field (system_info.cpp)
Good catch adding `mem.used` to NVML queries — the `NvmlMemory` struct already has the `used` field, so this is a safe addition.

### 5. AMD sysfs usage fields (system_info.cpp)
Adding `mem_info_vram_used` and `mem_info_gtt_used` reads is good. Verify these sysfs paths exist on all target kernel versions — they may not be available on older kernels.

CMakeLists.txt
```diff
@@ -2748,6 +2748,18 @@ if(BUILD_TESTING AND EXISTS "${_AUTO_TUNE_TEST_SRC}")
     add_cpp_ci_test(AutoTuneTest CI OFF COMMAND test_auto_tune)
 endif()
 
+set(_GPU_MEMORY_SELECTION_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_gpu_memory_selection.cpp"
+)
+if(EXISTS "${_GPU_MEMORY_SELECTION_TEST_SRC}")
+    add_executable(test_gpu_memory_selection test/cpp/test_gpu_memory_selection.cpp)
+    target_include_directories(test_gpu_memory_selection PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
```
kenvandine, line 2757: **Missing `BUILD_TESTING` guard.** Per AGENTS.md, every test registration must be wrapped in `if(BUILD_TESTING AND ...)`. Distro packaging fails with `BUILD_TESTING=OFF` — calling `add_cpp_ci_test()` without this guard is a fatal error.

Fix: change `if(EXISTS ...)` → `if(BUILD_TESTING AND EXISTS ...)`

src/cpp/include/lemon/auto_tune.h
```diff
@@ -85,70 +86,40 @@ static double get_used_memory_gb(DeviceType device_type) {
...
         auto amd_igpu = si->get_amd_igpu_device();
-        if (amd_igpu.available && amd_igpu.vram_gb > 0) {
-            // iGPU total = dedicated VRAM + GTT (system memory pool accessible by GPU)
-            double total_gb = amd_igpu.vram_gb + amd_igpu.virtual_gb;
-            double available = (std::max)(0.0, total_gb - used_gb);
-            LOG(DEBUG, "AutoTune") << "get_available_memory_gb: GPU (AMD iGPU) total="
-                                   << std::fixed << std::setprecision(2) << total_gb
-                                   << " GB (vram=" << amd_igpu.vram_gb
-                                   << " + gtt=" << amd_igpu.virtual_gb << "), used=" << used_gb
-                                   << " GB → " << available << " GB available"  << " ";
```
kenvandine, line 106: **Potential `json::get<std::string>()` exception.** `get_option()` can return a null json value. Calling `.get<std::string>()` on null throws `type_error::type_error_302`. Add a null check or use `.value()` pattern.

src/cpp/include/lemon/gpu_memory_selection.h
```diff
@@ -0,0 +1,130 @@
...
+
+enum class GpuMemoryVendor { Any, Amd, Nvidia, Metal };
+
+struct GpuMemoryPool {
+    double total_gb = 0.0;
+    double used_gb = -1.0;
+    std::string label;
+};
+
+inline std::vector<int> gpu_indices_for_target(const std::string& device,
```
kenvandine, line 20: **Strict `consumed` check.** `consumed == token.size() - prefix.size()` rejects tokens with suffixes like `CUDA0+1`. Consider `consumed > 0` for tolerance.

kenvandine (commented):
See inline comments below.

src/cpp/include/lemon/auto_tune.h
```diff
@@ -85,70 +86,37 @@ static double get_used_memory_gb(DeviceType device_type) {
...
-                                       << " GB → " << available << " GB available"  << " ";
-                return available;
-            }
+        auto apple_gpu = si->get_apple_silicon_device();
+        auto pool = select_gpu_memory_pool(gpu_vendor,
+                                           amd_igpu, amd_dgpus, nvidia_gpus, apple_gpu);
+        if (pool.total_gb > 0) {
+            double available = pool.label == "Metal"
+                ? (std::min)(pool.total_gb, (std::max)(0.0, apple_gpu.virtual_gb - used_gb))
+                : (std::max)(0.0, pool.total_gb - used_gb);
```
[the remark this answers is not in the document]
popey: Fixed in 84d683ad. The selected pool now carries per-device used memory from nvidia-smi/NVML or AMD sysfs. The generic metric is only used when per-device usage is unavailable.

CMakeLists.txt
popey: Fixed in 24390a2a. I also verified a fresh CMake configuration with BUILD_TESTING=OFF.

src/cpp/include/lemon/gpu_memory_selection.h
```diff
@@ -0,0 +1,61 @@
...
+    double total_gb = 0.0;
+    std::string label;
+};
+
+inline GpuMemoryVendor gpu_memory_vendor_for_target(const std::string& backend,
+                                                     const std::string& device) {
+    if (backend == "cuda" || device.rfind("CUDA", 0) == 0) return GpuMemoryVendor::Nvidia;
+    if (backend.rfind("rocm", 0) == 0 || device.rfind("ROCm", 0) == 0) return GpuMemoryVendor::Amd;
+    if (backend == "metal" || device.rfind("Metal", 0) == 0) return GpuMemoryVendor::Metal;
+    return GpuMemoryVendor::Any;
```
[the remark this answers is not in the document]
popey: Fixed in 84d683ad. Explicit CUDA and ROCm ordinals are preserved. For a comma-separated list, auto-tuning uses the selected GPU with the least available memory to avoid over-sizing.

src/cpp/include/lemon/auto_tune.h
popey: Fixed in 24390a2a. Both option values are checked with is_string() before conversion.

src/cpp/include/lemon/gpu_memory_selection.h
popey: I kept this strict intentionally. Treating CUDA0+1 as CUDA0 could tune against the wrong target when the suffix syntax is unknown. A malformed token now rejects the whole selection, including comma-separated lists, and the behavior has regression coverage.

popey:
Thanks for the review, @kenvandine. I have addressed the feedback and added coverage for the edge cases. Could you take another look when you have a chance?

bitgamma:
@popey, thanks for your PR. I had Qwen3.8 take a deeper look and found a few points to address:

1. CUDA<n> maps to the nvidia-smi index, but llama.cpp's --device CUDA<n> is an ordinal after the CUDA_VISIBLE_DEVICES filter. In select_nvidia(), CUDA1 matches gpu.index == 1 (physical). But apply_cuda_env_vars() respects an existing CUDA_VISIBLE_DEVICES in the environment, and lemonade itself sets one (nvidia-smi 
 indices of matching-arch GPUs) on mixed-arch multi-GPU hosts. So CUDA_VISIBLE_DEVICES=2,3 + --llamacpp-device CUDA1 runs llama.cpp on physical GPU 3 while auto-tune sizes against physical GPU 1 — possibly a GPU that isn't even visible. Same class of caveat for ROCm<n>: select_amd() indexes a synthetic vector (iGPU  
 first, then dGPUs in sysfs order), while llama.cpp's ROCm ordinal follows HSA agent enumeration — these can disagree on APU+dGPU hybrids. This is a big improvement over before, but the mapping assumes "no/passthrough CUDA_VISIBLE_DEVICES"; at minimum it deserves a comment (or reusing apply_cuda_env_vars' visibility 
 logic) so the next person doesn't assume it's exact.                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                              
 2. Bare device prefix degrades to the 4096 fallback. gpu_indices_for_target("CUDA", "CUDA") returns {}, and device.rfind("CUDA", 0) == 0 && indices.empty() → empty pool → 0.0 → AUTO_CTX_FALLBACK. Meanwhile validate_device_backend_match() lowercases and only checks the prefix — so --llamacpp cuda --llamacpp-device   
 CUDA (or backend system + CUDA) is accepted at load, but auto-tune silently clamps to ctx 4096/8192 instead of using that vendor's pool. A bare prefix isn't malformed, it's "vendor, no index" — I'd route it through the same first-available-device path the no-device case uses.                                         
                                                                                                                                                                                                                                                                                                                              
 3. Vendor detection is case-sensitive; the rest of the pipeline isn't. gpu_memory_vendor_for_target() uses case-sensitive rfind("CUDA"/"ROCm"/"Metal") and == "cuda"/"metal", but validate_device_backend_match() lowercases both sides, so cuda0/Cuda0 are valid load-time inputs. backend="system" + device="cuda0"        
 (lowercase) resolves to vendor Any → generic fallback order → exactly the original bug (AMD pool) via a lowercase spelling. Reuse the same case-insensitive matching.

popey:
@bitgamma Thanks, I've addressed these cases.

- CUDA ordinals are now resolved after `CUDA_VISIBLE_DEVICES`, for both numeric device IDs and GPU UUIDs. The tests include `CUDA_VISIBLE_DEVICES=2,3` with `CUDA1` selecting physical GPU 3.
- Bare `CUDA` and `ROCm` selections now use the first available pool for that vendor.
- Backend and device matching is now case-insensitive, consistent with load-time validation.

ROCm ordinals still use Lemonade's current iGPU-then-dGPU ordering because SystemInfo does not expose HSA runtime ordinals. That assumption is now documented next to the selection code.

jeremyfowers:

**Ready for review** — the checks below passed. No reviewers are suggested: this PR already has the review contribute.md's rung asks for (bitgamma, kenvandine).

**Attention level:** Elevated — needs 2 reviewers including 1 subject-area expert in llamacpp, Nvidia CUDA, ROCm

Why this rung: refactor that restructures existing code. The diff extracts GPU memory selection logic into a new module (gpu_memory_selection.h) and restructures how get_available_memory_gb resolves per-device memory — a contained refactor touching the auto-tune path, system info queries, and test infrastructure.

**Description:** Refactors context auto-tuning to select the correct GPU memory pool based on the llama.cpp backend or explicit device, fixing mixed AMD/NVIDIA systems where the wrong GPU was used for context sizing.

The PR body accurately describes the bug (issue #3017), the fix (backend/device-aware GPU memory selection), the test coverage, and the verified real-hardware result.

</details>

All changes serve one goal: making context auto-tuning use the correct GPU's memory based on the selected backend or explicit device. The new gpu_memory_selection.h module, the extended system_info queries, and the refactored get_available_memory_gb all exist to solve this single problem.

Checked: All six changed files serve the single goal: the new gpu_memory_selection.h module centralizes selection, auto_tune.h delegates to it, system_info.h/cpp add per-device used memory tracking, CMakeLists.txt adds the test, and test_gpu_memory_selection.cpp tests the new logic.

</details>

<details><summary>Alignment issues: <b>None found</b></summary>

Checked: PR links issue #3017 and improves auto-tune GPU selection within the existing working-group scope; inline reviews from kenvandine (subject-area maintainer) show design engagement; no philosophy or contribute.md clauses are violated.

</details>

Checked: Checked docs/dev/working-groups/auto-tune.md and docs/guide/configuration/llamacpp.md; changes are internal API improvements with no new user-facing configuration or behavior.

</details>

Checked: New test/cpp/test_gpu_memory_selection.cpp exercises vendor selection, explicit device selection, device lists, malformed input, unknown usage, and fallback order; registered in CMakeLists.txt with BUILD_TESTING guard and add_cpp_ci_test(GpuMemorySelectionTest CI ON ...); cpp-ci workflow will execute it.

</details>

<details><summary>Breaking changes: <b>None found</b></summary>

Checked: get_available_memory_gb gains two optional parameters with defaults; GPUInfo gains vram_used_gb and virtual_used_gb fields; nvidia-smi query adds memory.used; all internal surfaces, no user-visible API change.

</details>

_Reviewed at head `b9167a9` by repo-manager (tier: reviewers). Regenerate with `repo-manager review-pr 3018`._

bitgamma (approved):
Lgtm, @kenvandine do you want to have another look before merging?

kenvandine (changes requested):
Thanks for tightening up GPU memory selection for auto-tuning — the vendor-aware pool selection is a solid fix for the hybrid-GPU case. Ran the new gpu_memory_selection.h unit tests (18/18 pass). A few things worth addressing before merge:

  Correctness

  - parse_memory_sysfs() (system_info.cpp:3517) returns 0.0 on a read/parse failure instead of the -1.0 "unknown" sentinel the rest of the code relies on. A transient read failure (permission edge case, driver hiccup, GPU reset) would then read as "0 used / fully free" in get_available_memory_gb, overestimating available VRAM. The read_sysfs_u64 lambda near line 4206 already handles this correctly with a success flag —
  worth reusing that pattern here.
  - select_nvidia() (gpu_memory_selection.h:141) applies CUDA_VISIBLE_DEVICES filtering even in the generic/no-explicit-device fallback path. On a hybrid host running the Vulkan backend (which ignores that env var), a stray CUDA_VISIBLE_DEVICES left over from another tool would still filter the NVIDIA pool, skewing the "any vendor" memory estimate.
  - most_constrained() (gpu_memory_selection.h:101) treats an unknown used_gb (-1.0 sentinel) as 0 used rather than worst-case. With llamacpp_device="CUDA0,CUDA1" and GPU0's usage unreadable, GPU0 looks maximally free and gets picked over a GPU we actually know has more headroom.
  - CUDA_VISIBLE_DEVICES="" (explicit "hide all") is indistinguishable from unset ("show all") since both are .empty() — worth special-casing so an explicitly empty list doesn't fall back to enumerating every NVIDIA GPU.
  - Explicit ROCm<N> device mapping (gpu_memory_selection.h:122) uses SystemInfo's iGPU-then-dGPU detection order, which isn't guaranteed to match llama.cpp's actual HSA-agent enumeration order. Worth confirming these line up, or flagging the assumption more visibly if it's known-approximate.

  Scope

  - The vendor-aware fix in auto_tune.h:293 is gated on recipe == "llamacpp" only. sd-cpp/whisper.cpp/vllm follow the same {recipe}_backend/{recipe}_device convention and can hit the identical mis-selection bug on hybrid GPU hosts — is llama.cpp-only scoping intentional for this PR, with the others tracked separately?

  Nit

  - auto_tune.h:111 branches on pool.label == "Metal" (string match) rather than a typed field on GpuMemoryPool. A future rename or label collision would silently take the wrong branch with no compiler error — consider a vendor/is_unified_memory bool instead.

  Happy to take another pass once these are addressed.

kenvandine (approved):
@popey thanks for fixing those issues, this looks great!

bitgamma:
@popey merging failed because the test fails to compile in macos

popey:
@bitgamma Thanks — I reproduced the macOS failure. The test target was missing the CMake dependency that provides the `nlohmann/json.hpp` include path.

I added the dependency and verified the CMake target with AppleClang 21 on macOS ARM64; all 22 focused tests pass. I also rebased onto current main, built `lemond`, ran all 70 `cpp-ci` tests on Ubuntu 24.04 x86-64, and checked a fresh `BUILD_TESTING=OFF` configuration.

popey:
Looks like the Windows failure is in `test_021a_pull_sdcpp_import_preserves_merged_recipe_options`: the pull returned 500 while resolving/downloading the test GGUF.

I checked the other 29 recent C++ CI runs. I couldn't find this failure elsewhere; several PRs passed the same Windows endpoint job today, and the same test passed in the other recent failed runs. All the Windows build and packaging jobs on this PR passed too.

This looks like a one-off cache/download failure, and it happens outside the code changed here. Could someone please re-run the failed Windows job? If it fails the same way again, I'll dig into it further.

PR #3022, opened 2026-08-09
ci: make inference test matrix test-owned

fl0rianr:
## Summary

This is another step in the recent CI cleanup and consolidation work.

The inference test matrix is currently closely tied to the workflow definition, which makes it harder to reason about test ownership and adds maintenance effort as backend coverage changes.

This PR moves that ownership closer to the tests and derives the workflow matrix from a test-owned manifest instead.

## What changes

* Introduces a test-owned source of truth for inference coverage
* Derives the workflow matrix from that configuration
* Keeps the existing backend aliases and special-case mappings explicit
* Adds validation so the test manifest and generated CI configuration cannot silently drift apart
* Preserves the current inference coverage

## Why

Several recent changes have already simplified and consolidated the inference CI setup. This builds on that work by removing another layer of duplicated CI configuration.

The immediate benefit is less effort when adding, removing, or changing inference coverage: the test definition becomes the authoritative place to describe what needs to run.

More importantly, this also creates the foundation for the next step: using the same test-owned information together with runtime backend usage to avoid running inference CI that is not affected by a change.

So this PR is intentionally mostly structural. It reduces CI maintenance overhead now and enables more targeted CI execution in the follow-up.

### DIFF
--- a/.github/inference-tests.json
+++ b/.github/inference-tests.json
@@ -1,0 +1,482 @@
+{
+  "schema_version": 1,
+  "description": "Canonical inference-test definitions. Tests describe the backends they directly exercise and consume; platforms describe where each test runs.",
+  "tests": {
+    "llamacpp": {
+      "script": "server_llm.py",
+      "extra_args": "--wrapped-server llamacpp",
+      "impact": {
+        "direct": [
+          "llamacpp"
+        ],
+        "consumes": []
+      }
+    },
+    "omni": {
+      "script": "server_omni.py",
+      "extra_args": "--wrapped-server llamacpp",
+      "impact": {
+        "direct": [],
+        "consumes": [
+          "llamacpp",
+          "sdcpp",
+          "whispercpp",
+          "kokoro"
+        ],
+        "note": "The Omni collection integration test exercises its llamacpp planner and image, transcription, and speech components."
+      }
+    },
+    "ryzenai": {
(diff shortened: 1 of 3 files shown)


jeremyfowers:
> This builds on that work by removing another layer of duplicated CI configuration.

I'm confused by this and would appreciate a clearer explanation of the benefits and tradeoffs in the PR description.

This PR adds 1600 LoC and deletes none. I don't see how that removes a layer of duplicated CI configuration?


Additional Requests:

1. Delete the migration path. Drop apply, _tool_source, _ensure_check_job, build_manifest, _infer_impact, _git_dirty_targets, and the duplicated constants. The migration already happened; the JSON is the source of truth now. That should cut the script by ~40% and remove the self-installing behavior.
2. Wire the job into required checks, or it's decoration.
3. Support a per-leg comment or note field that sync preserves, and fix the misleading error message. (A stray field on a platform entry currently leaks into the GitHub matrix as a variable, so this needs an explicit field, not a convention.)
4. Decide on dmg — either generate it too, or drop it from the manifest and say why.

jeremyfowers:
> I'm confused by this and would appreciate a clearer explanation of the benefits and tradeoffs in the PR description.

Requesting an RFC here so that we can make an educated decision about this. Marking PR draft until RFC converges.

PR #3023, opened 2026-08-09
GUI3 fix: gui2 hugging face search parity

### LINKED ISSUE #2948: Copy the the model search filters from GUI2

fl0rianr:
**Tiny fix:** GUI2 keeps Hugging Face rows visible even when the checkpoint is already registered locally.

image

Follow-up on #2967 
Fixes #2948


PCAssistSoftware:
approved 

Tested and model search now working as expected and on par with GUI2 from what I can see, previous search of qwen3.6-35b correctly listing the unsloth models I would expect to see

image

jeremyfowers (approved):
Good enough for @PCAssistSoftware good enough for me!

PR #3024, opened 2026-08-09
GUI3 feat/fix: new design line concept and no backend colors

### LINKED ISSUE #2940: Reduce/unify the amount of colors used in the Models tab filters
Image

The colors do not appear to come from a unified pallate/theme between Task and Backends

fl0rianr:
Colors are now only for TASK.
We already have more backend the Tasks and I suspect then trend to go on.
So dropping color for backend reduces maintainer headache.
Saved 4px per model element by moving design line a bit higher.

Overall - I think we have a significant improvement here.

image

image

on light mode same TASK colors used, maybe needs thicker icon line with 

Router mode is added.

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -279,11 +279,24 @@ function mcpToolNamesForServers(servers: McpServerToolOption[], serverIds: strin
     .flatMap(server => server.toolOptions.map(tool => tool.runtimeName));
 }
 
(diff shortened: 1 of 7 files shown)


fl0rianr:
Now I'm satisfied with this one:
image
audio and music have the same color, same mode

image

The colored Task Icons have stroke width 2 which is much better, also in chat:
image

Also a visual bug fix:
current:
image

new:
image

jeremyfowers (approved):
This looks so much calmer now, great work!

PR #3029, opened 2026-08-10
fix(tts): stop dropping response_format on the streaming path

ZMXJJ:
## Summary

`/audio/speech` hardcoded `response_format = "pcm"` for any streaming request and never read the value the client sent. That silently dropped an explicit `response_format`, and it locked wav-only TTS backends out of streaming entirely — the forced `pcm` isn't in their supported set, so the request was rejected outright.

## What changes

- **`ITextToSpeechServer::supported_streaming_audio_formats()`** — a backend's streaming encoder is often narrower than its buffered one. Empty (the default) means the buffered list covers both transports.

- **Kokoro declares `{pcm}` for streaming.** Kokoros forces headerless s16le on that path whatever `response_format` asked for, so honoring anything else there would serve bytes the Content-Type doesn't match. Its buffered set is also narrowed to the four formats it actually implements — `aac` and `flac` deserialize fine but silently fall back to MP3 bytes.

- **Transport and container are now orthogonal in the gateway.** `stream_format` picks the transport, `response_format` the container. An explicit `response_format` wins on both paths and is rejected with 400 if the backend can't encode it, rather than silently swapped. Only the implicit default still varies by transport (`pcm` streaming, `mp3` buffered), and it yields to whatever the backend declares.

⚠️ **Behavior change:** `stream_format=audio` with `response_format=mp3` against Kokoro now returns 400 (`supported: pcm`) instead of silently returning PCM.

## Testing

`test/server_tts.py` gains a Content-Type and container-magic assertion on the streaming test, plus a new `test_007_tts_stream_rejects_unstreamable_format`. Both run against `kokoro-v1` in the existing TTS CI job — no new backend or runner needed.

### DIFF
--- a/docs/api/openai.md
+++ b/docs/api/openai.md
@@ -913,11 +913,11 @@ A typical workflow is to generate an image first, then upscale it:
 ## `POST /v1/audio/speech`
 <sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
 
-Speech Generation API. You provide a text input and receive an audio file. This API uses [Kokoros](https://github.com/lucasjinreal/Kokoros) as the backend.
+Speech Generation API. You provide a text input and receive an audio file. Which engine serves the request depends on the model.
 
 > **Note:** Supported models are `kokoro-v1` (fixed voices, [Kokoros](https://github.com/lucasjinreal/Kokoros) backend) and the OpenMOSS family — `OpenMOSS-TTS` (voice cloning from a reference WAV) and `MOSS-VoiceGen` (voice design from a text description).
 >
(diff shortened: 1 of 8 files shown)


fl0rianr (commented):
Thanks for contributing and the good work. This looks like the right direction overall. Separating the streaming transport from the audio format makes sense, and the Kokoro handling looks much clearer now.

Before merging, I think there are three small gaps worth addressing:

* The main regression described in the PR is that WAV-only backends could not stream because `pcm` was forced. It would be good to add an OpenMOSS streaming test for `stream_format="audio"` with WAV and verify `200`, `audio/wav`, and a RIFF header.
* The API docs still say that `stream_format=audio` outputs PCM and that streaming is only supported for Kokoro. With this change, that is no longer generally true, so the docs should reflect the new transport-vs-format behavior.
* After choosing the effective `response_format`, I think we should also write it back into `request_json` before forwarding it to the backend. Otherwise the gateway may validate/set the MIME type for one format while the backend still receives no `response_format` and falls back to its own default.

Other than that, the core fix looks solid to me.

fl0rianr (approved):
LGTM, thanks!

PR #3030, opened 2026-08-10
feat: add VoxCPM2 text-to-speech backend

ZMXJJ:
## Summary

Adds VoxCPM2 (OpenBMB) as an experimental text-to-speech backend, wrapping llama.cpp-omni's `llama-tts-server` subprocess to serve `/audio/speech`. Ships three catalog entries: a suggested Q8 default, an F16 variant, and a ModelScope-hosted mirror.

Depends on #3029 — VoxCPM2 emits wav only, which the streaming path rejects without it.

## Notes for review

- **Two GGUF files per model** (BaseLM + Acoustic), resolved through the `checkpoints` map the way sd-cpp handles flux.

- **Streaming forwards to the buffered `/v1/audio/speech` on purpose.** `llama-tts-server` also exposes `/v1/audio/speech/stream`, but it re-decodes a context window of latents every step, costing more than one buffered decode of the whole utterance. The comment at the call site says so, so nobody flips it and lands a slower path.

- **Backend binaries come from a fork** (`ZMXJJ/llama.cpp-omni`) that builds the per-variant `llama-tts-server` archives; upstream llama.cpp-omni only releases the Comni desktop installers. SHA-256 digests for all seven archives are pinned in `backend_versions.json`, per #2710. The `tts-` tag prefix keeps these clear of upstream's `v1.0.x` Comni tags.

- **The ModelScope entry is `suggested: false`** so default listings stay on Hugging Face. It exists because HF is unreachable in parts of the world VoxCPM2 targets. Note the owner differs per registry: `DennisHuang648` on HF, `DennisHuang` on ModelScope.

## Testing

New `test/server_tts_voxcpm2.py`, plus a `voxcpm2` entry in `test/utils/capabilities.py`. Wired into CI as the `tts-voxcpm2` job on vulkan, for both Windows and Linux.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1079,7 +1079,7 @@ jobs:
           - name: tts-voxcpm2
             script: server_tts_voxcpm2.py
             extra_args: "--wrapped-server voxcpm2"
-            backends: "vulkan"
+            backends: "vulkan cpu"
             runner: [Windows, rocm, vulkan, lemon-prod]
     env:
       LEMONADE_CI_MODE: "True"
@@ -1337,7 +1337,7 @@ jobs:
           - name: tts-voxcpm2
             script: server_tts_voxcpm2.py
             extra_args: "--wrapped-server voxcpm2"
-            backends: "vulkan"
+            backends: "vulkan cpu"
             runner: [Linux, vulkan, rocm, lemon-prod]
     env:
       LEMONADE_CI_MODE: "True"
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -1076,6 +1076,11 @@ jobs:
             extra_args: "--wrapped-server openmoss"
             backends: "vulkan rocm"
             runner: [Windows, rocm, vulkan, lemon-prod]
+          - name: tts-voxcpm2
+            script: server_tts_voxcpm2.py
+            extra_args: "--wrapped-server voxcpm2"
+            backends: "vulkan"
+            runner: [Windows, rocm, vulkan, lemon-prod]
     env:
       LEMONADE_CI_MODE: "True"
       HF_TOKEN: ${{ secrets.HUGGINGFACE_ACCESS_TOKEN }}
@@ -1328,6 +1333,11 @@ jobs:
             script: server_tts_openmoss.py
             extra_args: "--wrapped-server openmoss"
             backends: "vulkan rocm"
+            runner: [Linux, vulkan, rocm, lemon-prod]
+          - name: tts-voxcpm2
+            script: server_tts_voxcpm2.py
+            extra_args: "--wrapped-server voxcpm2"
+            backends: "vulkan"
             runner: [Linux, vulkan, rocm, lemon-prod]
     env:
       LEMONADE_CI_MODE: "True"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,6 +168,7 @@ set(LEMON_BACKENDS
     "onnxruntime|onnxruntime"
     "trellis|trellis"
     "openmoss|openmoss"
+    "voxcpm2|voxcpm2"
 )
 
 # ============================================================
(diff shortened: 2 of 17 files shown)


fl0rianr (changes requested):
Thanks for your contribution towards Lemonade!

> [!IMPORTANT]
> Generally I see sense in adding this backend, but adding backends towards Lemonade is not a simple decision and we would like to discuss this on Discord and agreed on by us maintainers before making such change with many implications e,g, maintaining, security. 

**That said, here a code-wise review:**
* The backend currently downloads and executes native binaries from a contributor-owned fork (`ZMXJJ/llama.cpp-omni`). The checksums are pinned, which is good, but the binary source and hashes still come from the same trust domain. Before merging, we should decide on a maintainer-controlled or otherwise trusted release/provenance process for these binaries.

* The backend advertises `metal`, `cuda`, `vulkan`, and `cpu`, but CI only tests Vulkan on Linux and Windows. Since these variants use different binaries/platform paths, the supported variants should ideally be covered by CI (expect `cuda`, we have no runner for that yet) or at least dedicated install/startup smoke tests.

* `voxcpm2` uses `DEVICE_GPU` as its default device while also supporting a CPU backend. This means `--voxcpm2 cpu` can still be reported/accounted as GPU internally. The backend should use the variant-aware device handling intended for this case.

* `test_003_streaming` currently sends `stream_format: audio`, but `requests.post()` itself is not using `stream=True`, so it only verifies the final WAV response. It would be better to actually consume the response as a stream and verify the transport path as well.

* Small consistency issue: the unsupported-backend error mentions ROCm, but ROCm is not declared or shipped for VoxCPM2.

* The TTS docs still say that the endpoint "uses Kokoros as the backend" and the `response_format` table reads as if all formats are generally supported. With OpenMOSS and VoxCPM2 now present, this should be described as backend-dependent.

Also, since this PR already includes the generic TTS streaming/format changes from #3029, I would prefer merging #3029 first and then rebasing this PR to keep the two changes cleanly separated.

The VoxCPM2 wrapper itself looks generally well structured, and the model/checkpoint and release wiring looks reasonable. These are the main points I would address before merging.

fl0rianr (commented):
Thanks for addressing the previous review. I went through the current state again and most of the original points are resolved now: CPU is covered in the CI matrix, device reporting is variant-aware, the streaming test actually consumes a streamed response, the ROCm error message is fixed, and the TTS docs have been updated.

There are still a few things I would like to see before approval:

* The PR currently conflicts with main, so it needs another rebase. Since some of the conflicts affect generated backend docs/config, please regenerate those from the current generator after rebasing rather than resolving the generated output manually.
* The new CPU/Vulkan backend tests are present, but they have not actually run on the current head yet. Please add ci:backends and get those rows green before merging.
* VoxCPM2 still advertises Metal support without testing it in Lemonade. There is already a macOS inference job (test-dmg-inference) and PRs can opt into it with ci:macos, so I think the VoxCPM2 Metal test should be added there and run once. I would not block on CUDA coverage alone since there currently is no suitable CUDA runner.
* The binary fork question: I checked the tts-v0.1.2 release more closely: the tag, successful release workflow SHA, published assets, and hashes pinned in Lemonade are consistent, so the provenance of this particular release looks reasonable. The remaining concern is ownership and long-term maintenance: Lemonade would automatically execute native binaries from a contributor-owned fork which also contains its own VoxCPM2 runtime changes. Will leave that to @ramkrishna2910 
* Very minor: the PR description still says the CI integration only covers Vulkan, while it now covers Vulkan and CPU.

Other than that, the wrapper itself looks much better after the changes and I don't see another clear code-level blocker at this point.

Geramy:
reference https://github.com/lemonade-sdk/lemonade/pull/3182 - new policy for adding backends that is not approved yet or finished but this will be a good example of a backend wanting to be added vs what the policy says and could help shape if the policy is correct as well with a real world example.

jeremyfowers (changes requested):
Adding a new backend requires:
- [ ] an analysis of why the added project scope justifies the maintenance burden
- [ ] a commitment from the author, or some other maintainer, to maintain the backend long-term

Added @Geramy as a reviewer to shepherd that process.

jeremyfowers:
> Adding a new backend requires:
> 
>     * [ ]  an analysis of why the added project scope justifies the maintenance burden
> 
>     * [ ]  a commitment from the author, or some other maintainer, to maintain the backend long-term
> 
> 
> Added @Geramy as a reviewer to shepherd that process.

We have a process for making these decisions now, see: https://github.com/lemonade-sdk/lemonade/pull/3521

Marking PR as draft until an RFC discussion converges.

Issue #2947, opened 2026-08-07
Models tab - in models list you get icons for Chat, Vision, Tool Use and Reasoning, but at the top of the model card it only shows Chat - see annotated screenshot


PCAssistSoftware:
### Platform

Linux/Ubuntu

### Lemonade Version

11.5.1

### GPU / APU Model

AMD Strix Halo

### Component

llama.cpp

### Bug Description

Models tab - in models list you get icons for Chat, Vision, Tool Use and Reasoning, but at the top of the model card it only shows Chat - see annotated screenshot

<img width="1579" height="617" alt="Image" src="https://github.com/user-attachments/assets/17185b04-19df-4697-81ff-08cdb58c95a3" />

### Steps to Reproduce

Open Models tab and select a model in the list

### Log Output

```shell

```

jeremyfowers:
+1 we need to ensure metadata is consistent

the list entry in the middle should show compressed/partial metadata; the details panel should be a superset of that, with everything clearly labeled.

jeremyfowers:
Removing this from release blockers; we should do it, it just isn't a blocker.

PR #3032, opened 2026-08-10
GUI3 fix: unify list styles

### LINKED ISSUE #2933: Unify the list styles in Chat History, Models list, Telemetry list, etc.
Image

Image

Image

These should be identical app-wide. Do a search for other instances of lists before declaring this done.

fl0rianr:
image

image

Model list was already in that style and the specific list in PR #3009 is not touched here.

Closes #2933

### DIFF
--- a/src/app/DESIGN.md
+++ b/src/app/DESIGN.md
@@ -19,6 +19,7 @@ This document is the contract for Lemonade's desktop and web UI. It describes th
 - `src/styles/styles.css` implements components and feature layout with those tokens.
 - Shared component classes use the `workspace-*` prefix. Feature classes may refine a shared component but must not redefine its typography, spacing scale, surface hierarchy, focus treatment, or control shape.
 - React implementations of the shared panels and controls live in `src/components/WorkspacePanels.tsx`, `WorkspaceRailHeader.tsx`, and `WorkspaceMobileMenuButton.tsx`.
+- Every selectable list renders `WorkspaceList` + `WorkspaceListRow` (`.workspace-list` / `.workspace-list-row`). A tab supplies the row's content, never its geometry, typography, status idiom, or selection treatment.
 - Dynamic values such as progress percentage, measured resizer position, chart geometry, and backend identity may be passed through inline custom properties. Repeated fixed values use tokens; feature-local geometry stays in its named class.
 
 ### Color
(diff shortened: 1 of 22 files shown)


jeremyfowers:
1. What about the search results for HF/MS models in the models tab? Those should have the same format as well. Also, remove the ability to click an arrow to view the list of quants; that is what the details panel is for.
2. Ensure that all lists and list row elements actually use the same CSS so that that they don't drift; there shouldn't be a unique style defined for each tab.
3. Show the models tab (catalog, and online search results) screenshots so we can assess consistency visually.

You should merge https://github.com/lemonade-sdk/lemonade/pull/3009 into this PR since I don't think we can resolve those issues separately.

jeremyfowers:
Depends on https://github.com/lemonade-sdk/lemonade/pull/3024

I am reviewing that one right now

fl0rianr:
Need to include fixes and rebase.

fl0rianr:
@jeremyfowers correctly rebased now you can review

jeremyfowers:
I just pulled, built, and launched, it looks like there are multiple styles still

image

image

image

image

Looking at these images, there's a lot of things that still need to be unified. Here is one concrete example: the status indicator:

- Regular models have status dot on the bottom-right. No pill background, no words.
- Online search has no status indicator, but that's ok
- Chat has it in the bottom-left, with a pill background and words
- Telemetry has it in the metadata content on the left (but not the bottom-left corner), no pull background, yes words. 


**Every shared element of every list row needs to be of the same style and same layout.** There's too many differences to type them all right now. Do you want to take a comprehensive pass? I could also take over this one.

fl0rianr:
Thanks, you can have a go.

My 2 cents:

They are quite alike, if you choose a list element it's highlighted (pill background upon selection).

Every list has a different intend ans the model list es a bit special an there the line makes sense.
I would avoid cramping all together by force since the are doing separate things but status to the right makes sense to me.

The real outsider here is the search result in the model list. Showing the cloud icon many time is to much.
It would be a good idea to have the result one whole section with one cloud icon and the result in the regular list style - pill when selected.

My specific idea on the special list on the details panel (quant list) I would keep unless there is a better idea.

jeremyfowers:
Ok I will work on this, I am taking over this branch and will push updates here.

> They are quite alike

The key to a professional design is to have every detail consistent and intentional. Having things like status, "Just now", etc. move around is unnecessary and detracts from this.

fl0rianr:
You're better suited doing that

jeremyfowers:
Making good progress, still work in progress

jeremyfowers:
1-models-catalog
2-registry-results
3-chat-history
4-composer-model-picker
5-telemetry-requests
6-monitor-loaded-models

Rose22:
these screenshots look *much* nicer to me imho. nice and unified. consistent design helps a lot :)

PCAssistSoftware:
I agree, the screenshots definitely make everything look more consistent and unified

fl0rianr:
I'm the culprit and will solve those merge conflicts

PCAssistSoftware:
1) Is it just me or does the highlighted part look a bit messy / overcomplicated?  so many different pills

image


2) Also not sure if relevant to this PR or not (but as it appears to have changed since I applied this PR I thought I would mention it here)

Basically same point as made here (https://github.com/lemonade-sdk/lemonade/issues/2947), but now seems to be showing less e.g. in the model list (left column) most models only now showing a Vision icon, and are missing other icons e.g. Chat,  Tool Use and Reasoning.   They also don't show in right hand pane amongst all the other many pills


3) Since applying this PR, my existing model in /home/darren/.cache/huggingface/hub is no longer showing within Lemonade at all


As I say these may not be relevant, as PR not finished, but as occurred after applying this PR to take a look, I wanted to mention it - just in case

fl0rianr:
I would opt for approve since are those are follow up issues.

The point with only vision is a server side issue. We have code to detect some things but it's not easy. Visions works since there is a separate file for it. Also follow up.

jeremyfowers:
@PCAssistSoftware I opened https://github.com/lemonade-sdk/lemonade/issues/3084 for your first part

Your second point seems to be a request for richer metadata on online search models, please open an issue if thats the case.

For your third point if its reproducible please open an issue.

jeremyfowers (approved):
self approve!

PCAssistSoftware:
> @PCAssistSoftware I opened #3084 for your first part
>
Thanks
>
> Your second point seems to be a request for richer metadata on online search models, please open an issue if thats the case.
>
Already covered by https://github.com/lemonade-sdk/lemonade/issues/2947 I think
>
> For your third point if its reproducible please open an issue.
>
Yes reproduced, so will open one now - https://github.com/lemonade-sdk/lemonade/issues/3091

PR #3036, opened 2026-08-10
Add support for a default model source

superm1:
The default will be HuggingFace.

If a user is in China they can set the default model source to Modelscope they can change the configuration option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2820,6 +2820,13 @@ if(BUILD_TESTING AND EXISTS "${_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC}")
     )
 endif()
 
+set(_CONFIG_DEFAULT_SOURCE_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_config_default_source.cpp")
+if(BUILD_TESTING AND EXISTS "${_CONFIG_DEFAULT_SOURCE_TEST_SRC}")
+    add_executable(test_config_default_source test/cpp/test_config_default_source.cpp)
+    target_link_libraries(test_config_default_source PRIVATE lemonade-server-core)
+    add_cpp_ci_test(ConfigDefaultSourceTest CI ON COMMAND test_config_default_source)
+endif()
+
 # ROCm root resolution (ROCM_PATH / rocm-sdk / /opt/rocm priority): covers the
 # external-ROCm detection that lets Lemonade skip the bundled TheRock download.
 set(_ROCM_ROOT_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_rocm_root_resolution.cpp")
--- a/docs/dev/model-registries.md
+++ b/docs/dev/model-registries.md
@@ -12,6 +12,10 @@ Remote provenance is stored separately from local origin:
 
 Hugging Face is the default for backward compatibility. A model and all of its checkpoints have one registry source. Collection components inherit the collection source unless an inline component definition explicitly declares another source.
 
+## Default source policy
+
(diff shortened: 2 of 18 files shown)


fl0rianr (changes requested):
I don't think this is ready to merge yet. A few cases need to be addressed:

1. **Desktop/Web manual model flow ignores `default_model_source`.** `AddModelPanel` still defaults `source` to `huggingface` and always sends it with `/pull`, so a server configured with `default_model_source=modelscope` will still use Hugging Face. The source should be omitted unless the user explicitly selects one, so the server default can apply.

2. **The server currently treats every non-empty checkpoint as a remote registry checkpoint.** This also affects self-managed backends such as FLM (`gemma3:4b`), which are not Hugging Face/ModelScope repo IDs. We shouldn't persist a registry source for checkpoints that are not actually registry-backed.

3. **Provider URL override is only handled by the CLI, not by the server.** The docs say provider URLs override the configured default and that source resolution is server-side, but `/pull` itself does not normalize/detect Hugging Face or ModelScope URLs. Either move that detection into shared/server code or adjust the documented contract.

4. **Docs still describe Hugging Face as the unconditional default.** After this change it is only the shipped/config default; source-less pulls should use the server's configured `default_model_source`.

5. **Tests don't cover the main `/pull` persistence path.** Please add coverage for source-less `/pull`, explicit override, provider URL, non-registry/self-managed checkpoints, and ideally the manual frontend flow so this behavior cannot regress.

The config validation and CLI `/pull/variants` handoff look good, but I would fix the cases above before merging.

fl0rianr (commented):
Thanks for addressing the previous review. The main issues are much improved, but I still see a few correctness problems before this is ready to merge:

1. **checkpoint vs checkpoints.main precedence is inconsistent.** `apply_default_pull_source()` checks `checkpoint` first, while `ModelManager` uses `checkpoints.main` whenever `checkpoints` is present. This can resolve/persist the registry from one checkpoint while actually downloading another. The helper should follow the same precedence as the existing model registration path, with a regression test covering both fields being present.

2. **Provider URL + explicit source is inconsistent across endpoints.** `/pull/variants` treats the provider URL as authoritative, while `/pull` keeps an explicitly supplied `source` even if it conflicts with the URL. The CLI rejects this conflict. I think the server should do the same and return `400` for conflicting `source`/`registry_source` vs provider URL.

3. **Only the primary checkpoint URL is normalized.** In multi-checkpoint registrations, URLs in entries such as `mmproj`, `vae`, or `text_encoder` remain raw and can later be parsed incorrectly. Either normalize every registry-backed checkpoint and enforce one registry per model, or reject provider URLs in secondary checkpoints.

There are also a few remaining docs references that still describe Hugging Face as the unconditional default; it is now only the shipped server default, while the effective default is configurable.

The new tests and Desktop/Web changes are good improvements, but I would still address the three cases above before merging.

fl0rianr (commented):
Thanks, the previous issues are fixed now. Round three is there:

1. **URL/source conflicts are still handled differently depending on the path.** Manual `--checkpoint` rejects a conflicting `--source`, but a normal `lemonade pull --source modelscope https://huggingface.co/...` lets the URL silently win. `/pull/variants` behaves similarly. Please use the same conflict rule everywhere, ideally rejecting mismatches consistently.

2. **A provider URL can currently hide an invalid explicit source.** In `/pull`, something like an HF URL with `source: "nexus"` gets normalized to `source: "huggingface"` before the normal source validation runs, so the invalid value is never rejected. Explicit `source`/`registry_source` should be validated first, or only filled from the URL when those fields are actually absent.

It would also be good to remove the duplicate provider-prefix detection in `main.cpp` and use the shared `detect_registry_url()` helper there too.

The rest of the previous feedback looks addressed.

fl0rianr (commented):
Thanks, the previous issues are addressed. I still see two small correctness gaps:

1. **Non-string `source` values can still be hidden by URL normalization.** For example, `source: 123` with an HF URL is not rejected early because validation only runs for strings, and the URL path can then replace it with `huggingface`. Please reject non-string `source`/`registry_source` values before normalization.

2. **`/pull/variants` rejects valid source aliases when a provider URL is used.** The conflict check compares the raw query value directly, so `source=hf` with a Hugging Face URL returns 400 even though `hf` is otherwise accepted. Please canonicalize through `parse_remote_registry_source()` before comparing.

Also, the PR currently conflicts with the latest `main`, so it will need a rebase/update and another CI run after resolving that.

Everything else from the previous review looks good.

fl0rianr (approved):
Thanks, LGTM.

PR #3037, opened 2026-08-10
fix(gui3): model detail panel refinement

### LINKED ISSUE #2935: Model detail panel refinement
Image

1. Non-LLM models should not display context size controls
2. Remove the Configure button, it doesnt do anything
3. Make the delete button quieter
4. Add a Pin button
5. Remove duplicate Load button
6. Checking "auto tune context size" should be enabled by default, and should hide the slider / text entry controls when checked.

fl0rianr:
image

1. Non-LLM models have no ctx size control
2. Removed configure button
3. Made the delete button quieter
4. Add a Pin button with the same desgin as favorites but not filled upon acivation
5. Removed the duplicate Load button (bottom)
6. Default "auto tune context size"  hides ctz settings

image
7. Save button becomes primary action once configuration is modified.
image

### DIFF
--- a/src/app/src/components/ModelDetailPanel.tsx
+++ b/src/app/src/components/ModelDetailPanel.tsx
@@ -1001,17 +1001,14 @@ const ModelConfigurationTab: React.FC<{
   loadedModel: LoadedModel | null;
   isActive: boolean;
   serverDefaultCtxSize: number;
-  isDownloaded?: boolean;
   isLoadingThis?: boolean;
   onReloadModel?: (model: LoadedModel, recipeOptions?: Record<string, unknown>) => Promise<void>;
-  onLoadWithOptions?: (recipeOptions: RecipeOptions) => Promise<void>;
   onDirtyChange?: (dirty: boolean) => void;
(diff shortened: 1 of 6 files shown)


jeremyfowers:
This looks really good, will test it locally

jeremyfowers:
image

image

We don't need a note and a tooltip, eliminate one of them.

jeremyfowers:
image

Remove redundant "Estimated at load time"

jeremyfowers:
Opened related issue: https://github.com/lemonade-sdk/lemonade/issues/3070

This prevents the context slider from showing on some LLMs; let's fix that in a separate PR.

jeremyfowers:
image

The "saved defaults" on here confuses me and needs to be fixed.

1. A default should be immutable, so "saved defaults" seems like a contradiction
2. If I change a setting, it still displays "saved defaults" even though that is no longer accurate; I'm literally no longer looking at the saved values nor the defaults

image

The button and text should just save "Save" and "Saved for future loads".

jeremyfowers (changes requested):
Big improvement over upstream, but more work is needed. See comments above.

fl0rianr:
solved all, opted for tool tip since it's the cleaner look and made it situation aware. BUT we need to design the tool tips, a round edge should to the trick (out of scope here).

jeremyfowers (approved):
Merging and will make a follow up PR

PR #3039, opened 2026-08-10
Debian packaging improvements

superm1:
## Summary
Debian packaging updates for lemonade.

- Drop quilt from build-depends (Closes: #1143219)
- Reformat d/control using cme
- Remove unnecessary licenses flagged by cme from d/copyright
- Bump standards version

### DIFF
--- a/contrib/debian/control
+++ b/contrib/debian/control
@@ -1,5 +1,5 @@
 Source: lemonade
-Standards-Version: 4.7.3
+Standards-Version: 4.7.4
 Maintainer: Mario Limonciello <superm1@debian.org>
 Section: utils
 Testsuite: autopkgtest-pkg-python
--- a/contrib/debian/control
+++ b/contrib/debian/control
@@ -1,7 +1,8 @@
 Source: lemonade
-Section: utils
-Priority: optional
+Standards-Version: 4.7.3
 Maintainer: Mario Limonciello <superm1@debian.org>
+Section: utils
+Testsuite: autopkgtest-pkg-python
 Build-Depends:
  cmake,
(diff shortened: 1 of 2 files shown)


superm1:
This aligns with what is in Salsa.

https://salsa.debian.org/debian/lemonade

superm1:
Ken is out this week, @fl0rianr could you review this?  It's mostly just aligning what's in Debian.

contrib/debian/control
```diff
@@ -1,7 +1,8 @@
 Source: lemonade
-Section: utils
-Priority: optional
```
fl0rianr, line 3: I think we should keep Priority: optional here for Ubuntu 24.04 compatibility. The implicit optional default was only added in dpkg 1.22.13, while Noble still ships dpkg-dev 1.22.6. Since our packaging workflow explicitly targets Noble and the actual distro build jobs were skipped on this PR (merge queue job - had this with an GCC version error on anohter PR before), this change hasn't been exercised there. Could we either keep the field or run ci:distros and verify the Noble package output?

fl0rianr (changes requested):

contrib/debian/control
superm1: Will trigger ci:distros to confirm.

superm1: @fl0rianr looks like the noble build is fine.

fl0rianr (approved):
LGTM, thanks!

To be fair, I was not totally wrong:
dpkg-genchanges: warning: missing Priority for binary package lemonade-desktop; using '-'

PR #3042, opened 2026-08-10
GUI3 fix: backend hide technical details

fl0rianr:
currently show technical details is useless, if applied:
image
If not no meaningfull difference:
image

With this change the user is now protected from detail he might not want to know:
image
Or he can opt it to know more:
image

### DIFF
--- a/src/app/src/components/BackendManager.tsx
+++ b/src/app/src/components/BackendManager.tsx
@@ -1111,32 +1111,36 @@ const BackendManager: React.FC<BackendManagerProps> = ({ isActive = true }) => {
                 </div>
 
                 <div className="backend-card__variant-meta">
-                  {version && (installedReleaseUrl ? (
-                    <a
-                      className="backend-card__version"
-                      href={installedReleaseUrl}
-                      target="_blank"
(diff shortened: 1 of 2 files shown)


jeremyfowers:
I'm not in favor of this change, the version number and link to upstream are valuable information for this screen to show.

fl0rianr:
should we the get rid of this useless button or turn it around like hide details? I get why this is interesting but one core idea is it should not matter to a certain portion of users. Current state is not desiderable IMO.

jeremyfowers:
This entire backends tab is optional and only needed for advanced users, though. A naive/new user can get almost all tasks done from the Chat and Models tab because loading a model will automatically set up the right backend. Therefore the backends tab is allowed to be more technical.

Another core idea I have here is that if we are showing an update button we need to also show that people are updating from what to what.

jeremyfowers:
Removing this from release blockers; we can keep iterating, I just dont think of this as a blocker.

kpoineal:
I agree we can just ignore this for now. I'd like to get Meghan's take on some of this before implementing.

PR #3043, opened 2026-08-10
GUI3 fix: use eject icon for model detail unload action

### LINKED ISSUE #3034: Broken/missing unload icon on models detail panel
Image

Make sure its identical to the model list in the chat panel:

Image

fl0rianr:
image
image

image

Used the eject icon since stop in the chat can be confused with stopping the response.
Opted for double stroke width so the icon is clearly visible.

Closes #3034


jeremyfowers (approved):
looks good, thanks!

PR #3047, opened 2026-08-11
feat(backends): ds4 (DwarfStar) backend for DeepSeek V4 Flash [experimental]

ramkrishna2910:
## What

Adds an **experimental** `ds4` recipe that serves DeepSeek V4 Flash through [antirez/ds4](https://github.com/antirez/ds4) (DwarfStar 4), a small self-contained C inference engine written specifically for that model. `Ds4Server`wraps `ds4-server` as a subprocess and forwards chat/completions/responses to its OpenAI-compatible API.

image

## Why

ds4 has a first-class `make strix-halo` ROCm/gfx1151 backend and is the only engine that runs the DeepSeek V4 Flash GGUFs (they use an asymmetric quant that llama.cpp cannot read). It also supports distributed inference across machines,
which opens up quants larger than one box can hold.

## Testing

Verified end to end on Strix Halo (Ryzen AI Max+ 395, gfx1151, ROCm 7.1) with
`DeepSeek-V4-Flash-IQ2XXS-...-0731` (81 GiB):

| step | result |
| --- | --- |
| model listed | `DeepSeek-V4-Flash-DS4` ok |
| load, ctx 32768 | 21.4 s, 100 GiB GTT |
| chat completion | correct |
| tool calling | clean `get_weather{"city":"Austin"}` |
| unload | GTT back to 0, `ds4-server` reaped |

Performance through lemonade is indistinguishable from driving `ds4-server`
directly (raw: 16.45 t/s decode, ~200 t/s prefill):

| prompt | decode | prompt processing |
| ---: | ---: | ---: |
| 26 tokens | 16.34 t/s | overhead-dominated |
| 1167 tokens | 16.11 t/s | 216 t/s |

## Known limitations

1. ds4 binaries now come from [lemonade-sdk/ds4-rocm](https://github.com/lemonade-sdk/ds4-rocm), which builds a pinned upstream commit and bundles the ROCm runtime, so users no  longer build ds4 themselves. `get_install_params()` resolves the release asset for the detected GPU arch, pinned to `b0001` with an asset checksum like every other backend. `rocm` is the only variant.

   Upstream has no releases, tags or build CI, so release identity is that repo's own `bNNNN` build number, with the ds4 commit SHA recorded in the release notes.

2. **Streaming is bursty.** For this backend lemonade delivers its deltas in one burst at the end rather than incrementally, so a streaming client sees no tokens until generation finishes. Throughput is unaffected, but it feels
   unresponsive in a streaming UI. Worth a separate look.

3. **Distributed is untested through lemonade.** `ds4_args` is a verbatim argv passthrough, so `--role coordinator --layers 0:20 --listen ...` should need no  code change, with the peer worker run out of band. Not yet exercised here.

## Notes

`server_models.json` registers `DeepSeek-V4-Flash-DS4` against the `-0731` checkpoint (the current generation, and the one benchmarked above).


fl0rianr (commented):
Thanks for putting this together — overall the integration looks clean.

I found two things I think should be fixed before merge:

1. The managed `ds4-rocm` bundle ships its ROCm `.so` files alongside `ds4-server`, but the wrapper starts the binary without adding its directory to `LD_LIBRARY_PATH`. The ds4-rocm README explicitly requires this, and llama.cpp already does the same setup for bundled ROCm/CUDA runtimes. This may work on the E2E machine because system ROCm is installed, but fail on the intended clean host without ROCm.

2. `ds4_args` is appended after Lemonade's own `-m`, `--host`, `--port`, and `-c` arguments. DS4 parses these left-to-right, so custom args can override Lemonade-managed values. In particular, overriding `--port` breaks readiness/proxying, `--host 0.0.0.0` could expose the backend directly, and `-m` can make Lemonade's loaded-model state disagree with the actual process. I'd reuse the existing `validate_custom_args()` / reserved-flags pattern from llama.cpp for these arguments.

Two small cleanup items as well: the descriptor comment still describes the removed PATH/system-backend approach, and `selectable_backend=true` is a little inconsistent with there being no declared `ds4_backend` option. The latter isn't a functional problem while ROCm is the only variant, but worth making consistent.

Other than those points, I didn't find anything else that would block this experimental backend. The readiness/process lifecycle, gfx1151 gating, install asset resolution, and generic streaming proxy path all look reasonable.

ramkrishna2910:
Thanks — both of these were real, and the first one was a genuine hole in my testing. Fixed in `9c3f0b4`.

**1. Bundled ROCm runtime was not being used.** You were right, and it was worse than "may work by accident": I checked, and `ds4-server` has no `RPATH`/`RUNPATH`, so the child was resolving `libamdhip64` / `libhipblaslt` / `librocblas` from `/usr/lib/x86_64-linux-gnu`. My end-to-end test passed only because this host has system ROCm installed. On a clean host it would not have started at all.

`load()` now prepends the binary's directory to `LD_LIBRARY_PATH` when spawning, following the `kokoro`/`thinksound`/`trellis` pattern. Verified from the running process:

```
LD_LIBRARY_PATH=/home/.../.cache/lemonade/bin/ds4/rocm
/home/.../bin/ds4/rocm/libamdhip64.so.7
/home/.../bin/ds4/rocm/libhipblaslt.so.1
/home/.../bin/ds4/rocm/liborigami.so.1
```

`liborigami.so.1` is the useful signal: it exists only in the bundle (TheRock's hipBLASLt pulls it in; distro ROCm has no such library), so its presence proves the bundled runtime is what is loaded. Before the fix it was not mapped at all.

**2. `ds4_args` could override managed flags.** Also correct — ds4 parses left-to-right, so anything appended after wins. Now validated with the existing `validate_custom_args()` helper against `{-m, --model, --host, --port, -c, --ctx}`:

```
--port 9999        -> Invalid custom ds4-server arguments:
                      Argument '--port' is managed by Lemonade and cannot be overridden.
                      Reserved arguments: --ctx, --host, --model, --port, -c, -m
--host 0.0.0.0     -> rejected
-m /tmp/other.gguf -> rejected
-c 999999          -> rejected
--power 100        -> accepted, reaches the process
```

**3. Stale descriptor comment.** Fixed — it still described the removed PATH lookup and claimed the backend has no version pin.

**4. `selectable_backend=true`.** I looked into this one and I think it is already consistent, so I have left it: `selectable_backend` *auto-creates* the `ds4_backend` option rather than requiring it to be declared, and `defaults.json` does contain `ds4.backend: "auto"`, which `load()` reads. `thenoise` sets the identical combination — `selectable_backend=true` with `rocm` as its only variant. Happy to flip it to `false` if you would rather it stay off until a second variant exists, but as written it matches the closest precedent and means adding a cuda/metal variant later needs no descriptor change.

Docs regenerated (`--check` passes) and the unit test still passes.

fl0rianr (commented):
Thanks for the follow-up. I re-checked the current head and the two original blockers look fixed now. The bundled ROCm runtime is actually used, and the managed model/host/port/context args can no longer be overridden.

I found one remaining issue I think is worth fixing before merge:

`ds4_args` can still override the compute backend itself. The reserved set currently covers `-m/--model`, `--host`, `--port`, and `-c/--ctx`, but ds4-server also accepts `--backend`, `--cpu`, `--metal`, and `--rocm`.

For example, `--ds4-args "--cpu"` would make the child run on CPU while Lemonade still tracks this as the `rocm` backend with `DEVICE_GPU`. That gives us the same kind of state mismatch we were trying to prevent for `-m`. I think those backend-selection flags should be reserved too.

I'd also strongly prefer a small regression test for the reserved custom args. The current DS4 test only covers install params / arch gating, so neither of the two review fixes is exercised automatically.

One smaller thing: the `selectable_backend=true` explanation doesn't quite match the current code. TheNoise explicitly declares a `thenoise_backend` option with `--thenoise`; DS4 currently only declares `ds4_args`. Since ROCm is the only variant this isn't blocking today, but I'd still make that consistent while we're here.

With the backend-override case fixed and the regression coverage added, I don't see another code-level blocker.

ramkrishna2910:
All three addressed in `35f748e4`.

**1. Backend-selection flags.** Confirmed against `ds4-server --help`:

```
--metal | --rocm | --cpu       Select the backend explicitly.
--backend NAME                 Backend name: metal, rocm, or cpu.
```

So `--ds4-args "--cpu"` really would have run the child on CPU while Lemonade tracked it as `rocm`/`DEVICE_GPU`. Those are reserved now (`--backend`, `--cpu`, `--metal`, `--rocm`, plus `--cuda` for when a CUDA build exists):

```
--cpu         -> Argument '--cpu' is managed by Lemonade and cannot be overridden.
--backend cpu -> Argument '--backend' is managed by Lemonade and cannot be overridden.
--rocm        -> Argument '--rocm' is managed by Lemonade and cannot be overridden.
--power 100   -> still accepted, reaches the process
```

The reserved set moved into the descriptor header as `ds4::reserved_custom_arg_flags()` so it is one list rather than a literal buried in `load()`, and so the test can assert on it.

**2. Regression coverage.** You were right that neither fix was exercised. `test_ds4_install_params` now covers both, 22 assertions:

```
PASS: rejects managed argument -m / --model / --host / --port / -c / --ctx
PASS: rejects backend-selection argument --backend / --cpu / --metal / --rocm / --cuda
PASS: allows arguments Lemonade does not manage
PASS: allows empty custom arguments
```

It runs under `cpp-ci`.

**3. `selectable_backend`.** You are right and my previous explanation was wrong. I had checked `defaults.json`, saw `ds4.backend: "auto"` appear, and concluded the option was auto-declared — but that flag only creates the *config key*; TheNoise separately declares `thenoise_backend` with `--thenoise` to get the CLI option, which DS4 never did. So it was advertising a knob with no way to set it.

Since `rocm` is the only published variant I have set `selectable_backend=false` and `load()` now uses `rocm` directly, which also drops the stray `backend` key from `defaults.json`. When a second variant lands, that is the point to declare the option properly.

Docs regenerated (`--check` passes).

fl0rianr (commented):
Thanks, the previous points look good now. The backend-selection args are reserved, the regression coverage is in place, and `selectable_backend` is consistent now.

I found one remaining thing before I'd approve:

`get_install_params()` currently builds the release asset name directly from `SystemInfo::get_rocm_arch()` without validating that the architecture is actually supported. The descriptor only supports `gfx1151`, but e.g. `gfx1150` would still produce `ds4-b0001-linux-rocm-gfx1150-x64.tar.gz` and proceed toward an asset that doesn't exist.

The normal model-filtering path catches unsupported systems, but the backend install path itself doesn't guarantee that this check has happened first. Since `RuntimeConfig::validate_backend_choice()` was removed when the backend became non-selectable, I'd make `get_install_params()` fail explicitly for unsupported/empty architectures and add a small test for that.

Other than this, the review issues I raised look resolved. Once the arch gating is hardened, I'm happy with this from the code side.

ramkrishna2910:
Good catch — and it was self-inflicted: dropping `validate_backend_choice()` along with `selectable_backend` removed the only thing standing between an unsupported host and a composed asset name. Fixed in `a1f7c2e`.

`get_install_params()` now checks the detected architecture against the descriptor's support matrix before building a filename:

```cpp
if (target_arch.empty()) {
    throw ... "requires a ROCm GPU, but no ROCm architecture was detected";
}
if (!SystemInfo::backend_supports_arch("ds4", "rocm", target_arch)) {
    throw ... "publishes no build for " + target_arch + "; only gfx1151 is validated upstream";
}
```

Using `backend_supports_arch` rather than a hardcoded string keeps this honest when ds4-rocm starts publishing a second target — the support row stays the single source of truth.

Tests added (25 assertions total now):

```
PASS: install refuses an unsupported architecture      (gfx1150)
PASS: install refuses a CDNA architecture              (gfx942)
PASS: install still resolves for the supported architecture
```

One gap worth flagging rather than hiding: I could not add a test for the empty-architecture branch. `set_rocm_arch_override("")` *clears* the override instead of setting an empty value, so the probe falls back to real hardware and the case is unreachable from a test on a GPU host. The runtime guard is still there and correct for a genuine no-ROCm machine; there is a comment at the call site saying why it is untested rather than leaving it looking overlooked.

Re-verified on a real gfx1151 host that a normal load still succeeds, and docs `--check` passes.

fl0rianr (commented):
Thanks, the arch gating fix looks good.

I found one last edge case in the install path: backend_supports_arch() checks the device/arch constraints, but not the descriptor's supported_os. Since the direct backend install endpoint calls install_backend() without going through model filtering, a Windows gfx1151 host could still pass the arch check and resolve the Linux asset.

Could we also reject unsupported OSes before composing the asset name? After that tiny point, I'm happy to approve.

docs/assets/models.js
```diff
@@ -28,6 +29,7 @@ const RECIPE_DISPLAY_NAMES = {
   'ryzenai-llm': 'Ryzen AI SW NPU',
   vllm: 'vLLM ROCm (experimental)',
   thenoise: 'thenoise',
+  ds4: 'DS4 DwarfStar (experimental)',
```
jeremyfowers:
```suggestion
  ds4: 'DwarfStar4 (experimental)',
```

The name "DS4 DwarfStar" feels redundant to me, the "DS" means "DwarfStar"

src/cpp/resources/server_models.json
```diff
@@ -2009,5 +2009,15 @@
...
+    },
+    "DeepSeek-V4-Flash-DS4": {
+        "checkpoint": "antirez/deepseek-v4-gguf:DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf",
+        "recipe": "ds4",
+        "suggested": true,
+        "labels": [
+            "reasoning",
+            "tool-calling"
+        ],
+        "size": 81.0
```
jeremyfowers: This is getting filtered out on my 128 GB linux strix halo, presumably because Lemonade thinks it wont fit?

I'm unable to test due to this.

src/cpp/resources/server_models.json
```diff
@@ -2009,5 +2009,15 @@
             "classification"
         ],
         "size": 0.27
+    },
+    "DeepSeek-V4-Flash-DS4": {
```
jeremyfowers:
```suggestion
    "DeepSeek-V4-Flash-IQ2XXS-DS4": {
```

Since it breaks our q4 convention.

src/cpp/resources/server_models.json
fl0rianr: How good you are using windows!!!

jeremyfowers: Why

- DeepSeek-V4-Flash-DS4 declares "size": 81.0 in src/cpp/resources/server_models.json:2013.
- The cap is max(largest_mem_pool_gb, 0.8 × system_ram_gb) — src/cpp/server/model_manager.cpp:2611. Here that's max(64.0 GB VRAM pool, 49.1 GB) = 64.0 GB, and 81 > 64, so it's dropped at cache-build time (model_manager.cpp:2685-2697).
- Filtering happens before the model cache is populated, so it's invisible to both the default listing and ?show_all=true — I confirmed: 45 models by default, 150 with show_all=true, DS4 in neither.

jeremyfowers (changes requested):
It needs to work on my 128 GB strix halo

src/cpp/resources/server_models.json
ramkrishna2910: This logic of size filtering does break for backends that support nvme streaming of experts.

ramkrishna2910:
Thanks both — all three points are addressed.

## @fl0rianr — unsupported OS in the install path

Correct, and confirmed by reading `backend_supports_arch()`: it validates the device constraints and `arch_gates`, but never the support row's `supported_os`. A Windows gfx1151 host would have passed the arch check and resolved the Linux-only asset. Fixed in `7676031`:

```cpp
if (!ds4::publishes_for_os(get_current_os())) {
    throw ... "publishes no build for " + current_os + "; Linux only";
}
```

It reads the descriptor's support row rather than hardcoding `"linux"`, so it stays correct if a Windows build is ever published. Exposed as `ds4::publishes_for_os()` so it is testable without faking the OS:

```
PASS: publishes for linux
PASS: does not publish for windows
PASS: does not publish for macos
```

28 assertions in `test_ds4_install_params` now.

## @jeremyfowers — filtered out on your 128 GB Strix Halo

Your diagnosis was exactly right, and the cause is now fixed on this branch — but by a commit that did not exist when you tested. Timing:

* your comment: **2026-08-12 13:28 UTC**
* `f6016c7e fix(system-info): count the iGPU carveout as physical memory on Linux`: authored **2026-08-12 22:09 UTC**, ~9 hours later, and merged here in `b681589f`

Your box has a 64 GB BIOS VRAM carve-out, and an integrated GPU's "VRAM" is system DRAM the firmware reserved before the kernel booted, so it never appears in `MemTotal`. Lemonade saw ~61 GB instead of ~125 GB, giving `max(64.0, 0.8 × 61.4) = 64.0` — and 81 > 64, so the model was dropped at cache-build time exactly as you traced.

With the carve-out counted, your box should report ~125 GB, so the cap becomes `max(64.0, ~100.3) = 100.3` and the 81 GB model passes.

I verified the arithmetic on this machine rather than assuming: `MemTotal/1024/1024` (122.79) + carveout (0.54) = **123.33 GB**, which is exactly what `/api/v1/system-info` now reports here, and the log line reads `System RAM: 123.3 GB (max model size: 98.7 GB)`.

I could not reproduce your configuration directly — this box is set up the other way round (minimal BIOS VRAM, GTT raised to 124 GiB), so its largest pool is 0.5 GB rather than 64 GB. If you pull the branch and it still filters, that is a real second bug and I would like the `System RAM:` / `Largest memory pool:` lines from your startup log.

Also applied both of your suggestions: the display name is now `DwarfStar4 (experimental)` and the model is `DeepSeek-V4-Flash-IQ2XXS-DS4`.

## On the size filter and SSD streaming

Agreed that the rule is wrong in principle for backends that stream experts from NVMe — ds4 has `--ssd-streaming`, so the resident working set can be far below the file size, and a pure `size > RAM` test will keep filtering models such a backend could run. I would rather not change core filtering inside this experimental-backend PR; happy to open a separate issue for it so it gets its own discussion.

## Re-verified on the merged branch (Linux, gfx1151, 128 GB)

* full `cpp-ci` suite: **34/34 pass**
* load 18.8 s, chat correct, tool calling correct, unload releases all GTT
* the managed binary runs with the bundled runtime (`liborigami.so.1` mapped from the bundle, a library system ROCm does not ship)
* decode 15.98 t/s, prompt processing 216 t/s @1.2k — unchanged from before the merge

fl0rianr (approved):
Looks good to me now. The two earlier blockers are addressed cleanly: `route_only` now fails closed on invalid types, and `routed: false` is limited to registered non-router models instead of masking lookup/dispatch failures. The added tests cover both cases, and the API/schema docs were updated accordingly.

A few small non-blocking nits:

* In `server.h`, the comment describing `apply_router_collection_dispatch()` now sits directly above `respond_route_only()`, so it reads like it belongs to the wrong function.
* `src/cpp/resources/schemas/README.md` still says “OpenAI chat body”, although the request schema now also applies to `completions` and `responses`.
* The docs say a dispatch failure returns “the same error a real request would”, while the dry-route path currently normalizes that case to a generic `500 server_error`. I think the behavior is fine, but wording like “returns an error” would be more precise.

None of these are blockers for me. Jeremy is blocking this anyway until it works on his machine. So from my side ok to say Approving.

src/cpp/server/system_info.cpp
```diff
@@ -3733,6 +3733,51 @@ double LinuxSystemInfo::get_amd_vram(const std::string& drm_render_minor) {
     return parse_memory_sysfs(drm_render_minor, "mem_info_vram_total");
 }
 
+double LinuxSystemInfo::get_igpu_carveout_gb() {
```
jeremyfowers: @superm1 can you review this?

jeremyfowers:
```
2026-08-13 13:10:59.737 [Debug] (ds4 Server) Found executable at: /home/jfowers/.cache/lemonade/bin/ds4/rocm/ds4-server
2026-08-13 13:10:59.738 [Debug] (WrappedServer) ds4-server will use port: 8001
2026-08-13 13:10:59.738 [Info] (DS4) Starting ds4-server (/home/jfowers/.cache/lemonade/bin/ds4/rocm/ds4-server) for /home/jfowers/.cache/huggingface/hub/models--antirez--deepseek-v4-gguf/snapshots/e7f04037032990db0346398d249baf9fb9df1ccc/DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf on port 8001
2026-08-13 13:10:59.738 [Debug] (ProcessManager) Starting process with filtered output: /home/jfowers/.cache/lemonade/bin/ds4/rocm/ds4-server -m /home/jfowers/.cache/huggingface/hub/models--antirez--deepseek-v4-gguf/snapshots/e7f04037032990db0346398d249baf9fb9df1ccc/DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf --host 127.0.0.1 --port 8001 -c 10970
2026-08-13 13:10:59.742 [Info] (ProcessManager) Process started successfully, PID: 2954633
2026-08-13 13:10:59.742 [Debug] (WrappedServer) Waiting for ds4-server to be ready...
2026-08-13 13:10:59.822 [Info] (Process) ds4: Linux rocm backend set oom_score_adj=1000
2026-08-13 13:11:00.268 [Info] (Process) ds4: ROCm backend initialized on AMD Radeon 8060S Graphics (sm_115)
2026-08-13 13:11:00.282 [Info] (Process) ds4: ROCm preparing model tensor mappings
2026-08-13 13:11:00.387 [Info] (Process) ds4: ROCm loading model tensors into device cache
2026-08-13 13:11:04.673 [Info] (Process) ds4: ROCm loading model tensors 16.02 GiB cached
2026-08-13 13:11:04.758 [Info] (Process) ds4: ROCm prepared model tensor mappings 16.37 GiB
2026-08-13 13:11:08.794 [Info] (Process) ds4: ROCm loading model tensors 32.06 GiB cached
2026-08-13 13:11:08.799 [Info] (Process) ds4: ROCm prepared model tensor mappings 32.07 GiB
2026-08-13 13:11:09.872 [Debug] (WrappedServer) Still waiting for ds4-server...
2026-08-13 13:11:13.832 [Info] (Process) ds4: ROCm loading model tensors 48.02 GiB cached
2026-08-13 13:11:13.931 [Info] (Process) ds4: ROCm prepared model tensor mappings 48.43 GiB
2026-08-13 13:11:18.274 [Info] (Process) ds4: ROCm model arena alloc failed for tensor-span:108 (528.00 MiB request): out of memory
2026-08-13 13:11:18.274 [Info] (Process) ds4: accelerator failed to prepare model tensor span 108 at offset 65244457792
2026-08-13 13:11:18.274 [Info] (Process) ds4: rocm failed to prepare optional model cache
2026-08-13 13:11:18.397 [Error] (WrappedServer) ds4-server process has terminated with exit code: 1
2026-08-13 13:11:18.397 [Error] (WrappedServer) This usually means:
2026-08-13 13:11:18.397 [Error] (WrappedServer)   - Missing required drivers or dependencies
2026-08-13 13:11:18.397 [Error] (WrappedServer)   - Incompatible model file
2026-08-13 13:11:18.397 [Error] (WrappedServer)   - Try running the server manually to see the actual error
2026-08-13 13:11:18.401 [Error] (Router) Retry also failed: ds4-server failed to start within timeout
2026-08-13 13:11:18.401 [Error] (Router) Failed to load model: ds4-server failed to start within timeout
2026-08-13 13:11:18.401 [Error] (Server) Failed to load model: ds4-server failed to start within timeout
2026-08-13 13:11:18.402 [Error] (Server) Error 500: POST /api/v1/load
2026-08-13 13:11:18.403 [Debug] (Server) POST /api/v1/load - 500
```

doesnt work for me on a basically idle 128 GB strix halo

docs/assets/models.js
```diff
@@ -5,6 +5,7 @@ const RAW_BASE = 'https://raw.githubusercontent.com/lemonade-sdk/lemonade';
 /* BEGIN GENERATED: models-js-recipes */
```
jeremyfowers, line 5: What is the behavior of this backend on a system that doesn't have enough RAM to run the model? Is the entire backend hidden from the user, since it has no supported models? Please share a screenshot on a ~32 GB laptop.

We dont have this problem with any other backend since they all have a model no bigger than 5 GB.

pdittaro:
Regarding the approach to ssd-streaming that was just introduced:

I've also been trying to get ds4-server running on a 128GB Strix Halo (Bosgame M5, gfx1151), which led me here while looking for a way to switch over to lemonade. I also ran into the same error that @jeremyfowers posted when trying to get it running initially. The ~48 GiB ceiling mentioned in the 2d5b1a0 commit isn't a platform limit - it's the BIOS UMA carve-out. 

I believe when the UMA carve-out is larger than the ttm.pages_limit, ROCm defaults to the carve size for the available memory. You can see it in `/sys/class/kfd/kfd/topology/nodes/*/mem_banks/*/properties`:

| BIOS UMA carve | ttm.pages_limit | Heap ROCm reports |
|----------------|-----------------|-------------------|
| 64 GiB         | 40 GiB          | 64 GiB            |
| 1 GiB          | 40 GiB          | 40 GiB            |
| 1 GiB          | 100 GiB         | 100 GiB           |

So the fix is to set the UMA Frame Buffer size to the BIOS minimum, then raise the GTT limits:
```
options ttm pages_limit=26214400      # 100 GiB in 4 KiB pages
options ttm page_pool_size=26214400
options amdgpu gttsize=102400         # MiB
```
I have this set in `/etc/modprobe.d/amdgpu.conf`, you'd also need to run `update-initramfs -u` and reboot afterwards for it to be picked up.

I don't know how something like this would be handled in the backend, other than a sanity check.

But using that ds4-server fully loads the 81 GiB IQ2XXS without --ssd-streaming. Peak GTT was 98.86 GiB against the 100 GiB limit at --ctx 98304; 65536 leaves more headroom. Roughly 150 t/s prefill, 15 t/s decode.

AMD's own Strix Halo guidance says the same about keeping the BIOS reservation small: https://rocm.docs.amd.com/en/docs-7.2.0/how-to/system-optimization/strixhalo.html

docs/assets/models.js
jeremyfowers: Should be solved in https://github.com/lemonade-sdk/lemonade/pull/3219

jeremyfowers:

**Not ready for review yet** — the PR bundles unrelated changes. Fixing the item below is the next step; the documentation, testing, and reviewer checks have not run yet and will follow once this is resolved.

**Attention level:** Elevated — any reviewer can take it, but the to-dos below need resolving before approval (needs 2 reviewers including a subject-area expert; unrelated changes bundled together)

### Unrelated changes bundled together

- [ ] Split the system_info.cpp and system_info.h changes into their own PR.
  - Why: The diff bundles two separate problems: the ds4 backend (primary feature) and a general system_info.cpp fix that counts the iGPU carveout as physical memory on Linux — a bug fix that benefits all backends, not just ds4.

### Review needed

**2 reviewers including 1 subject-area expert** — a new backend. The ds4 backend is a new backend, which the table places on the two-with-expert rung. A reviewer with new backends expertise and one with ROCm expertise are needed.
- Subject-area expertise this calls for: new backends, ROCm

### Not checked yet

The documentation, testing, alignment, and reviewer checks were skipped. Re-request a review once the item above is resolved and they will run.

_Reviewed at head `1921e0a` by repo-manager (tier: triage). Regenerate with `repo-manager review-pr 3047`._

jeremyfowers:
@ramkrishna2910 our contribute.md agent has flagged you :joy: :point_up: 

I'm inclined to agree, it would be nice to have the system-info change in its own PR, reviewed by @superm1, then rebase this PR and we'll go from there.

src/cpp/include/lemon/backends/ds4/ds4.h
```diff
@@ -0,0 +1,83 @@
...
+    /*slot_policy*/     SlotPolicy::Standard,
+    /*selectable_backend*/ false,  // rocm is the only variant; no option to declare
+    /*uses_ctx_size*/   true,
+    /*dynamic_models*/  false,
+    /*options*/ {
+        {"ds4_args", "--ds4-args", "", "ARGS",
+         "Custom arguments to pass to ds4-server", "DS4 Options"},
+    },
+    /*support*/ {
+        {"rocm", {"linux"}, {{"amd_gpu", {"gfx1151"}}}, "Prebuilt ds4 for AMD Strix Halo"},
```
superm1, line 54: Why only gfx1151

ramkrishna2910: Thats all the hardware I have access to for exhaustive testing. If we do have access to more hardware we can expand this list to other AMD machines, Apple silicon and NVIDIA.

ckuethe: because that's what can be and has been easily tested? has anyone tested this on one of those nvidia boxes or apple silicon?

with my end user hat can we stop bike shedding and just ship it? it's not gonna get any realistic testing while it rots in endless reviews.. it's already marked "experimental"

docs/assets/models.js
ramkrishna2910: Verified solved with #3219

src/cpp/include/lemon/backends/ds4/ds4.h
superm1: But do they build for other AMD GFS ISAs?  We should offer everything even if we haven't tested it, otherwise how could a user even test it?

superm1: > with my end user hat can we stop bike shedding and just ship it

This was my first review comment for this PR because I was just requested.  If my review isn't necessary I'll stop adding commentary.

src/cpp/server/system_info.cpp
```diff
@@ -3753,8 +3798,15 @@ std::string LinuxSystemInfo::get_physical_memory() {
         if(token == "MemTotal:") {
             // Get the token after "MemTotal:"
             if(double mem; file >> mem) {
+                // MemTotal excludes an integrated GPU's firmware-reserved
+                // carveout, which on a unified-memory APU is ordinary system
+                // DRAM the model can still be run out of. Report installed
+                // capacity, matching Windows, so model size filtering does not
+                // hide models that fit (a 128 GB Strix Halo with a 64 GB
+                // carveout otherwise reports ~61 GB).
```
superm1: if this backend is using ROCm this is incorrect.  ROCm we don't get to target both pools.  You get the bigger of the two pools.

src/cpp/server/system_info.cpp
superm1: I don't agree with this code.

iswaryaalex:
@ramkrishna2910 Works on my side!
Screenshot from 2026-08-24 10-18-56

src/cpp/resources/server_models.json
ramkrishna2910: Fixed by introducing logic for ssd streaming backends.

jeremyfowers (approved):
Works for me! Two things to be aware of as we move forward, since these will become problems fast:

1. **Context size**. There is no context window slider in the GUI. When I load on my strix halo 128 GB I get "Auto-tune ctx_size resolved to 13660" and I have no clue why. This is going to drive people crazy.
2. **Zero runtime testing**. We have no way to know if this breaks, including on backend tag upgrades or anything else. Consider creating a "tiny ds4" model for testing? Or just biting the bullet and running full sized ds4 in ci:backends?

jeremyfowers:
Opened https://github.com/lemonade-sdk/lemonade/issues/3343 since that's very important for maintainability.

superm1 (approved):

bitgamma:
https://github.com/lemonade-sdk/lemonade/actions/runs/32951134054/job/98122535927

fails because of: 

```
49/58 Test #61: Ds4InstallParamsTest ...................Subprocess aborted***Exception:   0.03 sec
libc++abi: terminating due to uncaught exception of type std::runtime_error: ds4 backend 'rocm' publishes no build for macos; Linux only
```

jeremyfowers:
taking a crack at it

PR #3048, opened 2026-08-11
GUI3: make Lemonade shockingly fast to load

### LINKED ISSUE #3035: Loading GUI3 takes ~10 seconds
This seems like a long time on my high end Strix Halo system. @fl0rianr or @kpoineal would you mind taking a pass to see if there is low-hanging fruit for optimizing this? IMO ideal load time is under 3 seconds.

fl0rianr:
## Summary

This PR significantly speeds up Lemonade GUI startup by keeping only the active workspace on the critical path and moving everything else into ordered background prewarming.

On my setup, the browser UI now reaches the **first usable frame in ~350 ms median**.

## What changed

* Production-optimized and split web renderer bundles
* Only the active workspace blocks initial rendering
* Remaining workspaces and interaction-critical modules are automatically prewarmed after the first frame
* Heavy Monitor views and charts no longer block opening the Monitor tab
* Models use a fast initial model snapshot while the full catalog loads in the background
* Model Details, Telemetry, Logs, Router Editor and Settings are prepared automatically instead of loading on first interaction
* Hidden workspaces are not mounted, avoiding unnecessary effects and polling
* Startup remains covered until styles are ready, preventing unstyled flashes
* Preserves async model-state and accessibility behavior

The result is a fast initial render without simply moving the delay to the first tab or panel interaction.

Tauri uses the same renderer-side startup changes, but I’d benchmark desktop startup separately before putting a number on that improvement.

Closes #3035

### DIFF
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -1,19 +1,22 @@
-import React, { useState, useEffect, useCallback, useMemo, useRef, Component, ErrorInfo, ReactNode } from 'react';
-import api, { friendlyErrorMessage, LoadedModel, ModelInfo } from './api';
-import { canSelectInComposer, capabilityFromModelInfo, selectPreferredLoadedModel } from './modelCapabilities';
-import { customModelToModelInfo, loadCustomModels } from './features/customModels/customModelStore';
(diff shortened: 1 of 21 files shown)


jeremyfowers:
I can confirm this boots up very fast, but @kpoineal should be the one to review this.

kpoineal (approved):

PR #3049, opened 2026-08-11
GUI3 fix: reintroduce missing HOT label and add it as TAG

fl0rianr:
image


jeremyfowers:
https://github.com/lemonade-sdk/lemonade/issues/3062 feels like an alternative solution to me, not a complementary one. Let's try to bottom out on the core problem: curating what new users see when they first visit the models tab.

fl0rianr:
This fix is needed so GUI actually knows what we define hot on server side - but I guess you got that already and maybe rethinking that "hot" concept as well.

jeremyfowers (approved):
Understood, lets merge.

PR #3050, opened 2026-08-11
GUI3 fix: router UI parity with gui3

fl0rianr:
Current UI router code was not yet confirm with the gui2 development.
This PR addressed that issue.

**Changes:**
Fixes router editor edge cases around switching routing modes and loading existing configurations.

* Warn before clearing existing rules or Natural Language Router progress
* Preserve implicit-AND match expressions correctly
* Tighten classifier and embedding model filtering
* Load the full router configuration when editing an existing router
* Add regression coverage for the affected cases

### DIFF
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -1184,8 +1184,8 @@ const App: React.FC = () => {
                 className="titlebar__window-btn titlebar__window-btn--close"
                 data-tauri-drag-region="false"
                 onClick={() => window.api?.closeWindow?.()}
-                aria-label="Close"
-                title="Close"
+                aria-label="Cancel"
+                title="Cancel"
               >
                 <Icon name="x" size={14} />
               </button>
--- a/src/app/src/components/EffectiveSettingsModal.tsx
+++ b/src/app/src/components/EffectiveSettingsModal.tsx
@@ -395,8 +395,8 @@ const EffectiveSettingsModal: React.FC<EffectiveSettingsModalProps> = ({
             iconOnly
             className="close-modal-btn"
             onClick={onClose}
-            aria-label="Close"
-            title="Close"
+            aria-label="Cancel"
+            title="Cancel"
           />
         </div>
 
--- a/src/app/src/components/ModelListPanel.tsx
+++ b/src/app/src/components/ModelListPanel.tsx
@@ -727,6 +727,7 @@ export const ModelListPanel: React.FC<ModelListPanelProps> = ({
               title="Update all downloaded models"
             />
           )}
+
         </WorkspaceActionGroup>
       )}
     >
--- a/src/app/src/components/ModelManager.tsx
+++ b/src/app/src/components/ModelManager.tsx
@@ -2895,15 +2895,11 @@ const ModelManager: React.FC<ModelManagerProps> = ({ onModelSelect, openModelReq
             ariaLabel={customFormTitle}
             leading={<Icon name="compose" size={20} aria-hidden="true" />}
             title={<h2 className="workspace-detail-panel__title custom-model-editor__title">{customFormTitle}</h2>}
-            metadata={(
-              <>
-                <WorkspaceMetadataChip emphasis="high" tone="accent">
-                  {isCustomOmniCollectionDraft ? 'Omni collection' : 'Custom model'}
-                </WorkspaceMetadataChip>
-                {editingCustomModelName && <WorkspaceMetadataChip emphasis="medium">Editing saved definition</WorkspaceMetadataChip>}
-              </>
-            )}
+            metadata={editingCustomModelName ? (
+              <WorkspaceMetadataChip emphasis="medium">Editing saved definition</WorkspaceMetadataChip>
+            ) : undefined}
             description={<p>Register a model or collection that is not included in the Lemonade catalog.</p>}
+            descriptionPlacement="identity"
             actions={(
               <WorkspaceActionGroup className="custom-model-editor__actions" label={`${customFormTitle} actions`}>
                 <WorkspaceActionButton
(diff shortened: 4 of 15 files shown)


meghsat:
The design looks sleek!

Few questions/ suggestions:
1. The use of window.confirm() for dialogs does not provide a branded, native-looking experience. This can result in inconsistent behavior and UX between Tauri's WebView2 and standard browsers. It would be better to standardize these dialogs in GUI3 so they align with the rest of the UI.
3. The New button (resetDraft) clears all in-progress work without any warning. It should show the same style of confirmation dialog as the mode-switch warnings, guarded by routerDraftHasRulesProgress / routerDraftHasLlmProgress so the prompt only appears when there is meaningful work to lose.
4. After "Save & register" succeeds, there is no immediate visual feedback in the editor, the router only appears silently under the Models panel. GUI2 showed a brief status indicator on completion. A similar success indicator in the bottom-right would make it clear the operation succeeded.
5. The .router-editor__rule-list and .router-editor__classifier-list containers expand unboundedly. When a user adds several classifiers, reaching the Ordered Rules section below requires scrolling past all of them. Capping each list to roughly 3–4 visible items with internal scroll (e.g. max-height + overflow-y: auto) would let users navigate between sections without losing context.
6. The model search dropdowns in Semantic, LLM signal classifiers, and NL router should include a search bar similar to the one available in the Candidate Models selector. The implementation does not need to be identical, but providing comparable search functionality would improve usability, especially when working with large model lists.

fl0rianr:
@meghsat Thank so much for the review, updated the PR, but dawn, this functionality is so complex and hard to test, appreciate any help!

fl0rianr:
Just rebased without changing it so you can see your significant UI progress within a day.

meghsat:
Thank you for the fixes! I have some UI/ UX suggestions:

1. Can we move the workspace-detail-panel__description divs (e.g., “Build and register a virtual model that routes requests across compatible candidates.”) next to workspace-detail-panel__metadata (collection.router)? The description currently seems mixed in with the router-editor__toolbar, which makes it easy to miss.
2. Do we need router-editor__backend-note? It seems redundant, and the router builder is intuitive enough on its own. I’m trying to declutter the panel, as there currently seems to be a lot of information on it.
3. Can we please make the headings title case? For example, “Default model” → “Default Model” and “JSON preview” → “JSON Preview.”
4. Selecting more than five candidate models causes router-editor__connections to overflow. Can we fix the div’s height so that it displays around 3–4 models and becomes scrollable?
5. Can we add a delete icon within each router-editor__connection for the selected candidate model? This would allow users to remove models directly without having to find and uncheck them in the router-editor__candidate-list. We could use the same trash icon used to delete classifiers.
6. I don’t see the context for “External endpoints are provider-wide and affect every model from that provider.” Can we remove it?
7. Doesn’t <div class="router-editor__connection-endpoint"><span>Managed by Lemonade</span><small>Local registered model</small></div> convey essentially the same meaning as <span class="router-editor__source-badge router-editor__source-badge--internal">Internal</span>? Let’s remove the former and move the latter, along with the backend, to the right edge of the div
8. For <div class="router-editor__connection-main"><div><strong>AMD-OLMo-1B-SFT-DPO-Hybrid</strong><span class="router-editor__default-badge">Default</span></div><small>AMD-OLMo-1B-SFT-DPO-Hybrid</small><small>Routing target</small></div>, can we remove the duplicate model ID by removing <small>AMD-OLMo-1B-SFT-DPO-Hybrid</small> and <small>Routing target</small>? Again, I’m trying to keep the UI minimal, as it currently feels a bit cluttered. This should make the div half.
9. We should cap router-editor__wide router-editor__concepts within a fixed-height div to prevent it from overflowing. Maybe show up to three concepts initially and then make the container scrollable?
10. In the JSON Preview section, clicking the copy icon currently displays a green “Copied” message at the bottom, which is outside the user’s immediate field of vision. Instead, can we change the copy button itself to a green “Copied” state for a few seconds, then revert it to the original copy button?
11. Within a gate, we should not allow duplicate conditions of the same type, since we could simply combine them into a single condition.
image
12. Can we distinguish gates using different colors, please? You could use AND-Blue, OR-Green, and NOT-Red
13. Can we add an expanded view for the graph builder, similar to GUI2?  
14. The <h3>Natural-language router</h3> is a duplicate of the NL Router in router-editor__strategy-option, so let’s remove the former and only put the router-editor__form-grid inside router-editor__section. This should also take care of the next pointer.
15. “The selected model must return exactly one candidate. Invalid output falls back to the default model.” is misleading. Since we already append the first part to every prompt by default, I don’t think we need to call this out.
16. We have 2 close buttons. Let's remove the workspace-detail-panel__close and read my next pointer
image
17. Let's move the router-editor__toolbar to the workspace-detail-panel__header. We could vertically stack these in place of the earlier removed workspace-detail-panel__close icon. I'm trying to make use of the empty space here by decluttering the router builder div.

fl0rianr:
Thank you very much! Really an extensive in-depth review - very very helpful!
I worked hard to address all 17 points - but still due to the complexity I can't assure we are fully there but for it's definitely a significant step forward. 
Let me know what is your verdict.

meghsat:
Thank you for addressing the long feedback! I made some tweaks to the GUI and created a PR here: https://github.com/lemonade-sdk/lemonade/pull/3149

1. Added collapsable panels to give the router builder more space
image

2. Added these separators
image

3. Made the trash icon always visible. 
Included the rule and model routes into this panel for better spacing
image

4. Earlier gates with a single condition would blindly collapse but now you'll get a warning that at least 2 conditions are needed to form a gate.
image

5. Fixed a bug where adding a single keywords-any gate would not display it on the canvas but only on the node inspector

6. Better aligned these
image

7. Cleaned the router and model headers
8. Expanded the graph builder size and added a scroll bar to adjust the node inspector and graph canvas size
image

9. Following a uniformed dropdown style
image

10. The dropdown for the latest classifier is cut off
image
to
image

meghsat:
LGTM!

meghsat (approved):

PR #3051, opened 2026-08-11
feat(llamacpp): auto-detect draft GGUF companions

fl0rianr:
## Summary

* Detect `dflash-*.gguf`, `dflash.gguf`, and `mtp-*.gguf` draft companions during GGUF registry discovery.
* Automatically register a single detected companion as the model's `draft` checkpoint during `lemonade pull`.
* Allow detected DFlash draft checkpoints to be enabled with the `dflash` label, while preserving the existing `draft` checkpoint type and MTP behavior.
* Keep draft companions out of the normal quantization/variant list.
* Avoid guessing when a repository contains multiple possible draft companions.
* Add regression coverage for DFlash, MTP draft companions, unquantized main models, and MTP-looking main filenames.

## Why

Custom GGUF repositories can contain a main model, an `mmproj`, and a separate draft model. Lemonade already supports `draft` checkpoints and passes them to llama.cpp using `--model-draft`, but registry pulls did not automatically discover and register DFlash or separate MTP draft companions.

This connects the existing draft-checkpoint support to registry discovery and keeps the current MTP behavior intact.

For a repository such as `unsloth/Muse-Glimmer-30B-GGUF`, a pull like:

```bash
lemonade pull unsloth/Muse-Glimmer-30B-GGUF:Q8_0
```

can now automatically associate the main model, multimodal projector, and `dflash-kquant.gguf` draft model.

### DIFF
--- a/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
@@ -137,16 +137,14 @@ static bool is_llamacpp_cuda_backend(const std::string& backend) {
     return backend == "cuda";
 }
 
-static std::string draft_checkpoint_label(std::string checkpoint) {
+static bool is_dflash_draft_checkpoint(std::string checkpoint) {
     std::transform(checkpoint.begin(), checkpoint.end(), checkpoint.begin(),
                    [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
     size_t separator = checkpoint.find_last_of("/:\\");
     std::string filename = separator == std::string::npos
                                ? checkpoint
                                : checkpoint.substr(separator + 1);
-    if (filename.rfind("dflash-", 0) == 0 || filename == "dflash.gguf") return "dflash";
-    if (filename.rfind("mtp-", 0) == 0) return "mtp";
-    return {};
+    return filename.rfind("dflash-", 0) == 0 || filename == "dflash.gguf";
 }
 
 static std::string trim_version_prefix(const std::string& version) {
@@ -384,17 +382,12 @@ void LlamaCppServer::load(const std::string& model_name,
     }
     push_reserved(reserved_flags, "--mmproj", std::vector<std::string>{"-mm", "-mmu", "--mmproj-url", "--no-mmproj", "--mmproj-auto", "--no-mmproj-auto"});
 
-    const bool uses_dflash =
+    const bool has_dflash_label =
         std::find(model_info.labels.begin(), model_info.labels.end(), "dflash") != model_info.labels.end();
-    const bool uses_mtp =
-        std::find(model_info.labels.begin(), model_info.labels.end(), "mtp") != model_info.labels.end();
-    const std::string draft_label =
-        draft_path.empty() ? std::string() : draft_checkpoint_label(model_info.checkpoint("draft"));
+    const bool is_dflash_draft =
+        !draft_path.empty() && is_dflash_draft_checkpoint(model_info.checkpoint("draft"));
     const bool use_draft_checkpoint =
-        !draft_path.empty() &&
-        (draft_label.empty() ||
-         (draft_label == "dflash" && uses_dflash) ||
-         (draft_label == "mtp" && uses_mtp));
+        !draft_path.empty() && (!is_dflash_draft || has_dflash_label);
 
     if (use_draft_checkpoint) {
         push_arg(args, reserved_flags, "--model-draft", draft_path);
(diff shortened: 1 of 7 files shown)


src/cpp/server/backends/llamacpp/llamacpp_server.cpp
```diff
@@ -379,7 +390,15 @@ void LlamaCppServer::load(const std::string& model_name,
     // Use legacy reasoning formatting
     push_overridable_arg(args, llamacpp_args, "--reasoning-format", "auto");
 
-    if (std::find(model_info.labels.begin(), model_info.labels.end(), "mtp") != model_info.labels.end()) {
+    const bool uses_dflash = !draft_path.empty() &&
```
bitgamma, line 393: can we use a label like for mtp? I like doing A/B testing with/without drafter but if this is merged as-is I think the only way I have to do it is to duplicate the profile and remove the draft checkpoint

bitgamma: also, while we are at it, maybe dspark could be added as well (if you feel like it, can be done in a separate pr)

iswaryaalex:
Great feature, was just adding hooks for this. Thanks for auto-detect

src/cpp/server/backends/llamacpp/llamacpp_server.cpp
fl0rianr: Thanks, done

fl0rianr:
I think we make a vision projector more explicit but draft - this is only IMO about performance not quality or function, auto is the right way to deal with it - who would want the slower version?

src/cpp/server/backends/llamacpp/llamacpp_server.cpp
bitgamma: this kinda goes in the opposite direction of what I meant, it is not possible to turn it off now. Since unfortunately the drafter does not always give the hoped results and takes some additional VRAM so there are cases where you want to manually control it. I understand it is a bit of a edge situation and there is a workaround (adding the same model without the draft checkpoint) so it is OK, but ideally the dflash label would be added during model registration like the mtp one (if I remember correctly how it was added)

sofiageo (approved):
Tested it partially with Muse-Glimmer-30B-GGUF:IQ2_XXS and it creates the correct user model and starts downloading the files. Because I'm having issues with slow internet at the moment I couldn't complete the testing.

I'm not really a fan of the draft detection algorithm. I tried to find more quants with separate draft checkpoints but there aren't many available. Maybe in the future we can use the metadata from the .gguf files to determine if a checkpoint is draft or not.

src/cpp/server/backends/llamacpp/llamacpp_server.cpp
fl0rianr: Sorry for misunderstanding - I hope this is the better approach now.

fl0rianr:
> I'm not really a fan of the draft detection algorithm. I tried to find more quants with separate draft checkpoints but there aren't many available. Maybe in the future we can use the metadata from the .gguf files to determine if a checkpoint is draft or not.

@sofiageo Does not sound very approving to me - do you think we have a bit better approach now?

sofiageo:
> @sofiageo Does not sound very approving to me - do you think we have a bit better approach now?

For the muse glimmer release both approaches are fine. I'm just not sure every model will follow the same naming convention in the future. But we don't have to handle this now. Please make sure to resolve the issues that bitgamma mentioned before merging since he is more familiar with the code base than I am.

src/cpp/server/backends/llamacpp/llamacpp_server.cpp
fl0rianr: Puh, got a bit confused about directions yesterday, I think I have it now.

bitgamma (approved):
lgtm

fl0rianr:
Let's go with this. My original motivation for auto-detection was also to avoid making DFlash another decision most users need to understand. But that's not in there anymore.

@bitgamma Your A/B use case makes sense though, so maybe as a follow-up we can revisit the activation policy: keep the common path automatic, while the label acts as an explicit override for users who want to disable or test the drafter. I'd rather avoid adding a new UI choice, I think most users would want the drafter enabled when it is available. The users who need to turn it off are more likely to be advanced users doing A/B testing or managing VRAM carefully. So i would lean more towards an auto active and manuell opt-out like with your context size estimator - but let's decide that later on.

bitgamma:
> Let's go with this. My original motivation for auto-detection was also to avoid making DFlash another decision most users need to understand. But that's not in there anymore.
> 
> @bitgamma Your A/B use case makes sense though, so maybe as a follow-up we can revisit the activation policy: keep the common path automatic, while the label acts as an explicit override for users who want to disable or test the drafter. I'd rather avoid adding a new UI choice, I think most users would want the drafter enabled when it is available. The users who need to turn it off are more likely to be advanced users doing A/B testing or managing VRAM carefully. So i would lean more towards an auto active and manuell opt-out like with your context size estimator - but let's decide that later on.

Yes I totally agree that defaulting to "on" is the right choice. We need to follow up with:

- Activation policy
- DSpark detection
- If possible, GGUF metadata based detection, also to account for models where the drafter weights might be included in the main model checkpoint

But this PR is already valuable as it is now, merging.

PR #3052, opened 2026-08-11
ci: stop building unused targets in the validate workflows

jeremyfowers:
Cuts ~13 minutes of CI time per merge by building only the targets the validation actually runs.

## Solution

The llama.cpp and vLLM validate jobs built **every** CMake target; they only need `lemond` (plus the CLI for sd.cpp/vLLM).

- `validate_llamacpp` — all targets → `lemond`, plus `BUILD_WEB_APP=OFF` and `/m:2`
- `validate_vllm` — all targets → `lemond lemonade`; now verifies `build/lemonade` alongside `build/lemond`
- `validate_sdcpp` — targets were already narrowed; adds `BUILD_WEB_APP=OFF` (skips the configure-time `CONFIGURE_DEPENDS` glob over `src/app`; the web app was never in this build graph) and `/m:2`

Also strips trailing whitespace on three lines of `validate_llamacpp.yml` (pre-commit).

## Results

Build-job wall clock. "After" is measured on `7c625bab3` ([run](https://github.com/lemonade-sdk/lemonade/actions/runs/31810829139)); "Before" lists every run of that job I could find in the current config, since these Windows runners are noisy enough that a single pair of numbers would overstate the precision.

| Job | Before | After |
|---|---|---|
| Validate llama.cpp — Build | 19m35s / 22m01s / 23m03s | **10m39s** |
| Validate vLLM — Build (Linux) | 10m30s / 10m41s | **8m15s** |
| Validate sd.cpp — Build | 9m57s / 10m18s / 11m53s | 10m59s |

**llama.cpp** is the real win — it was building all ~623 targets, including every unit test, to run an HTTP-driven validation that only ever touches `lemond`. Configure was 182.7s and compiling `lemond` alone took 7m05s.

**vLLM** drops from 623 targets to 504; compile went 8m30s → 6m06s, with `setup.sh` (1m56s) unchanged.

**sd.cpp is honestly a wash** — it lands inside the range of the unmodified job. Its targets were already narrowed on `main`, so all that is left is `BUILD_WEB_APP=OFF` shaving the `src/app` glob out of configure, and that is smaller than the run-to-run spread. Happy to drop the sd.cpp hunk if you would rather not carry a change that cannot be measured.

## No `_deps` cache

An earlier revision cached `build/_deps` in these three jobs. The warm-cache rerun @fl0rianr asked for shows it does not pay for itself, so it is dropped — along with the two changes that existed only to protect it (the `Remove-Item build` deletions and vLLM's direct `apt-get` in place of `setup.sh`).

- **2 of 3 jobs missed** on the rerun even though the prior attempt saved both keys 14 hours earlier. The repo's Actions cache sits at its 10 GB cap (10.12 GB across 67 entries), so ~550 MB entries are evicted within hours.
- **The job that hit got slower** — sd.cpp went 9m54s → 10m58s, i.e. within noise.
- **The ceiling is ~30s regardless.** In that job, configure was 131.9s cold vs 133.4s warm (FetchContent reconfigures either way) and dependency compilation was ~45s cold vs ~2s warm, while restoring the 549 MB entry costs ~15s.

This also resolves the Codex finding about `setup.sh` deleting the restored cache, and the cache-namespace/`pkg-config` notes from review, by removing the cache entirely.

### DIFF
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -902,7 +902,7 @@ jobs:
         uses: ./.github/actions/get-version
 
       - name: Cache FetchContent dependencies
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: build/_deps
           # vs2026 marker: see installer job — VS 2022-built _deps cannot be
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -889,7 +889,7 @@ jobs:
     outputs:
       version: ${{ steps.get_version.outputs.version }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v4
         with:
           clean: true
           fetch-depth: 0
@@ -902,7 +902,7 @@ jobs:
         uses: ./.github/actions/get-version
 
       - name: Cache FetchContent dependencies
-        uses: actions/cache@v5
+        uses: actions/cache@v4
         with:
           path: build/_deps
           # vs2026 marker: see installer job — VS 2022-built _deps cannot be
@@ -927,7 +927,7 @@ jobs:
           Get-ChildItem $archive
 
       - name: Upload embeddable archive
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@v4
         with:
           name: lemonade-embeddable-windows
           path: build/lemonade-embeddable-${{ env.LEMONADE_VERSION }}-windows-x64.zip
--- a/.github/workflows/cpp_server_build_test_release.yml
+++ b/.github/workflows/cpp_server_build_test_release.yml
@@ -889,7 +889,7 @@ jobs:
     outputs:
       version: ${{ steps.get_version.outputs.version }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           clean: true
           fetch-depth: 0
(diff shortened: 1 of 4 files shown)


fl0rianr (commented):
Overall this looks good to me. The target narrowing makes sense, BUILD_WEB_APP=OFF is appropriate for these validation jobs, and the vLLM change correctly avoids setup.sh deleting the restored _deps directory.

One thing I’d still like to see before relying on the cache changes: the benchmark runs on d555021c were all cold-cache runs. In the SD job the shared llama cache had not been saved yet, and llama/vLLM only seeded their caches at the end of the job. So the current results validate the target/setup optimizations, but not an actual warm-cache restore. Could we rerun the same head once and confirm an exact cache hit, successful configure/build/validation, and the resulting build time?

There’s also a small robustness concern with the cache namespace. _deps contains configured/built artifacts, while the compatibility boundary is currently the manual vs2026 / ubuntu2204 prefix. CMakePresets.json controls the generator/architecture but is not part of that boundary, so an incompatible preset/toolchain change relies on someone remembering to bump the marker. If that is intentional policy, the comments document it reasonably well. Otherwise it might be worth including the relevant preset/toolchain fingerprint in the restore namespace.

One minor vLLM nit: installing pkg-config on the Ubuntu runner replaces the already-present pkgconf provider and causes some unnecessary package churn. Not a correctness issue, just something that could be avoided in a CI-time optimization PR.

No blocking correctness issues from my side. With a successful warm-cache rerun, I’d be comfortable approving this.

jeremyfowers:
I just triggered a run-run of all 3 validation workflows

jeremyfowers:
Dropped the `_deps` cache after the warm-cache rerun showed it does not pay for itself (2 of 3 jobs missed because the repo cache is at its 10 GB cap, the one that hit got slower, and the ceiling is ~30s anyway), so this PR is now just the target narrowing — full CI is green on `7c625bab3`.

fl0rianr (approved):
LGTM

PR #3061, opened 2026-08-11
fix(gui3): router correct integration

fl0rianr:
## Summary

Fixes GUI3 router handling across Chat and Lemonade Tools.

* Treat `collection.router` as a virtual router instead of a directly loaded runtime model
* Add router dependency/preflight checks with clearer candidate/classifier errors
* Make Lemonade model tools router-aware for list/info/load/unload/delete
* Keep router selection/state in sync across Chat and Models
* Fix router routing trace/error attribution
* Fix Router task color in Chat

Designed to merge separately from #3050.

## Validation

* Router load regression tests
* Router editor tests
* Lemonade/MCP tool runtime tests
* Router SSE / route trace tests
* TypeScript transpile checks

### DIFF
--- a/src/app/src/components/ModelManager.tsx
+++ b/src/app/src/components/ModelManager.tsx
@@ -25,6 +25,7 @@ import {WorkspaceActionButton, WorkspaceActionGroup, WorkspaceDetailPanel, Works
 import Modal from './inspect/Modal';
 
 import { ROUTER_RECIPE, routerDisplayName, type RouterPullRequest } from '../features/router/routerTypes';
+import { isRouterModelInfo, preflightRouter, routerPreflightError } from '../features/router/routerRuntime';
 import {
   GLOBAL_MODEL_SETTINGS_EVENT,
   automaticUpdateIsDue,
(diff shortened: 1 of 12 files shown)


src/app/src/components/ModelManager.tsx
```diff
@@ -1603,7 +1616,17 @@ const ModelManager: React.FC<ModelManagerProps> = ({ onModelSelect, openModelReq
     setLoadError(null);
     setLoadingModel(name);
     try {
-      await loadWithGlobalPolicy(model);
+      if (isRouterModelInfo(model)) {
```
meghsat, line 1631: This unconditional call api.registerModelDefinition() on every Load/"Use Router" click, using info from potentially-stale local allModels state. A second client selecting the router to chat could silently overwrite another client's more recent router policy edit. 
I validated this from the GUI and noticed the stale router fire with the old routing payload, even though nothing in the routing was actually changed

meghsat:
LGTM

meghsat (approved):

PR #3073, opened 2026-08-12
fix(server): compare artifacts, not commit SHAs, when checking for model updates

### LINKED ISSUE #2542: check_for_model_updates ignores sticky refs/main — README-only upstream commits never clear the "Update available" flag
### Platform

Windows

### Lemonade Version

10.9.0

### GPU / APU Model

### Component

lemond (server)

### Bug Description

`check_for_model_updates` compares only the commit SHA from `refs/main` against the latest SHA from the HF API. When upstream makes a README-only commit (artifact files unchanged), `download_from_huggingface` correctly keeps `refs/main` on the old snapshot (`can_reuse_previous_hf_snapshot`, #2066). But `check_for_model_updates` (#2243) doesn't call that check — it sees SHA mismatch and sets `update_available=true`. Endless cycle: Pull clears flag → restart re-raises it.

### Steps to Reproduce

1. Pull a model where upstream has a README-only commit after the cached SHA (e.g. `Comfy-Org/z_image_turbo`: cached `2f862278568d`, latest `d24c4cf2a0cd`).
2. Restart the server.
3. Log: `Update available for ... 1 model(s)`.
4. Click Update / `lemonade pull`. Pull keeps `refs/main` — flag clears.
5. Restart → flag is back.

Expected: no `update_available` on restart when artifacts are unchanged.
Actual: flag on every restart because `check_for_model_updates` doesn't call `can_reuse_previous_hf_snapshot`.

### Log Output

```shell

```

### Additional Context

PR #2243 added SHA-only `check_for_model_updates` after #2066 already had `can_reuse_previous_hf_snapshot` in `download_from_huggingface` — the two never got synced. Fix: reuse `can_reuse_previous_hf_snapshot` in the startup check.

blackdeathdrow:
## Changes

- Compare the **per-model** artifact set between the cached ref and the latest commit before advertising a re-download, reusing the existing `can_reuse_previous_hf_snapshot` / `fetch_hf_file_metadata_for_ref` helpers.
- The comparison baseline is the **on-disk** snapshot (resolved checkpoint path, falling back to `refs/main`) — deliberately not the processed-at-pull sha recorded in `.lemonade_registry.json`, which can name a snapshot that was never materialized locally (pull keeps `refs/main` on an older snapshot when the selected artifacts are unchanged).
- Per-model file set is built from the model's resolved checkpoints (`main`/`mmproj`/`draft`/`config`), expanded across GGUF shard families — a sharded model resolves to a single shard, so the rest of the family must be included or a change confined to another shard would be missed.
- Pull and the update check share the selection via `lemon::registry_files::select_main_repo_files` (new `registry_files.h`, defined in `model_manager.cpp`), so a file pull would download is always a file the check compares.
- The comparison uses `registry_files::select_main_repo_files_union()` — the union of what's selected from the previous *and* current revision's listings — so a file removed upstream (e.g. from a directory checkpoint or a GGUF shard family) isn't silently dropped out of comparison just because it's absent from whichever single listing the selection would otherwise see.
- Pull's own snapshot-reuse fast path (`download_from_registry`) shared the identical blind spot independently of this update-check path — the union fix applies there too, for the comparison only; the actual download list stays scoped to files that exist in the current revision.
- Auxiliary checkpoints that live in another repository are verified under their own repository entry, tracked by `registry_files::DeterminationTracker`; a model counts as verified only once every repository it spans completes a determination, and a repository with no resolvable local baseline assumes changed rather than blocking the others.
- Gated on Hugging Face. ModelScope snapshots are tree fingerprints with no commit pin, so they keep the snapshot-id comparison.
- Hugging Face tree pages are memoized per (repo, ref, subdir) within one check run, so N models in a shared multi-quant repository cost one page fetch per ref, not 2N.
- An unreachable repository (404 / network error) assumes updates for its models instead of leaving the update flag stale.

## Why per-model and not the whole snapshot

Enumerating the snapshot directory would be simpler but wrong for shared multi-artifact repos: a repository hosting many quants would still report an update for every model whenever any *sibling* variant changed.

## Safety

Every indeterminate path assumes an update: non-HF source, missing file on disk, no resolvable local baseline, iteration error, metadata fetch failure, empty selection, unreachable repository. A real update is never hidden. The extra tree API calls happen only on SHA mismatch — exactly what the subsequent pull would do anyway.

## Testing

New `test/cpp/test_hf_update_check.cpp` (`HfUpdateCheckTest`, `CI ON`), 49 assertions, no network:

- only the resolved variant is selected; sibling variants in the same repo are not
- shard family expanded from a single resolved shard; sibling quants are not selected
- directory checkpoints select their subtree; files added upstream inside the directory are compared
- `select_main_repo_files_union` keeps a file removed from either revision's listing (directory checkpoint and GGUF shard cases), doesn't duplicate overlapping files, and still throws for a missing direct-file variant
- auxiliary checkpoints are grouped under their own repository
- the baseline is the on-disk snapshot, never the processed-at-pull sha
- `can_reuse_previous_hf_snapshot` detects content changes, added/missing files, partial downloads, and size mismatches
- every fallback path returns empty (missing file, empty input, nonexistent/empty snapshot, unresolved role)
- `DeterminationTracker` verifies a model only once every repository it spans has completed a determination, and `mark_determined()` stays idempotent past zero

Fixes #2542


fl0rianr (changes requested):
Thanks for tackling this. The overall direction looks good, especially making the update check artifact-aware instead of comparing repo SHAs only.

I do see a few correctness cases that I think should be addressed before merging:

* **Processed SHA vs local snapshot:** after a pull reuses an older snapshot because artifacts are unchanged, `processed_models.snapshot_id` can point to the newer remote SHA while that snapshot was removed locally. `check_for_model_updates()` prefers that processed SHA, then `hf_selected_artifacts_unchanged()` looks for `snapshots/<processed_sha>`. On the next unrelated upstream commit this falls back to “changed” again, so the false-positive cycle can return. I think the remote metadata baseline and the active local snapshot need to be tracked separately.

* **Directory checkpoints can miss newly added files:** `model_artifacts()` builds the selected set only from files present in the old local directory. If the new revision adds a file inside that directory while existing files stay unchanged, the new file is never compared and the update may be hidden. For directory artifacts we should compare the file set between revisions too.

* **The update check does not quite mirror download selection:** the download path can automatically include files such as `config.json`, `tokenizer.json`, `tokenizer_config.json`, etc., while the new check derives its set from `resolved_paths`. A change only to one of those extra selected files could therefore be missed. It would be safer to share the same artifact-selection helper between download and update checking.

* **Auxiliary checkpoints in another repo are skipped but not actually checked separately here.** `check_for_model_updates()` creates entries from the main checkpoint repo only, while `model_artifacts()` skips resolved paths outside that snapshot. So a changed external `mmproj`/draft/etc. can go unnoticed.

* **Fail-safe behavior:** `verified_models` is set before the artifact comparison, and not every filesystem operation in that comparison is covered by the local `try`. If one of those throws, the outer catch can still leave the model “verified” and clear `update_available`. Any indeterminate artifact check should consistently fall back to “assume changed”.

The new unit tests are useful, especially the shard-family and directory coverage. I’d also add regression tests for:

1. active snapshot A / processed SHA B / new upstream SHA C,
2. a file added inside a directory checkpoint,
3. a change only to an automatically selected tokenizer/config file,
4. an auxiliary checkpoint living in another repo.

So overall I like the approach, but I think the artifact selection and snapshot-baseline semantics need a little more tightening before this is safe to merge.

blackdeathdrow:
Thanks for the detailed review — pushed a follow-up that addresses all 5 points:
1. Baseline is now the on-disk snapshot only; processed sha is just a fast-path, never the comparison baseline.
2. Directory checkpoints compare against the new revision's tree, so upstream-added files aren't missed.
3. Selection is shared with pull via `registry_files::select_main_repo_files()`.
4. Aux checkpoints in another repo are verified under their own entry (`DeterminationTracker`) — a model is verified only once every repo it spans completes.
5. Any indeterminate result (including no local baseline) assumes changed and can't block another repo's determination.
Added regression tests for all of the above (43 assertions). PR description updated.

fl0rianr (commented):
Thanks for the follow-up, this is much stronger now. The five points from my previous review look addressed: the comparison now uses the local snapshot as baseline, pull/update share the same selection logic, aux repos are tracked separately, directory additions are covered, and indeterminate cases correctly fall back to “changed”.

I found one remaining correctness case: files removed upstream can still be missed. The update check builds selected only from latest_files, then compares both revisions using that set. If a directory checkpoint previously contained model.onnx + tokens.json, and the new revision removes tokens.json while leaving model.onnx unchanged, tokens.json is no longer selected at all. The comparison can therefore report the artifacts as unchanged.

The existing “missing from new revision metadata” test does not quite cover this, because it manually keeps the missing file in selected_files. I think we should derive the selection for both the active and latest revisions and treat any difference in those sets as a change, or compare their union so removals produce missing metadata.

A small regression test for old: model.onnx + tokens.json → new: model.onnx should catch it.

Other than that, the new structure looks good to me.

blackdeathdrow:
Done

fl0rianr (commented):
Thanks, the removed-file case looks properly covered now. The union across the previous and current selections handles directory files and GGUF shards symmetrically, and the added regression tests look good.

I found one remaining issue in the pull snapshot-reuse path: when an auxiliary checkpoint such as `mmproj` or `draft` lives in the **same repo as the main checkpoint**, it is initially included in `files_to_download[main_repo]`, but `comparison_files` is later replaced with `select_main_repo_files_union(...)`. That union only represents the main selection, so the same-repo auxiliary files are dropped from the reuse comparison.

This means a case like:

* main artifact unchanged
* same-repo `mmproj` changed

can still incorrectly reuse the previous snapshot.

The update-check path already does the right thing by appending the auxiliary variants after building the main union. I think the pull path should preserve the existing `files` set and merge the main old/new union into it rather than replacing it.

A regression test for `main + mmproj` in the same repo, with only `mmproj` changing, would cover this nicely.

Other than that, the previous findings look addressed. Once this case is covered, I don't see another code-level blocker from my review.

fl0rianr (approved):
thanks, LGTM now

PR #3074, opened 2026-08-12
GUI3 fix: effective settings visual correction

### LINKED ISSUE #3071: Put the icon next to the Effective Settings title
Image

Or just delete the icon

fl0rianr:
No only the icons on the title, the spinners are broken as well:
image

This PR fixes this:
image

When loaded the setting actual ctx setting shows up:
image

And the info about ctx (that I move higher) is now removed if we have a user configured fixed ctx:
image

Also removed visible Em Dashes, thanks @Rose22 

Not the final solution but visually fine now and ready for the functional follow-up.

Closes #3071


jeremyfowers (approved):
Nice improvement!

PR #3076, opened 2026-08-12
fix(gui): visually correct chat model options card

### LINKED ISSUE #3054: Refine the model options card in the chat tab
Image

Image

This card that shows for model options for image gen (and other model types) sticks out because it has a shadow under it, and it competes heavily with the text entry and model selection ui elements. It's not as bad in the dark theme because there's no shadow (or we just can't see the shadow). 

My suggestion to fully polish this would be to display the model controls inside the text entry box, at the top, rather than putting them in a separate card.

fl0rianr:
image
image
image

Closes #3054


jeremyfowers (approved):
Huge improvement!

PR #3079, opened 2026-08-12
Bump TheNoise to version v0.1.2

bitgamma:
Newer version of TheNoise, mostly fixes auto-detection of models when using custom checkpoints instead of the original weights.


fl0rianr (approved):
LGTM

Issue #3081, opened 2026-08-12
GUI3: get rid of the x's and +'s on the backend filters


jeremyfowers:
<img width="480" height="404" alt="Image" src="https://github.com/user-attachments/assets/2b00f265-a6f3-400f-8c1e-d59d9eda5c07" />

This is way too much mental load, let's just show everything. We can implement a more elegant system post-MVP.

Originally pointed out by @Rose22

PR #3083, opened 2026-08-12
GUI3 fix: get rid of the x's and +'s on the backend filters

jeremyfowers:
closes https://github.com/lemonade-sdk/lemonade/issues/3081

image


fl0rianr (commented):
I'm good with that but you are aware npm testing cost you only 2 min? There was a test for this and it will fail now. That's the only point.

jeremyfowers:
All tests are passing @fl0rianr please review

fl0rianr (approved):
LGTM

PR #3033, opened 2026-08-10
docs: add Muse Glimmer 30B blog post

jeremyfowers:
Desktop view of the Muse Glimmer blog post

Mobile view of the Muse Glimmer blog post

PR #3090, opened 2026-08-12
feat(models): add Muse Glimmer 30B to the model catalog

ramkrishna2910:
Adds `Muse-Glimmer-30B-GGUF` as a built-in model, as promised by the [Muse Glimmer blog post](https://lemonade-server.ai/news/muse-glimmer.html) (#3033), which states the model ships built-in in the August 12 release.

Uses the `checkpoints` form so the vision projector and the DFlash draft companion are pinned explicitly — the repo ships three mmproj candidates, and #3051 added DFlash discovery.

**Requires a llama.cpp re-pin before this is functional.** The current `b10360` pin (#3053, published Aug 11 13:17 UTC) includes upstream `model: Muse Glimmer Support` (ggml-org/llama.cpp#26841) but predates two follow-up fixes:

| Upstream PR | Merged (UTC) | In b10360 |
|---|---|---|
| #26841 Muse Glimmer Support | Aug 10 11:07 | yes |
| #26879 fix tool call detection after EOM | Aug 11 20:15 | no |
| #26900 disallow integer dflash sliding_window_pattern | Aug 12 11:24 | no |

#26879 affects tool calling (the model's agentic path) and #26900 affects the dflash draft path this entry uses. Latest upstream is `b10375` (Aug 12 12:18 UTC), which covers all three.

The `hot` label means `validate_llamacpp.py` will exercise this model, so a `validate_llamacpp` `workflow_dispatch` run after this merges will both re-pin and produce validation evidence.

Size `21.4` is the sum of main (15.88 GB) + mmproj BF16 (3.85 GB) + dflash (1.63 GB); flag if the convention is main-only.

### DIFF
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -617,6 +617,23 @@
         ],
         "size": 3.15
     },
+    "Muse-Glimmer-30B-GGUF": {
+        "checkpoints": {
+            "main": "unsloth/Muse-Glimmer-30B-GGUF:UD-Q4_K_XL",
+            "draft": "unsloth/Muse-Glimmer-30B-GGUF:dflash-kquant.gguf",
+            "mmproj": "unsloth/Muse-Glimmer-30B-GGUF:mmproj-Muse-Glimmer-30B-BF16.gguf"
+        },
+        "recipe": "llamacpp",
+        "suggested": true,
+        "labels": [
+            "hot",
+            "tool-calling",
+            "vision",
+            "reasoning",
+            "llamacpp"
+        ],
+        "size": 21.4
+    },
     "MiniCPM-o-4.5-Vision-GGUF": {
         "source": "modelscope",
         "checkpoint": "OpenBMB/MiniCPM-o-4_5-gguf:MiniCPM-o-4_5-Q4_K_M.gguf",
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -630,7 +630,8 @@
             "tool-calling",
             "vision",
             "reasoning",
-            "llamacpp"
+            "llamacpp",
+            "dflash"
         ],
         "size": 21.4
     },
--- a/src/cpp/resources/server_models.json
+++ b/src/cpp/resources/server_models.json
@@ -633,7 +633,7 @@
             "llamacpp",
             "dflash"
         ],
-        "size": 21.4
+        "size": 19.9
     },
     "MiniCPM-o-4.5-Vision-GGUF": {
         "source": "modelscope",
--- a/docs/dev/backends-reference.md
+++ b/docs/dev/backends-reference.md
@@ -207,7 +207,7 @@ the generator instead. Prose outside the markers is preserved. -->
 |-------|-----------|--------|
 | `kokoro-v1` | 0.354 | tts |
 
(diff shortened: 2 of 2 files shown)


src/cpp/resources/server_models.json
```diff
@@ -617,6 +617,23 @@
...
+            "mmproj": "unsloth/Muse-Glimmer-30B-GGUF:mmproj-Muse-Glimmer-30B-BF16.gguf"
+        },
+        "recipe": "llamacpp",
+        "suggested": true,
+        "labels": [
+            "hot",
+            "tool-calling",
+            "vision",
+            "reasoning",
+            "llamacpp"
```
bitgamma: let's add the "dflash" label so dflash is enabled automatically

ramkrishna2910:
Good catch — added in 5c25cde9b.

Confirmed against `llamacpp_server.cpp:385-405`: without the label, `is_dflash_draft` is true (the checkpoint matches the `dflash-` prefix) but `has_dflash_label` is false, so `use_draft_checkpoint` evaluates to false and `--model-draft` is never passed. The 1.63 GB companion would have been downloaded and then silently ignored, and `--spec-type draft-dflash` never set. With the label both now fire.

src/cpp/resources/server_models.json
ramkrishna2910: Done!

bitgamma (approved):

ramkrishna2910:
## Tested locally on Strix Halo (gfx1151, Radeon 8060S, 64 GB)

Built this branch (`lemond 11.5.2`, vs18) and ran it against the **current `b10360` pin** — i.e. *without* upstream ggml-org/llama.cpp#26879 and #26900. All green:

| Test | Result |
|---|---|
| Load | success in 15.3s, GPU, 131072 ctx |
| Chat | `content: "2+2 is 4."`, `finish_reason: stop` |
| Tool calling | `finish_reason: tool_calls`, `get_weather({"city":"Paris"})` |
| Vision (mmproj) | accurate description of `sample_image.png` |

**DFlash activation confirmed** — the label fix works end to end:

```
[Info] (LlamaCpp) Model uses DFlash, adding draft decoding defaults
common_speculative_impl_draft_dflash: adding speculative implementation 'draft-dflash'
  - n_max=3, n_min=0, p_min=0.00
  - block_size=16, mask_token_id=201818, n_extract=5
```

Speculative decoding is doing real work: draft acceptance **0.48–0.76**, mean accepted length 2.45–3.28 across requests.

Two notes:

1. **`size` corrected 21.4 → 19.9 in d25bba91b.** This answers the open question in the PR description: the convention is **GiB**, not GB. With the files on disk the server reported `size: 19.9` against my JSON's `21.4` — 21.36 GB = 19.89 GiB. It is the sum of main + mmproj + dflash, just in the other unit.

2. These two lines appear during load and are **benign** — lemonade's memory-fitting loop retries and the load then succeeds:

   ```
   E llama_init_from_model: failed to initialize the context: dflash requires ctx_other to be set
   W srv load_model: [spec] failed to measure draft model memory
   ```

Tool calling passing on b10360 is reassuring but not conclusive for #26879 — this was a single-turn call, not a tool call after EOM in a multi-turn agentic loop. I'll re-run against b10375 once the re-pin lands.

jeremyfowers:
It seems like a bug that docs/dev/backends-reference.md would have a complete list of models...

PR #3092, opened 2026-08-12
GUI3 fix: options block escape sequences and light-theme button contrast

### LINKED ISSUE #3068: Type your own \u2026
Type your own \u2026
Image

(need i say anything more?)

jeremyfowers:
Two rendering bugs in the tool-call options block.

**Escape sequences shown verbatim (#3068).** The block wrote `…` and `✓` in JSX text and a JSX attribute. Those are not JavaScript string literals and do not process backslash escapes, so users saw the raw sequence. Replaced with the literal characters, matching the 86 other literal `…` uses in the renderer.

**White text on a yellow button in the light theme.** The Send button and the choice buttons took their text color from `--bg-primary`, which is near-black in dark (`#0E0E0B`) but near-white in light (`#fdfbf6`). Switched all four accent-background rules to `--accent-on` (`#000000` in both themes, already used 13× elsewhere for this exact purpose). `critical.generated.css` is updated to match, since it is committed and inlined into the web app's first paint.

image

Both guarded by new static checks in `ui-regression-contracts.runtime.cjs`, each verified to fail on the pre-fix source:
- An AST walk over all 42 renderer `.tsx` files rejecting `\uXXXX`/`\xXX` inside JSX text or JSX attribute strings.
- A stylesheet scan rejecting any rule that paints an accent background and takes its text color from a theme-flipped `--bg-*` token.

Final result:

image

Closes #3068

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -4275,11 +4275,11 @@ const ToolCallsDisplay: React.FC<{ calls: ToolCallEntry[]; onOptionSelect?: (tex
                   ))}
                 </div>
                 {selectedChoice && (
-                  <div className="options-block__confirmation">\u2713 You chose: {selectedChoice}</div>
+                  <div className="options-block__confirmation">✓ You chose: {selectedChoice}</div>
                 )}
                 {!selectedChoice && allowCustom && (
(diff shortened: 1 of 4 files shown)


fl0rianr (approved):
LGTM

PR #3095, opened 2026-08-12
GUI3 fix: make plus and tools more detectable

### LINKED ISSUE #3080: GUI3: redundant/confusing subtitle on add files button
Image

Why does it say "upload images" right under "Add photos and files"? This redundancy is confusing.

Suggestion: 
- the tile should just be `Add files`
- the subtitle should list the file types supported by the selected model: images, audio files, etc.

fl0rianr:
Makes the plus more visible and detectable. Small change, huge difference.

image

Setting is unclear since there is no "Setting" at all.

image

Try it out locally...

Also added a fix that fits well the review together:

image
image

image

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -3892,17 +3892,17 @@ ${finalText}`
                 >
                   <span className="composer__add-icon"><Icon name="paperclip" size={16} /></span>
                   <span className="composer__add-text">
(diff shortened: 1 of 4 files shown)


PCAssistSoftware (commented):
Tested this PR in relation to #3080  and it looks fine to me, meets the requirements of that issue

jeremyfowers (approved):

PR #3099, opened 2026-08-12
feat(models): add an explicit `chat` label

### LINKED ISSUE #3070: Some LLMs classified as "unknown"
Image

This classification is very important since it determines icons, which settings are available, and more.

Ensure that there are no "unknown" class models anywhere in the app.

jeremyfowers:
## Why

Clients had no deterministic way to know whether a model is a chat model — chat-ness was inferred from whichever capability labels happened to be present, in five inconsistent places. Fixes #3070.

## Details

`chat` becomes a first-class deployment label, and each backend descriptor declares the deployment modes it can serve, replacing `default_labels`, `default_classification` and the editorial `modality` string.

- `chat` added to the LLM entries in `server_models.json`; `get_model_type_from_labels` keys off it instead of `vision`/`reasoning`/`tool-calling`
- Each descriptor declares `supported_modes` (most-default first) plus `default_capabilities` for non-mode labels. A model naming no mode is stamped with the recipe's default, so an LLM always carries `chat`
- Labels are reconciled against `supported_modes` on every ingest path — registry, FLM discovery, cloud, `extra_models_dir`, `/pull/variants`, and legacy `user_models.json`. `get_deployment_model_type()` and the copy-pasted non-chat denylists are deleted
- `POST /pull` returns 400 and registers nothing when a definition names an unservable mode, through `labels` or the `embedding`/`reranking` parameters; collection imports reject the same definition before any component registers. Persisted entries still normalize on load rather than blocking startup
- `/system-info` derives `modality` from the declared modes; `bench` reads them directly instead of parsing the display string back into a `ModelType`, fixing llamacpp embedding and reranking benchmarks
- `BackendModeContractTest` fails the build when a descriptor's declared modes disagree with the capability interfaces its server class implements
- Deprecated `Lite Collection` / `Ultra Collection` entries removed

**Not included:** `src/app` is untouched.


jeremyfowers:
Requires https://github.com/lemonade-sdk/lemonade/pull/3103

bitgamma (approved):
LGTM. The chatgpt comments do not seem to relevant to me, but you might want to take a look before merging

fl0rianr (changes requested):
I also always wanted to answer #3070 since this is a server error and change makes a lot of sense.  Clients need a positive, deterministic way to know that a model is an LLM instead of inferring it from whatever capability labels happen to be present.

I like the direction of making **chat** a first-class deployment label. But I think there are two edge cases worth locking down because they can break that exact invariant:

1. llamacpp + labels: ["classification"]

`normalized_definition_labels()` sees `classification` as an already-declared deployment mode, so it doesn't stamp the recipe default `chat`. Later, `get_deployment_model_type()` correctly recognizes that llamacpp cannot serve `/classify`, removes that classification claim logically, and treats the model as an LLM.

That leaves us with an effective `ModelType::LLM` that does **not** carry **chat**, while the new CLI/model-selection code identifies LLMs solely by **chat**.

There is already a regression test for this supported LLM-as-classifier behavior in `test_model_manager_collection_validation.cpp`, so it might be worth normalizing the unsupported classification label before stamping the deployment default. A test like `effective type == LLM => labels contain chat` would also protect the main goal of #3070 nicely.

2. Conflicting labels on fixed-modality recipes

Since `chat` wins in `find_deployment_mode()`, a user registration such as an `sd-cpp` model with an explicit `chat` label can end up with both `chat` and the descriptor-provided `image` label and then resolve as an LLM, even though the backend only serves image generation.

Since the new contract says a model has one deployment mode, maybe conflicting deployment labels should be rejected/normalized at registration, or fixed-modality recipes should retain their backend-defined mode.

I would no ignore what codex has brought up.

jeremyfowers:
I'm reworking a few things, will mark ready-for-review again when I'm done iterating.

jeremyfowers:
Made chat a first-class deployment label and gave each backend descriptor a single declared list of the deployment modes it can serve, so a model's labels can no longer claim a mode its backend would reject at inference time.

- Consolidated three descriptor fields into one. supported_modes (ordered, most-default first) replaces default_labels, default_classification, and the editorial modality string; default_capabilities holds the non-mode labels.
- Replaced the hardcoded classification special case with a general rule. get_deployment_model_type() is deleted; labels are reconciled against supported_modes at ingest, so labels and ModelType agree by construction on every path — including the built-in cache build, which previously bypassed normalization entirely.
- POST /pull rejects illegal definitions with 400. Applies to both the labels array and the embedding/reranking parameters, registers nothing, and rejects collection imports before any component registers; naming no mode remains valid and stamps the recipe's default.
- Migrated the modality consumers. /system-info derives the display string from the declared modes, and bench reads them directly instead of parsing that string back into a ModelType — which fixes llamacpp embedding and reranking benchmarks.
- Added a build-failing contract test. BackendModeContractTest compares each descriptor's declared modes against the capability interfaces its server class implements, verified to fail in both directions; the ingest cases cover both reviewer findings.

Will run automated AI review to make sure I didnt miss anything.

jeremyfowers:
Came back clean, please have another look!

fl0rianr (commented):
Thanks for the rework. I went through the current head again, and I think my two previous concerns are addressed cleanly now. The supported_modes approach and rejecting unsupported modes at registration make sense to me.

I still see two runtime issues before merge:

1. Multiple supported deployment modes can still be accepted together.
For example, llamacpp supports chat, embeddings, and reranking, so labels: ["chat", "embeddings"] keeps both. chat wins type resolution, meaning the model loads as an LLM, while LlamaCppServer::load() only enables --embeddings for ModelType::EMBEDDING. The labels can therefore advertise an endpoint the subprocess was not configured to serve. I agree with the new Codex finding here. Since the contract is one deployment mode per model, I think ingest should reject multiple distinct deployment modes.

2. FastFlowLM has the same kind of mismatch for reranking.
The descriptor declares reranking as a supported deployment mode, but fastflowlm_models.cpp explicitly says /v1/rerank is served by a chat model and there is no reranking mode to detect. FastFlowLMServer::reranking() also rejects requests unless model_type_ == LLM. So recipe: "flm", labels: ["reranking"] is accepted, becomes ModelType::RERANKING, and then cannot use its reranking handler.

This also means BackendModeContractTest cannot always equate an implemented capability interface with a standalone deployment mode. For FLM, removing reranking from supported_modes seems consistent with the runtime unless a separate reranking mode is intended.

The second Codex finding around collecting removed labels after std::remove_if also looks valid, although that seems limited to the error message rather than routing correctness.

Small docs nit: openai.md still says unsupported classification is dropped, while new registrations now reject unsupported modes. It also says every model has exactly one deployment mode before describing chat + transcription.

Overall I like the direction a lot more now. The explicit chat contract is a good foundation for #3070. I think fixing the two mode-selection cases above would make that invariant hold end-to-end.

jeremyfowers:
> FastFlowLM has the same kind of mismatch for reranking.

There is a bigger problem! FastFlowLM doesn't support reranking whatsoever. Removing that from the supported_modes list.

jeremyfowers:
Thanks — both are fixed, and the first one turned into a bigger change than a reject.

**1. Multiple deployment modes.** You and Codex were right that `["chat", "embeddings"]` was accepted. Rather than reject just that case, an illegal label set is now refused outright instead of being normalized down to a winner: a mode the recipe cannot serve, *or* two different modes, is an error. Nothing repairs a label set anymore, which removes the "which mode wins" question entirely — so labels and `ModelType` agree by construction rather than by later correction.

- `POST /pull` answers 400 and registers nothing. The legacy `embedding`/`reranking` booleans count as mode claims, so the rule can't be sidestepped by spelling.
- A stored `user_models.json` entry that predates the rule is skipped at startup with an error naming it, and the file is left untouched so it can be corrected by hand. Guessing which of its claims was meant would make every consumer trust an answer nobody wrote.
- Aliases of one mode (`embedding` beside `embeddings`) are one claim, and capability labels (`realtime-transcription`, `chat-transcription`) were never modes.

This deleted more than it added: the drop/keep logic, the "which mode wins" precedence question, and the two-branch error message are gone. It also makes the Codex `std::remove_if` finding moot — there's no compaction left to get wrong.

**2. FastFlowLM reranking.** Removed, along with `IRerankingServer` and the `reranking()` override on `FastFlowLMServer`. There is no separate reranking mode intended — FLM has no rerank endpoint at all. `flm-real` v0.9.46 registers only `/v1/chat/completions`, `/v1/completions`, `/v1/embeddings`, `/v1/audio/transcriptions`, `/v1/models`, `/v1/version`; the string `rerank` appears nowhere in the binary, and no model in `flm list --json` carries a reranking label. `FastFlowLMServer::reranking()` was forwarding to a 404. A rerank request against an FLM model now fails in the router's `dynamic_cast<IRerankingServer*>` with the same clean unsupported-capability error.

`BackendModeContractTest` passes on the truthful pair, so your point about it not always equating an interface with a mode holds here by removing the interface rather than loosening the test.

**3. Docs.** Fixed both: the `chat` + `transcription` omni sentence contradicted the one-mode rule (that case is `chat` + the `chat-transcription` capability), and the classification paragraph still said the label is dropped. `lemonade.md` now shows both 400s, verified against a running server.

Tests: multi-mode and legacy-flag rejection at registration in `ModelManagerCollectionValidationTest`; legal/illegal ingest cases plus stamping idempotence in `BackendModeContractTest`.

Not addressed, deliberately: the `ChatWindow.tsx` finding. The `chat` + `transcription` model it describes can no longer be registered, and `src/app` is out of scope for this PR.

fl0rianr (commented):
Thanks, I checked the latest head again. The previous multi-mode and FLM reranking issues look fixed to me.

I found one remaining hole in the deployment-mode invariant:

Collections currently accept any single deployment mode. Since recipes without a backend descriptor make serves() return true, collection.omni + labels:["transcription"] is considered valid, does not get chat stamped, and becomes ModelType::TRANSCRIPTION. BackendModeContractTest explicitly tests this as legal.

But both collection.omni and collection.router are chat-facing virtual models. They are invoked through /chat/completions, and the new clients rely on the positive chat label. So an explicitly non-chat collection can still disappear from chat selection, which seems to break the #3070 invariant.

I think the known collection recipes should effectively be chat-only at the parent level: bare collections get chat, and another deployment label on the collection itself is rejected.

One upgrade question as well: current main intentionally treats classification on llamacpp/flm/vllm as an LLM-as-classifier by ignoring that deployment claim. The new loader skips such persisted user models entirely. Since the old meaning of this specific case is deterministic, would it be worth migrating that legacy case to chat on read while keeping new registrations strict?

Minor cleanup: test_model_type_classifier.cpp and the comment in model_types.h still describe chat + transcription/embeddings/... precedence as valid behavior, while the new ingest contract rejects those combination

fl0rianr (approved):
I reconsidered if I can be ok with this specific collection case. I agree we can treat explicitly non-chat collections as an unsupported configuration for now rather than adding more validation in this PR.

My only concern was that the current contract test explicitly treats collection.omni + transcription as legal, which makes it look like supported behavior. I don't think this needs to block the PR, but I'd avoid codifying that case as part of the supported contract.

With that understood, I'm good with the current direction. The issues I was concerned about on the supported paths are addressed.

PR #3101, opened 2026-08-12
Add route_only request option: routing decision without inference (dry-route)

### LINKED ISSUE #3098: Router: route_only request option — return the routing decision without running inference (dry-route)
## Problem

There is no way to ask the router engine "which candidate would you pick for this request?" without actually running inference. Today the workaround is registering a shadow policy whose candidates are stand-in models, sending the request with `max_tokens: 1`, and reading `x-lemonade-route` — wasteful, fragile, and it still loads/invokes a model.

Decision-only access is generally useful:

- **Policy debugging/iteration** — edit a policy, replay a set of prompts, diff the decisions; no inference cost.
- **CI for policies** — assert routing decisions for representative requests in a test suite.
- **Cost pre-estimation** — predict the local/cloud split of a workload before running it.
- **External evaluation harnesses** — router benchmarks (e.g. RouterArena-style frameworks) need `prompt → chosen model` at scale.

## Proposal

A `route_only` request option on the OpenAI-compatible endpoints that already perform router dispatch (`chat/completions`, `completions`, `responses`):

```jsonc
POST /api/v1/chat/completions
{ "model": "user.SWE-Router", "route_only": true, "route_trace": true,
  "messages": [...] }
```

Behavior: run `apply_router_collection_dispatch` exactly as today (rules, classifiers, LLM-as-router — decisions are engine-real, and classifier helper models may load as usual), then **return the decision instead of forwarding to a backend**:

```jsonc
200 OK, x-lemonade-route: <rule-id|default>
{
  "requested_model": "user.SWE-Router",
  "selected_model": "Qwen3.6-27B-UD-Q4",
  "routed": true,
  "decision": { "version": "1", "route_to": "...", "matched_rule": "...",
                "default_used": false, "outputs": {}, "trace": [...] }
}
```

- `decision` reuses the existing `route_decision_to_json()` shape; `trace` appears when `route_trace: true` is also set.
- On a non-router model, respond `{"requested_model": m, "selected_model": m, "routed": false}` — uniform for harness callers.
- The selected model is **not** auto-loaded and no completion runs.

## Alternative considered

A dedicated `POST /v1/route` endpoint — rejected: duplicates request parsing/normalization, needs quad-prefix registration, and drifts from the dispatch actually used at serving time. Keeping it a request option guarantees the answer is exactly what a real request would do.

## Out of scope

`/v1/messages` gains this automatically once #3087 (Anthropic-path dispatch) lands.

ramkrishna2910:
Closes #3098.

Adds a `route_only: true` request option to `chat/completions`, `completions`, and `responses`: the server runs `collection.router` dispatch exactly as for a real request (rules, classifiers, LLM-as-router — classifier helpers load as usual), then returns the decision instead of forwarding to a backend. The selected candidate is not loaded and no completion runs.

- Response: `{requested_model, selected_model, routed, decision}` with `decision` in the existing `route_decision_to_json` shape; `x-lemonade-route` header set as usual; `decision.trace` appears with `route_trace: true`; non-router models answer `routed: false` uniformly.
- Complementary to `POST /routing/validate` (ad-hoc policy document + synthetic features): `route_only` answers what the server would do with a real request against a registered router. Documented in `docs/dev/router-policy.md`.
- Tests: new `test_021za` in `server_endpoints.py` covers rule match, trace opt-in, all three endpoints, `routed: false`, and asserts `all_models_loaded` stays empty (no load). Existing router dispatch tests and 13/13 C++ routing unit tests pass.

https://claude.ai/code/session_016yHZizzdKkiPz3WbPpUHu1

### DIFF
--- a/docs/dev/router-policy.md
+++ b/docs/dev/router-policy.md
@@ -212,11 +212,15 @@ is **not** loaded and no completion runs:
 ```
 
 The `x-lemonade-route` header is set as usual, `decision.trace` appears when
-`"route_trace": true` is also sent, and `stream` is ignored. On a model that is
-not a router collection the response is
+`"route_trace": true` is also sent, and `stream` is ignored. On a **registered**
+model that is not a router collection the response is
 `{"requested_model": m, "selected_model": m, "routed": false}` — uniform for
-callers sweeping mixed model lists. Use this for policy debugging and CI,
-pre-estimating a workload's local/cloud split, and router evaluation harnesses.
+callers sweeping mixed model lists. `routed: false` means exactly that; an
+unknown model or a router whose dispatch fails returns the same error a real
+request would, and a non-boolean `route_only` value fails closed with `400`
+(never silently degrading into a real completion). Use this for policy
+debugging and CI, pre-estimating a workload's local/cloud split, and router
+evaluation harnesses.
 
 Related: `POST /api/v1/routing/validate` evaluates an **unregistered policy
 document** against caller-supplied features (`prompt`, `has_tools`,
--- a/docs/dev/router-policy.md
+++ b/docs/dev/router-policy.md
@@ -192,6 +192,38 @@ also carries an **`x_lemonade_route`** object:
 `route_to` is the candidate that actually answered. For streaming responses the
 same object is attached to the first SSE event as `x_lemonade_route`.
 
+## Dry-route: decision without inference
+
+Add `"route_only": true` to a `chat/completions`, `completions`, or `responses`
+request and the server runs dispatch exactly as it would for a real request —
+rules, classifiers, LLM-as-router; classifier helper models load as usual — then
+returns the decision instead of forwarding to a backend. The selected candidate
+is **not** loaded and no completion runs:
+
+```json
+{
+  "requested_model": "user.My-Router",
+  "selected_model": "Big-GGUF",
+  "routed": true,
+  "decision": { "version": "1", "route_to": "Big-GGUF",
+                "matched_rule": "coding-or-long-to-big",
+                "default_used": false, "outputs": {} }
+}
+```
+
+The `x-lemonade-route` header is set as usual, `decision.trace` appears when
+`"route_trace": true` is also sent, and `stream` is ignored. On a model that is
+not a router collection the response is
+`{"requested_model": m, "selected_model": m, "routed": false}` — uniform for
+callers sweeping mixed model lists. Use this for policy debugging and CI,
+pre-estimating a workload's local/cloud split, and router evaluation harnesses.
+
+Related: `POST /api/v1/routing/validate` evaluates an **unregistered policy
+document** against caller-supplied features (`prompt`, `has_tools`,
+`metadata`). `route_only` answers the complementary question — what the
+server would do with a real request against a **registered** router, including
+the request-feature extraction a live request gets.
+
 ## Cloud candidates
 
 A candidate may be a cloud model — the router derives the route category from the
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -93,10 +93,10 @@ Malformed requests (invalid JSON, missing `input`/`text`, non-string fields, non
 <sub>![Status](https://img.shields.io/badge/status-experimental-orange)</sub>
 
 Naming a registered `collection.router` model in the `model` field of a
-`chat/completions` or `completions` request triggers the routing engine: the
-server picks a candidate by the policy's first-matching rule (fail-open to
-`default_model`) and forwards the request to it. No dedicated endpoint or `"auto"`
-model is involved.
+`chat/completions`, `completions`, or `responses` request triggers the routing
+engine: the server picks a candidate by the policy's first-matching rule
+(fail-open to `default_model`) and forwards the request to it. No dedicated
+endpoint or `"auto"` model is involved.
 
 The decision is reported on the response:
 
@@ -105,6 +105,13 @@ The decision is reported on the response:
   response body: `{ route_to, matched_rule, default_used, outputs, trace[] }`
   (`route_to` is the candidate that answered). For streaming responses it is
   attached to the first SSE event.
+- Request field **`route_only: true`** (dry-route) returns the decision instead
+  of running a completion: `{ requested_model, selected_model, routed,
+  decision }`. The selected candidate is not loaded, `stream` is ignored, and
+  `route_trace` composes as above. A registered non-router model answers
+  `routed: false`; an unknown model or a dispatch failure returns the same
+  error a real request would. A non-boolean `route_only` value is rejected
+  with `400`.
 
 See [Router Policies](../dev/router-policy.md) for authoring the policy.
 
--- a/src/cpp/include/lemon/server.h
+++ b/src/cpp/include/lemon/server.h
@@ -124,6 +124,9 @@ private:
     // If request_json addresses a collection.router model, rewrite its "model"
     // field in place to the engine-selected candidate and return the Decision.
     // No-op otherwise.
+    void respond_route_only(const nlohmann::json& request_json,
+                            const std::optional<RouterDispatchResult>& dispatch,
+                            httplib::Response& res);
     std::optional<RouterDispatchResult> apply_router_collection_dispatch(
         nlohmann::json& request_json);
     // Union of routing-helper models across every active router collection's
--- a/src/cpp/include/lemon/server.h
+++ b/src/cpp/include/lemon/server.h
@@ -124,9 +124,12 @@ private:
     // If request_json addresses a collection.router model, rewrite its "model"
     // field in place to the engine-selected candidate and return the Decision.
     // No-op otherwise.
-    void respond_route_only(const nlohmann::json& request_json,
-                            const std::optional<RouterDispatchResult>& dispatch,
-                            httplib::Response& res);
+    // Fully handles a route_only request: flag validation (fails closed on
+    // non-boolean values), a single model resolution reused for dispatch, and
+    // the decision or error response. Returns true when a response was written
+    // (the caller must return); false when the request is not route_only (a
+    // literal `false` is erased and normal forwarding proceeds).
+    bool try_respond_route_only(nlohmann::json& request_json, httplib::Response& res);
     std::optional<RouterDispatchResult> apply_router_collection_dispatch(
         nlohmann::json& request_json);
     // Union of routing-helper models across every active router collection's
--- a/src/cpp/resources/schemas/README.md
+++ b/src/cpp/resources/schemas/README.md
@@ -9,7 +9,7 @@ decision `trace` — never in the engine.
 | Schema | Describes |
 |--------|-----------|
 | `route_policy.schema.json` | The `routing` block embedded in a `collection.router` collection JSON. Invoked like `collection.omni`: point the OpenAI `model` field at the collection name. |
-| `request.schema.json` | The request-side extension fields on the OpenAI chat body: `metadata` (string-valued routing inputs) and the optional `route_trace`. Validates only those fields (`additionalProperties: true`); no `version` (rides on the stock OpenAI request). |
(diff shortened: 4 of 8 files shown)


docs/dev/router-policy.md
```diff
@@ -192,6 +192,38 @@ also carries an **`x_lemonade_route`** object:
 `route_to` is the candidate that actually answered. For streaming responses the
 same object is attached to the first SSE event as `x_lemonade_route`.
 
+## Dry-route: decision without inference
+
+Add `"route_only": true` to a `chat/completions`, `completions`, or `responses`
```
ramkrishna2910, line 197: Please add `route_only` to `src/cpp/resources/schemas/request.schema.json` (and update that schema's description plus the schemas README/lock as required). This file is the published schema for Lemonade-specific routing request fields, so leaving it unchanged means clients generated or validated from the repository contract will not discover/type-check the new option even though the server and prose docs support it.

ramkrishna2910 (commented):
The implementation otherwise matches #3098's requested endpoint behavior. I found one request-contract omission that should be addressed.

docs/dev/router-policy.md
ramkrishna2910: Done in f5cbdcfc: `route_only` added to `request.schema.json` (typed boolean with dry-route semantics in the description, incl. the `routed: false` non-router shape and `route_trace` composition), root schema description and schemas README updated, and `schema-lock.json` refreshed via `test/test_schema_lock.py --update` in the same diff (v1 is unreleased so the reviewed lock refresh is the sanctioned path). `test_schema_lock.py` (2/2) and `test_routing_fixtures.py` (6/6) pass.

fl0rianr (commented):
Overall this looks solid and matches #3098 well. The scope is focused, the schema update is now included, and the happy path is well covered across all three endpoints.

I found two things I think should be fixed before merge:

1. `route_only` should fail closed on invalid types. Right now values like `"true"`, `1`, or `null` are silently removed and the request continues normally, which can unexpectedly trigger real model loading/inference. Since this flag explicitly means "decision without inference", an invalid value should return `400` instead.

2. `routed: false` currently also covers cases where routing failed, not only valid non-router models. `apply_router_collection_dispatch()` returns `nullopt` for unknown models, missing policies, and caught routing errors too, so those can become a successful `200 { routed: false }`. It would be safer if `routed: false` only meant "this model exists and is not a router", while actual lookup/dispatch failures keep their normal error behavior.

Small docs follow-up: `docs/api/lemonade.md` still only mentions `chat/completions` and `completions` in the routing section and doesn't document `route_only` / `responses` yet. The request schema description also still refers specifically to a chat-completions body.

With the two error-handling points addressed, this looks very close to approval. Nice work overall.

ramkrishna2910:
Both points addressed in 1cee3bfd, thanks for the careful read:

1. **Fail closed on invalid types** — `route_only` is now validated before dispatch: any non-boolean value (`"true"`, `1`, `null`, ...) returns `400 invalid_request_error` instead of being silently dropped into a real completion. Covered in the test for all three literal shapes.

2. **`routed: false` reserved for registered non-routers** — the dry-route responder now distinguishes the nullopt cases: missing `model` → 400; unknown model → the same `create_model_error` / 404 a real request gets; a router whose dispatch produced no decision → 500 `server_error` instead of a disguised `routed: false`. Only "exists and is not a router collection" answers `200 routed: false`. Test asserts the unknown-model 404 carries no `routed` field.

Docs follow-up also done: `docs/api/lemonade.md` routing section now lists `responses` and documents `route_only` with the error semantics, and the request-schema description was broadened from chat-completions-specific to the three OpenAI request bodies (schema lock refreshed in the same diff).

Note: CI runs for this push were cancelled on purpose — runners are reserved for something else right now; will re-run the suite before undrafting.

fl0rianr:
Thanks, are you sure you have them addressed, sorry, got me a bit confused here.

SlawomirNowaczyk (changes requested):
Overall looks good, just a couple of minor suggestions for improvement:

### 1. Surface the real dispatch failure instead of a generic 500
In `server.cpp:3069-3073`, `apply_router_collection_dispatch` catches the routing exception, logs `e.what()` at WARNING, and returns `nullopt`. `respond_route_only` then produces a generic `500 "router dispatch failed for '<model>'"` with the actual reason discarded. For a *diagnostic* endpoint whose whole purpose is "tell me what the router would decide", the failure reason is the most valuable payload. Consider propagating the error message (e.g. capture it and include it in the 500 body) so callers can debug their policy instead of seeing an opaque error.

### 2. Avoid the redundant/TOCTOU double model lookup
When `dispatch` is `nullopt`, `server.cpp:3013-3032` re-runs `model_exists()`, `get_model_info()`, and `is_router_collection_recipe()` — all already performed by the caller (`apply_router_collection_dispatch` or the inline chat path). Besides the duplicated work, the router serves concurrent requests (invariant #8), so state can change between the two lookups. Passing the already-resolved `ModelInfo` (or a small result enum) into `respond_route_only` removes the second lookup and the race window.

### 3. Build error bodies with `nlohmann::json`, not raw string literals
`server.cpp:2971-2977` and the missing-model branch of `respond_route_only` emit hand-written `R"({"error": ...})"` literals, while the same function's 500 branch and the rest of the file use structured `nlohmann::json` / `create_model_error`. The raw literals are inconsistent and fragile (any future interpolation risks manual-escaping bugs). Prefer the structured builders everywhere for uniform error shapes.

### 4. De-duplicate the three-handler wiring
The sequence `take_route_only_flag` → `apply_router_collection_dispatch` → `if (route_only) respond_route_only(...)` is byte-for-byte identical in `server.cpp:3394-3411` and `server.cpp:5057-5074` (chat differs only by the omni branch). Extracting a single helper (e.g. `try_handle_route_only(...) -> bool`) reduces the risk of the three copies drifting as routing evolves.

### 5. Make the "response already written" contract harder to misuse
`take_route_only_flag` returns `bool` meaning "not rejected" while also writing to `res` and setting an out-param `route_only`. Every call site must remember the `if (!take_route_only_flag(...)) return;` guard; a forgotten guard silently proceeds into real inference — exactly what the flag is meant to prevent. A clearer contract (e.g. return `std::optional<bool>` where `nullopt` = "rejected, response written", or an explicit result enum) makes the failure mode self-documenting across all three call sites.

ramkrishna2910:
@SlawomirNowaczyk all five addressed in d39f224f — the shape you suggested in points 2/4/5 turned out to compose into one change:

The three per-handler sequences are now a single `Server::try_respond_route_only(request_json, res)` member; each handler reduces to `if (try_respond_route_only(...)) return;` right after model-name normalization.

1. **Real failure reasons** — dispatch runs inside the helper: an exception becomes `500 {"error":{"message":"router dispatch failed for '<m>': <e.what()>",...}}`, a decision-less return becomes `...: no routing decision (no parsed policy)`, and `RouterResidencyConflictException` serializes as the standard 409 instead of a generic 500.
2. **Single lookup, no TOCTOU** — one `get_model_info()` is reused for both the recipe check and `route_collection_request(request_json, info)`; the decision reflects one registry snapshot.
3. **Structured error bodies** — the raw string literals are gone; all errors go through a small `set_json_error()` builder or `create_model_error`.
4. **De-duplication** — as above; the chat handler also lost its special omni guard (the helper's exists-and-not-a-router path covers it).
5. **Misuse-resistant contract** — the flag extraction is a tri-state enum (`kAbsent/kEnabled/kInvalid`) fully internal to the helper; call sites only see "handled → return". A forgotten guard can no longer silently continue into inference, because the invalid-value rejection and the decision response are written by the same function that reports handled.

Verified: `test_021za` (route_only incl. fail-closed and error-contract assertions), `test_021z`/`test_021zh` (normal dispatch), and 13/13 C++ routing unit tests pass.

ramkrishna2910:
@fl0rianr sorry for the confusion — both of your points did land, and the code has since been consolidated per Slawomir's review, so the current state is easiest to check in one place: `Server::try_respond_route_only` in `src/cpp/server/server.cpp` (as of d39f224f). Concretely: (1) fail-closed — a non-boolean `route_only` hits the `kInvalid` branch and returns `400 invalid_request_error` before any dispatch or model loading; `test_021za` asserts 400 for `"true"`, `1`, and `null`. (2) `routed: false` only for registered non-routers — the helper resolves the model once: unknown → the normal `create_model_error`/404 path (test asserts the 404 body carries no `routed` field), router-with-failed-dispatch → 500 carrying the dispatch failure reason (or 409 for residency conflicts), and only "exists and is not a router collection" answers `200 routed: false`. Your docs note was also addressed earlier: `docs/api/lemonade.md` routing section documents `responses` + `route_only`, and the request-schema description now covers all three OpenAI request bodies.

fl0rianr (approved):
Rechecked the latest head, thanks. The refactor looks good to me and I don’t see a functional blocker anymore.

Just to clarify my earlier nits: I think there are still a couple of small cleanup items left:

the old comment above try_respond_route_only() in server.h still describes rewriting the model field, which no longer really matches that helper
the schemas README still says “OpenAI chat body”, although the request extension also applies to completions and responses

I also noticed the model_exists() → get_model_info() sequence still leaves a small check-then-use window, so I wouldn’t quite call the TOCTOU @SlawomirNowaczyk pointed out as fully eliminated. None of these are blockers from my side though.

So thanks for the follow-up and let be me the first one to be approving.

jeremyfowers:

**Ready for review** — the checks below passed, and suggested reviewers are at the end.

**Attention level:** Routine — needs any 1 reviewer

### Review needed

**any 1 reviewer** — a field or parameter on an existing route. The diff adds a single request option (`route_only`) to existing HTTP endpoints (`chat/completions`, `completions`, `responses`), which maps to 'a field or parameter on an existing route' in the review requirement table, yielding `one-reviewer`.


_Reviewed at head `d39f224` by repo-manager (tier: reviewers). Regenerate with `repo-manager review-pr 3101`._

jeremyfowers:
Triage-bot is pleased!

SlawomirNowaczyk:
Thanks, all the points are well handled. A few minor items left before I approve:

1. **Stale doc comment** — in `server.h`, the block above `try_respond_route_only` ("…rewrite its `"model"` field in place… No-op otherwise.") actually describes `apply_router_collection_dispatch` (which currently has no comment). `try_respond_route_only` doesn't rewrite `model`, so please move that comment down to the right function.

2. **Schemas README** — `README.md` still says the extension fields ride on "the OpenAI chat body", even though `request.schema.json` was correctly broadened to "chat/completions, completions, or responses". Please match the README wording.

3. **TOCTOU / unguarded lookup** — in `try_respond_route_only`, `model_exists()` → `get_model_info()` is still two lookups, and `get_model_info()` is outside the `try`. If the model is deleted in that window, `get_model_info()` throws and bubbles to the endpoint's generic catch → a raw `500 {"error": e.what()}` instead of the clean 404 `create_model_error`. Collapsing to a single guarded `get_model_info()` (map not-found → 404, keep an inner `try` for dispatch → 500/409) closes both the race and the double lookup, which fully resolves the earlier TOCTOU note.

SlawomirNowaczyk (changes requested):
Thanks, all the points are well handled. A few minor items left before I approve:

1. **Stale doc comment** — in `server.h`, the block above `try_respond_route_only` ("…rewrite its `"model"` field in place… No-op otherwise.") actually describes `apply_router_collection_dispatch` (which currently has no comment). `try_respond_route_only` doesn't rewrite `model`, so please move that comment down to the right function.

2. **Schemas README** — `README.md` still says the extension fields ride on "the OpenAI chat body", even though `request.schema.json` was correctly broadened to "chat/completions, completions, or responses". Please match the README wording.

3. **TOCTOU / unguarded lookup** — in `try_respond_route_only`, `model_exists()` → `get_model_info()` is still two lookups, and `get_model_info()` is outside the `try`. If the model is deleted in that window, `get_model_info()` throws and bubbles to the endpoint's generic catch → a raw `500 {"error": e.what()}` instead of the clean 404 `create_model_error`. Collapsing to a single guarded `get_model_info()` (map not-found → 404, keep an inner `try` for dispatch → 500/409) closes both the race and the double lookup, which fully resolves the earlier TOCTOU note.

PR #3102, opened 2026-08-12
feat(app): add cost metadata fields to the Add Model form

Bekhouche:
Lets a user declare cost_tier / cost_input_per_million / cost_output_per_million / latency_ms_hint when registering a custom model, consumed by a collection.router policy's outputs.estimated_cost reporting and cost_select routing. Also extends the server's user-model registration allowlist (model_manager.cpp) so these fields actually persist to user_models.json instead of being silently dropped.

Related to #3078 and #3100 - this is the UI piece that lets a user actually set the cost metadata (`cost_tier` / `cost_input_per_million` / `cost_output_per_million` / `latency_ms_hint`) that `cost_select` (#3100) ranks candidates by, and that #3078's future budget-tracking proposal would also read. Without this, those fields could only be set by hand-editing `user_models.json`.

### DIFF
--- a/src/app/src/renderer/AddModelPanel.tsx
+++ b/src/app/src/renderer/AddModelPanel.tsx
@@ -17,6 +17,8 @@ export interface AddModelInitialValues {
   embedding?: boolean;
 }
 
+export type ModelCostTier = 'free' | 'low' | 'medium' | 'high';
+
 export interface ModelInstallData {
   name: string;
   checkpoint: string;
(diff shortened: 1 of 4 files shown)


jeremyfowers:
FYI development of the GUI on `main`, aka "GUI2", is frozen. We are currently beta testing a complete replacement, which we call GUI3. Please find me on the discord to learn more.

jeremyfowers:
Closing PR to save on CI runners, let's discuss on discord!

PR #2495, opened 2026-06-30
Update backend reference docs for new vLLM models

superm1:
Regenerated docs/dev/backends-reference.md to include recently added vLLM models:
- GLM-4.7-Flash-FP16-vLLM
- Qwen3.6-27B-FP16-vLLM
- Qwen3.6-35B-A3B-FP16-vLLM

This fixes the backend-docs-drift CI check failure.

PR #3103, opened 2026-08-12
docs(backends): drop the model catalog tables from the backend reference

jeremyfowers:
Removes the `## Models` section from `docs/dev/backends-reference.md` — ~285 lines of markdown tables (58% of the file) that duplicated `server_models.json` and that nothing links to.

**Why:** the user-facing model catalog at [lemonade-server.ai/models.html](https://lemonade-server.ai/models.html) fetches `server_models.json` live from GitHub raw at the release tag (`docs/assets/models.js:456`), so it never read these tables. The only inbound link to the file at all is `docs/dev/adding-a-backend.md:111`, which cites the *backend* appearing there, not the model list.

**Cost it was imposing:** every model-catalog PR had to rebuild `lemond` in the `backend-docs-drift` job and regenerate the doc, or fail CI. #2495 is the pure case — a PR whose entire content was three table rows, body: "This fixes the backend-docs-drift CI check failure."

**Removed:**
- the `## Models` heading and the `backend-models` generated region in `backends-reference.md`
- `render_models()` in `docs/tools/gen_backend_boilerplate.py`
- the `SERVER_MODELS` constant (dead once `render_models` is gone — it was its only user)
- the `backend-models` marker pair in `DEFAULT_TEMPLATE` and the `targets` entry that drove it
- the `server_models.json` mention in the module docstring

The descriptor-backed regions — `backends-overview`, `backends-matrix`, `backend-options` — are untouched. Those are derived from the C++ descriptors and can't be read any other way, so the drift check still earns its keep for them.

**Verification:** `gen_backend_boilerplate.py --check` passes against a locally built `lemond`; a full write-mode run is idempotent (no diff beyond these two files); Black clean; pre-commit hooks pass.


ramkrishna2910 (approved):

jeremyfowers:
Requires https://github.com/lemonade-sdk/lemonade/pull/3090

PR #3104, opened 2026-08-12
feat(mcp): add Streamable HTTP client transport

fl0rianr:
## Summary

Adds Streamable HTTP support to the server-side MCP client host while keeping the existing stdio transport.

This allows Lemonade to connect directly to external MCP HTTP endpoints instead of requiring every external server to run as a local process.

## Changes

- add `streamable-http` as an MCP client transport
- keep existing `stdio` configs backwards compatible
- support MCP initialization and protocol negotiation over HTTP
- handle JSON and SSE responses
- track and clean up `Mcp-Session-Id` sessions
- support tool discovery and tool calls over HTTP
- respond to server-initiated MCP requests such as `ping`
- add URL and port validation
- require HTTPS for non-local endpoints unless insecure HTTP is explicitly enabled
- support bearer authentication through `${VARIABLE}` environment references
- add `POST /internal/mcp/servers/test` for testing a configuration without persisting it
- handle expired HTTP sessions and disconnect cleanup

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -37,6 +37,103 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/internal/aliases`](#get-internalaliases) | List all active model aliases |
 | `POST` | [`/internal/aliases`](#post-internalaliases) | Create or update a model alias |
 | `DELETE` | [`/internal/aliases/{alias}`](#delete-internalaliasesalias) | Remove a model alias |
+| `GET` | [`/internal/mcp/servers`](#external-mcp-client-host) | List configured external MCP servers and their current state |
+| `GET` | [`/internal/mcp/tools`](#external-mcp-client-host) | List tools exposed by connected external MCP servers |
+| `POST` | [`/internal/mcp/servers`](#external-mcp-client-host) | Create or update an external MCP server configuration |
+| `POST` | [`/internal/mcp/servers/test`](#external-mcp-client-host) | Test a server configuration without persisting it |
+| `DELETE` | [`/internal/mcp/servers/{id}`](#external-mcp-client-host) | Remove an external MCP server configuration |
+| `POST` | [`/internal/mcp/servers/{id}/connect`](#external-mcp-client-host) | Connect to an external MCP server |
+| `POST` | [`/internal/mcp/servers/{id}/disconnect`](#external-mcp-client-host) | Disconnect from an external MCP server |
+| `POST` | [`/internal/mcp/servers/{id}/refresh-tools`](#external-mcp-client-host) | Refresh the server tool catalogue |
+| `POST` | [`/internal/mcp/servers/{id}/tools/call`](#external-mcp-client-host) | Call a tool on an external MCP server |
+
+## External MCP client host
+<sub>![Status](https://img.shields.io/badge/status-experimental-orange)</sub>
+
+Lemonade can also act as an **MCP client host** for external MCP servers. This is separate from the [`POST /mcp`](./mcp.md) gateway, where Lemonade itself acts as an MCP server.
+
+External servers are managed through the `/internal/mcp/*` administration API. The client host supports two transports:
+
+- **`streamable-http`** for remote or local MCP HTTP endpoints.
+- **`stdio`** for locally launched MCP processes. Existing configurations remain backward compatible because an omitted `transport` still defaults to `stdio`.
+
+### Streamable HTTP configuration
+
+A Streamable HTTP server configuration uses the following fields:
+
+| Field | Required | Description |
+|-------|----------|-------------|
+| `id` | no on create/test | Stable server identifier. When omitted for create/test, Lemonade derives one from the name or URL. |
+| `name` | no | Human-readable name. Defaults to the resolved id. |
+| `transport` | yes | Use `streamable-http`. `http` is accepted as an alias and normalized to `streamable-http`. |
+| `url` | yes | MCP endpoint URL. HTTP and HTTPS are supported. |
+| `bearer_token` | no | Bearer credential as a `${VARIABLE}` environment reference. Raw tokens are rejected and are never persisted. |
+| `allow_insecure_http` | no | Allow plain HTTP to a non-loopback host. Plain HTTP to localhost/loopback is allowed without this flag. Defaults to `false`. |
+| `enabled` | no | Whether the configured server may be connected. Defaults to `true`. |
+| `timeout_ms` | no | Request timeout in milliseconds, from 1000 to 300000. Defaults to 30000. |
+
+Streamable HTTP configurations must not include stdio-only fields such as `command`, `args`, `env`, or `working_dir`.
+
+Example:
+
+```json
+{
+  "server": {
+    "id": "example-mcp",
+    "name": "Example MCP",
+    "transport": "streamable-http",
+    "url": "https://example.com/mcp",
+    "bearer_token": "${EXAMPLE_MCP_TOKEN}",
+    "timeout_ms": 30000,
+    "enabled": true
+  }
+}
+```
+
+### Local stdio configuration
+
+The existing stdio transport remains available for MCP servers that Lemonade should launch and supervise locally:
+
+```json
+{
+  "server": {
+    "id": "local-mcp",
+    "name": "Local MCP",
+    "transport": "stdio",
+    "command": "python",
+    "args": ["/absolute/path/server.py"],
+    "env": {
+      "SERVICE_TOKEN": "${SERVICE_TOKEN}"
+    },
+    "timeout_ms": 30000,
+    "enabled": true
+  }
+}
+```
+
+Values in `env` must also be `${VARIABLE}` references. HTTP-only fields such as `url`, `bearer_token`, and `allow_insecure_http` are rejected for stdio configurations.
+
+### Administration endpoints
+
+| Method | Endpoint | Purpose |
+|--------|----------|---------|
+| `GET` | `/internal/mcp/servers` | List configured servers and current connection state. |
+| `GET` | `/internal/mcp/tools` | List tools exposed by connected external servers. |
+| `POST` | `/internal/mcp/servers` | Create or update a persisted server configuration. |
+| `POST` | `/internal/mcp/servers/test` | Connect, initialize, discover tools, disconnect, and return the result without persisting the draft. |
+| `DELETE` | `/internal/mcp/servers/{id}` | Remove a persisted configuration. |
+| `POST` | `/internal/mcp/servers/{id}/connect` | Connect and perform the MCP initialization handshake. |
+| `POST` | `/internal/mcp/servers/{id}/disconnect` | Disconnect and clear local connection state. |
+| `POST` | `/internal/mcp/servers/{id}/refresh-tools` | Refresh `tools/list`. |
+| `POST` | `/internal/mcp/servers/{id}/tools/call` | Call a remote tool. The body requires `name` and may include an `arguments` object. |
+
+These are internal administrative routes and follow Lemonade's existing authentication rules for `/internal/*` endpoints.
+
+### Streamable HTTP behavior
+
+For Streamable HTTP connections, Lemonade performs the MCP initialization flow, negotiates a supported protocol version, and then sends `notifications/initialized`. It supports both JSON responses and `text/event-stream` responses, carries the negotiated `MCP-Protocol-Version` header, tracks `Mcp-Session-Id`, and performs best-effort session cleanup with HTTP `DELETE` on disconnect.
+
+If an HTTP session expires and the server returns `404`, Lemonade invalidates the local session so a later operation can establish a fresh connection. Incoming server requests are handled while waiting for a response; `ping` is answered directly and unsupported methods receive a JSON-RPC method-not-found response.
 
 ## `POST /v1/classify`
 <sub>![Status](https://img.shields.io/badge/status-experimental-orange)</sub>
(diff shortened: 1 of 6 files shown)


docs/api/lemonade.md
```diff
@@ -37,6 +37,103 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
...
 | `DELETE` | [`/internal/aliases/{alias}`](#delete-internalaliasesalias) | Remove a model alias |
+| `GET` | [`/internal/mcp/servers`](#external-mcp-client-host) | List configured external MCP servers and their current state |
+| `GET` | [`/internal/mcp/tools`](#external-mcp-client-host) | List tools exposed by connected external MCP servers |
+| `POST` | [`/internal/mcp/servers`](#external-mcp-client-host) | Create or update an external MCP server configuration |
+| `POST` | [`/internal/mcp/servers/test`](#external-mcp-client-host) | Test a server configuration without persisting it |
+| `DELETE` | [`/internal/mcp/servers/{id}`](#external-mcp-client-host) | Remove an external MCP server configuration |
+| `POST` | [`/internal/mcp/servers/{id}/connect`](#external-mcp-client-host) | Connect to an external MCP server |
+| `POST` | [`/internal/mcp/servers/{id}/disconnect`](#external-mcp-client-host) | Disconnect from an external MCP server |
+| `POST` | [`/internal/mcp/servers/{id}/refresh-tools`](#external-mcp-client-host) | Refresh the server tool catalogue |
+| `POST` | [`/internal/mcp/servers/{id}/tools/call`](#external-mcp-client-host) | Call a tool on an external MCP server |
```
jeremyfowers, line 48: Why are these marked as "internal"? That indicates that only first party clients should be allowed to use it. But we should want lots of clients to use our MCP?

docs/api/lemonade.md
```diff
@@ -37,6 +37,103 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/internal/aliases`](#get-internalaliases) | List all active model aliases |
 | `POST` | [`/internal/aliases`](#post-internalaliases) | Create or update a model alias |
 | `DELETE` | [`/internal/aliases/{alias}`](#delete-internalaliasesalias) | Remove a model alias |
+| `GET` | [`/internal/mcp/servers`](#external-mcp-client-host) | List configured external MCP servers and their current state |
```
jeremyfowers, line 40: This PR adds 9 endpoints but 0 tests. Please ensure all endpoints are tested in CI.

jeremyfowers:
Will send for maintainer review once my basic questions are answered.

docs/api/lemonade.md
fl0rianr: Good point. These routes started as GUI-internal, but the intended scope is broader: third-party apps should also be able to let Lemonade connect to and manage external MCP servers, e.g. a CAD MCP server.

So I agree /internal is probably the wrong namespace. I think these should move to the versioned public API, e.g. /v1/mcp/*, while remaining admin-authenticated. /mcp would stay separate for external MCP clients using Lemonade's own tools.

Would /v1/mcp/* be the preferred API shape?

If so I will adapt that way, adjust the docs and add tests and once we merged play it back towards GUI_merging, ok?

jeremyfowers: @siavashhub can you provide guidance here?

jeremyfowers:
Pinged maintainer on discord; will assign someone else if I don't hear back.

While we wait, @fl0rianr: can you provide a brief explanation of _why_ lemond needs to connect to external MCP servers? I don't understand what lemond itself would do with the tools from these MCPs.

fl0rianr:
The intent is not that lemond itself decides when to use external MCP tools. The GUI / chat client (e.g. someone deploying lemond) is still the component driving tool use.

The reason for putting the MCP client in lemond is to make it the transport and execution layer for those tools. lemond owns the MCP connection, discovers the tools, maintains the session, and executes tools/call on behalf of a client. The client can then treat external MCP tools much like any other tool exposed by Lemonade without having to implement MCP transports itself.

That becomes super useful beyond GUI3. A browser client otherwise has CORS restrictions, a web client cannot launch stdio MCP servers at all, and every client would need its own MCP session/auth/lifecycle implementation. Keeping that in lemond gives us one implementation for HTTP and stdio and makes the same external tools usable from GUI, CLI, or other Lemonade clients. Only this way GUI and TUI can have somewhat equal features.

So the flow I have in mind is roughly:

external MCP server -> lemond MCP client host -> Lemonade client/chat -> model tool call -> lemond -> external MCP tool

lemond is therefore not the agent consuming the tools; it is the host that makes them available and executes them for the actual consumer.

docs/api/lemonade.md
siavashhub: > @siavashhub can you provide guidance here?

Code quality is great, thanks @fl0rianr for the strong implementation. 

My comments below are mostly about scope and process, not code quality.

I'd like to split this into (1) one direction question and (2) implementation notes if we proceed.

## 1. Direction: should the MCP *client host* live in core `lemond`?

For anyone reading: there are two separate MCP features here. 

- The `/mcp` gateway (Lemonade *as a server*, exposing our inference as tools to agents) 
- The current implementation in the PR which is the inverse: Lemonade *as a client* reaching out to external MCP servers. 

My concern is blast radius. `LemonadeServer.exe` is an always-on, auto-started background service whose job is local inference. Being an MCP client host adds outbound egress to arbitrary endpoints, stored third-party credentials and (via stdio) local process launching to that always-on surface. That's a meaningful and permanent expansion of what a compromised or misconfigured `lemond` can do.

I recognize the stdio foundation already merged, so this PR is an extension, not the introduction, which is exactly why I'd rather confirm the direction now, before outbound HTTP + credential storage entrench it further.

I also want to concede the strongest counterargument up front: a browser/web client can't spawn stdio servers and hits CORS on remote HTTP, so if web-app parity is a goal, something server-side has to host MCP. That's a real point.

Given both, my suggestion is a middle path: keep the MCP client host first-party but **optional**: a sidecar or build-flagged module, co-deployable for web/CLI parity, but not compiled into the always-on core by default. That keeps core `lemond`'s blast radius small while still unblocking the parity use case.

@jeremyfowers I'm happy to go with whatever code owners decide.

## 2. If we proceed, implementation notes

- **Tests (blocking).** Already raised and beyond `test/cpp/test_mcp_client_config.cpp` (which only exercises stdio), we could add: config-parse cases for HTTP (valid URL, bad port, non-loopback plain-HTTP rejection, `allow_insecure_http` opt-in, raw-bearer rejection, stdio+HTTP field conflicts), plus an integration test hitting `/internal/mcp/servers/test` against a mock HTTP MCP server (JSON and SSE responses, 404 session-expiry, `ping` handling).
- **`/internal/` vs `/v1/mcp/*`.** I'd keep it under `/internal/` for now, it's an admin control plane like `/internal/set` and `/internal/config`. If we ever promote it to the public API, it has to:

  - (a) be registered under all four public prefixes (`/api/v0/`, `/api/v1/`, `/v0/`, `/v1/`), not just `/v1/`
  - (b) carry the fail-closed admin gate with it. `server.cpp` currently requires an admin key for `/internal/mcp/*` even on a keyless server and that must not regress or a keyless server would expose SSRF + process launch. I'd not bundle that move into this PR.

Really nice work overall, my pushback is about where this lives, not how it's built.

fl0rianr:
@siavashhub Thanks for your input, this is a really important discussion here.
I also think it helped to clarify that there are actually two separate questions here.

First, the MCP client host itself is not new in this PR. #2605 already introduced the server-side MCP client host in lemond with stdio, tool discovery and tools/call. Yes it was work for GUI3 form the start but also protected by the "/internal" standard like admin key. This PR extends that existing architecture with Streamable HTTP. I still think that placement is the right one: lemond is the shared transport/execution layer, while the GUI, CLI or any other client decides when and how tools are used. Keeping that in one place is what gives browser, desktop, CLI and third-party clients the same MCP capabilities without each of them implementing stdio, HTTP sessions, auth and lifecycle independently.

Second, I think we should be more explicit about what "/internal" means.

There are **two opposite MCP roles**:

/mcp exposes Lemonade's own capabilities to external MCP clients. This is Lemonade acting as an MCP server.

```
/mcp
External MCP Client
        │
        ▼
     lemond
        │
        ▼
Lemonade capabilities
(models, chat, image, audio, ...)
```

/internal/mcp/* manages Lemonade's connections to external MCP servers. This is an administrative control plane for the MCP client host.
```
Lemonade GUI / CLI / external app
              │
              ▼
           lemond
      MCP client host
          │      │
          ▼      ▼
       stdio    HTTP
          │      │
          ▼      ▼
     external MCP servers
```
The second role is also much more powerful, but that is exactly the value of the MCP client host. Lemonade itself has a bounded set of capabilities; connecting it to external MCP servers makes those capabilities extensible to the broader MCP ecosystem. That is a significant increase in what Lemonade can do - IMO we need this! But I also fully agree that it should come with a correspondingly stronger security boundary and remain fail-closed behind admin authentication.

That distinction above is important. If I'm not mistaken, @jeremyfowers we absolutely want third parties to build their own apps and GUIs on top of lemond, and trusted third-party clients may need to use this control plane too - or the never could actually leverage what we "show off" in our GUI3. So internal should not mean "first-party clients only" (or we might have a small misunderstanding here what first party is). It should mean something closer to "administrative/control-plane API, not part of the versioned public API compatibility contract".

For MCP specifically I agree that this is a privileged surface. It can create outbound connections and, for stdio, launch local processes. So keeping /internal/mcp/* fail-closed behind LEMONADE_ADMIN_API_KEY makes sense and should remain an explicit invariant.

My preference would therefore be to keep the current architecture and namespace, add the missing HTTP tests, and use this PR to also document the /internal/* contract clearly for the first time. I think that closes the real gap here without introducing a sidecar, build-time variants, or moving a privileged control-plane API under /v1/mcp/*, where it could easily be confused with the completely different /mcp server endpoint.

Geramy:
> @siavashhub Thanks for your input, this is a really important discussion here. I also think it helped to clarify that there are actually two separate questions here.
> 
> First, the MCP client host itself is not new in this PR. #2605 already introduced the server-side MCP client host in lemond with stdio, tool discovery and tools/call. Yes it was work for GUI3 form the start but also protected by the "/internal" standard like admin key. This PR extends that existing architecture with Streamable HTTP. I still think that placement is the right one: lemond is the shared transport/execution layer, while the GUI, CLI or any other client decides when and how tools are used. Keeping that in one place is what gives browser, desktop, CLI and third-party clients the same MCP capabilities without each of them implementing stdio, HTTP sessions, auth and lifecycle independently.
> 
> Second, I think we should be more explicit about what "/internal" means.
> 
> There are **two opposite MCP roles**:
> 
> /mcp exposes Lemonade's own capabilities to external MCP clients. This is Lemonade acting as an MCP server.
> 
> ```
> /mcp
> External MCP Client
>         │
>         ▼
>      lemond
>         │
>         ▼
> Lemonade capabilities
> (models, chat, image, audio, ...)
> ```
> 
> /internal/mcp/* manages Lemonade's connections to external MCP servers. This is an administrative control plane for the MCP client host.
> 
> ```
> Lemonade GUI / CLI / external app
>               │
>               ▼
>            lemond
>       MCP client host
>           │      │
>           ▼      ▼
>        stdio    HTTP
>           │      │
>           ▼      ▼
>      external MCP servers
> ```
> 
> The second role is also much more powerful, but that is exactly the value of the MCP client host. Lemonade itself has a bounded set of capabilities; connecting it to external MCP servers makes those capabilities extensible to the broader MCP ecosystem. That is a significant increase in what Lemonade can do - IMO we need this! But I also fully agree that it should come with a correspondingly stronger security boundary and remain fail-closed behind admin authentication.
> 
> That distinction above is important. If I'm not mistaken, @jeremyfowers we absolutely want third parties to build their own apps and GUIs on top of lemond, and trusted third-party clients may need to use this control plane too - or the never could actually leverage what we "show off" in our GUI3. So internal should not mean "first-party clients only" (or we might have a small misunderstanding here what first party is). It should mean something closer to "administrative/control-plane API, not part of the versioned public API compatibility contract".
> 
> For MCP specifically I agree that this is a privileged surface. It can create outbound connections and, for stdio, launch local processes. So keeping /internal/mcp/* fail-closed behind LEMONADE_ADMIN_API_KEY makes sense and should remain an explicit invariant.
> 
> My preference would therefore be to keep the current architecture and namespace, add the missing HTTP tests, and use this PR to also document the /internal/* contract clearly for the first time. I think that closes the real gap here without introducing a sidecar, build-time variants, or moving a privileged control-plane API under /v1/mcp/*, where it could easily be confused with the completely different /mcp server endpoint.

I'm not sure about any of this, Lemond being both a MCP Server or a MCP Client as well as a engine manager.
I believe what we should be doing is separating these processes out honestly. Even at AAI Day they acknowledged that a server cpu is meant for agentic tasks and you must have the right hardware for the right job, that's a given and it goes for software packages.

So let's think about lemonade, we are expanding lemonade now into other areas but using the same core installer for all of it. I believe there might be a point where we need to do something like

```
sudo apt install lemonade-mcp

or

sudo apt install lemonade-mcps
```

Perhaps we create a install button and it helps the user auto install these features.
In no way should this be in the same service honestly, how do we define access to such a thing and limits?

I am going to follow up on this later today when I have more time, but with this I am seeing huge risk in storing credentials, and the ability to use lemond to basically DDoS other people by using tools and the "client" gets shielded from responsibility in said attack. These are my current initial thoughts.

1. lemond becomes a proxy hop.

@jeremyfowers @fl0rianr @siavashhub

jeremyfowers:
Opening a thread on the maintainers private channel, please dont send more replies here until that converges.

jeremyfowers (changes requested):
@fl0rianr what are the next steps for this PR?

PR #3110, opened 2026-08-13
Manage per-model recipe options without loading

jeremyfowers:
## Why

It was impossible to create a user-friendly model settings menu for GUI3 without these changes:
- No way to enable/disable auto-tune context size
- No way to save settings without also loading the model

## Details

Clients had no way to save a model's recipe options without loading it, no way to read back the effective options a load would actually use, and no way to set `ctx_size` back to automatic once a number was saved. This adds a per-model options sub-resource, following the `/models/{id}/files` pattern:

- `GET|POST|DELETE /v1/models/{id}/options`. `GET` returns `saved` / `effective` / `defaults`; `POST` validates, merges, and persists (a `null` value clears a key); `DELETE` erases the model's entry, resetting it to defaults.
- `effective` includes `model_name` and is the exact request body `POST /v1/load` uses for the model right now, replayable verbatim. The response also carries `load_command`: that request rendered as a runnable curl command, with the base URL left as `$LEMONADE_BASE_URL` and, on a key-protected server, the credential as `$LEMONADE_API_KEY`. `lemond` only ever listens over plain HTTP, so it cannot know the scheme a client reached it through — a TLS-terminating proxy is invisible to it, and the `Host` header is only the caller's own claim — and a wrong scheme in a command meant to be run is worse than a placeholder. Clients know their own base URL (per the many-clients-one-server invariant) and substitute the two tokens before display; a shell user exports them once.
- To make the replay a no-op, the router's reload check now compares resolved rather than stored option sets: a request that spells out a running model's own options keeps the process instead of restarting it. An automatically sized model is part of that: the live process holds the concrete size auto-tune chose, which no request can spell, so a requested `ctx_size` of `-1` is recognized as the size it already resolved to. This is the only `/v1/load` behavior change.
- `resolved_ctx_size` reports the concrete size an automatic `ctx_size` computes to right now, and `POST` accepts `dry_run: true` to validate and resolve a change without persisting it. Together these cover everything GUI3_merging's `/load/command` serves, so that endpoint can be deleted at the merge, with `EffectiveSettingsModal` showing `load_command` — substituting its own base URL for `$LEMONADE_BASE_URL`, which is a one-line replace on state the modal already holds.
- `ctx_size: -1` now persists as an explicit "size from available memory" choice, so a model can be pinned to automatic sizing even when the server-wide `ctx_size` is an explicit number. `-1` is the only spelling of automatic; strings like `"auto"` are rejected.
- Fixed `/v1/load` with `save_options` overwriting the in-memory `ModelInfo` with request-only options, which dropped the model's own registry options from that load.
- `pinned` is not settable here (it belongs to `/v1/load` and `/internal/pin`) and is omitted from `effective`/`defaults`.
- `parse_required_json_body` now answers 400 for a numeric literal too large for a double. `nlohmann`'s number-overflow error is an `out_of_range`, a sibling of `parse_error`, so it was escaping the parse handler as a 500 on every endpoint that reads a JSON body.

## Testing

Eight new tests in `test/server_endpoints.py`: saving without loading (`test_012m`), merge/clear/reset semantics (`test_012o`), input rejection including literals no `int64` can hold (`test_012p`), a saved `-1` overriding an explicit global `ctx_size` (`test_012r`), `effective` replaying verbatim as a `/load` body that keeps the running process, for both an explicit and an automatic `ctx_size` (`test_012t`), `resolved_ctx_size`/`load_command`/`dry_run` (`test_012u`), `load_command` never repeating the caller's `Host` header (`test_012v`), and a keyed server rendering the `Authorization` header its own `/v1/load` demands while keeping the key itself out of the response (`test_012w`). The existing `/load` regression tests (`test_012`–`test_012c`, `test_013`) were re-run alongside them.

`test_012t`'s automatic case was checked against a build with the router change reverted, where it fails on a changed backend PID, so the test is known to hold the fix down. The returned `load_command` was also executed through a shell with `LEMONADE_BASE_URL` exported, and loaded the model. Full `server_endpoints.py` run: 130 tests, OK. Verified on Linux with llama.cpp/Vulkan; labeled `ci:backends` since this touches load-path option resolution.


bitgamma (changes requested):
This is a really nice feature to have. A few findings from my friend DS4

 ### 1. load_command — Host-header injection and hardcoded http:// (security + correctness)                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                              
 server.cpp:                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                              
 ```cpp                                                                                                                                                                                                                                                                                                                       
   const std::string host = req.get_header_value("Host");                                                                                                                                                                                                                                                                     
   const std::string base = "http://" + (host.empty() ? "localhost:" + ... : host);                                                                                                                                                                                                                                           
   const std::string load_command =                                                                                                                                                                                                                                                                                           
       "curl -X POST " + base + "/v1/load -H \"Content-Type: application/json\" -d " +                                                                                                                                                                                                                                        
       shell_single_quote(effective_json.dump());                                                                                                                                                                                                                                                                             
 ```                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
 The Host header is embedded into a command string unquoted and unsanitized, and the docs explicitly instruct clients to "Run or display it verbatim" (the test even executes it through a shell). A Host value containing shell metacharacters (e.g. example.com; touch /tmp/x) is a command-injection vector in any         
 deployment where a client blindly runs the returned command. The JSON body is safely single-quoted, but the host/base is not. Recommend sanitizing/validating the host (and it should be derived from the server's own bound address/TLS config rather than the client-supplied header).                                     
                                                                                                                                                                                                                                                                                                                              
 Related correctness issue: the scheme is hardcoded to http://. Behind TLS or a reverse proxy terminating HTTPS (and not preserving the scheme), the returned curl command points at http and fails. X-Forwarded-Proto/X-Forwarded-Host aren't considered.                                                                    
                                                                                                                                                                                                                                                                                                                              
 ### 2. Replay no-op only works for explicit ctx_size — docs/behavior mismatch                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                              
 The PR description and docs claim effective/load_command is "replayable verbatim" and "keeps the running process." But the reload comparison happens before ctx auto-resolution, so for the default automatic case (ctx_size: -1) the live process holds a concrete resolved ctx while the replayed request carries -1 →     
 reload on every replay. The authors implicitly acknowledge this in test_012t: "An auto-sized model restarts on every /load, so pin an explicit size."                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                              
 This is a real gap between the headline feature and actual behavior for the most common (automatic) configuration. Either the docs/load_command description should state the limitation (no-op replay only when ctx_size is explicit), or the reload check should be extended to compare against the resolved ctx.           
                                                                                                                                                                                                                                                                                                                              
 ### 3. ctx_size overflow → 500 instead of 400                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                              
 validate_option_value does value.get<int64_t>(). A ctx_size numeric literal exceeding int64_t throws, which is caught by the outer try/catch in respond_with_model_options and reported as a 500 internal_error rather than a 400 validation error. Minor, but it's an input-validation path and should be a 400.            
                                                                                                                                                                                                                                                                                                                              
 ### 4. TOCTOU on registry options in write_saved_model_options                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                              
 info and registry_options are gathered before taking recipe_options_write_mutex_:                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                              
 ```cpp                                                                                                                                                                                                                                                                                                                       
   info = get_model_info(cache_key);              // lock released                                                                                                                                                                                                                                                            
   const json registry_options = ...registry_recipe_options(cache_key);                                                                                                                                                                                                                                                       
   std::lock_guard<std::mutex> write_lock(...);                                                                                                                                                                                                                                                                               
 ```                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
 The comment acknowledges this. If the registry JSON is edited between the read and the write, the in-memory cache update can drift from what the file write implies. Low severity (registry options are rarely mutated concurrently), but the in-cache rebuild uses a possibly-stale registry_options snapshot. Worth a      
 defensive re-read inside the lock if cheap.                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                              
 ### 5. Minor doc nit: pinned in saved                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                              
 The docs note that saved "can also hold keys this endpoint does not accept, such as pinned written by /v1/load." However, with the handle_load fix, /v1/load's save_options now persists options.to_json() (recipe keys only; pinned is handled separately and not included), so /v1/load no longer writes pinned into saved 
 options. The note appears stale unless some other path persists it. Consider rewording.

jeremyfowers:
Thanks — useful list. Three are fixed as described, one is fixed but the diagnosis was off, and one I think is not actually stale. Details, plus the two Codex items.

**1. `load_command` — Host injection and hardcoded `http://`.** Fixed by removing the guess rather than sanitizing it. `lemond` only ever constructs a plain `httplib::Server` — no `SSLServer`, no cert option anywhere in the tree — so it knows its own listener is HTTP and nothing beyond that: a TLS-terminating proxy is invisible to it, `X-Forwarded-Proto` is exactly as forgeable as `Host`, and a wrong scheme in a command the docs tell you to run is worse than no scheme at all. The command now reads `curl -X POST $LEMONADE_BASE_URL/v1/load …` and the caller substitutes — clients know their own base URL, which the many-clients-one-server invariant already says is client-side state. No caller-supplied text reaches the string, so the injection vector is closed by construction rather than by filtering. `test_012v` sends a `Host` of `evil.example.com; touch …` and asserts none of it survives.

**2. Replay no-op only for explicit `ctx_size`.** Fixed. The router now records whether a live process's `ctx_size` came from auto-tune, and a requested `-1` against an auto-sized process compares equal instead of forcing a reload. `test_012t` runs the replay for both an explicit and an automatic size, and the workaround comment you quoted is gone. I checked the automatic case against a build with the router hunk reverted — it fails there on a changed backend PID — so it holds the fix down rather than just passing.

**3. `ctx_size` overflow → 500.** Right conclusion, different cause. `value.get<int64_t>()` doesn't throw: `is_number_integer()` is true for unsigned as well, and nlohmann's `get_arithmetic_value` `static_cast`s rather than range-checks. The 500 comes from `json::parse` itself — `{"ctx_size": 1e400}` raises `out_of_range.406`, and `out_of_range` is a *sibling* of `parse_error`, so `parse_required_json_body`'s `catch (const json::parse_error&)` never saw it. It now catches `json::exception`, which fixes this for every endpoint that reads a JSON body, not just this one.

Chasing it turned up an adjacent bug worth having: because the cast doesn't range-check, `{"ctx_size": 18446744073709551615}` wrapped to `-1` and was accepted as "size it automatically" (and `2^64 + 4096` would have arrived as `4096`). `validate_option_value` now rejects anything above `INT64_MAX` before converting. Both literals are in `test_012p`.

**4. TOCTOU on registry options.** Fixed as suggested. `registry_recipe_options()` is split so a `_locked` variant can run inside `models_cache_mutex_`, and the in-cache rebuild re-reads the registry layer there instead of using a snapshot taken before the lock. `get_model_info` is still read outside — it takes the same mutex — but it no longer feeds the option layering.

**5. `pinned` in `saved`.** I don't think this one is stale, so I've left the note. `pinned` is a recipe option key for *every* recipe — `get_keys_for_recipe` appends it unconditionally, `recipe_options.cpp:85` — so `RecipeOptions(info.recipe, request_json)` picks it up out of a `/v1/load` body and `to_json()` carries it into `set_saved_model_options`. `POST /v1/load {"model_name": …, "pinned": true, "save_options": true}` still writes `pinned` into the model's `recipe_options.json` entry, and older entries can hold it regardless. Happy to be shown otherwise.

**Codex — auth header.** Fixed: the command carries `-H "Authorization: Bearer $LEMONADE_API_KEY"` when the server requires a key. The variable, not the value — the key shouldn't land in a response body that a GUI renders or a user screenshots. `test_012w` spawns a keyed `lemond` and asserts both halves.

**Codex — Windows quoting.** Documented rather than fixed. It's real for `cmd.exe`, and for PowerShell before 7.3 which mangles embedded quotes when re-quoting native arguments. But the server can't render for the client's shell: a client may be on a different OS than the `lemond` it is talking to, so platform-specific quoting would be its own guess. Since a client now substitutes the base URL anyway, it can render for its own platform, and the docs state what the shipped quoting assumes.

bitgamma (approved):
LGTM

PR #3111, opened 2026-08-13
fix(alias): report the real reason for alias validation failures

ramkrishna2910:
Three small fixes surfaced while verifying the v11.6.0 release checklist (#3107).

### 1. Alias validation errors reported the wrong reason

`handle_aliases_{get,add,remove}` routed failures through `create_model_error()`. That helper rewrites its message whenever `model_exists(requested_model)` is false (`server.cpp:2190`) — always true for a *new* alias name — so the real cause was discarded:

```
POST /internal/aliases {"alias":"foo","target":"foo"}
before: {"code":"model_not_found","message":"Model 'foo' was not found. Available models include: ..."}
after:  {"code":"invalid_alias","type":"invalid_request_error","param":"alias","message":"Alias cannot point to itself"}
```

Adds `create_alias_error()` and converts **10** call sites. HTTP status codes were already correct (400/404/409) and are unchanged, but this is **more than message quality**: the structured `code`, `type`, and `param` fields change too, so API consumers keying on `code` now get a meaningful value instead of `model_not_found` on every alias failure.

The three 500 paths (`handle_aliases_get` catch, `AliasManager uninitialized`, `handle_aliases_remove` catch) carry `type: server_error`; the seven 400/404/409 paths carry `invalid_request_error`.

New in v11.6.0 via #2818.

### 2. CLI discarded structured server errors

`extract_server_error_message()` only unwrapped `{"error": "<string>"}`. The server emits `error` as an object, so `is_string()` failed and the CLI fell back to `"Request failed: <code>"`. This affected every CLI command hitting a structured error — 31 `create_model_error` call sites feed it — not just aliases.

```
before: Request failed: 400
after:  Error creating alias: Alias cannot point to itself
```

Not a regression (the string-only check dates to `a2e6a8a45`, 2026-04-04), but fixing #1 without this leaves CLI users seeing nothing useful.

### 3. Muse Glimmer post promised a specific date

`docs/news/muse-glimmer.html:539` said the model ships built-in "in the regular **Wednesday** Lemonade release on **August 12**". Aug 12 is a Wednesday but Aug 13 is a Thursday, so a slipped tag makes the sentence self-contradictory rather than merely late. Now refers to v11.6.0 instead of a date, so it cannot go stale again. The substance was already true — the shipped `builtin` pin (b10375) covers Muse Glimmer.

Note `publish-website.yml` triggers on `push: tags: v*` and checks out the tag, so this fix only reaches the live site once it is on the tagged commit.

### Verification

Built clean and tested against a local server on this branch:
- self-referencing alias → `invalid_alias` / `invalid_request_error`
- delete nonexistent alias → `alias_not_found` / `invalid_request_error`
- `lemonade alias add foo foo` → `Error creating alias: Alias cannot point to itself`
- `lemonade alias remove nope` → `Error removing alias: Alias not found: nope`

### Follow-up

Automated coverage for `extract_server_error_message()` (string-shaped, object-shaped, malformed JSON, fallback) and the alias HTTP error envelopes is not included here — `lemonade_client.cpp` pulls in httplib/curl so it needs a new test target and `add_cpp_ci_test` wiring, which is not something to land on release night. Tracking as a follow-up alongside the other test gaps found during checklist verification.

Needs cherry-picking onto `release-v11.6.0` after merge — that branch was cut at `3e5325960`.

### DIFF
--- a/docs/news/muse-glimmer.html
+++ b/docs/news/muse-glimmer.html
@@ -536,7 +536,7 @@
 
             <p class="mg-note"><strong>Note:</strong> Set the same key to <code>builtin</code> to return to the llama.cpp version Lemonade ships and tests, for example <code>lemonade config set llamacpp.vulkan_bin=builtin</code>.</p>
 
-            <p class="mg-note"><strong>Note:</strong> Muse Glimmer ships as a built-in model in the regular Wednesday Lemonade release on August 12, so this step becomes optional then.</p>
+            <p class="mg-note"><strong>Note:</strong> Muse Glimmer ships as a built-in model in Lemonade v11.6.0, so this step becomes optional once you update.</p>
           </div>
         </div>
 
--- a/src/cpp/cli/lemonade_client.cpp
+++ b/src/cpp/cli/lemonade_client.cpp
@@ -140,8 +140,15 @@ std::string extract_server_error_message(const HttpError& error) {
     if (!error.response_body().empty()) {
         try {
             auto parsed = json::parse(error.response_body());
-            if (parsed.contains("error") && parsed["error"].is_string()) {
-                return parsed["error"].get<std::string>();
+            if (parsed.contains("error")) {
+                const auto& err = parsed["error"];
+                if (err.is_string()) {
+                    return err.get<std::string>();
+                }
+                // OpenAI-style bodies nest the text under error.message.
+                if (err.is_object() && err.contains("message") && err["message"].is_string()) {
+                    return err["message"].get<std::string>();
+                }
             }
         } catch (const json::exception&) {
         }
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -5789,10 +5789,11 @@ void Server::handle_cleanup_cache(const httplib::Request& req, httplib::Response
 // message whenever the name is absent from the registry, which is always true for a
 // new alias, so routing these through it would report "model not found" instead of
 // the actual reason.
-static nlohmann::json create_alias_error(const std::string& message, const std::string& code) {
+static nlohmann::json create_alias_error(const std::string& message, const std::string& code,
+                                         const std::string& type = "invalid_request_error") {
     return nlohmann::json{{"error", {
         {"message", message},
-        {"type", "invalid_request_error"},
+        {"type", type},
         {"param", "alias"},
         {"code", code}
     }}};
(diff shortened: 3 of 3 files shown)


fl0rianr (approved):
Looks good overall, and the fixes make sense to me. The root cause around `create_model_error()` is clear, and the CLI change is a nice minimal fix that preserves the existing string-shaped error handling while also supporting the structured `error.message` format.

A couple of small things I noticed, none of them blockers:

* `create_alias_error()` always sets `type` to `invalid_request_error`, but it is also used for the 500/internal-error paths. That means a genuine server-side failure can still be classified as a client request error in the structured response. It might be cleaner to pass the type into the helper as well, or use a server-error type for those cases.
* Since this fixes two fairly specific regressions, some automated coverage would be nice. In particular, `extract_server_error_message()` looks easy to cover with string-shaped errors, object-shaped `error.message`, malformed JSON, and the fallback path. The alias HTTP error envelopes would also be worth locking down if there is a suitable endpoint test location. Given the manual verification and green CI, I would not block this release fix on that.
* Small PR description nit: this is a little more than "message quality only", since the structured `code`, `type`, and `param` fields also change. Also, I count 10 alias `create_model_error()` replacements in the diff rather than 7.

The Muse Glimmer wording change also looks better than simply moving the date, since tying it to `v11.6.0` avoids the same issue if the release date moves again.

No blocking issues from my side.

ramkrishna2910:
Thanks — all three points were fair, and two of them I had wrong.

**1. `type` on the 500 paths — fixed in 885cc8922.** You're right that a genuine server failure shouldn't be classified as a client request error. Added a defaulted `type` parameter and passed `server_error` at exactly the three 500 sites (`handle_aliases_get` catch, `AliasManager uninitialized`, `handle_aliases_remove` catch); the seven 400/404/409 paths keep `invalid_request_error`. Verified each `server_error` site is preceded by `res.status = 500`, and confirmed live that the 400 and 404 paths still report `invalid_request_error`.

**3. Both nits correct, PR description updated.** It is **10** replacements — I miscounted at 7. And "message quality only" did undersell it: `code`, `type`, and `param` are new structured fields, so consumers keying on `code` previously got `model_not_found` for every alias failure regardless of cause. Reworded.

**2. Tests — agreed in principle, deferring per your call.** `extract_server_error_message` is header-declared and `HttpError` is trivially constructible, so the cases you list are straightforward. The cost is that `lemonade_client.cpp` pulls in httplib/curl, so it needs a new test target plus `add_cpp_ci_test` wiring — new CI surface I'd rather not add to a release-night fix. Tracking it as a follow-up together with the other coverage gaps found during checklist verification (trellis reserved-flag rejection, `bench.cpp` containment guard, and the routing load-vs-policy-change interleaving that `test_routing_helper_reconcile.cpp`'s own header flags as untestable).

Happy to pull the tests into this PR instead if you'd prefer them to land together.

fl0rianr (approved):
Very nice! Regarding your test question do it as you may.

PR #3112, opened 2026-08-13
lemonade import <file.json>: --dry-run policy validation, --alias; pull redirects file args to import

### LINKED ISSUE #3088: CLI: register router policy from a local JSON file (lemonade pull <policy.json>)
## Problem

Router policies can only be registered by hand-crafting a `curl -X POST /api/v1/pull` call with the policy JSON. The CLI's manual pull path advertises `--recipe collection.router` in `lemonade pull --help`, but there is no flag capable of expressing `routing` (candidates, default_model, rules, classifiers) — so a policy JSON on disk has no CLI path into the server.

## Proposal

Accept a local policy file:

```bash
lemonade pull ./user.SWE-Router.json
# or explicitly:
lemonade pull --file ./user.SWE-Router.json
```

The CLI reads the file and POSTs it to `/api/v1/pull` unchanged — the endpoint already accepts exactly this document and the parser already validates it (recipe check, non-empty rules, candidate existence, `version` field). Registration is idempotent today (re-pull overwrites), which gives file-based edit→re-register iteration for free.

Nice-to-have: `--dry-run` that runs server-side validation without registering, for CI-checking policy files.

## Why

- Router policies are documents, not flag sets — file-based registration is the natural UX and matches how users iterate on them (edit JSON, re-register, test).
- Prerequisite for `lemonade launch <agent> --config <policy.json>` (filed separately), which needs a programmatic register-from-file step.
- Comparable tools ship this: NVIDIA Switchyard takes its whole deployment as a config file per run.

ramkrishna2910:
Closes #3088.

Router policies (and any model definition) register from a local JSON file via `lemonade import`, which gains validation and aliasing; `pull` stays registry-only and redirects file arguments with a hint. (Originally implemented on `pull`; reworked onto `import` per review — one file-ingestion entry point, no `.json` heuristics on pull's positional.)

```bash
lemonade import ./my-router.json            # register (idempotent: re-import to update)
lemonade import ./my-router.json --dry-run  # validate only; nothing is registered
lemonade import ./my-router.json --alias my-router
lemonade pull ./my-router.json              # → hint: "Use: lemonade import ..."
```

- `--dry-run`: the local structural checks, plus the server-side policy parse via existing `POST /routing/validate` for `collection.router` documents (rule/candidate cross-checks, classifier refs, `version`). Only HTTP 400 reads as a validation verdict; other statuses report as request failures. Component *existence* is only checked by real registration — documented.
- `--alias`: binds to the normalized name returned by the import (so a bare-name file that gains the `user.` prefix aliases correctly). `--dry-run`/`--alias` without a file argument are rejected.
- `import`'s existing semantics are untouched: `upgrade` stays off, remote-recipe flow unchanged.
- Docs: `router-policy.md` registering section leads with the CLI form; `cli.md` import section documents the new flags, pull section points to import (edits outside generated regions).
- Tests (`server_cli2.py`): import-from-file + alias resolution, dry-run registers-nothing + bad-policy verdict + no-file-arg rejection, pull-redirect hint for existing and typo'd paths; neighboring pull/import tests pass as regression.

https://claude.ai/code/session_016yHZizzdKkiPz3WbPpUHu1

### DIFF
--- a/docs/dev/router-policy.md
+++ b/docs/dev/router-policy.md
@@ -157,8 +157,8 @@ and every entry's `model` must be one of `components`:
 Register the policy like any collection — from the CLI:
 
 ```bash
-lemonade pull ./my-router.json            # register (idempotent: re-pull to update)
-lemonade pull ./my-router.json --dry-run  # validate only; nothing is registered
+lemonade import ./my-router.json            # register (idempotent: re-import to update)
+lemonade import ./my-router.json --dry-run  # validate only; nothing is registered
 ```
 
 `--dry-run` runs the same structural checks as registration plus the
--- a/docs/dev/router-policy.md
+++ b/docs/dev/router-policy.md
@@ -154,7 +154,19 @@ and every entry's `model` must be one of `components`:
 
 ## Registering and invoking
 
-Register the policy like any collection — `POST /v1/pull` with the policy JSON:
+Register the policy like any collection — from the CLI:
+
+```bash
+lemonade pull ./my-router.json            # register (idempotent: re-pull to update)
+lemonade pull ./my-router.json --dry-run  # validate only; nothing is registered
+```
+
+`--dry-run` runs the same structural checks as registration plus the
+server-side policy parse (rule/candidate cross-checks, classifier references,
+`version`); component *existence* in the registry is only checked by a real
+registration.
+
+Or `POST /v1/pull` with the policy JSON directly:
 
 ```bash
 curl -X POST http://localhost:13305/api/v1/pull \
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -175,7 +175,7 @@ The `pull` command downloads and installs models. It can also register a custom 
 Common forms:
 
 ```bash
-lemonade pull MODEL_OR_CHECKPOINT_OR_FILE [--alias ALIAS] [--checkpoint TYPE CHECKPOINT] [--recipe RECIPE] [--label LABEL] [--components MODEL ...] [--dry-run]
+lemonade pull MODEL_OR_CHECKPOINT [--alias ALIAS] [--checkpoint TYPE CHECKPOINT] [--recipe RECIPE] [--label LABEL] [--components MODEL ...]
 ```
 
 ```bash
@@ -190,19 +190,17 @@ lemonade pull unsloth/Qwen3-8B-GGUF:Q4_K_M
 
 # Register and pull a custom model with an alias
 lemonade pull user.MyModel --checkpoint main org/model:Q4_0 --recipe llamacpp --alias my-alias
-
-# Register a model or router policy from a local JSON file (idempotent —
-# re-pull the file to update; --dry-run validates without registering)
-lemonade pull ./my-router.json
-lemonade pull ./my-router.json --dry-run
 ```
+
+To register a model or router policy from a local JSON file, use
+[`lemonade import`](#options-for-import) — a `.json` path passed to `pull`
+prints a hint pointing there.
 
 | Option | Description | Required |
 |--------|-------------|----------|
-| `MODEL_OR_CHECKPOINT_OR_FILE` | Registered model name, `owner/repo[:variant]` Hugging Face/ModelScope checkpoint, or path to a local `.json` model/policy file (e.g. a `collection.router` policy) | Yes |
+| `MODEL_OR_CHECKPOINT` | Registered model name, or `owner/repo[:variant]` Hugging Face/ModelScope checkpoint | Yes |
 | `--source` | Remote registry for checkpoint pulls: `huggingface` or `modelscope`; when omitted, the server's configured `default_model_source` applies. Direct hub URLs are auto-detected | No |
 | `--alias ALIAS` | Add an alias for the model being registered or pulled. | No |
-| `--dry-run` | With a `.json` file argument: validate without registering — structural checks plus the server-side policy parse for `collection.router` files. | No |
 | `--checkpoint TYPE CHECKPOINT` | Manual registration: add a checkpoint entry. Repeat for multi-component models such as `main` + `mmproj` or `main` + `vae`. | No |
 | `--recipe RECIPE` | Manual registration: recipe to associate with the new `user.*` model (`llamacpp`, `flm`, `ryzenai-llm`, `vllm`, `whispercpp`, `sd-cpp`, `kokoro`, `collection.omni`) | No |
 | `--label LABEL` | Manual registration: add a label to the new model. Repeatable. Valid: `coding`, `embeddings`, `hot`, `mtp`, `reasoning`, `reranking`, `tool-calling`, `vision` | No |
@@ -259,11 +257,23 @@ lemonade import [JSON_FILE] [options]
 
 | Option | Description | Required |
 |--------|-------------|----------|
-| `JSON_FILE` | Path to a JSON configuration file | No |
+| `JSON_FILE` | Path to a JSON model/policy file (e.g. a `collection.router` policy) | No |
+| `--dry-run` | Validate `JSON_FILE` without registering — structural checks, plus the server-side policy parse for `collection.router` files | No |
+| `--alias ALIAS` | Add an alias for the imported model | No |
 | `--directory DIR` | Remote recipes directory to query (e.g., `coding-agents`) | No |
 | `--recipe-file FILE` | Specific recipe JSON filename from the selected directory | No |
 | `--skip-prompt` | Run non-interactively (requires `--directory` and `--recipe-file` for remote import) | No |
 | `--yes` | Alias for `--skip-prompt` | No |
+
+Import is idempotent — re-import the file to update an existing registration.
+Router policies (`collection.router` files) register the same way; `--dry-run`
+additionally runs the server-side policy parse (rule/candidate cross-checks,
+classifier references, `version`) without registering anything:
+
+```bash
+lemonade import ./my-router.json            # register / update
+lemonade import ./my-router.json --dry-run  # validate only
+```
 
 **Remote import notes:**
 - Running `lemonade import` without `JSON_FILE` starts interactive recipe browsing from GitHub.
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -175,7 +175,7 @@ The `pull` command downloads and installs models. It can also register a custom 
 Common forms:
 
 ```bash
-lemonade pull MODEL_OR_CHECKPOINT [--alias ALIAS] [--checkpoint TYPE CHECKPOINT] [--recipe RECIPE] [--label LABEL] [--components MODEL ...]
+lemonade pull MODEL_OR_CHECKPOINT_OR_FILE [--alias ALIAS] [--checkpoint TYPE CHECKPOINT] [--recipe RECIPE] [--label LABEL] [--components MODEL ...] [--dry-run]
 ```
 
 ```bash
@@ -190,6 +190,11 @@ lemonade pull unsloth/Qwen3-8B-GGUF:Q4_K_M
 
 # Register and pull a custom model with an alias
 lemonade pull user.MyModel --checkpoint main org/model:Q4_0 --recipe llamacpp --alias my-alias
+
+# Register a model or router policy from a local JSON file (idempotent —
+# re-pull the file to update; --dry-run validates without registering)
+lemonade pull ./my-router.json
+lemonade pull ./my-router.json --dry-run
 ```
 
 | Option | Description | Required |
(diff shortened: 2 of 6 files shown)


fl0rianr (commented):
Thanks, this is a solid direction overall. Reusing the existing JSON import/normalization path keeps the implementation nicely contained, and the new tests cover the main happy/error paths.

A few things I think should be addressed before merge:

* `validate_model_json_file()` currently treats every `HttpError` from `/routing/validate` as `Invalid policy`, although only HTTP 400 is actually a validation result. A 404/500/etc. should be reported as a request/server failure instead. `HttpError::status_code()` is already available for this.
* `--alias` is silently ignored for JSON file pulls because the file branch returns before the existing alias handling. Either support it, ideally using the normalized name returned by `import_model_from_json_file()`, or reject the combination explicitly.
* `lemonade pull <file.json>` currently goes through `pull_model(..., upgrade=false)`, while the normal explicit `lemonade pull` path deliberately uses `upgrade=true`. For JSON model definitions with remote checkpoints this changes the expected pull/update semantics.
* `is_local_json_file()` only recognizes existing files, so a typo like `lemonade pull ./missing.json` falls back to the model/registry path instead of producing a clear file-not-found error. It would be better to separate “looks like a JSON file argument” from “file exists”.
* Small docs mismatch: the usage line now says `MODEL_OR_CHECKPOINT_OR_FILE` and includes `--dry-run`, but the options table still documents `MODEL_OR_CHECKPOINT` and has no `--dry-run` entry.

Apart from these, the structure and test coverage look good. I’d address the first three before merging; the last two are smaller cleanup items.

ramkrishna2910:
All five addressed in dee352cd:

1. **HttpError scoping** — only `status_code() == 400` reads as `Invalid policy`; any other status now reports `Error: routing/validate request failed (HTTP <n>): <server message>`.
2. **`--alias` on file pulls** — supported, applied to the normalized name returned by `import_model_from_json_file()` (so a bare-name file that gets the `user.` prefix aliases correctly). Test verifies the alias resolves to the collection (alias lookups echo the alias as `id`, so the assertion checks recipe + components of the resolved object).
3. **Upgrade semantics** — `import_model_from_json_file()` gained an `upgrade` flag (default `false`, so `lemonade import` behavior is unchanged); the explicit `pull <file>` path passes `true`, matching the other explicit-pull call sites.
4. **Typo'd path** — split into `looks_like_json_file_argument()` (suffix only) + an existence check in the pull handler: `lemonade pull ./missing.json` now errors `JSON file not found: './missing.json'` instead of falling through to the registry flow. Covered in the test.
5. **Docs table** — positional renamed to `MODEL_OR_CHECKPOINT_OR_FILE` with the file form described, and a `--dry-run` row added.

Test tally after the changes: the three new pull-file tests plus `test_054`/`test_060` as regression — 5/5 pass against a live server.

fl0rianr (commented):
Thanks for addressing the previous review points. I re-checked the latest head and those changes look good, and CI is fully green now.

I still see one correctness issue before merge:

* looks_like_json_file_argument() currently treats any argument ending in .json as a local file before normal model/registry resolution. That means a valid registered model such as user.foo.json would now fail with JSON file not found instead of being pulled as a model. I think file detection should only claim the argument when the file actually exists, or when the argument is clearly path-like (./foo.json, ../foo.json, absolute path, etc.). An explicit --file would also avoid the ambiguity entirely.

One smaller UX point: --source is accepted together with a JSON file but is silently ignored by the file branch. It would be better to either reject that combination explicitly or define how it should behave.

Apart from that, the previous issues look resolved and the implementation looks good to me.

bitgamma:
one question, is there any particular reason why this goes through the `lemonade pull` command and not `lemonade import` which already handles importing JSON model recipes?

fl0rianr:
I agree, this seems to be the better option. I missed this in my earlier review.

ramkrishna2910:
@bitgamma @fl0rianr agreed — reworked in 66f4ea6d. The feature now lives on `lemonade import` (which was already the file-ingestion path): it gains `--dry-run` (server-side policy parse via `POST /routing/validate` for `collection.router` files, nothing registered) and `--alias` (bound to the normalized imported name). `pull` is reverted to registry-only; a `.json` argument prints a redirect hint (`Use: lemonade import <file> [--dry-run] [--alias ALIAS]`) instead of falling through to a registry lookup — that keeps Ollama-style pull muscle memory from dead-ending. The `upgrade` flag from the earlier round is gone entirely (it was pull-specific; `import`'s existing semantics are untouched). PR title/body updated; tests reworked to the import surface + a redirect test, all green with Black applied.

fl0rianr (commented):
One issue from my previous review still seems to remain: looks_like_json_file_argument() is still suffix-only and is checked before normal model/registry resolution. So a valid registered model such as user.foo.json would still be intercepted; it now gets the import hint instead of the previous file-not-found error, but the ambiguity itself is unchanged.

Since file ingestion now lives entirely under import, I think pull should either only show this hint for clearly path-like/existing file arguments, or drop the .json heuristic altogether. A regression test for a registered model ending in .json would also be useful.

jeremyfowers:
**Ready for review**

No to-dos.

The PR body accurately covers all three changes: --dry-run with local + server-side validation, --alias binding to the normalized import name, and the pull redirect with a hint. The code examples match the diff behavior.

Checked: PR body lists --dry-run, --alias, import semantics, docs, and tests; diff confirms all items — main.cpp adds the flags and redirect logic, recipe_import.cpp adds validate_model_json_file, docs/guide/cli.md and docs/dev/router-policy.md updated, test/server_cli2.py adds tests 057–059.

**Focus: focused**

All changes serve one goal: providing a CLI entry point for registering router policies and models from local JSON files. The --dry-run and --alias flags extend import, the pull redirect prevents accidental file arguments, and the doc/test updates all support this single feature.

Checked: The three changes are tightly coupled: --dry-run and --alias extend import for JSON file ingestion, the pull redirect closes the UX gap for accidental .json args on pull, and all docs/tests cover the same feature. No separable work identified.

**Alignment: none found**

Checked: Reviewed contribute.md AI Contribution Policy and Reviewer Expectations; no misaligned changes observed — the diff is focused on import/pull CLI flags and their tests.

**Documentation: adequate**

Checked: Reviewed documentation.md; the PR updates docs/guide/cli.md (import flags) and docs/dev/router-policy.md (registering section) in the same PR as the code, matching the same-PR requirement.

**Testing: adequate**

Checked: Reviewed testing.md; the PR adds three test methods (test_057, test_058, test_059) to test/server_cli2.py, the correct suite for CLI command changes per the Where Tests Go table.

Checked: No user-visible contract changes: new optional flags on import and a redirected error on pull are additive, not breaking; no existing usage pattern is altered.

</details>

**Review suggestion**

Suggested reviewers:

- bitgamma (cli) — The maintainer table lists cli as a subject area; this PR modifies src/cpp/cli/main.cpp and src/cpp/cli/recipe_import.cpp to add --dry-run and --alias flags to the import subcommand.
- sawansri (cli) — The maintainer table lists cli as a subject area; this PR modifies src/cpp/cli/main.cpp and src/cpp/cli/recipe_import.cpp to add --dry-run and --alias flags to the import subcommand.

**Why this rung**

Flag on an existing CLI command. Two flags added to an existing `import` command; does not introduce a new command, route, or subsystem. The maintainer table lists cli (bitgamma) and documentation (fl0rianr) as subject areas relevant to the surfaces changed.

**Reviewer search**

Checked: Searched the maintainers table in docs/dev/contribute.md for the term 'cli' (the changed surface); two rows matched: @bitgamma and @sawansri both list cli in their Subject Areas cell. The PR modifies src/cpp/cli/main.cpp and src/cpp/cli/recipe_import.cpp, landing squarely in that area. The code-authorship pass (get-pr-code-authors.sh) was also run: @sawansri has the most blamed lines (27) on the affected files, followed by @bitgamma (18), @jeremyfowers (16), and @ramkrishna2910 (8, the PR author — excluded per policy). Both @sawansri and @bitgamma appear in the maintainer table under cli and are confirmed as the top two code authors by hunk blame on the changed files.

</details>

_[AI-assisted review] Automated pre-review from repo-manager — flags for the human reviewer, not a replacement for one. Reviewed at head `66f4ea6`; regenerate with `repo-manager review-pr 3112`._

jeremyfowers:
Requesting @bitgamma since it hits `lemonade import`

bitgamma (commented):
one minor nit from my side: considering a method called `validate_and_transform_model_json` already exists, maybe `validate_model_json_file` should get a name that better describes its goal? `dry_run_model_json` or something 

Below, a report from DS4-Flash-0731

I've reviewed the PR by fetching the diff, reading the surrounding source, applying the patch to a clean tree, and compiling. Here are my findings.                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
### Summary                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                              
 The PR adds --dry-run and --alias to lemonade import <file.json>, and redirects .json file args passed to pull toward import. It compiles cleanly against the current base and the changes are well-scoped and documented. The /api/v1/routing/validate endpoint the dry-run depends on exists and is registered under all   
 four prefixes; the CLI uses the correct /api/v1/ path. Docs edits land outside the generated regions in cli.md (generated block is lines 375–462), so no CI drift.                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                              
 Findings (prioritized)                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                              
 ### 1. --dry-run is not side-effect free — it can trigger server-side model loads (design).                                                                                                                                                                                                                                      
 For collection.router files, validate_model_json_file POSTs the policy to /api/v1/routing/validate. That handler builds a RoutingPolicyEngine and calls engine.route(ctx, …), and its ClassifierServices wrap the live Router with auto_load_model_if_needed(..., LoadPurpose::RoutingDependency) (server.cpp:3685–3688,     
 routing_classifier_services_router.cpp). So a dry-run of a policy with classifier rules can load models on the server. A "validate only / nothing is registered" command that may pull models into memory is surprising and should be called out in the docs or avoided (e.g., validate without evaluating classifier        
 branches).                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                              
### 2. The bare-name alias case the PR explicitly claims is untested (coverage gap).                                                                                                                                                                                                                                             
 The PR body stresses "a bare-name file that gains the user. prefix aliases correctly," but test_057 writes a user.-prefixed name, so the user.-prefixing path is never exercised. The logic is correct (import_model_from_json_file returns the post-normalization model_name, and handle_import_command binds the alias to  
 it), but the specific scenario advertised isn't verified by a test.                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
### 3. looks_like_json_file_argument is suffix-only — over-broad redirect (minor risk).                                                                                                                                                                                                                                          
 pull rejects any argument ending in .json, regardless of existence (explicitly documented in the header comment). Since the pull positional also accepts URLs and owner/repo[:variant] checkpoints, a legit value ending in .json (e.g., a hub URL) is silently redirected with a hint instead of falling through. Low       
 likelihood, but the check runs before has_manual_pull_options, so combining .json with manual-pull flags also yields a confusing hint.                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                              
### 4. Misleading / dropped flags in edge combinations (UX).                                                                                                                                                                                                                                                                     
 - lemonade import --directory X --recipe-file Y --dry-run errors with "require a JSON file argument" even though a remote recipe was given — dry-run can't validate remote recipes, but the message doesn't say that.                                                                                                        
 - lemonade import file.json --dry-run --alias foo silently ignores --alias (the dry-run branch returns first) with no warning.                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
### Minor 
                                                                                                                                                                                                                                                                                                          
 - The success-path response_json.contains("error") branch in validate_model_json_file is effectively dead code — make_request throws HttpError on any non-200, and the server returns validation errors only as 400, so a 200-with-error never occurs. Harmless but misleading.                                              
 - The --dry-run/--alias rejection is checked only in handle_import_command; the --alias value is bound to the shared config.alias_name field used by pull/alias subcommands, which is fine since they're separate CLI::Apps — no runtime conflict.                                                                           
 - pull previously accepted any model string; rejecting .json-suffixed names is an intentional behavior change per #3088, but it could break scripts that pull a model whose name happens to end in .json (essentially never in practice).                                                                                    
                                                                                                                                                                                                                                                                                                                              
### Verified                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                              
 - Patch applies cleanly and the CLI builds (ninja lemonade) with no warnings from the changed files.                                                                                                                                                                                                                         
 - Manual checks: pull ./my-router.json → correct redirect hint; import --dry-run (no file) → correct rejection; local dry-run of a non-router model prints Dry run OK (local checks) and exits 0 without registering.                                                                                                        
 - import_model_from_json_file signature change (added imported_model_out) is consistently updated in header and all call sites (main.cpp, recipe_import.cpp:533).

PR #3117, opened 2026-08-13
refactor(gui3): persist custom models in lemond

fl0rianr:
Moves GUI3 custom model persistence from browser storage to lemond.

Custom models are now saved through /v1/models/register and read back from the server model registry. Import, delete, and selection follow the same server-owned model state.

Uses the new server registration endpoint from #3118

### DIFF
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -381,11 +381,11 @@ const App: React.FC = () => {
   const serverModelState = useServerModelState();
   const status = serverModelState.status;
   const serverModels = serverModelState.models?.data ?? EMPTY_MODELS;
+  const customModelInfos = useMemo(() => serverModels.filter(model => (model as any).custom === true), [serverModels]);
   const rawLoadedModels = serverModelState.health?.all_models_loaded ?? EMPTY_LOADED_MODELS;
   const [currentModel, setCurrentModel] = useState<string | null>(null);
   const [theme, setTheme] = useState<Theme>(loadTheme);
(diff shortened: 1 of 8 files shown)


jeremyfowers:
Why do we need a new endpoint? Why can't custom model registration in GUI3 be identical to GUI2?

fl0rianr:
It's due to the current implementation that you can register models but not download them yet. This is useful if you have a unstable connection and might want to download later. Current GUI2 registers and pulls in one go. But this are two operations. If you have a model with no weights etc. - so current main allows now to register and download separately.
This change is needed that the UI does not register a model in it's lokal browser storage but relies on the server. The server side change will be removed in the next PR update.

jeremyfowers:
I do not want to consider this new endpoint for adoption at this time. It is crucial that we stop adding features and focus on quality.

jeremyfowers:
Looks like you removed the cpp changes, thank you. Can you update the PR description as well?

fl0rianr:
> Looks like you removed the cpp changes, thank you. Can you update the PR description as well?

Did that already. It's a good foundation PR and I'll have a clean-up follow-up for local storage after that.

jeremyfowers:
Sorry I misunderstood! All good. Will review soon.

fl0rianr:
Made a deterministic diff reduction of this PR (removed unnecessary legacy model data local storage migration)  with the same end result after #3183 and tried the new GH stacked PR feature to improve review-ability code-wise. I suggest to choose the complete result of #3183 for local testing.

jeremyfowers (approved):
LGTM

PR #3118, opened 2026-08-13
feat(server): add model registration endpoint

fl0rianr:
Adds POST /v1/models/register for persisting user model definitions without downloading model files.

The endpoint reuses the existing server-side model registry and validation flow, updates the model cache, and returns the registered model name. /pull remains responsible for installation and downloads.

Includes endpoint tests and API documentation.

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -228,8 +228,8 @@ When `include_paths=true` is supplied, each file entry also includes `path`:
 
 Register or update a `user.*` model definition without downloading model files.
 Use this endpoint when registration and installation are separate actions.
-`POST /v1/pull` remains the install/download path and keeps its existing
-register-then-install behavior through the same internal registration path.
+`POST /v1/pull` remains the install/download path and performs the same internal
+registration step before downloading.
 
 The endpoint is available at:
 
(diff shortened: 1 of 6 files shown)


bitgamma (changes requested):
is all that duplication necessary? The risk of the `register` branch and `pull` diverging over time is concrete. I'd rather either have the `/pull` endpoint accept a `register_only` param or have both endpoints call the same helper method for everything they have in common.

fl0rianr:
I would rather have it separate since current server lags this and having register and pull in one operation is wrong. I agree the change could be more minimal but we would then make a fake pull. Do you see a way we can do it properly? See it as the start of the proper separation.

bitgamma:
> I would rather have it separate since current server lags this and having register and pull in one operation is wrong. I agree the change could be more minimal but we would then make a fake pull. Do you see a way we can do it properly? See it as the start of the proper separation.

@fl0rianr well, `handle_pull` already does this if the server is offline. What speaks against extracting everything that comes before that check into a helper function and call that helper both in `handle_model_register` and `handle_pull`? pull is register + download in my mind

fl0rianr:
Thanks @bitgamma! Got it. Refactored this according to your suggestion. Both /models/register and /pull now always go through the same internal registration path. /models/register returns after registration, while /pull continues with the download.

The duplicated validation/normalization from handle_pull is gone as well.

bitgamma (approved):

PR #3120, opened 2026-08-13
GUI3 fix: chat curser position

fl0rianr:
This is the fix, seems to work correct for me, but kinda hard to show since it's dynamic..


jeremyfowers (approved):
LGTM!

PR #3123, opened 2026-08-13
fix(packaging): add AppStream MetaInfo files for desktop entries

superm1:
## Summary
Resolves lintian `no-metainfo` warning by adding proper AppStream MetaInfo files for both desktop applications.

## Changes
- Added `data/com.lemonade.WebApp.metainfo.xml` for the web app
- Added `data/com.lemonade.App.metainfo.xml` for the Tauri desktop app
- Updated `CMakeLists.txt` to install MetaInfo files to `share/metainfo`
- MetaInfo for Tauri app only installs when `BUILD_TAURI_APP` is enabled

## Why
The lintian warning indicates that software components with desktop files should provide MetaInfo files instead of relying on deprecated metadata generation from desktop entries. MetaInfo files provide:
- Better descriptions for users in software centers
- Proper metadata structure (categories, keywords, URLs)
- AppStream-compliant packaging
- Version and release information

### DIFF
--- a/data/com.lemonade.App.metainfo.xml
+++ b/data/com.lemonade.App.metainfo.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.lemonade.App</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>Apache-2.0</project_license>
+  <name>Lemonade</name>
+  <summary>Local AI server with GPU and NPU acceleration</summary>
+  <description>
+    <p>
+      Lemonade is a local AI server that provides the same capabilities as cloud APIs,
+      but 100% free and private. Run the latest models for chat, coding, speech, and
+      image generation on your own NPU and GPU.
+    </p>
+    <p>
+      Features include:
+    </p>
+    <ul>
+      <li>Multi-modal AI support: chat, speech transcription, text-to-speech, and image generation</li>
+      <li>GPU and NPU acceleration for optimal performance</li>
+      <li>OpenAI, Anthropic, and Ollama compatible APIs</li>
+      <li>Model management with automatic download and optimization</li>
+      <li>Privacy-focused: all processing happens locally</li>
+    </ul>
+  </description>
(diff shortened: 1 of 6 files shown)


data/com.lemonade.App.metainfo.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.lemonade.App</id>
```
fl0rianr: If I'm not mistaken, the new AppStream component IDs should use a reverse-DNS namespace controlled by this project. com.lemonade.* corresponds to lemonade.com, while the project domain is lemonade-server.ai (and the developer ID below already reflects that). Since AppStream component IDs are intended to be globally unique and stable once published, can we switch these to something like ai.lemonade_server.app / ai.lemonade_server.webapp before shipping them? The existing desktop filenames can stay unchanged because <launchable type="desktop-id"> provides the association.

data/com.lemonade.WebApp.metainfo.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.lemonade.WebApp</id>
```
fl0rianr: same here

fl0rianr (changes requested):

fl0rianr (commented):
Looks good overall. One thing I think we should fix before merging: the new AppStream component IDs use `com.lemonade.*`, which points to the `lemonade.com` namespace rather than one controlled by this project.

Since AppStream IDs are meant to be globally unique and stable once published, could we use the project domain here instead, for example `ai.lemonade_server.app` and `ai.lemonade_server.webapp`?

The existing `.desktop` filenames can stay as they are since the `<launchable type="desktop-id">` entries already provide the association.

superm1:
Isn't that what was just changed?

fl0rianr (approved):
Yeah, you're right. I missed the latest commit when I re-checked the PR, a refresh problem, not sure. The namespace change addresses my concern. LGTM.

PR #3126, opened 2026-08-13
GUI3 fix(chat): handle active request cancellation correctly

fl0rianr:
Fixes two related cancellation issues in the chat UI:

1. Deleting a conversation now stops any active request before removing the chat. Before that an (also UX breaking extreme long computing) request could not be stopped anymore by the GUI.
2. 
3. User-initiated request aborts, such as pressing Stop during prefill, are no longer shown as connection errors.

This keeps cancellation behavior consistent and prevents requests from continuing after their conversation has been deleted.


jeremyfowers (approved):
tested locally

PR #3128, opened 2026-08-13
fix(server): preserve auxiliary checkpoints after interrupted updates

### LINKED ISSUE #3075: Interrupted update hides a working model: auxiliary checkpoint resolver ignores snapshots with a live .download_manifest.json
### Platform

Windows

### Lemonade Version

11.5.2

### GPU / APU Model

### Component

lemond (server)

### Bug Description

When a model update is interrupted, `refs/main` is left pointing at a snapshot that holds a leftover `.download_manifest.json` and a `.partial`. From that point the model is reported as `"downloaded": false` and disappears from `/api/v1/models` and the app list — even though a complete copy of every artifact it needs is still present in the previous snapshot.

The GGUF resolver already guards against this. `resolve_gguf_path()` skips any snapshot that has a live manifest (`is_uncommitted_snapshot_file`, `llamacpp_gguf.cpp:43-55`, applied at line 78) and falls back to every completed snapshot (`llamacpp_gguf.cpp:240-245`, added in #2300 for exactly this "variant stranded in another snapshot" case).

The shared resolver used for **auxiliary** checkpoints has no such filter. `BackendOps::resolve_checkpoint_path()` takes the active snapshot and returns the first exact-name match:

```cpp
// backend_ops.cpp:34-39
fs::path active_snapshot = hf_cache::active_snapshot_path(model_cache_path_fs);
fs::path direct_variant_path = active_snapshot / path_from_utf8(ctx.variant);
if (hf_cache::exists(direct_variant_path)) {
    return path_to_utf8(direct_variant_path);   // returns immediately
}
```

`llamacpp` routes `main` to the GGUF resolver and everything else to that shared one (`llamacpp_server.cpp:847-855`), so `mmproj` / `draft` are resolved without any awareness that the snapshot is mid-download. `is_checkpoint_path_complete()` then rejects the result because a manifest sits next to it (`model_manager.cpp:1981`), `are_required_checkpoints_complete()` returns false, and the model is marked as not downloaded.

Net effect: a single auxiliary file that is itself **fully downloaded** takes down a model whose artifacts are all available.

### Steps to Reproduce

1. Register a vision GGUF model with an `mmproj` checkpoint, e.g.

   ```json
   {
     "checkpoints": {
       "main": "unsloth/gemma-4-E4B-it-GGUF:UD-Q8_K_XL",
       "mmproj": "unsloth/gemma-4-E4B-it-GGUF:mmproj-F16.gguf",
       "draft": "unsloth/gemma-4-E4B-it-GGUF:mtp-gemma-4-E4B-it.gguf"
     },
     "recipe": "llamacpp"
   }
   ```

2. Pull it so it works, then trigger an update to a newer commit.
3. Let the small files (`config.json`, `mmproj-F16.gguf`) finish, then interrupt during the large main GGUF (cancel, kill the server, or drop the network).
4. Restart the server.
5. The model is gone from `lemonade list` / the app list, and `GET /api/v1/models` no longer returns it.

Expected: the previously working model stays visible and loadable. Every artifact is still present in the previous completed snapshot, and the GGUF resolver already knows how to fall back to it.

Actual: `"downloaded": false`, the model disappears from the list, and the only way back is to re-download the whole thing — for large models that means re-pulling tens of GB for files that are already on disk.

### Log Output

```shell

```

### Additional Context

Observed on two models from different repositories, so it is not specific to one download.

`models--unsloth--gemma-4-E4B-it-GGUF` (still in the broken state while writing this):

```
refs/main = bfc15c382204943c3a8fff0c750b94ae2364d7a3

snapshots/bfc15c38...             <- active
  .download_manifest.json                       leftover (removed only on success)
  config.json
  gemma-4-E4B-it-UD-Q8_K_XL.gguf.partial        2.4 GB of 8.7 GB
  mmproj-F16.gguf                               complete
  (mtp-gemma-4-E4B-it.gguf absent)

snapshots/e1d90e5f...             <- previous, complete
  gemma-4-E4B-it-UD-Q8_K_XL.gguf
  mmproj-F16.gguf
  mtp-gemma-4-E4B-it.gguf
  ...
```

Resolution in this state:

- `main` -> found in `e1d90e5f` via the whole-cache fallback. OK.
- `draft` -> absent from the active snapshot, so the subtree search finds it in `e1d90e5f`. OK.
- `mmproj` -> found by exact name in `bfc15c38`, next to the manifest -> reported incomplete. This one file is the whole failure.

Consistent with that, deleting **only** `.download_manifest.json` (keeping the `.partial`, so resume data is preserved) is enough to bring the model back with no download at all — the manifest, not any missing artifact, is what hides it.

The built-in `Gemma-4-E4B-it-GGUF` (`Q4_K_M`) from the same repository is hidden too, for the same reason: it also declares `mmproj-F16.gguf`.

A second model, `models--unsloth--gemma-4-26B-A4B-it-GGUF`, hit the same thing earlier — its active snapshot was missing the 16.9 GB main GGUF while the small files were present, it vanished after a server restart, and only a full re-pull brought it back.

Possible directions, in rough order of cost:

1. Give `BackendOps::resolve_checkpoint_path()` the same "skip snapshots with a live manifest" filter the GGUF resolver has, so auxiliary checkpoints fall back to a completed snapshot.
2. Alternatively, do not advance `refs/main` to a snapshot whose download has not completed, so an aborted update cannot strand the previous working state.

Related but distinct from #2542: that one is about the update flag being raised spuriously, this one is about what happens when the resulting update is interrupted.

Solaris-star:
## Summary

- Skip auxiliary checkpoint matches from snapshots with a live `.download_manifest.json`.
- Fall back to a completed snapshot so interrupted updates do not hide an otherwise usable model.
- Add a C++ regression test covering fallback, active-snapshot preference, and the interrupted-only case.

Fixes #3075

## Testing

- `ctest --test-dir build-ops -R 'BackendOpsTest|HfVariantsTest|ModelRegistryTest' --output-on-failure`
- Result: 3/3 tests passed (including 4 new BackendOps assertions, 16 HF variant assertions, and the model registry suite).
- `git diff --check`
- Tracked-content secret scan against `origin/main`: no new findings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2872,6 +2872,15 @@ if(BUILD_TESTING AND EXISTS "${_HF_VARIANTS_TEST_SRC}")
     add_cpp_ci_test(HfVariantsTest CI ON COMMAND test_hf_variants)
 endif()
 
+# Auxiliary checkpoint resolution must ignore snapshots that still carry the
+# downloader's live manifest, just like the GGUF resolver does.
+set(_BACKEND_OPS_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_backend_ops.cpp")
+if(BUILD_TESTING AND EXISTS "${_BACKEND_OPS_TEST_SRC}")
+    add_executable(test_backend_ops test/cpp/test_backend_ops.cpp)
+    target_link_libraries(test_backend_ops PRIVATE lemonade-server-core)
+    add_cpp_ci_test(BackendOpsTest CI ON COMMAND test_backend_ops)
+endif()
+
 # vLLM CDNA gating: the vllm descriptor support row + asset-family resolution for
 # AMD Instinct (gfx942), alongside the existing RDNA families.
 set(_VLLM_CDNA_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_vllm_cdna_gating.cpp")
--- a/src/cpp/server/backends/backend_ops.cpp
+++ b/src/cpp/server/backends/backend_ops.cpp
@@ -29,12 +29,29 @@ std::string BackendOps::resolve_checkpoint_path(const ModelInfo& info,
     fs::path model_cache_path_fs = path_from_utf8(ctx.model_cache_path);
 
     if (!ctx.variant.empty()) {
+        auto is_uncommitted_snapshot_file = [](const fs::path& candidate,
+                                               const fs::path& snapshots_path) {
+            const fs::path relative = candidate.lexically_relative(snapshots_path);
+            if (relative.empty()) {
+                return false;
+            }
+
+            auto first = relative.begin();
+            if (first == relative.end() || *first == "." || *first == "..") {
+                return false;
+            }
+
+            return hf_cache::exists(snapshots_path / *first / ".download_manifest.json");
+        };
+
         // Prefer refs/main for auxiliary checkpoints too (e.g. mmproj) so
         // companion files stay on the active snapshot as the main model.
+        const fs::path snapshots_path = model_cache_path_fs / "snapshots";
         fs::path active_snapshot = hf_cache::active_snapshot_path(model_cache_path_fs);
         if (!active_snapshot.empty()) {
             fs::path direct_variant_path = active_snapshot / path_from_utf8(ctx.variant);
-            if (hf_cache::exists(direct_variant_path)) {
+            if (hf_cache::exists(direct_variant_path) &&
+                !is_uncommitted_snapshot_file(direct_variant_path, snapshots_path)) {
                 return path_to_utf8(direct_variant_path);
             }
             std::error_code ec;
@@ -42,12 +59,14 @@ std::string BackendOps::resolve_checkpoint_path(const ModelInfo& info,
                  fs::recursive_directory_iterator(active_snapshot, hf_cache::dir_options(), ec)) {
                 if (ec) break;
                 if (entry.is_regular_file(ec)) {
-                    if (entry.path().filename().string() == ctx.variant) {
+                    if (entry.path().filename().string() == ctx.variant &&
+                        !is_uncommitted_snapshot_file(entry.path(), snapshots_path)) {
                         return path_to_utf8(entry.path());
                     }
                 } else if (entry.is_directory(ec)) {
                     fs::path variant_path = entry.path() / path_from_utf8(ctx.variant);
-                    if (hf_cache::exists(variant_path)) {
+                    if (hf_cache::exists(variant_path) &&
+                        !is_uncommitted_snapshot_file(variant_path, snapshots_path)) {
                         return path_to_utf8(variant_path);
                     }
                 }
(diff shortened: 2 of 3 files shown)


Solaris-star:
Live-server verification follow-up for head `9e339c9b524b313ee0641411fcc259111503b46d`:

- Installed `test/requirements.txt` into an isolated `/tmp/lemonade-3128-live-venv`.
- Ran `/tmp/lemonade-3128-live-venv/bin/python test/server_cli2.py` — **failed before tests ran**: the harness timed out after 30 seconds because no server was running on port `13305`; `Ran 0 tests`.
- Ran `/tmp/lemonade-3128-live-venv/bin/python test/server_endpoints.py` — **failed before tests ran** for the same reason; `Ran 0 tests`.

The clean PR checkout has neither `build/lemonade` nor `build/lemond`, so the harness could not start a live server. I am recording this as **unable to verify live-server integration in this environment**, rather than claiming a pass. The focused C++/CTest result in the PR description is unchanged.

jeremyfowers:
**Ready for review**

No to-dos.

All three claimed changes — the manifest-skip filter, the fallback behavior, and the C++ regression test — are present in the diff, and the Fixes #3075 reference is valid: the issue describes exactly this gap in BackendOps::resolve_checkpoint_path().

Checked: PR body lists three bullet points; the diff in backend_ops.cpp adds the is_uncommitted_snapshot_file lambda and applies it in five resolution paths, CMakeLists.txt adds the test registration, and test/cpp/test_backend_ops.cpp contains three test functions covering the three cases described.

**Focus: focused**

All three changed files serve one goal: adding the is_uncommitted_snapshot_file guard to every checkpoint-resolution path in resolve_checkpoint_path(), wiring the new test into CMakeLists.txt, and verifying the behavior with three test cases. Removing any part would leave the fix incomplete.

Checked: All three changed files — CMakeLists.txt (+9), backend_ops.cpp (+31/-7), test/cpp/test_backend_ops.cpp (+120) — touch the same function and same bug; splitting any out would leave the fix untestable or unregistered.

**Alignment: none found**

Checked: Reviewed contribute.md and philosophy.md; the change is a contained fix solving one problem with no scope creep.

**Documentation: not-applicable**

Checked: none observed — a bug fix restoring intended behavior with no new user-facing surface.

**Testing: adequate**

Checked: New suite test_backend_ops.cpp registered via add_cpp_ci_test(BackendOpsTest CI ON COMMAND test_backend_ops) in CMakeLists.txt; three test cases cover interrupted fallback, completed-preference, and interrupted-only resolution.

Checked: none observed — the change restores intended behavior by skipping interrupted snapshots rather than removing a deliberate contract.

</details>

**Review suggestion**

Suggested reviewers:

- jeremyfowers (backend_ops.cpp (#2287)) — Wrote 51 blamed lines in src/cpp/server/backends/backend_ops.cpp via the self-describing backends refactor (#2287) and last touched it 2026-06-29.
- ramkrishna2910 (smart router and orchestration) — This PR lands on backend checkpoint resolution logic, which falls under the smart router and orchestration subject area listed in the maintainer table.

**Why this rung**

A contained fix or extension. The diff confines to one function (BackendOps::resolve_checkpoint_path) in one module (backend_ops.cpp) plus its test. This is a contained fix on the backend surface, which maps to one-reviewer.

**Reviewer search**

Checked: Maintainer table lists @jeremyfowers, @ramkrishna2910, @kenvandine, @bitgamma, and @sawansri under new backends; code blame points to @jeremyfowers as primary author of backend_ops.cpp; @ramkrishna2910 listed under smart router and orchestration which covers checkpoint resolution.

</details>

_[AI-assisted review] Automated pre-review from repo-manager — flags for the human reviewer, not a replacement for one. Reviewed at head `9e339c9`; regenerate with `repo-manager review-pr 3128`._

jeremyfowers:
@Geramy would you be interested in reviewing this networking-related PR?

PR #3130, opened 2026-08-14
test(gui): detect em dash regression, fix UI test

### LINKED ISSUE #3066: em dashes in description/help texts across the UI. this is a bad look: em dashes are a hallmark of ai output
Image

most humans use simple hyphens (-), not em dashes. em dashes are a clear sign of AI writing, and depending on how you want your product to be perceived, you may want to do a pass to remove these em dashes and replace them with something more human, like hyphens, commas, or whatever else you may think of. I know em dashes aren't *strictly* ai generated, and before AI were often used in literature, but that's just it... it was mostly used in literature and science papers, not in everyday things, and humans barely if ever use them.

fl0rianr:
Adds a regression check for em dashes used as punctuation in rendered UI copy. Check it catches correctly. 

While keeping standalone `—` placeholders valid (e.g. used in the monitor tab). Internal non-UI strings are ignored.

Also fixed:
wires the existing UI regression suite into the default npm test flow without preventing the remaining Playwright suites from running after a failure.

Did not found any other critical em dashes but without a regression test we would need to maintain this very carefully.

Closes #3066

### DIFF
--- a/src/app/scripts/run-playwright-tests.mjs
+++ b/src/app/scripts/run-playwright-tests.mjs
@@ -10,6 +10,7 @@ const hasProject = args.some(arg => arg.startsWith('--project') || arg === '-p')
 const hasHelp = args.some(arg => arg === '--help' || arg === '-h' || arg === 'help');
 
 const npxCommand = process.platform === 'win32' ? 'npx.cmd' : 'npx';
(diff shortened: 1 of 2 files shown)


jeremyfowers:
@claude review

PR #3132, opened 2026-08-14
fix(server): normalize thinking controls for llama.cpp

fl0rianr:
Tested and disabling thinking does not work.

Fixes server-side handling of enable_thinking for llama.cpp by mapping it to the native reasoning controls instead of relying on /no_think.

The legacy fallback is kept in place, so existing backends and clients continue to work unchanged.


bitgamma (approved):
makes sense. The /no_think thing was specific only to Qwen3.0 models and maybe a few more but has been long abandoned.

PR #3133, opened 2026-08-14
fix: stop request during prefill now possible

fl0rianr:
Propagate streaming client disconnects to the backend while waiting for the first response bytes.

Previously, disconnects were only detected when the backend produced another chunk. During a long prefill, this could leave the backend request running even after the client had stopped the request, blocking hardware etc.

post_stream() now polls the downstream connection through libcurl's transfer callback and aborts the upstream request when the client disconnects.

This does not change model unload behavior.


bitgamma (approved):
lgtm

PR #3134, opened 2026-08-14
feat(gui): add thinking menu

fl0rianr:
One remaining GUI2 regression - no option to disable thinking (which is rightfully an auto-on).
Cramping this into setting would be the wrong move I rather opted for this approach:
image
image
image

This is the way like it's done in recent very well known chat client like chatGPT.

Using it that way actually enables to switch back and forth in chat between fast and more thought through answers - very helpful.

That it acutally works depends on #3132 

Also good ground laying work for future additions like Lemonade specific thinking modes or what we defently can't ignore more and more OSS model appearing with dedicated reasoning effort settings. 

Additionally move the mic of the audio mode also to the right where the user expects this action, also comparable with modern chat client websites.
image

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -121,6 +121,8 @@ function getDownloadStoreModule(): Promise<typeof import('../features/downloadMa
 
 const CHAT_LOGS_WIDTH_KEY = 'chat_logs_panel_width';
 const CHAT_THINKING_MODE_KEY = 'chat_thinking_mode';
+const CHAT_COMPOSER_INPUT_MIN_HEIGHT = 40;
+const CHAT_COMPOSER_INPUT_MAX_HEIGHT = 200;
 const CHAT_LOGS_DEFAULT_WIDTH = 520;
 const CHAT_LOGS_MIN_WIDTH = 340;
 const CHAT_LOGS_MAX_WIDTH = 920;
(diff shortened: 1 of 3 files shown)


jeremyfowers (changes requested):
The control needs to be labeled in a way that is intuitive to users. "Normal" and "Off" don't convey any information about what the control does.

Here's how the same control looks on Claude

Image

and ChatGPT

Image

also: my 2 cents is that we should stop adding things and focus on shipping the MVP.

fl0rianr:
Interesting, for me it's more lean on the ChatGPT side:
Only the current setting is displayed:
image
If I hover the background comes up as above and similar what I did.

Additionally during hover a "Thinking effort" pill  with Crtl + Shift + M is displayed.

And finally with clicking on it we see a effort control. But I guess once we click the claude solution above fits better to use.

So my proposal to make this working for you would be the following:
1. Add a pill upon over properly describing what that control those "Thinking [Crtl + Shift + M] " (the shortcut is a pill inside of the pill.)
2. Change the names: On / Off, with a short description within the dropdown.

Regarding MVP, this is MVP for me at least partly. Since you can set it in GUI2 globally and a good Design foundation for other thinking settings / reasoning modes I have no doubt users will demand them. Feedback on that approach also.

jeremyfowers:
My main issue with the label is that Normal/Off are not standard terms.

Thinking/Off would be intuitive.

High/Medium/Low/Instant is also intuitive, because its so common in cloud AI, except that we don't actually support such a granularity.

fl0rianr:
> Thinking/Off would be intuitive.

On and Off are both intuitive and a good way to start.

> High/Medium/Low/Instant is also intuitive, because its so common in cloud AI, except that we don't actually support such a granularity.

All true, but we will have to support that kind of granularity soon enough. That way we have a good foundation and users are used to it and will not be overpowered if more options are coming.

jeremyfowers:
The problem with simply "On" as the primary label is that its not part of the existing popular vocabulary, in the way that "High", "Thinking", "Low", "Instant, and "Off" are. A new user looking at the word "On" doesn't strictly understand what is turned on without hovering the tooltip.

fl0rianr:
image
image
Now it's clearly better. If you have a big long thinking model like the new Qwen3.8 27B you really wand to be able to switch it on and of - Really nice!

jeremyfowers:
I just tried this and it didnt work, is that because https://github.com/lemonade-sdk/lemonade/pull/3142 has to land first, and then update here?

fl0rianr:
Exactly as said in the PR. Thanks

jeremyfowers:
image


Not sure if caused by this branch or not, but there is a responsive design issue with the hint text.

edit: yes confirmed introduced by this branch. the text box needs to get taller to fit the text, or it needs to scroll.

Everything else works correctly in this PR. Please share a screenshot of correct responsive design and then we can merge.

jeremyfowers (changes requested):
see comment

fl0rianr:
Screenshot From 2026-08-16 00-46-27

jeremyfowers (approved):

PR #3135, opened 2026-08-14
feat(server,cli): add discovery and broadcast controls with config decoupling

abn:
Add configuration and CLI controls to enable or disable server UDP beacon broadcasting (`lemond --broadcast` / `--no-broadcast`) and client auto-discovery (`lemonade --discovery` / `--no-discovery`). Broadcast state is configured via a new `broadcast` boolean key (defaulting to `true`) with backward-compatible normalization for `no_broadcast`. Transient CLI startup overrides are isolated in process memory to prevent leaking into persistent `config.json` on dynamic configuration updates.

- **Server CLI (`lemond`)**: Add `--broadcast,!--no-broadcast` with tristate `std::optional<bool>` override semantics.
- **Client CLI (`lemonade`)**: Add `!--discovery,--no-discovery` flags for auto-discovery control.
- **Configuration & Migration**: Add `broadcast: true` to `defaults.json` and `RuntimeConfig`, with automatic on-disk migration of legacy `no_broadcast` in `ConfigFile::load`, dynamic side-effects for runtime beacon start/stop, and backward-compatible input normalization.
- **Persistence Isolation**: Decouple CLI startup overrides from `config.json` persistence via `RuntimeConfig::broadcast_override_`, ensuring `snapshot()` only serializes persistent settings.
- **Tests**: Add unit test suite `test_config_discovery` (`CI ON`), clean up telemetry test boilerplate into shared helper `test_config_helpers.h`, and add integration test `test_045_config_set_broadcast`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2872,11 +2872,18 @@ if(BUILD_TESTING AND EXISTS "${_CONFIG_MIGRATION_TEST_SRC}")
     add_cpp_ci_test(ConfigMigrationTest CI OFF COMMAND test_config_migration)
 endif()
 
+set(_CONFIG_DISCOVERY_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_config_discovery.cpp")
+if(BUILD_TESTING AND EXISTS "${_CONFIG_DISCOVERY_TEST_SRC}")
+    add_executable(test_config_discovery test/cpp/test_config_discovery.cpp)
+    target_link_libraries(test_config_discovery PRIVATE lemonade-server-core)
+    add_cpp_ci_test(ConfigDiscoveryTest CI ON COMMAND test_config_discovery)
+endif()
+
 set(_CONFIG_TELEMETRY_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_config_telemetry.cpp")
 if(BUILD_TESTING AND EXISTS "${_CONFIG_TELEMETRY_TEST_SRC}")
     add_executable(test_config_telemetry test/cpp/test_config_telemetry.cpp)
     target_link_libraries(test_config_telemetry PRIVATE lemonade-server-core)
-    add_cpp_ci_test(ConfigTelemetryTest CI OFF COMMAND test_config_telemetry)
+    add_cpp_ci_test(ConfigTelemetryTest CI ON COMMAND test_config_telemetry)
 endif()
 
 set(_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC
--- a/docs/dev/getting-started.md
+++ b/docs/dev/getting-started.md
@@ -641,7 +641,7 @@ Accepts a JSON object with one or more keys to update atomically. Returns `{"sta
 | `host` | string | HTTP rebind |
 | `log_level` | string (`trace`, `debug`, `info`, `warning`, `error`, `fatal`, `none`) | Reconfigures log filter |
 | `global_timeout` | int (positive) | Updates default HTTP client timeout |
-| `no_broadcast` | bool | Stops or starts UDP beacon |
+| `broadcast` | bool | Starts or stops UDP beacon |
 | `extra_models_dir` | string | Updates model manager search path |
 
 **Deferred keys** (affect the next model load or eviction decision, no immediate side effect):
--- a/docs/embeddable/runtime.md
+++ b/docs/embeddable/runtime.md
@@ -167,7 +167,7 @@ Accepts a JSON object with one or more keys to update atomically. Returns `{"sta
 | `host` | string | HTTP rebind |
 | `log_level` | string (`trace`, `debug`, `info`, `warning`, `error`, `fatal`, `none`) | Reconfigures log filter |
 | `global_timeout` | int (positive) | Updates default HTTP client timeout |
-| `no_broadcast` | bool | Stops or starts UDP beacon |
+| `broadcast` | bool | Starts or stops UDP beacon |
 | `models_dir` | string (`"auto"` or path) | Updates the primary model cache location |
 | `extra_models_dir` | string | Validates access and updates the external GGUF search path |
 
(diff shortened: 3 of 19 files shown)


fl0rianr (changes requested):
Overall this looks solid and the latest migration fix addresses the main backward-compatibility issue I initially spotted. I still see a few things I’d want fixed before merge:

* Blocking: clearing broadcast_override_ can change the effective broadcast state without adding anything to applied_diff. Example: persisted broadcast=false, start with --broadcast, then config set broadcast=false. The override is cleared, config.broadcast() becomes false, but the side-effect callback is skipped because the persisted value did not change, so the beacon keeps running. The inverse case exists with --no-broadcast + persisted true. The diff should reflect changes to the effective value, not only config_.
* Please add a regression test for that override → config set transition, including verifying that the side-effect callback fires.
* The legacy no_broadcast migration should also be tested through the real ConfigFile::load() + save path, not only through RuntimeConfig construction, since the defaults/user-config merge is the important upgrade path here.
* test_045_config_set_broadcast currently assumes port 13305. server_cli2.py supports LEMONADE_TEST_PORT, and --no-discovery status will fail or hit the wrong server on a non-default test port. Also, on the normal port the test does not actually prove discovery is enabled/disabled because localhost is already reachable.
* Smaller points: conflicting broadcast + no_broadcast inputs should probably be rejected rather than silently preferring broadcast; malformed persisted broadcast values currently fail open to broadcasting; and the replicated config CLI parser in the unit-test helper is weaker coverage than exercising the real path.

With the effective-state/side-effect bug fixed and the migration/discovery coverage tightened up, I think this is in good shape.

abn:
Thank you for the review and catching those subtle transitions! All feedback has been addressed in commit `f0d77b1e`:

1. **Effective broadcast state diffing & side-effect firing (Blocking)**:
   - In `RuntimeConfig::apply_changes`, diff tracking now compares against the pre-clearing effective state (`broadcast()`).
   - When transitioning from a transient CLI override (`--broadcast` or `--no-broadcast`) to a `config set` matching the underlying persisted value, the override is cleared and `applied_diff["broadcast"]`, triggering `apply_config_side_effects` to correctly start/stop the UDP beacon.

2. **Regression tests for override → set transitions**:
   - Added test cases in `test_config_discovery.cpp` for both `--broadcast` + persisted `false` $\to$ `set(broadcast=false)` and `--no-broadcast` + persisted `true` $\to$ `set(broadcast=true)`, asserting that `broadcast()` updates, `applied_diff` contains the transition, and the side-effect callback fires.

3. **On-disk `ConfigFile::load()` + `save()` migration test**:
   - Added tests in `test_config_discovery.cpp` creating a real temporary directory with legacy `{"no_broadcast": true}`, invoking `ConfigFile::load()`, and validating that both in-memory output and on-disk `config.json` are cleanly upgraded to `{"broadcast": false}` with `no_broadcast` removed.

4. **Integration test port parameterization**:
   - Updated `test_045_config_set_broadcast` in `test/server_cli2.py` to pass `--port str(PORT)` to all CLI invocations, ensuring compatibility with custom `LEMONADE_TEST_PORT`.

5. **Conflicting inputs, type validation & unit test cleanup**:
   - `normalize_config_set_changes` now explicitly rejects conflicting `broadcast` + `no_broadcast` payloads with `std::invalid_argument`.
   - `RuntimeConfig` constructor and `set()` validate that `broadcast` / `no_broadcast` are strictly booleans.
   - Refactored `test_config_discovery.cpp` to test `RuntimeConfig::set` and `lemon::CLIParser` directly.

fl0rianr (commented):
Thanks for the follow-up, the previous blocking issue is fixed and the added regression coverage looks good. I found one remaining compatibility issue plus two smaller points:

* Blocking: legacy no_broadcast in deployment defaults (/usr/share/lemonade/defaults.json / LEMONADE_DEFAULTS_PATH) can be ignored after the upgrade. Built-in defaults now already contain broadcast=true, so merging a legacy override like {"no_broadcast": true} produces both keys. RuntimeConfig only converts no_broadcast when broadcast is absent, so the legacy override is effectively lost and broadcasting becomes enabled. I’d normalize the legacy key per config source before merging so precedence is preserved, and add a regression test using LEMONADE_DEFAULTS_PATH.
* Related: malformed legacy values such as "no_broadcast": "true" are currently removed by ConfigFile::load() and replaced by the default broadcast=true instead of being rejected. Ideally the same normalization path should validate the type before migration.
* Non-blocking: the updated test_045_config_set_broadcast is portable now, but passing --port makes explicit_target=true, which bypasses discovery entirely. So the --discovery / --no-discovery assertions are only flag/parser smoke coverage, not actual discovery behavior.

Everything else from my previous review looks addressed, including effective-state diffing, side-effect firing, on-disk migration coverage, conflicting input validation, and removal of the replicated discovery parser test. With the deployment-default migration issue fixed, I’d be comfortable approving this.

abn:
Rebased on latest `upstream/main` and addressed all follow-up feedback in `7651cf1d`:

1. **Per-source normalization before defaults merge (Blocking)**:
   - Added `normalize_legacy_keys` in `src/cpp/server/config_file.cpp` to normalize legacy keys (`"no_broadcast": true` $\to$ `"broadcast": false`) per source before deep-merging in `ConfigFile::get_defaults()` and `ConfigFile::load()`.
   - Distro defaults (`/usr/share/lemonade/defaults.json`) and environment defaults (`LEMONADE_DEFAULTS_PATH`) containing legacy `no_broadcast` now cleanly override the base default `broadcast=true`.
   - Added automated regression tests in `test_config_discovery.cpp` using a temporary `LEMONADE_DEFAULTS_PATH` file.

2. **Malformed legacy value rejection**:
   - `normalize_legacy_keys` validates that `no_broadcast` and `broadcast` are strictly booleans, throwing `std::invalid_argument` if malformed non-boolean values are provided rather than silently discarding them.
   - Added unit test coverage in `test_config_discovery.cpp` verifying rejection of malformed values.

3. **Test verification**:
   - All 42 discovery unit tests in `test_config_discovery` and 39 CI tests under `ctest -L cpp-ci` build and pass cleanly.

fl0rianr (approved):
LGTM

ianbmacdonald:
This fixes #3143: the `--no-broadcast` flag the docs and log line referenced without implementing now exists on `lemond` via this PR. The body carries no `Fixes:` link, so #3143 is still open and can be closed.

PR #3137, opened 2026-08-14
fix(gui): chat log position and overlay behavior

fl0rianr:
image

This fixes the log overlay error in the chat page.
Starts well inside of the screen depending on it's size to be clearly visible. After Initial usage the position is remembered as originally intended.

The issue that image mode controls where on top or that moving the log moved the whole chat panel are resolved now.

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -119,15 +119,24 @@ function getDownloadStoreModule(): Promise<typeof import('../features/downloadMa
   return downloadStoreModulePromise;
 }
 
-const CHAT_LOGS_WIDTH_KEY = 'chat_logs_panel_width';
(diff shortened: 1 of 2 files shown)


PCAssistSoftware (approved):
Works perfectly for me, problem resolved :)

jeremyfowers (changes requested):
I'm sorry, this doesn't for for me as a solution to the issue.

The purpose of a logs view in the chat tab is to be able to observe logs while chatting. 

This solution obscures the chat, forcing me to switch back and forth between logs and chat. If I wanted to do that, I would just switch back and forth between the chat tab and logs tab.

To ship the MVP ASAP I suggest deleting the log overlay control entirely. We can revisit this if we feel a need for it.

jeremyfowers (approved):
Perfect!

PR #3138, opened 2026-08-14
GUI3 fix: remove the "Loaded right now" cards from the chat tab

### LINKED ISSUE #2978: Redesign the chat tab model loaded card
Image

Details tbd

jeremyfowers:
Closes #2978.

Deletes the "Loaded right now" card grid from the chat tab empty state, plus the dead code it leaves behind: the `EmptyState` props that only fed it, `handleLoadedCardUnload`, `modelModeBadge`, `ChatView`'s `onOpenModelDetails` prop and its `openModelDetails` callback in `App.tsx`, and the `.section-label` / `.active-models` / `.active-card*` rules in both `styles.css` and `critical.generated.css`.

There are already 3 other ways to see what models are loaded — the chat tab dropdown, the models tab list, and the performance dashboard models widget — so we shouldn't spend time trying to optimize the design of a 4th.

### DIFF
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -880,11 +880,6 @@ const App: React.FC = () => {
     return () => window.removeEventListener('keydown', onKeyDown);
   }, []);
 
-  const openModelDetails = useCallback((modelName: string) => {
-    setModelDetailsRequest({ modelName, nonce: Date.now() });
-    setView('models');
-  }, [setView]);
-
   // Desktop host deep-links use the same canonical routes as browser navigation.
   useEffect(() => {
     let cancelled = false;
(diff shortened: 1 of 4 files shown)


PCAssistSoftware (commented):
Looks good to me

fl0rianr:
I have a UX question. This prevents the direct route from chat into a specific model details card in the models tab. I can understand if that is intentional and would not object, but do you see sense in bringing the link inside the chat tab dropdown?
image
E.g. using an icon like this?

PCAssistSoftware:
> I have a UX question. This prevents the direct route from chat into a specific model details card in the models tab. I can understand if that is intentional and would not object, but do you see sense in bringing the link inside the chat tab dropdown? image E.g. using an icon like this?

Like most things, I suppose it is all down to personal preferences / expectations.  To be honest I didn't even notice you couldn't do that after the PR, and personally if I wanted more details on a model I had installed I wouldn't have even looked for an icon to do it in chat, I would have just automatically gone to models tab to do so, but as I say we all do things differently

PR #3141, opened 2026-08-14
GUI3 refactor(gui): loaded in chat view

fl0rianr:
image

Here is the first version ready to be improved.

For testing, commit 1 is the main change, commit 2 changes a click on the card switches the model does feel really smooth to change that way - I preferred this.
Please try it out an tell me your ideas to land it.

### DIFF
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -384,6 +384,7 @@ const App: React.FC = () => {
   const customModelInfos = useMemo(() => serverModels.filter(model => (model as any).custom === true), [serverModels]);
   const rawLoadedModels = serverModelState.health?.all_models_loaded ?? EMPTY_LOADED_MODELS;
   const [currentModel, setCurrentModel] = useState<string | null>(null);
+  const [modelSelectionEpoch, setModelSelectionEpoch] = useState(0);
   const [theme, setTheme] = useState<Theme>(loadTheme);
   const [clientDataResetNonce, setClientDataResetNonce] = useState(0);
   const [modelHelpers, setModelHelpers] = useState<ModelHelpers | null>(null);
@@ -948,6 +949,7 @@ const App: React.FC = () => {
 
   const handleModelSelect = useCallback((modelName: string) => {
     setCurrentModel(modelName);
+    setModelSelectionEpoch(epoch => epoch + 1);
     setView('chat');
   }, [setView]);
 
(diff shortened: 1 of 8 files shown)


PCAssistSoftware (approved):
I like it, looks nice and clean and works well, switching between models really nice

jeremyfowers:
Screenshot From 2026-08-18 14-31-15

image

Design notes:
- There are 3 UI elements in the gui (see above) that display a list of loaded models and provide eject controls. The chat tab `Loaded now` element needs to be visually aligned to the other two.
- The green "running" status indicator should be removed. The design convention app-wide is that a status indicator should only show to distinguish one member from the others; for example in the model select dropdown the green status distinguishes from the default status, which is "not loaded". On the monitor tab there is no status indicator, since the default is "loaded" and there are no other states.
- The per-model member elements are very large, consider shrinking them or using the space more effectively.
- The eject ghost button on the right side breaks from the app-wide convention.

One concrete suggestion would be to copy the loaded models content of the Monitor tab with only one exception, displaying different metadata. No Use button, always display the eject button, use the conventional yellow outline and fill to show selection state.

jeremyfowers:
One other thing I feel is a UX bug: if I select a chat model from the dropdown menu, the suggested prompts vanish from the top of the screen and they are replaced by a single-entry Loaded Now widget that only contains the model I just clicked. That isn't very helpful, I know what model I just clicked so I dont need a widget to tell me. The user has a higher chance of wanting to use one of the chat suggestions.

fl0rianr:
image

fl0rianr:
image

jeremyfowers:
> One other thing I feel is a UX bug: if I select a chat model from the dropdown menu, the suggested prompts vanish from the top of the screen and they are replaced by a single-entry Loaded Now widget that only contains the model I just clicked. That isn't very helpful, I know what model I just clicked so I dont need a widget to tell me. The user has a higher chance of wanting to use one of the chat suggestions.

Sorry for confusion. I meant that:

1. the Loaded Now widget should only display when the New Chat button is pressed or the gui is freshly loaded AND a model is loaded. 
2. picking a model from the model select dropdown (or the Models tab) should not trigger the Loaded Now widget to be displayed.

So we see either the Loaded Now widget or the chat suggestions, depending on that context.

jeremyfowers:
Tested and I like it! 

Please consider these small design changes before merging:
1. Move the "Chat", "Image", etc. modality metadata out from under the icon, and into the metadata row
2. Remove the pill around the device metadata, e.g., "GPU"
3. Move the model size metadata into the metadata row
4. Normalize the size of the buttons: right now the pin button looks really small to me (this may not be as much of an issue once the size metadata has moved, though.)

That will make this control match the unified design language from the rest of the gui.

jeremyfowers:
image

fl0rianr:
Those this work for you? Make the logo smaller again (in between?) was quite small after removing the language.

image

jeremyfowers (approved):
Love it!

PR #3144, opened 2026-08-14
Update FLM version

zaneni6:
Update FLM version to `1.0.1`.


src/cpp/resources/backend_versions.json
```diff
@@ -130,7 +130,7 @@
     "npu": "v1.7.0"
   },
   "flm": {
-    "npu": "v0.9.46"
+    "npu": "v1.0.1"
```
jeremyfowers, line 133: @zaneni6 can you please use this opportunity to also change the url pointers from your old github address to your new one under rocm org?

jeremyfowers (approved):
LGTM!

PR #3147, opened 2026-08-14
fix(server): add --no-broadcast CLI flag and LEMONADE_NO_BROADCAST env var

### LINKED ISSUE #3143: --no-broadcast CLI flag doesn't exist despite docs/logs claiming it does
## Bug Description

Documentation and a server log message both claim `lemond` has a `--no-broadcast` command-line flag to disable its UDP discovery beacon. **No such CLI flag exists.** The UDP beacon (used for LAN server discovery) can currently only be disabled by setting `"no_broadcast": true` in `config.json`, or by hitting the runtime `params`/management API after the server is already running. There is no way to disable it at container/process launch time via a flag or environment variable.

This matters in practice for containerized deployments (e.g. rootless Podman with the `pasta` network plugin — https://passt.top/passt/about/): with no way to disable broadcasting *before* the server starts, the beacon always attempts a UDP broadcast on first startup. Under rootless Podman, `pasta` cannot establish the broadcast flow and rejects it. `pasta` runs as a host-side userspace process (not inside the container), so this doesn't show up in `podman logs` — it spams the host's **journald** logs with a message every ~2 seconds (matching the beacon's hardcoded broadcast interval), until someone takes the separate, out-of-band step of editing the mounted `config.json` or calling the management API to turn it off:

```
Dropping datagram with no flow TAP 10.23.234.20:33503 -> 10.23.234.255:13305
Flow 1 (UDP flow): Couldn't connect flow socket
: Permission denied
```

## Where the documentation/log claims a CLI flag exists

- `docs/dev/getting-started.md:612`: *"The custom flag --no-broadcast is available in the command line to disable."*
- `src/cpp/server/server.cpp` (startup path, when `no_broadcast` is set in config): logs `"Broadcasting disabled by --no-broadcast option"`.

## Where it's actually implemented (config-only)

- `src/cpp/server/cli_parser.cpp` (`CLIParser::CLIParser()`): the only CLI args `lemond` registers via CLI11 are the positional `cache_dir`, `--port`, `--host`, and `-v/--version`. There is no `--no-broadcast` flag defined anywhere.
- `no_broadcast` is only readable from `config.json` via `RuntimeConfig::no_broadcast()` (`src/cpp/server/runtime_config.cpp`), seeded from `src/cpp/resources/defaults.json` (default `false`).
- The beacon start/stop gating on this config key lives in `src/cpp/server/server.cpp` (initial startup near `udp_beacon_.startBroadcasting(...)`, and the hot-reload handler for `no_broadcast` in `apply_config_side_effects`). This gating itself works correctly — the problem is purely that it's unreachable except via the config file or the params API.
- Beacon implementation: `src/cpp/server/utils/network_beacon.cpp` (`NetworkBeacon::broadcastThreadLoop`, interval hardcoded to 2s at the call sites).

## Requested Fix (preferred: add the flag)

Add a real `--no-broadcast` CLI flag to `lemond`, wired up in `cli_parser.cpp` the same way `--port`/`--host` already override `config.json`. Ideally also expose it as an environment variable (e.g. `LEMONADE_NO_BROADCAST`), consistent with the existing per-key env var pattern used elsewhere (`LEMONADE_API_KEY`, `LEMONADE_ADMIN_API_KEY`, `LEMONADE_ALLOWED_ORIGINS`, etc. — each individually read via `std::getenv()`, there is no generic env-to-config mechanism to piggyback on).

This is the preferred fix over just correcting the docs/log message, because container/orchestration deployments (Podman, Kubernetes, Docker Compose) need to be able to disable the beacon at launch time via flags/env vars/`docker run` args — not by pre-seeding a mounted config file or making a follow-up API call after the server is already up and has already attempted (and failed) to broadcast at least once.

If adding the flag is out of scope, the fallback is to fix the documentation (`docs/dev/getting-started.md:612`) and the misleading log message in `server.cpp` so neither claims a CLI flag exists — but this leaves the underlying rootless-Podman/pasta log-spam problem unsolved for anyone who can't pre-seed `config.json` before first launch.

## Environment (where this was observed)

- Deployment: rootless Podman, `pasta` network plugin
- Image: `ghcr.io/lemonade-sdk/lemonade-server:v11.5.2`
- Component: `lemond` (server)

petergil:
## Summary
- Adds a real `--no-broadcast` CLI flag to `lemond` and `LemonadeServer.exe`, persisting to `config.json` the same way `--port`/`--host` already do.
- Adds a `LEMONADE_NO_BROADCAST` env var for container deployments (e.g. rootless Podman with the `pasta` network plugin, which rejects the UDP beacon's broadcast and floods the host's journald). It's checked live via `RuntimeConfig::get_bool_opt` rather than written into `config.json`, so it never persists past the process that set it.
- Fixes a misleading log message and docs that referenced this CLI flag before it actually existed.

Fixes #3143

## Test plan
- [ ] `lemond --no-broadcast` and confirm the UDP beacon doesn't start, and that `config.json` gets `"no_broadcast": true`
- [ ] `LEMONADE_NO_BROADCAST=true lemond` and confirm the beacon doesn't start, and `config.json` is left untouched
- [ ] Verify `LemonadeServer.exe --no-broadcast` behaves the same as `lemond --no-broadcast`

### DIFF
--- a/docs/dev/getting-started.md
+++ b/docs/dev/getting-started.md
@@ -609,7 +609,7 @@ The client automatically:
 - Uses port 13305 to broadcast to the network that it exists
 - Clients can read the json broadcast message to add server to server picker.
 - Uses machine hostname as broadcast name.
-- The custom flag --no-broadcast is available in the command line to disable.
+- The `--no-broadcast` command-line flag disables it and persists to `config.json`, same as setting the `no_broadcast` key directly. `LEMONADE_NO_BROADCAST` (accepts exactly `1`, `true`, `True`, `TRUE`, `yes`, or `YES` — no other casing) disables it for the life of the process without touching `config.json`, and takes precedence over both whenever it's set — for container use.
 - Auto protection, doesnt broadcast on non RFC1918 Networks.
 
 ### Internal Endpoints
--- a/docs/guide/install/docker.md
+++ b/docs/guide/install/docker.md
@@ -56,6 +56,32 @@ docker run -d \
 
 > This will run the server on port 5000 inside the container, mapped to port 4000 on your host.
 
+### Docker/Podman Run with the Discovery Beacon Disabled
+
+`lemond` periodically broadcasts a small UDP beacon on port 13305 so LAN clients can auto-discover it. In some container networking setups — notably rootless Podman with the `pasta` network plugin — the container isn't permitted to send broadcast traffic, and every failed attempt logs a `Permission denied` message to the host's journal every couple of seconds.
+
+Disable the beacon with the `LEMONADE_NO_BROADCAST` environment variable (no need to override the container's command). Unlike the flag/config.json forms below, this is not persisted, so it must be passed on every `run`:
+
+> `LEMONADE_NO_BROADCAST` always takes precedence while it's set (over both `config.json` and any `--no-broadcast` flag). It accepts exactly `1`, `true`, `True`, `TRUE`, `yes`, or `YES` — other casings (e.g. `Yes`, `TRUe`) are not recognized and are treated as unset.
+
+```bash
+podman run -d \
+  --name lemonade-server \
+  -p 13305:13305 \
+  -e LEMONADE_NO_BROADCAST=true \
+  ghcr.io/lemonade-sdk/lemonade-server:latest
+```
+
+Equivalently, pass `--no-broadcast` on the command line (this requires specifying the full command, since it replaces the image's default `CMD`), or set `"no_broadcast": true` in `config.json`:
+
+```bash
+podman run -d \
+  --name lemonade-server \
+  -p 13305:13305 \
+  ghcr.io/lemonade-sdk/lemonade-server:latest \
+  ./lemond --host 0.0.0.0 --no-broadcast
+```
+
 ### Docker Run with CPU backend
 
 To use the CPU backend, create or modify the `config.json` file in the `lemonade-recipe` volume:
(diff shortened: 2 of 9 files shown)


jeremyfowers:
Makes sense to me, @Geramy would you kindly review?

Geramy (approved):
Yeah this looks good :)

superm1:
Do we really need another env variable?  Can't the container just launch lemond with `--no-broadcast`?

Geramy:
> Do we really need another env variable? Can't the container just launch lemond with `--no-broadcast`?

I would think the docker container file could be adjusted to add that depending on how it gets launched.

superm1:
> > Do we really need another env variable? Can't the container just launch lemond with `--no-broadcast`?
> 
> I would think the docker container file could be adjusted to add that depending on how it gets launched.

So perhaps add it to our Dockerfile too then to launch this way and drop the env variable.

Geramy:
> > > Do we really need another env variable? Can't the container just launch lemond with `--no-broadcast`?
> > 
> > 
> > I would think the docker container file could be adjusted to add that depending on how it gets launched.
> 
> So perhaps add it to our Dockerfile too then to launch this way and drop the env variable.

sounds good @petergil can you make the changes as Mario has suggested? 
We have moved away from environment variables and he has a point about that, we really dont need both a switch and env var.

Geramy (changes requested):
change PR to fix Dockerfile instead of introducing a new redundant environment variable, or submit a new PR and close this one.

petergil:
This PR is superseded by #3135 so closing this one.

PR #3149, opened 2026-08-14
GUI enhancements to the router

meghsat:
1. Added collapsable panels to give the router builder more space
image

2. Added these separators
image

3. Made the trash icon always visible. 
Included the rule and model routes into this panel for better spacing
image

4. Earlier gates with a single condition would blindly collapse but now you'll get a warning that at least 2 conditions are needed to form a gate.
image

5. Fixed a bug where adding a single keywords-any gate would not display it on the canvas but only on the node inspector

6. Better aligned these
image

7. Cleaned the router and model headers
8. Expanded the graph builder size and added a scroll bar to adjust the node inspector and graph canvas size
image

9. Following a uniformed dropdown style
image

10. The dropdown for the latest classifier is cut off
image
to
image

### DIFF
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -1191,8 +1191,8 @@ const App: React.FC = () => {
                 className="titlebar__window-btn titlebar__window-btn--close"
                 data-tauri-drag-region="false"
                 onClick={() => window.api?.closeWindow?.()}
-                aria-label="Close"
-                title="Close"
+                aria-label="Cancel"
+                title="Cancel"
               >
                 <Icon name="x" size={14} />
               </button>
--- a/src/app/src/components/EffectiveSettingsModal.tsx
+++ b/src/app/src/components/EffectiveSettingsModal.tsx
@@ -395,8 +395,8 @@ const EffectiveSettingsModal: React.FC<EffectiveSettingsModalProps> = ({
             iconOnly
             className="close-modal-btn"
             onClick={onClose}
-            aria-label="Close"
-            title="Close"
+            aria-label="Cancel"
+            title="Cancel"
           />
         </div>
 
--- a/src/app/src/components/ModelListPanel.tsx
+++ b/src/app/src/components/ModelListPanel.tsx
@@ -465,8 +465,6 @@ export interface ModelListPanelProps {
   onOpenCustomModels?: () => void;
   onOpenRouter?: () => void;
   onUpdateAllModels?: () => void;
-  onToggleListPanel?: () => void;
-  listPanelCollapsed?: boolean;
   /** Lowercased set of pinned model names. Pinned rows float to the top. Client-local. */
   pinnedNames?: Set<string>;
   /** Toggle a model's pinned state. Receives the model name. */
@@ -501,8 +499,6 @@ export const ModelListPanel: React.FC<ModelListPanelProps> = ({
   onOpenCustomModels,
   onOpenRouter,
   onUpdateAllModels,
-  onToggleListPanel,
-  listPanelCollapsed,
   pinnedNames,
   onTogglePin,
   favoriteNames,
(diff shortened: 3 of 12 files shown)


fl0rianr (changes requested):
Thanks for making this fixing PR! Very helpful!

Overall the UI direction looks good, but I still see a few issues that should be addressed before merging:

Out of Scope UI changes:

1. The second collapse option, this might be acceptable but I have to check this aspect with @jeremyfowers since it's not necessary outside of the specific router setup.
Since the height is also limited I thing the builder will be expanded anyway for more complex setups that's why I marked that design expand icon as primary action element.

2.  The added Separators: Not necessary, for the router UI. I'm opposing this change.

3. trash icon visible - it this way throughout the app and in other cases visible is not ok. I don't see this as an access issue for the delete option as following the same design logic for all lists.


Further critical points:

1. **`RouterSelect` is not fully keyboard accessible.** The native selects were replaced with a custom listbox, but there is no ArrowUp/Down navigation, active option focus, Enter/Escape handling, etc. In the expanded graph modal the options are also portaled outside the modal focus trap, and Escape currently closes the whole modal instead of just the dropdown.

2. **Dropdown positioning can become incorrect.** The popover uses `position: fixed`, while its coordinates include `window.scrollX/Y`. It is also only positioned when opened, so scrolling the inspector or resizing the window can leave the dropdown detached from its trigger. Long option lists also need a proper max-height/scroll treatment.

3. **The empty `keywords_any` fix is not fully represented in history/state.** Whether the blank node is the initial empty graph or an intentionally added condition is stored only in `hasCommittedRef`. Undo/redo and the inline/expanded graph instances do not preserve/share that state, so the empty state can be shown incorrectly and the original issue can still appear in some flows.

4. **The new inspector resizer is pointer-only.** It has `role="separator"` but no `tabIndex`, keyboard resize handling, or `aria-valuemin/max/now`. The existing `WorkspacePanelResizer` already implements this correctly and would be a good reference.

5. **Minor: some real Close actions were renamed to Cancel.** The window close button still calls `closeWindow()` and the Effective Settings X closes the dialog, so their accessible labels/titles should stay `Close`, not `Cancel`.

The single-condition AND/OR gate behavior itself looks good to me: preserving the gate and surfacing validation instead of silently collapsing it is consistent with the intended change.

6 ally npm test failed also.

meghsat:
Thanks for the feedback!
Left out the UI change suggestions: 1&2 until we get some clarity from others. Fixed the other issues.

fl0rianr (commented):
Thanks for addressing the previous feedback! I checked the latest version again and I do not want to hold this up over minor points.

Most of my earlier technical concerns look resolved now. I only see two things I would still fix before merging.

First, long dropdowns can close while scrolling inside the dropdown itself. This makes the new scrollable dropdown behavior unreliable for longer lists.

Second, the empty Keywords Any case still seems broken when switching to the expanded graph builder. It works in the current inline builder, but the expanded builder starts fresh and treats that intentionally added empty condition as the initial empty canvas again. Since this PR specifically addresses that bug, I think this one is worth fixing before merge.

Everything else from my previous review is either addressed or not something I would block the PR on. The remaining collapse and separator questions are design decisions from my side, not technical blockers.

jeremyfowers:
Please move out-of-scope model list panel changes to their own PR and see feedback I posted on discord https://discord.com/channels/1392562559122407535/1514989303594483802/1538194755052376074

PS. I would prefer to not change the models list panel UI at all at this point, we worked on it a lot already.

meghsat:
Thanks for the feedback!
I removed the model and filter panel UI changes. Fixed the remaining 2 issues: keywords-any and dropdown bugs @fl0rianr

fl0rianr:
Yery nice, right before approving but found one issue with the node inspector. Right after expanding the inspector took roughly 3/4 of the screen. When I closed it there is no option to bring it back. Even a restart of the setup did not work. A restart should always be a way around certain errors.

meghsat:
Hey @fl0rianr, To get the node inspector back, you would have to press on the gate/ condition, right?

fl0rianr:
correction, there is away to bring it back clicking at a node sorry.

fl0rianr (approved):
LGTM

PR #3156, opened 2026-08-15
fix(gui): classify models by their deployment label

jeremyfowers:
Closes https://github.com/lemonade-sdk/lemonade/issues/3070 using the new chat label from https://github.com/lemonade-sdk/lemonade/pull/3099

**Bug:** any model carrying a `registry_source` was treated as a Hugging Face search row; that field defaults to `"huggingface"`, so installed chat models rendered **Unknown**.

**Fix:** read the deployment label the server already stamps. `modelCapabilities.ts` 521 → 371 — recipe hint lists, checkpoint-name regexes, `capabilityFromName`/`Type`/`Recipe` and the `preferNonChat` ladder deleted.

- Collections aren't a capability: Omni/Router deploy as `chat`, identity from the recipe.
- `classification` added (two ONNX models were falling through): chip + filter, no panel.
- Remote search rows keep inference, isolated — `unknown` there is honest.
- Authoring writes one label; older records repaired on read, never rewritten on disk.
- Commit 2 fixes three regressions from commit 1 (collections lost from the composer, Omni shown as "Chat", old custom models Unknown) and deletes a duplicate name-guessing classifier in `lemonadeTools.ts`.

**Tests:** new suite pins the lookup to `docs/api/openai.md` and all 214 registry models; new spec covers every label end-to-end. Typecheck clean, 45 functional + 146 a11y pass. `default-chat-model`, `directory-settings` and `model-provider-search` already fail on base; the `inspect-interactions` optimizer is flaky.

### DIFF
--- a/src/app/package.json
+++ b/src/app/package.json
@@ -20,6 +20,7 @@
     "test:a11y": "npx playwright test tests/a11y.spec.ts",
     "test:model-providers": "node tests/model-provider-search.runtime.cjs",
     "test:chat-audio": "node tests/chat-audio-capability.runtime.cjs",
+    "test:model-kinds": "node tests/model-kind-classification.runtime.cjs",
     "test:mcp-external": "node scripts/test-external-mcp.mjs",
     "test:mcp-mock": "node tests/mock-mcp-server.runtime.cjs",
     "test:mcp-runtime": "node tests/mcp-runtime.runtime.cjs",
--- a/src/app/src/App.tsx
+++ b/src/app/src/App.tsx
@@ -341,7 +341,7 @@ function loadSavedRoute(): AppRoute {
   return { view: 'chat' };
 }
 
-type ModelHelpers = Pick<typeof import('./modelCapabilities'), 'canSelectInComposer' | 'capabilityFromModelInfo' | 'selectPreferredLoadedModel'>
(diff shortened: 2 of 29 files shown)


fl0rianr (commented):
Potential import regression: `normalizeImportedCapability()` is the compatibility boundary for older/arbitrary JSON, but label-only aliases like `image-generation`, `image-to-3d`, or `text-to-speech` now fall through to `chat` when no explicit `type` is present.

Example:

{ "recipe": "sd-cpp", "labels": ["image-generation"] }

This used to classify as `image`, but now becomes `chat`, which can later produce an invalid `sd-cpp` + `chat` registration.

I’d keep `deploymentKindFromLabels()` strict and handle this only at the import boundary by canonicalizing legacy labels, e.g.:

`image-generation` -> `image`
`image-to-3d` -> `3d`
`text-to-speech` -> `tts`
`asr` / `stt` -> `transcription`
`reranker` -> `reranking`

Then run the normalized labels through the normal strict classifier.

That preserves the new architecture while keeping backwards compatibility. A few regression tests with missing `type` would be useful too.

jeremyfowers:
We have handling on the server side that raises helpful warnings for illegal json. IMO this should be sufficient for the rare cases where someone accidentally produced illegal JSON.

> That preserves the new architecture while keeping backwards compatibility.

Backwards compatibility to what? I wouldn't count backwards compatibility within GUI3's history as a goal. Is this backwards compatibility to GUI2?

fl0rianr (approved):
Good point, LGTM.

PR #3158, opened 2026-08-15
Replace /load/command with /models/{id}/options

jeremyfowers:
The effective-settings modal was the only caller of the experimental `/load/command`. It now uses `/models/{id}/options`, and shows that endpoint's curl command instead of the synthesized `lemonade load` one.

- Client-staged options still have to reach the server for the command to describe the load the app would actually send, so the modal's calls go through `dry_run: true` (resolves, persists nothing). The staged set spans every recipe, so it's filtered against `defaults` — an unrecognized name is a 400.
- `$LEMONADE_BASE_URL` is substituted with the client's base URL; `$LEMONADE_API_KEY` deliberately stays a variable rather than pasting a live credential into copyable UI text.
- Dead after the removal, both deleted: `Router::resolve_effective_recipe_options` (the merge kept it alongside main's identically-bodied `resolve_effective_options`) and `RecipeOptions::to_cli_options` plus its CLI-flag table.

`router.cpp` now matches main's blob exactly.

Verified: `ctest -L cpp-ci` 36/36; `test/server_endpoints.py` 132 passed; `test/server_cli2.py` 63 passed; `tsc --noEmit` clean. Against a live server, `/v1/load/command` returns 404 and the dry-run preview reflects draft args while leaving `saved` empty.

image

The Effective Settings GUI now shows the actual HTTP request the GUI will send, not the CLI request (I did not understand why the GUI should reference the CLI before).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,7 @@ All core endpoints are registered under **4 path prefixes**:
 - `/v0/` — Legacy short
 - `/v1/` — OpenAI SDK / LiteLLM compatibility
 
-**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `health`, `pull`, `pull/variants`, `registry/search`, `load`, `load/command`, `unload`, `delete`, `params`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `log-level`, `logs/stream`, `jobs`, `jobs/{id}`, `jobs/{id}/pause`, `jobs/{id}/interrupt`, `jobs/{id}/resume`
+**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `models/{id}/options`, `health`, `pull`, `pull/variants`, `registry/search`, `load`, `unload`, `delete`, `params`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `log-level`, `logs/stream`, `jobs`, `jobs/{id}`, `jobs/{id}/pause`, `jobs/{id}/interrupt`, `jobs/{id}/resume`
 
 **Job engine** (`POST jobs`, `GET jobs`, `GET/DELETE jobs/{id}`, `POST jobs/{id}/{pause,interrupt,resume}`): server-side sequences of ops (`system_info`, `system_stats`, `models`, `sleep`, `load`, `unload`, `chat`) with data passing, forward-only branching, and a pause/interrupt/resume lifecycle persisted across restart. Exclusive ops hold a Router slot so normal traffic queues. See `docs/dev/job-system.md` and `docs/dev/job-expression-language.md`.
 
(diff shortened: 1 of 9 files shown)


fl0rianr (commented):
I think there’s a semantic mismatch here between the preview and the actual GUI load.

`POST /models/{id}/options` applies changes on top of the currently saved options, while `loadModel()` sends `save_options: true`, which replaces the saved option set.

That means the preview can keep an old saved option that the real GUI load would actually remove. For example, if `llamacpp_args="--no-mmap"` is currently saved on the server but the GUI has no local override for it, the dry-run can still include `--no-mmap` in `load_command`, while the subsequent GUI load drops it.

Since this modal is meant to show the exact request the GUI will send, I think the dry-run should also account for saved keys that are absent from the staged load, e.g. by sending those as `null`, or the server could expose a preview path with the same replacement semantics as `/load` + `save_options: true`.

A regression test for this case would be useful too.

jeremyfowers:
I'm confused now. If all we care about is showing what command was used to load the model, the GUI already has that information. Why would we need an endpoint to tell us what command we used? Can the GUI just save the command it used and display it?

fl0rianr:
Yeah, I think that would work if the goal is to show the exact request the GUI sent.

The reason I brought up the endpoint is that what it returns is slightly different: it gives us the fully resolved load, including server-side defaults that the GUI itself doesn't necessarily know about. So it's more of a reproducible "effective load command" than literally the request the GUI sends.

I think that's where my confusion came from too. If we only want the actual GUI request, saving it client-side seems simpler. If we want the fully resolved/replayable command (let's say you have a bug - originally you could have used that one directly towards the backend -mostly), then using the endpoint makes sense, but the dry-run should match the same replacement semantics as the real GUI load.

At the end effective should be what the backend receives, not only what was sended.

jeremyfowers:
Ok I misunderstood this. 

Actions needed:
- We can probably drop the "effective command" feature out of /models/id/options since it seems useless now.
- Should each model's  load command simply be recorded on the wrapped server and returned as part of the /health endpoint? All other similar items of live server status are already there.
- The UI should not say "effective load command" if it was the actual load command. "Effective" makes it sound like a prediction.
- The GUI should not reference the CLI.

Do these actions make sense?

fl0rianr:
Yeah, these actions make sense to me. I like the /health idea, since the command actually used to launch the backend is runtime state and the wrapped server is the authoritative place for that.

I would actually keep both views in the UI though. The GUI can show the exact request it sent, since it knows that without having to reconstruct anything. After the model is loaded, /health could provide the actual backend launch command and the UI could let you switch between the two. 
CLI had the cleaner "look" and this is exactly what power users want and @pwilkin intended here.
Both versions - if someone is a dev he learns form the UI, if someone wants to debug he also learns form the UI critical informaton..

You are right about the names, I’d probably call the second one “Backend launch command” rather than “Effective load command”, and since “effective” does sound like a prediction, we could use "utilized" since the backend launch command is the "actual" one.

Only thing I’d be careful about is making sure anything returned through /health is sanitized/redacted in case a backend command ever contains credentials or other sensitive values. Up to know I'm not aware of something like that.

jeremyfowers:
Opened issues:
- https://github.com/lemonade-sdk/lemonade/issues/3191
- https://github.com/lemonade-sdk/lemonade/issues/3192
- https://github.com/lemonade-sdk/lemonade/issues/3193

and closing this change

PR #3159, opened 2026-08-15
fix(debian): add libomp1 runtime dependency for ROCm backends

wariobot09:
## Problem

Several .deb test jobs are failing with:

```
ace-server: error while loading shared libraries: libomp.so: cannot open shared object file: No such file or directory
```

This happens in the **Inference backend tests**, **Test .deb - audio-gen-acestep**, **Test .deb - audio-gen-thinksound**, and **Test .deb - tts-openmoss** jobs.

## Root Cause

The ROCm-built binaries for **acestep** (`ace-server`), **thinksound** (`ts-server`), and **openmoss** (`moss-tts-server`) all depend on `libomp.so` at runtime. The Debian control file had `libgomp1` but was missing `libomp1`.

## Fix

Added `libomp1` to the lemonade-server package runtime dependencies in `contrib/debian/control`.

## Verification

- All affected tests only fail on the `rocm-stable` backend path (vulkan builds succeed - they don't link libomp.so)
- This is a single-line dependency addition, no code changes needed
- The dependency was already listed in the CI setup scripts but missing from the package manifest


superm1:
Changing too much, this isn't appropriate.  This is at most a one line diff.

PR #3171, opened 2026-08-16
update TheNoise v0.2.1

bitgamma:
update TheNoise to version v0.2.1:

- add support for Z-Image-Turbo
- forward non-mapped request parameter to the backend as-is
- add upscaler_dir backend config to support additional upscaler models
- remove upscale boolean as from recipe option and keep it as request parameter
- remove lora_dir from recipe options and keep it as backend config (better security)

### DIFF
--- a/docs/dev/backends-reference.md
+++ b/docs/dev/backends-reference.md
@@ -146,12 +146,10 @@ the generator instead. Prose outside the markers is preserved. -->
 | `height` | — | SIZE | 512 | Output image height |
 | `sampler` | — | ARGS | "" | Denoising solver (euler \| er_sde) |
 | `negative_prompt` | — | ARGS | "" | Negative prompt |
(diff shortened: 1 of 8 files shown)


fl0rianr (commented):
Thanks. The new passthrough can re-forward parameters that were already handled by the adapter but intentionally omitted from `body`.

For example, `lora_specs: ""` is parsed as empty and skipped, but the passthrough then adds the raw empty string back. TheNoise expects `lora_specs` as a list, so this can cause validation errors.

I’d suggest skipping all adapter-handled keys explicitly and only passing through truly unmapped backend parameters.

bitgamma:
@fl0rianr thanks for catching that! Fixed now.

fl0rianr (approved):
LGTM. thanks

PR #3172, opened 2026-08-16
feat(server): support transient null overrides on load

fl0rianr:
### Summary

Add transient per-option clearing semantics to `POST /v1/load`.

An explicitly provided `null` now means: ignore the saved per-model value for this load only and fall through to the lower server/default layers.

This gives `/v1/load` three distinct states for recipe options:

* omitted: keep using the saved per-model value
* `null`: ignore the saved per-model value for this load only
* concrete value: override it for this load

`ctx_size: -1` remains the explicit auto-sizing value and is not treated as a clear.

### Why

This lets clients clear individual saved options without mutating `recipe_options.json`.

For example, a client can temporarily clear saved `llamacpp_args` without also losing unrelated saved settings. This is more precise than a global `ignore_saved_options` flag.

### Tests

Added coverage for:

* saved args → transient `null` clear
* another saved option remains active
* `merge_args` behavior
* explicit `ctx_size: -1`

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -967,6 +967,8 @@ Explicitly load a registered model into memory. This is useful to ensure that th
 
 > Note: loading a collection (`recipe: "collection.omni"`) loads each of its components in turn. Per-model options like `ctx_size` or `llamacpp_backend` are not forwarded to components — set them on each component's own `recipe_options.json` entry instead.
 
+Recipe option fields on `/v1/load` have three-state semantics. Omitting a field keeps using its saved per-model value. Passing explicit `null` ignores only that saved key for this load and falls through to the lower default layers without changing `recipe_options.json`. Passing a concrete value overrides the saved value. `ctx_size: -1` is a concrete value meaning automatic context sizing, not a tombstone. With `save_options: true`, concrete values are persisted as usual while a `null` tombstone preserves the existing saved value for that key.
+
 ### Parameters
 
 | Parameter | Required | Applies to | Description |
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -5793,10 +5793,18 @@ void Server::handle_load(const httplib::Request& req, httplib::Response& res) {
 
         auto info = model_manager_->get_model_info(model_name);
 
-        // Extract optional per-model settings. An omitted or empty option falls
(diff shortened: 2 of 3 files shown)


bitgamma (commented):
there is one edge case that is probably worth addressing: tombstones are silently dropped when the model needs downloading                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                              
 The rebuild runs before the download block:                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                              
 ```cpp                                                                                                                                                                                                                                                                                                                       
   // rebuild info.recipe_options with tombstones removed                                                                                                                                                                                                                                                                     
   info.recipe_options = RecipeOptions(info.recipe, per_model_options);                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                              
   if (!model_manager_->is_model_downloaded(model_name) && !is_model_collection_recipe(info.recipe)) {                                                                                                                                                                                                                        
       ...                                                                                                                                                                                                                                                                                                                    
       info = model_manager_->get_model_info(model_name);   // <-- overwrites the tombstoned recipe_options                                                                                                                                                                                                                   
   }                                                                                                                                                                                                                                                                                                                          
 ```                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
If the model isn't already downloaded, info is re-fetched from get_model_info(), which rebuilds recipe_options from the full saved layer (including tombstoned keys). The transient null clear then has no effect — the model loads with the saved value. This is reachable because options can be saved without loading     
 (POST /models/{id}/options) and download state is independent of the saved options. It affects both the save_options and non-save_options paths.                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                              
Fix suggestion: move the tombstone-rebuild to after the download block (after both info reassignments), so it is applied to the final info used by router_->load_model.                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                              
 Minor note: the tombstone-rebuild block and the save-write preservation block both re-read get_saved_model_options and could be factored into a small helper, but this is cosmetic.

fl0rianr:
Good catch. Moved the transient tombstone overlay after the download/re-fetch so it is applied to the final ModelInfo passed to the router. Kept this as a minimal ordering fix without factoring the surrounding logic.

bitgamma (approved):

PR #3180, opened 2026-08-16
fix(GUI app): stop chat from bypassing collection.router routing

meghsat:
## Problem
Selecting a collection.router model in the chat window and sending a prompt always hit the same component model, regardless of the routing rules - routing only worked via direct curl against the server.

The GUI treated routers like omni collections:

1. getCollectionPrimaryChatModel() resolved every collection to a single component, so the request body's model field carried a fixed component name and the server's routing engine never saw the router.
2. ChatWindow also put routers into omni collectionMode, attaching the agentic tool loop and a system prompt to every turn.

## Fix
1. getCollectionPrimaryChatModel() now returns a router's own model name, so the server routes each request per its policy.
Routers are excluded from collectionMode and use the plain streaming chat path.
2. Routers no longer show omni-mode chat extras (unload-collection button, tool loop). This is intentional, the agentic tool loop is a collection.omni feature, and running it against a router would attach tools and a system prompt to every turn, skewing has_tools / length conditions in the routing policy.


fl0rianr (approved):
LGTM

ramkrishna2910 (changes requested):
Routing the chat request through the server is the right call, but `getCollectionPrimaryChatModel()` feeds more than the request body — its return value becomes `chatModelName`, which `sendMessage`/`submitEdit` also hand to `runPreFlight` → `ensureModelReady()`. With the router's own name now flowing through, `ensureModelReadyInternal` (`src/app/src/renderer/utils/backendInstaller.ts:707`) takes the `isCollectionModel()` branch — the router has non-empty `components` — and sequentially `/load`s **every** candidate before the request is sent. Previously only the single resolved component was loaded.

Two consequences:

**Load thrash.** With candidates `[A, B, C]` and the default `max_loaded_models = 1` (`src/cpp/include/lemon_tray/tray_ui.h:110`), every message loads A, then B (evicting A), then C (evicting B), and then `apply_router_collection_dispatch` → `auto_load_model_if_needed` loads the routed candidate again. Four loads where one is needed, with the UI parked in `pre_flight` the whole time.

**Failure amplification.** If any candidate fails to load — backend not installed, `RouterResidencyConflictException` (409), NPU exclusivity — `ensureModelReady` throws and the turn aborts with "Error preparing model", even when the policy would never have routed there. A router is now only as loadable as its least loadable member.

That component fan-out is an omni behavior (the branch's own error string says "Omni-model ... references missing component"). Suggest short-circuiting it for routers, e.g. in `ensureModelReadyInternal`, treat `isRouterCollection(modelInfo)` as a no-op / defer to the server, which already downloads and loads the routed candidate itself.

Two smaller follow-ons from the same overloading, non-blocking:

- `LLMChatPanel.tsx:503` — `buildErrorMessage` resolves `modelsData[chatModelName].recipe` into `systemInfo.recipes[...]`. `collection.router` isn't a hardware recipe, so backend-install failures lose their actionable help text and setup-page hand-off.
- `LLMChatPanel.tsx:800` — `isNewModelLoad = currentLoadedModel !== chatModelName` compares the router name against `/health`'s `model_loaded`, which reports the routed *component*. They never converge, so every turn fires a spurious `modelLoadEnd` and `currentLoadedModel` flaps on each 5s poll.

meghsat:
Thanks for pointing the fan-out bug @ramkrishna2910 
Fixed it in the latest commit: 
1. For routers, ensureModelReady still walks each candidate for backend install and download (keeps Download Manager progress and pull flows visible in the UI), but skips POST /load via a new internal skipLoad flag. Loading is deferred to the server's dispatch loading only the routed candidate.
2. /health reports the routed candidate, never the router name, so currentLoadedModel !== chatModelName was true on every turn and fired modelLoadEnd each message. "Loaded" now accepts the router name or any of its candidates.
3. buildErrorMessage looked up collection.router in systemInfo.recipes, which always missed since it isn't a hardware recipe. Backend setup hints now resolve from the candidates' recipes.

ramkrishna2910 (approved):
Verified all three against the code on `b92de20c6` — thanks for the quick turnaround. `tsc --noEmit` on `src/app` is clean on the branch.

- **Preflight fan-out**: `skipLoad` propagates into each component and short-circuits before Step 6's `POST /load`. The early `return` is inside the `try`, so `finally { visited.delete(...) }` still runs. `ModelManager.handleLoadModel` passes *component* names, so `isRouterCollection` is false there and the explicit Load button still loads — no collateral.
- **`buildErrorMessage`**: resolving the action from the candidates' recipes is right.
- **`isChatModelLoaded`**: all four comparison sites go through it now. `currentLoadedModel` still alternates between the router name and the routed candidate on the 5s poll, but that's inert — `fetchLoadedModel` skips `setSelectedModel` for collections via `keepCollectionSelection`, and `isChatModelLoaded` is its only remaining consumer.

Approving. Three residuals for a follow-up, none blocking:

1. **Candidate install failures can still abort a router turn.** Only `/load` was lifted out; Step 3 still runs per candidate, and `ensureBackendForRecipe` throws on `state === 'unsupported'` and on `installable` when the action carries a guide URL. A router spanning devices — an NPU/FastFlowLM candidate alongside a llama.cpp one, which is rather the point of routing — then fails *every* message on a machine where the NPU backend is unsupported, even though the policy would have routed to the working candidate. Treating per-candidate install failures as non-fatal and erroring only when all candidates are unusable would close this out.
2. **Preflight latency scales with candidate count.** Step 3 always calls `fetchSystemInfoData()`, so a 3-candidate router costs roughly six sequential round-trips (`/health` + `/system-info` per candidate) before the request goes out, on every message.
3. **No load indicator when the router switches candidates.** `isNewModelLoad` is false whenever *any* candidate is loaded, so a switch from A to B shows no spinner while B loads. Strictly better than the old spurious-event behavior; the server's route-decision SSE frame could drive this precisely if it's worth tightening.

Trivial: the `ensureModelReady` doc comment still lists steps 1-6 with no mention of `skipLoad`.

meghsat:
Thanks @ramkrishna2910 Fixed the 2/3 residuals
1. The candidate loop now wraps each candidate in a try/catch when in router mode: a candidate whose backend is unsupported (e.g. the NPU/FastFlowLM one on a machine without an NPU) is recorded as a failure and skipped, and preflight only throws if every candidate is unusable.
2. The collection branch now creates a shared snapshot passed to every candidate: one /health call builds a loaded-model set consulted by each candidate's Step 1, and one lazy /system-info fetch serves every candidate's Step 3. So a 3-candidate router costs 2 round-trips per message instead of ~6 (and 0 extra when all candidates are already loaded)
3. Pausing this for now as GUI2 is frozen

PR #3181, opened 2026-08-16
feat(router): add min_total_chars / max_total_chars whole-conversation length conditions (#2958)

### LINKED ISSUE #2958: [Router] Whole-conversation length condition (total_chars)
## Problem

`min_chars` / `max_chars` measure the UTF-8 byte length of the **routing input only**, which is the last user message: `build_route_context` takes the last `role == "user"` entry's text (`src/cpp/server/routing_classifier_services.cpp:383-390`) and sets `ctx.params.chars = ctx.input.size()` (`routing_classifier_services.cpp:451`); `CharsCondition` compares bounds against that single value (`src/cpp/server/routing_policy.cpp:875-894`). The unit is frozen for v1 as "input length in UTF-8 bytes" (`src/cpp/resources/schemas/README.md`, frozen-semantics table).

Consequence: "route long conversations to the big-context candidate" is inexpressible. A conversation whose accumulated history is 50k+ tokens routes by its final turn — a 5-byte "go on" matches the short-input rules and can land on a 4k-context local model, where the request then fails with `context_length_exceeded` (there is no preflight or fallback at dispatch; that is a separate issue). Conversation length is the single most decision-relevant feature for local-vs-cloud candidate selection, and no condition can see it.

## Proposed approach

Add two additive leaf condition keys, `min_total_chars` / `max_total_chars`: inclusive bounds on the **total UTF-8 byte length of all text content in the request**, with exactly the char-bounds semantics authors already know (greedy independent rules, inclusive comparison, bytes not code points).

What counts, per request form (mirroring the existing extraction paths in `build_route_context`, `routing_classifier_services.cpp:361-463`):

- Chat `messages`: sum of `collect_text_from_content(msg["content"])` byte lengths over **all** messages, every role (system/user/assistant/tool). All roles count because the value is a proxy for prompt size — the whole array is what gets prefilled. Non-text parts (images, audio) contribute nothing; this scans the same array the `has_images` detection already walks (`:377-382`).
- Legacy `prompt` (string or array-of-strings): identical to the routing input (`:391-401`), so `total_chars == chars`.
- Responses `input`: bare string — identical to `chars`; array — sum of text parts across **all** items, role-tagged or bare (`:402-449`).

Naming note: `total_chars` keeps the frozen `chars` unit (UTF-8 bytes) and does not touch the existing keys, so this is a compatible addition within schema major 1 per the compatibility rules in `src/cpp/resources/schemas/README.md`. Token-based bounds remain explicitly deferred (`src/cpp/include/lemon/routing_policy.h:57-59` — "token-based length is deferred to a future min_tokens/max_tokens, never a redefinition of chars"); byte totals are the tokenizer-free v1 proxy, with a documented chars-per-token rule of thumb (~4) for authors sizing against context windows.

Example policy fragment:

```json
{"id": "long-conversation", "match": {"min_total_chars": 60000}, "route_to": "cloud-model"}
```

## Implementation sketch

- `src/cpp/include/lemon/routing_policy.h` — add `std::size_t total_chars = 0;` to `RouteContext::Params` (`:60-65`) with a doc comment defining the unit and scope.
- `src/cpp/server/routing_classifier_services.cpp` — accumulate the total during the existing walks in `build_route_context`; no new passes needed for `messages` (fold into the loop at `:377-382` or a sibling loop), and reuse the existing per-form extraction for `prompt` / `input`.
- `src/cpp/server/routing_policy.cpp` — either generalize `CharsCondition` (`:876-894`) with a source selector (last-turn vs total) or add a sibling `TotalCharsCondition`; trace condition names are `min_total_chars` / `max_total_chars` via the existing `trace_leaf` call pattern (`:887`).
- `src/cpp/server/routing_policy_parser.cpp` — extend leaf validation (the `{"min_chars", "max_chars"}` loop at `:303-311` generalizes to the four keys: non-negative integer) and add both keys to `routing_match_expr_keys()` (`:550-556`).
- Schema + docs: additive keys in the major-1 route-policy schema; new row in the frozen-semantics table in `src/cpp/resources/schemas/README.md`; condition table in `docs/dev/router-policy.md`.

## Acceptance criteria

- Unit tests (`test/cpp/`): totals across roles and forms (messages multi-role, prompt array, Responses input array with role-tagged and bare items), multi-byte UTF-8 counting, parser acceptance/rejection (negative, non-integer).
- Conformance corpus: new group `test/conformance/routing/1/l1_conditions_total_char_bounds/` locking: inclusive `min`/`max` boundaries; bytes-not-code-points; long-history-short-last-turn matches `min_total_chars` but not `min_chars` (the motivating case); `prompt`-form equivalence (`total == chars`); trace emission for both condition names.
- `test/server_router.py` e2e: policy routing on `min_total_chars`; a multi-turn history with a short final message routes to the long-context candidate; the same final message with a short history falls through to default.

## Out of scope

- `min_tokens` / `max_tokens` (deferred per the header comment; a future issue once a tokenizer-backed estimate exists).
- Capacity-aware candidate filtering / overflow fallback at dispatch (separate issue; this condition is the policy-level building block it composes with).
- Counting non-text content (images/audio) toward the total.

## References

- vLLM Semantic Router `context` signal (routes by estimated token demand, `min_tokens`/`max_tokens`) — prior art for conversation-length routing; their docs define no overflow behavior, which Lemonade handles in the companion dispatch issue.
- Frozen v1 semantics and additive-evolution rules: `src/cpp/resources/schemas/README.md`; corpus conventions: `test/conformance/routing/README.md`.
- Related issues: #2714 (aggregate `max_context_window` metadata for collections), #2389 (tracker).

Size: S

meghsat:
Closes #2958.

Adds min_total_chars / max_total_chars - inclusive UTF-8-byte bounds over all text in the request, so a policy can route on conversation size instead of just the last turn. A 50k-byte history ending in "go on" now matches a min_total_chars rule where min_chars sees only 5 bytes.

Semantics (additive, schema major 1; frozen chars unit unchanged):
1. Chat messages: sum over every role (system/user/assistant/tool)
2. Legacy prompt (string or array): equals chars (no history)
3. Responses input: string equals chars; array sums all items, role-tagged or bare
4. Non-text parts (images/audio) contribute nothing; token-based bounds stay deferred to a future min_tokens/max_tokens

Changes (backend only):

1. routing_policy.h / routing_policy.cpp: total_chars in RouteContext::Params; CharsCondition generalized with a source selector; trace emits the authored op name
2. routing_classifier_services.cpp: totals accumulated in the existing build_route_context walks; the messages image scan no longer breaks at the first hit (latched flag) so totalling shares the pass
3. routing_policy_parser.cpp: validation + match-expr key registry for both keys
4. server.cpp: /routing/validate sets total_chars = prompt.size() so Router Builder test prompts can exercise the new conditions
5. Schema + frozen-semantics table + docs/dev/router-policy.md (includes the ~4 bytes/token sizing hint)

GUI: none in this PR. The new leaves are not yet exposed in the Router Builder's Quick editor - policies using them can be authored via raw JSON / POST /pull today; Quick-editor support can follow as a small UI change. The router chat bypass discovered while E2E-testing this feature (chat requests to a collection.router resolved to the first component, silently skipping routing) is fixed separately in #3180.

Testing:

1. Unit: multi-role/UTF-8/all-forms totals, boundary + trace-name coverage, parser rejections (test_routing_classifier_services, test_routing_policy_deterministic, test_routing_policy_parser)
2. Conformance: new 1/l1_conditions_total_char_bounds/ group - inclusive boundaries, bytes-not-code-points, long-history-short-last-turn (motivating case), prompt-form total == chars, trace emission for both ops
3. E2E: server_router.py multi-turn tests (long history + short turn routes to the capable candidate; short history falls through to default); server_endpoints.py covers /routing/validate
4. Manual: verified end-to-end against a live server (with #3180 applied for the chat path) - routed/default/multi-turn cases all return the correct x-lemonade-route header and candidate
5. Out of scope (per issue): min_tokens/max_tokens, overflow fallback at dispatch, counting non-text content, Quick-editor support for the new leaves.


fl0rianr (approved):
LGTM

jeremyfowers:

**Attention level:** Routine — needs any 1 reviewer. Already on this PR: fl0rianr, ramkrishna2910.

Why this rung: a field or parameter on an existing route. The diff adds two additive keys (min_total_chars, max_total_chars) to the existing route-policy schema and their corresponding condition implementations — a field/parameter on an existing schema, which the table places on one-reviewer.

The description accurately maps the diff: the semantic rules for messages, prompt, and responses forms match the implementation in build_route_context; the five change categories (header, classifier, parser, server.cpp, schema/docs) are all present; testing coverage across unit, conformance, and e2e suites is correctly enumerated; out-of-scope items (min_tokens, overflow fallback, Quick-editor support) are correctly excluded.

</details>

The diff solves one problem: adding whole-conversation length conditions for routing policy. The changes span the header, classifier service, parser, server validate endpoint, schema, docs, and tests — all of which are necessary to ship the feature. The one CI-formatting fix in server_endpoints.py is a pre-commit cleanup needed for the PR to pass CI, not a separate concern.

Checked: Every changed file serves the single feature: header declares the field, classifier computes it, parser validates it, server.cpp exposes it for testing, schema documents it, docs explain it, and tests cover all paths. No independent concern exists that could be split out.

</details>

<details><summary>Alignment issues: <b>none found</b></summary>

Checked: Reviewed contribute.md Merging a Contribution and AI Contribution Policy sections; PR closes issue #2958 with a detailed description mapping code to docs, and all four AI policy requirements (understands code, reviewed full diff, no unrelated changes, tested locally) are evidenced by the PR body and commit history.

</details>

Checked: Checked docs/dev/router-policy.md for the new total_chars conditions table entry and explanatory prose, src/cpp/resources/schemas/README.md for the parameter row, and route_policy.schema.json for the new field definitions — all ship with the code in this PR per documentation.md.

</details>

Checked: C++ unit tests land in test/cpp/test_routing_classifier_services.cpp (5 new test functions), test/cpp/test_routing_policy_deterministic.cpp (4 new functions + rejections + trace names), and test/cpp/test_routing_policy_parser.cpp (2 validation cases); conformance tests in test/conformance/routing/1/l1_conditions_total_char_bounds/ with 10 cases; Python integration in test/server_router.py (test_605, test_606) and test/server_endpoints.py (test_021zta); all in existing suites already wired to CI.

</details>

<details><summary>Breaking changes: <b>none found</b></summary>

Checked: min_total_chars and max_total_chars are additive schema fields; no existing policy, endpoint, CLI flag, config format, or GUI control is changed or removed.

</details>

_Reviewed at head `7388065` by repo-manager (tier: reviewers). Regenerate with `repo-manager review-pr 3181`._

jeremyfowers:
bot is happy :)

ramkrishna2910 (approved):
Approving. I reviewed this against the frozen v1 semantics rather than just the diff, and it is the model for how to extend the routing vocabulary — new ops, correct unit, precise schema descriptions, a lock refreshed in the same diff, and a conformance group that actually discriminates.

**No redefinition of `min_chars`/`max_chars`.** `CharsCondition` gains a `Source{Input,Total}` enum defaulting to `Input`, and both existing factories pass `Source::Input` explicitly. The net deletions in the header are comment rewrites and column realignment only; the deletions in `routing_classifier_services.cpp` are the two `break` statements, replaced by a latched `if (!ctx.params.has_images && ...)` so the totalling pass can share the walk — `has_images` is still true iff *any* message carries an image. `ctx.params.chars` is still assigned exactly once, unchanged. `l1_conditions_char_bounds/` is untouched and still passes.

**Conformance work is the right standard.** Ten cases, ten matrix rows, 1:1, auto-discovered by the runner — and they discriminate rather than merely pass: `long-history-short-last-turn` flips `matched_rule` under an input-only implementation, and `total_chars-counts-every-role` falls to default under a user-turns-only one.

Two follow-ups. Neither blocks the merge, and both are still cheap because `route_policy.schema.json` is `"released": false`:

1. **Tighten the prose rows.** `schemas/README.md` and `docs/dev/router-policy.md` say "all text in the request", which overstates what the code does — the top-level `tools[]` array and `tool_calls` arguments contribute zero, since only message `content` is summed. A 20 KB tool-schema array plus a two-line chat totals ~40 bytes, so a `max_total_chars` sized as a prefill budget would admit a request that prefills 20 KB. The schema field description is already precise ("the sum over every message of every role"); it's the two tables that are loose. "all **message content** text" fixes it. Worth doing before v1 flips to `released: true`.

2. **One corpus gap.** "Non-text content (images, audio) contributes nothing" is explicitly frozen in both the schema and the docs, but none of the ten cases carry an image part. It's unit-tested, though the corpus is what the README frames as the cross-version guarantee. One case — a message with a text part plus an image part, asserting the total equals the text bytes alone — closes it.

Minor: the header comment "They differ only for multi-turn requests" isn't quite true in the Responses bare-parts form, where `chars` joins parts with a newline while `total_chars` sums them without a separator, so `chars` can exceed `total_chars`. The behavior is deliberate and matches the frozen text; only the word "only" is wrong. Also, `collect_text_from_content(content).size()` materializes and discards a copy of every message's text purely to read a length — a size-only overload would be a clean cleanup, though it's noise next to the JSON DOM parse that already happened.

On `/v1/routing/validate`: thanks for remembering the `total_chars` line. That endpoint hand-assembles its `RouteContext`, so **every** new `Params` field silently defaults unless an author adds a line — this PR got it right and #3169 currently does not. I'll file that separately; having the handler synthesize a body and call `build_route_context` would kill the whole class.

PR #3182, opened 2026-08-16
docs: add Backend Engine Admission Policy

Geramy:
## Summary

Adds `docs/dev/backend-admission-policy.md`, which defines how a third-party inference engine gets admitted as a lemonade backend and at what tier:

- **Tiers** — Experimental vs. Supported; everything enters at Experimental or below.
- **Value proposition (gate zero)** — coverage matrix and qualifying paths a newcomer must meet before any engine review or scoring.
- **Engine review and scoring** — what we evaluate in the upstream project.
- **Approval quorums** — who signs off on admission, promotion to Supported, replacement, and demotion/removal.
- **Lifecycle** — absorbed-claim re-qualification and the demotion path.

Download/execute hardening is out of scope for this doc and applies to every backend at every tier.

## Doc wiring

- `mkdocs.yml` — nav entry "Backend Admission Policy" placed after "Adding a Backend".
- `docs/dev/contribute.md` — paragraph under "Adding a Backend" pointing prospective contributors to the policy.
- `docs/dev/adding-a-backend.md` — callout distinguishing wiring mechanics (that page) from admission/tiering (this policy).

## Test plan

- [ ] Docs-only change; verify `mkdocs` nav renders and links resolve.

### DIFF
--- a/docs/dev/adding-a-backend.md
+++ b/docs/dev/adding-a-backend.md
@@ -19,6 +19,8 @@ That covers the backend's wiring. You still write an integration test ([Testing]
 
 Everything for one backend lives in `lemon::backends::<stem>`. The descriptor is header-only so it links into **both** the `lemonade` CLI and `lemond`; the server class and `create()` are server-only (compiled into `lemond`).
 
+> **Wrapping a new third-party engine?** This page covers the mechanics of wiring a backend in. Whether the engine gets admitted at all — and at which tier — is governed by the [Backend Admission Policy](./backend-admission-policy.md). Read that first; the value-proposition gate is the most common reason a proposal is declined.
+
 ## The descriptor — `<stem>/<stem>.h`
 
 Plain data. The single object the registry, CLI, `/system-info`, and docs all read.
--- a/docs/dev/backend-admission-policy.md
+++ b/docs/dev/backend-admission-policy.md
@@ -1,0 +1,281 @@
+# Backend Engine Admission Policy
+
+Decides whether a third-party inference engine gets integrated as a lemonade backend, and at what tier.
+
+This is about admitting the **upstream project**. Download/execute hardening is a separate doc and applies to every backend at every tier.
+
+---
+
+## 1. Tiers
+
+| Tier | Meaning | Shipped how |
+|---|---|---|
+| **Experimental** | Vetted, not trusted for defaults | In-tree, off by default, marked experimental |
+| **Supported** | We stand behind it | Default-installable, in CI, release-blocking |
+
+Everything enters at Experimental or below. Nothing goes straight to Supported, including vendor backends.
+
+---
+
+## 2. Value proposition (gate zero)
+
+Answered before sponsorship, engine review, or scoring. Cheapest question to answer, most common reason to decline.
+
+Every backend costs us forever: CI time, per-release validation, conformance runs, triage, another download path to secure, one more option in front of a user who just wants to be told what to pick. Burden of proof is on the newcomer. "It works" is not a value proposition.
+
+### 2.1 Coverage matrix
+
+Describe the backend as the cells it serves:
+
+- **Device** — CPU, dGPU by API (Vulkan / ROCm / CUDA / Metal), iGPU, NPU
+- **Format / architecture** — GGUF quant families, ONNX, safetensors; dense, MoE, hybrid/SSM, vision, embedding
+- **Capability** — streaming, grammars, speculative decoding, batching, LoRA, tool calling, long context
+
+Compare against the union of all Supported backends.
+
+### 2.2 Qualifying paths
+
+Meet at least one.
+
+| Path | Threshold |
+|---|---|
+| **Coverage** | Uniquely serves ≥ 1 device class or ≥ 1 architecture family no Supported backend serves |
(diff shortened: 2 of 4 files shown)


Geramy:
I have added all the relevant people I know of for this policy, I had written it weeks ago as @ramkrishna2910 asked. Please let me know what you think and what changes are needed.

docs/dev/backend-admission-policy.md
```diff
@@ -0,0 +1,281 @@
...
+
+This is about admitting the **upstream project**. Download/execute hardening is a separate doc and applies to every backend at every tier.
+
+---
+
+## 1. Tiers
+
+| Tier | Meaning | Shipped how |
+|---|---|---|
+| **Experimental** | Vetted, not trusted for defaults | In-tree, off by default, marked experimental |
```
jeremyfowers, line 13: what does "off by default" mean here?

docs/dev/backend-admission-policy.md
```diff
@@ -0,0 +1,281 @@
...
+| **Supported** | We stand behind it | Default-installable, in CI, release-blocking |
```
jeremyfowers, line 14: we currently have lots of experimental backends in CI. I wouldn't merge anything to the project that wasnt in ci.

docs/dev/adding-a-backend.md
```diff
@@ -19,6 +19,8 @@ That covers the backend's wiring. You still write an integration test ([Testing]
 
 Everything for one backend lives in `lemon::backends::<stem>`. The descriptor is header-only so it links into **both** the `lemonade` CLI and `lemond`; the server class and `create()` are server-only (compiled into `lemond`).
 
+> **Wrapping a new third-party engine?** This page covers the mechanics of wiring a backend in. Whether the engine gets admitted at all — and at which tier — is governed by the [Backend Admission Policy](./backend-admission-policy.md). Read that first; the value-proposition gate is the most common reason a proposal is declined.
```
jeremyfowers, line 22: The markdown diffs in this PR violate docs/dev/documentation.md in a bunch of ways, please address.

docs/dev/backend-admission-policy.md
```diff
@@ -0,0 +1,281 @@
...
+
+Everything enters at Experimental or below. Nothing goes straight to Supported, including vendor backends.
```
jeremyfowers, line 16: > Everything enters at Experimental or below

below what?

jeremyfowers, line 16: > Nothing goes straight to Supported, including vendor backends.

Why?

jeremyfowers: My take on experimental is "if you use this in production there is nobody on call to support you." Vendor backends do have someone (the vendor) on call to support it.

docs/dev/backend-admission-policy.md
Geramy: Meaning the backend never is "turned on by default" basically its not a on by default engine such as llama.cpp.

docs/dev/backend-admission-policy.md
Geramy: That was a prior edit and had 3 levels, experimental was level 2 and in-tree or trusted was level 3 basically. I will correct it.

Geramy: > > Nothing goes straight to Supported, including vendor backends.
> 
> Why?

I believe its important to start everything as experimental period, that way people know it was introduced recently and has not been extremely tested by our base. The existing backends we have llama.cpp, vLLM etc... are all "Supported" already by this definition.

Geramy: > My take on experimental is "if you use this in production there is nobody on call to support you." Vendor backends do have someone (the vendor) on call to support it.

im going to think about this one, I do see the point on that, but some backends are just experimental you know. I think a backend gets upgraded, kind of like earning respect or street credit.

docs/dev/backend-admission-policy.md
jeremyfowers: What is the technical definition of "on" though?

docs/dev/backend-admission-policy.md
jeremyfowers: >  The existing backends we have llama.cpp, vLLM etc... are all "Supported" already by this definition.

vLLM is still marked experimental for one good reason though: there is no lead maintainer. If you have a problem you can go pound sand, there's no support guarantee.

kenvandine: I do think some backends should be able to bypass experimental. For example, for a vendor-supported backend, we will likely want it to be clearly and officially supported because we know a vendor will ensure bugs get fixed; they have a stake in its success.

jeremyfowers (changes requested):
Will continue reviewing after the current set of comments have been addressed.

docs/dev/backend-admission-policy.md
Geramy: Yeah I can see the point there actually, it does make sense in some cases to fast track past experimental. But even like llama.cpp wasn't perfect on day-1 and experimental doesn't just express the vendor is experimental, it also can be us who still do not have a perfect implementation of the new backend, or limited feature support even.

jeremyfowers:
IMO this has been subsumed by https://github.com/lemonade-sdk/lemonade/pull/3521

If you're still interested, I suggest opening a new PR that adds backend-specific guidance to spec-driven-dev.md.

Issue #3161, opened 2026-08-16
GUI3: client code must not use localStorage any time there is an equivalent server feature


jeremyfowers:
It will be very confusing for users, and a breaking change coming from GUI2, if any server-side features for storing persistent data are replicated into localStorage.

Examples include:
- [ ] lemond config already has `llamacpp.args` and `llamacpp.rocm_args`. GUI3 must not also introduce localStorage backend args.
- [ ] lemond's recipe_options.json already stores per-model options. GUI3 must not also introduce localStorage model options.
- [ ] lemonade's user_models.json already stores custom model definitions. GUI3 must not also introduce localStorage custom models.
- [ ] are there any more instances of this?

jeremyfowers:
> are there any more instances of this?

Yes. Here is every remaining `localStorage` call site in `src/app/src` at `3d2d061` (head of #3183), split by whether lemond already owns the state.

### Still duplicating server-owned state

| localStorage key | Where | lemond already has |
|---|---|---|
| `pinned_models` | `features/modelSettings/globalModelSettings.ts:91` | `POST /internal/pin`, the `pinned` load param (`server.cpp:468`), and `pinned_models` in `/stats`. Clearest remaining case — `server.cpp:3044` explicitly documents `pinned` as live-process state that `/v1/load` and `/internal/pin` own, and the model-options endpoint refuses to report or accept it. |
| `global_model_settings` → `automaticModelUpdates`, `lastAutomaticUpdateAt` | `globalModelSettings.ts:73` | `auto_check_model_updates` in config.json |
| `global_model_settings` → `loadingPolicy`, `evictionPolicy`, `resourceBudgetMode/Gb`, `autoEvictOnPressure`, `protectPinnedModels` | same | `max_loaded_models` in config.json, plus the Router's own LRU eviction and NPU-exclusivity logic. These are GUI-only knobs describing behavior lemond actually implements, so two clients can disagree about a policy only one of them enforces. |
| `lemonade_download_manager_items_v1`, `lemonade_download_manager_dismissed_v1` | `features/downloadManager/downloadStore.ts:494,523` | `GET /api/v1/downloads` + `POST /api/v1/downloads/control` are authoritative. The file's own comments already note that stale cached rows can block a model. |
| `model_filter_custom_tags_v1` | `components/ModelNavRail.tsx:74` | `labels` is part of the server model definition (`model_manager.cpp:372`) and is accepted by `models/register` |
[...]

PR #3183, opened 2026-08-17
fix(gui3): move server-owned persistent state out of localStorage

fl0rianr:
This PR addressed the demand of #3161 to remove server-owned states and data out of the local browser storage.
Build on top of #3117 as stacked PR.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


jeremyfowers (commented):
Direction is right — backend args, model options, custom models and routers all go through lemond now. Two things to fix before merge, inline.

src/app/src/components/BackendManager.tsx
```diff
@@ -225,6 +221,90 @@ const BACKEND_VIEW_FILTERS: Array<[BackendViewFilter, string, string, IconName]>
...
+    : Object.prototype.hasOwnProperty.call(section, 'args') ? 'args' : null;
+  if (!field) return null;
+
+  const value = section[field];
+  const effectiveValue = field !== 'args'
+    && (typeof value !== 'string' || value.trim() === '')
+    && typeof section.args === 'string'
+    ? section.args
+    : value;
+  return { flatKey: `${sectionName}_${field}`, effectiveValue };
```
jeremyfowers: **Blocker:** the global-`args` fallback is right for recipes that only have `args`, but wrong for a section that has *some* `<backend>_args` fields and not the one being asked for — and there it corrupts neighbouring backends.

`defaults.json` has `llamacpp.{cpu,rocm,vulkan}_args` but no `cuda_args` or `metal_args` (same shape in `whispercpp`, which only has `cpu_args`/`npu_args`, and `sdcpp`, which has no `metal_args`). So on a CUDA box or on macOS:

1. Saving args for `llamacpp:cuda` finds no `cuda_args`, falls back to `field = 'args'`, and `flatKey` becomes the section-**global** `llamacpp_args`.
2. `backendTuningsFromServerConfig` then evaluates `llamacpp:vulkan`, finds `vulkan_args` empty, and the `effectiveValue` fallback on line 268 substitutes `section.args` — so the CUDA args now render as user args on the cpu/rocm/vulkan rows too.
3. Worse, **Clear** becomes a silent no-op that reports success: clearing `llamacpp:vulkan` writes `vulkan_args=""`, the fallback immediately re-reads the still-set global, the row keeps showing the old args, and the toast says `Cleared backend arguments for llamacpp:vulkan`.

Suggest gating the fallback on the section having no per-backend `*_args` fields at all (`flm`, `vllm`, `trellis` — where the global genuinely is the target), and returning `null` otherwise so `saveBackendArgsToServer` raises its existing `No server config args field exists for ...` error instead of writing the global.

src/app/src/modelConfiguration.ts
```diff
@@ -281,211 +278,56 @@ function sanitizeBackendTuning(raw: Partial<BackendTuning> | null | undefined):
...
-
-const LEGACY_DEFAULT_RECORD_SUFFIX_ORDER = ['default', 's_default', 's__default'] as const;
-const LEGACY_DEFAULT_RECORD_SUFFIXES = new Set<string>(LEGACY_DEFAULT_RECORD_SUFFIX_ORDER);
-const LEGACY_RECORD_SEPARATOR = '@@';
-
-function compareStorageKeys(left: string, right: string): number {
-  return left < right ? -1 : left > right ? 1 : 0;
-}
-
-function migrateLegacyModelConfigurationStorage(): boolean {
```
jeremyfowers, line 357: **Blocker:** `src/app/tests/model-configuration-migration.runtime.cjs` still asserts this deleted `migrateLegacyModelConfigurationStorage` behavior together with the `recipe_options` half of `model_tunings`, and now fails.

Verified: it passes at the base commit `bbf2effb` and fails at head `3d2d061`:

```
AssertionError [ERR_ASSERTION] ... at tests/model-configuration-migration.runtime.cjs:76
  actual: undefined
  expected: { source: 'default', recipe_options: { ctx_size: 8192 } }
```

It is not wired into an npm script or a CI job so it will not redden the build, but it should be deleted alongside the code it covers (or rewritten against the new server-owned options path) rather than left as a test that cannot pass.

jeremyfowers:
BTW it looks like this PR doesn't comprehensively close #3161, see https://github.com/lemonade-sdk/lemonade/issues/3161#issuecomment-5318716185

edit: let's merge this PR ASAP by addressing the 2 blockers, and address the other opens in a follow-up PR. In turn that will unblock #3082 ASAP

fl0rianr:
Addressed. Hope that's it.

jeremyfowers (approved):
LGTM

PR #3186, opened 2026-08-17
Switch to bundler module resolution for tsconfig

superm1:
## Summary

Fixes the FTBFS in Debian.
```
 @ ../app/src/renderer/index.tsx 10:30-46

ERROR in ../app/src/renderer/hooks/useSystem.tsx
[tsl] ERROR
      TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
  Visit https://aka.ms/ts6 for migration information.
 @ ../app/src/renderer/App.tsx 51:20-48
 @ ../app/src/renderer/index.tsx 10:30-46
```

Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1144577

## Scope

## Testing

_Testing details:_

I ran `sbuild` with this patch in place.  I haven't tested other distros.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/app/tsconfig.json
+++ b/src/app/tsconfig.json
@@ -1,17 +1,16 @@
 {
   "compilerOptions": {
     "target": "ES2020",
-    "module": "commonjs",
+    "module": "es2020",
     "lib": ["ES2020", "DOM", "DOM.Iterable"],
     "jsx": "react-jsx",
-    "outDir": "./dist",
     "rootDir": "./src",
     "strict": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "sourceMap": true,
     "declaration": true,
     "declarationMap": true
--- a/src/app/tsconfig.json
+++ b/src/app/tsconfig.json
@@ -1,16 +1,17 @@
 {
   "compilerOptions": {
     "target": "ES2020",
-    "module": "es2020",
+    "module": "commonjs",
     "lib": ["ES2020", "DOM", "DOM.Iterable"],
     "jsx": "react-jsx",
+    "outDir": "./dist",
     "rootDir": "./src",
     "strict": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
-    "moduleResolution": "bundler",
+    "moduleResolution": "node",
     "sourceMap": true,
     "declaration": true,
     "declarationMap": true
--- a/src/web-app/BuildWebApp.cmake
+++ b/src/web-app/BuildWebApp.cmake
@@ -65,6 +65,25 @@ _stage_dir("${WEB_APP_SOURCE_DIR}" "${STAGED_WEB_APP_DIR}")
 # the staged web-app directory, which now sits at staging/web-app/.
 set(WEB_APP_BUILD_SOURCE_DIR "${STAGED_WEB_APP_DIR}")
 
+# The committed tsconfigs use `moduleResolution: bundler`, correct for the
+# bundled TypeScript (>= 5.3). System TypeScript spans a range where no single
+# value works: Debian trixie ships 4.9 (no `bundler`) while unstable ships 7.0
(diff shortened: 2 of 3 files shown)


superm1:
Hmm, this patch passed and fixed the FTBFS in Debian.  I guess that means it's got an implicit dependency on the newer node-typescript version or something.

https://salsa.debian.org/debian/lemonade/-/pipelines/1151424

superm1:
This now works both in Debian and in CI.

https://salsa.debian.org/debian/lemonade/-/pipelines/1152495

fl0rianr (commented):
Looks good overall, and nice to see it passing both Debian and CI.

One concern: I don't think removing `"outDir": "./dist"` is related to the module resolution fix. With `declaration`, `declarationMap`, and `sourceMap` still enabled, running `tsc` directly can emit generated files next to the sources, especially with the web config including `../app/src/**/*`. I'd keep `outDir` in both tsconfigs.

Also worth confirming that the system build intentionally rewrites both `moduleResolution` and `module` to `node16`. Since the web package has no `"type": "module"`, that can change the emit to CommonJS. It clearly builds, so this may be fine, but I'd like to make sure that behavior is intentional.

Otherwise the approach looks good to me.

jeremyfowers:

**Ready for review** — the checks below passed. No reviewers are suggested: this PR already has the review contribute.md's rung asks for (fl0rianr).

**Attention level:** Routine — needs any 1 reviewer

Why this rung: a version bump or a config option. Build configuration change (tsconfig moduleResolution and CMake rewrite logic). One reviewer is sufficient for a config option change under the Dependencies & packaging row in the review table.

_Reviewed at head `f90af47` by repo-manager (tier: reviewers). Regenerate with `repo-manager review-pr 3186`._

jeremyfowers:
Bot is happy :)

superm1:
> Bot is happy :)

You should make him his own GitHub account I suggest

fl0rianr (commented):
Looks much better, and my previous concerns are addressed: outDir is restored and module is no longer rewritten.

There is still one blocker though: the current head fails the Ubuntu 26.04 PPA build on both x86_64 and ARM64 with:

TS5110: Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.

So transpileOnly does not avoid validation of this compiler option combination. I think the clean fix is to rewrite moduleResolution to node16 only for TypeScript < 5.0, where bundler is unavailable, and leave bundler unchanged for newer TypeScript versions.

So you see, even if the Bot is happy, I'm not - yet.

superm1:
Fwiw your feedback is what broke things. I need to find a balance.

fl0rianr:
> Fwiw your feedback is what broke things. I need to find a balance.

Ah, I see what happened. My concern about changing the emit to CommonJS still stands, but I missed that newer TypeScript doesn't allow module: es2020 together with moduleResolution: node16. Sorry about that.

I think there is a fairly clean way to handle both sides though: keep the committed es2020 + bundler config unchanged for TypeScript >= 5, and only rewrite staged moduleResolution to node for older system TypeScript that doesn't support bundler yet. module can then stay es2020 in both cases.

That should avoid the TS5110 failure without changing the webpack emit format.

fl0rianr (commented):
This looks good now, and the latest CI coverage gives me a lot more confidence. Nice to see the Ubuntu 24.04/25.10/26.04 builds, including ARM64, all passing.

One small cleanup suggestion: `outDir` is removed from both tsconfigs again. I don't think that's needed for this fix, and keeping it avoids direct `tsc` runs potentially emitting files next to the sources. I'd prefer to keep `outDir`, but I wouldn't consider that worth blocking the PR over if there's a reason for removing it.

The `node16` rewrite for system builds is also broader than the normal `bundler` configuration, but given that it is isolated to the staged system-package build and the distro matrix is now green, I'm comfortable with the approach.

From my side, no remaining blocker.

superm1:
Given we're essentially doing a sed on the files for system nodejs, let me try just a cmake change.  I'll do a test build here and in salsa.

superm1:
Looks good in Salsa.

https://salsa.debian.org/debian/lemonade/-/jobs/10212096

fl0rianr (approved):

PR #3188, opened 2026-08-17
fix: support mbedTLS 4 on Arch Linux

fl0rianr:
## Summary

Adds support for the system mbedTLS 4 package used by current Arch Linux and fixes the cross-host portability issue exposed by that change.

The change:
- updates bundled cpp-httplib from v0.47.0 to v0.51.0, which includes upstream mbedTLS 4 support
- installs the system mbedTLS development package on Linux through `setup.sh`
- adds `LEMONADE_PORTABLE_BUILD` as a single policy for runtime dependencies Lemonade can bundle itself
- keeps normal Linux distro/package builds system-linked so Arch genuinely exercises mbedTLS 4
- uses portable mode for cross-host/raw artifacts, Linux embeddable builds and Docker so they do not inherit build-host versions of curl, zstd, cpp-httplib, libwebsockets or mbedTLS
- enables the same portable dependency policy automatically on Windows and macOS

This avoids fixing the vLLM failure by making mbedTLS a special case. The actual issue is that artifacts executed outside their build host must not accidentally depend on build-host runtime library versions.

The bundled mbedTLS revision itself is unchanged.

## Scope

## Testing

_Testing details:_

The Linux distro CI continues to build against system libraries. It now explicitly verifies that:
- the system `mbedtls` pkg-config module is available
- the Arch job is using mbedTLS 4+
- both `lemond` and `lemonade` are dynamically linked to the system mbedTLS libraries

This prevents the bundled fallback from accidentally masking the Arch compatibility test.

Portable/cross-host builds use `LEMONADE_PORTABLE_BUILD=ON` and verify with `readelf` that the resulting binaries do not directly depend on system copies of Lemonade-managed runtime dependencies. This covers:
- the vLLM artifact built on Ubuntu and executed on the self-hosted ROCm runner
- Linux x64/ARM64 embeddable archives
- the raw Linux ARM64 cross-job artifact
- the Docker build/runtime boundary

The existing runtime smoke tests still execute the produced artifacts in their target environments.

## Documentation

The embeddable build documentation now uses `LEMONADE_PORTABLE_BUILD=ON` on Linux so local portable builds follow the same dependency policy as release CI.

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -87,7 +87,7 @@ jobs:
       - name: Install ccache
         run: |
           if command -v pacman >/dev/null 2>&1; then
-            pacman -Syu --needed --noconfirm ccache
+            pacman -Syu --needed --noconfirm ccache mbedtls
           elif command -v zypper >/dev/null 2>&1; then
             zypper install -y ccache
           elif command -v apt-get >/dev/null 2>&1; then
@@ -166,10 +166,22 @@ jobs:
 
           echo "=== lemond dependencies ==="
           ldd lemond
+          if ldd lemond | grep -q "not found"; then
+              echo "ERROR: lemond has unresolved shared-library dependencies"
+              exit 1
+          fi
+          if [ "${{ matrix.distro }}" = "Arch" ] && ! ldd lemond | grep -q 'libmbedtls\.so'; then
+              echo "ERROR: Arch lemond is not linked against the system mbedTLS library"
+              exit 1
+          fi
 
           echo ""
           echo "=== lemonade dependencies ==="
           ldd lemonade
+          if ldd lemonade | grep -q "not found"; then
+              echo "ERROR: lemonade has unresolved shared-library dependencies"
+              exit 1
+          fi
 
           echo ""
           echo "All dependencies resolved successfully on ${{ matrix.distro }}!"
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -166,22 +166,10 @@ jobs:
 
           echo "=== lemond dependencies ==="
           ldd lemond
-          if ldd lemond | grep -q "not found"; then
-              echo "ERROR: lemond has unresolved shared-library dependencies"
-              exit 1
-          fi
-          if [ "${{ matrix.distro }}" = "Arch" ] && ! ldd lemond | grep -q 'libmbedtls\.so'; then
-              echo "ERROR: Arch lemond is not linked against the system mbedTLS library"
-              exit 1
-          fi
 
           echo ""
           echo "=== lemonade dependencies ==="
           ldd lemonade
-          if ldd lemonade | grep -q "not found"; then
-              echo "ERROR: lemonade has unresolved shared-library dependencies"
-              exit 1
-          fi
 
           echo ""
           echo "All dependencies resolved successfully on ${{ matrix.distro }}!"
--- a/.github/workflows/linux_distro_builds.yml
+++ b/.github/workflows/linux_distro_builds.yml
@@ -165,13 +165,49 @@ jobs:
           cd build
 
           echo "=== lemond dependencies ==="
-          ldd lemond
+          LEMOND_DEPS="$(ldd lemond)"
+          echo "$LEMOND_DEPS"
 
           echo ""
           echo "=== lemonade dependencies ==="
-          ldd lemonade
+          LEMONADE_DEPS="$(ldd lemonade)"
+          echo "$LEMONADE_DEPS"
+
+          # Prove this job really exercises the system-mbedTLS path.
+          # Otherwise a bundled fallback could make #3188 look fixed.
+          if ! pkg-config --exists mbedtls; then
+              echo "ERROR: system mbedTLS pkg-config module is unavailable"
+              exit 1
+          fi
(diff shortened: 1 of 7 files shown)


sofiageo:
@fl0rianr I have the feeling the correct file to do this is in setup.sh in the libraries section.

fl0rianr:
Thanks @sofiageo, I agree and addressed it.

sofiageo (dismissed):

fl0rianr:
The previous minimal fix trail did show a architecture weakness:
https://github.com/lemonade-sdk/lemonade/actions/runs/32544285439/job/96960813048
We use builds created one one host for multiple host in our test field creating AIB conflicts.

DLEMONADE_PORTABLE_BUILD now defines if this build should bundle what it needs to avoid AIB conflicts or if it is valid to rely on the targets host packages.
mbedTLS is only the first lib showing this problem it was not the fully correct approach IMO.

jeremyfowers:
@sofiageo please review the final changeset if you would like this merged

sofiageo (approved):
It looks to fine me, and it compiles fine in Arch Linux. I just don't have an easy way to test other distros or the CI locally.

jeremyfowers:
> It looks to fine me, and it compiles fine in Arch Linux. I just don't have an easy way to test other distros or the CI locally.

Thanks! @superm1 can you provide review for this PR outside of the Arch context that @sofiageo reviewed?

.github/workflows/linux_distro_builds.yml
superm1, line 1: I suggest stacking this on https://github.com/lemonade-sdk/lemonade/pull/3524, or drop linux distro builds changes.

CMakeLists.txt
```diff
@@ -465,14 +480,16 @@ endif()
...
 # ============================================================
 if(NOT TARGET lemonade-digest-crypto)
-    find_package(PkgConfig QUIET)
-    if(PkgConfig_FOUND)
-        pkg_check_modules(MBEDTLS QUIET mbedtls mbedx509 mbedcrypto)
+    if(NOT LEMONADE_PORTABLE_BUILD)
+        find_package(PkgConfig QUIET)
+        if(PkgConfig_FOUND)
+            pkg_check_modules(MBEDTLS QUIET mbedtls mbedx509 mbedcrypto)
+        endif()
```
superm1, line 487: I'm confused.  Why can't we copy the system one in for a portable build and patch the rpath?  Save build time then.

setup.sh
```diff
@@ -174,6 +174,11 @@ if command_exists pkg-config; then
 
     # Check for required libraries using pkg-config
     libs_to_check=("libcurl" "openssl" "zlib" "libsystemd" "libdrm" "libcap" "libwebsockets")
+    # Linux distro/package builds intentionally exercise system mbedTLS.
+    # Portable artifacts select bundled runtime dependencies at CMake configure time.
+    if [ "$OS" = "linux" ]; then
+        libs_to_check+=("mbedtls")
+    fi
```
superm1, line 181: I don't like this change being in `setup.sh`.  That's what we use for regular developers too and now it's going to be different policy.  If we are trying to test something in CI only - make it CI only change.

docs/embeddable/building.md
```diff
@@ -14,6 +14,8 @@ Contents:
 
 The `embeddable` CMake target builds the server, CLI, and required resource files, then packages them into a single archive. The [release workflow](https://github.com/lemonade-sdk/lemonade/blob/main/.github/workflows/cpp_server_build_test_release.yml) uses this target to produce the embeddable archives.
 
+On Windows and macOS, portable dependency mode is enabled automatically. On Linux, use `-DLEMONADE_PORTABLE_BUILD=ON` when producing an archive or raw binary that will run outside the build host. This selects bundled copies of runtime dependencies that Lemonade manages itself, matching release CI.
```
superm1, line 17: To me this is begging a new preset.  We could have an embedded preset and automatically set the define and turn off the web app

CMakeLists.txt
```diff
@@ -217,18 +217,31 @@ set(DOWNLOAD_ZSTD_VERSION "1.5.7")
...
 
+option(
+    LEMONADE_PORTABLE_BUILD
+    "Prefer bundled runtime dependencies for artifacts that run outside the build host"
+    OFF
+)
+
+# Windows and macOS release artifacts must not inherit developer-machine
+# package-manager libraries. Linux distro/package builds stay system-linked
+# by default and portable Linux artifacts opt in explicitly.
```
superm1, line 228: I think this the wrong surface area.  Why wouldn't we want to use system wide on windows and mac too if they're there?

fl0rianr:
@superm1 Please tell me if we can keep it this simpel. At the start it was just one line - then @sofiageo asked to go with setup.sh while you said don't do it. One way or the other, I need guidance since one has to win, I prefer the tiny version tbh.

sofiageo:
> @superm1 Please tell me if we can keep it this simpel. At the start it was just one line - then @sofiageo asked to go with setup.sh while you said don't do it. One way or the other, I need guidance since one has to win, I prefer the tiny version tbh.

I definitely don't care about winning :) So go ahead and do as @superm1 suggested. I only suggested for the change to be on setup.sh because it is also used by the CI and it would be better to have all dependencies in one place. Arch Linux has the dependency available so in my mind it is better to use the official packages than having to build them. `libwebsockets` is also one of those dependencies.

PR #3195, opened 2026-08-17
Add missing handler for lemonade:// URLs (Closes: #1144656)

superm1:
Thanks Peter Marschall.

## Summary
Make lemonade:// urls work.

Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1144656

## Scope

## Testing

- [ ] I described the testing performed below.

Tested by the submitter (peter Marschall)

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

- [ ] I checked for hallucinated APIs, unrelated changes, and incorrect assumptions.


fl0rianr (commented):
That's a HUGE change! 
Still please let me note:
 This registers lemonade-web-app as the handler for lemonade:// URLs, but Exec=lemonade-web-app does not pass the actual URL to the process. Should this be `Exec=lemonade-web-app %u`, with the app handling that URI on cold start as well? That would also match the approach used for the same scheme in #1957. Otherwise the desktop file can claim the handler without the app receiving the URL that was opened.

superm1:
Good suggestion, thanks

fl0rianr (commented):
Thanks for adding %u. One remaining issue: data/lemonade-web-app currently ignores its arguments and always opens the fixed /lemonade URL, so the actual lemonade://... URI is still dropped. For example, lemonade://open?view=logs&model=foo will only launch the base web app without preserving the deep-link payload. Could the launcher consume $1 and forward/translate the URI to the web UI, with an end-to-end xdg-open 'lemonade://open?view=logs' test? Otherwise this registers the scheme and launches the app, but doesn't actually handle the URL.

fl0rianr (commented):
Thanks for forwarding the URI. I think there's still one issue with the translation: lemonade://open?view=logs becomes /lemonade/#/open?view=logs, but the web app reads view from window.location.search, not from the hash.

With this URL window.location.search is empty, so view=logs won't open the logs view. Could we preserve the deep-link query as the actual HTTP query, e.g. /lemonade?view=logs, or explicitly add hash parsing on the web side? An end-to-end xdg-open lemonade://open?view=logs check would be useful here.

fl0rianr (commented):
Thanks and sorry for blocking here The new translation itself looks right now. One leftover block still undoes it though: after setting URL to /lemonade?view=logs#/open, the old # Handle lemonade:// URI passed via desktop file %u block runs immediately afterwards and overwrites it with /lemonade/#/open?view=logs again. So the browser still receives the old broken URL. I think that second block can simply be removed.

fl0rianr (approved):
LGTM, thanks!

PR #3197, opened 2026-08-17
Fail startup when one resolved address cannot bind

### LINKED ISSUE #2917: lemond serves IPv6-only when the IPv4 bind fails, and reports itself healthy
## Summary

When `lemond`'s IPv4 bind fails, it logs an error, binds IPv6 only, and keeps running. The process looks healthy — `localhost` health checks pass on macOS, where `localhost` resolves to `::1` first — while every client that connects to `127.0.0.1` gets `ECONNREFUSED`.

## Observed

```
[Info]  (Server) Host resolution: IPv4=127.0.0.1, IPv6=::1
[Info]  (Server) Binding IPv4 HTTP server to 127.0.0.1:13305...
[Error] (Server) Failed to bind IPv4 HTTP server to 127.0.0.1:13305
[Info]  (Server) Binding IPv6 HTTP server to [::1]:13305...
[Info]  (Server) IPv6 HTTP server listening on [::1]:13305
```

Startup then continues normally. Nothing downstream reports degraded state: `/api/v1/health` answers over IPv6, the UDP beacon broadcasts, and the log gives no further hint.

Reproduced on macos-latest in runs [31046661982](https://github.com/lemonade-sdk/lemonade/actions/runs/31046661982) and [31057852790](https://github.com/lemonade-sdk/lemonade/actions/runs/31057852790), where a previous `lemond` still held `127.0.0.1:13305` while the new one started.

## Why it is hard to spot

A half-bound server is worse than one that fails to start: it passes every readiness check written against `localhost`, so it is reported healthy and only fails at the point where some specific client uses a literal IPv4 address. In CI this presented as one test suite failing with `Connection refused` while ten others passed against the same process, which reads like a crash rather than a bind problem.

## Suggested behavior

Either:

- Treat a partial bind as fatal — if the host resolves to both families and only one binds, exit non-zero rather than serving, or
- Keep serving but make the degraded state discoverable: a `[Warn]`/`[Error]` at every request-handling entry point is too noisy, so at minimum surface the bound addresses in `/api/v1/health` (and `/api/v1/system-info`) so a client can tell what it is actually talking to.

The first is probably right for a server whose whole contract is "listening on this port".

## Notes

Found while unifying the macOS CI jobs in #2911. Supersedes #2916, which attributed the same symptom to a crash after Kokoro eviction; that was wrong — the server never died, and the requests that appeared to kill it simply never arrived.

popey:
Fixes #2917.

When a host resolves to both IPv4 and IPv6, wait for every requested bind attempt and abort startup if either one fails. This prevents `lemond` from remaining healthy on only one address family after a bind race.

Tests:

- Built `lemond` and passed all 39 `cpp-ci` tests on Ubuntu 24.04.
- Built natively on an M1 Mac and reproduced the reported IPv4-fails/IPv6-succeeds case; the patched process exited with status 1 and closed the IPv6 listener.
- Confirmed normal dual-stack startup remained healthy over both IPv4 and IPv6 on macOS.

### DIFF
--- a/src/cpp/include/lemon/utils/network_utils.h
+++ b/src/cpp/include/lemon/utils/network_utils.h
@@ -1,9 +1,43 @@
 #pragma once
 
+#include <atomic>
+#include <cstddef>
 #include <string>
 
 namespace lemon::utils {
 
 bool is_tcp_listener_active(int family, const std::string& host_ip, int port);
+
+class ListenerStartupState {
+public:
+    explicit ListenerStartupState(std::size_t expected_bind_attempts)
+        : expected_bind_attempts_(expected_bind_attempts) {}
+
+    void record_bind_success() {
+        completed_bind_attempts_.fetch_add(1);
+    }
+
+    void record_bind_failure() {
+        failed_.store(true);
+        completed_bind_attempts_.fetch_add(1);
+    }
+
+    void record_listen_failure() {
+        failed_.store(true);
+    }
+
+    bool bind_attempts_complete() const {
+        return completed_bind_attempts_.load() == expected_bind_attempts_;
+    }
+
+    bool failed() const {
+        return failed_.load();
+    }
+
+private:
+    const std::size_t expected_bind_attempts_;
+    std::atomic<std::size_t> completed_bind_attempts_{0};
+    std::atomic<bool> failed_{false};
+};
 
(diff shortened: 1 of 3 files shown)


kenvandine (approved):
Thanks, this looks good now!

PR #3198, opened 2026-08-17
Fix incompatible GGUF results in model search

### LINKED ISSUE #3136: GUI3: able to download incompatible model that then causes "http://localhost:13305/api/v1/load returned 500 Internal Server Error: Failed to load model 'Krea-2-Turbo-GGUF-Q8_0': llama-server failed to start"
### Platform

Linux/Ubuntu

### Lemonade Version

11.5.2

### GPU / APU Model

Strix Halo

### Component

Desktop App / UI

### Bug Description

Downloaded Krea-2-Turbo-GGUF-Q8_0, but it fails to load "http://localhost:13305/api/v1/load returned 500 Internal Server Error: Failed to load model 'Krea-2-Turbo-GGUF-Q8_0': llama-server failed to start"

### Steps to Reproduce

1. Models tab
2. Searched "krea 2 turbo"
3. Downloaded model listed by HuggingFace (as per image)
4. Tried to load it

Image

_No response_

### Log Output

```shell
14:24:12.098
INF
Server
Pulling model: Krea-2-Turbo-GGUF-Q8_0
14:24:12.098
INF
Server
recipe: llamacpp
14:24:12.098
ERR
Server
Error 400: POST /api/v1/pull
14:24:12.098
ERR
Server
400 Bad Request details - Body length: 181, Content-Type: application/json
14:24:12.111
INF
Server
Ensuring model loaded: Krea-2-Turbo-GGUF-Q8_0 
14:24:12.112
INF
Router
Auto-tune ctx_size resolved to 32768
14:24:12.112
INF
LlamaCpp
Loading model: user.Krea-2-Turbo-GGUF-Q8_0
14:24:12.112
INF
LlamaCpp
Using LlamaCpp Backend: vulkan
14:24:12.174
INF
LlamaCpp
Starting llama-server...
14:24:12.177
INF
ProcessManager
Process started successfully, PID: 1994438
14:24:12.215
INF
Process
0.00.031.654 I cmn  common_param: common_params_print_info: verbosity = 3 (adjust with the `-lv N` CLI arg)
14:24:12.215
INF
Process
0.00.031.877 I srv          init: The UI is disabled
14:24:12.215
INF
Process
0.00.031.879 I srv          init: Use --ui/--no-ui (or deprecated --webui/--no-webui) to enable/disable
14:24:12.215
INF
Process
0.00.031.977 W srv  llama_server: -----------------
14:24:12.215
INF
Process
0.00.031.981 W srv  llama_server: CORS is set to allow all origins ('*') and no API key is set
14:24:12.215
INF
Process
0.00.031.981 W srv  llama_server: this can be a security risk (cross-origin attacks)
14:24:12.215
INF
Process
0.00.031.981 W srv  llama_server: more info: https://github.com/ggml-org/llama.cpp/pull/25655
14:24:12.215
INF
Process
0.00.031.981 W srv  llama_server: -----------------
14:24:12.216
INF
Process
0.00.033.210 I srv    load_model: loading model '/home/darren/.cache/huggingface/hub/models--stduhpf--Krea-2-Turbo-GGUF/snapshots/ad39dc86347ec69c184299a002e197b9fac3cfed/krea2-turbo-Q8_0.gguf'
14:24:12.218
ERR
Process
0.00.035.333 E llama_model_load: error loading model: unknown model architecture: ''
14:24:12.218
INF
Process
0.00.035.339 E llama_model_load_from_file_impl: failed to load model
14:24:12.219
ERR
Process
0.00.035.366 E common_fit_params: encountered an error while trying to fit params to free device memory: failed to load model
14:24:12.220
ERR
Process
0.00.036.944 E llama_model_load: error loading model: unknown model architecture: ''
14:24:12.220
INF
Process
0.00.036.945 E llama_model_load_from_file_impl: failed to load model
14:24:12.220
INF
Process
0.00.036.947 E cmn  common_init_: failed to load model '/home/darren/.cache/huggingface/hub/models--stduhpf--Krea-2-Turbo-GGUF/snapshots/ad39dc86347ec69c184299a002e197b9fac3cfed/krea2-turbo-Q8_0.gguf'
14:24:12.220
INF
Process
0.00.036.950 E srv    load_model: failed to load model, '/home/darren/.cache/huggingface/hub/models--stduhpf--Krea-2-Turbo-GGUF/snapshots/ad39dc86347ec69c184299a002e197b9fac3cfed/krea2-turbo-Q8_0.gguf'
14:24:12.220
INF
Process
0.00.036.951 I srv    operator(): operator(): cleaning up before exit...
14:24:12.221
ERR
Process
0.00.037.881 E srv  llama_server: exiting due to model loading error
14:24:12.277
ERR
WrappedServer
llama-server process has terminated with exit code: 1
14:24:12.277
ERR
WrappedServer
This usually means:
14:24:12.277
ERR
WrappedServer
- Missing required drivers or dependencies
14:24:12.277
ERR
WrappedServer
- Incompatible model file
14:24:12.277
ERR
WrappedServer
- Try running the server manually to see the actual error
14:24:12.278
ERR
Router
Backend load failed: llama-server failed to start
14:24:12.278
WRN
Router
Load failed with non-file-not-found error, evicting all models and retrying...
```

### Additional Context

_No response_

popey:
## Summary

Fixes #3136.

Hugging Face GGUF repositories were treated as llama.cpp models based only on their filenames. `/pull/variants` now checks parsed GGUF architecture metadata against an allowlist generated from Lemonade's pinned llama.cpp source; file metadata and compatibility metadata use separate supported Hugging Face requests.

## Scope

## Testing

_Testing details:_

- Built `lemond`, `test_hf_variants`, and `test_model_registry` after rebasing onto current `main`.
- Passed `HfVariantsTest`, `ModelRegistryTest`, pre-commit, and workflow validation.
- Reproduced #3136 with released Lemonade 11.6.0.
- Verified the patched endpoint rejects the reported malformed Krea repository and an untagged `krea2` repository.
- Verified valid TinyLlama, Gemma 3 multimodal, and Nomic embedding repositories remain accepted, with positive variant sizes.

## Documentation

## Breaking Changes

## AI-assisted contribution


jeremyfowers:
**Ready for review**

No to-dos.

The PR title and body accurately describe the scope, mechanism, and surface. The diff adds parsed GGUF metadata to HuggingFace queries, rejects media tasks and missing architectures before /pull/variants, and preserves ModelScope/mirror behavior exactly as claimed.

Checked: The PR body states it adds parsed GGUF metadata to repository inspection and rejects media tasks, missing architectures, and models without positive text context length before /pull/variants; the diff confirms llamacpp_gguf_incompatibility() in hf_variants.cpp, expanded query params in model_registry.cpp, and the guard in fetch_pull_variants().

**Focus: focused**

All changes serve one goal: preventing incompatible GGUF repositories from being offered for download. The new llamacpp_gguf_incompatibility function, the expanded HuggingFace query parameters, the /pull/variants guard, and the unit + regression tests all exist to solve this single problem.

Checked: All six changed files serve the single goal of filtering incompatible GGUF models: two headers expose new functions, two source files implement the metadata check and guard, and two test files cover the new logic in C++ and Python.

**Alignment: none found**

Checked: reviewed contribute.md AI Contribution Policy and Reviewer Expectations; the PR describes its changes accurately and stays within one problem.

**Documentation: adequate**

Checked: no user-facing surface changed; the change is an internal filter applied before model download, so no docs update is required.

**Testing: adequate**

Checked: added 100-line C++ unit test suite in test/cpp/test_hf_variants.cpp covering valid, malformed, missing, media, mirror, and ModelScope cases; added opt-in integration test in test/server_endpoints.py using the reported Krea checkpoint.

Checked: none observed — this is a bug fix that makes the server reject models it should have always rejected, restoring intended behavior.

</details>

**Review suggestion**

Suggested reviewers:

- superm1 (llamacpp) — The PR adds GGUF compatibility filtering logic that determines which Hugging Face repositories are valid llama.cpp models — directly in the llamacpp subject area.
- pwilkin (llamacpp) — The PR changes how GGUF metadata is parsed and validated for llama.cpp compatibility, which falls under the llamacpp subject area.
- jeremyfowers (hf_variants.cpp/hf_variants.h) — Last edited 14 lines across src/cpp/include/lemon/hf_variants.h and src/cpp/server/hf_variants.cpp (refactor(backends): self-describing WrappedServer backends, 2026-06-29), the exact files this PR modifies to add GGUF incompatibility checking.

**Why this rung**

A contained fix or extension. A contained fix in the model registry and variant enumeration code. No new command, route, backend, pane, or policy — just a filter guard on an existing code path.

**Reviewer search**

Checked: Identified surfaces: llamacpp GGUF compatibility checking (hf_variants.cpp/hf_variants.h), model registry search expansion and filtering (model_registry.cpp). Blamed terms gguf, llamacpp_gguf_incompatibility, pipeline_tag, enumerate_gguf_variants, registry_task — top hits were @jeremyfowers (14 lines on hf_variants.h/hf_variants.cpp) and @fl0rianr (10 lines on model_registry.cpp/hf_variants.h). @superm1 and @pwilkin list llamacpp matching the primary surface. Code-authorship pass: run get-pr-code-authors.sh — script not found in repo; manual blame confirmed @jeremyfowers as code-author of the hf_variants files this PR modifies. @jeremyfowers added to suggested_reviewers on code-author basis.

</details>

_[AI-assisted review] Automated pre-review from repo-manager — flags for the human reviewer, not a replacement for one. Reviewed at head `7dc67d6`; regenerate with `repo-manager review-pr 3198`._

jeremyfowers:
@pwilkin interested in reviewing this one? it would be a really nice fix to have.

fl0rianr (commented):
The bot mentioned me in the explanations so here are my 2 cents:

The compatibility filtering itself looks good, and I couldn't find another issue in the decision logic. One thing I'd like to resolve before merging is the Hugging Face request: it now combines blobs=true with expand[]=..., while huggingface_hub explicitly treats files_metadata (blobs=true) and expand as mutually exclusive. Since we still rely on sibling size/LFS metadata for variant sizes, I'd prefer keeping the file-metadata request supported and fetching the GGUF/task metadata separately (or otherwise using a documented API combination). Please also add/retain coverage that a normal GGUF variant still reports a non-zero size_bytes.

pwilkin:
I would like this to be comprehensive.

Heuristic filtering based on criteria such as included in this pull request only narrows the scope of the incorrectly permitted models. The way I would like to see this work is:

- we extract the supported architectures from the llama.cpp source code (either by a script extracting from GitHub or by extracting this beforehand)
- we check the GGUF metadata and we parse by architecture
- for the future, we extend this behavior to sd.cpp and possibly other backends, but that would require support for complex recipes, so out of scope for this PR (but it should be easily extensible to support that)

fl0rianr (commented):
Thanks, looks much better, and my previous blobs=true + expand concern is fully addressed. One remaining issue:
The second compatibility-metadata request currently fails open. Any non-200 response, including 429/5xx, is silently ignored, after which missing gguf metadata is treated as compatible.
That can reintroduce exactly the incompatible-result path this PR is meant to close, especially now that marketplace validation makes an additional HF request per candidate.

I think metadata-unaware mirrors can remain permissive, but transient/rate-limit/server failures should not turn into “compatible”. As a smaller robustness improvement, it would also be good to issue the metadata request against the SHA returned by the first request rather than mutable main, so the architecture metadata and file tree describe the same snapshot.

popey:
@fl0rianr Thanks. I've updated the compatibility request to use the SHA from the file-tree response. Rate limits and server errors now fail closed for both Hugging Face and custom mirrors, and malformed responses from the official endpoint fail closed; custom mirrors remain permissive only where the compatibility metadata endpoint appears unsupported. I added coverage for those cases and rechecked the live Krea rejection and TinyLlama variant sizes.

fl0rianr (commented):
The previous two issues are fixed cleanly, thanks. I found two remaining concerns on the full re-review:

* The compatibility metadata request now lives inside generic HuggingFaceRegistry::fetch_repository(). That function is also used by normal downloads, update checks, auxiliary repos and snapshot comparisons, so every HF repository operation now gets an extra request and can fail because that compatibility-only request gets a 429/5xx. I think the extra metadata fetch should stay scoped to /pull/variants / GGUF validation rather than changing the reliability contract of generic repository discovery.
* The generated architecture list is based only on llamacpp.cpu (b10375), but Lemonade currently ships different pins per backend, e.g. rocm-stable = b10470. This already creates real false negatives: b10470 supports architectures such as bailingmoe3, minimax-01 and kimi-k3, which aren't present in the b10375-generated allowlist. We should either make compatibility backend/pin-aware, or explicitly define this as an intersection/common-denominator policy rather than implicitly using the CPU pin.

Other than those two points, the previous metadata-request, snapshot consistency and fail-open issues look resolved

popey:
@fl0rianr @pwilkin I’ve updated this on top of current `main`.

The compatibility metadata request is now limited to GGUF variant validation, so it no longer changes the behaviour of normal downloads, update checks, auxiliary repositories, or snapshot comparisons.

The generated architecture allowlist is now the common denominator across all llama.cpp revisions configured in `backend_versions.json`, rather than being based only on the CPU pin. The llama.cpp update workflow regenerates the list after updating those pins.

I rebuilt from a fresh `BUILD_TESTING=OFF` configuration, passed all 68 `cpp-ci` tests, pre-commit, and workflow validation. I also rechecked the live endpoint: the reported Krea repository is rejected, while TinyLlama returns all 12 variants with positive sizes and a Kimi-K3 repository supported by the newer shipped pin remains accepted.

Could someone approve the workflows and take another look when convenient?

fl0rianr (approved):
Thanks, LGTM.

But let's wait what @pwilkin says.

pwilkin (approved):
Yup, this is exactly what I had in mind, thanks!

PR #3199, opened 2026-08-17
refactor(server): drop load_command from the options endpoint

### LINKED ISSUE #3191: Remove the "load_command" field from /model/id/options before release

jeremyfowers:
Context: we realized that the design of this field was based on a misunderstanding. It was introduced after v11.6.0 released, so if remove before vNext we can avoid supporting it forever as a published API. FYI @superm1 this should be in the vNext release this week.

## Summary

Removes the `load_command` field from `/v1/models/{id}/options` before it ships, along with the shell-quoting helper that existed only to build it, its documentation, and the two tests that covered it. Clients that need the command can build it from `effective`, which is already the exact `/v1/load` body.

Closes #3191

## Scope

Note: the two 4-line Black reformats in `test/server_endpoints.py` are pre-existing on `main` (Black 26.1.0 flags them there too); the pre-commit hook applied them because it formats the whole file.

## Testing

_Testing details:_ Built `lemond` (Ninja/Release) and ran it against an isolated cache dir and port. `GET` and `POST {"dry_run": true}` on `/api/v1/models/Qwen3-0.6B-GGUF/options` both return `model_name`, `recipe`, `saved`, `effective`, `defaults`, and `resolved_ctx_size` with no `load_command`.

## Documentation

## Breaking Changes

The field was never in a release.

## AI-assisted contribution

### DIFF
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -272,22 +272,9 @@ curl http://localhost:13305/v1/models/Qwen3-0.6B-GGUF/options
     "merge_args": true,
     "model_name": "Qwen3-0.6B-GGUF"
   },
-  "resolved_ctx_size": 32768,
-  "load_command": "curl -X POST $LEMONADE_BASE_URL/v1/load -H \"Content-Type: application/json\" -d '{\"ctx_size\":-1, ...}'"
+  "resolved_ctx_size": 32768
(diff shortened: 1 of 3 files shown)


fl0rianr (approved):
LGTM

PR #3142, opened 2026-08-14
Merge 'main' into 'GUI3_merging'

fl0rianr:
Chore...

@bitgamma Please have a look, I hope theNoise works as expected. Thanks!

PR #3203, opened 2026-08-18
fix(gui3): derive recipe configuration from system-info

fl0rianr:
## Summary

Derive GUI3 recipe configuration from the recipe metadata already exposed by `/system-info` instead of maintaining separate frontend recipe tables.

This fixes the recipe coupling uncovered during #3142, where a new recipe such as `thenoise` could be classified incorrectly or inherit configuration fields from another recipe such as `sd-cpp`.

The Model Configuration tab now uses the server-provided recipe option schema for field discovery and field types, while keeping existing GUI labels and presentation overrides. Backend selection is also resolved from the recipe metadata rather than a frontend recipe allowlist.

As a follow-up to #3183, the Backends page now only exposes the backend arguments action when the selected backend variant has a concrete writable runtime-config target. This keeps the existing per-backend args safety behavior while allowing future backends to expose args without adding another GUI recipe allowlist.

No server API or C++ changes are required.

Related to #3161.
Follow-up to #3142 and #3183.

## Scope

## Testing

_Testing details:_

- npm run typecheck
- npm test
- manual UI tresting and inference

Regression coverage includes:
- server-defined recipes such as `thenoise` use their own option schema instead of `sd-cpp` fields
- unknown recipes are not silently classified as LLM
- backend selector fields are discovered from server recipe metadata
- future `*_args` fields can be discovered without adding a frontend recipe entry
- backend args actions require a concrete writable runtime-config target
- existing `llama.cpp` accessible labels remain unchanged
- model-specific defaults continue to come from lemond

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/app/src/components/BackendManager.tsx
+++ b/src/app/src/components/BackendManager.tsx
@@ -1,12 +1,10 @@
 import React, { useState, useEffect, useCallback, useMemo, useRef } from 'react';
 import api, { friendlyErrorMessage } from '../api';
-import {
-  BackendTuning,
-  backendSupportsArgs,
-} from '../modelConfiguration';
+import { BackendTuning } from '../modelConfiguration';
 import { Icon, type IconName } from './Icon';
 import { WorkspaceCatalogLayout, WorkspaceCatalogSection } from './WorkspaceCatalogLayout';
 import { useFocusTrap } from '../hooks/useFocusTrap';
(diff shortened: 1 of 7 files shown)


bitgamma (approved):
This indeed makes TheNoise display and save its actual settings

PR #3205, opened 2026-08-18
fix(gui): omni model bug

fl0rianr:
## Summary

Fixes a GUI3 regression where Omni collections can be downloaded and loaded successfully, but the chat view still reports that the collection "did not become ready for chat."

Omni collections are virtual models: the server loads their component models rather than adding the collection name itself to `all_models_loaded`.

The chat readiness check now uses the existing virtual collection logic so a fully loaded Omni collection is correctly recognized as ready for chat.

## Scope

## Testing

_Testing details:_

Reproduction target: `LMX-Omni-5.5B-Lite`.

Before the fix, the model downloads and its components load, but sending a chat message fails with:

`LMX-Omni-5.5B-Lite was downloaded but did not become ready for chat.`

Test:
- Load `LMX-Omni-5.5B-Lite` and send a normal chat message, also asked for an image, worked both.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -46,7 +46,7 @@ import {
 import { storageKey } from '../storage';
 import { CHAT_HISTORY_PREFERENCE_EVENT, loadChatHistoryPreference } from '../features/chatHistory/historySettings';
 import type { DownloadListItem } from '../features/downloadManager/downloadStore';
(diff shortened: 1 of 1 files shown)


jeremyfowers (approved):
LGTM

PR #3208, opened 2026-08-18
fix(pre-commit): make comment-slop runnable on Windows

blackdeathdrow:
`comment-slop` uses `language: script`, which on Windows can't launch a `.py` via `CreateProcess` (no shebang, no assoc) → hook exits `9009`. Switch to `language: python` so pre-commit provisions an interpreter and runs `python tools/check_comment_slop.py`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,8 +30,8 @@ repos:
     hooks:
       - id: comment-slop
         name: comment slop (repeated explanations, oversized blocks)
-        entry: tools/check_comment_slop.py
-        language: script
+        entry: python tools/check_comment_slop.py
+        language: python
         pass_filenames: false
         require_serial: true
         types_or: [c, c++, python]


.pre-commit-config.yaml
```diff
@@ -30,8 +30,8 @@ repos:
     hooks:
       - id: comment-slop
         name: comment slop (repeated explanations, oversized blocks)
-        entry: tools/check_comment_slop.py
-        language: script
+        entry: python tools/check_comment_slop.py
+        language: python
```
superm1, line 34: the problem with this is that on Linux you only have `/usr/bin/python` if you install `python-is-python3`

```
╰─❯ dpkg -S /usr/bin/python
python-is-python3: /usr/bin/python
```

blackdeathdrow: It resolves through the hook's venv, so /usr/bin/python isn't needed.

superm1: ah great!

superm1 (approved):

PR #3209, opened 2026-08-18
feat(server): enable cross-platform tcp keepalive socket options

abn:
### Summary

Enable OS-level TCP Keep-Alive socket options across `lemond`'s HTTP front server (`UpgradableFrontServer`) and WebSocket server (`WebSocketServer`) on Linux, macOS, and Windows.

### Rationale

Stateful L4 firewalls, NAT gateways, and intermediate routers often silently drop idle TCP connections after inactivity windows (often 60s or lower). For LLM inference workflows involving long time-to-first-token (TTFT) prefill or idle client connections between turns, silent drops cause broken connections without proper TCP RST or FIN notification.

*(Note: application-layer idle timeouts, such as nginx `proxy_read_timeout` or ALB idle timeout, require application data / SSE heartbeats.)*

### Changes

- **Cross-Platform Helper (`network_utils`)**: Implemented `lemon::utils::configure_tcp_keepalive(socket_t sock)` in `network_unix.cpp` and `network_windows.cpp`:
  - `SO_KEEPALIVE`: enabled
  - Idle time before probe: 15s (`TCP_KEEPIDLE` on Linux/Windows, `TCP_KEEPALIVE` on macOS)
  - Probe interval: 5s (`TCP_KEEPINTVL`)
  - Probe count: 3 probes (`TCP_KEEPCNT`)
- **HTTP Server (`UpgradableFrontServer`)**: Configures client sockets upon accept in `process_and_close_socket` before HTTP parsing.
- **WebSocket Server (`WebSocketServer`)**: Invokes `configure_tcp_keepalive()` on newly established client sockets (`LWS_CALLBACK_ESTABLISHED` via `lws_get_socket_fd`) as well as upgrade sockets adopted from the HTTP server.
- **Platform Separation**: Keeps server headers clean and free of OS-specific headers / `#ifdef`s.
- **Diagnostic Logging**: Emits debug logs identifying failed options with `errno` / `WSAGetLastError`.
- **Tests**: Consolidated loopback accepted-socket verification into `test/cpp/test_network_utils.cpp` running under `NetworkUtilsTest`.

### Verification

- `cmake --build build --preset default` (0 errors)
- `ctest --test-dir build -L cpp-ci --output-on-failure` (passed)
- `python3 docs/tools/gen_backend_boilerplate.py --check` (clean, 0 drift)

Relates-to: #2898

### DIFF
--- a/src/cpp/include/lemon/upgradable_http_server.h
+++ b/src/cpp/include/lemon/upgradable_http_server.h
@@ -34,6 +34,9 @@
 #include <string>
 #include <thread>
 
+#include "lemon/utils/aixlog.hpp"
+#include "lemon/utils/network_utils.h"
+
 namespace lemon {
 
 bool is_websocket_endpoint(const std::string& path);
@@ -139,6 +142,9 @@ public:
 
 private:
     bool process_and_close_socket(socket_t sock) override {
+        if (!lemon::utils::configure_tcp_keepalive(sock)) {
+            LOG(DEBUG, "Server") << "Failed to configure TCP keep-alive on accepted socket" << std::endl;
+        }
         if (upgrade_handler_ && detail::peek_websocket_upgrade(sock)) {
             if (upgrade_handler_(sock)) {
                 return true;  // ownership transferred to the WebSocket server
--- a/src/cpp/include/lemon/utils/network_utils.h
+++ b/src/cpp/include/lemon/utils/network_utils.h
@@ -4,6 +4,13 @@
 #include <cstddef>
 #include <string>
 
+#ifdef _WIN32
+#include <winsock2.h>
+using socket_t = SOCKET;
+#else
+using socket_t = int;
+#endif
+
 namespace lemon::utils {
 
 bool is_tcp_listener_active(int family, const std::string& host_ip, int port);
@@ -39,5 +46,13 @@ private:
     std::atomic<std::size_t> completed_bind_attempts_{0};
     std::atomic<bool> failed_{false};
 };
+
+// Enable TCP keepalive probes with 15s initial idle, 5s retry interval, and
+// 3 probe attempts (~30s total dead-peer detection) to prevent stateful L4
+// firewalls, NAT gateways, and intermediate routers from silently dropping
+// idle TCP connections during prolonged inactivity (e.g. prompt evaluation).
+// Note: application-layer idle timeouts (e.g. nginx proxy_read_timeout or ALB
+// idle timeout) require periodic application data / SSE heartbeats.
+bool configure_tcp_keepalive(socket_t sock);
 
 } // namespace lemon::utils
--- a/src/cpp/server/utils/platform/network_unix.cpp
+++ b/src/cpp/server/utils/platform/network_unix.cpp
@@ -1,15 +1,55 @@
+#include "lemon/utils/aixlog.hpp"
 #include "lemon/utils/network_utils.h"
 
 #include <arpa/inet.h>
+#include <cstring>
 #include <errno.h>
 #include <fcntl.h>
 #include <netinet/in.h>
+#include <netinet/tcp.h>
 #include <sys/select.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <unistd.h>
 
+#if defined(__APPLE__)
+#define LEMON_TCP_KEEPIDLE TCP_KEEPALIVE
+#elif defined(TCP_KEEPIDLE)
+#define LEMON_TCP_KEEPIDLE TCP_KEEPIDLE
+#endif
+
 namespace lemon::utils {
+
+bool configure_tcp_keepalive(socket_t sock) {
+    int opt = 1;
+    bool ok = true;
+    if (setsockopt(sock, SOL_SOCKET, SO_KEEPALIVE, &opt, sizeof(opt)) != 0) {
+        LOG(DEBUG, "Server") << "setsockopt(SO_KEEPALIVE) failed: " << std::strerror(errno) << " (" << errno << ")" << std::endl;
+        ok = false;
+    }
+#ifdef LEMON_TCP_KEEPIDLE
+    int idle = 15;
+    if (setsockopt(sock, IPPROTO_TCP, LEMON_TCP_KEEPIDLE, &idle, sizeof(idle)) != 0) {
+        LOG(DEBUG, "Server") << "setsockopt(LEMON_TCP_KEEPIDLE) failed: " << std::strerror(errno) << " (" << errno << ")" << std::endl;
+        ok = false;
+    }
+#endif
+#ifdef TCP_KEEPINTVL
+    int interval = 5;
+    if (setsockopt(sock, IPPROTO_TCP, TCP_KEEPINTVL, &interval, sizeof(interval)) != 0) {
+        LOG(DEBUG, "Server") << "setsockopt(TCP_KEEPINTVL) failed: " << std::strerror(errno) << " (" << errno << ")" << std::endl;
+        ok = false;
+    }
+#endif
+#ifdef TCP_KEEPCNT
+    int count = 3;
+    if (setsockopt(sock, IPPROTO_TCP, TCP_KEEPCNT, &count, sizeof(count)) != 0) {
+        LOG(DEBUG, "Server") << "setsockopt(TCP_KEEPCNT) failed: " << std::strerror(errno) << " (" << errno << ")" << std::endl;
+        ok = false;
+    }
+#endif
+    return ok;
+}
 
 bool is_tcp_listener_active(int family, const std::string& host_ip, int port) {
     int sock = socket(family, SOCK_STREAM, IPPROTO_TCP);
--- a/src/cpp/server/utils/platform/network_windows.cpp
+++ b/src/cpp/server/utils/platform/network_windows.cpp
@@ -1,9 +1,55 @@
-#include "lemon/utils/network_utils.h"
-
+// Windows header discipline — must precede all other includes.
+#ifdef _WIN32
+#ifndef _WINSOCKAPI_
+#define _WINSOCKAPI_
+#endif
+#define WIN32_LEAN_AND_MEAN
+#define NOMINMAX
 #include <winsock2.h>
 #include <ws2tcpip.h>
+#include <mstcpip.h>
+#include <windows.h>
+#pragma comment(lib, "ws2_32.lib")
+#ifdef ERROR
+#undef ERROR
+#endif
+#endif
+
+#include "lemon/utils/network_utils.h"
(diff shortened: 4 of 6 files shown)


fl0rianr (commented):
I think the feature itself makes sense, but I’d like to see the implementation tightened before merge.

My  concern is the amount of duplicated OS-specific code. Linux, macOS and Windows currently repeat almost the same `setsockopt()` sequence, even though the real differences are mostly `TCP_KEEPALIVE` vs `TCP_KEEPIDLE` and the Windows argument type. This goes against the cleanup direction we already took in #2142 and #2692 to reduce scattered `#ifdef`s. A small common helper with only the actual platform differences guarded should be enough.

Also, every `setsockopt()` result is currently ignored. Please make failures observable instead of silently continuing, especially since compile-time availability of an option does not guarantee runtime support. The hardcoded `15/5/3` policy should probably get a short rationale as well.

Finally, the new test mostly tests `configure_tcp_keepalive()` directly. CI runs that runtime check on Linux, while Windows/macOS mainly get build coverage. Given that this is specifically cross-platform socket behavior, it would be good to either add real platform coverage or keep the implementation small enough that the remaining platform-specific surface is minimal. And for added test group them e.g. test_streaming_proxy_cancel.cpp.

abn:
Thanks for the review @fl0rianr! Updated the implementation addressing all points:

1. **Deduplicated OS-specific code**: Introduced a generic template helper `set_socket_opt(sock, level, optname, value)` and a unified `LEMON_TCP_KEEPIDLE` alias (`TCP_KEEPALIVE` on macOS, `TCP_KEEPIDLE` elsewhere), collapsing repeated per-OS `setsockopt` calls into a single unified sequence.
2. **Observable socket options**: `configure_tcp_keepalive` now returns a boolean status verifying that options applied successfully.
3. **Rationale documented**: Added inline architectural documentation explaining the 15/5/3 timing policy (~30s total dead connection detection to prevent premature teardown by 60s intermediate NAT/proxy timeouts).
4. **Enhanced test suite**: Updated `test_tcp_keepalive.cpp` to verify both `set_socket_opt` helper behavior and full `configure_tcp_keepalive` option configuration with explicit assertions.

fl0rianr (commented):
Thanks, this is cleaner, but I don't think all of the previous points are addressed yet.

configure_tcp_keepalive() now returns a status, but the production caller still ignores it, so setsockopt() failures remain completely silent.

Also, I think we can avoid adding the platform abstraction to upgradable_http_server.h altogether. We already have network_utils.h with network_unix.cpp / network_windows.cpp, selected per platform by CMake. This seems like the natural place for the keepalive implementation and would keep the OS headers, Windows cast and LEMON_TCP_KEEPIDLE macro out of the server header, consistent with the #2142/#2692 cleanup direction.

The test point also remains: this still adds a separate 217-line test_tcp_keepalive.cpp, while test_network_utils.cpp already contains the cross-platform socket test infrastructure. And CI currently executes the new keepalive test only on Linux, so Windows/macOS still don't get runtime validation.

With those addressed, this looks much closer to merge-ready.

abn:
Thanks for the thoughtful follow-up @fl0rianr! Completely agree with these architectural points. Updated and pushed:

1. **Platform separation via network utils**: Moved the socket keepalive implementation into `lemon::utils::configure_tcp_keepalive(sock)` declared in `src/cpp/include/lemon/utils/network_utils.h` and defined separately in `src/cpp/server/utils/platform/network_unix.cpp` and `network_windows.cpp`. This completely cleans `src/cpp/include/lemon/upgradable_http_server.h` of all OS headers, macros, and socket `#ifdef`s.
2. **Observable failure in production**: `UpgradableFrontServer::process_and_close_socket` now logs failures via `LOG(DEBUG, "Server")` if socket option configuration fails.
3. **Consolidated tests in `test_network_utils.cpp`**: Merged all TCP keepalive assertions into `test/cpp/test_network_utils.cpp`, reusing existing cross-platform socket infrastructure and running under multi-platform CI, while eliminating the standalone test file and its CMake entry.
4. **Rebased**: Rebased cleanly onto the latest `main`.

abn:
Fixed the Windows build issue in `network_utils.h` by adding `using socket_t = SOCKET;` in the `_WIN32` branch. Tested and re-pushed.

fl0rianr (commented):
Thanks, this is much cleaner now. The move into network_utils and the consolidation into test_network_utils.cpp address my original platform/ifdef concerns.

I did another pass over the actual socket paths and there are two things I’d still like to address or clarify:

The keepalive test currently configures a fresh, unbound/unconnected socket, while production calls configure_tcp_keepalive() on an accepted/established socket. This matters especially on Windows, where TCP/IP options set before bind() are not necessarily validated by the TCP/IP stack until the bind happens. Since this test already lives in test_network_utils.cpp, could we use a real loopback listener/client pair and verify the options on the accepted socket? That would test the same socket state we use in production rather than only checking that setsockopt/getsockopt round-trip on a fresh socket.

Also, this currently only covers connections accepted through UpgradableFrontServer. The dedicated websocket_port listener still exists and is still used as a client fallback, but those sockets never go through configure_tcp_keepalive(). libwebsockets already exposes ka_time, ka_probes and ka_interval, so we could apply the same policy there without adding another OS-specific implementation. Otherwise I think the PR should explicitly scope the feature to the main HTTP port.

One documentation point: I’d narrow the reverse-proxy/load-balancer rationale a little. TCP keepalive is useful for dead-peer detection and L4 stateful devices, but it does not satisfy application-level idle/read timeouts such as nginx proxy_read_timeout or ALB idle timeout. That is what #3210’s SSE heartbeat addresses.

Minor: the failure log only says that configuration failed, without the failed option or errno/WSAGetLastError, which will make platform failures difficult to diagnose. The PR description is also stale and still references the removed test_tcp_keepalive and the old 38/38 count.

Other than these points, the original duplication / #ifdef concern is resolved for me.

abn:
Thanks for the thorough follow-up @fl0rianr! Updated all points:

1. **Accepted socket test verification**: Updated `test_configure_tcp_keepalive` in `test/cpp/test_network_utils.cpp` to establish a local loopback connection and verify options directly on the accepted socket, matching the production lifecycle.
2. **Dedicated WebSocket listener coverage**: Added `ka_time = 15`, `ka_interval = 5`, and `ka_probes = 3` to `lws_context_creation_info` in `src/cpp/server/websocket_server.cpp` so the fallback dedicated WebSocket port shares the exact keepalive policy.
3. **Diagnostic error logging**: Added option-specific logging with `strerror(errno)` (POSIX) and `WSAGetLastError()` (Windows) in `network_unix.cpp` and `network_windows.cpp` to pinpoint any failed socket options.
4. **Scoped rationale**: Refined the rationale comment in `network_utils.h` to clarify L4 dead-peer / stateful gateway detection vs L7 application idle timeouts.
5. **Updated PR description**: Refreshed the PR description with the current test target and test suite count (46/46 passed).

fl0rianr (commented):
Thanks, I went through the latest revision again. The points from my previous review are addressed now: the accepted-socket test, dedicated WebSocket keepalive, L4/L7 scope, diagnostics, and updated PR description all look good.

There is one new blocker on the current head though: both Windows C++ builds are failing. network_windows.cpp now includes aixlog.hpp before network_utils.h. On Windows, AixLog includes Windows.h, which can pull in the old Winsock 1 winsock.h before network_utils.h includes winsock2.h. The resulting CI errors are the expected Winsock redefinitions (sockaddr, fd_set, timeval, WSAData, socket functions, etc.).

We already handle this explicitly in process_windows.cpp, where the Windows/Winsock headers are required to come before all other includes, including AixLog. I think this file should follow the same header discipline.

Longer term, it may also be worth keeping winsock2.h out of the public network_utils.h entirely and using a platform-neutral socket handle at that boundary, similar to WebSocketServer::adopt_socket(intptr_t), but I don't consider that part blocking.

Once the Windows build is fixed, the implementation looks good to me. Running NetworkUtilsTest on Windows as well would still be useful for a cross-platform socket change, but the build regression is the only thing I would block on at this point.

abn:
Fixed the Windows header inclusion order in `network_windows.cpp` and `test_network_utils.cpp` following the `process_windows.cpp` header discipline (`WIN32_LEAN_AND_MEAN`, `_WINSOCKAPI_`, `winsock2.h` before `aixlog.hpp`/`windows.h`). Tested and force-pushed.

fl0rianr (commented):
Almost approved, but one more thing turned up when checking the actual libwebsockets 4.3.3 implementation we build against.

The ka_time / ka_interval / ka_probes settings don't give us the same 15/5/3 policy on all platforms.

In libwebsockets 4.3.3, the Unix socket implementation explicitly skips the per-socket keepalive timing options on __APPLE__. It only enables SO_KEEPALIVE, so the dedicated WebSocket listener on macOS falls back to the system TCP keepalive timings rather than 15/5/3. The macOS default keepalive idle time is much longer, so this can still miss the kind of ~60s L4 idle timeout this PR is meant to address.

Windows has a smaller version of the same issue: LWS 4.3.3 uses SIO_KEEPALIVE_VALS, which applies ka_time and ka_interval, but it never uses ka_probes, so the 3 probe count isn't actually applied there either.

The main HTTP port, including WebSockets adopted from it, is fine because those sockets already go through configure_tcp_keepalive().

I think the cleanest fix would be to reuse that same helper for WebSocket connections as well, e.g. from the LWS established/new-client callback using lws_get_socket_fd(wsi). That keeps one implementation of the policy and avoids relying on the platform-specific semantics of LWS ka_*.

With that fixed, I don't see another blocker. The Windows build regression from the previous revision is resolved and CI is green.

abn:
Updated the dedicated WebSocket listener to invoke `lemon::utils::configure_tcp_keepalive()` directly from `LWS_CALLBACK_ESTABLISHED` using `lws_get_socket_fd(wsi)`. This ensures our cross-platform 15/5/3 policy is applied identically across macOS (`TCP_KEEPALIVE`), Linux, and Windows (`TCP_KEEPCNT`) without relying on LWS's internal platform-specific `ka_*` handling.

fl0rianr (approved):
Rechecked the latest revision carefully. The direct configure_tcp_keepalive() call from LWS_CALLBACK_ESTABLISHED resolves the libwebsockets 4.3.3 platform discrepancy I raised, and I don't see a remaining functional blocker.

A few non-blocking cleanup points: the PR description still describes the old ka_* WebSocket configuration and says 46/46 tests, while the current implementation uses configure_tcp_keepalive() directly and cpp-ci is now 48/48. I'd also capture errno / WSAGetLastError() immediately after a failed setsockopt() before logging.

Cross-platform runtime coverage for NetworkUtilsTest would still be useful as a follow-up, but I don't think it needs to block this PR.

LGTM from my side.

PR #3211, opened 2026-08-18
fix(gui3): make pinned models server-owned

fl0rianr:
## Summary

Make runtime model pinning server-owned in GUI3.

- read the current pin state from lemond instead of maintaining a second `pinned_models` state in localStorage
- update pin state through `/internal/pin`
- remove the obsolete localStorage pin ownership and clean up the stale key
- keep Pin available in model details **only when the model is actually loaded**
- use **Favorite as the model-list secondary action**
- keep favorite state visible in the list with a small indicator when the hover action is not shown (seperate commit)
- update the affected storage, feature and accessibility tests

Related to #3161.

### Why the model-list action changes

This looks slightly wider than a storage fix, but it follows from making Pin reflect the server's runtime state.

A pin only applies to a loaded model. In practice that is normally one or two models, while the catalog can easily contain 100+ or more models and e.g. 15 favorites are likely numbers. Keeping Pin as the shared right-side action in the model list would therefore make that action unavailable for most rows.

The list row also intentionally has one shared secondary-action slot. Special-casing that slot so Pin only exists for loaded rows would break the interaction pattern used by the other lists (chat, telemetry etc.).

Favorite fits that slot better because it can be applied to every model in the catalog. It is also useful as the lightweight way to keep a small set of models easy to find, especially with the separate **My models** navigation planned to go away and **Favorite** seems to fit much better . Keeping Favorite only in the detail card would make that workflow unnecessarily slow for a larger catalog.

Pin state is not hidden by this change: pinned models still have their dedicated `Pinned` section in the model list and the state remains visible and controllable from model details.

There is one remaining workflow gap: loading a model takes the user into Chat, while Pin currently lives in the Models view. I plan to add the Pin control to the loaded-model UI in #3141, where it matches the server semantics well because that view only contains models that are actually loaded.

I hope in the light of this change make sense to you @jeremyfowers 

## Scope

## Testing

_Testing details:_

- npm test
- verify Pin is only offered for loaded models
- verify pin/unpin updates the server-owned state and the Pinned section
- verify Favorite can be toggled from any catalog row
- verify favorited rows show the small stationary star, replace it with the Favorite action on hover, and restore the backend label after hover
- verify favorite state persists across reload

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/app/src/components/ModelManager.tsx
+++ b/src/app/src/components/ModelManager.tsx
@@ -30,10 +30,8 @@ import {
   GLOBAL_MODEL_SETTINGS_EVENT,
   automaticUpdateIsDue,
   loadGlobalModelSettings,
-  loadPinnedModelNames,
   loadWithGlobalModelPolicy,
   saveGlobalModelSettings,
-  savePinnedModelNames,
   type GlobalModelSettings,
 } from '../features/modelSettings/globalModelSettings';
 import { backendLabel } from '../modelPresentation';
@@ -1127,7 +1125,10 @@ const ModelManager: React.FC<ModelManagerProps> = ({ onModelSelect, openModelReq
   const [dynamicRecipeOptions, setDynamicRecipeOptions] = useState<Partial<Record<CustomModelCapability, CustomRecipeOption[]>>>({});
   const [customRecipeAvailabilityLoaded, setCustomRecipeAvailabilityLoaded] = useState(false);
   const [systemInfo, setSystemInfo] = useState<Record<string, unknown> | null>(() => api.systemInfoData);
-  const [pinnedModels, setPinnedModels] = useState<string[]>(() => loadPinnedModelNames());
+  const pinnedModels = useMemo(
+    () => loadedModels.filter(model => model.pinned === true).map(model => model.model_name),
+    [loadedModels],
+  );
   const [favoriteModels, setFavoriteModels] = useState<string[]>(() => loadFavoriteModels());
   // Real disk usage for the storage meter (null until/unless lemond exposes it).
   const [storageInfo, setStorageInfo] = useState<import('../api').StorageInfo | null>(null);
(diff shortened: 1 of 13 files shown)


jeremyfowers:
ui/ux changes are fine by me

PCAssistSoftware (approved):
all looks good to me

jeremyfowers:
Will wait to review this one until https://github.com/lemonade-sdk/lemonade/pull/3262 merges and this one is rebased

jeremyfowers (approved):
Approving — PR faithfully moves pin state to the server: `LoadedModel.pinned` + `POST /internal/pin` replace client-local `localStorage`, the old `pinned_models` key and `loadPinnedModelNames`/`savePinnedModelNames` helpers are fully removed, and there's a regression test (`configuration-consistency.runtime.cjs`) guarding against dangling references. One real bug flagged inline below on `ModelManager.tsx` — worth a follow-up fix but not blocking.

src/app/src/components/ModelManager.tsx
```diff
@@ -2063,13 +2060,16 @@ const ModelManager: React.FC<ModelManagerProps> = ({ onModelSelect, openModelReq
...
   };
 
-  const togglePinnedModel = (name: string) => {
-    setPinnedModels(prev => {
-      const exists = prev.some(item => item.toLowerCase() === name.toLowerCase());
-      const next = exists ? prev.filter(item => item.toLowerCase() !== name.toLowerCase()) : [name, ...prev];
-      savePinnedModelNames(next);
-      return next;
-    });
+  const togglePinnedModel = async (name: string) => {
```
jeremyfowers, line 2063: Bug: this looks up the model in the raw \`loadedModels\` array, but the Pin button's enabled/disabled gate elsewhere (~line 3131) uses \`displayLoadedModels\`, which adds synthetic entries for fully-loaded Omni/router collections. Since \`collectionModels.ts\` filters out names already present in \`loadedModels\`, a fully-loaded collection's synthetic entry never matches here — \`loaded\` is \`undefined\`, and \`if (!loaded) return;\` silently no-ops with no API call and no user feedback.

Suggest having this lookup use \`displayLoadedModels\` instead (or dropping the redundant gate at line 3131, since \`ModelDetailPanel\` already applies the same \`isLoaded\` check internally).

PR #3212, opened 2026-08-18
feat(sandbox): add in-tree nono C FFI backend sandboxing

abn:
RFC: #3596

This changeset introduces kernel-level process sandboxing, capability confinement, and environment isolation for wrapped backend inference subprocesses (`llama-server`, `whisper-server`, `sd-server`, `flm`, `vllm`, etc.) using an in-tree C FFI static library wrapper around `nono` (0.73) alongside native OS security primitives.

> [!NOTE]
> Live manual verification has been conducted on Linux (x86_64). macOS and Windows will benefit from additional community validation.
> 
> This PR is experimental and currently submitted as a draft for testing and review.

### Key Architecture Highlights

* **Pure 4-Step Policy Assembly Line**:
  1. `Step 1 (System Runtime)`: Mounts standard dynamic linker paths, minimal device primitives, and base execution environment.
  2. `Step 2 (Hardware Profile)`: Adds GPU / NPU hardware device nodes and driver paths (`ROCm`, `CUDA`, `DirectML`, `XRT`).
  3. `Step 3 (Backend Workload Assets)`: Mounts the backend executable binary and the designated model checkpoint snapshot directory.
  4. `Step 4 (Declarative Recipe Delta)`: Merges per-model JSON `"sandbox"` overrides for companion weights (draft models, vision projectors) without C++ code changes.
* **Implicit Process Spawning**:
  - `WrappedServer::start_backend_process(...)` automatically builds the 4-step policy, validates boundaries, cleanses ambient environment, and launches the subprocess.
  - Backends with unique compiler caches (e.g. Triton in `vLLM`) declare them via the `WrappedServer::customize_sandbox_policy(...)` virtual hook.
* **Core Security Boundaries**:
  - **Default-Deny Outbound Egress**: External TCP/UDP egress is blocked while permitting local loopback communication (`127.0.0.1:<port>`).
  - **Filesystem Snapshot Isolation**: Positive allowlisting restricts filesystem visibility strictly to granted runtime paths and specific model snapshot subtrees; ungranted parent directories (including token roots) remain inaccessible.
  - **Zero-Privilege Environment Isolation**: Child processes execute in a scrubbed environment (`clearenv`) with a strict positive allowlist of runtime variables (`PATH`, `LD_LIBRARY_PATH`, GPU/NPU device indices), omitting ambient secrets and credentials by default.
* **Cross-Platform Primitives**:
  - **Linux**: Sandboxed via `nono` Rust C FFI (Landlock LSM + Seccomp BPF filters applied in-child).
  - **macOS**: Sandboxed via `nono` Rust C FFI (Apple Seatbelt profile generation and `sandbox_init`).
  - **Windows**: Native Win32 AppContainer isolation + transient NTFS ACEs + Anonymous Job Objects (`KILL_ON_JOB_CLOSE`), delegating to `nono` Landlock when executed inside WSL2.

### Technical Design Brief

The sandboxing subsystem decouples security policy construction from backend engine lifecycles through a data-driven pipeline:
1. **Policy Assembly**: An immutable baseline is computed in `WrappedServer::build_default_sandbox_policy` by composing the OS runtime, hardware device profile (`DEVICE_GPU`, `DEVICE_NPU`, `DEVICE_CPU`), and workload assets. Model recipes can append declarative filesystem, device, or network overrides via the `"sandbox"` key in `server_models.json`.
2. **Platform Dispatch & Governance**: Before process creation, `ProcessPlatform` executes validation gates to reject path traversals, cleanses the environment via `EnvScrubber`, and translates the policy to the target kernel sandbox engine (Landlock on Linux, Seatbelt on macOS, AppContainer on Windows).

### Build and Packaging Changes

* **In-Tree Rust Static Library (`src/rust/nono-c`)**:
  - Compiles an in-tree static C FFI library (`libnono_c.a` / `nono_c.lib`) wrapping `nono` (0.73).
  - Automatically honors vendored crates (`vendor/`) when present to support hermetic, air-gapped distro packaging.
* **C++ Stub Fallback (`-DLEMON_USE_NONO_RUST=OFF`)**:
  - Distro builders without a Rust toolchain cleanly fall back to `nono_stub.cpp`, preserving secret scrubbing while building standard `.deb`, `.rpm`, and Windows packages without Rust dependencies.
* **Target Packaging Hygiene**:
  - The macOS trampoline binary (`lemonade-sandbox-exec`) is conditionally built and packaged only on macOS (`if(APPLE)`).
  - Linux and Windows packages continue to ship only the single `lemond` server executable.

### Key Limitations and Tradeoffs

* **Linux Kernel Version Constraints**: Complete outbound network egress restriction via Landlock requires Landlock ABI v4+ (Linux kernel 6.7+). On older kernels (5.13 to 6.6), `SandboxMode::Auto` degrades gracefully to filesystem-only Landlock isolation and environment allowlisting.
* **macOS Egress Granularity**: macOS Seatbelt enforces strict filesystem and IPC boundaries, but per-port socket filtering is coarser compared to Linux Landlock ABI v4.
* **Windows NTFS ACL Lifecycles**: Native Windows AppContainer requires transient NTFS DACL grants on model cache directories. While managed via RAII (`AclGrantGuard`), abnormal termination of the host daemon may leave lingering AppContainer package ACEs.
* **JIT & Shader Cache Permissions**: Backends with runtime JIT compilers (e.g. Triton in vLLM, MIOpen, Mesa shader caches) require explicit writable directory grants declared via `customize_sandbox_policy` or model recipe options.
* **Live Hub Downloads (`hf_load`)**: Direct model streaming from Hugging Face requires full outbound network access and write access to `~/.cache/llama.cpp`, which must be explicitly requested via recipe configuration.

### Binary Size and Disk Footprint Impact

The following measurements reflect the exact stripped binary size of `lemond` on Linux (x86_64 Release build) across build configurations:

| Configuration | Stripped Binary Size | Net Delta vs Baseline | Notes |
| :--- | :---: | :---: | :--- |
| **Baseline (`LEMON_USE_NONO_RUST=OFF`)** | **10.01 MB** (`10,495,432 B`) | Baseline ($0$) | Pure C++ stub (`nono_stub.cpp`), zero Rust dependencies. |
| **`LEMON_USE_NONO_RUST=ON` (Default Cargo)** | **30.35 MB** (`31,829,304 B`) | **+20.34 MB** | Default `cargo build --release` staticlib. |
| **`LEMON_USE_NONO_RUST=ON` (Optimized Release Profile)** | **22.22 MB** (`23,301,464 B`) | **+12.21 MB** | With `opt-level = "z"`, `lto = true`, `codegen-units = 1`, `panic = "abort"`, and `strip = true` (saves **8.5 MB**). |
| **Windows Native (MSVC)** | — | **< 100 KB** | Pure C++ Win32 (`AppContainer` + `Job Objects`), zero Rust footprint. |

For comprehensive technical documentation, refer to:
* `docs/guide/sandboxing.md`: User, Operator & Administrator Sandboxing Guide
* `docs/dev/sandboxing-and-capabilities.md`: Developer & Backend Architecture Reference
* `docs/dev/adding-a-backend.md`: Backend Developer Onboarding Guide


docs/dev/getting-started.md
```diff
@@ -50,6 +50,7 @@ Lemonade consists of these main executables:
 **All Platforms:**
 - CMake 3.28 or higher
 - C++17 compatible compiler
+- Rust toolchain (Cargo & rustc 1.77+, for in-tree kernel sandboxing FFI and the desktop app)
```
superm1, line 53: is rustc 1.7.7 a hard sell?  because Ubuntu 24.04 doesn't have that new of a rustc in the repos.

https://launchpad.net/ubuntu/+source/rustc

abn: We should be okay with 1.75+.

CMakeLists.txt
```diff
@@ -203,6 +203,34 @@ set(WEB_APP_BUILD_DIR "${CMAKE_BINARY_DIR}/resources/web-app")
 # instead of installing project-local dependencies with npm for web-app builds.
 option(USE_SYSTEM_NODEJS_MODULES "Use system-installed Node.js modules for building the Web app" OFF)
 
+# Build and link the in-tree nono Rust C FFI static library by default when cargo is available.
+find_program(CARGO_EXECUTABLE cargo)
+option(LEMON_USE_NONO_RUST "Link nono Rust C FFI library (libnono_c)" ON)
```
superm1, line 209: I think we're going to want this mandatory, right?

src/rust/nono-c/src/lib.rs
superm1, line 1: When adding in this source, we probably need to update our LICENSE to reflect it

superm1:
I think at some point in the source package build process `cargo vendor` needs to be called to pull down all the crates if they're missing.  Otherwise this is going to fail when it goes to a real PPA (we can't pull down crates dynamically during build).

CMakeLists.txt
abn: Not necessarily, but it can be. Right now it just means you have a degraded sandbox - ie. just static things like scrubbing env var. Native windows builds will also not need them as the windows sandbox is something we rolled our own - atleast until we can offload to nono.

superm1: Okay, I guess that's fine.

abn:
Thanks for the review @superm1!

1. **Third-Party License Notice**:
   - Added `nono` (dual-licensed MIT / Apache-2.0) under `THIRD-PARTY NOTICES` in `LICENSE`.

2. **Offline Source Package Vendoring for PPA / Distros**:
   - Updated `.github/actions/prepare-debian-build/action.yml` to run `cargo vendor` during source package preparation. The generated `.orig.tar.xz` source tarball will contain `src/rust/nono-c/vendor/`.
   - Added `cargo` to `Build-Depends` in `contrib/debian/control`.
   - Updated `CMakeLists.txt` so that whenever `src/rust/nono-c/vendor/` is present in the source tree, CMake automatically passes `--config source.crates-io.replace-with="vendored-sources"` and `--offline` to `cargo build`.
   - In ordinary git clones (where `vendor/` is ignored by `.gitignore`), Cargo builds dynamically from `crates.io` over the network as usual.
   - For minimal builder environments without a Rust toolchain, passing `-DLEMON_USE_NONO_RUST=OFF` continues to compile cleanly using the in-tree fallback stub (`nono_stub.cpp`).

docs/dev/getting-started.md
abn: With recent changes I think 1.95+ is required for build. 

We could create a tool chain ppa with 1.95 and chain it to the lemonade package one. 

But otherwise, right now sandbox just operates in degraded state.

superm1: Or I guess as long as we're making it not mandatory distros with older rust tool chain would be degraded.

jeremyfowers:
Before we get too much further here @abn can you please share an impact analysis:
1. Has the size on disk of lemond been materially impacted by this change? It looks like yes, but please share by exactly how much, what that is going to, and whether it can be optimized down.
2. What is the UI/UX impact of the sandboxing? Will day-to-day lemonade users notice this change? Do the backends lose any functionality that was present before?

abn:
@jeremyfowers, here is the impact analysis:

### 1. Disk and Binary Size Impact (Empirically Measured on Linux x86_64 Release)

* **Baseline (`LEMON_USE_NONO_RUST=OFF`)**: Stripped `lemond` is **10.01 MB** (`10,495,432` bytes).
* **With `libnono_c.a` (Default `nono` 0.73)**: Stripped `lemond` was **30.35 MB** (+20.34 MB delta).
* **With Cargo Release Profile Optimizations**: By configuring `opt-level = "z"`, `lto = true`, `panic = "abort"`, and `strip = true` in `src/rust/nono-c/Cargo.toml` (pushed in latest commit), stripped `lemond` drops to **22.22 MB** (+12.21 MB delta, an **8.5 MB reduction** over default).
* **What that goes to**: `nono` 0.73 embeds Landlock LSM/seccomp ruleset generation along with its default capability stack (crypto verification, keyring, and ICU tables).
* **Windows (MSVC)**: Adds **< 100 KB** (Windows sandboxing is implemented natively in C++ using Win32 AppContainer, Job Objects, and ACL manipulation with zero Rust dependencies).
* **Distro / PPA Builds**: Targets building with `-DLEMON_USE_NONO_RUST=OFF` add **0 KB** Rust overhead.

### 2. UI/UX & Backend Functionality Impact

* **Day-to-day UX**: Completely invisible and frictionless. Users do not need to configure policies or approve permissions. The server automatically constructs the positive capability set (model file path, hardware device nodes, and loopback port) at spawn time.
* **Backend Functionality & Acceleration**: **No functionality is lost**.
  - Hardware acceleration on Linux (`/dev/dri`, `/dev/kfd`, `/dev/accel*`, `/dev/amdxdna`), Windows (DirectX/DirectML/Vulkan driver handles), and macOS (Metal) is explicitly granted via hardware profile presets.
  - All 15 inference backends (`llama.cpp`, `FastFlowLM`, `vLLM`, `RyzenAI`, `whisper.cpp`, `stable-diffusion.cpp`, `Kokoro`, `Moonshine`, etc.) pass CI inference testing under containment.
  - Recipes requiring dynamic downloading (such as `hf_load`) automatically receive network egress and cache write allowances.
* **Security Guarantees Added**:
  - **Secret Scrubbing**: Strips ambient `LEMONADE_*`, cloud provider keys, and system secrets from child process environments.
  - **Token Isolation**: Grants read access strictly to model snapshot files; parent token locations (`~/.cache/huggingface/token`) are blocked.
  - **Default-Deny Egress**: Restricts network access to `LoopbackOnly` so backend subprocesses cannot establish outbound connections.
  - **Graceful Fallback**: If run on legacy kernels without Landlock support, it logs a debug notice and falls back gracefully to unconfined execution with environment scrubbing. An admin escape hatch is also available via `LEMONADE_SANDBOX_MODE=disabled`.

jeremyfowers:
Thanks very much for the impact analysis @abn! We've certainly grown in size since the original embeddable lemond was only 3 MB. Some marketing claims will need to be updated.

Would you do me a favor and add documentation of `LEMON_USE_NONO_RUST=OFF` to https://github.com/lemonade-sdk/lemonade/blob/main/docs/embeddable/building.md

superm1:
> Thanks very much for the impact analysis @abn! We've certainly grown in size since the original embeddable lemond was only 3 MB. Some marketing claims will need to be updated.
> 
> Would you do me a favor and add documentation of `LEMON_USE_NONO_RUST=OFF` to https://github.com/lemonade-sdk/lemonade/blob/main/docs/embeddable/building.md

Do you mean discussing the merits of it?  Or telling people to do this?

"Traditional" packages at least run with systemd confinements.  I think _especially_ in an embedded context where you don't have that you want to be using sandboxing.

abn:
I agree with @superm1 it is definitely valuable in the embedded context. There is a pathway to use landlock ourselves in-tree, that will be smaller but it is a maintenance burden if we go down that path. I had a draft [here](https://github.com/abn/lemonade/blob/3a6a39799fb7c629c6bcaa78b27d718fdefc6c8f/src/cpp/server/sandbox/sandbox_engine.cpp), but it will be a lot of work going forward to keep it secure - work that the nono team is already doing.

Either way happy to add/link documentation with context and impact.

superm1:
As part of your research did you look into just using `bubblewrap` to launch backends and weigh that out against the nono integration?

(I'm not advocating for `bubblewrap` I'm just making sure this was evaluated and the trade off of the nono integration code is worth it)

abn:
> As part of your research did you look into just using `bubblewrap` to launch backends and weigh that out against the nono integration?

I am familiar with `bubblewrap` as it is what is flatpak builds on top of, I am a big fan :smile:.  The reason why I discarded it as an option for us were:

1. It is not cross-platform, it is really Linux only.
2. When comparing against nono, it is a level lower than we would want it. What nono brings to the table is the "dx" we can integrate against - on particular win is the declarative policies. `bubblewrap` would really compare more as a "backend" for nono on Linux.

Hope that helps.

sofiageo:
I hadn't noticed this PR was about nono, just wanted to add that I've been using nono with goose for some weeks now and it works great, and I feel pretty safe to leave my agents unattended. I had planned to test other solutions like https://github.com/fencesandbox/fence which is based on bubblewrap but didn't have the chance to compare yet. So +1 for nono

PR #3217, opened 2026-08-18
fix(rocm): only stage amdhip64_7.dll when TheRock runtime is newer than System32

### LINKED ISSUE #3213: ROCm backend does not detect GPU on Windows when AMD driver ships a newer HIP runtime than TheRock
### Platform

Windows

### Lemonade Version

11.6.0

### GPU / APU Model

AMD Ryzen AI MAX+ 395 w/ Radeon 8060S (Strix Halo, gfx1151)

### Component

llama.cpp

### Bug Description

ROCm backend does not detect the GPU on Windows when the installed AMD driver ships a **newer** HIP runtime than the bundled TheRock runtime.

PR #2768 added an unconditional force-copy of `amdhip64_7.dll` from TheRock into the backend's exe directory (`llamacpp\rocm-stable\`) to work around System32 DLL shadowing (Windows loads DLLs from the exe dir before System32, which is before PATH). On machines whose driver ships a newer runtime, that copy **downgrades** the loaded HIP runtime:

| `amdhip64_7.dll` | Version | Source |
|---|---|---|
| Copied by lemond into `llamacpp\rocm-stable\` | **10.0.3581** | TheRock 7.14.0 |
| In `System32` (from installed AMD driver 32.0.31035.1003) | **10.0.3679** | newer, driver-matched |

The stale copy wins the DLL search, and `hipGetDeviceCount()` returns 0 → `ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected`. llama.cpp then silently falls back to CPU.

Verification: deleting the copied `amdhip64_7.dll` from `llamacpp\rocm-stable\` makes ROCm detect `gfx1151` (102 GiB) and inference works (~47 tok/s).

### Steps to Reproduce

1. Fresh AMD driver (newer than the bundled TheRock 7.14 runtime) on Windows.
2. Install the `llamacpp:rocm` backend.
3. Load any ROCm model (e.g. `Ornith-1.0-35B-GGUF-Q8_0`) with `llamacpp_backend=rocm`.
4. Observe llama-server log: `ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected`; `/api/v1/system-info` shows no ROCm device.
5. Confirm a stale `amdhip64_7.dll` (10.0.3581) was copied into `bin\llamacpp\rocm-stable\`.

Expected: ROCm backend uses the driver-matched HIP runtime from System32 and detects the GPU.

Actual: lemond copies the older TheRock `amdhip64_7.dll` next to the exe, shadowing the newer System32 runtime, and ROCm reports no device.

### Log Output

```
[Info] (LlamaCpp) Using LlamaCpp Backend: rocm-stable
ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected
Available devices: (none)
```

(With debug logging, the backend load also logs `Copied amdhip64_7.dll from TheRock to ...`.)

### Additional Context

Root cause: the force-copy introduced in #2768 (commit `0e1eb208`) only handles the case where System32 is *older* than TheRock. When the driver is *newer*, the copy shadows the correct runtime. A fix on branch `fix/rocm-amdhip64-dll-version-staging` makes staging version-aware (compare `VS_FIXEDFILEINFO` of System32 vs TheRock; skip the copy and remove any previously staged one when System32 >= TheRock). Verified on the hardware above: with the fix, `amdhip64_7.dll` (10.0.3679) loads from System32, GPU is detected, and inference runs at ~47 tok/s.

blackdeathdrow:
## Summary

PR #2768 force-copies TheRock's `amdhip64_7.dll` into the backend exe directory unconditionally. When the AMD driver ships a newer HIP runtime in System32 (e.g. 10.0.3679 vs TheRock's 10.0.3581), the stale copy shadows it: Windows loads DLLs from the exe dir before System32, so llama-server's `hipGetDeviceCount()` returns 0 and ROCm reports "no ROCm-capable device is detected".
This PR makes staging version-aware: it copies System32's `amdhip64_7.dll` into the exe dir first, reads its version from that plain path (GetFileVersionInfoW returns stale cached data for System32 paths), and only overwrites it with TheRock's when TheRock is newer. The exe dir always ends up with the correct runtime, so the loaded DLL is always known.

Fixes #3213

The staging is only applied to llama.cpp.

## Scope

## Testing

Testing details:
- Built with `cmake --build build --preset vs18 --target wix_installer_minimal` — MSI built successfully.
- New unit test `BackendUtilsDllVersionTest` passes via `ctest --test-dir build -C Release -R BackendUtils` (verifies `read_dll_version` matches the direct API on a plain path and handles missing files).
- On Strix Halo (gfx1151): with System32 HIP runtime 10.0.3679 (newer than TheRock's 10.0.3581), the staged System32 copy is used and ROCm detects the GPU — inference at ~47.4 t/s (matches ROCm benchmark; Vulkan was ~50.4 t/s).

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2956,6 +2956,13 @@ if(BUILD_TESTING AND EXISTS "${_GPU_ENV_TEST_SRC}")
     add_cpp_ci_test(BackendUtilsGpuEnvTest CI ON COMMAND test_backend_utils_gpu_env)
 endif()
 
+set(_DLL_VERSION_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_backend_utils_dll_version.cpp")
+if(BUILD_TESTING AND EXISTS "${_DLL_VERSION_TEST_SRC}")
+    add_executable(test_backend_utils_dll_version test/cpp/test_backend_utils_dll_version.cpp)
+    target_link_libraries(test_backend_utils_dll_version PRIVATE lemonade-server-core)
+    add_cpp_ci_test(BackendUtilsDllVersionTest CI ON COMMAND test_backend_utils_dll_version)
+endif()
+
 set(_PIP_DL_PARSER_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_pip_download_parser.cpp")
 if(BUILD_TESTING AND EXISTS "${_PIP_DL_PARSER_TEST_SRC}")
     add_executable(test_pip_download_parser test/cpp/test_pip_download_parser.cpp)
--- a/src/cpp/include/lemon/backends/backend_utils.h
+++ b/src/cpp/include/lemon/backends/backend_utils.h
@@ -192,6 +192,16 @@ namespace lemon::backends {
          *  get_therock_lib_paths). Returns "" for empty input. */
         static std::string join_runtime_dirs(const std::vector<std::string>& dirs);
 
+        /** Stage TheRock's amdhip64_7.dll next to a ROCm backend exe, unless
+         *  System32 already ships a newer runtime. No-op off Windows. */
+        static bool stage_therock_hip_runtime(const std::string& rocm_arch,
+                                              const fs::path& target_dir);
+
+        /** Read a DLL's VS_FIXEDFILEINFO version (packed MS<<32|LS). Windows
+         *  only; returns 0 elsewhere. GetFileVersionInfoW returns stale data
+         *  for system32 paths, so callers read versions from plain paths. */
+        static uint64_t read_dll_version(const fs::path& dll);
+
         /** Get the path to the backend's binary. Gives precedence to the path set through environment variables, if set. Throws if not found. */
         static std::string get_backend_binary_path(const BackendSpec& spec, const std::string& backend);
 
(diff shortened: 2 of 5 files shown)


blackdeathdrow:
@superm1 Could you please take a look? With this change it works, GPU is detected. Otherwise it falls back to CPU.

superm1:
Did you test this (presumably) on real hardware and confirmed the failure previously and the pass after?

blackdeathdrow:
Exactly. I updated from origin/main, and the model using rocm degraded to 6 tps. After that, I made a fix, and now it runs at ~50 tps.

superm1 (approved):

superm1:
OK thanks.  Will land this then, much appreciated.

superm1:
I think that CI failure might be caused by your change, I'll try once more.

superm1:
Yeah I'm pretty sure your changes caused a bug, can you please take a look.

blackdeathdrow:
Apologies for the bug. I've restricted the fix to llama.cpp only. Also, I'm concerned this might be specific to my hardware. Could you re-run the CI to check?

PR #3218, opened 2026-08-18
fix(gui3): pin storage meter to the bottom of the models rail

falabellamichael:
## Summary

Pin the storage meter in the Models rail so it stays fixed at the bottom of the rail instead of moving when the filter sections expand or collapse.

The meter was rendered inside `.model-nav-rail__scroll`, so toggling the Task / Backends / Tags sections shifted its position (and it could scroll out of view). This moves it out of the scroll container into a pinned footer, and gives the scroll area `flex: 1` so the filters scroll independently above it. The meter stays hidden when the rail is collapsed (the existing `.workspace-rail.is-collapsed` rule covers it).

## Scope

## Testing

Built the web-app from source (`cmake --build build --config Release --target web-app`) and verified in the browser that the storage meter stays pinned at the bottom of the rail while the filter sections scroll independently. Also rebuilt the Tauri desktop app and confirmed it compiles and launches.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/src/app/src/components/ModelNavRail.tsx
+++ b/src/app/src/components/ModelNavRail.tsx
@@ -524,7 +524,9 @@ export const ModelNavRail: React.FC<ModelNavRailProps> = ({
         )}
       </section>
 
-      {/* 5. Storage meter */}
+      </div>
+
+      {/* 5. Storage meter — pinned footer (stays at the bottom of the rail) */}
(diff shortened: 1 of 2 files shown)


jeremyfowers (approved):
nice improvement, thank you!

PR #3219, opened 2026-08-18
feat(server): hide backends with no runnable models on this host

ramkrishna2910:
## Summary

When every one of a backend's built-in models is dropped by the system-memory filter, the backend has nothing runnable on this host — yet it was still listed as `installed`/`available` in the recipes/backends output (e.g. a large single-model backend on a smaller-RAM box). This hides such recipes from the `/system-info` listing, which is the single source for the CLI `backends` command, the web BackendManager, and `pull` backend-selection.

Scoped deliberately:
- Only recipes dropped by the **memory** heuristic are hidden. Recipes filtered for hardware/OS reasons keep their existing "unsupported" display.
- **Partial** filtering keeps the backend visible (≥1 model, built-in or user, still runnable).
- **Dynamic-model** backends (`cloud`, `flm`) are never hidden — they supply models at runtime.
- The load-time "requires N GB, your system has M GB" error is unaffected (still returned if the model is named directly).

## Scope

## Testing

_Testing details:_

- New unit test `test/cpp/test_recipe_hiding.cpp` (`RecipeHidingTest`, CI ON) covers the pure rule: all-filtered → hidden, partial → kept, not-size-filtered → never hidden, mixed recipes independent.
- End-to-end on a live server by inflating registry model sizes to force the memory filter:
  - single-model backend (`kokoro`) with its only model filtered → **hidden** from `/system-info`, `lemonade backends`, and `/models`.
  - multi-model backend (`llamacpp`) with 1 of 90 models filtered → **still listed**.
  - `flm` (dynamic_models) → **never hidden**.
  - loading the filtered model still returns the clear "requires N GB" error.
- Existing `ModelManagerCollectionValidationTest` still passes.

## Documentation

## Breaking Changes

## AI-assisted contribution

Implemented with Claude Code (Opus 4.8). Design, tests, and end-to-end verification reviewed by the author.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2805,6 +2805,18 @@ if(BUILD_TESTING AND EXISTS "${_MODEL_MANAGER_COLLECTION_VALIDATION_TEST_SRC}")
     add_cpp_ci_test(ModelManagerCollectionValidationTest CI ON COMMAND test_model_manager_collection_validation)
 endif()
 
+# Recipe hiding: a backend with every built-in model dropped by the system-memory
+# heuristic (and no user model left) is hidden from the recipe/backends listing.
+set(_RECIPE_HIDING_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_recipe_hiding.cpp")
+if(BUILD_TESTING AND EXISTS "${_RECIPE_HIDING_TEST_SRC}")
+    add_executable(test_recipe_hiding test/cpp/test_recipe_hiding.cpp)
+    target_link_libraries(test_recipe_hiding PRIVATE lemonade-server-core)
+    add_dependencies(test_recipe_hiding copy_resources)
+
+    include(CTest)
+    add_cpp_ci_test(RecipeHidingTest CI ON COMMAND test_recipe_hiding)
+endif()
+
 # Remote model registry source parsing, cache namespaces, snapshot fingerprints,
 # and persistence of ModelScope provenance in ModelInfo.
 set(_MODEL_REGISTRY_TEST_SRC
--- a/src/app/src/renderer/BackendManager.tsx
+++ b/src/app/src/renderer/BackendManager.tsx
@@ -141,9 +141,18 @@ const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError,
     window.open(url, '_blank', 'noopener,noreferrer');
   }, []);
 
+  // Mirrors BackendRow's `isInstalledLike`: something is on disk here, so this
+  // panel must keep showing it. It is the only place that offers uninstall
+  // (ModelManager renders the 'banner' variant, which has no uninstall action),
+  // so hiding an unavailable-but-installed backend would strand it.
+  const isInstalledLike = (info: BackendInfo) =>
+    info.state === 'installed' || info.state === 'update_available';
+
   const groupedBackends: Array<[string, Array<[string, BackendInfo]>]> = recipes
     ? Object.entries(recipes)
-      .filter(([recipeName]) => !unavailableRecipes.has(recipeName))
+      .filter(([recipeName, recipe]: [string, Recipe]) =>
+        !unavailableRecipes.has(recipeName) ||
+        Object.values(recipe.backends).some(isInstalledLike))
       .map(([recipeName, recipe]: [string, Recipe]) => {
         const backends = Object.entries(recipe.backends).filter(([, info]) => info.state !== 'unsupported');
         return [recipeName, backends] as [string, Array<[string, BackendInfo]>];
--- a/src/app/src/renderer/BackendManager.tsx
+++ b/src/app/src/renderer/BackendManager.tsx
@@ -47,6 +47,10 @@ const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError,
   }, [refresh]);
 
   const recipes = systemInfo?.recipes;
+  // `recipes` is canonical (docs are generated from it), so host-specific
+  // availability arrives separately: these have no model that fits this
+  // machine's memory, so there is nothing to install them for.
+  const unavailableRecipes = new Set(systemInfo?.unavailable_recipes ?? []);
 
   // Fetch asset sizes from GitHub Releases API
   useEffect(() => {
@@ -139,6 +143,7 @@ const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError,
 
   const groupedBackends: Array<[string, Array<[string, BackendInfo]>]> = recipes
     ? Object.entries(recipes)
+      .filter(([recipeName]) => !unavailableRecipes.has(recipeName))
       .map(([recipeName, recipe]: [string, Recipe]) => {
         const backends = Object.entries(recipe.backends).filter(([, info]) => info.state !== 'unsupported');
         return [recipeName, backends] as [string, Array<[string, BackendInfo]>];
--- a/src/app/src/renderer/BackendManager.tsx
+++ b/src/app/src/renderer/BackendManager.tsx
@@ -141,20 +141,29 @@ const BackendManager: React.FC<BackendManagerProps> = ({ searchQuery, showError,
     window.open(url, '_blank', 'noopener,noreferrer');
   }, []);
 
-  // Mirrors BackendRow's `isInstalledLike`: something is on disk here, so this
-  // panel must keep showing it. It is the only place that offers uninstall
-  // (ModelManager renders the 'banner' variant, which has no uninstall action),
-  // so hiding an unavailable-but-installed backend would strand it.
-  const isInstalledLike = (info: BackendInfo) =>
-    info.state === 'installed' || info.state === 'update_available';
+  // Matches the server's own `locally_installed` (system_info.cpp): these three
+  // states all mean the backend is on disk, `update_required` included.
+  const isLocallyInstalled = (info: BackendInfo) =>
+    info.state === 'installed' ||
+    info.state === 'update_available' ||
+    info.state === 'update_required';
+
+  // An unavailable recipe has no built-in model that fits this host, so it keeps
+  // only rows for an existing installation: this panel is the only one offering
+  // uninstall (ModelManager renders the 'banner' variant, which has none), so
+  // hiding those would strand them. `installable` and `action_required` are
+  // dropped — nothing is on disk yet, and setting one up would buy the user
+  // nothing. A recipe left with no rows falls out via the length check below.
+  const keepBackend = (recipeName: string) =>
+    unavailableRecipes.has(recipeName)
+      ? isLocallyInstalled
+      : (info: BackendInfo) => info.state !== 'unsupported';
 
   const groupedBackends: Array<[string, Array<[string, BackendInfo]>]> = recipes
     ? Object.entries(recipes)
-      .filter(([recipeName, recipe]: [string, Recipe]) =>
-        !unavailableRecipes.has(recipeName) ||
-        Object.values(recipe.backends).some(isInstalledLike))
       .map(([recipeName, recipe]: [string, Recipe]) => {
-        const backends = Object.entries(recipe.backends).filter(([, info]) => info.state !== 'unsupported');
+        const keep = keepBackend(recipeName);
+        const backends = Object.entries(recipe.backends).filter(([, info]) => keep(info));
         return [recipeName, backends] as [string, Array<[string, BackendInfo]>];
       })
       .filter(([, backends]) => backends.length > 0)
--- a/src/app/src/renderer/utils/systemData.ts
+++ b/src/app/src/renderer/utils/systemData.ts
@@ -4,6 +4,10 @@ export interface SystemInfo {
   processor: string;
   devices: Devices;
   recipes?: Recipes;
+  // Recipes with nothing runnable on this host: every built-in model was
+  // dropped by the system-memory heuristic. `recipes` stays canonical, so
+  // listings that reflect this host must subtract this set themselves.
+  unavailable_recipes?: string[];
 }
 
 interface Devices {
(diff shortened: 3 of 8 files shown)


fl0rianr (commented):
I think the overall direction makes sense, but there are two issues I’d fix before merging. 

First, /system-info is now host-dependent because recipes are erased based on the local RAM filter, while the docs generator uses /system-info recipes as the source for README/docs/models.js. That makes generated docs depend on the machine running the generator and appears related to the current Docs And Style failure.

Second, filter_models_by_backend() now mutates recipes_all_models_filtered_ on every call, but it is also called from add_model_to_cache() with a single-model temp map. That can overwrite the correct full-cache result with state derived from only one model. The disable_model_filtering early return also clears filtered_out_models_ but leaves recipes_all_models_filtered_ stale. I’d keep the canonical recipe list separate from host-specific availability, and only compute/commit recipes_all_models_filtered_ during a full cache build. The new test currently only covers the pure set-difference helper, so adding stateful coverage for full-cache → incremental update, disable_model_filtering, and the /system-info behavior would catch these cases.

ramkrishna2910:
Thanks, both good catches — fixed in 9eb9d3d.

**1. `/system-info` host-dependence / docs drift.** `handle_system_info` no longer prunes `recipes`; that dict stays canonical so the docs generator renders README/models.js identically regardless of the generating machine's memory. Host availability is now a separate `unavailable_recipes` field, and the `lemonade backends` CLI hides those recipes unless `--all` is passed (web can adopt the field later). Verified `gen_backend_boilerplate.py --check` is clean for the recipe regions.

**2. Incremental clobber + `disable_model_filtering` staleness.** `filter_models_by_backend()` now takes `track_recipe_availability`, and only the full-registry pass in `build_cache()` passes `true`. The single-model temp map from `add_model_to_cache()` no longer touches `recipes_all_models_filtered_`, and the `disable_model_filtering` early return clears it.

**Tests.** Added stateful coverage in `test_recipe_hiding.cpp`: a full pass commits and *recomputes* the side table (a now-fitting recipe is dropped), and an incremental single-model pass leaves it unchanged. The pure set-difference cases remain. End-to-end verified by forcing the filter (registry size inflation): `recipes` stays canonical, `unavailable_recipes` lists the backend, `lemonade backends` hides it while `--all` shows it.

One note on `/system-info` behavior in a unit test — it needs a live server with a doctored registry, so that path is covered by the manual e2e run rather than a C++ unit test; happy to add a Python integration test if you'd prefer.

fl0rianr (commented):
Thanks, the two original issues look fixed in the current head, and CI is green now. I think there’s still one remaining gap before merge:

`unavailable_recipes` is exposed by `/system-info` and consumed by the C++ CLI, but the web app does not use it yet. `BackendManager.tsx` still builds its list directly from `systemInfo.recipes` and only filters `state === 'unsupported'`, while `SystemInfo` does not currently define `unavailable_recipes`.

That means a backend whose built-in models are all filtered out by the RAM heuristic will still appear in the Web BackendManager, even though the PR description lists it as one of the consumers this change is intended to fix.

I’d either wire `unavailable_recipes` into the web consumer here (and consider the same signal for `ensureBackendForRecipe`), or narrow the PR scope/description and track the web behavior separately.

Aside from that, the follow-up fixes for canonical `/system-info` recipes and the full-cache vs incremental availability state look good.

ramkrishna2910:
Wired it in — 042eaad.

**Web consumer.** `SystemInfo` now declares `unavailable_recipes?: string[]`, and `BackendManager` subtracts that set before grouping, so a backend whose built-in models are all dropped by the RAM heuristic no longer shows up there. That matches the CLI's default (`lemonade backends`), which still has `--all` as the escape hatch; the web panel has no equivalent toggle today, so it just hides. `tsc --noEmit` on `src/app` is clean.

**`ensureBackendForRecipe` — deliberately not gated.** It runs on an explicit user action for one named model, not as a listing, and `unavailable_recipes` is a statement about *built-in* models only. User models are exempt from the memory filter (`user_controlled_model`), so a recipe can be "unavailable" while the model the user just asked for is perfectly loadable — gating there would block that flow for no gain. Built-in models that *are* filtered never reach this path (they're hidden from `/models`), and if one is named directly the load-time "requires N GB, your system has M GB" error is still the authoritative answer. This is the same split as the CLI: `hf_pull`'s pull preflight checks backend *install state* and ignores `unavailable_recipes` for exactly this reason.

`supportedRecipes` (AddModelPanel / ModelOptionsModal) is left alone on the same grounds — it drives recipe choice when registering a user model.

fl0rianr (commented):
Thanks, the two original issues look fixed. But I think there’s still one remaining gap before merge:

`unavailable_recipes` is exposed by `/system-info` and consumed by the C++ CLI, but the web app does not use it yet. `BackendManager.tsx` still builds its list directly from `systemInfo.recipes` and only filters `state === 'unsupported'`, while `SystemInfo` does not currently define `unavailable_recipes`.

That means a backend whose built-in models are all filtered out by the RAM heuristic will still appear in the Web BackendManager, even though the PR description lists it as one of the consumers this change is intended to fix.

I’d either wire `unavailable_recipes` into the web consumer here (and consider the same signal for `ensureBackendForRecipe`), or narrow the PR scope/description and track the web behavior separately.

Aside from that, the follow-up fixes for canonical `/system-info` recipes and the full-cache vs incremental availability state look good.

ramkrishna2910:
Both claims in this review are already false on the current head. `042eaad` landed at 18:12:37Z and this review was submitted at 18:21:16Z, so I believe it was drafted against the previous head.

At `042eaad`:

**`SystemInfo` does define it** — `src/app/src/renderer/utils/systemData.ts:10`:

```ts
recipes?: Recipes;
// Recipes with nothing runnable on this host: every built-in model was
// dropped by the system-memory heuristic. `recipes` stays canonical, so
// listings that reflect this host must subtract this set themselves.
unavailable_recipes?: string[];
```

**`BackendManager.tsx` does consume it** — read at `:53`, subtracted at `:146`, ahead of the `state !== 'unsupported'` filter at `:148`:

```ts
const unavailableRecipes = new Set(systemInfo?.unavailable_recipes ?? []);
...
const groupedBackends = recipes
  ? Object.entries(recipes)
    .filter(([recipeName]) => !unavailableRecipes.has(recipeName))
    .map(([recipeName, recipe]) => { ... })
```

So a backend whose built-in models are all dropped by the RAM heuristic is hidden from the web BackendManager, matching `lemonade backends`. The CLI keeps `--all` as an escape hatch; the web panel has no equivalent toggle today, so it just hides. `tsc --noEmit` on `src/app` is clean, and CI is green on this head.

For what it's worth, this review is byte-identical to #pullrequestreview-4977973213 except for the first sentence, which is what suggests a re-run against a stale base rather than new feedback.

The one sub-point that *is* still open is `ensureBackendForRecipe`, which I answered in https://github.com/lemonade-sdk/lemonade/pull/3219#issuecomment-5359892614 and deliberately left ungated: `unavailable_recipes` describes *built-in* models only, user models are exempt from the memory filter (`user_controlled_model`), so a recipe can be "unavailable" while the model the user explicitly named loads fine. Gating there would block that flow for no gain, and it mirrors `hf_pull`'s pull preflight, which checks backend install state and ignores `unavailable_recipes` for the same reason.

Could you re-review against `042eaad`? If you still want `ensureBackendForRecipe` gated after the user-model point above, I'll make that change.

fl0rianr (commented):
Thanks, I'm sorry for the confusion, I re-reviewed this against the current head 042eaad. The web consumer issue is fixed, and I agree `ensureBackendForRecipe` should stay ungated for explicit/user-model flows.

I do see one remaining functional issue: `BackendManager` now filters out an unavailable recipe entirely. If that backend is already installed, its row disappears too, which also removes the web UI’s uninstall action. The CLI has `--all` as an escape hatch, but the web panel currently has no equivalent. I think installed/uninstallable backends should remain reachable somehow, even if the recipe is unavailable for built-in models.

One smaller test concern: the new stateful test never positively asserts that the seeded unavailable recipe was actually added to the side table. If the full-pass commit were broken and the set stayed empty, the incremental “unchanged” check could still pass. A positive assertion on the seeded recipe would make that coverage much stronger.

Aside from those points, the current approach looks consistent.

ramkrishna2910:
Both fixed in b0dd893.

**1. Installed backends no longer disappear.** Confirmed the regression: `BackendManager` is the only caller that renders `ConnectedBackendRow` with `variant='full'`, and `onUninstall` is passed only when `isFull` — so dropping the group really did remove the sole web path to uninstall. The group is now kept whenever any of its backends is installed-like, mirroring `BackendRow`'s own predicate so the two can't drift:

```ts
// Mirrors BackendRow's `isInstalledLike`: something is on disk here, so this
// panel must keep showing it. It is the only place that offers uninstall
// (ModelManager renders the 'banner' variant, which has no uninstall action),
// so hiding an unavailable-but-installed backend would strand it.
const isInstalledLike = (info: BackendInfo) =>
  info.state === 'installed' || info.state === 'update_available';

.filter(([recipeName, recipe]) =>
  !unavailableRecipes.has(recipeName) ||
  Object.values(recipe.backends).some(isInstalledLike))
```

Net effect: an unavailable recipe is hidden only while it's *not* installed — i.e. we stop advertising something you'd have no use for, but never hide something already on your disk. I deliberately used `isInstalledLike` rather than replicating the full `canUninstall` predicate (`can_uninstall !== false && backendName !== 'system'`), since the rule that matters is "present on this machine", not "the uninstall button happens to be enabled".

**2. Positive assertion added.** Your read was right — `after_incremental == seeded` was `{} == {}` if the full pass committed nothing. The stateful tests now probe for a recipe that actually reaches the memory heuristic (a tiny model must survive and a huge one must not, so neither host support nor RAM readability is assumed) and assert membership directly.

I mutation-tested it by stubbing the commit to `if (false)`:

```
FAIL full pass records a recipe whose every model was size-filtered
PASS full pass recomputes: partially-filtered recipe is dropped
PASS full pass never records a recipe that kept a visible model
FAIL seeded side table actually holds the all-filtered recipe
PASS incremental pass leaves the availability side table unchanged   <-- the vacuous one you spotted
FAIL incremental pass does not drop the recorded recipe
```

The check you flagged still passes under the broken build; the three new ones catch it. Reverted, and all 11 checks pass (`exit=0`, probe recipe `llamacpp`). If no recipe reaches the heuristic on a given host the test now fails loudly rather than skipping, since a silent skip is the same vacuous pass in a different costume.

`tsc --noEmit` on `src/app` is clean.

One thing I did *not* do: annotate the retained rows with why the recipe is unavailable ("no built-in model fits this system's memory"). It'd be a genuine UX improvement, but it changes displayed copy on installed backends, so I left it out of this PR's scope — happy to add it here or as a follow-up if you'd prefer it in.

fl0rianr (commented):
Thanks, both previous points are fixed in b0dd893. I found one remaining edge case in the web filtering.

For an unavailable recipe, `BackendManager` currently keeps the whole recipe group if any backend is `installed` or `update_available`. That has two side effects:

* `update_required` is not included, even though the server treats it as locally installed and the UI provides an update/remediation action for it. An unavailable recipe in that state can still disappear completely.
* Once one backend keeps the group alive, all other non-unsupported backends are rendered too, including `installable` variants. That means the UI can advertise/install another backend variant for a recipe that is unavailable specifically because none of its built-in models fit this host.

I think the cleanest fix is to filter at the backend-row level for unavailable recipes: keep only rows representing an existing installation or required remediation (`installed`, `update_available`, `update_required`), instead of retaining the entire recipe group.

Other than that, the current head looks good to me.

ramkrishna2910:
Fixed in 5b4198c, using the row-level approach you suggested — it's the better shape and it fixes both points at once.

Rather than patch the two symptoms, I went back to the server for the authoritative definition. `system_info.cpp:1598` already has one:

```cpp
const bool locally_installed = effective_state == "installed"
    || effective_state == "update_available"
    || effective_state == "update_required";
```

So `update_required` is a local installation by the server's own reckoning, and my predicate had diverged from it. The web side now matches that definition exactly:

```ts
// Matches the server's own `locally_installed` (system_info.cpp): these three
// states all mean the backend is on disk, `update_required` included.
const isLocallyInstalled = (info: BackendInfo) =>
  info.state === 'installed' ||
  info.state === 'update_available' ||
  info.state === 'update_required';

const keepBackend = (recipeName: string) =>
  unavailableRecipes.has(recipeName)
    ? isLocallyInstalled
    : (info: BackendInfo) => info.state !== 'unsupported';
```

The group-level filter is gone entirely; a recipe left with no rows now falls out through the existing `backends.length > 0` check, so the common "unavailable and nothing installed" case still hides the whole group.

I enumerated the full state space this time instead of handling only the states named, and verified it by running the two predicates — extracted verbatim from the component — over all six states in both cases:

| State | Available recipe | Unavailable recipe |
|---|---|---|
| `unsupported` | hidden | hidden |
| `installable` | shown | **hidden** ← your point 2 |
| `action_required` | shown | **hidden** ← same class as `installable` |
| `update_required` | shown | **shown** ← your point 1 |
| `installed` | shown | shown |
| `update_available` | shown | shown |

`action_required` isn't in the server's `locally_installed` set — nothing is on disk, it's a "follow the setup guide" prompt — so it's dropped alongside `installable` on the same reasoning: no reason to walk someone through setting up a backend that has no runnable models. Nothing is stranded by dropping it, since there's nothing installed to manage.

Behavior for available recipes is byte-for-byte unchanged. `tsc --noEmit` on `src/app` is clean.

fl0rianr (approved):
Re-reviewed the current head 5b4198c. The row-level filtering looks right now: `update_required` is kept consistently with the server's `locally_installed` definition, while `installable`/`action_required` siblings no longer leak back into unavailable recipes. The stateful test coverage also looks much stronger now, and CI is green.

I don't see any remaining code blockers from my side.

**One small non-blocking cleanup:** the PR description still says recipes are hidden from `/system-info` and mentions `pull` backend-selection as a consumer. The final design keeps `recipes` canonical, exposes `unavailable_recipes` separately, and deliberately leaves explicit/user-model pull flows ungated. Might be worth updating the summary/testing section so it matches the final behavior.

Otherwise, LGTM.

PR #3222, opened 2026-08-18
feat(cloud): configurable auth header name and prefix per provider

### LINKED ISSUE #3220: Support cloud providers with a non-standard auth header name/prefix
### Feature Description

Cloud providers are hardcoded to authenticate with `Authorization: Bearer <key>`. Some OpenAI-compatible endpoints — API gateways in particular — expect the key in a differently named header (e.g. `X-Api-Key`), or with no prefix at all. Those providers can be installed today, but every request to them fails auth upstream.

Proposal: two optional per-provider settings, `auth_header_name` (default `Authorization`) and `auth_header_prefix` (default `Bearer `), exposed on `/v1/install` and `lemonade cloud install`, and persisted alongside the existing non-secret provider fields.

```bash
lemonade cloud install acme --base-url https://gateway.example.com/v1 \
  --api-key "$ACME_KEY" --auth-header-name X-Api-Key --auth-header-prefix ""
```

Both default to the current values, so existing providers and `config.json` files are unaffected.

### Use Case / Motivation

Organizations frequently put an API gateway in front of one or more OpenAI-compatible model backends, and those gateways commonly authenticate with their own header name rather than `Authorization: Bearer`. Lemonade can already reach such an endpoint's base URL and discover its models — the only blocker is the fixed header shape. This is a small, additive config change that makes a whole class of OpenAI-compatible endpoints reachable.

Key handling would be unchanged: keys stay in `LEMONADE_<PROVIDER>_API_KEY` or `lemond`'s process memory, never on disk (AGENTS.md invariant #11). Only the non-secret header *name* and *prefix* are persisted.

### Platform Relevance

All platforms

### Additional Context

Related to but distinct from #2758, which covers vision/tool-calling/model-discovery gaps for specific named providers.

While implementing this I also hit a latent bug in the `/v1/cloud/auth` insecure-http opt-in: it calls `install(provider, base_url, true)`, which resets every other field on the provider record back to its default. A dedicated setter fixes it.

I have a patch ready and will open a PR against this issue.

jgmelber:
## Summary

Cloud providers are currently hardcoded to authenticate with `Authorization: Bearer <key>`. Some OpenAI-compatible endpoints — API gateways in particular — expect the key in a differently named header, or with no prefix at all, and are unusable today.

This adds two optional per-provider settings, `auth_header_name` (default `Authorization`) and `auth_header_prefix` (default `Bearer `), threaded through the registry, the `/v1/install` cloud branch, and `lemonade cloud install`. Both default to today's values, so every existing provider and every `config.json` on disk behaves exactly as before.

```bash
lemonade cloud install acme --base-url https://gateway.example.com/v1 \
  --api-key "$ACME_KEY" --auth-header-name X-Api-Key --auth-header-prefix ""
```

Also fixes a latent bug in the `/v1/cloud/auth` insecure-http opt-in: it called `install(provider, base_url, true)`, which reset every other field on the record back to its default. It now flips just that one flag via a new `set_allow_insecure_http()`.

Key handling is unchanged — keys still live only in `LEMONADE_<PROVIDER>_API_KEY` or `lemond`'s process memory, never on disk.

Fixes #3220

## Scope

Non-secret fields only are persisted, and defaults are omitted from `config.json` so existing files are untouched until a non-default value is set. The UI (`CloudProvidersSection.tsx`) is deliberately not modified — per AGENTS.md, frontend changes are handled by core maintainers.

## Testing

_Testing details:_

- `cmake --build --preset default` — clean build, no new warnings.
- Extended `test/cpp/test_cloud_provider_registry_config.cpp` (renamed from the previous target name) with 17 checks: defaults, round-trip through `to_config_array()` → `load_from_config()`, an empty prefix surviving serialization, `install()` treating a changed header as a real update rather than a no-op, and `set_allow_insecure_http()` preserving the other fields. The test was also promoted from `CI OFF` to `CI ON`, since the registry is pure logic and needs no backend.
- `ctest -L cpp-ci -j4` — 43/43 pass.
- Live: registered a provider behind a gateway that requires a custom header name and an empty prefix; discovery returned its model list and chat completions round-tripped. Re-verified that a default `Authorization: Bearer` provider is unaffected.

## Documentation

`docs/guide/cli.md` (option rows + example), `docs/api/lemonade.md` (install params, install response, `system-info`), `docs/guide/configuration/README.md` (`cloud_providers` fields), `docs/guide/configuration/cloud.md` (new section on gateways with a non-standard auth header). No generated (`<!-- BEGIN GENERATED -->`) region was touched.

## Breaking Changes

Both fields default to the current behavior; absent config keys load as the defaults.

## AI-assisted contribution


superm1 (commented):
Can you please clean up the findings from the robot?

src/cpp/server/server.cpp
```diff
@@ -7565,11 +7566,17 @@ void Server::handle_install(const httplib::Request& req, httplib::Response& res)
...
+        //    auth_header_prefix: "Bearer ",       // optional, default shown
         //    api_key: "..."}  // optional
         if (request_json.value("backend", "") == "cloud") {
             const std::string provider = request_json.value("provider", "");
             const std::string base_url = request_json.value("base_url", "");
             const std::string api_key = request_json.value("api_key", "");
+            const std::string auth_header_name =
+                request_json.value("auth_header_name", std::string("Authorization"));
+            const std::string auth_header_prefix =
+                request_json.value("auth_header_prefix", std::string("Bearer "));
```
ramkrishna2910: **Re-installing an existing provider silently resets these to the defaults.**

`handle_install` always materializes `Authorization` / `Bearer ` when the fields are absent, and `install()` overwrites the whole record — so a provider previously configured with, say, `x-api-key` / `""` gets clobbered back to `Authorization` / `Bearer ` by any re-install that doesn't repeat the flags.

Triggers in practice:
- `lemonade cloud install <same provider> --base-url ...` without the new flags
- the desktop app's Add-provider form, which posts only `backend` / `provider` / `base_url` / `api_key`

The gateway then 401s with nothing surfacing why. This is the same field-clobbering class the PR fixes for `allow_insecure_http` — suggest reading the existing record first and only overwriting fields that were actually present in the request body (`request_json.contains(...)`), rather than defaulting unconditionally.

src/cpp/server/server.cpp
```diff
@@ -7565,11 +7566,17 @@ void Server::handle_install(const httplib::Request& req, httplib::Response& res)
...
         //    allow_insecure_http: false,
+        //    auth_header_name: "Authorization",   // optional, default shown
+        //    auth_header_prefix: "Bearer ",       // optional, default shown
         //    api_key: "..."}  // optional
         if (request_json.value("backend", "") == "cloud") {
             const std::string provider = request_json.value("provider", "");
             const std::string base_url = request_json.value("base_url", "");
             const std::string api_key = request_json.value("api_key", "");
+            const std::string auth_header_name =
+                request_json.value("auth_header_name", std::string("Authorization"));
```
ramkrishna2910: **`auth_header_name` is accepted unvalidated.**

Unlike the adjacent `provider` and `base_url` validators, there's no type / empty / header-token check here. Two concrete consequences:

1. An empty name persists and produces the malformed header line `": Bearer <key>"` through curl — every request then goes out effectively unauthenticated.
2. A value containing CRLF injects arbitrary extra headers into every forwarded request.

Suggest a shared validator (reject non-string, reject empty, restrict to RFC 7230 token chars) applied both here and in `load_from_config()`, which has the same gap for values arriving from `config.json`.

ramkrishna2910 (commented):
Two findings on the new `auth_header_name` / `auth_header_prefix` plumbing. The core threading looks right — all three outbound auth sites (`post_with_auth`, `forward_streaming_request`, `discover_models`) and both discovery call sites were updated, and no hardcoded `"Bearer "` remains on a cloud path. The `set_allow_insecure_http()` fix for `/v1/cloud/auth` is a good catch too.

jgmelber:
Thanks both — all findings addressed in 96a5e64.

**Header injection / unvalidated `auth_header_name` (@ramkrishna2910, Copilot ×3)**

Added `validate_auth_header_name()` (non-empty RFC 7230 token) and `validate_auth_header_prefix()` (rejects control chars; empty stays valid, since that's the point of the flag). Applied at `/v1/install` *and* in `load_from_config()`, which had the same gap for values arriving from a hand-edited `config.json` — a bad value there falls back to the default rather than disabling the provider.

I put the check at the registry boundary rather than sanitizing at the three curl sites: the registry is the only way a value reaches those sites, so validating on the way in means stored state is always safe, and there's one place to audit instead of three. `discover_models()` and `ResolvedCreds` now take the registry's `AuthHeader` struct instead of two loose strings, which makes that the type-level contract, and the three outbound sites share one header-building helper.

Both concrete consequences you listed are now covered by tests — the malformed `": Bearer <key>"` line and the CRLF case.

**Re-install silently resetting the fields (@ramkrishna2910)**

Good catch, and you're right that it's the same class as the `allow_insecure_http` bug. `InstallOptions` fields are `std::optional` now and `install()` only overwrites what was actually supplied, so a re-install that omits the flags keeps the stored record. This covers both triggers you named — `cloud install <same provider> --base-url ...` and the desktop form's `backend`/`provider`/`base_url`/`api_key` post. The `http://` opt-in gate reads the effective (stored-or-supplied) value so an omitted flag doesn't re-prompt. Same fix on the CLI side: `--auth-header-*` are optional and omitted from the request body when not passed.

**`auth_header_prefix` not surfaced (Copilot ×2)**

Now in both `system-info` and the install response, and `cloud list` prints the header when *either* value is non-default, quoting the prefix so an empty or trailing-space one is visible:

```
auth header: X-Api-Key: ""<key>
```

**Testing**

`ctest -L cpp-ci -j4` → 43/43. The registry test is up to 31 checks, adding the clobber case, the validator cases, and malformed-values-from-config. Also verified live against a running `lemond`: installed a provider with `X-Api-Key`/empty prefix, re-installed with only a new `--base-url` and confirmed the header survived; confirmed the four bad-input shapes 400 and don't persist; confirmed a default provider still writes no `auth_header_*` keys to `config.json`; and confirmed a hand-edited malformed entry loads as `Authorization`/`Bearer ` after restart.

ramkrishna2910 (approved):

PR #3223, opened 2026-08-18
feat(cloud): relay Anthropic-wire-format providers through /v1/messages

### LINKED ISSUE #3221: Relay cloud providers that speak the Anthropic Messages wire format through /v1/messages
### Feature Description

`CloudServer` assumes every registered cloud provider speaks the OpenAI wire format (`POST /chat/completions`). Providers that instead serve the Anthropic Messages format (`POST /messages`) can be installed and their models discovered, but every inference request against them fails upstream.

Proposal: an optional per-provider `wire_format` setting (`openai` default, or `anthropic`). For an `anthropic` provider, `POST /v1/messages` forwards the request to the provider's `/messages` endpoint — rewriting only the model id to the provider's checkpoint — and relays the response back unchanged, streaming or buffered.

```bash
lemonade cloud install acme --base-url https://gateway.example.com/v1 \
  --api-key "$ACME_KEY" --wire-format anthropic
```

Relaying rather than converting is both smaller and lossless: thinking blocks, tool use, `cache_control`, and future Anthropic fields all survive. A converter would silently drop whatever it doesn't model.

Notably this needs no new endpoint — `POST /v1/messages` already exists and is a documented exception to the quad-prefix invariant. Local models and existing OpenAI-format providers keep the current convert-to-OpenAI-and-back path untouched.

### Use Case / Motivation

Lemonade already exposes an Anthropic-compatible `/v1/messages`, so Anthropic-format clients can point at `lemond` today and reach local models. But if a user also wants to route one model slot to a cloud provider that is itself Anthropic-native, that request has to be translated to OpenAI and back for no reason — and currently just fails, since the provider doesn't serve the OpenAI shape at all.

Since all three participants (client, `lemond`, provider) already speak the same wire format, passing the request through is strictly better than translating it in both directions. This lets a single `lemond` serve a mix of local models and Anthropic-format cloud models over one endpoint, which is what Anthropic-format clients need since they accept only one base URL.

### Platform Relevance

All platforms

### Additional Context

Builds on the per-provider auth header work (separate issue), since these providers are commonly reached through a gateway.

Two design points worth review: a provider that is relayed but unusable (no key, no base URL, plaintext http without the opt-in) should report its own error rather than falling through to the conversion path and surfacing as a misleading "model not found" 404; and the OpenAI-shaped endpoints should reject these providers with a clear 400 pointing at `/v1/messages` rather than passing an unservable request upstream.

I have a patch ready and will open a PR against this issue.

jgmelber:
> Stacked on #3222. Review that first; this branch is rebased on its current head.

## Summary

`CloudServer` assumes every provider speaks the OpenAI wire format, so providers serving the Anthropic Messages format can be installed and discovered but fail on every inference request.

This adds a per-provider `wire_format` setting (`openai` default, or `anthropic`). For an `anthropic` provider, `POST /v1/messages` forwards the body to the provider's `/messages` **rewriting only the model id**, and relays the response back unchanged, streaming or buffered. Nothing else is translated, so thinking blocks, tool use, `cache_control`, and future Anthropic fields pass through intact — a converter would silently drop what it doesn't model.

```bash
lemonade cloud install acme --base-url https://gateway.example.com/v1 \
  --api-key "$ACME_KEY" --wire-format anthropic \
  --auth-header-name x-api-key --auth-header-prefix ""
```

Everything else is untouched: `/v1/messages` keeps its convert-to-OpenAI path for any provider not registered as `anthropic`. No new endpoint — `POST /v1/messages` is already a documented quad-prefix exception.

Fixes #3221

## Scope

Three things worth a reviewer's attention:

- **Errors don't fall through.** Upstream resolution returns "claimed" separately from "resolved", so a relayed-but-unusable provider (no base URL, no key, plaintext http without the opt-in) reports its own 500/401/400 instead of falling through to the conversion path and surfacing as a misleading "model not found" 404.
- **OpenAI endpoints reject these providers explicitly.** `chat_completion()`, `completion()`, and `forward_streaming_request()` return `UNSUPPORTED_OPERATION` (400) naming the wire format and pointing at `/v1/messages`, rather than sending an unservable body upstream for an opaque gateway error.
- **`upstream_url()` consolidates URL joining.** Three call sites built the upstream URL by hand and disagreed: only the streaming path stripped the router's local `/v1` prefix, and none normalized a trailing slash on `base_url`. They now share one helper. The `/v1` an endpoint carries is lemonade's own, not the provider's — a gateway may serve the OpenAI shape at a base with no version segment.

The relayed path contributes no `/v1/stats` telemetry — the response body is never parsed, which is the point. The UI is deliberately not modified, per AGENTS.md.

**`5df168b5` reaches every backend, not just this relay.** `HttpClient`'s POST methods (`post`, `post_multipart`, `post_stream`) were passing `timeout_seconds=0` straight to `CURLOPT_TIMEOUT`, which curl reads as "no timeout" — `get` already treated `0` as "use the default." Fixed via `effective_timeout()`, `kConnectTimeoutSeconds`, and `kStreamStallSeconds`, so it now applies repo-wide, introducing connect/stall timeouts on paths that previously had none.

### Folded in from #3222's review

This branch predates #3222's review; rebasing surfaced two issues, fixed in the commits here:

- **`InstallOptions` fields are `std::optional`.** #3222 made every install option optional so a partial re-install leaves unspecified fields alone. `wire_format` had been added as a plain `std::string`, reintroducing that clobbering bug: a base-url-only re-install — or the desktop add-provider form, which posts only `backend`/`provider`/`base_url`/`api_key` — would reset it to `openai`, stopping the relay. Now `std::optional<std::string>`, with a test.
- **`anthropic-version` is sent on discovery too.** Discovery reads `GET <base_url>/models`, and a strict gateway rejects it without the version header, so the provider would register zero models and the relay would be unreachable. Extracted into `CloudServer::upstream_headers()` so relay and discovery build the same map and can't drift.

## Testing

_Testing details:_

`cmake --build --preset default` — clean, no new warnings. `ctest -L cpp-ci -j4` — **44/44 pass.**

`test_cloud_provider_registry_config` **46/46**: `wire_format` default, config round-trip, absent key loading as `openai`, invalid persisted value falling back rather than poisoning the record, `validate_wire_format()` accept/reject, and the clobbering case above.

`test_cloud_discovery_policy` **15/15**, now covering `upstream_headers()` (no version header on `openai`; `anthropic-version` and nothing beyond it on `anthropic`; empty prefix sends the bare key; wire format and auth header vary independently; a provider setting neither auth field gets exactly `Authorization: Bearer <key>`) and `upstream_url()` (base with and without a version segment, trailing slash, discovery path).

> This test was `CI OFF` and never ran in packaging CI. #3222 flips it `CI ON`, which is what takes ctest from 43 to 44.

Verified end-to-end against a live gateway serving both formats behind a non-standard auth header with an empty prefix:

- Discovery on both surfaces returns models — the request only succeeds if the key arrives under the configured header name.
- `POST /v1/messages` to a relayed provider round-trips buffered and streaming, through `message_stop` and including `ping` frames, with provider-native content blocks preserved.
- `POST /v1/chat/completions` for the same model → 400 naming `/v1/messages`.
- An OpenAI-wire model still converts through `/v1/messages` unchanged.
- Regression: a provider configuring no auth header still receives byte-identical `Authorization: Bearer <key>`, exactly one auth header, and no `anthropic-version` on the OpenAI path — captured at the wire with a loopback provider, and confirmed to fail when the previously hardcoded literal is restored.

## Documentation

`docs/guide/cli.md` (`--wire-format` row, `cloud list` note), `docs/api/lemonade.md` (install param, install response, `system-info`), `docs/guide/configuration/README.md` (`cloud_providers` field), `docs/guide/configuration/cloud.md` (new section, auth-header guidance, `anthropic-version` disclosure, two troubleshooting rows), `docs/api/anthropic.md` (relayed providers bypass conversion, so the documented field limits don't apply). No generated region touched.

## Breaking Changes

`wire_format` defaults to `openai`; absent and invalid config values both load as the default.

## AI-assisted contribution


ramkrishna2910 (commented):
Reviewed the incremental delta over #3222 only (`9ba8991..af2eaa1`, 18 files) — this PR's base is `main`, so the GitHub diff also shows all of #3222's already-reviewed work. Retargeting the base to `pr1-cloud-auth-header` would make the diff match what's actually new here.

No blocking bug. The wire-format threading, the segment-aware `/v1` stripping in `upstream_url`, the config round-trip, and the streaming error-diversion logic all look correct, and every `discover_models` caller was updated.

Six findings below. The first and third are the ones I'd want addressed before merge — one breaks real Anthropic SDK / Claude Code traffic against a relayed provider, the other makes this PR's own troubleshooting guidance wrong.

src/cpp/server/anthropic_api.cpp
```diff
@@ -177,6 +183,168 @@ static void set_anthropic_residency_conflict_response(
...
+    // the provider's own id rather than lemonade's "<provider>.<id>" public one.
+    json forwarded = request_json;
+    forwarded["model"] = info.checkpoint();
+
+    const auto auth_header = registry->auth_header_for(info.cloud_provider);
+    AnthropicUpstream upstream;
+    upstream.url = backends::CloudServer::upstream_url(base_url, "/messages");
+    upstream.headers = backends::CloudServer::upstream_headers(auth_header, api_key,
+                                                               "anthropic");
+    upstream.headers["Content-Type"] = "application/json";
```
ramkrishna2910, line 288: **Client request headers are dropped — most consequentially `anthropic-beta`.**

`upstream_headers()` builds the map from scratch (auth header + a hardcoded `anthropic-version = kAnthropicVersion`), and only `Content-Type` is added here. Nothing from the inbound request survives.

The Anthropic SDK and Claude Code send `anthropic-beta` routinely — `context-1m-2025-08-07`, `interleaved-thinking-2025-05-14`, and others. Concrete failure: a client POSTs `/v1/messages` with `anthropic-beta: context-1m-2025-08-07` and a 400k-token prompt to an `--wire-format anthropic` provider; the header never reaches upstream, so the request is rejected as over-length. A client-supplied `anthropic-version` is silently overwritten rather than honored, and the `?beta=true` query param isn't propagated either.

This also contradicts the doc line added in this same PR (`docs/guide/configuration/cloud.md:82`): "relays byte-for-byte… future Anthropic fields pass through intact" holds for the *body*, but not for the request envelope. Suggest forwarding an allowlist of `anthropic-*` request headers (at minimum `anthropic-beta`), letting a client-supplied `anthropic-version` win over the default.

src/cpp/server/anthropic_api.cpp
```diff
@@ -177,6 +183,168 @@ static void set_anthropic_residency_conflict_response(
...
+            set_anthropic_error_response(
+                res, 502, "cloud request for '" + model + "' failed: " + e.what());
+        }
+        return;
+    }
+
+    res.set_header("Cache-Control", "no-cache");
+    res.set_header("Connection", "keep-alive");
+    res.set_header("X-Accel-Buffering", "no");
+    res.set_chunked_content_provider(
```
ramkrishna2910, line 344: **The streaming relay always returns HTTP 200, even when upstream rejects the request.**

`set_chunked_content_provider` is installed before the upstream POST runs inside the lambda, so the response status is committed before it is known. A `stream: true` request with an expired key gets `HTTP/1.1 200` plus an SSE `event: error`, while the non-streaming path just above correctly relays the 401.

That's a real asymmetry for callers that gate on the status code — and note the troubleshooting table this PR adds tells users to look for a 401 ("An `anthropic` provider 401s with a valid key"), which streaming will never show them. Peeking at the first chunk to discriminate before committing the status — the same trick `CloudServer::forward_streaming_request` already uses — would let a non-200 come through with its real status.

src/cpp/server/anthropic_api.cpp
```diff
@@ -177,6 +183,168 @@ static void set_anthropic_residency_conflict_response(
...
+static void forward_anthropic_upstream(AnthropicUpstream upstream,
+                                       bool stream,
+                                       const std::string& model,
+                                       httplib::Response& res) {
+    if (!stream) {
+        try {
+            auto response = utils::HttpClient::post(
+                upstream.url, upstream.body, upstream.headers, 0, upstream.policy);
+            res.status = response.status_code;
+            res.set_content(response.body, "application/json");
```
ramkrishna2910, line 328: **Upstream response headers are discarded on the non-streaming relay.**

Only `status_code` and the body are copied. If upstream returns `429` with `retry-after: 30` and `anthropic-ratelimit-requests-reset`, the client sees a bare 429 and its SDK backoff falls back to a local heuristic instead of honoring the provider's window. `request-id` is lost too, which makes escalating a bad response to the upstream provider impossible.

Relaying an allowlist (`retry-after`, `anthropic-ratelimit-*`, `request-id`) would keep the "relay" framing honest.

src/cpp/server/anthropic_api.cpp
```diff
@@ -177,6 +183,168 @@ static void set_anthropic_residency_conflict_response(
...
+    if (CloudProviderRegistry::is_http_base_url(base_url) && !allow_insecure_http) {
+        return fail(400, "provider '" + info.cloud_provider +
+                         "' uses an http:// base URL; re-install it with "
+                         "--allow-insecure-http to send the API key in plaintext");
+    }
+
+    // The body passes through byte-for-byte apart from "model", which must name
+    // the provider's own id rather than lemonade's "<provider>.<id>" public one.
+    json forwarded = request_json;
+    forwarded["model"] = info.checkpoint();
```
ramkrishna2910, line 276: **No empty guard on `checkpoint()`, unlike the equivalent OpenAI-shaped path.**

`CloudServer::load()` (`cloud_server.cpp:299`) throws a clear "missing the 'checkpoint' field" error for exactly this condition. Here a hand-authored or extra-models-dir entry with `recipe: cloud` and no `checkpoint` relays `"model": ""` upstream and surfaces an opaque provider 400 — so the same malformed model gives an actionable local error on `/v1/chat/completions` but a confusing remote one on `/v1/messages`.

src/cpp/server/anthropic_api.cpp
```diff
@@ -938,6 +1106,20 @@ void OllamaApi::handle_anthropic_messages(const httplib::Request& req, httplib::
...
+        auto match = resolve_anthropic_upstream(model_manager_, model, request_json);
+        if (match.claimed) {
+            if (!match.upstream) {
+                set_anthropic_error_response(res, match.error_status,
+                                             match.error_message);
+                return;
+            }
+            forward_anthropic_upstream(std::move(*match.upstream),
+                                       request_json.value("stream", false), model, res);
+            return;
```
ramkrishna2910, line 1184: **The relay branch returns before `X-Lemonade-Warning` is set, so accumulated warnings are silently dropped.**

`add_warning(warnings, "Ignored unsupported beta query value: " + beta_value)` is recorded at line 1105 and then thrown away by this early return. `POST /v1/messages?beta=maybe` against a relayed provider returns no warning header, while the identical request against a local model does.

docs/guide/configuration/cloud.md
```diff
@@ -48,6 +48,42 @@ lemonade cloud install fireworks \
...
+```
+
+The wire format and the auth header are independent settings, but usually go together: the Anthropic API authenticates with `x-api-key`, so leaving the header at its default returns `401`. Omit the auth flags only for a gateway that fronts the Anthropic format behind bearer auth.
+
+Every request to an `anthropic` provider, discovery included, carries `anthropic-version: 2023-06-01`.
+
+Discovery is unchanged — `GET <base_url>/models`, which these providers serve in the OpenAI envelope. Inference differs:
+
+- **`POST /v1/messages`** relays to `<base_url>/messages` byte-for-byte, rewriting only the `model` field. Nothing is converted, so thinking blocks, tool use, cache control, and future Anthropic fields pass through intact. Streaming relays the upstream SSE unmodified.
+- **`POST /v1/chat/completions` and `POST /v1/completions`** return `400` pointing at `/v1/messages`, since the provider does not serve the OpenAI shape.
```
ramkrishna2910: **The `400` claim holds only for non-streaming requests.**

`CloudServer::forward_streaming_request` (`cloud_server.cpp:552`) writes the wire-format error as an SSE `data:` frame into an already-started chunked response, so a *streaming* chat-completions request against an `anthropic` provider returns HTTP 200 with an error event, not `400`. Worth qualifying the sentence, since someone debugging will be looking for the status code.

src/cpp/server/anthropic_api.cpp
jgmelber: _Edited: this reply originally described a threaded relay as the fix for this finding. That approach was reverted in 7600d5d0 and the original text is no longer accurate, so it's replaced here rather than left standing._

**Your suggestion was the right one. This finding is not fixed.**

The mechanism I originally described is real: httplib writes the status line in `write_response_core` (`httplib.h:11423`) before invoking the content provider, so a first-chunk peek genuinely cannot change the status. But I used that to justify moving the upstream request *before* the handler returned, which put the blocking wait somewhere neither `DataSink::is_writable` nor `~Response`'s releaser can reach. A client that disconnected in that window stranded the worker and the upstream transfer with no way to recover. Under load that was enough to take the whole server down and keep it down; the peek path under identical load is unaffected.

So the tradeoff runs the other way from how I first presented it. Peek is safe by construction — every blocking wait sits where the peer is observable — and I traded that away for a status code. Not a good trade, and the resource-releaser I claimed had closed the hazard did not: it only runs after the handler returns, which was exactly what was blocked.

Streaming still surfaces an upstream error as an SSE `error` event on a `200`. The docs and both troubleshooting rows now state that instead of claiming the status is preserved.

I'd also retract the verification line from the original text. It cited a thread-count check across a handful of aborted streams, but the harness answered promptly, so a status always arrived before the abort and the window in question never opened. It read as clearing abort-safety when it tested nothing of the sort.

Happy to share the measurements and harness privately if useful.

src/cpp/server/anthropic_api.cpp
jgmelber: Fixed in efb47a5e.

Added an allowlist forwarded from the inbound request — `anthropic-beta` and `anthropic-version` (`src/cpp/include/lemon/anthropic_relay_headers.h`). A client value replaces the default `upstream_headers()` set, so a client-pinned `anthropic-version` now wins. Repeats of the same name are comma-joined, since the SDK may send `anthropic-beta` as several headers rather than one combined value.

`?beta=true` is handled separately: the API reads it from the query string, so it's reattached to the upstream URL rather than folded into a header.

Verified against a loopback provider that records the wire. Two `anthropic-beta` request headers arrived upstream joined as `interleaved-thinking-2025-05-14,context-1m-2025-08-07`; a client `anthropic-version: 2024-10-22` beat the default; the path was `/messages?beta=true`. Confirmed the exclusions still hold — `x-api-key` set from the provider config, client `authorization` absent.

You're right that the doc line overclaimed. It now scopes "byte-for-byte" to the body and describes the envelope separately.

docs/guide/configuration/cloud.md
jgmelber: Confirmed and documented in efb47a5e. Reproduced it exactly as described: non-streaming gives `400`, `stream: true` gives `200` with the message as an SSE error frame.

The bullet now reads: "A non-streaming request gets `400` pointing at `/v1/messages`; a `stream: true` request gets the same message as an SSE error frame on a `200`, because the response has already begun by the time the body is written." The troubleshooting row carries the same qualification, since that's where someone debugging by status code will land.

Note this is the OpenAI-shaped path in `CloudServer::forward_streaming_request`, which is out of scope for this PR — so it stays a doc fix here. The `/v1/messages` relay path does now report real upstream statuses (see the thread on the streaming relay).

src/cpp/server/anthropic_api.cpp
jgmelber: Fixed in efb47a5e — allowlist is `retry-after`, `request-id`, and prefix `anthropic-ratelimit-*`.

This needed a change one layer down: `HttpClient` was discarding response headers entirely. Added a `CURLOPT_HEADERFUNCTION` collector that lowercases names (so lookups don't depend on provider casing) and clears on each `HTTP/` status line, so a redirect or a 1xx doesn't leave headers from a superseded block. `post_stream` takes an optional out-param populated before `on_status` fires.

Verified an upstream 429 carrying `retry-after: 30`, `request-id`, and `anthropic-ratelimit-*` relays all three to the client, on both the buffered and streaming paths. A `set-cookie` on the same response was not relayed.

src/cpp/server/anthropic_api.cpp
jgmelber: Fixed in efb47a5e. Guard added before the body is built, returning 500 with the same "missing the 'checkpoint' field (provider's upstream model id)" wording `CloudServer::load()` uses.

Verified with a registry entry with an empty `checkpoint`: both `/v1/chat/completions` and `/v1/messages` now give the same actionable local error instead of one of each.

src/cpp/server/anthropic_api.cpp
jgmelber: Fixed in efb47a5e. Hoisted the warning emission into an `emit_warning_header()` lambda above the relay branch and called it on both paths.

Verified `POST /v1/messages?beta=maybe` against a relayed provider now returns `X-Lemonade-Warning`, matching the local-model path, and confirmed the rejected beta value is still not forwarded upstream.

jgmelber:
Thanks for the review — all six findings are addressed in efb47a5e, with a reply on each thread. `main` is merged in as of 710d98a7.

Summary:

1. **Client request headers dropped** — `anthropic-beta` / `anthropic-version` now forwarded via an allowlist, client `anthropic-version` wins, repeats comma-joined, `?beta=true` reattached to the upstream URL.
2. **Streaming always 200** — the upstream request now starts before the handler returns, so a non-200 relays its real status. The suggested first-chunk peek can't work: httplib commits the status line only after the handler returns, so the content provider runs strictly too late to affect it. Details in that thread.
3. **Response headers discarded** — `retry-after`, `request-id`, `anthropic-ratelimit-*` relayed on both paths. Required adding response-header capture to `HttpClient`, which was dropping them entirely.
4. **No `checkpoint()` empty guard** — same actionable 500 as `CloudServer::load()`.
5. **`X-Lemonade-Warning` dropped on the relay branch** — emitted on both paths now.
6. **Doc `400` claim** — qualified in the bullet and the troubleshooting row. Reproduced first; the underlying behavior is in the OpenAI-shaped path, so it stays a doc fix here.

Beyond the six: the threading rework in #2 introduced a hazard where a client disappearing before the response started would park the upstream thread on the pending-buffer wait with the connection open. Closed with a resource-releaser that httplib runs even when the content provider never does. Flagging it since it's the one change not traceable to a finding.

Testing: verified end-to-end against a loopback Anthropic-wire provider covering each finding, plus regressions on the `openai` wire path (byte-identical auth header, no `anthropic-version` leak, `/v1/messages` still converts and doesn't forward `anthropic-beta`) and a thread-count check across 8 aborted streams. `ctest -L cpp-ci` is 46/46. Added `test/cpp/test_anthropic_relay_headers.cpp` covering the allowlists — the exclusions especially, since forwarding a client `Authorization` or a stale `Content-Length` would break the relay outright.

On retargeting the base: #3222 has since landed on `main`, so the diff here should now show only this PR's work without a base change.

src/cpp/server/anthropic_api.cpp
jgmelber: **Correction: I was wrong, and I've reverted to your suggestion (7600d5d0).**

My earlier reply argued the peek couldn't work and presented a threaded handoff as a better fix. The mechanism claim was correct — httplib writes the status line in `write_response_core` (`httplib.h:11423`) before the content provider runs, so peeking genuinely cannot change the status. But I used that to justify a design whose failure mode I hadn't tested, and it was much worse than the problem it solved.

**The threaded version was a remotely-triggerable DoS.** It blocked the handler *before* it returned, waiting for the upstream status. In that window neither safety net exists: `DataSink::is_writable` isn't available yet, and `~Response` (which fires the resource releaser) hasn't run. A client that disconnected there stranded the httplib worker on an untimed wait with the upstream transfer still open — permanently.

Measured against a hung upstream with a burst of client-aborted streaming requests, the threaded version drove the server to the point where it stopped answering `/health` at all and never recovered, holding every upstream connection open. The peek path under the identical load held zero connections and stayed at `200` throughout. Exact figures and the harness are off-thread; happy to share them privately.

Not a smaller leak — no leak, and the server stayed fully healthy under the load that permanently killed the other version. The reason is structural: peek does all its blocking inside the content provider, where the peer is observable on every iteration. That's the property I traded away without noticing.

**What this means for the finding.** It stands, unfixed. A streaming request against an `anthropic` provider still returns `200` with an SSE `error` event when upstream rejects it. Rather than claim otherwise, the docs now say so directly — the bullet no longer asserts the status is preserved on both paths, and the 401 troubleshooting row you flagged now tells the reader to check the event body instead of the status code on `stream: true`.

I'd rather leave this open than ship the version that fixed it. A safe fix likely needs a bounded, liveness-aware wait using `Request::is_connection_closed` (`httplib.h:1292`), or a status hook at the httplib layer — worth doing deliberately, not under review pressure.

**Also correcting my testing claim.** I previously reported "thread count was 266 before and after 8 aborted streams." That measurement was real but tested the wrong thing: my harness responded promptly, so a status always arrived before the abort and the vulnerable window never opened. It read as if I'd cleared abort-safety generally. I hadn't.

The non-streaming path keeps the real status and the relayed `retry-after` / `request-id` / `anthropic-ratelimit-*` headers — that part of the work is unaffected.

Also folded in while testing this: `HttpClient`'s POST methods did not honor the `0`-means-default timeout convention that `get` uses, so a request that left the timeout unspecified could wait on upstream far longer than intended, and `global_timeout` had no effect on it. Fixed in 5df168b5 with a regression test (`test_http_client_timeout`). It is shared `HttpClient` code rather than relay code, so it is not one of the six findings — flagging it since it is in the diff.

ramkrishna2910: Leaving this one open deliberately — the other five are resolved.

I read `7600d5d07` and the rationale holds: running inside the content provider is what keeps every blocking wait somewhere `sink.is_writable` can observe the peer, so a client that disconnects can't strand the upstream transfer. Correct disconnect handling is worth more than an accurate status line, and the docs + both troubleshooting rows now say the error arrives as an SSE `error` event on a `200`, which was my actual objection.

So no change requested. Keeping the thread open only so the tradeoff is visible to the next reader rather than silently closed — feel free to resolve it yourself.

jgmelber:
Thanks for the review. Five of the six findings are fixed; #2 is deliberately left open. `main` is merged in as of 710d98a7.

1. **Client request headers dropped** — `anthropic-beta` / `anthropic-version` forwarded via an allowlist, client `anthropic-version` wins, repeats comma-joined, `?beta=true` reattached to the upstream URL.
2. **Streaming always 200** — **not fixed.** I first shipped a threaded relay for this, then reverted it in 7600d5d0: it fixed the status code by moving the blocking wait outside the reach of httplib's disconnect handling, which was a much worse problem than the one it solved. Your peek suggestion was the safer shape. Streaming still surfaces an upstream error as an SSE `error` event on a `200`, and the docs now say so rather than claiming otherwise. Detail on the thread.
3. **Response headers discarded** — `retry-after`, `request-id`, `anthropic-ratelimit-*` relayed on the non-streaming path. Required adding response-header capture to `HttpClient`, which was dropping them entirely.
4. **No `checkpoint()` empty guard** — same actionable 500 as `CloudServer::load()`.
5. **`X-Lemonade-Warning` dropped on the relay branch** — emitted on both paths now.
6. **Doc `400` claim** — qualified in the bullet and the troubleshooting row. Reproduced first; the underlying behavior is in the OpenAI-shaped path, so it stays a doc fix here.

Also in the diff, not traceable to a finding: 5df168b5 makes `HttpClient`'s POST methods honor the `0`-means-default timeout convention that `get` already used. They were passing `0` to `CURLOPT_TIMEOUT`, which curl reads as "no timeout", so a request that left the timeout unspecified could wait on upstream far longer than intended — and `global_timeout` had no effect on any of them, since it only ever reached `get`. `post_stream` bounds upstream silence rather than total duration, since a flat timeout would cut off a long but healthy generation. `test_http_client_timeout` covers it.

Testing: verified end-to-end against a loopback Anthropic-wire provider covering each finding, plus regressions on the `openai` wire path (byte-identical auth header, no `anthropic-version` leak, `/v1/messages` still converts and doesn't forward `anthropic-beta`). All findings re-verified against the reverted build. `ctest -L cpp-ci` is 47/47. Added `test/cpp/test_anthropic_relay_headers.cpp` covering the allowlists — the exclusions especially, since forwarding a client `Authorization` or a stale `Content-Length` would break the relay outright.

On retargeting the base: #3222 has since landed on `main`, so the diff here should now show only this PR's work without a base change.

src/cpp/server/anthropic_api.cpp
jgmelber: Sounds good. Thanks @ramkrishna2910

ramkrishna2910:
The only thing red on this PR is `pre-commit` — every one of the 41 checks that ran, C++ unit tests and integration tests included, passes. The hook is tripping `MAX_BLOCK_LINES = 12` in `tools/check_comment_slop.py` (not the repeated-explanation rule that came up on #3222):

```
test/cpp/test_cloud_discovery_policy.cpp:1: 15-line comment block (limit 12).
test/cpp/test_http_client_timeout.cpp:1: 13-line comment block (limit 12).
    Design notes belong in the PR description, not the source.
```

Both are file-header blocks in the test files added by the review-fix commits — my review asked for the explanations that then pushed these over the limit, so here's the trim rather than just the report.

Verified locally: `python tools/check_comment_slop.py --from-ref $(git merge-base origin/main HEAD) --to-ref HEAD` goes from `EXIT=1` to `EXIT=0`, leaving only the density `note:` (which doesn't fail).

What I kept vs. cut:

- **`test_cloud_discovery_policy.cpp`** — dropped the per-function walkthrough, since each paragraph restated what its own test case already asserts. Kept the one genuinely non-obvious invariant (`allow_insecure_http` must never apply to an `https://` provider) and the `NDEBUG` rationale for the pass/fail counter.
- **`test_http_client_timeout.cpp`** — only needed to lose one line. The `CURLOPT_TIMEOUT 0` regression note is the valuable part and is untouched.

```diff
diff --git a/test/cpp/test_cloud_discovery_policy.cpp b/test/cpp/test_cloud_discovery_policy.cpp
--- a/test/cpp/test_cloud_discovery_policy.cpp
+++ b/test/cpp/test_cloud_discovery_policy.cpp
@@ -1,18 +1,9 @@
-// Unit tests for the CloudServer statics that decide how an outbound provider
-// request is addressed: discovery_policy(), which selects the HTTP trust
-// boundary, and upstream_headers(), which builds the header map.
+// Unit tests for the CloudServer statics that address an outbound provider
+// request: discovery_policy(), upstream_headers(), and upstream_url().
 //
-// discovery_policy(): the AllowInsecureHttp opt-in must only apply to plaintext
-// http:// providers; an https:// provider must stay HTTPS-only even when
-// allow_insecure_http is stale or accidentally set, since the discovery request
-// carries an Authorization: Bearer header.
-//
-// upstream_headers(): an anthropic-wire-format provider needs anthropic-version
-// on every request, discovery included; a provider that configures no auth
-// header must still get exactly Authorization: Bearer <key>.
-//
-// upstream_url(): the "/v1" on an endpoint is lemonade's own, not the
-// provider's, so it must be stripped before joining.
+// The invariant worth guarding: the AllowInsecureHttp opt-in must apply only to
+// plaintext http:// providers. An https:// provider stays HTTPS-only even when
+// allow_insecure_http is stale, since the request carries the API key.
 //
 // Checks use an explicit pass/fail counter (not assert()) so the test stays
 // effective under the Release build the CI `default` preset uses, where
diff --git a/test/cpp/test_http_client_timeout.cpp b/test/cpp/test_http_client_timeout.cpp
--- a/test/cpp/test_http_client_timeout.cpp
+++ b/test/cpp/test_http_client_timeout.cpp
@@ -1,15 +1,13 @@
 // Unit tests for the timeout convention shared by lemon::utils::HttpClient's
 // request methods. A loopback handler that never responds stands in for a
-// silent upstream, so each method can be checked for whether it actually
-// abandons the transfer.
+// silent upstream.
 //
-// The regression these guard: curl reads CURLOPT_TIMEOUT 0 as "no timeout",
-// so a method that forwarded 0 straight through blocked forever and parked the
+// The regression these guard: curl reads CURLOPT_TIMEOUT 0 as "no timeout", so
+// a method that forwarded 0 straight through blocked forever and parked the
 // calling httplib worker with it.
 //
 // Checks use an explicit pass/fail counter (not assert()) so the test stays
-// effective under the Release build the CI `default` preset uses, where
-// -DNDEBUG would compile assert() to a no-op.
+// effective under the Release CI build, where -DNDEBUG no-ops assert().
 
 #include <atomic>
 #include <chrono>
```

Two notes to attach to the PR while you're in here, neither blocking:

1. **`5df168b57` deserves a line in the description.** The `HttpClient` change (`effective_timeout()`, `kConnectTimeoutSeconds = 30`, `kStreamStallSeconds = 120`) applies to every caller across every backend, not just the cloud path. It's a real fix — curl reads `CURLOPT_TIMEOUT 0` as "no timeout" and parked httplib workers — but it introduces connect and stall timeouts repo-wide where some paths had none. Worth surfacing under an Anthropic-relay title so it isn't a surprise later.
2. **The forwarded request-header values are unvalidated**, which is safe only because httplib's parser can't produce a value containing CRLF. That's a real guarantee, but it's implicit — and it sits right next to #3222's explicit CRLF rejection for config-supplied header values. A one-line comment saying why the two differ would be worth more than a validator.

Otherwise this looks good to me: the five code findings are fixed and verified, the streaming tradeoff is documented, and I traced the `cloud_providers` round-trip against the new `config.json` deep-merge (#3268) — arrays replace wholesale rather than merging element-wise, and a non-empty provider list is never pruned against the `[]` default, so nothing is lost across a restart.

ramkrishna2910 (approved):
Approving the code. All six review findings are settled — five fixed and verified against the branch (client `anthropic-beta`/`anthropic-version` forwarding plus `?beta=true`, the response-header relay, the `checkpoint()` guard, `X-Lemonade-Warning` on the relay branch, and the `cloud.md` corrections), and the streaming-status tradeoff is documented rather than changed, which I'm satisfied with. I've resolved the last thread so conversation-resolution doesn't block the merge.

Also checked and clean: the `cloud_providers` round-trip against #3268's `config.json` deep-merge (arrays replace wholesale, a non-empty provider list is never pruned against the `[]` default), and the security surfaces — credentials never reach a log or error path, no reachable CRLF injection in either header direction, and the relay target is admin-configured rather than client-controlled.

One thing to land first: `pre-commit` is still red on two oversized test-header comment blocks. The verified trim is in https://github.com/lemonade-sdk/lemonade/pull/3223#issuecomment-5399639719 — please apply it before merging. Note that `pre-commit` is **not** in the required-checks list, so nothing stops the merge button from going green while it's red; that's exactly why I'd rather it land than get waved through.

Heads up that the workflow runs will come back as `action_required` after you push and will need approving before CI actually executes.

Non-blocking, worth a line in the description either way: `5df168b57`'s `HttpClient` change (`effective_timeout()`, `kConnectTimeoutSeconds`, `kStreamStallSeconds`) reaches every backend, not just the cloud path.

src/cpp/server/utils/http_client.cpp
```diff
@@ -250,6 +269,40 @@ static size_t write_callback(void* contents, size_t size, size_t nmemb, void* us
     return total_size;
 }
 
+// Collects response headers into HttpResponse::headers, lowercasing names so
```
Geramy, line 272: Note to self I'm not sure this is needed like this, I almost feel like we can pull the headers and append them directly to HttpResponse::headers without doing much, lower effort, no first/last. That seems weird at first glance.

src/cpp/server/utils/http_client.cpp
```diff
@@ -701,11 +759,23 @@ HttpResponse HttpClient::post_stream(const std::string& url,
...
+    }
     if (!apply_http_security_policy(curl, policy, false)) {
         curl_easy_cleanup(curl);
         throw std::runtime_error("Failed to apply HTTP security policy");
     }
-    curl_easy_setopt(curl, CURLOPT_TIMEOUT, timeout_seconds);
+    // A total timeout would kill a long but healthy generation, so an
+    // unqualified request is bounded by upstream silence instead of duration.
+    if (timeout_seconds == 0) {
+        curl_easy_setopt(curl, CURLOPT_LOW_SPEED_LIMIT, 1L);
```
Geramy, line 773: this is just crazy town right here! How does a low bps kill a connection, this might make sense if you have a multi-stream downloader where a stream can get choked because the upstream server mirror is now crap/overloaded, but this doesn't apply to this scenario. The only reason I see this being needed is if llama.cpp stops responding because it has a bug, but its not our job to fix their bugs in our software that leads to bad practices.

src/cpp/server/server.cpp
```diff
@@ -7812,9 +7813,9 @@ void Server::handle_install(const httplib::Request& req, httplib::Response& res)
             }
 
             CloudProviderRegistry::InstallOptions install_options;
-            auto read_header_field = [&](const char* field,
```
Geramy, line 7815: this is really horrible, why did this even change at all its completely the same with a different variable name it makes no sense it could of just had the added if statement and that's it.

Geramy:
I just want to comment, since we operate a very unknown range of expectations in this software I don't see using timeouts as a tool, if a server software is soft crashing or not responding, a timeout is probably not the right fix. The correct fix would be to walk the problem upstream instead of patching it in lemonade, that could be why this existed in the first place? Second a timeout is very user driven and can actually be taken care of in the user software itself, they can decide to kill the connection, even the firewall generally can or even the reverse proxy if there is one. It's not exactly required for us to do it. Once a program closes, the connection closes. If this was a commercial feature then perhaps that's fine. I just don't get the original idea behind this timeout mechanism, the open connection timeout is fine, that shouldn't be more than a few hundred ms if that, ever.

PR #3229, opened 2026-08-19
feat(server): record the backend launch command in /health

### LINKED ISSUE #3192: Record the backend launch command into the /health endpoint
Replaces the prototype /load/command endpoint in GUI3_merging - should return the same information, just without the need for a new endpoint.

Data can be stored in the wrappedserver class, which is already read out into the /health result.

fyi @fl0rianr

anditherobot:
## Summary

Fixes #3192.

Before this change, backends ran as subprocesses and `/health` only showed the process ID. The command itself was never recorded, even though options get resolved at load time: `ctx_size: auto` becomes a number, and the server adds flags of its own for llama.cpp.

`WrappedServer` now keeps the executable and args with the process handle. The 14 backends pass them at spawn, and the router publishes them in `/health` as `launch_command`. Replaced on restart, cleared on stop. Cloud models spawn nothing, so they have no field.

`/metrics` is left out on purpose. Prometheus, the monitoring system that scrapes it, stores one time series per unique label value, so a command line as a label would grow the series count without bound.

```json
"pid": 30320,
"launch_command": [
  "~/.cache/lemonade/bin/llamacpp/vulkan/llama-server.exe",
  "-m", "~/.cache/huggingface/hub/.../gemma-3-270m-it-UD-IQ2_M.gguf",
  "--ctx-size", "2048",
  "--port", "8001",
  "--jinja", "--metrics",
  "--reasoning-format", "auto",
  "--no-ui"
]
```

The caller supplied the ctx size and the server filled in the rest, which is now visible.

## Testing

Built on Windows with the `vs18` preset.

Added `test/cpp/test_launch_command.cpp` for the parts `/health` can't reach. The router skips dead backends, so a restart replacing the old command, and cleanup wiping it, are only checkable in process.

Ran a server on llama.cpp/Vulkan with `Tiny-Test-Model-GGUF`. The JSON above is what came back. Reloading with `ctx_size: 4096` gave 12 entries with the new value instead of 24. `test_009_load_model_basic` passes.

Notes:

- No user interface changes, out of scope, but the data is available if it should be displayed.

No cloud provider is configured locally, so I couldn't exercise that path, and I didn't run the full endpoint suite.

## Documentation

Updated `docs/api/lemonade.md`.

## AI-assisted contribution

Used AI tools. Reviewed to make sure only the changes needed were made.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3078,3 +3078,12 @@ if(BUILD_TESTING AND EXISTS "${_ROUTING_HELPER_RECONCILE_TEST_SRC}")
     target_link_libraries(test_routing_helper_reconcile PRIVATE lemonade-server-core)
     add_cpp_ci_test(RoutingHelperReconcileTest CI ON COMMAND test_routing_helper_reconcile)
 endif()
+
+# Launch command bookkeeping: the command recorded at spawn must keep describing
+# the process currently held, so a restart replaces it and cleanup erases it.
+set(_LAUNCH_COMMAND_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_launch_command.cpp")
+if(BUILD_TESTING AND EXISTS "${_LAUNCH_COMMAND_TEST_SRC}")
+    add_executable(test_launch_command test/cpp/test_launch_command.cpp)
+    target_link_libraries(test_launch_command PRIVATE lemonade-server-core)
+    add_cpp_ci_test(LaunchCommandTest CI ON COMMAND test_launch_command)
+endif()
--- a/docs/api/lemonade.md
+++ b/docs/api/lemonade.md
@@ -1270,6 +1270,12 @@ curl http://localhost:13305/v1/health
       "pinned": true,
       "recipe": "ryzenai-llm",
       "pid": 12345,
+      "launch_command": [
+        "~/.cache/lemonade/bin/ryzenai/npu/ryzenai-server.exe",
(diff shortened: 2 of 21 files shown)


fl0rianr:
Please rebase to main, this is a server only change. (server changes main, GUI changes GUI3_merging).

fl0rianr (changes requested):
Looks good overall. Assuming this gets rebased onto current `main`, I only see a few things worth tightening before merge:

* `pid` and `launch_command` are written together under `process_mutex_`, but `/health` reads them separately. A watchdog cleanup/restart between those reads can produce a mismatched snapshot. It would be safer to expose a single process metadata snapshot and read PID + launch command atomically.
* After rebasing to `main`, please double-check the `/load/command` references in the PR description. That endpoint is not currently on `main`, so “works as before” may no longer be accurate.
* The `/health` docs show `launch_command` only for the llama.cpp example, although other local backends such as RyzenAI expose it too. I’d update the example so it’s clear this is generally available for local backends and only absent for cloud models.
* Minor security/privacy note: this exposes resolved executable/model paths and arbitrary backend argv through `/health`. That seems acceptable as debug information, but it would be good to treat argv as potentially sensitive and avoid ever putting credentials there.

Other than that, the implementation looks solid: all current `set_process_handle` call sites appear covered, the command is recorded from the actual spawn arguments, restart/cleanup semantics are sensible, and the added tests cover the important bookkeeping behavior.

fl0rianr (approved):
LGTM, thanks!

PR #3235, opened 2026-08-19
fix(gui3): make download state server-owned

fl0rianr:
# PR Description

## Summary

Make `/api/v1/downloads` the sole source of truth for GUI3 download state.

* Remove persistent download items from localStorage (`lemonade_download_manager_items_v1`).
* Stop local pull callbacks and download actions from mutating canonical download state. They now only trigger a refresh from the server.
* Keep `lemonade_download_manager_dismissed_v1` client-owned for presentation only. Dismissal only hides terminal history and never changes download or blocking state.
* Update regression coverage for stale localStorage rows, authoritative empty snapshots, callback races, and dismissed non-terminal downloads.

Adresses #3161

## Scope

## Testing

_Testing details:_

* npm test passed
* typecheck passed
* locally tested DL works correctly

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/app/src/components/DownloadManager.tsx
+++ b/src/app/src/components/DownloadManager.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import api, { friendlyErrorMessage } from '../api';
 import { Icon } from './Icon';
-import { DownloadListItem, DownloadStatus, downloadStore, isDownloadActive } from '../features/downloadManager/downloadStore';
+import { DownloadListItem, downloadStore, isDownloadActive } from '../features/downloadManager/downloadStore';
 
 interface DownloadManagerProps {
   isVisible: boolean;
@@ -61,32 +61,19 @@ function isFinalizing(download: DownloadListItem): boolean {
     && download.bytesDownloaded >= download.bytesTotal;
 }
 
-function applyControlSnapshot(
-  download: DownloadListItem,
-  result: unknown,
-  fallbackStatus: DownloadStatus,
-): void {
-  if (!result || typeof result !== 'object' || Array.isArray(result)) return;
-  const snapshot = result as Record<string, unknown>;
-  downloadStore.upsertFromPull(download.modelName, {
-    ...snapshot,
-    status: snapshot.status || fallbackStatus,
-  }, download.downloadType);
-}
-
 async function removeDownload(download: DownloadListItem): Promise<void> {
   downloadStore.remove(download.id);
   await api.controlDownload(download.id, 'remove').catch(() => undefined);
 }
 
 const DownloadManager: React.FC<DownloadManagerProps> = ({ isVisible, onClose }) => {
-  const [downloads, setDownloads] = useState<DownloadListItem[]>(() => downloadStore.snapshot());
+  const [downloads, setDownloads] = useState<DownloadListItem[]>(() => downloadStore.visibleSnapshot());
   const [expanded, setExpanded] = useState<Set<string>>(() => new Set());
   const [busyIds, setBusyIds] = useState<Set<string>>(() => new Set());
   const prevStatusRef = useRef<Map<string, Pick<DownloadListItem, 'status' | 'running'>> | null>(null);
   const [statusAnnouncement, setStatusAnnouncement] = useState('');
 
-  useEffect(() => downloadStore.subscribe(setDownloads), []);
+  useEffect(() => downloadStore.subscribeVisible(setDownloads), []);
 
   // Announce status transitions (start/complete/error/pause/resume) to screen readers.
   // Runs on every downloads change but only emits announcements on status transitions,
@@ -182,15 +169,13 @@ const DownloadManager: React.FC<DownloadManagerProps> = ({ isVisible, onClose })
 
   const handlePause = (download: DownloadListItem) => withBusy(download.id, async () => {
     downloadStore.markLocal(download.modelName, 'paused', download.downloadType);
-    const result = await api.controlDownload(download.id, 'pause').catch(() => undefined);
-    applyControlSnapshot(download, result, 'paused');
+    await api.controlDownload(download.id, 'pause').catch(() => undefined);
     await downloadStore.refresh();
   });
 
   const handleCancel = (download: DownloadListItem) => withBusy(download.id, async () => {
     downloadStore.markLocal(download.modelName, 'cancelled', download.downloadType);
-    const result = await api.controlDownload(download.id, 'cancel').catch(() => undefined);
-    applyControlSnapshot(download, result, 'cancelled');
+    await api.controlDownload(download.id, 'cancel').catch(() => undefined);
     await downloadStore.refresh();
   });
 
--- a/src/app/src/components/DownloadManager.tsx
+++ b/src/app/src/components/DownloadManager.tsx
@@ -62,8 +62,13 @@ function isFinalizing(download: DownloadListItem): boolean {
 }
 
 async function removeDownload(download: DownloadListItem): Promise<void> {
-  downloadStore.remove(download.id);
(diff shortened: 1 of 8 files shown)


jeremyfowers (dismissed):
LGTM

src/app/src/features/downloadManager/downloadStore.ts
```diff
@@ -616,112 +544,36 @@ class DownloadStore {
...
-      raw.status = 'error';
-      raw.complete = false;
-    }
-    const percent = optionalNumber(data.percent);
-    if (percent != null) raw.percent = percent;
-    const previous = this.findExisting(raw);
-    const normalized = normalizeDownload(raw, previous);
-    if (normalized) this.mergeDownloads([normalized], false);
-    return normalized;
+  upsertFromPull(_modelName: string, _data: Record<string, unknown> = {}, _type: DownloadType = 'model'): DownloadListItem | null {
```
jeremyfowers: Backend install can now hang forever: `waitForBackendDownloadTerminal` (BackendManager.tsx:376) has no timeout and relied on the client-authored error row this method used to write. If the install fails before the server registers a `backend:` row (`api.installBackend` swallows errors), `sawDownload` stays false, the terminal promise never settles, and the Install button spins indefinitely.

src/app/src/features/downloadManager/downloadStore.ts
```diff
@@ -616,112 +544,36 @@ class DownloadStore {
...
+    // Pull/install callbacks are only a wake-up signal. Download state itself is
```
jeremyfowers: Discarding the payload makes pull failures invisible in ModelManager: its `onError` handlers only call `upsertFromPull({status:'error', ...})` (ModelManager.tsx:1724, 1784) with no toast or banner. A pull that fails before the server registers a job now just clears the spinner with no error shown anywhere. (ChatView is fine — it rethrows.)

src/app/src/features/downloadManager/downloadStore.ts
```diff
@@ -616,112 +544,36 @@ class DownloadStore {
...
+    // always replaced from GET /api/v1/downloads.
+    void this.refresh();
```
jeremyfowers: This refresh is wired to unthrottled SSE progress callbacks (per SSE line in api.ts, plus the 500ms terminal-wait poll), so any pull issues back-to-back `GET /downloads` requests limited only by latency for the entire download — many per second on fast links, multiplied per pull on a remote lemond. Consider throttling/coalescing here beyond the in-flight check.

src/app/src/features/downloadManager/downloadStore.ts
```diff
@@ -616,112 +544,36 @@ class DownloadStore {
...
+    return null;
```
jeremyfowers: Always returning null regresses progress display: callers using `item?.percent ?? data.percent` (ModelManager.tsx:1716, 1764, 1822; ChatView.tsx:1829) now fall back to the server's per-file SSE percent, so multi-file pulls (sharded GGUF, mmproj, FLM folders) sawtooth 0→100 once per file instead of showing the aggregated overall percent the old `normalizeDownload` path returned — disagreeing with the Download Manager row for the same pull.

src/app/src/features/downloadManager/downloadStore.ts
```diff
@@ -616,112 +544,36 @@ class DownloadStore {
...
-        : 0,
-      speedSampleTime: startsNewAttempt ? timestamp : base.speedSampleTime,
-      speedSampleBytes: startsNewAttempt ? 0 : base.speedSampleBytes,
-      error: startsNewAttempt ? undefined : base.error,
-      updatedAt: timestamp,
-      terminalAt: isDownloadTerminal({ status, running }) ? (base.terminalAt || timestamp) : undefined,
-      raw: startsNewAttempt ? undefined : base.raw,
-    };
-    this.mergeDownloads([item], false);
+  markLocal(_modelName: string, _status: DownloadStatus, _type: DownloadType = 'model'): void {
```
jeremyfowers: Cleanup: `upsertFromPull` and `markLocal` are now identical no-op shims, yet ~34 call sites still build rich status/percent/error payloads that are silently discarded (e.g. the BackendManager.tsx:878-969 mapping). Suggest replacing both with one explicit refresh/wake call and deleting the dead payload construction, so the signatures stop promising behavior that no longer exists.

src/app/src/features/downloadManager/downloadStore.ts
```diff
@@ -548,22 +464,21 @@ function writeDismissed(dismissed: Record<string, number>): void {
 }
 
 function isDismissedTerminal(download: DownloadListItem, dismissed: Record<string, number>): boolean {
-  return Boolean(dismissed[download.id]) && !isDownloadActive(download);
+  return Boolean(dismissed[download.id]) && isDownloadTerminal(download);
```
jeremyfowers: Dismissals expire after `TERMINAL_TTL_MS` (1h) but the server retains `error`/`cancelled` rows indefinitely (only `completed` is auto-pruned). If the `controlDownload('remove')` alongside a dismissal fails silently (DownloadManager.tsx:66), the cleared row reappears an hour later with no explanation.

src/app/src/features/downloadManager/downloadStore.ts
```diff
@@ -731,81 +583,24 @@ class DownloadStore {
...
-        // renderer reload recovered an obsolete localStorage row.
-        const awaitingServerRegistration = item.status === 'downloading'
-          && timestamp - item.createdAt < SERVER_APPEAR_GRACE_MS;
-        if (!awaitingServerRegistration) map.delete(id);
-      }
+    const dismissed = readDismissed();
+    for (const item of map.values()) {
+      // A retry reusing the same stable id is a new visible row. Clearing this is
+      // visual only and never changes canonical download/blocking state.
+      if (isDownloadActive(item)) delete dismissed[item.id];
```
jeremyfowers: Un-dismissal requires observing the row while active, but ids are stable per model (`model:<name>`). If another client (e.g. CLI `lemonade pull`) re-downloads a dismissed model while this GUI is idle with polling stopped, every snapshot it sees is already terminal, so the new download's outcome stays hidden for up to the 1h TTL.

src/app/src/components/DownloadManager.tsx
```diff
@@ -61,32 +61,19 @@ function isFinalizing(download: DownloadListItem): boolean {
...
-  if (!result || typeof result !== 'object' || Array.isArray(result)) return;
-  const snapshot = result as Record<string, unknown>;
-  downloadStore.upsertFromPull(download.modelName, {
-    ...snapshot,
-    status: snapshot.status || fallbackStatus,
-  }, download.downloadType);
-}
-
 async function removeDownload(download: DownloadListItem): Promise<void> {
   downloadStore.remove(download.id);
```
jeremyfowers: 'Remove from list' on a **paused** row is a silent no-op with a ghost: `removeMany` only dismisses `isDownloadTerminal` rows (paused isn't terminal), there's no refresh after the server deletes the job, and polling is stopped because paused isn't active. The row lingers until a focus/online/reopen refresh, and `activeDownloadForModel` keeps blocking re-download meanwhile.

jeremyfowers (commented):
Reviewed against the question: does this PR faithfully replace client-owned download state with server-owned state? The core migration holds — `GET /api/v1/downloads` is authoritative, dismissal is presentation-only, and duplicate concurrent pulls are deduped server-side. But no-op'ing `upsertFromPull`/`markLocal` without migrating their call sites drops client-authored state some callers still depend on. 8 inline comments; the first 5 are worth addressing before merge.

jeremyfowers (commented):
LGTM

jeremyfowers (approved):
LGTM

PR #3241, opened 2026-08-19
Scope VRAM headroom to the target GPU when auto-tuning ctx-size

### LINKED ISSUE #3224: ctx-size auto-tune miscalculates VRAM headroom with multiple models on separate GPUs, silently falls back to 4096
# ctx-size auto-tune miscalculates VRAM headroom with multiple models on separate GPUs, silently falls back to 4096

## Summary

When `--ctx-size` is omitted from `lemonade load`, the server auto-tunes a
context size based on available VRAM. With a single model loaded, this works
correctly. But once a *second* model is already loaded (even on a completely
separate GPU), loading or reloading either model without an explicit
`--ctx-size` collapses to the 4096-token floor instead of computing the real
per-GPU headroom. There is no warning or error — the model reports `status:
ready` and the caller has no indication the requested/expected context size
wasn't honored.

## Repro

Environment: 2x GPU (ROCm), llamacpp backend, `max_loaded_models=2`.

```bash
lemonade config set llamacpp.backend=rocm max_loaded_models=2

# Load model A on GPU0 with a large context size — works fine.
lemonade load ModelA --llamacpp-device ROCm0 --ctx-size 131072
# recipe_options: {ctx_size: 131072, llamacpp_device: ROCm0}  -- correct

# Load model B on GPU1, no explicit --ctx-size.
lemonade load ModelB --llamacpp-device ROCm1
# recipe_options: {ctx_size: 131072, llamacpp_device: ROCm1}  -- correct (single/first auto-tune ok)

# Now reload model B without --ctx-size, while model A is still loaded on the other GPU.
lemonade load ModelB --llamacpp-device ROCm1
# recipe_options: {ctx_size: 4096, llamacpp_device: ROCm1, status: ready}  -- BUG: silently collapses to floor
```

Verified via `curl http://localhost:<port>/api/v1/health` — each loaded
model's `recipe_options.ctx_size` reflects the actual value the backend was
launched with.

## Expected behavior

Auto-tune should compute headroom per-device (i.e., based on the VRAM
available on the specific GPU the model is being placed on), not based on
some aggregate/first-model view that gets confused once a second model is
resident. At minimum, if the auto-tune can't determine safe headroom in this
situation, it should warn rather than silently fall back to 4096.

## Impact

This is a silent correctness/perf regression: a model that appears healthy
(`status: ready`) may be serving with a drastically smaller context window
than the user expects and requested previously, with nothing in the CLI
output flagging the discrepancy. Any workflow that reloads models
periodically (e.g. after a server restart, or refreshing multiple models in
a multi-GPU setup) will hit this. The only reliable workaround today is to
always pass `--ctx-size` explicitly on every load call.

## Suggested fix

- Compute VRAM headroom scoped to the specific device a model is being
  placed on, not globally/aggregated across all loaded models.
- If a safe auto-tuned value truly cannot be determined, emit a visible
  warning in the CLI/log output rather than silently using the 4096 floor.

## Related

- #3017 — same auto-tune subsystem, but a different root cause (cross-vendor
  device-type memory lookup picks the wrong GPU's pool on hybrid
  AMD/NVIDIA systems). This issue is same-vendor, multi-GPU, and triggered
  purely by a second *model* being resident, not by backend/device-type
  selection.
- #2063 — same symptom family (silent bad ctx-size with no warning), but
  scoped to single-model/single-GPU defaults rather than the multi-GPU
  regression described here.

jgmelber:
## Summary

Auto-tuned `ctx_size` mixed two unrelated measurements on multi-GPU hosts: total VRAM came from the *first* AMD dGPU, while used VRAM came from a machine-wide "busiest card" scalar. Once a second model was resident on another GPU, the subtraction produced near-zero headroom and silently collapsed to the 4096 floor. This scopes both numbers to the device the model is placed on, and warns on every remaining path to the floor.

Fixes #3224

## Scope

Four supporting changes, all required rather than optional:

- **Sort KFD nodes numerically** in `detect_amd_gpus()`. `fs::directory_iterator` order is unspecified; on the test host it returns `3 1 2 0`, so the AMD GPU vector came back reversed and `GPUInfo::index` was never set for AMD at all. Any per-device lookup keyed on vector position reads the wrong card.
- **Per-device usage counters** on `GPUInfo`, from `mem_info_{vram,gtt}_used` (AMD) and `memory.used`/NVML (NVIDIA). They default to `-1` ("unknown") so an unreadable counter is never mistaken for an idle card, which would overestimate free VRAM.
- **`static` → `inline` on `get_used_memory_gb`.** As `static` it was emitted per-TU at `-O0`, dragging an undefined `create_metrics_platform()` into any TU including the header.
- **Two extractions to create test seams** — the nvidia-smi CSV row splitter (`nvidia_smi_parse.h`) and the warning-branch decision (`classify_ctx_scope()`). The splitter was previously reachable only by shelling out to a real binary. Both behavior-preserving; verified bit-identical against live hardware.

Behavior is deliberately unchanged where no per-device signal exists (Windows AMD, macOS, CPU/NPU): those keep the aggregate path. An unresolvable device — a nonexistent ordinal, a malformed string, or `Vulkan<N>`, which enumerates across vendors — falls back rather than silently selecting a different GPU.

## Testing

_Testing details:_

Built with the `default` (Release) and `debug` (`-O0`, confirming the `inline` fix links) presets. `ctest -L cpp-ci -j4` passes 43/43, including the new `GpuDeviceMemoryTest` (71 assertions; topology is supplied by the test, so multi-GPU cases run on any host).

Verified on a 2× AMD dGPU ROCm host (gfx1201) matching the issue's topology:

| | free VRAM | `ctx_size` |
|---|---|---|
| `--llamacpp-device ROCm0` | 20.70 GB | 16131 |
| `--llamacpp-device ROCm1` | 4.10 GB | 4096 |
| aggregate path (before) | 20.68 GB for *both* | — |

Both figures match `mem_info_vram_{used,total}` for that card's `renderD` node. The 4096 on ROCm1 is now correct rather than spurious, and it is announced:

```
[Warn] (AutoTune) Muse-Glimmer-30B-GGUF: only 4.10 GB free on AMD dGPU 1 — not enough
for the model weights plus a larger KV cache, so ctx_size fell back to 4096. Free VRAM
or pass --ctx-size to override.
```

The ordinal mapping was confirmed against llama.cpp itself — `llama-server --list-devices` reports `ROCm0: 21362 MiB free` / `ROCm1: 4420 MiB free`, matching which card this patch attributes to each ordinal (llama.cpp reads slightly higher as it excludes its own reserve). `hipGetDevicePCIBusId` maps HIP device 0/1 to the same PCI addresses as sorted KFD nodes 2/3.

Also confirmed the reading tracks a second allocation on the same card: holding 8 GiB via `hipMalloc` on ROCm0 moved its headroom 20.70 → 12.50 GB while ROCm1 stayed at 4.10 GB, both recovering on release.

Confirmed end-to-end through `POST /v1/models/{id}/options`, which uses the same `resolve_auto_ctx_size()` path as `load`. All four warning paths fire at the default `info` log level; the healthy path stays silent. CPU and NPU results are bit-identical to before the patch.

Worth noting for reviewers: on that host both cards report `gpu_busy_percent = 3` and the old tie-break is a strict `>`, so the "busiest card" was decided by `/sys/class/drm` directory order — the pre-patch number was arbitrary, not merely stale.

## Documentation

## Breaking Changes

No API, schema, or CLI change. `GPUInfo` gains fields; none are removed or repurposed. `GPUInfo::index` was NVIDIA-only and left at `-1` for AMD, so populating it is additive.

## Related

Overlaps open PR #3018 (for sibling issue #3017), which rewrites the same function. I ran a harness around its `select_gpu_memory_pool()` against this host's sysfs; two findings:

1. It indexes AMD GPUs by vector position and never sorts KFD nodes, so here `select(ROCm0)` returns ROCm1's memory and vice versa — 4.1 GB vs 20.7 GB, inverted.
2. With `backend=rocm` and no explicit device (the #3224 repro), it falls through to "first available AMD dGPU", leaving the headline case unfixed. Its fallback paths are all `LOG(DEBUG)`, i.e. invisible at the default log level.

## AI-assisted contribution


bitgamma (changes requested):
Thanks for the PR. A few things that need to be addressed

### 1. The result == AUTO_CTX_FALLBACK warning is a false-positive-prone heuristic                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                              
 src/cpp/include/lemon/auto_tune.h (end of resolve_auto_ctx_size):                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                              
 ```cpp                                                                                                                                                                                                                                                                                                                       
   if (!is_embedding && result == AUTO_CTX_FALLBACK) {                                                                                                                                                                                                                                                                        
       LOG(WARNING, "AutoTune") << ... "only " << available_gb << " GB free ... not enough                                                                                                                                                                                                                                    
           for the model weights plus a larger KV cache, so ctx_size fell back to 4096" ...;                                                                                                                                                                                                                                  
   }                                                                                                                                                                                                                                                                                                                          
 ```                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
 compute_auto_context_size returns AUTO_CTX_FALLBACK (4096) for several distinct reasons, and 4096 is also a legitimate non-fallback value:                                                                                                                                                                                   
 - A model whose max_context_window is exactly 4096 gets clamped to 4096 even with abundant memory.                                                                                                                                                                                                                           
 - A computed value can land exactly on 4096.                                                                                                                                                                                                                                                                                 
 - kv_bytes_per_token <= 0 returns fallback for non-memory reasons.                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                              
 Because the function returns no flag distinguishing "fell back" from "computed/clamped to 4096", this warning will fire in cases where there is no memory shortage at all and the diagnosis is wrong. Better to have compute_auto_context_size return a fallback-reason and only warn on the memory-specific paths.          
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
 ### 2. New spurious warnings on platforms with a named device but no per-device data                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
 Windows AMD never populates index/vram_used_gb (stays -1), so any named device resolves as Unscoped → new WARNING even on a single-GPU host where the aggregate figure is perfectly fine. Metal0 and CPU0 device strings also hit this. The message "may be far smaller than this GPU can hold" is misleading   
 on such hosts. Consider suppressing the warning when there is only one reachable GPU.                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                              
 ### 3. std::endl in the LOG stream is inconsistent and forces a flush                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                              
 The new warnings use << ... << std::endl, while the rest of the file ends lines with << " ". Minor, but it changes newline/flush behavior in the aixlog stream.                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                              
 ### 4. Implementation-detail headers in the public include tree                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                              
 gpu_device_memory.h and nvidia_smi_parse.h are placed in src/cpp/include/lemon/ (the public header dir). gpu_device_memory.h pulls in the heavy <lemon/system_info.h> (and json) just for GPUInfo. Consider moving to an internal location or a minimal local struct.

### 5. The change is broader than the PR's framing — it alters ctx tuning for all GPU recipes on Linux                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                              
resolve_auto_ctx_size calls get_available_memory_detail(model_info.device, backend, device_string) for every model, not just llama.cpp. Since per-device counters are now populated on Linux (AMD sysfs, NVIDIA NVML), the new per-device path is active for vllm and other GPU recipes too. On any Linux GPU host (even  single-GPU) the "free memory" figure changes source: previously first-card-capacity − machine-wide-used, now target-card-total − that-card-used. On multi-GPU hosts it switches to "most constrained GPU" and emits a new Ambiguous warning — but non-llamacpp recipes have no device option, so the warning's only          
 actionable advice is --ctx-size

bitgamma:
since https://github.com/lemonade-sdk/lemonade/pull/3018 predates this and is likely going to be merged soon, if you think this one comes with additional improvements it would make sense stack your changes on top off those

jeremyfowers:
@jgmelber if this PR is ready for re-review can you request it?

PR #3248, opened 2026-08-19
feat(telemetry): configurable session ID ingestion and OpenInference tool-call capture

abn:
## Summary

Enhances OpenTelemetry and OpenInference telemetry ingestion and attribute conformance:
- Ingests session and client identifiers from request headers (`telemetry.session.headers.id` / `telemetry.session.headers.client` or built-in agent headers like `x-opencode-session`) when omitted from the request body, namespacing as `<client>/<session>`.
- Aligns telemetry keys with standard OpenInference (`session.id`, `user.id`, `llm.output_messages.*.tool_calls`) and OpenTelemetry GenAI (`gen_ai.conversation.id`) specifications.
- Captures streaming SSE and non-streaming tool calls into structured OpenInference output messages (`llm.output_messages.0.message.tool_calls.*`), formatting `output.value` as a fallback summary when text content is empty.
- Clamps attribute lengths to `telemetry.max_attribute_length` while preserving valid JSON for truncated tool-call function arguments (`{"_truncated": true, "value": "..."}`).
- Respects privacy controls: `telemetry.hide_outputs` (suppressing tool calls and redacting content) and `telemetry.hide_thinking` (stripping reasoning tags).

## Scope

## Testing

_Testing details:_
- Built with `cmake --build build --target lemond lemonade cpp-ci-tests test_telemetry_helpers`.
- Ran CI test suite: `ctest --test-dir build -L cpp-ci --output-on-failure` (43/43 tests passed, 100%).
- Ran C++ unit tests: `./build/test_telemetry_helpers` covering:
  - Session precedence (body > configured headers > well-known headers > none).
  - Explicit empty `"session_id": ""` body precedence.
  - Client namespacing (`<client>/<session>`) with session ID preservation on long client prefixes.
  - Canonical OpenInference tool-call attribute emission and JSON-preserving truncation.
  - Reasoning stripping under `hide_thinking = true`.
  - Tool-call suppression under `hide_outputs = true`.
  - Loopback SSE stream client-disconnect error handling.
- Ran integration tests: `python test/server_telemetry.py` passing session header and OTLP validation.
- Validated doc sync: `python3 docs/tools/gen_backend_boilerplate.py --check`.
- Live validated with Arize Phoenix (`http://localhost:6006`) against multi-turn OpenCode sessions.

## Documentation

## Breaking Changes

> **Removal of Non-Standard Telemetry Attributes:**
> Removed legacy non-standard attributes `openinference.session.id` and `openinference.user.id` in favor of standard OpenInference attributes `session.id` and `user.id`. Because these legacy keys were non-standard and not recognized by standard OpenInference consumers (e.g. Arize Phoenix, Langfuse, MLflow), we have opted to remove them rather than carry redundant aliases. Any custom queries or dashboards explicitly filtering on `openinference.session.id` or `openinference.user.id` will need to update to `session.id` or `user.id`.

## AI-assisted contribution


fl0rianr (commented):
Overall this looks good, but I found a few correctness issues I think we should address before merge:

* The new `gen_ai.output.messages.0.tool_calls.*` attributes don't match the current OTel GenAI semantic convention. `gen_ai.output.messages` is defined as a structured message value with `parts` containing `tool_call` entries, so standard OTel consumers may not recognize the current flattened keys.
* An explicit `"session_id": ""` currently falls back to the header session, although the documented precedence says body wins and headers are only used when the body field is omitted.
* Truncating the combined `<client>/<session>` value can remove the actual session-specific part and make distinct sessions collide, e.g. multiple sessions under the same long client prefix.
* Tool-call arguments and the tool-only `output.value` fallback don't appear to respect `telemetry.max_attribute_length`, which can produce unexpectedly large span attributes.

The OpenInference key changes and the general streaming/tool-call capture approach otherwise look solid.

abn:
Thanks for the thorough review, @fl0rianr! I've addressed all four points:

1. **OTel GenAI Semantic Conventions**:
   - Removed the non-standard flattened `gen_ai.output.messages.0.tool_calls.*` keys from the `otel_genai` semantic path. Standard `gen_ai.output.messages.0.role` / `.content` and canonical OpenInference `llm.output_messages.0.message.tool_calls.*` remain intact.

2. **Body vs Header Precedence**:
   - Updated `InferenceSpan` to check `request_json.contains("session_id") && request_json["session_id"].is_string()`. An explicit `"session_id": ""` in the payload now strictly takes precedence and will not fall back to header session extraction.

3. **Namespaced Session Truncation & Collision Prevention**:
   - Added `format_namespaced_session()` with independent length budgeting. When a combined `<client>/<session>` exceeds `max_attribute_length`, the client prefix is truncated first to ensure the unique session ID suffix is preserved (e.g. `opencode-cli-desktop-extra-long/sess-xyz` becomes `opencode-cl/sess-xyz`), avoiding session collisions across long client names.

4. **Attribute Length Clamping on Tool Calls**:
   - Applied `truncate_string(..., max_len)` to `output.value` and per-tool-call attributes (`id`, `function.name`, `function.arguments`).

fl0rianr (commented):
Thanks for addressing the previous points. The remaining issue is OTel GenAI conformance: removing the flattened tool-call keys is correct, but tool calls are now omitted entirely from the OTel path, and `gen_ai.output.messages.0.role/content` still isn't the current convention. `gen_ai.output.messages` should be emitted as structured messages with `parts`, including `type: "tool_call"` entries.

If full OTel support is out of scope here, I'd suggest narrowing the PR description/tests to OpenInference tool-call capture instead.

abn:
Agreed, @fl0rianr! I've updated the PR title and description to scope structured tool-call capture specifically to the OpenInference semantic convention (`llm.output_messages.0.message.tool_calls.*`), leaving OTel GenAI message handling unchanged so full `parts`-based message modernization can be addressed in a dedicated PR.

fl0rianr (commented):
After a deeper pass, two things still stand out before approval:

* `tool_call.function.arguments` is truncated as a raw string, which can turn valid JSON into invalid JSON (e.g. `{"com... [TRUNCATED]`). Since OpenInference expects this to remain a JSON string, oversized arguments should be handled without breaking JSON validity.
* Replacing `openinference.session.id` / `openinference.user.id` with the standard `session.id` / `user.id` is correct, but it is still an observable compatibility change for existing dashboards/queries. The “no breaking changes” wording should probably acknowledge that, or the old keys could be kept temporarily as aliases.

I didn't find any additional runtime, threading, session-leak, or streaming regressions so far.

abn:
Thanks @fl0rianr! I've addressed both items:

1. **Valid JSON for Truncated Tool Arguments**:
   - Implemented `truncate_json_string()` in `telemetry.cpp`. When `tool_call.function.arguments` exceeds `max_attribute_length`, it is wrapped into a valid JSON object (`{"_truncated": true, "value": "..."}`) so downstream OpenInference trace consumers (Arize Phoenix, Langfuse) continue to parse the argument payload without JSON decode errors.

2. **Breaking Change Callout for Legacy Keys**:
   - Updated the PR description to explicitly flag a breaking change for the removal of the non-standard legacy attributes `openinference.session.id` and `openinference.user.id`. Since these non-standard keys are not recognized by standard OpenInference consumers, we have opted to cleanly remove them in favor of the canonical `session.id` and `user.id` attributes.

Also rebased cleanly on latest `upstream/main`.

fl0rianr (commented):
One remaining blocker after check:

`truncate_json_string()` does preserve JSON syntax, but it doesn't reliably enforce `max_attribute_length`. The wrapper itself can exceed the configured limit, especially for small values. It also truncates byte-wise, so UTF-8 arguments can be cut mid-character and make `nlohmann::json::dump()` throw. In the metrics worker, `end_with_success()` is outside the existing try/catch, so that could escape the worker thread.

I'd add coverage for `result.size() <= max_len` and a Unicode boundary case, and make the worker resilient to serialization errors. Also, the new `TelemetryHelpersTest` is currently `CI OFF`, so these cases aren't covered by the green CI run.

Everything else from the previous reviews looks resolved.

abn:
Thanks @fl0rianr! I've addressed all remaining points:

1. **Strict `max_attribute_length` Bounds & UTF-8 Safety**:
   - Updated `truncate_string()` with `utf8_safe_len()` so multi-byte UTF-8 sequences (e.g. Japanese characters, emojis) are never cut mid-character, preventing `nlohmann::json` serialization throws.
   - Updated `truncate_json_string()` with tiered valid JSON fallbacks (`{"_truncated": true, "value": "..."}`, `{"_truncated": true}`, `"{}"`, `""`) ensuring `result.size() <= max_len` unconditionally for all `max_attribute_length` thresholds (including very small bounds).

2. **Metrics Worker Resilience**:
   - Wrapped `task.span->end_with_success()` in `MetricsWorker::run()` within a `try/catch` block to ensure serialization exceptions cannot escape the background worker thread.

3. **CI Test Coverage & Enablement**:
   - Switched `TelemetryHelpersTest` to `CI ON` in `CMakeLists.txt` so it runs as part of `cpp-ci-tests` / `ctest -L cpp-ci` (now 47/47 passing).
   - Added unit test cases verifying Unicode multi-byte tool arguments, JSON parse validity, and strict `result.size() <= max_len` checks across bounds [0, 2, 10, 18, 30, 50, 100].

fl0rianr (approved):
Looks good now. The previous blockers are resolved and CI is green.

One minor edge case remains: max_attribute_length = 1 can emit an empty tool_call.function.arguments, which is no longer valid JSON. I wouldn't block the merge on this, but it may be worth either requiring >= 2 or omitting the attribute for such tiny limits.

Approve from my side.

abn:
Good catch! Updated `end_with_success()` to omit the `tool_call.function.arguments` attribute entirely when `max_attribute_length < 2`, ensuring any emitted argument attribute is always valid JSON. Added a test case for `max_attribute_length = 1` to `TelemetryHelpersTest`.

PR #2465, opened 2026-06-27
fix: skip optional draft checkpoint in are_required_checkpoints_complete (#2435)

bong-water-water-bong:
## Description

Fixes #2435 — adding a \`draft\` checkpoint to \`user_models.json\` caused the entire model to become unreadable.

### Root Cause

\`are_required_checkpoints_complete()\` iterates over **all** checkpoint types, including the optional \`draft\` MTP checkpoint. If the draft model hasn't been downloaded yet (or has a stale/invalid path), the entire model is marked as not-downloaded — making it invisible in the UI and unusable.

The \`draft\` checkpoint is purely optional: the model runs without MTP speculative decoding just fine. Only \`main\` is truly required.

### Fix

Skip \`draft\` checkpoint validation in \`are_required_checkpoints_complete()\`, the same way \`npu_cache\` was already skipped.

### Testing

- Compiles and links cleanly against \`main\`
- Models with a \`draft\` checkpoint entry but no downloaded draft GGUF will now appear as available
- Models without a draft checkpoint are unaffected

Issue #2435, opened 2026-06-26
10.8.1 breaks MTP with external MTP heads on user models


alucryd:
### Platform

Linux/Arch

### Lemonade Version

10.8.1

### GPU / APU Model

AMD Ryzen AI Max 395+

### Component

llama.cpp

### Bug Description

Lemonade now forbids passing -md to llama.cpp, without providing a way to pass the heads in user_models.json, essentially making us unable to use a custom Gemma 4 or other models with separate heads.

### Steps to Reproduce

1. Download the QAT versions of Gemma 4 from Unsloth
2. Download the associated MTP checkpoints
3. Try to load the model and pass the MTP head as argument

### Expected vs Actual Behavior

Lemonade should allow passing -md with user models, or at least use the same format in both server_models.json and user_models.json so we have at least one way to pass an MTP head.

### Log Output

```shell

```

bitgamma:
this is an intentional change and works the same as for the model file itself and mmproj file. You can specify the draft model by adding it to the model definition itself, as "draft" checkpoint. Example of json that can be imported (either via the webui or with `lemonade import` and defines the draft checkpoint:

```
{
  "checkpoints": {
    "draft": "unsloth/gemma-4-26B-A4B-it-qat-GGUF:mtp-gemma-4-26B-A4B-it.gguf",
    "main": "unsloth/gemma-4-26B-A4B-it-qat-GGUF:UD-Q4_K_XL",
    "mmproj": "unsloth/gemma-4-26B-A4B-it-qat-GGUF:mmproj-BF16.gguf"
  },
  "labels": [
    "custom",
    "vision",
    "mtp",
    "tool-calling"
  ],
  "recipe": "llamacpp",
  "recipe_options": {
    "ctx_size": 262144,
    "llamacpp_args": "--temp 1.0 --top_p 0.95 --top_k 64 --spec-type draft-mtp --spec-draft-n-max 4 --presence_penalty 0.0 --reasoning on",
    "llamacpp_backend": "vulkan",
    "merge_args": true,
    "pinned": false
  },
  "size": 14.4,
  "model_name": "user.gemma-4-26B-A4B-it-qat-GGUF-UD-Q4_K_XL"
}
```

alucryd:
I did try this, and my user_models.json file became unreadable, none of my user models came up in the web ui. I'm quite sure syntax was correct, but I'll give it another try.

alucryd:
[...]

PR #3253, opened 2026-08-20
fix: skip optional draft checkpoint in are_required_checkpoints_complete (#2435)

bong-water-water-bong:
Re-opened cleanly from the closed stacked PR #2465 (that branch was stacked on #2452's base; this is just the draft-checkpoint commit on current main).

Follow-up to #2435.

`are_required_checkpoints_complete()` iterated all checkpoint types including the optional `draft` MTP checkpoint, so a missing/stale draft model made the whole model appear not-downloaded (invisible in UI, unusable). Now skips `draft` checkpoints the same way `npu_cache` is skipped.

To keep the runtime consistent with that optional-draft semantics, `llamacpp_server.cpp` now only passes `--model-draft` / `--spec-type draft-mtp` (and the dflash equivalent) when the draft file actually resolves on disk — a missing drafter now loads and runs normally instead of failing llama-server startup.

Added regression coverage in `test_multi_checkpoint_completeness.py` asserting a main-only model is reported downloaded while main remains required.

---

**Update: merged the GLM4-MOE crash guard from #3255 into this PR** (both touched the same draft-activation path).

`compute_draft_activation()` now also gates MTP on mmproj presence (see #2451): GLM4-MOE / ChatGLM4 embed MTP layers in the GGUF but their draft graph builder accesses vision tensors that only exist when an mmproj companion is present — without one, MTP speculative decoding crashes llama-server. For those architectures MTP stays inactive unless mmproj resolves or the model is loaded via `-hf`. Text-only MTP models (Gemma-4, Qwen3.x) are unaffected. Case-insensitive architecture match; covered by new unit cases in `test_llamacpp_draft.cpp`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1179,6 +1179,19 @@ if(BUILD_TESTING AND EXISTS "${LLAMACPP_REQUEST_TEST_SOURCE}")
     add_cpp_ci_test(llamacpp_request CI OFF COMMAND test_llamacpp_request)
 endif()
 
+set(LLAMACPP_DRAFT_TEST_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_llamacpp_draft.cpp")
+if(BUILD_TESTING AND EXISTS "${LLAMACPP_DRAFT_TEST_SOURCE}")
+    add_executable(test_llamacpp_draft
+        ${LLAMACPP_DRAFT_TEST_SOURCE}
+        src/cpp/server/backends/llamacpp/llamacpp_draft.cpp
+    )
+    target_include_directories(test_llamacpp_draft PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+    add_cpp_ci_test(llamacpp_draft CI ON COMMAND test_llamacpp_draft)
+endif()
+
 set(MCP_CLIENT_CONFIG_TEST_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_mcp_client_config.cpp")
 if(BUILD_TESTING AND EXISTS "${MCP_CLIENT_CONFIG_TEST_SOURCE}")
     find_package(Python3 COMPONENTS Interpreter QUIET)
--- a/src/cpp/include/lemon/backends/llamacpp/llamacpp_draft.h
+++ b/src/cpp/include/lemon/backends/llamacpp/llamacpp_draft.h
@@ -19,16 +19,27 @@ struct DraftActivation {
 bool is_dflash_draft_checkpoint(const std::string& checkpoint);
 
 // Compute draft activation from the model's labels, the raw draft checkpoint
-// reference, and whether the resolved draft file actually exists on disk.
+// reference, whether the resolved draft file actually exists on disk, and the
+// GGUF architecture (used to gate GLM4-MOE, see below).
 //
 // Registration and activation are kept separate: the `draft` checkpoint only
 // describes where the drafter lives, while the `mtp` / `dflash` label is the
 // switch that enables speculative decoding. A drafter present on disk without
 // its matching label does not activate anything, and a label without a present
 // drafter also stays inactive (the base model runs normally).
+//
+// GLM4-MOE / ChatGLM4 models embed MTP layers in the GGUF but their draft
+// context construction accesses vision tensors that only exist when an mmproj
+// file is present; without one, MTP speculative decoding crashes llama-server
+// (see #2451). For those architectures MTP stays inactive unless the mmproj
+// path resolves (or the model is loaded via -hf, where llama-server resolves
+// the companion itself). Text-only MTP models (Gemma-4, Qwen3.x) are unaffected.
 DraftActivation compute_draft_activation(const std::vector<std::string>& labels,
                                          const std::string& draft_checkpoint,
-                                         bool draft_file_present);
+                                         bool draft_file_present,
+                                         const std::string& architecture,
+                                         const std::string& mmproj_path,
+                                         bool hf_load);
 
 } // namespace backends
 } // namespace lemon
--- a/src/cpp/include/lemon/backends/llamacpp/llamacpp_draft.h
+++ b/src/cpp/include/lemon/backends/llamacpp/llamacpp_draft.h
@@ -19,21 +19,15 @@ struct DraftActivation {
 bool is_dflash_draft_checkpoint(const std::string& checkpoint);
 
 // Compute draft activation from the model's labels, the raw draft checkpoint
-// reference, whether the resolved draft file actually exists on disk, and the
-// GGUF architecture (used to gate GLM4-MOE, see below).
+// reference, whether the resolved draft file exists on disk, the GGUF
+// architecture, the mmproj path, and hf_load.
 //
 // Registration and activation are kept separate: the `draft` checkpoint only
 // describes where the drafter lives, while the `mtp` / `dflash` label is the
-// switch that enables speculative decoding. A drafter present on disk without
-// its matching label does not activate anything, and a label without a present
-// drafter also stays inactive (the base model runs normally).
-//
-// GLM4-MOE / ChatGLM4 models embed MTP layers in the GGUF but their draft
-// context construction accesses vision tensors that only exist when an mmproj
-// file is present; without one, MTP speculative decoding crashes llama-server
-// (see #2451). For those architectures MTP stays inactive unless the mmproj
-// path resolves (or the model is loaded via -hf, where llama-server resolves
-// the companion itself). Text-only MTP models (Gemma-4, Qwen3.x) are unaffected.
+// switch that enables speculative decoding. An `mtp` label without an external
+// drafter means the MTP weights are embedded in the main GGUF: emit
+// --spec-type draft-mtp without --model-draft. GLM4-MOE embedded MTP crashes
+// without an mmproj companion (see #2451).
 DraftActivation compute_draft_activation(const std::vector<std::string>& labels,
                                          const std::string& draft_checkpoint,
                                          bool draft_file_present,
--- a/src/cpp/include/lemon/backends/llamacpp/llamacpp_draft.h
+++ b/src/cpp/include/lemon/backends/llamacpp/llamacpp_draft.h
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace lemon {
+namespace backends {
+
+// Decision about whether/how to activate an optional `draft` (MTP / DFlash)
+// checkpoint when launching llama-server.
+struct DraftActivation {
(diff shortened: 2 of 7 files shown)


fl0rianr (changes requested):
Thanks for splitting this out into a clean single-purpose PR.

I'm still not comfortable merging this without regression coverage. #2435 was already resolved after the reporter confirmed that importing a model with a draft checkpoint works, so this changes behavior beyond the bug that was originally reported.

The important part here is that skipping draft in are_required_checkpoints_complete() changes the model's downloaded state. With main present but the configured draft missing/stale, auto-load can now consider the model fully downloaded and skip fetching the missing checkpoint.

At the same time, the mtp label still enables --spec-type draft-mtp independently of whether draft_path resolved successfully. So I don't think we can assume that a missing external draft simply means "run normally without MTP".

bong-water-water-bong:
Thanks for the review @fl0rianr — addressed both points:

1. **Runtime alignment**: `llamacpp_server.cpp` now only adds `--model-draft` and `--spec-type draft-mtp` (and the dflash equivalent) when the draft file actually resolves on disk. A configured-but-missing drafter — or a bare `mtp` label with no draft — now loads and runs normally instead of failing llama-server startup.
2. **Regression coverage**: added `test_optional_draft_checkpoint_is_not_required` in `test/test_multi_checkpoint_completeness.py` (main-only model is downloaded; main is still required).

Also changed the PR body to `Follow-up to #2435` since the issue was already closed.

bong-water-water-bong:
@fl0rianr friendly nudge — the regression coverage and runtime gating are both in now: `llamacpp_server.cpp` only passes `--model-draft` / `--spec-type draft-mtp` when the draft file actually resolves on disk (so a missing drafter loads normally instead of failing startup), and `test_optional_draft_checkpoint_is_not_required` covers main-only→downloaded plus main-still-required. Ready for another look when you get a chance.

fl0rianr (commented):
Thanks, this is moving in a much better direction. I think there are still two things we should settle before merging though.

The external drafter clearly belongs in the model definition as a draft checkpoint, so that part of this approach makes sense. What I'm less sure about is the completeness semantic. An explicitly configured but missing or stale draft is now always treated as complete, which effectively makes draft optional at the model-manager level.

If the intended contract is that draft is optional acceleration and the base model must remain usable when it is unavailable, then I think this approach is reasonable. I'd just like that behavior confirmed before we bake it into are_required_checkpoints_complete().

I also think registration and activation should remain separate concerns. Having a draft checkpoint available should not by itself force the runtime to use it. The activation label should remain the switch for speculative decoding, since users may intentionally want to run the base model without the additional drafter.

There is also a concrete Windows concern in the runtime guard: please avoid using std::filesystem::exists() directly here. The repo already has hf_cache::exists() specifically because HF cache symlinks/reparse points can produce false negatives with std::filesystem on Windows. A false negative here would silently disable MTP even though the drafter is present.

Finally, the new regression test proves the downloaded/completeness state, but not the runtime behavior introduced in this PR. Could we also cover:

draft present + no mtp label -> no --model-draft and no --spec-type draft-mtp
draft missing + mtp label -> no --model-draft and no --spec-type draft-mtp
draft present + mtp label -> both arguments are emitted

With the intended optional-draft semantics confirmed, the Windows-safe existence check, and runtime coverage for the activation behavior, I'd be comfortable re-reviewing this for approval.

bong-water-water-bong:
@fl0rianr confirming the contract: `draft` is optional acceleration and the base model must stay usable without it — that's the semantic now baked into `are_required_checkpoints_complete()`. Also:

- made the `mtp`/`dflash` label the activation switch (a present drafter without its matching label no longer emits `--model-draft`/`--spec-type`),
- switched the guard to `hf_cache::exists()` for the Windows reparse-point case, and
- added `test_llamacpp_draft` covering the three activation cases (present+no-label → nothing, missing+mtp → nothing, present+mtp → both args) plus the dflash path.

`ctest` passes and the existing completeness integration test still passes. Ready for another look.

CMakeLists.txt
```diff
@@ -1179,6 +1179,19 @@ if(BUILD_TESTING AND EXISTS "${LLAMACPP_REQUEST_TEST_SOURCE}")
...
+if(BUILD_TESTING AND EXISTS "${LLAMACPP_DRAFT_TEST_SOURCE}")
+    add_executable(test_llamacpp_draft
+        ${LLAMACPP_DRAFT_TEST_SOURCE}
+        src/cpp/server/backends/llamacpp/llamacpp_draft.cpp
+    )
+    target_include_directories(test_llamacpp_draft PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/include
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
+    add_cpp_ci_test(llamacpp_draft CI OFF COMMAND test_llamacpp_draft)
```
```suggestion
    add_cpp_ci_test(llamacpp_draft CI ON COMMAND test_llamacpp_draft)
```

fl0rianr (commented):
With that we can have the workflow run.

fl0rianr (changes requested):
I dug into this again before approving, and I think there is still one real blocker.

The new compute_draft_activation() currently returns inactive whenever no external draft file is present. That is not valid for all MTP models: llama.cpp supports --spec-type draft-mtp without --model-draft when the MTP weights are embedded in the main GGUF.

Lemonade already has models using exactly that setup, e.g. Qwen3.5-4B-MTP-GGUF, Qwen3.5-122B-A10B-MTP-GGUF, Qwen3.6-27B-MTP-GGUF, etc. They have the mtp label but no external draft checkpoint. #2451 also shows the GLM path running with --spec-type draft-mtp and no --model-draft.

So I think the activation logic needs to distinguish:

mtp + no configured external draft → embedded MTP, keep --spec-type draft-mtp
mtp + external draft present → use both --model-draft and --spec-type draft-mtp
configured external draft missing → optional fallback without MTP
GLM4 embedded MTP → keep the mmproj guard

The current test draft missing + mtp label -> inactive unfortunately locks in the wrong behavior for embedded-MTP models, so that case should be adjusted as well.

Apart from that this is looking much better. The remaining comment-slop CI failure is just the 17-line design comment in llamacpp_draft.h; shortening that to the non-obvious invariant should be enough.

bong-water-water-bong:
Thanks for the detailed review — all four cases are now handled in `compute_draft_activation()`:

- **mtp + no external draft** → embedded MTP: emits `--spec-type draft-mtp` without `--model-draft` (covers Qwen3.5/3.6-*-MTP-GGUF, GLM path from #2451)
- **mtp + external draft present** → both `--model-draft` and `--spec-type draft-mtp`
- **configured external draft missing** → falls back to running without MTP
- **GLM4 embedded MTP** → mmproj guard kept (#2451), including embedded (no-external-draft) GLM4

The embedded path is also covered by new unit tests: `mtp label + no external draft -> embedded MTP (spec-type only)`, plus GLM4 embedded with/without mmproj and via `-hf`.

The 17-line design comment in `llamacpp_draft.h` is trimmed to 12 lines (comment-slop check passes now), and `load()`'s comment now spells out that `--model-draft` requires the drafter on disk while embedded MTP only emits `--spec-type`.

One thing to confirm: for a **configured-but-missing external draft** I kept the fallback as *no MTP at all* (base model runs), per your "optional fallback without MTP" — not a silent retry of embedded. Let me know if you'd rather it fall back to embedded MTP when the arch supports it.

jeremyfowers:
**Not ready for review yet**

**To-dos**

- [ ] Replace the reference to test_multi_checkpoint_completeness.py with the actual test path test/cpp/test_model_download_state.cpp in the PR body.

**Description vs. the diff: discrepancies**

The description references a Python test file that was not changed.

- Replace the reference to test_multi_checkpoint_completeness.py with the actual test path test/cpp/test_model_download_state.cpp in the PR body.
  - Described: Added regression coverage in test_multi_checkpoint_completeness.py
  - In the diff: The regression test was added to test/cpp/test_model_download_state.cpp, a C++ test file, not a Python test file.
  - Reference: PR body: 'Added regression coverage in test_multi_checkpoint_completeness.py asserting a main-only model is reported downloaded while main remains required.' — files list shows test/cpp/test_model_download_state.cpp with +31 additions and a new test_optional_draft_checkpoint_not_required() function.

**Focus: focused**

All changes serve one goal: making the draft checkpoint optional and its activation robust. The new llamacpp_draft module, the checkpoint skip in model_manager.cpp, the runtime gating in llamacpp_server.cpp, and the GLM4-MOE crash guard all address the same underlying problem of draft/MTP being treated as required rather than optional.

**Alignment: none found**

Checked: Reviewed contribute.md, philosophy.md, documentation.md, and testing.md; the change is a contained refactor with linked issues and comprehensive test coverage.

**Documentation: adequate**

Checked: none observed — the change is internal code restructuring with no user-facing surface; testing.md and documentation.md have no obligations for pure C++ logic changes.

**Testing: adequate**

Checked: New test/cpp/test_llamacpp_draft.cpp (191 lines, 20+ cases) exercises compute_draft_activation() across all activation branches; CMakeLists.txt registers it via add_cpp_ci_test(llamacpp_draft CI ON COMMAND test_llamacpp_draft) so the cpp-ci job runs it. test/cpp/test_model_download_state.cpp adds a regression case for the optional-draft skip.

Checked: none observed — the behavior shifts (skipping optional draft, gating MTP on presence) are fixes that restore robustness under the philosophy's 'User Error is a Bug' tenet, not removals of deliberate user-facing contracts.

</details>

**Review suggestion**

Attention level: Elevated — needs 2 reviewers including 1 subject-area expert in llamacpp.

Suggested reviewers:

- superm1 (llamacpp) — Maintainer table lists llamacpp as a subject area; this PR restructures the llamacpp draft/MTP activation path and adds the crash guard that affects llama-server startup behavior.
- pwilkin (llamacpp) — Maintainer table lists llamacpp as a subject area; this PR restructures the llamacpp draft/MTP activation path and adds the crash guard that affects llama-server startup behavior.

**Why this rung**

A refactor that restructures existing code. The diff introduces a new module (llamacpp_draft.h and llamacpp_draft.cpp) and restructures llamacpp_server.cpp to use it, which the table classifies as a refactor that restructures existing code. The llamacpp subject area covers this surface.

**Reviewer search**

Checked: Maintainer table rows for llamacpp: @superm1 and @pwilkin; both are subject-area experts for this PR's touched surface. Code-authorship pass (blamed via changed hunks) found @fl0rianr (relevance 62), @anditherobot (7), @jeremyfowers (6), @bitgamma (5), @sujikathir (3), @Phqen1x (3), @SlawomirNowaczyk (1), @Geramy (1) — none of these overlap with the suggested maintainers, confirming the authorship pass did not surface additional reviewers.

</details>

_[AI-assisted review] Automated pre-review from repo-manager — flags for the human reviewer, not a replacement for one. Reviewed at head `245bb57`; regenerate with `repo-manager review-pr 3253`._

jeremyfowers:
@bong-water-water-bong please fix the PR body as suggested above

Also please wait to hear from @bitgamma (who does a lot of MTP work) before merging.

bong-water-water-bong:

PR #3257, opened 2026-08-20
fix: add .completed sentinel to is_checkpoint_path_complete with backfill (#1341, #1999)

### LINKED ISSUE #1341: False positive downloaded flag for built-in model database.

bong-water-water-bong:
Re-opened from the closed stacked PR #2463 (that PR bundled 3 unrelated fixes; only the sentinel piece is still missing on main — omni size filtering conflicts with the deliberate \"collections always visible\" design, and the disk-space pre-check landed in stronger per-file form in download_from_manifest).

Fixes #1341, #1999.

Adds a `.completed` sentinel as the authoritative download-complete marker in `is_checkpoint_path_complete()`: a crash between `fs::rename` and manifest removal previously left a directory that looked complete. Backfills the sentinel for pre-existing downloads (content + no manifest/partials → write sentinel + accept), preventing re-download loops after upgrades.

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -2199,7 +2199,43 @@ static bool is_checkpoint_path_complete(const std::string& path_str) {
         return !safe_exists(path_from_utf8(path_str + ".partial"));
     }
 
-    return !has_partial_files(resolved);
+    if (has_partial_files(resolved)) return false;
+
+    // .completed sentinel: authoritative marker that a download finished
+    // successfully. Without it, a crash during download (between fs::rename
+    // and manifest removal) leaves a corrupt directory that appears complete
+    // by other checks alone.
+    fs::path completed_path = marker_dir / ".completed";
+    if (safe_exists(completed_path)) {
+        return true;
+    }
+
+    // Backward-compat: pre-existing downloads (before .completed was
+    // introduced) won't have the sentinel. If the directory has content
+    // and no manifest/partial files, write the sentinel and accept it.
+    // Also prevents re-download loops after upgrades. See #1999.
+    std::error_code ec;
+    bool has_content = false;
+    for (auto it = fs::directory_iterator(resolved, ec); it != fs::directory_iterator(); ++it) {
+        std::string name = it->path().filename().string();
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@fl0rianr this PR touches the download manager you worked on, would you like to review?

fl0rianr (changes requested):
I would not merge this as-is.

The main issue is that `.completed` is described as authoritative, but the fallback still treats `content + no manifest/partials` as complete and then writes the sentinel retroactively. So the state this is supposed to protect against is effectively accepted just like before, only now it gets permanently marked as complete.

On the current download path we also already have stronger commit semantics: files are validated before the manifest is removed, and modern snapshot caches without a valid `refs/main` are treated as uncommitted. That makes it unclear which real failure mode this sentinel still fixes.

A couple of related concerns:

* `Fixes #1341` doesn’t appear applicable here. That issue is about GGUF variant resolution, and the sentinel code only affects directory checkpoints.
* The completeness check now mutates the filesystem and ignores errors when writing `.completed`.
* There are no regression tests showing a case that fails on `main` and passes with this change.

I’d want to see a concrete reproducer first, ideally as a test. If there isn’t one against current `main`, I think closing this PR may be better than introducing a second completion mechanism alongside the existing manifest/`refs/main` flow.

bong-water-water-bong:

PR #2469, opened 2026-06-27
docs: document draft checkpoint key and add MTP model example (#2435)

bong-water-water-bong:
## Summary

Documents the `draft` checkpoint key and adds a complete `user_models.json` example for MTP (Multi-Token Prediction) models with a separate draft head.

Fixes #2435 (documentation component).

## Changes

**`docs/guide/configuration/custom-models.md`**

- **Checkpoint keys table** — added `draft` row (`llamacpp` | `Draft/MTP head for speculative decoding`)
- **New MTP example** — `My-Gemma-4-MTP` entry showing `"draft"` checkpoint + `"mtp"` label + `"mmproj"` in a single `user_models.json` definition
- **Explanatory note** — documents that the `draft` checkpoint is optional (model loads without MTP if the file is missing) and that the `mtp` label enables `--spec-type draft-mtp` defaults

## Background

The `draft` checkpoint key has been supported in code via `load_checkpoints()` and `llamacpp_server.cpp`'s `--model-draft` resolution since #2317, but it was never documented in the custom-models guide. Users with models like Gemma 4 (which ship separate MTP head GGUFs) had to guess the format or dig through source code.

The companion code fix (`7048b17f`) ensures that having a `draft` checkpoint doesn't break `are_required_checkpoints_complete()` when the draft GGUF hasn't been downloaded yet.

PR #3258, opened 2026-08-20
docs: document draft checkpoint key and add MTP model example (#2435)

bong-water-water-bong:
Re-opened as the clean docs-only PR @fl0rianr requested on #2469 (that branch carried the unmerged mlx-engine backend; the code half — skipping the optional `draft` checkpoint — is now its own PR #3253).

Follow-up to #2435.

- Checkpoint keys table: adds `draft` as a known type alongside `main`, `npu_cache`, etc.
- New "Multi-Token Prediction (MTP) models" section: `user.Gemma-4-MTP` registration example with `draft` checkpoint, `mtp` label, and `mmproj` — using the verified `unsloth/gemma-4-26B-A4B-it-GGUF` repo.
- `--checkpoint` flag help: documents `draft` and lists known types (including `text_encoder`).

### DIFF
--- a/docs/guide/configuration/custom-models.md
+++ b/docs/guide/configuration/custom-models.md
@@ -101,7 +101,7 @@ Supported registration flags:
 | Flag | Description |
 |------|-------------|
 | `--source SOURCE` | Remote registry for every checkpoint in this model: `huggingface` or `modelscope`. When omitted, the server's configured `default_model_source` applies. |
-| `--checkpoint TYPE CHECKPOINT` | Add a checkpoint entry. Repeat for multi-file models such as `main` + `mmproj`, `main` + `vae`, or `main` + `draft`. Known types: `main`, `mmproj`, `vae`, `draft`, `text_encoder`, `npu_cache`. |
+| `--checkpoint TYPE CHECKPOINT` | Add a checkpoint entry. Repeat for multi-file models. Known types: `main`, `mmproj`, `vae`, `draft`, `text_encoder`, `npu_cache`. |
 | `--recipe RECIPE` | Recipe to associate with the new `user.*` model. Common values: <!-- BEGIN GENERATED: recipe-values -->`llamacpp`, `whispercpp`, `moonshine`, `kokoro`, `sd-cpp`, `flm`, `ryzenai-llm`, `vllm`, `thenoise`, `thinksound`, `acestep`, `onnxruntime`, `trellis`, `openmoss`, `collection.omni`<!-- END GENERATED: recipe-values -->. |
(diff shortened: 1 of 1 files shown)


fl0rianr (changes requested):
Thanks for recreating this as a clean docs-only PR. The scope looks much better now, but I don't think the new docs are safe to merge yet.

A few things need fixing:

1. I'm not aware of a Gemma4 27B as there is only a similar sounding, well known 26B around. A current Unsloth example is `unsloth/gemma-4-26B-A4B-it-GGUF`, where the MTP drafter is `mtp-gemma-4-26B-A4B-it.gguf` in the same repo and the vision projector is `mmproj-F16.gguf`.

2. The statement that omitting `draft` makes the model "run normally but without MTP acceleration" is stronger than what the current Lemonade code guarantees. The `mtp` label still enables `--spec-type draft-mtp` even when no draft path resolves. This should be aligned with the actual behavior, ideally after #3253 has regression coverage.

3. The new "Known types" list omits `text_encoder`, even though the supported checkpoint table below lists it.

I'd also prefer `Follow-up to #2435` rather than `Fixes #2435`, since #2435 was already closed after the reporter confirmed the documented `draft` import approach works.

Please use a verified copy/pasteable model example and make the optional-draft wording match tested runtime behavior before we can consider merging.

bong-water-water-bong:
Thanks @fl0rianr — all four points addressed:

1. **Gemma 4 26B**: example now uses the verified `unsloth/gemma-4-26B-A4B-it-GGUF` repo with `mtp-gemma-4-26B-A4B-it.gguf` drafter and `mmproj-F16.gguf` projector (matches the built-in `Gemma-4-26B-A4B-it-MTP-GGUF` entry).
2. **Wording**: clarified that `mtp` only enables `--spec-type draft-mtp` when a `draft` checkpoint is present (this is now the actual behavior, per #3253).
3. **Known types**: added `text_encoder` to the `--checkpoint` known-types list.
4. **`Follow-up to #2435`**: updated the PR body.

jeremyfowers:
@bitgamma this PR documents features you added, please review before this merges.

fl0rianr:
This one has to land after #3253 IMO, it is already in a much better condition now.

bong-water-water-bong:
@fl0rianr thanks — agreed on the ordering. All four points are addressed here, so this is ready to land right after #3253.

bitgamma (approved):
Looks good to me. I took the liberty to change a little bit the wording on the `--checkpoint` documentation as it could have been misleading (since the examples were always listing only two checkpoints)

bong-water-water-bong:

PR #3259, opened 2026-08-20
feat(cli): return per-model detail from status --json

### LINKED ISSUE #3238: `lemonade status --json` returns almost no information; per-model detail only available via undocumented HTTP API
# `lemonade status --json` returns almost no information; per-model detail only available via undocumented HTTP API

## Summary

`lemonade status --json` currently returns just:

```json
{"port": 8123}
```

All of the actually useful per-model runtime state — `ctx_size`,
`llamacpp_device`, `status` (ready/loading/error), `pid`, `backend_url`,
`pinned`, etc. — is only obtainable by hitting the server's own HTTP endpoint
directly (`GET /api/v1/health`), which is undocumented as a public/stable
CLI-adjacent interface. Scripts that want to check "is model X loaded with
the config I expect" have to shell out to `curl` + parse JSON from an
internal API rather than using a documented CLI surface.

## Repro

```bash
lemonade status --json
# {"port": 8123}   -- basically nothing

curl -s http://localhost:$PORT/api/v1/health | python3 -m json.tool
# ... actual per-model detail lives here: model_name, recipe_options
#     (ctx_size, llamacpp_device, llamacpp_backend), status, pid, pinned, etc.
```

## Expected behavior

`lemonade status --json` should include the same per-model detail already
exposed by `/api/v1/health` — model name, recipe/backend, recipe_options
(ctx_size, device, backend), load status, pinned state — as a documented,
stable CLI output. This is the natural place to look for this data; needing
to know about and hand-parse an internal HTTP endpoint is a surprising
discovery for anyone scripting against lemonade.

## Impact

Any automation that needs to verify model load state (e.g., "reload only if
ctx-size or device changed from what I want") has to either shell out to
curl against an internal API, or guess/re-derive state instead of asking the
CLI. This also blocks portable tooling since the HTTP API's shape isn't
guaranteed stable/documented the way CLI output would be.

## Suggested fix

Extend `lemonade status --json` (and probably the human-readable `status`
output too) to include the loaded-model array with recipe_options and status,
mirroring what `/api/v1/health` already computes internally.

anditherobot:
## Summary

Fixes #3238.

`status --json` returned only the port. It now reads `/api/v1/health`, the same source the human output already used.

```json
// before
{"port": 13306}

// after
{
  "port": 13306,
  "version": "11.7.0",
  "websocket_port": 9000,
  "models": [
    {
      "model_name": "Tiny-Test-Model-GGUF",
      "checkpoint": "unsloth/gemma-3-270m-it-GGUF:gemma-3-270m-it-UD-IQ2_M.gguf",
      "type": "llm", "device": "gpu", "recipe": "llamacpp",
      "recipe_options": { "ctx_size": 2048, "llamacpp_backend": "vulkan" },
      "status": "ready", "pinned": false, "pid": 30932,
      "backend_url": "http://127.0.0.1:8001/v1"
    }
  ]
}
```

The human table gains `Ctx` and `Status` columns, which the issue also asks for.

Verified against an unreachable server (`--port` pointed at a port nothing is listening on):

**Human** (`lemonade status --port 19999`):
```
Error fetching health status: Could not connect to Lemonade server (Connection timed out).
Make sure the server is running and try again.
```
Exit code: `1`, stdout: empty

**JSON** (`lemonade status --port 19999 --json`):
```
Error fetching health status: Could not connect to Lemonade server (Connection timed out).
Make sure the server is running and try again.
```
Exit code: `1`, stdout: empty

Same output in both cases. Both paths now go through the same error-handling code (`print_health_fetch_error()` in `lemonade_client.cpp`), so this is guaranteed, not coincidental.

`launch_command` is left out so this does not depend on #3229.

## Testing

Built on Windows with the `vs18` preset.

- `status --json` against no server, an idle server, and a loaded model
- stdout carries JSON only, and stays empty when the fetch fails
- added `test_010a_status_json`; it and `test_010_status` pass

## Documentation

`docs/guide/cli.md` gains an `Options for status` section describing the JSON output. The command previously had only a one-line entry in the command table.

## AI-assisted contribution

Used AI tools for research and implementation guidance. Reviewed to make sure only the changes needed were made.


jeremyfowers (changes requested):
Overall this looks great to me, just 2 things needed.

1. Resolve merge conflict
2. Please re-evaluate the claim `"An unreachable server used to print "Server is not running" from a liveness ping. It now prints the same message the human path prints, with stdout empty and exit code 1."` Would you mind putting exactly what gets printed in both the human and --json cases for an unreachable server into the PR description? My reading of the code is that they are not identical.

jeremyfowers (approved):
Thanks @anditherobot! 

FYI our ci system is flakey so this might take a little while to merge today.

PR #3262, opened 2026-08-20
fix(gui): refactor memory settings to be server owned

fl0rianr:
## Summary

Align GUI3's global model lifecycle settings with lemond-owned state and policy.

The Memory settings page previously maintained its own model residency and eviction behavior in the browser. This removes that second lifecycle layer and uses the existing lemond runtime configuration instead:

- `max_loaded_models`
- `auto_evict`
- `auto_evict_threshold_pct`
- `auto_check_model_updates`

Model loading now sends the intended load request without performing client-side pre-eviction or OOM-based eviction/retry decisions.

The Memory settings UI keeps the existing GUI3 layout, but only exposes lifecycle controls backed by server state. Runtime values are refetched from lemond, save failures are surfaced, and Discard restores the authoritative server values instead of maintaining another set of defaults in the client.

System Vitals also shows the configured eviction threshold when automatic eviction is enabled. A marker is used when the visualization has a reliable memory total; otherwise the threshold is shown as text rather than drawing a misleading position.

The old browser-owned automatic model-update schedule is replaced by lemond's existing startup update-check setting.

Related to #3161

## Scope

But opted for an additional storage reduction...

## Testing

_Testing details:_

- Added runtime coverage for server-backed global model settings and localStorage cleanup.
- Added Playwright coverage for reading, saving, discarding, and failed writes of lifecycle runtime config.
- Covered `max_loaded_models = -1` and the custom number controls without allowing `0`.
- Covered conversion between the displayed eviction percentage and `auto_evict_threshold_pct`.
- Covered the 5 percentage-point eviction-threshold stepper.
- Covered System Vitals threshold marker/text fallback behavior.
- Verified model load paths no longer use the removed client-side lifecycle/eviction policy.
- Verified legacy browser lifecycle values are not migrated into server configuration.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -66,7 +66,6 @@ import {
   GLOBAL_MODEL_SETTINGS_EVENT,
   loadGlobalModelSettings,
   loadPinnedModelNames,
-  loadWithGlobalModelPolicy,
   savePinnedModelNames,
 } from '../features/modelSettings/globalModelSettings';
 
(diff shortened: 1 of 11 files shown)


jeremyfowers (approved):
LGTM!

PR #3265, opened 2026-08-20
fix(server): resolve custom args by scope

### LINKED ISSUE #3226: Problematic load/save semantics for args in the server
The Lemonade API spec allows for models to be load``ed with highly customized CLI options, which we call `*_args` (e.g., `llamacpp_args`). I will refer to these as "args" for short. We also support saving args, as well as server-provided default args.

In trying to implement GUI3, I have discovered that it is very difficult for a Lemonade client to navigate the semantics of loading and saving with args. Here are the problems in lemond on `main` as of this writing:

## Architectural Defaults

There is a JSON file that defines 6 per-model-architecture default sampling parameters (temp, top-k, etc.). These are automatically applied to models, and there is no way to disable this.

Users can override these args, for example by passing their own `--temp`. However, omitting one such value semantically means "I want the architectural default". So there is no way to override one arg _and_ pass no value for the others.

## Saved Value Merge on Load

Lemonade allows for args to be saved, for example `--no-mmap --threads 8`. These saved args are _merged_ with args passed in the load body (ie, the load body does not wholesale replace the saved args).

So if I saved `--no-mmap --threads 8` and then pass `--threads 4` in the load body, the result is `--no-mmap threads 4`. There is no way for me to pass my `--threads 4` _and_ tell Lemonade to omit the `--no-mmap` except by clearing the entire saved entry for that model.

## Saved Value Merge on Save

Saving new args does not wholesale replace previously saved args. For example, if I saved `--no-mmap` and then save `--threads 8`, now my saved args are `--no-mmap --threads 8`.

## Hidden and Conditional Defaults

The WrapedServer C++ for Lemonade also automatically applies a variety of args at runtime, some of which are static (like `-ngl 99`) and some of which are dynamic (like `--no-mmap` on APUs only). These arguments are not disclosed by the API, not even the `/models/{id}/options` endpoint, except that there are some cases where lemond will raise an HTTP error if the load body contradicts one of these args.

## Conclusion

It is a bug that clients are unable to exercise control, or gain visibility, over the args they are loading. We should propose a comprehensive solution to allow full control and visibility.

fl0rianr:
## Summary

Fix the `*_args` resolution semantics by separating backend/machine arguments from model-scoped and request-scoped arguments.

The effective custom-argument order is now:

`backend/machine -> model defaults -> explicit request -> runtime defaults`

The important behavior changes are:

- without request `*_args`, model and architecture args are inherited normally
- explicit request `*_args` replace the model/architecture `*_args` layer while preserving backend/machine args
- `merge_args=false` disables inherited custom args completely
- explicit empty `*_args` clears the model-scoped args for that load
- existing transient `null` semantics are preserved
- overridable llama.cpp runtime defaults are resolved into the effective options before backend startup instead of being added later as hidden arguments

This also makes the effective options a better representation of the backend configuration that will actually be launched.

The existing per-model save behavior remains key-based: supplying a new `*_args` value replaces the stored value rather than merging the argument strings.

Adds focused C++ regression coverage for the argument precedence and override rules.

Fixes #3226

## Scope

## Testing

_Testing details:_

```bash
cmake --build --preset default --target test_recipe_arg_resolution
ctest --test-dir build -R '^RecipeArgResolutionTest$' --output-on-failure
git diff --check
```

Regression coverage includes:

- backend + model args when no request `*_args` are supplied
- backend + request args when request `*_args` are supplied
- explicit empty request args
- transient `null` request behavior
- `merge_args=false` with and without request args
- conflicting flags such as `--threads`
- boolean negation pairs such as `--mmap` / `--no-mmap`
- `--flag=value` syntax
- repeatable arguments
- llama.cpp runtime-default overrides and aliases
- runtime defaults appearing in the resolved effective args

## Documentation

## Breaking Changes

- [ ] This PR does not introduce breaking changes.

The precedence of custom `*_args` is intentionally changed.

Clients that explicitly provide request `*_args` will no longer also receive model or architecture `*_args`. Backend/machine arguments are still inherited when `merge_args=true`.

Setting `merge_args=false` now means no lower-level custom arguments are inherited.

These changes are the intended semantics discussed in #3226.

## AI-assisted contribution

If AI tools were used:

### DIFF
--- a/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
@@ -119,12 +119,12 @@ static std::string resolve_llamacpp_runtime_args(const ModelInfo& model_info,
         {"--reasoning-format auto", "--reasoning-format", {}, false},
     };
 
-    const std::string draft_path = model_info.resolved_path("draft");
+    const std::string draft_checkpoint = model_info.checkpoint("draft");
     const bool has_dflash_label =
         std::find(model_info.labels.begin(), model_info.labels.end(), "dflash") !=
         model_info.labels.end();
     const bool is_dflash_draft =
-        !draft_path.empty() && is_dflash_draft_checkpoint(model_info.checkpoint("draft"));
+        !draft_checkpoint.empty() && is_dflash_draft_checkpoint(draft_checkpoint);
     const bool uses_mtp =
         std::find(model_info.labels.begin(), model_info.labels.end(), "mtp") !=
         model_info.labels.end();
--- a/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
@@ -115,9 +115,7 @@ static std::string resolve_llamacpp_runtime_args(const ModelInfo& model_info,
                                                  bool merge_args) {
     if (!merge_args) return custom_args;
 
-    std::vector<RuntimeArgDefault> defaults = {
-        {"--reasoning-format auto", "--reasoning-format", {}, false},
-    };
+    std::vector<RuntimeArgDefault> defaults;
 
     const std::string draft_checkpoint = model_info.checkpoint("draft");
     const bool has_dflash_label =
@@ -130,21 +128,9 @@ static std::string resolve_llamacpp_runtime_args(const ModelInfo& model_info,
         model_info.labels.end();
 
     if (is_dflash_draft && has_dflash_label) {
-        defaults.push_back({"--spec-type draft-dflash", "--spec-type", {}, false});
+        defaults.push_back({"--spec-type draft-dflash", "--spec-type"});
     } else if (uses_mtp) {
-        defaults.push_back({"--spec-type draft-mtp", "--spec-type", {}, false});
-    }
-
-    defaults.push_back({"--no-ui", "--no-ui", {}, true});
-
-    if (SystemInfo::get_has_igpu()) {
-        defaults.push_back({
-            "--load-mode none",
-            "--load-mode",
-            {"--direct-io", "--no-direct-io", "-dio", "-ndio", "--mmap",
-             "--no-mmap", "--mlock", "-lm"},
-            false,
-        });
+        defaults.push_back({"--spec-type draft-mtp", "--spec-type"});
     }
 
     return append_runtime_arg_defaults(custom_args, defaults);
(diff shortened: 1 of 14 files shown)


src/cpp/server/backends/llamacpp/llamacpp_server.cpp
```diff
@@ -147,6 +110,46 @@ static bool is_dflash_draft_checkpoint(std::string checkpoint) {
     return filename.rfind("dflash-", 0) == 0 || filename == "dflash.gguf";
 }
 
+static std::string resolve_llamacpp_runtime_args(const ModelInfo& model_info,
+                                                 const std::string& custom_args,
+                                                 bool merge_args) {
+    if (!merge_args) return custom_args;
+
+    std::vector<RuntimeArgDefault> defaults = {
```
bitgamma: this matches llamacpp's default. I think this can be dropped

src/cpp/server/backends/llamacpp/llamacpp_server.cpp
```diff
@@ -147,6 +110,46 @@ static bool is_dflash_draft_checkpoint(std::string checkpoint) {
...
+        std::find(model_info.labels.begin(), model_info.labels.end(), "mtp") !=
+        model_info.labels.end();
+
+    if (is_dflash_draft && has_dflash_label) {
+        defaults.push_back({"--spec-type draft-dflash", "--spec-type", {}, false});
+    } else if (uses_mtp) {
+        defaults.push_back({"--spec-type draft-mtp", "--spec-type", {}, false});
+    }
+
+    defaults.push_back({"--no-ui", "--no-ui", {}, true});
```
bitgamma: `--no-ui` is not really important, I would drop it

src/cpp/server/backends/llamacpp/llamacpp_server.cpp
```diff
@@ -147,6 +110,46 @@ static bool is_dflash_draft_checkpoint(std::string checkpoint) {
...
+
+    if (SystemInfo::get_has_igpu()) {
```
bitgamma: this behavior has recently become the default in upstream llamacpp (https://github.com/ggml-org/llama.cpp/pull/26081), so I would say we can also drop this one

bitgamma (approved):
Looks good to me. I left a few comments but up to you if you want to address them here.

Below also a few comments from Qwen3.8, again mostly nits.

 1. Redundant re-application: the resolve_llamacpp_runtime_args call in LlamaCppServer::load is dead weight today (router always materializes first). Harmless/idempotent, but a comment noting it's defensive — or removal — would prevent confusion.                                                                        
 2. Collection components get subtly different semantics: ensure_collection_loaded (server.cpp:2453) passes comp_info.recipe_options as the request layer for each component, so a component's saved *_args now replace the architecture layer, whereas for a regular model saved *_args are the model layer and still merge  
    with arch defaults. Defensible under the new global semantics (per-component saved options are the component's explicit layer), but worth a sentence in the PR description.                                                                                                                                               
 3. Empty-string injection: the resolver unconditionally set_option(key, resolved) even when resolved is "", so effective options for arg-less recipes can now contain "x_args": "" (the constructor would have dropped it). Cosmetic, but visible in the options GET endpoint.                                               
 4. Fragile invariant: inherit() constructs its result via the RecipeOptions constructor, which records every present key as "explicit" — so has_explicit_option is true for inherited values on inherit-derived objects. It works because only the original request object is ever queried, but the header comment           
    ("constructor input named this option") doesn't capture that inherited options also qualify.

bitgamma (approved):

fl0rianr:
Thanks for the review! I droped the three points noted and also addressed some of the Qwen3.8 suggestions, but NOT number 3 since we explicitely want that "save to empty" option.
I would appreciate if you can take a look the approval ist still valid - oh thanks, was too slow.

PR #3277, opened 2026-08-20
fix(llamacpp): pin --parallel 1 so ctx_size is not oversubscribed

### LINKED ISSUE #3276: llama.cpp: --ctx-size is applied per slot, so concurrent requests exhaust the KV cache
### Platform
Linux/Ubuntu

### Lemonade Version
11.7.0

### GPU / APU Model
2x AMD gfx1201 (ROCm0 / ROCm1)

### Component
llama.cpp

### Bug Description
Lemonade never passes `--parallel` to `llama-server`, so the slot count falls to llama.cpp's auto heuristic. On this host that resolves to 4:

```
srv load_model: initializing, n_slots = 4, n_ctx_slot = 131072, kv_unified = 'true'
```

`--ctx-size` is applied **per slot**, so the `ctx_size` recipe option is advertised to every slot while a single shared KV buffer backs all of them. Concurrent requests then oversubscribe the cache and fail as HTTP 500 `Context size has been exceeded` (with `truncated = 0`, i.e. rejected rather than trimmed), plus `failed to find free space in the KV cache` warnings and a batch-size death spiral (2048 -> 1024 -> ... -> 32).

Because the failure surfaces to the client as a request that never returns, it reads as a client timeout rather than a backend resource error. My log had 0 curl timeouts, 50 context-exceeded 500s, and 190 KV-pressure warnings.

### Steps to Reproduce
1. Load a model with a large explicit context:
   `lemonade run <model> --ctx-size 131072 --llamacpp-device ROCm0`
2. Confirm the backend log shows `n_slots = 4`.
3. Send 4 concurrent `/v1/chat/completions` requests with ~25k-token prompts to that one model.
4. Requests hang and/or return 500 `Context size has been exceeded`.

Adding `--llamacpp-args "--parallel 1"` makes the same test pass.

Expected: `ctx_size` is the context each request actually gets; concurrent requests queue.
Actual: `ctx_size` is promised to N auto-chosen slots at once; concurrent requests fail.

A/B on the same host and model:

| | `--parallel 1` | auto (`n_slots = 4`) |
|---|---|---|
| 4 concurrent 25k-token prompts | 4/4 -> 200 | 0/4 completed in 10 min |
| New ctx-exceeded errors | 0 | prompt eval collapsed 480 -> 6.53 tok/s |

With `--parallel 1` the log reads `n_slots = 1, n_ctx_slot = 131072, kv_unified = 'false'`.

### Log Output
```shell
# oversubscribed (default, no --parallel)
srv load_model: initializing, n_slots = 4, n_ctx_slot = 131072, kv_unified = 'true'
srv update_slots: decode() failed: Context size has been exceeded.
... failed to find free space in the KV cache

# after --llamacpp-args "--parallel 1"
srv load_model: initializing, n_slots = 1, n_ctx_slot = 131072, kv_unified = 'false'
```

### Additional Context
`grep -rn -- "--parallel\|n_parallel\|n_slots\|kv-unified" src/cpp/` returns nothing — no backend sets a slot/concurrency flag.

Proposed fix, one line in `src/cpp/server/backends/llamacpp/llamacpp_server.cpp` next to the existing `--ctx-size` push (~line 349):

```cpp
push_overridable_arg(args, llamacpp_args, "--parallel", "1", {"-np"});
```

`push_overridable_arg` is the right helper rather than `push_arg`: it is skipped when the user already passes `--parallel`/`-np` in `llamacpp_args`, and it does not add the flag to `reserved_flags`, so overriding stays legal. Same pattern already used for `--reasoning-format`, `--spec-type`, and `--load-mode`.

Note `compute_auto_context_size` in `src/cpp/include/lemon/auto_tune.h` (~L250-264) divides available memory by `kv_bytes_per_token` with no slot term — it sizes for a single sequence. Pinning one slot makes that assumption true, so no second change is needed there.

**Related issues**

- #3224 / #3241 — auto-tune *choosing* a bad ctx size on multi-GPU hosts. Complementary, not overlapping: that is the value being chosen, this is the chosen value being divided at runtime. #3241 arguably makes this more visible, since today the 4096 auto-tune floor can accidentally mask the oversubscription.
- #3160 — Qwen3.8-27B generation corruption under 8 concurrent requests at ctx 262144. Possibly the same root cause: 262144 per slot across auto-chosen slots is a much larger overcommit than the one here. Worth re-testing that repro with `--parallel 1` before pursuing an arch-specific explanation.
- #1836 — also concurrent-request failures, but caused by model eviction under `max_loaded_models=1`. Distinct mechanism; noting it so the two are not conflated.
- #3247 — KV cache quantization to fit the configured context. Same memory budget, different lever; a correct slot count should land before quantization is tuned against it.
- #1715 — request for `/slots` endpoint support. Pinning `--parallel 1` is a constraint on that design: if slots become user-visible, the ctx-size-per-slot accounting has to be surfaced with it.

Happy to open a PR for this.

jgmelber:
## Summary

`llama-server` auto-picks its slot count (`--parallel` defaults to `-1`), and an auto count also enables the unified KV buffer. The result is that the full `--ctx-size` is advertised to every slot off one shared allocation, so `ctx_size` is oversubscribed as soon as two requests overlap and they fail with `Context size has been exceeded` rather than queueing. Pinning `--parallel 1` makes `ctx_size` the context a request actually gets.

Fixes #3276

## Scope

## Testing

_Testing details:_

Built with `cmake --build --preset default`; `ctest -L cpp-ci` is 47/47.

Verified against a live `llama-server` (b10469) on a separate `lemond` instance, reading the resolved slot count out of the backend log:

| Load | `srv load_model: initializing` |
|---|---|
| before this change | `n_slots = 4, n_ctx_slot = 131072, kv_unified = 'true'` |
| after, no user args | `n_slots = 1, n_ctx_slot = 8192, kv_unified = 'false'` |
| after, `--llamacpp-args "--parallel 3"` | `n_slots = 3, n_ctx_slot = 2816, kv_unified = 'false'` |
| after, `--llamacpp-args "-np 2"` | `n_slots = 2, n_ctx_slot = 4096, kv_unified = 'false'` |

The last two confirm the override path: `push_overridable_arg` is skipped when either `--parallel` or the `-np` alias is present in `llamacpp_args`, and the flag is deliberately not added to `reserved_flags` so overriding stays legal. Same pattern as the existing `--reasoning-format`, `--spec-type`, and `--load-mode` defaults.

Those rows also show the mechanism precisely: with an explicit slot count llama.cpp *divides* the context (8192/3 -> 2816) and leaves `kv_unified` off. It is only the auto path that turns on the shared buffer and hands the undivided 131072 to all four slots.

Behavior under load, measured on 2x gfx1201 at `ctx_size 131072` with 4 concurrent 25k-token prompts to one model:

| | pinned to 1 slot | auto (`n_slots = 4`) |
|---|---|---|
| requests completing | 4/4 -> 200 | 0/4 in 10 min |
| new ctx-exceeded errors | 0 | prompt eval collapsed 480 -> 6.53 tok/s |

## Documentation

## Breaking Changes

Deployments relying on the implicit multi-slot default lose concurrent batching within a single model and get serialization instead. That default was not deliverable at the advertised `ctx_size` in the first place, and `--llamacpp-args "--parallel N"` restores it explicitly for anyone who sized their context for it.

## AI-assisted contribution

### DIFF
--- a/src/cpp/include/lemon/utils/recipe_arg_resolver.h
+++ b/src/cpp/include/lemon/utils/recipe_arg_resolver.h
@@ -27,6 +27,7 @@ struct ScopedCustomArgs {
 struct RuntimeArgDefault {
     std::string args;
     std::string flag;
+    std::vector<std::string> aliases = {};
 };
 
 inline bool is_custom_args_option(const std::string& key) {
@@ -72,8 +73,10 @@ inline std::string append_runtime_arg_defaults(
     std::vector<std::string> resolved_tokens = parse_custom_args(custom_args);
 
     for (const auto& runtime_default : defaults) {
-        const bool overridden =
-            custom_args_has_flag(resolved_tokens, runtime_default.flag);
+        bool overridden = custom_args_has_flag(resolved_tokens, runtime_default.flag);
+        for (const auto& alias : runtime_default.aliases) {
+            overridden = overridden || custom_args_has_flag(resolved_tokens, alias);
+        }
         if (overridden || runtime_default.args.empty()) continue;
 
         if (!result.empty()) result += " ";
--- a/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
@@ -338,6 +338,11 @@ void LlamaCppServer::load(const std::string& model_name,
     }
     push_arg(args, reserved_flags, "--ctx-size", std::to_string(ctx_size), std::vector<std::string>{"-c"});
 
+    // An auto slot count also enables the unified KV buffer, which advertises the
+    // full ctx_size to every slot instead of dividing it. Pinning the count keeps
+    // ctx_size the context a request actually gets.
+    push_overridable_arg(args, llamacpp_args, "--parallel", "1", {"-np"});
+
     if (!llamacpp_device.empty()) {
         BackendUtils::validate_device_backend_match(llamacpp_backend, llamacpp_device);
         push_arg(args, reserved_flags, "--device", llamacpp_device);
--- a/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
+++ b/src/cpp/server/backends/llamacpp/llamacpp_server.cpp
@@ -134,6 +134,11 @@ static std::string resolve_llamacpp_runtime_args(const ModelInfo& model_info,
         defaults.push_back({"--spec-type draft-mtp", "--spec-type"});
     }
 
+    // An auto slot count also enables the unified KV buffer, which advertises the
+    // full ctx_size to every slot instead of dividing it. Pinning the count keeps
+    // ctx_size the context a request actually gets.
+    defaults.push_back({"--parallel 1", "--parallel", {"-np"}});
+
     return append_runtime_arg_defaults(custom_args, defaults);
 }
 
@@ -337,11 +342,6 @@ void LlamaCppServer::load(const std::string& model_name,
         push_arg(args, reserved_flags, "-m", gguf_path, std::vector<std::string>{"--model"});
     }
     push_arg(args, reserved_flags, "--ctx-size", std::to_string(ctx_size), std::vector<std::string>{"-c"});
-
-    // An auto slot count also enables the unified KV buffer, which advertises the
-    // full ctx_size to every slot instead of dividing it. Pinning the count keeps
-    // ctx_size the context a request actually gets.
-    push_overridable_arg(args, llamacpp_args, "--parallel", "1", {"-np"});
 
     if (!llamacpp_device.empty()) {
         BackendUtils::validate_device_backend_match(llamacpp_backend, llamacpp_device);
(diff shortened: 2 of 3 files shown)


bitgamma (changes requested):
Thanks for your PR!

while I agree that this is a better default for most people, this is a breaking change as some users might be running parallel queries. @jeremyfowers this seems like something that needs to be flagged in the release notes. What is the correct way to do this?

Additionally, the PR needs to be rebased and the new default argument needs to be also added to `resolve_llamacpp_runtime_args`

jeremyfowers:
> while I agree that this is a better default for most people, this is a breaking change as some users might be running parallel queries. @jeremyfowers this seems like something that needs to be flagged in the release notes. What is the correct way to do this?

@jgmelber already did this correctly! The breaking change is clearly documented in the PR body, so repo-manager will pick that up and include it in the release notes.

bitgamma (approved):
LGTM, thanks!

PR #3280, opened 2026-08-20
feat(server): add context_length to the models endpoints

### LINKED ISSUE #3046: `/v1/models` response missing `context_length` — OpenAI-compatible clients cannot determine effective runtime context
### Platform

Linux/Arch

### Lemonade Version

11.5.2

### GPU / APU Model

AMD Ryzen AI MAX+ 395 (gfx1151 / Strix Halo), 96 GB unified LPDDR5X reported as GPU VRAM by ROCm

### Component

lemond (server)

### Bug Description

The `/v1/models` (and `/v1/models/{model_id}`) response does not include a `context_length` field. OpenAI-compatible clients use `context_length` to determine session budget, compression thresholds, and overflow handling. Without it, clients fall back to a hardcoded guess — Hermes, for example, defaults to 256,000 tokens — which causes context overflow mid-session when the actual loaded context is smaller.

`max_context_window` is returned, but it reflects the **GGUF native training maximum** (e.g. 1,048,576 for Laguna S 2.1), not the effective runtime context the server was launched with. The effective loaded context (`ctx_size`) is present in `recipe_options` but not surfaced as a top-level field clients can discover.

### Steps to Reproduce

1. Register a model with an explicit `ctx_size` in `recipe_options` (e.g. `ctx_size: 131072`)
2. Query the models endpoint:
   ```bash
   curl -s http://localhost:13305/v1/models/<model-id> | python3 -m json.tool
   ```
3. Observe that `context_length` is absent and `max_context_window` reflects the GGUF training maximum, not the loaded context

**Expected:** `context_length` is present and reflects the effective runtime context window — what the server will actually accept.

**Actual:** `context_length` is absent. `max_context_window` is set to the GGUF training maximum (1,048,576), not the loaded context (131,072).

**Proposed priority order for populating `context_length`:**
1. `recipe_options.ctx_size` — explicit per-model override (highest priority)
2. Auto-calculated value from VRAM-based auto-sizing (when `ctx_size: -1`)
3. Global `ctx_size` config value
4. `max_context_window` — GGUF native maximum (last resort; always present, always has a value)

This means `context_length` is always populated — never absent.

### Log Output

```json
{
  "id": "Laguna-S-2.1-APEX-compact",
  "max_context_window": 1048576,
  "recipe_options": {
    "ctx_size": 131072
  }
}
```

Hermes log when connecting to Lemonade:
```
WARNING agent.model_metadata: Could not determine context length for model
'Laguna-S-2.1-APEX-compact' — falling back to 256,000 tokens. Set
model.context_length in config.yaml to override.
```

Runtime overflow when Hermes compresses to ~132K and sends to a 131K context:
```
request (132,951 tokens) exceeds the available context size (131,072 tokens),
try increasing it
```

### Additional Context

`context_length` is the de facto field used by OpenAI-compatible clients (OpenRouter, together.ai, and others). Hermes's own warning message explicitly says to set it as a client-side override — but the current workaround forces a single value at the **provider level**, meaning all models served by that Lemonade instance inherit the same context length. If different models are loaded with different context sizes, there is no per-model workaround available to the client — one value will always be wrong.

The data is always knowable at serve time — Lemonade chose the `ctx_size` when loading the model. It just isn't being surfaced in the API response.

Related: Lemonade issue #249 (closed) — context window auto-sizing.

anditherobot:
## Summary

Fixes #3046.

`/v1/models` only returned `max_context_window`, which is the model's
built-in maximum rather than the size it is actually served with.
Clients that size a session from `context_length` had nothing to read,
so they guessed and ran out of room mid-session.

`context_length` now resolves in order:

1. the running backend's `ctx_size`, when the model is loaded
2. the configured `ctx_size`, per model then server-wide
3. `max_context_window` as a floor

I did not implement the VRAM auto-sizing step listed in the issue. It
may be a little out of scope here and probably belongs in its own
issue. Would you prefer it in this PR, or tracked separately?

Models with no context window, such as Whisper and Kokoro, currently
omit the field.

## Testing

Builds clean. Four new cases in `test/server_endpoints.py` cover an
explicit `ctx_size`, an inherited global, the `-1` automatic sentinel,
and list/retrieve agreement, and all pass.

## Documentation

Documentation is affected and has been updated: `docs/api/openai.md`.

## AI-assisted contribution

Used AI tools for research and implementation guidance. Reviewed to
make sure only the changes needed were made.

### DIFF
--- a/docs/api/openai.md
+++ b/docs/api/openai.md
@@ -1045,7 +1045,7 @@ curl http://localhost:13305/v1/models?show_all=true
   - `recipe` - Backend/device recipe used to load the model (e.g., `"ryzenai-llm"`, `"llamacpp"`, `"flm"`)
   - `size` - Model size in GB (omitted for models without size information)
   - `max_context_window` - Optional integer indicating the maximum model-supported text context discovered from local static metadata. Currently populated for downloaded GGUF/llama.cpp models and installed FLM text-context models.
-  - `context_length` - Integer indicating the effective context window the model is served with, rather than the native maximum in `max_context_window`. Populated from the loaded context when the model is running and the configured `ctx_size` otherwise (omitted when neither is known).
+  - `context_length` - Number of tokens the model can handle in one request. Uses the loaded value when the model is running and the configured `ctx_size` otherwise (omitted when neither is known).
   - `downloaded` - Boolean indicating if the model is downloaded and available locally
   - `update_available` - Boolean indicating a newer commit exists on HuggingFace for this model. Only set for downloaded HF-backed models. `false` otherwise.
   - `suggested` - Boolean indicating if the model is recommended for general use
--- a/docs/api/openai.md
+++ b/docs/api/openai.md
@@ -956,7 +956,7 @@ The generated audio file is returned as-is.
 ## `GET /v1/models`
 <sub>![Status](https://img.shields.io/badge/status-fully_available-green)</sub>
 
-Returns a list of models available on the server in an OpenAI-compatible format. Each model object includes extended fields like `checkpoint`, `recipe`, `size`, `downloaded`, `labels`, and, when known, `max_context_window`.
+Returns a list of models available on the server in an OpenAI-compatible format. Each model object includes extended fields like `checkpoint`, `recipe`, `size`, `downloaded`, `labels`, `context_length`, and, when known, `max_context_window`.
 
 By default, only models available locally (downloaded) are shown, matching OpenAI API behavior.
 
@@ -993,6 +993,7 @@ curl http://localhost:13305/v1/models?show_all=true
       "recipe": "llamacpp",
       "size": 0.38,
       "max_context_window": 40960,
+      "context_length": 8192,
       "downloaded": true,
       "suggested": true,
       "update_available": false,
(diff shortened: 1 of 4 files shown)


src/cpp/server/server.cpp
```diff
@@ -2892,6 +2917,14 @@ nlohmann::json Server::model_info_to_json(const std::string& model_id, const Mod
         model_json["max_context_window"] = info.max_context_window;
     }
 
+    // OpenAI-compatible clients read context_length to budget a session.
```
bitgamma: the first line of the comment is enough, the comment about max_context_window is unnecessary

src/cpp/server/server.cpp
```diff
@@ -2843,6 +2840,34 @@ void Server::handle_model_register(const httplib::Request& req, httplib::Respons
...
+int64_t Server::resolve_context_length(const std::string& model_id, const ModelInfo& info) const {
+    // ctx_size stores -1 for "size this automatically", so only a positive
+    // value answers; anything else falls through to the next source.
+    auto ctx_size_of = [](const RecipeOptions& options) -> int64_t {
+        const nlohmann::json ctx_json = options.get_option("ctx_size");
+        return ctx_json.is_number() ? ctx_json.get<int64_t>() : 0;
+    };
+
+    if (router_) {
+        // A live backend was started with a resolved ctx_size, so it answers
```
bitgamma: this comment is unnecessary, behavior is clear from the code

docs/api/openai.md
```diff
@@ -1044,6 +1045,7 @@ curl http://localhost:13305/v1/models?show_all=true
   - `recipe` - Backend/device recipe used to load the model (e.g., `"ryzenai-llm"`, `"llamacpp"`, `"flm"`)
   - `size` - Model size in GB (omitted for models without size information)
   - `max_context_window` - Optional integer indicating the maximum model-supported text context discovered from local static metadata. Currently populated for downloaded GGUF/llama.cpp models and installed FLM text-context models.
+  - `context_length` - Integer indicating the effective context window the model is served with, rather than the native maximum in `max_context_window`. Populated from the loaded context when the model is running and the configured `ctx_size` otherwise (omitted when neither is known).
```
bitgamma: "rather than the native maximum in `max_context_window`" is unnecessary

bitgamma (changes requested):
Thanks for the PR! Looks good, just a few things worth addressing:

1. Alias entries can report an inconsistent value. In handle_models, alias entries call model_info_to_json(alias_id, target_info). get_model_recipe_options(alias_id) resolves names only through public_model_aliases_ (not user aliases), so alias entries typically fall to the configured/floor value rather than the canonical entry's loaded value.

2. Some noisy comments/documentation entries.

docs/api/openai.md
anditherobot: That makes sense . I have fixed it :)

bitgamma (approved):
lgtm

PR #3281, opened 2026-08-21
feat(config): add configurable download rate limit

blackdeathdrow:
## Summary

Adds a configurable download rate limit for backend/model downloads. The limit is configured in `config.json`:

- `download_rate_limit` — active limit as a string (e.g. `"10M"`, `""` or `"0"` = unlimited).

Fixes #1585

## Scope

## Testing

_Testing details:_
- C++ unit test `test/cpp/test_config_download_rate_limit.cpp`
- Python integration tests
  `test_056` (validate + round-trip via `/internal/set`),
  `test_057` (invalid config value does not crash server startup),
  `test_058` (runtime vs persisted semantics).
- `python docs/tools/gen_backend_boilerplate.py --check`: all generated files up to date.

## Documentation

  (configuration/README.md config-example + table, runtime.md, getting-started.md)

## Breaking Changes

## AI-assisted contribution

If AI tools were used:

### DIFF
--- a/src/cpp/resources/defaults.json
+++ b/src/cpp/resources/defaults.json
@@ -14,6 +14,7 @@
   "ctx_size": -1,
   "default_model_source": "huggingface",
   "disable_model_filtering": false,
+  "download_rate_limit": "",
   "enable_dgpu_gtt": false,
   "extra_models_dir": "",
   "flm": {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -979,6 +979,7 @@ set(SOURCES_CORE
     src/cpp/server/utils/json_utils.cpp
     src/cpp/server/utils/process_manager.cpp
     src/cpp/server/utils/path_utils.cpp
+    src/cpp/server/utils/rate_limit_utils.cpp
     src/cpp/server/utils/url_utils.cpp
     src/cpp/server/utils/version_utils.cpp
     src/cpp/server/utils/wmi_helper.cpp
@@ -2974,6 +2975,19 @@ if(BUILD_TESTING AND EXISTS "${_CONFIG_TELEMETRY_TEST_SRC}")
     add_executable(test_config_telemetry test/cpp/test_config_telemetry.cpp)
     target_link_libraries(test_config_telemetry PRIVATE lemonade-server-core)
     add_cpp_ci_test(ConfigTelemetryTest CI ON COMMAND test_config_telemetry)
+endif()
+
+set(_CONFIG_DOWNLOAD_RATE_LIMIT_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_config_download_rate_limit.cpp"
+)
+if(BUILD_TESTING AND EXISTS "${_CONFIG_DOWNLOAD_RATE_LIMIT_TEST_SRC}")
+    add_executable(test_config_download_rate_limit
+        test/cpp/test_config_download_rate_limit.cpp
+    )
+    target_link_libraries(test_config_download_rate_limit PRIVATE lemonade-server-core)
+    add_cpp_ci_test(ConfigDownloadRateLimitTest CI ON
+        COMMAND test_config_download_rate_limit
+    )
 endif()
 
 set(_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC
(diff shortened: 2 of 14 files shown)


src/cpp/resources/defaults.json
```diff
@@ -14,6 +14,8 @@
   "ctx_size": -1,
   "default_model_source": "huggingface",
   "disable_model_filtering": false,
+  "download_rate_limit": "",
+  "download_rate_limit_options": [],
```
jeremyfowers: This is making the rate options displayed in the tray configurable by the user? It might be more user friendly to detect this automatically and offer 100%, 50%, and 25% of the user's peak bandwidth instead.

jeremyfowers:

**Attention level:** Routine — needs any 1 reviewer. Covered by: jeremyfowers.

Why this rung: a field or parameter on an existing route. The diff adds a field on the existing POST /api/v1/params route and two config options, both listed as one-reviewer in the table.

The description accurately covers the two config keys, the runtime-vs-persisted semantics, and the tray submenu — all confirmed in the diff.

</details>

All changes serve the single goal of adding a configurable download rate limit: config parsing, HTTP endpoint handling, tray submenu, C++ and Python tests, and documentation updates.

Checked: Every changed file serves the single goal: new rate_limit_utils module, config parsing, HTTP endpoint handling, tray submenu, CMakeLists additions, C++ and Python tests, and three documentation files.

</details>

<details><summary>Alignment issues: <b>none found</b></summary>

Checked: Reviewed contribute.md AI Contribution Policy and Merging a Contribution sections; the author checked all four AI-assisted coding items and the second commit (hardening menu callback, rejecting zero-rate options, deduplicating harness) demonstrates self-review of generated code.

</details>

Checked: Reviewed docs/dev/documentation.md; inspected docs/guide/configuration/README.md for the example block and parameter table, docs/dev/getting-started.md and docs/embeddable/runtime.md for the runtime config table entries, and src/cpp/resources/defaults.json for matching defaults — all four surfaces updated in the same PR.

</details>

Checked: Reviewed testing.md Where Tests Go table and CI rules; C++ test registered via add_cpp_ci_test(ConfigDownloadRateLimitTest CI ON COMMAND test_config_download_rate_limit) with if(BUILD_TESTING AND ...) guard in CMakeLists.txt; Python tests test_056/test_057/test_058 added to test/server_endpoints.py which runs on every PR.

</details>

<details><summary>Breaking changes: <b>none found</b></summary>

Checked: Scanned all 18 changed files for removed or renamed API endpoints, CLI flags, config keys, persisted data formats, and GUI elements — none found.

</details>

_Reviewed at head `a51b5e8` by repo-manager (tier: reviewers). Regenerate with `repo-manager review-pr 3281`._

jeremyfowers:
@abn can I interest you in reviewing this one? You've been working with some related code lately.

src/cpp/resources/defaults.json
blackdeathdrow: Yes, exactly. You are right that a list of fixed percentages would look nicer in the menu. However, it would require periodic bandwidth measurements and still wouldn't account for traffic consumed by other hosts on the network. That being said, if you prefer the automatic approach, I am open to changing the implementation.

abn: @blackdeathdrow We should consider whether the tray submenu is necessary. Download rate limiting is typically a set-and-forget preference rather than an on-the-fly tray toggle. If we drop the tray submenu and `download_rate_limit_options`, we can simplify this down to a single clean `download_rate_limit` key in `config.json` / `/internal/set` / `lemonade config` (and let the desktop app expose a proper input field in its settings panel in a follow-up), keeping the tray menu lean.

abn (commented):
The changes looks good, however I have to ask if the added systray bloat is worth it as it might not be a common quick action use case. See my comment ~above~ below.

It would be great to get some context if I am missing something here.

src/cpp/resources/defaults.json
blackdeathdrow: @abn Done. Now we can set only `download_rate_limit` key in `config.json`, without using the tray submenu.

abn (changes requested):
Thanks for updating this and removing the tray submenu! The configuration surface is much simpler and more focused now.

However, in Lemonade's security model, global resource allocation and host quota controls belong under `/internal/*` protected by `LEMONADE_ADMIN_API_KEY` (or `config.json` / `lemonade config set`) to prevent unprivileged consumers from causing resource starvation in shared or multi-client environments.

Because `download_rate_limit` is **process-global**, mutating `HttpClient` state and serializing background transfers across all clients, automated model updates, and backend installs. It should **not** be mutable via the public `POST /api/v1/params` endpoint by standard `LEMONADE_API_KEY` holders. 

**Suggestion**: Restrict `download_rate_limit` updates exclusively to `POST /internal/set` (and inspectable via `GET /internal/config`), disallowing it on `POST /api/v1/params`.

blackdeathdrow:
#3322 retired the undocumented `/api/v1/params`, so `download_rate_limit` is now mutable only via POST `/internal/set` (+ `config.json` / `lemonade config set`).

abn (approved):
LGTM! 🚀

PR #3282, opened 2026-08-21
fix(server): report NVIDIA GPU and VRAM usage

fl0rianr:
## Summary

Add NVIDIA GPU utilization and VRAM usage to `/api/v1/system-stats` using dynamically loaded NVML on Windows and Linux.

The existing AMD Linux metrics path remains unchanged. NVML is loaded at runtime, so this does not add a build-time NVIDIA dependency.

## Scope

## Testing

_Testing details:_
- Tested on an NVIDIA system.
- Verified `/api/v1/system-stats` reports `gpu_percent` and `vram_gb`.
- Verified the Monitor view of GUI3 displays NVIDIA GPU utilization and VRAM usage.
- Confirmed the existing AMD Linux metrics path is preserved.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:


src/cpp/server/platform/nvidia_metrics.h
```diff
@@ -0,0 +1,260 @@
...
+
+    void unload_library() {
+        if (library_) {
+            FreeLibrary(library_);
+            library_ = nullptr;
+        }
+    }
+#elif defined(__linux__)
+    static LibraryHandle open_library() {
+        for (const char* name : {"libnvidia-ml.so.1", "libnvidia-ml.so"}) {
```
superm1: Are these normally in LD_LIBRARY_PATH?

fl0rianr: Not necessarily, dlopen by soname also uses the loader cache/default library paths. Good point though: system_info.cpp already has an explicit /usr/lib/x86_64-linux-gnu fallback for NVML, so I'll mirror that here.

src/cpp/server/platform/metrics_linux.cpp
```diff
@@ -120,16 +122,17 @@ class LinuxMetricsPlatform : public SystemMetricsPlatform {
...
         try {
             std::string drm_path = "/sys/class/drm";
 
             if (!fs::exists(drm_path)) {
-                return -1.0;
+                return nvidia.vram_used_gb;
             }
 
-            double highest_usage = -1.0;
+            double highest_usage = nvidia.vram_used_gb >= 0.0 ? nvidia.gpu_percent : -1.0;
```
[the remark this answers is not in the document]
fl0rianr: Good catch. GPU selection should stay based on utilization independently of whether the memory query succeeded. I'll seed highest_usage directly from nvidia.gpu_percent, so an unavailable NVIDIA memory sample can't make us report VRAM from a less-active DRM device.

src/cpp/server/platform/nvidia_metrics.h
```diff
@@ -0,0 +1,261 @@
...
+            if (void* library = dlopen(name, RTLD_NOW | RTLD_LOCAL)) {
+                return library;
+            }
+        }
+        return dlopen(
+            "/usr/lib/x86_64-linux-gnu/libnvidia-ml.so.1", RTLD_NOW | RTLD_LOCAL);
```
superm1: this probably won't work on DGX spark

kenvandine: This definitely wouldn't work on the DGX Spark. It's also not going to work in the snap, and probably the flatpak as well.

fl0rianr: Good point. The explicit x86_64 fallback I added after the earlier review is too platform-specific. The initial soname lookup is the portable path and lets the system loader handle architecture-specific and sandbox-provided library locations. I'll drop the absolute fallback rather than add more architecture or packaging-specific paths.

kenvandine (changes requested):
Thanks for adding NVIDIA VRAM metrics via NVML — useful signal for /system-stats and downstream sizing. A few correctness concerns in the multi-GPU and lifecycle handling, plus a duplication note:

  Correctness

  - get_vram_usage_gb() (metrics_linux.cpp:133) can silently discard a correct NVIDIA VRAM reading in favor of an unrelated AMD card. The selection loop seeds highest_usage/highest_card_memory from the NVIDIA reading, but overrides both as soon as any AMD GPU/iGPU reports gpu_usage > 0 — which happens routinely from desktop compositing even when the AMD card is idle for compute. On a hybrid AMD+NVIDIA box with a model
  loaded on the NVIDIA GPU, this replaces the real VRAM figure with the AMD card's near-zero usage. This isn't just a UI cosmetic issue — it feeds router.cpp's get_available_memory_gb() and auto_tune.h's layer-offload sizing, so it can affect load/eviction decisions.
  - Related: in nvidia_metrics.h:228, when NVML's utilization query fails (e.g. virtualized/passthrough GPU) but the memory query succeeds, gpu_percent is forced to -1.0 even though vram_used_gb is valid. That -1 then trips the highest_usage < 0 override branch in get_vram_usage_gb() and discards the valid VRAM reading in favor of the first AMD card found, regardless of that card's actual usage.
  - On multi-GPU NVIDIA systems, sample_uncached() (nvidia_metrics.h:196) picks a single "best" GPU by utilization rather than aggregating across all NVIDIA GPUs. An idle GPU holding 20GB of a loaded model loses to a lightly-busy GPU holding 0GB, so vram_used_gb reports ~0 despite 20GB actually being in use.
  - NvmlProbe's destructor (nvidia_metrics.h:45) calls shutdown_() during static destruction at process exit. If a background thread is still mid-poll on /system-stats when shutdown starts, this races NVML/driver state teardown against a live call into the same probe — a known-fragile pattern for GPU driver libraries invoked from atexit-time destructors in a multithreaded process. Worth shutting NVML down explicitly and
  deterministically (e.g. from the server's shutdown path) rather than relying on static destruction order.
  - query_nvidia_nvml() in system_info.cpp:2489 calls nvmlInit()/nvmlShutdown() around every invocation, independent of the long-lived NvmlProbe singleton in nvidia_metrics.h, which expects NVML to stay initialized for the process lifetime. If a system-info cache refresh runs concurrently with a /system-stats poll, the transient nvmlShutdown() can tear down shared driver state out from under the persistent probe on
  implementations where init/shutdown ref-counting isn't safe across independent dlopen handles — could leave metrics stuck returning -1 until restart. Might be worth having both paths share one NVML lifecycle owner.

  Duplication

  - nvidia_metrics.h reimplements the NVML dynamic-loading pattern (ABI structs, symbol-name fallbacks, dlopen/dlsym) that already exists in system_info.cpp's query_nvidia_nvml(). The two copies have already drifted — system_info.cpp tries libnvidia-ml.so.1 then a hardcoded x86_64 path, while nvidia_metrics.h tries libnvidia-ml.so.1 then libnvidia-ml.so. Worth consolidating into one shared NVML-loading utility so future
  ABI/symbol fixes don't need to be applied twice.

  Happy to take another look once these are sorted, especially the hybrid-GPU override and the shutdown race.

kenvandine (changes requested):
Thanks for the update — this round restructures things to merge NVML data directly into Server::get_gpu_usage()/get_vram_usage(). A few correctness issues in the new merge/lifecycle logic, plus some cleanup items:

Correctness

  - get_vram_usage() (server.cpp:6779) combines platform_vram_gb + nvidia.vram_used_gb, but these two numbers aren't computed the same way: on Linux, metrics_platform_->get_vram_usage_gb() collapses multiple AMD cards down to just the busiest card's VRAM, while aggregate_nvidia_metrics() sums VRAM across every NVIDIA device. So on a hybrid AMD+NVIDIA box the result is "busiest AMD card" + "sum of all NVIDIA cards" — a
  number that doesn't correspond to any real quantity, and it's inconsistent with the max()-based combination used for gpu_percent a few lines above. Worth using a consistent rule (max or sum) for both, and applying it uniformly per-vendor.
  - The Linux NVML loader (nvidia_metrics.h:206) dropped the hardcoded absolute-path fallback (/usr/lib/x86_64-linux-gnu/libnvidia-ml.so.1) that the old system_info.cpp code used, replacing it with the unversioned libnvidia-ml.so soname. The comment still sitting above query_nvidia_nvml() explains that fallback exists specifically for "AppArmor restrictions in snap strict confinement" — under that exact scenario, NVML
  may now fail to load entirely instead of falling back successfully, silently disabling GPU telemetry on affected installs.
  - The NVML library handle (nvidia_metrics.h:268) is loaded once into a process-lifetime singleton and never reloaded, unlike the old per-call dlopen/dlclose. If the NVIDIA driver is upgraded/reinstalled while lemond keeps running, the mapped library can go ABI-mismatched with the running kernel module — NVML calls then silently fail or return stale data until a full restart, which is a new failure mode this change
  introduces.
  - A single global mutex (nvidia_metrics.h:310) serializes every NVML query across all callers. Per AGENTS.md's many-clients-one-server topology, multiple desktop/tray/CLI clients plus a Prometheus scraper can poll concurrently — NVML is known to stall for hundreds of ms during GPU power-state transitions, and while the mutex is held for that stall, every other client's /system-stats//metrics/job system_stats op blocks
  behind it.

Efficiency

  - The 250ms cache TTL (nvidia_metrics.h:301) barely amortizes anything against real polling intervals — the desktop app polls /system-stats every 5s, so nearly every poll is a cache miss paying a full nvmlInit → query → nvmlShutdown round trip. Since the library handle is now kept persistent anyway, consider keeping NVML initialized persistently too and just issuing lightweight per-poll queries.

Simplification

  - The NVML merge logic is bolted directly onto Server::get_gpu_usage()/get_vram_usage() rather than living inside SystemMetricsPlatform. This leaves the platform abstraction stale — metrics_windows.cpp's "GPU usage monitoring not implemented for Windows" comment is now inaccurate on NVIDIA/Windows — and means a third GPU vendor would require copy-pasting the merge branch into both functions again.
  - get_gpu_usage() and get_vram_usage() duplicate the same three-way merge branching with only the combining operator (max vs. +) differing — worth factoring into a shared helper so a future fix to merge semantics (e.g. the VRAM aggregation issue above) doesn't need to be applied twice.
  - The new BYTES_PER_GIB constant in nvidia_metrics_detail duplicates the existing lemon::BYTES_PER_GIB in auto_tune.h:25 — worth reusing instead of a second source of truth.

Style

  - #include "platform/nvidia_metrics.h" in server.cpp:35 (and system_info.cpp:3) is out of alphabetical order, per AGENTS.md's C++ include-ordering rule.
  - The new comments on get_gpu_usage()/get_vram_usage() (server.cpp:6765) just restate the function names (WHAT), while the actually non-obvious part — why utilization uses max() but VRAM uses summation, and whether that risks double-counting — is left undocumented. Per AGENTS.md, comments should explain the non-obvious WHY, not the WHAT.

Happy to take another pass once these land, particularly the VRAM combination-rule inconsistency and the confinement fallback regression.

kenvandine:
Reviewed this again — found 4 issues worth addressing before merge:

  1. Auto-tune still uses the old VRAM lookup (src/cpp/include/lemon/auto_tune.h:70)
  get_used_memory_gb() calls the legacy single-device get_vram_usage_gb() rather than the new get_system_gpu_metrics() this PR introduces. On Windows that old path is hardcoded to return -1.0 (never implemented), so auto-tune falls through to using system RAM as a VRAM proxy — even though this PR
  now has real NVML VRAM data available on the same class. Worth wiring auto-tune to the new metrics path too, or at least tracking it as a follow-up.

  2. query_nvidia_nvml_devices() reloads the NVML library on every call (src/cpp/server/platform/nvidia_metrics.h:301)
  No caching — each call does a full dlopen/nvmlInit/nvmlShutdown/dlclose cycle. This is now on three hot paths: the job-engine system_stats op, /metrics (Prometheus scrapes), and /system-stats (polled by the GUI). On non-NVIDIA Linux boxes, open_library() also retries 6 hardcoded dlopen
  candidates that are guaranteed to fail — repeated on every poll. Given how frequently these endpoints get hit, some form of caching (e.g. keep the library handle alive, or cache "no NVIDIA present" after first failure) seems worth adding.

  3. LinuxMetricsPlatform::get_gpu_usage() is now dead code (src/cpp/server/platform/metrics_linux.cpp:155)
  Nothing calls it anymore — LinuxMetricsPlatform now overrides get_system_gpu_metrics() directly instead of going through the base class default that used to call this. It's still a plausible-looking 33-line method, which risks a future fix landing here and silently not affecting the real
  /system-stats//metrics output. Suggest removing it (or its sibling get_vram_usage_gb() at line 189, which still duplicates the sysfs walk done by the new query_drm_system_gpu_metrics()) if it's confirmed unused.

  4. Looser NVML readiness gate could desync /system-info vs /system-stats (src/cpp/server/platform/nvidia_metrics.h:294)
  The new gate only requires one of get_name_/get_utilization_/get_memory_ to resolve via dlsym, vs. the old code which hard-required nvmlGetName. If name resolution fails on some driver/NVML combo but utilization/memory still resolve, NvidiaNvmlDevice.name is empty. system_info.cpp's
  query_nvidia_nvml() filters out empty-name devices (reports 0 GPUs), while aggregate_nvidia_metrics() (used by /system-stats and /metrics) has no such filter and keeps reporting utilization/VRAM — so a GPU could show up as present in stats but absent in system-info. Worth confirming this is
  intentional or tightening the gate back to match.

kenvandine:

jeremyfowers:
@kenvandine what's the deal here? Please request further changes or approve.

PR #3283, opened 2026-08-21
test(telemetry): prevent dangling RuntimeConfig pointer and worker race in telemetry helpers test

abn:
## Summary

Fixes intermittent segmentation fault in `TelemetryHelpersTest` caused by a race condition with stack-allocated `RuntimeConfig` pointers and asynchronous background workers:

- Reset the global `RuntimeConfig` pointer on each iteration of the Unicode bound truncation test loop before stack-allocated config objects go out of scope.
- Set `telemetry.enabled: false` in test configs so unit test spans are handled synchronously via span listeners without enqueueing live network exports to the background worker thread.
- Explicitly flush and shut down telemetry workers (`lemon::telemetry::flush()` and `lemon::telemetry::shutdown()`) at test completion to ensure clean background thread teardown.

## Scope

## Testing

_Testing details:_
- Executed 20-iteration stress run of `test_telemetry_helpers`: 20/20 runs passed with zero segfaults.
- Ran `ctest -L cpp-ci`: all C++ unit tests passed.
- Verified backend boilerplate sync via `python3 docs/tools/gen_backend_boilerplate.py --check`: 0 drift.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/test/cpp/test_telemetry_helpers.cpp
+++ b/test/cpp/test_telemetry_helpers.cpp
@@ -492,7 +492,7 @@ int main() {
             {"port", 13305},
             {"host", "localhost"},
             {"telemetry", {
-                {"enabled", true},
+                {"enabled", false},
                 {"max_attribute_length", 20},
                 {"otlp", {
                     {"semantics", {"openinference", "otel_genai"}}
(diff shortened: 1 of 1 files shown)


fl0rianr (approved):
Thanks! LGTM

PR #3285, opened 2026-08-21
test: migrate model registry label validation to C++

fl0rianr:
## Summary

Move model registry deployment-mode validation from the standalone Python test into the existing C++ backend mode contract test.

The registry is now validated against the real `find_deployment_mode()` and backend descriptor logic, removing the duplicated Python mapping and regex parsing of C++ headers. The dedicated Python CI job is removed because the replacement runs through the normal `cpp-ci` CTest path.

## Scope

## Testing

*Testing details:*

* `BackendModeContractTest` passes.
* Full `ctest --test-dir build -L '^cpp-ci$' --output-on-failure` passes: 51/51 tests.
* Confirmed `BackendModeContractTest` is included in the `cpp-ci` label set.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:


abn (approved):
LGTM

PR #3287, opened 2026-08-21
test: move llamacpp system backend policy coverage to C++

fl0rianr:
## Summary

This PR moves the pure `llamacpp` system-backend checks out of the server-heavy Python test and into direct C++ coverage.

The practical result is that `test_llamacpp_system_backend.py` goes from **14 `lemond` starts/stops to 1**.

Previously, the suite repeatedly restarted the full server just to change things like `PATH`, `prefer_system`, fallback conditions, or the HIP plugin location, then read the resulting backend state through HTTP. Those decisions are C++ backend-selection logic and do not need a running server.

The new C++ tests exercise the real production paths directly. They cover system `llama-server` discovery, installed/unsupported state, `prefer_system`, explicit system selection, fallback behavior, and HIP plugin/path validation.

This also avoids reproducing the backend-selection logic in test-side mocks. If that logic changes, the tests now exercise the changed production code directly instead of relying on a parallel mocked version that could drift from the implementation.

The Python suite keeps the cases where the process boundary actually matters: launching an external `llama-server`, request forwarding, thinking-flag translation, streaming, error forwarding, and process/environment behavior.

So this is not just a test speedup. It removes unnecessary server lifecycle complexity while moving the regression coverage closer to the code it is supposed to protect.

No backend-selection behavior is intentionally changed.

## Scope

## Testing

_Testing details:_

```bash
cmake --build build --target test_llamacpp_system_backend
ctest --test-dir build -R '^LlamaCppSystemBackendTest$' --output-on-failure

python3 test/test_llamacpp_system_backend.py --cli-binary "$PWD/build/lemonade"

cmake --build build --target cpp-ci-tests
ctest --test-dir build -L '^cpp-ci$' --output-on-failure
```

The Python suite now uses a single lemond lifecycle for the remaining integration tests instead of restarting it for each backend-policy case.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3002,6 +3002,21 @@ if(BUILD_TESTING AND EXISTS "${_GPU_ENV_TEST_SRC}")
     add_cpp_ci_test(BackendUtilsGpuEnvTest CI ON COMMAND test_backend_utils_gpu_env)
 endif()
 
+set(_LLAMACPP_SYSTEM_BACKEND_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_llamacpp_system_backend.cpp"
+)
+if(BUILD_TESTING AND CMAKE_SYSTEM_NAME STREQUAL "Linux" AND EXISTS "${_LLAMACPP_SYSTEM_BACKEND_TEST_SRC}")
+    add_executable(test_llamacpp_system_backend
+        test/cpp/test_llamacpp_system_backend.cpp
+    )
+    target_include_directories(test_llamacpp_system_backend PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/cpp/server
+    )
+    target_link_libraries(test_llamacpp_system_backend PRIVATE lemonade-server-core)
+    add_dependencies(test_llamacpp_system_backend copy_resources)
+    add_cpp_ci_test(LlamaCppSystemBackendTest CI ON COMMAND test_llamacpp_system_backend)
+endif()
+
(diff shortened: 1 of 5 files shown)


abn (approved):
This looks great as it is and is a significant improvement to test speed and CI reliability by dropping 13 unnecessary `lemond` lifecycles while expanding C++ coverage across non-AMD CI environments. Approving!

The following are not blockers at all, just minor nits/observations that could be considered in a follow-up:

1. **Atomic write for mock control IPC (`test_llamacpp_system_backend.py`)**:
   `_write_mock_control` opens `control_path` with `"w"`, which truncates the file before writing. While tests currently run sequentially, writing to a temp file and doing an atomic `os.replace` is safer for file-based mock IPC to prevent `ReusableHTTPServer` from ever reading an empty/truncated file under concurrent operations.

2. **Unset vs. Explicit `prefer_system: false` in C++ fixture**:
   In `test/cpp/test_llamacpp_system_backend.cpp`, `build_recipes()` sets `config_json["llamacpp"]["prefer_system"]` explicitly to `true` or `false`. Adding a test case that leaves `prefer_system` omitted entirely from `config.json` would give 1:1 parity with the "unset" check from the original Python `test_005` (verifying `ConfigFile::base_defaults()` fallback).

3. **Missing PATH binary with `prefer_system=false`**:
   The `empty_bin` (missing `llama-server`) scenario is currently tested with `prefer_system=true`. Adding a test case with `prefer_system=false` when the binary is missing would mirror the original `test_001` to ensure `state == "unsupported"` is reported identically regardless of the preference setting.

Thanks for cleaning this up!

PR #3288, opened 2026-08-21
test: fix dangling CLI temp path in runtime override test

fl0rianr:
## Summary

Fixes the macOS failure in `CliRuntimeOverrideTest`.

The test passed `temp_dir.string().c_str()` as a CLI argument, leaving a dangling pointer after the temporary string was destroyed. On macOS/libc++ this could result in an empty path and eventually fail in `std::filesystem::create_directories("")`.

Keep the path string alive through CLI parsing, add assertions for the parsed cache/config directories, and use `config_dir` for persistence to match the current config layout.

No production code is changed.

## Scope

## Testing

_Testing details:_

Targeted validation:

```bash
cmake --build --preset default --target test_cli_runtime_override
ctest --test-dir build -R '^CliRuntimeOverrideTest$' --output-on-failure
```

Full C++ CI suite:

```bash
cmake --build --preset default --target cpp-ci-tests
ctest --test-dir build -L '^cpp-ci$' --output-on-failure
```

## Documentation

## Breaking Changes

* [ ] This PR introduces breaking changes.

## AI-assisted contribution

### DIFF
--- a/test/cpp/test_cli_runtime_override.cpp
+++ b/test/cpp/test_cli_runtime_override.cpp
@@ -53,9 +53,12 @@ int main() {
         ConfigFile::save(temp_dir.string(), initial_cfg);
 
         CLIParser parser;
-        const char* argv[] = {"lemond", temp_dir.string().c_str(), "--port", "8888", "--host", "127.0.0.1"};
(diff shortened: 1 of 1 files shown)


abn (approved):
LGTM 🚀

PR #3289, opened 2026-08-21
fix(server): count the iGPU firmware carveout as physical memory on Linux

ramkrishna2910:
## Summary

Split out of #3047 (ds4 backend) at reviewer request — this is a general model-size-filtering fix, not ds4-specific.

On a unified-memory APU (e.g. AMD Strix Halo), the firmware reserves a block of system DRAM as the iGPU's "VRAM" before the kernel boots, so it never appears in `MemTotal`. That DRAM is ordinary memory a model can still run out of, but reporting only `MemTotal` makes model-size filtering hide models that actually fit — a 128 GB Strix Halo with a 64 GB carveout reports ~61 GB, so anything over ~49 GB (0.8 × 61) is wrongly filtered.

This adds the detected iGPU carveout (read from the KFD topology, `/sys/class/kfd`) back into the reported physical memory, matching how Windows already reports installed capacity. Discrete GPU VRAM is separate silicon and is deliberately **not** counted.

Once this merges, #3047 will pick it up via a merge from `main` and drop these lines.

## Scope

## Testing

_Testing details:_

- Built `lemond` on Linux and confirmed `GET /api/v1/system-info` reports `Physical Memory: 126.44 GB` on a 128 GB Strix Halo (gfx1151) whose `MemTotal` is ~62 GB (64 GB carveout added back). Previously it reported ~62 GB.
- The carveout reader is hardware/`/sys`-specific (returns 0 where no KFD iGPU topology exists, e.g. CPU-only CI), so `get_physical_memory()` degrades to `MemTotal` unchanged on non-APU hosts. Happy to extract the KFD-node parsing into a pure helper with unit coverage if the reviewer would like that.

## Documentation

## Breaking Changes

## AI-assisted contribution

Implemented with Claude Code (Opus 4.8). Change was originally authored in #3047 and split out here.

### DIFF
--- a/src/cpp/include/lemon/system_info.h
+++ b/src/cpp/include/lemon/system_info.h
@@ -228,6 +228,7 @@ private:
     double get_amd_vram(const std::string& drm_render_minor);
     double get_amd_gtt(const std::string& drm_render_minor);
     bool get_amd_is_igpu(const std::string& drm_render_minor);
+    double get_igpu_carveout_gb();
 
 private:
     double parse_memory_sysfs(const std::string& drm_render_minor, const std::string& fname);
--- a/src/cpp/server/system_info.cpp
+++ b/src/cpp/server/system_info.cpp
@@ -3653,6 +3653,51 @@ double LinuxSystemInfo::get_amd_vram(const std::string& drm_render_minor) {
     return parse_memory_sysfs(drm_render_minor, "mem_info_vram_total");
 }
 
+double LinuxSystemInfo::get_igpu_carveout_gb() {
+    // An integrated GPU's "VRAM" is system DRAM the firmware reserved before the
+    // kernel booted, so it never appears in MemTotal. Discrete VRAM is separate
+    // silicon and must not be counted here.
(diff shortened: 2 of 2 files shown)


superm1 (changes requested):
I really don't expect this to be correct.  We can't address all memory with ROCm.  We target the bigger of the two pools.

ramkrishna2910:
Closing in favor of a different approach.

@superm1's review [flagged](https://github.com/lemonade-sdk/lemonade/pull/3047#discussion_r3832551973) that adding the iGPU carveout to physical memory is incorrect for ROCm: you don't get to address system RAM **and** the carveout as one pool — you get the larger of the two. Our own OOM repro confirms it (the ds4 model tops out at ~48 GiB on a box this "fix" reports as 126 GB).

The replacement filters streaming backends on the model's resident **working set** vs the device's largest memory pool (exactly the "bigger of the two pools" superm1 described), and drops the carveout change entirely. Opening that as a separate PR shortly; will link it here.

ramkrishna2910:
Replacement opened: #3328 (working-set filter, off `main`).

PR #3292, opened 2026-08-21
fix: guide external custom model deletion

falabellamichael:
## Summary

Models discovered through `extra_models_dir` are user-managed and cannot be deleted through Lemonade's managed-delete API. This PR intercepts that Delete action, avoids calling the delete endpoint, and shows a notification telling the user to remove the model directly from their external models folder.

No linked issue.

## Scope

## Testing

_Testing details:_

- `npm run typecheck`
- `npm run test:ui-regressions`
- `npm run build:nobundle`
- Focused Playwright manual UI test was not run because Playwright Chromium is not installed.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

### DIFF
--- a/src/app/src-tauri/src/commands.rs
+++ b/src/app/src-tauri/src/commands.rs
@@ -6,7 +6,6 @@ use crate::settings::{self, AppSettings};
 use crate::tray_launcher;
 use serde::Serialize;
 use serde_json::Value;
-use std::path::PathBuf;
 use tauri::{AppHandle, Emitter, Manager, WebviewWindow};
 
 // Note: `discover_server_port` deliberately does NOT spin up a second UDP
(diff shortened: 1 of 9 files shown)


fl0rianr (changes requested):
Nice UX improvement overall and thanks for tackeling this. I like intercepting extra_models_dir deletion before calling the delete API. I think the native “Open folder” part needs a little more work before merge though:

* I agree with the existing Windows review: `canonicalize()` can produce a `\\?\...` path that Explorer does not reliably handle.
* `isLoopbackServer()` is not enough to prove the desktop app and `lemond` share a filesystem. 127.0.0.1 may still be an SSH tunnel, container/WSL forwarding, etc. This conflicts with Lemonade's many-clients/one-server topology and could open an unrelated local path.
* The server accepts relative extra_models_dir values, while the new Tauri command rejects anything non-absolute.
* extra_models_dir is scanned recursively, so opening only the configured root may still leave the user far away from the actual model. The server already knows the concrete checkpoint/model location.

For what is basically “explain why this model cannot be deleted”, this also feels like a bit more native plumbing than the change really needs. I would keep the simple delete interception + notice, and either leave “Open folder” out for now, or make it a properly local-only operation based on the resolved model/checkpoint location rather than passing the server's configured root into a new generic native path opener. That would also remove most of the cross-platform/path/topology edge cases here.

falabellamichael:
@fl0rianr alright no problem, should be fixed now without the open folder icon 👌

fl0rianr (approved):
LGTM. thanks.

PR #3305, opened 2026-08-23
fix: recover repo name for extra models kept in HF-cache layout

### LINKED ISSUE #3297: Models moved from "default lemonade model directory" TO "external custom models directory" don't display correctly
### Platform

Linux/Ubuntu

### Lemonade Version

11.6

### GPU / APU Model

_No response_

### Component

Other / Not sure

### Bug Description

Issue when models moved from default lemonade model directory TO external custom models directory

As per screenshot below, model appears fine when in the default location, but once moved to custom location it loses its name

Image

### Steps to Reproduce

download model into default location
move model to custom location
refresh models page

expected - should appear showing correct name in either location

### Log Output

```shell

```

### Additional Context

_No response_

Davedave001:
## Summary

`discover_extra_models_in_directory` named an extra-directory model after its immediate containing folder. When a model is moved into `extra_models_dir` with its Hugging Face cache layout intact (`models--org--repo/snapshots/<hash>/...`), that folder is the snapshot hash, so the model showed up with an unreadable hash instead of its name once moved out of the default cache location.

This walks up from the discovered file to the `models--`/`modelscope--models--` ancestor (bounded by the configured `extra_models_dir` root) and decodes it back to `org/repo` when that layout is present, falling back to the previous folder-name behavior otherwise.

Fixes #3297

## Scope

## Testing

_Testing details:_
No C++ toolchain was available in the environment this change was authored in, so I was not able to build or run this locally — please let me know if CI surfaces anything and I'll fix it quickly. I traced the change carefully against the existing "walk up to `models--` ancestor" pattern already used elsewhere in this file (the delete-path cache lookup a few hundred lines down) to keep the decode logic consistent with it, and checked that the recovered name flows correctly through both the split-variant alias path and the non-split id path later in the function.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1171,6 +1171,37 @@ ModelInfo ModelManager::init_extra_model_info(const std::string& name) const {
     return info;
 }
 
+// A model moved (with its full hub layout) into extra_models_dir keeps its
+// "models--org--repo/snapshots/<hash>/..." shape. Walking up from a discovered
+// file to that "models--"/"modelscope--models--" ancestor recovers "org/repo"
+// so the model keeps a readable name instead of the snapshot hash that would
+// otherwise be read as the containing folder's name. Stops at `stop_at` (the
+// configured extra_models_dir root) so an unrelated ancestor name can't match.
+static std::string hf_cache_repo_name_for_path(const fs::path& path, const fs::path& stop_at) {
+    static constexpr const char kHfPrefix[] = "models--";
+    static constexpr const char kMsPrefix[] = "modelscope--models--";
+    for (fs::path cur = path; cur.has_filename() && cur != stop_at; cur = cur.parent_path()) {
+        std::string dirname = cur.filename().string();
+        std::string encoded;
+        if (dirname.rfind(kMsPrefix, 0) == 0) {
+            encoded = dirname.substr(sizeof(kMsPrefix) - 1);
+        } else if (dirname.rfind(kHfPrefix, 0) == 0) {
+            encoded = dirname.substr(sizeof(kHfPrefix) - 1);
+        } else {
+            continue;
+        }
+        // registry_repo_cache_dir_name() encodes "org/repo" as "org--repo";
+        // repo names don't contain "--", so the first occurrence is the
+        // namespace boundary.
+        size_t sep = encoded.find("--");
+        if (sep == std::string::npos || sep == 0 || sep + 2 >= encoded.size()) {
+            return encoded;
+        }
+        return encoded.substr(0, sep) + "/" + encoded.substr(sep + 2);
+    }
+    return "";
+}
+
 // Record a discovered model without ever overwriting one already found. Two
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Nice focused fix and the general approach makes sense. One issue I think we should handle before merge: a full HF cache can contain multiple `snapshots/<hash>` directories. Discovery currently scans all of them, while this change gives each snapshot the same recovered `org/repo` name, so an older snapshot may win the plain `extra.org/repo` ID and the active `refs/main` snapshot gets collision-qualified instead.

Could we detect the actual cache layout and prefer only the active `refs/main` snapshot, with fallback to normal extra-directory discovery if no valid ref exists? A regression test with two snapshots would be valuable here. It would also avoid accidental `models--*` folder matches and cover this new naming behavior explicitly.

PCAssistSoftware (changes requested):
That seems to work for me, models names in external storage now showing correctly

Related suggestion: on the Files tab could we show file location of a model? as then it would be easy at a glance to see on the model card which model is located in your custom external storage location?

Image

PCAssistSoftware (commented):
just spotted another issue - after models have been moved they then lose any settings you had configured for them e.g. temp, top-k etc

think this is because in recipes file before moving it appears as:

  "user.Llama-3.2-3B-Instruct-GGUF-Q4_K_M": {
    "llamacpp_args": "--temp 0.6 --top-p 0.9"
  },
  
and after it appears as 

"extra.unsloth/Llama-3.2-3B-Instruct-GGUF": {
    "llamacpp_args": "--temp 0.6"
  },
  
  so not sure what can really be done about this

PCAssistSoftware (commented):
One other comment on this:

I like how for my custom model directory it now shows name as

unsloth/Llama-3.2-3B-Instruct-GGUF

rather than

Llama-3.2-3B-Instruct-GGUF

This make it easier to see different variants at a glance

Can we replicate this functionality for models stored in default location as well, so if I have say Qwen 3.8 from two different providers I can clearly see them e.g.

instead of seeing:

Qwen3.8-27B-GGUF
Qwen3.8-27B-GGUF-Q4_K_M

I see

unsloth/Qwen3.8-27B-GGUF
bartowski/Qwen3.8-27B-GGUF-Q4_K_M

jeremyfowers (changes requested):
No C++ PRs into GUI3_merging, please.

Kindly retarget the PR to main and send me a ping on discord when its ready for another look.

falabellamichael:
i will relay this to dave!

Davedave001:
Retargeting onto main per @jeremyfowers — opened #3361 with the same two commits plus the refs/main fix @fl0rianr flagged and regression tests. Closing this one in its favor.

Davedave001:
I retargeted onto main per your request with a new PR: https://github.com/lemonade-sdk/lemonade/pull/3361 and also fixed the multi-snapshot collision fl0rianr flagged, with regression tests. Thanks

PR #3306, opened 2026-08-23
fix: download TheRock ROCm backend for all detected GPU architectures

### LINKED ISSUE #2371: ROCm backend downloaded for first GPU architecture detected only, gfx1036 will break ROCm backends.
### Platform

Linux/Arch

### Lemonade Version

617d85ec

### GPU / APU Model

RX7900XTX

### Component

Desktop App / UI

### Bug Description

ROCm backends are downloaded for only the first GPU architecture found. If you are running an AMD iGPU, it will be for that not any dGPU's because SystemInfo::get_rocm_arch() deliberately checks iGPU first, then dGPUs and returns the first match.

### Steps to Reproduce

1. Download ROCm backend while running both an AMD iGPU and dGPU.
2. Attempt to run a model using ROCm backend.
3. Receive error because it only downloaded TheRock for gfx1036.

ROCm backend download either downloads requirments to run on all detected architectures, or provides the user with a choice of which ones to download.

### Log Output

```shell

```

### Additional Context

The other PR's worked well, it does let you download the ROCm backends again.

bong-water-water-bong:
Split from the closed stacked PR #2452 (first of four clean PRs).

Closes #2371 — multi-GPU systems only got ROCm for the first GPU arch.

Previously `get_rocm_arch()` returned only the first AMD GPU arch, so on systems with an iGPU and a dGPU of different architectures TheRock was only installed for one of them.

- `get_rocm_arches()` is now the single entry point, returning ALL detected AMD GPU ROCm architectures (deduplicated, discrete-first so `front()` preserves single-arch semantics)
- `install_therock_if_needed()` installs the ROCm runtime for every detected arch
- Per-thread arch override feeds `get_rocm_arches()` for asset-URL resolution without a GPU present
- Pure `collect_rocm_arches(json)` helper + unit coverage for ordering, dedup, skip and override behavior

### DIFF
--- a/src/cpp/include/lemon/system_info.h
+++ b/src/cpp/include/lemon/system_info.h
@@ -120,9 +120,9 @@ public:
     // over an integrated one on a hybrid host (e.g. Strix Halo APU + MI300X dGPU).
     static std::string select_rocm_arch(const json& amd_gpu_devices);
 
-    // Map an "amd_gpu" device array to ALL ROCm architectures, deduplicated and
-    // ordered discrete-first then integrated, so front() is the single arch a
-    // hybrid host should use (matches select_rocm_arch()). Pure — no probing.
+    // Pure version of get_rocm_arches() over a caller-supplied "amd_gpu" device
+    // array — same ordering/dedup contract, no probing. Throws
+    // std::invalid_argument on non-array input (caller bug).
     static std::vector<std::string> collect_rocm_arches(const json& amd_gpu_devices);
 
     // Collapse a concrete ROCm ISA (e.g. gfx1201) to the family target name the
--- a/src/cpp/include/lemon/system_info.h
+++ b/src/cpp/include/lemon/system_info.h
@@ -108,19 +108,22 @@ public:
     static std::vector<RecipeStatus> get_all_recipe_statuses();
 
     // Device support detection
-    // Return the first (primary) ROCm architecture — typically the iGPU.
-    // Used for rocm_channel selection and single-arch contexts.
-    static std::string get_rocm_arch();
-
-    // Return ALL detected AMD GPU ROCm architectures, ordered iGPU first
-    // then dGPUs. Used by backend download paths to install ROCm binaries
-    // for every GPU on the system, not just the first one detected.
+    // Return ALL detected AMD GPU ROCm architectures (deduplicated). The list
+    // is ordered discrete-first, then integrated, so front() is the arch a
+    // single-arch consumer should use — matching select_rocm_arch()'s
+    // preference on a hybrid host. Used by backend download paths to install
+    // ROCm binaries for every GPU on the system, not just the first one.
     static std::vector<std::string> get_rocm_arches();
     static std::string get_cuda_arch();
 
     // Picks the ROCm compute target from an "amd_gpu" device array: a discrete GPU wins
     // over an integrated one on a hybrid host (e.g. Strix Halo APU + MI300X dGPU).
     static std::string select_rocm_arch(const json& amd_gpu_devices);
+
+    // Map an "amd_gpu" device array to ALL ROCm architectures, deduplicated and
+    // ordered discrete-first then integrated, so front() is the single arch a
+    // hybrid host should use (matches select_rocm_arch()). Pure — no probing.
+    static std::vector<std::string> collect_rocm_arches(const json& amd_gpu_devices);
 
     // Collapse a concrete ROCm ISA (e.g. gfx1201) to the family target name the
     // GitHub release repos publish their assets under (e.g. gfx120X), per the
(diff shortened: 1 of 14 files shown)


src/cpp/include/lemon/system_info.h
```diff
@@ -108,7 +108,14 @@ class SystemInfo {
     static std::vector<RecipeStatus> get_all_recipe_statuses();
 
     // Device support detection
+    // Return the first (primary) ROCm architecture — typically the iGPU.
+    // Used for rocm_channel selection and single-arch contexts.
     static std::string get_rocm_arch();
```
superm1: I have to question the purpose of `get_rocm_arch` and `get_rocm_arches`.  I think it would be better to just always return a list (IE `get_rocm_arches`) and if you need the "first" you just access the first index.

src/cpp/server/system_info.cpp
```diff
@@ -2171,6 +2171,39 @@ std::string SystemInfo::get_rocm_arch() {
...
+            for (const auto& gpu : devices["amd_gpu"]) {
+                if (gpu.value("available", false)) {
+                    std::string name = gpu.value("name", "");
+                    if (!name.empty()) {
+                        std::string arch = identify_rocm_arch_from_name(name);
+                        if (!arch.empty()) {
+                            // Avoid duplicates (e.g., identical iGPU/dGPU arch strings)
+                            if (std::find(arches.begin(), arches.end(), arch) == arches.end()) {
+                                arches.push_back(arch);
+                            }
```
superm1: To avoid so much deep indentation could you instead skip loop iterations with `continue;` on the negative of what you do now?

src/cpp/include/lemon/system_info.h
bong-water-water-bong: Done — `get_rocm_arch()` is removed entirely and `get_rocm_arches()` is now the single entry point; single-arch consumers take `front()`.

One subtlety I kept in mind: on a hybrid host (e.g. Strix Halo iGPU + MI300X dGPU with different archs) the old `get_rocm_arch()` deliberately preferred the *discrete* GPU (via `select_rocm_arch()`). So the new list is ordered **discrete-first, then integrated** — `front()` therefore preserves the exact single-arch semantics the removed API had for all backend call sites (rocm_channel, TheRock lib pathing, launch policy), while the full list still feeds the multi-arch install path.

The per-thread override (`set_rocm_arch_override()`) now feeds `get_rocm_arches()` (returns just that arch), so the asset-URL resolution endpoint keeps working without a GPU present.

I also extracted the device-array mapping into a pure `collect_rocm_arches(json)` and added unit coverage in `test_vllm_cdna_gating.cpp`: hybrid returns both arches discrete-first, identical iGPU/dGPU arch deduped, unavailable/unknown devices skipped, override short-circuit.

src/cpp/server/system_info.cpp
bong-water-water-bong: Done — the detection loop now skips early with `continue` on the negatives (unavailable GPU, empty arch), which flattened the nesting to one level of conditionals. Combined with the discrete/integrated bucket split the function is a straightforward single pass now.

src/cpp/server/system_info.cpp
```diff
@@ -2148,27 +2150,65 @@ std::string SystemInfo::select_rocm_arch(const json& amd_gpu_devices) {
...
 
-std::string SystemInfo::get_rocm_arch() {
+std::vector<std::string> SystemInfo::collect_rocm_arches(const json& amd_gpu_devices) {
+    // Pure mapping from the "amd_gpu" device array to ALL supported ROCm
+    // architectures, deduplicated and ordered discrete-first then integrated,
+    // so front() matches what select_rocm_arch() picks on a hybrid host.
+    std::vector<std::string> discrete;
+    std::vector<std::string> integrated;
+    if (!amd_gpu_devices.is_array()) {
+        return discrete;
```
superm1: this seems like a caller mistake.  Shouldn't it throw an error?   Why would you call collect_rocm_arches with an empty array?

src/cpp/server/system_info.cpp
```diff
@@ -2148,27 +2150,65 @@ std::string SystemInfo::select_rocm_arch(const json& amd_gpu_devices) {
...
+        if (arch.empty()) {
+            continue;
+        }
+        std::vector<std::string>& bucket =
+            gpu.value("integrated", false) ? integrated : discrete;
+        if (std::find(bucket.begin(), bucket.end(), arch) == bucket.end()) {
+            bucket.push_back(arch);
+        }
+    }
+    // Dedup across buckets (e.g. identical iGPU/dGPU arch strings), keeping
```
superm1: why would you ever have more than one iGPU?

superm1:
Directionally it looks right.  You can see CI failed because of your excessively wordy comments.

Please install `pre-commit` on your side to help provide better quality PRs in the future.

src/cpp/server/system_info.cpp
bong-water-water-bong: Agreed — silent empty-return was the wrong shape for a caller bug. `collect_rocm_arches()` now throws `std::invalid_argument` on non-array input (commit 33dd8aa). The outer `get_rocm_arches()` already guards the `devices`/`amd_gpu` shape and absorbs any detection failure, so the throw only fires on direct misuse of the pure helper — exactly the case you flagged. Test updated to assert the throw.

src/cpp/server/system_info.cpp
bong-water-water-bong: The dedup is not about multiple iGPUs — a system has at most one. It is about the same arch string appearing in both buckets: an iGPU and a discrete GPU can report the same family, e.g. a Strix Halo iGPU (gfx1151) next to an RDNA4 dGPU that is also gfx1151. The per-bucket dedup handles repeats within a bucket; the cross-bucket merge keeps the final list unique while preserving the discrete-first order (covered by a test: identical iGPU/dGPU arch -> one entry).

bong-water-water-bong:
Thanks — pre-commit feedback addressed in 33dd8aa. The comment-slop hook now passes locally (the "explained at 3 sites" issue is collapsed to one canonical explanation on `get_rocm_arches()` with cross-references from `collect_rocm_arches()`), and the comment density note is gone. Re: installing pre-commit — agreed, will wire it up for future PRs.

bong-water-water-bong:

PR #3307, opened 2026-08-23
fix: model-level resume for interrupted HF downloads

### LINKED ISSUE #1546: Lemonade should be more resilient downloading models
### Platform

Linux/Ubuntu

### Lemonade Version

10.0.1

### GPU / APU Model

strix halo

### Component

Desktop App / UI

### Bug Description

1) in the event of a network interruption (frequently happening here) occurring during a "lemonade(server) pull"  for some model, eventual  subsequent retries to download the same not yet fully retrieved model, lemonade should not erase the already partially retrived model but it should instead just continue to download it from where it was left, appending to the remaining part of the same model

2) when a network interruption as above described occurs leaving a partially downloaded model in a uncompleted state, lemonade should not show the model as if it was fully retrieved and ready to be run/served 

3) should the user deliberately interrupt some model downloading process using ctrl-c, model file and other files relics are left dangling into the relative directory such that when invoking lemonade (even for just printing the list of available/downloaded models names), it crashes in that precise listing position requiring  the user to remove the uncompleted  model directory in order to be able to run lemonade stuff again

### Steps to Reproduce

1. lemonade-server serve
2. lemonade(-server) pull <model> *
*. btw, why NPU models listed/pulled/run/served from lemonade(-server) and flm aren't named exactly the same?
3. [optional] disconnect from the net or interrupt w/ ctrl-c
 
Lemonade should check if a model has been really fully retrieved (maybe checksumming it or by creating a lock file at pull start and removing it after downloading 100%); in case it was not completed it should ask the user to continue retriving it, to clear the model's directory or to skip. In any case and especially in this last case, lemonade should also skip a model directory if it contains partially retrieved files affecting lemonade itself ability to run

### Log Output

```shell

```

### Additional Context

_No response_

bong-water-water-bong:
Split from the closed stacked PR #2452 (second of four clean PRs).

Fixes #1546.

Before fetching the HF API and rebuilding the file list, check for an existing `.download_manifest.json` with incomplete files. If found, resume downloading from the partial state instead of starting over. Avoids re-downloading already-completed files after a network interruption or Ctrl-C during model pull.

(Per-file retry/resume already exists on main; this adds the model-level manifest fast-path on top.)

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -4834,6 +4834,41 @@ void ModelManager::download_from_registry(const ModelInfo& info,
     repositories.emplace(main_repo_id, registry.fetch_repository(main_repo_id));
 
     std::map<std::string, std::vector<std::string>> files_to_download;
+
+    // Resume fast-path: if a download manifest exists, resume downloading
+    // incomplete files instead of re-fetching the HF API and rebuilding the
+    // file list. This handles the common case of a network interruption.
+    {
+        fs::path model_cache_path = path_from_utf8(get_hf_cache_dir()) / repo_id_to_cache_dir_name(main_repo_id);
+        fs::path manifest_path = model_cache_path / "snapshots" / ".download_manifest.json";
+        if (!safe_exists(manifest_path)) {
+            manifest_path = model_cache_path / ".download_manifest.json";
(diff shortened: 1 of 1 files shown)


bong-water-water-bong:

PR #3308, opened 2026-08-23
fix: ensure content+role fields always present in streaming deltas for OpenAI compat

### LINKED ISSUE #1370: feat: OpenCode compatibility — streaming reasoning_content breaks @ai-sdk/openai-
## Summary

When using Lemonade as a backend for OpenCode via the `@ai-sdk/openai-compatible` adapter, all requests fail with **Connection reset by server**. The root cause is that Lemonade's streaming responses use a non-standard `reasoning_content` field in delta chunks instead of the standard OpenAI `content` field, which the adapter does not recognise and drops the connection.

## Environment

- Lemonade version: 10.0.0
- Model: Qwen3-8B-Hybrid (ryzenai-llm backend)
- Hardware: AMD RyzenAI NPU
- Client: OpenCode with `@ai-sdk/openai-compatible`

## What happens

Every streaming chunk sent by Lemonade uses `reasoning_content` in the delta:

```
{"choices":[{"delta":{"reasoning_content":"..."}, "finish_reason":null}]}
```

The OpenAI spec expects:

```
{"choices":[{"delta":{"content":"..."}, "finish_reason":null}]}
```

Because `@ai-sdk/openai-compatible` never sees a `content` field, it resets the connection immediately.

Passing `"thinking": false` in the request body does not suppress the behaviour — Lemonade ignores this flag.

## Expected behaviour

One or more of the following:
1. Lemonade respects `"thinking": false` and streams standard `content` deltas only
2. Lemonade streams `content` alongside `reasoning_content` so OpenAI-compatible clients do not break
3. Official docs or config option for connecting OpenCode to Lemonade

## Workaround

Using `lemonade-server launch claude` works via the Anthropic compatibility endpoint. However this requires Claude Code specifically — OpenCode users on RyzenAI hardware have no working path currently.

## Steps to reproduce

1. Start Lemonade with any ryzenai-llm model
2. Configure OpenCode with `baseURL: http://127.0.0.1:8000/v1` using `@ai-sdk/openai-compatible`
3. Send any message — observe "Connection reset by server"
4. Run: `curl http://127.0.0.1:8000/v1/chat/completions -d '{"model":"...","messages":[...],"stream":true}'`
5. Observe all chunks use `reasoning_content` instead of `content`

bong-water-water-bong:
Split from the closed stacked PR #2452 (third of four clean PRs).

Closes #1370 — OpenCode / @ai-sdk/openai-compatible streaming crash when reasoning models emit `reasoning_content` without `content`.

**Streaming proxy** (`streaming_proxy.cpp`): intercepts each SSE `data: {...}` line in `forward_sse_stream()`; injects `content: ""` when `reasoning_content` is present without content, and `role: "assistant"` when null/missing on assistant deltas. Only applies to `chat.completion.chunk` objects — non-chat SSE passthrough.

**Non-streaming REST** (`server.cpp`): same content injection on the response message.

Notes: the `thinking` → `enable_thinking` normalization half of the original fix already exists on main via `normalize_thinking_controls`, so it is not duplicated here.

Unit tests: role normalization, reasoning content normalization, CRLF, multi-choice, multi-line streams (11 cases).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3118,6 +3118,14 @@ if(BUILD_TESTING AND EXISTS "${_STREAMING_PROXY_CANCEL_TEST_SRC}")
     add_cpp_ci_test(StreamingProxyCancelTest CI OFF COMMAND test_streaming_proxy_cancel)
 endif()
 
+set(_STREAMING_PROXY_REASONING_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_streaming_proxy_reasoning_content.cpp")
+if(BUILD_TESTING AND EXISTS "${_STREAMING_PROXY_REASONING_TEST_SRC}")
+    add_executable(test_streaming_proxy_reasoning_content
+        test/cpp/test_streaming_proxy_reasoning_content.cpp)
+    target_link_libraries(test_streaming_proxy_reasoning_content PRIVATE lemonade-server-core)
+    add_cpp_ci_test(StreamingProxyReasoningContentTest CI ON COMMAND test_streaming_proxy_reasoning_content)
+endif()
+
 set(_STREAMING_HEARTBEAT_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_streaming_heartbeat.cpp")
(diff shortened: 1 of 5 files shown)


bong-water-water-bong:

PR #3309, opened 2026-08-23
fix: pre-load OOM memory warning + hard block at 2x headroom

bong-water-water-bong:
Split from the closed stacked PR #2452 (fourth of four clean PRs).

Addresses #1804 — large model loads that can't fit now fail fast with a clear error instead of hanging or being OOM-killed.

**Pre-load check** (`router.cpp`): compares `model_info.size` against `get_available_memory_gb()` for the target device (after eviction, so freed VRAM/RAM is visible).

- Model fits with < 2x headroom: warn-only (GGUF file size ≠ load-time memory; auto-tune ctx_size may still fit it)
- Model exceeds 2x headroom: reject the load with a clear `runtime_error` instead of letting the OS OOM killer kill the whole process


bong-water-water-bong:

PR #3314, opened 2026-08-24
Conversation momentum implementation

### LINKED ISSUE #2956: [Router] Conversation momentum: stateless hysteresis to prevent per-turn route flapping
## Problem

The routing engine evaluates every request from scratch, and the routing input is the **last user message only**: `build_route_context` walks `messages` backward and takes the last `role == "user"` entry (`src/cpp/server/routing_classifier_services.cpp:383-390`), and `ctx.params.chars` is the UTF-8 byte count of just that turn (`routing_classifier_services.cpp:451`, documented in `src/cpp/include/lemon/routing_policy.h:57-64`). There is no session state, decision cache, or hysteresis anywhere in the routing path.

Consequence: in a real conversation the turn shape oscillates, so the route flaps. A policy like

```json
{"id": "big-model", "match": {"min_chars": 2000}, "route_to": "user.Qwen3-32B-GGUF"}
```

escalates a long coding question to the strong model, then the follow-up "yes, do it" (11 bytes) falls through to the small default — mid-task. Every flap pays a switch cost (model load/eviction, KV re-prefill, provider cache miss), and switching itself measurably degrades multi-turn outcomes: arXiv 2603.03111 measures −8 to +13 pp outcome drift when the model changes mid-conversation, attributing it to the new model conditioning on a prefix authored by a different model. vLLM Semantic Router hit this exact failure in production (their issue #1439: a coding session routes to a large model, then "looks good, commit it" alone gets classified and lands on a 0.5B model) and shipped stateless "Conversational Routing Momentum" as the fix (their issue #1458, released in their v0.3).

## Proposed approach

Add an opt-in `routing.momentum` block that replaces the raw last-turn length with a **momentum-filtered effective length** wherever `min_chars` / `max_chars` conditions evaluate. The filter is recomputed per request from the request's own message history — zero server-side session state, fully deterministic, so it stays conformance-corpus testable.

Signal and filter (the concrete design):

1. Extract the UTF-8 byte lengths `c_1 .. c_n` of every **user** turn's text content, in order, using the same `collect_text_from_content` extraction the routing input already uses (`routing_classifier_services.cpp:148`). User turns only, matching the existing principle that earlier assistant/system turns can't skew routing (`routing_classifier_services.cpp:418-419`).
2. Asymmetric exponential filter over that series (IEEE double): `m_1 = c_1`; for `i > 1`, `m_i = m_prev + attack * (c_i - m_prev)` when rising (`c_i >= m_prev`), `m_i = m_prev + release * (c_i - m_prev)` when falling.
3. When `momentum.enabled` is true, `CharsCondition` (`src/cpp/server/routing_policy.cpp:876-894`) compares its threshold against `m_n` instead of `params.chars`. All other conditions are untouched.

With `attack = 1.0` (default) a suddenly long/complex turn escalates immediately; with `release = 0.3` (default) a short follow-up only decays the effective length by 30% per turn, so the conversation stays on the escalated route for several turns before de-escalating. `attack = release = 1.0` degenerates exactly to today's behavior.

Schema (additive within major 1, per the compatibility rules in `src/cpp/resources/schemas/README.md` — a compatible addition stays within major 1; the frozen-semantics table entry "`min_chars`/`max_chars` = input length in UTF-8 bytes" is not redefined because the block is opt-in and absent-block behavior is bit-identical):

```json
"routing": {
  "momentum": {"enabled": true, "attack": 1.0, "release": 0.3},
  ...
}
```

Validation: `attack`/`release` are numbers in (0, 1]; reject anything else at parse time with the usual clear 400. vLLM Semantic Router's additional `threshold` / `switch_margin` knobs are deliberately **not** adopted: in Lemonade's rule engine the thresholds already live in each rule's `min_chars`/`max_chars`, so a second global threshold layer would duplicate them. A classifier-scored momentum variant (filtering classifier scores rather than byte lengths) is future work.

Trace: when momentum is active, char-condition trace entries keep their condition names (`min_chars`/`max_chars`) and additionally record the effective length in the entry's `rationale` (e.g. `"effective_chars=1490.5 raw_chars=11"`), so `route_trace: true` explains why a short turn still matched the long-input rule.

## Implementation sketch

- `src/cpp/include/lemon/routing_policy.h` — add `struct Momentum {bool enabled; double attack; double release;}` to `RoutePolicy`; add `double effective_chars` alongside `chars` in `RouteContext::Params` (`routing_policy.h:60-65`).
- `src/cpp/server/routing_classifier_services.cpp` — in `build_route_context` (`:361-463`), collect the per-user-turn byte lengths during the existing message walk (chat `messages`; for Responses `input` arrays use role-tagged user items; legacy `prompt` and bare-string `input` are single-turn so `m_1 = c_1`). Store the series (or compute the filter inline once the policy's coefficients are known — since `build_route_context` is policy-agnostic today, prefer storing the length series in `RouteContext` and folding it in the engine).
- `src/cpp/server/routing_policy.cpp` — compute `effective_chars` once per request in `RoutingPolicyEngine::route` before rule evaluation; switch `CharsCondition::evaluate` (`:883-889`) to read the effective value when the policy has momentum enabled; extend the trace rationale.
- `src/cpp/server/routing_policy_parser.cpp` — accept `momentum` in `routing_block_keys()` (`:527-531`); validate the block (types, (0,1] bounds).
- `src/cpp/resources/schemas/` — additive keys in the route-policy schema for major 1 (unreleased-minor rules per the README); regenerate whatever the schema CI checks.
- Docs: `docs/dev/router-policy.md` section with the escalate-then-follow-up worked example.

## Acceptance criteria

- Unit tests (`test/cpp/`): filter arithmetic (rise/fall/degenerate coefficients, single-turn, non-chat forms), parser validation rejects out-of-range coefficients, absent block leaves `params.chars` behavior untouched.
- Conformance corpus: new group `test/conformance/routing/1/l1_momentum/` with golden cases locking: multi-turn decay keeps the escalated route for N follow-up turns; `attack=1.0` immediate escalation; `release` decay eventually de-escalates; absent block ≡ today. Choose byte counts and coefficients with exact double representations (e.g. 0.5, 0.25) so goldens are stable.
- `test/server_router.py` e2e: policy with momentum enabled; long first turn escalates; an 11-byte follow-up with full history still routes to the escalated candidate; the same follow-up without momentum falls through to default. Assert the trace rationale carries `effective_chars`.
- `route_trace` output remains schema-valid against `decision.schema.json`.

## Out of scope

- Any server-side session store, client headers, or decision caching (stateless by design; a session-affinity layer is a separate concern).
- Classifier-scored momentum, `threshold`/`switch_margin` knobs, and token-based lengths (`min_tokens`/`max_tokens` stay deferred per `routing_policy.h:57-59`).
- Applying momentum to `keywords`/`regex`/`metadata`/classifier conditions.

## References

- vLLM Semantic Router: issue #1439 (multi-turn conversations bounce between models without session affinity), issue #1458 (Conversational Routing Momentum — stateless asymmetric low-pass filter, shipped v0.3 as `momentum: {enabled, attack, release, threshold}`).
- arXiv 2603.03111 — "Evaluating Performance Drift from Model Switching in Multi-Turn LLM Systems": −8 to +13 pp strict-success drift from mid-conversation switches.
- Related issues: #2389 (tracker), and the companion whole-conversation length condition issue (total_chars) — momentum and total_chars compose but are independent.

Size: M

QuMuzafferEge:
**Summary**

Adds an opt-in `routing.momentum` policy block to the router engine to fix route flapping in multi-turn conversations. Today `min_chars`/`max_chars` only ever measure the latest user turn, so a long coding question correctly escalates to a strong model via `min_chars`, but the very next short follow-up ("yes, do it") falls straight back to `default_model` mid-task — paying a real switch cost (model load/eviction, KV re-prefill, provider cache miss) every time.

`routing.momentum` replaces the raw last-turn byte length with a momentum-filtered "effective length," recomputed fresh from each request's own message history (no server-side session state — the conversation is already resent every turn, same as any OpenAI-style chat client):

json
"routing": {
  "momentum": { "enabled": true, "attack": 1.0, "release": 0.3 }
}

An asymmetric EMA over each user turn's UTF-8 byte length: attack (default 1.0) weights a rising turn so a sudden long/complex turn escalates immediately, release (default 0.3) weights a falling turn so a short follow-up only decays the effective length by 30% per turn — keeping the conversation on the escalated route for several turns before de-escalating. attack = release = 1.0 degenerates exactly to today's behavior. Applies only to min_chars/max_chars; absent or enabled: false is behavior-identical to a policy without the block.

Fixes [#2956](https://github.com/lemonade-sdk/lemonade/issues/2956)

**Scope**
This PR addresses one clear issue or change.
I reviewed the full diff myself before submitting.
I removed unrelated local changes.

**Testing**
Code builds without errors locally.
I tested this change locally.
I described the testing performed below.

Testing details:

Built and tested on Linux (Ubuntu 24.04 container, matching the cpp-unit-tests CI job's recipe: cmake --preset default → cmake --build --preset default --target cpp-ci-tests → ctest -L "^cpp-ci$").

50/52 cpp-ci tests passed, including every test this change touches:
RoutingPolicyMomentumTest (new) — filter arithmetic (rise/fall/tie/degenerate coefficients, IEEE-exactness), CharsCondition/EvalContext::effective_chars wiring, engine-level fold, trace rationale format.
RoutingPolicyParserTest (extended) — momentum block validation (required enabled, (0,1] coefficient bounds, unknown-key rejection) and schema/parser key parity.
RoutingClassifierServicesTest (extended) — build_route_context collecting user_turn_chars across chat messages, legacy prompt, and Responses input forms.
RoutingConformanceCorpusTest — two new golden groups (l1_momentum_effective_chars, l1_momentum_absent_is_noop) plus every pre-existing group re-validated byte-for-byte unchanged, confirming no regression when the block is absent.
2 pre-existing failures (TelemetryHelpersTest, CliRuntimeOverrideTest) are unrelated to this change — neither file is touched by this diff, and both look like environment artifacts of a from-scratch container vs. the pinned CI image (locale/encoding and a streaming-timing sensitivity respectively).
python test/test_schema_lock.py and python -m unittest test.test_routing_fixtures both pass (schema-lock hash refreshed for the additive routing.momentum schema keys and the decision.schema.json rationale-field description update).

**Documentation**
Documentation is affected and has been updated.

Added a "Momentum (routing.momentum)" section to docs/dev/router-policy.md (field table, formula, worked example, trace format, explicit non-goals), footnoted the min_chars/max_chars schema descriptions and the frozen-semantics table in src/cpp/resources/schemas/README.md to reference the new opt-in mechanism without redefining the existing frozen unit/comparator.

Breaking Changes
This PR does not introduce breaking changes.

Fully additive and opt-in: new RoutePolicy/RouteContext/EvalContext fields (C++-internal), one new optional routing.momentum schema property, and use of the already-optional trace_entry.rationale field. No existing field is redefined, and every code path is gated behind momentum.enabled (default false), so any policy without the block is behavior-identical to before this change.


jeremyfowers:
Please give this PR a descriptive title before merging.

PR #3315, opened 2026-08-24
Handle llama.cpp new release scheme

bitgamma:
## Summary

Fixes two issues:

1. if backend update fails but a previous version is installed, this doesn't block the model from launching anymore. A warnign is logged and the prev version is used. This is not strictly related to llamacpp but is a nice fallback.
2. resolve latest release on GitHub including pre-releases. Of course if the version is resolved for the brief period when the vSEMVER is the latest release this will fail (it doesn't have assets), but point 1. saves the day until the next bBUILDNUM release drops.

This could be probably be extended, but then we'd need llamacpp-specific rules. I think since the current behavior completely breaks `llamacpp.vulkan_bin=latest` it is a good idea to merge this and then eventually refine the approach later.

## Scope

## Testing

Use llamacpp and see that new released are picked up correctly.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/cpp/server/backend_manager.cpp
+++ b/src/cpp/server/backend_manager.cpp
@@ -736,10 +736,29 @@ void BackendManager::install_backend(const std::string& recipe, const std::strin
     const bool backend_install_dir_existed_before = fs::exists(backend_install_dir);
 
     bool completion_reported = false;
+    bool backend_committed = false;
+
+    auto handle_install_failure = [&](const std::string& message) -> bool {
+        if (backend_committed || !has_existing_backend || force) {
+            if (!backend_install_dir_existed_before) {
+                std::error_code cleanup_ec;
+                fs::remove_all(backend_install_dir, cleanup_ec);
+            }
+            return false;
+        }
+        LOG(WARNING, "BackendManager")
+            << "Backend update for " << recipe << ":" << resolved_backend
+            << " failed (" << message << "); falling back to installed backend";
+        report_backend_ready(recipe, resolved_backend, progress_cb);
+        return true;
+    };
 
     try {
         backends::BackendUtils::install_from_github(
             *spec, params.version, params.repo, params.filename, resolved_backend, backend_progress_cb);
+        // install_from_github only returns after atomically swapping the new
+        // backend into place; from here on the working install is the new one.
+        backend_committed = true;
 
         const int logical_total_files = backend_total_files + static_cast<int>(runtime_steps.size());
         for (size_t i = 0; i < runtime_steps.size(); ++i) {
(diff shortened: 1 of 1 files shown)


fl0rianr (commented):
Thanks so much for tackeling this! I like it, especially keeping an already working backend usable when an update cannot be fetched, but I think we should tighten the design a bit before merging.

There seem to be two issues:

1. `/releases?per_page=1` changes the meaning of `latest` globally to include prereleases. For the new llama.cpp scheme this can actually select the asset-less `vSEMVER` release, while `/releases/latest` skips prereleases and keeps pointing at the binary-bearing `bBUILDNUM`. If llama.cpp eventually needs special handling, I think it should be backend/repo-specific or select the latest *usable* release rather than changing the generic resolver.

2. The new catch around `install_backend()` is too broad. It also catches failures from the follow-up runtime steps. For example, `llamacpp:rocm-stable` can successfully swap in the new backend, then fail installing/repairing TheRock, and we would still report the backend as ready because an old backend existed before the update. At that point the old backend may already be gone.

I also think there is useful synergy with #2279 here. That already established the policy “if resolving latest fails, reuse the installed version”, and the atomic staging code already preserves the old backend when the actual backend download/install fails. Rather than adding another broad fallback layer, could we keep those responsibilities separate and only fall back where we know the existing installation is still intact?

Do you think we should add regression coverage for failed backend update vs failed required-runtime install?

bitgamma:
@fl0rianr 

1. I know, but llamacpp is 90% of what is being used, and aligning behavior for all backends is good imho. Having different behavior for each repo is not great. Defining usable release is more complex because we should know the name of artifacts to look for already at the resolver.  BTW the latest release for llamacpp is always the asset-less semver, never the build, so current behavior is ALWAYS broken
2. Let's not have perfect be the enemy of good. The case you specified is an edge case, while what we address here is currently the most common failure scenario. After all both the "checked" and "unchecked" path lead to failure to launch, just with a different message.

I agree both points can be addressed better, but the issue is urgent and we release  in two days. I'd prefer to get a low-risk/low-surface fix in and then have all the time to follow-up with a "definitive" solution to the issue without having to rush it.

fl0rianr:
Fair point on keeping this small before release. I'm fine with a pragmatic hotfix now and a cleaner follow-up.

I think the important part is to preserve a simple rule:

* offline → don't update
* lookup/update failed before changing the working install → keep the known-good backend
* new backend was already committed and runtime/setup then fails → fail, don't report the backend as ready

That is also why I'd still slightly narrow this PR. The broad catch crosses that last boundary, and changing the generic `latest` resolver affects every backend.

For llama.cpp, if we need it urgent, I'd keep the temporary workaround local to that backend for now. After the release we can properly follow its new semver → nightly build mapping and fold that into the existing offline/fallback behavior from #2279 and #2612.

bitgamma:
@fl0rianr. I agree about the installation check thing, but I really cannot agree about the `latest` behavior. People selecting `latest` really want the latest version, not the latest-stable. The concept of latest-stable has been introduced now, indirectly, by llama.cpp, so this PR actually preserves the old behavior conceptually speaking. 

Also as soon as a new model drops and llamacpp merges the related PR we'll get hundreds of "when will model X be supported in Lemonade?" if we redefine "latest" to "latest-stable". So I think long-term we should still include pre-releases in `latest` and eventually a) introduce `latest-stable` that does not and b) define the concept of "usable release" to filter all releases that do not contain useful binaries.

fl0rianr:
@bitgamma I agree. Sorry for mixing GitHub's releases/latest semantics with what Lemonade's latest should mean. For us, latest should indeed mean the newest usable release, including pre-releases. latest-stable can be a separate concept later, together with proper usable-release filtering.

I'm dropping that concern.

My remaining concern is really just the install/fallback boundary. I'd like us to preserve one simple invariant: if we say we're falling back, the previous known-good installation must still actually be intact.

As pointed out already above.

If we can narrow the catch around that boundary, I'm good with this as the release fix.

bitgamma:
@fl0rianr the install/fallback boundary is now respected. The fallback path is taken only if the error happens before the new version is commited.

fl0rianr (approved):
LGTM now. Thanks

Issue #3274, opened 2026-08-20
GUI3: discover Lemonade capabilities through MCP tools/list


fl0rianr:
## Goal

Use Lemonade's MCP server as the source of truth for MCP tool contracts consumed by GUI3 when those capabilities are already exposed through `/mcp`.

Related to #3270 and #3271.

## Motivation

GUI3 currently contains custom Lemonade tool definitions and client-side execution code.

At the same time, `lemond` exposes its own MCP tool catalogue through `/mcp` and `tools/list`.

As Lemonade capabilities become available through MCP, GUI3 should not maintain a second copy of the same MCP tool names, descriptions, and input schemas.

## Scope

For Lemonade capabilities that are exposed through `/mcp`:

- Discover tools with `tools/list`
- Use the returned tool name
- Use the returned tool description
- Use the returned input schema
- Invoke the tool with `tools/call`
- Keep GUI-specific presentation metadata local where needed, such as:
  - icons
  - grouping
  - display labels
  - ordering
  - UI hints

## Important distinction

This issue does not require every existing GUI3-native tool to immediately move through MCP.

Migration should happen only for capabilities that are actually exposed through Lemonade `/mcp`.

This also does not necessarily remove the existing OmniRouter `toolDefinitions.json` or other tool definitions used by separate OpenAI-compatible orchestration paths.

Those files may serve consumers other than the GUI3 MCP client.

The first goal is to eliminate duplicate MCP contracts where GUI3 is consuming a capability that Lemonade already exposes as an MCP tool.

## Dependencies

This depends on the GUI3 TypeScript MCP client.

Migration of model-management capabilities additionally depends on those operations being exposed through Lemonade MCP where they are not already available.

## Acceptance criteria

- GUI3 does not hard-code MCP contracts for Lemonade capabilities that are discoverable through `/mcp`
- Tool contract changes in `lemond` become visible to GUI3 through `tools/list`
- GUI3 invokes migrated Lemonade capabilities through `tools/call`
- GUI-specific metadata remains independent from the MCP wire contract
- Existing non-MCP tool/orchestration paths are not removed unless they are independently obsolete

## Merge impact

Follow-up cleanup and architecture unification.

This should not block the initial GUI3 merge.

PR #3319, opened 2026-08-24
MCP: expose canonical Lemonade capability hub

### LINKED ISSUE #3275: MCP: expose model management capabilities used by GUI3
## Goal

Expose the Lemonade model-management capabilities needed by GUI3 through the Lemonade MCP server.

Related to #3270 and #3271 also #3274 

## Motivation

GUI3 should progressively be able to consume Lemonade capabilities through the same `/mcp` interface available to other MCP clients.

This also makes model-management functionality reusable by third-party MCP applications without duplicating GUI-specific contracts.

## Existing capability

Lemonade already exposes model discovery through the existing `lemonade_list_models` MCP tool.

That tool should be reused or extended where necessary rather than adding a second overlapping model-listing MCP tool.

## Candidate additional capabilities

Exact tool boundaries and names should be agreed separately, but explicit model lifecycle operations likely include:

API calls like
- pull/download a model
- explicitly load a model
- unload a model
- delete a model

If additional model discovery information is required by GUI3, prefer extending `lemonade_list_models` rather than creating a duplicate listing capability.

## Architecture

MCP handlers should adapt to the same underlying Lemonade model-management implementation used by the server's existing APIs.

MCP must not introduce a second implementation of model lifecycle behavior.

For example:

```text
GUI3 / third-party MCP client
        |
        v
      /mcp
        |
        v
MCP model-management handler
        |
        v
existing Lemonade model-management implementation
```

The MCP layer should be an adapter, not a second source of business logic.

## Security

These are control-plane operations, not just inference operations.

The implementation must account for:

- authentication and authorization
- argument validation
- model identifier validation
- download and resource limits
- destructive operations such as deletion
- concurrency and conflicting lifecycle operations
- error propagation
- logging or auditing where appropriate

Destructive operations should have clearly defined semantics and must not silently behave differently from the equivalent server API.

## Out of scope

- Replacing the underlying model-management implementation
- Creating MCP-specific model lifecycle state
- Duplicating the existing `lemonade_list_models` capability
- Requiring GUI-specific business logic inside MCP handlers

## Acceptance criteria

- Required explicit model-management operations are available through `/mcp`
- Existing model discovery continues to use `lemonade_list_models`, extended if necessary
- `tools/list` accurately describes every exposed model-management capability
- `tools/call` invokes the same underlying model-management logic as Lemonade's other interfaces
- MCP lifecycle operations return clear errors for invalid or conflicting requests
- No GUI-specific business logic is required to execute these operations
- MCP does not maintain a second model-management implementation

## Merge impact

Follow-up by default.

Only a specific model-management operation should become merge-enabling if GUI3 requires that operation through MCP on day one and no existing server API can provide the required behavior.

fl0rianr:
## Summary

Expand `/mcp` into the canonical Lemonade capability hub so clients can discover server-owned tool contracts through `tools/list` and execute them through `tools/call`.

This exposes model management, system/backend management, and multimodal capabilities while reusing the existing server/API implementation instead of duplicating business logic. It also prepares GUI3 to consume Lemonade capabilities dynamically rather than maintaining a separate tool contract.

Fixes #3275
Related to #3274

## Scope

## Testing

_Testing details:_

- `python test/server_mcp.py`
- 24 tests passed in 33.789s
- Verified the complete MCP tool catalogue is exposed through `tools/list`
- Verified advertised tools have matching `tools/call` dispatch

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -1211,7 +1211,8 @@ nlohmann::json Server::handle_mcp_server_tool(
 
     auto call_binary_handler = [this](const nlohmann::json& body,
                                       JsonHandler handler,
-                                      const std::string& label) {
+                                      const std::string& label,
+                                      const std::string& fallback_mime_type) {
         httplib::Request request;
         request.method = "POST";
         request.body = body.dump();
@@ -1221,10 +1222,15 @@ nlohmann::json Server::handle_mcp_server_tool(
         if (response.status >= 400) {
             return mcp_model_http_result(response);
         }
-        const std::string mime_type = response.get_header_value("Content-Type");
+        std::string mime_type = response.get_header_value("Content-Type");
+        if (mime_type.empty()) {
+            mime_type = fallback_mime_type.empty()
+                ? "application/octet-stream"
+                : fallback_mime_type;
+        }
         const std::string encoded = utils::JsonUtils::base64_encode(response.body);
         nlohmann::json payload = {
-            {"mime_type", mime_type.empty() ? "application/octet-stream" : mime_type},
+            {"mime_type", mime_type},
             {"data_base64", encoded},
         };
         nlohmann::json content = nlohmann::json::array();
@@ -1503,7 +1509,7 @@ nlohmann::json Server::handle_mcp_server_tool(
                 if (arguments.contains(key)) body[key] = arguments[key];
             }
             return call_binary_handler(
-                body, &Server::handle_audio_generations, "Generated audio.");
+                body, &Server::handle_audio_generations, "Generated audio.", "audio/wav");
         }
 
         if (tool_name == "lemonade_text_to_speech") {
@@ -1517,8 +1523,23 @@ nlohmann::json Server::handle_mcp_server_tool(
             for (const char* key : {"voice", "speed", "response_format"}) {
                 if (arguments.contains(key)) body[key] = arguments[key];
             }
+            std::string fallback_mime_type = "audio/mpeg";
+            const std::string response_format =
+                body.value("response_format", std::string("mp3"));
+            if (response_format == "wav") {
+                fallback_mime_type = "audio/wav";
+            } else if (response_format == "opus") {
+                fallback_mime_type = "audio/ogg";
+            } else if (response_format == "aac") {
+                fallback_mime_type = "audio/aac";
+            } else if (response_format == "flac") {
+                fallback_mime_type = "audio/flac";
+            } else if (response_format == "pcm") {
+                fallback_mime_type = "audio/l16";
+            }
             return call_binary_handler(
-                body, &Server::handle_audio_speech, "Generated speech audio.");
+                body, &Server::handle_audio_speech, "Generated speech audio.",
+                fallback_mime_type);
         }
 
         if (tool_name == "lemonade_generate_3d") {
@@ -1533,7 +1554,8 @@ nlohmann::json Server::handle_mcp_server_tool(
                 if (arguments.contains(key)) body[key] = arguments[key];
             }
             return call_binary_handler(
-                body, &Server::handle_3d_generations, "Generated GLB model.");
+                body, &Server::handle_3d_generations, "Generated GLB model.",
+                "model/gltf-binary");
         }
 
         return mcp_model_tool_result({{"error", "Unknown Lemonade server tool: " + tool_name}}, true);
--- a/src/cpp/server/server.cpp
+++ b/src/cpp/server/server.cpp
@@ -350,6 +350,99 @@ nlohmann::json get_model_storage_stats(const std::string& model_storage_path) {
     };
 }
 
+std::string mcp_model_result_text(const nlohmann::json& payload) {
+    if (payload.is_object()) {
+        auto error = payload.find("error");
+        if (error != payload.end()) {
+            if (error->is_string()) return error->get<std::string>();
+            if (error->is_object()) {
+                auto message = error->find("message");
+                if (message != error->end() && message->is_string()) {
+                    return message->get<std::string>();
+                }
+            }
+        }
+        auto message = payload.find("message");
+        if (message != payload.end() && message->is_string()) {
+            return message->get<std::string>();
+        }
+    }
+    if (payload.is_string()) return payload.get<std::string>();
+    return payload.dump();
+}
+
+nlohmann::json mcp_model_tool_result(nlohmann::json payload,
+                                     bool is_error,
+                                     const std::string& message = "") {
+    nlohmann::json structured = payload.is_object()
+        ? payload
+        : nlohmann::json{{"result", std::move(payload)}};
+    const std::string text = message.empty()
+        ? mcp_model_result_text(structured)
+        : message;
+    return {
+        {"content", nlohmann::json::array({{
+            {"type", "text"},
+            {"text", text},
+        }})},
+        {"structuredContent", std::move(structured)},
+        {"isError", is_error},
+    };
+}
+
+nlohmann::json mcp_model_http_result(const httplib::Response& response) {
+    nlohmann::json payload;
+    if (response.body.empty()) {
+        payload = nlohmann::json::object();
+    } else {
+        try {
+            payload = nlohmann::json::parse(response.body);
+        } catch (const std::exception&) {
+            payload = {{"body", response.body}};
+        }
+    }
+    if (!payload.is_object()) {
+        payload = {{"result", std::move(payload)}};
+    }
+    if (response.status > 0) {
+        payload["http_status"] = response.status;
+    }
+    return mcp_model_tool_result(std::move(payload), response.status >= 400);
+}
+
+void validate_mcp_model_arguments(
+        const nlohmann::json& arguments,
+        const std::set<std::string>& allowed) {
+    if (!arguments.is_object()) {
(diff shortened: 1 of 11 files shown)


falabellamichael:
Local verification on Windows (RX 6750 XT, VS 2026 / CMake 4.4.2) at head `bb185e46`:

- Built `lemond` from source: clean compile.
- Ran the server on an isolated cache/config dir and port 13305: `/mcp` `tools/list` advertises all 19 tools.
- `python test/server_mcp.py` against that server: **Ran 24 tests in 15.7s — OK** (all pass).

Also confirmed all three review comments are addressed at head: the `assumend` typo, the subset-based catalog assertion in `test/server_mcp.py`, and the per-tool fallback `mimeType` in `call_binary_handler` (audio/wav, per-format TTS types, model/gltf-binary).

Ready for maintainer review.

kpoineal (approved):
LGTM!

jeremyfowers:
Let's give @siavashhub until morning CT to reply.

edit: confirmed on discord that review is coming

siavashhub:
Nice work on this! delegating to the existing REST implementations instead of duplicating logic is clean.

## TLDR
non-blocking, but I'd prefer gating the five mutating tools behind a default-off flag before merge, explicit opt-in over default.

If this is intentional and we're comfortable giving agents lifecycle control by default, I won't block.

### Context / concern
One thing I want to flag on the expanded catalog, I deliberately kept model/backend *lifecycle* operations out of /mcp back then. The original gateway exposed inference-style tools (chat, transcribe, generate) and read-only discovery only, I intentionally did **not** expose destructive or state-changing actions, because I wanted the user to manage models and backends **on the server**, not hand those privileges to an agent (Copilot, Claude Desktop, Cursor).

This PR changes that. It now exposes at the /mcp tier, so a confused or injected agent could delete a user's models or trigger a backend install. confirm: true guards against accidents, but it isn't an authorization control:

- lemonade_delete_model (destructive)
- lemonade_pull_model (arbitrary registry download)
- lemonade_install_backend (installs/updates backend binaries, although it's constrained to known recipes, but still fetches and runs installers)
- lemonade_load_model / lemonade_unload_model (less riskier but still lifecycle)

The fair counterpoint to that might be that these REST endpoints (/v1/delete, /v1/pull, /api/v1/install) already exist at the same tier, so this isn't a raw privilege escalation for a direct HTTP client. I get it. I'm just raising awareness on the *agent-invokable* additions through the interface most likely to fire them unintentionally.

My ask: gate the five mutating tools (delete, pull, install, load, unload) behind a default-off capability flag ideally before merge, so handing model/backend lifecycle to agents is an explicit, documented opt-in rather than the default.

Concretely:
- Keep the read-only tools where they are (lemonade_list_models, get_*, list_backends, health, system-info), no concern there.
- Put the mutating tools behind a config flag that defaults to off. When it's off, they shouldn't just be refused on call, they should be dropped from tools/list entirely, so an injected agent never even sees them. I'd keep the flag as the primary control rather than leaning on the admin key alone, since an agent handed that key for convenience is still the prompt-injectable party

### Non-blocking issues if we were to merge:
- I think we should update these files server_mcp_smoke.py, mcp-smoke-test.yml, persona-demo.css and apps.js where they are mentioning only 5 tools
- New binary tools only return inline base64. lemonade_edit_image, lemonade_generate_audio and lemonade_generate_3d always inline base64, whereas the existing lemonade_generate_image/lemonade_omni support writing to the MCP image sandbox (output_dir) precisely to avoid huge JSON-RPC payloads. I did design it like that earlier on the gateway since a GLB or audio clip inline can get large. So my recommendation would be to consider offering the same sandbox-output option for consistency (follow-up, not a blocker).
- Pre-existing media tools go through McpServer::tool_*, the new ones go through Server::handle_mcp_server_tool. Fine given "keep refactoring separate" but just a note for future consolidation.

jeremyfowers:
> My ask: gate the five mutating tools (delete, pull, install, load, unload) behind a default-off capability flag ideally before merge, so handing model/backend lifecycle to agents is an explicit, documented opt-in rather than the default.

If there is a convenient way to gate these I'm in favor of this. Especially for `delete` (and `uninstall` if that is ever considered), since it would be incredibly annoying for an agent to delete GBs of data against my wishes.

GUI3 would also ideally ask for user permissions before letting the LLM run them.

edit: `install` should be gated too. any kind of generate, load, etc. can still install as a side-effect, but it would be very annoying and potentially harmful for the LLM to go around unnecessarily installing things.

fl0rianr:
> > My ask: gate the five mutating tools (delete, pull, install, load, unload) behind a default-off capability flag ideally before merge, so handing model/backend lifecycle to agents is an explicit, documented opt-in rather than the default.
> 
> If there is a convenient way to gate these I'm in favor of this. Especially for `delete` (and `uninstall` if that is ever considered), since it would be incredibly annoying for an agent to delete GBs of data against my wishes.
> 
> GUI3 would also ideally ask for user permissions before letting the LLM run them.
> 
> edit: `install` should be gated too. any kind of generate, load, etc. can still install as a side-effect, but it would be very annoying and potentially harmful for the LLM to go around unnecessarily installing things.

Now you confused me - Did we not agree on that's API? We want it easy for users - we can then decide for an user ok or not depending on UX.
Change it or not??

jeremyfowers:
> > > My ask: gate the five mutating tools (delete, pull, install, load, unload) behind a default-off capability flag ideally before merge, so handing model/backend lifecycle to agents is an explicit, documented opt-in rather than the default.
> > 
> > 
> > If there is a convenient way to gate these I'm in favor of this. Especially for `delete` (and `uninstall` if that is ever considered), since it would be incredibly annoying for an agent to delete GBs of data against my wishes.
> > GUI3 would also ideally ask for user permissions before letting the LLM run them.
> > edit: `install` should be gated too. any kind of generate, load, etc. can still install as a side-effect, but it would be very annoying and potentially harmful for the LLM to go around unnecessarily installing things.
> 
> Now you confused me - Did we not agree on that's API? We want it easy for users - we can then decide for a user ok or not depending on UX. Change it or not??

after considering it, let’s leave it enabled (no change). Users will let us know if it has problematic outcomes.

SlawomirNowaczyk:
I suggest merging this now, but there are two minor issues I'd like to see fixed as near-future follow-ups:

1. **`server.cpp` `resolve_media_model` duplicates model-selection policy (thin-wrapper violation).** It hand-rolls the loaded→downloaded precedence that `McpServer::resolve_model_for_tool` already owns, but *diverges*: `resolve_media_model` has no `allow_download`/default-model step and throws instead of returning a structured `needs_model` result, so `edit_image`/`generate_audio`/`text_to_speech`/`generate_3d` behave differently from `chat`/`transcribe`/`generate_image` when no model is present. Suggested fix: extract a single capability-based selector (ideally on `Router`/`ModelManager`, per #3426) and have both dispatch tables call it; at minimum reuse `resolve_model_for_tool` rather than adding a third variant.

2. **Media tools lack MCP annotations.** All eleven model-management tools carry `readOnlyHint`/`destructiveHint`/`openWorldHint`, but `lemonade_edit_image`, `lemonade_generate_audio`, `lemonade_text_to_speech`, and `lemonade_generate_3d` have none (see mcp_server.cpp tools_descriptor). Agents can't see these are non-read-only/open-world. The completeness test only checks annotations for the 11 mgmt tools, so the gap is untested. Suggested fix: add annotations to the four media tools.

src/cpp/server/server.cpp
```diff
@@ -1092,6 +1185,385 @@ httplib::Server::HandlerResponse Server::authenticate_request(const httplib::Req
 }
 
 
+
+nlohmann::json Server::handle_mcp_server_tool(
```
jeremyfowers, line 1189: Is this not completely redundant with the `McpServer` class that already exists?

One example of why this is problematic: the pre-existing `lemonade_generate_image` method in `McpServer` correctly handles blob data, but the new blob-related tools here do not.

jeremyfowers (changes requested):
This is a really tricky PR to merge. On the one hand, the new capabilities are awesome and people are in favor of the general idea of more MCP capabilities.

On the other hand, this PR introduces a very large amount of surface area backed by a lot of boilerplate code. I am concerned that the implementation has not been fully vetted (see https://github.com/lemonade-sdk/lemonade/pull/3319#discussion_r3971712082). I'm also concerned about the amount of boilerplate documentation that is being introduced that must be kept in sync with the code.

Here are the changes I want to see that would make this, and related PRs, review-able at a deep and convincing level:

1. Apply the following changes to the pre-existing MCP tools **only** as a starting point (refactor first), then open subsequent smaller PRs that introduce a minimal number of new tools on the new pattern.
2. Our MCP architecture should follow the pattern set by our backends: information should have a single source of truth, a standardized place to live, and all boilerplate should be auto-generated from that.
3. Come up with a base class (or struct) for an MCP tool and re-map our 5 existing tools to that data structure. **Large amounts of code should not spill into server.cpp.**
4. Organize the class instances into a list so that we don't need all these code blocks that list all the tools.
5. Generate the documentation by starting a live server and calling the `tools/list` endpoint, just like we do for backend documentation by calling the `system-info` endpoint. Protect this with the docs-drift CI job.

Next, in one PR per tool, add individual tools in priority order. This will allow reviewers to 1) very easily see that the new tool is following the MCP tool-adding pattern, and 2) test the new tool in real-world scenarios in an interactive session with an agent.

PR #3322, opened 2026-08-24
Retire undocumented /params, /log-level, /api/v1/test, and /status routes

jeremyfowers:
## Background

These 4 endpoints accidentally made it into the modern lemonade codebase during the change from python to C++. I've been meaning to get rid of these for ages, but never got around to it.

Cleaning this up is an important part of ensuring we have accurate docs and practices.

## Summary

Retires four undocumented HTTP routes and moves their first-party callers onto the existing `/internal/*` config endpoints.

- `POST /{api/}v{0,1}/params` and `POST /{api/}v{0,1}/log-level` were thin wrappers over the same `RuntimeConfig::set()` that `POST /internal/set` calls; `GET /{api/}v{0,1}/params` was an alias of `GET /internal/config` that exposed the full config snapshot on the public prefix.
- `POST /api/v1/test` was a debug stub ("TEST POST endpoint hit!") with no callers, v1-only, breaking the quad-prefix invariant.
- `GET /status` was a second mount of the legacy status page; the page is still served at `GET /api/v1`.

Client migrations: the tray context-size menu (`tray_ui.cpp`) and the desktop app's log-level selector (`LogsWindow.tsx`) now `POST /internal/set`. `LogsWindow` already read the current level from `GET /internal/config`, so both directions now use one API.

@fl0rianr please note a minor change to GUI2 here.

Note: `/internal/set` persists to `config.json`, which `/params` and `/log-level` did not. The tray's context size and the app's log level now survive a restart — intended, but calling it out as an observable behavior change.

## Scope

## Testing

_Testing details:_

- `cmake --build --preset default --target lemond` — clean.
- `npx tsc --noEmit` in `src/app` — clean.
- Ran `lemond` and confirmed `GET/POST /api/v1/params`, `/v1/params`, `/api/v1/log-level`, `/api/v1/test`, and `/status` all return 404, while `/api/v1`, `/api/v1/health`, `/app`, and `GET /internal/config` still return 200.
- Confirmed the replacement calls: `POST /internal/set {"ctx_size":8192}` and `POST /internal/set {"log_level":"debug"}` both return `{"status":"success","updated":{...}}`.
- Migrated `test/server_telemetry.py` (18 `/api/v1/params` call sites → `/internal/set` for writes, `/internal/config` for reads) and `test/validate_llamacpp.py`.
- **Not built locally:** `tray_ui.cpp` — the Linux tray target is gated off in my configuration (`REQUIRE_LINUX_TRAY=OFF`). The change there is a one-line endpoint string swap.

## Documentation

All four routes were undocumented; the only references were the endpoint list in `AGENTS.md`, which is updated here.

## Breaking Changes

`/v1/params`, `/v1/log-level`, `/api/v1/test`, and `/status` are removed. None were documented, so no third-party integration should be relying on them, but any that is will need to move to `POST /internal/set` / `GET /internal/config` (admin-key gated when `LEMONADE_ADMIN_API_KEY` is set).

## AI-assisted contribution

Please select one:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,7 @@ All core endpoints are registered under **4 path prefixes**:
 - `/v0/` — Legacy short
 - `/v1/` — OpenAI SDK / LiteLLM compatibility
 
-**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `health`, `pull`, `pull/variants`, `registry/search`, `load`, `unload`, `delete`, `params`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `log-level`, `logs/stream`, `jobs`, `jobs/{id}`, `jobs/{id}/pause`, `jobs/{id}/interrupt`, `jobs/{id}/resume`
+**Core endpoints:** `chat/completions`, `completions`, `embeddings`, `reranking`, `models`, `models/{id}`, `health`, `pull`, `pull/variants`, `registry/search`, `load`, `unload`, `delete`, `install`, `uninstall`, `audio/transcriptions`, `audio/speech`, `images/generations`, `images/edits`, `images/variations`, `responses`, `stats`, `system-info`, `system-stats`, `logs/stream`, `jobs`, `jobs/{id}`, `jobs/{id}/pause`, `jobs/{id}/interrupt`, `jobs/{id}/resume`
 
 **Job engine** (`POST jobs`, `GET jobs`, `GET/DELETE jobs/{id}`, `POST jobs/{id}/{pause,interrupt,resume}`): server-side sequences of ops (`system_info`, `system_stats`, `models`, `sleep`, `load`, `unload`, `chat`) with data passing, forward-only branching, and a pause/interrupt/resume lifecycle persisted across restart. Exclusive ops hold a Router slot so normal traffic queues. See `docs/dev/job-system.md` and `docs/dev/job-expression-language.md`.
 
--- a/docs/assets/persona-demo/developers/interfaces.js
+++ b/docs/assets/persona-demo/developers/interfaces.js
@@ -144,7 +144,7 @@
     { title: 'System', icon: 'monitor_heart', eps: [['GET', '/v1/system-info'], ['GET', '/v1/system-stats']] },
     { title: 'Models', icon: 'inventory_2', eps: [['POST', '/v1/pull'], ['POST', '/v1/load'], ['POST', '/v1/unload'], ['POST', '/v1/delete']] },
     { title: 'Backends', icon: 'memory', eps: [['POST', '/v1/install'], ['POST', '/v1/uninstall']] },
-    { title: 'Config & scale', icon: 'tune', eps: [['POST', '/v1/params'], ['POST', '/v1/cloud/auth'], ['POST', '/internal/set']] },
+    { title: 'Config & scale', icon: 'tune', eps: [['POST', '/v1/cloud/auth'], ['POST', '/internal/set'], ['GET', '/internal/config']] },
     { title: 'Lifecycle', icon: 'restart_alt', eps: [['GET', '/v1/health'], ['GET', '/v1/logs/stream'], ['POST', '/internal/shutdown']], wide: true }
   ];
 
--- a/src/app/src/renderer/LogsWindow.tsx
+++ b/src/app/src/renderer/LogsWindow.tsx
@@ -61,9 +61,7 @@ const LogsWindow: React.FC<LogsWindowProps> = ({ isVisible, height }) => {
       setIsInitialized(true);
 
       try {
-        // Using serverFetch with the full URL to reach the root-level /internal/config
-        // endpoint while benefiting from automatic authentication and init-waiting.
-        const response = await serverFetch(`${url}/internal/config`);
+        const response = await serverFetch('/internal/config');
         if (!response.ok) return;
 
         const config = await response.json();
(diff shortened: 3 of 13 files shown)


fl0rianr (commented):
Overall this looks like a good cleanup and the server-side removal/migration seems solid. I found one issue I think is worth fixing before merge, plus a couple of smaller follow-ups:

* `serverConfig.fetch()` handles `/internal/*` correctly on the first request, but after port rediscovery the retry always rebuilds relative URLs via `getApiBaseUrl()`. So `/internal/set` becomes `/api/v1/internal/set` and the GUI log-level update will fail with 404 after a port change/restart. I’d reuse the same endpoint resolver for both the initial request and retry.
* The telemetry tests now use persistent `/internal/set` instead of runtime-only `/params`, but teardown only disables telemetry. That means mock endpoint/protocol/etc. values can remain in `config.json` after a local test run. It would be better to snapshot and restore the original telemetry config.
* Minor: internal test callers currently mostly use `LEMONADE_API_KEY`; with a distinct `LEMONADE_ADMIN_API_KEY`, `/internal/*` calls can fail. Using admin-key-with-API-key-fallback would match the server auth model.
* A small negative regression test asserting the retired `/params`, `/log-level`, `/test`, and `/status` routes stay 404 would also be useful.

With the retry-path fix, I’d be comfortable with this.

jeremyfowers:
Addressed in 40cb32d0b:

- `/internal/*` path rewriting on port-rediscovery retry — shared `resolveUrl()` now used by both the initial request and the retry.
- Telemetry teardown — snapshots and restores the original config instead of only disabling.
- Admin API key preferred in the shared test auth helpers (including the `/internal/telemetry/flush` call).
- Added `test_000b_retired_endpoints_removed` as a negative regression test.

fl0rianr (approved):
Nice, thanks!

One small non-blocking thing I noticed: validate_llamacpp.py now uses /internal/set for rocm_channel, but its request_json() still sends no auth header. This will fail if LEMONADE_ADMIN_API_KEY is configured, so it would be nice to reuse the shared auth helper there as well.

Nothing blocking from me. Looks good now.

kenvandine (approved):
Thanks, I'm happy with this now.

ramkrishna2910:
Looks good!

ramkrishna2910 (approved):
Approving. I verified the removal is complete and the migration is faithful rather than taking the description on trust, and I checked that the items raised in the earlier round are actually closed at head.

**Removal is complete across all four prefixes.** The `register_post`/`register_get` helpers fan out to `/api/v0/`, `/api/v1/`, `/v0/`, and `/v1/`, so deleting the registration removes all four mounts. Handler declarations are gone from `server.h`, and the `register_post` special case is correctly narrowed from `endpoint != "params" && endpoint != "jobs"` to just `jobs` — easy to leave behind as dead weight, and it wasn't. I swept the whole repo for surviving references and found none: the remaining `log_level` hits are the *config key* (`/internal/set`, `lemonade config set`, `config.json`), and the remaining `params` hits in `server.cpp` are the job engine's unrelated JSON field.

**The migration preserves behavior.** The claim that `/params` was a thin wrapper over the same call holds: `handle_config_set` invokes the identical `config_->set(body, apply_config_side_effects)` that `handle_params` did, so no side effect is dropped in the move. Persistence to `config.json` is the only behavioral delta and you already call it out. The response-shape and request-key changes are handled on the client side too (`data.level` → `data?.updated?.log_level`, `level` → `log_level`).

**One thing the description undersells: this is a security fix, not just a cleanup.** `GET /params` aliased `GET /internal/config` — the full config snapshot — on the *public* prefix, reachable with the regular API key. The POST side is the same story: mutating global server config now requires the admin key when a distinct one is configured. That's the strongest argument for this change and it reads as a footnote right now.

**Earlier review items all land at head**, which is worth stating since both approvals carried caveats:

- The port-rediscovery retry bug is fixed by extracting `resolveUrl` and using it on both paths. That also closes a genuine latent bug — the retry previously rebuilt every relative URL under `getApiBaseUrl()`, so `/internal/config` would have become `/api/v1/internal/config` on any retry, not just the newly migrated calls.
- Telemetry snapshot/restore is in.
- `_auth_headers()` now prefers `LEMONADE_ADMIN_API_KEY` in both `server_base.py` and `reset_server_state.py`.
- The negative regression test is thorough: 3 endpoints x 4 prefixes x 2 methods, plus `/status` returning 404 and `/api/v1` still returning 200. That last assertion is the one that matters most — it pins that the status page survived the move.

I also checked the CLI, since `lemonade status` shares a name with a removed route: it calls `/api/v1/health`, so there's no breakage there.

Two non-blocking notes:

1. **The breaking-changes section covers third-party integrations but not first-party clients.** With `LEMONADE_API_KEY` and a distinct `LEMONADE_ADMIN_API_KEY` both set, a *remote* desktop app configured with only the regular key loses the log-level and context-size controls, because `/internal/*` requires the admin key while `/v1/params` did not. The tray is unaffected — it prefers the admin key and reads it from the local environment. This is the correct posture given point 3 above, so I'm not asking you to change it; it just deserves a sentence, since the app is explicitly supported as a remote client. Worth noting `LogsWindow` surfaces the rejection as a `console.error` and a silently reverted dropdown rather than a message, which would make that case hard to diagnose in the field.

2. **`validate_llamacpp.py` still sends no auth header**, as flagged earlier. Confirming it is pre-existing rather than a regression: that file never sent auth for `/params` either, so a configured key would have 401'd before this change too. Fine to leave, though the shared helper is right there. Small separate nit while you're in that file: `base_url.rsplit('/api/v1', 1)[0]` silently yields a doubled slash if `base_url` ever carries a trailing slash.

Nice cleanup — the endpoint sweep this enables is the real payoff.

jeremyfowers:
> validate_llamacpp.py now uses /internal/set for rocm_channel, but its request_json() still sends no auth header. This will fail if LEMONADE_ADMIN_API_KEY is configured, so it would be nice to reuse the shared auth helper there

Opened https://github.com/lemonade-sdk/lemonade/issues/3336

PR #3325, opened 2026-08-24
docs: document POST /internal/simulate-vram-pressure

jeremyfowers:
## Summary

`POST /internal/simulate-vram-pressure` was the one `/internal/*` endpoint missing from the internal endpoints table in `docs/dev/getting-started.md`. It synchronously fires the eviction engine's VRAM pressure callback, which is how `test/server_eviction.py` and `test/server_pinning.py` get a deterministic eviction pass without a real VRAM reading. Documents the `pct` semantics — including the `-1` idle-only evaluation that the tests use — and notes that pinned models, `auto_evict: false` models, and active exclusive job sessions are all still respected.

Docs only; no code changes.

cc @clemperorpenguin — you added this endpoint in #2183 as part of dynamic VRAM management, so please confirm the `pct` semantics (threshold comparison vs. the `-1` idle-only path) read correctly.

Fixes #

## Scope

## Testing

_Testing details:_

- Against a running `lemond`: `{"pct": 0.95}`, `{"pct": -1.0}`, and `{}` all return `200 {"status": "ok"}`; a non-JSON body and `{"pct": "x"}` both return `400` with an `error` field — matching what the section documents.
- The three `pct` branches were traced through `EvictionEngine::on_vram_pressure` / `evaluate_servers`, and the protections (pinned, per-model `auto_evict`, `exclusive_active_`) read off `evaluate_servers` directly.
- The `-1` idle-only behavior matches how `test/server_eviction.py` uses it (`IDLE_EVALUATION_PCT = -1.0`, `_evaluate_idle_now`).

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

### DIFF
--- a/docs/dev/getting-started.md
+++ b/docs/dev/getting-started.md
@@ -630,6 +630,7 @@ Internal endpoints accept connections from any address, so first-party clients o
 | `POST` | `/internal/pin` | Pin or unpin a loaded model |
 | `POST` | `/internal/models/sync` | Sync/update downloaded models to the latest version (async or dry-run) |
 | `GET`  | `/internal/models/sync/status` | Retrieve status/progress of background model synchronization |
+| `POST` | `/internal/simulate-vram-pressure` | Test hook: synchronously run one eviction-engine evaluation |
 
 #### `POST /internal/set`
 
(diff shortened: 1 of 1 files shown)


jeremyfowers:
@fl0rianr @clemperorpenguin  PR https://github.com/lemonade-sdk/lemonade/pull/2183 added a new endpoint without adding documentation for it. Please shepherd this PR and ensure that the endpoint is fully documented.

The draft PR here is fully AI generated, feel free to adjust it in any way needed before bringing it out of draft.

clemperorpenguin (changes requested):
I read through this against the source. and the `pct` semantics look right to me.

`pct == -1.0` is exact equality (`eviction_engine.cpp:39`), so "Exactly `-1`" and the other-negatives bullet both hold. Sub-threshold returns before `evaluate_servers` is reached (`eviction_engine.cpp:43-47`), so that bullet is right too. The synchronous claim traces cleanly from `handle_simulate_vram_pressure` through `Router::simulate_vram_pressure` (`router.cpp:674`) and `GlobalVramMonitor::simulate_pressure` (`global_vram_monitor.h:31`) into `evaluate_servers`, with the downsizes and `evict_if_committed` finishing before the response.

A few things we could change:

**Exclusive sessions.** The code reads `if (router_->exclusive_active_) return;`, which looks more like a skip than a deferral, since I don't see anything rescheduling it. The endpoint still returns `200`, so I think a test could get a green response for a sweep that never ran. Suggested wording:

> If an exclusive job session is active the evaluation is skipped and not rescheduled. The endpoint still returns `200`, so a successful response does not on its own mean an evaluation ran.

**Scoring direction.** "Highest-scoring" is accurate but the score isn't defined anywhere, and I read it backwards at first:

> the score is `idle_time / (load_duration * evict_weight_factor)`, so higher means more disposable. Models that are idle and cheap to reload go first, and a larger `evict_weight_factor` protects a model.

**Hard idle ordering.** The loop breaks on the first server past its `evict_idle_timeout` in `loaded_servers_` order rather than the most idle one, so "the first model found past its hard idle timeout" may be more precise.

One thing potentially worth adding: the per-model `auto_evict` option overrides the global configuration, and the current phrasing reads as per-model only. If the global is off with no override, both paths end up no-ops. Naming `evict_idle_timeout`, `downsize_idle_timeout`, and `evict_weight_factor` would also help, since those seem to be what determines what the `-1` sweep actually does.

Otherwise this looks good, and sorry I missed documenting the endpoint in #2183.

fl0rianr:
Thanks! Hope it's better now.

clemperorpenguin (approved):
cool, thank you! looks good!

fl0rianr (approved):

PR #3333, opened 2026-08-24
fix(server): classify extra models using reserved directories: chat, embeddings, reranking instead of treating all model as chat

### LINKED ISSUE #1667: Bug: Models imported via extra_models_dir are all incorrectly identified as LLMs
### Platform

Windows

### Lemonade Version

10.2.0

### GPU / APU Model

_No response_

### Component

Other / Not sure

### Bug Description

When importing local models (especially embedding models) using the lemonade config set extra_models_dir command, all models are incorrectly identified as LLM (Large Language Model) models. This prevents the embedding models from functioning properly as they lack the embeddings label.

### Steps to Reproduce

1. Download an embedding model (e.g., Qwen3-Embedding-0.6B-Q8_0.gguf).
2. Run lemonade config set extra_models_dir="./path/to/your/models".
3. Restart Lemonade Server.
4. Check the model list in the Web UI or via API.
5. See error: The embedding model is listed as an LLM.

Lemonade should intelligently recognize the model type based on the GGUF file's metadata or filename, and assign the appropriate labels (e.g., embeddings). Alternatively, it should provide a mechanism (like a config file) to manually specify the model type during import.

### Log Output

```shell

```

### Additional Context

_No response_

anditherobot:
## Summary

Fixes #1667.

Models discovered through `extra_models_dir` were always classified as chat models. Embedding and reranking GGUFs therefore started llama.cpp in the wrong runtime mode and could not use their intended endpoints.

**The convention: the folder a model sits in tells the server how to run it. Filenames are never used to guess.**

| Directory | Runtime mode |
|---|---|
| `chat/` | Chat completion |
| `embeddings/` | Embeddings |
| `reranking/` | Reranking |
| Root or any other directory | Chat completion |

For example:

```text
extra_models_dir/
├── chat/
│   ├── Qwen3-8B-Instruct.gguf
│   └── Gemma-3-4B-It.gguf
├── embeddings/
│   ├── nomic-embed-text-v2.gguf
│   └── all-MiniLM-L6-v2.gguf
└── reranking/
    ├── bge-reranker-v2-m3.gguf
    └── Qwen3-Reranker-0.6B.gguf
```

Files directly inside a reserved directory are discovered separately. Nested folder models and split variants inherit the reserved directory's mode.

> **Fallback:** Models placed at the root of `extra_models_dir`, or under any non-reserved top-level directory, continue to use chat completion mode.

**Exact matching.** Reserved directory names must match exactly. `embeddings` is reserved; `Embedding`, `embedding`, and `embeddings 2` are ordinary directories. The server does not try to infer near-matches because doing so could accidentally select the wrong runtime behavior. Filenames are not used to infer model type for the same reason.

**Naming order.** IDs are assigned in this order so they remain stable across machines and restarts:

| # | Group | Order within the group |
|---|---|---|
| 1 | Files at the root of `extra_models_dir` | Filename |
| 2 | Files inside a reserved directory | Path |
| 3 | Directory models | Path |

If a later model would use an ID that is already taken, its directory is added as a qualifier. This means adding a reserved-directory model does not rename an existing root model.

Existing explicit registrations in `user_models.json` are unaffected.

**Legacy directory IDs.** Existing requests using `extra.chat`, `extra.embeddings`, or `extra.reranking` must continue to resolve to the same model after upgrading. These IDs are therefore reserved for backward compatibility.

If a model filename would create one of these IDs, its directory is added as a qualifier. For example, `embeddings/embeddings.gguf` is listed as `extra.embeddings-embeddings`, leaving `extra.embeddings` available as the legacy alias. Root files keep their existing naming priority.

## Scope

## Testing

_Testing details:_

Windows, Visual Studio 2022, Release:

- Rebased onto current `origin/main`.
- Built the `cpp-ci-tests` target.
- Ran:

  ```text
  ctest --test-dir build -C Release -L cpp-ci --output-on-failure
  ```

- 63/63 tests passed.
- Added filesystem coverage for:
  - reserved `chat/`, `embeddings/`, and `reranking/` directories;
  - root-file and ordinary-directory chat fallback;
  - direct files and nested folder models;
  - split variants;
  - filenames that conflict with the selected directory;
  - case-sensitive directory matching;
  - same-named models in different reserved directories;
  - equivalent directory paths (`models`, `models/`, `models/.`, `parent/../models`);
  - stable IDs for same-named models at the root and inside a reserved directory;
  - `extra.<reserved-directory>` still resolving after upgrade;
  - filenames that collide with legacy reserved-directory IDs, including collisions across reserved directories;
  - multimodal models and `mmproj`.

## Documentation

The convention is now stated explicitly rather than implied by examples, so users are told what each location means instead of the server inferring it:

| File | Published | What it now states |
|---|---|---|
| `docs/embeddable/models.md` | Yes, Embeddable → Models | The folder convention up front, the exact-matching rule, and an upgrade note for anyone who already has a `chat`, `embeddings`, or `reranking` folder |
| `docs/guide/configuration/README.md` | Yes, Server → Service | The `extra_models_dir` row points at the convention |
| `src/cpp/Extra-Models-Dir-Spec.md` | No, internal spec | The convention as a scanning rule, the naming order, and how a reserved directory's previous ID is preserved |

`src/cpp/Extra-Models-Dir-Spec.md` is not in `mkdocs.yml`, so anything a user needs to know is also in the published pages.

## Breaking Changes

This fixes incorrect automatic classification. Existing root files, ordinary directories, and explicit `user_models.json` registrations retain their previous behavior. The top-level names `chat`, `embeddings`, and `reranking` now have documented reserved meanings inside `extra_models_dir`.

Files in reserved directories are now listed individually rather than as a single model named after the directory. That ID is retained as a hidden input alias resolving to the same file, so existing requests continue to work.

## AI-assisted contribution

Please select one:

If AI tools were used:


jeremyfowers:
**Ready for review**

No to-dos.

The description accurately covers all changes: the classification fix, the secondary label-ordering fix, the /pull preview consistency change, and the documented limitations; Fixes #1667 is verified against the issue body.

Checked: Read PR body and all 8 commits against the diff; verified Fixes #1667 by reading the issue body and comments which describe embedding models being misclassified as LLMs in extra_models_dir.

**Focus: focused**

All changes serve one goal: fixing the extra_models_dir model classification. The /pull preview change ensures consistency with model registration, and the documentation updates describe the new behavior.

Checked: All touched files serve the classification fix: model_types.h adds infer_labels_from_name(), model_manager.cpp and hf_variants.cpp apply it at all discovery sites, CMakeLists.txt registers the new test, and docs reflect the new behavior.

**Alignment: none found**

Checked: reviewed contribute.md and philosophy.md; the change is a contained fix with a linked issue (#1667) and no scope creep.

**Documentation: adequate**

Checked: reviewed docs/embeddable/models.md and src/cpp/Extra-Models-Dir-Spec.md; both updated in the same PR to reflect the new classification behavior.

**Testing: adequate**

Checked: reviewed CMakeLists.txt, test/cpp/test_extra_model_discovery.cpp, and test/cpp/test_model_type_classifier.cpp; the new test is registered with add_cpp_ci_test and covers root files, split variants, folder models, filename-vs-folder conflict, mixed folders, and multimodal folders.

Checked: none observed — the change fixes misclassified models to their correct mode but does not remove or rename any user-facing surface.

</details>

**Review suggestion**

Suggested reviewers:

- jeremyfowers (model classification and registration) — Author of 62 blamed lines across model_types.h, model_manager.cpp, Extra-Models-Dir-Spec.md, and test_model_type_classifier.cpp — last commit 2026-08-14 on feat(models): add an explicit chat label (#3099).
- ramkrishna2910 (model types and model manager) — Author of 10 blamed lines in model_manager.h, model_types.h, and test_model_type_classifier.cpp — last commit 2026-06-14 on Cloud offload: provider-agnostic OpenAI-compat backend (#1785).
- No maintainer listed for: model classification
- No maintainer listed for: model discovery

**Why this rung**

A contained fix or extension. The diff is a contained fix to the model classification logic in model_manager.cpp and model_types.h, which maps to 'a contained fix or extension' on the code structure surface.

**Reviewer search**

Checked: @jeremyfowers (62 blamed lines) and @ramkrishna2910 (10 blamed lines) are the top code authors by relevance after the PR author. @bitgamma and @sawansri both cover 'LLMs' in the maintainer table and are appropriate for model classification changes.

</details>

_[AI-assisted review] Automated pre-review from repo-manager — flags for the human reviewer, not a replacement for one. Reviewed at head `37b2b4f`; regenerate with `repo-manager review-pr 3333`._

jeremyfowers:
This is a little outside my wheelhouse and I know Krishna is overloaded right now, so I posted on discord to solicit a reviewer: https://discord.com/channels/1392562559122407535/1394414791593820252/1542164758457884863

bitgamma:
@anditherobot I feel like the file-name approach is very fragile and I would prefer to remove it to avoid strange naming conflicts and the false expectations that we are actually detecting the model kind in a reliable way. But the directory based approach can work as long as it is properly documented.

bitgamma:
additionally, we have a PR template that we introduce to streamline the reviewing process. Please correct the PR description by using the template.

fl0rianr (commented):
Thanks! The underlying fix is good: moving the default `chat` label after classification is correct, the discovery paths are wired consistently, and the new filesystem tests are valuable.

My main concern is the filename-based classification. This does not only affect display labels: it changes `ModelInfo::type`, which directly controls whether llama.cpp starts with `--embeddings` or `--reranking`. A false positive therefore starts the model in the wrong runtime mode. Matching arbitrary substrings like `embed`, `rerank`, plus family-specific exceptions such as `bge-`, feels too fragile and will likely grow into an incomplete family whitelist.

I would prefer an explicit directory-based contract for `extra_models_dir`, e.g. reserved `chat/`, `embeddings/`, and `reranking/` folders, with everything unspecified continuing to default to chat. That is deterministic, backwards-compatible, and expresses user intent without guessing from filenames. The existing discovery tests can mostly stay, just with the policy expectations changed.

I would also keep the `/pull` change separate unless we explicitly want to change its behavior here, since #1667 is specifically about `extra_models_dir`.

So I suggest:

* keep the label-ordering fix and filesystem regression tests
* replace filename inference with explicit directory-based mode selection
* keep unspecified models defaulting to chat
* either drop or split the `/pull` classifier change
* update the user-facing docs accordingly
* update the PR description to the current template
* rebase/update on current `main` and run CI

bitgamma (changes requested):
This new approach is good. There are just a couple of small things that I'd like to see fixed.
                                                                                                                                                                                                                                                                     
1. Fragile path equality in directly_in_category. The check `parent_dir.parent_path() == search_path` is a raw `fs::path` comparison. The deployment label uses lexically_relative() (robust), but this branch relies on exact path equality. A search_path given with ./, trailing separators, or .. normalization differences could make reserved-dir direct files get grouped instead of standalone, inconsistently with the label logic. Normalizing search_path once would harden this.                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                       
2. Test hook in production class. friend struct ExtraModelDiscoveryTestHook is added to ModelManager purely for tests.  A small public/guarded accessor would be cleaner.

fl0rianr (commented):
This looks much better and the previous concerns are addressed. I found one remaining determinism issue: same-named direct models in different reserved directories can receive different public IDs depending on recursive_directory_iterator order. standalone_files is not sorted, while add_extra_model() lets the first occurrence keep extra.<name> and only qualifies later collisions. The current same-name test checks count/types but not the resulting IDs, so both orderings pass. I would make that ordering deterministic and pin the exact IDs in the test.
After that, this looks good to me.

bitgamma (approved):
lgtm now, @fl0rianr want to double check that your concern was addressed as you intended?

fl0rianr (commented):
The determinism fix looks good and addresses my previous concern as intended. I found one small backward-compatibility edge case though:

```text
extra_models_dir/
├── model.gguf
└── embeddings/
    └── model.gguf
```

Because standalone files are now sorted lexicographically by relative path, `embeddings/model.gguf` is processed before the root `model.gguf`. That means the embedding model gets `extra.model`, while the existing root model gets the qualified collision ID.

Previously, root files were processed first, so `extra.model` referred to the root model. This can therefore change an existing public model ID, or worse, make `extra.model` suddenly refer to a different model type.

I would keep root standalone files first, then sort the remaining reserved-directory files deterministically, and add a regression test for this root-vs-category collision.

After that, I’m happy to approve.

fl0rianr (commented):
The previous ID stability issue is fixed correctly now. I found one remaining functional edge case before I’m comfortable approving:

Direct multi-file models inside a reserved directory no longer go through the existing folder-model path.

For example:

```text
extra_models_dir/
└── chat/
    ├── model.gguf
    └── mmproj-model.gguf
```

On `main`, these files are grouped as one directory model and the `mmproj` is attached. With this PR, both files are treated as standalone candidates, then the `mmproj` file is skipped. The old `extra.chat` alias still resolves, but it now points to a model without its vision checkpoint.

The same structural issue can affect sharded GGUFs placed directly inside a reserved directory, since they also bypass `discover_extra_models_in_directory()`.

The current multimodal test uses an extra subdirectory (`chat/gemma-vision/...`), so it does not cover this case.

I would either preserve the existing multi-file grouping for direct reserved-directory models, or explicitly make the subdirectory structure part of the contract and document the compatibility change. I’d strongly prefer preserving the old behavior.

After that and I’m happy to approve.

fl0rianr (commented):
I rechecked the current head. One small backward-compatibility collision remains:

If `embeddings/` contains both `a.gguf` and `embeddings.gguf`, the legacy `extra.embeddings` alias is assigned to `a.gguf`, but `embeddings.gguf` also creates a real model with the canonical ID `extra.embeddings`. The real model wins, so existing requests to `extra.embeddings` no longer resolve to the same file as before.

Please reserve the legacy `extra.chat` / `extra.embeddings` / `extra.reranking` IDs and qualify a direct model that would collide with them, plus add a regression test for this case.

After that I’m good to approve.

fl0rianr (commented):
The direct alias collision is fixed now. I found one remaining version of the same issue with nested folder models though.

For example, `embeddings/embeddings/model.gguf` can still claim the real `extra.embeddings` ID because the directory-model path does not use the new reserved-ID protection.

Could you apply the same legacy-ID reservation there as well and add a small regression test for the nested-folder case?

After that I’m good to approve.

fl0rianr (approved):
Yes, that compatibility contract is what I had in mind. I rechecked the latest changes and the nested case is handled correctly now as well.

I also went back through the overall discovery, naming, shard/mmproj and compatibility paths and I don’t see another functional blocker.

Looks good to me now, thanks for working through all the edge cases.

fl0rianr:
ExtraModelDiscoveryTest C++ unit test is at least flaky, please fix it.

PR #3338, opened 2026-08-25
test(server): eliminate timing flakiness in StreamingHeartbeatTest

abn:
## Summary

I have stabilized `StreamingHeartbeatTest` in `test/cpp/test_streaming_heartbeat.cpp` by replacing fixed-duration wall-clock sleeps with atomic synchronization against received SSE `: ping\n\n` comments. This eliminates race conditions against libcurl's ~1000ms progress callback polling interval and avoids CI flakiness under system load.

Key updates:
- Prefill and inter-token pause tests synchronize dynamically on received heartbeats (`heartbeat_count >= 2`) with a 5s fallback deadline.
- In the inter-token pause test, the 5s deadline starts only after `chunk1_seen` is observed downstream, and heartbeats are counted strictly during the pause window.
- Deterministic TTFT assertion (`>= 2.0s`) verified under two 1s heartbeat intervals.
- Dropped the scheduler-dependent wall-clock disconnect assertion, relying strictly on the exact callback-abort telemetry error check.

## Scope

## Testing

_Testing details:_
- Ran 50 consecutive iterations of `./build/test_streaming_heartbeat` with 0 failures.
- Ran full C++ CI test battery: `ctest --test-dir build -L cpp-ci --output-on-failure` (59/59 passed).
- Verified backend boilerplate sync: `python3 docs/tools/gen_backend_boilerplate.py --check` (0 drift).

## Documentation

## Breaking Changes

## AI-assisted contribution

If AI tools were used:

### DIFF
--- a/test/cpp/test_streaming_heartbeat.cpp
+++ b/test/cpp/test_streaming_heartbeat.cpp
@@ -23,18 +23,17 @@ static void check(bool condition, const char* name) {
 
 static void test_heartbeat_emission_during_prefill() {
     httplib::Server backend;
-    std::atomic<bool> backend_started{false};
     std::atomic<bool> release_backend{false};
+    std::atomic<int> heartbeat_count{0};
 
     backend.Post("/v1/chat/completions",
         [&](const httplib::Request&, httplib::Response& res) {
-            backend_started.store(true, std::memory_order_release);
-
             res.set_chunked_content_provider(
                 "text/event-stream",
                 [&](size_t, httplib::DataSink& sink) {
-                    const auto deadline = std::chrono::steady_clock::now() + 2200ms;
-                    while (std::chrono::steady_clock::now() < deadline) {
+                    const auto deadline = std::chrono::steady_clock::now() + 5000ms;
+                    while (heartbeat_count.load(std::memory_order_acquire) < 2 &&
+                           std::chrono::steady_clock::now() < deadline) {
                         if (release_backend.load(std::memory_order_acquire)) {
                             return false;
                         }
@@ -65,7 +64,6 @@ static void test_heartbeat_emission_during_prefill() {
     backend.wait_until_ready();
 
     std::vector<std::string> received_writes;
-    int heartbeat_count = 0;
     int data_chunk_count = 0;
     int done_count = 0;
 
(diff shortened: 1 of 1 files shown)


fl0rianr (commented):
Looks good overall, this addresses the actual flake instead of just increasing the sleep.

Two small things I’d still tighten:

* In the inter-token test, count heartbeats only after `Chunk1` was actually observed downstream, so it strictly proves heartbeats during the pause.
* I’d keep the TTFT assertion at `>= 2.0`; with two 1s heartbeat intervals the new synchronization should make that deterministic.

The `< 5000ms` disconnect check is probably fine, but could also be dropped since the expected disconnect telemetry already proves the important behavior.

jeremyfowers:
Thank you for working on this! I just had a PR get dropped from merge queue due to this issue.

abn:
I have updated the PR to address the review feedback:

* **Inter-token pause counting**: In `test_heartbeat_emission_during_pause_between_tokens()`, `heartbeat_count` is now incremented only after `Chunk1` has arrived downstream (`chunk1_count > 0`), strictly verifying that heartbeats were received during the pause window.
* **TTFT assertion**: Restored the TTFT check to `>= 2.0` in `test_heartbeat_emission_during_prefill()`, which deterministically holds with the atomic synchronization on two 1s heartbeat intervals.

fl0rianr (commented):
The previous points look good now. One timing edge remains: the 5s pause deadline starts when the backend writes Chunk1, while heartbeats are only counted after Chunk1 reaches downstream. Under heavy runner descheduling, that delay can consume the timeout budget and still cause a false failure. I’d start the heartbeat deadline only after an atomic chunk1_seen is observed.

I’d also drop the < 5000ms disconnect assertion rather than relaxing it. The exact Client disconnected during stream telemetry already verifies the callback-abort path, while the wall-clock bound remains scheduler-dependent.

abn:
I have applied both refinements:

* **Pause deadline synchronization**: In `test_heartbeat_emission_during_pause_between_tokens()`, the backend now waits until `chunk1_seen` is observed before starting the 5s pause deadline, ensuring that client TCP read latency does not count against the pause heartbeat budget.
* **Disconnect assertion cleanup**: Dropped the wall-clock elapsed check from `test_heartbeat_client_disconnect_aborts_upstream()`, relying directly on the exact `telemetry_error == "Client disconnected during stream"` callback-abort check.

fl0rianr (approved):
LGTM, thanks!

Geramy (dismissed):
@abn please make the timeout a constant above or in the header.

fl0rianr:
@Geramy Please remove your Request changes for such a tiny non functional nit as we need this PR to merge ASAP to make to queue working again. You can do a follow up etc. but I would not change this one - IMO.

PR #3339, opened 2026-08-25
fix(router): filter candidates by context window at dispatch

### LINKED ISSUE #2959: [Router] Capacity-aware candidate filtering and context-overflow fallback at dispatch
## Problem

A router collection can select a candidate whose context window cannot fit the request, and nothing catches this before or after dispatch — the client gets a hard 400 mid-conversation.

There is no preflight length check anywhere on the dispatch path:

- `route_collection_request` (`src/cpp/server/server.cpp:2752`) evaluates the policy and returns `Decision.route_to`; `apply_router_collection_dispatch` (`src/cpp/server/server.cpp:2894`) rewrites `request_json["model"]` at `server.cpp:2915` and falls through to the normal completion path. No size comparison happens.
- The over-long prompt is forwarded to llama-server, which rejects it with `"request (N tokens) exceeds the available context size (M tokens), try increasing it"`. Lemonade only string-matches that error after the fact: `is_context_window_error` (`src/cpp/server/wrapped_server.cpp:91-95`) and maps it to `error.code = "context_length_exceeded"` (`wrapped_server.cpp:144-147`). The 400 is returned to the client; there is no re-route, retry, or fallback.
- Capacity metadata exists but is never consulted for routing: per-model `ModelInfo::max_context_window` (`src/cpp/include/lemon/model_manager.h:89`) is populated from GGUF metadata (`src/cpp/server/backends/llamacpp/llamacpp_server.cpp:837`), cloud discovery (`src/cpp/server/backends/cloud/cloud_server.cpp:861`), and FLM (`src/cpp/server/backends/fastflowlm/fastflowlm_server.cpp:457`). The runtime window is resolved at load by `resolve_auto_ctx_size` (`src/cpp/include/lemon/auto_tune.h:307`, applied at `src/cpp/server/router.cpp:757-762`) but is not visible to dispatch afterward.

Why this matters for router collections specifically: local candidates commonly resolve to 4k–32k windows (`AUTO_CTX_FALLBACK = 4096`, `AUTO_CTX_UNKNOWN_MAX = 32768`, `auto_tune.h:18,22`) while cloud candidates advertise 128k+. A conversation that grew during cloud turns hard-fails the moment a short follow-up ("go on") matches a rule that routes to a small local candidate — the routing input is the last user message, so rule matching cannot see the history size at all.

## Proposed approach

Capacity-aware candidate filtering at dispatch, with a bounded inference-time backstop. No policy-schema change; behavior is additive and fails open when capacity is unknown.

1. **Preflight estimate.** Before applying the decision, estimate the prompt footprint: `est_prompt = ceil(utf8_bytes(serialized messages + system + tools) / 4)`. Add generation headroom: requested `max_tokens`/`max_completion_tokens` when present, else a `DEFAULT_GENERATION_HEADROOM` constant (proposed: 1024). A candidate fits when `SAFETY_MARGIN * est_prompt + headroom <= effective_window`, with `SAFETY_MARGIN = 1.25` (chars/4 undercounts for code, non-Latin scripts, and chat-template overhead; the margin must err toward skipping).
2. **Effective window per candidate.** If the candidate is loaded, use its resolved runtime `ctx_size` (surface it from `Router` — see sketch). If not loaded, use the same estimate the load path would produce: `compute_auto_context_size` (`auto_tune.h:195`) clamped by registry `max_context_window` when `> 0`. If no capacity information exists (e.g., cloud provider reported nothing, window is `0`), treat the candidate as unconstrained — never skip on unknown.
3. **Fall-through on predicted overflow.** If the selected candidate does not fit, re-run resolution with that candidate excluded: extend `RoutingPolicyEngine::route` with an optional `excluded_candidates` set (engine API only, not policy schema). First-match-wins then lands on the next matching rule whose candidate fits, else the policy default, else the no-fit error below.
4. **Trace visibility.** Record each capacity skip as an additive `trace[]` entry using the existing `TraceEntry` shape: `{condition: "capacity:<model>", result: false, rationale: "estimated <N> tokens (+<H> headroom) > window <W>"}`. Existing conformance-corpus cases carry no capacity metadata and never trigger the filter, so recorded v1 `policy → Decision` goldens are unchanged. Validate emitted decisions against `decision.schema.json` (trace `condition` is a free string; confirm no schema tightening is needed).
5. **No candidate fits.** Return 400 with `error.code = "context_length_exceeded"` (same shape clients already handle from `wrapped_server.cpp:144-147`) and, when `route_trace` was requested, the trace listing every skipped candidate.
6. **Inference-time backstop.** The estimate can be wrong. If a routed local candidate still returns a context-window error at inference (`is_context_window_error`), re-enter dispatch exactly once with that candidate added to the exclusion set, then forward to the new selection. Bound: one retry per request, router-collection requests only; direct (non-collection) requests keep today's behavior.

## Implementation sketch

1. `Router`: expose resolved runtime `ctx_size` for loaded servers (extend the loaded-model info alongside `max_context_window`, `src/cpp/server/router.cpp:1039-1041`), so dispatch can read live windows.
2. New helper (proposed `src/cpp/server/routing_capacity.cpp` + header): `estimate_request_tokens(const json& request)` and `effective_context_window(const ModelInfo&, const Router&)`, both pure and unit-testable.
3. `RoutingPolicyEngine::route(ctx, want_trace, const std::set<std::string>* excluded = nullptr)`: rules whose `route_to` is excluded are treated as non-matching; default falls through the same way. Append capacity trace entries via the caller (the engine stays capacity-agnostic; dispatch owns the estimate).
4. `Server::route_collection_request` / `apply_router_collection_dispatch` (`server.cpp:2752`, `2894`): loop — decide, check fit, exclude, repeat; emit trace entries; produce the no-fit 400.
5. Backstop: at the two call sites that forward router-dispatched requests (chat completions at `server.cpp:3220` region and responses at `server.cpp:4826` region), catch `context_length_exceeded` backend errors and re-dispatch once with the exclusion set.
6. Optional refinement (follow-up-friendly, keep out of first PR if it grows): exact token count via the loaded candidate's `/tokenize` passthrough (`LlamaCppServer::tokenize`, `src/cpp/server/backends/llamacpp/llamacpp_server.cpp:659-660`) instead of the chars/4 estimate when the candidate is already resident.

## Acceptance criteria

- `test/server_router.py` e2e:
  - A collection with a small-window local candidate and a larger/cloud-mock default: a request whose history exceeds the small candidate's window routes past it to the default; `x_lemonade_route.trace` (with `route_trace: true`) contains the `capacity:<model>` skip entry; a short request still routes to the small candidate.
  - No-candidate-fits: 400 with `error.code == "context_length_exceeded"`; skipped candidates visible in trace.
  - Backstop: candidate whose live window is smaller than the estimate predicted (force via small `ctx_size` load option) triggers exactly one re-route, and the response carries the final model.
- `test/cpp` unit tests: estimator (UTF-8 byte counting over messages/system/tools, headroom, margin), `effective_context_window` (loaded vs unloaded vs unknown=0 → unconstrained), engine exclusion-set resolution (first-match with exclusions, default fall-through).
- Routing conformance corpus (`test/conformance/routing/1/**`) passes unmodified — no Decision drift for existing cases.
- No behavior change for non-collection requests.

## Out of scope

- Compaction/summarization of history to fit a candidate (separate research issue).
- Token-exact length conditions in the policy schema (`min_tokens`/`max_tokens` — deferred per `routing_policy.h`; this issue is dispatch-side only).
- Changing #2714's aggregate collection metadata (sibling: #2714 computes the collection-level minimum for `/v1/models` display; this issue reads per-candidate windows at dispatch and works whether or not #2714 has landed).
- Cloud-side truncation parameters (e.g. forwarding provider `truncation` options).

## References

- #2714 — aggregate `max_context_window` for `collection.router`/`collection.omni` in `/v1/models` (sibling metadata fix; not a prerequisite).
- LiteLLM `context_window_fallbacks` — prior art for overflow fallback in a model router: https://docs.litellm.ai/docs/proxy/reliability
- llama-server rejection string (verified against the vendored b9747 binary): `request (%d tokens) exceeds the available context size (%d tokens), try increasing it`, error type `exceed_context_size_error`.

Size: M

ao-batouche:
## Summary

Router collections could select a candidate whose context window cannot fit the request, and the client got a hard 400 mid-conversation. Dispatch now estimates the request's token footprint, skips candidates that cannot fit, and falls through to the next matching rule or the policy default. Fixes #2959.

## Scope
- New routing_capacity helper: prompt-token estimator, effective-window resolution (live ctx_size when loaded, else what a load would compute), and context-overflow error detection. Pure and unit-tested.
- RoutingPolicyEngine::route() takes an optional exclusion set (engine API only, not policy schema); excluded rules count as non-matching.
- Router::get_loaded_ctx_size() exposes a live model's resolved window to dispatch.
- No candidate fits → 400 with error.code = "context_length_exceeded" and the skipped candidates in the trace.
- Inference-time backstop: a routed candidate that still rejects for length triggers exactly one re-route. Non-streaming only (on the streaming path headers are already sent, and the preflight filter covers those requests).
- Unknown window (e.g. a cloud provider reporting none) is treated as unconstrained; a candidate is never skipped on missing data.
- No behavior change for non-collection requests, and none for existing policies until a window is actually exceeded.

## Testing

- [X] Code builds without errors locally.
- [X] I tested this change locally.

_Testing details:_
Builds clean (Ninja/GCC, Linux). ctest -L cpp-ci: 58/58 pass, including the new RoutingCapacityTest and four new engine exclusion checks; the routing conformance corpus passes unmodified. test/server_router.py against a live server: 15/15 pass (1 pre-existing skip needing onnxruntime), with three new e2e tests covering the capacity skip, the no-fit 400, and the one-shot backstop. 
## Documentation

- [X] Documentation is affected and has been updated.

Added a "Context-window fitting" section to docs/dev/router-policy.md.

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:


SlawomirNowaczyk (changes requested):
Looks solid; here are a few improvement suggestions:

## 1. Streaming requests get no inference-time backstop — undercounts fail hard mid-stream

The whole justification for the runtime backstop is that the preflight estimate "can be wrong in the other direction." Yet `retry_dispatch_on_context_overflow` is wired only into `handle_chat_completions`/`handle_responses` non-streaming paths. The Scope section admits streaming is excluded "because headers are already sent."

That means the exact failure mode the PR set out to fix (#2959: "client got a hard 400 mid-conversation") still happens for streaming clients whenever the estimator undercounts — which is common for code and non-Latin scripts, the very cases the `SAFETY_MARGIN` comment calls out.

Suggestion: detect overflow before the first SSE event is emitted. Most backends reject on the initial request, before any token is produced, so you can buffer/peek the first upstream chunk and, if it is a `context_length_exceeded` error, run the same one-shot re-route before writing status/headers. That closes the gap without needing to "un-send" headers.

## 2. `is_context_overflow_error` substring-matches the entire serialized error → spurious re-routes

```cpp
const std::string text = lower_copy(error.dump());
return text.find("exceeds the available context size") != ... ||
       text.find("context length exceeded") != ... ||
       text.find("context_length_exceeded") != ...;
```

Lower-casing and searching the whole `error.dump()` means any error whose message *echoes user input or a model id* containing those phrases triggers a re-route. A user prompt like "why did I get context_length_exceeded?" reflected back in a validation error would spuriously evict a working candidate.

Suggestion: scope the text search to the fields that actually carry backend rejections — `error.message` and the nested `error.details.response.error.message` — instead of the entire dumped object. The normalized-`code` check is already precise; only the raw-text fallback needs tightening.

## 3. `compute_auto_context_size` + `get_available_memory_gb` runs per-candidate on the request hot path

In `candidate_context_window`, for any unloaded llama.cpp candidate you call `get_available_memory_gb(info.device)` and `compute_auto_context_size(...)` on every dispatch. In a collection with several unloaded local candidates, that's a system memory probe + auto-tune computation per candidate, per request, inside the resolution `while` loop.

Suggestion: memoize the resolved unloaded window keyed by model (invalidated on load/unload), or at least hoist the `get_available_memory_gb` call out so it's computed at most once per dispatch rather than once per skipped candidate.

## 4. The estimator (bytes/4 of the JSON dump × 1.25) can silently push traffic to cloud

`estimate_prompt_tokens` sums `dump().size()` of `messages`/`system`/`tools`/`prompt`/`input` — so JSON structural bytes (keys, quotes, braces, escapes) are counted as prompt content, then multiplied again by `SAFETY_MARGIN = 1.25`. That compounds two conservative fudge factors and can *over*-skip a perfectly capable local candidate, falling through to a cloud default. That's a cost/latency regression that's invisible unless `route_trace` is on.

Suggestion: estimate from message *content* length rather than the full serialized envelope, and expose `SAFETY_MARGIN` / `DEFAULT_GENERATION_HEADROOM` as policy-level knobs (they're currently hard-coded `constexpr`). Operators tuning a collection that mixes a small local model with an expensive cloud fallback need to control that boundary.

## 5. The one-shot backstop can't recover when the *replacement* is also unconstrained-and-too-small

`retry_dispatch_on_context_overflow` re-routes exactly once, excluding only the failed candidate. But the reason a candidate reaches inference despite being too small is that it reported an unknown window (treated as unconstrained). If the re-routed replacement is *also* a no-window candidate that's too small (e.g. a second cloud provider), it overflows again — and now there's no further fallback, so the client gets the hard failure the PR was meant to eliminate.

Suggestion: allow the backstop to iterate up to the number of remaining candidates (accumulating the exclusion set) instead of capping at one, or record that a candidate rejected for length so subsequent dispatches treat its window as bounded. At minimum, document this residual failure mode alongside the "re-routes once" note.

ao-batouche:
Thank you @SlawomirNowaczyk  for the review and comments, I've now addressed those. Two things worth your attention are flagged at the end.

1. Streaming backstop. Done as suggested. A backend length rejection arrives as a single SSE event with nothing written before it, so a new ContextOverflowProbeSink withholds an attempt's bytes until the first complete event is known: an overflow event is swallowed and the request re-routed, and only the surviving attempt reaches the client. Both outcomes are 200/text/event-stream, so no status or header needs revising. The three streaming call sites now share one provider (the old free-function helper is gone).

Two things surfaced getting this green:

The withheld attempt's done() was closing the client sink before the retry could write. The probe now swallows done() when it has withheld.
The cloud streaming path discarded the provider's error body, so a length rejection was indistinguishable from any other 400 (the backstop could never fire for a cloud candidate). forward_streaming_request now nests the provider body under details.response, matching what non-streaming post_with_auth already does. clients were losing that information too.
2. Scoped overflow detection. No longer searches error.dump(). It reads only error.code/error.message and the nested error.details.response.error.{code,message}. I also dropped the bare context_length_exceeded token from the text fallback, the code check covers it precisely, and that token was the likeliest thing to appear in echoed input. Your exact scenario is now a unit test, alongside four other spurious-match cases.

3. Memory probe. New per-dispatch CandidateWindowCache memoizes resolved windows per model and get_available_memory_gb per device. The probe and auto-tune now also only run when a candidate is genuinely unloaded and unpinned, so both are skipped entirely in the common cases.

4. Estimator + knobs. The estimator counts message content text plus a per-message template allowance instead of the serialized envelope, so structural bytes are no longer compounded with the margin. Image parts are excluded entirely, a base64 image previously inflated the estimate enormously. tool schemas still count from their JSON, since that is how they reach the model.

safety_margin and generation_headroom are now policy knobs under routing.capacity, validated rather than clamped (a margin below 1.0, or a negative headroom, is rejected loudly). Schema, parser, and the key-parity test are updated.

5. Backstop iterates. Both backstops now re-route past each rejecting candidate, bounded by the collection's candidate count.

I did not add the cross-request "this candidate rejected for length" memory. Bounding a window from an observed rejection needs an observed token count, and it goes stale as soon as the model is reloaded with a different ctx_size (it seemed to want its own change rather than being folded in here). The residual (every candidate rejects -> the last rejection is surfaced unchanged) is documented.

Two to flag:

This adds policy-schema keys, which the issue explicitly ruled out ("No policy-schema change; behavior is additive"). Your #4 asked for the knobs, so I went with the review, happy to revert them to constants if you'd rather keep the v1 schema frozen.
x-lemonade-route on a re-routed stream is wrong, and I could not fix it: it is a header, committed before streaming begins, so it still names the initially matched rule. The first SSE event's x_lemonade_route is authoritative for which candidate answered. Noted in docs/dev/router-policy.md.

Testing. C++ 58/58 (46 checks in test_routing_capacity, 11 new in the parser test). Router e2e 17/17, with new tests for the streaming backstop and the knobs, the server log confirms real re-routes rather than incidental passes. LLM suite 31/31, run specifically because the SSE provider refactor touches all streaming, not just router traffic. Endpoints 53/54, the one failure being test_020_web_app_root (web app not built in my environment), identical before and after.

SlawomirNowaczyk (changes requested):
All previous issues are fixed, thanks.
However, I see some new problems to look into before merge:

**1. The probe leaks error bytes if the producer ignores a `false` write return (robustness/info-leak).**
In `ContextOverflowProbeSink::write`, once overflow is decided:
```cpp
if (decided_) { return inner_.write(data, length); }
```
After an overflow it sets `decided_ = true` and returns `false` to stop the relay — but if the streaming source keeps writing (doesn't honor the `false`), every subsequent chunk now takes the `decided_` fast-path and is forwarded straight to the client, defeating the "withhold the rejection" guarantee. Post-overflow the sink should keep swallowing, not pass through. Worth confirming `StreamingProxy`/`*_stream` actually abort on `false`; if any path doesn't, the rejection detail reaches the client.

**2. The whole streaming design rests on "the rejection is the first SSE event with nothing before it."**
If any backend emits an initial event (e.g. a role/`delta` opener) before its length rejection, the probe releases that first event, sets `decided_`, and the following error passes through mid-stream — exactly issue #2959 for streaming. The author documents the assumption, but it should be validated against each real streaming source (cloud providers, `llama-server`, whisper) rather than assumed. This is the single biggest correctness risk in the commit.

**3. Work amplification on pathological policies (latency / model thrashing).**
Both backstops loop up to `candidate_count`, and each re-route calls `auto_load_model_if_needed`. For a collection of N *local* candidates that all reject, one request can trigger N sequential upstream calls **and** N evict+load cycles while the client blocks with no output. Consider a tighter bound, or not counting reload-heavy retries the same as cloud retries.

**4. Responses API structured `input` is under-counted to ~0.**
`text_bytes` extracts only a top-level `text` key from array elements. Responses `input` items are message objects (`{"role","content":[{"type":"input_text","text":...}]}`), so their nested content isn't summed — the estimate collapses toward the per-message overhead. Preflight won't skip these; it silently relies on the inference-time backstop... but the estimator claims to handle `input` and doesn't for the array form.

ao-batouche:
Thank you @SlawomirNowaczyk , all four are fixed: 

1. Probe leak. Post-overflow the sink now swallows everything instead of passing through, so a producer that ignores the false return can't push the rejection to the client. I also traced the paths you flagged: the StreamingProxy relay, the cloud SSE relay, and both chat_completion_stream / responses_stream wrappers do all honor false, and StreamingProxy writes nothing after the error event. The sink no longer relies on that being true.

2. The "first event" assumption. Validated rather than assumed, and you were right to push on it. StreamingProxy initializes backend_status = 200, so a : ping keep-alive can precede a rejection, under the old code that settled the decision and the error then leaked mid-stream. Comments are now forwarded immediately (so a long prefill can't time the client out) without spending the re-route opportunity.

Capturing llama-server's real rejection also turned up a separate bug: its payload carries a numeric code: 400, and json::value("code", std::string()) throws on a type mismatch. So a local length rejection in a router collection was becoming a 500 with no backstop at all. Now type-guarded, with the real payload pinned as a test.

On the sources: llama-server is verified empirically (HTTP 200, rejection is its first and only event). Cloud is verified from the code path plus a mock provider, not a capture from a real one, and a provider that answers 200 then reports overflow in-band after content falls into the mid-stream case, which is forwarded rather than re-routed since it can't be un-sent. Whisper can't reach the backstop; it's only wired into chat/completions, completions and responses.

3. Work amplification. Did both things you suggested: a hard cap of 3 re-routes (and never more than candidates - 1), plus reload-heavy retries counted separately with a budget of 1. An already-resident or unmetered (cloud) replacement is cheap and only counts against the first cap. Affordability is checked before touching the router, so a refused retry costs nothing. Worst case is now 4 upstream calls and 1 load, regardless of how many candidates a policy lists.

4. Responses under-count. text_bytes now recurses through nested content, so {"role","content":[{"type":"input_text","text":...}]} is summed properly; images still contribute nothing, and Responses items get the same per-message overhead as chat messages. Worth noting the asymmetry behind your point: build_route_context already read that shape correctly, so routing saw the text while the estimator didn't.

Testing. The probe is extracted to its own header so #1 and #2 are directly unit-tested (39 checks: withholding, post-overflow swallowing, keep-alives, split events, CRLF framing, mid-stream pass-through). C++ 59/59. Router e2e 19/19, including two new tests that drive the backstop against a real llama-server instead of a mock, 1800 chars of hex tokenizes to ~1600 real tokens but estimates at 450, so the preflight passes and the backend then rejects.

ao-batouche:
Force-pushed : my mistake on the previous push: it accidentally carried five
unrelated frontend files under this commit message, and left the actual fixes
out. Both sorted now. The frontend files are off this branch (moved to where
they belong), and the four review fixes are in `f23ad4eb`.

One wording change while I was in there: the streaming section of
`docs/dev/router-policy.md` now separates what I verified empirically
(llama-server) from what I established from the code path plus a mock provider
(the cloud path), instead of claiming both equally.

Green on the final tree: C++ 59/59, router e2e 19/19, endpoints 141/141.

SlawomirNowaczyk (approved):
LGTM now

ramkrishna2910 (changes requested):
Reviewed this specifically for blast radius outside the router, given the size of the diff.

**The containment is good.** Of ~1,090 production lines (the other ~1,360 are tests, docs and CMake), the router-owned files account for roughly 600. The `server.cpp` additions are properly gated — both `retry_dispatch_on_context_overflow` and `stream_with_context_overflow_backstop` early-out on `!route_dispatch.has_value()`, and `set_router_backstop_sse_content_provider` replaces `set_route_decision_sse_content_provider`, which already took `route_dispatch`. A non-router request pays one `has_value()` check and behaves exactly as before. `error_types.h` and `router.cpp` are purely additive.

**One change before this lands.** `cloud_server.cpp` is the only edit that changes behavior for traffic which never touches the router:

```cpp
json extra = {
    {"status_code", result.status_code},
    {"details", {{"status_code", result.status_code},
                 {"response", provider_body}}},
};
```

Every streaming cloud request that gets a non-200 now carries the provider's raw error body in the SSE payload, router or not. That's a payload-shape change for clients using a cloud model directly, and it forwards whatever the upstream returned — quota details, account hints, internal messages — verbatim. The non-streaming path does this already, so the direction is defensible, but it shouldn't widen as a side effect of a routing change. Please gate the `details` block on the backstop actually needing it, so a plain cloud stream keeps its current error shape.

**Tests, docs and CMake all check out.** The two new targets are `BUILD_TESTING`-guarded and correctly registered `CI ON`; the doc sections cover the `capacity_settings` knobs the schema addition introduces; and `start_mock_cloud_provider` gains a `context_limit` parameter rather than being duplicated. Nothing there is surplus. One thing to confirm once checks run: `test_routing_capacity` links only `nlohmann_json` while `routing_capacity.h` pulls in `model_manager.h`.

Two blockers that are independent of the above: no CI has run on this branch at all yet, and `schema-lock.json` was not refreshed for the new `capacity_settings` block — I ran `test/test_schema_lock.py` at head and it fails. `python test/test_schema_lock.py --update` fixes it.

ao-batouche:
Thanks @ramkrishna2910, the cloud change is gated, and both blockers are cleared.

**cloud_server.cpp.** The `details` block is now attached only when the provider's
body is a context-length rejection:

```cpp
json extra = {{"status_code", result.status_code}};
if (routing_capacity::is_context_overflow_error(provider_body)) {
    extra["details"] = {{"status_code", result.status_code},
                        {"response", provider_body}};
}
```

A plain cloud stream that 429s, 401s or 500s keeps exactly the shape it had before
this branch, no provider text relayed. Only an overflow carries the body, which is
the single case the backstop reads it for.

I gated on the error rather than plumbing a "router dispatch" flag down to the
backend: that flag would have to cross `WrappedServer::forward_streaming_request`
and every override, widening the diff in the direction you're asking me to narrow.

`schema-lock.json`. Refreshed with `python test/test_schema_lock.py --update`;
The test passes at head.

`CI`. It has run since your review and found more than the lock. Four jobs
failed, all now fixed:

Routing schema + fixtures, the lock above.
pre-commit, the comment-slop hook, 6 issues. Each repeated concept is now explained once where the surprising thing happens, with the other sites pointing at it; the probe-sink header and the probe test's header are inside the 12-line limit.
Test CLI/Endpoints (ubuntu + windows), `test_654`, a bug in my test rather than in the code. `safety_margin`: 500.0 inflates the 11-token estimate to 5508, which also exceeds the default candidate's 4096 window, so every candidate is skipped and dispatch correctly returns 400 instead of falling back. Now 200.0 → 2208: past the matched candidate's 2048, inside the default's.
`test_routing_capacity` linkage. It holds: `compute_auto_context_size` is
inline in `auto_tune.h` and ModelInfo is only read for its data members, so
`nlohmann_json` is the only library symbol the test needs. The C++ unit tests job
passed at 5m17s on the last run.

Verified locally: the schema-lock test, comment-slop (0 issues), black, and
`-fsyntax-only `on `cloud_server.cpp` and both C++ tests. `test_654` needs a live
server, so that one will prove out in the next CI run.

PR #3344, opened 2026-08-25
fix(vllm): scale gpu-memory-utilization to free memory on shared-memory devices

jeremyfowers:
## Summary

vLLM refuses to start unless `gpu_memory_utilization` of the **total** memory pool is free. The `--kv-cache-memory-bytes 4G` cap that shared-memory devices already get bounds what vLLM *uses*, but not what it *demands be free*, so the default 0.92 still applies at startup. On a 32 GiB gfx1151 any co-tenant process holding more than ~2.5 GiB rejects a 4B model that fits with room to spare.

This scales `--gpu-memory-utilization` to the free share of **the device vLLM will actually run on**, read from the amdgpu driver's own per-device accounting. An idle device keeps vLLM's own default; an explicit user `--gpu-memory-utilization` and discrete-HBM parts are untouched.

This is the intermittent `Validate vllm (rocm)` failure in the merge queue. Three runs, each with the identical reading `Free memory on device cuda:0 (27.07/32.0 GiB) on startup is less than desired GPU memory utilization (0.92, 29.44 GiB)` in its `server-logs-rocm` artifact:

- [Aug 25 14:35](https://github.com/lemonade-sdk/lemonade/actions/runs/32860463383/job/97845930366)
- [Aug 25 05:33](https://github.com/lemonade-sdk/lemonade/actions/runs/32813275201/job/97698221197)
- [Aug 25 01:29](https://github.com/lemonade-sdk/lemonade/actions/runs/32797764433/job/97653926622)

That reading now yields 0.79 — 25.28 GiB requested against 27.07 GiB free — which loads.

### Why the reading is per-device

`SystemInfo::get_global_vram_usage_pct()` is the eviction engine's machine-wide pressure signal. It folds every GPU in the box into one worst-case ratio, and its APU branch falls back to the dedicated VRAM carve-out whenever its libdrm probe cannot open a render node — reachable in normal deployments, since `lemond.service` runs as `User=lemonade` and the postinst only adds that user to `render` when the group already exists. Either path can describe a device vLLM never touches, which would clamp the request low enough to fail loads that succeed today.

`SystemInfo::get_rocm_device_memory(arch, ...)` instead reads the single GPU matching the ISA the launch policy was derived from, through world-readable KFD topology and sysfs — no render-node access required. It reports nothing unless exactly one GPU matches, so an ambiguous or unreadable topology leaves vLLM on its own default rather than on another device's numbers.

The emitted value only steps in where vLLM's default would actually be rejected, truncates rather than rounds so it never asks for more than is free, and is withheld entirely when the resulting budget could not hold the 4 GiB cache cap emitted beside it — vLLM's own pre-flight names the byte counts in that case, which is a better diagnostic than a request known in advance to fail.

## Scope

- [ ] I kept refactoring separate unless it is required for this change.

_Scope note:_ the per-device probe (`system_info_utils.h`, `system_info.h`, `test_system_info_helpers.cpp`) is new surface this fix requires — the codebase had no per-device memory reading. Two smaller items are judgment calls, easy to drop on request: `identify_rocm_arch_from_name()` now shares the packed-decimal gfx conversion with the probe instead of duplicating it (behavior preserved, including `"0"` → `"gfx000"`), and the CMake change below.

## Testing

_Testing details:_ `cmake --build --preset windows --target cpp-ci-tests lemond`, then `ctest --test-dir build -C Release -L cpp-ci` — 57/57 pass.

- 13 cases in `test_vllm_arg_resolver` pin literal values rather than restating the formula: the observed CI reading, the truncation boundary (0.7959 must emit 0.79, never 0.80), vLLM's exact 0.92 threshold, an exhausted device, a nonsensical reading, and the kv-cap floor on both a small and a large pool.
- 20 cases in `test_system_info_helpers` drive the sysfs probe against a fixture topology: a single match, an exhausted second GPU that must not influence the result, duplicate ISAs, a GPU with no render node, and missing / zero / inconsistent readings.

`add_cpp_ci_test(vllm_arg_resolver ...)` moves from `CI OFF` to `CI ON`. The target compiles only `vllm_arg_resolver.cpp` plus nlohmann and needs no GPU or backend; without it these regression tests never execute in CI, leaving the flake this PR targets unguarded. The `OFF` appears to date from the bulk helper migration in #2877 rather than a deliberate exclusion.

Removing the `#ifdef __linux__` in favour of a path-parameterized probe means the logic now compiles and is unit-tested on every platform, not only on Linux builds.

_Not verified locally:_ that sysfs `mem_info_vram_total` equals the total vLLM reports on gfx1151 — that needs the runner, so `Validate vllm (rocm)` is the real check. A mismatch would make the sysfs total the smaller of the two, which lowers the emitted request further below vLLM's default; it can therefore leave the flake unfixed but cannot introduce a new failure.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1166,7 +1166,7 @@ if(BUILD_TESTING AND EXISTS "${VLLM_ARG_RESOLVER_TEST_SOURCE}")
         src/cpp/server/backends/vllm/vllm_arg_resolver.cpp
     )
     target_link_libraries(test_vllm_arg_resolver PRIVATE nlohmann_json::nlohmann_json)
-    add_cpp_ci_test(vllm_arg_resolver CI OFF COMMAND test_vllm_arg_resolver)
+    add_cpp_ci_test(vllm_arg_resolver CI ON COMMAND test_vllm_arg_resolver)
 endif()
 
 set(LLAMACPP_REQUEST_TEST_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_llamacpp_request.cpp")
--- a/src/cpp/include/lemon/backends/vllm/vllm_arg_resolver.h
+++ b/src/cpp/include/lemon/backends/vllm/vllm_arg_resolver.h
@@ -43,5 +43,10 @@ DeviceClassLaunchPolicy device_class_launch_policy(const std::string& arch,
                                                    bool has_memory_budget_arg,
                                                    bool has_enforce_eager = false);
 
+// The --gpu-memory-utilization to emit alongside the shared-memory kv-cache cap, or a
+// negative value to leave vLLM's own default in place. `global_vram_usage_pct` is
+// used/total in [0, 1] (SystemInfo::get_global_vram_usage_pct), or negative if unknown.
+double shared_memory_gpu_utilization(double global_vram_usage_pct);
+
 } // namespace backends
 } // namespace lemon
--- a/src/cpp/include/lemon/backends/vllm/vllm_arg_resolver.h
+++ b/src/cpp/include/lemon/backends/vllm/vllm_arg_resolver.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstdint>
 #include <nlohmann/json.hpp>
 #include <string>
 #include <vector>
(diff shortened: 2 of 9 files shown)


sree-ram-11:
Tested on real `gfx1151` hardware (Strix Halo, 64 GiB unified VRAM) at Krishna's latest commit.

**Unit tests** — both suites pass on the box:

- `ctest -R "vllm_arg_resolver|SystemInfoHelpers"` → 100% passed, 0 failures. 
- **Also confirmed the per-device probe targets the right node:** node 1 = gfx1151 (`gfx_target_version 110501`, `drm_render_minor 128`), and vLLM's own `Initial free memory` reading matched the `mem_info_vram` pool the probe reads.

**Idle machine (no-regression)** — with ~99% free, the scaler correctly stayed out: no `gpu_memory_utilization` in vLLM's args, just the 4 GiB kv-cache cap. `Qwen3.5-4B-FP16-vLLM` loaded and served `OK`.

**Under memory pressure (the actual fix)**  
- Held ~10 GiB of VRAM with a co-tenant process, dropping free to 53.85 GiB (~83%), below vLLM's 0.92 default demand.
- vLLM launched with `gpu_memory_utilization=0.78` (≈49.9 GiB demanded, fits), and the model loaded and served `OK`.
- The 0.92 default would have demanded ~58.9 GiB free and been rejected, so this is exactly the failing condition the PR targets, now loading cleanly.

LGTM.

sree-ram-11 (approved):

PR #3345, opened 2026-08-25
ci(vllm): keep the HF model cache outside the workspace

jeremyfowers:
## Summary

The vLLM validate job re-downloads the 9.3 GB model on every run. Its `Run validation` step sets `HOME` to `${{ github.workspace }}/.home`, so the HF cache lands inside the workspace, which `actions/checkout` with `clean: true` wipes on the next run. The logs show a cold start every time: `Cache built: 125 total, 0 downloaded`, then a 2-minute weight download before the load is even attempted.

This points `HF_HOME` at the persistent location the release workflow already uses (`cpp_server_build_test_release.yml`, added for #2103). The vLLM backend tarball is deliberately left uncached: this workflow exists to validate backend releases, so re-fetching it each run is the point.

## Scope

## Testing

_Testing details:_ Confirmed lemond honors `HF_HOME` by running the same binary with the same lemonade cache dir against two `HF_HOME` values: an empty dir reports `Cache built: 136 total, 0 downloaded`, the populated one reports `136 total, 32 downloaded`. `path_utils.cpp:239` resolves the hub cache as `$HF_HOME/hub`. The workflow YAML parses and passes the repo's workflow linters.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/.github/workflows/validate_vllm.yml
+++ b/.github/workflows/validate_vllm.yml
@@ -140,6 +140,16 @@ jobs:
           python-version: '3.10'
           requirements-file: 'test/requirements.txt'
 
+      - name: Set HF_HOME environment variable
+        run: |
+          # Persistent HF cache outside the workspace so model weights survive across
+          # jobs. The checkout clean wipes untracked workspace files (#2103), and the
+          # step below points HOME into the workspace, so the default cache location
+          # would be re-downloaded on every run.
+          HF_HOME="$HOME/.cache/lemonade-ci/huggingface"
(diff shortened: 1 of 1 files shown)


sree-ram-11:
LGTM.

The fix is correct. The new `Set HF_HOME` step runs *before* the `HOME` override, so `HF_HOME` resolves against the real persistent home and lands at an absolute path outside the workspace.

sree-ram-11 (approved):

PR #3346, opened 2026-08-25
server: make registered model discovery canonical and queryable

fl0rianr:
## Summary

Adds canonical server-side model discovery that can be reused by API consumers such as MCP without duplicating model filtering or capability heuristics.

This PR:

- adds `ModelManager::query_models()` as the canonical query path for models registered and supported by the running server
- supports case-insensitive token search across model IDs, checkpoints, recipes, and labels
- adds `capability`, `downloaded`, `limit`, and `cursor` filtering
- extends the existing `/v1/models` endpoint instead of introducing a parallel discovery API
- preserves the existing downloaded-only default and `show_all=true` behavior
- exposes canonical `capability` / `capabilities` metadata on model records
- derives primary capabilities from model type / collection type and additional capabilities from recognized per-model labels
- maps the per-model `edit` label to `image-edit`, rather than assuming every model on an image backend supports editing
- persists user-defined labels so registered-model discovery remains consistent
- rejects unknown capability predicates with `400 Bad Request`
- adds platform-independent capability regression coverage and endpoint coverage
- updates the OpenAI API documentation

**Why:**
Model discovery should be a server-owned contract. MCP and other clients (e.g. GUI3 with currently it's own logic) should be able to ask the server which supported registered models satisfy a capability without reimplementing model-name, backend, or GUI heuristics.

Capability is deliberately model-specific where necessary. For example, an image model has primary capability `image`; only a model carrying the `edit` label gains the additional `image-edit` capability. This avoids advertising generation-only image models as edit-capable simply because they share a backend with models that support editing.

Unknown capability names are rejected rather than returning an empty successful result, so callers can distinguish an invalid predicate from a valid query with zero matching models.

Canonical discovery uses the server's supported model set. It therefore describes models registered and usable on this running server, not the raw unfiltered global registry.

This is intentionally server-side foundation only. MCP-specific tool schemas, MCP response formatting, and loaded-model state remain outside this PR and can consume this canonical query contract in follow-up work.

## Scope

## Testing

_Testing details:_

- `ModelDiscoveryCapabilitiesTest`
  - verifies an image model remains `image` without an edit capability
  - verifies the per-model `edit` label normalizes to `image-edit`
  - verifies `image` remains the primary capability for editable image models
- `test_000c_registered_model_query_contract`
  - case-insensitive multi-token search
  - label-based discovery
  - capability filtering
  - downloaded-state filtering
  - pagination and cursors
  - virtual `collection.router` discovery
  - invalid cursor handling
  - unknown capability rejection with `400 Bad Request`
  - legacy downloaded-only `/v1/models` behavior

## Documentation

The `/v1/models` documentation now describes the canonical filters, pagination fields, capability metadata, unknown-capability errors, and the distinction between the supported server registry and the raw unfiltered registry.

## Breaking Changes

Existing `/v1/models` behavior remains compatible: the default still returns downloaded models only and `show_all=true` remains supported. Canonical filtering and pagination are opt-in through the new query parameters.

## AI-assisted contribution

Please select one:

If AI tools were used:


jeremyfowers (changes requested):
I want the opportunity to review this PR before it merges. I may not have bandwidth for a while, so thanks in advance for your patience.

@fl0rianr in the meantime, please update the PR description for _why_ Lemonade's users need this change. Right now the PR description only states _what_ this change is. **What are the user stories that are unlocked once this change is merged?**

PR #3347, opened 2026-08-25
feat(server): serve bundled API docs at /docs

### LINKED ISSUE #1700: API documentation included with lemonade server

https://lemonade-server.ai/docs/server/server_spec/

### What's Missing or Wrong

Following on to #1616 it would be nice to have a local copy of the API documentation, especially the lemonade-specific parts. I see two main benefits:
1. the API documentation would inherently match the running server version. Maybe there's a new endpoint, or one was removed in the development/latest version. That doesn't matter as much as long as the server can describe the API it supports.
2. improved offline support. It seems like everybody has agreed on OpenAI-compatible endpoints, with a little bit of Ollama-compatible pixie dust, and there are bindings for every language. We don't really care about which REST methods are available for the standard operations .. but the lemonade-specific details are definitely worth documenting.

### Suggested Improvement

Include a copy of the API documentation for the running server version, and make it accessible over the network. This could be an OpenAPI spec, a pretty swagger endpoint, a `/docs` path that just returns a markdown file, some server-side magic to automatically return HTML or JSON depending on the value of the `Content-Type` header ... I could work with any of these.

anditherobot:
## Summary

Serves the server's own API documentation. Machine clients fetch an
index of the available pages and then each page as markdown, agents get the same
through an MCP tool.

| Route | Returns |
|---|---|
| `GET /v1/docs` | the list of documentation pages this server has |
| `GET /v1/docs/api/lemonade` | the Lemonade API reference, as markdown |
| `POST /mcp` -> `lemonade_docs` | the same list, or one page, for MCP clients |

Fixes #1700

## Scope

- **Documentation API** — index plus one page per URL, on all four
  prefixes; all 7 files in `docs/api` instead of 3, read from the folder rather
  than a list in the code.
- **MCP tool** — reuses the HTTP endpoint's code.

## Testing

## Documentation

`docs/api/lemonade.md` documents both endpoints, `docs/api/mcp.md` documents the
new tool, `docs/api/README.md` was updated to match.

## Breaking Changes

## AI-assisted contribution


ckuethe:
Nice. It works for me, and degrades gracefully when the documentation path is inaccessible.

jeremyfowers:
Awesome! I would like to review this before it merges. Thanks for your patience :)

jeremyfowers (changes requested):
So, at the risk of scope-creeping you very badly...

How would you feel about changing this PR so that it builds the entire docs site with Zensical and serves that over the `/docs` endpoint, just like https://lemonade-server.ai/docs/ does? That would be much more comprehensive. It would also mean, later on, we could load it in an iframe in the GUI and display the complete up-to-date project documentation for the user's current install.

Alternatively, if you don't want to take that on, this PR still needs to be durable in the likely event people will want more in `/docs` than just `api/lemonade.md`. I think the simplest way to achieve that would be to make the current endpoint `/docs/api/lemonade`, and then it makes sense that it serves exactly that doc, at the same relative path that https://lemonade-server.ai/docs/api/lemonade/

Otherwise if `/v1/docs` serves the document `/docs/api/lemonade.md` it will be a breaking change when we inevitably decide to add more documentation to the endpoint.

Thoughts?

ckuethe:
what I would find most useful is to have `/api/docs` serve up a listing of markdown docs available. it can be html, json, or markdown - just something easy to parse. that tells the consumer which documentation pages the server knows about, eg. lemonade.md, openai.md, ollama.md,... . I can point any reasonably competent agent+model at it and it'll do the rest.

I don't think this is a breaking change because you first query `/api/docs` to see what's available *on this server*, and then you can fetch `/api/docs/{doc}`

I use this "here's the top of the docs, you figure it out" pattern a lot for making python api bindings. 

you can serve a rendered copy of the docs from `/docs` for human consumption.

ckuethe:
Lovely. Versioned doc API is fine, I'm just happy that there'll be an endpoint I can point my agent at, allowing it to figure out what endpoints are available and how to use them.

jeremyfowers:
Thanks @anditherobot the changes sound really nice! Sorry I haven't had a chance to have another look yet. I will soon.

> should I open a separate issue under GUI 3 for a docs panel? 

Please open an RFC discussion (see https://github.com/lemonade-sdk/lemonade/pull/3521) and tag @kpoineal

jeremyfowers:
Marked PR as "rfc required" for training purposes, but we've already had sufficient discussion on the PR itself.

docs/api/lemonade.md
```diff
@@ -1405,6 +1407,80 @@ curl -X POST http://localhost:13305/v1/3d/generations \
...
+### Reading the files directly
+
+The same files are installed on disk, so they can be read without a running server:
+
+| Platform | Path |
+|----------|------|
+| Windows (per-user) | `%LOCALAPPDATA%\lemonade_serverin
esources\docs\` |
+| Windows (all users) | `C:\Program Files\Lemonade Serverin
esources\docs\` |
```
jeremyfowers: What's going on with the funky escape characters here?

jeremyfowers:
> Lovely. Versioned doc API is fine, I'm just happy that there'll be an endpoint I can point my agent at, allowing it to figure out what endpoints are available and how to use them.

@ckuethe with the big expansions coming to the MCP server soon your agent might not even need to know about endpoints. Everything will be available as a tool!

docs/api/lemonade.md
```diff
@@ -27,6 +27,8 @@ We have designed a set of Lemonade-specific endpoints to enable client applicati
 | `GET` | [`/v1/models/{id}/options`](#get-v1modelsidoptions) | Read a model's saved, effective, and default recipe options |
 | `POST` | [`/v1/models/{id}/options`](#post-v1modelsidoptions) | Save recipe options for a model without loading it |
 | `DELETE` | [`/v1/models/{id}/options`](#delete-v1modelsidoptions) | Reset a model's recipe options to defaults |
+| `GET` | [`/v1/docs`](#get-v1docs) | List the API reference pages bundled with the running server |
+| `GET` | [`/v1/docs/{page}`](#get-v1docspage) | Read one bundled API reference page |
```
jeremyfowers, line 31: Does this PR also add a `/docs` endpoint that ships the rendered zensical output? If so it needs to be documented here.

Also, `/v1/docs` vs. `/docs` feels too similar of a name for me. Can you come up with something more distinct, or combine them into a single endpoint tree?

.github/workflows/cpp_server_build_test_release.yml
```diff
@@ -74,6 +74,16 @@ jobs:
...
           .\setup.ps1
 
+      - name: Bundle the documentation site
+        shell: PowerShell
+        run: |
+          $ErrorActionPreference = "Stop"
+          # zensical is pip-only, so BUILD_DOCS_SITE stays off by default
+          # and is turned on here, where the dependency can be installed.
+          python -m pip install --disable-pip-version-check -r docs/assets/docs_requirements.txt
+          cmake build -DBUILD_DOCS_SITE=ON
```
jeremyfowers: This flag needs to be documented in the developer guide

jeremyfowers: Related: I'm not able to get this to work. I'll try again when its documented.

jeremyfowers: Also, when this does't work, navigating to `BASE_URL/docs` falls through to the regular web ui. Should it 404 instead?

jeremyfowers:
image

hehe amazing!

jeremyfowers (changes requested):
This is AWESOME. Love the new design. I wasn't able to get Zensical output working, though. Please see comments.

I'm also completely happy with this implementation without Zensical, if you want to drop the Zensical part and add it back in as a follow up PR.

jeremyfowers:
image

:point_up: this is good

image

:point_up: seems like something broke though?

edit: `http://localhost:13399/v1/docs/api/README` works. So I can get specific pages. Was the root-level index lost?

src/cpp/server/server.cpp
```diff
@@ -1598,6 +1611,20 @@ void Server::setup_static_files(httplib::Server &web_server) {
...
+    // Mounted with the trailing slash so a bare /docs reaches the redirect:
+    // httplib matches mounts ahead of routes, and serving index.html at /docs
+    // would resolve the page's relative links one level too high.
+    std::string docs_site_dir = utils::get_resource_path("resources/docs-site");
+    if (web_server.set_mount_point("/docs/", docs_site_dir)) {
+        web_server.Get("/docs", [](const httplib::Request&, httplib::Response& res) {
+            res.set_redirect("/docs/", 301);
+        });
+        LOG(INFO, "Server") << "Serving bundled documentation site at /docs/" << std::endl;
+    }
```
jeremyfowers: Stale code and comment, remove?

docs/api/lemonade.md
```diff
@@ -1405,6 +1407,80 @@ curl -X POST http://localhost:13305/v1/3d/generations \
...
+### Example response
+
+```json
+{
+  "version": "11.8.0",
+  "format": "text/markdown",
+  "docs": [
+    {
+      "id": "api/README",
+      "title": "Lemonade Server API Spec",
```
jeremyfowers:
```suggestion
      "title": "Lemonade Endpoints Spec",
```

Should it be this?

jeremyfowers (changes requested):
Two very minor changes needed and then we can merge!

jeremyfowers (approved):
Thanks for the awesome work @anditherobot !

PR #3351, opened 2026-08-25
docs: add smart router subject area and two router maintainers

ramkrishna2910:
## Summary

Updates the Maintainers table in `docs/dev/contribute.md` so the smart router has listed subject-area experts.
## Scope

One note on the diff size: both new usernames are 17 characters, which is wider than the previous 15-character name column, so every row is re-padded to keep the table aligned. That is why a 3-line content change shows as 18 insertions / 16 deletions. The only content changes are the three listed above; everything else is whitespace. A side effect of the re-pad is that `@sawansri`'s row, which had extra trailing padding and was 24 characters wider than every other row, is now normalized with the rest.

## Testing

_Testing details:_

No build — this PR changes one Markdown file and touches no code.

Verified locally:
- Every row of the table is exactly 176 characters and parses as 3 cells, so the Markdown table is well formed and visually aligned.
- New rows are inserted in the existing case-insensitive alphabetical order within the non-admin block: `@eddierichter-amd` between `@bitgamma` and `@fl0rianr`, `@SlawomirNowaczyk` between `@siavashhub` and `@sofiageo`.
- Neither username was already present in the table.
- Checked the pre-commit hooks that apply to a Markdown file by hand, since `pre-commit` is not installed in my environment: no trailing whitespace, file ends with a newline, LF line endings only.

## Documentation

This PR is the documentation change.

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

@fl0rianr @eddierichter-amd @SlawomirNowaczyk — flagging you since this lists you as reviewers for router work. Say so if you would rather not be on the hook for it, or if you want the subject area worded differently; `@ramkrishna2910`'s row uses "smart router and orchestration" and I kept the shorter "smart router" for these three.

### DIFF
--- a/docs/dev/contribute.md
+++ b/docs/dev/contribute.md
@@ -78,19 +78,21 @@ While each maintainer is welcome to work on any part of the Lemonade codebase, e
 
 "Admin maintainer" means that individual is a repository admin who can tag releases and take other administrative actions.
 
-| Maintainer      | Admin | Subject Areas                                                                                                                                    |
-|-----------------|-------|--------------------------------------------------------------------------------------------------------------------------------------------------|
-| @jeremyfowers   | Yes   | new endpoints, new backends, large new features, GUI design language, new CLI commands, website, governance, Lemonade Mix (LMX) omni models, ci  |
-| @kenvandine     | Yes   | snaps, linux, new backends, hardware vendor support, Nvidia CUDA, ARM                                                                            |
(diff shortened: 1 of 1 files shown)


jeremyfowers (approved):

fl0rianr (approved):
Fine, by me, will be asked anyway.
Disabled auto merge to give the other two also a chance to respond.

eddierichter-amd (approved):
Works with me!

PR #3353, opened 2026-08-25
ci: retry apt-get on transient arm64 mirror 404s

kenvandine:
## Summary

The [Build Embeddable Lemonade (Linux ARM64) job](https://github.com/lemonade-sdk/lemonade/actions/runs/32888173142/job/97933468430) failed with a 404 fetching `libssl-dev`/`libssl3t64`/`openssl` from `ports.ubuntu.com` — a transient mirror-sync issue where the package index was briefly ahead of the pool. This cascaded and cancelled the sibling x86_64 embeddable job via `fail-fast`. Adds a 3-attempt retry with backoff around `apt-get update && apt-get install` in the two workflow steps that hit this arm64 mirror.

## Scope

## Testing

_Testing details:_ Validated workflow YAML with `python3 -c "import yaml; yaml.safe_load(...)"` and pre-commit's `check-yaml`/`Validate GitHub Workflows` hooks. This is a CI-only change; correctness will also be observed on the next PR/merge-queue run.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/.github/actions/build-debian-package/action.yml
+++ b/.github/actions/build-debian-package/action.yml
@@ -25,11 +25,24 @@ runs:
       shell: bash
       run: |
         set -e
-        if [ "${{ inputs.arch-only }}" = "true" ]; then
-          apt-get update && apt-get build-dep --arch-only . -y
-        else
-          apt-get update && apt-get build-dep . -y
-        fi
+        # The Ubuntu ports mirror occasionally serves a package index that
+        # is briefly out of sync with its pool (esp. on arm64 runners),
+        # causing 404s on individual .deb files. Retry a few times.
+        success=false
+        for i in 1 2 3; do
+          if [ "${{ inputs.arch-only }}" = "true" ]; then
+            cmd=(apt-get build-dep --arch-only . -y)
+          else
+            cmd=(apt-get build-dep . -y)
+          fi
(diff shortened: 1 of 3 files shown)


jeremyfowers (dismissed):
I am happy that CI works

jeremyfowers (approved):
Let's try again!

fl0rianr (commented):
I think this still misses an important path for the same failure mode.

`.github/actions/build-debian-package/action.yml` still runs `apt-get update && apt-get build-dep ...` only once. The failure we're seeing now ends with `Failed to process build dependencies`, so the mirror 404 can still hit after the new workflow-level retries have already succeeded. There is also another single-shot `apt-get build-dep` in the Launchpad `Catch missing build-deps` step.

Could we cover the shared `build-debian-package` path as part of this fix as well? Ideally I'd put the retry in a shared helper rather than keep duplicating the loop. On retry, I'd also discard `/var/lib/apt/lists/*` before updating again and use `Acquire::Retries=3`.

A couple of smaller things: the current retry uses a fixed 10s delay rather than backoff, and it sleeps after the final failed attempt even though there is no further retry.

The current change fixes the original embeddable ARM64 failure, but I don't think it fixes the broader ports.ubuntu.com mirror-sync flakiness yet.

PR #3357, opened 2026-08-25
ci: shared apt-get retry helper + reuse build-environment containers on arm64

kenvandine:
## Summary

Follow-up to #3353 (temporary fix) per review feedback: replaces the duplicated inline `apt-get` retry loops with a shared `retry-apt-get` composite action (exponential backoff, clears `/var/lib/apt/lists` before retrying, passes `Acquire::Retries=3`), extends retry coverage to `build-debian-package`'s `apt-get build-dep` and launchpad-ppa's `Catch missing build-deps` step, and — per the suggestion to reuse existing container infrastructure — switches `build-lemonade-linux-arm64` and launchpad-ppa's `package-arm64` job to reuse the same multi-arch `ghcr.io/lemonade-sdk/lemonade/build-environment` image the x86_64 jobs already pull (confirmed it now has an arm64 manifest), instead of apt-get-installing a toolchain against the flaky `ports.ubuntu.com` mirror on every run.

Fixes #

## Scope

## Testing

_Testing details:_ Validated all changed workflow/action YAML with `python3 -c "import yaml; yaml.safe_load(...)"` and pre-commit's `check-yaml`/`Validate GitHub Actions`/`Validate GitHub Workflows` hooks. Manually simulated the retry loop's success/failure/backoff behavior in bash outside of CI. Confirmed via the ghcr.io registry API that `build-environment:ubuntu24.04` (and other tags) publish an `arm64` manifest, which the launchpad-ppa `package-arm64` job now relies on. Real-world correctness will also be observed once this workflow runs in CI.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/.github/actions/build-debian-package/action.yml
+++ b/.github/actions/build-debian-package/action.yml
@@ -22,14 +22,10 @@ runs:
   using: composite
   steps:
     - name: Install build dependencies
-      shell: bash
-      run: |
-        set -e
-        if [ "${{ inputs.arch-only }}" = "true" ]; then
-          apt-get update && apt-get build-dep --arch-only . -y
-        else
-          apt-get update && apt-get build-dep . -y
-        fi
+      uses: ./.github/actions/retry-apt-get
+      with:
+        command: >-
+          ${{ inputs.arch-only == 'true' && 'build-dep --arch-only . -y' || 'build-dep . -y' }}
 
     - name: Build Debian package
       id: build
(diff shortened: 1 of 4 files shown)


kenvandine:
@copilot please fix the merge conflicts in this pull request.

superm1:
This didn't fix the issue and is a bandaid for an infrastructure issue.  If we continue to have problems, please instead switch to a full container workflow that isn't dependent upon 3rd party (non Github) infrastructure.

PR #3361, opened 2026-08-26
fix: recover repo name for extra models kept in HF-cache layout

### LINKED ISSUE #3297: Models moved from "default lemonade model directory" TO "external custom models directory" don't display correctly

Davedave001:
## Summary

Retargets #3305 onto `main` per @jeremyfowers's request ("No C++ PRs into GUI3_merging, please"). Same change, rebuilt as a clean two-commit branch off `main` instead of `GUI3_merging` (cherry-picked cleanly, no conflicts). #3305 is being closed in favor of this one.

`discover_extra_models_in_directory` named an extra-directory model after its immediate containing folder. When a model is moved into `extra_models_dir` with its Hugging Face cache layout intact (`models--org--repo/snapshots/<hash>/...`), that folder is the snapshot hash, so the model showed up with an unreadable hash instead of its name once moved out of the default cache location.

This walks up from the discovered file to the `models--`/`modelscope--models--` ancestor (bounded by the configured `extra_models_dir` root) and decodes it back to `org/repo` when that layout is present, falling back to the previous folder-name behavior otherwise.

Per @fl0rianr's review on #3305: a cache directory can hold more than one `snapshots/<hash>` revision, and recovering `org/repo` for all of them would make them race for the same model id. The second commit fixes this — only the snapshot `refs/main` actually points at is trusted with the recovered name; any other snapshot falls back to the prior hash-named behavior. Added two regression tests: single-snapshot recovery, and the multi-snapshot case confirming only the active snapshot claims the name.

Fixes #3297

## Scope

## Testing

_Testing details:_
No C++ toolchain was available in the environment this change was authored in, so I could not build or run the C++ side locally. I added two regression tests to `test/server_endpoints.py` following the existing `extra_models_dir` test patterns in that file (`test_021yb_extra_hf_cache_layout_recovers_repo_name`, `test_021yc_extra_hf_cache_multiple_snapshots_prefers_active_ref`) covering both the single-snapshot recovery and the multi-snapshot collision case @fl0rianr flagged. I verified the diff against `main` is scoped to exactly the two intended files with no incidental changes (`git diff --stat upstream/main..HEAD`). Please let me know if CI surfaces anything and I'll fix it quickly.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -1162,6 +1162,37 @@ ModelInfo ModelManager::init_extra_model_info(const std::string& name) const {
     return info;
 }
 
+// A model moved (with its full hub layout) into extra_models_dir keeps its
+// "models--org--repo/snapshots/<hash>/..." shape. Walking up from a discovered
+// file to that "models--"/"modelscope--models--" ancestor recovers "org/repo"
+// so the model keeps a readable name instead of the snapshot hash that would
+// otherwise be read as the containing folder's name. Stops at `stop_at` (the
+// configured extra_models_dir root) so an unrelated ancestor name can't match.
+static std::string hf_cache_repo_name_for_path(const fs::path& path, const fs::path& stop_at) {
+    static constexpr const char kHfPrefix[] = "models--";
+    static constexpr const char kMsPrefix[] = "modelscope--models--";
+    for (fs::path cur = path; cur.has_filename() && cur != stop_at; cur = cur.parent_path()) {
+        std::string dirname = cur.filename().string();
+        std::string encoded;
+        if (dirname.rfind(kMsPrefix, 0) == 0) {
+            encoded = dirname.substr(sizeof(kMsPrefix) - 1);
+        } else if (dirname.rfind(kHfPrefix, 0) == 0) {
+            encoded = dirname.substr(sizeof(kHfPrefix) - 1);
+        } else {
+            continue;
+        }
+        // registry_repo_cache_dir_name() encodes "org/repo" as "org--repo";
+        // repo names don't contain "--", so the first occurrence is the
+        // namespace boundary.
+        size_t sep = encoded.find("--");
+        if (sep == std::string::npos || sep == 0 || sep + 2 >= encoded.size()) {
+            return encoded;
+        }
+        return encoded.substr(0, sep) + "/" + encoded.substr(sep + 2);
+    }
+    return "";
+}
+
 // Record a discovered model without ever overwriting one already found. Two
 // extra_models_dir folders can hold identically named files; qualifying the
 // newcomer with its folder keeps both and leaves the first model's id alone.
@@ -1283,6 +1314,11 @@ void ModelManager::discover_extra_models_in_directory(
     std::map<std::string, ModelInfo>& discovered) const {
 
     std::string dir_name = dir_path.filename().string();
+    std::string hf_repo_name = hf_cache_repo_name_for_path(
+        dir_path, path_from_utf8(extra_models_dir_));
+    if (!hf_repo_name.empty()) {
+        dir_name = hf_repo_name;
+    }
     fs::path main_model_path; // File the old folder-based discovery would have selected.
     std::vector<fs::path> mmproj_files;
     double total_size = 0.0;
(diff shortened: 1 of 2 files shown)


fl0rianr (changes requested):
Thanks for retargeting this to main. The multi-snapshot case I mentioned on #3305 looks correctly addressed now.

I found one remaining edge case that I think we should fix before merge:

`hf_cache_repo_name_for_path()` currently finds any ancestor starting with `models--` / `modelscope--models--`, but it does not require the discovered model to actually live under `cache_root/snapshots/<hash>`.

The `refs/main` check only runs if such a snapshot directory happens to be found while walking up the path. If none is found, the loop finishes and we still decode the `models--org--repo` directory name.

So for example a perfectly normal custom layout like:

```text
extra_models_dir/
  models--acme--my-model/
    model.gguf
```

would be renamed to `extra.acme/my-model`, even though this is not an HF cache layout at all.

I think recovery should only happen after we have positively identified:

```text
<cache_root>/snapshots/<hash>/...
```

and that `<hash>` matches `refs/main`. If no `snapshots/<hash>` ancestor exists, or `refs/main` is absent/invalid, we should just return `""` and keep the normal extra-directory naming behavior.

It would be good to add a negative regression test for this too, e.g. a regular `models--acme--ordinary-folder/model.gguf` directory must remain `extra.models--acme--ordinary-folder`.

One smaller edge case worth deciding on: the walk currently stops before inspecting `stop_at` itself. So if `extra_models_dir` is configured directly to a `models--org--repo` cache root, recovery will not happen. I don't consider that as important as the false-positive case above, but it would be good to either support it or make the intended boundary explicit.

Also, the large explanatory comment above `hf_cache_repo_name_for_path()` will currently fail our `comment-slop` pre-commit hook. It is a 15-line contiguous comment block, while `tools/check_comment_slop.py` has `MAX_BLOCK_LINES = 12`. Most of that explanation is already in the PR description, so I would shorten the source comment to just the non-obvious invariant, something along the lines of:

```cpp
// Recover org/repo only for the active snapshots/<hash> entry of a
// recognized HF/ModelScope cache. Other paths keep normal directory naming.
```

The two new positive tests are useful, especially the active-vs-stale snapshot case. Once the actual cache-layout check and the negative test are added, and pre-commit/CI are green, this looks good to me.

Davedave001:
Thanks for the catch — fixed. `hf_cache_repo_name_for_path()` now only recovers the name after positively identifying a `cache_root/snapshots/<hash>` ancestor whose hash matches `refs/main`; a plain `models--*` folder with no snapshots/ layout keeps its normal folder name. Added a negative regression test for that case (`test_021yd`), and shortened the comment block to fit the comment-slop pre-commit limit. Left the `stop_at`-as-cache-root edge case as-is per your note that it's minor — happy to add explicit support if you'd rather.

fl0rianr (commented):
Thanks, good fixes.

I found one remaining compatibility issue before I’d approve this:

`dir_name` is replaced with the recovered `org/repo` name before the existing hidden-alias logic runs. For split folders this means the old folder name, i.e. the snapshot hash, is no longer preserved as an input alias even though that is an existing compatibility contract from #2107.

So for a cache layout with multiple GGUF variants, requests using either:

```text
<snapshot_hash>
extra.<snapshot_hash>
```

would stop resolving after this change.

I think the cleanest fix is to keep the physical folder name separately:

```cpp
const std::string folder_name = dir_path.filename().string();
std::string dir_name = folder_name;
```

then use `dir_name` for the new readable `org/repo` identity, but keep `folder_name` / `extra.<folder_name>` as hidden aliases whenever the recovered name differs.

It would be good to add one regression test combining the HF-cache layout with multiple GGUF variants and verify that both old hash-based aliases still resolve to the same main variant while the model is listed under the new readable name.

A few smaller things:

* Please don't forget updating the PR description after the next fix
* #3333 is modifying the same extra-model discovery area, so this will likely need a careful rebase/reconciliation depending on merge order.

jeremyfowers:
Bumping :) since this sounds like a nice fix to have

PR #3366, opened 2026-08-26
test(gpu-hang): scale request timeouts to the model-operation budget

jeremyfowers:
## Summary

`test_gpu_hang_retry.py` exercises a path where lemond resets a hung backend, reloads the model, and replays the request, but it capped every request at a fixed 60 seconds. On a loaded runner the reload does not finish inside that budget, so the test raises `requests.exceptions.ReadTimeout` instead of seeing the recovery it is meant to verify. That failed the `Test streaming-errors` job in the merge queue on 2026-08-26.

The six request timeouts now use `TIMEOUT_MODEL_OPERATION`, which is **500 seconds** (`test/utils/test_models.py:138`). That is the shared constant `server_watchdog_lifecycle.py` and `server_base.py` already use for requests that can span a model load, so this raises the budget from 60s to 500s and stops hard-coding it in a third place.

## Scope

## Testing

_Testing details:_

No C++ changes, so nothing to build. Verified locally that the new import resolves to 500 and that `pre-commit run` passes including black.

The test itself needs a built `lemond` plus the mock backend harness, so the run is left to CI on this PR.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

### DIFF
--- a/test/test_gpu_hang_retry.py
+++ b/test/test_gpu_hang_retry.py
@@ -20,7 +20,11 @@ import concurrent.futures
 
 # Add test/ to path
 sys.path.append(os.path.dirname(os.path.abspath(__file__)))
-from utils.test_models import ENDPOINT_TEST_MODEL, get_default_lemond_binary
+from utils.test_models import (
+    ENDPOINT_TEST_MODEL,
+    TIMEOUT_MODEL_OPERATION,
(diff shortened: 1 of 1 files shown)


bitgamma (approved):
Lgtm

PR #3374, opened 2026-08-26
feat(backends): hrx backend [experimental]

AaronStGeorge:
## Summary

Adds an `hrx` backend, this is an experimental llama.cpp backend making its way into llama.cpp mainline. See https://github.com/ggml-org/llama.cpp/discussions/27219 for more information. 

Currently hrx only supports `Qwen3-30B-A3B-Instruct-2507-Q4_K_M`, this will evolve rapidly but _today_ model support is limited. To avoid burdening the `llama.cpp` recipe with additional functionality to limit model support per backend, this PR introduces hrx as a separate recipe vs a `llama.cpp` backend. With a separate recipe we can scope hrx to the qualified models "for free." When hrx support advances to be competitive with other `llama.cpp` backends the hrx specific recipe will be removed and it will become a regular `llama.cpp` backend.

Hrx recipe simply wraps the existing llama.cpp recipe, so there's not much to it.

image

## Testing

I verified locally on a W7900 `gfx1100` machine. The `llama-server` artifact is verified in CI on strix halo `gfx1151` machine ([link](https://github.com/ROCm/ggml-staging-automation/actions/runs/32521144748)).

### DIFF
--- a/docs/assets/models.js
+++ b/docs/assets/models.js
@@ -7,7 +7,7 @@ const RECIPE_PRIORITY = [
   'acestep',
   'ds4',
   'flm',
-  'hrx',
+  'llamacpp-hrx',
   'kokoro',
   'llamacpp',
   'moonshine',
@@ -24,7 +24,7 @@ const RECIPE_PRIORITY = [
 
 const RECIPE_DISPLAY_NAMES = {
   llamacpp: 'llama.cpp GPU',
-  hrx: 'HRX GPU (experimental)',
+  'llamacpp-hrx': 'HRX GPU (experimental)',
   whispercpp: 'whisper.cpp',
   'sd-cpp': 'stable-diffusion.cpp',
   flm: 'FastFlowLM NPU',
--- a/docs/assets/models.js
+++ b/docs/assets/models.js
@@ -7,6 +7,7 @@ const RECIPE_PRIORITY = [
   'acestep',
   'ds4',
   'flm',
+  'hrx',
   'kokoro',
   'llamacpp',
   'moonshine',
@@ -23,6 +24,7 @@ const RECIPE_PRIORITY = [
 
 const RECIPE_DISPLAY_NAMES = {
   llamacpp: 'llama.cpp GPU',
+  hrx: 'HRX GPU (experimental)',
   whispercpp: 'whisper.cpp',
   'sd-cpp': 'stable-diffusion.cpp',
   flm: 'FastFlowLM NPU',
--- a/src/cpp/include/lemon/backends/hrx/hrx.h
+++ b/src/cpp/include/lemon/backends/hrx/hrx.h
@@ -43,10 +43,10 @@ inline const std::set<std::string>& reserved_custom_arg_flags() {
 }
 
 inline const BackendDescriptor descriptor = {
-    /*recipe*/          "hrx",
+    /*recipe*/          "llamacpp-hrx",
     /*display_name*/    "HRX GPU (experimental)",
     /*binary*/          "llama-server",
-    /*config_section*/  "",  // defaults to recipe
+    /*config_section*/  "hrx",
     /*default_device*/  DEVICE_GPU,
     /*slot_policy*/     SlotPolicy::Standard,
     /*selectable_backend*/ false,
--- a/src/cpp/include/lemon/backends/hrx/hrx.h
+++ b/src/cpp/include/lemon/backends/hrx/hrx.h
@@ -9,14 +9,6 @@ namespace lemon {
 namespace backends {
 namespace hrx {
 
-inline constexpr char kQualifiedCheckpoint[] =
-    "unsloth/Qwen3-30B-A3B-Instruct-2507-GGUF:"
-    "Qwen3-30B-A3B-Instruct-2507-Q4_K_M.gguf";
-
-inline bool is_qualified_checkpoint(const std::string& checkpoint) {
-    return checkpoint == kQualifiedCheckpoint;
-}
-
 inline const std::set<std::string>& reserved_custom_arg_flags() {
     static const std::set<std::string> kReservedCustomArgFlags = {
         "-c",
--- a/src/cpp/include/lemon/backends/hrx/hrx.h
+++ b/src/cpp/include/lemon/backends/hrx/hrx.h
@@ -1,0 +1,84 @@
+#pragma once
+
+#include "lemon/backends/backend_descriptor.h"
+
+#include <set>
+#include <string>
+
+namespace lemon {
+namespace backends {
+namespace hrx {
+
+inline constexpr char kQualifiedCheckpoint[] =
+    "unsloth/Qwen3-30B-A3B-Instruct-2507-GGUF:"
+    "Qwen3-30B-A3B-Instruct-2507-Q4_K_M.gguf";
+
+inline bool is_qualified_checkpoint(const std::string& checkpoint) {
+    return checkpoint == kQualifiedCheckpoint;
+}
+
+inline const std::set<std::string>& reserved_custom_arg_flags() {
+    static const std::set<std::string> kReservedCustomArgFlags = {
+        "-c",
+        "-dev",
+        "-dr",
+        "-hf",
+        "-hff",
+        "-hfr",
+        "-m",
+        "-mu",
+        "--ctx-size",
+        "--device",
+        "--docker-repo",
+        "--hf-file",
+        "--hf-repo",
+        "--jinja",
+        "--metrics",
+        "--model",
+        "--model-url",
+        "--no-jinja",
+        "--port",
+    };
+    return kReservedCustomArgFlags;
+}
+
+inline const BackendDescriptor descriptor = {
+    /*recipe*/          "hrx",
+    /*display_name*/    "HRX GPU (experimental)",
+    /*binary*/          "llama-server",
+    /*config_section*/  "",  // defaults to recipe
+    /*default_device*/  DEVICE_GPU,
+    /*slot_policy*/     SlotPolicy::Standard,
+    /*selectable_backend*/ false,
+    /*uses_ctx_size*/   true,
+    /*dynamic_models*/  false,
+    /*options*/ {
+        {"hrx_args", "--hrx-args", "", "ARGS",
+         "Custom arguments to pass to the HRX llama-server", "HRX Options"},
+    },
+    /*support*/ {
+        {"hrx", {"linux"},
+         {{"amd_gpu", {"gfx1100", "gfx1151"}}},
+         "AMD GPUs (gfx1100, gfx1151)",
+         {{"gfx1100", {/*os*/ {"linux"}, /*channels*/ {}}},
+          {"gfx1151", {/*os*/ {"linux"}, /*channels*/ {}}}}},
+    },
+    /*supported_modes*/ {"chat", "embeddings", "reranking"},
+    /*required_checkpoints*/ {"main"},
+    /*default_capabilities*/ {},
+    /*experimental*/    true,
+    /*web_display_name*/ "",
+    /*rocm_channels*/   {},
+    /*exposes_prometheus_metrics*/ true,
+    /*rocm_requires_cwsr_fix*/ false,
+    /*version_policy*/  VersionPolicy::Exact,
+    /*self_manages_downloads*/ false,
+    /*takes_args*/      true,
+    /*arg_variants*/    {},
+    /*bin_variants*/    {"hrx"},
+    /*config_extra*/    nlohmann::json::object(),
+};
+
+}  // namespace hrx
+}  // namespace backends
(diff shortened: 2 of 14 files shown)


docs/assets/models.js
```diff
@@ -23,6 +24,7 @@ const RECIPE_PRIORITY = [
 
 const RECIPE_DISPLAY_NAMES = {
   llamacpp: 'llama.cpp GPU',
+  hrx: 'HRX GPU (experimental)',
```
superm1: `llamacpp-hrx` maybe instead?

jeremyfowers:
34 MB backend archive downloaded instantly :face_holding_back_tears:

jeremyfowers:
image

Up and running!

docs/assets/models.js
jeremyfowers: +1, will help people understand what it is

docs/assets/models.js
```diff
@@ -7,6 +7,7 @@ const RECIPE_PRIORITY = [
   'acestep',
   'ds4',
   'flm',
+  'hrx',
```
jeremyfowers: My only real question with this PR is whether we want to add it to inference backend testing in https://github.com/lemonade-sdk/lemonade/blob/main/.github/workflows/cpp_server_build_test_release.yml

It would be really easy to do, since there is already an established pattern for llamacpp testing. 

The main downsides would be 
1. if HRX isn't ready for extensive functional testing yet (I dont know the answer, please inform). although we could mitigate this by disabling capabilities in https://github.com/lemonade-sdk/lemonade/blob/main/test/utils/capabilities.py
2. The 30B model is a lot bigger than other other test models, and would be slower, however this isn't a huge issue since it runs fast on strix halo CI runners

We'll also certainly want to add this testing eventually. @AaronStGeorge please advise if you think this is the right time.

src/cpp/server/backends/hrx/hrx_server.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    std::string validate_registration_checkpoint(
+        const std::string& checkpoint) const override {
+        return llamacpp::ops()->validate_registration_checkpoint(checkpoint);
+    }
+
+    std::string validate_checkpoint_file(
+        const std::string& resolved_path) const override {
+        return llamacpp::ops()->validate_checkpoint_file(resolved_path);
+    }
```
jeremyfowers, line 188: `HrxOps` delegates `populate_metadata`, `resolve_checkpoint_path`, `find_imported_checkpoint`, and both validators to `llamacpp::ops()`, but not `resolve_runtime_options` — so it falls back to the no-op `BackendOps` default and HRX loses the `--parallel 1` pin.

`LlamaCppOps::resolve_runtime_options` (`llamacpp_server.cpp:787`) appends `--parallel 1` via `resolve_llamacpp_runtime_args` specifically because an auto slot count enables the unified KV buffer, which advertises the full `ctx_size` to every slot instead of dividing it. HRX launches the same `llama-server` binary with the same `--ctx-size`, so on a build whose `-np` defaults to auto, the router's auto-tuned `ctx_size` gets divided across slots and a request silently gets less context than lemonade computed and reports.

Delegating here would also need the option key remapped (`llamacpp_args` → `hrx_args`), which is presumably why it was skipped — but the `--parallel 1` default still needs to land somewhere, e.g. directly in `hrx::build_server_argv`. (`--parallel`/`-np` is also absent from `reserved_custom_arg_flags()`, so a user can work around it, but the default is wrong.)

jeremyfowers:
@AaronStGeorge I just set the CI checks to run. The docs check will most likely fail, you just need to run `python docs/tools/gen_backend_boilerplate.py` to auto-generate the correct documentation.

jeremyfowers (approved):
Worked great for me out of box! Easy approve. Please see the open conversations and lets get those resolved before merging.

I also want to see sign-off from @superm1 @Geramy and @iswaryaalex before we merge (hopefully this is straightforward, looks that way to me, I just want us to all be aligned).

docs/assets/models.js
Geramy: I would agree on gating this based on its hardware capabilities during runtime.

docs/assets/models.js
Geramy: +1

src/cpp/include/lemon/backends/hrx/hrx.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include <set>
+#include <string>
+
+namespace lemon {
+namespace backends {
+namespace hrx {
+
+inline constexpr char kQualifiedCheckpoint[] =
+    "unsloth/Qwen3-30B-A3B-Instruct-2507-GGUF:"
```
Geramy: Just wondering, I know this was built for Qwen3_MoE, can we not just gate on that kernel here instead and Q4? It would give users flexibility of running any flavor model that would be in compliance with that kernel and quantize which is supported by llamacpp-hrx, this I think is more preferred by the users. Not a requirement please let me know if you think this is easily possible.

docs/assets/models.js
AaronStGeorge: https://github.com/lemonade-sdk/lemonade/blob/main/.github/workflows/cpp_server_build_test_release.yml looks like it's all windows at the moment? If there's a linux pattern we can follow we can definitely add it to CI.

> 1. if HRX isn't ready for extensive functional testing yet (I dont know the answer, please inform). although we could mitigate this by disabling capabilities in https://github.com/lemonade-sdk/lemonade/blob/main/test/utils/capabilities.py

As far as I know it's ready, I'm sure we would find some gaps, but finding and filling the gaps isn't a bad thing😄.

> 2. The 30B model is a lot bigger than other other test models, and would be slower, however this isn't a huge issue since it runs fast on strix halo CI runners
> 
> We'll also certainly want to add this testing eventually. @AaronStGeorge please advise if you think this is the right time.

I didn't look into adding CI coverage yet simply because of the single large model currently supported. But if you're OK with the CI time cost of the large download we could add it now.

src/cpp/server/backends/hrx/hrx_server.cpp
AaronStGeorge: 👍 addressed in https://github.com/lemonade-sdk/lemonade/pull/3374/changes/88391a9acd025f8d1c86b09ae568fb31c876aebd

docs/assets/models.js
AaronStGeorge: Done in https://github.com/lemonade-sdk/lemonade/pull/3374/changes/2816e38fb568e95d58dc0bde91048b5ac4ba85f1

src/cpp/include/lemon/backends/hrx/hrx.h
```diff
@@ -0,0 +1,84 @@
...
+         "Custom arguments to pass to the HRX llama-server", "HRX Options"},
+    },
+    /*support*/ {
+        {"hrx", {"linux"},
+         {{"amd_gpu", {"gfx1100", "gfx1151"}}},
+         "AMD GPUs (gfx1100, gfx1151)",
+         {{"gfx1100", {/*os*/ {"linux"}, /*channels*/ {}}},
+          {"gfx1151", {/*os*/ {"linux"}, /*channels*/ {}}}}},
+    },
+    /*supported_modes*/ {"chat", "embeddings", "reranking"},
```
iswaryaalex: should we limit this to "chat" since we only currently support a given chat model?

src/cpp/resources/defaults.json
```diff
@@ -26,6 +26,10 @@
   },
   "global_timeout": 600,
   "host": "localhost",
+  "hrx": {
+    "args": "",
```
iswaryaalex, line 30: should this be `args` or actually `hrx_args` ?

src/cpp/include/lemon/backends/hrx/hrx.h
AaronStGeorge: Currently hrx needs `qwen3moe` architecture and `Q4_K_M` quant specifically.

That said, I think we should just remove this gate entirely. The `llamacpp-hrx` recipe ships one [server model](https://github.com/lemonade-sdk/lemonade/pull/3374/changes#diff-ac5802948a3b43c44ff2ebffbcab0ac9fa3bdd2764996ae6199c78366f290f29) (fully qualified), so this check only prevents someone using a local gguf file. A local gguf file may not work, but some will, `Qwen3-Coder-30B-A3B-Instruct-Q4_K_M` for example. Given that the backend is marked experimental allowing any local model someone wants to try makes more sense to me.

Geramy: That would be great, I know people love playing with custom tunes and not just unsloth, for example some people do the undistilled models too so that would be a ``nice to have`` before this merges, I appreciate it!

AaronStGeorge: ^ pushed [here](https://github.com/lemonade-sdk/lemonade/pull/3374/changes/579cc20c22403cd99f7f8746064b2f27790162d8) let me know if you disagree and I can back that out.

Geramy: if you dont mind, it would be nice if https://github.com/lemonade-sdk/lemonade/pull/3382 could land real quick and you could use this for your arch and quant filtering :)

Geramy: it should take care of arch and quant filtering so kernels and quants can be added as it gets implemented, reducing user friction also.

Geramy (approved):
I approve it as it is, but I would love to see https://github.com/lemonade-sdk/lemonade/pull/3382 being used in this, it would help users pick supported models and quants from the search which would be amazing utilization for a up and coming backend so as archs get added and supported there are nice guards in place for users.

iswaryaalex:
Works smooth on my end!

iswaryaalex (approved):

src/cpp/resources/defaults.json
AaronStGeorge: It looks like `hrx_args` [gets converted to `args`](https://github.com/lemonade-sdk/lemonade/blob/7dbac3c52827a36d459d143731b6e524c2b5504b/src/cpp/server/runtime_config.cpp#L716-L722), so I think we want `args`.

src/cpp/include/lemon/backends/hrx/hrx.h
AaronStGeorge: Do we really need a guard in load at all? For a random model a user wanted to load it would be great if the UI was filtered to avoid sending in something the backend can't handle, but that would be well upstream of load. A load guard might produce a better error message user but if load fails or if the model fails I think the user is getting an error either way right?

src/cpp/include/lemon/backends/hrx/hrx.h
AaronStGeorge: [done](https://github.com/lemonade-sdk/lemonade/pull/3374/changes/e7d718f2917e70ea5b9e5b82c1a8c15d8a848373)

src/cpp/include/lemon/backends/hrx/hrx.h
AaronStGeorge: Just trying to get a handle on things, looking at https://github.com/lemonade-sdk/lemonade/pull/3382 (I'm a lemonade newbie so correct me if I'm wrong) it could be used to drive the UI but wouldn't _yet_ (it could in the future). So changes to this PR would be adding the `ModelConstraint` to the backend definition + re-adding the load guard?

If so works for me👍, I'm not sure we need a load guard but doesn't hurt for sure.

Geramy: > Just trying to get a handle on things, looking at #3382 (I'm a lemonade newbie so correct me if I'm wrong) it could be used to drive the UI but wouldn't _yet_ (it could in the future). So changes to this PR would be adding the `ModelConstraint` to the backend definition + re-adding the load guard?
> 
> If so works for me👍, I'm not sure we need a load guard but doesn't hurt for sure.

That is correct #3382 does drive the UI and puts safeguards around user interfaces, currently lemonade will error when trying to load a model that doesn't load, its not too friendly in terms of it will say "error model not compatible with backend" so that could generate friction, but its also experimental so it has the possibility of generating low friction. I am under the impression the current model implemented is qwen3_moe + q4 so that to me means anyone can run any model with those specifications? If that's true I think it would be a good add. The usage you are expressing is correct that is how you would use it. @jeremyfowers let me know that he thinks it's fine the way it is now, I am okay with that. So I will let you and Jeremy decide if its worth it, at the end of the day you would know the hrx scope more than I currently do so I believe the decision lands on you two.

jeremyfowers: Discussed offline; agreed to keep 3382 open as a draft and defer it until we're convinced it would help a specific scenario.

PR #3379, opened 2026-08-26
Fix(#3336) improve validate llamacpp.py

### LINKED ISSUE #3336: Testing: Improve validate_llamacpp.py
As of https://github.com/lemonade-sdk/lemonade/pull/3322#top validate_llamacpp.py now uses /internal/set for rocm_channel, but its request_json() still sends no auth header. This will fail if LEMONADE_ADMIN_API_KEY is configured, so it would be nice to reuse the shared auth helper there

GabrielReusRodriguez:
## Summary

This PR fixes #3336 : Testing: Improve validate_llamacpp.py
I have added  the _auth_headers to the request_json function  to get LEMONADE_ADMIN_API_KEY or LEMONADE_API_KEY  and then pass it to the headers  of request. I saw  validate_vllm.py had the same issue so I added the same solution to it.

The solution was  add the _auth_headers function call and check if the returned dictionary was not empty. If it was NOT empty, I update the headers dictionary in **kwargs with the returned value.

image

Fixes #3336 

## Scope

- [ ] I kept refactoring separate unless it is required for this change.

## Testing

_Testing details:_

I called it without my changes and it failed:

image

Then I applied the changes and it worked
image

I used the same  testing with validate_vllm.py
Without changes:
image

With changes:

image

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

- [ ] I used AI tools for this PR.
- [x] I did not use AI tools for this PR.

### DIFF
--- a/test/validate_llamacpp.py
+++ b/test/validate_llamacpp.py
@@ -27,7 +27,7 @@ import tempfile
 
 import requests
 
-from utils.server_base import unload_all_models, wait_for_server
+from utils.server_base import _auth_headers, unload_all_models, wait_for_server
 from utils.test_models import PORT, TIMEOUT_DEFAULT
 
 TIMEOUT_HEALTH = 60
@@ -61,6 +61,11 @@ def collect_server_logs(output_dir):
 
 def request_json(method, url, timeout, **kwargs):
     """Perform an HTTP request and parse the JSON response when present."""
+    api_key = _auth_headers()
+    if api_key: 
+        headers = kwargs.get("headers", {})
+        headers.update(api_key)
+        kwargs["headers"] = headers
     response = requests.request(method, url, timeout=timeout, **kwargs)
     body = {}
     if response.content:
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
Overall this looks correct and I think including validate_vllm.py makes sense since it has the same auth issue.

One small blocker: both files have trailing whitespace after `if api_key:`, which should fail the repo's trailing-whitespace pre-commit hook.

Also a minor non-blocking nit: `headers.update(api_key)` mutates a caller-provided headers dict in place. Something like `{**(kwargs.get("headers") or {}), **auth_headers}` would avoid that and `auth_headers` would also be a clearer name than `api_key`.

Otherwise the fix looks good to me.

GabrielReusRodriguez:
Hello, I made some changes  after reading the review:

Removed trail spaces
renamed api_key to auth_headers
Now I get the auth_header , I join to the headers dict and then I create a new dictionary with everything that was at kwargs  plus the  headers  (including auth_header)

fl0rianr (approved):
Thanks, fine by me now.

Please apply the very simple Black format pre-commit fix.

superm1 (approved):

PR #3380, opened 2026-08-26
Fix compact chat model selector and modality display

falabellamichael:
## Summary

This keeps the chat model selector, Effective Settings, and Logs controls on one row at narrow widths while making the selected model and its supported input modalities easier to understand.

The model selector now:

- shows Chat with Image and Audio modality icons when the selected chat model supports those inputs
- keeps the selected model name visible when that model is not currently loaded
- hides the longer mode label when the Models rail collapses, while retaining the modality icons
- uses container-aware compaction so the model name, status badge, and caret remain stable from desktop down to 320 px

## Scope

## Testing

Testing details:

- git diff --check
- npm.cmd run typecheck
- npm.cmd run test:chat-audio
- npm.cmd run test:ui-regressions
- npm.cmd run build:renderer
- Playwright A186 responsive group using installed Chrome: 3 passed
- Browser assertions cover 851, 769, 768, 480, 400, 360, and 320 px states, loaded Chat + Image + Audio display, unloaded selected-model display, same-row containment, menu positioning, and caret alignment.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/app/src/components/ChatView.tsx
+++ b/src/app/src/components/ChatView.tsx
@@ -3634,7 +3634,10 @@ ${finalText}`
       <div className="composer" onDrop={handleDrop} onDragOver={handleDragOver}>
         <div className="composer__toolbar">
           {(modelPickerOptions.length > 0 || modelPickerOpen) && (
-            <div className="composer__model-picker" ref={modelPickerRef}>
+            <div
+              className={`composer__model-picker${currentLoadedModel ? ' composer__model-picker--loaded' : ''}`}
+              ref={modelPickerRef}
+            >
               <span className="composer__model-label">Model</span>
               <button
                 type="button"
@@ -3642,6 +3645,7 @@ ${finalText}`
                 onClick={() => { setModelPickerOpen(v => !v); setModelPickerError(null); }}
                 aria-haspopup="listbox"
                 aria-expanded={modelPickerOpen}
+                aria-label={currentModel ? `Select model, current ${currentModel}` : 'Select model'}
               >
                 {currentLoadedModel ? (
                   <span className={`composer__model-mode composer__model-mode--${modelModeBadge(currentCapability, currentRecipe)}`}>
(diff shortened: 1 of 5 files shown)


falabellamichael:
i see, ok understood, Codex. lol :o i did not realize that's what it was for, i can fix this

falabellamichael:
@jeremyfowers i have the fix for this in mind already, models becomes an ^ arrow at the smallest and final breakpoint, that's how Hermes** itself does it actually and others i've seen.  i can add the wrapper back but make sure it won't even be needed ;)

falabellamichael:
Follow-up pushed in 8e56b2a5. This update addresses the narrow-width issue reported in review and consolidates the responsive toolbar behavior.

Full changes:

- Keeps Parameter Settings and Logs together on the right side of the chat toolbar instead of allowing Logs to wrap below.
- Preserves the normal MODEL label and full model selector above the phone breakpoint.
- Uses one responsive transition at 480 px, removing the intermediate state where the selector progressively narrowed and then expanded again.
- Hides the MODEL label at 480 px and below to make room for the remaining toolbar controls.
- Makes the compact selector state-aware: when a model is loaded, it displays the normal mode, model name, available-model count, and caret pill; when no model is loaded, it collapses to a 30 px caret-only control.
- Keeps the loaded pill at its intrinsic content width, capped at min(260px, 100%), so it does not add unused white space or participate in unnecessary flex shrinking.
- Truncates long model names only when the parent genuinely has insufficient room, including the 320 px safety case.
- Retains the original Lemonade down-caret style at 16 px.
- Places the caret in a fixed 16 by 16 rotor with a centered transform origin, so only the glyph rotates 180 degrees when the model menu opens and closes.
- Preserves reduced-motion behavior through the existing global motion preference handling.
- Adds an accessible model-selector label that includes the current model and hides the decorative caret from assistive technology.
- Keeps the model menu fully contained inside the viewport at the narrowest tested width.
- Mirrors the source CSS changes into critical.generated.css.

Verification completed locally:

- Focused Playwright coverage for 851, 480, 400, and 320 px layouts.
- Loaded compact selector, unloaded caret-only selector, steady settled width, right-side Settings and Logs grouping, menu containment, and centered open/close animation assertions.
- npm run typecheck.
- npm run test:ui-regressions.
- Production renderer builds for both the regular and AMD Dev clients.
- Native Windows builds for both clients.
- git diff --check.

No base branch was merged or modified; this is an additional commit on the existing PR branch for review.

PR #3368, opened 2026-08-26
server: fix MTP draft companion handling

fl0rianr:
## Summary

Fix automatic GGUF draft companion selection for remote registry pulls.

Lemonade already detects MTP and DFlash GGUF companions, but currently returns them only as a flat `draft_files` list. When a repository contains multiple draft variants, the CLI refuses to select one and GUI clients have no deterministic way to know which draft belongs to the selected main model.

This change resolves the best draft companion per GGUF variant and exposes it as `draft_file`.

The selection is generic and follows llama.cpp's model-sidecar resolution behavior rather than containing model- or repository-specific rules:

- prefer the closest shared directory with the selected main model
- prefer an exact quantization match
- otherwise prefer the closest quantization bit width

The existing `draft_files` field remains available for compatibility.

The CLI now consumes the per-variant `draft_file` when available and retains the existing single-entry `draft_files` fallback for compatibility.

Regression coverage includes multiple draft companions and quantizations without hard-coding behavior for a specific model repository.

Related to #3290

## Scope

- [x] This PR addresses one clear issue or change.
- [x] I reviewed the full diff myself before submitting.
- [x] I removed unrelated local changes.
- [x] I kept refactoring separate unless it is required for this change.

## Testing

- [ ] Code builds without errors locally.
- [ ] I tested this change locally.
- [x] I described the testing performed below.

_Testing details:_

```bash
cmake --build --preset default --target test_hf_variants
ctest --test-dir build -R '^HfVariantsTest$' --output-on-failure
```

Regression cases cover:

- a single draft companion
- multiple MTP quantizations
- selection of the matching draft for a selected main-model quantization
- fallback to the closest available quantization when an exact draft quantization does not exist
- preservation of existing DFlash detection
- preservation of the full draft_files compatibility list

The server patch can also be applied temporarily on GUI3_merging together with the GUI-only #3367 patch for an end-to-end download test.

## Documentation

- [x] Documentation is not affected by this change.
- [ ] Documentation is affected and has been updated.
- [ ] Documentation is affected but will be handled in a separate PR or issue.

## Breaking Changes

- [ ] This PR introduces breaking changes.
- [x] This PR does not introduce breaking changes.

## AI-assisted contribution

Please select one:

- [x] I used AI tools for this PR.
- [ ] I did not use AI tools for this PR.

If AI tools were used:

- [x] I verified that I understand the changes.
- [x] I checked for hallucinated APIs, unrelated changes, and incorrect assumptions.

PR #3382, opened 2026-08-27
feat(backends): declare per-backend model architecture/quant constraints

Geramy:
## Summary

Lets a backend declare which model shapes it is qualified to serve, so an experimental backend can be scoped to validated models without a separate recipe to carry the restriction. A backend adds `supported_models` to its descriptor; everything else is shared machinery.

```cpp
struct ModelConstraint {
    std::string architecture;       // "qwen3moe" or "qwen3_moe" — separators are folded
    std::set<std::string> quants;   // empty = any quant of that architecture
};
std::vector<ModelConstraint> supported_models;
```

Enforced at three tiers, all calling one predicate (`model_constraints_allow`):

| Tier | Where | Behavior |
|---|---|---|
| Load | `Router::load_model` | Requires `Supported`; refuses `Unsupported` **and** `Unknown`. Runs before the load mutex, so backends that override `WrappedServer::load()` cannot bypass it. |
| Pre-download | `GET /pull/variants?recipe=` | Drops unserveable variants; reports `constraint` and `filtered_out`. |
| Search | `GET /registry/search?recipe=` | Drops results whose architecture the registry reports and the backend cannot serve. |

Compatibility is tri-state — `Supported` / `Unsupported` / `Unknown`. The discovery tiers pass `Unknown`, so incomplete registry metadata never hides a serveable model; the load gate requires `Supported`, so a renamed GGUF or an `hf_load` model with no local metadata is refused rather than waved through. Worth knowing: Hugging Face's *search* endpoint reports no architecture for GGUF repos (verified — neither `config=true` nor `full=true` returns it), so the search tier is inert against HF today; the per-repo endpoint behind `/pull/variants` *does* return `gguf.architecture`, which is where narrowing actually happens.

`supported_models` defaults to empty, so **every existing backend is unchanged** and `recipe=` is optional on both endpoints. The only response change when it is omitted is an additive `architecture` field on search results; nothing is filtered.

### Example

Scoping hrx (#3374) to Qwen3-MoE Q4_K_M — `src/cpp/include/lemon/backends/hrx/hrx.h`:

```cpp
inline const BackendDescriptor descriptor = {
    /*recipe*/          "llamacpp-hrx",
    /*display_name*/    "HRX GPU (experimental)",
    // ... unchanged ...
    /*arg_variants*/    {},
    /*bin_variants*/    {"hrx"},
    /*config_extra*/    nlohmann::json::object(),
    /*streams_model_from_storage*/ false,
    /*supported_models*/ {{"qwen3_moe", {"Q4_K_M"}}},
};
```

Last field; `streams_model_from_storage` is named only because the aggregate is positional. No JSON changes.

Fixes #

## Scope

Two small consolidations were required rather than optional, both to avoid a second copy of existing logic: the quant tokenizer in `hf_variants.cpp` is exported as `extract_gguf_quant` (it was already the canonical one, just not in the header), and the registry architecture lookup moved into `model_registry` so search and variants share one implementation.

## Testing

_Testing details:_

- `ctest -L cpp-ci` — **62/62 pass** on macOS (arm64, Ninja/AppleClang).
- `python3 docs/tools/gen_backend_boilerplate.py --check` — no drift.
- **21 new unit tests**, extending existing binaries (no new CMake targets):
  - `test_model_registry` — architecture-hint extraction (`gguf.architecture` preferred over `config.model_type`); constraint matching including case-insensitivity, quant allow/refuse, and the `qwen3moe` vs `qwen3_moe` spelling fold; unknown-recipe and unconstrained-recipe no-ops; and search filtering proving a reported mismatch is **dropped** while an unreported architecture is **kept**.
  - `test_hf_variants` — `extract_gguf_quant` over plain, sharded, `UD-` prefixed, lowercase, and quant-less names; `filter_variants` narrowing 3 variants to the single qualified quant, dropping all on a mismatched architecture, and keeping all when the architecture is unreported or constraints are empty.
- Because no shipping backend declares constraints yet, the search and variant filters take constraints directly (the recipe overloads resolve the descriptor and delegate) so the **filtering paths are exercised for real** rather than only on the no-op path.
- Verified the Hugging Face metadata claims above against the live API.

## Documentation

`docs/api/lemonade.md`: the `recipe` parameter on both endpoints, the new `architecture` field on search results, the `constraint`/`filtered_out` response fields, and an explicit note that the search filter is inert for HF GGUF repos.

## Breaking Changes

New descriptor field defaults to empty; both query parameters are optional; `BackendDescriptor` is an aggregate and the field is appended last, so descriptors that omit it still compile.

## AI-assisted contribution

## Note for maintainers — UI

`src/app/src/renderer/ModelManager.tsx` gains the plumbing (`recipeQuery()` on the three search/variant fetches) but `targetRecipe` is left unset, so behavior is unchanged. Binding it to a backend selector is the one remaining step, and I left it unbound deliberately: UI changes are maintainer-owned, and the GUI3 branches are actively reworking this surface. Happy to drop the `.tsx` hunk entirely if you'd rather take it there.


jeremyfowers:
I’m not sure if we need this right now? We have HRX covered for the time being by declaring it as a recipe and listing specific supported models in server_models.json.

Geramy:
> I’m not sure if we need this right now? We have HRX covered for the time being by declaring it as a recipe and listing specific supported models in server_models.json.

I was saying it would be nice to lock it to the arch and quant so users can use other models and other versions like distilled, a lot of users use more than unsloth so for user adoption and testing it would be beneficial in my opinion. If we want to merge hrx first anyways that's fine, I can later update it to fit this PR.

This also fits other experimental engines which will also have similar requirements where they are highly optimized for only a handful of models.

fl0rianr (commented):
I like the direction, but I don't think this is merge-ready yet.

Main issue: the PR calls the load check a hard guarantee, but it currently isn't. `model_constraints_allow()` intentionally treats unknown architecture/quant as allowed, which is correct for search/pre-download but not for a final load gate. Quant is also inferred from the filename, so an incompatible GGUF renamed to `model.gguf` becomes `quant=""` and passes. `hf_load` can similarly have no local metadata at all and fail open.

There is also a structural issue with HRX specifically: `HrxServer` overrides `LlamaCppServer::load()` completely, so the new check in `LlamaCppServer::load()` is bypassed by the backend this feature is primarily meant for. The compatibility check should live in a common pre-load path that every constrained backend must pass.

I would also avoid calling `architecture + quant` a model-shape constraint. `qwen3_moe + Q4_K_M` allows every Qwen3 MoE of that quant, while HRX is currently qualified much more narrowly. If we want arbitrary repos/fine-tunes of the same supported shape, the constraint should encode the actual relevant GGUF shape, not just architecture.

Smaller issues:

* `/registry/search` now always adds `architecture`, so responses are not actually byte-identical when `recipe` is omitted.
* `/pull/variants?recipe=llamacpp-hrx` filters for HRX but still returns `"recipe": "llamacpp"`, which is confusing for clients.
* This needs a rebase onto current `main`; #3368 changed the same variant/quant code substantially.

I think the clean model is a tri-state compatibility result: `supported / unsupported / unknown`. Search and variant discovery keep `unknown`; the final load path for a constrained backend should require `supported`. Then add one real constrained-backend integration test covering wrong architecture, wrong quant/shape, and unknown metadata.

fl0rianr (commented):
The two original blockers are fixed well, but I still see two issues before I'd call this a hard compatibility guarantee:

1. The load gate still derives the quant from `model_info.resolved_path()`. That is filename/path metadata, not GGUF metadata. A Q8_0 file named `foo-Q4_K_M.gguf` would be accepted as Q4_K_M, and a quant token in a parent path can also influence classification. For a hard load guarantee, the final check should use the actual GGUF contents rather than the path name.

2. `architecture + quant` still isn't really a model-shape constraint. HRX is currently qualified specifically for `Qwen3-30B-A3B-Instruct-2507-Q4_K_M`, while `qwen3_moe + Q4_K_M` admits any Qwen3 MoE with that quant, regardless of dimensions/expert layout. If the goal is to allow compatible fine-tunes/repos while preserving backend qualification, the constraint needs enough GGUF shape metadata to identify the actually supported model shape.

Once those two points are addressed, I don't see a fundamental problem with the overall design anymore.

Geramy:
@fl0rianr this will be converted into a draft PR as per @jeremyfowers

PR #3391, opened 2026-08-27
fix(gui3): move setLogLevel off the retired /v1/log-level route

### LINKED ISSUE #3389: GUI3: stop using deprecated endpoints
Lemonade v11.8 removed 4 endpoints, GUI3 uses 2 of them. Replace.

jeremyfowers:
## Summary

`main` branch removed 4 undocumented endpoints. I did a sweep of the GUI code and found that there is only 1 instance of any of those being used. This PR migrates that 1 instance.

Deliberately not in scope: any non-gui code. 

Fixes #3389 

## Scope

## Testing

Confirmed the server gets the signal:

image

_Testing details:_

- `npx tsc --noEmit -p tsconfig.json` in `src/app` — clean.
- Verified the replacement contract against this branch's server: `POST /internal/set` routes to `handle_config_set` → `RuntimeConfig::set`, which handles the `log_level` key (`server.cpp`, `reconfigure_application_logging`) and returns `{"status":"success","updated":{...}}` (`runtime_config.cpp`). The new return value reads `updated.log_level` and falls back to the requested level.
- `setLogLevel`'s return type is unchanged, so its only caller (`LogViewer.tsx:278`) needs no change.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

### DIFF
--- a/src/app/src/api.ts
+++ b/src/app/src/api.ts
@@ -1916,10 +1916,12 @@ class LemonadeAPI {
   }
 
   async setLogLevel(level: string): Promise<{ status: string; level: string }> {
-    return this._json<{ status: string; level: string }>('/api/v1/log-level', {
-      method: 'POST',
-      body: { level },
-    });
+    const data = await this.setRuntimeConfig({ log_level: level });
+    const updated = (data.updated as Record<string, unknown> | undefined) ?? {};
+    return {
+      status: String(data.status ?? 'success'),
+      level: typeof updated.log_level === 'string' ? updated.log_level : level,
+    };
   }
 
   // ── Log stream (WebSocket) ──────────────────────────────────────


fl0rianr (commented):
Looks good overall, but I think there’s one auth issue we should fix before merging. setLogLevel() now goes through /internal/set, but setRuntimeConfig() still uses the default auth: 'api'. If LEMONADE_ADMIN_API_KEY differs from LEMONADE_API_KEY, this will send the regular API key and fail with 401. Since /internal/* is meant to use admin auth, I’d fix setRuntimeConfig() (and ideally getRuntimeConfig() too) to explicitly use auth: 'admin'. A small test with different API/admin keys would also cover this nicely. Otherwise the change looks clean to me.

fl0rianr (commented):
Looks good overall, but I think there’s one auth issue we should fix before merging. setLogLevel() now goes through /internal/set, but setRuntimeConfig() still uses the default auth: 'api'. If LEMONADE_ADMIN_API_KEY differs from LEMONADE_API_KEY, this will send the regular API key and fail with 401. Since /internal/* is meant to use admin auth, I’d fix setRuntimeConfig() (and ideally getRuntimeConfig() too) to explicitly use auth: 'admin'. A small test with different API/admin keys would also cover this nicely. Otherwise the change looks clean to me.

kpoineal:
@Geramy Do you agree with Florians assessment? If so I can make that change and we can get this merged.

Geramy:
> @Geramy Do you agree with Florians assessment? If so I can make that change and we can get this merged.

Some questions:
@claude Answer these questions.
Can the user with the user API see the logs without an admin key?
Does setRuntimeConfig block the regular api user to see logs at all or it only limits?

jeremyfowers:
I wont have time to come back to this for a while. Someone else feel free to take over.

Geramy:
> **Claude finished @Geramy's task in 1m 39s** —— [View job](https://github.com/lemonade-sdk/lemonade/actions/runs/33556493239)
> 
> ### Answers
> * [x]  Read PR diff and `api.ts` auth plumbing
> * [x]  Trace server-side auth for `/logs/stream`, `/api/v1/log-level`, `/internal/*`
> * [x]  Answer the two questions
> 
> **Q1: Can a user with the regular API key see the logs without an admin key?**
> 
> **Yes.** Viewing logs is a completely different path from what this PR touches. `LogViewer` reads logs over the WebSocket `/logs/stream` ([`api.ts:1953`](https://github.com/lemonade-sdk/lemonade/blob/6146687/src/app/src/api.ts#L1953)), and the client puts the **regular** key in the `api_key` query param ([`api.ts:1153`](https://github.com/lemonade-sdk/lemonade/blob/6146687/src/app/src/api.ts#L1153)). Server-side, the WS upgrade accepts either credential:
> 
> ```c++
> // src/cpp/server/websocket_server.cpp:412
> if (token_str == api_key_ || (!admin_api_key_.empty() && token_str == admin_api_key_)) {
>     out_authenticated = true;
> ```
> 
> Full matrix for reading logs vs. changing the log level:
> 
> env config	view logs (`/logs/stream`)	read/set log level (`/internal/*`)
> neither key set	open	open (`admin_api_key_` empty → no gate, `server.cpp:948`)
> only `LEMONADE_API_KEY`	regular key	**regular key works** — `admin_api_key_ = api_key_` (`server.cpp:651`)
> only `LEMONADE_ADMIN_API_KEY`	open (`api_key_` empty, `websocket_server.cpp:395`)	admin key required
> both, **distinct**	regular key OR admin key	**admin key only**
> So the only configuration where anything changes is the last row — and even there, log _viewing_ is unaffected.
> 
> **Q2: Does `setRuntimeConfig` block the regular API user from seeing logs at all, or does it only limit?**
> 
> **It only limits, and narrowly.** `setRuntimeConfig` → `POST /internal/set` ([`api.ts:1893`](https://github.com/lemonade-sdk/lemonade/blob/6146687/src/app/src/api.ts#L1893)), which falls under the admin gate at [`server.cpp:947-955`](https://github.com/lemonade-sdk/lemonade/blob/6146687/src/cpp/server/server.cpp#L947). It has no bearing on the log stream. The blast radius is exactly two controls in the Logs panel:
> 
> * `getLogLevel()` → `GET /internal/config` — **already** on `/internal/*` before this PR ([`api.ts:1890`](https://github.com/lemonade-sdk/lemonade/blob/6146687/src/app/src/api.ts#L1890)), so reading the current level was admin-gated already.
> * `setLogLevel()` → now `POST /internal/set` — this PR moves the write onto the same gate.
> 
> Both fail gracefully. `LogViewer.tsx:258` swallows the read failure with `.catch(() => {})` (dropdown just shows the default), and `LogViewer.tsx:277-281` `console.error`s the write failure. Logs keep streaming either way; the user just can't change the server-side verbosity from the UI.
> 
> ### On Florian's / Codex's point
> Their finding is correct, and there are actually **two** distinct things behind it:
> 
> 1. **Client sends the wrong credential (a real bug, worth fixing).** `_headers` picks `this.apiKey` under the default `auth: 'api'` ([`api.ts:1038`](https://github.com/lemonade-sdk/lemonade/blob/6146687/src/app/src/api.ts#L1038)), so an explicit admin key the user typed into the client is ignored on this path → 401. Note the `adminApiKey` getter already falls back to the regular key (`api.ts:832-834`), so switching to `auth: 'admin'` is **strictly a superset** — it's identical in the single-key case and only helps when a distinct admin key exists. Zero regression risk. `getRuntimeConfig()` has the same pre-existing bug, so fixing both (as Florian suggests) is the right call.
> 2. **Privilege change that the client fix cannot undo.** In the both-keys-distinct config, a user holding only `LEMONADE_API_KEY` could previously change the log level via `POST /api/v1/log-level` and now cannot, because `/internal/*` is admin-only by design. That's inherent to retiring the route, not a client bug. It seems defensible to me — mutating server-wide config is reasonably an admin operation — but it is a behavior change worth a line in the PR description so it isn't a surprise later.
> 
> So: 👍 to making `getRuntimeConfig`/`setRuntimeConfig` use `auth: 'admin'`, with the caveat that it fixes #1 only.
> 
> [Fix this →](https://claude.ai/code?q=On%20branch%20fix%2Fretire-params-log-level-callers%20in%20src%2Fapp%2Fsrc%2Fapi.ts%2C%20make%20getRuntimeConfig()%20(line%201889)%20and%20setRuntimeConfig()%20(line%201893)%20explicitly%20pass%20auth%3A%20'admin'%20to%20_json%2F_fetch%2C%20since%20they%20call%20%2Finternal%2Fconfig%20and%20%2Finternal%2Fset%20which%20are%20gated%20by%20LEMONADE_ADMIN_API_KEY%20server-side.%20Currently%20they%20use%20the%20default%20auth%3A%20'api'%20and%20send%20the%20regular%20API%20key%2C%20causing%20401%20when%20the%20admin%20key%20differs.%20Run%20npx%20tsc%20--noEmit%20-p%20tsconfig.json%20in%20src%2Fapp%20to%20verify.&repo=lemonade-sdk/lemonade)
> 
> No code was changed — this was a question, not an implementation request. • branch [`fix/retire-params-log-level-callers`](https://github.com/lemonade-sdk/lemonade/tree/fix/retire-params-log-level-callers)

so your saying its required to make this change or it won't work? @claude

PR #3393, opened 2026-08-27
fix: recover cache and models lost by 11.8.0 FHS split (#3387)

### LINKED ISSUE #3387: Almost everything lost after upgrade from 11.7.0 to 11.8.0 by apt upgrade
### Platform

Linux/Ubuntu

### Lemonade Version

11.8.0

### GPU / APU Model

_No response_

### Component

lemond (server)

### Bug Description

previously installed rocm-stable shown as not installed in 11.8.0.
previously workable configuration under /var/lib/lemonade/.cache/lemonade/config.json ignored by 11.8.0. previously downloaded models shown as No by command 'lemonade list' in 11.8.0.

is 11.8.0 a serious version?

### Steps to Reproduce

with Ubuntu 26.04, lemonade version 11.7.0 installed, apt update then apt upgrade

_No response_

### Log Output

```shell

```

### Additional Context

_No response_

superm1:
## Summary

The 11.8.0 FHS split (#3028) relocated the systemd service's config, models, and binaries all at once — config to `StateDirectory` (`/var/lib/lemonade`), models to `HF_HOME` (`/var/cache/huggingface`), binaries to `CacheDirectory` (`/var/cache/lemonade`) — but the startup migration only moved JSON files and read them from the *new, empty* cache dir. So on upgrade, existing installs found no config and no downloaded models ("almost everything lost").

This fixes it in two parts:

1. **`migrate_legacy_paths()`** recovers the pre-3028 home-based (`$HOME/.cache/lemonade`) config JSON into the config dir and the cache payload (downloaded backends, registry blobs) into the relocated `CacheDirectory`, using `rename()` with a recursive-copy fallback across filesystems. It never overwrites existing target files.

2. **Models are no longer relocated.** `HF_HOME=/var/cache/huggingface` is dropped from the systemd unit, so models resolve via `HOME` (`/var/lib/lemonade/.cache/huggingface` for the service) and stay exactly where 11.7.0 left them. This avoids a cross-filesystem copy of a potentially tens-of-GB model set (which could fail on a full `/var/cache`) and needs no migration at all. The postinst no longer creates `/var/cache/huggingface`, and the configuration docs + `migrate-to-systemd.sh` example are updated to match.

The migration is a no-op for plain-user, user-service, macOS, and Windows installs (those layouts didn't move).

Fixes #3387

## Scope

## Testing

_Testing details:_
- Built `lemond` and `test_config_dir_migration` with the `default` preset (Ninja/Release).
- `test_config_dir_migration` passes, including: `test_systemd_relocation_recovers_cache_and_models` (config + downloaded backends migrate into the state/cache dirs while a model under `~/.cache/huggingface` is left untouched) and `test_relocation_preserves_existing_target_files` (a current `config.json` is not clobbered by a stale legacy copy).
- Pre-commit hooks pass (comment-slop, trailing whitespace, EOF).

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:


abn (approved):
Thanks @superm1! This cleanly addresses the upgrade regression from #3028.

Verified the migration flow:
- Preserves Hugging Face snapshot symlinks across filesystem boundaries via `copy_symlinks`.
- Non-destructive behavior when target files exist (retains target and prevents clobbering).
- Clean no-op for Windows, macOS, and un-relocated user installs.
- C++ migration test suite passes with full environment isolation.

LGTM! 🚀

jeremyfowers (changes requested):
This hotfix needs to address the following problems as well:
1. Moving terabytes of model data without at least clearly informing the user is wrong; asking permission would be ideal.
2. Ensuring that the new HF_HOME location is on the same drive as the prior one

jeremyfowers:
Additional linux maintainers @bitgamma @sofiageo @kenvandine please test and sign off as well.

abn (changes requested):
Found one cross-platform scoping issue we should address before merging:

**HF-hub migration is ungated by platform / systemd context**
In `path_utils.cpp`:
```cpp
const fs::path legacy_hub = path_from_utf8(default_hf_cache_dir());
const fs::path new_hub = path_from_utf8(resolve_hf_cache_dir());
```
`default_hf_cache_dir()` resolves on macOS and Windows as well as Linux. If any user on macOS, Windows, or non-systemd Linux sets `HF_HOME` or `HF_HUB_CACHE` in their environment (e.g. for Python/transformers), `legacy_hub != new_hub` evaluates to true, causing `move_tree_into` to migrate and delete their default `~/.cache/huggingface/hub`.

Could we gate the HF model migration to only run when legacy cache relocation is actually active (e.g., when `get_legacy_cache_dir()` is non-empty and differed from `cache_dir`, or under Linux systemd context)?

**Minor polish:**
- In `move_tree_into()`, `fs::directory_iterator(src, ec)` populates `ec` on read failure, but `ec` is not checked before proceeding (if reading `src` fails, `drained` remains `true`).

abn:
> This hotfix needs to address the following problems as well:
> 
>     1. Moving terabytes of model data without at least clearly informing the user is wrong; asking permission would be ideal.
> 
>     2. Ensuring that the new HF_HOME location is on the same drive as the prior one

If `HF_HOME` is explicitly set that would be respected. I would assume that folks who use mounts have this set explicitly. Asking for a permission in a startup flow might be difficult and rather friction inducing in unattended boxes. Just my 2c.

superm1:
I decided to pull our `HF_HOME` setting and directory creation.  I think that's the most pragmatic thing to not migrate models based on above discussion.

abn (approved):
Thanks @superm1! Dropping the `HF_HOME` relocation and leaving models in `$HOME/.cache/huggingface` is much cleaner given the circumstances and avoids any risk of moving or deleting model data across platforms as was flagged earlier.

Verified the updated branch:
- Recovers JSON configuration to `StateDirectory` and backend binaries to `CacheDirectory`.
- Keeps Hugging Face models untouched in place.
- All migration unit tests pass.

LGTM! 🚀

jeremyfowers:
Before merging this PR, please address https://github.com/lemonade-sdk/lemonade/issues/3395, show that it would have caught the bug, merge the test to main, and show that this PR passes that test.

ckuethe:
> Thanks @superm1! Dropping the `HF_HOME` relocation and leaving models in `$HOME/.cache/huggingface` is much cleaner given the circumstances and avoids any risk of moving or deleting model data across platforms as was flagged earlier.
> 
> Verified the updated branch:
> 
> * Recovers JSON configuration to `StateDirectory` and backend binaries to `CacheDirectory`.
> * Keeps Hugging Face models untouched in place.
> * All migration unit tests pass.
> 
> LGTM! 🚀

wait, does that mean my migrated machines are going to have another migration?

bitgamma:
`/var/cache/huggingface` makes more sense long term. Considering part of the userbase has already migrated I don't think migrating back is a good idea.

jeremyfowers:
> > Thanks @superm1! Dropping the `HF_HOME` relocation and leaving models in `$HOME/.cache/huggingface` is much cleaner given the circumstances and avoids any risk of moving or deleting model data across platforms as was flagged earlier.
> > Verified the updated branch:
> > 
> > * Recovers JSON configuration to `StateDirectory` and backend binaries to `CacheDirectory`.
> > * Keeps Hugging Face models untouched in place.
> > * All migration unit tests pass.
> > 
> > LGTM! 🚀
> 
> wait, does that mean my migrated machines are going to have another migration?

Yes, set `HF_HOME=/var/cache/huggingface` prior to installing v11.8.1 and you should have a smooth experience.

jeremyfowers (commented):
(written by CC... obviously)

The recovery leg needs to be gated on the systemd relocation, otherwise this introduces a new data-relocation surprise for anyone who passes an explicit cache dir.

The new migration decides whether to move data by asking one question: is the folder lemond is using right now different from the old standard folder? But those differ for two different reasons:

1. The package upgrade relocated things — the case this PR is fixing, where moving is correct.
2. The user simply asked lemond to use a different folder — where nothing should move.

The code can't tell them apart, so it treats both as case 1. Concretely, `lemond /tmp/scratch` on Linux drains the user's real `~/.cache/lemonade` (JSON to the config dir, `bin/` and the rest into the scratch dir). Someone pointing lemond at a scratch folder for a quick test silently relocates their actual install — the same class of surprise as 11.8.0. Our own CI hits this: `install-lemonade-deb` starts lemond with `CACHE_DIR="./"`.

`get_legacy_cache_dir()` is non-empty for every Linux user, not just the service, so the `legacy_cache_path != new_cache_path` check isn't enough on its own. `cli_parser.cpp:40` already tracks whether the user supplied the argument (`cache_dir_opt_->count() > 0`) — skipping the legacy leg when they did, or gating on `CACHE_DIRECTORY`/`STATE_DIRECTORY` being set (the actual relocation trigger), would limit this to upgrades.

Two smaller things while you're in here:

- The PR description doesn't match the diff. It describes a fallback hub recorded for `BackendOps::resolve_checkpoint_path` to retry against, and says the FHS `/var/cache/huggingface` model location is retained. The diff does the opposite and simpler thing — it removes `Environment=HF_HOME` from the unit so models resolve back to `$HOME/.cache/huggingface` and never move (no `backend_ops.cpp` change in the PR). The simpler design is the better one, but the description will land in the release notes as-is.
- `move_tree_into` passes `ec` to `fs::directory_iterator` and never checks it. On a permission error the range comes back empty, `drained` stays `true`, and the caller proceeds to remove the legacy dir. No data loss (the remove fails on a non-empty dir), but recovery silently does nothing with no warning.

src/cpp/server/utils/path_utils.cpp
```diff
@@ -257,6 +257,101 @@ std::string resolve_hf_cache_dir() {
...
+
+    // Pre-3028 systemd installs kept everything under $HOME/.cache. Recover the
+    // cache payload (downloaded backends, registry blobs) into the relocated
+    // cache dir and the JSON into the config dir.
+    const std::string legacy_cache = platform()->get_legacy_cache_dir();
+    if (!legacy_cache.empty()) {
+        const fs::path legacy_cache_path = path_from_utf8(legacy_cache);
+        const fs::path new_cache_path = path_from_utf8(cache_dir);
+        std::error_code exists_ec;
+        if (legacy_cache_path != new_cache_path &&
```
jeremyfowers, line 353: This condition is true both when the systemd upgrade relocated the cache and when the user just passed an explicit cache dir on the command line. `get_legacy_cache_dir()` returns `$HOME/.cache/lemonade` for every Linux user, so `lemond /tmp/scratch` drains the real install into the scratch dir.

Gate on the relocation itself — `CACHE_DIRECTORY`/`STATE_DIRECTORY` being set — or skip the legacy leg when `cache_dir_opt_->count() > 0` (see `cli_parser.cpp:40`).

kenvandine (commented):
## Review

Built `lemond` and `test_config_dir_migration` locally and ran the new tests — they pass. Traced the full path chain (postinst → `lemond.service.in` → `path_linux.cpp` → `path_utils.cpp`) and confirmed the design is internally consistent: pre-#3028 systemd installs had `HOME=/var/lib/lemonade` (set via `useradd -d`), so `$HOME/.cache/lemonade` (legacy cache/config) and `$HOME/.cache/huggingface` (models) both already lived there. Dropping the `HF_HOME=/var/cache/huggingface` env var (2nd/3rd commits) means models never actually moved — the "recovery" is really just reverting the broken redirect, which is a safe, minimal hotfix. `migrate_legacy_paths()` correctly recovers config JSON + downloaded backend binaries via rename-with-copy-fallback, never clobbers existing target files (covered by test), and is a no-op elsewhere (plain-user, macOS, Windows).

**One regression risk worth flagging before merge:** `contrib/debian/lemonade-server.postinst` still runs `chown -R lemonade:lemonade /var/lib/lemonade` on every install/upgrade. Since models now permanently live under `/var/lib/lemonade/.cache/huggingface` (this PR removes the `/var/cache/huggingface` redirect), that recursive chown will walk potentially tens of GB of model blobs on **every future package upgrade**, not just once — a new perf/latency cost for `apt upgrade`/`dpkg configure` that didn't exist under the (buggy) 3028 design. Worth a fast follow-up (e.g. skip `.cache/huggingface` in the chown, or scope it to just the config subtree). Not a blocker for this urgent hotfix.

**Minor:** the PR description's opening paragraph (about recording a "fallback hub" and `BackendOps::resolve_checkpoint_path` retrying against it) describes an approach the later commits in this same PR abandoned — the final code just leaves models in place with no fallback-hub logic at all. Worth trimming the description so it doesn't mislead reviewers/changelog readers.

No functional bugs found in the C++ migration logic itself (`move_tree_into` merge/skip/drain logic checked for correctness). Safe to ship as a hotfix.

kenvandine (approved):
This looks good to me now, thanks! The snap wasn't affected by the issue, but I also built your branch into a snap and confirmed it still works as expected.

sofiageo (approved):
I tested this PR on Arch Linux by

* upgrading from 11.7.0 -> 11.8.1 (branch). models visible as before
* downgrading from 11.8.1 -> 11.8.0. models missing
* upgrading from 11.8.0 -> 11.8.1 (branch). models visible again

It looks to be fixing the issue.

jeremyfowers (approved):
In you all I trust. I will test the bleeding ppa when its out.

Issue #3394, opened 2026-08-27
[Known Issue] Lemonade 11.8.0 Migration - Configuration Loss - Instructions included


bitgamma:
### Component

lemond (server)

### Bug Description

Upgrading to Lemonade 11.8.0 changes where lemond looks for configuration, models, etc. There is no automatic migration. This is an issue. 

The manual steps to run to fix your installation are as follows, but please note **it is much better to wait for v11.8.1**:

```
sudo mv /var/lib/lemonade/.cache/lemonade/config.json /var/lib/lemonade/
sudo mv /var/lib/lemonade/.cache/lemonade/recipe_options.json /var/lib/lemonade/
sudo mv /var/lib/lemonade/.cache/lemonade/user_models.json /var/lib/lemonade/
sudo mkdir /var/cache/lemonade
sudo mv /var/lib/lemonade/.cache/lemonade/bin /var/cache/lemonade/bin
sudo mv /var/lib/lemonade/.cache/huggingface /var/cache/huggingface
sudo chown -R lemonade:lemonade /var/lib/lemonade
sudo chown -R lemonade:lemonade /var/cache/lemonade
sudo chown -R lemonade:lemonade /var/cache/huggingface
```

If you had an API key, you will now need to set it in `/etc/default/lemond`

## Update: v11.8.1 has released, see the comment below for more information https://github.com/lemonade-sdk/lemonade/issues/3394#issuecomment-5453720866

rugosi:
I saw after installing 11.8.0 that all drivers shall be downloaded + all models. But not really:
I stopped and disabled lemond service and started the lemond as previously (as user) and everything went in place, and worked like a charm.

mhantsch:
I stopped lemond, moved the files and directories as to the above instructions, and restarted lemond. The models were still not showing as downloaded. `lemonade update-models` also had no effect.

After a bit of investigation, I found that the 'mv's had created a double-nested directory path `/var/cache/huggingface/huggingface/hub/` (huggingface was in there twice). After clearing this, `lemonade list --downloaded` showed the models again.

jeremyfowers:
[Lemonade v11.8.1](https://github.com/lemonade-sdk/lemonade/releases/tag/v11.8.1) has been released, which contains a single change meant to mitigate the problems this issue discusses. If you were waiting to upgrade from 11.7.0 you should be good to go with v11.8.1.

[...]

PR #3396, opened 2026-08-27
ci: test PPA package upgrade path (config + model persistence)

### LINKED ISSUE #3395: Enhanced CI testing for the PPA
Lemonade CI builds the PPA but does not test it, which allowed bug https://github.com/lemonade-sdk/lemonade/issues/3394 to merge undetected.

Lemonade CI needs to test the PPA by:
1. Installing the last stable release
2. Applying some configuration and downloading a model
3. Install the new PPA build
4. Assert that the new installation is functional and that the saved configuration and model from step 2 are still present

superm1:
## Summary

CI builds the PPA `.deb` but never installs or upgrade-tests it, which let the 11.7.0→11.8.0 FHS-split regression (#3387 / #3394) — lost config + models on upgrade — merge undetected.

This adds an `upgrade-test` job to `.github/workflows/launchpad-ppa.yml` that exercises the real upgrade path per #3395:
1. Installs the last stable release from `ppa:lemonade-team/stable`.
2. Sets a config value (`lemonade config set log_level=debug`) and downloads a model (`lemonade pull Tiny-Test-Model-GGUF`, ~0.18 GB).
3. Installs the freshly-built `.deb` over it.
4. Asserts the upgraded server is healthy and that the config value and the model both survived.

A new composite action `.github/actions/start-lemond-service` launches `lemond` the way its systemd unit would — it parses the installed `lemond.service` for `StateDirectory`/`CacheDirectory`/`RuntimeDirectory`/`Environment=` (e.g. `HF_HOME`) and runs `lemond` as the `lemonade` user with that environment (no positional cache dir). This reproduces the systemd service layout where the migration logic lives, which the existing `install-lemonade-deb` (extract-only) and `test-lemonade-debian13-smoke` (positional cache dir) jobs do not. It re-reads the unit for each installed version, so the stable run and the upgraded run each use their own declared layout.

Fixes #3395

## Scope

## Testing

_Testing details:_
- Validated both YAML files and ran the repo's pre-commit GitHub Actions / Workflows validators (pass).
- Simulated the unit-file parsing in the composite action against `data/lemond.service.in`: resolves `STATE_DIRECTORY=/var/lib/lemonade`, `CACHE_DIRECTORY=/var/cache/lemonade`, `RUNTIME_DIRECTORY=/run/lemonade`, forwards `HF_HOME`, and safely no-ops the directives a pre-3028 unit omits.
- End-to-end validation requires GitHub Actions (real `apt` install + PPA + model pull), which cannot run locally. To run the job on this PR, apply the **`ci:distros`** label (the binary `.deb` and this job are gated to the merge queue / `ci:distros`, matching the existing distro jobs).

Notes for reviewers:
- Coverage is `noble` (24.04) and `resolute` (26.04).
- The job's ability to reproduce the *specific* #3394 bug depends on the current stable's layout. If the current stable predates the FHS split, this job will fail on `main` until the migration fix (#3393) lands; if stable is already ≥11.8, it passes today and guards future layout changes. Either way it is a lasting guard against upgrade data loss.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/.github/workflows/launchpad-ppa.yml
+++ b/.github/workflows/launchpad-ppa.yml
@@ -226,6 +226,148 @@ jobs:
           path: ${{ steps.build_binary.outputs.output-dir }}/*.deb
           retention-days: 30
 
+  upgrade-test:
+    name: Upgrade test (${{ matrix.codename }})
+    needs: [prepare-matrix, package]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    # Needs the binary .deb, which only builds in the merge queue or on a
+    # ci:distros-labelled PR (see the `package` job's build steps).
+    if: github.event_name == 'merge_group' || contains(github.event.pull_request.labels.*.name, 'ci:distros')
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - release: "26.04"
+            codename: resolute
+    container:
+      image: ghcr.io/lemonade-sdk/lemonade/build-environment:ubuntu${{ matrix.release }}
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    env:
+      HF_TOKEN: ${{ secrets.HUGGINGFACE_ACCESS_TOKEN }}
+      DEBIAN_FRONTEND: noninteractive
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Configure git safe directory
+        run: git config --global --add safe.directory $(pwd)
+
+      - name: Download new build (.deb)
+        uses: actions/download-artifact@v7
+        with:
+          name: lemonade-${{ matrix.codename }}-deb
+          path: new-deb
+
+      - name: Install prerequisites
+        shell: bash
+        run: |
+          set -e
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            ca-certificates curl procps software-properties-common util-linux
+
+      - name: Install last stable from PPA
+        id: stable
+        shell: bash
+        run: |
+          set -e
+          add-apt-repository -y ppa:lemonade-team/stable
+          apt-get update
+          apt-get install -y lemonade-server
+          echo "Installed stable version: $(dpkg-query -W -f='${Version}' lemonade-server)"
+
+      - name: Start stable lemond
+        uses: ./.github/actions/start-lemond-service
+        with:
+          port: '13305'
+          log-name: 'lemonade-server-stable.log'
+
+      - name: Seed config and download a model
+        shell: bash
+        run: |
+          set -e
+          # Persist a benign, non-default config value (does not affect binding).
+          lemonade config set log_level=debug
+
+          # Download the purpose-built tiny CI model, retrying transient HF errors.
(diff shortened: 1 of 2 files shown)


superm1:
This is good that it failed; that means it flagged properly.  Once https://github.com/lemonade-sdk/lemonade/pull/3397 and https://github.com/lemonade-sdk/lemonade/pull/3393 merge will make sure it goes green as part of a rebase.

superm1:
@jeremyfowers here's your config upgrade test

jeremyfowers (changes requested):
I would like to review this

.github/workflows/launchpad-ppa.yml
```diff
@@ -226,6 +226,163 @@ jobs:
...
+            ca-certificates curl procps software-properties-common util-linux
+
+      - name: Install last stable from PPA
+        id: stable
+        shell: bash
+        run: |
+          set -e
+          add-apt-repository -y ppa:lemonade-team/stable
+          apt-get update
+          # Skip gracefully on releases the stable PPA does not build for yet
```
jeremyfowers: This seems like a dangerous policy. Ubuntu releases are not frequent, can we hardcode the releases we want to test into the job matrix instead of allowing the test logic to quietly skip testing?

jeremyfowers:
Marked rfc:required for training purposes, but in reality this was agreed to offline not in a discussion.

.github/workflows/launchpad-ppa.yml
```diff
@@ -226,6 +226,163 @@ jobs:
...
+    # Needs the binary .deb, which only builds in the merge queue or on a
+    # ci:distros-labelled PR (see the `package` job's build steps).
+    if: github.event_name == 'merge_group' || contains(github.event.pull_request.labels.*.name, 'ci:distros')
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - release: "26.04"
+            codename: resolute
+    container:
```
jeremyfowers, line 244: This seems to contradict the PR body:

> Notes for reviewers:
>    Coverage is noble (24.04) and questing (25.10).

jeremyfowers (changes requested):
This looks really nice, we can merge as soon as my 2 comments are addressed. Checked the logs myself and everything 🫡

.github/workflows/launchpad-ppa.yml
superm1: It changed due to us deleting 25.10 from EOL (see offline discussion).   I'll update PR body.

jeremyfowers (approved):
sry for slow review, LGTM!

fyi I edited the PR body to remove the claim about 24.04 being part of the test matrix, since it isn't.

jeremyfowers:
> [73](https://github.com/lemonade-sdk/lemonade/actions/runs/35616344698/job/106389391449#step:8:474)

> stderr: Error: Failed to download llama-server from: https://github.com/ggml-org/llama.cpp/releases/download/b10825/llama-b10825-bin-ubuntu-x64.tar.gz - Download failed after 6 attempts.

Good grief, github doesnt want to serve 50 MB artifacts today?

PR #3398, opened 2026-08-27
Merge 'main' into 'GUI3_merging'

fl0rianr:
## Summary

Merge current `main` into `GUI3_merging` and reconcile GUI3 with the server/API changes that landed on main since the branch diverged.

The main cleanup here is removing some GUI3-specific server functionality that is now redundant and moving the GUI onto the APIs already provided by main.

## Scope

Changes made while resolving the merge:

- Remove the GUI3-specific `/api/v1/load/command` endpoint and its server handler/declaration.
- Use the existing per-model `/api/v1/models/{id}/options` endpoint with `dry_run: true` to resolve effective load settings.
- Update the GUI effective-settings preview to consume the resolved `effective` options instead of the removed command/args response.
- Keep GUI3 custom-model metadata required by the UI: `display_name`, `component_roles`, and `custom_tools`.
- Drop the redundant serialized `custom: true` flag; GUI3 already identifies user/custom models from the server-owned model metadata.
- Drop redundant server changes that are already covered by current main rather than carrying parallel GUI3 implementations.
- Preserve the existing GUI3 behavior and model configuration layering while adapting it to the current main API.

## Testing

_Testing details:_

- Rebased/merged against current `main` and reviewed the conflict resolutions individually.
- Ran `npm run typecheck` while reconciling the frontend API changes.
- Reviewed remaining GUI3 references to the removed load-command response/API.
- Effective Settings work correctly

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:


fl0rianr:
@claude review. pre-commit fail is fine since the target is to reduce main changes of the PR beside src/app

jeremyfowers:
`+19,271 -3,122` the diff is basically impossible to review as a human. I could throw agents at it, but before I do I want to ask: is there a more gradual way to accomplish this?

fl0rianr:
> `+19,271 -3,122` the diff is basically impossible to review as a human. I could throw agents at it, but before I do I want to ask: is there a more gradual way to accomplish this?

I think it is possible to review. There was a merge conflict I solved that manually. I know every line I touched.
In order to do so the best option was to remove code sections of GUI3 server changes.

I did look exactly at the diff and removed those sections.

This removal had to have a slight GUI3 adaption easy to review since it's only a change under src/app

Then I tested it and it's safe to merge, also claude says so. There are some changes I could add or do as a follow-up.

PR #3404, opened 2026-08-27
ci(sdcpp): raise lemond global_timeout to 3600s for validation legs

jeremyfowers:
## Summary

The sd-cpp validation legs start `lemond` against a fresh CI cache dir, so they inherit the default `global_timeout` of 600s. The `Validate cpu` leg's Flux-2-Klein-4B 1024x1024 generation runs **419-520s across 11 recent runs** (mean 471s, max 520s) — only 13% below that ceiling. This writes `{"global_timeout": 3600}` into the cache dir before launching `lemond`, giving the CPU leg real headroom.

This is a CI-side configuration change only; no server behavior changes.

## Scope

## Testing

_Testing details:_ Verified against a local `lemond` that a partial `config.json` in the cache dir is deep-merged with defaults and takes effect — `lemonade --port <p> config` reports `global_timeout 3600` (and `1` for a control run written with `{"global_timeout": 1}`). Also confirmed the file the workflow writes is parsed correctly with a UTF-8 BOM and CRLF line endings, since `Set-Content -Encoding utf8` under Windows PowerShell 5.1 emits a BOM. YAML validated.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/.github/workflows/validate_sdcpp.yml
+++ b/.github/workflows/validate_sdcpp.yml
@@ -372,6 +372,10 @@ jobs:
           New-Item -ItemType Directory -Force -Path $logsDir | Out-Null
           New-Item -ItemType Directory -Force -Path $cacheDir | Out-Null
 
+          # A 1024x1024 CPU generation runs 419-520s against lemond's 600s default.
+          Set-Content -Path (Join-Path $cacheDir "config.json") `
+            -Value '{"global_timeout": 3600}' -Encoding utf8
+
           $stdoutLog = Join-Path $PWD "$logsDir\lemond.stdout.log"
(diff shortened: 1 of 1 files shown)


fl0rianr (approved):
Looks good to me

One non-blocking info: the existing sd-cpp comment saying `0` is used to “avoid timeout” is stale. `0` now means use the global default; `kNoTimeout` is `-1`. Not caused by this PR though.

Approved from my side.

fl0rianr:
We need to land https://github.com/lemonade-sdk/lemonade/pull/3453 before this one due to a test failing - Actually the test worked correct the server operation is flawed.

PR #3405, opened 2026-08-27
fix(sdcpp): stage HIP runtime by version and fail validation on a CPU fallback

### LINKED ISSUE #3401: sd-cpp validation: fail loudly when the rocm backend is requested but no ROCm device is active
The `rocm-stable` leg of the sd-cpp validation workflow silently falls back to CPU when ggml reports `no ROCm-capable device is detected`, so the job passes while validating the wrong backend.

`sd-server` (or the validation harness) should fail the job when the requested backend does not initialize, instead of falling back to CPU.

- https://github.com/lemonade-sdk/lemonade/actions/runs/33084860623/job/98565289531
- https://github.com/lemonade-sdk/lemonade/actions/runs/33068459126/job/98507016318
- https://github.com/lemonade-sdk/lemonade/actions/runs/33065352735/job/98496534626

ramkrishna2910:
## Summary

The `rocm-stable` leg of the sd-cpp validation workflow has been passing while generating every image on CPU. Two independent bugs, plus a third that fixing them uncovered.

- **The ROCm backend never initialized.** `sdcpp_server.cpp` force-copies TheRock's `amdhip64_7.dll` next to `sd-server.exe`. When the AMD driver ships a newer HIP runtime in System32, that stale copy shadows it (the exe dir precedes System32 in the loader search order), `hipGetDeviceCount()` returns 0, and ggml falls back to CPU. This is #3213, fixed for llama.cpp in #3217 via `BackendUtils::stage_therock_hip_runtime()` — that PR scoped the fix to llama.cpp. This uses it for sd-cpp too.
- **Nothing checked which backend actually came up.** `validate_sdcpp.py` only asserted HTTP 200 plus a decodable PNG, so the fallback was invisible. It now asserts the active ggml backend from the captured server log via `--server-log`, which the workflow passes.
- **Uncovered by the above: ROCm 7.14.0's rocBLAS is broken on gfx1151** (#3409). The `rocm-stable` leg is therefore non-blocking, so it reports without gating merges.

Evidence from run [33084860623](https://github.com/lemonade-sdk/lemonade/actions/runs/33084860623/job/98565289531) — SD-Turbo at 1024x1024 took **105.7s** on `rocm-stable` against 159.6s on `cpu` and 10.0s on `vulkan`, with byte-identical PNGs across three consecutive runs.

## Verification on this PR

The fix works. The `rocm-stable` leg now reports:

```
[Info] (BackendUtils) System32 amdhip64_7.dll is at least as new as TheRock's; staged it at ...
ggml_cuda_init: found 1 ROCm devices (Total VRAM: 44920 MiB):
  Device 0: AMD Radeon(TM) 8060S Graphics, gfx1151 (0x1151), Wave Size: 32
[DEBUG] Initializing backend: ROCm0
... total params memory size = 1929.50MB (VRAM 1929.50MB, RAM 0.00MB)
[OK] sd-server initialized ['ROCm0'] on rocm-stable
```

Weights now land in VRAM instead of RAM, and the new assertion confirms it.

The leg still fails, on the newly visible ROCm regression. Isolated locally on a Radeon 8060S (gfx1151), holding the HIP runtime constant at the driver's `System32\amdhip64_7.dll` 10.0.3679 and varying only the rocBLAS/hipBLAS libraries:

| sd-cpp build | ROCm libraries | Result |
|---|---|---|
| `master-721-8caa3f9` | 7.13.0 | generates, ~11 it/s on GPU |
| `master-827-97d2990` (current pin) | 7.13.0 | generates on GPU |
| `master-827-97d2990` (current pin) | 7.14.0 | `hipErrorInvalidDeviceFunction` |
| `master-721-8caa3f9` | 7.14.0 | `hipErrorInvalidDeviceFunction` |
| `master-827-97d2990` | 7.14.0 + 7.13.0 Tensile kernels | `hipErrorInvalidDeviceFunction` |

It tracks the ROCm version, not the sd-cpp build, and is not the Tensile kernel data. llama.cpp is unaffected — its `rocm (stable)` and `rocm (nightly)` legs pass on the same gfx1151 runner in this PR's run. Full detail and repro in #3409.

## Notes on the assertion

- sd-server logs its ggml device selection to stdout, which lemond inherits into the files the workflow already redirects, so no new capture path was needed.
- It runs once at the end of a leg: sd-server's stdout is fully buffered when redirected to a file, so the early device line is only reliably present after a run's worth of output.
- A log with **no** evidence fails too. An unverified leg silently passing is the failure this exists to prevent.
- The pure log parsing lives in `test/utils/sdcpp_backend_log.py` so it is unit-testable without the server or `requests`.
- A matrix leg with no `blocking` value gates by default, so new legs are blocking unless explicitly opted out.

Follow-up not in scope: lemond itself still cannot see this fallback, so users get a silent CPU downgrade with no warning. Worth a separate issue.

Fixes #3401

## Scope

## Testing

_Testing details:_

- Compiled the changed translation unit: `MSBuild lemonade-server-core.vcxproj /t:ClCompile /p:SelectedFiles=...\sdcpp_server.cpp` — clean.
- New unit tests: `python -m unittest test.test_sdcpp_backend_log -v` — 13 tests pass. Wired into the `pre-commit` job in `docs_and_style.yml`.
- Replayed the assertion against the real `sdcpp-server-logs-*` artifacts from run 33084860623: `rocm` -> FAIL with the HIP reason quoted, `vulkan` -> PASS `['Vulkan0', 'Vulkan0']`, `cpu` -> PASS `['CPU', 'CPU']`.
- Confirmed in CI on this PR that the staging fix brings up `ROCm0` with weights in VRAM (see above).
- Reproduced and isolated the ROCm 7.14.0 regression locally on gfx1151 with `sd-cli.exe`, per the matrix above.
- Verified on Windows that a file held open by `Start-Process -RedirectStandardOutput` is readable concurrently, which the in-run check depends on.
- `black` and `python tools/check_comment_slop.py` are clean.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/.github/scripts/render_sdcpp_pr_body.py
+++ b/.github/scripts/render_sdcpp_pr_body.py
@@ -49,6 +49,11 @@ def read_records(labels: list[str]) -> list[dict[str, Any]]:
         for item in data:
             if not isinstance(item, dict):
                 raise SystemExit(f"{path} contains a non-object record")
+            # Non-image diagnostics (e.g. the active-backend assertion) share the
+            # array but are not per-model/size results, so they must not count
+            # toward the pass tallies or the expected-image assertion.
+            if item.get("check"):
+                continue
             item.setdefault("label", label)
             item["artifact"] = f"sdcpp-validation-{label}"
             records.append(item)
--- a/.github/workflows/docs_and_style.yml
+++ b/.github/workflows/docs_and_style.yml
@@ -79,6 +79,9 @@ jobs:
       - name: Run the comment-slop unit tests
         run: python -m unittest test.test_comment_slop -v
 
+      - name: Run the sd-cpp active-backend assertion unit tests
+        run: python -m unittest test.test_sdcpp_backend_log -v
+
       - name: Run pre-commit on the changed files
         # Scoped to what the change touches; running repo-wide would fail on drift that
         # predates it and is not the contributor's to fix.
--- a/.github/workflows/validate_sdcpp.yml
+++ b/.github/workflows/validate_sdcpp.yml
@@ -33,7 +33,11 @@ env:
   # lemonade-sdk fork and may intentionally be on a different release tag.
   SDCPP_BASE_UPDATE_BACKENDS: cpu,vulkan,rocm-stable,metal
   SDCPP_CUDA_UPDATE_BACKENDS: cuda
-  SDCPP_VALIDATED_LABELS: cpu,vulkan,rocm-stable
+  # rocm-stable is omitted while its leg is non-blocking (see #3409): the update
+  # PR body asserts image evidence for every label listed here, and the leg
+  # cannot produce any. It stays in SDCPP_BASE_UPDATE_BACKENDS, so the pin is
+  # still updated and the body reports it as updated-but-not-validated.
+  SDCPP_VALIDATED_LABELS: cpu,vulkan
   SDCPP_TEST_PROMPT: "A small glass of lemonade on a clean table, product photo, high detail"
   SDCPP_TEST_MODELS: "SD-Turbo-GGUF,Flux-2-Klein-4B"
   SDCPP_TEST_SIZES: "512x256,1024x1024"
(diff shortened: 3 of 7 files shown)


superm1 (approved):

superm1:
There's two more libraries you should try to stage, the kpack one and comgr one.  If you can reproduce locally please try copying those in too.

jeremyfowers (changes requested):
> The rocm-stable leg is therefore non-blocking, so it reports without gating merges.

> The fix works. The rocm-stable leg now reports:

@ramkrishna2910 these two statements seem contradictory to me, can you reconcile them? If the fix works why would we want the test to be non-blocking?

Also, can you please rewrite the PR description by dropping the archeology and focusing on the new state of testing and ROCm support? I'm finding it too difficult to infer this information from the current description.

jeremyfowers:
Also can you clarify if sd-cpp:rocm is completely broken on Windows + gfx1151 on main right now, or not? And does this PR fix it completely, or not? I went into this thinking we had a stale driver in our CI lab or something.

If its actually completely broken we need to escalate fixing it or drop it from the support table or something...

jeremyfowers:
sdcpp:rocm works correctly on my Windows gfx1151 with the 26.1.1 driver, which calls into question some of the premise here. Moving PR back to draft until you can thoroughly investigate @ramkrishna2910

PR #3408, opened 2026-08-27
fix(test): pre-download the model in test_gpu_hang_retry before the timed load

### LINKED ISSUE #3407: Address  test_concurrent_gpu_hang_recovery
details tbd

example failure:   https://github.com/lemonade-sdk/lemonade/actions/runs/33109845467/job/98653274108

jeremyfowers:
## Summary

`test_gpu_hang_retry.py` spawns its own `lemond` and times `/load` at 60 s to observe the reload, but never pre-downloads the model — so on a cold cache that call is measuring bandwidth and times out. This applies the standard `pull_model_with_retry()` pattern in `setUp`, and additionally aligns the macOS CI job's two servers on one HF cache via `HF_HUB_CACHE` so they stop maintaining separate copies.

Fixes #3407

## Scope

The test fix is what removes the flake. The workflow change is CI hygiene for the same root cause — on macOS the `.pkg` daemon takes its cache from the plist while test-spawned servers inherit the job's `HF_HOME`, so models get downloaded twice. Happy to split it out.

## Testing

_Testing details:_ `black` 26.1.0 clean, module compiles, pre-commit hooks pass. The macOS path needs a `macos-latest` runner, so this PR carries `ci:macos` to opt the job back in; the added verify step asserts via `ps eww` that the daemon actually received `HF_HUB_CACHE` and fails the job otherwise.

Note on the workflow half: the daemon writes the shared cache as root and test-spawned servers write it as `runner`. A mode on the cache root would not reach the model directories the daemon creates later, so the step sets an `everyone allow ... file_inherit,directory_inherit` ACL on `/Users/Shared/hf-cache` and echoes `ls -lde` for the record. This is a throwaway single-tenant runner, so the ACE is deliberately broad.

## Documentation

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:


abn (approved):
Looks good to me. Pre-warming the cache via `pull_model_with_retry()` in `setUp()` eliminates the race with the 60s `/load` timeout, and setting `HF_HUB_CACHE` in the launchd domain along with the inherited ACL cleanly aligns the daemon and test-spawned instances on macOS without modifying the plist template.

Once CI is green you are good to go.

superm1 (approved):
Can we also persist the cache between runs?  I think GH allows up to 10GB to be persisted in cache, right?

jeremyfowers:
> Can we also persist the cache between runs? I think GH allows up to 10GB to be persisted in cache, right?

@superm1 please feel free to open an issue to discuss this. Caching is challenging because if we are haphazard it is very easy to overflow the 10 GB allowance. The proposal needs to identify specific artifacts, like the test GGUF files, that are likely to stay within the cache budget and are likely to be cache hits.

PR #3414, opened 2026-08-28
fix(routing): don't drop the whole router policy over one hardware-filtered classifier

### LINKED ISSUE #2748: [Router] collection.router policy dropped when a classifier component is hardware-filtered or unregistered
Follow-up from the capability-validation work in #2727 (found in a comprehensive review). Two related resolver issues, both about how a `collection.router`'s classifier/semantic component type is resolved at cache-build time.

## 1. build_cache drops the whole policy when a classifier component is filtered/missing

`ModelManager::build_cache()`'s routing-policy resolver (`model_manager.cpp` ~1938) is a bare `models_cache_.find(name)` with no fallback. `filter_models_by_backend()` runs **before** the cache is populated, so a classifier component that is filtered out (no NPU, insufficient memory, not-yet-downloaded, hardware mismatch) is absent from `models_cache_` → the resolver returns `nullopt` → the parser throws "unresolvable type" → the **entire** routing policy is dropped (`route_policy` stays null), including its deterministic rules.

At runtime `route_collection_request` then sees no `route_policy` and returns `nullopt`, leaving the request's `model` as the collection name → it tries to load a `collection.router` recipe with no backend → the collection is unroutable.

**Impact:** a router whose classifier model needs hardware the host lacks takes down the *whole* router, not just the classifier condition. This is a consequence of the (intentional, per review) error-on-unresolvable behavior; the open question is whether a filtered/unavailable classifier component should **degrade gracefully** (that classifier's conditions fail-open per `on_error`, deterministic rules keep working) rather than drop the policy.

## 2. Resolver asymmetry: registry-backed collections are validated only on the weak path

`validate_collection_request()` has an inline-`models[]` fallback (resolves types even for unregistered inline components); `build_cache()` does not, and its reconstructed policy doc doesn't even carry `models[]`. Pointer-only / registry-backed router collections return early from `validate_collection_request` (before the routing parse), so their classifier capability is validated **only** at build_cache — the path in (1). Net: a policy can pass registration validation and later be silently dropped at build_cache, or never be validated at registration.

## Also worth a look (lower severity, from the same review)
- **Silent fail-open masks all of the above.** A mis-typed classifier surfaces as an `UnsupportedOperationException` error response → `parse_classifier_scores` throws → `on_error` (default `match_false`) → the classifier becomes a silent no-op and all traffic goes to `default_model`, with no runtime signal. Worth a telemetry/log flag.
- **Display inconsistency:** `ollama_api.cpp` / `cli/model_selection.cpp` / `cli/main.cpp` still derive type via `get_model_type_from_labels()` directly rather than the backend-aware `get_deployment_model_type()` added in #2727, so an onnxruntime/sd-cpp model can report a different type in `/api/tags` / `/api/show` / CLI than its runtime type.

Not blocking #2727 (whose capability check is correct for the resolvable cases); filing so the graceful-degradation design decision and the asymmetry are tracked.

Bekhouche:
## Summary

`ModelManager::build_cache()`'s routing-policy resolver only consulted the post-filter `models_cache_`, so a classifier component whose backend needs hardware this host lacks (no NPU, insufficient memory, not yet downloaded) was indistinguishable from a name that isn't a model at all. The parser threw "unresolvable type" and the entire policy was dropped, including unrelated deterministic rules that had nothing to do with the unavailable classifier.

Snapshot every model's type before `filter_models_by_backend()` removes some of them, and fall back to that snapshot in the resolver. A filtered classifier's type is still known, so the policy still parses; that specific classifier now fails only at `evaluate()` time, where its own `on_error` already applies — the existing, correct mechanism for an unavailable backend, just no longer bypassed by a parse-time abort.

Fixes #2748 (the two confirmed bugs: `build_cache`'s no-fallback resolver, and the "policy silently dropped" consequence of its asymmetry with `validate_collection_request`'s inline-model fallback). The lower-priority "worth a look" items from that issue are unrelated to this fix: one (display inconsistency across `ollama_api.cpp`/CLI) was already resolved as a side effect of #3099; the other (telemetry for silent classifier fail-open) is untouched here.

## Scope

## Testing

_Testing details:_
Added a regression test (`test_model_manager_collection_validation.cpp`) that registers a `collection.router` whose classifier is backed by a `ryzenai-llm` model (reliably hardware-unsupported on any CI host) alongside an unrelated deterministic rule, and asserts `route_policy != nullptr` after `build_cache()` runs. Verified the test actually catches the bug: reverted just the `model_manager.cpp` fix and confirmed the same test fails via `ctest` (0/1 passing), then restored it (1/1 passing). Ran the full `cpp-ci` suite via `ctest -L cpp-ci` on WSL/GCC — 66/66 pass. Built and ran the affected test binary natively on MSVC (VS 2026) — passes. `tools/check_comment_slop.py` is clean against `main`.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -3037,16 +3037,19 @@ void ModelManager::build_cache() {
         }
     }
 
-    // Snapshot every known model's type before hardware filtering removes some
-    // of them from all_models (#2748): a router's classifier component can name
-    // a model this host can't run right now (no NPU, insufficient memory, not
-    // yet downloaded), which must not be indistinguishable from a name that
-    // isn't a model at all. The routing-policy resolver below falls back to
-    // this so that case still resolves a type and parses; the classifier then
-    // fails at evaluate() time instead, where on_error already applies per rule.
+    // Snapshot type + bare-name alias before hardware filtering removes some
+    // models from all_models (#2748): the resolver below falls back to this so
+    // a router referencing a currently-unavailable model still parses; that
+    // classifier then fails only at evaluate() time, where on_error applies.
     std::map<std::string, ModelType> pre_filter_types;
+    std::map<std::string, std::string> pre_filter_bare_aliases;
     for (const auto& [name, info] : all_models) {
         pre_filter_types[name] = info.type;
+        std::string bare = name;
+        if (auto canon = parse_canonical_id(name)) {
+            bare = canon->bare_name;
+        }
+        pre_filter_bare_aliases.emplace(bare, name);
     }
 
     // Step 2: Filter by backend availability. This is the full-registry pass, so
@@ -3099,23 +3102,22 @@ void ModelManager::build_cache() {
     cache_valid_ = true;
 
     // Parse each collection.router model's routing policy now, while the cache
-    // and its alias map are fully built and the lock is still held. The parser's
-    // component resolver needs only alias resolution, which is a direct lookup
-    // into public_model_aliases_ here - exactly what resolve_model_name() does,
-    // minus its own build_cache()+lock. Calling resolve_model_name() here would
-    // re-lock this non-recursive mutex and deadlock, so the lookup is inlined.
+    // and its alias map are fully built and the lock is still held (inlined
+    // rather than resolve_model_name()/get_model_info(), which would re-lock
+    // this non-recursive mutex). Both resolvers fall back to the pre-filter
+    // snapshots above (#2748) so a component this host can't run right now
+    // still resolves and the policy still parses; that classifier then fails
+    // only at evaluate() time, where on_error already applies.
     RoutingPolicyParseOptions policy_options;
     policy_options.resolve_component =
-        [this](const std::string& name) -> std::optional<std::string> {
+        [this, &pre_filter_bare_aliases](const std::string& name) -> std::optional<std::string> {
         auto it = public_model_aliases_.find(name);
-        return it != public_model_aliases_.end() ? it->second : name;
+        if (it != public_model_aliases_.end()) {
+            return it->second;
+        }
+        auto pre_it = pre_filter_bare_aliases.find(name);
+        return pre_it != pre_filter_bare_aliases.end() ? pre_it->second : name;
     };
-    // Direct lookup into the map already being built, same rationale as
-    // resolve_component above: models_cache_ is populated and the lock is
-    // still held, so this avoids re-entering get_model_info()'s own locking.
-    // Falls back to pre_filter_types (#2748) for a component this host
-    // currently can't run: its type is still known, so the policy still
-    // parses instead of being dropped whole over one unavailable classifier.
     policy_options.get_model_type = [this, &pre_filter_types](
                                          const std::string& name) -> std::optional<ModelType> {
         auto it = models_cache_.find(name);
--- a/src/cpp/server/model_manager.cpp
+++ b/src/cpp/server/model_manager.cpp
@@ -3037,6 +3037,18 @@ void ModelManager::build_cache() {
         }
     }
 
+    // Snapshot every known model's type before hardware filtering removes some
+    // of them from all_models (#2748): a router's classifier component can name
+    // a model this host can't run right now (no NPU, insufficient memory, not
+    // yet downloaded), which must not be indistinguishable from a name that
+    // isn't a model at all. The routing-policy resolver below falls back to
+    // this so that case still resolves a type and parses; the classifier then
+    // fails at evaluate() time instead, where on_error already applies per rule.
+    std::map<std::string, ModelType> pre_filter_types;
+    for (const auto& [name, info] : all_models) {
+        pre_filter_types[name] = info.type;
+    }
+
     // Step 2: Filter by backend availability. This is the full-registry pass, so
     // it also refreshes the recipe availability side table used to hide backends
     // that have nothing runnable on this host.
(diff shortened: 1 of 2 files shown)


fl0rianr (commented):
I think the core fix is right, but there is still one edge case that looks unresolved.

`pre_filter_types` preserves the type, but component alias resolution is still rebuilt from the post-filter `models_cache_`. So if an imported component is stored as `user.npu-clf` in `components`, while the routing policy still refers to the public/bare name `npu-clf`, filtering the model removes the `npu-clf -> user.npu-clf` alias. The parser then fails the declared-component check before `get_model_type()` can use the new fallback.

The new test uses `user.npu-clf` in both places, so it does not cover this case.

I think we should also preserve/fallback to the pre-filter alias resolution here, and add a regression test with canonical `components` but a bare classifier model name.

Finally a nit, there is too much of commenting even if the pre-commit does not catch it - verbose comments tend to confuse AI and might diverge form what the code does in the future. Only explain briefly what the codes isn't already showing but is important.

Bekhouche:
@fl0rianr you're right. public_model_aliases_ is also rebuilt from models_cache_, so the bare-name alias was lost in the same way as the type.
Pushed a fix: resolve_component now falls back to a snapshot of the bare aliases taken before filtering, just like get_model_type does. I also added a regression test with the canonical ID in components and the bare name in the classifier, which is the case your example needed and the existing test did not cover.
Also trimmed the comments as you suggested. Ready for re-review.

fl0rianr (commented):
The original issue is fixed now and the new bare-name regression test covers the case I raised.

I think there is one remaining edge case though: `pre_filter_bare_aliases.emplace(bare, name)` does not preserve the existing alias precedence (`Registered > Imported > Builtin`). Also `resolve_component` checks the post-filter alias map first, so if e.g. `user.X` normally owns bare `X` but gets filtered while a builtin `X` survives, the routing policy can suddenly resolve `X` to the builtin instead of the filtered user model. That can either fail the declared-component check again or, worse, bind the classifier to a different model.

I think the policy resolver should use a real pre-filter alias snapshot with the same precedence rules as `rebuild_public_model_aliases_locked()`, and that mapping should win for names that existed before filtering.

Small test nit: please also make the test verify that the NPU model was actually filtered, otherwise it can become vacuously green on a future NPU-capable CI runner.

Bekhouche:
@fl0rianr  fixed. Extracted the precedence logic (Registered > Imported > Builtin) into compute_bare_name_precedence() and ran it against the pre-filter model set; the resolver now checks that first, so it can't silently rebind to a lower-precedence survivor. Also added an assertion to both tests that the NPU model is actually filtered before checking the policy still parses, so they can't go vacuously green on a future NPU-capable runner, per your nit. Ready for re-review.

fl0rianr (commented):
The precedence fix looks correct now, and the tests no longer go vacuously green.

There is still one alias path missing though: the pre-filter snapshot only preserves bare-name aliases, while `public_model_aliases_` also provides `builtin.X -> X` and `ModelInfo::input_aliases`.

For example, if builtin `X` is hardware-filtered and a router explicitly references `builtin.X`, `pre_filter_types` contains only `X`, `pre_filter_bare_aliases` only contains `X -> X`, and the post-filter alias map no longer contains `builtin.X -> X`. So `resolve_component("builtin.X")` stays `builtin.X`, `get_model_type()` still cannot resolve it, and the whole policy is dropped again.

I think the pre-filter snapshot needs to preserve the full input-alias resolution semantics, not only the bare-name winner. A regression test with a filtered canonical `builtin.X` would cover this nicely.

Bekhouche:
@fl0rianr Confirmed and fixed. Extracted the real alias algorithm into compute_model_alias_maps() and run it against both the pre- and post-filter model sets, so they can't drift apart again. Added a regression test for the builtin.X case.

fl0rianr (approved):
Thanks, LGTM now

SlawomirNowaczyk (changes requested):
The PR is solid, but I see some room for improvement:

## 1. Unify the two component resolvers so they provably can't diverge (correctness)

The issue #2748 root cause was an *asymmetry* between two resolvers. This PR fixes the one in `build_cache()` to use a pre-filter snapshot, but the second resolver in `validate_collection_request()` still uses live `resolve_model_name()` / `get_model_info()` plus an inline-`models[]` fallback:

```cpp
// build_cache() — now pre-filter snapshot (PR)
policy_options.resolve_component = [&pre_filter_aliases](...) { ... };

// validate_collection_request() — still live resolution + inline fallback
options.resolve_component = [this](const std::string& name) {
    try { return resolve_model_name(name); } catch (...) { return name; }
};
```

These two paths now differ along a *new* axis: `validate` sees inline-but-unregistered components; `build_cache` sees pre-filter aliases. A policy could pass one and be rejected by the other. Since the whole bug class is "two resolvers that are supposed to agree but don't," extracting a single shared resolver helper (parameterized by which model-set/fallback it consults) would close the door on the next drift instead of patching one side. See `model_manager.cpp:3231` and `model_manager.cpp:6267`.

## 2. Close the remaining half of #2748: telemetry for the silent fail-open

The PR explicitly leaves the issue's "telemetry for silent classifier fail-open" item untouched. After this fix the policy parses fine, then that classifier silently takes its `on_error`/`default_label` path at `evaluate()` time — the operator gets **no signal** that routing is degraded on this host. But `build_cache()` already knows, at parse time, exactly which classifier component is hardware-filtered (it's in `pre_filter_types`/`pre_filter_aliases` but not in `models_cache_`). A single one-time `LOG(WARNING, ...)` there naming the specific classifier that will fail-open (or a stat counter) makes the degraded routing observable and fully closes #2748 rather than half of it.

## 3. Skip the snapshot work when there are no router collections (performance)

`compute_model_alias_maps(all_models)` (an O(n log n) rebuild over the full registry) plus the `pre_filter_types` copy now run on **every** `build_cache()`. That function is called from ~15 sites — `register`, `delete`, `pull`, `/cloud/auth`, etc. — and the snapshots are only consumed inside the `is_router_collection_recipe` loop that follows. In the common case (no `collection.router` models registered), this is pure waste. Guard both snapshots behind a quick "any router-collection recipe present?" check so non-router hosts pay nothing.

## 4. Test strength

The three new tests only assert `route_policy != nullptr`. The PR's stated value is that *unrelated deterministic rules survive*. Consider evaluating the parsed policy and asserting the `code-remote` keyword rule still routes to `remote` while the filtered classifier rule falls back to its `default_label` — that directly verifies the behavior the fix promises, not just that parsing didn't abort.

Bekhouche:
@SlawomirNowaczyk  thanks, all four addressed.
`ctest -L cpp-ci` 66/66 on GCC, and built + run on MSVC.

SlawomirNowaczyk (approved):

PR #3417, opened 2026-08-28
fix(cloud): restore public model name in provider responses

jenbrewerworks:
## Summary
- Normalize cloud streaming/non-streaming responses to the public model name.
- Flush trailing SSE line buffer; set `[DONE]` from reassembled lines.
- Guard new C++ tests with `BUILD_TESTING AND EXISTS(...)`.
- Add regression tests for SSE buffering and model-name rewrite.

Replaces #2980 (closed after branch recreation while resolving conflicts).

Thanks @fl0rianr for the review feedback on the prior PR.


fl0rianr:
Thanks for going at this again, was already wondering. I hope you don't mind if I kindly ask for a rebase, you are 151 commits behind current main. After this and the resolved merge conflict this will be in a much easier state to review it.

fl0rianr:
Thanks for rebasing. Unfortunately the merge resolution seems to have brought a lot of old branch state back over current main.

The PR now reverts unrelated changes across CI, CMake, cloud auth/wire-format handling, telemetry/streaming logic, etc. — even the project version goes from 11.8.1 back to 11.6.0. There also appears to be a compile break where `TelemetryCallback` is still the current single-argument API, but `cloud_server.cpp` calls the old 5-argument form.

The actual model-name/SSE fix looks good now and addresses the previous review points. I think the safest path is to recreate/rebase from current main and reapply only the intended cloud/SSE changes plus the two tests, without carrying over the unrelated old state.

Let me know if you need help.

jeremyfowers (changes requested):
@jgmelber there appears to be some significant structural issue with this PR? Surely it isn't intended to be +1199 -2625 ?

Also please have a look at PR scoping. Is it really necessary to have both of these in a single PR? They seem like separate changes that should have different reviewers.

- Normalize cloud streaming/non-streaming responses to the public model name.
- Flush trailing SSE line buffer; set [DONE] from reassembled lines.

PR #3419, opened 2026-08-28
fix(server): allow same-origin HTTP requests matching request Host

abn:
## Summary

When hosting Lemonade on a non-loopback network interface (such as `0.0.0.0`) and accessing the built-in web application at `http://<LAN_IP>:<PORT>/app` or via mDNS `http://<hostname>.local:<PORT>/app`, browser requests send matching `Host` and `Origin` headers. Because LAN IP addresses and local hostnames were not in the loopback allowlist, requests previously failed with `403 Origin not allowed`.

This change strengthens origin validation using server self-identity enumeration ("enumerate, don't guess"):
1. **Rule Precedence**: Loopback & approved desktop schemes (Layer 1) -> Explicit allowlist (Layer 2, strictly authoritative without fallthrough) -> Same-origin validated against `server_self_set` (Layer 3) -> Default reject (Layer 4).
2. **Server Self Set (`server_self_set`)**: The server dynamically enumerates its active network interface IPs (IPv4 & IPv6 via `getifaddrs`/`GetAdaptersAddresses` checking `IFF_UP`), machine hostname, and `.local` mDNS name with a thread-safe 60-second TTL cache. Same-origin requests (`Origin` matching `Host`) are verified against `server_self_set`, enabling zero-config LAN and mDNS access while completely blocking DNS-rebinding attacks (e.g. `Host: evil.com` matching `Origin: http://evil.com` is rejected).
3. **WebSockets & Desktop Schemes**: WebSockets in `websocket_server.cpp` now share this validation pipeline via centralized `lemon::utils::resolve_allowed_origins()`, allowing same-origin WebSockets over LAN/mDNS while remaining protected against CSWSH. Native desktop application schemes (including `lemonade://`) and non-browser clients without `Origin` headers are explicitly supported.

### When Zero Configuration is Possible
- **Local & Loopback**: Loopback addresses (`localhost`, `127.0.0.1`, `[::1]`, `tauri.localhost`) are permitted automatically.
- **Native Desktop Apps**: Custom desktop URI schemes (`lemonade://`, `file://`, `app://.`, `vscode-webview://`, `jan://`, etc.) are permitted automatically.
[...]

PR #3420, opened 2026-08-28
feat(server): add allowed_origins to global configuration schema

abn:
## Summary

Adds `allowed_origins` to the global configuration schema in `config.json` and CLI (`lemonade config set allowed_origins="..."`), deprecates the `LEMONADE_ALLOWED_ORIGINS` environment variable, and unifies origin resolution across HTTP and WebSocket paths via `utils::resolve_allowed_origins()`.

**Key Behavior:**
- **Automatic Migration**: When `LEMONADE_ALLOWED_ORIGINS` is set at startup and `allowed_origins` is unset or empty in `config.json`, Lemonade automatically writes the value into `config.json` and emits a one-time migration warning.
- **Boot Precedence**: At startup, `LEMONADE_ALLOWED_ORIGINS` takes interim precedence over `config.json`. If `config.json` contains a non-empty value that conflicts with the environment variable, Lemonade warns of the conflict and instructs the user to unset `LEMONADE_ALLOWED_ORIGINS`.
- **Runtime Overrides**: Running `lemonade config set allowed_origins="..."` updates `config.json` and applies an in-memory session override immediately, allowing runtime changes to take effect even if the server was started with an environment variable set.
- **Recommended Action**: Unset or remove `LEMONADE_ALLOWED_ORIGINS` from your shell or service environment to manage allowed origins exclusively via `config.json` and the CLI.

## Scope

## Testing

_Testing details:_
- Unit test suite `test_config_allowed_origins.cpp` (38 assertions) covering defaults, getters, setters, type rejection, CLI arg parsing, dynamic runtime updates, active environment variable overrides, and disk migration scenarios.
- Unit test suite `test_origin_utils.cpp` (79 assertions) covering same-origin resolution, LAN/mDNS detection, desktop schemes, and explicit allowlists.
- Executed full `cpp-ci` test battery via `ctest -L cpp-ci` (69/69 test suites passed).
- Executed Python WebSocket telemetry security tests via `python test/server_websocket_telemetry.py` (12/12 passed).
- Validated doc/defaults synchronization with `python3 docs/tools/gen_backend_boilerplate.py --check`.

## Documentation

## Breaking Changes

## AI-assisted contribution

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3030,6 +3030,20 @@ if(BUILD_TESTING AND EXISTS "${_CONFIG_DOWNLOAD_RATE_LIMIT_TEST_SRC}")
     )
 endif()
 
+set(_CONFIG_ALLOWED_ORIGINS_TEST_SRC
+    "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_config_allowed_origins.cpp"
+)
+if(BUILD_TESTING AND EXISTS "${_CONFIG_ALLOWED_ORIGINS_TEST_SRC}")
+    add_executable(test_config_allowed_origins
+        test/cpp/test_config_allowed_origins.cpp
+    )
+    target_link_libraries(test_config_allowed_origins PRIVATE lemonade-server-core)
+    add_dependencies(test_config_allowed_origins copy_resources)
+    add_cpp_ci_test(ConfigAllowedOriginsTest CI ON
+        COMMAND test_config_allowed_origins
+    )
+endif()
+
 set(_RUNTIME_CONFIG_BACKEND_VALIDATION_TEST_SRC
     "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_runtime_config_backend_validation.cpp"
 )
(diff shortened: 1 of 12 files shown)


superm1 (approved):
I would say we should drop the environment variable support and then document in the release notes to move to config option instead.

abn:
> I would say we should drop the environment variable support and then document in the release notes to move to config option instead.

Personally, I would like to keep the env var, as it is of use for me. But if there is strong resistance to making this available as an env var, I can live with just adding another snippet to my ansible :smile:

superm1:
It's just we spent a bunch of time getting rid of them so it would be better to be consistent.

And yes I know I introduced this one. I should have done a config at the time and that's on me 😞.

Geramy (approved):

abn:
I have updated the branch with an additional commit that deprecates `LEMONADE_ALLOWED_ORIGINS` and adds automatic startup migration into `config.json`:

**Key updates:**
- **Automatic Migration**: When `LEMONADE_ALLOWED_ORIGINS` is present on startup and `allowed_origins` is unset or empty in `config.json`, Lemonade automatically writes the value into `config.json`.
- **Preserve Existing Config**: If `config.json` already contains a non-empty `allowed_origins`, it is left untouched and not overwritten.
- **Deprecation Warnings**: A single migration log is emitted when `config.json` is modified. If disk config is preserved and the environment variable is still present, a warning is logged noting that the environment variable takes interim precedence while set.
- **Documentation & Tests**: Updated configuration docs with the deprecation notice and interim precedence rules, and added migration test cases to `test_config_allowed_origins.cpp`.

abn:
@superm1 @Geramy added the deprecation change as an [independent commit for review](https://github.com/lemonade-sdk/lemonade/pull/3420/changes/8fb69bec0f7c14a0c12a091aab35d3a4ac6bf7a4).

Geramy (approved):

abn:
This will require a rebase once #3419 merges.

PR #3435, opened 2026-08-30
fix(anthropic): resolve model aliases on /v1/messages and the Ollama endpoints

### LINKED ISSUE #3433: Anthropic Messages endpoint does not resolve model aliases
### Platform

Linux/Ubuntu

### Lemonade Version

11.8.1

### GPU / APU Model

AMD R9700 AI Pro

### Component

lemond (server)

### Bug Description

`POST /v1/messages` returns 404 for a registered model alias. The same alias works on `POST /api/v1/chat/completions` (200), and `/v1/messages` works with the alias's target model ID (200).

Error body: `{"error":{"message":"model 'ALIAS' not found, try pulling it first","type":"not_found_error"}}`

The server appears to resolve the alias before rejecting it — the log shows `(OllamaApi) Auto-loading model: ALIAS` immediately before `Error 404: POST /v1/messages`.

The docs state "When an alias appears in API requests or CLI commands, Lemonade resolves the alias to its underlying target model ID", with no surface qualification.

Impact: Anthropic-API clients (e.g. Claude Code via `ANTHROPIC_BASE_URL`) cannot use aliases at all, and the failure reads as a missing model or bad credentials rather than an unresolved alias.

### Steps to Reproduce

```bash
BASE=http://localhost:13305
ALIAS=my-alias
MODEL=target-model-id
JSON='content-type: application/json'
AV='anthropic-version: 2023-06-01'
BODY='{"model":"%s","max_tokens":4,"messages":[{"role":"user","content":"hi"}]}'

# alias is registered
curl -s "$BASE/internal/aliases"
# {"aliases":[{"alias":"ALIAS","downloaded":true,"recipe":"llamacpp","target":"MODEL"}]}

# 1. Messages + alias -> 404
curl -s -w ' HTTP %{http_code}\n' -X POST "$BASE/v1/messages" -H "$JSON" -H "$AV" \
  -d "$(printf "$BODY" "$ALIAS")"
# {"error":{"message":"model 'ALIAS' not found, try pulling it first","type":"not_found_error"},"type":"error"} HTTP 404

# 2. Messages + target -> 200
curl -s -w ' HTTP %{http_code}\n' -X POST "$BASE/v1/messages" -H "$JSON" -H "$AV" \
  -d "$(printf "$BODY" "$MODEL")"
# {"content":[{"text":"...","type":"text"}],"model":"MODEL",...} HTTP 200

# 3. Chat completions + alias -> 200
curl -s -w ' HTTP %{http_code}\n' -X POST "$BASE/api/v1/chat/completions" -H "$JSON" \
  -d "$(printf "$BODY" "$ALIAS")"
# {"choices":[{"message":{"content":"...","role":"assistant"}}],"model":"MODEL"} HTTP 200
```

**Expected behaviour:**

`POST /v1/messages` resolves a registered alias to its target and answers, as the other inference surfaces do.

**Actual behaviour:**

`POST /v1/messages` with a registered alias returns 404 `not_found_error`, "model 'ALIAS' not found, try pulling it first". The ST /api/v1/chat/completions`, and `/v1/messages` returns 200 for the alias's target model ID.

### Log Output

```log
[Info]  (ModelManager) Loading "user_models.json"
[Info]  (OllamaApi) Auto-loading model: ALIAS
[Error] (Server) Error 404: POST /v1/messages
```

### Additional Context

llamacpp backend, alias target loaded and pinned at the time of the requests.

The log above was captured while an Anthropic-API client was hitting the endpoint; the alias appears to be resolved before the 404.

Documented behaviour, without surface qualification: "When an alias appears in API requests or CLI commands, Lemonade resolves trget model ID."

Client impact: Anthropic-API clients such as Claude Code (`ANTHROPIC_BASE_URL`) can only use `/v1/messages`, so aliases are unusports it as "There's an issue with the selected model (ALIAS). It may not exist or you may not have access to it", which pointsat credentials, base URL or routing rather than at alias resolution — a route-missing and a model-missing 404 are indistinguisha

Related: #3413 — same alias gap on `/metrics`.

boecks:
## Summary

`POST /v1/messages` returned 404 for a registered alias while the same alias
succeeded on `/v1/chat/completions`. Alias resolution lives in
`Server::normalize_and_resolve_request_model`, which only the OpenAI-surface
handlers call, and `OllamaApi` (which serves `/api/*` and `POST /v1/messages`)
held no reference to `AliasManager`. `OllamaApi` now takes an `AliasResolver`
callback and resolves the request model once per request, mirroring what
`Server` already does.

Fixes #3433

## Scope

The Ollama endpoints (`/api/chat`, `/api/generate`, `/api/embed`,
`/api/embeddings`) are fixed by the same change. They are methods on the same
class reached through the same unresolved name, so fixing only `/v1/messages`
would have meant special-casing one method of `OllamaApi` against its siblings.

Why a second helper rather than teaching `normalize_model_name` to resolve, which
would be a smaller diff: its three remaining callers are `handle_show`,
`handle_delete` and `handle_pull`. `POST /api/delete {"name": "<alias>"}` would
then delete the target model, which is a bad surprise for "delete the alias".
`pull` would be harmless and `show` would arguably improve, but the set resolved
here deliberately matches the set `Server::normalize_and_resolve_request_model`
already resolves (inference, load, unload) rather than widening it in a bug fix.

Resolution happens once, in the handler, rather than at each point that reads the
model name. Resolving in both the handler and the converter would read mutable
alias state twice, so an alias edited between the two reads could load one model
and route the request to another.

Two disclosures rather than changes:

- Responses now echo the resolved model name, not the requested alias. That
  matches what `/v1/chat/completions` already does for an alias, but it is a new
  behaviour on the Ollama compatibility layer, where upstream Ollama echoes the
  name it was given.
- `/api/show` returns 404 for an alias and `/api/tags` omits it, while
  `/api/v1/models` and `/api/v1/models/{id}` expose it. That behaviour predates
  this change, but this PR makes the asymmetry visible: chat, generate and embed
  now resolve while show does not. I have not verified whether the Ollama CLI
  probes `/api/show` before chatting, which would make it user-visible for
  `ollama run <alias>`. Happy to fold it in if you want it here rather than in a
  separate issue.

## Testing

_Testing details:_

- `ctest -L cpp-ci`: 67/67 pass.
- `python test/test_ollama.py`: 32/32 pass, including two new tests,
  `test_025b_anthropic_messages_resolves_alias` and
  `test_025c_ollama_chat_resolves_alias`. Both fail before this change.

Containers built from `main` and from this branch, alias
`ALIAS -> Tiny-Test-Model-GGUF`:

| request | main | this branch |
|---|---|---|
| `POST /v1/messages` with alias | 404 `model 'ALIAS' not found, try pulling it first` | 200 |
| `POST /api/chat` with alias | 404, same message | 200 |
| `POST /v1/messages` with target id | 200 | 200 |
| `POST /api/v1/chat/completions` with alias | 200 | 200 |

Also run against a live server with a 35B llamacpp/ROCm model behind an alias:
`/v1/messages` streaming and non-streaming, `/api/chat`, `/api/generate`, and a
chained alias all resolve to the final target.

## Documentation

`docs/guide/configuration/custom-models.md:341` already states that an alias in
"any API request" resolves to its target. This change makes the implementation
match.

## Breaking Changes

Requests that previously returned 404 now succeed. Nothing that previously
succeeded changes behaviour.

## AI-assisted contribution

### DIFF
--- a/src/cpp/include/lemon/ollama_api.h
+++ b/src/cpp/include/lemon/ollama_api.h
@@ -16,7 +16,10 @@ using json = nlohmann::json;
 
 class OllamaApi : public std::enable_shared_from_this<OllamaApi> {
 public:
-    OllamaApi(Router* router, ModelManager* model_manager);
+    using AliasResolver = std::function<std::string(const std::string&)>;
+
+    OllamaApi(Router* router, ModelManager* model_manager,
+              AliasResolver resolve_alias);
 
     // Must be called on a shared_ptr instance (uses shared_from_this internally)
     void register_routes(httplib::Server& server);
@@ -24,6 +27,7 @@ public:
 private:
     Router* router_;
     ModelManager* model_manager_;
+    AliasResolver resolve_alias_;
 
     // Endpoint handlers
     void handle_chat(const httplib::Request& req, httplib::Response& res);
@@ -44,6 +48,8 @@ private:
     void auto_load_model(const std::string& model, const json& request_options = json::object());
     static json extract_auto_load_options(const json& request);
     std::string normalize_model_name(const std::string& name);
+    // Not called from show/pull/delete: resolving there would act on the target.
+    void resolve_request_model(json& request_json);
     json build_ollama_model_entry(const std::string& id, const ModelInfo& info);
     json convert_openai_chat_to_ollama(const json& openai_response, const std::string& model);
     json convert_openai_delta_to_ollama(const json& openai_chunk, const std::string& model);
--- a/src/cpp/server/anthropic_api.cpp
+++ b/src/cpp/server/anthropic_api.cpp
@@ -1143,6 +1143,7 @@ void OllamaApi::handle_anthropic_messages(const httplib::Request& req, httplib::
         auto request_json = json::parse(req.body);
         std::vector<std::string> warnings;
 
+        resolve_request_model(request_json);
         std::string model = normalize_model_name(request_json.value("model", ""));
         if (model.empty()) {
             res.status = 400;
--- a/src/cpp/server/ollama_api.cpp
+++ b/src/cpp/server/ollama_api.cpp
@@ -149,8 +149,10 @@ static void map_ollama_options(const json& ollama_request, json& openai_req) {
     }
 }
 
-OllamaApi::OllamaApi(Router* router, ModelManager* model_manager)
-    : router_(router), model_manager_(model_manager) {
+OllamaApi::OllamaApi(Router* router, ModelManager* model_manager,
+                     AliasResolver resolve_alias)
+    : router_(router), model_manager_(model_manager),
+      resolve_alias_(std::move(resolve_alias)) {
 }
 
 void OllamaApi::register_routes(httplib::Server& server) {
@@ -232,6 +234,22 @@ std::string OllamaApi::normalize_model_name(const std::string& name) {
         return name.substr(0, name.size() - suffix.size());
     }
     return name;
+}
+
+// Once per request: a second lookup downstream could see a concurrent alias edit.
+void OllamaApi::resolve_request_model(json& request_json) {
+    if (!resolve_alias_ || !request_json.contains("model") ||
+        !request_json["model"].is_string()) {
+        return;
+    }
+    const std::string normalized = normalize_model_name(request_json["model"].get<std::string>());
+    if (normalized.empty()) {
+        return;
+    }
+    const std::string resolved = resolve_alias_(normalized);
+    if (resolved != request_json["model"].get<std::string>()) {
+        request_json["model"] = resolved;
+    }
 }
 
 // ============================================================================
@@ -710,6 +728,7 @@ void OllamaApi::handle_chat(const httplib::Request& req, httplib::Response& res)
     try {
         auto request_json = json::parse(req.body);
 
+        resolve_request_model(request_json);
         std::string model = normalize_model_name(request_json.value("model", ""));
         if (model.empty()) {
             res.status = 400;
@@ -846,6 +865,7 @@ void OllamaApi::handle_generate(const httplib::Request& req, httplib::Response& 
     try {
         auto request_json = json::parse(req.body);
 
+        resolve_request_model(request_json);
         std::string model = normalize_model_name(request_json.value("model", ""));
         if (model.empty()) {
             res.status = 400;
@@ -1307,6 +1327,7 @@ void OllamaApi::handle_embed(const httplib::Request& req, httplib::Response& res
     try {
         auto request_json = json::parse(req.body);
 
+        resolve_request_model(request_json);
         std::string model = normalize_model_name(request_json.value("model", ""));
         if (model.empty()) {
             res.status = 400;
@@ -1377,6 +1398,7 @@ void OllamaApi::handle_embeddings(const httplib::Request& req, httplib::Response
     try {
         auto request_json = json::parse(req.body);
 
+        resolve_request_model(request_json);
         std::string model = normalize_model_name(request_json.value("model", ""));
         if (model.empty()) {
             res.status = 400;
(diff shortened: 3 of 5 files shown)


abn:
Good catch on this, @boecks. The `/api/show` 404 you called out in the description is actually the clue here.

Across this PR, #3437 (MCP), and #3436 (metrics), the common issue is that `ModelManager` and `Router` have zero visibility into aliases. Because resolution only lives as a JSON-rewriting helper at the HTTP server edge, we have to keep threading callbacks into every new adapter.

For context on how we got here: in #2818, aliases were originally in `ModelManager`, but during review it was decided to move them out to `Server` so they wouldn't be tied to model definitions. That separation still makes complete sense: `aliases.json` shouldn't pollute model configs. But pushing lookup logic out to the HTTP perimeter is what created this gap.

We can keep both subsystems decoupled without the perimeter whack-a-mole: `ModelManager` can simply expose an optional resolver hook (`set_alias_resolver`), which `Server` wires up to `AliasManager` at startup. `ModelManager` stays completely agnostic of `aliases.json`, but `resolve_model_name`, `model_exists`, and `get_model_info` handle aliases naturally across all routes. That drops the callback plumbing here and in #3437, fixes `/api/show` without special-casing, and gives metrics a clean source for mappings.

cc @bitgamma: does a resolver hook on `ModelManager` feel like a clean middle ground to keep that separation?

If that sounds good to you both, @boecks would you be up for consolidating this into a single PR to replace the three? Happy to outline the hook changes if helpful.

bitgamma:
@abn adding a hook is just a more convoluted way of adding a reference to the AliasManager inside the ModelManager. I see the advantages of wiring it there but I think this should be part of a larger ModelManager split-up so that its many concerns are in self-contained components (I can see at least the model registry, the model downloader and the checkpoint resolver). AliasManager is a good example of a well-defined, self-contained component that can interact with the ModelManager. Want to write an RFC about this or should I do it?

PS: note that the anti-AI-prose guidelines went into effect.

abn:
> adding a hook is just a more convoluted way of adding a reference to the AliasManager inside the ModelManager.

Well I proposed the hook to keep majority of the resolution logic out of the `ModelManager`, as this was a pushback earlier. My first recommendation would have been to add the logic into it. And tackle refactor more broadly later. 

> I see the advantages of wiring it there but I think this should be part of a larger ModelManager split-up so that its many concerns are in self-contained components (I can see at least the model registry, the model downloader and the checkpoint resolver). AliasManager is a good example of a well-defined, self-contained component that can interact with the ModelManager. Want to write an RFC about this or should I do it

You can do it.

While I agree with you that this is likely where we want to get to, the question really is if it is the best short term course of action. My recommendation is as indicated above, do the minimal coupling now and broader refactor later.

boecks:
Sounds good, ping me when you've settled on the approach and I'll write it. Happy to consolidate the three PR (#3436 and #3437 and this one) into one PR if that's what you want, or leave them separate. And whenever the bigger refactor happens, I can be a test bench for it — Linux/ROCm, R9700 AI Pro, 35B through llamacpp, and real Anthropic-API and Ollama clients hitting it through aliases daily, so I can catch regressions on the surfaces these three touched.

abn:
Thanks for creating #3545 @bitgamma :tada:.

@jeremyfowers @bitgamma My recommendation from above still stands. I would suggest that we take the simplest route here, ie. add the coupling into `ModelManager` so it resolves this and the three other PRs in one-go as these are genuine bugs. And we can focus on refactor more holistically later. We should not block this on the acceptance and work that needs to be done for #3545. That said, I am happy with whatever direction you choose on this one.

bitgamma:
@abn on the other hand, the currently open PR come with a small, self contained solution to the problem. What speaks against merging them as-is?

abn:
> @abn on the other hand, the currently open PR come with a small, self contained solution to the problem. What speaks against merging them as-is?

It's scattered fixes and there might be more we haven't seen yet. This would also mean more things to keep an eye on and clean up rather than one thing to tidy in one place once you get to the refactor.

PR #3436, opened 2026-08-30
feat(metrics): relate model aliases to their target in /metrics

### LINKED ISSUE #3413: Expose model aliases in the Prometheus metrics
### Platform

Linux/Other (please specify)

### Lemonade Version

11.8.0 (still present on 11.8.1)

### GPU / APU Model

AMD R9700 AI Pro

### Component

lemond (server)

### Bug Description

A registered model alias is accepted as a model id by `/api/v1/models` and by `POST /api/v1/chat/completions`, but is absent from the Prometheus metrics.

`lemonade alias add` / `POST /internal/aliases` register an alias that is listed by `/api/v1/models` and served by chat completions. Every `lemonade_model_*` series on `/metrics`, however, is labelled with the canonical model name only. The alias appears nowhere, so a consumer that refers to a model by its alias cannot observe that model at all.

This makes an alias usable for inference but unusable for any check based on the metrics. In our case that is readiness gating: we warm a model and wait for `lemonade_model_loaded{model_name=...} == 1` before dispatching work. Pointed at an alias, that condition is never satisfied and the model reads as "no such model" rather than "not loaded yet". The same applies to any dashboard or alert written against an alias — it matches nothing, with nothing indicating why.

The effect is that adopting aliases costs the one thing they provide: every consumer still has to know the canonical name.

### Steps to Reproduce

1. Register an alias for a downloaded model: `lemonade alias add ALIAS MODEL`
2. Confirm it is registered: `lemonade alias list`
3. Confirm the alias is a valid model id in the API: `curl -s localhost:13305/api/v1/models | jq -r '.data[].id'` → the list contains both MODEL and ALIAS
4. Load the model BY ITS ALIAS, then read the metrics:
   ```bash
   curl -s localhost:13305/api/v1/chat/completions \
     -d '{"model":"ALIAS","messages":[{"role":"user","content":"hi"}],"max_tokens":1}'
   curl -s localhost:13305/metrics | grep lemonade_model_loaded
   ```
   → the request is served, but the only label present is `model_name="MODEL"`
5. grep the whole endpoint for the alias: `curl -s localhost:13305/metrics | grep ALIAS` → no match

(ALIAS and MODEL are placeholders; 13305 is the default API port.)

**Expected:** an alias that the API accepts as a model id is reachable in the metrics somehow — so that a consumer holding only an alias can find the corresponding `lemonade_model_*` series.

**Actual:** aliases are invisible on `/metrics`. Only canonical model names appear, on `lemonade_model_loaded`, `lemonade_model_info` and the rest of the `lemonade_model_*` family, and there is no series relating an alias to a model name.

We haven't looked at how the metrics are assembled, so no suggestion as to the right shape — a label, a separate mapping series, or something else entirely.

### Log Output

```console
$ lemonade alias list
Registered model aliases:
  ALIAS -> MODEL [llamacpp, downloaded=yes]

# The alias is a valid model id in the API:
$ curl -s localhost:13305/api/v1/models | tr ',' '\n' | grep '"id"'
"id":"MODEL"
"id":"OTHER-MODEL"
"id":"ALIAS"

# Nothing loaded yet:
$ curl -s localhost:13305/metrics | grep -E '^lemonade_(model_loaded|loaded_models)'
lemonade_loaded_models 0
lemonade_model_loaded{checkpoint="CHECKPOINT",device="gpu",model_name="MODEL",recipe="llamacpp",type="llm"} 0

# Load the model BY ITS ALIAS — served correctly, but the response body
# reports the canonical name, not the alias:
$ curl -s localhost:13305/api/v1/chat/completions -H 'Content-Type: application/json' \
    -d '{"model":"ALIAS","messages":[{"role":"user","content":"hi"}],"max_tokens":1}'
{"choices":[{"finish_reason":"length","message":{"content":"Hello","role":"assistant"}}],"model":"MODEL","object":"chat.completi

# Now loaded — and the metrics carry the canonical name only:
$ curl -s localhost:13305/metrics | grep -E '^lemonade_(model_loaded|loaded_models|model_info)'
lemonade_loaded_models 1
lemonade_model_info{checkpoint="CHECKPOINT",device="gpu",model_name="MODEL",recipe="llamacpp",type="llm"} 1
lemonade_model_loaded{checkpoint="CHECKPOINT",device="gpu",model_name="MODEL",recipe="llamacpp",type="llm"} 1

# The alias appears nowhere in the metrics:
$ curl -s localhost:13305/metrics | grep ALIAS
$
```

### Additional Context

Platform: Linux, running the official container image on Kubernetes (k3s).

Aliases themselves work correctly and persist across a restart (written to `aliases.json` in the server's config directory). Thi from the metrics.

Verified: loading the model via its alias does not change the emitted labels — the metrics still carry the canonical `model_name` only. The completion response body likewise reports the canonical name rather than the alias, so the server resolves alias to canonical at the edge and the alias survives only in the alias registry and the `/api/v1/models` listing.

That may well be intentional normalisation. The request is not that existing labels change, but that an alias be observable at all.

Correction to the original wording: an alias is **not** a first-class model id everywhere in the API. It resolves on `/api/v1/mocompletions`, and does **not** resolve on `POST /v1/messages` (404 `not_found_error`) — filed separately as #3433 — nor on`/metrics`, which is this issue.

boecks:
## Summary

Every `lemonade_model_*` series carries the canonical model name, because the
rows come from the Router snapshot, which is canonical by construction. No series
related an alias to the model it names, so a readiness check written against an
alias could not tell "not loaded yet" from "no such model". `/metrics` now emits
`lemonade_model_alias{alias,model_name}`, one sample per registered alias,
present whether or not the target has ever been loaded.

Fixes #3413

## Scope

Notes on the shape, since it was a real choice:

- An info-metric rather than duplicated series. An alias is a name for a model,
  not a second model. Emitting `lemonade_model_loaded` again under the alias name
  would make an alias indistinguishable from a model and would corrupt `count()`
  and `sum()` over those series. The mapping joins instead:
  `lemonade_model_loaded * on(model_name) group_right lemonade_model_alias`.
- Counters are not mirrored, so cumulative totals stay single-counted.
- Chains collapse to the final model. Unresolvable and cyclic aliases are
  omitted.
- An alias whose target does not exist locally is still exported. `POST
  /internal/aliases` does not require the target to be pulled, and the docs say
  an alias may target a model before it exists, so the row is deliberate: it is
  what distinguishes "this alias names a model that is not loaded" from "no such
  alias". It joins to nothing until the target is first loaded.
- Known limitation, same root: an alias whose target has never been loaded has no
  `lemonade_model_loaded` row to join against, because the Router only reports
  models it has observed. Emitting `loaded=0` for every registered model would
  change what `/metrics` means, so it is not in scope here.

## Testing

_Testing details:_

- `ctest -L cpp-ci`: 68/68 pass, including the new `MetricsAliasesTest`
  (`test/cpp/test_metrics_aliases.cpp`) covering a single alias, a chained alias,
  several aliases sharing a target, a cyclic alias, an empty alias name, a
  self-referential alias, and a null resolver.
- `python test/server_endpoints.py`: the new
  `test_002c_metrics_expose_model_aliases` passes. It asserts the mapping exists
  for a plain and a chained alias, that the target it names is a label the model
  series actually carries, and that the alias appears in neither
  `lemonade_model_loaded` nor the cumulative counters.
- Both new tests fail before this change.
- One unrelated failure in that suite,
  `test_012d_cloud_install_then_auth_then_chat` (`models_discovered: 0 != 1`),
  reproduces identically on unmodified `main` in the same environment. It needs a
  reachable cloud provider.

`curl -s $BASE/metrics | grep alias` returns nothing on `main`. On this branch:

```
lemonade_model_alias{alias="ALIAS",model_name="Tiny-Test-Model-GGUF"} 1
```

Also run against a live server with a 35B llamacpp/ROCm model behind an alias.
The mapping is present while no model is loaded, which is the case the issue is
about. Once loaded, the join from alias to target to `lemonade_model_loaded`
resolves to 1. A chained alias reports the final model, and an alias to a
never-loaded model still appears.

## Documentation

docs/guide/configuration/custom-models.md gains an "Alias Mapping in /metrics" subsection under the alias docs, explaining lemonade_model_alias_info and how to join it onto the model series.

## Breaking Changes

A new metric family is added. No existing series, label, or value changes.

## AI-assisted contribution

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3222,6 +3222,13 @@ if(BUILD_TESTING AND EXISTS "${_ANTHROPIC_ERROR_TEST_SRC}")
     add_cpp_ci_test(AnthropicErrorTest CI ON COMMAND test_anthropic_error)
 endif()
 
+set(_METRICS_ALIASES_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_metrics_aliases.cpp")
+if(BUILD_TESTING AND EXISTS "${_METRICS_ALIASES_TEST_SRC}")
+    add_executable(test_metrics_aliases test/cpp/test_metrics_aliases.cpp)
+    target_link_libraries(test_metrics_aliases PRIVATE lemonade-server-core)
+    add_cpp_ci_test(MetricsAliasesTest CI ON COMMAND test_metrics_aliases)
+endif()
+
 set(_ANTHROPIC_RELAY_HEADERS_TEST_SRC "${CMAKE_CURRENT_SOURCE_DIR}/test/cpp/test_anthropic_relay_headers.cpp")
 if(BUILD_TESTING AND EXISTS "${_ANTHROPIC_RELAY_HEADERS_TEST_SRC}")
     add_executable(test_anthropic_relay_headers test/cpp/test_anthropic_relay_headers.cpp)
--- a/docs/guide/configuration/custom-models.md
+++ b/docs/guide/configuration/custom-models.md
@@ -347,6 +347,18 @@ When an alias is specified in any API request (e.g., `/v1/chat/completions`, `/v
 
 Every registered alias is exposed as an independent model entry in `/v1/models`, `/v1/models/{id}`, and `lemonade list`. Alias entries set `id` to the alias name while sharing the target model's recipe, downloaded status, and backend configuration.
 
+#### Alias Mapping in `/metrics`
+
+`/metrics` emits one `lemonade_model_alias_info{alias,model_name}` sample per registered alias, present whether or not the target model has been loaded. Alias chains collapse to the final target, and unresolvable or cyclic aliases are omitted so `/metrics` never advertises a mapping the inference APIs would refuse to follow.
+
+Model series stay labeled with the canonical `model_name` only — they are not duplicated under the alias — so cumulative counters remain single-counted. Join the mapping onto them to query by alias:
+
+```promql
+lemonade_model_loaded * on(model_name) group_right lemonade_model_alias_info
+```
+
+The `lemonade_model_*` series are built from the Router's snapshot of models it has observed, so a model that has not been loaded in the current process has no series for the mapping to join onto. The `lemonade_model_alias_info` sample is present either way, so an alias that resolves but joins to nothing indicates a target that is not currently tracked, while an absent sample indicates the alias is not registered.
+
 ### Five reference cases
 
 | Sources                                         | `/v1/models` ids                                      | Resolution                                                                 |
--- a/src/cpp/include/lemon/prometheus_metrics.h
+++ b/src/cpp/include/lemon/prometheus_metrics.h
@@ -1,6 +1,9 @@
 #pragma once
 
+#include <functional>
+#include <map>
 #include <string>
+#include <unordered_map>
 
 #include "router.h"
 
@@ -13,6 +16,14 @@ struct SystemMetrics {
     double npu_percent = -1.0;
 };
 
-std::string build_prometheus_metrics(Router& router, const SystemMetrics& system_metrics);
+using MetricsAliasMap = std::map<std::string, std::string>;
+
+// `resolve` returns "" when the alias does not resolve; those entries are dropped.
+MetricsAliasMap build_metrics_alias_map(
+    const std::unordered_map<std::string, std::string>& raw_aliases,
+    const std::function<std::string(const std::string&)>& resolve);
+
+std::string build_prometheus_metrics(Router& router, const SystemMetrics& system_metrics,
+                                     const MetricsAliasMap& aliases = {});
 
 } // namespace lemon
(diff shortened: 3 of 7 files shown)


boecks:
Sure it could be useful. I was just not sure if it is actually desired. If you want, i could make the change in the docs and force push on this branch?

boecks:
I added it to docs/guide/configuration/custom-models.md under the alias section: the lemonade_model_alias_info{alias,model_name} series, chain collapsing, why model series aren't duplicated under the alias, and the PromQL join. Plus a note that lemonade_model_* comes from the Router's snapshot of observed models, so an alias to a never-loaded target has nothing to join onto while its own sample is still present.

While writing it up I noticed the family was named lemonade_model_alias, without the _info suffix that lemonade_server_info and lemonade_model_info already use for the same constant-1 join-metric pattern. Renamed it to lemonade_model_alias_info in the same force-push — free to change now, breaking once released. Happy to revert if you'd rather keep the shorter name, or for any other change you might think it needs so this PR can get merged.

jeremyfowers:
This seems like more of a fix than a feature to me, so I'm saying RFC is not required.

@abn would you like to review this alias-related PR? Please ensure the ~200 LoC of new testing adds value.

PR #3437, opened 2026-08-30
fix(mcp): resolve model aliases in tool calls

### LINKED ISSUE #3434: POST /mcp does not resolve model aliases
### Platform

Linux/Other (please specify)

### Lemonade Version

11.8.1

### GPU / APU Model

AMD R9700 AI Pro

### Component

lemond (server)

### Bug Description

The MCP gateway does not resolve model aliases. When I pass a registered alias as
the `model` argument of an MCP tool call, the call fails with "Model not found",
while the same alias on `/v1/chat/completions` succeeds against the same running
server.

The alias documentation does not qualify which surfaces resolve aliases:

> When an alias is specified in any API request (e.g., `/v1/chat/completions`,
> `/v1/embeddings`, `/v1/load`) or CLI command, Lemonade resolves the alias to its
> underlying target model ID.
> — `docs/guide/configuration/custom-models.md:341`

### Steps to Reproduce

1. Start `lemond` and pull `Tiny-Test-Model-GGUF`.

2. Register an alias:
   ```bash
   curl -X POST http://localhost:13305/internal/aliases \
     -d '{"alias":"ALIAS","target":"Tiny-Test-Model-GGUF"}'
   ```
   ```json
   {"alias":"ALIAS","status":"ok","target":"Tiny-Test-Model-GGUF"}
   ```

3. Call `lemonade_chat` with the alias:
   ```bash
   curl -X POST http://localhost:13305/mcp \
     -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{
           "name":"lemonade_chat",
           "arguments":{"model":"ALIAS","messages":[{"role":"user","content":"hi"}],"max_tokens":8}}}'
   ```

4. Send the same alias to `/api/v1/chat/completions` — returns 200.

Expected: the alias resolves to `Tiny-Test-Model-GGUF`, as it does on
`/v1/chat/completions`.

Actual, from step 3:

```json
{"id":1,"jsonrpc":"2.0","result":{"content":[{"text":"Error: Model not found: ALIAS","type":"text"}],"isError":true}}
```

Step 4 with the same alias on the same server returns HTTP 200.

### Log Output

```shell
2026-08-30 09:53:15.325 [Info] (AliasManager) Set alias: ALIAS -> Tiny-Test-Model-GGUF
2026-08-30 09:53:20.451 [Info] (McpServer) tools/call: lemonade_chat
2026-08-30 09:53:20.451 [Info] (Server) Auto-loading model: ALIAS
2026-08-30 09:53:20.451 [Error] (McpServer) Tool 'lemonade_chat' failed: Model not found: ALIAS
```

### Additional Context

Code pointers on 11.8.1:

- `McpServer::resolve_model_for_tool` (`src/cpp/server/mcp_server.cpp:443`), step 1,
  returns an explicit `model` argument verbatim without consulting `AliasManager`.
- `McpServer::tool_omni` (`src/cpp/server/mcp_server.cpp:819`) reimplements the same
  precedence inline for collections and has the same gap.
- `McpServer` is constructed at `src/cpp/server/server.cpp:1518` with `Router*`,
  `ModelManager*` and a load callback, and holds no reference to `AliasManager`,
  so resolution is not reachable from inside the class.

I found this while investigating #3433 and #3413, and I think all three share one
root cause: alias resolution lives in `Server::normalize_and_resolve_request_model`,
a private `Server` method that only the OpenAI-surface handlers call. `/mcp` is the
third surface that cannot reach it. In the same repro, `POST /v1/messages` and the
Ollama `POST /api/chat` also return 404 on the alias, and `/metrics` carries no
series relating the alias to its target.

I have a fix and am happy to open a PR.

boecks:
## Summary

Passing a registered alias as the `model` argument of an MCP tool call failed
with "Model not found", while the same alias succeeded on
`/v1/chat/completions`. `McpServer::resolve_model_for_tool` returned an explicit
`model` argument verbatim, `tool_omni` reimplemented that precedence inline with
the same gap, and `McpServer` held no reference to `AliasManager`. `McpServer`
now takes an `AliasResolver` callback bound to `Server::resolve_alias_target`,
applied through one helper that both call sites share.

Fixes #3434

## Scope

The resolver parameter is required rather than defaulted, matching the sibling
`EnsureLoadedFn`, so a future construction site cannot silently omit it and
reintroduce this bug without a compile error. Both call sites go through one
private helper rather than repeating the resolution inline, since `tool_omni`
duplicating the precedence rule is why the gap existed in two places.

## Testing

_Testing details:_

- `ctest -L cpp-ci`: 67/67 pass.
- `python test/server_mcp.py`: 19/19 pass, including the new
  `test_031_chat_tool_resolves_alias`, which registers an alias and calls
  `lemonade_chat` with it. It fails before this change.

Containers built from `main` and from this branch, alias
`ALIAS -> Tiny-Test-Model-GGUF`, calling `lemonade_chat`:

- main: `{"result":{"content":[{"text":"Error: Model not found: ALIAS"}],"isError":true}}`
- this branch: `isError:false`

Server log on main, showing the alias reaching the load path unresolved:

```
(McpServer) tools/call: lemonade_chat
(Server) Auto-loading model: ALIAS
(McpServer) Tool 'lemonade_chat' failed: Model not found: ALIAS
```

Also run against a live server with a 35B llamacpp/ROCm model behind an alias:
`lemonade_chat` succeeds through the alias and through a chained alias,
resolving to the final target.

`tool_omni` is patched for the same gap but is not covered by a regression test,
since it needs an Omni collection model that is not available in the test
environment.

## Documentation

`docs/guide/configuration/custom-models.md:341` already states that an alias in
"any API request" resolves to its target. This change makes the implementation
match.

## Breaking Changes

Tool calls that previously failed now succeed. A tool call naming a real model id
is unaffected, since a non-alias name resolves to itself.

## AI-assisted contribution

### DIFF
--- a/src/cpp/include/lemon/mcp_server.h
+++ b/src/cpp/include/lemon/mcp_server.h
@@ -22,8 +22,10 @@ using json = nlohmann::json;
 class McpServer : public std::enable_shared_from_this<McpServer> {
 public:
     using EnsureLoadedFn = std::function<void(const std::string&)>;
+    using AliasResolver = std::function<std::string(const std::string&)>;
 
-    McpServer(Router* router, ModelManager* model_manager, EnsureLoadedFn ensure_loaded);
+    McpServer(Router* router, ModelManager* model_manager, EnsureLoadedFn ensure_loaded,
+              AliasResolver resolve_alias);
     ~McpServer();
 
     // Must be called on a shared_ptr instance — handlers capture shared_from_this().
@@ -46,7 +48,7 @@ private:
     json tool_list_models(const json& arguments);
 
     // Resolve which model a tool should use when `model` is omitted. Precedence:
-    //   1. an explicit `model` argument always wins;
+    //   1. an explicit `model` argument always wins, resolved through aliases;
     //   2. a model of the right type that's already LOADED (zero cost);
     //   3. a model of the right type that's already DOWNLOADED (no network);
     //   4. the tool's hard-coded default, but only when the caller passed
@@ -60,6 +62,8 @@ private:
         const char* type_str,
         const char* default_model,
         bool allow_download);
+
+    std::string resolve_requested_model(const std::string& model) const;
 
     static json text_content_block(const std::string& text);
     static json make_error_response(const json& id, int code, const std::string& message);
@@ -76,6 +80,7 @@ private:
     Router* router_;
     ModelManager* model_manager_;
     EnsureLoadedFn ensure_loaded_;
+    AliasResolver resolve_alias_;
 };
 
 }  // namespace lemon
--- a/src/cpp/server/mcp_server.cpp
+++ b/src/cpp/server/mcp_server.cpp
@@ -215,10 +215,12 @@ std::string unique_token() {
 
 }  // namespace
 
-McpServer::McpServer(Router* router, ModelManager* model_manager, EnsureLoadedFn ensure_loaded)
+McpServer::McpServer(Router* router, ModelManager* model_manager, EnsureLoadedFn ensure_loaded,
+                     AliasResolver resolve_alias)
     : router_(router),
       model_manager_(model_manager),
-      ensure_loaded_(std::move(ensure_loaded)) {}
+      ensure_loaded_(std::move(ensure_loaded)),
+      resolve_alias_(std::move(resolve_alias)) {}
 
 McpServer::~McpServer() = default;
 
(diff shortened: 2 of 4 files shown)


jeremyfowers:
@abn : correct usage of the aliases feature?

@fl0rianr : correct integration with MCP?

fl0rianr (approved):
Looks fine for me

PR #3442, opened 2026-08-31
Prompt Validator: accept real request body

### LINKED ISSUE #3332: routing/validate: accept a real request body so multi-turn policies can be previewed (and document the endpoint)
## Summary

`POST /v1/routing/validate` takes a **flattened, one-field-per-signal** request shape rather than a real request body:

```json
{ "policy": {...}, "prompt": "…", "has_images": false, "has_tools": false, "metadata": {...} }
```

and `Server::handle_routing_validate` hand-assembles its `RouteContext` field by field instead of deriving it from a request the way production does via `build_route_context`.

Two consequences, one structural and one immediate.

## 1. Every new routing signal has to be re-plumbed by hand, or it silently reads as absent

Because the context is hand-built, **any field added to `RouteContext::Params` defaults to `0`/`false` in the policy-preview path unless the author remembers to add a line.** The failure mode is a silent wrong answer, not an error: a rule evaluates `false` in the Router Builder while the identical request matches in production.

This is not hypothetical — it has already diverged across the current batch of routing PRs:

| PR | New `Params` field | Plumbed into validate? |
|---|---|---|
| #3181 | `total_chars` | ✅ yes (the `+5/-0` hunk in `server.cpp`) |
| #3100 | cost lookups | ✅ yes (`make_router_cost_services` wired in) |
| #3169 | `turn_count` | ❌ **no** |
| #3163 | `expected_output_tokens` | ❌ no (currently harmless — nothing reads it yet) |

For #3169 the practical effect is that the policy from its own new documentation:

```json
{"all": [{"min_turns": 4}, {"metadata": {"key": "tool_error_streak", "gte": 2}}]}
```

reports `min_turns` **false** in the preview endpoint and falls through to `default_model`, while the same request through `/v1/chat/completions` matches. The `metadata gte` half works, which makes it harder to diagnose rather than easier. That is being fixed in #3169 itself by adding a `turns` input; this issue is about the pattern that produced it.

## 2. Multi-turn and multi-message policies cannot be previewed at all

The endpoint accepts only a single `prompt` string, so it can represent exactly one turn. It therefore **under-reports rather than misreports** — but the whole point of `min_turns`/`max_turns` and `min_total_chars`/`max_total_chars` is conversation shape, and none of it is reachable here. A `{"min_total_chars": 100}` policy tested with a 50-byte prompt reports "default", while in production a 3-turn conversation totalling 120 bytes with a 50-byte last turn matches the rule.

## Proposal

Let the endpoint accept a real request body as an alternative input — e.g. an optional `messages` array (and `input`/`prompt` for the other two request forms) — and derive the context through `build_route_context`, keeping the existing flattened shape working for back-compat.

That does two things at once: it makes multi-turn policies previewable, and it makes the preview path derive its signals from the *same* function production uses, so the drift class disappears permanently instead of being re-fixed per field.

**Note on a tempting shortcut:** simply calling `build_route_context` on a synthesized body while keeping the current input shape is *worse*, not better — `has_tools` and `has_images` arrive as booleans, so you would have to fabricate a dummy `tools[]` array and a fake image content part purely to have the helper re-derive the booleans you started with. The value is in accepting a real body, not in routing the existing flattened shape through a different function.

## Documentation

Per the project guideline that all endpoints are documented, `routing/validate` should be documented as part of this work — it currently is not.

I swept all 38 endpoints registered via `register_post`/`register_get`/`register_delete` in `server.cpp` against `docs/`. Three have no documentation hits:

- `install/dry-run` — already being addressed by #3324
- `log-level` — being retired by #3322
- **`routing/validate` — the only genuine open gap**

So this is the last one outstanding. Registration itself is fine: `register_post` correctly registers all four prefixes (`/api/v0/`, `/api/v1/`, `/v0/`, `/v1/`) plus 405s on GET, satisfying the quad-prefix invariant.

## Scope

Deliberately **not** folded into #3169 — that PR should land the minimal `turns` input matching the endpoint's existing per-signal design. Accepting a real request body is an API feature (new expressiveness, docs, tests), not a refactor, and shouldn't gate two condition ops from an external contributor. If this lands later it subsumes the per-field additions (`turns`, `total_chars`, and the two booleans).

## Context

Found while reviewing the routing PR batch (#3100, #3163, #3169, #3181).

QuMuzafferEge:
## Summary

POST /v1/routing/validate endpoint now accepts the real message body as an alternative input. Also updated the documentation about the endpoint.

Fixes #3332

## Scope

- [X] This PR addresses one clear issue or change.
- [X] I reviewed the full diff myself before submitting.
- [X] I removed unrelated local changes.
- [X] I kept refactoring separate unless it is required for this change.

## Testing

- [X] Code builds without errors locally.
- [X] I tested this change locally.

_Testing details:_ 14 tests: the pre-existing back-compat tests plus 4 new real-body tests run against a live server. All tests pass.

## Documentation

- [X] Documentation is affected and has been updated.

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:


ramkrishna2910:
This PR has a number of issues:
1. `/v1/routing/validate` is already documented on main and this PR adds a duplicate index row .
2. "```" bash fence nested inside the existing example code block breaks markdown rendering.
3.  `Dockerfile.momentum-tests` and `Dockerfile.momentum-tests.dockerignore` are unrelated and needs to be removed.
4. Missing type validation on the new inputs.
5. The flattened path still hand-builds `RouteContext`, so a new `Params` field still reads silently absent there and the drift isnt solved.

QuMuzafferEge:
> 5\. The flattened path still hand-builds `RouteContext`, so a new `Params` field still reads silently absent there and the drift isnt solved

I pushed a new commit for these.

PR #3444, opened 2026-08-31
Minor follow-up cleanup to #3330

SlawomirNowaczyk:
## Summary

Two minor, non-functional follow-up cleanups to #3330 (`fix(routing): llm classifier no longer treats has_tools as risk`), raised as non-blocking nits during review:
- The inline comment in `build_context_payload()` still said the fields were withheld "unless the policy actually has a use for these fields," implying a general policy-driven decision. They're actually exposed only when `expose_request_features_` is set — currently just the `routing.router` sugar's synthesized classifier. Reworded to match.
- Renamed `test_feature_serialization_opt_in` → `test_author_classifier_never_serializes_request_features`. Since #3330 removed the sibling-rule opt-in path, normal classifiers never receive the fields, so the old "opt_in" name was misleading about what the test asserts.

No behavior change — comment and test-name only.

### DIFF
--- a/src/cpp/server/routing_policy.cpp
+++ b/src/cpp/server/routing_policy.cpp
@@ -355,7 +355,9 @@ private:
             {"metadata", std::move(metadata)},
         };
         // Withheld entirely (not just disclaimed) unless expose_request_features_
-        // is set — currently only the routing.router sugar's synthesized classifier.
+        // is set. When constructed through the policy parser this is only the
+        // routing.router sugar's synthesized classifier; the make_classifier(s)
+        // helpers default it to true for direct callers.
         if (expose_request_features_) {
             payload["has_tools"] = request.params.has_tools;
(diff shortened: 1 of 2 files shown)


fl0rianr (approved):
Looks good overall.

One tiny non-blocking nit:
The new wording is slightly more absolute than the API actually guarantees. `routing.router` is currently the only parser path that sets `expose_request_features = true`, but the public `make_classifier()` / `make_classifiers()` helpers still default that argument to `true` for direct callers/tests/tooling.

So currently only the routing.router sugar and `test_author_classifier_never_serializes_request_features` are not universally true outside the parser path. Maybe qualify the comment with "when constructed through the policy parser" and rename the test to something like `test_parsed_author_classifier_never_serializes_request_features`.

Not blocking from my side. Approve.

Bekhouche:
LGTM

Bekhouche (approved):

PR #3445, opened 2026-08-31
feat(app): router collection panel and add-model flow

### LINKED ISSUE #2654: [Router] Add “New Model” Capability Within Hybrid Router Candidate Models
### Feature Description

The Candidate Models dropdown within the Hybrid Router currently displays only the existing Lemonade models. To improve usability, we should add a "New Model" button directly within the Candidate Models section.

This button should reuse the same logic as the "New Model" option available from the main menu, enabling users to:
- Create and upload a new model manually.
- Import a model configuration from JSON.
- Add the newly created model directly to the router's candidate model list.

### Use Case / Motivation

Currently, if a user wants to add a model that is not already available, they must leave the Router Builder, navigate back to the landing page, select "New Model" from the main menu, and then return to the router configuration. This interrupts the workflow.

Embedding the model creation capability directly within the Router Builder would provide a more seamless user experience and allow users to manage candidate models.

### Platform Relevance

All platforms

### Additional Context

_No response_

ao-batouche:
## Summary

Adds a "+ New Model" button to the Candidate Models section of the Router Builder,
opening the same two entry points as the File menu — a manual form (`AddModelPanel`)
or a JSON import. The newly registered model is selected as a router candidate
automatically once the server reports it, so adding a candidate no longer means
leaving the Router Builder and losing the in-progress router.

The panel renders its own `AddModelPanel` portal instead of dispatching `openAddModel`,
because `ChatWindow` (which owns that listener) unmounts when the chat pane is hidden.
ChatWindow's inline install logic moved to `utils/addModel.ts` so both entry points
register models through one code path.

Fixes #2654

## Scope

- [X] This PR addresses one clear issue or change.
- [X] I reviewed the full diff myself before submitting.
- [X] I removed unrelated local changes.
- [X] I kept refactoring separate unless it is required for this change.

## Testing

_Testing details:_

- Router Builder > Candidate Models > **+ New Model** > **Manually**: register a
  llama.cpp model; it is checked as a candidate once the pull completes.
- **+ New Model** > **From JSON**: import a model JSON exported from the Model
  Manager; same result.
- Register a non-chat model (e.g. an embedding model) from the button: a warning
  toast explains it cannot route chat requests and it is not added as a candidate.
- Empty state (no compatible models registered): the button is reachable and the
  message points at it.
- Existing File > New Model paths (Manually and From JSON) still register models
  unchanged.

## Documentation

- [X] Documentation is not affected by this change.

## Breaking Changes

## AI-assisted contribution

Please select one:

If AI tools were used:

### DIFF
--- a/src/app/src/renderer/ChatWindow.tsx
+++ b/src/app/src/renderer/ChatWindow.tsx
@@ -4,6 +4,7 @@ import {
   AppSettings,
   mergeWithDefaultSettings,
 } from './utils/appSettings';
+import { installModelFromForm, installModelFromJSON, readModelJSONFile } from './utils/addModel';
 import { serverFetch } from './utils/serverConfig';
 import { useModels } from './hooks/useModels';
 import { useInferenceState } from './hooks/useInferenceState';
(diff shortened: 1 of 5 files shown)


jeremyfowers (changes requested):
@ramkrishna2910 help me understand why we have a PR into GUI2 on a milestone? GUI2 should be frozen.

Unless there is a high justification IMO this should be converted into a GUI RFC to be evaluated after GUI3 is live.

